PR #230, opened 2020-07-23
Use map when Import onnx node

chentong319:
Previously, a long list of 'IF' statements are used to check which builder function to use when a onnx node is imported. 
The new implementation uses a map between the node name and the builder function. Exploit the lookup in STL map and hopefully could be optimized.

Changes include:

1. generate the map initialization by OpBuildTable.inc in gen_pnnx_mlir.py

2. set up and use the map in FrontendGenImpl class.

3. Give a name to the previous anonymous namespace, to follow the style of tensor flow. Make it possible to change OpBuildTable.inc into a .cpp file.


caoimhinuibrian:
Shouldn't you commit the output .inc files too? Otherwise the build results are not based on the changed tool.

chentong319:
I checked in the OpBuildTable.inc, but not ONNXOps.td.inc because it is not changed by this patch.

caoimhinuibrian:
LGTM

PR #236, opened 2020-07-25
include stdlib.h when compiling on macOS

mshr-h:
On macOS, `<stdlib.h>` contains malloc. So `<stdlib.h>` should be included instead of `<malloc.h>`.
In this PR, I added `#ifdef` to switch include header.


tjingrant:

PR #238, opened 2020-07-28
Cleanup rtmemref api

gongsu832:
-  use forward declaration to hide private data fields
    -  RtMemRef.h: external user header, C/C++
    - _RtMemRef.h: internal user header, C++ only
    -  RtMemRef.hpp and RtMemRef.cpp: implementation header and file
    -  add external APIs
           OrderedRtMemRefDict *ormrd_create(RtMemRef **rmrs, int n)
           RtMemRef **ormrd_getRmrs(OrderedRtMemRefDict *ormrd)
           int ormrd_getNumOfRmrs(OrderedRtMemRefDict *ormrd)
       for creating and querying OrderedRtMemRefDict with RtMemRef arrays
    -  data buffer installed by rmr_setData() will be managed by user
    -  unique_ptr<RtMemRef> must use custom deleter <RtMemRef,decltype(&rmr_destroy)>


src/Runtime/DataType.h
```diff
@@ -4,37 +4,82 @@
...
+    {"h", ONNX_TYPE_UINT8},  // uint8_t  -> UINT8,  unsigned char  -> UNIT 8
+    {"s", ONNX_TYPE_INT16},  // int16_t  -> INT16,  short          -> INT16
+    {"t", ONNX_TYPE_UINT16}, // uint16_t -> UINT16, unsigned short -> UINT16
+    {"i", ONNX_TYPE_INT32},  // int32_t  -> INT32,  int            -> INT32
+    {"j", ONNX_TYPE_UINT32}, // uint32_t -> UINT32, unsigned int   -> UINT32
+    {"l", ONNX_TYPE_INT64},  // int64_t  -> INT64,  long           -> INT64
+    {"m", ONNX_TYPE_UINT64}, // uint64_t -> UINT64, unsigned long  -> UINT64
+    {"f", ONNX_TYPE_FLOAT},  // float    -> FLOAT
+    {"d", ONNX_TYPE_DOUBLE}, // double   -> DOUBLE
+};
```
AlexandreEichenberger, line 85: There is a method where you can build a data-base of related info. It is then much easier to update / add operations. It goes something like this...

say in `header.h` you first define your database under a given macro. When 
```
# if defined OMPT_CALLBACK
/* CALLBACK(ompt level at which it is supported, callback name, how it is supported) */
/* Mandatory Callbacks */
OMPT_CALLBACK(1, thread_begin            , always)
OMPT_CALLBACK(1, thread_end              , always)
OMPT_CALLBACK(1, parallel_begin          , always)
OMPT_CALLBACK(1, parallel_end            , always)
OMPT_CALLBACK(1, task_create             , always)

# else

// normal header file

# endif
```

and then you can use this macro as follow, in the same header file or somewhere else

```
const int ompt_callback_num = 0
  #define OMPT_CALLBACK(level, name, support) + (level<=OMPT_TIER ? 1 : 0)
  #include "header.h"
  #undef OMPT_CALLBACK
  ;
```

which you can call elsewhere with a different macro, to extract different parts of the database.

In the same `header.h` you can build multiple database by adding before the default `# else` other `# elif defined OMPT_FCT`

tjingrant:
@AlexandreEichenberger thanks! working on it.

tjingrant:
@gongsu832 my updates:
- Rename OrderedRtMemRefDict -> RtMemRefList for simplicity. One can still sort of use RtMemRefList like a dictionary (keyed by rmr names) since we can iterate over the list of rmrs and find the first match instead of relying on an actual dictionary.
- Change API names to conform to the naming styles used by the rest of the project (LLVM style).
- Clarify memory management responsibilities and create constructors that allow user to specify data ownership.
- Apply alex's comments to build a database of data type metadatas.

I'm still working on the docs, but these are all my code changes; I hope to hear your opinions.

src/Conversion/KrnlToLLVM/KrnlToLLVM.cpp
```diff
@@ -538,27 +546,41 @@ class KrnlEntryPointOpLowering : public OpRewritePattern<KrnlEntryPointOp> {
...
+                                outRmrPtrsArr, ArrayRef<Value>({idxVal}))
+                            .getResult();
+
+      rewriter.create<LLVM::StoreOp>(loc, outRtMemRef, rmrPtrAddr);
     }
+
+    // Create wrapped output.
+    auto wrappedOutput = callApi(rewriter, loc, apiRegistry,
+        API::CREATE_ORDERED_RTMEMREF_DICT, {outRmrPtrsArr, numOutput});
+
```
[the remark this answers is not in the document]
tjingrant: > both memRef and outRtMemRef have _owningData=true

This is not exactly the case, RtMemRefs are user-facing wrappers for MemRef and regular MemRefs are compiler internal constructs that have no concept of data ownership; part of the reason is that memory management can get complicated within the compiler and data ownership may be too simplified a model to determine malloc/dealloc responsibility. The good thing is that the memory management for regular MemRef is done by the compiler, presumably, it is or will be done in ways that are as optimal as it can be according to static analysis, so we don't really need to worry about managing MemRef's memory (other than making sure that it is not released by the time we need to transfer data out of it) and by extension of this reason, we needn't worry about its data ownership.

In other words, the concept of ownership disappears as we enter the inference function and re-emerges as we exit the function. It only needs to exist where we cannot manage memories automatically.

src/Runtime/PyExecutionSession.cpp
```diff
@@ -19,74 +19,72 @@ namespace onnx_mlir {
...
       dtype = py::dtype("uint64");
     else {
-      fprintf(stderr, "Unsupported ONNX type in RtMemRef.onnx_dtype.");
+      fprintf(stderr, "Unsupported ONNX type in RtMemRef.");
       exit(1);
     }
 
-    outputPyArrays.emplace_back(py::array(dtype, shape, dynMemRef->data));
+    outputPyArrays.emplace_back(py::array(dtype, shape, rmrGetData(rmr)));
   }
```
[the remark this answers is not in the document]
tjingrant: It's a bit of a technical reason if I recall correctly. Otherwise, we need to create another struct where the pointer `void* data` is a const pointer, etc...

src/Conversion/KrnlToLLVM/KrnlToLLVM.cpp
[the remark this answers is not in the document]
gongsu832: OK as long as we know what happens with `memRef` (i.e., compiler should manage it), that's fine.

src/Runtime/PyExecutionSession.cpp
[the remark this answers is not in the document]
gongsu832: So this makes destroying `wrappedInput` a bit tricky. When data is writable, `rmrListDestroy(wrappedInput)` should free the rmr structs and the rmrList struct **without** freeing data. Since the data belongs to the caller of `pyRun`. When data is not writable, data should be freed since we malloc-ed and copied it.

So I think when data is writable, we should create `inputRtMemRef` with `rmrCreateWithNameAndOwnership()` and set ownership to false. This means that `inputRtMemRef` does not own the data. If we do that, `rmrListDestroy(wrappedInput)` should do the right thing for both cases.

src/Runtime/include/OnnxMlir.h
```diff
@@ -0,0 +1,250 @@
...
+RtMemRefList *rmrListCreate(RtMemRef **rmrs, int n);
+
+/**
+ * RtMemRefList destroyer
+ *
+ * @param ormrd, pointer to the RtMemRefList to be destroyed
+ *
+ * Destroy the RtMemRefList struct.
+ */
+void rmrListDestroy(RtMemRefList *ormrd);
```
gongsu832, line 228: Change `ormrd` (also in comment) to `rmrl`? Same for `rmrListGetPtrToRmrs()` and `rmrListGetNumRmrs()`, as well as internal functions in `OnnxMlirInternal.h`.

src/Runtime/PyExecutionSession.cpp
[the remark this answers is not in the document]
tjingrant: That sounds reasonable! Will update the code as you suggested.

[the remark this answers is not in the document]
gongsu832: Thanks. That's all I got for now. JNI code needs minor update but I will do that with a separate patch once this gets merged.

src/Runtime/include/OnnxMlir.h
```diff
@@ -0,0 +1,289 @@
...
+ *
+ * @param rmr, pointer to the RtMemRef
+ * @return the number of elements in the data buffer.
+ */
+INDEX_TYPE rmrGetNumElems(RtMemRef *rmr);
+
+/*---------------------------------------- */
+/* C/C++ API for RtMemRefList calls */
+/*---------------------------------------- */
+
```
AlexandreEichenberger, line 246: please add a description of what the list is, not necessarily the fields but what it is intended for. This is where I would indicate that in this list, RtMemRefs can be accessed by index or names. Is it supposed to be consistent? If one name is provided, all must have names? 

Also define that list appears to be static in size.

src/Runtime/include/OnnxMlir.h
AlexandreEichenberger: What does the destroy do? Does it also destroy RMR that have ownership? I.e. is it a deep or shallow destructor. Needs to be explicit.
I would rename the parameter to "list"... the type say what type of list it is. Same for calls below.

src/Runtime/include/OnnxMlir.h
```diff
@@ -0,0 +1,289 @@
...
+ *
+ * @param rmrs, array of pointers to RtMemRef
+ * @param n, number of elements in rmrs array
+ * @return pointer to the RtMemRefList created, NULL if creation failed.
+ *
+ * Create an RtMemRefList with specified RtMemRef array.
+ * If a RtMemRef has a name, in addition to be accessed by its index,
+ * the RtMemRef can also be accessed by its name.
+ */
+RtMemRefList *rmrListCreate(RtMemRef **rmrs, int n);
```
AlexandreEichenberger, line 258: When dealing with RtMemRefs, you omitted the MemRef from the function names, like "rmrCreate" "rmrGetXXX" which I think is fine.  I know that the prefix "rmr" is supposed to say RtMemRef, some folks may be parsing as a qualifier to prevent name conflict, not really something with a meaning.

I would continue to have the rmr being a prefix, and then just say rmrCreateList.

maybe size should be "unsigned int size" instead of "int n"

src/Runtime/include/OnnxMlir.h
```diff
@@ -0,0 +1,289 @@
...
+ * Destroy the RtMemRefList struct.
+ */
+void rmrListDestroy(RtMemRefList *ormrd);
+
+/**
+ * RtMemRefList RtMemRef array getter
+ *
+ * @param ormrd, pointer to the RtMemRefList
+ * @return pointer to the array of RtMemRef pointers.
+ */
```
AlexandreEichenberger, line 274: This is confusing... why return the list of pointers? I would have preferred to have a 

rmrGetFromList(RtMemRefList *list, uint index)
and rmrGetFromListWithName(RtMemRefList *list, char *name)

src/Runtime/include/OnnxMlir.h
```diff
@@ -0,0 +1,289 @@
...
+RtMemRef **rmrListGetPtrToRmrs(RtMemRefList *ormrd);
+
+/**
+ * RtMemRefList number of RtMemRefs getter
+ *
+ * @param ormrd, pointer to the RtMemRefList
+ * @return number of elements in the RtMemRef array.
+ */
+int rmrListGetNumRmrs(RtMemRefList *ormrd);
```
AlexandreEichenberger, line 283: if you look at std::list, it uses "size." I would use it in the constructor and here, i.e. "rmrGetListSize"

src/Runtime/include/OnnxMlir.h
```diff
@@ -0,0 +1,289 @@
...
+ *
+ * \section intro_sec Introduction
+ *
+ * ONNX-MLIR project comes with an executable `onnx-mlir` capable
+ * of compiling onnx models to a shared library. In this documentation, we
+ * explain and document ways to interact programmatically with the compiled
+ * shared library using the API's defined in `src/Runtime/include/OnnxMlir.h`.
+ *
+ * \section Important Data Structures
+ * \subsection RtMemRef
```
AlexandreEichenberger, line 31: in the list of functions operating on RtMemRef, I did not find one that accesses the data

rmrGetElement(rmr, indices)
rmrSetElement(rmr, indices, value)

test/numerical/TestConv.cpp
```diff
@@ -90,38 +90,39 @@ bool isOMConvTheSameAsNaiveImplFor(const int N, const int C, const int H,
   OwningModuleRef moduleRef(module);
 
   compileModule(moduleRef, ctx, SHARED_LIB_BASE, EmitLib);
-  onnx_mlir::ExecutionSession sess(
-      SHARED_LIB_BASE + ".so", "_dyn_entry_point_main_graph");
+  onnx_mlir::ExecutionSession sess(SHARED_LIB_BASE + ".so", "run_main_graph");
 
-  std::vector<unique_ptr<RtMemRef>> inputs;
-  auto xRmr = unique_ptr<RtMemRef>(getRndRealRmr<float>({N, C, H, W}));
+  std::vector<unique_ptr<RtMemRef, decltype(&rmrDestroy)>> inputs;
```
AlexandreEichenberger, line 95: Where is the C++ interface to RMR defined? I saw the C version; is there a "manual" for the C++ version too?

src/Runtime/include/OnnxMlir.h
```diff
@@ -0,0 +1,289 @@
...
+ *
+ * ```cpp
+ * void *_data;            // data buffer
+ * void *_alignedData;     // aligned data buffer that the rmr indexes.
+ * INDEX_TYPE _offset;     // offset of 1st element
+ * INDEX_TYPE *_dataSizes; // sizes array
+ * int64_t *_dataStrides;  // strides array
+ * int _dataType;          // ONNX data type
+ * int _rank;              // rank
+ * std::string _name;      // optional name for named access
```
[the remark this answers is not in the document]
gongsu832: Using char * for getter/setter is because the API needs to work for both C and C++. Using std::string for _name internally is because the RtMemRef destructor will automatically call string destructor to free _name (done by the compiler) when the RtMemRef struct is destroyed. If char * is used for _name, RtMemRef destructor needs to free _name explicitly to avoid memory leak.

src/Runtime/include/OnnxMlir.h
```diff
@@ -0,0 +1,289 @@
...
+ * bool _owningData;       // indicates whether the Rmr owns the memory space
+ *                            referenced by _data. Rmr struct will release the
+ * memory space refereced by _data upon destruction if and only if it owns it.
```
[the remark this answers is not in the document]
tjingrant: Is the user supposed to know about how to use this? No, and I've changed the doc so that we don't reveal things like this for users. Code has always been such that user has no access to internal details. Gong achieved this using forward declarations for API function call deaclarations.

src/Runtime/include/OnnxMlir.h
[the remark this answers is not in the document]
tjingrant: It feels like adding name-based query is too complicated for C runtime API (the implementation as well as specification of semantics are both way more complicated than we should be worrying about at this moment, I think, due to the lack of dictionary-like standard container in c), maybe I'll make that happen for C++ API next.

src/Runtime/include/OnnxMlir.h
```diff
@@ -0,0 +1,289 @@
...
+ */
+void rmrSetName(RtMemRef *rmr, char *name);
+
+/**
+ * RtMemRef number of elements getter
+ *
+ * @param rmr, pointer to the RtMemRef
+ * @return the number of elements in the data buffer.
+ */
+INDEX_TYPE rmrGetNumElems(RtMemRef *rmr);
```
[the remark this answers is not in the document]
tjingrant: I explicitly asserted that no skipping strides are specified when making this calculation.

src/Runtime/include/OnnxMlir.h
tjingrant: Thanks, new API looks like "omTensorListXXX".

src/Runtime/include/OnnxMlir.h
tjingrant: Now there're both options. More specifically, we use to have index-based mutator and accessor for private use. But I dropped the index-based mutator (because I can think of no legitimate use) and made the accessor public.

src/Runtime/include/OnnxMlir.h
tjingrant: done.

src/Runtime/include/OnnxMlir.h
tjingrant: Probably a thing to add next, if there's a use case for them.

test/numerical/TestConv.cpp
tjingrant: This patch concerns the C99 API (important in and of itself, and more significantly building block for JNI). Will add C++ runtime API in another patch shortly.

AlexandreEichenberger (approved):
Left mostly requests for clarification, for a user that does not want to look at the code to understand memory alloc policies, and alignment.

code and functionality: LGTM

include/onnx-mlir/Runtime/OMTensor.h
```diff
@@ -0,0 +1,387 @@
...
+
+/**
+ * OMTensor number of elements getter
+ *
+ * @param tensor, pointer to the OMTensor
+ * @return the number of elements in the data buffer.
+ */
+int64_t omTensorGetNumElems(OMTensor *tensor);
+
+#ifdef __cplusplus
```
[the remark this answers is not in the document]
gongsu832: Also, the `#ifdef __cplusplus` surrounded code are meant to be "internal" only, i.e., they are only called by ONNX MLIR itself, not meant for end users. So I don't think they should appear in user header OMTensor.h.

include/onnx-mlir/Runtime/OMTensor.h
```diff
@@ -0,0 +1,387 @@
...
+ * @param allocatedPtr allocated pointer to tensor content.
+ * @param alignedPtr aligned pointer to tensor content. If NULL will be set to
+ * allocatedPtr.
+ *
+ *
+ */
+void omTensorSetPtr(
+    OMTensor *tensor, int owning, void *allocatedPtr, void *alignedPtr);
+
+/**
```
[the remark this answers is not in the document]
gongsu832: Same comment as below for `#ifdef cplusplus`. Calls involving internal fields like `alignedPtr` really shouldn't appear in a header that will be included by an end user.

src/Runtime/OMTensorList.inc
```diff
@@ -0,0 +1,80 @@
...
+   * to address ReMemRefs by name.
+   */
+  OMTensor **_omts; // OMTensor array
+
+  size_t _size; // Number of elements in _omts.
+};
+
+/* OMTensorList creator */
+OMTensorList *omTensorListCreate(OMTensor **tensors, int n) {
+  OMTensorList *list = (OMTensorList *)malloc(sizeof(struct OMTensorList));
```
gongsu832, line 56: Need to check for `malloc` failure.

src/Runtime/OMTensor.inc
```diff
@@ -0,0 +1,434 @@
...
+               // and only if it owns it.
+};
+
+// Create a OMTensor.
+OMTensor *omTensorCreate(
+    void *data_ptr, int64_t *shape, int64_t rank, OM_DATA_TYPE dtype) {
+  OMTensor *tensor = (OMTensor *)malloc(sizeof(OMTensor));
+  if ((tensor->_shape = (int64_t *)malloc(rank * sizeof(int64_t))) &&
+      (tensor->_stride = (int64_t *)malloc(rank * sizeof(int64_t)))) {
+    tensor->_allocatedPtr = data_ptr;
```
gongsu832, line 99: Need to check each `malloc` for failure. And if one fails, the previous successful `malloc` must be freed.

src/Runtime/OMTensor.inc
```diff
@@ -0,0 +1,434 @@
...
+  OMTensor *tensor = omTensorCreate(data_ptr, shape, rank, dtype);
+  tensor->_owning = owning;
+  return tensor;
+}
+
+// Create a OMTensor.
+OMTensor *omTensorCreateEmptyDeprecated(int rank) {
+  OMTensor *omt = (OMTensor *)malloc(sizeof(struct OMTensor));
+  if ((omt->_shape = (int64_t *)malloc(rank * sizeof(int64_t))) &&
+      (omt->_stride = (int64_t *)malloc(rank * sizeof(int64_t)))) {
```
gongsu832, line 128: Assume this will be removed so no need to fix?

src/Runtime/OMTensor.inc
```diff
@@ -0,0 +1,434 @@
...
+    omt->_owning = false;
+  }
+  return omt;
+}
+
+OMTensor *omTensorCreateEmpty(
+    int64_t *shape, int64_t rank, OM_DATA_TYPE dtype) {
+  OMTensor *tensor =
+      omTensorCreateWithOwnership(NULL, shape, rank, dtype, true);
+  void *dataPtr = malloc(omTensorGetNumElems(tensor) * getDataTypeSize(dtype));
```
gongsu832, line 143: Need to check for `omTensorCreateWithOwnership` and `malloc` failure. And if `malloc` fails, tensor must be destroyed.

src/Runtime/OMTensor.inc
```diff
@@ -0,0 +1,434 @@
...
+      tensor->_stride[i] = tensor->_stride[i + 1] * tensor->_shape[i + 1];
+  }
+  return tensor;
+}
+
+// Create a OMTensor.
+OMTensor *omTensorCreateWithOwnership(void *data_ptr, int64_t *shape,
+    int64_t rank, OM_DATA_TYPE dtype, int owning) {
+  OMTensor *tensor = omTensorCreate(data_ptr, shape, rank, dtype);
+  tensor->_owning = owning;
```
gongsu832, line 120: Need to check for `omTensorCreate` failure.

tjingrant:
@gongsu832, my response to the high-level question of why I dropped the distinction between external/internal APIs: I think some internal api's are kept internal because communicating its semantics to users is hard and one can get oneself into an inconsistent state if it is used. However, the same rationale applies to stop onnx-mlir developers to use these troublemakers. So I removed some of them (like OMTensorList mutator) and made rest of them public.

Do you see other compelling reasons to have a internal/external separation? The difference between C/C++ corresponds loosely with the current designation of external/internal APIs and that's because we haven't introduced C++ APIs yet.

gongsu832:
@tjingrant I think we do have to make a distinction between end users and developers. Developers can be educated to use the internal APIs properly. In a sense, they actually have to. End users on the other hand should not be expected to know anything that are not meant for them.

Also purely from a good coding practice point of view, if something is not meant for the end user, then it should not appear in any user visible API header. So as much as I hate to split the header, I don't see a better way around this unless we advertise those internal APIs as end user callable APIs, which in my opinion is really not what we want to do.

tjingrant:
@gongsu832 how should we decide which API is for users to know and which not? I tried to tell the difference, but for every API function in this patch, I can always think of (maybe rare but still valid) reasons that users may still want to use it: just to give you some examples, 
- `omTensorCreateWithRandomData` is useful if users decide to benchmark their model/device with random input.
- `omTensorAreTwoOmtsClose` is useful if users decide to check whether the model is compiled correctly by comparing it with the reference value they obtained through PyTorch or Tensorflow.

Furthermore, I think we do not have to choose between:
- Keeping things simple and not to overwhelm users.
- Making available as many bells and whistles to user as we have, to save user time/effort in case they need it.

Because the former is documented through tutorials/how-tos, and the latter is documented through references, and I do keep them separate (tutorial is included as in OnnxMlirRuntime.h, which is translated to HTML pages through `doxygen`).

Taking a step back, I actually don't know what is the motivation for having an internal API at this point - now that those trouble-maker API functions are either removed or deprecated. If there is a good motivation, it must entail a simple rule that divides these APIs into two classes (user-facing and developer-facing). If this rule of separation is complicated, then it's actually worse off for users and developers alike, because then users/developers will have to find them from two, instead of one places.

Taking another step back, if what you mean by developer-facing API functions are helper functions like `computeStridesFromSizes` then I totally agree they should be moved out of the user-facing headers; in fact I think they shouldn't be in any header files at all and I was being sloppy when I put them in `OMTensor.h`.

gongsu832:
@tjingrant I think the rule is fairly simple. However we decide which functions are useful for end users and should be exposed, they stay in the user visible headers; otherwise, they don't. I think that's a fairly standard coding practice.

So for the examples you gave, they stay in user visible headers. But what about calls like

- `omTensorGetAlignedPtr`
- `omTensorSetPtr`

that deal with alignedPtr? I think we definitely don't want the users to have to worry about alignment, which is an entirely internal matter. So it makes no sense for them to stay in the user visible header. And yes helper functions like `computeStridesFromSizes` definitely don't belong in the user visible header.

Perhaps a slightly easier way is that, if we are not entirely sure which functions should be exposed to the end users, we can start by removing those that we know definitely don't belong in the user visible header.

tjingrant:
@gongsu832 , that makes sense, so it looks like the majority of the API functions will stay public. If this is the case, then we needn't declare the internal APIs in any header at all. At least for my use case (calling them from within LLVM), header files are not useful. Do you need these declarations in JNI?

gongsu832:
@tjingrant No. The JNI wrapper is essentially a "normal user".

include/onnx-mlir/Runtime/OMTensor.h
```diff
@@ -0,0 +1,387 @@
...
+ * @param dtype tensor element data type.
+ * @param owning whether OMTensor owns the data, if set to true, OMTensor will
+ * release the data_ptr upon destruction.
+ * @return pointer to OMTensor created, NULL if creation failed.
+ *
+ */
+OMTensor *omTensorCreateWithOwnership(void *data_ptr, int64_t *shape,
+    int64_t rank, OM_DATA_TYPE dtype, int owning);
+
+/**
```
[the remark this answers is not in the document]
tjingrant: Yes, it ought to be removed in favor of more disciplined ctors.

include/onnx-mlir/Runtime/OMTensor.h
```diff
@@ -0,0 +1,387 @@
...
+ * Destroy the OMTensor struct.
+ *
+ * @param tensor pointer to the OMTensor
+ *
+ */
+void omTensorDestroy(OMTensor *tensor);
+
+/**
```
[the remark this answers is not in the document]
tjingrant: That's a question I do not have answers to, and I presume we will only get answers when we actually get to deal with alignment issues. Current approach is to leave alignment out of the user API design and documentation, because the current assumptions are:
- If data pointer is passed in from an external context, we assume alignment is not an issue (or it has been taken care of, via functions like `_aligned_malloc`). When using `_aligned_malloc`, I don't think it is customary for users to have access to or keep track of both aligned pointers and allocated pointers, so in this sense we don't really have to distinguish between two pointers.
- If data pointer is created by us and we do not have access to things like _aligned_malloc, then we align the data pointer we malloced ourselves and deal with the alignment issue internally and don't involve users in the decision process.

In either case, there is no need for users to know about the potential alignment issue. BTW, the specific function you are referring to is removed now because it is only used by the wrapper code converting MemRef -> OMTensor at the end of inference execution, and I just realized, reading Gong's comment, that I don't need to expose it in the header.

include/onnx-mlir/Runtime/OMTensor.h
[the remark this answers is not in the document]
tjingrant: Thanks for the super-detailed explaination, let me instead include this explanation to the decalaration of `allocatedPtr` in `struct OMTensor`.

include/onnx-mlir/Runtime/OMTensor.h
```diff
@@ -0,0 +1,387 @@
...
+ * Set the ONNX data type of the data buffer elements.
+ */
+void omTensorSetDataType(OMTensor *tensor, OM_DATA_TYPE dataType);
+
+/* Helper function to get the ONNX data type size in bytes */
+static inline int getDataTypeSize(OM_DATA_TYPE dataType) {
+  return OM_DATA_TYPE_SIZE[dataType];
+}
+
+/**
```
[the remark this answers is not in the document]
tjingrant: It seems somewhat obvious to me that the logical size is the one that makes sense here since we are calculating the number of elements in the tensor, not buffer size. The only thing that could go wrong is when you have gaps in the array (in the form of non-trivial strides), and I explicitly ensure that strides contain no gap.

include/onnx-mlir/Runtime/OMTensor.h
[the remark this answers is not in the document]
tjingrant: Currently, there're only a subset of C++ APIs used by the numerical tests. As per our (Gong and I) discussion, since they provide value for potential use case, they can stay inside the OMTensor header. 

Let me try to clean things up a bit by moving these unpolished C++ functions to a separate file inside an internal directory for the moment. In the future, I will:
- Provide full API spec for C++ user operating on OMTensor.
- Document them to the same level of detailedness as C99 API.
- Try to move implementation out of the header (this may be difficult with template going on).

include/onnx-mlir/Runtime/OMTensorList.h
```diff
@@ -0,0 +1,59 @@
...
+ * OMTensorList creator
+ *
+ * @param tensors array of pointers to OMTensor
+ * @param n number of elements in tensors array
+ * @return pointer to the OMTensorList created, NULL if creation failed.
+ *
+ * Create an OMTensorList with specified OMTensor array.
+ * If a OMTensor has a name, in addition to be accessed by its index,
+ * the OMTensor can also be accessed by its name.
+ */
```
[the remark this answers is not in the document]
tjingrant: I recorded this information as requested, but I find little principles to follow when making decisions as such, do you have any suggestions on how to make these decisions (what to reuse v.s. what to copy)?

src/Runtime/OMTensor.inc
tjingrant: If we bail out entirely (i.e., assertion failure with no recovery), it seems like we can get away with letting the OS free up the memory we malloc'ed? Otherwise the code gets complicated since we have three malloc's in a row here.

tjingrant: nvm, saw your comments on the return value, we actually don't bail out.

src/Conversion/KrnlToLLVM/KrnlToLLVM.cpp
```diff
@@ -455,7 +475,7 @@ class KrnlEntryPointOpLowering : public OpRewritePattern<KrnlEntryPointOp> {
         op.getAttrOfType<SymbolRefAttr>(
               KrnlEntryPointOp::getEntryPointFuncAttrName())
             .getLeafReference();
-    auto dynEntryPointName = "_dyn_entry_point_" + staticEntryPointFuncName;
+    auto dynEntryPointName = "run_" + staticEntryPointFuncName;
```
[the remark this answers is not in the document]
tjingrant: Yes, it's in the introduction to the runtime API.

src/Runtime/OMTensor.inc
gongsu832: Yes assertions are for catching bugs, i.e., internal code logic error. Malloc failures are not bugs. And APIs typically don't end abruptly (unless it's an assertion failure due to a bug of course) since they should give the callers a chance to clean up.

BTW, the logic for checking the malloc failures should be fairly simple. It can be some thing like:
```
if (tensor == NULL) return NULL;
if ((tensor->_shape = (int64_t *)malloc(rank * sizeof(int64_t))) == NULL) {
  free(tensor);
  return NULL;
}
if ((tensor->_stride = (int64_t *)malloc(rank * sizeof(int64_t))) == NULL) {
  free(tensor->_shape);
  free(tensor);
  return NULL;
}
...
```

AlexandreEichenberger:
Will look at it. Is there anyway that the doxygen files be generated and not included in the GitHub? Like "make doc" would gen them? If its too much work, forget about it.

include/onnx-mlir/Runtime/OMTensorList.h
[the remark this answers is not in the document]
AlexandreEichenberger: either policies are fine, the selected behavior has to be clear. It is now as the behavior is explicit in the comments.

AlexandreEichenberger (approved):
Looks great. Just asked for you to consider not putting all the oxygen in the GitHub, since it appears to be constructed from existing docs. But if not practical, please ignore.

Issue #234, opened 2020-07-24
Windows CI pipeline is failing silently


NathanielMcVicar:
As far as I can tell, the Windows CI pipeline isn't checking for and returning errors from the script, causing builds that fail to appear as green. An example is here: https://dev.azure.com/onnx-pipelines/onnx/_build/results?buildId=3115&view=logs&j=4060f429-1919-5440-1fce-77a91697330b&t=91a0b9d9-34dd-5af7-633c-c7e5328fed2f

Notice that the builder fails with the following error, but shows success.
```
MainUtils.cpp
D:\a\1\llvm-project\llvm\include\llvm/IR/User.h(48,32): warning C4141: 'inline': used more than once [D:\a\1\onnx-mlir\build\src\MainUtils.vcxproj]
D:\a\1\onnx-mlir\src\MainUtils.cpp(231,54): error C2065: 'outputBaseName': undeclared identifier [D:\a\1\onnx-mlir\build\src\MainUtils.vcxproj]
D:\a\1\onnx-mlir\src\MainUtils.cpp(233,61): error C3536: 'constPackFileName': cannot be used before it is initialized [D:\a\1\onnx-mlir\build\src\MainUtils.vcxproj]
D:\a\1\onnx-mlir\src\MainUtils.cpp(233,61): error C2664: 'std::error_code llvm::sys::fs::rename(const llvm::Twine &,const llvm::Twine &)': cannot convert argument 2 from 'int' to 'const llvm::Twine &' [D:\a\1\onnx-mlir\build\src\MainUtils.vcxproj]
D:\a\1\onnx-mlir\src\MainUtils.cpp(233,61): message : Reason: cannot convert from 'int' to 'const llvm::Twine' [D:\a\1\onnx-mlir\build\src\MainUtils.vcxproj]
```

I believe this is due to the way cmd.exe scripts returns (or, more often doesn't return) error codes. I am currently working on a local fully .yml implementation of the build, and it might be worth incorporating something like this into the project CI builds, depending on the philosophy around such things. Thanks!

edit:
I haven't experimented with this, but I think the script should internally check for failures between stages with something like:
```
if %ERRORLEVEL% NEQ 0 (
   echo ERROR Failed to build LLVM
   exit /b %ERRORLEVEL%
)
```

AlexandreEichenberger:
[...]

PR #239, opened 2020-07-28
Fix mainutils win32

gongsu832:


gongsu832:
@AlexandreEichenberger I think @jcwchen setup the Windows CI so he should be the best person to look at it as I have no idea how the pipeline is setup or which script is involved.

caoimhinuibrian:
@gongsu832 @AlexandreEichenberger This issue has been resolved - it's just in the Windows CI and @NathanielMcVicar is addressing it ... see the comments in the two relevant issues

jcwchen:
@gongsu832 @AlexandreEichenberger
As mentioned here https://github.com/onnx/onnx-mlir/issues/234#issuecomment-665417580, I can help to fix the Windows CI bug if needed. Thanks.

src/MainUtils.cpp
```diff
@@ -223,6 +223,11 @@ void genConstPackObj(const mlir::OwningModuleRef &module,
       .exec();
 
 #else
+  /* FIXME: final object file should be set in constPackObjPath
```
tjingrant, line 226: Hi Gong, I'm not seeing what needs fixing. The patch presented here seems to resolve the issue fully?

Also, the constant pack for platforms that are **not** Linux or Mac shouldn't be deleted. This is because baking the parameters into the shared library is a very platform-specific process, and I only coded up the mac & Linux version; for platforms like Windows, we resort to having separate files for shared library and parameter constant packs.

So to keep the naming consistent between the constant pack & the shared library, they will share the same name (conveyed via `outputBaseName`) except for the suffix (e.g., `mnist.so` for shared library and `mnist.bin` for parameters) on platforms like Windows.

gongsu832: @tjingrant Great. That's exactly why you need to take a lot and make sure that my fix didn't just work by luck. I was obviously under the wrong impression that the constant pack on Windows should be deleted once it's embedded into the shared library like on Linux and MacOS. I will update the comment to reflect this. Thanks.

PR #241, opened 2020-07-29
Catch build failure on Windows-CI (Azure Pipelines)

jcwchen:
**Description** 
This PR fixes error reporting for Windows-CI.

**Motivation and Context**
https://github.com/onnx/onnx-mlir/issues/234
Since sometimes Azure Pipelines cannot catch the exit code normally, Azure Pipelines needs to manually catch the error after building scripts.

cc @NathanielMcVicar


tjingrant:
@jcwchen please disregard DocCheck error, I'll fix it later.

jcwchen:
> @jcwchen please disregard DocCheck error, I'll fix it later.

That would be great. Thank you!

tjingrant:
Hi @jcwchen do you think this issue is specific to Azure Pipeline or is it something that people interested in using onnx-mlir on Windows should pay attention to in general?

In other words, I'm trying to decide whether we want to update our Windows user installation guide to include this check? Or is it sufficient to keep it within our CI infrastructure.

jcwchen:
Hi @tjingrant,
I suggest it's kind of a common "bug" in Azure Pipelines. The check is only for the CI.
Since exit codes of certain commands in Azure Pipelines sometimes cannot normally be caught, it's better to catch them manually. 
So we don't need to update installation guide for Windows in this case. Thanks!

jcwchen:
BTW, I think the files under `/utils` are for installation instruction purpose?
In that case, maybe I need to decouple my modification (only for CI) from them?
If yes, I will make a modification for it.

tjingrant:
Sorry for my late response.

> BTW, I think the files under /utils are for installation instruction purpose?
> In that case, maybe I need to decouple my modification (only for CI) from them?
> If yes, I will make a modification for it.

Yes, this current state of the patch is ideal.

tjingrant:

PR #243, opened 2020-07-30
Simplify cmpi operations with identical operands

doru1004:
This patch enables the following operation:

```
%3 = cmpi <comp_op>, %0, %0 : index
```

to be replaced with a constant operation signifying true or false depending on the type of <comp_op> employed. Currently this type of operations are not being simplified away.


doru1004:
The pass is used to simplify standard dialect operations. It can be applied at any level that contains standard library operations. In this particular case I apply it at the level of the Krnl dialect i.e. once the Krnl operations have been emitted and standard dialect operations are present in the code.

AlexandreEichenberger:
Would you consider this experiement? What if you added the case you have here (namely compare a result with self) in the constant folding of the op: Look at mlir/lib/dialect/StandardOps/IR/Ops.cpp

```
// Constant folding hook for comparisons.
OpFoldResult CmpIOp::fold(ArrayRef<Attribute> operands) {
  assert(operands.size() == 2 && "cmpi takes two arguments");

 // DORU: check here the case compare with self here

  auto lhs = operands.front().dyn_cast_or_null<IntegerAttr>();
  auto rhs = operands.back().dyn_cast_or_null<IntegerAttr>();
  if (!lhs || !rhs)
    return {};

  auto val = applyCmpPredicate(getPredicate(), lhs.getValue(), rhs.getValue());
  return IntegerAttr::get(IntegerType::get(1, getContext()), APInt(1, val));
}
```

It may seem to be a more natural fit than a custom phase. We could do the same with a `select(cond=?, then=a, else=a)`

doru1004:
We don't need to worry about the `select` op. Once the the input to the select is a constant (true/false) then the select will be replaced by its result for us.

doru1004:
The optimization cannot be added to constant folding in MLIR like suggested above since fold assumes attributes as inputs. The optimization should be applied to Value typed operands.

The only place it could be added is as a pattern attached to MLIR's canonicalization pass.

doru1004:
I've submitted this patch as an MLIR patch here: https://reviews.llvm.org/D84990
If it gets approved we can close this one.

doru1004:
Patch due to land in MLIR: https://reviews.llvm.org/D84990

doru1004:
Patch due to land in MLIR: https://reviews.llvm.org/D84990

PR #244, opened 2020-07-30
Assorted fixes for memory pool related passes

doru1004:
Minor changes to make memory pool passes more robust.


src/MainUtils.cpp
```diff
@@ -384,6 +384,7 @@ void addONNXToMLIRPasses(mlir::PassManager &pm) {
 
 void addONNXToKrnlPasses(mlir::PassManager &pm) {
   pm.addPass(mlir::createLowerToKrnlPass());
+  pm.addPass(mlir::createCanonicalizerPass());
```
AlexandreEichenberger, line 387: Can you add a comment indicating what the goal of this pass is, thanks

doru1004: Not needed anymore actually. Good catch.

PR #245, opened 2020-07-31
Call LLVM opt to optimize bitcode

tungld:
This patch is to call LLVM `opt` to optimize the generated bitcode before creating an object file.

The inference time for ResNet50 decreased by a factor of 3.


src/MainUtils.cpp
```diff
@@ -251,15 +251,27 @@ void genConstPackObj(const mlir::OwningModuleRef &module,
...
 
   llvm::WriteBitcodeToFile(
       *mlir::translateModuleToLLVMIR(*module), moduleBitcodeStream);
   moduleBitcodeStream.flush();
+
+  // Optimize the bitcode.
+  Command optBitcode(/*exePath=*/kOptPath);
+  optBitcode.appendStr("-O3")
+      .appendList({"-o", bitcodePath})
+      .appendStr(bitcodeTempPath)
```
tjingrant, line 273: Can we use more expressive variable names? I.e., 
`bitcodePath` -> `optimizedBitCodePath`
`bitcodeTempPath` -> `unoptimizedBitCodePath`

tungld: Tian, it's done. Thanks!

PR #250, opened 2020-08-05
Move Shape Inference Pass before Constant Propagation

alberto-magni:
Before this change, running "onnx-mlir --EmitONNXIR" on this pattern:

```
%c = "onnx.Constant"() {value = dense<"..."> : tensor<2x2xf32>} : () -> tensor<2x2xf32>
%t = "onnx.Transpose"(%32) {perm = [1, 0]} : (tensor<2x2xf32>) -> tensor<*xf32>
```

would result in this cast (in ConstPropTranspose):

`ShapedType resType = resOperand.getType().cast<RankedTensorType>();`

to fail.

This is because the `resOperand `(the output of onnx.Transpose)
is an `UnrakedTensor `at this point.

This failure is fixed by moving the Shape Inference transformation to
run before Constant Propagation.


tjingrant:
Ping @AlexandreEichenberger , I think how shape inference gets invoked needs more work in general, but I wonder if you have any concerns w.r.t. this patch?

doru1004:
@AlexandreEichenberger do you have any objections regarding this patch?

chentong319 (commented):
As @tjingrant commented previously that the pass ordering is a general issue. Different test cases may require different sequence. By switch the order, this patch will work on the mentioned example. Obviously there may be cases in which shape inference should go after constant propagation. 
If this patch is really needed, it is fine to me. Otherwise, I suggest close this PR and remember this issue .

PR #251, opened 2020-08-05
Remove benchmark submodule

NathanielMcVicar:
This PR is for discussion of the Google benchmark submodule. I'm not sure, in terms of project etiquette, if a PR or an issue is better (or if there is some other discussion forum for ONNX MLIR) @doru1004 @tjingrant please let me know if that is preferred and I will abandon this. Thanks!

Currently ctest (and thus ``cmake --build . --target test``) has many failures in Google benchmark (and submodule gtest) [across platforms, I replicated on Windows and Ubuntu], making it less useful for adding additional tests (the only ONNX MLIR specific test currently is ``OMTestConv`` which passes). As far as I can tell benchmark isn't actually used in the project. Since ONNX also includes the same submodule and both are chained through cmake ``add_subdirectory`` there can be version mismatches between them leading to build problems. I observed this on Windows, where Visual Studio has imperfect submodule support.


tjingrant:
@NathanielMcVicar thanks for looking into the unit tests.

From what I understood, the (potential) test failure is most likely due to test executables not being produced by default. If you choose to invoke "make all", then all unit tests drivable by ctest will be produced & run successfully. 

OTOH, even though the benchmark subproject is not utilized right now, I've relied heavily on it in the past to evaluate key micro-benchmark performance achieved by compilers; so I expect this subproject to be increasingly useful as we mature.

NathanielMcVicar:
As far as I can tell the failures are mostly just due to paths. If you run tests from your main build directory the tests binaries go in bin (on Linux) which isn't in the default ctest search path. Do you see something different locally? There are also some tests in the gtest default that are known to fail on some system/compiler combinations (I'm seeing this one https://github.com/google/benchmark/issues/578 with Ubuntu 18.04/clang9).

One option to keep benchmark and allow ctest to work, would be to not build the default gtest suite by setting ``BENCHMARK_ENABLE_GTEST_TESTS`` to OFF in the top level ONNX MLIR cmake.

tjingrant:
> As far as I can tell the failures are mostly just due to paths. If you run tests from your main build directory the tests binaries go in bin (on Linux) which isn't in the default ctest search path. Do you see something different locally? 

Yes, we do see similar issues here and we have an ugly workaround, for example, in our jenkins CI:
https://github.com/onnx/onnx-mlir/blob/dbe0d734b5687e0aa7da911684912163cea07bd2/.buildbot/z13.sh#L75

> One option to keep benchmark and allow ctest to work, would be to not build the default gtest suite by setting BENCHMARK_ENABLE_GTEST_TESTS to OFF in the top level ONNX MLIR cmake.

That would be reasonable default behavior since most users wouldn't have a need for micro-benchmarking.

NathanielMcVicar:
I can abandon this PR and create one to turn off the gtest suite by default, if you think that's an acceptable solution. As I understand it, this won't prevent benchmarking, as it will still be possible to ``add_test`` the benchmarks from any ONNX MLIR CMake file.

Otherwise, if you prefer the current approach, I'll just abandon the PR and we'll address our own unit tests locally. Thanks!

tjingrant:
> I can abandon this PR and create one to turn off the gtest suite by default, if you think that's an acceptable solution. 

That sounds reasonable to me!

NathanielMcVicar:
@tjingrant Unfortunately there are some fixes to Google benchmark/test that are more recent than the fork we are using. Is there a reason the project is using a fork for the benchmark submodule instead of the master repo? Thanks!

PR #256, opened 2020-08-08
Fix build config validation on Travis CI

mshr-h:
Fixed build config validation shown in https://travis-ci.org/github/onnx/onnx-mlir/builds/715883913/config.
> Build config validation 
> root: deprecated key sudo (The key `sudo` has no effect anymore.)
> root: missing os, using the default linux


tjingrant:

PR #261, opened 2020-08-11
Add krnl.dim operation and lowering pass

doru1004:
This patch enables the lowering of a newly added `krnl.dim` operation to code which is independent of the alloc operation which defines the MemRef which participates in the `dim` operation.


src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
```diff
@@ -514,3 +514,20 @@ Value getDynamicMemRefSizeInBytes(
 
   return result;
 }
+
```
AlexandreEichenberger, line 517: Just add some comments here.

Consider the following "alloc(p0, p1, p2) : memref<1x?x2x?x?x3xf32, map>" with 3 parameters (in general pn parameters).

We are looking for the parameter p0..pn that correspond to the "?" dynamic parameter pointed by index in the memref. To find the right parameter, scan the memref to find the number of '?` that precede the one pointed to by index. Return that parameter index number

doru1004: Add in header file.

src/Transform/DisconnectKrnlDimFromAlloc.cpp
```diff
@@ -0,0 +1,104 @@
...
+class DisconnectKrnlDimFromAlloc : public OpRewritePattern<KrnlDimOp> {
+public:
+  using OpRewritePattern<KrnlDimOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(
+      KrnlDimOp krnlDimOp, PatternRewriter &rewriter) const override {
+    auto loc = krnlDimOp.getLoc();
+
+    // If index is not constant, return failure.
+    // TODO: support dynamic index case.
```
AlexandreEichenberger, line 48: I can't imagine why we would need to support a dynamic index: since we are looking at the memref, looking at the '?', and then adding them to the alloc function as parameter.

PR #263, opened 2020-08-12
Added builder case for int8 tensors in the dialect

lszinv:
Adding a select case in the helper to pick up int8 tensors


tjingrant:

PR #265, opened 2020-08-14
OneHotEncoder Shape Inference

hleu:


src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2775,6 +2775,54 @@ LogicalResult ONNXDropoutOp::inferShapes() {
   return success();
 }
 
+//===----------------------------------------------------------------------===//
+// OneHotEncoder
+//===----------------------------------------------------------------------===//
+
+LogicalResult ONNXOneHotEncoderOp::inferShapes() {
+  ShapedType inputType = X().getType().dyn_cast<ShapedType>();
+  if (!X())
```
chentong319, line 2784: not needed. You already called getType() on it.

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2775,6 +2775,54 @@ LogicalResult ONNXDropoutOp::inferShapes() {
...
+                       "no cats_strings attribute");
+    outDim = ArrayAttrSize(cats_strings());
+  }
+
+  // Encoded output data, having one more dimension than X
+  // total category count will determine the size of the extra dimension
+  int8_t flag = 0;
+  SmallVector<int64_t, 2> dims;
+  for (int i = 0; i != shape.size(); ++i) {
+    if (i == shape.size() - 1 && shape[i] == 1) {
```
chentong319, line 2811: not sure whether this special case need to be "optimized" or not. Uniformly attach the one hot size to the last dimension may be better .

PR #266, opened 2020-08-14
Add shape inference and names

amanwithfruit:
- Add shape inference for PRelu
 - Fix shape inference for group conv for ConvTranspose
 - Add input and output names for graphs (functions)
 - Add support for uint8 tensor attributes
 - Add Slice op importing for some cases where an (older version, maybe) .onnx file gets imported with only attrs and none of the required operands, and so fails verification.
 - Add name importing to make it easier to match up compiled model I/O with ONNX model I/O
 - Other small point fixes


src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -406,7 +494,9 @@ class FrontendGenImpl {
     // the generic operator is used
     // one known reeason is the optional input
 
-    (this->*(import_handler_map_[opName.str()]))(node);
+    auto found = import_handler_map_.find(opName.str());
+    assert(found != import_handler_map_.end() && "Could not find op importer");
```
amanwithfruit, line 498: We ran into the case where this just faulted and tracking it down was tricky - which is why we added this code.

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -452,20 +542,39 @@ class FrontendGenImpl {
...
-    auto mainFunc =
-        mlir::FuncOp::create(UnknownLoc(), name, funcType, /* attrs = */ {});
+    auto mainFunc = mlir::FuncOp::create(UnknownLoc(), name, funcType,
+        /* attrs = */ llvm::makeArrayRef(funcAttrs));
 
     // Emit the entry point operation which specifies the number of user
     // inputs and outputs.
     auto entryPoint = mlir::ONNXEntryPointOp::create(UnknownLoc(), mainFunc,
-        /*numInputs=*/graph.input().size() - graph.initializer().size(),
+        /*numInputs=*/graph.input().size() < graph.initializer().size()
```
amanwithfruit, line 575: Lots of our inputs had huge negatives in this attribute and I decided to fix it. I'd be more than happy to remove this change if it's unpopular :)

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -1500,7 +1511,7 @@ LogicalResult ONNXConvOp::inferShapes() {
 LogicalResult ONNXConvTransposeOp::inferShapes() {
   // Generic shape for data input X, weight tensor W, and optional bias B
   // X: (N x C x D1 x D2 ... x Dn)
-  // W: (M x C/group x k1 x k2 x ... x kn)
+  // W: (C x M/group x k1 x k2 x ... x kn)
```
amanwithfruit, line 1514: This follows the pytorch/onnx spec more closely - looks like it was a copy-paste error before.

src/Builder/FrontendDialectTransformer.cpp
doru1004: What corner case made this code error? Missing operation?

src/Builder/FrontendDialectTransformer.cpp
doru1004: How does the initializer have more elements than the number of inputs in your case? I'm assuming that's what you mean by "huge negatives".

src/Builder/FrontendDialectTransformer.cpp
amanwithfruit: IIRC the case was that during development we had added a custom ImportNodeSlice (and the corresponding code in gen_onnx_mlir.py) but had accidentally commented the function itself out, which caused a segfault.

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -1164,14 +1174,13 @@ LogicalResult ONNXBatchNormalizationTestModeOp::inferShapes() {
...
+  // 2-D tensors are assumed to be of shape NxC
   // Shapes of scale, bias, mean and variance must be C.
   int64_t c = -1;
   if (inputTensorTy.getShape().size() == 1) {
     c = 1;
-  } else if (inputTensorTy.getShape().size() > 2) {
+  } else if (inputTensorTy.getShape().size() >= 2) {
     c = (inputTensorTy.getShape()[1] != -1) ? inputTensorTy.getShape()[1] : -1;
-  } else {
-    return emitError("Wrong rank for the input");
```
doru1004, line 1174: Why was rank == 2 a problem before? @AlexandreEichenberger @tjingrant @tungld

src/Builder/FrontendDialectTransformer.cpp
amanwithfruit: You're right - The initializer had things like weights, whereas the inputs just had the input tensors. This might be a difference in how the models are exported - training vs inference mode.

doru1004:
Thanks for all the fixes. In general it's faster for us to review this kind of patches if they are more focused on a particular type of fix.

Regarding your list of changes:

(OK with me) Add shape inference for PRelu
(OK with me) Fix shape inference for group conv for ConvTranspose
(en example of what you mean by this?) Add input and output names for graphs (functions)
(OK with me, you also add support for DOUBLE attributes) Add support for uint8 tensor attributes

(@AlexandreEichenberger @tjingrant ) Add Slice op importing for some cases where an (older version, maybe) .onnx file gets imported with only attrs and none of the required operands, and so fails verification.

(an example of what you mean would be nice) Add name importing to make it easier to match up compiled model I/O with ONNX model I/O

(@chentong319  there are some fixes to Pad operation, can you have a look?) Other small point fixes

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -155,6 +155,12 @@ class FrontendGenImpl {
       auto attr = node.attribute(i);
       attributes.push_back(convertOnnxAttributeProtoToMlirNamedAttribute(attr));
     }
+
+    // If the node has a name, then import it.
+    if (node.has_name()) {
+      attributes.push_back(
+          builder_.getNamedAttr("name", builder_.getStringAttr(node.name())));
+    }
```
doru1004, line 163: @tjingrant

amanwithfruit:
Responses inline

> Thanks for all the fixes. In general it's faster for us to review this kind of patches if they are more focused on a particular type of fix.

Sorry about that - I have to get approval for each patch so I have to submit them somewhat in bulk sadly. I'd much rather submit smaller patches too!

> Regarding your list of changes:
> 
> (OK with me) Add shape inference for PRelu
> (OK with me) Fix shape inference for group conv for ConvTranspose
> (en example of what you mean by this?) Add input and output names for graphs (functions)

Many of the graphs we run into have names on the inputs and outputs that need to be propagated forward in the compiler pipeline. Basically if I have a graph with an input of name "animal_image" and an output of name "animal_kind" then these would be added as attributes onto the FuncOp that is created from the graph. This is so that if there are 2 inputs (say, "image" and "words") the compiler can re-order them if necessary and we can still write the correct data to the correct input.

> (OK with me, you also add support for DOUBLE attributes) Add support for uint8 tensor attributes
Yes thanks - I forgot to update the commit message.

> (@AlexandreEichenberger @tjingrant ) Add Slice op importing for some cases where an (older version, maybe) .onnx file gets imported with only attrs and none of the required operands, and so fails verification.
> 
> (an example of what you mean would be nice) Add name importing to make it easier to match up compiled model I/O with ONNX model I/O

Some of the nodes have names - the onnx::NodeProto has a `name` field. It's the same spirit as the input/output names, it makes it easier to match up which operation is which even if the compiler reorders them. This happens a lot with the MLIR folding, so having names on named nodes is super helpful for debugging, at least for us.

> (@chentong319 there are some fixes to Pad operation, can you have a look?) Other small point fixes

src/Builder/FrontendDialectTransformer.cpp
doru1004: I see, in the examples I've seen, everything is an input and a subset of those inputs have initializers which means they are constant. You have initializers which are not also inputs? Or you have multiple initializers for the same input?

amanwithfruit: Yes - it seems like our weights are treated as initializers that are not also inputs.

doru1004: How are those weights used by your model then? Are they global?

src/Builder/FrontendDialectTransformer.cpp
doru1004: I'm ok if this leads to a better error message which it looks to me like it does. @tjingrant

src/Builder/FrontendDialectTransformer.cpp
amanwithfruit: They get used inside the FuncOp that is created for the graph, or am I misunderstanding? They are just normal weights, I'm not sure why they are in the initializer and not in the inputs other than that they are constant at export time.

doru1004: Ah my fault, I didn't mean used, I meant more like consumed by your model. How are they passed to your model. I'm trying to understand whether there is something going on in the ONNX model itself where you use the weights inside the main function without passing them in the argument list (or something like that), i.e. they are just globals.

amanwithfruit: ah - yeah they're essentially globals. We load them out-of-band with regards to ONNX

doru1004: Got it, in that case the actual fix would have to look a little but different.

Out of all the actual input arguments to the main function, you would have to exclude the ones which have an initializer from counting towards `numInputs`.  If in your case it happens to be that none of your actual input arguments have initializers (i.e. you can get away with returning `graph.input().size()`) then that's just a happy corner case but in general you need to support the case where you have a mix. To do that you'd have to find which initializers initialize one of the input args and only subtract that number from the total number of input args.

amanwithfruit: Thank you for clarifying - I'll fix that :)

amanwithfruit: I think the latest commit should be the actual fix as you described.

doru1004: Looks good now thanks for that! :)

amanwithfruit: Happy to help! :)

src/Dialect/ONNX/ONNXOps.cpp
tungld: We never have an input of rank 2 as described in ONNX (https://github.com/onnx/onnx/blob/master/docs/Operators.md#BatchNormalization). Only N or NxCxD1xD2x...xDn.

amanwithfruit: We do see networks that have this though...exported from pytorch (which is why we needed to change this)

tungld: Have you confirmed that the onnx model exported from pytorch is runnable, say, with onnxruntime? It is helpful if you can share the model here. I just wanted to be convinced about this case since the ONNX document does not mention it. Thanks!

src/Builder/FrontendDialectTransformer.cpp
chentong319: It is OK to use "name". But a more special name for this attribute may be better. At least let's document this choice.

src/Dialect/ONNX/ONNXOps.cpp
doru1004: That would be great! Thanks a lot for looking into that!

src/Builder/FrontendDialectTransformer.cpp
amanwithfruit: Just so I know what's going on - would you like me to add a comment here or should that documentation go elsewhere?

src/Dialect/ONNX/ONNXOps.cpp
amanwithfruit: From looking in the code I don't see anything obvious that would break. I'm trying to add a test now, but I'm getting git submodule issues so I will have to resolve those before I can add the test.

src/Builder/FrontendDialectTransformer.cpp
chentong319: Could you change "name" to "onnx_node_name"?

amanwithfruit: Sure, no problem.

src/Dialect/ONNX/ONNXOps.cpp
amanwithfruit: @doru1004 I added a 2D batchnorm test and made the changes in Conversion/NN/Normalization to make it work (it was a very simple modification) - let me know what you think?

doru1004:
@amanwithfruit in your last commit message it says you added a test but I don't see any code changes inside test files, did you forget to add the new test file? :-)

amanwithfruit:
@doru1004 sorry old habit - I usually merge with `squash + merge` so I like to summarize the last few commits so I can craft the squashed commit message easier. I'd be happy to change it if you'd prefer :)

src/Dialect/ONNX/ONNXOps.cpp
AlexandreEichenberger: thanks, it was an error indeed

test/mlir/onnx/onnx_shape_inference.mlir
```diff
@@ -350,22 +350,59 @@ func @test_conv_12(%arg0 : tensor<1x2x32xf32>, %arg1 : tensor<5x2x6xf32>, %arg2
...
+  %cst = constant unit
+  %0 = "onnx.ConvTranspose"(%arg0, %arg1, %cst) {dilations = [1, 1], kernel_shape = [2, 2], pads = [0, 0, 0, 0], strides = [2, 2]} : (tensor<1x64x36x48xf32>, tensor<64x1x2x2xf32>, none) -> tensor<*xf32>
+  "std.return"(%0) : (tensor<*xf32>) -> ()
+
+  // CHECK-LABEL: test_conv_transpose_1
+  // CHECK: [[RES_ATTR:%.+]] = "onnx.ConvTranspose"(%arg0, %arg1, %cst) {auto_pad = "NOTSET", dilations = [1, 1], group = 1 : i64, kernel_shape = [2, 2], output_shape = [1, 1, 72, 96], pads = [0, 0, 0, 0], strides = [2, 2]} : (tensor<1x64x36x48xf32>, tensor<64x1x2x2xf32>, none) -> tensor<1x1x72x96xf32>
+  // CHECK: return [[RES_ATTR]] : tensor<1x1x72x96xf32>
+}
+
+func @test_conv_transpose_2(%arg0 : tensor<1x64x36x48xf32>, %arg1 : tensor<64x1x2x2xf32>) -> tensor<*xf32> {
```
AlexandreEichenberger, line 367: Thanks for including a test with group>1

doru1004:
> @doru1004 sorry old habit - I usually merge with `squash + merge` so I like to summarize the last few commits so I can craft the squashed commit message easier. I'd be happy to change it if you'd prefer :)
> 
> The test is the lowering test for 2d batchnorm I added in [ffcb47d](https://github.com/onnx/onnx-mlir/commit/ffcb47d192b4a4fb4b79b8220ab410257cb539e6) but it actually works with [e18c750](https://github.com/onnx/onnx-mlir/commit/e18c750c2ff506a197e90e008131be6eda55923f)

Ah, it's my fault I only checked the last commit :-) No need to change it, no worries.

doru1004:

PR #267, opened 2020-08-14
Introduce krnl.shape operation and its lowering

doru1004:
This patch adds a new operation called krnl.shape which returns the shape of a MemRef (with or without a map) using the `shape` type of the MLIR Shape Dialect:

```
%shape = krnl.shape(%alloc) : (memref<?x10xf32, #map>) -> !shape.shape
```

This patch also add the following transformation pass:

```
    %1 = alloc(%0) : memref<?x10xf32>
    %shape = "krnl.shape"(%1) : (memref<?x10xf32>) -> !shape.shape
```

to

```
    %1 = alloc(%0) : memref<?x10xf32>
    %2 = "krnl.dim"(%1, %c0) : (memref<?x10xf32>, index) -> index
    %3 = "krnl.dim"(%1, %c1) : (memref<?x10xf32>, index) -> index
    %4 = shape.from_extents %2, %3
```


src/Transform/LowerKrnlShape.cpp
```diff
@@ -0,0 +1,85 @@
...
+    ConversionTarget target(getContext());
+    OwningRewritePatternList patterns;
+    patterns.insert<LowerKrnlShape>(&getContext());
+
+    applyPatternsAndFoldGreedily(function, patterns);
+  }
+};
+} // namespace
+
+std::unique_ptr<Pass> mlir::createLowerKrnlShapePass() {
```
AlexandreEichenberger, line 83: Please add a TODO: integrate in other passes

doru1004: Done

PR #268, opened 2020-08-16
Build llvm nightly prereq images & test onnx-mlir against it regularly.

tjingrant:
1. Enable building llvm prereq docker on a nightly basis on x86 (though Github Action) and s390x (though jenkins).
2. On Travis, enable testing onnx-mlir with nightly llvm prereq images.
3. Allow these builds with nightly llvm images to fail on Travis.
4. Rebuild all outdated Docker prereq, now Travis build should succeed.

Nit:
- Actually use the docker build arg BASE to determine the base image.
- Rename amd64 -> x86 as x86 is more widely recognized.
- On docker hub, the image tagged "initial" should no longer be used, a more accurately tagged image "x86" should be used.


.travis.yml
```diff
@@ -22,12 +38,12 @@ script:
   - echo "about to execute build inside docker"
   - docker exec build df -h
   - docker exec build compile-onnx-mlir.sh
-  - docker commit build onnxmlirczar/onnx-mlir-build:$TRAVIS_CPU_ARCH
+  - docker commit build onnxmlirczar/onnx-mlir-build:$CPU_ARCH
 after_success: 
-  - if [ $TRAVIS_PULL_REQUEST == false ] && [ $TRAVIS_BRANCH == "master" ]; then
+  - if [ $TRAVIS_PULL_REQUEST == false ] && [ $TRAVIS_BRANCH == "master" ] && [ $NIGHTLY == fase ]; then
```
caoimhinuibrian, line 43: change fase to false

.travis.yml
```diff
@@ -7,11 +7,27 @@ arch:
   - s390x
   - ppc64le
   - amd64
+env:
+  - NIGHTLY=true
+  - NIGHTLY=false
```
caoimhinuibrian, line 12: Why set NIGHTLY twice here?

tjingrant: It's to specify all values of NIGHTLY that needs to be tested. So the combination of arch and env expands to:
```
s390x-true
s390x-false
ppc64le-true
ppc64le-false
...
```
resulting in 6 jobs.

PR #272, opened 2020-08-19
Make krnl-to-llvm patterns reusable

tungld:


src/Conversion/KrnlToLLVM/KrnlToLLVM.cpp
```diff
@@ -22,6 +21,7 @@
 #include "onnx/onnx_pb.h"
 #include "llvm/ADT/Sequence.h"
 
+#include "src/Conversion/KrnlToLLVM/KrnlToLLVM.hpp"
```
tjingrant, line 24: We should drop the path prefix and make it:
```
#include "KrnlToLLVM.hpp"
```
And perhaps move it to a lower position; because this is the corresponding header file declaring the function signatures implemented in the function definition file.

tjingrant (approved):
LGTM!

PR #273, opened 2020-08-25
change tag to x86

caoimhinuibrian:
tag of build image was changed by PR #268 from amd64 to x86 --- fix here to pick up latest build

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ seen in the docs folder. The Dockerfile is shown here.
 
 [same-as-file]: <> (docs/docker-example/Dockerfile)
 ```
-FROM onnxmlirczar/onnx-mlir-build:amd64
+FROM onnxmlirczar/onnx-mlir-build:x86
 
 WORKDIR /build
 ENV HOME=/build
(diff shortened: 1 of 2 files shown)


gongsu832:
@caoimhinuibrian Any particular reason to make the change? amd64 is the official docker arch name for 64-bit Intel. x86 typically refers to 32-bit Intel and docker arch name is i386 (https://github.com/docker-library/official-images#architectures-other-than-amd64). Although this is for the tag instead of the manifest, it's probably a good idea to keep the two the same.

caoimhinuibrian:
@gongsu832 PR #268 by @tjingrant  changed the tag from amd64 to x86 in the build. So this docker file was picking up the old image from before that change. I'm not sure what the rationale for the build change was, but they need to agree.

gongsu832:
@caoimhinuibrian @tjingrant OK. The proper arch name should be amd64. In fact, for a true multi-arch image, the tag should not have the arch in it, i.e., all archs should have the same tag. instead, you build and push an image + a manifest for each arch and the manifest specifies the arch of the image.

caoimhinuibrian:
@gongsu832 @tjingrant OK ... we can let this land and go back and revisit the underlying problem

doru1004:
@gongsu832 if a more involved fix is required for this, would you be ok with opening an issue on this just to keep track of this problem?

caoimhinuibrian:
@doru1004 sure, will do that

PR #277, opened 2020-08-27
Update available python3 version on AzurePipelines

jcwchen:
**Description**
Update available python3 version (3.7.9) on AzurePipelines

**Motivation and Context**
CI on Azurepipelines is failing because the python3 version (3.7.7) s not available.


jcwchen:
The commit log is confusing. Will open another one.

PR #281, opened 2020-08-27
 shape inference for Size Op

chentong319:
A simple PR of shape inference for Size. [https://github.com/onnx/onnx/blob/master/docs/Operators.md#Size](url)
The Size op always returns a scalar of i64, namely,  tensor<i64>. 
Our current importer of Size Op gives the output type unknown shape of i64 tensor, tensor<*xi64>. It is an alternative approach to define a special builder for SizeOp.


doru1004:
I'm guessing that the changes to the submodules are not intentional. Could you remove them from the patch if that's the case? If they are intentional you should post a separate patch with just those changes.

AlexandreEichenberger:
@chentong319 please determine if the PR needs to change the third_party files; if not, I suggest to make two different PR and keeping this one only for the Size related changes

PR #284, opened 2020-08-31
Update llvm commit ID to 1d01fc1

imaihal:
This PR includes fixes for LLVM revisions: [D85495](https://reviews.llvm.org/D85495), [D85622](https://reviews.llvm.org/D85622), and [D86121](https://reviews.llvm.org/D86121) to use newer LLVM commit.

- D85495
`TypeID` should be set to `mlir::Dialect` by LLVM revision [D85495](https://reviews.llvm.org/D85495) I updated as it is done at `ToyDialect` in the revision https://reviews.llvm.org/D85495#change-M9kXqI8sOFkn

- D85622
This revision includes change in how to register dialects. I updated as it is done in the revision. For example. https://reviews.llvm.org/D85622#change-2drP7BXp5CDr

- D86121
`type.getKind()` was removed by LLVM revision [D86121](https://reviews.llvm.org/D86121). So, I needed to modify it as described in https://llvm.discourse.group/t/removing-kinds-from-attributes-and-types/1547
I used `TypeSwitch<Type>(type)` instead of `switch (type.getKind())` 

I confirmed this works on [1d01fc1](https://github.com/llvm/llvm-project/commit/1d01fc100bb5bef5f5eaf92520b2e52f64ee1d6e) (Aug 31, 2020)

### DIFF
--- a/src/Dialect/Krnl/KrnlOps.hpp
+++ b/src/Dialect/Krnl/KrnlOps.hpp
@@ -17,6 +17,7 @@
 #include "mlir/IR/Function.h"
 #include "mlir/IR/OpDefinition.h"
 #include "mlir/IR/StandardTypes.h"
+#include "llvm/ADT/TypeSwitch.h"
 
 #include "KrnlHelper.hpp"
 #include "KrnlTypes.hpp"
@@ -38,11 +39,17 @@ public:
 
   /// Print a type registered to this dialect.
   void printType(Type type, DialectAsmPrinter &os) const override {
+    TypeSwitch<Type>(type).Case<LoopType>([&](Type) {
+      os << "loop";
+      return;
+    });
+    /*
     switch (type.getKind()) {
     case KrnlTypes::Loop:
       os << "loop";
       return;
     }
+    */
   }
 };
 
--- a/src/Dialect/Krnl/KrnlOps.hpp
+++ b/src/Dialect/Krnl/KrnlOps.hpp
@@ -43,13 +43,6 @@ public:
       os << "loop";
       return;
     });
-    /*
-    switch (type.getKind()) {
-    case KrnlTypes::Loop:
-      os << "loop";
-      return;
-    }
-    */
   }
 };
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,34 +14,39 @@ env:
 jobs:
   allow_failures:
     - env: NIGHTLY=true
-script:
-  - export CPU_ARCH=$TRAVIS_CPU_ARCH
-  - if [ "$CPU_ARCH" == "amd64" ] ; then
-      export CPU_ARCH="x86";
-    fi
-  - echo "CPU Architecture is " $CPU_ARCH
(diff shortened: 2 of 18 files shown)


AlexandreEichenberger:
Thanks for updating the LLVM level. In addition to the usual updates to the commit level, I see a lot of work on types. Can you comment on whether that is due to upgrades in MLIR or separate changes.

imaihal:
Thanks for your comments. All of the work in this PR is due to upgrade MLIR.

Regarding the types,  `type.getKind()` was removed by LLVM revision [D86121](https://reviews.llvm.org/D86121). So, I needed to modify it as described in https://llvm.discourse.group/t/removing-kinds-from-attributes-and-types/1547
I used `TypeSwitch<Type>(type)` instead of `switch (type.getKind())` 

I added descriptions about other works above.

imaihal:
I'm not sure how to fix other errors in travis and docker.  I guess this errors come from using old llvm images, but I don't know how to fix them. Could you give me your comments or suggestions?

AlexandreEichenberger:
@caoimhinuibrian Since the LLVM level is changed, I believe you mentioned that some of the tests will fail. Can you just check the list to see if the failures are expected? Thanks

caoimhinuibrian:
@AlexandreEichenberger @imaihal When the LLVM commit level changes, the TravisCI builds will always fail. The prereq docker images are for the previous version of LLVM. If you have verified that the build works in your fork, then it should be fine to merge the PR and subsequently update the prereq images for 390x and ppc64le. If you can't do that yourself, ask me to do it.

src/Dialect/Krnl/KrnlOps.hpp
```diff
@@ -38,11 +39,17 @@ class KrnlOpsDialect : public Dialect {
 
   /// Print a type registered to this dialect.
   void printType(Type type, DialectAsmPrinter &os) const override {
+    TypeSwitch<Type>(type).Case<LoopType>([&](Type) {
+      os << "loop";
+      return;
+    });
+    /*
```
AlexandreEichenberger, line 46: Please remove the old commented code

imaihal: I removed this code.

imaihal:
I confirmed FileCheck test and backend test passed on my s390 environment. However, `linux-s390x-docker` of this PR  fails in addition to `travis-ci`. Is this expected results?  If so, could you merge this?

@caoimhinuibrian I don't know how to create the prereq images. Could you create them?

tjingrant:
@imaihal I'm dealing with building pre-req images now, I will base my commits on this patch for convenience.

AlexandreEichenberger:
@imaihal could you update and try the test again? there was a problem with the linux-s390x and Gong did some work to fix it, Thanks

tjingrant:
Closing in favor of https://github.com/onnx/onnx-mlir/pull/292, we need privileged information on Travis to push new prereq docker images so it has to be a branch in the master repo, not a pull request from a forked repo.

PR #285, opened 2020-08-31
Rewrite shape and size OP 

chentong319:
When the input tensor of Shape and Size Op is ShapedTensor,  rewrite them with its constant output.

### DIFF
--- a/src/Transform/ONNX/Rewrite.cpp
+++ b/src/Transform/ONNX/Rewrite.cpp
@@ -27,36 +27,30 @@ DenseElementsAttr createDenseElementsAttrFromFloatAttr(
   return mlir::DenseElementsAttr::get(tensorType, llvm::makeArrayRef(values));
 }
 
+// Create a DenseElementsAttr based on the shape of type.
 DenseElementsAttr createDenseElementsAttrFromShape(
     PatternRewriter &rewriter, Value value) {
-  auto inType = value.getType().dyn_cast<ShapedType>();
+  auto inType = value.getType().cast<ShapedType>();
   ;
   if (!inType)
     llvm_unreachable("Shaped type is execptd\n");
   auto shape = inType.getShape();
-  SmallVector<int64_t, 1> dims(1, inType.getRank());
-  SmallVector<int64_t, 4> values;
-  for (auto s : shape) {
-    values.push_back(s);
-  }
+  SmallVector<int64_t, 1> dims = {inType.getRank()};
+  SmallVector<int64_t, 4> values(shape.begin(), shape.end());
   auto tensorType =
       mlir::RankedTensorType::get(dims, rewriter.getIntegerType(64));
   return mlir::DenseElementsAttr::get(tensorType, llvm::makeArrayRef(values));
 }
 
+// Create a DenseElementsAttr based on the size of type.
 DenseElementsAttr createDenseElementsAttrFromSize(
     PatternRewriter &rewriter, Value value) {
-  auto inType = value.getType().dyn_cast<ShapedType>();
+  auto inType = value.getType().cast<ShapedType>();
   ;
   if (!inType)
     llvm_unreachable("Shaped type is execptd\n");
-  auto shape = inType.getShape();
   SmallVector<int64_t, 1> dims(1, 1);
-  int64_t size = 1;
-  for (auto s : shape) {
-    size *= s;
-  }
-  SmallVector<int64_t, 1> values(1, size);
+  SmallVector<int64_t, 1> values = {inType.getNumElements()};
   auto tensorType =
       mlir::RankedTensorType::get(dims, rewriter.getIntegerType(64));
   return mlir::DenseElementsAttr::get(tensorType, llvm::makeArrayRef(values));
--- a/src/Transform/ONNX/Rewrite.cpp
+++ b/src/Transform/ONNX/Rewrite.cpp
@@ -27,6 +27,37 @@ DenseElementsAttr createDenseElementsAttrFromFloatAttr(
   return mlir::DenseElementsAttr::get(tensorType, llvm::makeArrayRef(values));
 }
 
+DenseElementsAttr createDenseElementsAttrFromShape(
+    PatternRewriter &rewriter, Value value) {
+  auto inType = value.getType().dyn_cast<ShapedType>();;
+  if (!inType) 
+    llvm_unreachable("Shaped type is execptd\n");
+  auto shape = inType.getShape();
+  SmallVector<int64_t, 1> dims(1, inType.getRank());
+  SmallVector<int64_t, 4> values;
+  for(auto s : shape) {
+    values.push_back(s);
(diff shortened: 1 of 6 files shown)


src/Transform/ONNX/Rewrite.cpp
```diff
@@ -27,6 +27,41 @@ DenseElementsAttr createDenseElementsAttrFromFloatAttr(
...
+  auto inType = value.getType().dyn_cast<ShapedType>();
+  ;
+  if (!inType)
+    llvm_unreachable("Shaped type is execptd\n");
+  auto shape = inType.getShape();
+  SmallVector<int64_t, 1> dims(1, 1);
+  int64_t size = 1;
+  for (auto s : shape) {
+    size *= s;
+  }
```
bondhugula, line 58: There is a `getNumElements()` accessor already available.

src/Transform/ONNX/Rewrite.cpp
```diff
@@ -27,6 +27,41 @@ DenseElementsAttr createDenseElementsAttrFromFloatAttr(
...
 }
 
+DenseElementsAttr createDenseElementsAttrFromShape(
+    PatternRewriter &rewriter, Value value) {
+  auto inType = value.getType().dyn_cast<ShapedType>();
+  ;
+  if (!inType)
+    llvm_unreachable("Shaped type is execptd\n");
+  auto shape = inType.getShape();
+  SmallVector<int64_t, 1> dims(1, inType.getRank());
```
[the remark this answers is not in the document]
chentong319: Thanks Uday. I updated the file for all your comments.

src/Transform/ONNX/Rewrite.cpp
```diff
@@ -27,6 +27,35 @@ DenseElementsAttr createDenseElementsAttrFromFloatAttr(
   return mlir::DenseElementsAttr::get(tensorType, llvm::makeArrayRef(values));
 }
 
+// Create a DenseElementsAttr based on the shape of type.
+DenseElementsAttr createDenseElementsAttrFromShape(
+    PatternRewriter &rewriter, Value value) {
+  auto inType = value.getType().cast<ShapedType>();
+  ;
```
AlexandreEichenberger, line 34: remove this lone `;` please

src/Transform/ONNX/Rewrite.cpp
```diff
@@ -27,6 +27,35 @@ DenseElementsAttr createDenseElementsAttrFromFloatAttr(
...
+  if (!inType)
```
AlexandreEichenberger, line 35: I though Uday stated that by using the "cast" it would assert if it is not the right type. Please verify.

src/Transform/ONNX/Rewrite.cpp
```diff
@@ -27,6 +27,35 @@ DenseElementsAttr createDenseElementsAttrFromFloatAttr(
...
+  auto tensorType =
+      mlir::RankedTensorType::get(dims, rewriter.getIntegerType(64));
+  return mlir::DenseElementsAttr::get(tensorType, llvm::makeArrayRef(values));
+}
+
+// Create a DenseElementsAttr based on the size of type.
+DenseElementsAttr createDenseElementsAttrFromSize(
+    PatternRewriter &rewriter, Value value) {
+  auto inType = value.getType().cast<ShapedType>();
+  ;
```
AlexandreEichenberger, line 49: same comments for the next 3 lines

test/mlir/onnx/onnx_canonicalization.mlir
```diff
@@ -191,8 +191,6 @@ func @test_conv_batchnormtestmode_fusion(%arg0 : tensor<1x3x224x224xf32>, %arg1
     // CHECK: return [[RES]] : tensor<1x64x112x112xf32>
 }
```
AlexandreEichenberger, line 193: I thought the 
```
// -------
```
were required to get some parallelism. Please add back unless it was discontinued

test/mlir/onnx/onnx_canonicalization.mlir
```diff
@@ -222,3 +216,49 @@ func @test_transpose_fusion_removal(%arg0: tensor<10x11x12x13xf32>) -> tensor<10
   // CHECK-NEXT: return %arg0 : tensor<10x11x12x13xf32>
   "std.return"(%1) : (tensor<10x11x12x13xf32>) -> ()
 }
+
+// -----
+
```
AlexandreEichenberger, line 221: Please add one comment for each new test to state what they do. LLVM requires it, and I think its a good practice.

src/Transform/ONNX/Rewrite.cpp
AlexandreEichenberger: Checked for you:

cast<>:
The cast<> operator is a “checked cast” operation. It converts a pointer or reference from a base class to a derived class, causing an assertion failure if it is not really an instance of the right type. This should be used in cases where you have some information that makes you believe that something is of the right type. An example of the isa<> and cast<> template is:

static bool isLoopInvariant(const Value *V, const Loop *L) {
  if (isa<Constant>(V) || isa<Argument>(V) || isa<GlobalValue>(V))
    return true;

  // Otherwise, it must be an instruction...
  return !L->contains(cast<Instruction>(V)->getParent());
}
Note that you should not use an isa<> test followed by a cast<>, for that use the dyn_cast<> operator.

dyn_cast<>:
The dyn_cast<> operator is a “checking cast” operation. It checks to see if the operand is of the specified type, and if so, returns a pointer to it (this operator does not work with references). If the operand is not of the correct type, a null pointer is returned. Thus, this works very much like the dynamic_cast<> operator in C++, and should be used in the same circumstances. Typically, the dyn_cast<> operator is used in an if statement or some other flow control statement like this:

if (auto *AI = dyn_cast<AllocationInst>(Val)) {
  // ...
}
This form of the if statement effectively combines together a call to isa<> and a call to cast<> into one statement, which is very convenient.

Note that the dyn_cast<> operator, like C++’s dynamic_cast<> or Java’s instanceof operator, can be abused. In particular, you should not use big chained if/then/else blocks to check for lots of different variants of classes. If you find yourself wanting to do this, it is much cleaner and more efficient to use the InstVisitor class to dispatch over the instruction type directly.

AlexandreEichenberger (commented):
minor changes requested

src/Transform/ONNX/Rewrite.cpp
chentong319: This description is what I understood cast and dyn_cast. We can choose:
1. dyn_cast follows by a if test, or
2. cast without if test
The constraint in the rewrite rule guarantees that the type is a shaped tensor. Thus, I will change code the just cast without test.

AlexandreEichenberger (approved):
LGTM, please add one liner for each tests

```
// test where size expand to one constant
// test where sizes does not expand to one constant, remain unchanged
```

Issue #252, opened 2020-08-06
Improving macOS support


mshr-h:
I've been using ONNX MLIR on macOS 10.15.6 for a few weeks now.
Basic functionality works correctly on my machine.
It looks like macOS is already supported but there are few areas which I can do to improve macOS support.
The following are my ideas.

- [x] Add macOS CI pipeline
- [x] Add CI status badge to README

If it's ok, I'll start working on it soon.

tjingrant:
Hi @mshr-h thanks for reaching out!

Some of us use macOS as our primary development platform so using onnx-mlir on macOS should be supported most of the time (however, as you obviously noted, it does break from time to time because we don't have CI testing for it). I'm fine with the proposed work items; as for 

> Add macOS CI pipeline,

 could you make it happen on Travis?

mshr-h:
> could you make it happen on Travis?
Ok.

caoimhinuibrian:
10.15.5 seems to be the latest version of OSX that Travis has. I can add that to our TravisCI build if there is a consensus.

mshr-h:
Here is the macOS versions which is supported by Travis.
https://docs.travis-ci.com/user/reference/osx#jdk-and-macos

There are two type of the latest osx_image, `osx_image: xcode12u` and `osx_image: xcode12`.
I'm not sure which one should be used for onnx-mlir.

mshr-h:
I looked at README and I noticed that there is already a chapter `Installation on UNIX` which include macOS installation.
So we don't need to do `Add build instruction on macOS to document`.

tjingrant:
@mshr-h to add travis support for osx, you'll need to have:
- A docker file for building dependent software (mostly llvm/mlir like here: https://github.com/onnx/onnx-mlir/blob/master/docker/prereq.amd64.Dockerfile), we call this docker image prereq image. 
- You'll need another much simpler dockerfile using the prereq image as a starting point to build onnx-mlir (like this one: https://github.com/onnx/onnx-mlir/blob/master/docker/x86.Dockerfile). The second image is used to build onnx-mlir every time Travis-CI is invoked.

This way, we effectively re-use llvm/mlir build across multiple Travis-CI runs, saving time & compute resources. The prereq image needs to be uploaded to docker hub (https://hub.docker.com/r/onnxmlirczar/onnx-mlir-llvmimage) once it's built, in order for Travis to access it and I can help you with the upload if you show me the dockerfile.

mshr-h:
[...]

PR #287, opened 2020-09-01
Added CI configuration for x86 macOS

mshr-h:
Added macOS CI configuration.
#252


tjingrant (approved):
LGTM, thanks!

tjingrant:
@mshr-h I invoked the mac os build twice, it seems like it's not making use of the MLIR cache?

mshr-h:
@tjingrant You means [this one](https://github.com/onnx/onnx-mlir/actions/runs/240408199) and [this one](https://github.com/onnx/onnx-mlir/actions/runs/240492662)?
My understanding is that if build was failed, `actions/cache` doesn't cache anything.
So, once build succeeded, caching will work.
I found [stackoverflow](https://stackoverflow.com/questions/60491837/saving-cache-on-job-failure-in-github-actions) post which mentions similar behavior of `actions/cache`.

It looks like testing is failed but I'm not sure why it happens.

tjingrant:
@mshr-h it seems like caching isn't fully working yet?
c.f. log here: https://github.com/onnx/onnx-mlir/pull/287/checks?check_run_id=1092414205

Seems like the cached files are not restored.

mshr-h:
@tjingrant I checked the below log and it looks like cache key collision was occured. I'll try to find the way to solve it ASAP.

[Merge branch 'master' into pr-add-macos-ci · onnx/onnx-mlir@472b531](https://github.com/onnx/onnx-mlir/runs/1092073891?check_suite_focus=true)
```
 Post cache MLIR directory
[warning]reserveCache failed: Cache already exists. Scope: refs/pull/287/merge, Key: macOS-build-cache-mlir-directory-9530df47a7667c57417ee817992e9114e00b1c649326d0aa871eae059a503898, Version: 40e68259f83fefa397b9a495fd46a09d779d54f12e7acec4bdc3dda75418c01b
```

tjingrant:
@mshr-h my guess is that when two jobs run concurrently building cached directories for the first time, they interfere with one another. If this is the cases, we can either ask folks to be aware of this pitfall, or have some mechanisms to cancel redundant jobs.

mshr-h:
@tjingrant I think you're right. I found the similar issue [here](https://github.com/actions/cache/issues/144).
I read through the discussion and there's no way to avoid this issue. 
But there are workarounds here:
- Change the cache key.
- Wait 7 days, and don't use that cache entry.

I think we have 2 choices:
- Every time we hit this issue, manually change the cache key.
- Not use cache mechanism.

Even though it takes much longer build time (15 min vs 1 hour), I suggest not to use the cache mechanism.
What do you think? @tjingrant

caoimhinuibrian:
@gongsu832 Is there a reason the s390x builds are not reporting?

caoimhinuibrian:

mshr-h:
It looks like caching is working but restoring the cache is failing.
I'll try to find out the cause of the problem.

- First CI run.
  - https://github.com/onnx/onnx-mlir/runs/1216548809?check_suite_focus=true
- Second CI run.
  - https://github.com/onnx/onnx-mlir/runs/1218038641?check_suite_focus=true

mshr-h:
@tjingrant 
The CI error was fixed. I think this should be good.
Could you review it when you have a time?

The error was occurred because '$HOME' was in the path.
After replacing '$HOME/' with '~/', the error was fixed.

Below are CI log: 
- First CI run
  - https://github.com/onnx/onnx-mlir/runs/1251430352?check_suite_focus=true
- Second CI run
  - https://github.com/onnx/onnx-mlir/runs/1252212966?check_suite_focus=true

tjingrant:

PR #288, opened 2020-09-01
initial code for handling custom ops

caoimhinuibrian:


caoimhinuibrian:
This currently doesn't do anything except issue a Warning that it found an unknown op (assumed to be a custom op). Future work will add that op to ONNX dialect level graph, and the lower it to a function call.

doru1004 (approved):
LGTM

PR #293, opened 2020-09-04
Fix wrong type for llvm::loadop

### LINKED ISSUE #291: lowering kernel dialect
I got these while converting onnx to MLIR using -EmitMLIR.
1. "krnl.packed_const"() {file_name = "/tmp/packed_const-b91861.tmp", is_le = true, size_in_bytes = 0 : i64} : () -> i64
2."krnl.global"() 
3."krnl.getref"
These are not lowered to affine dialect while doing ----convert-krnl-to-affine.

How can I lower these ?

and where can I get list of all kernel dialect operations.

Thank You

tungld:
Resolves #291


AlexandreEichenberger (approved):
LGTM, thanks @tungld

PR #294, opened 2020-09-04
Gather ONNX to Kernel Lowering

AlexandreEichenberger:
Lowering for Gather op. The best explanation that I found is from numpy.take.

```python
Ni, Nk = a.shape[:axis], a.shape[axis+1:]
Nj = indices.shape
for ii in ndindex(Ni):
    for jj in ndindex(Nj):
        for kk in ndindex(Nk):
            out[ii + jj + kk] = a[ii + (indices[jj],) + kk]
```

As I was looking at Transpose for inspiration, it felt to me that Transpose probably does not handle dynamic parameters as advertised. I added a TODO. As a matter of fact, there are no lit tests for dynamic shaped transpose, so it is untested at this stage anyway.

ONNXOps.cpp also had a missing return, I flagged that path as unreachable to get an error.


pashu123:
> Lowering for Gather op. The best explanation that I found is from numpy.take.
> 
> ```python
> Ni, Nk = a.shape[:axis], a.shape[axis+1:]
> Nj = indices.shape
> for ii in ndindex(Ni):
>     for jj in ndindex(Nj):
>         for kk in ndindex(Nk):
>             out[ii + jj + kk] = a[ii + (indices[jj],) + kk]
> ```
> 
> As I was looking at Transpose for inspiration, it felt to me that Transpose probably does not handle dynamic parameters as advertised. I added a TODO. As a matter of fact, there are no lit tests for dynamic shaped transpose, so it is untested at this stage anyway.
> 
> ONNXOps.cpp also had a missing return, I flagged that path as unreachable to get an error.

If we apply affine.load on   `a[ii + (indices[jj],) + kk]` it won't be an affine access? Right?

AlexandreEichenberger:
@pashu123 You are making a very valid point... let me change the code. Glad I added the python code to the doc... Thanks for looking at the patch and catching this flaw.

src/Conversion/ONNXToKrnl/Tensor/Gather.cpp
```diff
@@ -0,0 +1,117 @@
...
+using namespace mlir;
+
+struct ONNXGatherOpLowering : public ConversionPattern {
+  ONNXGatherOpLowering(MLIRContext *ctx)
+      : ConversionPattern(mlir::ONNXGatherOp::getOperationName(), 1, ctx) {}
+
+  LogicalResult matchAndRewrite(Operation *op, ArrayRef<Value> operands,
+      ConversionPatternRewriter &rewriter) const final {
+    ONNXGatherOpAdaptor operandAdaptor(operands);
+    ONNXGatherOp gatherOp = llvm::dyn_cast<ONNXGatherOp>(op);
```
chentong319, line 22: use cast?

src/Conversion/ONNXToKrnl/Tensor/Gather.cpp
```diff
@@ -0,0 +1,117 @@
...
+    // get input operands, shapes, and rank
+    Value data = operandAdaptor.data();
+    auto dataShape = data.getType().cast<MemRefType>().getShape();
+    int64_t dataRank = dataShape.size();
+    Value indices = operandAdaptor.indices();
+    auto indicesShape = indices.getType().cast<MemRefType>().getShape();
+    int64_t indicesRank = indicesShape.size();
+    int64_t axisIndex = gatherOp.axis().getSExtValue();
+    // get output info
+    auto outputMemRefType = convertToMemRefType(*op->result_type_begin());
```
chentong319, line 33: Question not specific to this PR: Is there a pass to translate all tensors, including inputs and outputs,  to memref?

test/backend/test.py
```diff
@@ -157,6 +157,11 @@ def supports_device(cls, device):
     "test_exp_cpu",
     "test_exp_example_cpu",
 
+    # Gather Op:
```
chentong319, line 160: not finish?

src/Conversion/ONNXToKrnl/Tensor/Gather.cpp
pashu123: May be there should be two ONNX dialects, ONNX on tensors and ONNX on buffers but unfortunately it's not there.

chentong319 (approved):
LGTM

I used this python code to verify your first test case:

import numpy as np
  
a = np.arange(6) 
a = np.reshape(a, [3,2])
index = np.array([[0, 1], [1,2]])
print(a.shape)
print(a)
print(index.shape)
print(index)
b = np.take(a, index, axis=0)
print(b.shape)
print(b)

src/Conversion/ONNXToKrnl/Tensor/Gather.cpp
AlexandreEichenberger: My understanding is that ONNX is for tensors, and krnl is for memrefs, and that the translations occurs during the lowering. If there is a need for ONNX-level transformation that can only done with memrefs, then we should think about this.

src/Conversion/ONNXToKrnl/Tensor/Gather.cpp
AlexandreEichenberger: thanks; I saw many other examples in the ONNXToKrnl, someone should probably do a pass

test/backend/test.py
AlexandreEichenberger: I had no way of testing locally, uncommented and will see if it goes through fine.

src/Conversion/ONNXToKrnl/Tensor/Gather.cpp
pashu123: I mean, if we have ONNX on memrefs then we can directly move from ONNX on memrefs to affine or linalg etc. We can even use stuff from tensorflow's lmhlo dialect. If there is a better way do suggest.

test/backend/test.py
AlexandreEichenberger: It failed to recognize the test names. Will have to investigate locally first. Disabled for now. Please approve the patch, will work on it once I get my environment to successfully run any backend tests. Make fails for me on my Mac at this time

PR #299, opened 2020-09-11
ShapeInference for SizeOp

chentong319:
Reopen PR#281 [with](https://github.com/onnx/onnx-mlir/pull/281) a clean local branch.
A simple PR of shape inference for Size. https://github.com/onnx/onnx/blob/master/docs/Operators.md#Size
The Size op always returns a scalar of i64, namely, tensor.
Our current importer of Size Op gives the output type unknown shape of i64 tensor, tensor<*xi64>. It is an alternative approach to define a special builder for SizeOp.


AlexandreEichenberger (approved):
LGTM, please fix formatting issue reporting by the checks, add one comment for the test.

PR #307, opened 2020-09-19
temporarily allow ppc64le build to fail

caoimhinuibrian:
There is some problem with the Travis server for ppc64le, so until it gets fixed, we should allow ppc64le builds to fail


tjingrant (approved):
Sounds reasonable!

PR #312, opened 2020-09-22
Use ubuntu focal as ppc64le base image.

tjingrant:
ppc64le non-nightly build restored.


caoimhinuibrian (approved):
Is there a reason not to move to focal for s390x and amd64?

tjingrant:
s390x and amd64 are already using ubuntu focal, i think.

PR #313, opened 2020-09-22
Gather: fix for negative indices

AlexandreEichenberger:
In the gather, the indices to gather are passed in a tensor; these values can be negative, in which case a negative index "starts counting" from the end of the array.

Had to add a 
```
get dimension along axis of the input data
```
before the loop, and inside the loop
```
if (index value negative) add axis dimension to index
```

Obviously I used a select to generate more compact, branch free code at the MLIR level.

Patch also enables the backend tests for gather.


AlexandreEichenberger:
WIP because I use the std.dim, which is a problem when we have maps in memrefs. Will change to krnl.dim once support by @doru1004 is added for the krnl.dim lowering for input arguments (that are constant sizes or have no maps).

AlexandreEichenberger:
Added the use of the krnl.dim after the PR enabling its use with memrefs passed as parameters.

It is now ready to be reviewed.

doru1004 (approved):
LGTM

PR #314, opened 2020-09-23
Improve shape inference of  flatten Op

chentong319:
The new implementation supports all the valid value of axis, other than only 1. Now the negative and 0 are supported.

Another relaxation is that the input tensor can be any shaped tensor, not necessarily static shaped.


AlexandreEichenberger (commented):
I believe there is an error in computing the axis for negative offsets. Do you mind checking?

Thanks

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2153,32 +2153,48 @@ LogicalResult ONNXSplitOp::inferShapes() {
...
+
+  // Negative axis is counting dimension from back
+  if (axisValue < 0)
+    axisValue = inputRank + axisValue + 1;
+
+  // Determine the size of the first dimension of output
+  uint64_t firstDim = 1;
+  for (auto i = 0; i < axisValue; i++) {
+    if (inputShape[i] == -1) {
+      firstDim = -1;
```
tjingrant, line 2178: Seems like you are assigning a negative value to an unsigned variable (firstDim is declared as uint64_t).

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2153,32 +2153,48 @@ LogicalResult ONNXSplitOp::inferShapes() {
...
-    SmallVector<int64_t, 2> dims;
-    // https://pytorch.org/docs/master/generated/torch.nn.Flatten.html
-    dims.emplace_back(inShape[0]);
-    dims.emplace_back(outDim);
-    getResult().setType(RankedTensorType::get(dims, outTy.getElementType()));
+  // Determine the size of the second dimension of output
+  uint64_t secondDim = 1;
+  for (auto i = axisValue; i < inputRank; i++) {
+    if (inputShape[i] == -1) {
+      secondDim = -1;
```
tjingrant, line 2189: Same here.

tjingrant (changes requested):

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2153,32 +2153,48 @@ LogicalResult ONNXSplitOp::inferShapes() {
...
-    auto inShape = inTy.getShape();
-    assert(inShape.size() >= 1 && "ONNXFlattenOp inShape.size() should be > 0");
-    uint64_t outDim = 1;
-    for (auto it = inShape.begin() + 1; it < inShape.end(); it++) {
-      outDim *= *it;
+  SmallVector<int64_t, 2> dims;
+
+  // Negative axis is counting dimension from back
+  if (axisValue < 0)
+    axisValue = inputRank + axisValue + 1;
```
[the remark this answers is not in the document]
chentong319: The document says: 
axis : int (default is 1)
Indicate up to which input dimensions (exclusive) should be flattened to the outer dimension of the output. The value for axis must be in the range [-r, r], where r is the rank of the input tensor. Negative value means counting dimensions from the back. When axis = 0, the shape of the output tensor is (1, (d_0 X d_1 ... d_n), where the shape of the input tensor is (d_0, d_1, ... d_n).

axis = -1 should be equal to axis = r. For your example, value of axis for the first dimension should be 1. That's why value 0 for axis is specially defined. Therefore, I think +1 is needed.

src/Dialect/ONNX/ONNXOps.cpp
chentong319: fixed

src/Dialect/ONNX/ONNXOps.cpp
chentong319: Fixed

src/Dialect/ONNX/ONNXOps.cpp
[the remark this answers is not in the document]
AlexandreEichenberger: I used the small python example from the doc
```
shape = (2, 3, 4, 5)
a = np.random.random_sample(shape).astype(np.float32)
i = 3
new_shape = (np.prod(shape[0:i]).astype(int), -1)
b = np.reshape(a, new_shape)
i_neg = -1
new_shape_neg = (np.prod(shape[0:i_neg]).astype(int), -1)
b_neg = np.reshape(a, new_shape_neg)
print(a.shape, b.shape, b_neg.shape)
```

which results in

```
(2, 3, 4, 5) (24, 5) (24, 5)
```

so axis= 3 is the same as axis=-1

-1 + rank = -1 + 4 = 3

AlexandreEichenberger (approved):
The equivalent python example seems to indicate that the "+rank" formula is correct. If you are not convinced, we can wait for the lowering and the backend test to give us the final result.

chentong319:
@AlexandreEichenberger Yes, let's check the unit test when the lowering pass is added.

AlexandreEichenberger:
Sure

PR #316, opened 2020-09-23
Fix substitution order in parse_type_str

gramalingam:
The conversion of an ONNX type string to an MLIR type string uses a sequence of substitutions provided by a dictionary. In many versions of python, the order of enumeration of entries in a dictionary is not guaranteed. This results in unexpected transformations (e.g., of float16 to F3216 when the prefix float is replaced by F32) depending on the version of python used.

Apply substitutions in decreasing order of key length to avoid these problematic cases.


chentong319 (approved):
LGTM

tjingrant:

tjingrant:

PR #317, opened 2020-09-23
Improve support for krnl.dim

doru1004:
This patch improves the range of cases supported by krnl.dim.

krnl.dim now works on non-alloc MemRef passed in as input arguments. Both static and dynamic cases are supported.

krnl.dim should not be used to get the dynamic dimension of an input MemRef with a map. This will count as a mis-use of krnl.dim and an error will occur.


AlexandreEichenberger (approved):
Please fix the comments.

Also (and I am sorry I am late on this), I find the name of this op surprising. While its true that you "disconnect" the krnl dim from the alloc, I would just call this "convert-krnl-shape-to-std" where shape includes shape and dim... If you agree, just refactor the code with that new name, if you don't mind.

AlexandreEichenberger (approved):
LGTM, thanks

PR #321, opened 2020-09-24
Lower RNNOp to Krnl dialect

tungld:
TODO:
- [X] Lowering code
- [X] End-to-end tests: passed
- [x] LIT tests

~~I have to disable `bundle-memory-pools` pass because of a bug reported here #323.~~


doru1004 (approved):
LGTM

PR #322, opened 2020-09-24
Normalize memrefs with affine_map in krnl.memcpy

imaihal:
This PR enable us to normalize memrefs with affine_map in use-case including `krnl.memcpy` .
Test case in this PR has `krnl.memcpy`. Without this PR, memrefs in this test case are not normalized because `krnl.memcpy` is not normalizable operation.
To have `krnl.memcpy` normalizable, this PR adds a trait of `MemRefsNormalizable`.

Reference: https://reviews.llvm.org/D86236


AlexandreEichenberger (approved):
LGTM, were you able to check that the resulting copying code is correct? It does not have to be a new benchmark, but was it used somewhere and it worked?

Thanks

imaihal:
@AlexandreEichenberger Thanks for your review.
I created this PR because MNIST example didn't work, but the test code was artificial one and I didn't confirm it really works(just confirmed normalization worked). 
I replaced it with more simple one. Since new one includes only alloc, dealloc, and krnl.memcpy, it should be correct.

imaihal:
@AlexandreEichenberger Thanks for your review.
I created this PR because MNIST example didn't work, but the test code was artificial one and I didn't confirm it really works(just confirmed normalization worked).
I replaced it with more simple one. Since new one includes only alloc, dealloc, and krnl.memcpy, new one should be correct.

Issue #323, opened 2020-09-25
Error when calling the pass bundle-memory-pools


tungld:
I have an MLIR program for RNN, which was generated by the PR #321. When I called `./bin/onnx-mlir-opt --bundle-memory-pools rnn.mlir`, I got the following error:
```shell
rnn.mlir:29:9: error: operand #1 does not dominate this use
        affine.store %cst, %8[] : memref<f32>
        ^
rnn.mlir:29:9: note: see current operation: "affine.store"(%cst, %14) {map = affine_map<() -> ()>} : (f32, memref<f32>) -> ()
rnn.mlir:50:15: note: operand defined here
        %15 = alloc() : memref<4xi8>
```

Without `bundle-memory-pools` the program can be lowered to LLVM bitcode and run correctly.

This is `rnn.mlir`:
```mlir
#map0 = affine_map<()[s0, s1, s2] -> (s0, s1, s2)>
#map1 = affine_map<() -> (0)>
#map2 = affine_map<() -> (1)>
#map3 = affine_map<() -> (3)>
#map4 = affine_map<() -> (4)>
#map5 = affine_map<()[s0, s1] -> (0, s0, s1)>
#map6 = affine_map<() -> ()>
#map7 = affine_map<() -> (2)>


module {
  func @main_graph(%arg0: memref<1x3x2xf32>, %arg1: memref<1x4x2xf32>, %arg2: memref<1x4x4xf32>) -> memref<1x3x4xf32> attributes {input_names = ["X", "W", "R"], output_names = ["Y"]} {
    %cst = constant 0.000000e+00 : f32
    %c0_i64 = constant 0 : i64
    %0 = alloc() : memref<1x3x4xf32>
    %1:3 = krnl.define_loops 3
    krnl.iterate(%1#0, %1#1, %1#2) with (%1#0 -> %arg3 = 0 to 1, %1#1 -> %arg4 = 0 to 3, %1#2 -> %arg5 = 0 to 4) {
      affine.store %cst, %0[symbol(%arg3), symbol(%arg4), symbol(%arg5)] : memref<1x3x4xf32>
    }
    %2 = krnl.define_loops 1
    krnl.iterate(%2) with (%2 -> %arg3 = 0 to 1) {
      %3:2 = krnl.define_loops 2
      krnl.iterate(%3#0, %3#1) with (%3#0 -> %arg4 = 0 to 3, %3#1 -> %arg5 = 0 to 4) {
[...]

PR #325, opened 2020-09-25
Support per-block bundling for static memory pools.

doru1004:
This patch ensures that static memory pools get aggregated on a per-block basis.

This also fixes issue https://github.com/onnx/onnx-mlir/issues/323


src/Transform/BundleMemoryPools.cpp
```diff
@@ -124,6 +128,9 @@ bool isBlockArgument(Block *block, Value operand) {
 KrnlGetRefOp getUnbundledGetRef(AllocOp *memPool) {
   auto parentBlock = memPool->getOperation()->getBlock();
 
+  parentBlock->dump();
+  printf("\n\n");
```
AlexandreEichenberger, line 132: may need to remove printfs?

AlexandreEichenberger (commented):
When you have a branch

```
if () A else B
```

and you add a block for handling allocs in A, do you have to change the control flow jumping from the test to A-alloc instead of A?

Also, the test that the German folks were looking into

```
block A: %1 = alloc()
if ? goto B
bloc B: %2 = alloc; free %1, return %2
block C %3 = merge (%1 on else (skip block B) path, %2 on then (through block B) path)
free %3.
```
That is a case where the allocs are escaping their blocks...

Can you just confirm that you can safely "not handle" such cases. Thanks

doru1004:
> When you have a branch
> 
> ```
> if () A else B
> ```
> 
> and you add a block for handling allocs in A, do you have to change the control flow jumping from the test to A-alloc instead of A?
> 
> Also, the test that the German folks were looking into
> 
> ```
> block A: %1 = alloc()
> if ? goto B
> bloc B: %2 = alloc; free %1, return %2
> block C %3 = merge (%1 on else (skip block B) path, %2 on then (through block B) path)
> free %3.
> ```
> 
> That is a case where the allocs are escaping their blocks...
> 
> Can you just confirm that you can safely "not handle" such cases. Thanks

Such cases are outside the scope of the bundling algorithm for now. Moreover we rely on MLIR for inserting the deallocs so if there is an error due to wrongly introduced dealloc that means the support in MLIR is not sufficient (yet). There is work by the community to improve this aspect of MLIR.

AlexandreEichenberger (approved):
LGTM, thanks also for the big new test

PR #327, opened 2020-09-28
Normalize memrefs with affine_map in a function containing krnl.getref

imaihal:
This PR enable us to normalize memrefs with affine_map in a function containing `krnl.getref` .
Without this PR, memrefs in this test case are not normalized because krnl.memcpy is not normalizable operation.
To have krnl.getref normalizable, this PR adds a trait of MemRefsNormalizable.

This PR should have been included in https://github.com/onnx/onnx-mlir/pull/322

Reference: https://reviews.llvm.org/D86236

### DIFF
--- a/test/mlir/krnl/krnl_normalization.mlir
+++ b/test/mlir/krnl/krnl_normalization.mlir
@@ -14,3 +14,17 @@ func @test_krnl_memcpy_norm(%arg0: memref<1x16384xf32>) -> memref<1x16x4x4xf32, 
   return %0 : memref<1x16x4x4xf32, #map_tile>
   // CHECK: return [[ALLOC]] : memref<1x16x1x1x32x32xf32>  
 }
+
+// CHECK-LABEL: test_getref_norm
+func @test_getref_norm() ->  () {
+  %c0_i64 = constant 0 : i64
+  %0 = alloc() : memref<1x81920xf32>
+  %1 = alloc() : memref<1x16x4x4xf32, #map_tile>
+  // CHECK: [[ALLOC:%.+]] = alloc() : memref<1x16x1x1x32x32xf32>
+  %2 = "krnl.getref"(%0, %c0_i64) : (memref<1x81920xf32>, i64) -> memref<1x16x4x4xf32>
+  // Do something using %1 and %2
+  dealloc %1: memref<1x16x4x4xf32, #map_tile>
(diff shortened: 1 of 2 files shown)


imaihal:
@AlexandreEichenberger Could you review this PR?

test/mlir/krnl/krnl_normalization.mlir
```diff
@@ -14,3 +14,17 @@ func @test_krnl_memcpy_norm(%arg0: memref<1x16384xf32>) -> memref<1x16x4x4xf32,
...
   // CHECK: return [[ALLOC]] : memref<1x16x1x1x32x32xf32>  
 }
+
+// CHECK-LABEL: test_getref_norm
+func @test_getref_norm() ->  () {
+  %c0_i64 = constant 0 : i64
+  %0 = alloc() : memref<1x81920xf32>
+  %1 = alloc() : memref<1x16x4x4xf32, #map_tile>
+  // CHECK: [[ALLOC:%.+]] = alloc() : memref<1x16x1x1x32x32xf32>
+  %2 = "krnl.getref"(%0, %c0_i64) : (memref<1x81920xf32>, i64) -> memref<1x16x4x4xf32>
```
doru1004, line 24: How does this test normalization trait for getref?

imaihal: @doru1004  All arguments and results of `getref` don't have map in the memrefs, but in current MLIR implementation, if there is an non-normalizable operation (in this case `getref`) in the func op, all of map in the func op are not normalized by `--normalize-memrefs`.  So, `MemRefsNormalizable` need to be added in the dialect to have it normalizable operation.

- This is normalized (comment out `getref`), but if `getref` is inserted, this is not normalized. 
```
func @test_getref_norm() ->  () {
  %c0_i64 = constant 0 : i64
  %0 = alloc() : memref<1x81920xf32>
  %1 = alloc() : memref<1x16x4x4xf32, #map_tile>
  // %2 = "krnl.getref"(%0, %c0_i64) : (memref<1x81920xf32>, i64) -> memref<1x16x4x4xf32>
  dealloc %1: memref<1x16x4x4xf32, #map_tile>
  dealloc %0: memref<1x81920xf32>
  return
}
```

doru1004: Got it! Thanks for the explanation! This makes sense now!

doru1004 (approved):
LGTM

AlexandreEichenberger (approved):
LGTM

PR #329, opened 2020-09-28
Mac instructions for backend check

AlexandreEichenberger:
steps I had to perform to instal third_party/onnx needed by the check-onnx-backend tests


tjingrant:
I'm a bit surprised that you need to manually install pybind, maybe I have it already installed as a dependency of other packages; regardless, it shouldn't hurt putting this extra information under FAQ. Thanks for the explanation!

tjingrant (approved):
LGTM!

chentong319 (commented):
These instructions, the pybind11_DIR and the change for cmake, are needed for PowerPC machines, too. Install onnx is also needed to run gen_onnx_mlir.py. 
Should we put these instructions with a separately?

chentong319:
I have a particular change to build on Mac. Since the default python on Mac is 2.7. I installed the new python. Even after I added the path to the new python and define alias of python, I still need to set PYTHON_EXECUTABLE in my make file. Otherwise check package for PyInterp will fail. Is there a simple solution to be included in the document?

AlexandreEichenberger:
@chentong319 Thanks for the input; I modified the text to incorporate your input.

Issue #326, opened 2020-09-25
Ensure dynamic memory pools are aggregated on a per block basis.


doru1004:
Ensure dynamic memory pools are aggregated on a per block basis.

PR #330, opened 2020-09-28
Enable bundling of dynamic memory pools on a block basis.

doru1004:
This patch enables the aggregation of individual dynamic memory pools on a block basis.

This fixes issue https://github.com/onnx/onnx-mlir/issues/326


AlexandreEichenberger (approved):
LGTM

PR #331, opened 2020-09-30
Lower Less op to Krnl dialect

tungld:
- Lower less op to Krnl dialect
- Fix a bug in type conversion between LLVM and ONNX


src/Runtime/PyExecutionSession.cpp
```diff
@@ -56,13 +56,13 @@ std::vector<py::array> PyExecutionSession::pyRun(
     py::dtype dtype;
     if (dynMemRef->onnx_dtype == onnx::TensorProto::FLOAT)
       dtype = py::dtype("float32");
-    else if (dynMemRef->onnx_dtype = onnx::TensorProto::UINT8)
+    else if (dynMemRef->onnx_dtype == onnx::TensorProto::UINT8)
```
AlexandreEichenberger, line 59: Good catch

AlexandreEichenberger (approved):
LGTM

PR #332, opened 2020-10-01
Lower Flatten to Krnl

chentong319:
Flatten Op is intended to change the shape of the input tensor into a 2D output tensor, with first axis() dimensions for the first dimension and the rest for the second dimension. 
A loop nested is generated with the shape of the input() of Flatten Op. The index of the output tensor is calculated with the index and dimension size of the input tensor.
test cases are added to check-onnx-lit and all the node tests for Flatten Op are enable.


chentong319:

AlexandreEichenberger (approved):
LGTM, thanks

Issue #184, opened 2020-06-24
Help on ExecutionSession


feiyuvl:
Is there any example to run the exeuction session ?

PR #334, opened 2020-10-01
Add pyruntime docs

gperrotta:
Hello,

I am studying MLIR and ML compilers in my undergraduate research project and was trying to run inference with models compiled with onnx-mlir, so I studied your code base a little and figured how to use the runtime python bindings.

Here is the how-to documentation I came up with and would be glad if it helps somebody else.

Issue #184 is related to this.


tjingrant:

tjingrant (approved):
LGTM!

PR #336, opened 2020-10-05
Shape inference for QLinearConv, QLinearMatMul, & GlobalAveragePool

thebyrd:
Hello, 

I was trying to get `onnx-mlir --EmitONNXIR resnet50v2.quantized.onnx` to work and it required adding shape inference for QLinearConv & QLinearMatMul & GlobalAveragePool. It also required updating the TransformValueToONNXData template for int8_t and uint8_t because they use int32_t for the underlying data storage.
[resnet50v2.quantized.onnx.zip](https://github.com/onnx/onnx-mlir/files/5331421/resnet50v2.quantized.onnx.zip)


src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -1693,6 +1930,31 @@ LogicalResult ONNXAveragePoolOp::inferShapes() {
...
+  if (!X().getType().isa<RankedTensorType>())
+    return emitError("Input tensor not ranked");
+
+  auto builder = mlir::Builder(getContext());
+
+  // Get shape of input.
+  auto xTy = X().getType().cast<RankedTensorType>();
+
+  // Assumes NCHW format.
+  auto xShape = xTy.getShape();
```
tjingrant, line 1948: We should probably assert that xTy is of rank 4 here.

tjingrant (approved):
LGTM, but it seems like some code duplication can be easily avoided.
For instance, the case-insensitive diff between shape inference code for `ONNXMatMulOp` and `ONNXQLinearMatMulOp` is literally only in the function name:

```diff
1c1
< LogicalResult ONNXMatMulOp::inferShapes() {
---
> LogicalResult ONNXQLinearMatMulOp::inferShapes() {
```
And similarly, the diff between shape inference code for `ONNXQLinearConvOp` and `ONNXConvOp` is also minimal:
```diff
1c1
< LogicalResult ONNXQLinearConvOp::inferShapes() {
---
> LogicalResult ONNXConvOp::inferShapes() {
30c30
<   int64_t group = ONNXQLinearConvOp::group();
---
>   int64_t group = ONNXConvOp::group();
34c34,35
<     groupAttr(builder.getI64IntegerAttr(group));
---
>     groupAttr(IntegerAttr::get(builder.getIntegerType(64, /*isSigned=*/true),
>         APInt(64, group, /*isSigned=*/true)));
38,39c39,42
<       xShape[1] != (weightShape[1] * group))
<     return emitError("Channel dimension mismatch");
---
>       xShape[1] != (weightShape[1] * group)) {
>     return emitOpError("Channel dimension mismatch")
>            << xTy << " " << weightTy << " " << group;
>   }
```

Seems like we can fairly easily extract some common shape inference patterns between quantized and non-quantized conv, matmul?

tungld:
Another PR #339 has a template for Global pooling ops' shape inference. Could we remove GlobalAveragePool here to avoid duplication and focus on QLinear ops only. Thanks!

tjingrant:
@tungld @thebyrd it is unfortunate that we ran into situations where engineering efforts are duplicated; especially when it seems like everyone is properly following the rules/best practices. @thebyrd let me do this: let me merge this patch right away, so that your contribution will get recorded & reflected in the project history, even though #339 may overwrite a part of your code contribution later.

I will follow up with fixes to reduce code duplication unless you feel strongly about doing it yourself.

tjingrant:

PR #340, opened 2020-10-07
Find installed tools

gongsu832:
Find llc and opt from the directory where onnx-mlir is run to account for the case when both onnx-mlir and llvm-project are installed system wide (typically under /usr/local/...).


caoimhinuibrian (commented):
The general concept seems fine, we shouldn't be forcing users to have an LLVM build environment, but I have a few concerns.

I'm not sure about forcing CMAKE_INSTALL_PREFIX to be the same as LLVM_PROJ_BUILD. LLVM_PROJ_BUILD is where the version of MLIR and LLVM that we use as source in onnx-mlir lives, not where a system wide LLVM is installed. Why do we look for llc and opt in the same place that onnx-mlir is executed from? Shouldn't we just look for them on the PATH? If they aren't on the path, then use our default locations.

gongsu832:
@caoimhinuibrian Searching PATH is a shell concept since the primary job of a shell is to run commands from different places. onnx-mlir can certainly do that too but it really is not a shell. It just happens to rely on a couple of commands from llvm-project. So it is in general not worth the trouble searching through a bunch of directories since most binaries produced by cmake will be installed under CMAKE_INSTALL_PREFIX which is typically /usr/local/bin. And since onnx-mlir logically depends on llvm-project so I think it's reasonable to assume that binaries for onnx-mlir and llvm-project are typically  installed under the same directory.

caoimhinuibrian:
@gongsu832 This is a further manifestation of the problem you started looking into - making an image that has just the stuff an end user will want. Currently our whole build process is just aimed at people who are developing onnx-mlir, so we assume that they have all the pieces needed to run the resulting code in the expected places. The LLVM tools like llc and opt are not usually provided in a binary install of LLVM (like Clang). If we want to use them in onnx-mlir then for a binary install (end user) we should package them with all the other stuff we need and provide an rpm or something like that. Alternatively, we could incorporate the bits we actually want in our driver rather than using these tools.

For now, I guess I'm OK with your change.

caoimhinuibrian (approved):
LGTM for now, but see my previous comment.

gongsu832:
@caoimhinuibrian I agree that we need to have some way to package up all the llvm-project binaries that we depend on together with onnx-mlir binaries. It appears to me that llvm-project specifically delegates this task to the "frontend" since llvm-project itself doesn't even have an "install" target. I don't know what the standard practice is here but maybe we should install llc and opt as part of "make install" for onnx-mlir if that's what's typically done? Maybe @tjingrant have some inputs?

If we take this route, then obviously we don't need to force our CMAKE_INSTALL_PREFIX to be the same as that of llvm-project (note this is NOT LLVM_PROJ_BUILD) since llc and opt will always be installed alongside onnx-mlir.

PR #341, opened 2020-10-08
Fix AzurePipeline Failure: disable conda-forge channel for conda install

jcwchen:
**Description**
To fix the conda-forge issue on AzurePipeline, simply use the package from conda instead of conda-forge.

**Motivation**
Windows-CI is failing on AzurePipeline. conda-forge channel is broken on AzurePipeline.


jcwchen:
It seems that this bug in Azure Pipeline has been fixed itself... Closing it now because it is not necessary anymore.

PR #346, opened 2020-10-14
Optimize static memory pools

doru1004:
This patch enables the optimization of the static memory pools emitted by the bundling pass.

Two `krnl.getref` operations can share the same slot in the memory pool if several conditions are met:
1. Same memory footprint: this includes the case where the MemRefs they produce have different shapes. Example: `memref<3x4xf32>` and `memref<2x6x4xi8>` occupy the same memory.
2. Not in the same `krnl.memcpy`: there is no `krnl.memcpy` which uses both. A `krnl.memcpy` is equivalent to a load from the source of the memcpy followed by a store to the destination of the memcpy.
3. Mutually disjoint load-store chains: the stored values in one of the resulting MemRefs must not have been computed using loads from the other resulting MemRefs. When a new `krnl.getref` is added to the same memory slot, it needs to be mutually disjoint with each of the other MemRefs already sharing that memory slot.
4. Live ranges must not intersect: a live range is the chain of instructions between the first and last load/store/memcpy instruction that uses a `krnl.getref`. This covers the case where `krnl.getref` uses are not linked by a load-store chain but are live at the same time.
5. Live range endpoints cannot appear in the same loop nest: live ranges as defined above may not intersect but instead they may start or end under the same loop (in our case `krnl.iterate`). To mitigate this case we ensure that the endpoints of the two `krnl.getref` operations do not share the same outermost loop.

Example:

Considering the following static memory pool usage:
```
    %1 = alloc() : memref<2000xi8>
    %2 = "krnl.getref"(%1, %c1600_i64) : (memref<2000xi8>, i64) -> memref<10x10xf32>
    %3 = "krnl.getref"(%1, %c1200_i64) : (memref<2000xi8>, i64) -> memref<10x10xf32>
    %4 = "krnl.getref"(%1, %c800_i64) : (memref<2000xi8>, i64) -> memref<10x10xf32>
    %5 = "krnl.getref"(%1, %c400_i64) : (memref<2000xi8>, i64) -> memref<10x10xf32>
    %6 = "krnl.getref"(%1, %c0_i64) : (memref<2000xi8>, i64) -> memref<10x10xf32>
```

Example of MemRefs which do not satisfy this rule (MemRefs `%4` and `%5`, the values of MemRef `%4` are computed using values from MemRef `%5`):
```
        %14 = affine.load %arg0[symbol(%arg2), symbol(%arg4)] : memref<10x10xf32>
        %15 = affine.load %5[symbol(%arg4), symbol(%arg3)] : memref<10x10xf32>
        %16 = affine.load %4[symbol(%arg2), symbol(%arg3)] : memref<10x10xf32>
        %17 = mulf %14, %15 : f32
        %18 = addf %16, %17 : f32
        affine.store %18, %4[symbol(%arg2), symbol(%arg3)] : memref<10x10xf32>
```

Example of MemRefs which satisfy the second rule (MemRefs `%4` and `%6`):
```
    %7:2 = krnl.define_loops 2
    krnl.iterate(%7#0, %7#1) with (%7#0 -> %arg2 = 0 to 10, %7#1 -> %arg3 = 0 to 10) {
      affine.store %cst, %6[symbol(%arg2), symbol(%arg3)] : memref<10x10xf32>
      %13 = krnl.define_loops 1
      krnl.iterate(%13) with (%13 -> %arg4 = 0 to 10) {
        %14 = affine.load %arg0[symbol(%arg2), symbol(%arg4)] : memref<10x10xf32>
        %15 = affine.load %arg1[symbol(%arg4), symbol(%arg3)] : memref<10x10xf32>
        %16 = affine.load %6[symbol(%arg2), symbol(%arg3)] : memref<10x10xf32>
        %17 = mulf %14, %15 : f32
        %18 = addf %16, %17 : f32
        affine.store %18, %6[symbol(%arg2), symbol(%arg3)] : memref<10x10xf32>
      }
    }
    %8:2 = krnl.define_loops 2
    krnl.iterate(%8#0, %8#1) with (%8#0 -> %arg2 = 0 to 10, %8#1 -> %arg3 = 0 to 10) {
      %13 = affine.load %arg0[symbol(%arg2), symbol(%arg3)] : memref<10x10xf32>
      %14 = affine.load %6[symbol(%arg2), symbol(%arg3)] : memref<10x10xf32>
      %15 = addf %13, %14 : f32
      affine.store %15, %5[symbol(%arg2), symbol(%arg3)] : memref<10x10xf32>
    }
    %9:2 = krnl.define_loops 2
    krnl.iterate(%9#0, %9#1) with (%9#0 -> %arg2 = 0 to 10, %9#1 -> %arg3 = 0 to 10) {
      affine.store %cst, %4[symbol(%arg2), symbol(%arg3)] : memref<10x10xf32>
      %13 = krnl.define_loops 1
      krnl.iterate(%13) with (%13 -> %arg4 = 0 to 10) {
        %14 = affine.load %arg0[symbol(%arg2), symbol(%arg4)] : memref<10x10xf32>
        %15 = affine.load %5[symbol(%arg4), symbol(%arg3)] : memref<10x10xf32>
        %16 = affine.load %4[symbol(%arg2), symbol(%arg3)] : memref<10x10xf32>
        %17 = mulf %14, %15 : f32
        %18 = addf %16, %17 : f32
        affine.store %18, %4[symbol(%arg2), symbol(%arg3)] : memref<10x10xf32>
      }
    }
```

The number of `krnl.getref` operations must be preserved because each one of the getref operations sharing a memory pool slot may return a MemRef with a different shape (dimensions or even rank). When collapsing getrefs into the same memory slot, more than one getref may be collapsed at a time.


src/Transform/OptimizeMemoryPools.cpp
```diff
@@ -0,0 +1,905 @@
...
+        stores.emplace_back(op);
+  });
+
+  // The list contains at least one use.
+  return stores;
+}
+
+/// Returns a list of distinct krnl.getref operations in the current
+/// block that use the memory pool.
+SmallVector<KrnlGetRefOp, 4> getAllDistinctGetRefsForAlloc(AllocOp *allocOp) {
```
AlexandreEichenberger, line 113: The preferred class is SmallVectorImpl, see https://llvm.org/docs/ProgrammersManual.html#llvm-adt-smallvector-h

It say that the Impl version is not really an implementation in the traditional sense and is compatible with the SmallVect... 

Though in this case, you return it... so it might be fine. Still, look at the page and make an opinion for yourself... It applies equally to the other usage of this same pattern below.

src/Transform/OptimizeMemoryPools.cpp
```diff
@@ -0,0 +1,905 @@
...
+
+  return alloc;
+}
+
+/// Get the number of GetRef ops associated with this AllocOp.
+int64_t getAllocGetRefNum(AllocOp *allocOp) {
+  auto parentBlock = allocOp->getOperation()->getBlock();
+
+  int64_t numGetRefs = 0;
+  parentBlock->walk([&numGetRefs, allocOp](KrnlGetRefOp op) {
```
AlexandreEichenberger, line 50: Nice use of walks....

src/Transform/OptimizeMemoryPools.cpp
```diff
@@ -0,0 +1,905 @@
...
+///    b = f(a)
+///    store b secondGetRef[]
+///
+/// The check must be done both ways:
+///
+///    a = load secondGetRef[]
+///    b = f(a)
+///    store b firstGetRef[]
+///
+bool getRefUsesAreMutuallyDisjoint(SmallVector<KrnlGetRefOp, 4> firstGetRefList,
```
AlexandreEichenberger, line 270: I think here you want to use the Impl version, and you want to pass as reference.

src/Transform/OptimizeMemoryPools.cpp
```diff
@@ -0,0 +1,905 @@
...
+
+  for (auto getRef : firstGetRefList) {
+    if (!getRefUsesAreDisjoint(secondGetRefList, getRef)) {
+      return false;
+    }
+  }
+  return true;
+}
+
+/// Op is a LoadOp or AffineLoadOp.
```
AlexandreEichenberger, line 286: I am surprised that there is no query like this in MLIR. A cursory search could not find one. I wonder if we are not doing ourselves a disservice by burying these code deep into a .cpp. I would suggest to do a src/Support directory with a couple of system-wide useful queries like these. What do you think?

src/Transform/OptimizeMemoryPools.cpp
```diff
@@ -0,0 +1,905 @@
...
+  while (!llvm::dyn_cast_or_null<FuncOp>(parentBlockOp)) {
+    if (llvm::dyn_cast_or_null<KrnlIterateOp>(parentBlockOp))
+      outermostLoop = parentBlockOp;
+    parentBlockOp = parentBlockOp->getBlock()->getParentOp();
+  }
+
+  return outermostLoop;
+}
+
+/// Returns true if two operations share the same outermost krnl.iterate.
```
AlexandreEichenberger, line 454: General question: if you have something like this

for(i...) {
   a = alloc(100), b = alloc(100), c = alloc(100)
   for(j) define (a)
   for(j) define(b) use(a)
   for(j) define(c) use(b)
}

would a and c be able to use the same memory? Asking this as I see you query about shared outer loops.

src/Transform/OptimizeMemoryPools.cpp
```diff
@@ -0,0 +1,905 @@
...
+
+    // Static memory pool type must be byte.
+    if (getMemRefEltSizeInBytes(memPoolType) != 1)
+      return failure();
+
+    // Rank of the static memory pool must be 1.
+    if (memPoolShape.size() != 1)
+      return failure();
+
+    // Determine if the static memory pool is bundled i.e. participates in more
```
AlexandreEichenberger, line 599: if a, b, and c have mutually disjoint liveranges, could they share the same memory? In principle they can.. say you merge a&b. Then the liveranges becomes the union of their respective ranges.

This is where doing graph coloring is useful. You determine the interferences of the liveranges, and then you can merge things without "thinking" about the original loads/store, because all info is incapsulated into the graph.

src/Transform/OptimizeMemoryPools.cpp
```diff
@@ -0,0 +1,905 @@
...
+  }
+
+  // If all getRef live ranges are independent then no intersection exists.
+  return false;
+}
+
+//===----------------------------------------------------------------------===//
+// Rewrite patterns.
+//===----------------------------------------------------------------------===//
+
```
AlexandreEichenberger, line 569: Your comments have been very useful. Do you think you could add one here too for each patterns? Say at high level what they do?

I say this is because I see two transforms... and in the future, maybe more of them will come too.

AlexandreEichenberger (commented):
Left a couple of questions.  At a general level, can you also comment on why you requires the sizes to be the same for merging? I can see how one would first try to merge all identical ones, and then progressively the ones that are most similar in sizes (e.g. 1000 would fit in 1001 bytes).

src/Transform/OptimizeMemoryPools.cpp
doru1004: They cannot share the same memory because they have a common for as their ancestor i.e. the outermost for, `for(i...)`

doru1004: > Left a couple of questions. At a general level, can you also comment on why you requires the sizes to be the same for merging? I can see how one would first try to merge all identical ones, and then progressively the ones that are most similar in sizes (e.g. 1000 would fit in 1001 bytes).

Very good point. For now I have only tried precise sizes for simplicity. This way I can compare the new incoming candidate size with any of the existing getRefs sharing that slot instead of having to keep track of the maximum MemRef size for a given slot.

doru1004:
> Left a couple of questions. At a general level, can you also comment on why you requires the sizes to be the same for merging? I can see how one would first try to merge all identical ones, and then progressively the ones that are most similar in sizes (e.g. 1000 would fit in 1001 bytes).

Very good point. For now I have only tried precise sizes for simplicity. This way I can compare the new incoming candidate size with any of the existing getRefs sharing that slot instead of having to keep track of the maximum MemRef size for a given slot.

src/Transform/OptimizeMemoryPools.cpp
doru1004: Yes they can and will share the same slot provided they satisfy all the conditions.

In your graph coloring example they would all 3 have the same color meaning they can be merged.

The current algorithm looks at how individual live ranges compare with each other.

Let's say the order in your source code is:
a
b
c

And the pattern rewriter has reached "a".

It will now go through all the candidates i.e. b and c.
When b is analyzed the live range of b (LR(b)) is compared with LR(a). if they are disjoint (assuming all other conditions are met) then b is a valid reuser and it is saved as a valid reuser.

When we then analyze candidate "c" then LR(c) is compared against the LR(a) if it is disjoint it is then compared against the LR(b) (because b is in the list of valid reusers) and if it is disjoint with LR(b) also then and only then is it added as a valid reuser.

In this way it doesn't matter if the order in the code is:

a
b
c

or 

a
c
b

"c" will still be identified as a valid reuser of the same slot as a & b.

Once all candidates have been analyzed and hence all valid reusers collected, the actual replacements are done. At that point the code becomes:
a
a
a

It's only at this point that LR(a) increases.

src/Transform/OptimizeMemoryPools.cpp
AlexandreEichenberger: Maybe my example was too sparse:
```
for(i...) {
  a = alloc(100), b = alloc(100), c = alloc(100) 
  for(j) define (a)
  for(j) define(b) use(a)
  for(j) define(c) use(b)
  free a, b, c
}
```

Since the liveranges are confined within the I loop, there is no reason to disable this... I am thinking that the LOOP operator will introduce such things. It would be nice to have it. If you don't want to do this now, please note in the code where this constraint is expressed with a TODO.

doru1004: I see your point now. The code to handle that case is the same as the code we have today. I have temporarily disabled the optimization within sub-blocks but I can re-enable that and this case should then be covered. Slight change to the "same outermost loop" definition: the outermost loop needs to be the one in the current block rather than one at top level block i.e. the analysis should stop at the for j loops when looking to optimize this block. This will then result in a, b and c having no common outermost loop which would lead to them being able to be safely reused. The pre-condition is that the alloc and the dealloc are not hoisted outside the for i loop. If you do that, then they will be handled at the top block level.

Perhaps hoisting needs to happen after this pass.

If you hoist the allocs before this pass, the allocs would be discovered at the top level (even without enabling the sub-block level optimization). They would then be checked and pass pretty much all conditions except the outermost for loop condition. I'm not sure I see a straightforward relaxation of this condition without some heavy analysis of the loop nest.

AlexandreEichenberger: > If you hoist the allocs before this pass, the allocs would be discovered at the top level (even without enabling the sub-block level optimization).

I think its the opposite; as there are here, the liveranges are easily contained inside the outer loop... easier to analyze locally.

doru1004: Nope, it's not the opposite, if you hoist the alloc, it will be considered for optimization in the context that contains the outermost loop i.e. the main block of the model function.

doru1004: I've added a TODO to enable to optimization for sub-blocks.

src/Transform/OptimizeMemoryPools.cpp
doru1004: Done

src/Transform/OptimizeMemoryPools.cpp
doru1004: Done

doru1004:
@AlexandreEichenberger I have gathered together in KrnlSupport.cpp/hpp all the helper functions that are used at the level of abstraction of the Krnl Dialect. As a consequence passes in Transforms no longer have a dependency on the ONNXToKrnlCommon.cpp/hpp files which was not desirable. This means we can drop the ONNX dialect dependency that we had when working on krnl transformations due to the use of the ONNXToKrnlCommon.cpp/hpp files.

src/Support/KrnlSupport.hpp
```diff
@@ -0,0 +1,118 @@
+//====---------- KrnlSupport.hpp - Krnl-level support functions -----------===//
```
AlexandreEichenberger, line 1: Very nice, this should promote reuse... very clean.

AlexandreEichenberger (approved):
LGTM

PR #348, opened 2020-10-15
Updating Onnx to v1.7.0

lszinv:


tjingrant:
@lszinv I think you need to re-generate ONNX Dialect using the new ONNX spec, as documented here: http://onnx.ai/onnx-mlir//ImportONNXDefs.html .

lszinv:
I've updated the new tablegen file where seemingly the only new changes are in the docstrings. This PR is still failing the circle CI test to check that the documentation is out uouf date. 

Did I misunderstand what needs to be updated and am missing  other changes?

chentong319:
The log showed the error is in the document. Did you run `make onnx-mlir-doc' to copy the build/docs/Dialect/onnx.md to docs/Dialect?
By the way, this change is intended only to update onnx to 1.7, not really to use the new version of operations if there is any. Right?

lszinv:
> The log showed the error is in the document. Did you run `make onnx-mlir-doc' and copy the build/docs/Dialect/onnx.md to docs/Dialect?
> By the way, this change is intended only to update onnx to 1.7, not really to use the new version of operations if there is any. Right?

the main motivation for updating this was actually mainly because when trying to build the project off the master branch, it fails on the `OMInstallOnnx` target. Locally, I updated the onnx commit to be 1.7 and was then able to compile, so I figured it was useful to push upstream. 

So you're correct. This is only meant to update the ONNX version, and not to use any new versions of operations

chentong319:

Xatter (approved):
👍

tjingrant:

tjingrant (approved):
LGTM, seems like it does introduce any changes to operator spec.

PR #354, opened 2020-10-16
add dummy function for shape inference

chentong319:

Add shape inference function for all operations if they are not implemented.
Error message will be generated if such operation is encountered at runtime.

The purpose is to release other users from the burden of installing the third-party/onnx, understand, modify and run the utils/gen_onnx_mlir.py.  Now  implementation of shape inference can be simply added into ONNXOps.cpp.


AlexandreEichenberger (approved):
LGTM; you may want to add in the comment section at the top that the minimum for supporting an op is 1) implement shape inference, 2) implement lowering from ONNX to Krnl, and add lit tests for 1 & 2 as well as unlock backend tests when suitable.

PR #355, opened 2020-10-17
Support Loop Op

tjingrant:
ONNX Loop Op
------

**Demo:**

An onnx model that computes the the sum from 0 to N:
```mlir
  func @main_graph(%arg0: tensor<i64>, %arg1: tensor<i1>, %arg2: tensor<1xi64>) -> tensor<1xi64> attributes {input_names = ["trip_count", "cond", "y"], output_names = ["res_y"]} {
    %0 = "onnx.Loop"(%arg0, %arg1, %arg2) {body = @loop_body} : (tensor<i64>, tensor<i1>, tensor<1xi64>) -> tensor<1xi64>
    return %0 : tensor<1xi64>
  }
  "onnx.EntryPoint"() {func = @main_graph, numInputs = 3 : i32, numOutputs = 1 : i32} : () -> ()
  func @loop_body(%arg0: tensor<i64>, %arg1: tensor<i1>, %arg2: tensor<1xi64>) -> (tensor<i1>, tensor<1xi64>) attributes {input_names = ["iter_count", "cond_in", "y_in"], output_names = ["cond_out", "y_out"]} {
    %0 = "onnx.Identity"(%arg1) : (tensor<i1>) -> tensor<i1>
    %1 = "onnx.Add"(%arg2, %arg0) : (tensor<1xi64>, tensor<i64>) -> tensor<1xi64>
    return %0, %1 : tensor<i1>, tensor<1xi64>
  }
```

Lowered Krnl Dialect IR:
```mlir
#map0 = affine_map<()[s0] -> (s0)>
#map1 = affine_map<() -> (0)>
#map2 = affine_map<() -> (1)>
#map3 = affine_map<() -> ()>
module {
  func @main_graph(%arg0: memref<i64>, %arg1: memref<i1>, %arg2: memref<1xi64>) -> memref<1xi64> attributes {input_names = ["trip_count", "cond", "y"], output_names = ["res_y"]} {
    // Allocate memory for storing current y value, this memory is used to host the current values of 
    // loop-carried dependencies, in this case it's just y. 
    // 
    // We store the intermediate value of loop carried dependencies at this function-scope
    // memref because this memref will be accessible for all iterations (as opposed to memref within 
    // the for loop scope which is only visible within a single iteration).
    %0 = alloc() : memref<1xi64>
    %1 = krnl.define_loops 1
    // Copy the initial value of y to the aforementioned function-scope buffer.
    krnl.iterate(%1) with (%1 -> %arg3 = 0 to 1) {
      %5 = affine.load %arg2[symbol(%arg3)] : memref<1xi64>
      affine.store %5, %0[symbol(%arg3)] : memref<1xi64>
    }

    // Load maximum trip count, use that as the upper bound to build a for-loop.
    %2 = krnl.define_loops 1
    %3 = load %arg0[] : memref<i64>
    %4 = index_cast %3 : i64 to index
    krnl.iterate(%2) with (%2 -> %arg3 = 0 to %4) {
      // Loop body expects iteration count to be passed in as a scalar memref, 
      // thus convert induction var to a scalar memref.
      %5 = index_cast %arg3 : index to i64
      %6 = alloc() : memref<i64>
      store %5, %6[] : memref<i64>
      // Call loop body function (repeatedly) with current iteration number, 
      // termination condition (not used), current values of y.
      %7:2 = call @loop_body(%6, %arg1, %0) : (memref<i64>, memref<i1>, memref<1xi64>) -> (memref<i1>, memref<1xi64>)
      // Copy the values produced by loop body to our function-level memref so that this 
      // result can be used as the starting point for next iteration.
      %8 = krnl.define_loops 1
      krnl.iterate(%8) with (%8 -> %arg4 = 0 to 1) {
        %9 = affine.load %7#1[symbol(%arg4)] : memref<1xi64>
        affine.store %9, %0[symbol(%arg4)] : memref<1xi64>
      }
      dealloc %7#1 : memref<1xi64>
    }
    return %0 : memref<1xi64>
  }
  "krnl.entry_point"() {func = @main_graph, numInputs = 3 : i32, numOutputs = 1 : i32} : () -> ()
  func @loop_body(%arg0: memref<i64>, %arg1: memref<i1>, %arg2: memref<1xi64>) -> (memref<i1>, memref<1xi64>) attributes {input_names = ["iter_count", "cond_in", "y_in"], output_names = ["cond_out", "y_out"]} {
    %0 = alloc() : memref<1xi64>
    %1 = krnl.define_loops 1
    krnl.iterate(%1) with (%1 -> %arg3 = 0 to 1) {
      %2 = affine.load %arg2[0] : memref<1xi64>
      %3 = affine.load %arg0[] : memref<i64>
      %4 = addi %2, %3 : i64
      affine.store %4, %0[symbol(%arg3)] : memref<1xi64>
    }
    return %arg1, %0 : memref<i1>, memref<1xi64>
  }
}
```

I'll probably implement a few more advanced features in a separate patch:
- Support early termination.
- Plan/manage memory optimally for intermediate values produced by loop body.
  - Deallocating them is difficult - we cannot dealloc everything produced by the loop body function because there can be aliases due to returning the same memref.
  - How can we plan/optimize memory usage such that we can have loop body write to the memref shared across iterations directly? Currently a copy is used to copy content from memrefs within loop scope to memrefs outside loop scope (i.e., sitting at func scope).
- Verbosity of numerical tests is getting out of hand, need to simplify.

Minor features I want to do in this patch but haven't already done so:
- [ ] Do not hard-code the entry point function name (main_graph); should make a constant out of it in some header file and refer to the constant variable.
- [x] Fix s390x build issue, likely because i'm passing reference to stack variables as data pointer, causing deallocation to fail.


src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -447,40 +474,47 @@ class FrontendGenImpl {
...
-    if (inVals[4] == nullptr) {
-      SmallVector<int64_t, 1> vals(startsDim, 1);
-      auto constantDenseAttribute =
-          mlir::DenseElementsAttr::get(tensorType, llvm::makeArrayRef(vals));
-      auto constantOp = builder_.create<mlir::ONNXConstantOp>(
-          UnknownLoc(), mlir::Attribute(), constantDenseAttribute);
-      mlir::Value constantResult = constantOp.output();
-      inVals[4] = constantResult;
-    }
+    inVals[3] = none();
```
chentong319, line 480: Where did the previous code go? Is the default value for none input no longer needed?

tjingrant: I moved it under shape inference. I don't think any shape-aware computation should happen at the time we import onnx graph into ONNX Dialect.

AlexandreEichenberger:
To help reviewing, can you summarize if you were able to use Doru's suggestion to change the function signature and then lower, or you had to keep with your original approach, thanks.

tjingrant:
@AlexandreEichenberger I used my original approach and while I was able to experiment with Doru's suggested method, as I explained earlier, it feels like the right thing to do to associate a unique set of types with each level of dialect whenever possible. For this reason, since both methods seem to involve some extent of hackiness (by that I mean code specifically written to workaround oddity of MLIR), the original approach seems more appropriate.

doru1004:
Can we have the call to the loop body as a function with no return value?

Instead of this:
```
%7:2 = call @loop_body(%6, %arg1, %0) : (memref<i64>, memref<i1>, memref<1xi64>) -> (memref<i1>, memref<1xi64>)
```

We could have this:
```
call @loop_body(%6, %arg1, %0) : (memref<i64>, memref<i1>, memref<1xi64>, %return_val, %return_val) -> (memref<i1>, memref<1xi64>)
```

The fact that we return the value means we don't have an exposed alloc in the caller which leads to %7 not being included in any memory optimizations in the caller. %7 will only be considered for optimization inside the body of the loop_body function if any such opportunities exist.

tjingrant:
@doru1004 thanks for taking a close look at the IR example! I think your point makes a lot of sense. Since this optimization is not specific to the Loop operation (and I imagine similar operations who make calls to sub-computation graphs will all benefit from optimization patterns like this), I think this feature may be well-suited for a separate patch.

doru1004:
@tjingrant I understand your point but we have full control over the emission of this code I don't see why we shouldn't emit it in the form that we want from the start. We should emit the function call in this form for all other operations which require loops.

Moving the allocs outside the function later on will require re-emitting the function which is something we are already doing here.

In my view this is a problem that's easier to do right from the start rather than repair it up later on.

AlexandreEichenberger (approved):
I second @doru1004 's opinion that if we could pass the address of the parameters so that we don't need to alloc/copy/free, that would be good. I am happy if you believe that it would be just as efficient to do it as a separate pass. 

It would be interesting to see if @caoimhinuibrian attempt to do custom function ops, and/or other graph importing features would indeed benefit from this pattern as well.

So let's put this one down with the intent to change it.

Another question: is the condition argument present in our test benchmarks? If so, we should probably add support for it too relatively soon.

tjingrant:
> So let's put this one down with the intent to change it.

For sure.

> Another question: is the condition argument present in our test benchmarks? If so, we should probably add support for it too relatively soon.

I have not seen the condition argument being used (modified); full support of `condition` is tricky, mostly because this would mean allowing for early termination, which comes with two issues:
- We need to adjust the size of the scan output, since we allocated for the amount of scan outputs based on the maximum trip count (without early termination); early termination means that we won't be able to fill every entry in the scan output.
- It's unclear to me what's the best way to lower it to affine/std dialect, the semantics is essentially this:
```
for (int i=0; i<max_trip_count && cond; i))
```
Supporting early termination improperly could obfuscate loop optimization opportunities, and I need to spend more time thinking about how to best support it while enabling a bunch of possible optimizations we may need down the road (e.g., for RNN, we may want to experiment with wavefront parallelism).

src/Conversion/KrnlToLLVM/KrnlToLLVM.cpp
```diff
@@ -327,9 +327,15 @@ class KrnlGlobalOpLowering : public ConvertToLLVMPattern {
     auto constantElementType =
         typeConverter.convertType(memRefTy.getElementType());
     auto globalType = constantElementType;
-    for (int i = shape.size() - 1; i >= 0; i--)
-      globalType = LLVM::LLVMType::getArrayTy(
-          globalType.cast<LLVM::LLVMType>(), ArrayAttrIntVal(shape, i));
+
+    if (shape.empty()) {
```
tjingrant, line 331: @doru1004 this is the necessary modification to get LLVM-level translation to work. Basically, LLVM global op lowering thinks that the value attribute with type tensor<i64> is a splat elements attribute, which is true, but still somewhat surprising (and I don't know whether not treating a scalar tensor as a splat element would solve this issue more elegantly, but if that is the case then it would be something we can contribute to MLIR upstream). And to lower a splat element attribute, the attribute needs to have either array type or vector type whereas the scalar type is not supported. To workaround this issue I convert the scalar to become an LLVM array of size 1. This seems fine because the shape information of the lowered LLVM array is never used to re-construct the memref anyways.

c.f. https://github.com/llvm/llvm-project/blob/047400ed8204ebcc0b361ca9285b34ea91479b69/mlir/lib/Target/LLVMIR/ModuleTranslation.cpp#L124

doru1004: I think the change makes sense :)

PR #357, opened 2020-10-21
Change check-onnx-backend

chentong319:
Purpose of the change is to make check-onnx-backend easier to use

1. Use the model name for each test case, instead of temp_model for all. E,g name for test_add_cpu is "test_add". A side effect is that the *.onnx and *.so for each test case are kept.

2. Add error report if the expected result of a step is not generated. There are two added failed message: "Failed save model", and "Failed ONNX MLIR"

3. User can invoke test for one test case by:  `make BACKEND_TEST=test_add_cpu check-onnx-backend`

4. A target is added to clean the *.onnx and *.so in build/test/backend. `make clean-onnx-backend`


test/backend/CMakeLists.txt
```diff
@@ -15,5 +15,11 @@ add_custom_target(check-onnx-backend
         COMMAND ${PYTHON_EXECUTABLE}
         ${CMAKE_CURRENT_BINARY_DIR}/test.py)
 
+add_custom_target(clean-onnx-backend
```
AlexandreEichenberger, line 18: Is the cleanup called when doing a new "full" check-onnx-backend? Otherwise, we may have mixures of new and old files.

AlexandreEichenberger (approved):
LGTM, personally, I would add the "cleanup" when starting a new full run, because that way we would avoid mixing models from different runs, in case something crashes.

test/backend/test.py
```diff
@@ -87,11 +87,22 @@ class DummyBackend(onnx.backend.base.Backend):
     @classmethod
     def prepare(cls, model, device='CPU', **kwargs):
         super(DummyBackend, cls).prepare(model, device, **kwargs)
+        name = model.graph.name
+        model_name = name+".onnx"
+        exec_name = name + ".so"
+        # Clean the temporary files in case
+        os.system("rm -f " + model_name)
```
tjingrant, line 94: It'll be safer to use temporary directories in python: https://stackoverflow.com/a/55104228 . If there's a bug in code logic like this the consequence can be severe. So we might want to delegate things like this to existing python modules.

To aid readability, you can give it a readable prefix/suffix as per TempDirectory documentation (https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory).

For a more complete example, you can see python code in our debug utility:
https://github.com/onnx/onnx-mlir/blob/50fca37336c1f8e8590949fb916abda1f60fdb85/utils/debug.py#L64

test/backend/CMakeLists.txt
chentong319: For each test, the temporary files will be deleted before the new test. This is done in the test.py.  I  did not clean the whole directory in case someone need some previous result.

test/backend/test.py
chentong319: The temporary result, *.onnx or *.so, may be needed for debugging, if a test fails.  These files are not temporary files to CMakefile
To avoid explicit delete in python, I think that we can call the clean automatically before each check-onnx-mlir, as Alex suggested. 
Since the entanglement of unitest and cmake, it seems no perfect solution.

tjingrant: Just want to add that leaving *.onnx or *.so files on the disk by default may not be a good idea, given that model tests require *.onnx files with size between 10-500MB. This means test artifacts will take up gigabytes of space by default.

Any chance we can have them cleaned up by default, and in the case of failure or user request (e.g., through environment variable), keep the files?

PR #358, opened 2020-10-22
Add a lit test for krnl shape lowering when its input is a MemRef with affine_map

tungld:


AlexandreEichenberger (approved):
LGTM, thanks

PR #359, opened 2020-10-23
Cleanup runtime api

s390xlinux:
@tjingrant I started to update the JNI wrapper code to the new runtime API and noticed a few inconsistencies. So I fixed them with this patch. One thing I left untouched is that we are still using `omTensorCreateEmptyDeprecated` in `KrnlToLLVM.cpp`. I assume you have some plan to fix that in the future.


s390xlinux:
Closing, wrong github ID used.

PR #360, opened 2020-10-23
Cleanup runtime api

gongsu832:
@tjingrant I started to update the JNI wrapper code to the new runtime API and noticed a few inconsistencies. So I fixed them with this patch. One thing I left untouched is that we are still using `omTensorCreateEmptyDeprecated` in `KrnlToLLVM.cpp`. I assume you have some plan to fix that in the future.


src/Runtime/OMTensor.inc
```diff
@@ -113,20 +121,17 @@ OMTensor *omTensorCreate(
...
-        free(tensor->_stride);
-      return NULL;
-    }
     tensor->_allocatedPtr = data_ptr;
     tensor->_alignedPtr = data_ptr;
     tensor->_rank = rank;
     tensor->_dataType = dtype;
     tensor->_owning = false;
+  } else {
+    if (tensor->_shape)
```
AlexandreEichenberger, line 130: Smart to test only the first one... as if the first one fails, then the second pointer is garbage, and if the second one fails, then we have to free only the first one. 

Why are some failure to allocate memory resulting here in returning NULL, whereas in `OMTensor(int rank)` above we throw an assert?

Also, do you check if pointers are not null before usage? If so, you may want to set the pointers (such as `tensor->_shape` back to null in such cases.

include/onnx-mlir/Runtime/OMTensor.h
```diff
@@ -52,14 +52,6 @@ struct OMTensor;
 typedef struct OMTensor OMTensor;
 #endif
 
-/* Helper function to compute the number of data elements */
```
AlexandreEichenberger, line 55: Did you just move the code below? Or was there a substantive change, I could not tell.

AlexandreEichenberger (commented):
LGTM, since Tian was so involved with the original code, it would probably be best if Tian is the one approving the patch. Thanks.

src/Runtime/OMTensor.inc
gongsu832: `OMTensor(int rank)` is a C++ constructor which does not return value so it throws a runtime_error.

`tensor->_shape` and `tensor->_stride` are guaranteed to be not NULL if a tensor is created successfully, i.e., they are always initialized upon successful tensor creation so there is no need to set them to NULL upon failure.

Actually I think I just saw a bug in `omTensorDestroy` that `tensor->_shape` and `tensor->_stride` were not freed. I will fix that. Thanks.

include/onnx-mlir/Runtime/OMTensor.h
gongsu832: Just moved the helper function `getNumOfElems` from user header into `OMTensor.inc` since it really is an internal function not meant for the user.

src/Runtime/OMTensor.inc
```diff
@@ -192,11 +222,10 @@ OMTensor *omTensorCreateEmpty(
 
 /* OMTensor destroyer */
 void omTensorDestroy(OMTensor *tensor) {
-  if (tensor->_owning) {
+  if (tensor->_owning)
     free(tensor->_allocatedPtr);
```
tjingrant, line 226: Might still want to set allocated/aligned ptrs to NULL? This gives us a way to check whether these pointers have been initialized (i.e., initialized if and only if not NULL).

tjingrant (approved):
LGTM! Thanks for improving the runtime API!
Just have a minor comment.

src/Runtime/OMTensor.inc
gongsu832: Setting those pointers to NULL is basically a "waste" since once the `tensor` is freed, its content is considered garbage, i.e., whatever we are setting immediately becomes garbage and nobody should be looking at the content of a freed `tensor` structure. Setting them to NULL is only useful if we intend to reuse the `tensor` structure later on but that's not the case here.

AlexandreEichenberger (approved):
LGTM too

PR #361, opened 2020-10-23
Slice lowering and Support for IndexExpr for combined Shape Inference and Index Computation in Krnl Lowering 

AlexandreEichenberger:
Slice lowering, at this time only for constant start/end/step sizes as support is missing for CeilDIv and FloorDiv in MLIR. I tested it with a patched version of MLIR (PR pending in MLIR) and it works for dynamic start/end/step and sizes.

Also introduce IndexExpr. In the IndexExpr.h, I have added this text.

***1) IndexExpr***

IndexExpr is a single data structure that holds either an integer, an affine
expression or a Value. It is used to compute shape inference, loop bounds, and
index expressions in memory accesses. The main purpose of this data structure is
to use a single function to either determine shape inference or the actual
shapes when allocating/looping during the lowering.

During Shape inference, no code is generated; the IndexExpr will only be used to
eitehr determine the actual constant size or a questionmark (signifying unknown
at compile time).

During lowering, code can be generated, and if fact it must, to fill in the
information that might be missing at compile time. The same IndexExpression
computation are actually used to determine the sizes, indices, and access
functions. Because AffineExpr have several advantages over more generic Value
computations, the IndexExpr maintain computations as AffineExpr as long as
possible. For example a "dim / literal const" is affine and would be represented
as such in the IndexExpr, but if the denominator was in fact another symbol or
computation, such as "dim / shape[3]", then the same IndexExpr would lower
its representation to a Value computation.

IndexExpr can be querried to determine if they are represented at any times
as an Integer Literal, an AffineExpr, or a generic Value. IndexExpr can be
operated on with operations typically found in index computations, namely.

Add, Sub, Mult, Mod, CeilDiv, FloorDiv with the usual mathematical meanings.

Clamp(val, min, max) which forces val to be contained within inclusively min and
exclusively max. Clamp can use AffineMaxOp but the result is affine only when
all inputs are integer literals.

Select(compA, compareOperator, compB, trueVal, falseVal) which corresponds to
"compA operator compb ? trueVal : falseVal". The result can be statically
determined when the compare can be evaluated at compile time.

***2) IndexExprContext***

Each IndexExpr must be part of a single context which holds all of the symbols
and Dim associated with them. Symbols are variables that are guaranteed to be
constant during the scope of the IndexExpre. Dim are typically runtime
dimensions of memrefs/tensors during computations to determine the shape of a
memref/tensor; or dims are typically the dynamic loop indices inside loop
structures.

A typical pattern is as follow for a kernel that a) determine the shape of the
output and computations, followed by b) the access pattern within the loop
iterations.

In a), the dims are runtime dimensions of inputs memrefs/tensors, and the
symbols are runtime parameters to the functions that are known to be constant.

In b), the dims are dynamic loop indices, and symbols are any of the
computations derived before the loop to compute the output bounds/shape of the
loop iterations.

When all the computations in a) are constant or affine, then the same
IndexExprContext can be reused between a) and b). It is recommended as it
enables bigger AffineExpr. But when the computations in a) are not affine, then
a new context can be started for the b) part. The non-affine parts of a)
becomes symbols.

Note that in a computation, all expressions must use IndexExpr originating from
the same context.

***3) Code Sample***

3a) Create a context:

```
// During shape inference: no rewriter.

  IndexExprContext context(nullptr, getLoc());

// During lowering.

    IndexExprContext outerloopContex(&rewriter, sliceOp.getLoc());
```
3b) Computations on IndexExpr

```
// IN ONNXShapeHelper.cpp

// Get a value from an input operand (either a constant or a value to load).

    startInput = context.CreateSymbolIndexFromArrayAtIndex(
        op, operandAdaptor.starts(), i);

// Get a dimension from a memref.
    dimInput = context.CreateDimIndexFromMemref(data, dataShape, ii);

// Perform calculations.

    startPlusDim.Add(startInput, dimInput);
    startPos.Select(startInput, CmpIPredicate::slt, 0, startPlusDim,
      startInput);
    // Step < 0: clamp(0, start, dim -1) else clamp(0, start, dim)
    dimMinOneInput.Sub(dimInput, 1);
    neg.Clamp(startPos, 0, dimMinOneInput);
    pos.Clamp(startPos, 0, dimInput);
    startFinal.Select(stepInput, CmpIPredicate::slt, 0, neg, pos);
```

3c) Look at Slice in ONNXOps.cpp on how to use IndexExpr for shape inferences.

```
// Extract the shape of the output.

  SmallVector<int64_t, 4> outputDims;
  IndexExprContext::GetOutputDimsForType(outputDimIndices, outputDims);
  getResult().setType(RankedTensorType::get(outputDims, elementType));
```

3d) Look at Slice.cpp on how to use IndexExpr for lowering.

```
// Create an alloc using dimensions as indices.

        rewriter, op, outputMemRefType, loc, outputDims);

// Use indices to set loop sizes.

outputLoops(rewriter, loc, outputRank);
    outputLoops.createDefineOp();
    for (int ii = 0; ii < outputRank; ++ii)
      outputLoops.pushBounds(outerloopContex, 0, outputDims[ii]);

// Create a sub-context for computations inside the loop iteration.
    IndexExprContext childContext(outerloopContex);

// Create indices with computations for a load.

    for (int ii = 0; ii < outputRank; ++ii) {
      Value loopVal = outputLoops.getInductionVar(ii);
      IndexExpr loopIndex, start, step, actualIndex;
      loopIndex = childContext.CreateDimIndex(loopVal);
      start = childContext.CreateSymbolIndexFromParentContext(starts[ii]);
      step = childContext.CreateSymbolIndexFromParentContext(steps[ii]);
      actualIndex.Mult(step, loopIndex).IncBy(start);
      loadIndices.emplace_back(actualIndex.GetValue());
    }
```


chentong319:
Better to split this PR into two: one for IndexExpr support and one for slice lowering.
I read through the document and code. I think it would be better you can explain your design on Monday meeting or a separate meeting.
Questions I have so far:
1. Can the shape helper defined as a trait of operation? 
2. Explain how the reused code is used in both shape inference and lower.
3. Some design questions: should IndexExpr be subclass of AffineExpr? Should the container be given in IndexExpr constructor?
4. How to pass the loop structure to indexexpr for lowering?

AlexandreEichenberger:
`Better to split this PR into two: one for IndexExpr support and one for slice lowering.
I read through the document and code. I think it would be better you can explain your design on Monday meeting or a separate meeting.
Questions I have so far:

Can the shape helper defined as a trait of operation?
Explain how the reused code is used in both shape inference and lower.
Some design questions: should IndexExpr be subclass of AffineExpr? Should the container be given in IndexExpr constructor?
How to pass the loop structure to indexexpr for lowering?`

@chentong319 I agree that in principle it is two tickets, but I would have a hard time veryfying the correctness of IndexExpr without a test case. Thus the merging with Slice.

I am trying to make shape helper a class that could be merged with the operations, but it lead to a strange behavior. I will continue to try, as a revision of this patch.

Same function is used to call for the computations in shape inference and lowering (`HandleSliceOpParams`). In shape inference, no rewriter is given to the context, as there is none (not possible to rewrite code). During lowering, a rewriter is given and code will be generated as needed.

IndexExpr is not a subclass of AffineExpr as IndexExpr can morph into a Value (operations) once you have a non-affine results. Nonaffine can be gotten in many ways, e.g. divide by non-literal, divide by negative number, select, result of a min/max.

I rewrote the way sub-context are used in loops, big thanks for your insight on this. Makes a lot of sense.

src/Dialect/ONNX/ONNXShapeHelper.hpp
```diff
@@ -0,0 +1,42 @@
...
+int64_t ArrayAttrIntVal(ArrayAttr a, int i);
+int64_t ArrayAttrIntVal(Optional<ArrayAttr> a, int i);
+// Returns the ConstantOp which defines an MLIR Value or null.
+ONNXConstantOp getONNXConstantOp(Value value);
+
+DenseElementsAttr getDenseElementAttributeFromValue(Value value);
+
+bool getIntegerLiteralFromValue(Value value, int64_t &intLit);
+
+LogicalResult HandleSliceOpParams(ONNXSliceOp *sliceOp,
```
chentong319, line 37: What is shared between shape inference and lowering is the outputDims. The other IndexExprs are auxiliary expressions that are used only in lowering. Can we have a uniform interface across the operations?  Make the OutputDim the first SmallVectorImpl and others with default NULL.

src/Dialect/ONNX/IndexExpr.hpp
```diff
@@ -0,0 +1,348 @@
...
+  IndexExpr &InitAsQuestionmark(IndexExprContext &context);
+  IndexExpr &InitAsLiteral(IndexExprContext &context, int64_t val);
+  IndexExpr &InitAsSymbol(IndexExprContext &context, Value val);
+  IndexExpr &InitAsDim(IndexExprContext &context, Value val);
+  IndexExpr &InitAsValue(IndexExprContext &context, Value val);
+  IndexExpr &InitAsAffineExpr(IndexExprContext &context, AffineExpr val);
+  // Higher-level initiation calls that extract info
+  IndexExpr &InitAsDimFromMemref(IndexExprContext &context, Value memref,
+      ArrayRef<int64_t> memrefShape, int index);
+  IndexExpr &InitAsSymbolFromArrayAtIndex(
```
chentong319, line 310: The term Dim and Index is not very clear to me. 
1. We may want to use a different name for IndexExpr.
2. AtIndex: means for a dimension specified, or the position of the affine map?
3. AsDim: is it for loop index or array dimension?

src/Dialect/ONNX/IndexExpr.hpp
```diff
@@ -0,0 +1,348 @@
...
+  // Getters.
+  int64_t GetLiteral() const;
+  AffineExpr GetAffineExpr();
+  Value GetValue();
+  IndexExprContext &GetContext() const;
+  IndexExprContext *GetContextPtr() const;
+  Location GetLocation() const;
+
+  // Possibly Affine Operations.
+  IndexExpr &Add(IndexExpr &a, IndexExpr &b);
```
chentong319, line 263: If all the operands are provided, are these operation static to the class? I assume that  we will define operator  for them later.

chentong319 (commented):
Most of current questions are related to the definition of interfaces.

src/Dialect/ONNX/ONNXShapeHelper.hpp
AlexandreEichenberger: I changed the interface in the last edition to hide the long list of parameters inside a struct. To respond to your question: during shape inference, only outputDims are needed; the other one are used to compute the outputDim. Here is the new interface
```
/// When defining support for a new op, add one such stuct which must minimally
/// compute the outputDims present in the parent class. Computation should be
/// performed using a `Compute` function. Return success on successful
/// computation of all the IndexExpr. During shape inference, object is built
/// using a null-ptr rewriter; during lowering, the rewriter is nonnull and will
/// be used to generate code.
template <class OP>
struct ONNXOpShapeHelper {
  ONNXOpShapeHelper(OP *newOp, ConversionPatternRewriter *rewriter);

  // Define in every children.
  LogicalResult Compute(ONNXSliceOpAdaptor operandAdaptor) {
    llvm_unreachable("implement in child structs");
  }

  // Data that must be present for every ShapeHelper operation. Op and context
  // are initialized in the constructor, and outputDims is computed by the
  // child's struct `Compute` function.
  OP *op;
  IndexExprContext context;
  SmallVector<IndexExpr, 4> outputDims;
};

struct ONNXSliceOpShapeHelper : public ONNXOpShapeHelper<ONNXSliceOp> {
  ONNXSliceOpShapeHelper(
      ONNXSliceOp *newOp, ConversionPatternRewriter *rewriter);

  LogicalResult Compute(ONNXSliceOpAdaptor operandAdaptor);

  // Additional data for SliceOp.
  SmallVector<IndexExpr, 4> starts;
  SmallVector<IndexExpr, 4> ends;
  SmallVector<IndexExpr, 4> steps;
};
```

What is in `ONNXOpShapeHelper` is the minimum that is needed for shape inference. When generating a subclass for a specific op, we can add additional parameters, but during the shape inference, only the first set is needed.

Inside the IndexExpression, the difference between shape inference and lowering is that during shape inference, expressions are either constant or '?'. We cannot/do not want to gen code during shape inference. so we will not go beyond the '?'. Note the expressions can go back to constant

res = (1 > 0) ? 3 : runtime-expo

will result in a constant...

During lowering, we will generate code.

Note that in the interface above, the `Compute(...)` function can detect if we are in shape inference or not by using the  `context.IsShapeInferencePass()` test and thus can customize the computations, if desired.

Bottom line, during inference: must compute `outputDims`. During lowering, compute that one plus any others needed for lowering. Hope this helps.

doru1004:
Could you separate the support for IndexExpr for combined Shape Inference and Index Computation in Krnl Lowering in a separate patch? I think that's a big enough change worthy of a standalone patch.

src/Dialect/ONNX/IndexExpr.hpp
AlexandreEichenberger: 1 & 3) Dim and Symbol are concept from AffineExpr. It is confusing and I don't like it much too. Dim are "runtime input sizes" during shape inferences, and Dim are "loop index" during lowering of code inside the loop iterations. 

I can rename Dim to be (both) "InputDim" and "LoopInductionVar" if you prefer, in the IndexExprContext.create* methods.

2) index of the array. Let me rename the parameter names to make it clearer.

src/Dialect/ONNX/IndexExpr.hpp
AlexandreEichenberger: No, it writes into the *this object, and then returns the *this.

That way you can have chains: 

```
newObj.add(a, b).divby(c)
```

First newObj becomes `a+b` and then that result is divided by `c`.

I did not want to use dynamic memory to alloc/free on demand...

src/Conversion/ONNXToKrnl/Tensor/Slice.cpp
```diff
@@ -0,0 +1,98 @@
...
+    t1.Sub(i1, i1);
+    t1.debugPrint("index loop i -i");
+#endif
+
+    // Proceed with the load data["i * step + start} for all dim].
+    Value loadVal;
+    SmallVector<Value, 4> loadIndices;
+    bool loadIsAffine = true;
+    for (int ii = 0; ii < outputRank; ++ii) {
+      Value loopVal = outputLoops.getInductionVar(ii);
```
chentong319, line 60: Does user need to know the order of Dims used in Helper.cpp? Or is there a default order?

src/Dialect/ONNX/IndexExpr.cpp
```diff
@@ -0,0 +1,1062 @@
...
+  parentContext = &newParentContext;
+}
+
+//===----------------------------------------------------------------------===//
+// IndexExprContext builder for IndexExpr.
+//===----------------------------------------------------------------------===//
+
+IndexExpr IndexExprContext::createUndefinedIndex() {
+  IndexExpr res;
+  return res.initAsUndefined();
```
chentong319, line 60: It seems to me that the res.initasUndefined() returns the object res. Will this cause any memory issue?

test/backend/test.py
```diff
@@ -351,6 +351,9 @@ def supports_device(cls, device):
     "test_reciprocal_cpu",
     "test_reciprocal_example_cpu",
 
+    # Slice, deactivated as axis is unsuported as runtime param.
+    #"test_slice_cpu",
```
chentong319, line 355: Any issue in running test_slice_cpu backend test?

AlexandreEichenberger:
New version with new interface is in a new PR #361. As a result, I am closing this one, which we can still look at for reference

PR #367, opened 2020-10-29
Fix an issue in lowering LSTM, RNN, and GRU

tungld:
The fusion of loops for matrix multiplications (e.g. `Ht-1*(Ri^T)`) and gate computations (e.g. hidden state ht, cell state ct) in LSTM, RNN, and GRU lowerings caused wrong results.

The reason for the wrong results was that gate computations updated an element of the current state `Ht-1` during a loop iteration which led to wrong results for `Ht-1*(Ri^T)` in the next loop iteration.

This patch
- fixes the above issue by using two separate loops for matrix multiplications and gate computations.
- adds numerical tests using rapidcheck for LSTM, RNN, and GRU, which compares onnx-mlir implementations against naive implementations 
- fixes some issues related to dynamic dimensions.


doru1004 (approved):
Thanks for fixing this. LGTM!

doru1004 (changes requested):
Can we have a runtime test added for this? I see you have lit tests which is great but since LSTM, RNN and GRU are such complex operations, can we have some runtime tests which are complex enough to showcase this particular wrong-result error you're fixing in this patch?

tungld:
I guess numerical tests are what you mean. I wrote a numerical test for LSTM (TestLSTM.cpp). Will do the same for RNN and GRU and ping you when they are done.

tungld:
@doru1004 Hi Doru, the patch is ready now. I added numerical tests, similar to what we have done for Conv.

doru1004:
Waiting for tests to finish.

doru1004 (approved):
LGTM

PR #369, opened 2020-10-30
Make test.py reusable

tungld:

This patch makes test.py reusable for a different backend. In such a case, only CMakeLists.txt is enough to write.


test/backend/CMakeLists.txt
```diff
@@ -1,3 +1,4 @@
+set (ONNX_MLIR_BACKEND "onnx-mlir")
```
tjingrant, line 1: I would rename it to something like `TEST_DRIVER`. The word _backend_ has many meanings in the context of both ONNX and MLIR. Probably don't want to confuse users any futher.

tungld: Sure, I often struggle with naming things. Changed the name, and just removed a redundant comment in test.py.

tjingrant (approved):
LGTM!

PR #370, opened 2020-10-30
Avoid model copy during import

gramalingam:
Pass ONNX model via const-reference to avoid unnecessary copy of model when importing the model.


chentong319 (approved):
LGTM

tjingrant:

PR #371, opened 2020-10-30
Prevent slot reuse when operators are passed to unknown operations 

doru1004:
In the current version of onnx-mlir, when two `krnl.getref` outputs are passed to a `krnl.memcpy` operation they cannot share the same slot. This rule needs to be expanded to operations from other dialects which may end up at this level of abstraction. For these operations we assume the same restriction as for `krnl.memcpy` so that they cannot share the same slot.


AlexandreEichenberger (approved):
LGTM, thanks for adding this case

PR #373, opened 2020-11-03
Improved version of IndexExpr (and slice for testing)

AlexandreEichenberger:
**1) IndexExpr**

IndexExpr is a single data structure that holds either an integer, an affine
expression or a Value. It is used to compute shape inference, loop bounds, and
index expressions in memory accesses. The main purpose of this data structure is
to use a single function to either determine shape inference or the actual
shapes when allocating/looping during the lowering.

During Shape inference, no code is generated; the IndexExpr will only be used to
eitehr determine the actual constant size or a questionmark (signifying unknown
at compile time).

During lowering, code can be generated, and if fact it must, to fill in the
information that might be missing at compile time. The same IndexExpression
computation are actually used to determine the sizes, indices, and access
functions. Because AffineExpr have several advantages over more generic Value
computations, the IndexExpr maintain computations as AffineExpr as long as
possible. For example a "dim / literal const" is affine and would be represented
as such in the IndexExpr, but if the denominator was in fact another symbol or
computation, such as "dim / shape[3]", then the same IndexExpr would lower
its representation to a Value computation.

IndexExpr can be querried to determine if they are represented at any times
as an Integer Literal, an AffineExpr, or a generic Value. IndexExpr can be
operated on with operations typically found in index computations, namely.

Add, Sub, Mult, Mod, CeilDiv, FloorDiv with the usual mathematical meanings.

Clamp(val, min, max) which forces val to be contained within inclusively min and
exclusively max. Clamp can use AffineMaxOp but the result is affine only when
all inputs are integer literals.

Select(compA, compareOperator, compB, trueVal, falseVal) which corresponds to
"compA operator compb ? trueVal : falseVal". The result can be statically
determined when the compare can be evaluated at compile time.

**2) IndexExprContext**

Each IndexExpr must be part of a single context which holds all of the symbols
and Dim associated with them. Symbols are variables that are guaranteed to be
constant during the scope of the IndexExpre. Dim are typically runtime
dimensions of memrefs/tensors during computations to determine the shape of a
memref/tensor; or dims are typically the dynamic loop indices inside loop
structures.

A typical pattern is as follow for a kernel that a) determine the shape of the
output and computations, followed by b) the access pattern within the loop
iterations.

In a), the dims are runtime dimensions of inputs memrefs/tensors, and the
symbols are runtime parameters to the functions that are known to be constant.

In b), the dims are dynamic loop indices, and symbols are any of the
computations derived before the loop to compute the output bounds/shape of the
loop iterations.

When all the computations in a) are constant or affine, then the same
IndexExprContext can be reused between a) and b). It is recommended as it
enables bigger AffineExpr. But when the computations in a) are not affine, then
a new context can be started for the b) part. The non-affine parts of a)
becomes symbols.

Note that in a computation, all expressions must use IndexExpr originating from
the same context.

**3) Code Sample**

**3a) Create a context:**

```
// During shape inference: no rewriter.

  IndexExprContext context(nullptr, getLoc());

// During lowering.

    IndexExprContext outerloopContex(&rewriter, sliceOp.getLoc());
```

**3b) Computations on IndexExpr**

```
// IN ONNXShapeHelper.cpp

// Get a value from an input operand (either a constant or a value to load).

    startInput = context.CreateSymbolIndexFromArrayAtIndex(
        op, operandAdaptor.starts(), i);

// Get a dimension from a memref.
    dimInput = context.CreateDimIndexFromMemref(data, dataShape, ii);

// Perform calculations.

    // Calculation for start: start < 0 ? start + dim : start.
    IndexExpr startPos =
        IndexExpr::select(startInput < 0, startInput + dimInput, startInput);
    // Step < 0: clamp(0, start, dim -1) else clamp(0, start, dim)
    IndexExpr neg = startPos.clamp(0, dimInput - 1);
    IndexExpr pos = startPos.clamp(0, dimInput);
    IndexExpr startFinal = IndexExpr::select(stepInput < 0, neg, pos);
```

**3c) Look at Slice in ONNXOps.cpp on how to use IndexExpr for shape inferences.**

```
// Extract the shape of the output.

  SmallVector<int64_t, 4> outputDims;
  IndexExprContext::GetOutputDimsForType(outputDimIndices, outputDims);
  getResult().setType(RankedTensorType::get(outputDims, elementType));
```

**3d) Look at Slice.cpp on how to use IndexExpr for lowering.**

```
// Create an alloc using dimensions as indices.

        rewriter, op, outputMemRefType, loc, outputDims);

// Use indices to set loop sizes.

  outputLoops(rewriter, loc, outputRank);
    outputLoops.createDefineOp();
    for (int ii = 0; ii < outputRank; ++ii)
      outputLoops.pushBounds(outerloopContex, 0, outputDims[ii]);

// Create a sub-context for computations inside the loop iteration.

    IndexExprContext childContext(outerloopContex);

// Create indices with computations for a load.

    for (int ii = 0; ii < outputRank; ++ii) {
      Value loopVal = outputLoops.getInductionVar(ii);
      IndexExpr loopIndex = childContext.createLoopIterIndex(loopVal);
      IndexExpr start = childContext.createSymbolIndexFromParentContext(
          shapeHelper.starts[ii]);
      IndexExpr step = childContext.createSymbolIndexFromParentContext(
          shapeHelper.steps[ii]);
      IndexExpr actualIndex = (step * loopIndex) + start;
      loadIndices.emplace_back(actualIndex.getValue());
    }
```


AlexandreEichenberger:
This version has the new interface that allows us to do this:
```
IndexExpr res = a * 3 + b
```
All memory associated with the IndexExpr reside in the IndexExprContext.

AlexandreEichenberger:
@chentong319 In response to an earlier comment, I have many dynamic lit tests and a few mlir-to-exec that I check on my own. I will be able to include them only once MLIR handling of Ceil/Floor are in. The patch was accepted but requires further changes.

chentong319:
The code looks good enough to be accepted so that we can use this functionality. Further improvements may be needed with the feedback of users.
When I tried dynamic dimension for tile with AffineExpr directly, I found another benefit of IndexExpr. In my the lowering code, I first generate semi-affine-expr to capture the computation and then call apply to feed the input. For some case, the semi-affine-expr will become non affine and the result will be used for ordinary load/store, not affine load/store. This works for check-onnx-lit, but failed for check-onnx-backend at compile time. Such affine.apply is not supported. With IndexExpr, we can capture the computation uniformly even just for lowering.

chentong319 (approved):
Ready to be used. Refine later with feedback.

AlexandreEichenberger:
@chentong319 thanks for the input again; and yes, the interface will continue to evolve. I have started working on the GEMM op and there are some changes here to reflect this already. Looking forward to continued input on this.

doru1004 (approved):
LGTM

src/Dialect/ONNX/ONNXShapeHelper.cpp
```diff
@@ -0,0 +1,170 @@
...
+LogicalResult ONNXSliceOpShapeHelper::Compute(
+    ONNXSliceOpAdaptor operandAdaptor) {
+  // Shape inference indicated by passing a null rewriter pointer.
+  Operation *genericOp = reinterpret_cast<Operation *>(op);
+
+  // Get info about input data operand.
+  Value data = operandAdaptor.data();
+  int64_t dataRank = data.getType().cast<ShapedType>().getShape().size();
+
+  // Get each of the axes, and save the litteral values in axesIntLit.
```
doru1004, line 54: Litteral -> literal

PR #375, opened 2020-11-05
Handle TileOp with dynamic dimension

chentong319:
There are two changes:

1. In shape inference, remove unnecessary checks and allow dynamic dimensions.

2. In lowering to Krnl, avoid AffineMap if  the result is not affine. Previous code used semi-affine map. If the result of Affine.apply is not Affine, the result is used for ordinary load/store, not affine one. Such code works for check-onnx-lit. But for check-onnx-backend, it failed. 

This PR is tested with the new functionality offered by PR354.
`OM_FORCE_FIRST_DIM_UNKNOWN=1 make BACKEND_TEST=test_tile_cpu check-onnx-backend` 

This patch provides a base for us to try IndexExpr on TileOp for static and dynamic shape.


AlexandreEichenberger (commented):
Quick question: is this code intended to be permanent, or will you try to move it to IndexExpr? If you intend to move it, then I am happy to approve it.

If you want it to be more permanent, I would ask you to remove many of the auto. Folks in MLIR complain about them, and its really easy to use the actual type for pretty much all but complicated types.

AlexandreEichenberger:
If you intend to move it to index expr, here are a couple of examples on how things are done quickly.

Once you have defined a shape helper, alloc a new array:
```
    // Insert an allocation and deallocation for the output of this operation.
    MemRefType outputMemRefType = convertToMemRefType(*op->result_type_begin());
    Type elementType = outputMemRefType.getElementType();
        rewriter, op, outputMemRefType, loc, shapeHelper.outputDims);
```

create a loop using the same dimensions as in a vector of index expressions:

```
    outputLoops.createDefineOp();
    outputLoops.pushAllBounds(shapeHelper.outputDims);
```

E.g. test if a dim is >1, in which case we use the index, or else we use the index zero:

```
        IndexExpr dim = outerContext.createSymbolIndexFromParentContext(
            shapeHelper.cDims[x]);
        cAccessFct.emplace_back(IndexExpr::select(dim > 1, resAccessFct[x], 0));
```

load and store values using vectors of expressions, automatically generating affine or std load depending on the index expressions being all fine:

```
    SmallVector<IndexExpr, 4> aAccessFct, bAccessFct;
    if (gemmOp.transA() != 0)
      aAccessFct = {k, n};
    else
      aAccessFct = {n, k};
    if (gemmOp.transB() != 0)
      bAccessFct = {m, k};
    else
      bAccessFct = {k, m};
    // Add mat mul operation.
    Value loadedA = outerContext.createLoadOp(operandAdaptor.A(), aAccessFct);
    Value loadedB = outerContext.createLoadOp(operandAdaptor.B(), bAccessFct);
    Value loadedY = outerContext.createLoadOp(alloc, resAccessFct);
```

chentong319:
@AlexandreEichenberger I will try IndexExpr for the TileOp later. Thanks for the note of new way of allocation. 
I agree on reducing the use of auto: not friends for human reader.

AlexandreEichenberger (approved):
LGTM, no need to remove auto now if you will do a revision with IndexExpr

PR #376, opened 2020-11-06
Fix the numerical stability issue in ReduceLogSumExp

### LINKED ISSUE #365: Logsumexp is not numerically stable
[Rewriting](https://github.com/onnx/onnx-mlir/blob/master/src/Transform/ONNX/Decompose.td#L46) of logsumexp in terms of simpler onnx ops need to take into account the numerical stability.

To reproduce the issue,
```python
import torch
import torch.nn as nn

class MyModule(nn.Module):
    def forward(self, x):
        return torch.logsumexp(x, dim=1)

model = MyModule()
x = torch.randn(3, 4)
torch.onnx.export(model, x, "logsumexp.onnx")

import sys
sys.path = ['test/backend'] + sys.path
import test

import onnx
onnx_model = onnx.load("logsumexp.onnx")
session = test.DummyBackend.prepare(onnx_model)
import numpy as np
x = np.random.normal(0, 1, (3, 4)).astype(np.float32) + 100.0
out = session.run(x)[0]
expected = model(torch.tensor(x)).numpy()
print(f"out={out}")
print(f"logsumexp(x)={expected}")
np.testing.assert_allclose(out, expected, rtol=1e-6)
```

The result was
```
out=[inf inf inf]
logsumexp(x)=[102.05343  101.04576  101.345764]
```
and the assertion fails. Note that if I comment out `+ 100.0`, then the test passes.

To make logsumexp numerically stable, we need to implement it as `logsumexp(x, dim=axis) = log(sum(exp(x - max), dim=axis)) + max` where `max=reduce_max(x, dim=axis, keepdims=True)`.

Similarly it would be better to implement `logsoftmax(x)` as `logsoftmax(x, dim=axis) = x - logsumexp(x, dim=axis)` to avoid computing the logarithm of a very small number.

tungld:
Resolves #365


tungld:
@tjingrant Any ideas on why the check-doc test failed? It passed on my local environment. Thanks!

src/Dialect/ONNX/ONNXOps.td.inc
```diff
@@ -188,8 +188,8 @@ def ONNXArgMaxOp:ONNX_Op<"ArgMax",
   let summary = "ONNX ArgMax operation";
   let description = [{
   "Computes the indices of the max elements of the input tensor's element along the "
-  "provided axis. The resulting tensor has the same rank as the input if keepdims equal 1. "
-  "If keepdims equal 0, then the resulting tensor have the reduced dimension pruned. "
+  "provided axis. The resulted tensor has the same rank as the input if keepdims equal 1."
```
doru1004, line 191: resulting .... equals 1.

src/Dialect/ONNX/ONNXOps.td.inc
```diff
@@ -188,8 +188,8 @@ def ONNXArgMaxOp:ONNX_Op<"ArgMax",
...
+  "If keepdims equal 0, then the resulted tensor have the reduced dimension pruned. "
```
doru1004, line 192: resulting tensor has

src/Dialect/ONNX/ONNXOps.td.inc
```diff
@@ -214,8 +214,8 @@ def ONNXArgMinOp:ONNX_Op<"ArgMin",
   let summary = "ONNX ArgMin operation";
   let description = [{
   "Computes the indices of the min elements of the input tensor's element along the "
-  "provided axis. The resulting tensor has the same rank as the input if keepdims equal 1. "
-  "If keepdims equal 0, then the resulting tensor have the reduced dimension pruned. "
+  "provided axis. The resulted tensor has the same rank as the input if keepdims equal 1."
```
doru1004, line 217: resulting ... equals

src/Dialect/ONNX/ONNXOps.td.inc
```diff
@@ -214,8 +214,8 @@ def ONNXArgMinOp:ONNX_Op<"ArgMin",
...
+  "If keepdims equal 0, then the resulted tensor have the reduced dimension pruned. "
```
doru1004, line 218: equals ... resulting .. has

src/Dialect/ONNX/ONNXOps.td.inc
```diff
@@ -2407,49 +2407,29 @@ def ONNXLoopOp:ONNX_Op<"Loop",
...
-  "      // The values below are bound to the output variables of the loop and therefore accessible"
-  "      // b_out; user_defined_vals; keepgoing_out;"
+  "      // These below values are live-out from the loop and therefore accessible"
+  "      b_out; user_defined_vals; keepgoing_out;"
   "    }"
   ""
   "There are several things of note in this code snippet:"
   ""
-  "1) Values from the enclosing scope (i.e. variable \"a\" here) are in scope and can"
+  "1) Values from the enclosing scope (i.e. variable a here) are in scope and can"
```
doru1004, line 2426: `\"a\"`

src/Dialect/ONNX/ONNXOps.td.inc
```diff
@@ -3626,7 +3604,7 @@ def ONNXRangeOp:ONNX_Op<"Range",
   [NoSideEffect, DeclareOpInterfaceMethods<ShapeInferenceOpInterface>]> {
   let summary = "ONNX Range operation";
   let description = [{
-  "Generate a tensor containing a sequence of numbers that begin at `start` and extends by increments of `delta`"
+  "Generate a tensor containing a sequence of numbers that begin at `start` and extends by increments of `delta` "
```
doru1004, line 3607: white space at end of line before the closing \".

src/Dialect/ONNX/ONNXOps.td.inc
```diff
@@ -4915,13 +4917,11 @@ def ONNXSliceOp:ONNX_Op<"Slice",
   "Slices uses `starts`, `ends`, `axes` and `steps` inputs to specify the start and end"
   "dimension and step for each axis in the list of axes, it uses this information to"
   "slice the input `data` tensor. If a negative value is passed for any of the"
-  "start or end indices, it represents number of elements before the end of that"
+  "start or end indices, it represent number of elements before the end of that"
```
doru1004, line 4920: represents the

doru1004 (approved):
LGTM but please fix the comments.

tungld:
@doru1004 `NNXOps.td.inc ` was automatically generated by `make OMONNXOpsIncTranslation`. Fixed the problem by using onnx 1.7.0 (which has recently used in onnx-mlir). Thanks for your comments!

PR #379, opened 2020-11-07
test dynamic shape with check-onnx-backend

chentong319:
Our importer can change static shaped tensor for input parameters into dynamic shaped tensor when the env variable IMPORTER_FORCE_DYNAMIC is defined. Currently, not all the test cases in check-onnx-backend can pass for dynamic shape. This patch adds a list (`test_not_for_dynamic`) of test cases that failed and skip them for dynamic shape. As a result, when we run `IMPORTER_FORCE_DYNAMCE=1 make check-onnx-backend`, the test should succeed. 
We will gradually reduce the list when more support for dynamic shape is added.


AlexandreEichenberger (approved):
LGTM, thanks for adding some explanation on how to run the test

PR #382, opened 2020-11-09
Add support for GEMM dynamic support

AlexandreEichenberger:
Added dyn support as well as lit tests where dim are partially or fully undefined.

In doing so, I also added support for creating load/store operations using list of IndexExpr. This allows to test if all indices are affine, in which case affine load/store are generated.

```
   // Code Create for possibly affine load and store. Memref shape is expected to
   // be of the same dimension than the indices array size. Each index expression
   // will be transformed to a value to be used as indices to the memref. When
   // all index expressions are affine, then an affine memory operation is
   // generated. Otherwise, a standard memory operation is generated.
   Value createLoadOp(Value memref, SmallVectorImpl<IndexExpr> &indices);
   void createStoreOp(
       Value val, Value memref, SmallVectorImpl<IndexExpr> &indices);

```

I also simplified a bit how Slice code is used, using some of the new interface.


src/Conversion/ONNXToKrnl/Math/Gemm.cpp
```diff
@@ -20,181 +21,97 @@ struct ONNXGemmOpLowering : public ConversionPattern {
...
+
+    // Compute the access functions for res[n,m].
+    IndexExpr n =
+        outerContext.createLoopIterIndex(outputLoops.getInductionVar(0));
+    IndexExpr m =
+        outerContext.createLoopIterIndex(outputLoops.getInductionVar(1));
+    SmallVector<IndexExpr, 4> resAccessFct({n, m});
+
+    // Insert res[n,m] = 0.
+    // rewriter.create<AffineStoreOp>(loc, zero, alloc, resAccessFct);
```
doru1004, line 61: Debug commment ?

doru1004: Actually it looks like it's needed based on the comment above.

src/Dialect/ONNX/ONNXShapeHelper.hpp
```diff
@@ -63,6 +64,21 @@ struct ONNXSliceOpShapeHelper : public ONNXOpShapeHelper<ONNXSliceOp> {
...
+      ONNXGemmOp *newOp, ConversionPatternRewriter *rewriter);
+
+  LogicalResult Compute(ONNXGemmOpAdaptor operandAdaptor);
+
+  // Additional data for GemmOp: output = a * b.
+  SmallVector<IndexExpr, 4> aDims; // Dim of A, after applying transpose.
+  SmallVector<IndexExpr, 4> bDims; // Dim of B, after applying transpose.
+  SmallVector<IndexExpr, 4> cDims; // Dim of C, padding "1" when broadcast.
+  bool hasBias; // Whether ther eis a bias (aka C exists).
+  int cRank; // Dim of the original C (not padding dims by 1).
```
doru1004, line 79: Something doesn't look right here. You're ending your block with an `int cRank` declaration.

doru1004: Actually it's fine I thought this was the inferShape method. :)

AlexandreEichenberger: I believe I need that init back. Have added it and I am running the backend tests.

src/Conversion/ONNXToKrnl/Math/Gemm.cpp
AlexandreEichenberger: I believe I need that init back. Have added it back.

src/Dialect/ONNX/ONNXShapeHelper.hpp
AlexandreEichenberger: I added here params that I need both in the computation inside of shape helper as well as in the lowering. Instead of redundantly computing them, I just added them to the struct.

AlexandreEichenberger:
Used Tong's test on a few backend example, e.g.

`OM_FORCE_FIRST_DIM_UNKNOWN=1 make BACKEND_TEST=test_gemm_default_matrix_bias_cpu check-onnx-backend`

AlexandreEichenberger:
@chentong319 Added the dyn tests using your latest PR, and it worked on my end. Please review when you have a chance

src/Conversion/ONNXToKrnl/Math/Gemm.cpp
AlexandreEichenberger: It was needed, it was a bug not to have it. Tian confirmed it.

doru1004: Nice!

doru1004 (approved):
LGTM

PR #389, opened 2020-11-11
Add a modification that was missing when updating to the new llvm commit id

tungld:


AlexandreEichenberger (approved):
LGTM,tx

PR #390, opened 2020-11-11
Control which input to be changed to dynamic shape

chentong319:
Use the value of env variable IMPORTER_FORCE_DYNAMIC to control which input tensor is changed.

- IMPORTER_FORCE_DYNAMIC=-1 : all the inputs will be changed
- IMPORTER_FORCE_DYNAMIC=0 : the first input will be changed
- IMPORTER_FORCE_DYNAMIC=n : inputs[n] will be changed

Docs/Testing is updated for this change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ arch:
   - ppc64le
   - amd64
 env:
-  - NIGHTLY=true
   - NIGHTLY=false
 addons:
   apt:
--- a/docs/Testing.md
+++ b/docs/Testing.md
@@ -20,13 +20,23 @@ With BACKEND_TEST specified, the intermedia result, the .onnx file and .so file,
 
 When the ONNX-to-Krnl conversion of an operator is added, the corresponding backend tests for this operator should be added to test.py. The available test cases can be found in third_part/onnx/onnx/backend/test/case/node. Please note to add suffix `_cpu` to the onnx test name. 
 
-The onnx node tests usually have known dimension size for input tensors. To test tensor with unknown dimension, the model importer (Build/FrontendONNXTransformer.cpp) provides a functionality to generate such cases. When the environment variable, `IMPORTER_FORCE_DYNAMIC`, is set, the frontend import will turn the first dimension of every input tensor of the model into -1. For example:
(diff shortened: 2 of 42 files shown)


AlexandreEichenberger:
also, generally the value 0 means disabled. Right now we have to unset the variable, which is a bit unconventional

AlexandreEichenberger:
Looking at the output of the tool, and I see this:
```
module {
  func @main_graph(%arg0: tensor<?x4x3x2xf32>, %arg1: tensor<?xi64>) -> tensor<3x4x3x2xf32> attributes {input_names = ["data", "indices"], output_names = ["y"]} {
    %0 = "onnx.Gather"(%arg0, %arg1) {axis = 0 : si64} : (tensor<?x4x3x2xf32>, tensor<?xi64>) -> tensor<3x4x3x2xf32>
    return %0 : tensor<3x4x3x2xf32>
  }
  "onnx.EntryPoint"() {func = @main_graph, numInputs = 2 : i32, numOutputs = 1 : i32} : () -> ()
```

I wonder if we should not force the output to be ? only

chentong319:
@AlexandreEichenberger My first version of implementation use 0 as disable. But the convention for inputs index starts from 0. If 0 is used as not change, the test.py has to interpret the meaning of the environment variable. So I changed to the current implementation. I have no strong preference between the two.
I did not touch the output tensor. The ? may come from the input tensor through shape inference.

AlexandreEichenberger:
@chentong319 , zero is fine. I wonder if you would not consider also mapping the variable that right now say if a test is eligible for dyn test to a map

test-name -> ((0,1,2), (1), (), (0))

where the first arg can take ? for the 0/1/2 dim, second take 1 dim, third cannot, and 4th take 1st dim.

that way we can make tests that take as much as runtime param as legal (or wanted to be supported).

then tests that cannot be handled are either commented out, or simply set to -> ()

We should then also provide a way for testing to give in an env var the setting that we want to test. By the way, parsing a json param is super easy in python... just did in in the mlir2FileCheck.py that I added in utils.

AlexandreEichenberger (approved):
LGTM, let you decide if you want this first, and add functionality later

PR #391, opened 2020-11-12
TileOp implementation with  IndexExpr

chentong319:
The shape inference and ONNX to Krnl is implemented with IndexExpr. Different from the previous example of Slice, I put all the lowering code in lowering. The ShapeHelper for TileOp only shares the computation of outputDims. I think that this approach divides the code closer to their semantics:  all the loops for lowering are in the lowering code. In the Shape, there is no need for loops. Consequently, in the lowering, there is no need to inherit IndexExpr context from the ShapeHelper. 
The patch failed some test with dynamic shape because the llvm patch for % is not ready yet. I removed that test case.

### DIFF
--- a/src/Conversion/ONNXToKrnl/Tensor/Tile.cpp
+++ b/src/Conversion/ONNXToKrnl/Tensor/Tile.cpp
@@ -61,8 +61,8 @@ struct ONNXTileOpLowering : public ConversionPattern {
 
     ONNXTileOpShapeHelper shapeHelper(&tileOp, &rewriter);
 
-    if (failed(shapeHelper.Compute(operandAdaptor)))
-      return op->emitError("Failed to scan Tile parameters successfully");
+    assert(
+        !failed(shapeHelper.Compute(operandAdaptor)) && "expected to succeed");
 
     auto resultOperand = tileOp.output();
     auto outputMemRefType = convertToMemRefType(*op->result_type_begin());
--- a/src/Conversion/ONNXToKrnl/Tensor/Tile.cpp
+++ b/src/Conversion/ONNXToKrnl/Tensor/Tile.cpp
@@ -9,6 +9,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.hpp"
+#include "src/Dialect/ONNX/ONNXShapeHelper.hpp"
 
 using namespace mlir;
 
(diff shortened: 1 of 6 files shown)


src/Conversion/ONNXToKrnl/Tensor/Tile.cpp
```diff
@@ -58,84 +59,62 @@ struct ONNXTileOpLowering : public ConversionPattern {
...
 
-    // get input operands, shapes, and rank
-    Value input = operandAdaptor.input();
-    auto inputMemRefType = input.getType().cast<MemRefType>();
-    auto inputShape = inputMemRefType.getShape();
-    int64_t inputRank = inputShape.size();
-    Value repeats = operandAdaptor.repeats();
+    ONNXTileOpShapeHelper shapeHelper(&tileOp, &rewriter);
+
+    if (failed(shapeHelper.Compute(operandAdaptor)))
```
AlexandreEichenberger, line 64: nit: since it already succeeded in shape inference, you could just do an assert here

assert(succeeded(shapeHelper.Compute(operandAdaptor)) && "expected success");

test/mlir/onnx/onnx_lowering.mlir
```diff
@@ -2995,20 +2995,17 @@ func @test_tile1(%arg0 : tensor<4x8xf32>) -> tensor<*xf32> {
   %0 = "onnx.Constant"() { value = dense<[3, 2]> : tensor<2xi64>} : () -> tensor<2xi64>
   %1 = "onnx.Tile"(%arg0, %0) : (tensor<4x8xf32>, tensor<2xi64>) -> tensor<*xf32>
   return %1 : tensor<*xf32>
-  // CHECK: [[INDEX_MAP:#.+]] = affine_map<(d0)[s0] -> (d0 mod s0)>
+  // CHECK: [[INDEX_MAP0:#.+]] = affine_map<(d0) -> (d0 mod 4)>
```
AlexandreEichenberger, line 2998: I would recommend to move this test to the onnx_lowering_with_canonicalize.mlir because it will get rid of the maps and move it into the computation. I don't know how safe it is to check things before the CHECK-LABEL. But I am a very naive FileCheck user, so it may be safe. Anyway, I think its better to canonicalize and see the more readable output.

AlexandreEichenberger (commented):
Looks great, thanks for trying out IndexExpr. I made a few suggestions that will simplify the code even more... see the detailed comments.

AlexandreEichenberger (approved):
LGTM, I will do a rebase and let's try to get it in soon.

PR #393, opened 2020-11-12
Enabled dynamic handling for slice, now that MLIR support floor/ceil ops

AlexandreEichenberger:
Enabled the dynamic handling for slices, which needs the MLIR support that I added. I added examples, unfortunately we cannot have backend test without modification at this time as axis cannot be dynamic. I tested the examples here by using a mill-run approach printing values using vector support.

Also removed debugging calls that are not needed anymore


doru1004 (approved):
LGTM

PR #394, opened 2020-11-13
Onnx function translation

gramalingam:
(1) Extend the ONNX-to-MLIR frontend translator to handle ONNX functions.
(2) Add a test-case to illustrate and test its use.

An ONNX function is an op ```foo``` whose semantics is defined in terms of other (more primitive) ONNX ops. Thus, any call ```y = foo(x)``` can be effectively replaced by another sequence of ONNX-op calls (nodes). This allows for an alternative way to translate such a call into ONNX-MLIR.

This PR adds such functionality. Currently, it is invoked only when there is no standard-translation available. This provides one way to translate "custom ops" whose semantics can be defined via ONNX functions.


src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -45,6 +48,7 @@ class FrontendGenImpl {
   }
 
   mlir::ModuleOp ImportONNXModel(const onnx::ModelProto &model) {
+    SetOpSetImport(model); // Determines which opsets to use.
```
doru1004, line 51: Could you give us a few more details on what this does and how it particularly helps with handling function ops?

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -66,16 +70,50 @@ class FrontendGenImpl {
       const onnx::NodeProto &);
 
   std::map<std::string, ImportHandlerType> import_handler_map_;
+  std::map<std::string, int64_t> opset_map_;
+  std::map<std::string, onnx::ValueInfoProto> value_info_map;
```
doru1004, line 74: Could you explains what these data structures do?  Would be nice to have a comment in the code about it.

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -66,16 +70,50 @@ class FrontendGenImpl {
...
+
+  void SetOpSetImport(const onnx::ModelProto &model) {
+    opset_map_.clear();
+    for (auto &binding : model.opset_import()) {
+      opset_map_[binding.domain()] = binding.version();
+    }
+  }
+
+  // TODO: Update ImportInputTensorSymbol to call this
+  void BindOnnxName(const std::string &onnx_name, mlir::Value symbol) {
```
doru1004, line 90: @chentong319 don't we have a similar mechanism in place already? Could this part use it instead of re-creating it?

doru1004:
@caoimhinuibrian @tjingrant @chentong319 please have a look at this patch.

doru1004:

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -20,6 +20,9 @@
 #include <mpark/variant.hpp>
 namespace bstd = mpark;
 
+#include "mlir/Ir/Module.h"
```
doru1004, line 23: I'm seeing this error:

```
/home/circleci/project/onnx-mlir/src/Builder/FrontendDialectTransformer.cpp:23:10: fatal error: mlir/Ir/Module.h: No such file or directory
 #include "mlir/Ir/Module.h"
```

src/Builder/FrontendDialectTransformer.cpp
gramalingam: I added comments to explain this. It essentially maintains a map from each op-set domain (like the standard ONNX domain) to the version of that domain (e.g., 11 or 12) used in the current model. It should be used to resolve any operator name (including function name).

src/Builder/FrontendDialectTransformer.cpp
gramalingam: Added comment to the source file.

src/Builder/FrontendDialectTransformer.cpp
gramalingam: Yes, but there is a minor difference (the other method takes a ValueInfo as a parameter). I have updated the other method to call this (more primitive method).

src/Builder/FrontendDialectTransformer.cpp
gramalingam: Fixed, thanks.

doru1004:

gramalingam:
I am looking into fixing the circleci failure. I am somewhat unclear about the dependencies between the libraries: e.g., it looks like code in Dialect/ONNX has dependencies on "Krnl"; is that correct? Ideally the ONNX dialect itself should not have dependencies on the Krnl dialect.

tjingrant:
> Ideally the ONNX dialect itself should not have dependencies on the Krnl dialect.

Looking at the problematic code, it looks like there's a helper function that is shared between shape inference and lowering because we recently unified shape inference and aspects of code-generation related to output memory allocation, they are now a bit tied together.

@AlexandreEichenberger I feels like you probably know what's the best way to avoid a dependency from ONNX dialect to Krnl dialect. It occurs to me that the function extracting dense attribute values from constant ops introduces a dependency to Krnl Dialect, i.e., as you lower, constants become krnl.global ops, and we extract constants from krnl.global during lowering whereas during shape inference, we extract them from onnx.Constant ops.

gramalingam:
On a slightly different note: I still don't understand why the build fails in CI. I am including OMLibs, which seems to contain all relevant libraries. Is there some other Krnl-related library that is not included in OMLibs? Thanks for any suggestions.

AlexandreEichenberger:
> > Ideally the ONNX dialect itself should not have dependencies on the Krnl dialect.
> 
> 
> @AlexandreEichenberger I feels like you probably know what's the best way to avoid a dependency from ONNX dialect to Krnl dialect. It occurs to me that the function extracting dense attribute values from constant ops introduces a dependency to Krnl Dialect, i.e., as you lower, constants become krnl.global ops, and we extract constants from krnl.global during lowering whereas during shape inference, we extract them from onnx.Constant ops.

@tjingrant This indeed was a cause of grief. The ONNXShapeHelper is a single piece of code that needs to read constants, and these constants are stored differently in shape inference and ONNX to Krnl lowering. I did not like this, but I did not know how to avoid this problem. If there was a way to keep operands that are literal tensors in the same format, that would be great. I believe I asked you about the "stuffing the constant tensor into the operation parameters" and you mentioned that it was not really used; maybe it would be a use case? What do you think? 

I see it used for gather (indices tensor), slice (start/end/step/axis)... and I am sure there are other places where the constant have to be read in order to define the shape and lowering to Krnl.

Looking forward to your suggestions.

AlexandreEichenberger:
We are looking into the failure of "Build x86 onnx-mlir on macOS / build-x86-onnx-mlir-macOS (pull_request)". It might be due to a mismatch of python levels. We are seeing this error in all of our PR, so its unlikely linked with your code change.

tjingrant:

gramalingam:
Great. Glad it is passing all tests now. Please let me know if you have any more questions about this PR or if it needs something more.

A question about the signoff: are the instructions shown in DCO valid? I think there could be some complications, since changes from master were merged into the PR. I don't want to confuse issues by signing off other merged commits.

tjingrant:

gramalingam:
Thanks. The DCO issue has been resolved. Please let me know if you have any questions or if this can be merged in.

AlexandreEichenberger (approved):
LGTM

tjingrant:

tjingrant:

tjingrant:
Sorry for the delay, just confirmed with @gongsu832 that the s390x-docker issue is expected for the time being.

PR #386, opened 2020-11-10
Make matmul work with dynamic dimensions

AlexandreEichenberger:
Currently, 3D and 4d test fails when forced to dynamic, e.g. using 
`IMPORTER_FORCE_DYNAMIC=1 BACKEND_TEST=test_matmul_4d_cpu make check-onnx-backend`

PR #395, opened 2020-11-13
Support equally spliting on unknown dimensions in the SplitOp lowering

tungld:
SplitOp lowering has already supported dynamic dimensions but missed the case of equally splitting where its split attribute is default.
 
This patch is to add support for that situation, making dynamic tests passed.


AlexandreEichenberger (approved):
LGTM, thanks for handling this. It was a small fix, so it is fine. You should really look at how simpler things become using the new IndexExpr approach. All of the special handling of static/dynamic is relegated to IndexExpr.  Look at a couple of recent PRs, e.g.
#386. All of the affine ops are overloaded so you get IndexExpr and then `(ii * steps) % n' works straight out of the box, generating literals when all are literals, affine maps when possible, and otherwise generic Value expr.. no need for you to manually do all this.

PR #397, opened 2020-11-16
Use IndexExpr to handle dynamic dimensions in lowering SplitOp

tungld:
This patch uses IndexExpr to handle dynamic dimensions in lowering SplitOp.

It also changes ONNXOpShapeHelper to support multiple outputs.  Using `getDimsForOutput(n)` to get and set IndexExpr(s) for the n-th output.


src/Dialect/ONNX/ONNXShapeHelper.hpp
```diff
@@ -45,12 +47,24 @@ struct ONNXOpShapeHelper {
     llvm_unreachable("implement in child structs");
   }
 
+  // Return output dims for the N-th output.
+  DimsExpr &getDimsForOutput(int n) {
```
AlexandreEichenberger, line 51: Nit: Get/Set goes typically in pairs, and here we have a get that is used for both as you return the reference. So maybe just name it 'dimsForOutput(int n)'. Note that I believe SmallVector already assert for out of bound accesses.

test/mlir/onnx/onnx_lowering.mlir
```diff
@@ -2744,41 +2744,41 @@ func @test_split_unknown_dimension_equal_split(%arg0 : tensor<?x?x64xf32>) -> (t
   %0, %1 = "onnx.Split"(%arg0) { axis = 1 : si64 } : (tensor<?x?x64xf32>) -> (tensor<*xf32>, tensor<*xf32>)
   "std.return"(%0, %1) : (tensor<*xf32>, tensor<*xf32>) -> ()
 
-  // CHECK: [[INDEX_MAP:#.+]] = affine_map<(d0, d1) -> (d0 + d1 ceildiv 2)>
+  // CHECK: [[ALLOC_MAP0:#.+]] = affine_map<(d0, d1) -> (d0 ceildiv 2)>
```
AlexandreEichenberger, line 2747: It is better to move the test to onnx_lowering_with_canonicalize as the map's math is moved into the code.

AlexandreEichenberger: Also, if you did the FileCheck format by hand, try my new `mlir2FileCheck` utility... it works quite well.

src/Dialect/ONNX/ONNXShapeHelper.cpp
```diff
@@ -364,5 +383,81 @@ LogicalResult ONNXMatMulOpShapeHelper::Compute(
...
+  // Save the final result.
+  getDimsForOutput(0) = outputDims;
+  return success();
+}
+
+//===----------------------------------------------------------------------===//
+// ONNX Split Op Shape Helper
+//===----------------------------------------------------------------------===//
+
+ONNXSplitOpShapeHelper::ONNXSplitOpShapeHelper(
```
AlexandreEichenberger, line 395: Glad it seems to have worked well for your example.

AlexandreEichenberger (commented):
Minor fixes, thanks for adding the multiple dim for output.

src/Dialect/ONNX/ONNXShapeHelper.hpp
tungld: Done.

test/mlir/onnx/onnx_lowering.mlir
tungld: `mlir2FileCheck` is quite useful. I used it. Thanks! However, it seemed it did not support local variables that have the same name, e.g. arguments of two krnl loops may use the same name. In the case of SplitOp, I manually changed variable names to make `mlir2FileCheck`work.

AlexandreEichenberger (approved):
LGTM

doru1004:
This should not have been merged since one of the tests was failing.

PR #401, opened 2020-11-16
Support lowering of krnl.getref with dynamic sizes

doru1004:
This patch enables the lowering of `krnl.getref` operations with dynamic sizes to LLVM.


AlexandreEichenberger (commented):
@doru1004 , can you leave a comment that explain the overall approach? I believe, when dyn getref you generate the data structure that LLVM expect from every memrefs, is that right?

Also, in the lit test, could you just comment "begin"/"end" of part associated with the getref? Some of the code appears to be normal "alloc" code, would be nice to be able to determine where we need to start to look at. Tx

doru1004:
> Also, in the lit test, could you just comment "begin"/"end" of part associated with the getref? Some of the code appears to be normal "alloc" code, would be nice to be able to determine where we need to start to look at. Tx

I have this comment:

```
/// Definition of the krnl.getref output memref<2x?xf32>
```

Which marks the paragraph in the test which handles the getref.

Is that what you're looking for?

doru1004:
> can you leave a comment that explain the overall approach?

There's not much to explain other than the current comments which are inline. It's just a lowering of the MemRef to the LLVM struct that represents a MemRef. What am I missing?

AlexandreEichenberger (approved):
LGTM, if you find issues with the lit tests on s390, consider using the mlir2FileCheck of PR #410 that adds the CHECK-DAG when statements can be reordered.

Issue #400, opened 2020-11-16
Dynamic shape for concat, reshape and conv


chentong319:
These operation can accept dynamic shape for certain input or certain dimension of the tensor. I tested using PR390 to select the input.
## reshape and conv:  OK
Only the first input can be dynamic. All the backend test cases for two operations passed with`IMPORTER_FORCE_DYNAMIC=0`, which made only the first input dynamic.

## concat: failed
Concat operation requires all the inputs have the same shape except the dimension to be concatenated. Since current compiler only change the first dimension to dynamic, I can test those cases that concatenate the first dimension. 
Shape inference for concat requires that the concatenated dimension has to be constant and failed the compiler.  It seems to me that this check should be removed. I checked the code of lowering for this operation and the dynamic size is not handled.

## new error exposed
With only one input is changed to dynamic dimension, a bunch of test, such as add, div, sub, min and max, failed with " 'affine.load' op index must be a dimension or symbol identifier".  The lowering result for test_add_cpu is listed below. The `select` may be the source.
```
module {
  func @main_graph(%arg0: memref<?x4x5xf32>, %arg1: memref<3x4x5xf32>) -> memref<3x4x5xf32> attributes {input_names = ["x", "y"], output_names = ["sum"]} {
    %0 = alloc() : memref<3x4x5xf32>
    %c0 = constant 0 : index
    %1 = dim %arg0, %c0 : memref<?x4x5xf32>
    %c1 = constant 1 : index
    %2 = cmpi "eq", %1, %c1 : index
    %3:3 = krnl.define_loops 3
    krnl.iterate(%3#0, %3#1, %3#2) with (%3#0 -> %arg2 = 0 to 3, %3#1 -> %arg3 = 0 to 4, %3#2 -> %arg4 = 0 to 5) {
      %c0_0 = constant 0 : index
      %4 = select %2, %c0_0, %arg2 : index
      %5 = affine.load %arg0[%4, %arg3, %arg4] : memref<?x4x5xf32>
      %6 = affine.load %arg1[%arg2, %arg3, %arg4] : memref<3x4x5xf32>
      %7 = addf %5, %6 : f32
      affine.store %7, %0[%arg2, %arg3, %arg4] : memref<3x4x5xf32>
    }
    return %0 : memref<3x4x5xf32>
  }
  "krnl.entry_point"() {func = @main_graph, numInputs = 2 : i32, numOutputs = 1 : i32} : () -> ()
}
```

[...]

PR #402, opened 2020-11-17
Fix checking unknown dimensions for broadcasting

tungld:
This is partially related to #400 


chentong319 (approved):
The change looks correct to me.
Can we figure out the performance impact? The condition of the select is loop invariant. Not sure whether there is performance difference if IF statement with two loop nests is generated. In that case, the load in either loop nest is an affine load.

tungld:
@chentong319 Right, IF statement with two loops would be better for performance. Will consider to use IF statement in another patch. Thanks!

chentong319:

chentong319:
@tungld When @AlexandreEichenberger  tests test_add_cpu with only the first operand changed to dynamic, the test failed. I confirmed this error.
The test_add.mlir is:
```
module {
  func @main_graph(%arg0: tensor<?x4x5xf32>, %arg1: tensor<3x4x5xf32>) -> tensor<3x4x5xf32> attributes {input_names = ["x", "y"], output_names = ["sum"]} {
    %0 = "onnx.Add"(%arg0, %arg1) : (tensor<?x4x5xf32>, tensor<3x4x5xf32>) -> tensor<3x4x5xf32>
    return %0 : tensor<3x4x5xf32>
  }
  "onnx.EntryPoint"() {func = @main_graph, numInputs = 2 : i32, numOutputs = 1 : i32} : () -> ()
}
```
The lowering output from `onnx-mlir-opt --convert-onnx-to-krnl` is:
```
module {
  func @main_graph(%arg0: memref<?x4x5xf32>, %arg1: memref<3x4x5xf32>) -> memref<3x4x5xf32> attributes {input_names = ["x", "y"], output_names = ["sum"]} {
    %0 = alloc() : memref<3x4x5xf32>
    %c0 = constant 0 : index
    %1 = dim %arg0, %c0 : memref<?x4x5xf32>
    %c1 = constant 1 : index
    %2 = cmpi "eq", %1, %c1 : index
    %3:3 = krnl.define_loops 3
    krnl.iterate(%3#0, %3#1, %3#2) with (%3#0 -> %arg2 = 0 to 3, %3#1 -> %arg3 = 0 to 4, %3#2 -> %arg4 = 0 to 5) {
      %c0_0 = constant 0 : index
      %4 = select %2, %c0_0, %arg2 : index
      %5 = affine.load %arg0[%4, %arg3, %arg4] : memref<?x4x5xf32>
      %6 = affine.load %arg1[%arg2, %arg3, %arg4] : memref<3x4x5xf32>
      %7 = addf %5, %6 : f32
      affine.store %7, %0[%arg2, %arg3, %arg4] : memref<3x4x5xf32>
    }
    return %0 : memref<3x4x5xf32>
  }
  "krnl.entry_point"() {func = @main_graph, numInputs = 2 : i32, numOutputs = 1 : i32} : () -> ()
}
```
Notice that both load is Affine.load, which causes trouble in later stage. The first load is indexed from the result of a `select` operation.  Could you take a look?

AlexandreEichenberger:
@chentong319 Good Catch... 

@tungld  By the way, if you were to use the index expression, the system does that for you.... if the expression going into the indices of a load are affine, affine.load; otherwise, std.load.

tungld:
@chentong319 will fix this problem here: #428.
@AlexandreEichenberger I will move to index expression in near future, it takes a bit time to write index expression for broadcasting.

PR #403, opened 2020-11-17
[WIP] Initial version of support-instance-normalization

xintin:


xintin:
@tjingrant Request you to review. Thanks.

tjingrant:
@xintin , we usually include two kinds of test to enable a new operation:
- LIT test (e.g., https://github.com/onnx/onnx-mlir/blob/3404af0201da167abfde3eb57dad83cd0891a303/test/mlir/onnx/onnx_lowering.mlir#L3135), where we verify that the IR generated by the compiler for lowering an operation is as expected. You may find documentation on FileCheck useful (https://llvm.org/docs/CommandGuide/FileCheck.html).
- ONNX backend test (https://github.com/onnx/onnx-mlir/blob/3404af0201da167abfde3eb57dad83cd0891a303/test/backend/test.py#L129). This is for checking whether the values produced by running this op implementation conforms with values produced by ML frameworks like PyTorch. Simply search for relevant tests using the name of the operation you're enabling, and then add their name into test_to_enable variable in test.py and the tests will run on our CI.

PR #406, opened 2020-11-18
check-onnx-backend-dynamic

chentong319:
Add the script to conduct End-to-End test with dynamic shape.

- Dynamic test is triggered by option `--dynamic` to test/backend/test.py. The arguments is parsed and removed so that unitest can do its own argument parsing.

- By default, the dynamic test changes the first dimension of all inputs to -1 (namely ?) in tensor shape. The default value can be changed by adding the desired value for the test case in either dictionary `dynamic_input_dict` or `dynamic_dim_dict`

- Since cmake does not have clear support to add env variable for a command execution, the env variables are set at subprocess.run

- A target check-onnx-backend-dynamic is added. It runs test.py with `--dynamic` option.

- check-onnx-backend-dynamic is called to `.circleci/config.yml`


AlexandreEichenberger (approved):
LGTM, like the solution

doru1004:
One test is failing due to failure in master.

AlexandreEichenberger:
Failure is unrelated, seen in all of the other PRs

PR #409, opened 2020-11-18
Print error message.

tjingrant:


doru1004 (approved):
LGTM

gongsu832:
@tjingrant Can we also print out the _path of the command that's being executed?

PR #410, opened 2020-11-19
added support for DAG in mlir2FileCheck

AlexandreEichenberger:

Add DAG support. Script does def use analysis and insert DAG whenever there are operations that could go in parallel due to lack of def-use chains that would otherwise serialize them.

```
//CHECK-LABEL: func @test_gather_axis0neg
//CHECK-SAME:  ([[PARAM_0_:%.+]]: memref<3x2xf32>) -> memref<2x2x2xf32> {
//CHECK-DAG:    [[CST_3_:%.+]] = constant 3 : index
//CHECK-DAG:    [[CST_0_:%.+]] = constant 0 : index
//CHECK-DAG:    [[RES_:%.+]] = alloc() : memref<2x2x2xf32>
//CHECK-DAG:    [[VAR_1_:%.+]] = "krnl.global"() {name = "constant_0", shape = [2, 2], value = dense<{{.}}[0, -1], [1, 2]{{.}}> : tensor<2x2xi64>} : () -> memref<2x2xi64>
//CHECK-DAG:    [[LOOP_0_:%.+]]:3 = krnl.define_loops 3
//CHECK:        krnl.iterate([[LOOP_0_]]#0, [[LOOP_0_]]#1, [[LOOP_0_]]#2) with ([[LOOP_0_]]#0 -> [[I_0_:%.+]] = 0 to 2, [[LOOP_0_]]#1 -> [[I_1_:%.+]] = 0 to 2, [[LOOP_0_]]#2 -> [[I_2_:%.+]] = 0 to 2) {
//CHECK:        [[LOAD_VAR_1_MEM_:%.+]] = affine.load [[VAR_1_]][symbol([[I_0_]]), symbol([[I_1_]])] : memref<2x2xi64>
//CHECK:        [[VAR_4_:%.+]] = index_cast [[LOAD_VAR_1_MEM_]] : i64 to index
//CHECK-DAG:    [[VAR_5_:%.+]] = addi [[VAR_4_]], [[CST_3_]] : index
//CHECK-DAG:    [[VAR_6_:%.+]] = cmpi "slt", [[VAR_4_]], [[CST_0_]] : index
//CHECK-NOT: separator of consecutive DAGs
//CHECK-DAG:    [[VAR_7_:%.+]] = select [[VAR_6_]], [[VAR_5_]], [[VAR_4_]] : index
//CHECK-DAG:    [[CST_3_1_:%.+]] = constant 3 : index
//CHECK-DAG:    [[CST_0_1_:%.+]] = constant 0 : index
//CHECK:        [[VAR_8_:%.+]] = load [[PARAM_0_]]{{.}}[[VAR_7_]], [[I_2_]]{{.}} : memref<3x2xf32>
//CHECK:        affine.store [[VAR_8_]], [[RES_]][symbol([[I_0_]]), symbol([[I_1_]]), symbol([[I_2_]])] : memref<2x2x2xf32>
//CHECK:        }
//CHECK:        return [[RES_]] : memref<2x2x2xf32>
//CHECK:        }
```

### DIFF
--- a/test/backend/test.py
+++ b/test/backend/test.py
@@ -14,10 +14,17 @@ from onnx.backend.base import Device, DeviceType
 import subprocess
 import test_config
 import tempfile
+import argparse
 
 VERBOSE = bool(os.environ.get("VERBOSE"))
 TEST_DYNAMIC = os.environ.get("IMPORTER_FORCE_DYNAMIC")
     
+parser = argparse.ArgumentParser(description='with dynamic shape or not.')
+parser.add_argument('--dynamic', action='store_true',
+    help='enable dynamic (default: false)')
+parser.add_argument('unittest_args', nargs='*')
+args = parser.parse_args()
+sys.argv[1:] = args.unittest_args
 
 TEST_CASE_BY_USER = os.environ.get("BACKEND_TEST")
 if TEST_CASE_BY_USER is not None and TEST_CASE_BY_USER != "" :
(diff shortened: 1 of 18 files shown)


AlexandreEichenberger:
Sorry I am adding things to the same PR, but since its sooooo slow to get them through, I don't have the patience to have them individually. I added the following:

1. Attempt to fix the MacOS check, which Tian think is due to a mismatch of python support and cached LLVM. So I bumped the count in the script
1. It was really hard to keep track of which backend operation is tested because the ops where not listed alphabetically, and we had a inclusion list for static case, and a separate exclusion list for the dynamic case. So I made a single map where the key is the benchmark, and the value is which benchmark is supported: static and/or dynamic. I kept the same interface by translating the dictionary into the arrays that @chentong319 was using before, so minimal need of changes.

```
test_disabled = 0
test_static = 1
test_dynamic = 2
test_static_dynamic = test_static + test_dynamic

test_to_enable_static_dynamic = {

    ############################################################
    # Elementary ops, ordered alphabetically.

    # Abs
    "test_abs_cpu": test_static_dynamic,

    # Acos

    # Acosh

    # Adagrad

    # Adam

    # Add
    "test_add_cpu": test_static_dynamic,
    "test_add_bcast_cpu": test_static_dynamic,

...


    ############################################################
    # Model (alphabetical order)

    "test_shufflenet_cpu": test_static,
    "test_resnet50_cpu": test_static,
    "test_vgg19_cpu": test_static,
}

# test for static
test_to_enable = [ key for (key, value) in test_to_enable_static_dynamic.items() if value & test_static ]
```

I added comment covering every ONNX backend file that currently exist, so that as we add them, they already have a "home".

I noted that there are many operations that we support (presumably) but not test. We should make a separate PR that add all the successful tests for ops that we are currently not testing.

doru1004 (approved):
LGTM

test/backend/test.py
```diff
@@ -181,763 +181,569 @@ def supports_device(cls, device):
...
+}
+
+# test for static
+test_to_enable = [ key for (key, value) in test_to_enable_static_dynamic.items() if value & test_static ]
+
+# Specify the test cases which currently can not pass for dynamic shape
+# Presumably, this list should be empty
+# Except for some operation too difficult to handle for dynamic shape
+# or big models
+test_not_for_dynamic = [ key for (key, value) in test_to_enable_static_dynamic.items() if value == test_static ]
```
chentong319, line 745: Should test_disabled be considered:  `if value == test_static or value == test_disabled`
Or do not define test_disabled at all.

AlexandreEichenberger: Good point, will update in the next round.

AlexandreEichenberger: But maybe we could switch to a positive logic, add a variable that includes just the tests to run the dynamic, as opposed to list of tests that cannot go dynamic :-)

PR #411, opened 2020-11-20
Replace the tanh lowering by using std.tanh

tungld:
TanhOp is in the std dialect, so we lower ONNXTanhOp to TanhOp instead of manually compute tanh using exp.

### DIFF
--- a/test/mlir/onnx/onnx_lowering.mlir
+++ b/test/mlir/onnx/onnx_lowering.mlir
@@ -215,14 +215,22 @@ func @test_tanh(%arg0 : tensor<?x10xf32>) -> tensor<*xf32> {
   // CHECK: [[C0_0:%.+]] = constant 0 : index
   // CHECK: [[DIM_2:%.+]] = dim %arg0, [[C0_0]] : memref<?x10xf32>
   // CHECK: krnl.iterate([[DEF_LOOPS]]#0, [[DEF_LOOPS]]#1) with ([[DEF_LOOPS]]#0 -> %arg1 = 0 to [[DIM_2]], [[DEF_LOOPS]]#1 -> %arg2 = 0 to 10) {
-  // CHECK: [[LOAD:%.+]] = affine.load %arg0[%arg1, %arg2] : memref<?x10xf32>
-  // CHECK: [[ZERO:%.+]] = constant {{0.+}} : f32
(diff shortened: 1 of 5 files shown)


test/mlir/onnx/onnx_lowering.mlir
```diff
@@ -215,14 +215,22 @@ func @test_tanh(%arg0 : tensor<?x10xf32>) -> tensor<*xf32> {
...
   // CHECK: krnl.iterate([[DEF_LOOPS]]#0, [[DEF_LOOPS]]#1) with ([[DEF_LOOPS]]#0 -> %arg1 = 0 to [[DIM_2]], [[DEF_LOOPS]]#1 -> %arg2 = 0 to 10) {
-  // CHECK: [[LOAD:%.+]] = affine.load %arg0[%arg1, %arg2] : memref<?x10xf32>
-  // CHECK: [[ZERO:%.+]] = constant {{0.+}} : f32
-  // CHECK: [[NLOAD:%.+]] = subf [[ZERO]], [[LOAD]] : f32
-  // CHECK: [[EXP:%.+]] = exp [[LOAD]] : f32
-  // CHECK: [[NEXP:%.+]] = exp [[NLOAD]] : f32
-  // CHECK: [[DIVIDEND:%.+]] = subf [[EXP]], [[NEXP]] : f32
-  // CHECK: [[DIVISOR:%.+]] = addf [[EXP]], [[NEXP]] : f32
-  // CHECK: [[TANH:%.+]] = divf [[DIVIDEND]], [[DIVISOR]] : f32
+  // CHECK: [[X:%.+]] = affine.load %arg0[%arg1, %arg2] : memref<?x10xf32>
```
AlexandreEichenberger, line 218: While it appears to have worked so far, I also cases where the order of operations was switched on s390. It may have been due to the phases called, not sure.

In general, the testing would be more resilient if we do not use strict ordering between consecutive operations without dependences (like the two constant below). The latest mlir2FileCheck.py automatically detect the "parallel" operations and insert CHECK-DAG to reflect that.

AlexandreEichenberger (approved):
LGTM, in the future (or now, your choice), it may be good to use CHECK-DAG for independent operations. mlir2FileCheck.py now add these also automatically.

PR #412, opened 2020-11-20
Enable some dynamic backend tests

tungld:
Enable dynamic backend tests for some operations.

### DIFF
--- a/.github/workflows/build-x86-onnx-mlir-macOS.yml
+++ b/.github/workflows/build-x86-onnx-mlir-macOS.yml
@@ -21,7 +21,7 @@ jobs:
         cache-name: cache-mlir-directory
       with:
         path: ~/work/onnx-mlir/llvm-project
-        key: V7-${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/clone-mlir.sh', '**/build-mlir.sh') }}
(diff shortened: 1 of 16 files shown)


chentong319 (commented):
For unsqueezed, can the second input, `axis`, be dynamic? I guess all the test cases have a constant for axis. For us, better to specify that only the first input to be dynamic in `dynamic_input_dict` for unsqueeze tests.

AlexandreEichenberger:
@tungld The error on macOS is not the same as the one we use to have and ignore; its in the lit tests, for krnl_permute. Please look into it, keeping in mind that it might be simply a problem of operation order (which might be the case as it works on the other platforms but not that one).

AlexandreEichenberger:
@tungld Just a reminder to look at the failure for MacOS, as the previous issue is fixed for that test, and this error is different. Tx

tungld:
> For unsqueezed, can the second input, axis, be dynamic? 

@chentong319 In onnx-mlir, unsqueeze's axes is an attribute, not an input. It's different from unsqueeze in the latest onnnx version: https://github.com/onnx/onnx-mlir/blob/master/src/Dialect/ONNX/ONNXOps.td.inc#L5675

@AlexandreEichenberger I cannot reproduce the problem on my MacOs. Just tried to trigger the test again to see whether it still happens or not.

tungld:
Enable dynamic backend test for LSTM/GRU/ RNN also.

AlexandreEichenberger (approved):
LGTM

PR #413, opened 2020-11-23
Add backend tests

AlexandreEichenberger:
Added backend tests of operations that we had, but not tested. This is a complement of #412.

Operations that don't work for static/dynamic are:
* constant
* shape
* size

These look like strange errors, as I would have expected them to work. Maybe they all have to do with constants, could it be that the compilation for the backend tests does not allow our constant scheme? @tjingrant ?


doru1004 (approved):
LGTM

tjingrant:
I think an interesting issue is that Size operation fails because the lowering involves using output shape to alloc output memref, e.g. [here](https://github.com/onnx/onnx-mlir/blob/4cc16aceb73ca401f0e692c4cebbdcba97f5b089/src/Conversion/ONNXToKrnl/Tensor/Size.cpp#L39). This seems to be fine with static shape, since the shape info is known fully at compile-time; but it's probably wrong to do so with dynamic shape.

AlexandreEichenberger:
> I think an interesting issue is that Size operation fails because the lowering involves using output shape to alloc output memref, e.g. [here](https://github.com/onnx/onnx-mlir/blob/4cc16aceb73ca401f0e692c4cebbdcba97f5b089/src/Conversion/ONNXToKrnl/Tensor/Size.cpp#L39). This seems to be fine with static shape, since the shape info is known fully at compile-time; but it's probably wrong to do so with dynamic shape.

Tian, looking at the def of size:

```
Size

Takes a tensor as input and outputs a int64 scalar that equals to the total number of elements of the input tensor.
```

isn't the output always a single int ?

chentong319:
The implementation of Size is incorrect at the allocate for dynamic. I can rewrite it with IndexExpr.

chentong319:
@AlexandreEichenberger @tjingrant I just realized that the size op failed for static. That bug code for allocate does not matter, even for dynamic because that branch should not be executed. The error for static shape may come after lowering to krnl. I will take a investigation.

chentong319:
The test `make BACKEND_TEST=test_size_cpu check-onnx-backend-dynamic` passed. This indicates that the error is highly possible related to constant handling.

PR #415, opened 2020-11-24
Consolidate dockerfiles

gongsu832:
- add Dockerfile for building user image
- consolidate prereq and build Dockerfiles for different archs
- rename x86 back to amd64 to conform to docker arch naming


gongsu832:
@tjingrant @caoimhinuibrian it looks like travis amd64 build fails because it's getting a bad onnxmlirczar/onnx-mlir-llvmimage:amd64 that was built at some point? Not sure why the stage job didn't rebuild and push it.

gongsu832:
@tjingrant @caoimhinuibrian it looks like travis amd64 build fails because it's getting a bad onnxmlirczar/onnx-mlir-llvmimage:amd64 that was built at some point? Not sure why the stage job didn't rebuild and push it.

gongsu832:
test this please

tjingrant:
Trying to see if problem persists after merging with latest changes from master.

tjingrant:
@gongsu832 would it help if i delete the old amd64 image?

tjingrant:
BTW, to rebuild & push to Dockerhub the new docker images, you need to push this branch to the main repo (as opposed to being a branch in your forked repo); this allows you to access the Dockerhub token. Then, you can push an empty commit with the commit message being "Rebuild prereq docker." to retrigger a pre-req image rebuild, after rebuilds succeed they'll get pushed to Dockerhub automatically.

Sign-off message can mess up those magic words, so please don't sign off the commit with the magic message ("Rebuild prereq docker.").

caoimhinuibrian:
@tjingrant I did this and I've replaced it ---- going to rerun

gongsu832:
> BTW, to rebuild & push to Dockerhub the new docker images, you need to push this branch to the main repo (as opposed to being a branch in your forked repo); this allows you to access the Dockerhub token. Then, you can push an empty commit with the commit message being "Rebuild prereq docker." to retrigger a pre-req image rebuild, after rebuilds succeed they'll get pushed to Dockerhub automatically.
> 
> Sign-off message can mess up those magic words, so please don't sign off the commit with the magic message ("Rebuild prereq docker.").

@tjingrant Thanks. Now I remember seeing you giving the same advise to someone else. I've always wondered why it's so cumbersome to push images. Combined with the resource restriction which leads to very slow build, I think having our own Jenkins build server may be easier to manage things. I mentioned in our Tue. internal meeting that there is an open source PowerCI cluster we can potentially use. I signed up for an account but they say it will take 7 days so we will see how that goes.

@caoimhinuibrian thanks for triggering the rerun. It appears that the Travis amd64 build now succeeds.

docker/Dockerfile.prereq
```diff
@@ -52,6 +53,6 @@ RUN if [ ! -f "/build/llvm-project/build/CMakeCache.txt" ]; then \
     fi
 
 ENV MAKEFLAGS=-j2
-RUN if timeout 30m cmake --build . --target -- ${MAKEFLAGS} ; then \
-      cmake --build . --target check-mlir; \
+RUN if cmake --build . --target -- ${MAKEFLAGS} ; then \
```
tjingrant, line 56: Hi @gongsu832 this timeout is important because Travis has a 60m time limit for each job. So to workaround this timelimit, I broke down the llvm-project compilation into several stages, each stage is roughly doing "make all" for a maximum of 30 min, and then it gets killed, the intermediate builds are sent to DockerHub, and the next Travis job picks up from where the previous one lefts off from DockerHub. After a few iterations, we will have a complete llvm/mlir build on DockerHub.

So I don't think we can omit the timeout.

tjingrant:
> @tjingrant Thanks. Now I remember seeing you giving the same advise to someone else. I've always wondered why it's so cumbersome to push images. To be honest, even with your instructions, I'm not confident that I'll get it right. Combined with the resource restriction which leads to very slow build, I think having our own Jenkins build server may be easier to manage things. I mentioned in our Tue. internal meeting that there is an open source PowerCI cluster we can potentially use. I signed up for an account but they say it will take 7 days so we will see how that goes.

I agree. So does the Power CI cluster include a x86 machine as well? 

For the time being, I think we should nonetheless push this branch to under the main repo, and test whether semi-automatically building llvm-project and updating them to DockerHub still works.

docker/Dockerfile.prereq
gongsu832: Good catch. Yes I see it's needed for the 5-stage build. I was playing around and thought I added it back but apparently forgot it. Thanks.

gongsu832:
> > @tjingrant Thanks. Now I remember seeing you giving the same advise to someone else. I've always wondered why it's so cumbersome to push images. To be honest, even with your instructions, I'm not confident that I'll get it right. Combined with the resource restriction which leads to very slow build, I think having our own Jenkins build server may be easier to manage things. I mentioned in our Tue. internal meeting that there is an open source PowerCI cluster we can potentially use. I signed up for an account but they say it will take 7 days so we will see how that goes.
> 
> I agree. So does the Power CI cluster include a x86 machine as well?
> 
> For the time being, I think we should nonetheless push this branch to under the main repo, and test whether semi-automatically building llvm-project and updating them to DockerHub still works.


Unfortunately I don't think the PowerCI cluster has x86 machines. But getting an x86 machine is relatively easy. If it turns out that we can indeed use the PowerCI for our own Jenkins server, I can look for an x86 machine. Then we could use our own Jenkins on all 3 archs which should make things a lot easier to deal with.

tjingrant:
Closing in favor of the upstream & tested version of this patch. #417

PR #416, opened 2020-11-25
#3 --- Fix the dynamic dimension issue when lowering TransposeOp

tungld:


src/Conversion/ONNXToKrnl/Tensor/Transpose.cpp
```diff
@@ -19,30 +19,38 @@ struct ONNXTransposeOpLowering : public ConversionPattern {
...
+      }
+      alloc = rewriter.create<AllocOp>(loc, memRefType, allocOperands);
+
+      // Insert dealloc if necessary.
+      auto *parentBlock = alloc.getDefiningOp()->getBlock();
+      if (insertDealloc) {
+        auto dealloc = rewriter.create<DeallocOp>(loc, alloc);
+        dealloc.getOperation()->moveBefore(&parentBlock->back());
+      }
+    }
```
chentong319, line 53: These lines are correct in semantics, but duplicated with shape inference and insertAllocAndDeallocSimple. This is a perfect example to use IndexExpr.

tungld: @chentong319 Done. Thanks!

chentong319 (approved):
LGTM!

PR #417, opened 2020-11-28
[Upstream] Consolidate dockerfiles

tjingrant:
Originally https://github.com/onnx/onnx-mlir/pull/415 authored by @gongsu832 , pushed to upstream for testing and debugging purpose.


tjingrant:
@gongsu832 mlir test suite is run and returns non-zero value on s390x.

gongsu832:
@tjingrant This is bizarre. I suspect the CPU_ARCH isn't getting passed in. Can you add a line like this:
```
diff --git a/docker/Dockerfile.prereq b/docker/Dockerfile.prereq
index f716142..baa6d63 100644
--- a/docker/Dockerfile.prereq
+++ b/docker/Dockerfile.prereq
@@ -54,5 +54,6 @@ RUN if [ ! -f "/build/llvm-project/build/CMakeCache.txt" ]; then \

 ENV MAKEFLAGS=-j2
 RUN if timeout 30m cmake --build . --target -- ${MAKEFLAGS} ; then \
+      echo "check-mlir on CPU_ARCH=${CPU_ARCH}"; \
       cmake --build . --target check-mlir || [ "${CPU_ARCH}" = "s390x" ]; \
     fi
```
I fetched your branch and tried to push the change but git says I'm not on a branch. Grrrrh.

tjingrant:
@gongsu832 did I do anything wrong? Not seeing the debug output.

gongsu832:
@tjingrant It appears that Travis is not running mlir-check at all on all platforms. I don't expect adding the `ENV CPU_ARCH ${CPU_ARCH}` would change anything but we can narrow things down by removing it and see what happens.

tjingrant:
> @gongsu832 did I do anything wrong? Not seeing the debug output.

Sorry, I was a bit confused, it did run check-mlir, but only on the final stage where make actually succeeded (as opposed to being killed due to using more than 30min), which is expected.

gongsu832:
@tjingrant It looks like Travis docker build is broken and not passing in the CPU_ARCH. Sigh...I don't know how to fix this.
```
[311/311] Linking CXX static library lib/libbenchmark_main.a
check-mlir on CPU_ARCH=
[1/30] Building CXX object tools/mlir/unittests/Analysis/Presburger/CMakeFiles/MLIRPresburgerTests.dir/MatrixTest.cpp.o
```

AlexandreEichenberger (commented):
Probably a very naive question: in the .travis.yml, I see a lot of repeating commands for each arch? I though the goal was to have one command with parametrized parameters, such as arch (already there) and target

onnxmlirczar/onnx-mlir-llvmimage-partial:s390x 
onnxmlirczar/onnx-mlir-llvmimage-partial:ppc64le
onnxmlirczar/onnx-mlir-llvmimage-partial:amd64

tjingrant:
> Probably a very naive question: in the .travis.yml, I see a lot of repeating commands for each arch? 

Yeah, I looked at ways to remove duplication of commands in the yaml file, but meta-programming the yaml file is hacky and pretty un-readable, so I gave up.

> I though the goal was to have one command with parametrized parameters, such as arch (already there) and target

The goal is to combine Dockerfiles for each arch into one that is shared across all archs.

AlexandreEichenberger (approved):
LGTM, readability is good too

gongsu832:
@tjingrant I guess you probably already tried this. You can't just list the 3 archs in the `arch` key for each stage in the jobs? Like:
```
jobs:
...
  include:
  - stage: prereq-init
     if: commit_message = "Rebuild prereq docker."
    arch:
    - s390x
    - ppc64le
    - amd64
    ...
```

tjingrant:
@gongsu832 yes I tried that, the job matrix is only expanded for the main job (the one with the name test).

gongsu832:
@tjingrant According to this https://docs.travis-ci.com/user/multi-cpu-architectures/ the matrix should also be available for includes although it's still in beta. Perhaps you also tried it already. In any case, I really hope the PowerCI thing works out so we don't have to deal with all these mess.

AlexandreEichenberger:
@tjingrant clean run, I will let you or @gongsu832 pull the plug

tjingrant:
> @tjingrant According to this https://docs.travis-ci.com/user/multi-cpu-architectures/ the matrix should also be available for includes although it's still in beta. Perhaps you also tried it already. In any case, I really hope the PowerCI thing works out so we don't have to deal with all these mess.

I'm somewhat ambivalent about spending more time working on a cosmetic one-off fix; Travis has been like a blackhole sucking away my time every time I try to debug something related to it. And yes, I'm looking forward to the PowerCI thing too.

PR #418, opened 2020-11-30
Make initializedTensors a class member

gramalingam:
Change global variable "initializedTensors" into a  member variable of FrontendGenImpl.

The current implementation does not allow a user to translate multiple ONNX models into MLIR one after another (because the initializers of earlier translations are carried over to subsequent translations). So, cannot use a test-suite with multiple translations, one after another. This is a minimal fix to enable that.


gramalingam:
Any feedback on this PR? It is a one-line change and should be easy to review, appreciate your help, thanks!

doru1004 (approved):
LGTM

gramalingam:
Is this okay to merge now?

tjingrant:

PR #421, opened 2020-12-01
Rewrite GlobalAvaragePool and GlobalMaxPool

tungld:
This patch rewrites GlobalAvaragePool and GlobalMaxPool into ReduceMean and ReduceMax, respectively.


AlexandreEichenberger (approved):
LGTM, elegant solution.

PR #423, opened 2020-12-01
Fix of Index Expr, more concise ONNXOps.cpp pattern, spelling fixes

AlexandreEichenberger:
Implementation of global average and max pool, and some updates to make code more concise.


AlexandreEichenberger:
Global Pools were done simultaneously in a different patch by @tungld , so I decided to use his, but there were several changes that I wanted to keep. So I retooled this PR to do anything but the Global Pools... Minor changes, some spelling mistakes. One thing is that I had a trick to prevent doing constants again and again in IndexExpr for 0, -1, 1, but it was not enabled. I changed the code to make it work.

doru1004 (approved):
LGTM

PR #424, opened 2020-12-01
#4 lower shape

chentong319:

Modified on top of PR#306, which is too old to be merged automatically.


AlexandreEichenberger (commented):
Tong, I would like you to add a test of shape for a memref that has a map that changes the size:

(d0, d1) -> (d0 floordiv 8, d1 floordiv 8, 8, 8)

My understanding is that @doru1004 added the krnl.dim for handling the sizes in presence of such maps. Check your new example, see that the shape remains the logical shape (not the mapped shape). If it does not work, check with lowering to krnl.dim.

chentong319:
@AlexandreEichenberger Should all the lowering code be change to use krnl.dim? If krnl.dim is to ignore the map, it should be test separately, not in lowering of every operations.

src/Conversion/ONNXToKrnl/Tensor/Shape.cpp
```diff
@@ -0,0 +1,68 @@
...
+    // Iterate along the data shape storing dim value to result.
+    for (int i = 0; i < dataRank; i++) {
+      // Create store index value.
+      Value storeVal;
+      Value storeIndex =
+          emitConstantOp(rewriter, loc, rewriter.getIndexType(), i);
+
+      // Checking for dynamic dimensions.
+      if (dataShape[i] == -1) {
+        Value shapeVal = rewriter.create<DimOp>(loc, data, storeIndex);
```
AlexandreEichenberger, line 51: Though we would use the krnl.dim ops

AlexandreEichenberger: @chentong319 any action on this one?

chentong319: I will push new commit after PR#443 is merged.

AlexandreEichenberger: Tong, I though we would use the Krnl.Dim ops instead of the std ones.

chentong319: The new version uses IndexExpr, which, I assume, uses Krnl.Dim.

AlexandreEichenberger (approved):
LGTM

src/Conversion/ONNXToKrnl/Tensor/Shape.cpp
AlexandreEichenberger: @chentong319  Busted... it is not; will make a patch to correct this.

chentong319: We may need two interfaces: logic dim size or allocated dim size.

PR #426, opened 2020-12-02
Avoid concurrency issue.

tjingrant:


doru1004 (approved):
LGTM

I see a fail with one of the docker tests, is that expected?

tjingrant:
@doru1004 yes it's exepceted, we're in the middle of working on the Dockerfiles.

PR #428, opened 2020-12-02
Enhance IMPORTER_FORCE_DYNAMIC and fix dynamic dims issues

tungld:
This patch changes the environment variable IMPORTER_FORCE_DYNAMIC and fixes issues about supporting dynamic dims (for variadic ops, and for second/third/... dynamic dimensions).

For the former one, the patch enhances the use of IMPORTER_FORCE_DYNAMIC, where it allows defining multiple inputs and multiple dimensions that can be changed to unknown dimensions, e.g.
```
1. IMPORTER_FORCE_DYNAMIC='-1:-1'
    - change all dimensions in all inputs to unknown dimensions.
2. IMPORTER_FORCE_DYNAMIC='-1:0'
   - change the first dimension in all inputs to unknown dimensions.
3. IMPORTER_FORCE_DYNAMIC='1:-1'
  - change all dimensions in the second input to unknown dimensions.
4. IMPORTER_FORCE_DYNAMIC='1:0,1'
  - change the first and second dimensions in the second input to unknown dimensions.
5. IMPORTER_FORCE_DYNAMIC='0:1|1:0,1'
  - change the second dimension in the first input to unknown dimensions, and
  - change the first and second dimensions in the second input to unknown
 dimensions,
```
By changing IMPORTER_FORCE_DYNAMIC, `test/backend/test.py` was changed where we now have only a single dict of test names, `test_to_enable_static_dynamic`. The single dict is the union of three dicts: `dynamic_input_dict`, `dynamic_dim_dict`, and `test_to_enable_static_dynamic`.

In the single dict, we can define a test like `"test_add_cpu": (test_static_dynamic,{0: {0}, 1:{1}}),` to say we want to test Add op whose the first dim of the first input and the second dim of the second input are dynamic.

The use of only single dict is also to avoid mistakes in inputting test names to the three dicts. `dynamic_input_dict` and `dynamic_dim_dict` use test names without postfix "_cpu" while `test_to_enable_static_dynamic` uses test names with "_cpu". This happened to me when I tested LSTM with dynamic dims. I wrongly used `test_lstm_with_initial_bias_cpu` in  `dynamic_input_dict` and `dynamic_dim_dict`, so tests for LSTM passed.

~~It seems there were tests failed if setting the second/third/... dimension to unknown. The ones I know they failed are:~~(FIXED)
- ~~IMPORTER_FORCE_DYNAMIC="0:1" BACKEND_TEST=test_matmul_2d_cpu make check-onnx-backend~~
- ~~IMPORTER_FORCE_DYNAMIC="0:1" BACKEND_TEST=test_matmul_3d_cpu make check-onnx-backend~~
- ~~IMPORTER_FORCE_DYNAMIC="0:2" BACKEND_TEST=test_matmul_4d_cpu make check-onnx-backend~~
- ~~All gemm test cases with IMPORTER_FORCE_DYNAMIC="0:1"~~
- ~~Relu with IMPORTER_FORCE_DYNAMIC="0:1"~~

~~An interesting issue is that Relu failed even though it is quite simple and its generated code looks correct. Nothing special to handle unknown dimensions except using “std.alloc” with dim.~~

~~While matmul 2d and 3d failed with the second dimension unknown, matmul 4d did not but with the third dimension unknown.~~


chentong319:
@tungld When I worked on concat, I also ran into the similar problem: backend test failed for dynamic size for second or third dimension of 3D tensor. All other combinations passed. I can not tell what's wrong at MLIR level IR and am working at the llvm level IR.

src/MainUtils.cpp
```diff
@@ -448,10 +448,12 @@ void addONNXToKrnlPasses(mlir::PassManager &pm) {
...
-  pm.addNestedPass<FuncOp>(mlir::createKrnlBundleMemoryPoolsPass());
-  pm.addPass(mlir::createCanonicalizerPass());
-  pm.addNestedPass<FuncOp>(mlir::createKrnlOptimizeMemoryPoolsPass());
+  // These passes causes error when compling LSTM/GRU/RNN ops with unknown dims.
+  // Temporary disable them for testing purpose.
+  // pm.addNestedPass<FuncOp>(mlir::createKrnlEnableMemoryPoolPass());
+  // pm.addNestedPass<FuncOp>(mlir::createKrnlBundleMemoryPoolsPass());
+  // pm.addPass(mlir::createCanonicalizerPass());
+  // pm.addNestedPass<FuncOp>(mlir::createKrnlOptimizeMemoryPoolsPass());
   pm.addPass(mlir::createCanonicalizerPass());
```
chentong319, line 457: Will these changes go into the patch or do not run those test cases until the source of error is fixed?

chentong319 (commented):
It seems that some changes are caused by clang-format.  What's the reason?

test/backend/test.py
```diff
@@ -45,177 +45,6 @@
...
-        d = Device(device)
-        if d.type == DeviceType.CPU:
-            return True
-        return False
-
-
-backend_test = onnx.backend.test.BackendTest(DummyBackend, __name__)
-
 # Test directories:
 # https://github.com/onnx/onnx/tree/master/onnx/backend/test/data/node
```
AlexandreEichenberger, line 49: Thanks for unifying the dict, much better to have one repo. Please add a quick comment on the format, presumably
{arg:{which dim to make unknown} }

Are the -1 allowed there? (just a question, not a request to add)

AlexandreEichenberger (commented):
This is a great help, thanks.

Will accept once the new IMPORTER format is also described here:

https://github.com/onnx/onnx-mlir/blob/master/docs/Testing.md

Thanks

test/backend/test.py
tungld: Added a comment for this. Yes, -1 is allowed, meaning all inputs or all dimensions.

src/MainUtils.cpp
tungld: I narrowed down the issue to the BundleMemoryPools pass and reported here #431. I am not sure which one to go. It seems to me there is an issue in BundleMemoryPools pass rather than in LSTM lowering code, since LSTM tests passed when the pass is turned off.

tungld:
> It seems that some changes are caused by clang-format. What's the reason?

Fixed this. Thanks!

src/MainUtils.cpp
chentong319: I suggest you leave the LSTM out of the dynamic test currently but do not change the intended passes.

tungld: OK. I updated according to your comments. Thanks!

tungld:
@AlexandreEichenberger 

> Will accept once the new IMPORTER format is also described here:
https://github.com/onnx/onnx-mlir/blob/master/docs/Testing.md

Done this. Thanks!

tungld:
> @tungld When I worked on concat, I also ran into the similar problem: backend test failed for dynamic size for second or third dimension of 3D tensor. All other combinations passed. I can not tell what's wrong at MLIR level IR and am working at the llvm level IR.

Fixed this problem.

AlexandreEichenberger (approved):
LGTM, big thanks for spotting the issue with https://github.com/onnx/onnx-mlir/issues/438

chentong319 (approved):
All the changes look good to me, except OMTensor.inc. I am not familiar with that code.   I guess the change is related to the strides value got from narray. It would be better to change the name of this set function as something like SetStrideWithNarrayStride. We should add comments for the semantics of the fields of tensor ( around line 97).

PR #430, opened 2020-12-02
generate ONNX Dialect op for custom node

caoimhinuibrian:
This patch takes an unknown op and turns it into an ONNX Dialect custom operation. ONNXCustomOp ... later this will be lowered to a function call.


caoimhinuibrian:
from this input:
ir_version: 6
producer_name: "pytorch"
producer_version: "1.5"
graph {
  node {
    output: "1"
    name: "Constant_0"
    op_type: "Constant"
    attribute {
      name: "value"
      t {
        dims: 4
        data_type: 7
        raw_data: "\000\000\000\000\000\000\000\000\001\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\001\000\000\000\000\000\000\000"
      }
      type: TENSOR
    }
  }
  node {
    input: "0"
    input: "1"
    output: "2"
    name: "NewOp_1"
    op_type: "NewOp"
  }
  name: "torch-jit-export"
  input {
    name: "0"
    type {
      tensor_type {
        elem_type: 7
        shape {
          dim {
            dim_value: 4
          }
        }
      }
    }
  }
  output {
    name: "2"
    type {
      tensor_type {
        elem_type: 7
        shape {
          dim {
            dim_value: 4
          }
        }
      }
    }
  }
}
opset_import {
  version: 11
}


we get this output:
func @main_graph(%arg0: tensor<4xi64>) attributes {input_names = ["0"], output_names = ["2"]} {
  %0 = "onnx.Constant"() {onnx_node_name = "Constant_0", value = dense<[0, 1, 0, 1]> : tensor<4xi64>} : () -> tensor<4xi64>
  %1 = "onnx.Custom"(%arg0, %0) {function_name = "NewOp", onnx_node_name = "NewOp_1"} : (tensor<4xi64>, tensor<4xi64>) -> tensor<4xi64>
  return %1 : tensor<4xi64>
}

src/Dialect/ONNX/AdditionalONNXOps.td.inc
```diff
@@ -0,0 +1,24 @@
+def ONNXCustomOp:ONNX_Op<"Custom",
```
AlexandreEichenberger, line 1: .inc files are reserved for auto-generated files. We used to have an additional file for the extra ops not in the standard. Looks like we need it again

AlexandreEichenberger (approved):
LGTM after renaming of output file and possibly changes of return value for shape inference.

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -674,6 +678,36 @@ class FrontendGenImpl {
...
           "represents a custom operator.");
+    mlir::emitWarning(UnknownLoc(), "Could not find op importer: assuming this "
+                                    "represents a custom operator.");
+    int nOps = node.input().size();
+    auto funcName = opName.str();
+    std::vector<mlir::Type> outputTypes;
+    std::vector<mlir::Value> inputs;
+    auto attributes = ImportNodeAttributes(node);
+    auto mlirAttr = builder_.getStringAttr(funcName);
+    auto funcAttr = builder_.getNamedAttr("function_name", mlirAttr);
```
gramalingam, line 687: It would be useful to include the domain-name of the op also as an attribute.

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -674,6 +678,36 @@ class FrontendGenImpl {
...
+    attributes.push_back(funcAttr);
+    //auto customOp = builder_.create<ONNXCustomOp>(UnknownLoc(), outputTypes, inputs, attributes);
+    int nIn = 0;
+    int nOut=0;
+    //mlir::Value customResult = *(customOp.getODSResults(0).begin());
+    //std::vector<mlir::Value> inputs;
+    for (const auto &item : node.input()) {
+      ++nIn;
+      if (initializedTensors.ContainKey(legalize_name(item))) {
```
gramalingam, line 696: Nit: it may be worthwhile extracting the logic in lines 424 to 432 as a reusable method and call it here. That will also help handle optional inputs. It looks like this logic is repeated in several places.

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -838,6 +872,7 @@ class FrontendGenImpl {
     builder_.create<mlir::ReturnOp>(UnknownLoc(), ret_vals);
     // Update main function signature to reflect types of newly imported
     // output tensors.
+    mainFunc.dump();
```
gramalingam, line 873: Was this accidentally included?

caoimhinuibrian: yep ... will remove

src/Dialect/ONNX/AdditionalONNXOps.td.inc
caoimhinuibrian: done

src/Builder/FrontendDialectTransformer.cpp
caoimhinuibrian: done

src/Builder/FrontendDialectTransformer.cpp
caoimhinuibrian: yep. done.

PR #432, opened 2020-12-03
[Fix] Use pass-local maps when bundling memory pools.

doru1004:
Adding and deleting functions from the functions map requires thread-safe insertion and deletion.


AlexandreEichenberger (approved):
LGTM

tjingrant:
I wonder if these data structures need to be shared across threads. If not, a better solution might be to make them member variables of the Pass, and make them thread-local. It seems like for `OptimizeMemoryPools.cpp`, they can be thread-local given that the data structure is cleared at the end of pass execution.

AlexandreEichenberger (approved):
LGTM:

for ref:

>https://mlir.llvm.org/docs/PassManagement/#operation-pass
>
>Specifically:
>
> Maintain any global mutable state, e.g. static variables within the source file. All mutable state should be maintained by an instance of the pass.​

PR #433, opened 2020-12-04
[DEBUG] [DO NOT MERGE] Do not dealloc shape/strides.

tjingrant:


AlexandreEichenberger (commented):
@tjingrant Are we creating a memory leak? Or is it because we sometimes make a shallow copy of these shape and strides? I feel uncomfortable with this solution long-term.

> My guess is that Test GRU/LSTM/RNN did not pass in shapes/strides correctly (they need to be heap variables because they're freed when tensors get destroyed). Given the relatively small size of these arrays (shape/strides), it might be easier to make a heap copy of these variables when creating new OMTensors.

I agree that making a fool-proof interface is better... namely allocating data when creating the data so that the data of the OMTensor is fully owned by it. Its better if the user fail to dealloc its structure (and resulting into a leak in his/her code) rather than having an interface where (because we don't know if the user is doing the right thing or not) we have a permanent leak.

You may remember that I was trying to get a clear understanding of who owned what data exactly because of these potential issues.

tjingrant:
@AlexandreEichenberger I'm simply debugging using our CI.. For some reason, I cannot trigger the same error directly on the s390x machine.

AlexandreEichenberger:
@tjingrant makes perfect sense, thanks for the clarification.

tjingrant:
Not particularly convinced that this is the issue, trying another approach: https://github.com/onnx/onnx-mlir/pull/436 .

Issue #431, opened 2020-12-03
Error in bundling memory pool


tungld:
I have an LSTM program generated by onnx-mlir up to `enable-memory-pool` pass enabled, but it did not pass the `bundle-memory-pools` pass. Without `bundle-memory-pools`, the program can be compiled and run correctly.

This is the program:
```mlir
#map0 = affine_map<(d0, d1, d2) -> (d0, d1, d2)>
#map1 = affine_map<() -> (0)>
#map2 = affine_map<() -> (1)>
#map3 = affine_map<()[s0] -> (s0)>
#map4 = affine_map<() -> (3)>
#map5 = affine_map<(d0, d1) -> (d0, d1)>
#map6 = affine_map<(d0)[s0, s1] -> (d0 + s0 * s1)>
#map7 = affine_map<() -> (2)>
#map8 = affine_map<() -> ()>
module {
  %0 = "krnl.packed_const"() {file_name = "/tmp/packed_const-77d95f.tmp", is_le = true, size_in_bytes = 0 : i64} : () -> i64
  func @main_graph(%arg0: memref<?x?x2xf32>, %arg1: memref<1x12x2xf32>, %arg2: memref<1x12x3xf32>) -> memref<1x?x3xf32> attributes {input_names = ["X", "W", "R"], output_names = ["Y"]} {
    %c0 = constant 0 : index
    %c2 = constant 2 : index
    %c1 = constant 1 : index
    %cst = constant 1.000000e+00 : f32
    %cst_0 = constant 2.000000e+00 : f32
    %cst_1 = constant 0.000000e+00 : f32
    %c4 = constant 4 : index
    %c3 = constant 3 : index
    %c0_i64 = constant 0 : i64
    %1 = dim %arg0, %c1 : memref<?x?x2xf32>
    %2 = alloc(%1) : memref<1x?x3xf32>
    %3 = dim %arg0, %c1 : memref<?x?x2xf32>
    %4 = muli %3, %c4 : index
    %5 = muli %4, %c3 : index
    %6 = alloc(%5) : memref<?xi8>
    %7 = "krnl.getref"(%6, %c0_i64, %3) : (memref<?xi8>, i64, index) -> memref<1x?x3xf32>
    %8:3 = krnl.define_loops 3
[...]

PR #434, opened 2020-12-04
[Fix] Fix the bundling for LSTM operation

doru1004:
This patch is a fix to issue: https://github.com/onnx/onnx-mlir/issues/431


doru1004:
@tungld for some reason I still can't add you as reviewer.

This patch should fix the error you reported about bundling.

tungld:
LGTM! It fixed the issue. Thanks!

test/mlir/krnl/krnl_bundle_lstm_memory_pool.mlir
```diff
@@ -0,0 +1,346 @@
...
+    }
+    dealloc %6 : memref<?xi8>
+    return %2 : memref<1x?x3xf32>
+  }
+  "krnl.entry_point"() {func = @main_graph, numInputs = 3 : i32, numOutputs = 1 : i32} : () -> ()
+}
+
+// CHECK-LABEL: bundle_lstm_dyn_mem_pools
+
+// CHECK: [[C0:%.+]] = constant 0 : index
```
AlexandreEichenberger, line 269: I see a lot of operations that can go in parallel... It would be safer to use the CHECK-DAG

AlexandreEichenberger (approved):
LGTM, I would prefer seen more CHECK-DAG in the mlir files as order has been seen as changing on some platforms.

test/mlir/krnl/krnl_bundle_lstm_memory_pool.mlir
doru1004: Right, the only part which could be DAG'ed are the constants, otherwise for all the other operation the order is important.

PR #439, opened 2020-12-07
Install libs

gongsu832:
- "make install" should install onnx-mlir libraries
- set ONNX_HOME for check-onnx-backend otherwise downloaded models are stored in user's home directory


AlexandreEichenberger:
@gongsu832 , could you indicate where the "installed" libraries are going.
Couldn't the downloaded models go by default somewhere under the build/test/backend sub dirs

gongsu832:
@AlexandreEichenberger The installed libraries go into `/usr/local/lib` typically unless the user changes it with `CMAKE_INSTALL_PREFIX`.
And yes, by setting ONNX_HOME=${CMAKE_CURRENT_BINARY_DIR} the downloaded models will be stored under `build/test/backend/models` instead of user's home directory by default.

tjingrant:
@gongsu832 feels like it will make our cmake script less repetitive if we iterate over the OMLibs variable and install all targets within it?

gongsu832:
@tjingrant that would be another approach although I think they both have pros and cons. The main difference between the two is basically whether the installation is done in a separate place or it's localized to individual cmake files where the library is built. Typically, I'd prefer to have the cmake logic that builds and installs a library localized in a single cmake file, which is why I did it the current way. But in our case, when we add a new library, besides the individual cmake file for building and installing the library, we need to update the OMLibs anyway. So having some logic to loop over OMLibs would also be OK.

Overall, I don't think either of the approach has overwhelming advantages over the other so I'm fine with either.

tjingrant (approved):
LGTM!

PR #441, opened 2020-12-08
#2 --- [Fix] Remove mutex and use pass-local map when optimizing memory pools

doru1004:
This patch removes the mutex introduced previously and replaced the global map with a pass-local map.


AlexandreEichenberger (approved):
LGTM

doru1004:
@AlexandreEichenberger This one is further along than the first patch in line, it should finish soon.

PR #443, opened 2020-12-09
#1 ---[Fix] Global emission uses malloc when returned by a function

doru1004:
This patch returns the result of a malloc when global value is returned.


src/Conversion/ONNXToKrnl/Tensor/Constant.cpp
```diff
@@ -25,6 +39,8 @@ struct ONNXConstantOpLowering : public ConversionPattern {
     auto loc = op->getLoc();
     auto constantOp = llvm::dyn_cast<ONNXConstantOp>(op);
 
+    printf("Operation is returned: %d\n", checkOpResultIsReturned(&constantOp));
+
```
chentong319, line 43: print for debugging only?

doru1004: done

chentong319 (approved):
Can you enable the test for SizeOp in backend (uncomment line 573 in test/backend/test.py)? This patch should make these test go through. Otherwise, the patch is good.

doru1004:
> Can you enable the test for SizeOp in backend (uncomment line 573 in test/backend/test.py)? This patch should make these test go through. Otherwise, the patch is good.

Just enabled them.

PR #444, opened 2020-12-09
Use IndexExpr to compute divisor for ReduceMean

tungld:


AlexandreEichenberger (commented):
@tungld I wonder if you would not want to use IndexExpr on this one. This is precisely why I wrote it, not to have lengthy special code for testing for constants.

tungld:
> @tungld I wonder if you would not want to use IndexExpr on this one. This is precisely why I wrote it, not to have lengthy special code for testing for constants.

Oh, right. We can use IndexExpr whenever we have index-related calculation. Will update this patch now.

tungld:
@AlexandreEichenberger I changed the patch to use IndexExpr. Thanks for your comments!

AlexandreEichenberger (approved):
LGTM, thanks

PR #446, opened 2020-12-09
Use type/shape information from input ONNX model for import

gramalingam:
(1) Introduce options to allow callers to customize translation of ONNX model to ONNX-MLIR
(2) Add option to use type/shape information from input ONNX model for import

An ONNX model allows type-annotations to be stored for all variables in a model (including intermediate temporaries). ONNX type/shape inference can produced a fully annotated ONNX model. The translation of such a fully-annotated model can use the type/shape information contained in the input ONNX model. This is useful in cases where ONNX-MLIR does not have a corresponding type/shape inference method (eg. for custom-ops) or for cases where there is a mismatch between the type/shape inference methods.

The test-case in this PR shows how this can be used with custom-ops.

### DIFF
--- a/test/onnx2mlir/CustomFnTest.cpp
+++ b/test/onnx2mlir/CustomFnTest.cpp
@@ -16,6 +16,8 @@ using namespace ONNX_NAMESPACE;
 #define ONNX_OPSET_VERSION 11
 
 void RegisterFunSchema() {
+  static bool registered = false;
+  if (registered) return;
   ONNX_NAMESPACE::OpSchema schema;
   schema.SetName("SquareFn")
       .SetDomain(ONNX_DOMAIN)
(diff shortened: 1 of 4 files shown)


test/onnx2mlir/CustomFnTest.cpp
```diff
@@ -91,6 +89,59 @@ int main(int argc, char *argv[]) {
...
+  node = graph->add_node();
+  node->add_input("t");
+  node->add_output("y");
+  node->set_op_type("CustomFn2");
+
+  check(model_proto);
+}
+
+int main(int argc, char *argv[]) {
+  // testCustomFunTranslation();
```
tjingrant, line 143: I'm guessing this test should be enabled?

tjingrant (approved):
LGTM, just double-checking whether you actually intend to disable the test case.

test/onnx2mlir/CustomFnTest.cpp
gramalingam: Yes, will do that. I think I did that because of the use of a static-variable that caused problems with doing two different translations, one after another (fixed in a different PR https://github.com/onnx/onnx-mlir/pull/418 ).

gramalingam: Thanks. This looks ready to be merged in (I guess the linux-s390x CI is not required, is it?)

tjingrant:

PR #447, opened 2020-12-09
Parse Cast op's "To" attribute as mlir type, to avoid onnx dialect depends on onnx

souptc:
Currently when convert from ONNX model to ONNX dialect, the "Cast" op's "To" attribute is using the default behavior that parsed as int attribute, which will depends on the ONNX to interpret what does that int attribute means, for example, 1 is mapping to fp32. 
This make the ONNX dialect not self-contain, the consumer need to take a dependency of onnx to interpret it. It also increase the risk that if the consumer take a different version of onnx that the onnx dialect is genereate, we may run into trouble. 

This PR change the Cast op's parser behavior to parse the "To" attribute as mlir type, then the onnx dialect is self-contained, and we can get rid of the onnx dependency.


tjingrant:
Thanks @souptc , the CircleCI failure is caused by ONNX Dialect documentation being deprecated by your change. You can update it by running (in the build directory):
```
cmake --build . --target onnx-mlir-doc
# Copy generated documentation to the corresponding places in the source tree.
cp -r docs/Dialects ../docs/Dialects
```

souptc:
Thanks, will do

souptc:
Hi, i run that command and update the dialect doc, but the cirleci still fail. could you help to double check?

alberto-magni:
Hi @souptc, the reason why the doc check still fails is likely because a new file onnx.md has been placed in folder docs/Dialects/Dialects/ rather than updating the existing one in docs/Dialects/. 
Hope this helps.

tjingrant:
@souptc just a heads up, your patch didn't pass the clang-format bot test. Could you run clang-format on your patch? 

Format file is here:
https://github.com/onnx/onnx-mlir/blob/master/.clang-format

tjingrant:

souptc:
It seems hit a docker issue:
"docker: Error response from daemon: Conflict. The container name "/onnx-mlir-test" is already in use by container"

anything i need to do for that?

tjingrant:
@souptc I think linux-s390x check is more comprehensive than the linux-s390x-docker. We're in the middle of upgrading our docker-based jenkins build system, so this issue is probably not worth fixing anyways.

tjingrant (approved):
LGTM!

souptc:
thank you!

PR #448, opened 2020-12-11
ShapeHelper for broadcasting operations using IndexExpr

tungld:
This patch uses IndexExpr for broadcasting operations such as binary ops, variadic ops that support multi-directional broadcasting.

### DIFF
--- a/src/Dialect/ONNX/ONNXShapeHelper.hpp
+++ b/src/Dialect/ONNX/ONNXShapeHelper.hpp
@@ -63,6 +63,41 @@ private:
   SmallVector<DimsExpr, 1> outputsDims;
 };
 
+/// Compute a broadcasted shape from the shapes of given operands. Operands must
+/// be ranked in advance.
+struct ONNXOpBroadcastedShapeHelper {
+  ONNXOpBroadcastedShapeHelper(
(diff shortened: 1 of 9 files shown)


src/Dialect/ONNX/ONNXShapeHelper.hpp
```diff
@@ -63,6 +63,41 @@ struct ONNXOpShapeHelper {
   SmallVector<DimsExpr, 1> outputsDims;
 };
 
+/// Compute a broadcasted shape from the shapes of given operands. Operands must
+/// be ranked in advance.
+struct ONNXOpBroadcastedShapeHelper {
```
AlexandreEichenberger, line 68: My first reaction would be to have this as a subclass of the main ShapeHelper class; thinking a bit more, maybe it does not really have to, because I don't really think we will use them in a generic way. So its probably fine as is.

AlexandreEichenberger (approved):
LGTM, let me know if using IndexExpr for a bit now makes you want changes in its interface. For example, I believe we only need Affine Expr and not Semi-affine expressions. When taking semi affine out of the equations (like I think we are), there is no need to distinguish between dimensions and symbols... So if that is the case, we could remove the distinction of it in our interface as well...

PR #452, opened 2020-12-15
Support casting from integer to integer

tungld:

Add support for casting from integer to integer. No backend test is added because there is no backend test for this type of casting in ONNX.


tjingrant (approved):
LGTM!

PR #457, opened 2020-12-29
Update onnx commit

gongsu832:
Update onnx commit to include Haruki san's big endian fix for numpy_helper.py


tjingrant (commented):
LGTM!

tjingrant (approved):
LGTM! 

Ideally we probably want to stick to only the release version of ONNX, but that's not possible because onnx v1.8 introduces spec changes that cause our backend tests to fail. Need to support opset versioning at some point.

gongsu832:
@tjingrant Yes and unfortunately the big endian fix was not backported to official v1.7. Hopefully once we support v1.8 this would be a nonissue.

PR #458, opened 2021-01-01
Bugfix loop op

tjingrant:
Changes:
- Supports pushing and popping variable scopes in the frontend symbol table, this is useful for managing nested variable scopes, something required by `Loop` operation importing.
- Supports `HasOnnxSubgraphOpInterface` interface, which allows for accessing op-associated region by name.
- Update `gen_onnx_mlir` to provide implementation for `HasOnnxSubgraphOpInterface` interface.
- Supports importing subgraph attribute to regions associated with an operation; I also refactored the code so that this new function shares much code with the main computation graph importing procedure.
- Update recursive shape inference mechanism, such that shape inference operation is given a handler of the shape inference function to invoke on its sub-regions.

Nits:
- Removed a bunch of redundant header guards in TableGen.


docs/Dialects/onnx.md
```diff
@@ -4253,6 +4240,27 @@ ONNX Resize operation
 | :----: | ----------- |
 `Y` | tensor of 8-bit unsigned integer values or tensor of 16-bit unsigned integer values or tensor of 32-bit unsigned integer values or tensor of 64-bit unsigned integer values or tensor of 8-bit signless integer values or tensor of 16-bit signless integer values or tensor of 32-bit signless integer values or tensor of 64-bit signless integer values or tensor of 16-bit float values or tensor of 32-bit float values or tensor of 64-bit float values or tensor of stirng type values or tensor of 1-bit signless integer values or tensor of complex type with 32-bit float elements values or tensor of complex type with 64-bit float elements values or memref of any type values
 
+### `onnx.Return` (::mlir::ONNXReturnOp)
```
AlexandreEichenberger, line 4243: quick question: these changes come from upgrading onnx to a new version?

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -592,7 +592,7 @@ ONNXEntryPointOp ONNXEntryPointOp::create(mlir::Location location,
 /// Infer the output shape of the ONNXExpOp. This method is required by the
 /// shape inference interface.
 LogicalResult ONNXExpOp::inferShapes(
-    std::function<void(mlir::FuncOp)> shapeInferenceFunc) {
+    std::function<void(mlir::Region &)> shapeInferenceFunc) {
```
AlexandreEichenberger, line 595: minor: maybe rename parameter to shapeInferenceRegion

AlexandreEichenberger (approved):
LGTM, would appreciate a few slides to understand impact of loops on the rest of optimizations and how to deal with them.

docs/Dialects/onnx.md
tjingrant: Standard return op specifies that its parent must be the mlir::FuncOp, so we need a different terminator op for subgraph (imported as region now).
https://github.com/llvm/llvm-project/blob/dfb558b82d3f2ca228799b03d2be00b163d04ea5/mlir/include/mlir/Dialect/StandardOps/IR/Ops.td#L2295

src/Dialect/ONNX/ONNXOps.cpp
tjingrant: Maybe doShapeInference? The Func suffix is there to suggests it's something callable.

AlexandreEichenberger: sure

AlexandreEichenberger:
@tjingrant looks like a failure associated with krnl load/store.

PR #460, opened 2021-01-06
Dynamic backend test s390x

gongsu832:
Enable backend tests with dynamic shape on s390x Jenkins CI

### DIFF
--- a/test/backend/test.py
+++ b/test/backend/test.py
@@ -30,13 +30,16 @@ TEST_DYNAMIC = os.getenv("TEST_DYNAMIC")
 parser = argparse.ArgumentParser(description='with dynamic shape or not.')
 parser.add_argument('--dynamic', action='store_true',
     default=(strtobool(TEST_DYNAMIC) if TEST_DYNAMIC else False),
-    help='enable dynamic (default: false)')
+    help='enable dynamic shape tests (default: false if TEST_DYNAMIC env var not set)')
 parser.add_argument('-i', '--input', type=int,
     default=os.getenv("TEST_INPUT", -1),
-    help='input whose dimensions to be changed to unknown (default: all inputs')
+    help='inputs whose dimensions to be changed to unknown (default: all inputs if TEST_INPUT env var not set)')
 parser.add_argument('-d', '--dim', type=int,
     default=os.getenv("TEST_DIM", -1),
-    help='dimension to be changed to unknown (default: all dimension')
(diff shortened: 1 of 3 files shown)


chentong319 (approved):
LGTM

test/backend/test.py
```diff
@@ -16,21 +17,31 @@
...
-VERBOSE = bool(os.environ.get("VERBOSE"))
-TEST_DYNAMIC = os.environ.get("IMPORTER_FORCE_DYNAMIC")
-    
+# Casting with "bool" does not work well. When you specify VERBOSE=xxx,
+# regardless of the value of xxx (e.g., true, false, y, n, etc.) the
+# casted bool value will be true. Only if xxx is empty, the casted bool
+# value will be false. This is a bit counter intuitive. So we use strtobool
+# to do the conversion. But note that strtobool can't take an emtpy string.
+
+VERBOSE = os.getenv("VERBOSE")
```
tjingrant, line 26: Missing a conversion to strtobool?

test/backend/test.py
```diff
@@ -16,21 +17,31 @@
...
+TEST_DYNAMIC = os.getenv("TEST_DYNAMIC")
+
 parser = argparse.ArgumentParser(description='with dynamic shape or not.')
 parser.add_argument('--dynamic', action='store_true',
+    default=(strtobool(TEST_DYNAMIC) if TEST_DYNAMIC else False),
     help='enable dynamic (default: false)')
-parser.add_argument('-i', '--input', type=int, default=-1,
+parser.add_argument('-i', '--input', type=int,
+    default=os.getenv("TEST_INPUT", -1),
     help='input whose dimensions to be changed to unknown (default: all inputs')
```
tjingrant, line 36: Seems like the help info needs an update, same with the other argument.

tjingrant (commented):
Some minor comments.

test/backend/test.py
gongsu832: Not here but when the VERBOSE is used in `execute_commands`.

test/backend/test.py
gongsu832: Not sure. The default values didn't change. They are still -1 for both. I just added the env vars as an extra check if the command line arguments are not specified. Feel free to make changes.

tjingrant: My suggestion was to advertise this feature to users through the help message, since `--help` output is the most useful documentation for features. Feel free to ignore if this is only intended for a specific purpose you have in mind.

PR #461, opened 2021-01-06
Fix bundling

doru1004:
Take into consideration affine loads and krnl dim operations.


doru1004:
@tungld this should fix the problem you were experiencing with your model. Let me know if it now works for you.

tungld:
> @tungld this should fix the problem you were experiencing with your model. Let me know if it now works for you.

It works now. Thanks!

AlexandreEichenberger (approved):
@doru1004 thanks for the fix; @tungld thanks for checking that it works.

PR #462, opened 2021-01-07
Support MemRefType in constant propagation pass

tungld:
This patch allows doing constant propagation with MemRefType (only RankedTensorType was supported before).
A new LIT test for memref is added by duplicating the original lit test, where `tensor` is replaced by `memref`.


src/Dialect/ONNX/ONNXOpsHelper.cpp
```diff
@@ -87,3 +87,25 @@ bool getIntegerLiteralFromValue(Value value, int64_t &intLit) {
...
+
+//===----------------------------------------------------------------------===//
+// Get a broadcasted type for RankedTensorType and MemRefType.
+//===----------------------------------------------------------------------===//
+Type getBroadcastedRankedType(Type type1, Type type2) {
+  if (type1.isa<RankedTensorType>() && type2.isa<RankedTensorType>())
+    return OpTrait::util::getBroadcastedType(type1, type2);
+  if (type1.isa<MemRefType>() && type2.isa<MemRefType>()) {
+    // Contruct RankedTensorType(s).
+    auto elementType = type1.cast<MemRefType>().getElementType();
```
AlexandreEichenberger, line 99: If you don't mind having explicit types here, would be appreciated.

AlexandreEichenberger (approved):
LGTM, thanks for the additional support. Minor nit, if you could remove the "auto" in the new code that you wrote, MLIR folks prefer actual types unless they are very complicated, and this kind of make sense.

src/Dialect/ONNX/ONNXOpsHelper.cpp
tungld: Done. Thanks!

PR #459, opened 2021-01-05
Update LLVM commit to 496fb70b

NathanielMcVicar:
Support LLVM commit 496fb70b141ccbfaba9761294f3b4b97717096a3

- Many changes related to LLVMType (starting with https://reviews.llvm.org/D93680)
- OpBuilder deprecated in some cases (https://reviews.llvm.org/D90038)
- Context to start of param list (https://reviews.llvm.org/D93432)
- Type classes to BuiltinTypes.h (https://reviews.llvm.org/D92435)
- Required libs for ArmNeon and AVX (https://reviews.llvm.org/D92171, https://reviews.llvm.org/D92614)
- Headers for greedy patterns (https://reviews.llvm.org/D89103)
- alloc() uses new sizeBytes functionality of getMemRefDescriptorSizes (https://reviews.llvm.org/D91549)

Patch primarily by: Stella Stamenova <stilis@microsoft.com>

PR #463, opened 2021-01-08
Do not call ShapeHelper Compute functions within assert

yangchen-MS:
ShapeHelper's Compute functions such as ONNXOpShapeHelper::Compute
and ONNXOpBroadcastedShapeHelper::Compute have side-effects.
They cannot be placed within an ``assert'' macro, which becomes a
no-op in the release build. Consequently, we would end up with not
calling these functions at all and hitting subsequent failures.

### DIFF
--- a/src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
+++ b/src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
@@ -616,7 +616,9 @@ struct ONNXElementwiseBinaryOpLowering : public ConversionPattern {
 
     // Shape helper.
     ONNXOpBroadcastedShapeHelper shapeHelper(&rewriter, loc);
-    assert(succeeded(shapeHelper.Compute(operands)));
+    bool computeSucceeded = succeeded(shapeHelper.Compute(operands));
+    (void)computeSucceeded;
+    assert(computeSucceeded && "Failed to compute shape!");
     IndexExprContext outerContext(shapeHelper.context);
 
     // Insert an allocation and deallocation for the result of this operation.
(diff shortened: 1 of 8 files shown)


tjingrant:
Hi @yangchen-MS thanks for the patch, I believe this has already been addressed in another patch from Microsoft by @NathanielMcVicar :
https://github.com/onnx/onnx-mlir/pull/459 .

yangchen-MS:
> 
> 
> Hi @yangchen-MS thanks for the patch, I believe this has already been addressed in another patch from Microsoft by @NathanielMcVicar :
> #459 .

@tjingrant Thank you for pointing it out. I wasn't aware of that. Closing the PR since it has been taken care of. Thanks for reviewing the code, Tian.

KeDengMS:
Should we separate the bug fix from LLVM updates?

tjingrant:
@KeDengMS I agree it's ideal to have each patch address a distinct issue; it's not strictly enforced only because some collaborators point out that their PR submission requires a lengthy internal review (not sure whether it's the case for #459). 

Furthermore, due to the way our CI is set up (with LLVM as a base docker image), it'll be a complicated process to merge any patch before #459 goes in, because the new LLVM base image has been pushed to Dockerhub, and any patch based on the current master will fail our Docker-based CI. Thus it's worth considering that breaking up the patch won't expedite things.

With such understanding, if you have a compelling reason to break up the patch, it'll probably be best to chat with @NathanielMcVicar .

NathanielMcVicar:
@KeDengMS I did it as part of the LLVM patch because it was causing our internal builds that I was using to verify the correctness of that patch to fail. If you would prefer it separately I don't have any objections to that, but I will second the note that the CI system makes that cumbersome (same reason for splitting the loop lowering tests, even though it's probably not correct from a software engineering perspective).

KeDengMS:
@tjingrant @NathanielMcVicar thanks for the explanation. I didn't realize that other PRs are blocked by #459 because of the CI setup cannot select LLVM base docker image. Please go ahead with #459 as one patch then.

PR #464, opened 2021-01-08
Add build instructions for Big Sur

AlexandreEichenberger:
Added instructions for macOS Big Sur. Its currently an additional option to the cmake, eventually we may want to migrate this to the cmake logic, if need be.

No code was changed, just one md file.


tjingrant (approved):
LGTM.

PR #466, opened 2021-01-11
Additional ops

AlexandreEichenberger:
Added shape/lower/lit tests/backend tests for ceil/floor/sin, shape/lower/lit tests for atan (mising std to llvm lowering it seems)

Powf is described in MLIR but is not in our current version. I suspect that it will become avail once we upgrade to the latest MLIR. So there is some code for it but I disabled it by returning an "unavailable" error when activated.


AlexandreEichenberger:
Added support for Pow (but only version with float only), as well as Clip.

src/Conversion/ONNXToKrnl/Math/Clip.cpp
```diff
@@ -0,0 +1,104 @@
+//===----------------- Clip.cpp - Lowering Clip Op
+//-------------------------===//
```
doru1004, line 2: Adjust line length to fit.

src/Conversion/ONNXToKrnl/Math/Clip.cpp
```diff
@@ -0,0 +1,104 @@
...
+    SmallVector<Value, 4> loopIVs;
+    // Only create krnl.iterate if one of the operands is not scalar tensor.
+    if (!hasAllScalarValues(operands)) {
+      // Create iterateOp & get block within iterate op.
+      BuildKrnlLoop loops(rewriter, loc, memRefType.getRank());
+      loops.createDefineAndIterateOp(input);
+      Block *iterationBlock = loops.getIterateBlock();
+
+      // Insert instructions inside the KernelIterateOp body.
+      rewriter.setInsertionPointToStart(iterationBlock);
```
doru1004, line 52: I think you might need to use an insertion point guard so that the insertion point is restored to the previous one when this block ends.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -91,6 +91,36 @@ struct ScalarOp<ONNXSqrtOp> {
   using IOp = SqrtOp; // not use
 };
 
+template <>
+struct ScalarOp<ONNXAtanOp> {
+  using FOp = AtanOp;
+  using IOp = AtanOp; // not use
```
doru1004, line 97: What does "not use" mean?

src/Conversion/ONNXToKrnl/Math/Clip.cpp
AlexandreEichenberger: I copied the code from Elementwise.cpp; I really don't know much how the insertion point is set before entering such functions, I think it never need to be reset for coherent use after the function is over.

Here, the use is fine: if there is a loop, emit the loop and set insertion point inside; if there is no loop, insert where we would have put the loops.

doru1004: As long as it's correct I'm ok with it :) I though it would be worth another glance in case there is something wrong with it. Thanks for double checking.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
AlexandreEichenberger: It should be "not used"; the struct request one op for int, one for float. Some operations exist only for int, others only for float, some for both. Comment indicate which one is not used, and should never be called

doru1004: Thanks for the clarification! Makes sense.

doru1004 (approved):
LGTM

PR #467, opened 2021-01-11
update testing doc

AlexandreEichenberger:
Update testing doc by using the correct env var TEST_CASE_BY_USER instead of BACKEND_TEST, and list the cmake command that is used to force all the dynamic dimensions, as used by our test scripts.


chentong319 (approved):
Thanks

PR #468, opened 2021-01-12
Support padding and dilation in lowering ONNXConvOp

tungld:
To support padding for ONNXConvOp, the input of ONNXConvOp is padded with zeros before passing to ONNXConvOp whose lowering does not support padding. This way is easy to implement by using pattern matching rules, but has poor performance.

This patch directly supports padding in the ONNXConvOp lowering, and enables dilation support. It uses the same approach that has been used in lowering pooling ops, where the convolution window is exactly captured via affine maps and IndexExpr with the presence of pads, strides and dilations.

- [x] Edit lit tests


tungld:
@AlexandreEichenberger So far we are using IndexExpr for computing access index, and I do the same thing in this patch. However, I also want to use it for computing loop bounds (e.g. upperbound for krnl.iterate), but have not yet succeeded. That's why I manually used AffineMap for such purpose in this patch. Any thoughts on using IndexExpr for loop bounds?

AlexandreEichenberger:
@tungld you are raising a good point. Are the bounds of a given type? or they can be signed / unsigned short/int/long/long long?

Is otherwise the patch ready for review?

tungld:
> Are the bounds of a given type? or they can be signed / unsigned short/int/long/long long?

Bounds for `affine.for` are in form of `max(l)` for lower bound or `min(u)` for upper bound, where l, u are affine maps, and `min` and `max` are automatically inserted when we emit the `affine.for`. So what we need to prepare are affine maps. There are some limitations related to affine maps used here, e.g. its dims and symbols cannot be the output of Affine.Max or Affine.Min. 

> Is otherwise the patch ready for review?

Yes, it's ready.

src/Dialect/ONNX/ONNXOpsHelper.cpp
```diff
@@ -45,6 +46,110 @@ AffineMap getConvDimMap(Builder &builder, bool ceilMode) {
...
+  kernelOffsetExpr = IndexExpr::min(kernelExprs);
+
+  return std::vector<IndexExpr>{
+      windowStartExpr, windowEndExpr, kernelOffsetExpr};
+}
+
+/// The conv/pooling window can be smaller than the kernel when slicing it over
+/// the border edges. This function returns an AffineMap to compute the size of
+/// one edge of the window.
+AffineMap getWindowAffineMap(Builder &builder, bool ceilMode, bool isDilated) {
```
tungld, line 113: @AlexandreEichenberger FYI, this function produces an affine map that will be used as the upperbound for krnl.iterate. It's great if we can use IndexExpr here.

AlexandreEichenberger (approved):
LGTM

PR #470, opened 2021-01-13
Extend ONNX function translation to handle optional parameters

gramalingam:
ONNX functions permit optional parameters (just like ONNX ops). Extend the translation of ONNX function to MLIR function to handle missing optional parameters. Add a test-case to test it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,8 @@ export LIT_OPTS=-v
 cmake --build . --target check-onnx-lit
 ```
 
+If you are running on OSX Big Sur, you need to add `-DCMAKE_CXX_COMPILER=/usr/bin/c++`
+to the `cmake ..` command due to changes in the compilers.
 After the above commands succeed, an `onnx-mlir` executable should appear in the `bin` directory. 
 
 ## Installation on Windows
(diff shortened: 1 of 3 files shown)


chentong319:
At line 392 of FrontendTransformer.cpp, the comment says that there are two ways to leave optional input/output: 1) not provide, 2) give empty string. This patch handles case#2, while only case#1 is handled for ordinary onnx operation. 
Q1. Should this patch handle case#1 for function?
Q2. Should we add handling case#2 for ordinary onnx operation(perhaps by another patch)?

gramalingam:
@chentong319 : this patch handles both kinds of omitted optional inputs. Here, we are mapping an ONNX call to a MLIR function and a call to that MLIR function. The solution adopted here is that the MLIR function has inputs that correspond only to the inputs that are actually present.

gramalingam:
Regarding Q2: isn't it being handled for the empty-string case by pushing a NoneType value? There is the question of how trailing missing optional inputs are handled ... I assume that this is handled by the specification of the corresponding ONNX-MLIR dialect instruction ... if that MLIR instruction also allows the trailing optional inputs to be omitted, things should be fine. But if that MLIR instruction does not, then a patch may be required.

gramalingam:
Hi, any further comments or questions about this PR? Thanks!

chentong319 (approved):
LGTM

tjingrant:

gramalingam:
Great! I think it is ready to be merged, thanks!

tjingrant:

tjingrant:

Issue #471, opened 2021-01-14
Using an initializer for the Transpose Op


kwu91:
Hello, 

I have a very simplified use case where I have a model with a Transpose node using an initializer as its input. If I try to run `onnx-mlir --EmitONNXIR` on the model, I get the following error:

```
onnx-mlir: onnx-mlir/src/Transform/ONNX/ConstProp.cpp:54: mlir::RankedTensorType {anonymous}::constructRankedTensorType(mlir::ShapedType): Assertion `type.hasRank() && "Not a ranked type"' failed.
```

If I add a ShapeInferencePass at the very beginning, I am able to successfully generate a file to ONNX IR. 

```
diff --git a/src/MainUtils.cpp b/src/MainUtils.cpp
index 09b0d2f..9aef926 100644
--- a/src/MainUtils.cpp
+++ b/src/MainUtils.cpp
@@ -424,6 +424,7 @@ void registerDialects(mlir::MLIRContext &context) {
 }

 void addONNXToMLIRPasses(mlir::PassManager &pm) {
+  pm.addPass(mlir::createShapeInferencePass());
   pm.addNestedPass<FuncOp>(mlir::createDecomposeONNXToONNXPass());
   pm.addNestedPass<FuncOp>(mlir::createConstPropONNXToONNXPass());
   pm.addPass(mlir::createShapeInferencePass());
```

I'm curious to know if there might be any consequences if I add this pass there?

Below is the script I used to generate the aforementioned model.

```
import torch

torch.set_default_dtype(torch.float32)
SZ = 4

class TransposeModel(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.a = torch.nn.Parameter(torch.randn(SZ, SZ))

    def forward(self, x):
        t = torch.transpose(self.a, 0, 1)
        return t

model = TransposeModel()
x = []

dynamic_axes = {0 : "batch_size", 1: "batch_size"}
torch.onnx.export(model, x,
        'transpose.onnx',
	verbose=True,
        do_constant_folding=False,
        input_names = ['input'],
        output_names = ['output'],
        dynamic_axes={'input': dynamic_axes,
            'output': dynamic_axes})

```

Thanks!

tungld:
@kwu91 Thanks for reporting this issue!

[...]

PR #472, opened 2021-01-15
Constant propagation for SplitOp

tungld:
This patch does constant propagation for SplitOp.

It also removes the call to constant propagation that happens before shape inference. This solves the issue #471. Because constant propagation now supports both tensor and memref (by PR #462), it always expects a ranked data and is able to return ranked data.

### DIFF
--- a/src/Transform/ONNX/ConstProp.cpp
+++ b/src/Transform/ONNX/ConstProp.cpp
@@ -398,7 +398,7 @@ DenseElementsAttr ConstPropUnsqueeze(
 
   // Unqueeze does not change the order of access, so just copy the whole data.
   std::vector<Attribute> resVector;
-  for (auto value : denseAttr.getValues<Attribute>()) {
+  for (Attribute value : denseAttr.getValues<Attribute>()) {
     resVector.emplace_back(value);
   }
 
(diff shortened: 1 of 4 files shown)


src/Transform/ONNX/ConstProp.cpp
```diff
@@ -406,6 +406,127 @@ DenseElementsAttr ConstPropUnsqueeze(
...
+
+  LogicalResult matchAndRewrite(
+      ONNXSplitOp op, PatternRewriter &rewriter) const override {
+    auto loc = op.getLoc();
+    // A dense attribute that contains constant values of the split op's input.
+    Attribute denseAttr;
+
+    // Match
+    auto splitOp = ::llvm::dyn_cast_or_null<::mlir::ONNXSplitOp>(&op);
+    (void)splitOp;
```
AlexandreEichenberger, line 485: Is this needed? I see the splitOp being used below. If it's needed, do you mind just letting me know so that I understand the code pattern, tx.

src/Transform/ONNX/ConstProp.cpp
```diff
@@ -406,6 +406,127 @@ DenseElementsAttr ConstPropUnsqueeze(
...
+    {
+      auto *producerOp = splitOp->input().getDefiningOp();
+      auto castedProducerOp =
+          ::llvm::dyn_cast_or_null<::mlir::ONNXConstantOp>(producerOp);
+      (void)castedProducerOp;
```
AlexandreEichenberger, line 490: same here, used it below.

AlexandreEichenberger (approved):
Thanks, left a few questions. Also, I would prefer to have fewer reference to "auto". Maybe only for trivial things like Value/Operations, and having the types for the other ones?

If you are using Visual studio, you can type "auto" and then when you hover over the variable, you can actually get the type. 

I know LLVM frown over uses of auto, just trying to keep in line with their practices, if you don't mind.

tungld:
@AlexandreEichenberger I just changed `auto` to a specific type. Thanks for your comment! I will pay attention to `auto` from now on.

PR #474, opened 2021-01-19
Fix checking of traces for bundling dynamic memory pools

doru1004:
Fix checking of traces for bundling dynamic memory pools.


AlexandreEichenberger (approved):
LGTM, let's make sure we can still do bundling for the important cases

PR #475, opened 2021-01-19
update dropout Op to the new version

chentong319:
Changes include:

1.  update the version of dropout in gen_onnx_mlir.py
2. add the default valid for optional inputs
3.  Update the document
4. change of test case for check-onnx-lit and check-onnx-backend

Future work: onnx-mlir failed other dropout tests in onnx unit test. Some are due to the data format and the rest may be related to our rewrite rule. We need new lowering function for the more complicated case in the new version. I will take a look into it.

### DIFF
--- a/docs/Dialects/onnx.md
+++ b/docs/Dialects/onnx.md
@@ -1089,31 +1089,39 @@ ONNX Div operation
 
 ONNX Dropout operation
 
-"Dropout takes one input floating tensor and produces two tensor outputs,"
-"output (floating tensor) and mask (`Tensor<bool>`). Depending on whether it is"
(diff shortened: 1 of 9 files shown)


AlexandreEichenberger (approved):
LGTM, if you do one more pass, maybe you can add in the test.py comments that we currently don't support generating a mask, as it seems a feature for training (if you think that is correct, I think so, but not 100% sure).

And the endianness issue should be corrected eventually, if it plage our tests. We should ask the TRL folks as they dealt with such issues before and have thus experience on the matter.

AlexandreEichenberger:
thanks

PR #476, opened 2021-01-20
Introduce krnl.load and krnl.store

tungld:
This patch introduces krnl.load and krnl.store  ops. These ops will replace std.load/store or affine.load/store when we lower ONNX ops to krnl dialect. 

With krnl.load/store, we don't need to care about whether access indices are affine or not. The lowering of krnl.load/store to affine and std dialects will automatically check access indices and decide which one to use.

Also, we can remove the dependency of the krnl.iterate op to affine (say, no need of AffineScope trait in the krnl.iterate definition). affine.load/store will be emitted at the same time affine.for is emitted, which avoids potential errors related to affine related optimization passes.

### DIFF
--- a/src/Conversion/KrnlToAffine/KrnlToAffine.cpp
+++ b/src/Conversion/KrnlToAffine/KrnlToAffine.cpp
@@ -269,9 +269,9 @@ public:
         !llvm::any_of(indices, [](Value v) { return !isValidDim(v); });
 
     if (affineIndices)
-      rewriter.replaceOpWithNewOp<StoreOp>(op, value, memref, indices);
-    else
       rewriter.replaceOpWithNewOp<AffineStoreOp>(op, value, memref, indices);
+    else
+      rewriter.replaceOpWithNewOp<StoreOp>(op, value, memref, indices);
 
     return success();
   }
--- a/src/Conversion/KrnlToAffine/KrnlToAffine.cpp
+++ b/src/Conversion/KrnlToAffine/KrnlToAffine.cpp
@@ -212,6 +212,71 @@ LogicalResult interpretOperation(Operation *op, OpBuilder &builder,
   return success();
 }
 
+//===----------------------------------------------------------------------===//
+// Krnl to Affine Rewrite Patterns: KrnlLoad operation.
+//===----------------------------------------------------------------------===//
+
+/// KrnlLoad will be lowered to std.load or affine.load, depending on whether
+/// the access indices are all affine maps or not.
+class KrnlLoadLowering : public OpRewritePattern<KrnlLoadOp> {
+public:
+  using OpRewritePattern<KrnlLoadOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(
+      KrnlLoadOp op, PatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+    KrnlLoadOpAdaptor operandAdaptor = KrnlLoadOpAdaptor(op);
+
+    // Prepare inputs.
+    Value memref = operandAdaptor.memref();
+    SmallVector<Value, 4> indices = operandAdaptor.indices();
+
+    // Check whether all indices are affine maps or not.
+    bool affineIndices =
+        !llvm::any_of(indices, [](Value v) { return !isValidDim(v); });
+
+    if (affineIndices)
+      rewriter.replaceOpWithNewOp<AffineLoadOp>(op, memref, indices);
+    else
+      rewriter.replaceOpWithNewOp<LoadOp>(op, memref, indices);
+
+    return success();
+  }
+};
+
+//===----------------------------------------------------------------------===//
+// Krnl to Affine Rewrite Patterns: KrnlStore operation.
+//===----------------------------------------------------------------------===//
+
+/// KrnlStore will be lowered to std.store or affine.store, depending on whether
+/// the access indices are all affine maps or not.
+class KrnlStoreLowering : public OpRewritePattern<KrnlStoreOp> {
+public:
+  using OpRewritePattern<KrnlStoreOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(
+      KrnlStoreOp op, PatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+    KrnlStoreOpAdaptor operandAdaptor = KrnlStoreOpAdaptor(op);
+
+    // Prepare inputs.
+    Value value = operandAdaptor.value();
+    Value memref = operandAdaptor.memref();
+    SmallVector<Value, 4> indices = operandAdaptor.indices();
+
+    // Check whether all indices are affine maps or not.
+    bool affineIndices =
+        !llvm::any_of(indices, [](Value v) { return !isValidDim(v); });
+
+    if (affineIndices)
+      rewriter.replaceOpWithNewOp<StoreOp>(op, value, memref, indices);
+    else
+      rewriter.replaceOpWithNewOp<AffineStoreOp>(op, value, memref, indices);
+
+    return success();
+  }
+};
+
 void ConvertKrnlToAffinePass::runOnFunction() {
   OpBuilder builder(&getContext());
   mlir::Operation *funcOp = getFunction();
@@ -253,8 +318,14 @@ void ConvertKrnlToAffinePass::runOnFunction() {
   // krnl.dim operations must be lowered prior to this pass.
   target.addIllegalOp<KrnlDimOp>();
   target.addLegalOp<AffineYieldOp>();
+  target.addLegalOp<AffineLoadOp>();
+  target.addLegalOp<AffineStoreOp>();
+  target.addLegalOp<LoadOp>();
+  target.addLegalOp<StoreOp>();
   OwningRewritePatternList patterns;
   patterns.insert<KrnlTerminatorLowering>(&getContext());
+  patterns.insert<KrnlLoadLowering>(&getContext());
+  patterns.insert<KrnlStoreLowering>(&getContext());
   DenseSet<Operation *> unconverted;
   if (failed(applyPartialConversion(
           getFunction(), target, std::move(patterns), &unconverted)))
--- a/src/Dialect/ONNX/IndexExpr.cpp
+++ b/src/Dialect/ONNX/IndexExpr.cpp
@@ -18,6 +18,7 @@
 #include "mlir/Support/LLVM.h"
 #include "mlir/Support/MathExtras.h"
 #include "src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.hpp"
+#include "src/Dialect/Krnl/KrnlOps.hpp"
 #include "src/Dialect/ONNX/ONNXShapeHelper.hpp"
 #include "llvm/ADT/ArrayRef.h"
 #include "llvm/ADT/Sequence.h"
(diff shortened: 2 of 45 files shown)


src/Transform/BundleMemoryPools.cpp
```diff
@@ -320,8 +320,7 @@ class KrnlBundleDynamicMemoryPools : public OpRewritePattern<AllocOp> {
             // Check if the current operation is a DimOp or a LoadOp.
             if (llvm::dyn_cast<DimOp>(definingOperation) ||
                 llvm::dyn_cast<KrnlDimOp>(definingOperation) ||
-                llvm::dyn_cast<LoadOp>(definingOperation) ||
-                llvm::dyn_cast<AffineLoadOp>(definingOperation)) {
+                llvm::dyn_cast<KrnlLoadOp>(definingOperation)) {
```
doru1004, line 323: When is the lowering of krnl.load/store performed?

tungld: When we lower krnl to affine, i.e. inside convert-krnl-to-affine pass.

AlexandreEichenberger:
While technically changing the Load/store to krnl operations is feasible, I believe we need a better understanding of why we do it, what we gain, and what we may lose. In particular, having custom operation may prevent optimizations that are written in MLIR as it does not know about the krnl dialect. I believe you had an example of a strange behavior witnessed with the affine/std memory ops, and I think we need to collectively think about that example, and the pluses/minues to go to that approach.

On an unrelated ops, I tried to systematically use krnl.dim and it lost a lot of std.dim optimizations, which are crucial to bundling. I am not saying the same thing will happen here, but just want to raise the awareness of the cost of going custom, when there are alternatives. 

We can start the discussion here, and maybe have an in-depth presentations at some time so that we are all onboard with this PR. Thanks

doru1004:
@AlexandreEichenberger I agree.

I think one thing we need to test is if we use krnl.load/krnl.store and then we end up lowering them to std.load/std.store and canonicalize then we end up with the exact same code as if we just used std.load/std.store directly.

This observation is also valid for krnl.dim. We can "abuse" krnl.dim as long as we lower it fairly early and give the std optimization a chance to run on the std.dims we obtain from the lowered krnl.dims.

tungld:
As you are aware of, we had to disable affine-fusion in [MainUtils.cpp](https://github.com/onnx/onnx-mlir/blob/master/src/MainUtils.cpp#L460), it is because the program we emit is not compatible with MLIR affine related optimizations such as affine-loop-fuse, affine-loop-tile, etc. So we cannot do any MLIR affine related optimizations. To make it compatible, I had to disable all canonicalizations that happen before convert-krnl-to-affine, which is a very bad thing.

The same thing happens when we run our LSTM model where we need to call `memref-normalize`, we also need to disable all canonicalizations before convert-krnl-to-affine to compile the model.

With the above issues, I think it is not good to have affine.load/affine.store when affine.for has not appeared. So, by introducing krnl.load/krnl.store, we lower them to affine.load/store at the same time we lower krnl.iterate to afffine.for, which, I believe, makes MLIR optimization valid. With krnl.load/krnl.store, I can call arbitrary MLIR affine related optimizations without any problem.

doru1004:
@tungld just to make sure, if you emit affine load/stores, canonicalize and then emit affine.for loops later and only after that you attempt to perform affine-loop-fuse and affine-loop-tile affine optimizations, they fail? Does this failure happen in all cases or just some special cases?

It would be good to know what it is that canonicalization removes that those optimizations need so we can avoid such situations in the future.

tungld:
>  if you emit affine load/stores, canonicalize and then emit affine.for loops later and only after that you attempt to perform affine-loop-fuse and affine-loop-tile affine optimizations, they fail?
Yes. It happens all the time. The root cause is affine.load/store's canonicalization. The ops are inside krnl.iterate's body which is marked with AffineScope, so canonicalization will promote all access indices of affine.load/store to symbol. So far, in MLIR, marking a custom op with AffineScope is the only way to allow affine.load/store uses variables as access indices.

doru1004:
> so canonicalization will promote all access indices of affine.load/store to symbol

Sounds like this might be the actual problem.

src/Dialect/ONNX/IndexExpr.hpp
```diff
@@ -285,6 +285,12 @@ class IndexExprContext {
   void createStoreOp(
       Value val, Value memref, SmallVectorImpl<IndexExpr> &indices);
 
+  // Emit a krnl.load to load an element from the memref.
+  Value createKrnlLoadOp(Value memref, SmallVectorImpl<IndexExpr> &indices);
```
AlexandreEichenberger, line 289: Is there still a need for the regular "createLoadOp" and "createStoreOp"? I would think that they can be removed

src/Conversion/KrnlToAffine/KrnlToAffine.cpp
```diff
@@ -212,6 +212,71 @@ LogicalResult interpretOperation(Operation *op, OpBuilder &builder,
   return success();
 }
 
+//===----------------------------------------------------------------------===//
+// Krnl to Affine Rewrite Patterns: KrnlLoad operation.
+//===----------------------------------------------------------------------===//
+
+/// KrnlLoad will be lowered to std.load or affine.load, depending on whether
+/// the access indices are all affine maps or not.
+class KrnlLoadLowering : public OpRewritePattern<KrnlLoadOp> {
```
AlexandreEichenberger, line 221: Saw that you checked that the right affine/std op is generated, thanks!

src/Dialect/ONNX/IndexExpr.cpp
```diff
@@ -295,6 +296,25 @@ void IndexExprContext::createStoreOp(
...
+}
+
+void IndexExprContext::createKrnlStoreOp(
+    Value val, Value memref, SmallVectorImpl<IndexExpr> &indices) {
+  SmallVector<Value, 4> storeIndices;
+  for (IndexExpr ie : indices)
+    storeIndices.emplace_back(ie.getValue());
+  getRewriter().create<KrnlStoreOp>(getLoc(), val, memref, storeIndices);
+}
+
```
AlexandreEichenberger, line 317: I have seen other occurrence of create<AffineLoadOp> in IndexExpr.cpp, can you double check that they are all converted to the krnl versions?

test/mlir/krnl/krnl_bundle_lstm_memory_pool.mlir
```diff
@@ -30,8 +30,8 @@ module {
     %7 = "krnl.getref"(%6, %c0_i64, %3) : (memref<?xi8>, i64, index) -> memref<1x?x3xf32>
     %8:3 = krnl.define_loops 3
     krnl.iterate(%8#0, %8#1, %8#2) with (%8#0 -> %arg3 = 0 to 1, %8#1 -> %arg4 = 0 to %1, %8#2 -> %arg5 = 0 to 3) {
-      affine.store %cst_1, %2[%arg3, %arg4, %arg5] : memref<1x?x3xf32>
-      affine.store %cst_1, %7[%arg3, %arg4, %arg5] : memref<1x?x3xf32>
+      krnl.store %cst_1, %2[%arg3, %arg4, %arg5] : memref<1x?x3xf32>
```
AlexandreEichenberger, line 33: Could you take this test, and feed it to krnl to affine, and make sure all of the krnl stores are transformed to the affine as seen on the original test? Visual inspection would be enough... if you wanted to make it an official test, that would be fine too.

AlexandreEichenberger (commented):
@tungld I now better understand the approach, and it does indeed feel like it's the right thing to do. That was quite a change. From checking on my own upstream/master, there seem to be some create<AffineLoadOp> that are in IndexExpr that did not show up as modified to KrnlLoadOp; do you mind checking that you covered all of them?

Also, is it true that you now don't anticipate to see any std or affine load during krnl gen? If so, should we mark these operations are forbidden? Just for sanity? Because someone in the future may erroneously gen the old load/store?

doru1004:
When you emit krnl.load/store do you emit it just in place of the affine load/stores?

Do we have operations where we emit both affine and standard load stores or do we emit only affine load/stores?

It would be good to emit standard loads and store as we do now and only provide krnl load/store for affine load/stores. Benefits:
- the lowering of the krnl load/store would become very easy, just transform to affine load/store
- standard load/stores can still be optimized

test/mlir/krnl/krnl_bundle_lstm_memory_pool.mlir
tungld: @AlexandreEichenberger I fed it krnl-to-affine and saw all krnl.load/store were transformed to affine.load/store as seen on the original test. Memory pool bundling is done at the krnl level, so it is good to see that krnl.load/store does not affect the correctness of the bundling here, instead of lower to affine.

src/Dialect/ONNX/IndexExpr.hpp
tungld: I removed them. Thanks!

src/Dialect/ONNX/IndexExpr.cpp
tungld: Oops, thanks for pointing it out! Fixed them.

tungld:
@doru1004, krnl.load/store will replace both std.load/store and affine.load/store. The lowering of krnl.load/store is able to detect affine-compatiable access indices and decide to lower to std or affine. The detection is very easy by just calling a built-in MLIR method.

> It would be good to emit standard loads and store as we do now and only provide krnl load/store for affine load/stores.

I am afraid that, in this situation, users who emits code for onnx-to-krnl need to care about the difference between using std and krnl load/store.

tungld:
> Also, is it true that you now don't anticipate to see any std or affine load during krnl gen? If so, should we mark these operations are forbidden? Just for sanity? Because someone in the future may erroneously gen the old load/store?

Yes, marked them illegal when generating krnl.

doru1004:
> The detection is very easy by just calling a built-in MLIR method.

Is there any difference between the cases we would emit by hand and the cases the MLIR method would successfully identify i.e. are there situation which the method would not be able to identify as being affine?

test/mlir/krnl/krnl_bundle_lstm_memory_pool.mlir
AlexandreEichenberger: great, thanks for visually inspecting the code.

AlexandreEichenberger (approved):
LGTM, @doru1004 do you want to approve as well?

doru1004:
@AlexandreEichenberger waiting for one more answer on a question I ask earlier.

AlexandreEichenberger:
@tungld , FYI, Doru is waiting for one more answer. Got the answers for mine, so once Doru is on board, we should be good to go.

tungld:
> Is there any difference between the cases we would emit by hand and the cases the MLIR method would successfully identify as affine i.e. are there situations which the method would not be able to identify as being affine?

@doru1004 As I saw when changing lit tests in this patch, I saw no difference so far, the method correctly produced std or affine as we did by hand. The method has been using in MLIR affine-related optimization passes to do affine transformation, so I expect that it is able to identity affine well.

doru1004:
> > Is there any difference between the cases we would emit by hand and the cases the MLIR method would successfully identify as affine i.e. are there situations which the method would not be able to identify as being affine?
> 
> @doru1004 As I saw when changing lit tests in this patch, I saw no difference so far, the method correctly produced std or affine as we did by hand. The method has been using in MLIR affine-related optimization passes to do affine transformation, so I expect that it is able to identity affine well.

Sounds great! Thanks for the clarification.

doru1004 (approved):
LGTM

AlexandreEichenberger:
@tungld sorry #484 went before you, and I believe it must be introducing a Load/Store that result in complains. Will make our best to not have PR go before yours again.

doru1004:
@AlexandreEichenberger just a gentle reminder to revert the title change of the patch before you merge. :)

tungld:
@AlexandreEichenberger No problem, we could see failures when some lowering tries to use std and affine load/store.

AlexandreEichenberger:
Thanks for your patience. You may now resume merging to master.

PR #479, opened 2021-01-22
Upsample shape inference

dhan-groq:


dhan-groq:
Oops, I meant to create this PR at a fork, not the parent repo.

PR #487, opened 2021-01-26
Lower ONNX Erf to math library function call

doru1004:
Lower erf to math library function in 2 steps
1. ONNX Dialect `onnx.Erf` to Krnl Dialect `krnl.erf`
2. Krnl Dialect `krnl.erf` to LLVM IR Dialect `call erff()`


AlexandreEichenberger (commented):
Could I ask you to add one mlir test since its the first to use this functionality, which we will hopefully reuse a lot for the other trig functions

doru1004:
> Could I ask you to add one mlir test since its the first to use this functionality, which we will hopefully reuse a lot for the other trig functions

Absolutely, the patch is not ready hence why it's marked with WIP :)

AlexandreEichenberger (approved):
LGTM except for all the "auto". Talked about it earlier, think we should be explicit when not complicated. Though fine as is, would prefer seeing fewer of them.

doru1004:
@tjingrant is travis stuck? It's been more than 4 hours now and it hasn't finished.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -609,6 +615,20 @@ Value emitScalarOpFor<ONNXLessOp>(ConversionPatternRewriter &rewriter,
   }
 }
 
+//===----------------------------------------------------------------------===//
+// Scalar unary ops for lowering ONNXErfOp
+//===----------------------------------------------------------------------===//
+
+template <>
+Value emitScalarOpFor<ONNXErfOp>(ConversionPatternRewriter &rewriter,
```
tungld, line 623: This looks redundant since we already have `struct ScalarOp<ONNXErfOp>`.

AlexandreEichenberger (approved):
LGTM, nice use of C++

PR #488, opened 2021-01-27
Set split attribute of Split Op in constant prop when it is not specified

imaihal:
This PR sets split attribute of SplitOP in constant propagation when it is not specified. The attribute is computed from input.

I saw this case when the Split Op is generated in lowering pass after shape inference is done.


src/Transform/ONNX/ConstProp.cpp
```diff
@@ -503,9 +503,26 @@ class ConstPropSplitPattern : public OpRewritePattern<ONNXSplitOp> {
 
     // Rewrite
     unsigned outputNum = splitOp->getNumResults();
-    int64_t rank = splitOp->input().getType().cast<ShapedType>().getRank();
+    Value splitInput = splitOp->input();
+    int64_t rank = splitInput.getType().cast<ShapedType>().getRank();
     IntegerAttr axisAttr = splitOp->axisAttr();
     ArrayAttr splitAttr = splitOp->splitAttr();
+    if (!splitAttr) {
+      // If split parameter is not specified, it is constructed from output.
```
tungld, line 511: I guess you mean "... constructed from the input".

src/Transform/ONNX/ConstProp.cpp
```diff
@@ -503,9 +503,26 @@ class ConstPropSplitPattern : public OpRewritePattern<ONNXSplitOp> {
...
+      SmallVector<Attribute, 4> splits;
+      for (int i = 0; i < outputNum; ++i) {
+        Value splitOutput = splitOp->getResults()[i];
```
tungld, line 514: `splitOutput` is not used.

src/Transform/ONNX/ConstProp.cpp
```diff
@@ -503,9 +503,26 @@ class ConstPropSplitPattern : public OpRewritePattern<ONNXSplitOp> {
...
+        uint64_t splitAxis = axisAttr.getValue().getSExtValue();
+        RankedTensorType inType =
+            constructRankedTensorType(splitInput.getType().cast<ShapedType>());
```
tungld, line 517: How about using SmallVector constructor instead of a for-loop, like:
```
ArrayRef<int64_t> shape = splitInput.getType().cast<ShapedType>().getShape();
Attribute splitSize =  rewriter.getIntegerAttr(rewriter.getIntegerType(64), shape[splitAxis] / outputNum));
SmallVector<Attribute, 4> splits(outputNum, splitSize);
```

tungld (changes requested):
@imaihal thanks for the patch! Just some comments to shorten the code.

src/Transform/ONNX/ConstProp.cpp
imaihal: I forgot to update.. Thanks!

src/Transform/ONNX/ConstProp.cpp
imaihal: Updated. Thanks!

src/Transform/ONNX/ConstProp.cpp
imaihal: Updated. Thanks!

AlexandreEichenberger:
@tungld @tjingrant @chentong319 @doru1004 
This fix is due to a default attribute (which is computed, not a simple value) not being set when the op is created later. Is there a place where this code can be set and where it would be systematically called when creating a new op?

tungld:
Currently, default attributes are set when importing ONNX ops (where attributes are constants) or during shape inference (where attributes are calculated from input dimensions).  We can set default attributes in the op builder also. To avoid boilerplate, we may need to have a separate pass for attribute setting, or to do canonicalization with attributes.

AlexandreEichenberger:
In the constructor, it may not work too well as we may not have the input shape. Do you know if shape inference is called on one op when that op is created? If yes, then we can set the default attribute there. I added code to table gen to add attribute setters. If not, Haruki's approach maybe the only one that make sense.

tungld:
Shape inference is called after an op is created. In Haruki's case, the op is created after shape inference and using MemRef instead of Tensor for the input, while shape inference works on Tensor only.

imaihal:
I applied shape inference again after generating new ops, but it failed. 
> Making shape inference work with MemRef is also one way to go.

I'm not sure if this is easy to implement, but if I can use this, I think my issue will be solved.

AlexandreEichenberger (approved):
LGTM, thanks for having looked for possibilities of using a common approach to setting defaults. It looks like this is not possible at this time.

PR #491, opened 2021-01-28
Remove symbol legalization.

tjingrant:
Symbol legalization is not longer needed; currently symbol names only appear in the symbol table and is not imported to be the names of IR values.


src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -182,14 +182,18 @@ class FrontendGenImpl {
   }
 
   void BindOnnxName(const std::string &onnx_name, mlir::Value symbol) {
-    auto input_tensor_legalized_name = legalize_name(onnx_name);
+    std::string result;
+    result = onnx_name;
+    auto input_tensor_legalized_name = result;
```
gramalingam, line 187: Why the extra lines to do nothing? Won't ```auto input_tensor_legalized_name = onnx_name``` or ```std::string input_tensor_legalized_name = onnx_name``` work (in case a local copy is required)? Similar questions several times below.

tjingrant: Sorry for this confusion, the last commit is generated using automatic refactoring tools.  I should've marked this patch as WIP.

tjingrant:
@gramalingam would appreciate your review again.

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -947,7 +938,7 @@ class FrontendGenImpl {
     auto *fnEntryBlock = funcOp.addEntryBlock();
 
     // Save caller context, while generating callee function body.
-    SymbolMapping callerScope(std::move(frontend_symbols_));
+    frontend_symbols_.pushScope(func_name_prefix);
```
gramalingam, line 941: Just for my understanding: what exactly does this scope mechanism do? Strictly speaking, the callee function body is not a nested-scope and it cannot reference symbols defined in the caller function. Will this potentially cause an issue (in how erroneous input models are handled ... I guess it may not matter in a correct model)?

gramalingam:
LGTM, thanks ... what is the motivation for dropping the name legalization? It seems common for ONNX models to violate the naming requirement for tensors, which will likely cause invalid mlir names, won't it?

src/Builder/FrontendDialectTransformer.cpp
tjingrant: Sorry I was mistaken. I assumed the scoping rules of the function body are similar to those of the subgraphs associated with control-flow operations like Loop and conditionals.

It seems to me that here we need a fresh scope with only function input variables in the symbol table.

tjingrant:
@gramalingam symbol name legalization was here mostly because previously we emitted C code with symbol names  as variable names. Presently, symbol names are only imported as an array of string attributes associated with each operation and are not the names of IR values (AFAIK there's no named IR value mechanism in MLIR), so they can be imported as is.

This will also likely aid the inspection of the ONNX model, given that you can search for symbols in tools like Netron, knowing that symbol names are unchanged.

src/Builder/FrontendDialectTransformer.cpp
gramalingam: Do you mean switch back to the previous implementation? Anyway this seems orthogonal to the main goal of the PR. Or, do you mean some other approach?

tjingrant: I will switch back to your original implementation shortly. I don't see a need for another approach presently, the ad-hoc symbol table works reasonably well under most circumstances.

AlexandreEichenberger:
@tjingrant do you still need a review?

tjingrant:
@AlexandreEichenberger yes, but please don't merge yet because a previous review comment is pending resolution and I've been procrastinating on it; will get it addressed soon.

AlexandreEichenberger:
@tjingrant NP, let me know when its ready and I will give it a quick review then.

tjingrant:
@AlexandreEichenberger the requested changes have been made.

AlexandreEichenberger (approved):
LGTM

gramalingam (approved):
Thanks!

chentong319 (approved):
LGTM!

PR #494, opened 2021-01-29
lower various trig ops to math lib

caoimhinuibrian:


caoimhinuibrian:
atan was being lowered to an mlir intrinsic, but there is no lowering of that to LLVM and the discussion on the MLIR channel seems to say there is no plan to add the lowering ... so, I just lowered it directly to an LLVM call like all the others.

test/backend/test.py
```diff
@@ -125,14 +127,17 @@
     # Argmin
 
     # Asin
+    "test_asin_cpu": (test_static_dynamic,),
 
     # Asinh
+    "test_asinh_cpu": (test_static_dynamic,),
 
     # Atan
     #"test_atan_example_cpu": (test_static_dynamic,),
```
AlexandreEichenberger, line 136: Are the _example test not working? If they are, I would enable them too

src/Conversion/KrnlToLLVM/KrnlToLLVM.cpp
```diff
@@ -531,7 +531,84 @@ struct MathFunctionName {
 
 template <>
 struct MathFunctionName<KrnlErfOp> {
-  static std::string functionName() { return "erff"; }
+  static std::string functionName(mlir::Type type) { return "erff"; }
```
AlexandreEichenberger, line 534: I would use the same template as you used, asserting an error if type is not float. If the erf is used for double format, that would be great too.

src/Conversion/KrnlToLLVM/KrnlToLLVM.cpp
```diff
@@ -531,7 +531,84 @@ struct MathFunctionName {
...
+};
+
+template <>
+struct MathFunctionName<KrnlAcosOp> {
+  static std::string functionName(mlir::Type type) {
+    if (type.isF32())
+      return "acosf";
+    if (type.isF64())
+      return "acos";
+    assert(false && "Unsupported type for acos");
```
AlexandreEichenberger, line 544: The proper form for `assert(false)`  is `llvm_unreachable("message")`.

AlexandreEichenberger (commented):
Great, couple of minor changes requested.

AlexandreEichenberger (approved):
LGTM, thanks for the changes, adding the test, and supporting ERF double.

AlexandreEichenberger:
Looks like there is still a formatting error, otherwise all good.

PR #498, opened 2021-02-02
Support lowering reshape when its input is promoted to attribute

### LINKED ISSUE #490: Assertion failed during Reshape Op
It failed to compile this model:
```
Assertion failed: (isa<U>()), function cast, file /Users/shizy/github/llvm-project/mlir/include/mlir/IR/Types.h, line 241.
```

图片

I think it is input `rnn_h` with shape [1, -1, 256] that cause the problem.

tungld:
This patch is to fix #490.

When the shape input of Reshape is constant, it is promoted to attribute. Thus, the lowering should read shape information from the attribute instead of the shape input (which became None after promotion).

For ops that have attribute promotion, we should handle two cases: an input is 1) promoted and 2) not promoted.

### DIFF
--- a/src/Conversion/ONNXToKrnl/Tensor/Reshape.cpp
+++ b/src/Conversion/ONNXToKrnl/Tensor/Reshape.cpp
@@ -12,21 +12,6 @@
 
 using namespace mlir;
 
-Value getDimOrConstant(ConversionPatternRewriter &rewriter, Location loc,
-    Value operand, int64_t axis) {
-  ArrayRef<int64_t> shape = operand.getType().cast<ShapedType>().getShape();
-  Value dimVal;
-  if (shape[axis] < 0) {
-    Value dim = rewriter.create<DimOp>(loc, operand, axis);
-    dimVal =
-        rewriter.create<IndexCastOp>(loc, dim, rewriter.getIntegerType(64));
-  } else {
-    dimVal =
-        emitConstantOp(rewriter, loc, rewriter.getIntegerType(64), shape[axis]);
-  }
-  return dimVal;
-}
-
 struct ONNXReshapeOpLowering : public ConversionPattern {
   ONNXReshapeOpLowering(MLIRContext *ctx)
       : ConversionPattern(mlir::ONNXReshapeOp::getOperationName(), 1, ctx) {}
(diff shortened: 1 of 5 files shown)


src/Conversion/ONNXToKrnl/Tensor/Reshape.cpp
```diff
@@ -12,16 +12,46 @@
 
 using namespace mlir;
 
+Value getDimOrConstant(ConversionPatternRewriter &rewriter, Location loc,
```
AlexandreEichenberger, line 15: This functionality stood go in the helper, if it does not exist already. In fact, IndexExpr has a lot of this things too, but I understand if you don't want to switch the implementation to that.

src/Conversion/ONNXToKrnl/Tensor/Reshape.cpp
```diff
@@ -31,71 +61,73 @@ struct ONNXReshapeOpLowering : public ConversionPattern {
...
           } else {
-            dimVal =
-                emitConstantOp(rewriter, loc, loadedValType, inputShape[i]);
+            auto dimVal = getDimOrConstant(rewriter, loc, data, i);
+            auto zero =
+                emitConstantOp(rewriter, loc, rewriter.getIntegerType(64), 0);
+            auto isZero = rewriter.create<CmpIOp>(
+                loc, CmpIPredicate::eq, loadedVal, zero);
+            loadedVal =
+                rewriter.create<SelectOp>(loc, isZero, dimVal, loadedVal);
```
chentong319, line 122: The loadedValue are assigned multiple times and its semantic changes in the process. Use different name, such as loadedValue then outputDimValue? 
The loadedValue is compared with 0, how about -1?  Will it make the tensorSizeFromShape become negative?

chentong319 (commented):
You may test the code with promoteConstantToAttribute disabled.

src/Conversion/ONNXToKrnl/Tensor/Reshape.cpp
tungld: @chentong319 thanks for your comments!
I reorganized the code to make it clear. 

> how about -1? Will it make the tensorSizeFromShape become negative?

Yes, it becomes negative. We need to make it positive later.

> You may test the code with promoteConstantToAttribute disabled.
it is already in the onnx_lowering.mlir test.

src/Conversion/ONNXToKrnl/Tensor/Reshape.cpp
tungld: @AlexandreEichenberger, thanks for your comments! I moved the function to a common place.

Here, we need to compute an output dim from values in a tensor. So we load a value from a tensor, and the value is always unknown. I am not sure it is appropriate to use IndexExpr here.

AlexandreEichenberger (approved):
LGTM, @chentong319 can you make sure you are happy too, tx

PR #505, opened 2021-02-03
Handle alignment for memory pooling

doru1004:
This patch does the following:
- prevents AllocOps with maps from participating in the optimization
- prevents AllocOps which return MemRefs with `index` element types to participate in the optimization.
- enables emission of individual memory pools to propagate alignment information.
- enables the bundling of individual memory pools to consider alignment. Memory pools with the same alignment are bundled together.
- enables optimization infrastructure for static memory pools to consider the presence of several memory pools with different alignments.


src/Support/KrnlSupport.cpp
```diff
@@ -308,6 +308,15 @@ int64_t getAllocArgIndex(AllocOp allocOp, int64_t index) {
   return -1;
 }
 
+/// Get alignment of an AllocOp if it exists else return zero.
+int64_t getAllocAlignment(AllocOp allocOp) {
+  int64_t alignment = 0;
+  if (IntegerAttr alignmentAttr = allocOp.alignmentAttr())
+    alignment = alignmentAttr.getInt();
```
AlexandreEichenberger, line 315: Minor, but could transform this to 
if ( ) return ;
return 0;

src/Transform/BundleMemoryPools.cpp
```diff
@@ -34,7 +34,8 @@ namespace {
 //===----------------------------------------------------------------------===//
 
 // Data structure for managing memory pools.
-typedef std::map<Block *, AllocOp> BlockToMemPool;
+typedef std::map<int64_t, AllocOp> AlignmentToMemPool;
```
AlexandreEichenberger, line 37: Do you mind adding a little more explanation on these two maps? Describing data goes a long way toward making the code more readable. My understanding is that each block has a unique map of alignment to a mem pool". So we have one alloc pool per alignment and block.

It would be great if you could also write at the top of this (or the header file) what are the high level characteristics currently handled by your approach:

Also currently bundle data that satisfy x) y) z) Currently bundling of buffers that w) is prevented.
This would help us better understand the code as it evolves.

src/Transform/BundleMemoryPools.cpp
```diff
@@ -160,20 +161,34 @@ class KrnlBundleStaticMemoryPools : public OpRewritePattern<AllocOp> {
...
     }
 
-    // If this parent block has been found present in the map, it means
-    // a static memory bundle already exists. Fetch it.
-    AllocOp staticMemPoolAlloc = blockToStaticPool->at(parentBlock);
+    // Static memory pool for this alignment exists, fetch it.
+    AllocOp staticMemPoolAlloc = alignmentToMemPool->at(alignment);
 
     // If this is the alloc representing the memory pool and the function
     // already has an init block, pattern matching must fail to avoid
```
AlexandreEichenberger, line 194: Avoid what?

src/Transform/OptimizeMemoryPools.cpp
```diff
@@ -29,7 +30,8 @@ using namespace mlir;
 namespace {
 
 // Handling of static memory pool on a block-basis in each function.
-typedef std::map<Block *, bool> BlockToCompactedFlag;
+typedef std::map<Block *, llvm::SmallSet<int64_t, 16>>
```
AlexandreEichenberger, line 33: Same thing here, a bit more explanation of what is being tracked. Is it that you are keeping track for each block of the alignment constraints that you have been seen?

AlexandreEichenberger (commented):
I understand that each alloc are binned into pools of equal alignment constraints. What happens when there are no constraints? Are they all binned into a "0" pool? A default one for which there are no constraints?

doru1004:
> Are they all binned into a "0" pool? A default one for which there are no constraints?

That's right.

src/Transform/BundleMemoryPools.cpp
doru1004: The changes are minimal, instead of having a single memory pool as before now there is one per alignment.

src/Transform/BundleMemoryPools.cpp
doru1004: the answer is on the next line :) (although it should say static memory pool not dynamic)

AlexandreEichenberger (approved):
LGTM

PR #508, opened 2021-02-07
Lowering imperfectly nested Krnl Statements.

tjingrant:
- Support lowering imperfectly nested krnl statements.
- Support `krnl.get_induction_var_value` operation, which converts optimized loop references to their corresponding induction variable values.
- Update softmax to use new loop builder API, this is relevant to this patch because it was re-using optimized loop references, which is incompatible with the lowering procedure introduced in this patch. 

Consider the following Krnl IR :

```mlir
  %ii, %ij = krnl.define_loops 2
  %ib, %il = krnl.block %ii 5 : (!krnl.loop) -> (!krnl.loop, !krnl.loop)
  %jb, %jl = krnl.block %ij 4 : (!krnl.loop) -> (!krnl.loop, !krnl.loop)
  krnl.permute(%ib, %il, %jb, %jl) [0, 2, 1, 3] : !krnl.loop, !krnl.loop, !krnl.loop, !krnl.loop
  krnl.iterate(%ib, %jb) with (%ii -> %i = 0 to 10, %ij -> %j = 0 to 20) {
    %ib_idx, %jb_idx = krnl.get_induction_var_value(%ib, %jb) : (!krnl.loop, !krnl.loop) -> (index, index)
    %bar = addi %ib_idx, %jb_idx : index

    %alloc = alloc() : memref<10 x f32>
    krnl.iterate(%il, %jl) with () {
      %foo = addi %i, %j : index
    }
    dealloc %alloc : memref<10 x f32>
  }
```

this will be lowered to:

```mlir
#map0 = affine_map<(d0) -> (d0)>
#map1 = affine_map<(d0) -> (d0 + 5)>
#map2 = affine_map<(d0) -> (d0 + 4)>
module  {
  func @test_2d_tiling_imperfectly_nested() {
    affine.for %arg0 = 0 to 10 step 5 {
      affine.for %arg1 = 0 to 20 step 4 {
        %0 = addi %arg0, %arg1 : index
        %1 = alloc() : memref<10xf32>
        affine.for %arg2 = #map0(%arg0) to #map1(%arg0) {
          affine.for %arg3 = #map0(%arg1) to #map2(%arg1) {
            %2 = addi %arg2, %arg3 : index
          }
        }
        dealloc %1 : memref<10xf32>
      }
    }
    return
  }
}
```


src/Conversion/KrnlToAffine/KrnlToAffine.cpp
```diff
@@ -25,6 +25,215 @@
 using namespace mlir;
 
 namespace {
+// Since Krnl Dialect allows optimizations to be specified in the form of
```
AlexandreEichenberger, line 28: The explanation below is great, and I am trying to clarify it a bit more. If I look at the example given in permute:
```mlir
    %ii, %jj, %kk = krnl.define_loops 3
    // Blocking each loop by a factor of 4.
    %ib, %il = krnl.block %ii 4 : (!krnl.loop) -> (!krnl.loop, !krnl.loop)
    %jb, %jl = krnl.block %jj 4 : (!krnl.loop) -> (!krnl.loop, !krnl.loop)
    %kb, %kl = krnl.block %kk 4 : (!krnl.loop) -> (!krnl.loop, !krnl.loop)
    // Move iteration over tile coordinates to be the outer loops and iterateion over
    // the inter-tile elements to be the inner loops.
    krnl.permute(%ib, %il, %jb, %jl, %kb, %kl) [0, 3, 1, 4, 2, 5] : !krnl.loop, !krnl.loop, !krnl.loop, !krnl.loop, !krnl.loop, !krnl.loop
    krnl.iterate(%ib, %il, %jb, %jl, %kb, %kl) with (%ii -> %i = 0 to 1024, %jj -> %j = 0 to 2048, %kk -> %k = 0 to 4096)  {
    }
```

there is a single iterate that list all of the blocked and intra-tile iterations `%ib, %il, %jb, %jl, %kb, %kl`. As an aside, the order in the iterate and permute input parameter has not been defined: what is it? Because ii came before jj in the define output, does it mean that the ib and il must be before the jb and jl in the permute input? Clarifying this would be helpful. I understand that the actual order is fixed by the permute literal `[0, 3, 1, 4, 2, 5]`.

Now if I compare this generic tiling to the example you give below:
```mlir
  %ii, %ij = krnl.define_loops 2
  %ib, %il = krnl.block %ii 5 : (!krnl.loop) -> (!krnl.loop, !krnl.loop)
  %jb, %jl = krnl.block %ij 4 : (!krnl.loop) -> (!krnl.loop, !krnl.loop)
  krnl.permute(%ib, %il, %jb, %jl) [0, 2, 1, 3] : !krnl.loop, !krnl.loop,
      !krnl.loop, !krnl.loop
  krnl.iterate(%ib, %jb) with (%ii -> %i = 0 to 10, %ij -> %j = 0 to 20) {
     %alloc = alloc() : memref<10 x f32>
     krnl.iterate(%il, %jl) with () {
        %foo = addi %i, %j : index
     }
     dealloc %alloc : memref<10 x f32>
  }
```

The first iterate contains only the two blocked iterations. This thus indicate that the alloc operation will be located directly inside of the two outer blocked iterations.  Then the two inner tile loop iterators are given with the add in the innermost. Thus the locations of the allocate/add are fully defined by the iterates and which loop iterators they entailed.

If my understanding is correct, that is good. I still wonder what the permute was about. To me, it is still redundant before, because we could have used the "order" in which the loop iterations are given as parameter to the `iterate` operation. Now it looks that since the original `%ib, %il, %jb, %jl` can be further separated into two distinct iterate operation, there is a second order of redundancy.

Ideally there would be no need for the permute in these cases; best would be that they become implicit (that is, if you need them,  they be automatically generated, as the order in the iterate would directly reflect the order of the loops. If that is too much work, we minimally need to be explicit that the order in the iterate is lexicographical (how they are listed in the define, permute outputs), and that the permute operation that reorder them must match the use across the iterate operation that may materialize them.

src/Conversion/KrnlToAffine/KrnlToAffine.cpp
```diff
@@ -25,6 +25,215 @@
...
+// affine.for %i = 0 to 10 {
+//   affine.for %j = 0 to 20 {
+//     %foo = addi %i, %j : index
+//   }
+// }
+//
+// Since the tiling has not taken place yet, tile coordinat iteration loops have
+// not been materialized, therefore the alloc and dealloc operations do not fit
+// in the IR presently yet. Instead, they will be placed within a krnl.movable
+// op region, to indicate that their positioning is subject to change.
```
AlexandreEichenberger, line 93: Again, thanks for a very thorough description. If my understanding is correct, the movable is never generated by the user (aka the person writing the lowering of ONNX to KRNL), it is generated during the process of lowering. I would add this "This operation is automatically generated by the lowering of KRNL to affine to assist with patterns found, for example, during the tiling of imperfectly nested loops".

This mechanism is very ingenious.

src/Conversion/KrnlToAffine/KrnlToAffine.cpp
```diff
@@ -25,6 +25,215 @@
...
+// krnl.permute(%ib, %il, %jb, %jl) [0, 2, 1, 3] : !krnl.loop, !krnl.loop,
+//     !krnl.loop, !krnl.loop
+// krnl.iterate(%ib, %jb) with (%ii -> %i = 0 to 10, %ij -> %j = 0 to 20) {
+//   %alloc = alloc() : memref<10 x f32>
+//   krnl.iterate(%il, %jl) with () {
+//     %foo = addi %i, %j : index
+//   }
+//   dealloc %alloc : memref<10 x f32>
+//  }
+//
```
AlexandreEichenberger, line 55: Another question: with krnl.matmul (which compute a panel of NxMxK), it will implement something akin to the 
```mlir
   krnl.iterate(%il, %jl) with () {
      %foo = addi %i, %j : index
    }
```
code above. So is it ok not to omit this inner iterate and generate the krnl.matmul in its place? Again, what does it say about the permute?

test/mlir/krnl/get_induction_var_value.mlir
```diff
@@ -0,0 +1,53 @@
+// RUN: onnx-mlir-opt --convert-krnl-to-affine %s -split-input-file | FileCheck %s
+
+func @test_simple() {
+  %ii = krnl.define_loops 1
+  krnl.iterate(%ii) with (%ii -> %i = 0 to 10) {
+    %idx = krnl.get_induction_var_value(%ii) : (!krnl.loop) -> (index)
```
AlexandreEichenberger, line 6: This is a loop, right, as `%ii` is already fully specified?

src/Dialect/Krnl/KrnlOps.td
```diff
@@ -595,3 +595,38 @@ def KrnlStoreOp : Op<Krnl_Dialect, "store",
...
+
+}
+
+def KrnlGetInductionVariableValueOp : Op<Krnl_Dialect, "get_induction_var_value"> {
+  let summary = "Krnl ";
+  let description = [{
+     Krnl operation to convert loop references to corresponding induction
+     variable values. This is useful for accessing optimized loop induction
+     variables, as they are not otherwise accessible during Krnl Dialect.
+  }];
```
AlexandreEichenberger, line 623: Can you be a bit more specific on the value returned in a blocked loop? It is not the block id, but the index value corresponding to the block?

AlexandreEichenberger (approved):
LGTM after a couple of additional comments in the code to clarify its behavior

caoimhinuibrian:
@tjingrant Is this ready to merge?

tjingrant:
@caoimhinuibrian not quite. I will follow up with @AlexandreEichenberger to answer his questions today and probably let him decide (based on whether his questions are addressed) when to merge this patch.

src/Conversion/KrnlToAffine/KrnlToAffine.cpp
tjingrant: To clarify when/how the order of appearance of loop handlers are meaningful:

- The order in which loop handlers are defined by `define_loops` operation is irrelevant and arbitrary. This ordering information is not used anywhere in any ways.

- The order in which loop handlers appear **after** `with` keyword stipulates the ordering of loops in the unoptimized case, and if an optimization recipe is present, optimizations will be applied successively to this unoptimized piece of code.

- The order in which loop handlers appear **before** `with` keyword **corresponds to, but does not determine** the ordering of loops after the original program has been optimized. So you will see the appearance of tiled loop handlers here. However, you can't put in arbitrary ordering of loop handlers here because there is only 1 single correct loop ordering, which corresponds to how loops are ordered in the optimized code after applying recipes to the unoptimized code. 

- The ordering of loop handlers within the input of krnl.permute operation is irrelevant and arbitrary. Since we only care about which new positions to send these loops to.

This is just the way it is right now, I don't think the third point makes total sense.


> The first iterate contains only the two blocked iterations. This thus indicate that the alloc operation will be located directly inside of the two outer blocked iterations. Then the two inner tile loop iterators are given with the add in the innermost. Thus the locations of the allocate/add are fully defined by the iterates and which loop iterators they entailed.

Your understanding is right.

> If my understanding is correct, that is good. I still wonder what the permute was about. To me, it is still redundant before, because we could have used the "order" in which the loop iterations are given as parameter to the iterate operation. 

I agree. Permute makes implementation a bit easier, so it's probably not their for the right reason. I agree that ultimately this redundancy should be eliminated by removing permute operation.

> Now it looks that since the original %ib, %il, %jb, %jl can be further separated into two distinct iterate operation, there is a second order of redundancy.

Why is there a second order of redundancy?

> Ideally there would be no need for the permute in these cases; best would be that they become implicit (that is, if you need them, they be automatically generated, as the order in the iterate would directly reflect the order of the loops. 

This is a worthwhile goal to work towards.

> If that is too much work, we minimally need to be explicit that the order in the iterate is lexicographical (how they are listed in the define, permute outputs), and that the permute operation that reorder them must match the use across the iterate operation that may materialize them.

Yes, they are and should be lexicographical in iterate op. However, they don't need to correspond to how they're listed in define/permute simply because of the way define and permuate are interpreted (define create loops, and says nothing about the ordering; permute argument specifies the permuted loop ordering, so don't care about how inputs are ordered). Please let me know if this makes sense or not.

src/Conversion/KrnlToAffine/KrnlToAffine.cpp
tjingrant: Yes. Will do.

src/Conversion/KrnlToAffine/KrnlToAffine.cpp
tjingrant: > So is it ok not to omit this inner iterate and generate the krnl.matmul in its place?

You mean is it okay to **omit** this inner iterate right? The best way to implement this is to lay this `krnl.matmul` down as an optimization recipe, and let the krnl dialect lowering process to do the substitution for you, as opposed to modifying the code directly (which amounts to doing part of the lowering outside Krnl lowering).

This means there's probably a need to implement another optimization primitive called `substitute_with_kernel`.

test/mlir/krnl/get_induction_var_value.mlir
tjingrant: Yes, this is a degenerate case where you can just use `i`. Instead of going to great length trying to convert the loop handler into an induction variable.

src/Dialect/Krnl/KrnlOps.td
tjingrant: Ahh, good point. Yes, it's the tile starting indices.

AlexandreEichenberger:
@tjingrant thanks for your helpful comments. We will need to talk about the substitution of the kernel, for a later patch. Please squash and merge once you feel the PR is ready. Thanks again.

src/Conversion/KrnlToAffine/KrnlToAffine.cpp
tjingrant: Just to update for record-keeping, per our latest discussion, directly substituting the inner loop with kernel makes more intuitive sense.

PR #509, opened 2021-02-08
Update LLVM commit to 980d28d Feb 7

tungld:


src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2068,7 +2068,7 @@ LogicalResult ONNXPadOp::inferShapes(
   SmallVector<int64_t, 4> outputShape(dataShape.begin(), dataShape.end());
 
   // Get pads from valueAttribute.
-  Attribute padattr = getAttr("pads");
+  Attribute padattr = (*this)->getAttr("pads");
```
tjingrant, line 2071: Hi @tungld do you mind explaining why these patterns are popping up? i.e., why is the object `this` is referring to becomes something like a pointer to a pointer? Any patch/background discussion in mlir linked to this change?

tungld: Hi @tjingrant, sorry for my late reply!

I guess, `this` refers to `OpState::operator`. I found this MLIR patch which may help understand the change: https://github.com/llvm/llvm-project/commit/1ffc1aaa09674824ce66600ee0b77d6a2c042e68#diff-9185f7f3f3628467e62e1712b01efc482eecc3643b67746c1eae73f5fd272138

tjingrant: Got it, thanks! I have no concerns now.

tjingrant (approved):
LGTM!

PR #511, opened 2021-02-08
Fix memleak reports from cppcheck

seroyer:
Two of the issues are unlikely in practice because they are only exposed
if you get a memory allocation failure.  The other is actually a bug
that manifests as a memleak, and only applies to the `__APPLE__`
environment.

Signed-off-by: Steven Royer <seroyer@us.ibm.com>


tjingrant:

caoimhinuibrian:
test this please

PR #512, opened 2021-02-09
Dockerized jenkins ci

gongsu832:
This patch adds a set of Jenkins build scripts (written in python)
    driven by the main Jenkinsfile script to automate the build and
    test of ONNX-MLIR github repo whenever a pull request is submitted,
    updated, and merged, etc. The pipeline job is designed such that
    multiple builds for different pull requests can run concurrently
    but only one build at a time can run for the same pull request.

    The build and test are done inside docker containers and verified
    docker images will be published to the dockerhub under the username
    onnxmlirczar when a pull request is merged and if necessary (i.e.,
    source repos used to build these images have newer commit sha1
    than those in the dockerhub). Four images are published with tags
    indicating their target architecture:

      onnx-mlir-llvm-static:{s390x|amd64} (llvm-project static libs)
      onnx-mlir-llvm-shared:{s390x|amd64} (llvm-project shared libs)
      onnx-mlir-dev:{s390x|amd64}         (onnx-mlir developer image)
      onnx-mlir:{s390x|amd64}             (onnx-mlir user image)

    For onnx-mlir-dev and onnx-mlir images, a multiarch manifest list
    tag "latest" is created so that they can be pulled without having
    to explicitly specify the architecture tag.


caoimhinuibrian (approved):
LGTM

Issue #506, opened 2021-02-04
How to generate the Executable File in x86-linux


CHUNERnubi:
I've finished the build add.onnx,  and then step by step down the dialect, and finally down to llvm IR, get the llvm IR file. If I want to generate executable files on the x86-Linux platform, what should I do?

I try to use the tool chain in llvm `./mlir-translate`. I want to talk about that llvm IR is first converted into an object file, then C program is used to call the C interface in the object file, and finally link and compile to generate an executable file.

Thank you in advance.

tjingrant:
My find this tutorial helpful: http://onnx.ai/onnx-mlir//doxygen_html/index.html

CHUNERnubi:
@tjingrant thanks a lot~

CHUNERnubi:
@tjingrant Thanks for the web link, I have successfully generated the executable file, and can run the correct results. However, this method can not see the whole process of gradual decline of dialect clearly. I also read your paper about this project. I still want to reduce dialect step by step, and be able to see the result of every decline. Do you have any suggestions?
Thank you very much.

tjingrant:
@CHUNERnubi i'm not sure what's stopping you from seeing the intermediate results of gradual lowering. Do you mean that you want to "edit" the intermediate results and then continue? (which is only possible currently using onnx-mlir-opt)

CHUNERnubi:
I just want to know is there any way or method can generate .so file from LLVM IR Dialect file, using the toolchain which is builded by this project. Thanks a lot.

caoimhinuibrian:
@CHUNERnubi Starting with an onnx file, "fn", you can use "onnx-mlir --EmitLib  fn.onnx" to generate a "fn.so". If you want to start from an arbitrary mlir file, you would need to use onnx-mlir-opt, specifying all the passes you want to run and ending with an LLVM IR bitcode file. You could then take the bitcode file and use llopt and llc to generate the .so. Currently the end-to-end compilation only starts from and onnx file. If you feed it an mlir file, it wouldn't in general know what passes to run on that.

If you want just to start from an onnx file and see the intermediate results after each pass, you can use these options:  --print-ir-after=<pass-arg>                       - Print IR after specified passes
  --print-ir-after-all                              - Print IR after each pass
  --print-ir-after-change                           - When printing the IR after a pass, only print if the IR changed
  --print-ir-before=<pass-arg>                      - Print IR before specified passes
  --print-ir-before-all                             - Print IR before each pass
[...]

PR #516, opened 2021-02-11
allow pass manager options

caoimhinuibrian:


caoimhinuibrian:
For debugging, its useful to be able to see the IR after each pass. This PR enables taht.

tjingrant:
Hi Kevin, the `onnx-mlilr` executable was conceptualized to be a simple user-facing frontend, that's why it contains a short help message description (https://github.com/onnx/onnx-mlir#using-onnx-mlir). I suspect this patch may obfuscate that concise help message.

OTOH, I feel like the proper thing to do may be to register our usual compiler pass pipelines in `onnx-mlir-opt`, which is supposed to be an `onnx-mlir` debugger, although I understand that as we amass more compiler passes, it is increasingly impossible to type out all the passes executed by our main compiler frontend as command-line options, therefore the need to register them as pass pipelines mentioned earlier. 

OTOH again, if you still think this is the right thing to do, maybe we should think instead about merging `onnx-mlir` and `onnx-mlir-opt` into one, given this patch will make two utilities overlap more in functionalities.

chentong319:
I assume that this change is to enable all the pass related flags defined in mlir for onnx-mlir.  The flags include IR printing, timing, statistic and crash reproducer. They are the debug flags provided by MLIR.  I tried to use them once and the flags did not work. I will be glad to have these flags enabled. 
As for Tian's concern about the interface, I think we will not, and should not, put these flags in our command line document. We just add a link to MLIR and say onnx-mlir accepts them.  It is part of MLIR functionalities.

caoimhinuibrian:
@tjingrant The way I see it, onnx-mlir-opt is usefule for running one or two passes, maybe starting from an intermediate stage in the compile process. This is certainly a good tool to have, and I don't think we should eliminate it. The other use-case is running the complete compile (or running from ONNX input up to some desired point), this is what onnx-mlir does. When running like this and trying to find a problem in the series of transformations, it is useful (actually necessary) to be able to print out the IR after all or some subset of passes so that one can inspect it to find erroneous transformations.

I don't know how to eliminate the extra output from the --help flag, which I agree is problematic.

Somewhat tangentially, I think this is what @CHUNERnubi is asking for in issue #506 .

caoimhinuibrian:
@tjingrant Maybe another solution would be to have onnx-mlir write out a pipeline description file which onnx-mlir-opt could then read in.

AlexandreEichenberger (approved):
LGTM, @caoimhinuibrian I let you decide if it is still your preferred way to solve this.

PR #520, opened 2021-02-14
Post build status as comments

gongsu832:
Since merged pull requests are closed and no longer update commit status,
we post merging and publishing status as comments on the pull request page.
They are viewable even if the pull requests are closed.


tjingrant (approved):
LGTM! Could you give some suggestions/guidelines on what to do in the event of a post-PR closure failure?

gongsu832:
@tjingrant Short answer, we shouldn't need to do anything.

Before a pull request is merged, it would have been tested by merging the master into the pull request so the final push on the master really shouldn't fail. If it did fail, it shouldn't be due to the source repo error but rather some other external factors such as running out of memory, disk space, etc. So the failure should be transient. When the merge build does fail, two things may need some attention:

- cleanup of related docker containers, images, and Jenkins workspaces. The Jenkinsfile has been written to handle this so we shouldn't need to worry about this, unless of course there are bugs in Jenkinsfile itself to cause the cleanup not being done properly.
- failure of publishing the docker images. This shouldn't be a big problem since it just means the docker images in the docker registry may not be built from the latest source repo. The next successful merge build of another pull request will correct that. In the rare event that the merge build was for updating the llvm-project commit sha1 and it failed to publish the llvm-project images, the build scripts have logic to rebuild the llvm-project images with updated commit sha1 locally instead of pulling them. So the impact is just that the build time for all the pull requests will be a bit longer until a successful merge build publishes the llvm-project images with the updated commit sha1.

BTW, you can force a pull request to publish all the docker images UNCONDITIONALLY by commenting on the pull request page with a phrase containing "publish this please", even after the pull request is closed. So if you REALLY want to immediately publish the docker images of a pull request in case the merge build fails and don't want to wait for the merge build of another pull request, you can just do that.

PR #522, opened 2021-02-16
Remove quotes when posting github status and comment

gongsu832:
'A quoted commit message that is longer than MAXLEN (24) characters'

When posting github status or comment, we take the first MAXLEN characters
to construct a one-liner, which might end in the middle of a quoted phrase
and cause JSON error in the REST call. This patch should fix that.


tjingrant (approved):
LGTM!

PR #523, opened 2021-02-16
Remove squeeze/unsqueeze pairs

tungld:
This patch adds DDR rules for removing squeeze/unsqueeze pairs.

### DIFF
--- a/src/Dialect/ONNX/ONNXOps.td.inc
+++ b/src/Dialect/ONNX/ONNXOps.td.inc
@@ -5223,6 +5223,7 @@ def ONNXSqrtOp:ONNX_Op<"Sqrt",
 
 def ONNXSqueezeOp:ONNX_Op<"Squeeze",
   [NoSideEffect, DeclareOpInterfaceMethods<ShapeInferenceOpInterface>]> {
+  let hasCanonicalizer = 1;
   let summary = "ONNX Squeeze operation";
   let description = [{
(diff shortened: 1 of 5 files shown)


AlexandreEichenberger (approved):
LGTM.

When I think of the squeeze or unsqueezed operation, I believe we do lower them to copy operations. But when it comes to it, aside from possibly needing a different buffer because the live range of the squeezed vs original buffer may be different, a squeeze operation is really a no-op as the data in the unsqueezed or squeezed location are exactly at the same offset. Does this PR eliminate these operations as well?

AlexandreEichenberger:
@tung, will squash this PR shortly, it would be great for you to add a response to my above question, thanks.

tungld:
@AlexandreEichenberger 

> Does this PR eliminate these operations as well?

No, this PR only deals with the situation where we squeeze and then untsqueeze with the same axes. It means the input and output have the same shape.

To avoid data copy (e.g. in squeeze, untsqueeze, reshape), I plan to use `memref_reshape`(https://mlir.llvm.org/docs/Dialects/Standard/#stdmemref_reshape-memrefreshapeop) which finally becomes no-op I believe. What do you think?

AlexandreEichenberger:
Yes, I think that would work, esp if the input parameter to squeeze is "dead" (meaning has no other uses)

PR #524, opened 2021-02-16
Map ONNX operator names to instructions

kile0:
This change maps the ONNX operator names (e.g. onnx.Add) to their corresponding MLIR instructions as  named location information.

Downstream, the operator names can be used to more accurately trace the kernel operations that were responsible for generating different MLIR instructions. To avoid extra code churn I only modified a subset of the common operators.

### DIFF
--- a/src/Conversion/ONNXToKrnl/ControlFlow/Loop.cpp
+++ b/src/Conversion/ONNXToKrnl/ControlFlow/Loop.cpp
@@ -23,7 +23,9 @@ struct ONNXLoopOpLowering : public ConversionPattern {
 
   LogicalResult matchAndRewrite(Operation *op, ArrayRef<Value> operands,
       ConversionPatternRewriter &rewriter) const final {
-    auto loc = op->getLoc();
+    auto loc = NameLoc::get(
+        Identifier::get(ONNXLoopOp::getOperationName(), op->getContext()),
+        op->getLoc());
     auto loopOp = dyn_cast<ONNXLoopOp>(op);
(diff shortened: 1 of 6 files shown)


tjingrant:
@kile0 could you format your code using provided style file https://github.com/onnx/onnx-mlir/blob/master/.clang-format ?

Many thanks for your help!

src/Conversion/ONNXToKrnl/ControlFlow/Loop.cpp
```diff
@@ -23,7 +23,9 @@ struct ONNXLoopOpLowering : public ConversionPattern {
 
   LogicalResult matchAndRewrite(Operation *op, ArrayRef<Value> operands,
       ConversionPatternRewriter &rewriter) const final {
-    auto loc = op->getLoc();
+    auto loc = NameLoc::get(
```
AlexandreEichenberger, line 26: Since this pattern is going to be used in a lot of place, would you be able to make it a templated function
``` C++
template<OP_TYPE>
Location ONNXLoc(Operation *op) { return NameLoc::get(OP_TYPE::getOperationName(), op->getContext()),
         op->getLoc()); }
```
(apologies if there are C++ error in the code above). You may need another version where the op type is OP_TYPE as sometime it might be a typed operation.

tjingrant:

caoimhinuibrian (approved):
LGTM, thanks for your contribution

caoimhinuibrian:

PR #527, opened 2021-02-24
add options for target triple and cpu (-mtriple and -mcpu) for LLVM

caoimhinuibrian:

### DIFF
--- a/src/MainUtils.cpp
+++ b/src/MainUtils.cpp
@@ -65,8 +65,8 @@ llvm::cl::opt<bool> useOnnxModelTypes("useOnnxModelTypes",
     llvm::cl::init(false), llvm::cl::cat(OnnxMlirOptions));
 
 llvm::cl::opt<string> mtriple("mtriple", llvm::cl::desc("Target architecture"),
-    llvm::cl::value_desc("<llvm target triple>"), llvm::cl::cat(OnnxMlirOptions),
-    llvm::cl::ValueRequired);
+    llvm::cl::value_desc("<llvm target triple>"),
+    llvm::cl::cat(OnnxMlirOptions), llvm::cl::ValueRequired);
 
 llvm::cl::opt<string> mcpu("mcpu", llvm::cl::desc("Target cpu"),
     llvm::cl::value_desc("<llvm cpu value>"), llvm::cl::cat(OnnxMlirOptions),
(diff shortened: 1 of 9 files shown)


caoimhinuibrian:
This is an example:
Command -

root@00b70efeab49:~/workspace# onnx-mlir --EmitLib -mtriple=s390x-ibm-loz -mcpu=z15 cos_op.onnx

Results in:

llc: error: : error: unable to get target for 's390x-ibm-loz -unknown-mcpu=z15', see --version and --triple.
llc -filetype=obj -relocation-model=pic --mtriple=s390x-ibm-loz --mcpu=z15 -o cos_op.o cos_op.bc

This is expected because we build LLVM with only host backend support, so on my mac the only targets are x86 triples and cpus.

src/MainUtils.cpp
```diff
@@ -288,6 +317,15 @@ void genConstPackObj(const mlir::OwningModuleRef &module,
...
 
+string getTargetOptions() {
+  string targetOptions = "";
+  if (mtriple != "")
+    targetOptions = "--mtriple=" + mtriple;
+  if (mcpu != "")
+    targetOptions += " --mcpu=" + mcpu;
+  return targetOptions;
+}
+
```
AlexandreEichenberger, line 328: It turns out that the same triple/cpu option does work for both the O3 and llc, so its great that the same option is passed for both.

src/MainUtils.cpp
```diff
@@ -321,6 +360,7 @@ void genModelObject(const mlir::OwningModuleRef &module, string bitcodePath,
   Command llvmToObj(/*exePath=*/!llcPath.empty() ? llcPath : kLlcPath);
   llvmToObj.appendStr("-filetype=obj")
       .appendStr("-relocation-model=pic")
+      .appendStr(getTargetOptions())
```
AlexandreEichenberger, line 363: Any reasons that we also add "-O3" here?

AlexandreEichenberger (approved):
LGTM, if you don't mind, I think we should also add the -O3 to llc, which does not appears to be on at this time.

Tx

src/MainUtils.cpp
caoimhinuibrian: the pipeline is to call opt to optimize the llvm IR (with -O3) and subsequently call llc to take the optimized bitcode and convert to machine code .... so, we don't need -O3 on the llc patrt

PR #529, opened 2021-02-25
update ONNX ConstantOp interface

chentong319:
Update ConstantOp to version 13 in onnx 1.8. The major change is 6 more attributes are add: float, list of floats, int, list of ints, string and list of strings.
This PR update the customized builder for constant and rewrite rules according to the new interface. Still, only the dense element attribute is generated  and consumed in onnx-mlir. No need to change analysis and transformation inside onnx-mlir. 
Details about changes:
1. Define util function getONNXConstantOpFromDenseAttr in ONNXOpsHelper.cpp
2. In rewrite rules, use this util function explicitly.
3. Put the definition in ONNXOps.td, instead of redundant copy in each .td file  
4. Convert constant with other attributes into dense_value in canonicalize pass
5. Test cases are added to check-onnx-lit


src/Builder/FrontendDialectHelper.cpp
```diff
@@ -126,8 +126,9 @@ mlir::Value InitializedTensorMapping::EmitInitializerForInputTensor(
       onnxTensorProtoToDenseElmAttr(builder, initializer);
 
   // Create ConstantOp for dense array.
-  return builder.create<mlir::ONNXConstantOp>(
-      loc, denseElmAttr.getType(), nullptr, denseElmAttr);
+  return builder.create<mlir::ONNXConstantOp>(loc, denseElmAttr.getType(),
```
AlexandreEichenberger, line 129: @chentong319 would there be a way to add a custom constructor with fewer parameters, where all the nullptr parameters below are set to null inside the constructor? I think there are example of custom builder already.

chentong319: Yes, I wrote that constructor and forgot to use it here.

src/Dialect/ONNX/ONNXOps.td.inc
```diff
@@ -657,10 +664,12 @@ def ONNXConstantOp:ONNX_Op<"Constant",
...
    if (value) {
     auto tensorType = value.getType();
-    build($_builder, $_state, tensorType, sparse_value, value);
+    build($_builder, $_state, tensorType, sparse_value, value,
+      FloatAttr(), ArrayAttr(), IntegerAttr(), ArrayAttr(), StringAttr(), ArrayAttr());
    } else {
     auto tensorType = sparse_value.getType();
-    build($_builder, $_state, tensorType, sparse_value, value);
+    build($_builder, $_state, tensorType, sparse_value, value,
+      FloatAttr(), ArrayAttr(), IntegerAttr(), ArrayAttr(), StringAttr(), ArrayAttr());
```
AlexandreEichenberger, line 672: @chentong319 Are the parameters after "value": these 6 attributes just null ptrs? If so, could we not have a custom builder where they are not needed as params?

Maybe a better question is when are they needed? From ONNX dag, there will be one that is used and the others are not?

This is just a small stylistic issue, so no worry

AlexandreEichenberger (approved):
LGTM, thanks for the welcome upgrade to version 13.

PR #530, opened 2021-02-25
Optimize memory consumption of constant propagation

tungld:
Constant propagation consumes much memory because it generates a lot of Attributes which are immortal. This patch optimizes memory consumption of constant propagation by using temporary files.

Some results when compiling resnet152:
- Memory consumption:
    - master branch: 7.1 GB
    - with this patch: 0.6 GB
- Constant propagation time:
    - master branch: 6 minutes 25 seconds
    - with this patch: 1 minutes 20 seconds


src/Transform/ONNX/ConstProp.cpp
```diff
@@ -38,508 +40,739 @@ namespace {
...
 // Then you need to add rules on how to transform the patterns; look into
 // ConstProp.td for example.
 //
 
-/// A helper function to contruct a RankedTensorType from a ShapedType.
-RankedTensorType constructRankedTensorType(ShapedType type) {
-  assert(type.hasRank() && "Not a ranked type");
-  return RankedTensorType::get(type.getShape(), type.getElementType());
-}
+const StringRef FILE_NAME_ATTR = "file_name";
```
AlexandreEichenberger, line 54: @tungld Can you comment on how saving in a separate file works with onnx-mlir-opt when we go from one phase to the next? Is that handled automatically, or do we need to do something separate about the file?

tungld: It is handled automatically and done inside a single phase of `constprop-onnx`. In a single phase of `constprop-onnx`, we create a file at the first time we reach a constant, and remove this file and return a constant with a dense attribute at the end of the phase. Users of onnx-mlir-opt do not see this file, but a normal constant op.

AlexandreEichenberger: I am not familiar with consprop-onnx: is that one phase of onnx-mlir? So you are saying that each time that constant-prop is called (and it may be called many times), a file is created, store temporary info in it, and then at the end we pluck from that file the constant that are still live, and put them back into mlir as dense attributes? 
In the old approach, you reported mem footprints of several GBs; is that the case now for this file? Or because it is used only for one phase of constant-prop, it is released between such phases and thus is smaller in general.

tungld: [`constprop-onnx`](https://github.com/onnx/onnx-mlir/blob/master/src/InitOMPasses.hpp#L26) is a pass in onnx-mlir. Creating a file attribute and removing it happen **inside** a single call of `constprop-onnx` pass only, it does not propagate to next calls.

>In the old approach, you reported mem footprints of several GBs; is that the case now for this file? Or because it is used only for one phase of constant-prop, it is released between such phases and thus is smaller in general.

Both approaches do not deal with sharing information between multiple calls of `constprop-onnx`, and `bin/onnx-mlir` command currently calls `constprop-onnx` once. 
The old approach is only for eliminating attribute creation when reading attribute values from constants, but dense attributes are still created for intermediate result constants, while this new approach uses external files for the intermediate result constants.

chentong319:
It is a surprise to me that using file is faster than using memory.  What's the fundermental reason for the inefficiency of using memory (namely Attribute) here? My impression is that the Attributes are not promptly freed by MLIR when they are not used any more. Is it true?
If we have to tackle this issue coming from MLIR design inside onnx-mlir, my view is to define a constant OP (not the one to match the onnx interface), say ConstantImplOp, This ConstantImplOp have the functionality this patch provides: operations with files under certain condition and flag control. And also copy the lowering functionality for  Krnl.Constant into this ConstantImplOp. So we can wrap all the file-related implementation for constant inside one dialect Op. In onnx-mlir, we only use ConstantImplOp after canonicalization.
My idea is still pre-mature. Since this patch does not have global impact, I am OK with this patch as a current solution. I assume there is a flag to turn it off easily.

tungld:
Hi @chentong319,

>It is a surprise to me that using file is faster than using memory

Running time improvement came from multiple reason, I think, such as: not creating each Attribute for each element of  tensor, inefficient recursion, etc. It's hard to say it was because of using files :)

>What's the fundermental reason for the inefficiency of using memory (namely Attribute) here? My impression is that the Attributes are not promptly freed by MLIR when they are not used any more. Is it true?

It's Attribute that is immortal by its design as discussed here in MLIR: https://llvm.discourse.group/t/attribute-storage-memory-consumption/2218

AlexandreEichenberger:
@tungld I think you may not quite understand the question of Tong: we understand that attributes are both slow and because their memory is never released, quite bad with respect to memory footprint. I believe the question that Tong is asking is why using a file vs a heap memory. Somewhere in your code, when looking for a temporary container for the constants, you do a "create file", and when that tensors becomes dead (either because its being combined with another constant tensor, generating yet another new tensor, or because it is being converted back to an attribute) you have a "remove file" command. So, if we replaced the create file by a malloc, and the remove file by a free, wouldn't that be faster because we don't have to go to disk? This is related to my earlier question of what is the max size of the temporary file space. Its probably less than the 7GB because not all variables are live at any instant of time.

tungld:
> I think you may not quite understand the question of Tong: we understand that attributes are both slow and because their memory is never released, quite bad with respect to memory footprint. I believe the question that Tong is asking is why using a file vs a heap memory. 

Oh, I see. Thanks for the clarification! My intention of using file was not for performance improvement, but to minimize memory consumption. So, I was confused when Tong mentioned about performance of using files vs. using memory.

There are several reasons. First, assume we use heap and we have N constants at the end of the propagation, we need to store these N constants in heap until the propagation finished (we don't know when a constant will be used or not later in order to remove it and create the final dense attributes). Heap memory for these N constants will be remained and released back to OS until the whole compilation finished as many online threads discussed about this problem. Second, for each constant in the N constants, each time we allocate it on heap, OS may search a buffer for reuse, and may double the buffer if necessary. At the end, the total buffer is quite large.

AlexandreEichenberger:
Generally, if there is heap memory that is not used, the associated memory get's paged out, so it is back to disk also, unless there is bad fragmentation on how the memory is allocated, with hot and cold memory subsections intermixed. So if you continue to feel strongly than saving to file is best, let's go with your solution; but let's hear from others too on this.

tungld:
@AlexandreEichenberger I did use heap before and did not manage to reduce memory consumption much. How about letting me move back to that version? Maybe I did some mistakes in memory management, so that memory was not released.

chentong319:
How about execution time? Did you allocate/free the memory in the same place you create and delete the files? I think you may keep the two implementation with flag control.

tungld:
Hi Tong,

> How about execution time?

The version that did not use files was using OMTensor to manage temporary tensors. Execution time is about 2minutes 40 seconds, memory consumption was about 4 GB. Again, I would say the poor performance was not because of not using files.

>  Did you allocate/free the memory in the same place you create and delete the files?

No, the version that did not use files used OMTensor, and OMTensor supports access elements in a convenient way using indices. In other words, the OMTensor-based version is not exactly same as the current patch. So it is difficult to compare. If everything is the same and only the difference is reading/writing from/to files vs heap,  I believe using heap is faster, never complain about that and don't want to rewrite the CS history :)

> I think you may keep the two implementation with flag control.
Yes, it's OK to provide the flag, and I need some additional works to use heap (malloc/free memory) in the current patch.

I just wanted to say the main purpose of this patch (as its title says) is to reduce memory consumption, and I am not sure about whether using heap can help reduce memory consumption as I haven't succeeded yet. As I know, we can reduce memory consumption by: 1) using external files 2) using stack 3) using mmap, and I chose using external files because it's easy to implement and performance is not so bad. Stack is not good for very large data in our case. mmap should be complicated. If you both strongly believe that using heap can help reduce memory consumption, I will try again and may need your help. Please let me know your decision. I am happy to try it. Thanks!

chentong319:
@tungld I just want to understand why we need to use file explicitly. As you pointed out previously, the root of this issue came from the implementation of Attribute in MLIR: its live range.  Therefore, I assumed naturally that your fix is to limit the live range of Attribute with self-managed memory or files. That's where my previous comment came from: the two implementation could be similar. Since you are using OMTensor and code might be quite different, it is not necessary to merge the two implementations together.
I hesitated on this patch because replacing variables in memory with files sounds like we want to do memory paging by ourself, but not just rely on OS.  I just wonder whether we run into such situation here. I hope I explained the origin of my concern clearly now. Your explanation helped me in understanding the problem.

AlexandreEichenberger:
I kind of agree with @chentong319 sentiments; but if it turns out that it is better to use files rather than memory, then I would go for the high-test performance.

We just need to be clear of one thing: absolute footprint is not necessarily the right metric: say I allocate 4GM and free it. It may be that the "nominal" footprint remains 4G until the end of the run, but since it is not in use, all the pages are paged out to disk, so it does not really matter. And on the other hand, if we have to read/write again and again to disk because each variable is saved on disk, that would appear to take more time. 

So if we want to settle this, it there an easy way to simply try both methods (managed memory pointers vs managed files) and see the difference in time?

tungld:
I added a flag to switch between using file and heap, and kept other parts the same.
Using file:
- Memory consumption: 723 MB, took 1m 4 seconds
Using heap
- Memory consumption: 723 MB, took 1m 3 seconds.

Good news is that memory consumption is the same as using file. It seemed somehow using OMTensor did not help reduce memory consumption, or I did something wrong. Performance is not significantly better.

So, please let me know the final decision: keep the flag, use heap only, use file only, or something else. Thanks!

AlexandreEichenberger (approved):
LGTM, my personal preference is to use the !USE_FILE option, what is yours @chentong319 @tungld ?
My recommendation would be to choose one and stick to that one, as #defined code is not tested

chentong319 (approved):

tungld:
Thanks for your comments! 

Please do not merge now, let me completely remove the files related part, and some redundant copies since we use memory.

tungld:
@AlexandreEichenberger @chentong319 I cleaned up file-related parts. It's ready to be merged once the checks passed. Thanks!

AlexandreEichenberger:
@tung, thank you so much for the work you did for this patch, really much much appreciated.

PR #532, opened 2021-02-28
Index Expression interface update

AlexandreEichenberger:
Updated the IndexExpr interface to be more like EDSC like. At the big picture level, I created two new data structure to capture memref bounds and array values as index expressions:

```mlir
  // Scan dimensions of A with/without transpose.
  MemRefBoundIndexCapture ABounds(A);
  if (op->transA() == 0) {
    aDims = {ABounds.getDim(0), ABounds.getDim(1)};
  } else {
    aDims = {ABounds.getDim(1), ABounds.getDim(0)};
  }
```

or 
```mlir
  ArrayValueIndexCapture startsCapture(genericOp, operandAdaptor.starts());
  for (uint64_t i = 0; i < sliceRank; i++) {
    int ii = axesIntLit[i];
    // Get start, end, step, and dim index expressions.
    // Get start.
    SymbolIndexExpr startInput(startsCapture.getSymbol(i));
   // ...
}
```

To create IndexExpr, we have now 5 subclasses of IndexExpr that are use solely to create Index Expressions. E.g.

``` mlir
    SymbolIndexExpr endInput(endsCapture.getSymbol(i));
    if (endInput.isUndefined())
      return op->emitError("end input parameter could not be processed");
    // Get step.
    SymbolIndexExpr stepInput(stepsCapture.getSymbol(i));
    if (stepInput.isUndefined())
      return op->emitError("step input parameter could not be processed");
    if (stepInput.isLiteral() && stepInput.getLiteral() == 0)
      return op->emitError("step input parameter cannot be zero");
    // Get dim.
    DimIndexExpr dimInput(dataBounds.getDim(ii));

   // create a literal
  LiteralIndexExpr three(3);
```

Scopes are now implicit, so they need to be created but don't need to be passed around. Internally, its a thread private.

``` mlir
  IndexExprScope outerScope; // outer scope 
  DimIndexExpr d1(dataBounds.getDim(2); // outer scope variable
  // ...
  {
     IndexExprScope innerScope;
     SymbolIndexExpr s1(d1); // in the inner scope, make a symbol out of the outer scope dim index expr d1.
    // ...
    // innerScope is deleted, and its enclosing scope becomes the active scope again.
}

// Back to the outer scope
```

In addition to the new interface, I also detected some redundant AffineExpr's dim and symbol being created. I changed the code to prevent the rewriting of these redundant symbols. Thus some lit tests were updated to reflect these changes. Otherwise, there were no changes in the code being produced as this PR is mainly about the IndexExpr interface.

### DIFF
--- a/src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
+++ b/src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
@@ -735,7 +735,7 @@ struct ONNXElementwiseBinaryOpLowering : public ConversionPattern {
     auto shapecomputed = shapeHelper.Compute(operands);
     (void)shapecomputed;
     assert(succeeded(shapecomputed));
-    IndexExprContext outerContext(shapeHelper.context);
+    IndexExprScope outerContext(shapeHelper.scope);
 
     // Insert an allocation and deallocation for the result of this operation.
     Value alloc = insertAllocAndDeallocSimple(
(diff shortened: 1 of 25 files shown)


doru1004 (approved):
LGTM

chentong319:
@AlexandreEichenberger I read your description of the PR. To my understanding, you introduced new subclass of IndexExpr and scope to address the issue we discussed previously.  As a result, the invariant symbols out side of the scope can be treated differently so that the result of multiplication or division from them can still be affine, as long as the computation is inserted outside of the scope.  Other changes are just interface improvement. If my understanding is roughly correct, I will go ahead read the code.
You may need to update the document for IndexExpr to make it easier to understand.

AlexandreEichenberger:
> @AlexandreEichenberger I read your description of the PR. To my understanding, you introduced new subclass of IndexExpr and scope to address the issue we discussed previously. As a result, the invariant symbols out side of the scope can be treated differently so that the result of multiplication or division from them can still be affine, as long as the computation is inserted outside of the scope. Other changes are just interface improvement. If my understanding is roughly correct, I will go ahead read the code.

@chentong319 you got it, added new subclass to make reuse of enclosing scope's index expression in current scope more straightforward. Note that scope is not really new, it used to be called "context", but scope is the more appropriate name (and it matches what EDSC also uses). I also made it implicit, like in EDSC, for example, and like the "insertion points" which are also implicit. It just got too heavy to pass/use the scope explicitly for each each explicit creation of index expression variables.

> You may need to update the document for IndexExpr to make it easier to understand.

Indeed; I did a significant rewrite of the doc in the IndexExpr.hpp. Feel free to suggest additional info int there.  At some later  time, we should have better documentation for this as well as krnl dialect. Totally agree.

chentong319 (approved):
LGTM!

PR #538, opened 2021-03-03
IndexExpr better support for Predicate literals

AlexandreEichenberger:
Better support for predicate literals. Before we could not create them. It turns out that it is useful to generate code. Now you may for example add

```C++
  PredicateIndexExpr compare(true);
```

Also added IndexExpr operators for "and", "or", and "not" for predicates. Note that the arithmetic operator is used as this is not the disjunctive logics used in C/C++ as these require branches. Its just the and/or-ing of the true/false values.
 
Also fix a bug for compares, where `<=` was mistakenly implemented by `<`. Changes in the lit tests represent the results of this fix.


doru1004 (approved):
LGTM

PR #540, opened 2021-03-06
Further changes to IndexExpr

AlexandreEichenberger:
IndexExpr were generated in a lazy manner, generating new values only when needed in other computations. This became a problem when having index expressions used in if-the-else blocks where a lazy def may have materialized in the "then" and thus be unavailable elsewhere. 

This fix changes that; as a result, some of the lit tests had to be modified and/or moved to the "--canonicalize" version of their tests.


src/Conversion/ONNXToKrnl/Tensor/Gather.cpp
```diff
@@ -64,15 +64,16 @@ struct ONNXGatherOpLowering : public ConversionPattern {
     int jIndexStart = iIndexStart + axisLit;
     int kIndexStart = jIndexStart + indicesRank - (axisLit + 1);
 
-    // Load the constants for the tests
+    // Insert code inside the loop.
+    rewriter.setInsertionPointToStart(outputLoops.getIterateBlock());
+    IndexExprScope innerLoopScope;
```
chentong319, line 69: No usage of innerLoopScope?

AlexandreEichenberger: @chentong319  

That is the new use; scopes are used implicitly, just like insertion points. They use a thread private pointers, and so do I now.

chentong319: Can we have a better way, such as a static method? To me, it is against intuition to just declare a local variable.

src/Dialect/ONNX/IndexExpr.hpp
```diff
@@ -458,8 +407,10 @@ class IndexExpr {
   bool isSymbol() const;
   bool isDim() const;
   bool isPredType() const;
-  bool isIndexType() const { return !isPredType(); }
-  bool isShapeInferencePass() const;
+  bool isIndexType() const;
+  bool isShapeInferencePass() const {
+    return getScope().isShapeInferencePass();
+  }
```
chentong319, line 413: When I read the fold function for canonicalization, I feel that isShapeInferencePass might be generalized to something like isForFolding. Just comment, not need to change now.

src/Dialect/ONNX/IndexExpr.hpp
```diff
@@ -677,6 +630,26 @@ class ArrayValueIndexCapture {
   bool hasDefault;
 };
 
+// Capture array of values given by attributes.
+class ArrayAttributeIndexCapture {
```
chentong319, line 634: I did not find the usage of this class.

src/Dialect/ONNX/IndexExpr.hpp
AlexandreEichenberger: Right, maybe the name is misleading. To indexExpr, is is more relevant to know if there is a rewriter and operations can be generated or not. Right now, inference is the only place where we cannot generate new operations, just update the type. If you have a better name, I am happy to use it.

src/Dialect/ONNX/IndexExpr.hpp
AlexandreEichenberger: Good catch, it is used in my support for matmul. I am trying to separate the IndexExpr from the Krnl.matmul PRs, so that they may be easier to digest.

src/Conversion/ONNXToKrnl/Tensor/Gather.cpp
AlexandreEichenberger: No, it need to be a variable; when created, it add itself to the top context, and when removed, it reinstall the parent context.

It is the same pattern as the EDSC scope.

chentong319 (approved):
LGMT

PR #541, opened 2021-03-09
Lower optimization time

doru1004:
This patch lowers the time it takes to apply the memory pool optimization pass.

The  reuse analysis on a given getref aims to find all the reuses in one application of the pattern. This means that subsequent application to an already optimized getref will not yield new reuses so we can exit early.

For resnet50 this patch lowers the runtime of the optimization pass from 73s to 13s.


AlexandreEichenberger (approved):
Bravo! Please ask the folks running the larger resnet to evaluate to see if it is acceptable there as well.

PR #542, opened 2021-03-09
clean PadConstant ops

chentong319:
At the beginning of the project, I tried to combine Pad operation with certain constant input to generate new operations. Now this approach is thought to be unnecessary since we already have the code to handle the constant input for Pad operation directly. This patch is intended to clean all the related code:
1. Dialect definition
2. Rewrite rule
3. Shape inference
4. Lowering
5. tests


AlexandreEichenberger (approved):
LGTM, can you just confirm that we did not loose functionality, or if we did, can you just state it so that we may add it in the future again

PR #543, opened 2021-03-10
Fix a wrong equation in the LRN lowering

tungld:
LRN equation in the comment is correct but implementation is incorrect. So all inception models in the model zoo which use LRN would be failed. This patch is to fix this problem and enable an inception model.


src/Conversion/ONNXToKrnl/Math/LRN.cpp
```diff
@@ -72,14 +72,15 @@ struct ONNXLRNOpLowering : public ConversionPattern {
     Value cValue = outputLoops.getInductionVar(loopIndexForC);
     DimIndexExpr cIE(cValue);
     MemRefBoundIndexCapture inputBounds(input);
-    DimIndexExpr sizeIE(inputBounds.getDim(loopIndexForC));
+    DimIndexExpr CIE(inputBounds.getDim(loopIndexForC));
+    SymbolIndexExpr sizeIE = LiteralIndexExpr((int)sizeLit);
```
chentong319, line 76: Attribute size should be int. Could you change the initialization of sizeLit?

tungld: Done. Thanks for your comment!

chentong319 (approved):
LGTM! Thanks for fixing the bug.

PR #546, opened 2021-03-15
Clean attribute promote

chentong319:
The experimental implementation to promote constant input into attributes turned out to be not necessary. Instead, we can uniformly check the definition of the input. This patch cleanse the code related to attribute promotion:
1. generation of ONNXOps.td.inc
2. Interface for constant promotion
3. pass for constant promotion
4. Shape inference and lowering that uses  promoted attribute
5. test cases


AlexandreEichenberger (approved):
LGTM, that was a lot of changes, thanks for doing it.

PR #547, opened 2021-03-16
Use MLIR_SYSTEM_LIBRARIES to link against the required system libraries

sstamenova:
Several system library dependencies that are needed are not included in a consistent manner. This change introduces an MLIR_SYSTEM_LIBRARIES variable (which could also be just SYSTEM_LIBRARIES) to collect the system libraries coming from mlir as dependencies and then add them as appropriate.

This also removes the requirement for the curses library on Windows - it is not actually needed since llvm does not use it on Windows.


AlexandreEichenberger:
@caoimhinuibrian or @gongsu832 do you mind looking into this PR? thanks

gongsu832:
LGTM

sstamenova:
Can someone commit this if it's approved? Thanks!

caoimhinuibrian (approved):
LGTM

AlexandreEichenberger:
Thanks @sstamenova for the contribution, and @gongsu832 and @caoimhinuibrian for the review. Much appreciated

PR #549, opened 2021-03-17
Change the equation for validating calculation results in numerical tests

imaihal:
The equation for validating calculation results in numerical tests is different from onnx backend test. This PR changes the equation to use the same one with onnx backend test. I think this equation is generally used, but is there any specific reason to use current one?

Onnx backend test uses `absolute(a - b) <= (atol + rtol * absolute(b))` by using [`testing.assert_allclose`](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html) . Source code is [here.](https://github.com/onnx/onnx/blob/v1.7.0/onnx/backend/test/runner/__init__.py#L174) Equation is written [here]( https://numpy.org/doc/stable/reference/generated/numpy.allclose.html)

When I changed ATOL and RTOL both in onnx backend test and numerical test to meet my requirement, I got errors only in numerical test. So, I found this difference. 

### DIFF
--- a/src/Runtime/OMTensor.inc
+++ b/src/Runtime/OMTensor.inc
@@ -473,7 +473,7 @@ inline bool omTensorAreTwoOmtsClose(
   // (rtol * b + atol) - abs(a - b)
   std::transform(eqAllclose.begin(), eqAllclose.end(), absoluteDiff.begin(),
       eqAllclose.begin(), std::minus<>());
-  bool satisfied = all_of(
+  bool satisfied = std::all_of(
       eqAllclose.begin(), eqAllclose.end(), [&](T eq) { return eq >= 0; });
 
   if (satisfied) {
--- a/src/Runtime/OMTensor.inc
+++ b/src/Runtime/OMTensor.inc
@@ -445,24 +445,38 @@ inline bool omTensorAreTwoOmtsClose(
     return false;
   }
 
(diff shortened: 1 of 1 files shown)


src/Runtime/OMTensor.inc
```diff
@@ -445,24 +445,38 @@ inline bool omTensorAreTwoOmtsClose(
...
+      (T *)b->_allocatedPtr, static_cast<T (*)(T)>(&std::abs));
+  std::transform((T *)b->_allocatedPtr, (T *)b->_allocatedPtr + bnum,
+      (T *)rtolT->_allocatedPtr, eqAllclose.begin(), std::multiplies<>());
+  // rtol * abs(b) + atol
+  std::transform(eqAllclose.begin(), eqAllclose.end(),
+      (T *)atolT->_allocatedPtr, eqAllclose.begin(), std::plus<>());
+  // (rtol * b + atol) - abs(a - b)
+  std::transform(eqAllclose.begin(), eqAllclose.end(), absoluteDiff.begin(),
+      eqAllclose.begin(), std::minus<>());
+  bool satisfied = all_of(
```
chentong319, line 476: Use std::all_of to keep the code consistency with std::transform

imaihal: Fixed. Thanks!

chentong319 (approved):
LGTM!

PR #551, opened 2021-03-20
Set ONNX_USE_PROTOBUF_SHARED_LIBS and Protobuf_USE_STATIC_LIBS 

sstamenova:
The onnx build on windows requires the provision of either static or shared libs for protobuf and to achieve that there are several different flags that need to be set in protobuf, onnx as well as cmake (for find_package). At the very minimum ONNX_USE_PROTOBUF_SHARED_LIBS and Protobuf_USE_STATIC_LIBS need to be opposites of each other, but onnx does not currently set them up that way (onnx doesn't set Protobuf_USE_STATIC_LIBS at all and cmake interprets that as setting Protobuf_USE_STATIC_LIBS to OFF). If/when onnx is updated to correctly set up both variables, this can be removed (see https://github.com/onnx/onnx/issues/3345).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,9 +27,38 @@ if (MSVC)
   # like onnx
   set(CMAKE_CXX_FLAGS "-wd4244 -wd4267 -wd4530 -wd4624 /bigobj")
 endif()
+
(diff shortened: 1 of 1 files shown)


caoimhinuibrian (approved):
LGTM

PR #554, opened 2021-03-22
Signature

caoimhinuibrian:


caoimhinuibrian:
The signature is created and stored in a global byte array, zero terminated. To interrogate it one needs a function that refers to _signature and which is linked with the .so file for the model.

caoimhinuibrian:
#include <OnnxMlirRuntime.h>
#include <stdio.h>
#include <string.h>

extern char _signature;

int getModelSignatureSize() {
  return strlen(&_signature);
}

void getModelSignature(char * buffer) {
  int size = strlen(&_signature);
  memcpy(buffer, &_signature, size);
}


OMTensorList *run_main_graph(OMTensorList *);
int main() {
  /***
  *   Bind this file with addop.so (adds two vectors). It will print the Model signature
  ***/
  int signatureSize = getModelSignatureSize();
  void *sig = malloc(signatureSize);
  getModelSignature(sig);

  printf("Model Signature = %s\n",sig);
  free(sig);

  // Shared shape & rank.
  int64_t shape[] = {2, 2};
  int64_t rank = 2;
  // Construct x1 omt filled with 1.
  float x1Data[] = {1., 1., 1., 1., 1., 1.};
  int64_t *x1Shape = {2, 2};
  OMTensor *x1 = omTensorCreate(x1Data, shape, rank, ONNX_TYPE_FLOAT);
  // Construct x2 omt filled with 2.
  float x2Data[] = {2., 2., 2., 2., 2., 2.};
  int64_t *x2Shape = {2, 2};
  OMTensor *x2 = omTensorCreate(x2Data, shape, rank, ONNX_TYPE_FLOAT);
  // Construct a list of omts as input.
  OMTensor *list[2] = {x1, x2};
  OMTensorList *input = omTensorListCreate(list, 2);
  // Call the compiled onnx model function.
  OMTensorList *outputList = run_main_graph(input);
  // Get the first omt as output.
  OMTensor *y = omTensorListGetOmtByIndex(outputList, 0);
  float *outputPtr = (float *)omTensorGetDataPtr(y);
  // Print its content, should be all 3.
  for (int i = 0; i < 6; i++)
    printf("%f ", outputPtr[i]);
  return 0;
}

AlexandreEichenberger:
@Kevin, I have seen a similar behavior: I had a patch that worked perfectly fine on Mac, but got all sorts of compile errors (symbols not found) on the Linux... in my case, the cause of the errors were namespaces, which worked flawlessly on Mac but could not get anything going on the Linux. And because no docker was created on compile failure, I was unable to really see what was up. It may/may not be the same

test/numerical/TestRNN.cpp
```diff
@@ -107,9 +107,11 @@ bool isOMRNNTheSameAsNaiveImplFor(const int direction, const int S, const int B,
 
   // Emit the entry point operation which specifies the number of user
   // inputs and outputs.
+  std::string signature("");
```
AlexandreEichenberger, line 110: I added a test: TestMatMul2D, do you mind to add the same thing to it. That is probably why the compilation failed...

src/Dialect/ONNX/ONNXOps.td
```diff
@@ -74,15 +74,16 @@ def ONNXEntryPointOp: ONNX_Op<"EntryPoint"> {
     The "onnx.EntryPoint" function indicates the main entry point of ONNX model.
   }];
 
-  let builders = [OpBuilderDAG<(ins "FuncOp":$function, "int":$numInputs, "int":$numOutputs)>];
+  let builders = [OpBuilderDAG<(ins "FuncOp":$function, "int":$numInputs, "int":$numOutputs, "std::string":$signature)>];
```
AlexandreEichenberger, line 77: Minor, if you could format to roughly be 80 chars

AlexandreEichenberger (approved):
LGTM, please add the signature to the file I added as part of my patch TestMatMul2D, same deal as you did for TestCONV, for example.

AlexandreEichenberger:
Would be great if you could add a lit test.

caoimhinuibrian:

src/Dialect/ONNX/ONNXOps.td
caoimhinuibrian: done

test/numerical/TestRNN.cpp
caoimhinuibrian: done

caoimhinuibrian:

PR #556, opened 2021-03-24
Fix memory leak with C/C++ interface

tungld:

Memory leak issue was originally reported by @seroyer. Thanks, Steven!

Test program (for MNIST inference): 
```mlir
OMTensor *tensor = omTensorCreate(input, shape.data(), shape.size(), ONNX_TYPE_FLOAT);
OMTensor *tensorList[] = {tensor};
OMTensorList *tensorListIn = omTensorListCreate(tensorList, 1);

OMTensorList *tensorListOut = run_main_graph(tensorListIn);

omTensorListDestroy(tensorListOut);
omTensorListDestroy(tensorListIn);
```

Using valgrind to check memory leak: `valgrind --leak-check=full ./memcheck`
- run one inference:
```
==372468== LEAK SUMMARY:
==372468==    definitely lost: 8 bytes in 1 blocks
==372468==    indirectly lost: 0 bytes in 0 blocks
==372468==      possibly lost: 0 bytes in 0 blocks
==372468==    still reachable: 0 bytes in 0 blocks
==372468==         suppressed: 0 bytes in 0 blocks
```

- run 100 inferences, the leak grows by each call.
```
==372469== LEAK SUMMARY:
==372469==    definitely lost: 800 bytes in 100 blocks
==372469==    indirectly lost: 0 bytes in 0 blocks
==372469==      possibly lost: 0 bytes in 0 blocks
==372469==    still reachable: 0 bytes in 0 blocks
==372469==         suppressed: 0 bytes in 0 blocks
```

There are two issues that cause the leak:
1) Global constant
In the generated code, we malloc memory for global constant but not free.
```
llvm.mlir.global internal @packedConst() : !llvm.ptr<i8>
%17 = llvm.mlir.constant(0 : i64) : i64
%18 = llvm.call @getEmbeddedConstPool(%17) : (i64) -> !llvm.ptr<i8>
```
In particular, memory is allocated (by malloc) inside getEmbeddedConstPool and pointed to by `%18`, but there is no intrinsic to free `%18`.

2) OMTensorList
In the generated code, we malloc memory for the list of OMTensor pointers which is to store output OMTensorList.
```
%26 = llvm.call @malloc(%25) : (i64) -> !llvm.ptr<i8>
```
This memory is expected to be destroyed by `omTensorListDestroy`. Document for `omTensorListDestroy` said the memory will be released but there is no code for it.

This patch fixes the two above issues:
- for issue 1: Emit `free` for the global constant if the global constant exists.
- for issue 2: Add a member of OMTensorList to indicate the OMTensorList owns the pointer to OMTensor array or not, and the pointer will be freed if owning. By default, OMTensorList does not own the pointer.

With this patch, we do not see memory leak:
- run 100 inferences:
```
==373203== HEAP SUMMARY:
==373203==     in use at exit: 0 bytes in 0 blocks
==373203==   total heap usage: 1,625 allocs, 1,625 frees, 12,184,032 bytes allocated
==373203==
==373203== All heap blocks were freed -- no leaks are possible
```

### DIFF
--- a/src/Conversion/KrnlToLLVM/KrnlToLLVM.cpp
+++ b/src/Conversion/KrnlToLLVM/KrnlToLLVM.cpp
@@ -877,8 +877,10 @@ public:
     }
 
     // Create wrapped output.
+    auto one = rewriter.create<LLVM::ConstantOp>(
+        loc, int32Ty, rewriter.getI32IntegerAttr(1));
     auto wrappedOutput = callApi(rewriter, loc, apiRegistry,
-        API::CREATE_OMTENSOR_LIST, {outOmtPtrsArr, numOutput});
+        API::CREATE_OMTENSOR_LIST, {outOmtPtrsArr, numOutput, one});
 
     // Clean the global constant.
     auto globalBase = module.lookupSymbol<LLVM::GlobalOp>("packedConst");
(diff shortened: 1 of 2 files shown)


caoimhinuibrian (approved):
LGTM

PR #558, opened 2021-03-25
[cmake] Allow pointing to LLVM without using environment variables

stephenneuendorffer:
When integrating as an external project, this becomes awkward.
It's generally preferable to use cmake variables to configure
things like this.  This patch extends the existing mechanism to
allow the use of cmake variables.

### DIFF
--- a/MLIR.cmake
+++ b/MLIR.cmake
@@ -2,29 +2,33 @@
 
 # Path to LLVM source folder.
 if(DEFINED LLVM_PROJ_SRC)
+  # Just use it...
 elseif(DEFINED ENV{LLVM_PROJ_SRC})
   set(LLVM_PROJ_SRC $ENV{LLVM_PROJ_SRC})
 else()
-  message(FATAL_ERROR "env variable LLVM_PROJ_SRC not set")
+  message(FATAL_ERROR "LLVM_PROJ_SRC is not configured.  Please set the env variable "
+  "LLVM_PROJ_SRC or the corresponding cmake configuration option to reference an LLVM source tree.")
 endif()
 if(EXISTS ${LLVM_PROJ_SRC})
-message(STATUS "LLVM_PROJ_SRC           : " ${LLVM_PROJ_SRC})
+  message(STATUS "LLVM_PROJ_SRC           : " ${LLVM_PROJ_SRC})
 else()
-message(FATAL_ERROR "The path specified by LLVM_PROJ_SRC does not exist: "
+  message(FATAL_ERROR "The path specified by LLVM_PROJ_SRC does not exist: "
         ${LLVM_PROJ_SRC})
 endif()
 
 # Path to LLVM build folder
 if(DEFINED LLVM_PROJ_BUILD)
+  # Just use it...
 elseif(DEFINED ENV{LLVM_PROJ_BUILD})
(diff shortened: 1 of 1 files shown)


MLIR.cmake
```diff
@@ -1,30 +1,32 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Path to LLVM source folder.
-if(DEFINED ENV{LLVM_PROJ_SRC})
+if(DEFINED LLVM_PROJ_SRC)
```
sstamenova, line 4: Is there a reason to have to separate statements both here and below? I think it would be better to say:
```if(NOT DEFINED LLVM_PROJ_SRC AND DEFINED ENV{LLVM_PROJ_SRC})```

File: MLIR.cmake [hunk 2]
```diff
@@ -1,30 +1,32 @@
...
+elseif(DEFINED ENV{LLVM_PROJ_SRC})
   set(LLVM_PROJ_SRC $ENV{LLVM_PROJ_SRC})
-  if(EXISTS ${LLVM_PROJ_SRC})
-    message(STATUS "LLVM_PROJ_SRC           : " ${LLVM_PROJ_SRC})
-  else()
-    message(FATAL_ERROR "The path specified by LLVM_PROJ_SRC does not exist: "
-            ${LLVM_PROJ_SRC})
-  endif()
 else()
   message(FATAL_ERROR "env variable LLVM_PROJ_SRC not set")
```
sstamenova, line 8: This message (below as well) is not longer accurate since now the variable can be specified as an environment variable through a cmake file or as part of the cmake command with -DLLVM_PROJ_SRC=...

stephenneuendorffer: good point...

MLIR.cmake
stephenneuendorffer: I'm not sure that I follow... isn't this equivalent to `elseif(DEFINED ENV{LLVM_PROJ_SRC})`?

sstamenova: Right now you have:

```
if(DEFINED LLVM_PROJ_SRC)
elseif(DEFINED ENV{LLVM_PROJ_SRC})
```

I am suggesting that you change your statements to avoid the first (empty) if condition:

```
if(NOT DEFINED LLVM_PROJ_SRC AND DEFINED ENV{LLVM_PROJ_SRC})
  set(LLVM_PROJ_SRC $ENV{LLVM_PROJ_SRC})
elseif(NOT DEFINED LLVM_PROJ_SRC)
  message(FATAL_ERROR "env variable LLVM_PROJ_SRC not set")
endif()
```

or something similar.

chentong319: Do you mean this?
```
if(NOT DEFINED LLVM_PROJ_SRC AND DEFINED ENV{LLVM_PROJ_SRC})
  set(LLVM_PROJ_SRC $ENV{LLVM_PROJ_SRC})
endif()
if(NOT DEFINED LLVM_PROJ_SRC)
  message(FATAL_ERROR "env variable LLVM_PROJ_SRC not set")
endif()
```

stephenneuendorffer: Looking closer at this, is seems like the condition has to appear somewhere in order to have the error message work, so it's not clear to me that changing this actually makes it easier to understand.

caoimhinuibrian:

chentong319 (approved):
LGTM!

PR #566, opened 2021-03-31
Support opaque attribute and alignment in KrnlGlobalOp

tungld:
This patch adds opaque attribute and alignment attribute to KrnlGlobalOp, which is to support constants with alignment and opaque constants.

Opaque constant is used when we want to store data that compiler does not care about its content, e.g data in specific format in deep learning.


AlexandreEichenberger (approved):
LGTM

PR #573, opened 2021-04-03
[WIP, strange namespace errors on Linux] Krnl matmul, enabling dynamic shapes

AlexandreEichenberger:
An issue that prevented the use of the optimized matmul for dynamic shape is fixed. 

Also, I changed how we use EDSC to be more inline with MLIR, which uses 2 scopes: edsc and edsc::intrinsics to activate the use of EDSC. Intrinsics are also in their own header/cpp file.

Finally, there were a lot of included header files that were not necessary; I removed as many as I could find in the lowering of krnl to affine.

Finally, MLIR uses an "s" in `MemRefBoundsCapture`, I added it too for our `MemRefBoundsIndexCapture` to increase familiarity between onnx-mlir and mlir


AlexandreEichenberger:
Have no idea why, but the build that works on Mac fails because of namespaces on linux and other places.

PR #574, opened 2021-04-03
Added support for matmul with unkown dim

AlexandreEichenberger:

Added support for matmul with unkown dim

Also change a name to be more inline with what MLIR uses


doru1004 (approved):
LGTM

Issue #568, opened 2021-04-01
parallel interference cause crash


shizy818:
Hello, we've done parallel predictions on the compiled model by `onnx-mlir --EmitLib`. Unfortunately it crashed:
```
g++ main.cc -std=c++11 -pthread -lbenchmark libmodel.so -o main && ./main
main() : creating thread, 0
main() : creating thread, 1
main() : creating thread, 2
main() : creating thread, 3
main() : creating thread, 4
main() : creating thread, 5
main() : creating thread, 6
main() : creating thread, 7
main() : creating thread, 8
main(1482,0x700002825000) malloc: *** error for object 0x7ff007809400: pointer being freed was not allocated
main(1482,0x7000028a8000) malloc: *** error for object 0x7ff008009400: pointer being freed was not allocated
main(1482,0x7000029ae000) malloc: *** error for object 0x7ff008809400: pointer being freed was not allocated
main(1482,0x700002825000) malloc: *** set a breakpoint in malloc_error_break to debug
main(1482,0x7000028a8000) malloc: *** set a breakpoint in malloc_error_break to debug
main() : creating thread, 9
main(1482,0x7000029ae000) malloc: *** set a breakpoint in malloc_error_break to debug
main(1482,0x70000292b000) malloc: *** error for object 0x7ff008809400: pointer being freed was not allocated
main(1482,0x70000292b000) malloc: *** set a breakpoint in malloc_error_break to debug
```

Is it a known issue? If needed, I can upload the test code. Thank you!

tjingrant:
@shizy818 this is not a case we've tested, could you post the content of main.cc?

gongsu832:
Running under gdb and produce a backtrace would also help.

shizy818:
main.cc
```cc
#include <OnnxMlirRuntime.h>
#include <stdio.h>
#include <benchmark/benchmark.h>
#include <pthread.h>
using namespace std;
#define NUM_THREADS 10
extern "C" OMTensorList *run_main_graph(OMTensorList *);

static void *run(void *args)
{
  // Perform setup here
  int64_t shape[] = {1, 4}; // NHWC
  int64_t rank = 2;

  // Construct x1 omt filled with 1.
  int sz = 1 * 4;
  float x1Data[sz];
  for (int i = 0; i < sz; i++)
    x1Data[i] = 0.5f;
  OMTensor *x1 = omTensorCreate(x1Data, shape, rank, ONNX_TYPE_FLOAT);

[...]

PR #577, opened 2021-04-05
[WIP] Add two functions for loading and destroying packed constants

tungld:


tjingrant:
Could you update the documentation example here, too?
https://github.com/onnx/onnx-mlir/blob/1d0c09d2cac9ff050f383f2c7220a8b1a9ea2458/include/OnnxMlirRuntime.h#L73

Thanks!

tungld:
> Could you update the documentation example here, too?

Done. Thanks for your comment!

seroyer:
Do we need changes to the jni code to allow passing through these new calls?

caoimhinuibrian:
I have a queasy feeling about this approach. To solve an internal problem, we are pushing an API change out to the end user. Is this really what we want to do?

tjingrant:
I think there're two issues to be addressed:

- Parallel inference segfault.
- Constant pool gets copied per call to inference function.

The answer may very well be one that solves both with one patch, but I agree with @caoimhinuibrian that we need to look at the first principles and examine the design rationales to decide. I think we having a meeting with Tung to discuss precisely this.

tungld:
> To solve an internal problem, we are pushing an API change out to the end user. Is this really what we want to do?

I agree, we don't like the change unless it is the right thing to do. It's not really an internal problem when we want to provide the best performance for different use cases.

Though this patch is not the final approach and we'll have a meeting for that, I'd like to clarify that if users use ExecutionSession in C++ or Python, these load and destroy are hidden inside the constructor and destructor of ExecutionSession. So, no change out to the end user.

gongsu832:
I'm don't know much about the internals of the compiler but let me just play devils advocate. The segfault is clearly an internal problem and the solution to it shouldn't require any user visible change. The constant pool problem sounds like an internal problem as well, at least from what your patch is trying to address. Why can't the `load_constants`/`destroy_constants` be sucked into `run_main_graph`, and use a reference counter to load/destroy only once?

tungld:
> Why can't the load_constants/destroy_constants be sucked into run_main_graph, and use a reference counter to load/destroy only once?

Do you mean we will have a race condition on the reference counter? If so, it introduces overhead, for example in the parallel example in issue #568.

gongsu832:
> > Why can't the load_constants/destroy_constants be sucked into run_main_graph, and use a reference counter to load/destroy only once?
> 
> Do you mean we will have a race condition on the reference counter? If so, it introduces overhead, for example in the parallel example in issue #568.

No. I meant we should be able to suck `load_constants`/`destory_constants` into `run_main_graph` with a reference counter. Concurrent updating reference counter is a no issue with a simple compare-and-swap or atomic add/subtract.

tungld:
> No. I meant we should be able to suck load_constants/destory_constants into run_main_graph with a reference counter. Concurrent updating reference counter is a no issue with a simple compare-and-swap or atomic add/subtract.

I see. Thanks! In this case, how do we know the last use of run_main_graph to clean up the constants automatically?

caoimhinuibrian:
Why do we copy the constants into a buffer? Why can't we use them in-place? This would solve the segfault and avoid the performance overhead of copying a potentially large piece of memory. It would also reduce the footprint by eliminating the buffer.

tungld:
> Why do we copy the constants into a buffer? Why can't we use them in-place?

@tjingrant any thought on this? it's related to the original design of the packed constants.

gongsu832:
> > No. I meant we should be able to suck load_constants/destory_constants into run_main_graph with a reference counter. Concurrent updating reference counter is a no issue with a simple compare-and-swap or atomic add/subtract.
> 
> I see. Thanks! In this case, how do we know the last use of run_main_graph to clean up the constants automatically?

Each call to `run_main_graph` atomically increment/decrement the counter. Only the first call (when counter goes from 0 to 1) and the last call (when counter goes from 1 to 0) will do whatever setup/teardown you need.

tjingrant:
> Why do we copy the constants into a buffer? Why can't we use them in-place? This would solve the segfault and avoid the performance overhead of copying a potentially large piece of memory. It would also reduce the footprint by eliminating the buffer.

We may be able to get away with copying into a buffer for Mac & Linux. But on platforms like Windows, our fallback/default path (in which weights are stored as files on a harddrive) can't take place without allocating a buffer and then reading file content into it. So there may be value in designing our runtime interface for the most generic case possible.

tungld:
> Each call to run_main_graph atomically increment/decrement the counter. Only the first call (when counter goes from 0 to 1) and the last call (when counter goes from 1 to 0) will do whatever setup/teardown you need.

May I confirm my understanding with the following examples?
With a single call to run_main_graph:
```
run_main_graph(...);
```
Its the first and last call, so the counter is increased to 1 and decreased to 0, and the cleanup is performed.

but, with three calls like this:
```
run_main_graph(...);
run_main_graph(...);
run_main_graph(...);
```
I feel we load and destroy the constants at each call. In other words, how do we know during the first call to run_main_graph that we will have another calls so that it does not decrease the counter to 0 and cleans up the constants? Sorry for many questions, I just wanted to understand your approach.

caoimhinuibrian:
look at PR 581. The code runs without segfaaulting, so I suspect we don't need to copy the constants.

caoimhinuibrian:
> > Each call to run_main_graph atomically increment/decrement the counter. Only the first call (when counter goes from 0 to 1) and the last call (when counter goes from 1 to 0) will do whatever setup/teardown you need.
> 
> May I confirm my understanding with the following examples?
> With a single call to run_main_graph:
> 
> ```
> run_main_graph(...);
> ```
> 
> Its the first and last call, so the counter is increased to 1 and decreased to 0, and the cleanup is performed.
> 
> but, with three calls like this:
> 
> ```
> run_main_graph(...);
> run_main_graph(...);
> run_main_graph(...);
> ```
> 
> I feel we load and destroy the constants at each call. In other words, how do we know during the first call to run_main_graph that we will have another calls so that it does not decrease the counter to 0 and cleans up the constants? Sorry for many questions, I just wanted to understand your approach.

if the second call enters before the first exits, it will up the count. If it doesn't then its ok for the first to free - when the second does enter it will realloc ...just need the locks to be correct

gongsu832:
Here is an example using gcc atomic built in. There is a global `refcnt` initialized to 0. Each `run_main_graph` will do the following:
```
run_main_graph() {
    int oldval = __atomic_fetch_add(&refcnt, 1, __ATOMIC_SEQ_CST);
    if (oldval == 0) {
        /* do you setup */
    }

    /* do whatever */

    int oldval = __atomic_fetch_sub(&refcnt, 1, __ATOMIC_SEQ_CST);
    if (oldval == 1) {
        /* do you teardown */
    }
}
```
The atomic operations guarantee that only one `run_main_graph` (the first one entered) will ever see `refcnt` goes from 0 to 1, and only one `run_main_graph` (the last one exited) will ever see `refcnt` goes from 1 to 0.

tungld:
@caoimhinuibrian @gongsu832 thank you for your explaination! Now I understand how it works in a parallel setting.

Back to a sequential setting, if we call run_main_graph multiple times one by one, e.g. in Python REPL, each call will load and destroy the constants, which introduces a big overhead if the constants is large.

gongsu832:
> @caoimhinuibrian @gongsu832 thank you for your explaination! Now I understand how it works in a parallel setting.
> 
> Back to a sequential setting, if we call run_main_graph multiple times one by one, e.g. in Python REPL, each call will load and destroy the constants, which introduces a big overhead if the constants is large.

Then in the python function, do the same:
```
# Python
def python_func:
    __load_constants()
    run_main_graph()
    ...
    run_main_graph()
    __destroy_constants()

/* C/C++ */
run_main_grap() {
    __load_constants()
    /* do whatever */
    __destroy_constants()
}

__load_constants() {
    int oldval = __atomic_fetch_add(&refcnt, 1, __ATOMIC_SEQ_CST);
    if (oldval == 0) {
        /* do you setup */
    }
}

__destroy_constants() {
    int oldval = __atomic_fetch_sub(&refcnt, 1, __ATOMIC_SEQ_CST);
    if (oldval == 1) {
        /* do you teardown */
    }
}
```
Basically, if `run_main_graph` is called by another function which already called `__load_constants`/`__destory_constants` the calls inside `run_main_graph` become noops.

tungld:
So, we still need explicit `load_constants` and `destroy_constants` to be used with Python.

```
run_main_graph() {
    int oldval = __atomic_fetch_add(&refcnt, 1, __ATOMIC_SEQ_CST);
    if (oldval == 0) {
        /* do you setup */
    }

    /* do whatever */

    int oldval = __atomic_fetch_sub(&refcnt, 1, __ATOMIC_SEQ_CST);
    if (oldval == 1) {
        /* do you teardown */
    }
}
```

What happens if we have two parallel runs: one goes to `do you setup` and the other, while the first one is doing setup, goes to `go whatever` and tries to access the buffer the first one is setting up? Should we introduce a sync after all `do you setup`. I feel we need to handle all sequential/parallel issues here.

gongsu832:
Having `__load_constants`/`__destroy_constants` is fine. The key is not to expose them to the users. If the setup/teardown itself needs to be atomic, you can just use a blocking lock to update the reference counter:
```
__load_constants() {
    pthread_mutex_lock();
    if (refcnt == 0) {
        /* do your setup */
    }
    refcnt++;
    pthread_mutex_unlock();
}

__destroy_constants() {
    pthread_mutex_lock();
    if (refcnt == 1) {
        /* do your teardown */
    }
    refcnt--;
    pthread_mutex_unlock();
}
```

gongsu832:
> look at PR 581. The code runs without segfaaulting, so I suspect we don't need to copy the constants.

@caoimhinuibrian The build is not failing likely because none of the tests is multi-threaded. We have concurrent tests but each test I believe is single-threaded. To see the segfault you need to manually run the multi-threaded driver posted by Zhen Yu in https://github.com/onnx/onnx-mlir/issues/568#issuecomment-813008756

tungld:
Thanks @gongsu832 for your useful answer! I understand much now.

The final question with a C program calling your `run_main_graph` three times one by one:
```
run_main_graph();
run_main_graph();
run_main_graph();
```

If users do not explicitly call `load_constants` at the beginning and `destroy_constants` at the end, the above program will load and destroy the constants three times, doesn't it? If so, do you think it is fine, especially when the constants is large, or should we recommend users to call `load_constants` to make the `load_constants` inside the first `run_main_graph` noop? Or anyway to hide the load and destroy from users in this case. Many thanks!

AlexandreEichenberger:
I concur with @caoimhinuibrian and @tjingrant ; If we do need to have some code run before a first instance / after the last instance, then we could have a "prepare" and "cleanup" kind of calls that would do the work under the cover, without the user having to know what is needed. 

My understanding was that because we pushed a lot of this work to the compiler (e.g handle constant propagation of tensor), that prepare/cleanup was not necessary anymore. Ideally, constant should be reused from instance to instance, even in parallel instances, since they are constants!

@tungld it would be great if you could explain what lead us to this solution to begin with, if you don't mind. 

If I recall, we used to send all the constants through LLVM, and that was very very slow. So we took them out, but I though that we had a way to put them back into the final binary.  If that were the case, then it should be as simple as following a pointer to a constant area of memory.

If we cannot avoid some pre-processing, then we should have a thread-safe prepare/cleanup. Because then we can run inferences concurrently without interference. To run inference lock free, we can have something like a test and test&set.

gongsu832:
> Thanks @gongsu832 for your useful answer! I understand much now.
> 
> The final question with a C program calling your `run_main_graph` three times one by one:
> 
> ```
> run_main_graph();
> run_main_graph();
> run_main_graph();
> ```
> 
> If users do not explicitly call `load_constants` at the beginning and `destroy_constants` at the end, the above program will load and destroy the constants three times, doesn't it? If so, do you think it is fine, especially when the constants is large, or should we recommend users to call `load_constants` to make the `load_constants` inside the first `run_main_graph` noop? Or any way to hide the load and destroy from users in this case. Many thanks!

This usage scenario cannot be handled completely within `run_main_graph` since the constants persist across multiple `run_main_graph` calls. Still, I don't think the users should be burdened to worry about such low level details, unless of course this is the prevalent usage scenario.

I don't know what `load_constants` does exactly. If it's loading from files, such back-to-back calls to `run_main_graph` should benefit from system caching. And if we really really want to handle such usage scenario, we can internally implement a caching system ourselves.

seroyer:
> 
> > If users do not explicitly call `load_constants` at the beginning and `destroy_constants` at the end, the above program will load and destroy the constants three times, doesn't it? If so, do you think it is fine, especially when the constants is large, or should we recommend users to call `load_constants` to make the `load_constants` inside the first `run_main_graph` noop? Or any way to hide the load and destroy from users in this case. Many thanks!
> 
> This usage scenario cannot be handled completely within `run_main_graph` since the constants persist across multiple `run_main_graph` calls. Still, I don't think the users should be burdened to worry about such low level details, unless of course this is the prevalent usage scenario.
> 
> I don't know what `load_constants` does exactly. If it's loading from files, such back-to-back calls to `run_main_graph` should benefit from system caching. And if we really really want to handle such usage scenario, we can internally implement a caching system ourselves.

I suspect that both the parallel and the sequential usages scenarios will be very common.  Imagine you have an inference service you are providing to some downstream users.  You can get multiple overlapping calls during peak usage or you could get sequential calls (or gaps between calls) when usage is lower or when you have a single client requesting inference sequentially for many things.

In order to achieve the best performance, wouldn't we either need to provide a setup/tear-down API to avoid allocating/copying/dellocating the constants on each call, or use an approach similar to what @caoimhinuibrian and @AlexandreEichenberger suggested and just use the constants directly where they are in the library?  Sounds like maybe the latter doesn't work on Windows, though?  So that leaves the former.

It's not uncommon for a library to require users to call a setup before first use and tear-down after last use and I don't think that's too much of a burden.  It just needs to be clearly documented that it's required.  In that case no locking is necessary in the library since it's the caller's responsibility to ensure safe usage in multi-threaded scenarios for the setup/tear-down functions and concurrent calls to the run function would become safe.

AlexandreEichenberger:
>We may be able to get away with copying into a buffer for Mac & Linux. But on platforms like Windows, our fallback/default path (in which weights are stored as files on a harddrive) can't take place without allocating a buffer and then reading file content into it. So there may be value in designing our runtime interface for the most generic case possible.

Given @tjingrant comment, I think we need to go to a prepare/cleanup scenario. We can make it lock-free if the cleanup does not need to be synchronized, meaning the cleanup is done when shutting down the whole inference. If so, here is how to do it lockless on the main path.

```C+
static bool isReady=false;
if (! isReady) {
  mutex.aquire();
  if (! isReady) {
    prepare();
    flush all changes; // machine dependent
    isReady = true;
  }
  mutex.release();
}
```

This follow a test and test and set pattern. "flush all changes" might be easiest performed by another internal mutex (as typically mutex flush changes before releasing it).

If we need to synchronize with shutdown, then we need a counter, which requires an atomic increment/decrement. They don't perform great under contention on POWER/Z...

AlexandreEichenberger:
Maybe @seroyer can comment if shutdown needs to be synchronized... Can we expect a shutdown to be called only once the inference engine (e.g. WMLz) knows that there are no "inflight" inferences?

gongsu832:
If we think repeated calling of `run_main_graph` is common then why do we even bother with prepare/shutdown? Why don't we just load it once and let it be, until the process ends? Maybe I'm not understanding something.

seroyer:
> 
> 
> Maybe @seroyer can comment if shutdown needs to be synchronized... Can we expect a shutdown to be called only once the inference engine (e.g. WMLz) knows that there are no "inflight" inferences?

I think we could assert that the inference engine call shutdown exactly once after all run calls are complete.  That's not too much of a burden.

seroyer:
> 
> 
> If we think repeated calling of `run_main_graph` is common then why do we even bother with prepare/shutdown? Why don't we just load it once and let it be, until the process ends? Maybe I'm not understanding something.

There are two scenarios I can think of where cleaning up is good.
1. If you dynamically load the library and unload it in some sort of long running application, you want to ensure you don't have a leak
2. If you are trying to debug memory leaks, leaving all that memory around complicates the analysis.

I suspect that scenario one is probably uncommon.  So these are probably weak reasons...

gongsu832:
> > If we think repeated calling of `run_main_graph` is common then why do we even bother with prepare/shutdown? Why don't we just load it once and let it be, until the process ends? Maybe I'm not understanding something.
> 
> There are two scenarios I can think of where cleaning up is good.
> 
>     1. If you dynamically load the library and unload it in some sort of long running application, you want to ensure you don't have a leak
> 
>     2. If you are trying to debug memory leaks, leaving all that memory around complicates the analysis.
> 
> 
> I suspect that scenario one is probably uncommon. So these are probably weak reasons...

And there are ways to run code when you load/unload a library, like the gcc init constructor and fini destructor where you can stick prepare/shutdown there.

seroyer:
I thought of another option...

Have a setup that increments the refcount and loads the constants if needed (0->1 transition).
Have a cleanup that decrements the refcount and deletes the constants if needed (1->0 transition).
Have run call setup and cleanup automatically.
But also document and expose the setup/cleanup functions as optional so that if the user wants the best performance for repeated use, they can explicitly call setup once before all run calls and cleanup once after all run calls.  It would end up calling setup multiple times but the subsequent times would just touch the refcount and not have to actually load/delete the constants.
But it would still support the simple scenario of calling run once without having to worry about calling into setup/cleanup methods.

AlexandreEichenberger:
> Have a setup that increments the refcount and loads the constants if needed (0->1 transition).
Have a cleanup that decrements the refcount and deletes the constants if needed (1->0 transition).
Have run call setup and cleanup automatically.

You need a mutex, or you need a ref count per constant, as one inference can start in the middle of the cleanup of a previous one. 

From past experience, atomic increment/decrement are not great for arch that have not direct support for it. It internally rely on an LARX and STCX (POWER) CompareAndSwap (Z) loop where you need an exponential backoff on conflict. 

The approach I suggested, if we can live with a non-synchronized cleanup, requires no synchronization whatsoever EXCEPT when two threads are simultaneously trying to both initialize at the same time (presumably a very rare instance).

gongsu832:
> > Have a setup that increments the refcount and loads the constants if needed (0->1 transition).
> > Have a cleanup that decrements the refcount and deletes the constants if needed (1->0 transition).
> > Have run call setup and cleanup automatically.
> 
> You need a mutex, or you need a ref count per constant, as one inference can start in the middle of the cleanup of a previous one.
> 
> From past experience, atomic increment/decrement are not great for arch that have not direct support for it. It internally rely on an LARX and STCX (POWER) CompareAndSwap (Z) loop where you need an exponential backoff on conflict.
> 
> The approach I suggested, if we can live with a non-synchronized cleanup, requires no synchronization whatsoever EXCEPT when two threads are simultaneously trying to both initialize at the same time (presumably a very rare instance).

I'm not worried about locks. The overhead of acquiring/releasing a user level mutex like pthread_mutex_lock/unlock would be peanuts comparing to the rest of the computation, unless of course the rest of the computation is on the order of microseconds.

To me, the more important question is, do we have to burden the user with extra calls for this particular optimization. A burden is a burden no matter how small we devs think it might be. And I'm not convinced that we must burden the user. For example, as I mentioned above, one potential solution is to put the prepare/shutdown logic in the init/fini constructor/deconstructor (supported by both gcc and clang, and I believe Windows also has equivalent). So the constants get loaded/unloaded when the model .so is loaded/unloaded and we never have to worry about anything while run_main_graph runs.

caoimhinuibrian:
> > look at PR 581. The code runs without segfaaulting, so I suspect we don't need to copy the constants.
> 
> @caoimhinuibrian The build is not failing likely because none of the tests is multi-threaded. We have concurrent tests but each test I believe is single-threaded. To see the segfault you need to manually run the multi-threaded driver posted by Zhen Yu in [#568 (comment)](https://github.com/onnx/onnx-mlir/issues/568#issuecomment-813008756)

Gong, yes I know ... I did run Zhen Yu's example, and it doesn't segfault. The segfault is caused by the double free, and since PR 581 removes both the malloc and the free, there is no longer the possibility of that case occurring.

I ran also a simple test that shows we get the same results as before.

On Linux and OSX there is no reason not to use the constants directly. On Windows we are reading them from a file ... not sure why this is required, but I'll let the Microsoft people look at that, maybe @NathanielMcVicar can chime in?

The discussion earlier in this PR about the treatment of the constants at compile time is irrelevant. When we build the shared lib for Linux, we embed the constants in the binary, so they are loaded at runtime as part of loading the lib.

tjingrant:
> On Linux and OSX there is no reason not to use the constants directly. On Windows we are reading them from a file ... not sure why this is required, but I'll let the Microsoft people look at that, maybe @NathanielMcVicar can chime in?

It's because linking binary data into a shared library requires very platform-specific methods. And I did not have access to a Windows machine to experiment and/or implement required mechanisms, nor do I know what's the Windows-specific way to embed binary data into shared library.

Taking a step back, storing weights in files is the only de-facto portable way to deal with constant pools. And I think this path should probably be kept as a fallback regardless of whether anything smarter can be done for weight storage on Windows.

NathanielMcVicar:
> On Windows we are reading them from a file ... not sure why this is required, but I'll let the Microsoft people look at that

I believe this is the preferred method, but I don't have any experience with it: https://docs.microsoft.com/en-us/cpp/windows/working-with-resource-files?view=msvc-160

The argument that having files as a platform-agnostic fallback option is important seem compelling.

gongsu832:
> Gong, yes I know ... I did run Zhen Yu's example, and it doesn't segfault. The segfault is caused by the double free, and since PR #581 removes both the malloc and the free, there is no longer the possibility of that case occurring.
> 

OK that's good to know. Sorry I missed your comment at the top of PR #581 saying you tried the parallel inference.

messerb5467:
Hi everyone! I work with @seroyer as a tester and wanted to chime in here since I've seen numerous thoughts on API stability that are top of everyone's mind. I'd definitely prefer to see API stability at the forefront like we're talking about, but if it can't be helped, we should be ok until we go point release IIRC. Past that we have to start doing paperwork letting people know about the api changes and it starts getting onerous.

tungld:
Thank you everyone for a very fruitful discussion here! We finally come up with a solution that keeps the API unchanged and runs well on different scenarios. It's ready on the master branch now. Let me close this patch.

PR #578, opened 2021-04-05
Krnl operations support for broadcast

AlexandreEichenberger:
Added support for Variadic parameter passing in MatMul, CopyFromBuffer, CopyToBuffer, in order to facilitate the use of broadcast in matrix multiplication. Before this, memrefs were restricted to 2D data structures.

EDSC and MLIR seems to use predominantly `ValueRange` instead of `ArrayRef<Value>` and having the Krnl EDSC rely on ArrayRefs was making the integration more difficult. This PR changes them to facilitate integration. 

Finally, moved asserts that were checking the validity of parameter for MatMul from the builder (which was limited to one specific builder) to a verify() function for more broad use.


doru1004 (approved):
LGTM

PR #579, opened 2021-04-06
update to onnx 1.8.1

chentong319:
Update submodule onnx to 1.8.1.
The version of operations does not change yet.
Due to the interface change, ReduceSum, Unsqueezed, Squeez and split are removed from backend test temporarily. Also some softmax and logsoftmax are removed due to unmatched result  probably because the new softmax is defined with new ReduceSum.  
Upgrading operation to this version will be put in separate patches later.


caoimhinuibrian (approved):
LGTM

PR #581, opened 2021-04-07
[WIP] Test pthreads

caoimhinuibrian:
I eliminated the copy of the constants to the buffer; the allocation of the buffer; the free of the buffer. The parallel inference runs without segfaulting. I have no idea if it gets the correct answer, however.

### DIFF
--- a/src/Builder/FrontendDialectTransformer.cpp
+++ b/src/Builder/FrontendDialectTransformer.cpp
@@ -73,8 +73,8 @@ public:
     }
   }
 
-  mlir::ModuleOp ImportONNXModel(
-      const onnx::ModelProto &model, ImportOptions options) {
+  mlir::ModuleOp ImportONNXModel(const onnx::ModelProto &model,
+                                 ImportOptions options) {
     options_ = options;
     SetOpSetImport(model); // Determines which opsets to use.
     importGraph(model.graph());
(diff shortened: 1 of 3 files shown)


tungld:
It seems only the cleanup was eliminated. I still see the copy here: https://github.com/onnx/onnx-mlir/blob/c6f908de2fbe14459057eec2ad453bcd69ffff2e/src/Conversion/KrnlToLLVM/KrnlToLLVM.cpp#L461 Please correct me if I am wrong. Thanks!

caoimhinuibrian:
#elif __linux__
extern char _binary_param_bin_start;
extern char _binary_param_bin_end;

void *getEmbeddedConstPool(int64_t _) { 
  checkEndianness();
  //auto size = (unsigned int)(&_binary_param_bin_end - &_binary_param_bin_start);
  //void *buffer = malloc(size);
  //memcpy(buffer, &_binary_param_bin_start, size);
  return &_binary_param_bin_start;
}
this is what I tested .... at least I think so

tungld:
Could you please comment out lines in getEmbeddedConstPool for Apple and Windows also to see if it works too? Thanks!

caoimhinuibrian:
> Could you please comment out lines in getEmbeddedConstPool for Apple and Windows also to see if it works too? Thanks!

I don't have an environment for either of these ... looking at the code, it should work for Apple. For windows we read a file in, so this doesn't work. I think for Linux this is the way to go.

tjingrant:
>  I think for Linux this is the way to go.

I think so too. 

On platforms where we store weights in files, we however will have to manage buffers somehow.

PR #584, opened 2021-04-08
access constants directly, remove free - linux only

caoimhinuibrian:


caoimhinuibrian:
The parallel segfault was resolved by #586, this was an inferior solution, closing it.

PR #586, opened 2021-04-08
Clean up packed constants and use native constants

tungld:
Do not packed constants, they are directly encoded in the generated code.


tjingrant:
So is the endianness of raw data always native with respect to the machine where code is generated?

caoimhinuibrian:

tungld:
@caoimhinuibrian Jenkins was failed because I did not edit lit tests. Fixed it. Thanks!

caoimhinuibrian:

caoimhinuibrian (approved):
LGTM

tjingrant:
@tungld I think we need to understand the performance implication here, could you post some results comparing the compilation time of ResNet before/after this patch?

tungld:
@tjingrant compilation time for resnet50:

Before: 50 s
After: 60 s

Before we packed constants at earlier stage while this patch leaves the constant in IRs until lowering to LLVM. So, passes would benefit a bit from a smaller IRs. We can introduce a pass to use opaque attributes at earlier stage to make it faster. In that case it may be not good for debugging if we want to see values.

tjingrant:
@tungld that CircleCI vgg test is actually erroring. Do you have any idea what might have gone wrong? The CircleCI machine is reasonably large.

tjingrant:
> Before we packed constants at earlier stage while this patch leaves the constant in IRs until lowering to LLVM. So, passes would benefit a bit from a smaller IRs. We can introduce a pass to use opaque attributes at earlier stage to make it faster. In that case it may be not good for debugging if we want to see values.

That seems like something worth trying. Adding 10s on an already very slow compilation time is disconcerting.

The inclusion of a large blob of constant in the model only hinders debuggability. I believe previously krnl constants with size above a certain threshold are packed away, maybe something similar can be done here.

tungld:
@tjingrant I cannot reproduce the error on CircleCI. Do you think it is related to stack since we push constants to LLVM constants?

tjingrant:
@tungld this sounds plausible. Could we change all the alloca to malloc?

tungld:
It seems VGG19 failed at compilation time. .so file was not generated. During compilation, I think we do not call alloca (I know we allocate buffers in constprop, but we use malloc there). Is there any way to debug on CircleCI?

tungld:
@tjingrant Just observed VGG16 compilation again on my machine, and I saw `check-onnx-backend` needed up to 7 GB when compiling VGG. So it looks there was memory shortage on CircleCI.

gongsu832:
Free CircleCI plan appears to have a 4GB limit so that's a plausible cause of failure on CircleCI. But given that it was working before, do we know what caused the significant increase in memory demand? I'm guessing it might have something to do with the new way of handling constants? So it looks like the change had more ramifications than we might have expected and we probably want to look into a bit more carefully. We certainly don't want to run into the situation where we can't even compile a model due to the significantly increased memory demand.

tungld:
The high memory footprint was because of attributes which are created and cannot be freed. This behavior is a known issue in MLIR. The new way of handling constants consumes more memory at the time of lowering to LLVM since we create opaque attributes for LLVM IR. Yes, I agree to minimize memory demand as much as possible.

As you know, we saw ResNet152 consumed up to +7GB  due to a lot of dense attributes generated during constant propagation, and we fixed that by avoiding creating intermediate attributes, which resulted in less than 1GB. I believe if we enabled ResNet152 on CircleCI a few weeks ago, we should see CircleCI failed also.

gongsu832:
Sorry I'm a bit confused. I thought the CircleCI failed for VGG? As far as I can tell, CircleCI started failing after the this patch. Before this patch, CircleCI build always succeeded. For example, look at https://github.com/onnx/onnx-mlir/pull/591 which was merged right before this patch.

tungld:
We don't have ResNet152 in our backend tests. I mean ResNet152 consumed +7GB during compilation before this patch #530. At that time, if we had ResNet152 in our backend tests, I though CircleCI would failed also due to the high memory demand.

gongsu832:
OK so the current failure of CircleCI is not related to resnet152, correct? So I guess you are saying that this patch is known to consume more memory (which I think caused CircleCI to fail for VGG) and you agree we should try to minimize the memory demand as much as possible. If that's the case, then we are on the same page. Thanks.

tungld:
>OK so the current failure of CircleCI is not related to resnet152, correct? So I guess you are saying that this patch is known to consume more memory (which I think caused CircleCI to fail for VGG) and you agree we should try to minimize the memory demand as much as possible. If that's the case, then we are on the same page. Thanks.

Exactly, sorry for the confusion! Thanks!

PR #588, opened 2021-04-08
add JNI signature calls

gongsu832:
- add JNI signature calls
- reorganize signature call files to be more consistent with other
  API implementation, expose the calls to user header, and add them
  to libcruntime and libOMTensorUtils
- rename Java class DynEntryPoint to OMModel to have better naming
  consistency
- Add SPDX-License-Identifier: Apache-2.0 line to some files


gongsu832:
@caoimhinuibrian Is there any particular reason why we need to make a copy of the signatures? I made the signature const char[] and simply returned pointers. Let me know what you think.

AlexandreEichenberger:
Obviously, if they can be passed back without malloc, that simplifies the interface (no freeing necessary).

gongsu832:
@caoimhinuibrian @tjingrant Here are some additional thoughts. Currently, the signature globals live inside model.so and the API calls returning the signatures live in libcruntime. This is a strange situation because libcruntime can potentially be used without the model.so, i.e., the program is generating and loading model.so dynamically. In this case, if you link the program against libcruntime, you will get undefined symbol error because the signatures aren't available yet until the program runs and generates and loads model.so. This was what happened with the numerical tests using libOMTensorUtils (the internal version of libcruntime).

So both the signature globals and the API calls returning the signatures really should live in model.so, i.e., the API calls should be generated when the model.so is generated, just like the run_main_graph call.

tjingrant:
@gongsu832 there may be a reason to return a copy of the signature, which effectively guarantees our internal record of signature is immutable.

gongsu832:
@tjingrant I thought about that. But why do we care if the signatures are immutable or not? I mean after we declare them const, compiler will catch unintentional writes to it. Of course you can coerce the returned pointer to be a writeable one and store into it. But if you really want to shot yourself in the foot, there are lots of other easier ways to do it.

AlexandreEichenberger:
Maybe a commonality with the constants. If we have a way to put them also on Read-only pages, we get the same benefits for the constants and for the signatures... though the constants have a real impact (they are large) and this is probably very small and rarely used... Nevertheless, if the same mechanism can be used, we may as well.

tjingrant:
> @tjingrant I thought about that. But why do we care if the signatures are immutable or not? I mean after we declare them const, compiler will catch unintentional writes to it. Of course you can coerce the returned pointer to be a writeable one and store into it. But if you really want to shot yourself in the foot, there are lots of other easier ways to do it.

Makes sense to me.

tjingrant:
> So both the signature globals and the API calls returning the signatures really should live in model.so, i.e., the API calls should be generated when the model.so is generated, just like the run_main_graph call.

The other way to look at this issue is to separate runtime into two parts:
- One that helps users interact with objects required for model inference.
- The other is simply stuffs that we don't want to write LLVM IR for because we can instead write the high-level C++ code that does the same thing.

And make sure that user links with the first kind of runtime, and we embed the code for the second kind of runtime within our model. The second part is tricky, there're several ways to do it:
- Link these code statically into the shared library.
- Upon installation, run clang on these codes to generate LLVM IR, and include them in our model LLVM module whenever they're needed.

caoimhinuibrian (approved):
LGTM

gongsu832:
> Maybe a commonality with the constants. If we have a way to put them also on Read-only pages, we get the same benefits for the constants and for the signatures... though the constants have a real impact (they are large) and this is probably very small and rarely used... Nevertheless, if the same mechanism can be used, we may as well.

I think this is already the case with Linux. When the constants are embedded in the .so, it gets its own segment and likely be marked as read-only. @tjingrant should know better. Windows is another story of course.

gongsu832:
> > So both the signature globals and the API calls returning the signatures really should live in model.so, i.e., the API calls should be generated when the model.so is generated, just like the run_main_graph call.
> 
> One possible solution is to separate runtime into two parts:
> 
>     * One that helps users interact with objects required for model inference.
> 
>     * The other is simply stuffs that we don't want to write LLVM IR for because we can instead write the high-level C++ code that does the same thing.
> 
> 
> And make sure that user links with the first kind of runtime, and we embed the code for the second kind of runtime within our model. The second part is tricky, there're several ways to do it:
> 
>     * Link these code statically into the shared library.
> 
>     * Upon installation, run clang on these codes to generate LLVM IR, and include them in our model LLVM module whenever they're needed.

I think this could be a general solution if we have a lot of functions to deal with. But I think splitting the external cruntime and internal OMTensorUtils already made things a bit tricky and this will make it even trickier. So I was going to suggest that perhaps we just bite the bullet and special case the two signature calls and generate them in LLVM since they are very simple. Funny thing is, while I'm typing this Kevin actually already did it! So I guess that settles it, at least for now. :-D

PR #589, opened 2021-04-08
Various Krnl Dialect Lowering Fixes

tjingrant:


AlexandreEichenberger (approved):
LGTM, glad you found the cause of the problems.

Issue #559, opened 2021-03-26
Avoid find_package(PythonInterp...)


stephenneuendorffer:
This mechanism is not robust in the presence of python2 and python3 installs and is deprecated in recent versions of cmake.  In our environment it tends to find python2 which seems to fail in various ways.  Unfortunately, this usage seems to be endemic in many of the sub-projects used in onnx.
The preference is to use find_package(Python3 ...) if python3 is needed.  (see https://cmake.org/cmake/help/git-stage/module/FindPython3.html)

I started down the path of fixing this, but because of the number of sub-projects, this isn't a simple patch, so in the short term I've resorted to setting -DPYTHON_EXECUTABLE to something sane.

sstamenova:
@stephenneuendorffer : I believe #590 takes care of this. Can you confirm it resolved the issues in your environment?

stephenneuendorffer:
Yes, thanks!

Issue #247, opened 2020-08-04
CMake paths are sometimes incorrect on Windows


NathanielMcVicar:
Some paths in CMake on Windows assume binaries can be found at LLVM_PROJ_BUILD/bin whereas on Windows they are typically at LLVM_PROJ_BUILD/Release/bin. The LLVM_PROJ_BIN variable is correctly set across platforms, but is not always used (for example for finding llc in src/CMakeLists.txt:51). On a related note, tools that are built with extensions (i.e. llc.exe on Windows) aren't accounted for.

NathanielMcVicar:
I wanted to call attention to some additional Windows issues here (aggregating to avoid spamming the issue tracker). Consider this list a work in progress, and please let me know if I'm simply making a mistake in our Windows builds! Thanks.

- There is an include of dlfcn.h in ExecusionSession.hpp (should this be Execution?). This is causing failures on some of our Windows build systems (all target, not built in onnx-mlir target). In particular the Windows version is 1) not POSIX compliant and 2) not included in the standard Windows 10 SDK that is suggested in the install instructions. I think this class may need some #ifdefs or other special handling for Windows.
- There is an interaction between ONNX version, Python version and pybind11 version on Windows. I haven't spent a lot of time characterizing it, but I have only been able to get things fully building using ONNX 1.7. This may very well be a problem on my end, but 1.7 does resolve it.

doru1004:
@NathanielMcVicar would you be open to contributing the fixes for some of these issues? I think seeing the fixes as code would be helpful.

NathanielMcVicar:
@doru1004 We do have a number of fixes locally we are planning to upstream soon. There didn't seem to be much demand for the library execution stuff on Windows, but if this is helpful I can look into adding them as well.

doru1004:
Thanks a lot for the update! That sounds great! Please feel free to PR any fixes you have locally, others may find them useful as well.

sstamenova:
#616 Fixed a few more of the issues on Windows. Of this list, the only remaining one is dlfcn.h (the need for equivalent on Windows).

sstamenova:
#722 addresses the last issue left here. That's not to say all is well on Windows after, but all the bugs listed here will have been fixed.

NathanielMcVicar:
Thanks so much Stella for resolving the rest of these! Future Windows problems can be new issues.

Issue #582, opened 2021-04-07
Release build of replace2x2Matmul2d creates blocks with improper alignment


NathanielMcVicar:
After #569 Release builds on Linux asserts in LLVM due to a null TypeRange as follows (block at src/Conversion/ONNXToKrnl/Math/MatMul.cpp:153).

```
onnx-mlir-opt: /mnt/e/llvm-project-github/mlir/lib/IR/TypeRange.cpp:19: mlir::TypeRange::TypeRange(ArrayRef<mlir::Type>): Assertion `llvm::all_of(types, [](Type t) { return t; }) && "attempting to construct a TypeRange with null types"' failed.
PLEASE submit a bug report to https://bugs.llvm.org/ and include the crash backtrace.
Stack dump:
0.      Program arguments: /mnt/e/onnx-mlir-github/build_linux/bin/onnx-mlir-opt --shape-inference --convert-onnx-to-krnl --canonicalize /mnt/e/onnx-mlir-github/test/mlir/onnx/onnx_lowering_with_canonicalize.mlir -split-input-file
 #0 0x000000000126a0e3 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) (/mnt/e/onnx-mlir-github/build_linux/bin/onnx-mlir-opt+0x126a0e3)
 #1 0x0000000001267d4e llvm::sys::RunSignalHandlers() (/mnt/e/onnx-mlir-github/build_linux/bin/onnx-mlir-opt+0x1267d4e)
 #2 0x000000000126a5b6 SignalHandler(int) (/mnt/e/onnx-mlir-github/build_linux/bin/onnx-mlir-opt+0x126a5b6)
 #3 0x00007f1f421f28a0 __restore_rt (/lib/x86_64-linux-gnu/libpthread.so.0+0x128a0)
 #4 0x00007f1f40beef47 raise /build/glibc-2ORdQG/glibc-2.27/signal/../sysdeps/unix/sysv/linux/raise.c:51:0
 #5 0x00007f1f40bf08b1 abort /build/glibc-2ORdQG/glibc-2.27/stdlib/abort.c:81:0
 #6 0x00007f1f40be042a __assert_fail_base /build/glibc-2ORdQG/glibc-2.27/assert/assert.c:89:0
 #7 0x00007f1f40be04a2 (/lib/x86_64-linux-gnu/libc.so.6+0x304a2)
[...]

PR #590, opened 2021-04-08
Use find_package(Python3) and bump min cmake version

sstamenova:
LLVM uses cmake version 3.13.4 and has dropped support for python2. This change brings onnx-mlir to the same versions of cmake and python as LLVM since onnx-mlir is including cmake files directly from LLVM and using the lit infrastructure
Without this change, the version of python used across the project is not consistent causing random build failures on (mostly Windows) machines that have both python2 and python3

This addresses part of #247 and #559


tjingrant:

tjingrant:
@sstamenova based on your description of this patch, it looks like we should probably merge this patch despite Windows-CI failure? Given that the failures were not due to this patch, and enabling this patch will help us catch future Windows-specific problems.

sstamenova:
> @sstamenova based on your description of this patch, it looks like we should probably merge this patch despite Windows-CI failure? Given that the failures were not due to this patch, and enabling this patch will help us catch future Windows-specific problems.

I think that's accurate. It does mean that the Windows-CI will stay red until #582 is resolved (to fix two of the tests) and the dependency issue for the third test is figured out as well (which I am looking at right now).

tjingrant:

PR #592, opened 2021-04-09
Fix usage of -- /m (for parallelism) when building on Windows

sstamenova:
```-- /m``` should be added where possible in order to improve the parallelism of the builds on Windows - by default cmake uses a single thread on Windows, this tells cmake to use the ```/m``` option in msbuild

Also fixes a couple of instances where the scripts are currently invoking cmake with ```--target -- /m``` (so without actually specifying a target) which results in cmake errors like: ```CMake Error: Invalid value used with --target …```


caoimhinuibrian:

tjingrant:

tjingrant (approved):
LGTM!

PR #593, opened 2021-04-09
Support Scan Op

tjingrant:
A low-effort implementation of basic scan op functionality. Code is largely copied and modified from Loop op. Significant refactoring and more extensive testing are desirable in the future to reduce code duplication and support more complicated use cases.


test/backend/test.py
```diff
@@ -91,527 +91,528 @@
...
-    # RNN
-    "test_rnn_seq_length_cpu": (test_static_dynamic,{0:{0,1,2}}),
-    "test_simple_rnn_defaults_cpu": (test_static_dynamic,{0:{0,1,2}}),
-    "test_simple_rnn_with_initial_bias_cpu": (test_static_dynamic,{0:{0,1,2}}),
+    #
+    # ############################################################
+    # # Elementary ops, ordered alphabetically.
+    #
+    # # Abs
+    # "test_abs_cpu": (test_static_dynamic,),
```
AlexandreEichenberger, line 99: Why are all the tests disabled?

AlexandreEichenberger (commented):
@tjingrant code looks good, why are all the lit tests disabled?

AlexandreEichenberger (approved):
LGTM

PR #597, opened 2021-04-09
Add missing dependency of check-onnx-lit on onnx-mlir

sstamenova:
All of the CI builds (and in our normal workflows) generally build onnx-mlir first and then proceed to build the tests, so onnx-mlir usually exists which is why this hadn't been caught before. The Windows CI build was not running the tests correctly because of python issues (see #590 ), so it thought they always passed marking the Windows CI build as succeeded because it is the last step to run. However, that was masking the fact that the build of onnx-mlir is failing!

This command:

```call cmake --build . --config Release --target onnx-mlir -- /m```

would run and fail. Then this command:

```call cmake --build . --config Release --target check-onnx-lit```

would run but not execute the tests and mark the build as succeeded. Since check-onnx-lit doesn't have a dependency on onnx-mlir, it does not build onnx-mlir and the earlier failure of the onnx-mlir build gets lost.

Long story short, check-onnx-lit has a dependency on onnx-mlir which this change adds. After the addition of the dependency, the Windows CI build is now correctly reporting that the Windows build of onnx-mlir in master is broken.


sstamenova:
@tjingrant @NathanielMcVicar 

So this is great - the python fix revealed that the onnx-mlir executable is not being built for the tests, which revealed a missing dependency of the tests and that in turned revealed that the onnx-mlir build on Windows has been failing for some amount of time but the combination of the python issue and the missing dependency have successfully hidden both the failing tests and the failing build!

caoimhinuibrian:

caoimhinuibrian (approved):
LGTM

PR #598, opened 2021-04-09
Fix the windows build of onnx-mlir

sstamenova:
The onnx-mlir (specifically, onnx-mlir.exe) build is failing on Windows with:

```
D:\a\1\llvm-project\llvm\include\llvm/IR/PassInstrumentation.h(277,16): error C2248: 'llvm::PassInstrumentationCallbacks::BeforeNonSkippedPassCallbacks': cannot access private member declared in class 'llvm::PassInstrumentationCallbacks' [D:\a\1\onnx-mlir\build\src\MainUtils.vcxproj]
D:\a\1\llvm-project\llvm\include\llvm/IR/PassInstrumentation.h(143): message : see declaration of 'llvm::PassInstrumentationCallbacks::BeforeNonSkippedPassCallbacks' [D:\a\1\onnx-mlir\build\src\MainUtils.vcxproj]
D:\a\1\llvm-project\llvm\include\llvm/IR/PassInstrumentation.h(65): message : see declaration of 'llvm::PassInstrumentationCallbacks' [D:\a\1\onnx-mlir\build\src\MainUtils.vcxproj]
```

This change cleans up the includes for onnx-mlir and MainUtils, so that we get the correct headers (headers from mlir and llvm were conflicting in the build).


caoimhinuibrian:
> This change cleans up the includes for onnx-mlir and MainUtils, so that we get the correct headers (headers from mlir and llvm were conflicting in the build).

Why is there a difference between Windows build and Linux build with respect to the order of include files? Is this a compiler problem?

sstamenova:
> > This change cleans up the includes for onnx-mlir and MainUtils, so that we get the correct headers (headers from mlir and llvm were conflicting in the build).
> 
> Why is there a difference between Windows build and Linux build with respect to the order of include files? Is this a compiler problem?

The VS compiler has subtle differences with clang/gcc and sometimes that results in build failures on Windows that don't reproduce on other platforms. Sometimes it catches bugs that other compilers don't too. In this case it was not to our advantage though.

caoimhinuibrian (approved):
LGTM

Issue #600, opened 2021-04-10
commit 694cdb033214e6baa314ae396fbeff1ce6a77ca0 breaks mnist model on s390x


gongsu832:
I have a mnist model and driver that worked fine before this patch but now is getting the wrong results. This looks like an endian problem. I remember Tian asked about the endian issue for this patch but there was no response.

Both native library and JNI jar are giving the wrong results (expected since JNI simply calls the native library).

Commands to compile the native library and run the inference:
```
# onnx-mlir --EmitLib model.onnx
# gcc -g -Iinclude_path mnist.c -o mnist model.so -Llib_path -lcruntime
# LD_LIBRARY_PATH=. ./mnist
```
`include_path` and `lib_path` are where runtime headers and libraries are, respectively. You should see the following output:
```
input  signature: [    { "type" : "float" , "dims" : [1 , 1 , 28 , 28]  }

]
output signature:  [   { "type" : "float" , "dims" : [1 , 10]  }

]
prediction[0] = -118029507121385983488744457502720.000000
prediction[1] = -71.833557
prediction[2] = 1230729779023607448758206332928.000000
prediction[3] = 112626513890916027531402412032.000000
prediction[4] = -22.549494
prediction[5] = 9.097518
prediction[6] = 1172105306786765666460368896.000000
prediction[7] = -13.723551
prediction[8] = -2.159079
prediction[9] = -3669.105713
The digit is 2
```
The correct results should be:
```
input  signature: [    { "type" : "float" , "dims" : [1 , 1 , 28 , 28]  }

]
output signature:  [   { "type" : "float" , "dims" : [1 , 10]  }

]
prediction[0] = 84.358864
prediction[1] = -71.825768
prediction[2] = -2.669611
prediction[3] = -23.258825
prediction[4] = -22.675903
prediction[5] = 9.237737
prediction[6] = 18.885262
prediction[7] = -13.772935
prediction[8] = -2.074757
prediction[9] = 0.823274
The digit is 0
```

Commands to compile the JNI jar and run the inference:
```
# onnx-mlir --EmitJNI model.onnx
# javac -cp model.jar MnistTestDriver.java
# ONNX_MLIR_JNI_LOG_LEVEL=debug java -cp .:model.jar MnistTestDriver
```
You should see the following output:
```
[    { "type" : "float" , "dims" : [1 , 1 , 28 , 28]  }

]
 [   { "type" : "float" , "dims" : [1 , 10]  }

]
[...]

PR #601, opened 2021-04-12
Use -O2 in bitcode opt

tungld:
Using -O3 for opt causes unexpected behavior on BE machines. One issue was reported here: #600.

This patch changes to use -O2 for safety.
 

caoimhinuibrian (approved):
Approving this, but we should figure out why O3 breaks stuff and move back to O3 when a fix becomes available

PR #602, opened 2021-04-12
JNI log improvement and Jenkins fix

gongsu832:
- use snprintf and additional logic to make sure we don't overrun
  output buffer
- fix Jenkins script where an old build can abort a new build,
  due to slow running of builds (likely caused by slow git clone)


caoimhinuibrian (approved):
LGTM

PR #603, opened 2021-04-12
Directly use a global constant for the returned MemRef

tungld:
This patch is complementary to #586, which
- cleans up getEmbeddedConstPool since it is not used anymore.
- directly uses a global constant's pointer as MemRef's pointers, so that it avoids creating a buffer and copying the constant value to the buffer. In case of aligned MemRef, we still need to create an aligned buffer and do copy.


tjingrant:
@tungld could you include the a short example IR snippet for each case to help with review:

- directly uses a global constant's pointer as MemRef's pointers, so that it avoids creating a buffer and copying the constant value to the buffer. 
- In case of aligned MemRef, we still need to create an aligned buffer and do copy.

It seems rather tricky to reason about the implications of code changes like this by staring at the lowering logic.

tungld:
@tjingrant I reorganized lit tests so that they are easy to follow. Please look at constant.mlir.

src/Conversion/KrnlToLLVM/KrnlToLLVM.cpp
```diff
@@ -375,80 +375,86 @@ class KrnlGlobalOpLowering : public ConvertToLLVMPattern {
...
-          // LLVM::GlobalOp does not support OpaqueElementsAttr.
-          // Both StringAttr and OpaqueElementsAttr use StringRef for internal
-          // data array. Thus, it looks safe to use StringAtrr instead of
-          // OpaqueElementsAttr.
-          StringRef data = krnlGlobalOp.value()
-                               .getValue()
-                               .cast<OpaqueElementsAttr>()
-                               .getValue();
+    if (!krnlGlobalOp.value().hasValue())
+      llvm_unreachable("Krnl Global must always have a value");
```
tjingrant, line 389: Looks like an assertion to me.

src/Conversion/KrnlToLLVM/KrnlToLLVM.cpp
```diff
@@ -375,80 +375,86 @@ class KrnlGlobalOpLowering : public ConvertToLLVMPattern {
...
-      // Allocate the memory where the constants will be used from.
-      // This is a region of local memory and needs to be emitted as an alloca.
-      int alignment = 0;
-      if (krnlGlobalOp.alignmentAttr())
-        alignment = krnlGlobalOp.alignmentAttr().getValue().getSExtValue();
-
       auto one = rewriter.create<LLVM::ConstantOp>(
           loc, llvmI64Ty, rewriter.getI64IntegerAttr(1));
-      alloc = rewriter.create<LLVM::AllocaOp>(loc,
+      Value alloc = rewriter.create<LLVM::AllocaOp>(loc,
```
tjingrant, line 457: This comment is not specific to this patch, but it seems like allocating parameters for weights on the stack is a risky idea, given the possibility of stack overflow. Plus weight tensors are pretty big.

tjingrant (approved):
LGTM! TODO note for myself: an unlikely corner case implied by this patch is that since our memory management policy currently mandates that users are responsible for freeing all output tensors, if a constant tensor is for some unlikely reason returned, it must not be freed. I need to change to krnl to llvm lowering mechanism such that if constant tensors are returned, then the OMTensor struct does not own the underlying memory buffer.

PR #609, opened 2021-04-13
Remove googletest and benchmark

sstamenova:
googletest is currently not used directly by onnx-mlir anywhere. There's a single function defined in test/unit/CMakeLists.txt that is not utilized.

Rather than adding googletest as a submodule and then adding custom defines, if any tests based on googletest are to be included in the future, they should leverage the infrastructure (add_unittest) exposed by LLVM through AddLLVM.cmake. The onnx-mlir build strictly depends on a build of LLVM, so it can be guaranteed to have access to add_unittest.

This change also removes benchmark which is similarly unused. onnx itself has benchmark and googletest for its tests, so if they ever need to run, onnx will manage those dependencies.


sstamenova:

caoimhinuibrian:

tjingrant (approved):
I think this is also a dependency for google benchmark, although we're not using benchmark either; and even if we do, we should probably let google benchmark resolve dependency itself. So it's probably okay to drop googletest.

caoimhinuibrian:

PR #610, opened 2021-04-13
Support scan - release branch

caoimhinuibrian:


AlexandreEichenberger (approved):
LGTM, if you want to merge it, go ahead. You need to be also aware that there is a lot of other code, including GEMM, and a few other things. So it is really a snapshot of what we have on the master branch.

PR #613, opened 2021-04-14
Simplify configuration and leverage cmake exports from LLVM and MLIR.

stephenneuendorffer:
This sequence of patches significantly refactors the toplevel cmake infrastructure, generally with the goal of avoiding duplication with information and rules that are already exported by MLIR.


caoimhinuibrian:

CMakeLists.txt
```diff
@@ -60,9 +60,9 @@ elseif ((ONNX_USE_PROTOBUF_SHARED_LIBS AND Protobuf_USE_STATIC_LIBS)
     "ONNX_USE_PROTOBUF_SHARED_LIBS and Protobuf_USE_STATIC_LIBS must be opposites of each other.")
 endif()
 add_subdirectory(third_party/onnx)
-
-add_subdirectory(third_party/googletest)
-set(BENCHMARK_ENABLE_GTEST_TESTS OFF)
+# Use googletest from llvm instead.
```
sstamenova, line 63: This is already gone in master, benchmark too

sstamenova:
I started doing the same thing and have a local patch working. I noticed that there were a number of things that need to be done and the first patch I created is #616, so that doesn't actually conflict with the first one you chose.

MLIR.cmake
```diff
@@ -32,6 +16,25 @@ else()
...
+set(LLVM_DIR ${LLVM_CMAKE_DIR})
+set(MLIR_DIR ${LLVM_CMAKE_DIR}/../mlir)
+
+find_package(MLIR REQUIRED CONFIG)
+
+message(STATUS "Using MLIRConfig.cmake in: ${MLIR_DIR}")
+message(STATUS "Using LLVMConfig.cmake in: ${LLVM_DIR}")
+
+list(APPEND CMAKE_MODULE_PATH "${LLVM_CMAKE_DIR}")
+include(AddLLVM)
```
sstamenova, line 33: Should you AddMLIR as well? HandleLLVMOptions too?

MLIR.cmake
```diff
@@ -52,300 +55,59 @@ message(STATUS "LLVM_PROJECT_LIB        : " ${LLVM_PROJECT_LIB})
...
-  set(LLVM_SRC_INCLUDE_PATH ${LLVM_PROJ_SRC}/include)
-else()
-  set(LLVM_SRC_INCLUDE_PATH ${LLVM_PROJ_SRC}/llvm/include)
-endif()
-set(LLVM_BIN_INCLUDE_PATH ${LLVM_PROJ_BUILD}/include)
-set(MLIR_SRC_INCLUDE_PATH ${LLVM_PROJ_SRC}/mlir/include)
-set(MLIR_BIN_INCLUDE_PATH ${LLVM_PROJ_BUILD}/tools/mlir/include)
-set(MLIR_TOOLS_DIR ${LLVM_PROJ_BIN})
+# FIXME: Remove after migration
+set(MLIR_TOOLS_DIR ${LLVM_TOOLS_BINARY_DIR})
```
sstamenova, line 63: I don't think this is needed

MLIR.cmake
```diff
@@ -52,300 +55,59 @@ message(STATUS "LLVM_PROJECT_LIB        : " ${LLVM_PROJECT_LIB})
...
-if(MSVC)
-  if (CMAKE_BUILD_TYPE)
-    set(ONNX_MLIR_TOOLS_DIR ${CMAKE_BINARY_DIR}/bin/${CMAKE_BUILD_TYPE})
-  else()
-    set(ONNX_MLIR_TOOLS_DIR ${CMAKE_BINARY_DIR}/bin/Release)
-  endif()
-else()
-  set(ONNX_MLIR_TOOLS_DIR ${CMAKE_BINARY_DIR}/bin)
-endif()
+set(ONNX_MLIR_TOOLS_DIR ${CMAKE_BINARY_DIR}/${CMAKE_CFG_INTDIR}/bin)
```
sstamenova, line 66: Same, and also LIB_DIR

MLIR.cmake
```diff
@@ -52,300 +55,59 @@ message(STATUS "LLVM_PROJECT_LIB        : " ${LLVM_PROJECT_LIB})
...
-        ${LLVM_SRC_INCLUDE_PATH};${LLVM_BIN_INCLUDE_PATH};${MLIR_SRC_INCLUDE_PATH};${MLIR_BIN_INCLUDE_PATH}
-)
-include_directories(${MLIR_INCLUDE_PATHS})
+include_directories(${LLVM_INCLUDE_DIRS};${MLIR_INCLUDE_DIRS})
 
-if (NOT USE_INSTALLED_LLVM)
-  # Force CMAKE_INSTALL_PREFIX and BUILD_SHARED_LIBS to be the same as LLVM build
+# If we're pointing at an LLVM build, then assume we want to install into the same
+# install location
+if(EXISTS ${LLVM_PROJ_BUILD}/CMakeCache.txt)
```
sstamenova, line 78: If CMAKE_INSTALL_PREFIX was explicitly specified, I don't think it should be overwritten.

MLIR.cmake
```diff
@@ -52,300 +55,59 @@ message(STATUS "LLVM_PROJECT_LIB        : " ${LLVM_PROJECT_LIB})
...
-  message(STATUS "BUILD_SHARED_LIBS       : " ${BUILD_SHARED_LIBS})
 endif()
 
+# Force BUILD_SHARED_LIBS to be the same as LLVM build
+set(shared ${LLVM_ENABLE_SHARED_LIBS})
+set(BUILD_SHARED_LIBS ${shared} CACHE BOOL "" FORCE)
+message(STATUS "BUILD_SHARED_LIBS       : " ${BUILD_SHARED_LIBS})
+
 # Threading libraries required due to parallel pass execution.
+# FIXME Should be added to LLVMConfig.cmake
```
sstamenova, line 92: I am not sure this is needed - each of the libs coming from LLVM/MLIR carry with them their dependencies, so there's no need to explicitly find these packages. Am I missing something?

MLIR.cmake
```diff
@@ -52,300 +55,59 @@ message(STATUS "LLVM_PROJECT_LIB        : " ${LLVM_PROJECT_LIB})
...
-  set(LLVM_CMAKE_DIR
-        "${LLVM_PROJ_BUILD}/lib/cmake/llvm"
-        CACHE PATH "Path to LLVM cmake modules")
-endif()
-list(APPEND CMAKE_MODULE_PATH "${LLVM_CMAKE_DIR}")
-include(AddLLVM)
-include(TableGen)
+# FIXME: Remove after migration
+#set(MLIRLibs ${MLIR_ALL_LIBS})
+get_property(MLIRLibs GLOBAL PROPERTY MLIR_ALL_LIBS)
```
sstamenova, line 105: Not all of these libs are needed, you can do:

  get_property(dialect_libs GLOBAL PROPERTY MLIR_DIALECT_LIBS)
  get_property(conversion_libs GLOBAL PROPERTY MLIR_CONVERSION_LIBS)

to limit the list

src/Tool/BinaryDecoder/CMakeLists.txt
```diff
@@ -2,8 +2,8 @@
 
 add_executable(binary-decoder BinaryDecoder.cpp)
 target_link_libraries(binary-decoder
```
sstamenova, line 4: target_link_libraries(binary-decoder
  PRIVATE
  LLVMSupport
  onnx
  )

is enough

stephenneuendorffer:
@sstamenova Do you want to take this over, or would you prefer that I update this?

MLIR.cmake
gongsu832: We currently have a dependency that llvm and onnx-mlir bins either both stay in their respective build directory, or both get installed to the same system wide directory (typically /usr/local/bin). See the comments above `getToolPath` in `src/MainUtils.cpp`.

sstamenova: > We currently have a dependency that llvm and onnx-mlir bins either both stay in their respective build directory, or both get installed to the same system wide directory (typically /usr/local/bin). See the comments above `getToolPath` in `src/MainUtils.cpp`.

With a complete switch to using MLIR as a package, I don't think there's a need for this requirement because the build of onnx-mlir can be pointed to an installation of LLVM and the paths (like kOptPath) can point to the correct location for the tools. Having said that, if you want to enforce that they be installed together regardless of whether they have to be, then it can stay that way.

gongsu832: I'm just pointing out that the current logic of `getToolPath` depends on the assumption made above. If you remove that assumption, the logic of `getToolPath` needs to change accordingly.

sstamenova:
> @sstamenova Do you want to take this over, or would you prefer that I update this?

I can take care of it since I already have it working locally. I just want to split it into smaller logical units (which I guess is what you were doing too) that make sense to commit individually.

stephenneuendorffer:
> > @sstamenova Do you want to take this over, or would you prefer that I update this?
> 
> I can take care of it since I already have it working locally. I just want to split it into smaller logical units (which I guess is what you were doing too) that make sense to commit individually.

Fine with me.  I had a few more patches to start with, but most of them didn't actually result in much simplification until the whole series was done.

caoimhinuibrian:
@sstamenova are you planning to resolve the conflicts here?

sstamenova:
> @sstamenova are you planning to resolve the conflicts here?

@caoimhinuibrian : This patch is superseded by https://github.com/onnx/onnx-mlir/pull/631.

caoimhinuibrian:
@sstamenova so shall we close it?

sstamenova:
> @sstamenova so shall we close it?

That seems fine to me.

caoimhinuibrian:
@sstamenova @stephenneuendorffer OK, closing this.

PR #616, opened 2021-04-15
Generate files per configuration & change build output directories

sstamenova:
This is a step towards fully supporting multi-config generators such as VS and Xcode. Right now, the generated files as well as path to files in the project point to a single location which often does not exist unless CMAKE_BUILD_TYPE was specified on the command line which defeats the purpose of multi-config. It also adds build_type to the output directory for single-config generators and changes the order of build_type in the output directory for multi-config generators.

1) Files that are generated by the build now reside in the config directory when using a multi-config generator. For example, instead of: src/ExternalUtil.hpp, when using VS or Xcode, we have: src/Debug/ExternalUtil.hpp, etc.
2) The targets that use the generated files now have an explicit dependency and the generated files have correct paths to onnx-mlir and the build directories regardless of whether the generator supports multi-config or no
3) The lit configuration now correctly refers to the binaries in the onnx-mlir build tree per configuration
4) Executables now have the right extensions when they are referenced in tools and scripts
5) A couple of unnecessary variables are removed
6) Use generator expressions for the CMAKE_*_OUTPUT_DIRECTORY properties when using a multi-config generator. The CMAKE_*_OUTPUT_DIRECTORY set now fully controls where all of the outputs are located which means that if we wanted to align with LLVM for the binary layout, it is a simple 3-line change.
7) For single-config generators, the output directories are now: build_dir/build_type/{bin|lib} instead of build_dir/{bin|lib}
8) For multi-config generators, the output directories are not: build_dir/build_type/{bin|lib} instead of build_dir/bin/build_type and build_dir/lib/build_type

After this change, the path to LLVM_PROJ_BIN is still not correct always since it still refers to CMAKE_BUILD_TYPE which is not always known at configuration time. This can be fixed by using find_package(LLVM) or find_package(MLIR). Using find_package is optimal, but it is a bigger change, so I am splitting it from the file generation for simplicity. The find_package(MLIR) change will be sent for review after this change (and maybe a couple more simple ones) is in.


caoimhinuibrian:

caoimhinuibrian:

caoimhinuibrian:

caoimhinuibrian:

test/backend/CMakeLists.txt
```diff
@@ -1,15 +1,42 @@
...
+    ${CMAKE_CURRENT_BINARY_DIR}/test.py
+    COPYONLY
+    )
+
+  configure_file(
+    ${CMAKE_CURRENT_SOURCE_DIR}/test_config.py.in
+    ${CMAKE_CURRENT_BINARY_DIR}/test_config.py
+    @ONLY
+    )
+endif()
```
gongsu832, line 39: Can we have some comments for this block of code? e.g., when does `CMAKE_CFG_INTDIR` resolve to a subdirectory and when does it resolve to a "."? What's the purpose of the `string(REPLACE ...)`?

sstamenova: I originally pushed an update which adds some more details to explain the block of code, but after spending some more time thinking through the problem, it is possible to simplify the logic and use generator expressions more heavily, so the latest change does just that.

CMakeLists.txt
```diff
@@ -12,12 +12,26 @@ endif()
...
+if (NOT "${CMAKE_CFG_INTDIR}" STREQUAL ".")
+  set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib/$<CONFIG>)
+  set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib/$<CONFIG>)
+  set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin/$<CONFIG>)
+else()
+  set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
+  set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
+  set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
+endif()
+
```
gongsu832, line 27: Do we even need to check for "." here? Since "." is basically a noop in a path, we can just set everything to `.../$<CONFIG>` (and indicate such in the comments)?

src/Runtime/jni/CMakeLists.txt
```diff
@@ -11,9 +11,11 @@ if(Java_Development_FOUND AND JNI_FOUND)
   # Target for Java runtime jar
   add_jar(javaruntime
           src/com/ibm/onnxmlir/OMModel.java
-	  src/com/ibm/onnxmlir/OMTensorList.java
-          src/com/ibm/onnxmlir/OMTensor.java
-          OUTPUT_DIR ${CMAKE_LIBRARY_OUTPUT_DIRECTORY})
```
gongsu832, line 16: Why not just change this to `${ONNX_MLIR_LIBRARY_PATH}`?

CMakeLists.txt
sstamenova: Yes, we do. CMAKE_CFG_INT_DIR is . with single-config generators, but $<CONFIG> always expands to all configurations. So even with make or ninja, the output directories would contain Debug/Release/etc. in the path. My assumption is that we would rather not do that and maintain the current pattern of having build_dir/bin and build_dir/lib when using ninja or make rather than build_dir/bin/release and build_dir/lib/release.

src/Runtime/jni/CMakeLists.txt
sstamenova: Because ${ONNX_MLIR_LIBRARY_PATH} right now has the generator expression $<CONFIG>, this fails. Other add_* functions like add_library support generator expressions, but add_jar does not. It tries to make sure that the directory exists, so with multi-config generators, it tries to create a path containing $<CONFIG> before it has been resolved to each of the configurations and fails.

CMakeLists.txt
gongsu832: I actually don't see what's wrong with having Debug/Release/etc for make or ninja. At least we would then be consistent across the board. But maybe a better path is `${CMAKE_BINARY_DIR}/$<CONFIG>/{bin|lib}`?

src/Runtime/jni/CMakeLists.txt
gongsu832: Thanks for the clarification. So it would be nice to have these as the comments about the copy command below.

CMakeLists.txt
sstamenova: LLVM uses build_dir/{bin|lib} for ninja/make and build_dir/build_flavor/{bin|lib} for Xcode/Visual Studio. It is definitely more convenient as a user as well as for scripts and tools to have the build flavor as a common parent for bin and lib, so I was going to suggest that in a separate code review - since making the change is not required for the multi-config generators to work. I can make it as part of this pull request though if you prefer.

As for having the build flavor for make or ninja as well - that is generally not how projects that use them are set up, but it is certainly convenient if anyone is going to build more than one flavor. I was trying to maintain the setup as it was for those generators, but I can change it if you would prefer.

gongsu832: OK up to you whether you want to make the change for the multi-config generators in this PR or another one.

@tjingrant @caoimhinuibrian I vote for having Debug/Release/... etc for make/ninja as well. It would make it easier to compare different flavors if needed. And it also makes the cmake file logic simpler and consistent across the board. Cast your vote and we can go with the majority.

caoimhinuibrian: I'm fine with doing that

NathanielMcVicar: For what it's worth, the behavior of prepending the Config for multi-config generators, but not single-config, is baked into CMake, as [described here](https://cmake.org/cmake/help/latest/prop_tgt/RUNTIME_OUTPUT_DIRECTORY.html). That doesn't mean ONNX MLIR can't decide to override the default here, but it would be non-standard.

gongsu832: Seeing that we are actually having issues with Debug vs Release build (so we likely need to compare them), and with Kevin's vote, let's go with having build flavor for make/ninja as well, i.e., same for both single and multi config across the board. Thanks.

sstamenova: I made the change to add the build flavor to the single-config generators. I also switched where the build flavor is in the path to build_dir/build_type/{bin|lib}. I would have preferred to do that in a separate change, but that would require adding support for having bin/build_type and lib/build_type to MainUtils.cpp which seems unnecessary if the plan is to just switch the order later.

sstamenova:
@gongsu832 : I'd like to get this checked in today if at all possible, so that I can send the next batch towards find_package(mlir)

caoimhinuibrian (approved):
LGTM

PR #627, opened 2021-04-19
Better Simd

AlexandreEichenberger:
Prior version uses affine.vector_load and store, which results in horrible sequences of code on z
This version uses a method that converts memref of scalar element to memref of vector of scalar element. Then, regular affine load/store are used to load vector of elements.

The problem with that new approach is that the memrefs must be multiples of the vector length being simonized. To address this, 2 things were done. First was to condition the use of SIMD inside matmul only when the proper conditions hold. That way, we don't have erroneous SIMD code. Second, I changed the GEMM code to tile the B or C matrices when the sizes of the original B or C are either runtime or compile time (in their last dim) but not multiple of the vector length. Applying them matrix mult to the tiled data then works perfectly with SIMD, as the buffers are sized appropriately to satisfy the conditions.


AlexandreEichenberger:
When running with 
```
#define DEBUG_MALLOC 1

#define DEBUG_SIMD_OFF 1
```
all but the circleCI worked (window did not complete).

AlexandreEichenberger:
@doru1004 this patch only works when bundling is disabled on Mac/x86. This run had it disabled, and it passed both Jenkins.

doru1004:
> @doru1004 this patch only works when bundling is disabled on Mac/x86. This run had it disabled, and it passed both Jenkins.

@AlexandreEichenberger the bundling optimization happens at Krnl level before any architecture specific decision is taken.

AlexandreEichenberger:
That version has -O3 disabled for a test on Z, but otherwise has bundling and simd. To actually enable SIMD on z, one needs the following options 

s390x-ibm-linux and z14. 

The options can be passed to onnx-mlir as follows: -mtriple= and -mcpu=. For tests such as backend tests, these same options can also be passed to test.py. The environment variable TEST_MTRIPLE and TEST_MCPU can also be used. For example, 

```
TEST_MCPU=z14 TEST_MTRIPLE= s390x-ibm-linux make check-onnx-backend
```

will trigger the option in `test.py`, which in turn will pass it to `onnx-mlir`

AlexandreEichenberger:
On Z, to get simd on a z14 loz machine, you need to do this:
```
TEST_MTRIPLE=s390x-ibm-loz TEST_MCPU=z14 make check-onnx-backend
```

or to generate a single model:

```
onnx-mlir --EmitLib --mtriple=s390x-ibm-loz --mcpu=z14 inception_v1.onnx
```

For some undetermined reason at this time, both triple and cpu are the default options on the z14 machine that I test, but unless they are explicitly mentioned, the llvm `opt` optimizer does not generate SIMD. This may be an issue to discuss with the LLVM compiler team,

AlexandreEichenberger:
Working, ready for reviews

doru1004 (approved):
LGTM

AlexandreEichenberger:
Thanks @caoimhinuibrian and @doru1004, much appreciated. Next patches will be smaller

PR #628, opened 2021-04-20
Fix constant duplication during bundling

doru1004:
Fix constant duplication during bundling.

Prior to auto-canonicalization, move all constants to the top of the block to ensure the squishing of constants does not lead to instruction dominance issues.


src/Transform/BundleMemoryPools.cpp
```diff
@@ -473,6 +473,28 @@ class KrnlBundleDynamicMemoryPools : public OpRewritePattern<AllocOp> {
...
+class KrnlMoveConstantsUp : public OpRewritePattern<ConstantOp> {
+public:
+  using OpRewritePattern<ConstantOp>::OpRewritePattern;
+  LogicalResult matchAndRewrite(
+      ConstantOp constOp, PatternRewriter &rewriter) const override {
+    // Get parent block.
+    auto parentBlock = constOp.getOperation()->getBlock();
+
+    // Ensure it's the top block.
+    if (!llvm::dyn_cast_or_null<FuncOp>(parentBlock->getParentOp()))
```
chentong319, line 489: Will this condition hold for nested Loop or If?

doru1004: At the moment this is only enabled for the main block since I've only observed the error there.

chentong319 (approved):
LGTM!

PR #631, opened 2021-04-21
Use find_package(mlir) to find the dependencies from llvm-project

sstamenova:
mlir (and llvm) both support import as a cmake package via find_package. This changes the onnx-mlir cmake setup to use find_package on mlir and then reference llvm and mlir libs and variables as normal cmake variables.


caoimhinuibrian:

sstamenova:
@gongsu832 @caoimhinuibrian : The break in the build with shared libs is because HandleLLVMOptions.cmake correctly interprets the value of LLVM_ENABLE_ABI_BREAKING_CHECKS in AddLLVM.cmake and sets the appropriate defines. I can do one of two things to fix the build:
1) (Temporarily) remove the include of HandleLLVMOptions.cmake. This will bring the build back to parity in terms of handling the options with the existing build. However, HandleLLVMOptions.cmake really should be included, so that the options are correctly interpreted, so this is a short term solution. It is, however, a way to keep the initial find_package change as small as possible without sacrificing functionality.
2) Make the changes necessary to support including HandleLLVMOptions.cmake. This includes adding LLVMSupport as a dependent library to pretty much every lib or exe built in the project. This will touch a lot of files, but it is the right long term change.

gongsu832:
I prefer to having things fixed properly instead of quick and dirty patch. And we are not in a big hurry to have this patch so taking a bit of time to get it right is fine with me.

sstamenova:
> I prefer to having things fixed properly instead of quick and dirty patch. And we are not in a big hurry to have this patch so taking a bit of time to get it right is fine with me.

I ran into an interesting problem. HandleLLVMOptions adds a number of handy options for warnings, compilation and linkage. For example:

```
# Pass -Wl,-z,defs. This makes sure all symbols are defined. Otherwise a DSO
# build might work on ELF but fail on MachO/COFF.
if(NOT (CMAKE_SYSTEM_NAME MATCHES "Darwin|FreeBSD|OpenBSD|DragonFly|AIX|SunOS|OS390" OR
        WIN32 OR CYGWIN) AND
   NOT LLVM_USE_SANITIZER)
  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,-z,defs")
endif()
```

This is great, except that now all of the shared libs built by ONNX MLIR want to verify that symbols are defined. We can easily add the references to the correct LLVM and MLIR libraries, however, the circular reference between OMONNXOps and OMKrnlOps is a problem because they each want to resolve the symbols, but they can't.

I think at this point it is better to remove the include of HandleLLVMOptions for the moment and then add it later when the circular dependency is resolved. Since HandleLLVMOptions is not required for find_package, this will give the immediate benefit of the majority of the find_package change and all the additional niceties of HandleLLVMOptions can be added at a later time.

sstamenova:
@stephenneuendorffer : Here's the find_package change so far.

caoimhinuibrian:

sstamenova:
@gongsu832 : Do you have any remaining concerns with this patch?

s390xlinux:
Sorry being busy and haven't looked at it. Looks fine to me.

caoimhinuibrian (approved):
@s390xlinux (Gong) approved, so this is g2g
LGTM

PR #643, opened 2021-04-26
Better simd on release

AlexandreEichenberger:
Better SIMD on Z, still break on some x86 system, being investigated.

To actually enable SIMD on z, one needs the following options

s390x-ibm-linux and z14.

The options can be passed to onnx-mlir as follows: -mtriple= and -mcpu=. For tests such as backend tests, these same options can also be passed to test.py. The environment variable TEST_MTRIPLE and TEST_MCPU can also be used. For example,

```
TEST_MCPU=z14 TEST_MTRIPLE= s390x-ibm-linux make check-onnx-backend
```

will trigger the option in test.py, which in turn will pass it to onnx-mlir


AlexandreEichenberger:
It now works on z and x. 

On Z, to get simd on a z14 loz machine, you need to do this:

TEST_MTRIPLE=s390x-ibm-loz TEST_MCPU=z14 make check-onnx-backend
or to generate a single model:

onnx-mlir --EmitLib --mtriple=s390x-ibm-loz --mcpu=z14 inception_v1.onnx
For some undetermined reason at this time, both triple and cpu are the default options on the z14 machine that I test, but unless they are explicitly mentioned, the llvm opt optimizer does not generate SIMD. This may be an issue to discuss with the LLVM compiler team

caoimhinuibrian (approved):
LGTM

PR #647, opened 2021-04-30
Handle nested initialized tensor

chentong319:
Currently, there is one list for initialized tensors. An error occurred when a model contains same name of initialized tensor for different graphs in control blocks.
The fix is to utilize the utility (SymbolMapping) for tensors of operations. SymbolMapping handles the nested scope.
1. Extend the SymbolMapping into a template.
2. The original SymbolMapping is initialized with type `onnx::Value` with a new name `ValueSymbolMapping'
3. The InitializedTensorMapping is initialized with type 'onnx::TensorProto' as base and extra operations originally defined for InitializedTensorMapping is kept.


src/Builder/FrontendDialectHelper.cpp
```diff
@@ -103,6 +103,7 @@ static std::vector<T> CreateArrayAttribute(onnx::TensorProto initializer) {
   return std::vector<T>(&data[0], &data[0] + size);
 }
 
+#if 0
```
caoimhinuibrian, line 106: Why not delete this block, rather than ifdef'ing it out?

caoimhinuibrian (approved):
LGTM, except for the comment about the \#if

src/Builder/FrontendDialectHelper.cpp
chentong319: Forgot. Deleted now.

PR #653, opened 2021-05-05
Removing in IndexExpr any references to ONNX and KRNL, Removing in ONNX any references to KRNL

AlexandreEichenberger:
As the title say. The most "difficult" was that ONNX Shape Helper need to generate some code, which for us is KRNL ops. So I added two lambda functions that are passed as parameters to the ONNX shape helper to retrieve and generate KRNL ops, so that ONNX may be deployed (for shape inference) totally free of KRNL, and when we want to use it with KRNL, we just passed the appropriate functions. Similar functions are trivial to generate using affine dialect, if need be.

Lots of tedious small changes

### DIFF
--- a/src/Builder/FrontendDialectHelper.cpp
+++ b/src/Builder/FrontendDialectHelper.cpp
@@ -15,7 +15,6 @@
 #include <llvm/Support/SwapByteOrder.h>
 
 #include "src/Builder/FrontendDialectHelper.hpp"
-#include "src/Dialect/ONNX/ONNXOps.hpp"
 
 namespace onnx_mlir {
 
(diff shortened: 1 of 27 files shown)


AlexandreEichenberger:
It is my intention to eventually move IndexExpr away from ONNX dialect, and to put it into an IR subdirectory, but I wanted to move the files after the changes inside them are approved, to simplify the reviewing.

caoimhinuibrian (approved):
LGTM ... will the cmake changes come in a separate PR?

AlexandreEichenberger:
> will the cmake changes come in a separate PR?

Yes, I would like to take it one step at a time.

PR #654, opened 2021-05-06
Leak namespace

AlexandreEichenberger:
there was a
```
using namespace mlir
```
in a header file. removed it, had to add more mlir:: that I wish I had to, but it works now.

### DIFF
--- a/src/Builder/FrontendDialectHelper.cpp
+++ b/src/Builder/FrontendDialectHelper.cpp
@@ -15,7 +15,6 @@
 #include <llvm/Support/SwapByteOrder.h>
 
 #include "src/Builder/FrontendDialectHelper.hpp"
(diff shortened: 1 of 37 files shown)


caoimhinuibrian (commented):
couldn't we just put the "using mlir" into the cpp files that needed it, rather than changing to explict mlir:: everywhere?

AlexandreEichenberger:
> couldn't we just put the "using mlir" into the cpp files that needed it, rather than changing to explict mlir:: everywhere?

good point, I did this plus added the `using mlir` in the numerical tests that failed to compile.

caoimhinuibrian (approved):
LGTM

AlexandreEichenberger:
Thanks @caoimhinuibrian for the suggestion, it was a good one.

PR #656, opened 2021-05-06
Move some operations to onnx1.8.1

chentong319:
1. Changes for lots of operations are the new data type and documentation
2. Support the new type for LOOP and IF
3. Big change is for ReduceSum: axes is changed from attribute to input. This change involves shape inference, decomposer, lowering and test
4. only axes defined with Constant is handled now. This is the reason that backend test for ReduceSum is not enabled yet


NathanielMcVicar:
We have a fix for v13 unsqueeze, which I could add to this PR or create a new PR, but I'm curious what the general goal is here. The backend test models, for example, depend on specific opsets and this mix and match approach is going to make it hard to find models that can run, especially given that there is no checking and some semantics have changed significantly.

chentong319:
@NathanielMcVicar You can open another PR. The general rule is to maintain onnx-mlir for the latest onnx and use onnx_converter to handle the old version. For the change of attribute to input, we are thinking of add the change in the onnx importer (src/Build/

NathanielMcVicar:
> For the change of attribute to input, we are thinking of add the change in the onnx importer (src/Build)

Does that mean that for older models we will convert the attribute to input on import, or does it mean converting the opset 13 input to an attribute before processing using the existing code? Either way, being able to mix opsets seems like a valuable feature, and we can wait to push our updates until that's in place to avoid breaking the current tests.

src/Dialect/ONNX/ONNXShapeHelper.cpp
```diff
@@ -21,10 +21,6 @@
 
 using namespace mlir;
 
-//===----------------------------------------------------------------------===//
```
AlexandreEichenberger, line 24: Thanks, forgot to remove it :-)

src/Transform/ONNX/Decompose.td
```diff
@@ -24,159 +24,177 @@ include "src/Dialect/ONNX/ONNXOps.td"
...
-    (ONNXReduceLogSumOp
-      (ONNXExpOp
-         (ONNXSubOp $oprd, (ONNXReduceMaxOp:$max $oprd, $axes, (KeepdimsTrue)))),
-      $axes,
-      $keepdims),
-    (ONNXSqueezeOp $max, $axes)),
-  [],
-  (addBenefit 0)
->;
+def ReduceLogSumExpOpPattern1
```
AlexandreEichenberger, line 84: Did you add a lit test for these new patterns?

src/Conversion/ONNXToKrnl/Math/Reduction.cpp
```diff
@@ -150,16 +151,40 @@ struct ONNXReductionOpLowering : public ConversionPattern {
...
-      for (auto axisAttr : axisAttrs.getValue()) {
-        int64_t axis = axisAttr.cast<IntegerAttr>().getInt();
-        axis = axis >= 0 ? axis : (inRank + axis);
-        assert(axis >= -inRank && axis <= inRank - 1);
-        if (std::find(axes.begin(), axes.end(), axis) == axes.end())
-          axes.push_back(axis);
+    if (definedAxes.size()) {
+      for (auto axis : definedAxes) {
+        int64_t newaxis = axis >= 0 ? axis : (inRank + axis);
+        assert(newaxis >= -inRank && newaxis <= inRank - 1);
```
AlexandreEichenberger, line 185: When the inputs are bad, don't we generally emit an error instead of an assert? I though that was our policy. Since this come from the "user".

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -156,6 +156,56 @@ RankedTensorType getReductionOutputType(RankedTensorType operandTy,
...
+    DenseElementsAttr axesAttrs, uint64_t keepdims,
+    uint64_t noop_with_empty_axes) {
+  int64_t rank = operandTy.getRank();
+
+  SmallVector<int64_t, 4> axes;
+  if (axesAttrs) {
+    for (auto element : axesAttrs.getValues<IntegerAttr>()) {
+      int64_t axis = element.getInt();
+      axis = axis >= 0 ? axis : (rank + axis);
+      assert(axis >= -rank && axis <= rank - 1);
```
AlexandreEichenberger, line 171: If this come from the user, I would use an op error, for sure a message.

AlexandreEichenberger (commented):
A few minor changes, thanks for all the good work.

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -156,6 +156,56 @@ RankedTensorType getReductionOutputType(RankedTensorType operandTy,
   return RankedTensorType::get(dims, operandTy.getElementType());
 }
 
+// Reduction with axes is from ConstantOp.
+// Only ReduceSum call this function now.
+RankedTensorType getReductionOutputType(RankedTensorType operandTy,
```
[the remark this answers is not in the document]
chentong319: sure

src/Transform/ONNX/Decompose.cpp
```diff
@@ -33,15 +33,49 @@ using namespace mlir;
...
-  for (int i = 0; i < nElements; ++i) {
-    wrapper[i] = origAttrs.getValue()[i].cast<FloatAttr>().getValueAsDouble();
+
+  assert(origAttrs && "handle EXISTING ArrayAttr only");
+  if (origAttrs.getValue()[0].dyn_cast<FloatAttr>()) {
+    mlir::Type elementType = rewriter.getF32Type();
+    int nElements = origAttrs.getValue().size();
+    SmallVector<float, 4> wrapper(nElements, 0);
+    for (int i = 0; i < nElements; ++i) {
+      wrapper[i] = origAttrs.getValue()[i].cast<FloatAttr>().getValueAsDouble();
```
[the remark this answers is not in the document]
chentong319: The other way is `getValue().convertToFloat()`. I think FloatAttr is generic for float or double value.

src/Conversion/ONNXToKrnl/Math/Reduction.cpp
chentong319: This is the case for optional input. The axes input may be missing or empty. We have to handle it.

src/Transform/ONNX/Decompose.td
chentong319: Not new patterns. Just replaced the old ReduceSum with the new version. Backend test contains cases for this pattern.

src/Conversion/ONNXToKrnl/Math/Reduction.cpp
AlexandreEichenberger: First, newaxis>=0 && <= inRank-1  so the check must be updated. Or the current test should be tested on "axis" and not "newaxis"

Second, you have an assert here. What I am saying is that our policy is to use assert for compiler errors, and emitError for model errors. When checking values that come from the model, we should thus use emitError.

chentong319: fixed.

src/Dialect/ONNX/ONNXOps.cpp
AlexandreEichenberger: @chentong319 same issue here as above. after correction, axis >=0 && <= rank -1

chentong319: You are right. Fixed. 
Since this code is in a utility function, some extra work is needed. I tested the code with some examples.

PR #658, opened 2021-05-06
Removed dependences to KRNL in ONNX CMakeLists.txt

AlexandreEichenberger:
Removed dependence, and there was 2 spurious includes to KRNL files that were still in the ONNX subdirectories. They are now removed. I believe there are no more references to Krnl in Dialect/ONNX except in comments...


caoimhinuibrian (approved):
LGTM, but what about changing src/CMakelists.txt too? There is a kludge there to force dependency resolution.

AlexandreEichenberger:
> LGTM, but what about changing src/CMakelists.txt too? There is a kludge there to force dependency resolution.

@caoimhinuibrian happy to do it but I need a hint.

AlexandreEichenberger:
> I think Kevin is referring to this:

@gongsu832 : it was hiding in plain sight! Thanks, it now works without the repeated target.

sstamenova:
Was this supposed to get committed? I thought it had gone into the master branch, but it looks like it did not.

sstamenova:
> > Was this supposed to get committed? I thought it had gone into the master branch, but it looks like it did not
> 
> Not sure what happened. Indeed I should get it back to master. Let me run the tests again and merge if all goes well

Thanks!

PR #659, opened 2021-05-06
added use for the return value of passes

AlexandreEichenberger:

MLIR requires the return value of passes to be tested. Most are now, with some where its ok to succeed or fail.


chentong319 (approved):
LGTM!

PR #660, opened 2021-05-07
Fix aligned dynamic memory pool

doru1004:
Fix aligned dynamic memory pool. This patch moves the aligned dynamic memory pool to its correct location.


AlexandreEichenberger (approved):
LGTM, thanks for the fix

PR #661, opened 2021-05-07
replaced assert false by llvm-unreachable, squash no return warning

AlexandreEichenberger:

minor change, remove one warning by the compiler


utils/OpBuildTable.inc
```diff
@@ -0,0 +1,354 @@
+//********************************************************
```
chentong319, line 1: This file should be move to src/Build

chentong319 (commented):
Do not add util/OpBuildTable.inc. It should be in src/Build. In your case, this file is not changed.

chentong319 (approved):
LGTM!

AlexandreEichenberger:
> Do not add util/OpBuildTable.inc. It should be in src/Build. In your case, this file is not changed.

@chentong319 Thanks, removed the file in util. The one in OpBuildTable.inc is unchanged from master.

PR #662, opened 2021-05-07
Move aligned dynamic memory pool to the correct position.

doru1004:
Move aligned dynamic memory pool to the correct position.


AlexandreEichenberger (approved):
LGTM

PR #667, opened 2021-05-11
Bad input file in onnx-mlir[-opt] clearly indicated

AlexandreEichenberger:
Before this fix, onnx-mlir-opt just crashed deep into mlir if the input file did not exist, without any warnings
Onnx-mlir did give a warning but then experienced a segfault.

After this patch, missing files will be clearly indicated to the user and a clean exit is performed.


doru1004 (approved):
LGTM

Issue #619, opened 2021-04-16
Different version of Slice Op


chentong319:
@amanwithfruit I am debugging a model and found the error was in ImportNodeSlice. The size for DenseElementAttribute is assumed to be   `{1}`, while the attribute in the test case has two elements.  The bug can be fixed by using the size of the attribute instead of `{1}` for tensor type in DenseElementAttribute.
However, I noticed that the current ImportNodeSliceOp (by #266) is using an old version (v1): axes is an attribute. In the new versions (v10, v11 and v13) of slice, the default value for optional input axes is specified in the specification.  There is no attribute for axes.
For this special case,  ImportNodeSlice can check the name of attributes first. If "axes" attribute is not found,  the default value can be added according to the specification. The code will be able for handle old and new versions of Slice, though a little bit messy.

amanwithfruit:
Let me page this back into my memory and I'll get started on this. Can you put up a failing test case I can use to motivate the fix part of this? The other parts I will refer to the spec.

chentong319:
Thanks @amanwithfruit . I can not give you that model directly. I asked the owner to simplify the model so that the buggy part can be shared.

amanwithfruit:
I know that pain 😅 I'm still paging things back in so take your time.

chentong319:
By the way, the code handling axes for SliceOp may be used by other operations, such as ReduceSum, Squeeze and Unsqueezed, all of which changes their axes from attribute to input in the new version. 
As for the default value, I am thinking whether some support should be added into the table gen. At least, we may define that function in the Op definition using table gen so that the code is more declarative.

amanwithfruit:
> By the way, the code handling axes for SliceOp may be used by other operations, such as ReduceSum, Squeeze and Unsqueezed, all of which changes their axes from attribute to input in the new version.
Sure - I'll take a look and see if I can re-use anything that already exists

> As for the default value, I am thinking whether some support should be added into the table gen. At least, we may define that function in the Op definition using table gen so that the code is more declarative.
Since they're inputs, I'm not sure what we can do? Unless you're suggesting that we hoist them from ONNX/protobuf inputs to MLIR attrs? Or am I misunderstanding your suggestion?

chentong319:
I am thinking of adding a function to generate the default value in the Op definition, hook it with the optional input with default value, and the constructor invokes that function automatically when the corresponding optional input is missing.

amanwithfruit:
[...]

PR #668, opened 2021-05-11
Fix for Issue #619

amanwithfruit:
When converting Slice op attrs to constant inputs, we need to use the size of the attribute as the shape of the constant.

Signed-off-by: aman <alachapelle@apple.com>


chentong319 (commented):
Should we check the name of the attribute to be safe?

amanwithfruit:
I'm not sure I understand what you mean, exactly? Why would this change depend on the name of the attribute?

chentong319 (approved):
LGTM!

chentong319:
I updated your branch with the master and will merge after tests are passed.

PR #672, opened 2021-05-11
Release mode2

AlexandreEichenberger:
This PR uncovered one return function without a value, which is fixed.
Also, when running the Release build on Z, I located the error to issues in the ONNXOps.cpp.inc.

When a printf is added anywhere into the function below:
```
void ONNXAddOp::build(::mlir::OpBuilder &, ::mlir::OperationState &odsState, ::mlir::TypeRange resultTypes, ::mlir::ValueRange operands, ::llvm::ArrayRef<::mlir::NamedAttribute> attributes) {
  assert(operands.size() == 2u && "mismatched number of parameters");
  odsState.addOperands(operands);
  odsState.addAttributes(attributes);
  assert(resultTypes.size() == 1u && "mismatched number of return types");
  odsState.addTypes(resultTypes);
}
```

is added there (tried first and last line), then the test below:
```
// CHECK-LABEL: @test_add_constant_1(%arg0: tensor<3xf32>) -> tensor<3xf32>
func @test_add_constant_1(%arg0 : tensor<3xf32>) -> tensor<3xf32> {
  %0 = "onnx.Constant"() {value = dense<[0.0, 1.0, 2.0]> : tensor<3xf32>} : () -> tensor<3xf32>
  %1 = "onnx.Add"(%0, %arg0) : (tensor<3xf32> , tensor<3xf32>) -> tensor<3xf32>
  "std.return"(%1) : (tensor<3xf32>) -> ()
  // CHECK-NEXT: [[CONST:%.+]] = "onnx.Constant"() {value = dense<[0.000000e+00, 1.000000e+00, 2.000000e+00]> : tensor<3xf32>} : () -> tensor<3xf32>
  // CHECK-NEXT: [[ADD:%.+]] =  "onnx.Add"(%arg0, [[CONST]]) : (tensor<3xf32>, tensor<3xf32>) -> tensor<3xf32>
}
```

works when called as below.

```
eichen@bitperf:~/Onnxcode/atests/release$ onnx-mlir-opt --shape-inference --constprop-onnx test-constprop1.mlir
```

using the release version. Also, all the lit tests passes. 

The printf is a symptom that some thing might be wrong with the compiler, and indeed, when adding the -O1 option, it works.

If the Release is needed, then we would need to add the `-O1` option for Z for Release build until the bug is located and fixed.


AlexandreEichenberger:
@gongsu832 could you try this PR with the compile option uncommented and see if it works in your environment? Tx

AlexandreEichenberger:
Fixes here resulted in a clean build in the #677 related patch (same solution, forced Release) on all platforms. So it looks like the combination of @tungld and my fixes are getting us to a happy place (for now)

AlexandreEichenberger:
@tungld Big thanks again for the fix on the library issue, it would have taken me a while to suspect the int32 bs 64.
@gongsu832 Big thanks also for helping us along and providing context and explanations.

Much appreciated.

AlexandreEichenberger:
There is an error in the latest s390x build #677 , investigating

doru1004 (approved):
LGTM

Issue #669, opened 2021-05-11
Parallelize ctest by default


NathanielMcVicar:
Would it make sense to parallelize the ctest target by default ("make test" or "cmake --build . --target test") using a method similar to this https://cmake.org/cmake/help/latest/module/ProcessorCount.html#module:ProcessorCount? On some platforms these tests can take a long time to run serially, so it might be nice to let each test run on a different core (as opposed to making multiple cores available to each test as in the above link).

Before I submit this change, I wanted to confirm that the current serially behavior isn't preferred. Thanks!

AlexandreEichenberger:
@gongsu832 Please respond to this one, not sure what is best, since we share some of these machines with others for the CIs

gongsu832:
All tests including ctest are already parallelized in Dockerfiles. Specifically, ctest is parallelized with `make ARGS=-j${NPROC} test`. NPROC is set by `.buildbot/jenkins-build-*.py` using an algorithm that takes into account both CPU numbers and memory size.
```
# Set parallel jobs based on both CPU count and memory size.
# Because using CPU count alone can result in out of memory
# and get Jenkins killed. For example, we may have 64 CPUs
# (128 threads) and only 32GB memory. So spawning off 128
# cc/c++ processes is going to quickly exhaust the memory.
#
# Algorithm: NPROC = min(2, # of CPUs) if memory < 8GB, otherwise
#            NPROC = min(memory / 4, # of CPUs)
MEMORY_IN_GB                = (os.sysconf('SC_PAGE_SIZE') *
                               os.sysconf('SC_PHYS_PAGES') / (1024.**3))
NPROC                       = str(math.ceil(min(max(2, MEMORY_IN_GB/4), os.cpu_count())))
```

NathanielMcVicar:
That makes sense, although -j will still override any defaults, so changing it won't result in a different behavior in CI that sets it manually. Regardless, if you prefer the current approach feel free to close this issue.

sstamenova:
@gongsu832 : Does it make sense to add a default in the cmake files for people who don't use the dockerfiles? I would personally expect the tests to run in parallel *unless* I did something to make them not as opposed to the other way around.

gongsu832:
@NathanielMcVicar Not sure I follow what you are saying. Can you elaborate?

@sstamenova `make test` not running in parallel by default really is a `make` behavior. People who don't use the dockerfiles obviously need to run the command themselves and might be expecting the default behavior.

[...]

PR #673, opened 2021-05-12
Run backend tests with python3 exe

NathanielMcVicar:
The backend tests currently check for pytest using pip but execute using pytest. On some platforms this can create a mismatch where pytest-xdist appears to be available but isn't found when called.


caoimhinuibrian:

NathanielMcVicar:
@gongsu832 This was the cleanest solution I could come up with to handle the docker and not-docker scenarios. It is a behavioral change in that this won't capture changes to $NPROC made after generator time if you call make directly. I don't believe that will be an issue with the current docker scripts, but please let me know if we need a different solution.

test/backend/CMakeLists.txt
```diff
@@ -33,9 +33,16 @@ execute_process(
...
 )
 if (${PYTEST_XDIST_FOUND} EQUAL 0)
-  message(STATUS "Parallel backend tests  : ON")
-  set(BACKEND_TEST_COMMAND "pytest")
-  set(BACKEND_TEST_ARGS "--forked" "-n" "$\{NPROC\}")
+  if (NOT DEFINED NPROC)
+    if (DEFINED ENV{NPROC})
+      set(NPROC $ENV{NPROC} CACHE STRING "Number of processors for pytest" FORCE)
+    else()
+      set(NPROC "auto" CACHE STRING "Number of processors for pytest" FORCE)
```
gongsu832, line 40: I assume "auto" is basically the number of CPUs? I typically avoid that due to the same reason mentioned in https://github.com/onnx/onnx-mlir/issues/669#issuecomment-839293610. So I would just use a conservative constant like 4.

test/backend/CMakeLists.txt
```diff
@@ -33,9 +33,16 @@ execute_process(
...
+    endif()
+  endif()
+  message(STATUS "Parallel backend tests  : ON (Using ${NPROC} thread(s))")
+  set(BACKEND_TEST_COMMAND "${Python3_EXECUTABLE}" "-m" "pytest")
+  set(BACKEND_TEST_ARGS "--forked" "-n" "${NPROC}")
```
gongsu832, line 45: If we just use a constant then I'd think `$\{NPROC:-4\}` would be sufficient? Note the backslashes make NPROC **not** to be interpreted by cmake but rather by shell at runtime.

NathanielMcVicar: The problem is that doesn't evaluate to an environment variable across platforms (i.e. Windows) [the {} format for non-lists isn't even handled uniformly across unix shells]. I agree that losing the evaluation at runtime is a disadvantage but I don't see a workaround better than handling it in cmake at generator time.

gongsu832: So out of all the archs we build on, Windows is the only one that can't handle `${NPROC:-4}`? Then I'm wondering if it's better to just special case Windows. It's a bit of an eye sore having to write multiple ifs just to set an env var. Actually, do we even run pytest on Windows?

gongsu832: Also on 2nd thought about `auto` vs a constant (e.g., 4), I think it's OK to use auto. Since inside a Jenkins CI docker environment, NPROC will also be set. So `auto` will only apply when you run pytest manually.

NathanielMcVicar: I think you're right. This file is going to require significant overhaul to work on Windows, and worrying about this particular case is not valuable. If the ${var} format works across all other platforms we care about, that seems reasonable. I'll restore the auto per your comment as well.

gongsu832:

gongsu832:

gongsu832 (approved):
LGTM

PR #674, opened 2021-05-12
Various small fixes

sstamenova:
Some typos; some white space; removing duplicate lib in OMLibs; removing a couple of unused include directories; removing unused definition INCLUDE_ONNX_ML; moving a dependency to the correct cmake; removing add_c_unit_test function which is only used in one place; making debug.py work (better) on windows


gongsu832:

src/Builder/FrontendDialectHelper.hpp
```diff
@@ -35,12 +35,9 @@
 #include "llvm/Support/raw_ostream.h"
 
 #include "onnx/onnx_pb.h"
+#include "src/Builder/SymbolTable.hpp"
 #include "src/Dialect/ONNX/ONNXOps.hpp"
 #include "src/Dialect/ONNX/ONNXOpsHelper.hpp"
-#if INCLUDE_ONNX_ML == 1
```
AlexandreEichenberger, line 40: Can you explain if the #if include is still present elsewhere, or not. I believe we can have some of the ML operations and thus I assume that this would be needed.

@chentong319 since you did this work, please comment as you see fit.

AlexandreEichenberger (commented):
Can you just confirmed that the removed dependences where truly superfluous?

src/Builder/FrontendDialectHelper.hpp
sstamenova: INCLUDE_ONNX_ML was in two places - the cmake and the header file. The cmake had:

```
if (INCLUDE_ONNX_ML)
  add_definitions(-DINCLUDE_ONNX__ML=1)
endif()
```

Notice that the -D statement has multiple underscores instead of just one, so it doesn't actually define the same define that is used in the header.

The header then did:
```
#if INCLUDE_ONNX_ML == 1
#include "src/Dialect/MLONNX/MLONNXOps.hpp"
#endif
```

The header that is being included does not exist.

These two together make me think that this has been unused for some time.

sstamenova:
> Can you just confirmed that the removed dependences where truly superfluous?

Yes, I believe everything that has been removed is superfluous. The builds and tests work correctly on Windows and Linux, so I don't expect any issues.

src/Builder/FrontendDialectHelper.hpp
caoimhinuibrian: @sstamenova we are planning to support at least some of the ONNX_ML ops, so it would be better to fix the inconsistencies, rather than deleting them

sstamenova: @caoimhinuibrian : What does "fix" mean in this case? The referenced header doesn't exist in onnx-mlir, so even if I fixed the cmake file to use the same name for both defines, this would still be dead code. I think it is generally better to remove dead code like this and then when the support for ONNX_ML ops is ready, it can be done correctly. Otherwise, we are risking the dead code staying dead. If you have an idea of how you would like the support to look like, I'm happy to update it instead, but until there's something concrete to include, I don't think it's the best course.

AlexandreEichenberger: @chentong319 you did the OnnxML work, can you tell us what your intention was? @sstamenova indicates that the header file does not exist. Was it not included, or is it included under other circumstances? A bit of background would be appreciated, tx.

caoimhinuibrian (approved):
LGTM

src/Builder/FrontendDialectHelper.hpp
chentong319: Initially, we want to separate ops for onnx and onnx_ml and selectively support them. Now it seems that there is no need to support only onnx not onnx_ml, the related code can be deleted.

PR #676, opened 2021-05-12
Make populateVectorXXX reusable

tungld:
KrnlToLLVM pass will be reused by other projects via `populateAffineAndKrnlToLLVMConversion`.
This patch puts `populateVectorToVectorCanonicalizationPatterns`, `populateVectorSlicesLoweringPatterns`, and `populateVectorContractLoweringPatterns` into `populateAffineAndKrnlToLLVMConversion`, so that the other projects can reuse these populateVectorXXX.


caoimhinuibrian (commented):
@tungld So, do I understand correctly? Previously we applied the Vector patterns greedily first, and then we applied the other patterns. Now, with this patch, we will apply all of the patterns at once. Does this have any implications on the final code we get?

tungld:
Kevin, yes, you are right. We have applied all of the pattern about krnl, affine, and std at once. Now with vector also. I think applyFullConversion is able to handle the order automatically. @AlexandreEichenberger you are using vector for Gemm and Matmul, any thoughts?

caoimhinuibrian (approved):
LGTM

PR #679, opened 2021-05-12
Remove variant

sstamenova:
This submodule is currently unused, so it can be removed. The header and lib are referenced (as in the header is included and the lib is linked against), however, they are never actually used in code.


caoimhinuibrian:

caoimhinuibrian (approved):
LGTM

PR #682, opened 2021-05-13
add_onnx_mlir_library and add_onnx_mlir_executable

sstamenova:
Add two cmake functions to simplify the addition of libraries and executables in onnx mlir. This is based on the add* functions in llvm/mlir. The goal is to capture commonality between the setup for the various libraries/executables in onnx mlir and simplify the cmake management. There are many things that can be added to the functions but I started with a couple of simple ones that strictly improve the code without introducing too many changes:
1) The add* functions add the default include directories rather than each library having to add them. The function can be changed not to do that or to add them as private includes, but at the moment it does it for each library (since each library already did) and it adds them as public - this way any executables or tests in the project don't have to. If onnx mlir is shipped as a package in the future and the headers are installed, this can be modified to have a different set of include directories for the package interface and the change will only need to happen in the one place.
2) All libraries (unless they opt out) are added to a single global property ONNX_MLIR_LIBS. Once the libraries use target_link_libraries rather than add_dependency to manage their dependency, this global property can replace the static list OMLibs. This cannot happen yet because the libraries are added to ONNX_MLIR_LIBS in the order in which they are declared which is not the order of dependencies needed by the linker (on Linux, Windows is happy).
3) All libraries and executables are installed (unless they opt out).

Lastly, this change removes the header files from the add_library calls. They are not needed since tey are included in the cpp files and they are generally ignored by add_lirary


sstamenova:
Now would be a good time to also decide if the tools (such as onnx-mlir-opt) should be installed. Both onnx-mlir-opt and binary-decoder are currently not installed.

caoimhinuibrian:

caoimhinuibrian:
> Now would be a good time to also decide if the tools (such as onnx-mlir-opt) should be installed. Both onnx-mlir-opt and binary-decoder are currently not installed. We've found cases where we would have liked onnx-mlir-opt to be installed, for example.

Something to consider is the audience for these things. onnx-mlir does an end-to-end compilation of the onnx file and so is relevant to an end-user - the others are more for people modifying or testing the code. But I don't have a strong opinion either way.

sstamenova:
> > Now would be a good time to also decide if the tools (such as onnx-mlir-opt) should be installed. Both onnx-mlir-opt and binary-decoder are currently not installed. We've found cases where we would have liked onnx-mlir-opt to be installed, for example.
> 
> Something to consider is the audience for these things. onnx-mlir does an end-to-end compilation of the onnx file and so is relevant to an end-user - the others are more for people modifying or testing the code. But I don't have a strong opinion either way.

I think both approaches make sense and I did not change the logic in this PR. We have found the llvm approach of shipping most everything to the install directories very useful, but it does mean that a lot of tools are installed that are rarely needed.

caoimhinuibrian (approved):
Cool .... looks good

caoimhinuibrian:
@sstamenova need to resolve conflicts

caoimhinuibrian:
> > > Now would be a good time to also decide if the tools (such as onnx-mlir-opt) should be installed. Both onnx-mlir-opt and binary-decoder are currently not installed. We've found cases where we would have liked onnx-mlir-opt to be installed, for example.
> > 
> > 
> > Something to consider is the audience for these things. onnx-mlir does an end-to-end compilation of the onnx file and so is relevant to an end-user - the others are more for people modifying or testing the code. But I don't have a strong opinion either way.
> 
> I think both approaches make sense and I did not change the logic in this PR. We have found the llvm approach of shipping most everything to the install directories very useful, but it does mean that a lot of tools are installed that are rarely needed.

OK ... following llvm practice seems right

sstamenova:
> > > > Now would be a good time to also decide if the tools (such as onnx-mlir-opt) should be installed. Both onnx-mlir-opt and binary-decoder are currently not installed. We've found cases where we would have liked onnx-mlir-opt to be installed, for example.
> > > 
> > > 
> > > Something to consider is the audience for these things. onnx-mlir does an end-to-end compilation of the onnx file and so is relevant to an end-user - the others are more for people modifying or testing the code. But I don't have a strong opinion either way.
> > 
> > 
> > I think both approaches make sense and I did not change the logic in this PR. We have found the llvm approach of shipping most everything to the install directories very useful, but it does mean that a lot of tools are installed that are rarely needed.
> 
> OK ... following llvm practice seems right

@caoimhinuibrian: I did not change the logic to install the two tools, should I do it in order to match llvm or leave it as it is?

caoimhinuibrian:
> > > > > Now would be a good time to also decide if the tools (such as onnx-mlir-opt) should be installed. Both onnx-mlir-opt and binary-decoder are currently not installed. We've found cases where we would have liked onnx-mlir-opt to be installed, for example.
> > > > 
> > > > 
> > > > Something to consider is the audience for these things. onnx-mlir does an end-to-end compilation of the onnx file and so is relevant to an end-user - the others are more for people modifying or testing the code. But I don't have a strong opinion either way.
> > > 
> > > 
> > > I think both approaches make sense and I did not change the logic in this PR. We have found the llvm approach of shipping most everything to the install directories very useful, but it does mean that a lot of tools are installed that are rarely needed.
> > 
> > 
> > OK ... following llvm practice seems right
> 
> @caoimhinuibrian: I did not change the logic to install the two tools, should I do it in order to match llvm or leave it as it is?

@sstamenova Sure, go ahead and add it here

caoimhinuibrian:

PR #691, opened 2021-05-18
Squeeze and Unsqueeze op for onnx1.8.1

chentong319:
Update Squeeze and Unsqueeze to op version 13. Changes are for their shape inference, rewrite rules, lowering and check-onnx-lit.


NathanielMcVicar:
Is the plan to move forward with this approach? If so, I'm happy to do the required merge to help. We have a conflicting solution for opset 13 squeeze/unsqueeze locally, and if I'm going to merge anyway, might as well benefit everyone :)

chentong319:
@NathanielMcVicar  This patch currently failed model test for  densenet121.onnx. The plan is to update onnx-mlir to version 1.9 with PR#717 and then we can use the version converter to handle the old model.
Another pending issue is whether we should keep the code for old version, discussed in #708.

chentong319:
This is an experimental PR.

PR #692, opened 2021-05-19
Reimplement RNN ops to utilize SIMD

tungld:
This patch is to reimplement LSTM, GRU, and RNN using MatMul, Split, Squeeze, Transpose, Add, Mul, which helps boost the inference performance.

- Weights and biases are computed during compile time using constant propagation. 
- ConstPropHelper is created to allow calling constant propagation from lowering patterns. Do not use tryToFold mechanism since tryToFold communicates via Attributes which will cause a large memory footprint again.
- Matrix multiplications are done via ONNXMatMulOp. Experimental fusion of multiple MatMuls, which uses the lowering pattern from GEMM, is added for LSTM but not used since it is currently slower than non-fusion.
- Other element-wise computations are fused into a single nested loop as the previous approach.
- Regarding constants, use threshold 512 to trigger the use of DenseElementsAttr for constants. This value is obtained by experimenting with the value ranges in the numerical tests for LSTM. Threshold less than 512 would trigger the loop unroll in -O1/2/3, causing the wrong result we saw before.


tungld:

tungld:

tungld:
@AlexandreEichenberger this patch is ready to be reviewed. Could you please review it? Thanks!

src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
```diff
@@ -369,3 +369,170 @@ Value getDimOrConstant(ConversionPatternRewriter &rewriter, Location loc,
...
+    return constVal;
+  } else {
+    return rewriter
+        .create<ONNXUnsqueezeOp>(
+            loc, resultType, input, rewriter.getI64ArrayAttr(axis))
+        .getResult();
+  }
+}
+
+// Only support evenly splitting.
```
AlexandreEichenberger, line 462: Could you just copy over the header file comment here too.

maybe the function name could be "splitConstantOrEmitSplitONNXOp"

Same comments for other operations below...

src/Conversion/ONNXToKrnl/ConvertONNXToKrnl.cpp
```diff
@@ -88,6 +97,16 @@ void FrontendToKrnlLoweringPass::runOnOperation() {
   // TODO: add any other ops which are considered legal.
   // Some operations can be marked as being still legal.
   // Example: target.addLegalOp<mlir::OpName>();
+  if (testRNNOps) {
```
AlexandreEichenberger, line 100: Can you add a comment on the approach. Let me say what I think it is. You enable ONNX operations to be generated, as you transform ONNX ops into a sequence of other ONNX ops that will then be later lowered and/or combined when operating on constants

test/numerical/TestGRU.cpp
```diff
@@ -26,6 +26,22 @@ using namespace mlir;
 // Sigmoid
 float sigmoid(float x) { return 1 / (1 + exp(-x)); }
 
+// Build an ONNXConstantOp from an OMTensor.
```
AlexandreEichenberger, line 29: Would it make sense to have a Helper.cpp, to be shared among the numerical tests? Maybe it is too much trouble, or just a header file with the inline implementations...

AlexandreEichenberger (approved):
LGTM, mainly asked for a few clarifications, where hopefully you can add the comments directly in the code for future references.

AlexandreEichenberger:
@tungld Note that the PR that moved the transforms for canonicalization in the dialect dir. Hopefully not too disruptive to your PR. Note also that apparently MLIR is going away from EDSC... we will have to have a strategy for that.

src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
```diff
@@ -369,3 +369,170 @@ Value getDimOrConstant(ConversionPatternRewriter &rewriter, Location loc,
   }
   return dimVal;
 }
+
+/// A helper fucntion to check whether a value is produced by a dense
```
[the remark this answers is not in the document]
tungld: Move it to the ONNX helper files. Also move another one to the Krnl helper file.

src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
tungld: Done.

src/Conversion/ONNXToKrnl/ConvertONNXToKrnl.cpp
tungld: Done.

test/numerical/TestGRU.cpp
tungld: Done.

tungld:
@AlexandreEichenberger Thanks for your comments! I addressed them in the above commits. 

> @tungld Note that the PR that moved the transforms for canonicalization in the dialect dir. Hopefully not too disruptive to your PR. Note also that apparently MLIR is going away from EDSC... we will have to have a strategy for that.

Most of them can be rewritten back to `rewrite.create` easily. However, `krnl_iterate` may take more times to be replaced I think.

src/Conversion/ONNXToKrnl/ConvertONNXToKrnl.cpp
```diff
@@ -52,7 +52,27 @@ class ONNXEntryPointLowering : public OpRewritePattern<ONNXEntryPointOp> {
...
+  // RNN ops are lowered to other ONNX ops such as ONNXMatMulOp, ONNXSplitOp,
+  // ONNXTransposeOp, etc. These ONNX ops are then lowered into krnl ops in this
+  // pass.
+  //
+  // To write LIT tests for RNN ops, we need not to check the final generated
+  // krnl code that is lengthy but the intermediate generated code including
+  // ONNX ops. We trust the lowering of the other ONNX ops.
+  //
+  // This flag is used in LIT tests to stop the lowering of the other ONNX ops.
+  // Usage: onnx-mlir-opt --convert-onnx-to-krnl='check-rnn-ops-lowering'
```
AlexandreEichenberger, line 72: Thanks, that is exactly the type of super useful comments that will help us in the long run. Much appreciated

src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
```diff
@@ -369,3 +369,131 @@ Value getDimOrConstant(ConversionPatternRewriter &rewriter, Location loc,
   }
   return dimVal;
 }
+
+/// Emit an ONNXSqueezeOp. If the input is constant, do const propagation, and
+/// return a constant.
+Value foldOrEmitONNXSqueezeOp(ConversionPatternRewriter &rewriter, Location loc,
```
AlexandreEichenberger, line 375: great name, super clear now

AlexandreEichenberger (commented):
thanks

src/Transform/ONNX/ConstPropHelper.hpp
```diff
@@ -0,0 +1,88 @@
...
+#include "mlir/Transforms/DialectConversion.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+#include "llvm/ADT/ArrayRef.h"
+
+#include "src/Dialect/ONNX/ONNXOps.hpp"
+#include "src/Pass/Passes.hpp"
+
+#include <math.h>
+
+using namespace mlir;
```
bxing-groq, line 29: I recommend adding some automation to catch namespace leaks like this! :P

PR #693, opened 2021-05-20
Use unit constants as none value

tungld:
Rewriting rules for Reduce ops used ONNXConstantOp as none value to pass to other ops as follows:
```
%1 = "onnx.Constant"() {value = dense<> : tensor<0xi64>} : () -> tensor<0xi64>
%2 = "onnx.ReduceSum"(%0, %1) {keepdims = 1 : si64, noop_with_empty_axes = 0 : si64} : (tensor<3x2x2xf32>, tensor<0xi64>) -> tensor<1x1x1xf32>
    return %2 : tensor<1x1x1xf32>
```
This is not a real none value, since the type of `%1` is `tensor<0xi64>`, not none.

A proper way should be using std.constant with unit attribute. We will get a none value.
```
%cst = constant unit
%1 = "onnx.ReduceSum"(%0, %cst) {keepdims = 1 : si64, noop_with_empty_axes = 0 : si64} : (tensor<3x2x2xf32>, none) -> tensor<1x1x1xf32>
```

This patch is to use unit constants for none value.


chentong319 (approved):
LGTM!

PR #694, opened 2021-05-20
Add NoSideEffect trait to KrnlGlobalOp

tungld:
KrnlGlobalOp is to define constants. Constants become dangling once there is not use, but they are not removed.
This patch marks KrnlGlobalOp with NoSideEffect so that dangling constants will be removed automatically when canonicalization is called.
 

AlexandreEichenberger (approved):
LGTM pending successful tests

PR #695, opened 2021-05-23
Tag dangling image when build fails (master)

gongsu832:
- Tag dangling image when build fails for easier debugging
- Build user image with Release type


caoimhinuibrian (approved):
LGTM

PR #699, opened 2021-05-25
Added return of nonzero value when errors occur

AlexandreEichenberger:

Current numerical may assert in the `rc::check` silently using the `make test` check. The fix here enforce a return of a non-zero values on `rc::check` error.

It will be enabled also for TestGemm, which currently fails.


doru1004 (approved):
LGTM

PR #700, opened 2021-05-25
 Support multiple versions of onnx op

chentong319:
When multiple version of an op is to be support in dialect, the old version will have their version number added in the operation name. 
Use ReduceSum as an example.
- Previously, version_dict["ReduceSum"] = 11; now version_dict["ReduceSum"] = {13, 11}
- Two operands, ONNXReduceSumOp and ONNXReduceSumV11Op, are defined in ONNXOps.td.inc. Only the non-head version will have version number attached.
- When importing a node from onnx model, choose the right operand based on the Opset of that node
- Replace the existing ONNXReduceSumOp in code with ONNXReduceSumV11Op and test.
- Add new code for ONNXReduceSumOp, which is version 13.
- Later remove ONNXReduceSumV11Op code when the new version stable. This cleanup can reduce the code to be maintained.

Even though converter can be used, this PR has the benefits for the process of upgrading by keep the existing code and handle model with mixed operations.


src/Conversion/ONNXToKrnl/Math/Reduction.cpp
```diff
@@ -55,6 +61,12 @@ struct ScalarOp<ONNXReduceProdOp> {
   using IOp = MulIOp;
 };
 
+template <>
+struct ScalarOp<ONNXReduceSumV11Op> {
```
AlexandreEichenberger, line 65: An alternative naming scheme would be to add the version numbering (with/without) the `V` for versioning at the end: `ONNXReduceSumOpV11`.  I leave it up to you.

utils/gen_onnx_mlir.py
```diff
@@ -47,193 +47,193 @@
 # run this script with --check-operation-version flag.
 # Update this dictionary when a newer version is implemented
 # TODO: how to keep the old version
```
AlexandreEichenberger, line 49: Yes, it may be a good place to add what to do when you want to keep the old ones.
Also, what do you recommend when there is an other operation, should it be -1... or some value that indicate "not supported?" Or not being there indicate not supported?

AlexandreEichenberger (commented):
Added a few minor questions.

src/Conversion/ONNXToKrnl/Math/Reduction.cpp
chentong319: I thought of that too. But I feel it is better to keep the ONNX....Op format for all ONNX operands.

utils/gen_onnx_mlir.py
chentong319: When searching for operation with version (in FrontendDialectTransformer.cpp), it is assumed that all versions are supported. If the version is not directly supported, the closest higher version is used. For example, if version 1, 3 and 5 are supported, version 3 is used for version 2 op in model. Can add an array for not supported op+version.

AlexandreEichenberger (approved):
LGTM

PR #701, opened 2021-05-25
[DO NOT COMMIT] Convert attributes to input tensor

chentong319:
One common change in onnx is that an attribute of an op is changed to input tensor in the new version. This PR is intended to handle that change so that the onnx-mlir is able to handle ops in old version with the new version op.
1. Specify which attribute is converted to which input in gen_onnx_mlir.py so that the op dialect has a mapping.
2. In frontend, convert the attributes according to the mapping


chentong319:
No need anymore. Better to use converter.

PR #702, opened 2021-05-25
Preserve mlir files when using  onnx-mlir or compileModule 

AlexandreEichenberger:

Enable the saving of the input and output mlir files when using the compileModule function (either called in drivers such as in numerical, or onnx-mlir). 

It can also be enabled by setting the KEEP_TEMP_FILES define in src/MainUtils.cpp. This enables to more easily look at the input ONNX mlir files generated by numerical or any other models using onnx-mlir, as it does not requires printout to stderr and/or adding a flag (for which the functionality does not appear to be here in the direct interfaces (only present via the onnx-mlir program)


caoimhinuibrian (approved):
LGTM

PR #704, opened 2021-05-26
Only enable pooling for top level allocs

doru1004:
This patch limits the memory pool creation to top level allocs where most of the space should be used. Loop invariant allocs that can be hoisted to the top level will still participate in pooling. Loop-local allocs will not be considered to avoid interference with loop optimizations.


AlexandreEichenberger (approved):
LGTM. Did you evaluate if the total usage of memory increased significantly?

doru1004:
> LGTM. Did you evaluate if the total usage of memory increased significantly?
No increase as far as I could tell.

PR #705, opened 2021-05-26
Fix issues of onnx.Split builder

negiyas:
Update utils/gen_onnx_mlir.py to generate builder for onnx.Split in src/Dialect/ONNX/ONNXOps.td.inc according in oder to calculate the output memref with considering the split parameter of the onnx.Split op.
 
 The following two files are updated.
1. **utils/gen_onnx_mlir.py** (updated to fix the onnx.Split builder issue)
2. **src/Dialect/ONNX/ONNXOps.td.inc** (replaced with one generated from the updated utils/gen_onnx_mlir.py ).


negiyas:

tungld:

utils/gen_onnx_mlir.py
```diff
@@ -425,6 +425,29 @@
...
+ '''   let builders = [
+  OpBuilder<(ins "Value":$input, "IntegerAttr":$axis, "ArrayAttr":$split), [{
+   Type elementType = input.getType().cast<ShapedType>().getElementType();
+   std::vector<mlir::Type> outputTypes;
+   if (split.size() == 0) {
+     outputTypes.emplace_back(UnrankedTensorType::get(elementType));
+   } else {
+     std::vector<int64_t> inputShape = input.getType().cast<MemRefType>().getShape();
+     int64_t splitAxis = axis.cast<IntegerAttr>().getSInt();
+     int64_t splitNum = (split.size() == 0) ? inputShape[splitAxis] : split.size();
```
chentong319, line 439: split.size() == 0 is checked at 434. Will this condition always be false here?

utils/gen_onnx_mlir.py
```diff
@@ -425,6 +425,29 @@
...
+  ];'''),
+  ('Split',
+ '''   let builders = [
+  OpBuilder<(ins "Value":$input, "IntegerAttr":$axis, "ArrayAttr":$split), [{
+   Type elementType = input.getType().cast<ShapedType>().getElementType();
+   std::vector<mlir::Type> outputTypes;
+   if (split.size() == 0) {
+     outputTypes.emplace_back(UnrankedTensorType::get(elementType));
+   } else {
+     std::vector<int64_t> inputShape = input.getType().cast<MemRefType>().getShape();
```
chentong319, line 437: How is the cast<MemRefType>() guaranteed to be successful? For MemRefType, do we need to worry unknown dimension?

chentong319 (commented):
It seems to me that this PR mixed shape inference for MemRef with builder. Does such situation occur for other ops? Should we have a general solution, such as construct and fold?

gongsu832:
@negiyas I added you to the admin list so your PR should now build automatically. And you should be able to comment with "@jenkins-droid test this please" to trigger a build manually.

utils/gen_onnx_mlir.py
negiyas: Its shapeInference code guarantees that the input tensor is RankedTensorType. I fixed the code to convert the input tensor to UnrankedTensorType.

utils/gen_onnx_mlir.py
negiyas: You are right. I removed the check. Thank you.

negiyas:
@gongsu832 Thank you!

negiyas:
@chentong319 Thank you very much for your useful comments.
Let me explain my motivation for this PR.
I am searching for the way to implement it, and your comments and suggestions are very welcome.

I want to write a [Table-driven Declarative Rewrite Rule (DRR)](https://mlir.llvm.org/docs/DeclarativeRewrites/) like as follows.
 
```
def : Pattern<(XXXXOp ...),
              [(ONNXSplitOp:$results, ...), // Bind the results to "$results"
               (AOp (BOP $results__0), (COp $results__1)) // Refer the results as "$results__N" in different operations.
               ...
              ]
             >;
```
We can build the ONNXSplitOp with NativeCodeCall without this PR, but the current DRR implementation does not allow to use NativeCodeCall as "auxiliary ops." We cannot write it as follows.
```
class NativeOp<string s> :  NativeCodeCall<"$_builder.create<" # s # ">($_loc, $0, $1, ...)">;
def : Pattern<(XXXXOp ...),
              [(NativeOp<"ONNXSplitOp">:$results, ...),  // Bind the results to "$results"
               (AOp (BOP $results__0), (COp $results__1)) // Refer the results as "$results__N" in different operations.
               ...
              ]
             >;
```

chentong319:
I understand that we need to define customized builder when the op is generated in DRR. The PR enhanced the existing customized builder with shape inference for split. Obviously, that part of code is duplicated. Can we call shape inference after rewrite? Which rewrite rule, in onnx or lowering, uses this builder? I will approve this PR if it is really needed for some model and we do not have a better solution.

negiyas:
@chentong319 I am closing this PR, because we have found another way to solve this situation without updating the onnx dialect definition. 
(1) Create a new operation similar to onnx.Split in another dialect with a custom builder.
(2) Write DRR rule to emit the new operation without using NativeCodeCall.
(3) Write another rule to lower the new operation to onnx.Split.

Thank you very much for your comments, and sorry for troubling you.

PR #709, opened 2021-05-28
add util tool to run a model.so with random inputs

AlexandreEichenberger:

Retooled my prior tool to use the LLVM::JSON reader. I changed it so that the tool read a dynamically linked library at runtime so that the tool can be compiled once and run multiple times with different example.

This tool can be used as simply as this:
```
TEST_CASE_BY_USER=test_add_cpu make check-onnx-backend

cd test/backend
run-onnx-lib test_add.so
```

This is helpful to investigate a model that may crash, and can be also use to run gdb, or memcheck, as it is purely a C++ invocation method.

Below are the current options.
```
Usage: run-onnx-lib [options] model.so

  Program will instantiate the model given by "model.so"
  with random inputs, launch the computation, and ignore
  the results. A model is typically generated by lowering
  an ONNX model using a "onnx-mlir --EmitLib model.onnx"
  command. When the input model is not found as is, the
  path to the local directory is also prepended.

  Options:
    -e name | --entry-point name
         Name of the ONNX model entry point.
         Default is "run_main_graph".
    -n NUM | --iterations NUM
         Number of times to run the tests, default 1
    -v | --verbose
         Print the shape of the inputs and outputs
    -h | --help
         help
```

Instructions are given in the file on how to compile the tool (tested on MacOS and Z at this time)


doru1004 (approved):
LGTM

tungld:
@AlexandreEichenberger, as a user of this tool, may I ask whether this tool supports a model whose inputs have unknown dimensions? Thanks!

As you know, we have a python [debug.py](https://github.com/onnx/onnx-mlir/blob/master/utils/debug.py) which also runs a model with random inputs (but its then compares with a reference backend, e.g. onnxruntime, we can make this optional). What is the difference between this tool and debug.py?

AlexandreEichenberger:
@tungld 

their goals are different, as debug.py uses a lot of python and other things to verify the accuracy.
This tool uses a straightforward call in C++ so that you can run the debugger and / or valgrind on it without processes and other things coming in the way. Convenient to investigate runtime failures / crashes.

At this time, it supports only known dimensions. I can add support for runtime dimensions, if needed. Let me know.

PR #711, opened 2021-06-01
Invoke onnx version converter

chentong319:
Invoke onnx version converted before the onnx model is translated into our onnx mlir dialect. By default, the converter is not called. The flag to turn on the converter is `--invokeOnnxVersionConverter`. The current onset supported by onnx-mlir is defined in `ONNXOps.hpp` 
The converter is not called by default because converter generated some assertion errors when I tried it on our check-onnx-backend models.


AlexandreEichenberger (approved):
Smart. I would like you to open defect on the ONNX converter repo about the errors we get, to let them know of it. Can you add a section in the doc on handling versions, and one sentence on this option? Tx


tx

PR #713, opened 2021-06-02
removed files that belongs to Dialect/ONNX to be all under that directory

AlexandreEichenberger:
Currently, there were two optimizations, listed under Transform/ONNX/ Combine & Rewrite that were really just an implementation of the ONNX XXX Ops definition. It was awkward to have the implementations of the ONNX Ops in two different libraries. Now all the ONNX Ops classes are in the Dialect/ONNX library.


AlexandreEichenberger:
@sstamenova If you have the time, feel free to check the CMakeList.txt. I am by far not an expert, I migrated the file and added the new .td file to the list of dependences to be performed for the compilation in Dialect/ONNX to happen. I worry that there is a dependence between first building the ONNXOps.td before the Combine.td and Rewriter.td... but it has not occurred in my parallel build. Maybe the system is smart enough?

chentong319:
Should we move the common constructor utility functions in combine.cpp, rewrite.cpp and etc into ONNXOpsHelper.cpp? Also how about some common definition in *.td file?

src/Dialect/ONNX/CMakeLists.txt
```diff
@@ -14,12 +14,19 @@ add_custom_target(OMONNXOpsInc
         DEPENDS OMONNXOpsIncGen
                 OMResultTypeInferenceOpInterfaceIncGen
                 OMShapeInferenceOpInterfaceIncGen
-                OMHasOnnxSubgraphOpInterfaceIncGen)
+                OMHasOnnxSubgraphOpInterfaceIncGen
+                OMONNXRewriteIncGen
+                OMONNXCombineIncGen)
+
+add_onnx_mlir_rewriter(Rewrite)
```
sstamenova, line 21: Move these before adding the custom target since the custom target depends on the two targets created by the add calls

sstamenova:
> @sstamenova If you have the time, feel free to check the CMakeList.txt. I am by far not an expert, I migrated the file and added the new .td file to the list of dependences to be performed for the compilation in Dialect/ONNX to happen. I worry that there is a dependence between first building the ONNXOps.td before the Combine.td and Rewriter.td... but it has not occurred in my parallel build. Maybe the system is smart enough?

My expectation is that tablegen in the same file will happen sequentially, which is why it is working; however, if you want to guarantee it, you can add the dependencies like so:

```
add_dependency(OMONNXRewriteIncGen OMONNXOpsIncGen)
add_dependency(OMONNXCombineIncGen OMONNXOpsIncGen)
```

This is not going to impact the build negatively (since that's the order it needs to happen anyway), so there's no downside to adding it.

AlexandreEichenberger:
>Should we move the common constructor utility functions in combine.cpp, rewrite.cpp and etc into ONNXOpsHelper.cpp? Also how about some common definition in *.td file?

Done, migrated the supports in combine and rewrite into ops helper. for the td file, I am going to leave it as is for the moment.

src/Dialect/ONNX/CMakeLists.txt
AlexandreEichenberger: done

AlexandreEichenberger:
> This is not going to impact the build negatively (since that's the order it needs to happen anyway), so there's no downside to adding it.

make sense, I like it, will add,  thanks

src/Dialect/ONNX/CMakeLists.txt
```diff
@@ -31,6 +31,8 @@ add_onnx_mlir_library(OMONNXOps
         IndexExprDetail.cpp)
 
 add_dependencies(OMONNXOps OMONNXOpsInc)
+add_dependencies(OMONNXRewriteIncGen OMONNXOpsIncGen)
```
sstamenova, line 34: I would keep the dependencies closer to where the targets are declared. In this case the add_onnx_mlir_rewriter calls. It makes the file easier to read when they are co-located.

tungld:
Is it reasonable to merge Combine and Rewrite into a single file? They both implements canonicalization patterns for ONNX ops.

doru1004 (approved):
LGTM

AlexandreEichenberger:
>  it reasonable to merge Combine and Rewrite into a single file? They both implements canonicalization patterns for ONNX ops.

Did not see your comment before I merged. Let me check, this seems reasonable.

PR #715, opened 2021-06-04
Support .mlir files as inputs to onnx-mlir command

tungld:
This patch is to allow passing .mlir files as inputs to onnx-mlir command.

Currently, `onnx-mlir` only accepts .onnx files as its input. But we often emit intermediate .mlir files, say, for debugging, (e.g. using --EmitMLIR), do some modifications, and want to continue compiling to LLVMIR or a binary library. 
This patch extends `onnx-mlir` to accept .mlir files as well, so we can take intermediate .mlir files and continue the compilation.

Minor change: remove a redundant parameter in method `processInputFile`.


AlexandreEichenberger (approved):
LGTM, let me know if it is ready to squash into master

tungld:
Yes, it is ready. Thanks!

PR #716, opened 2021-06-04
Create initial scope when translating functions

gramalingam:
Fix issue relating to the introduction of nested-name-scopes that affects translation of ONNX functions. (Need to call pushScope/popScope for the initial scope in a function context).


chentong319:

chentong319:
Can you explain why a new scope is needed for function call?

gramalingam:
ONNX functions are used to help translate ONNX ops (e.g., custom ops) which can be expanded into an ONNX sub-graph. The translation of this expanded sub-graph, however, should NOT have access to the outer-scope names. Hence, a fresh new scope is used for the translation of this sub-graph.

chentong319 (approved):
LGTM!

PR #717, opened 2021-06-06
Onnx1.9

chentong319:
Move onnx to 1.9.
The operators is not changed to upset 14 for 1.9 yet.  There is no interface change. 
onnx 1.9 has an important fix for version converter. The patch will allow we to invoke converter to handle the old models.


caoimhinuibrian (approved):
LGTM, but correct spelling of BatchNorm

test/backend/test.py
```diff
@@ -896,6 +903,10 @@ def prepare(cls, model, device='CPU', **kwargs):
             command_list.append("--mcpu="+args.mcpu)
         if args.mtriple:
             command_list.append("--mtriple="+args.mtriple)
+        if args.converter:
+            command_list.append("--invokeOnnxVersionConverter=true")
+        elif name in test_need_converter :
+            command_list.append("--invokeOnnxVersionConverter=true")
```
gongsu832, line 909: Can simply as
```
if args.converter or name in test_need_converter:
```

test/backend/test.py
```diff
@@ -780,6 +784,9 @@
 # or big models
 test_for_dynamic = [ key for (key, value) in test_to_enable_static_dynamic.items() if value[0] & test_dynamic ]
 
+# Specify the test cases which need version converter
+test_need_converter = []
```
gongsu832, line 788: Perhaps do something similar to `TEST_CASE_BY_USER` so that one can specify the tests on command line without having to change `test.py`
```
test_need_converter = []

TEST_NEED_CONVERTER = os.getenv("TEST_NEED_CONVERTER")
if TEST_NEED_CONVERTER is not None and TEST_NEED_CONVERTER != "" :
    test_need_converter = TEST_NEED_CONVERTER.split()
```

test/backend/test.py
chentong319: sure

test/backend/test.py
chentong319: The env variable TEST_NEED_CONVERTER serves this purpose: just set the env variable without modifying test.py.

gongsu832: Not sure I understand. I'm suggesting adding above code. I don't see TEST_NEED_CONVERTER in the current patch.

gongsu832: But setting `TEST_CASE_BY_USER` does not change the list variable `test_need_converter`. In the current patch, `test_need_converter` is always empty. So not sure why we need it then.

chentong319: The test_need_converter will not be empty when more op (e.g. squeeze) is updated to the new version. I noticed some module test failed in the PR for squeeze/unsqueeze. 
The env variable will be used when we debug one test case with TEST_CASE_BY_USER.

gongsu832: So `test_need_converter` will be hardwired in `test.py` to hold the list of tests we know requiring converter?

chentong319: yes. Otherwise our check-onnx-backend can not pass.

gongsu832: OK I thought it was supposed to be changeable by users through env var like `test_to_enable` (via `TEST_CASE_BY_USER`). Then it's fine. We can worry about that later.

PR #718, opened 2021-06-09
HandleLLVMOptions and library dependencies

sstamenova:
This change adds HandleLLVMOptions to the cmake file and then makes the necessary changes to support it. More specifically:
1) HandleLLVMOptions enforces that shared libs be able to link. So for each of the OM libraries, add the correct linking dependencies
2) To make 1) easier, extend the add_onnx_mlir_* functions to support dependencies, link libraries and include directories
3) Once all the OM libs link correctly, we no longer need to support a manual OMLibs list - we can rely on the global list created by add_onnx_mlir_library for linking
4) Similarly, once all the OM libs link correctly, we no longer need to support a manual list of MLIRLibs - they are all added to the appropriate OMLibs or executables instead and propagated elsewhere
5) Since onnx uses exceptions, using HandleLLVMOptions requires that we set LLVM_REQUIRES_EH to ON so that fno-exceptions is not set for onnx-mlir
6) Once all the dependencies are resolved, the custom IncGen targets for krnl and onnx are no longer needed
7) We no longer need to set /bigobj since HandleLLVMOptions does this. The same is true for /Md

A couple of caveats: there's an issue with assertions with llvm, clang and protobuf. This change sets LLVM_ENABLE_ASSERTIONS to OFF when the compiler is clang to work around it. This change also does not touch the Runtime/JNI targets. They both need some additional attention so that they can build correctly on both Windows and Linux, so they have been left alone in anticipation of getting some of those issues fixed.


src/Dialect/ONNX/CMakeLists.txt
```diff
@@ -12,29 +12,23 @@ add_onnx_mlir_dialect_doc(onnx ONNXOps.td)
...
+  Rewrite.cpp
+  Combine.cpp
+  IndexExpr.cpp
+  IndexExprDetail.cpp
 
-add_dependencies(OMONNXOps OMONNXOpsInc)
-add_dependencies(OMONNXRewriteIncGen OMONNXOpsIncGen)
-add_dependencies(OMONNXCombineIncGen OMONNXOpsIncGen)
+  DEPENDS
+  OMHasOnnxSubgraphOpInterfaceIncGen
```
stephenneuendorffer, line 25: It has a header dependency here, but really doesn't have a linkage dependency?  Usually, the corresponding libraries would be in LINK_LIBS PUBLIC

stephenneuendorffer (approved):
Looks reasonable to me.  Thanks alot for doing all this!

src/Dialect/ONNX/CMakeLists.txt
sstamenova: That's what I expected also, but it is happy to build without the library, at least in my testing on Ubuntu and Windows. I'm curious to see if any of the automated builds will fail, but as long as they don't, I think this is fine.

stephenneuendorffer: Usually BUILD_SHARED_LIBS=on is stricter about such dependencies..  It's possible it's not necessary if it's only using #defines and such.

caoimhinuibrian:

MLIR.cmake
```diff
@@ -19,27 +19,35 @@ include(TableGen)
...
+# When LLVM_ENABLE_ASSERTIONS is set to ON, the build is not Debug and the compiler
+# is clang, onnx_proto will fail to link to libprotobuf due to missing symbols when
+# we let HandleLLVMOptions actually set all the expected properties correctly (such
+# as removing NDEBUG from the compiler options). This appears to be a bug in the way
+# the libs interact with clang, because everything links fine with gcc in this situation
+# and everything links fine with clang when LLVM_ENABLE_ASSERTIONS is OFF. In order to
+# support the build with clang until this can be resolved, we set LLVM_ENABLE_ASSERTIONS
+# to OFF before including HandleLLVMOptions. This does not reproduce with clang on
+# Windows.
+if ("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang" AND (NOT CMAKE_BUILD_TYPE STREQUAL "Debug"))
```
caoimhinuibrian, line 31: @sstamenova  How should we resolve this?

caoimhinuibrian (approved):
LGTM, assuming the builds work

MLIR.cmake
sstamenova: This is not trivial to fix (a.k.a. I haven't figured it out yet despite several hours spent on it). Unless you definitely want LLVM_ENABLE_ASSERTIONS to be possible to be set to ON for onnx-mlir, I think it is OK for the moment until it can be resolved later. I don't think LLVM_ENABLE_ASSERTIONS is required for onnx-mlir.

sstamenova:
@caoimhinuibrian : I fixed the issue with ExecutionSession, I believe the builds should be fine now.

caoimhinuibrian:

src/Dialect/ONNX/CMakeLists.txt
sstamenova: All the builds work correctly without the lib dependency, so I believe it is unnecessary.

MLIR.cmake
sstamenova: @caoimhinuibrian: I can actually remove this workaround in the flags PR - cleaning up the flags that were being set explicitly in the cmake lets clang work correctly with LLVM_ENABLE_ASSERTIONS=ON. So once this change is committed, I will update the flags PR to remove the workaround.

PR #723, opened 2021-06-16
Shape inference fix

gramalingam:
A couple of fixes to enable the translation of ONNX functions using MLIR functions:

(a) The existing shape-inference pass has a hack that checks if the module contains any function with a name matching the pattern "[a-zA-Z0-9_]*main_graph" and, if so, restricts inference to those functions. Adding an option analyzeAllFunctions to skip this filter, since when translating ONNX functions into MLIR functions it is useful to apply shape-inference to all functions (even though it has a main_graph).

(b) Refine the translation process when using the option useOnnxModelTypes (which enables the translator to use type-information contained in the input model during translation) to use the resultTypeInference method if the input model doesn't contain type information.


gramalingam:
Closing this PR and recreating it due to signoff/DCP issues.

PR #725, opened 2021-06-17
Add shape inference pass option to analyze all functions

gramalingam:
(Closing PR #723 and recreating it due to signoff/DCO issues.)

A couple of fixes to enable the translation of ONNX functions using MLIR functions:

(a) The existing shape-inference pass has a hack that checks if the module contains any function with a name matching the pattern "[a-zA-Z0-9_]*main_graph" and, if so, restricts inference to those functions. Adding an option analyzeAllFunctions to skip this filter, since when translating ONNX functions into MLIR functions it is useful to apply shape-inference to all functions (even though it has a main_graph). Also, allow function-calls in code being checked.

(b) Refine the translation process when using the option useOnnxModelTypes (which enables the translator to use type-information contained in the input model during translation) to use the resultTypeInference method if the input model doesn't contain type information.

### DIFF
--- a/src/Builder/FrontendDialectTransformer.cpp
+++ b/src/Builder/FrontendDialectTransformer.cpp
@@ -232,13 +232,14 @@ private:
     return RankedTensorType::get(tensor_dims, elementType);
   }
 
-  Type ConvertOnnxType(const std::string &onnx_name) {
-    auto it = value_info_map.find(onnx_name);
-    if (it != value_info_map.end()) {
-      return ImportTensorType(it->second);
-    } else {
-      return builder_.getNoneType();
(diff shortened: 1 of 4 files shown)


AlexandreEichenberger:
@gramalingam thanks for the update, @chentong319 mind looking at it since its frontend?

I have a more general question: is there a reason to restrict analysis only to not having `analyzeAllFunctions` the default and only option? What was the motivation behind only analyzing one function if matching *main_graph*?

Similarly, why do we not always use the onnx types.

It would be good to have our reasoning documented, and if the options are necessary, let's go with them, and add the proper help for users to know which one to use.

chentong319 (approved):
LGTM!

PR #726, opened 2021-06-18
Resize Op with static shape and constant scales

chentong319:
This PR support shape inference and lowering for Resize Op with static shape and constant scales.  Only the nearest mode is supported. 

1. Shape inference. Use `output_dimension = floor(input_dimension * (roi_end - roi_start) * scale) if input "sizes" is not specified.` to compute the output_dimension. The standard says `It (roi) only takes effect when coordinate_transformation_mode is "tf_crop_and_resize"`. Current implementation is for "asymmetric" and rio_end and rio_start are ignored.
2. In lowering, the index is computed as 
```
if coordinate_transformation_mode is "asymmetric", 
x_original = x_resized / scale, 
```
3. examples from standard:
resize_downsample_scales_nearest and resize_upsample_scales_nearest
Please note that in resize_downsample_scales_nearest, for the last dimension, when x_resized = 1 and scale = 0.6, the x_original is 2, not 1. Need to figure out why.
4. The targeted model may have dynamic shape input tensor for Resize. It seems that IndexExpr should be used to handle the dynamic shape. @AlexandreEichenberger  There are float and integer type casting for index computation in Resize. Can IndexExpr handle those operations?

### DIFF
--- a/src/Conversion/ONNXToKrnl/Tensor/Resize.cpp
+++ b/src/Conversion/ONNXToKrnl/Tensor/Resize.cpp
@@ -44,13 +44,14 @@ struct ONNXResizeOpLowering : public ConversionPattern {
     if (hasAllConstantDimensions(memRefType))
       alloc = insertAllocAndDealloc(memRefType, loc, rewriter, insertDealloc);
     else
-      llvm_unreachable("unknown shape for output");
+      // TODO: handle dynamic shape in alloc
+      return emitError(loc, "unknown shape for output");
 
     // Get the scales
     DenseElementsAttr scalesAttrs =
         getDenseElementAttributeFromONNXValue(resizeOp.scales());
     if (!scalesAttrs)
-      llvm_unreachable("Not implemented yet");
+      return emitError(loc, "Not implemented yet");
     SmallVector<float, 4> scalesConstant;
     for (auto scaleAttr : scalesAttrs.getValues<FloatAttr>()) {
       scalesConstant.emplace_back(scaleAttr.getValueAsDouble());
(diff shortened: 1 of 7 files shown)


src/Conversion/ONNXToKrnl/Tensor/Resize.cpp
```diff
@@ -0,0 +1,95 @@
...
+    Location loc = op->getLoc();
+    Value alloc;
+    bool insertDealloc = checkInsertDealloc(op);
+    ONNXResizeOp resizeOp = llvm::cast<ONNXResizeOp>(op);
+
+    // Check implementation constraints
+    if (resizeOp.mode() != "nearest" ||
+        resizeOp.coordinate_transformation_mode() != "asymmetric" ||
+        resizeOp.nearest_mode() != "floor")
+      llvm_unreachable("not implemented yet");
```
AlexandreEichenberger, line 36: When we don't support something, don't we want to give a more friendly error message? Also, I am not sure what remains from such messages in the Release mode. Please use an emitError message

src/Conversion/ONNXToKrnl/Tensor/Resize.cpp
```diff
@@ -0,0 +1,95 @@
...
+    ONNXResizeOpAdaptor operandAdaptor(operands);
+    Value data = operandAdaptor.X();
+
+    MemRefType memRefType = convertToMemRefType(*op->result_type_begin());
+    int64_t rank = memRefType.getShape().size();
+
+    if (hasAllConstantDimensions(memRefType))
+      alloc = insertAllocAndDealloc(memRefType, loc, rewriter, insertDealloc);
+    else
+      llvm_unreachable("unknown shape for output");
```
AlexandreEichenberger, line 47: Same here; please add a TODO about dyn shapes.

AlexandreEichenberger (commented):
> The targeted model may have dynamic shape input tensor for Resize. It seems that IndexExpr should be used to handle the dynamic shape. @AlexandreEichenberger There are float and integer type casting for index computation in Resize. Can IndexExpr handle those operations?

not yet. My suggestion is to force code gen by using an 
``` 
Value val = indexVar.getValue();
Value floatVal =  <<conversion operation>>>
```

AlexandreEichenberger (approved):
LGTM

caoimhinuibrian:

PR #727, opened 2021-06-18
Add Range operation shape inference and lowering to Krnl

doru1004:
Add Range operation shape inference and lowering to Krnl.


src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -3300,7 +3300,78 @@ LogicalResult ONNXRandomUniformLikeOp::inferShapes(
 
 LogicalResult ONNXRangeOp::inferShapes(
     std::function<void(mlir::Region &)> doShapeInference) {
-  return emitError(NOT_IMPLEMENTED_MESSAGE);
+  // Cannot infer shape if no shape tensor is specified.
```
AlexandreEichenberger, line 3303: Can you add a TODO if some of the restrictions are simplifications to get a PR quickly... i.e. be explicit below which restrictions should be lifted to be standard compliant. Same with dynamic sizes, if supported. Tx

AlexandreEichenberger (commented):
Please confirm that there is no runtime checks; if there is, please enable.

Tx

AlexandreEichenberger (approved):
LGTM

AlexandreEichenberger:
Thanks for adding the backend test

PR #728, opened 2021-06-18
removed EDSC in matmul

AlexandreEichenberger:
First steps of removing EDSC, tested removing it in matmul which used to use structured krnl iterate. Was able to preserve the structured approach, but using create<KrnlIterateOp> instead of the previous EDSC krnl_iterate.

Note that I lifted some of the useful new additions in MLIR to support the EDSC-free approach, so that I may use it in the code without having to merge yet with MLIR. Once we are totally free of EDSC, I will remove this file at the same time as actually merging with MLIR. 


NathanielMcVicar:
LGTM, I still don't fully understand the krnl loop lambda, but I like the approach to replacing EDSC a lot. I can help when it's time to update LLVM and remove the helper, since we have all the other updates done.

AlexandreEichenberger:
@NathanielMcVicar Thanks, that will be great. But there remain a difficult part, namely handling the IndexExpr, as they also use a thread private builder. I will have to give it special consideration, as they also use the Scopes... Once I have a proof of concept for that one, we can go crazy changing the code...

AlexandreEichenberger:
@NathanielMcVicar The lambda is relatively easy: think of it as a function that you call to fill in the loop body. That function will be invoked when the time is right, namely the loop structure has been established and the "insert cursor" is inside the loop. That is pretty much it. Lambda vs function is convenient because it let you use variables from the context where the function is defined... simplifies a lot for the parameter passing... That is it in a nutshell.

NathanielMcVicar:
Sounds like a good plan for the others! I should have been more clear, the part I was having trouble with wasn't the lambda itself, it's why the indices are what they are.

PR #731, opened 2021-06-20
Handle dynamic input of Resize Op

chentong319:
This PR handles non-constant scales() and unknown size of input data X().
Extend the implementation to handle nearest_mode() == "round_prefer_floor" and coordinate_transformation_mode() == "half_pixel", the default value for the optional attributes. With these support, some backend test for resize in onnx can be passed. 
IndexExpr utilities are used and mixed with std ops due to the type conversions. The IndexExpr code is directly added in lowering currently.


AlexandreEichenberger (approved):
LGTM, thanks for the quick turnaround.

PR #734, opened 2021-06-22
Search .so inside jar

gongsu832:
- Search .so inside jar to load instead of hardcode .so name
- Remove bogus --target option on cmake when building MLIR


caoimhinuibrian (approved):
LGTM

src/Runtime/jni/src/com/ibm/onnxmlir/OMModel.java
```diff
@@ -1,58 +1,81 @@
...
+			// Copy .so to the temporary directory
+			Files.copy(jf.getInputStream(jf.getEntry(libFile)), lib,
+				   StandardCopyOption.REPLACE_EXISTING);
+
+			// z/OS USS requires "x" permission bit
+			Files.setPosixFilePermissions(lib,
+			      PosixFilePermissions.fromString("rwxr-xr-x"));
+
+			// Load the temporary .so copy
+			System.load(lib.toString());
```
seroyer, line 56: Is there any concept of ordering the loading of libraries?  For example if you want library foo.so loaded before bar.so?  Right now you seem to be at the mercy of the order of the files in the jar.

src/Runtime/jni/src/com/ibm/onnxmlir/OMModel.java
```diff
@@ -1,58 +1,81 @@
...
+                                              .getLocation().toURI());
+	    Path jarDir = jar.getParent();
+
+	    // All .so files found will be copied under the ./native subdirectory
+	    // where the jar file resides.
+	    Path tmpDir = Files.createTempDirectory(jarDir, "native_");
+
+            // Open jar file to load all .so libs inside. The .so is loaded
+            // by OS so it must be extracted to the file system.
+            JarFile jf = new JarFile(jar.toFile());
```
seroyer, line 36: nit, there appears to be a minor indentation error here.

gongsu832: No problem in the file. Github is somehow displaying the diff wrong.

src/Runtime/jni/src/com/ibm/onnxmlir/OMModel.java
gongsu832: Nothing is being resolved here. The libraries are simply being loaded to make the symbols available so it makes no difference which library is loaded first.

PR #735, opened 2021-06-22
Remove nearly all build warning

AlexandreEichenberger:
A few are left behind, like unused functions. I did not cleanup any of the thrid_party warnings, but removed unused variables.


doru1004 (approved):
LGTM

PR #736, opened 2021-06-22
Prepare IndexExpression Scope for EDSC-free MLIR

AlexandreEichenberger:
rewriters will become "local", so I added the rewriter in the IndexExprScope definitions.


doru1004 (approved):
LGTM

PR #737, opened 2021-06-23
Handle scalar attributes in constprop

### LINKED ISSUE #733: Asserting error for model yolo
Error message is:
```
Assertion failed: (numElements != 1 && "splat of 1 element should already be detected"), function getKey, file /Users/shizy/Projects/github/llvm-project/mlir/lib/IR/AttributeDetail.h, line 115.
```
This is called from ConstPropHelper.cpp:143. I dumped the resType is <f32> and the sizeInBytes is 4. 
@tungld I think you are familiar with the constant from raw buffer. Could you take a look?
Currently, I commented out the assert in AttributeDetail.h:115 and the model is parsed. I think this assert is not correct.

tungld:
This patch fixes an issue in constant propagation when creating a DenseElementsAttr whose type is scalar tensor using a raw buffer. For scalar tensor, we should mark the attribute with splat.

Resolves #733 


AlexandreEichenberger (approved):
LGTM, thanks for the quick turnaround

PR #738, opened 2021-06-24
Rewrite BatchNormTestMode into x*a+b

tungld:
This patch is to rewrite BatchNormTestMode into `x * a + b` where `a` and `b` are computed from the constant inputs (scale, bias, mean, var).


AlexandreEichenberger:
@tung: looks good, why "Test Mode"?

tungld:
> why "Test Mode"?

It refers to "inference mode". It's a bit confusing, so if you are agree, we can change the op name to BatchNormInferenceMode in this patch also.

AlexandreEichenberger (approved):
LGTM, approved pending change to "Inference" mode. Test means something else :-) Let me know when the change is done. Saw at least a test in the lit-tests

tungld:
@AlexandreEichenberger I changed TestMode to InferenceMode. It is ready for merging.

PR #739, opened 2021-06-24
removed edsc from all matmul onnx lowering

AlexandreEichenberger:
Added support for krnl constructors for krnl iterate with index expression, and used it in lowering of matmul from onnx to krnl IR.

migrated matmul lit test to the ones using canonicalize


doru1004 (approved):
LGTM

PR #742, opened 2021-06-26
change shape inference pass

chentong319:
Two changes:
1. Add unknown size tensor as candidates. For example, refine tensor<?xf32> to tensor<10xf32>.
2. Remove the check of tensor of unknown dimension. Even though I can exclude tensor with ? dim in the check, I think we should not do this check in case we can gradually refine the shape information.

This change is useful to do experiment with yolo model.


src/Transform/ONNX/ShapeInferencePass.cpp
```diff
@@ -166,8 +154,12 @@ class ShapeInferencePass : public mlir::PassWrapper<ShapeInferencePass,
    */
   static bool returnsDynamicShape(Operation *op) {
```
tjingrant, line 155: The naming of this function is vague. I think according to the current definition, it is better to name this function as "hasExplicitResultShape", and negate all usages of the original function output.

tjingrant (approved):
LGTM, with nit.

src/Transform/ONNX/ShapeInferencePass.cpp
```diff
@@ -164,10 +152,14 @@ class ShapeInferencePass : public mlir::PassWrapper<ShapeInferencePass,
...
-  static bool returnsDynamicShape(Operation *op) {
+  static bool returnsDynamicOrUnknownShape(Operation *op) {
     return llvm::any_of(op->getResultTypes(), [](Type result_type) {
-      return !result_type.isa<NoneType>() &&
-             !result_type.isa<RankedTensorType>();
+      if (result_type.isa<RankedTensorType>()) {
+        return llvm::any_of(result_type.dyn_cast<RankedTensorType>().getShape(),
+            [](int64_t dim) { return dim < 0; });
+      } else {
+        return !result_type.isa<NoneType>();
```
gramalingam, line 161: Wouldn't it be better to explicitly test if this is an unranked tensor-type? Is there an assumption right now that the only types of values are either tensors or NoneType? I was wondering what happens in the future, when there is a mix of code returning primitive types, or other types.

chentong319: @gramalingam I read your comment after I merged the PR. Yes, it is better to check unranked tensor-type.

PR #724, opened 2021-06-16
Build JNI on Windows

sstamenova:
This change makes JNI buildable on Windows. It also cleans up the logic for determining when to build JNI and adds the necessary dependency to MainUtils

PR #744, opened 2021-06-28
Implement thread local storage with pthread_setspecific/pthread_getspecific

gongsu832:
Implement thread local storage with pthread_setspecific/pthread_getspecific for archs that do not support __thread


sstamenova:
This won't work correctly on Windows. I spent some time getting the jni libs compiling on Windows in https://github.com/onnx/onnx-mlir/pull/724 and I'd like to make sure we don't get the Windows build broken again.

gongsu832:
Do you mean there will be merging conflicts or Windows does not support pthread_setspecific/pthread_getspecific?

sstamenova:
> Do you mean there will be merging conflicts or Windows does not support pthread_setspecific/pthread_getspecific?

I suspect there will be merge conflicts. The main concern though is that Windows does not support pthread at all. There's a different set of APIs to achieve the same on Windows and llvm exposes a consistent set of APIs that use pthread on Ubuntu and Tls on Windows via ThreadLocal.

gongsu832:
Merge conflict should be easy to deal with. Does Windows support `__thread`? If not, then we need to use whatever the equivalent of pthread calls on Windows. There are just 3 calls needed: `pthread_once`, `pthread_setspecific`, and `pthread_getspecific`.

sstamenova:
> Merge conflict should be easy to deal with. Does Windows support `__thread`? If not, then we need to use whatever the equivalent of pthread calls on Windows. There are just 3 calls needed: `pthread_once`, `pthread_setspecific`, and `pthread_getspecific`.

The equivalent on Windows is: https://docs.microsoft.com/en-us/cpp/parallel/thread-local-storage-tls?view=msvc-160

Is there a reason not to use the thread local implementation from LLVM: https://github.com/llvm/llvm-project/blob/d480f968ad8b56d3ee4a6b6df5532d485b0ad01e/llvm/include/llvm/Support/ThreadLocal.h?

gongsu832:
> Is there a reason not to use the thread local implementation from LLVM: https://github.com/llvm/llvm-project/blob/d480f968ad8b56d3ee4a6b6df5532d485b0ad01e/llvm/include/llvm/Support/ThreadLocal.h?

The JNI wrapper is not part of LLVM lowering. Since Windows has a `__thread` equivalent, it makes things much simpler. The pthread implementation is only effective when `USE_PTHREAD_SPECIFIC` is defined (I missed a couple in the patch which I will fix).

@caoimhinuibrian @AlexandreEichenberger we can get PR https://github.com/onnx/onnx-mlir/pull/724 merged first, then I will get this PR fixed to resolve any conflicts.

sstamenova:
> @caoimhinuibrian @AlexandreEichenberger we can get PR #724 merged first, then I will get this PR fixed to resolve any conflicts.

After #724 is committed, JNI will also build as part of the Windows CI, so it should catch any changes that cause failures.

AlexandreEichenberger:
@sstamenova that makes sense, since @gongsu832 is our JNI guy, let's just wait for his final ok on #724 before merging it. Thanks for your patience.

caoimhinuibrian:
> > @caoimhinuibrian @AlexandreEichenberger we can get PR #724 merged first, then I will get this PR fixed to resolve any conflicts.
> 
> After #724 is committed, JNI will also build as part of the Windows CI, so it should catch any changes that cause failures.

@sstamenova @AlexandreEichenberger I'm cool with merging #724 first. It looks fine to me, but let's wait for @gongsu832

caoimhinuibrian:
@gongsu832 is this ready to go now? or still waiting on a Windows solution?

gongsu832:
@caoimhinuibrian This has conflicts with the recently merged PR https://github.com/onnx/onnx-mlir/pull/724 that I need to resolve.

src/Runtime/jni/jnilog.c
```diff
@@ -29,6 +32,63 @@
 
 #include "jnilog.h"
 
+#define INLINE static __inline
```
sstamenova, line 35: Is there a reason to use a define instead of just ```static __inline``` directly in the signatures?

src/Runtime/jni/jnilog.c
```diff
@@ -140,21 +214,25 @@ static FILE *get_log_file_by_name(char *name) {
  */
 void log_init() {
 
-  if (log_inited)
+#ifdef USE_PTHREAD_SPECIFIC
```
sstamenova, line 217: Why not add another helper function to encapsulate this? Then log_init won't have any ifdefs and will just call helper functions.

src/Runtime/jni/jnilog.c
gongsu832: The `#define` is a reminder that `static __inline` is platform dependent and not all platforms necessarily support it. Currently all the platforms we support support it so it may seem redundant. But if we ever have to deal with a platform that doesn't support inline or uses some other directive, `INLINE` can be defined accordingly.

src/Runtime/jni/jnilog.c
gongsu832: Was lazy to deal with it since it's just one line. But it's good to get rid of the #ifdef so I used a PTHREAD_ONCE macro. Thanks.

src/Runtime/jni/jnilog.c
sstamenova: I think it would be better to be explicit and use `static __inline`, so that the code is easier to read. And if there's ever a need to support a platform that does not support it, `INLINE` can be defined at that time.

gongsu832: ...and doing double amount of work to get what's there now. I'm not sure what's not explicit about INLINE. As mentioned I intentionally use it as a reminder and I don't see it makes the code any harder to read.

sstamenova: Macros (even simple ones) require an extra step to consider when reading the code. For example, in this case when I first looked at the code, I had to go back and check how the macro was defined because I was confused that it existed at all since `static __inline` generally works.

I find that there's little value in preparing for the possibility of a use (like another platform that doesn't support `static __inline`) at the expanse of current readability, but if you have strong feelings about using the macro, that's fine too.

src/Runtime/jni/jnilog.c
```diff
@@ -140,21 +218,23 @@ static FILE *get_log_file_by_name(char *name) {
  */
 void log_init() {
 
-  if (log_inited)
+  PTHREAD_ONCE(&key_once, key_init);
```
sstamenova, line 221: I think it is confusing to have a single PTHREAD_* macro in an otherwise generic function. I think it would be cleaner if the functionality was hidden inside a function that can be a noop on Windows (or other platforms that don't support pthread). Having said that, this is better than using the ifdef, so if you want to leave it this way, that's fine too.

src/Runtime/jni/jnilog.c
gongsu832: See PTHREAD_ONCE comment below.

src/Runtime/jni/jnilog.c
gongsu832: I sense you are in general against macros. Since there are no clear rules on when one should use macros and when one should use functions, this mostly comes down to personal coding style. My general rule of thumbs is that if something is relatively simple and can be determined at compile time I would use macros. So unless the code vastly deviates from "common standards", I suggest we don't debate on coding style too much since it quickly becomes fruitless.

AlexandreEichenberger:
@sstamenova Let us know when you believe that this PR addresses your concerns about Window, or when it is in a state that let you adapt this PR with a later PR for window. Thanks for your input.

sstamenova:
> @sstamenova Let us know when you believe that this PR addresses your concerns about Window, or when it is in a state that let you adapt this PR with a later PR for window. Thanks for your input.

LGTM. It also builds cleanly as part of the Windows CI build.

src/Runtime/jni/jnilog.c
sstamenova: I don't mind macros, they can be really useful. My comment here was more about the naming - on Windows, we don't really use pthread* but a completely different set of functions, so having the macros be named PTHREAD_ONCE or PTHREAD_SELF, etc. makes the code read as less platform-agnostic. My suggestion would be to name them something that doesn't have pthread in the name since they don't always refer to actual pthread functions/objects.

sstamenova (approved):

src/Runtime/jni/jnilog.c
gongsu832: When everyone uses pthread except Windows, it would be bending backwards to not use pthread/PTHREAD in the name.

PR #746, opened 2021-06-29
EDSC free GEMM and MatMul using new scheme

AlexandreEichenberger:
Preferred way is to use the ArithBuilder and KrnlBuilder to create new operations. Compared to the `builder.create<OP>`, that new approach allow us to not have to specify as many type casts as the  `builder.create<OP>` is easily confused by types.

I made on purpose the KrnlBuilder methods to have zero-processing, just a direct feed through to the builder.create ops. So if adding a new Krnl op, just add the equivalent method there with the same parameter, and the new op will then be super easy to use.

This change is prompted because MLIR does not want to have "implicit" context scope anymore.


AlexandreEichenberger:
A bit of extra info. I added the KrnlBuilder (using the same approach as MLIR ArithBuilder. Below is a small extract of that class

```C++
struct KrnlBuilder : public DialectBuilder {
  KrnlBuilder(OpBuilder &b, Location loc) : DialectBuilder(b, loc) {}
  KrnlBuilder(ImplicitLocOpBuilder &lb) : DialectBuilder(lb) {}
  KrnlBuilder(DialectBuilder &db) : DialectBuilder(db) {}

  Value load(Value memref, ValueRange indices = {});
  Value loadIE(Value memref, ArrayRef<IndexExpr> indices);
  void store(Value val, Value memref, ValueRange indices = {});
  void storeIE(Value val, Value memref, ArrayRef<IndexExpr> indices);

  Value vectorTypeCast(Value sourceMemref, int64_t vectorLen);

  ValueRange defineLoops(int64_t originalLoopNum);
  ValueRange block(Value loop, int64_t blockSize);
  void permute(ValueRange loops, ArrayRef<int64_t> map);
  ValueRange getInductionVarValue(ValueRange loops);

  void iterate(ValueRange originalLoops, ValueRange optimizedLoops,
      ValueRange lbs, ValueRange ubs, ValueRange iterArgs,
      function_ref<void(KrnlBuilder &createKrnl, ValueRange args)>
          bodyBuilderFn);

...
};
```

and it can be used as follow. Note that the lambda passes again a KrnlBuilder that is updated for the inner region, so it works pretty smoothly. 
```C++
                createKrnl.iterate({}, {jj2, ii2}, {}, {}, {},
                    [&](KrnlBuilder &createKrnl, ValueRange args) {
                      ValueRange j2_i2_indices =
                          createKrnl.getInductionVarValue({jj2, ii2});
                      Value j2(j2_i2_indices[0]), i2(j2_i2_indices[1]);
                      for (unsigned int n = 0; n < bBuffs.size(); ++n) {
                        createKrnl.matmul(aBuff, {i1, k1}, bBuffs[n], {k1, j1},
                            cBuffs[n], {i1, j1},
                            /*loops*/ {ii3, jj3, kk2},
                            /*compute start*/ {i2, j2, k1},
                            /*ubs*/ {I, J, K},
                            /*compute tile*/ {iRegTile, jRegTile, kCacheTile},
                            /* a/b/c tiles*/ {}, {}, {}, simdize, unrollAndJam,
                            false);
                      }
                    });
```

chentong319:
I went through the code and here is my understanding how the EDSC is removed
1. The Context is nested through lambda function. The nesting of function implements the nesting of builder. This seems right to me. Only concern is the complexity o deeply nested constructing code. Will that happen?
2. Another criticism of EDSC is that explicit constructor need to be defined for each op and can not just use builder.create<op>. It seems that this patch still has that drawback.

AlexandreEichenberger:
> The Context is nested through lambda function. The nesting of function implements the nesting of builder. This seems right to me. Only concern is the complexity o deeply nested constructing code. Will that happen?

You are correct, and the nesting only changes how we get at the nested builders. Before it was done by "attaching" a nested builder to a thread private stack; now it is done explicitly. Note that the DialectBuilders (examples: ArithBuilder, KrnlBuilder) are very small object, just carrying a ref to the builder and a location.

> Another criticism of EDSC is that explicit constructor need to be defined for each op and can not just use builder.create. It seems that this patch still has that drawback.

That is correct. MLIR folks talked about a automatically generated builder... but this has not happened, to the best of my knowledge. This is why all of the smarts are in the builder/create functions attached to the original ops. KrnlBuilder/ArithBuilder are very very thin layers with no smarts in it... so very easy to add.

To better explain what we gain by using this layer, let me paste the code with/without using the native builder

```C++
                rewriter.create<KrnlIterateOp>(loc, ValueRange{}, ValueRange{jj2, ii2},
                    ValueRange{}, ValueRange{}, ValueRange{},
                    [&](ImplicitLocOpBuilder &lb, ValueRange args) {
            
```

vs

```C++
                createKrnl.iterate({}, {jj2, ii2}, {}, {}, {},
                    [&](KrnlBuilder &createKrnl, ValueRange args) {
```

So while it is true that we have to add a bit of code to support a new op via the KrnlBuilder, its improved readability is well worth it in my opinion.

Note too that the `rewriter.create<OP>` is available by default, and only for the ops that we gen a lot, we can choose to add it to the KrnlBuilder interface.

NathanielMcVicar:
LGTM

caoimhinuibrian (approved):
LGTM

PR #749, opened 2021-06-30
Put aligned constants on the heap instead of the stack

tungld:
This patch is to put aligned constants on the heap to avoid stack overflow.

### DIFF
--- a/src/Conversion/KrnlToLLVM/KrnlToLLVM.cpp
+++ b/src/Conversion/KrnlToLLVM/KrnlToLLVM.cpp
@@ -350,6 +350,7 @@ public:
     auto loc = op->getLoc();
 
     auto krnlGlobalOp = llvm::dyn_cast<KrnlGlobalOp>(op);
+    auto alignmentAttr = krnlGlobalOp.alignmentAttr();
 
     // Get module.
     ModuleOp module = op->getParentOfType<ModuleOp>();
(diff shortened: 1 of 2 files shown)


AlexandreEichenberger (commented):
@tungld is there performance implications with putting constants on the heap? Is it done early enough for the memory bundler to catch them?

Does it apply only to "Aligned" constants and not unaligned one?

Thanks for providing a bit of context.

tungld:
@AlexandreEichenberger we are currently using LLVM global constants to store our constants (produced by krnl.global). For non-aligned constants, they are accessed directly from the LLVM global constants. But for aligned constants, we need to allocate an aligned buffer and copy constant values into the buffer. 

> Does it apply only to "Aligned" constants and not unaligned one?

Yes, this applies to aligned constants only, in particular, aligned buffers are created on the heap instead of the stack.

>  is there performance implications with putting constants on the heap?
No, it is mainly for avoiding stack overflows.

> Is it done early enough for the memory bundler to catch them?

Uhm, it's a good point. It seems to me, in such case, we also need to copy constant values to a local aligned buffer. At a high level, say before memory bundler, the copy should be like reading constant values one-by-one, and emitting `krnl.store` one-by-one. This looks like a regression that the generated IR was full of `krnl.store`. Still not clear what would be a good way to utilize the memory bundler.

AlexandreEichenberger:
@tungld how large are these constants? If large, then should we plan enabling mem bundling? If so, let me know if you prefer to merge now and fix later, or if you want to investigate first.

tungld:
@AlexandreEichenberger please merge it now. I don't think I can have a complete solution in near future (say, a month or so).

AlexandreEichenberger (approved):
LGTM fo now. I would suggest that constant should eventually go in the binary itself.

PR #750, opened 2021-06-30
Prototype of Instrumentation

chentong319:
Experimental support instrumentation on Ops: onnx.Conv,  element binary op and variadic element op (such as onnx.Mul).

The usage is in docs/Instrument.md.

There two parts of the code:
1. Compiler.
2. Runtime lib.

Limitations:
1. Instrumentation is represented as KrnlInstrumentOp. Currently two attributes, ID and tag, are defined for this Op. It is planned to use ID to index the debug info for this op and tag to represent some runtime control. But currently, ID stores the name of op and tag just for before or after the op.
2. Currently the instrumentation is added in lowering onnx to Krnl. A separate pass may be developed later.
3. Runtime lib is as a part of OMTensor lib. Can be put a separate lib.
4. Runtime lib can be optimized, such as using buffer instead of printf, using performance counter and etc.


src/Conversion/KrnlToLLVM/KrnlToLLVM.cpp
```diff
@@ -501,6 +521,48 @@ class KrnlGlobalOpLowering : public ConvertToLLVMPattern {
...
+      : ConversionPattern(KrnlInstrumentOp::getOperationName(), 1, context) {}
+
+  LogicalResult matchAndRewrite(Operation *op, ArrayRef<Value> operands,
+      ConversionPatternRewriter &rewriter) const override {
+    auto *context = op->getContext();
+    KrnlInstrumentOpAdaptor operandAdaptor(operands);
+    auto loc = op->getLoc();
+    KrnlInstrumentOp instrumentOp = llvm::dyn_cast<KrnlInstrumentOp>(op);
+
+    // Get a symbol reference to the memcpy function, inserting it if necessary.
```
caoimhinuibrian, line 536: change memcpy in comment to OMInstrumentPoint

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -725,6 +725,7 @@ struct ONNXElementwiseBinaryOpLowering : public ConversionPattern {
         NameLoc::get(Identifier::get(ElementwiseBinaryOp::getOperationName(),
                          op->getContext()),
             op->getLoc());
+    rewriter.create<KrnlInstrumentOp>(loc, op, 1);
```
caoimhinuibrian, line 728: can we have an enum for the tag argument? then we could print something meaningful?

src/Runtime/OMInstrument.inc
```diff
@@ -0,0 +1,86 @@
...
+     mypid = getpid();
+     sprintf(memCommand, "ps -o vsz --noheader -p %d", mypid);
+  }
+}
+
+void OMInstrumentPoint(int64_t id, int64_t tag) {
+  if (timeEnabled) {
+    struct timeval newTimeValue, result;
+    gettimeofday(&newTimeValue, NULL);
+    printf("ID=%s TAG=%ld ",(char *)&id, tag);
```
caoimhinuibrian, line 73: see earlier comment about having an enum for the tag. Also, for the op, can we print something more informative than a number?

src/Conversion/KrnlToLLVM/KrnlToLLVM.cpp
chentong319: fixed

src/Runtime/OMInstrument.inc
chentong319: The ID is changed to op name string. 
Enumerate is not used for tag because I am think tag may be a bit mask to provide more control on instrumentation.

caoimhinuibrian (approved):
LGTM

PR #755, opened 2021-07-01
Implement compileModuleToSharedLibrary on Windows

sstamenova:
compileModuleToSharedLibrary and friends currently do not support building on Windows - all of the default flags and properties do not work with the Windows build infrastructure. This change implements compileModuleToSharedLibrary (and any necessary functions) on Windows.

cruntime had to be updated to only depend on .c files since the .cpp file that was part of the library was not being built on Windows resulting in missing definitions when linking.

The numerical tests are also updated to work on Windows and two new build targets have been added: numerical (to build all numerical tests) and check-numerical (to run all numerical tests).

This change touches some of the JNI functions in MainUtils, but it is not sufficient to make them work on Windows. They've only been updated as a side-effect of changes to function common to both JNI and shared libs.


test/numerical/CMakeLists.txt
```diff
@@ -1,86 +1,82 @@
...
+
+add_numerical_unittest(TestGRU
+  TestGRU.cpp
+  LINK_LIBS PRIVATE ${TEST_LINK_LIBS}
+  )
+
+add_numerical_unittest(TestLoop
+  TestLoop.cpp
+  LINK_LIBS PRIVATE ${TEST_LINK_LIBS}
+  )
```
gongsu832, line 82: Function `add_numerical_unittest` is defined with one parameter but called with three arguments?

sstamenova: Yes, the rest of the arguments are passed through to add_onnx_mlir_executable. By not formally declaring all the arguments, we don't have to specify what they are and how to parse them and can use ARGN to get the whole collection of arguments past the first and pass them to the function that creates the executable.

gongsu832: Why not just write
```
add_onnx_mlir_executable(${test_name} NO_INSTALL ${test_name}.cpp LINK_LIBS PRIVATE ${TEST_LINK_LIBS})
```
and then simply write
```
add_numerical_unittest(TestConv)
add_numerical_unittest(TestMatMul2D)
...
```

sstamenova: @gongsu832 : I wanted the usage to look more like add_executable or add_library. If you prefer me to change the usage to just the name of the test, I can do that.

gongsu832: No that's fine. Just wondering the rational behind. Perhaps add what you just said as a comment to `add_numerical_unittest`? Thanks.

gongsu832:

AlexandreEichenberger:
@gongsu832 looked at it and would like you to determine if its ok or of any changes that you may want to request.

gongsu832:
@AlexandreEichenberger LGTM

sstamenova:
@AlexandreEichenberger: Do you need anything else before this can be committed?

caoimhinuibrian:

caoimhinuibrian (approved):
LGTM

PR #756, opened 2021-07-02
Update the equation used for printing out wrong results.

imaihal:
PR #549 updated the equation for validating calculation results, but, the equation used for printing out wrong results was not updated. This PR updates it.


tjingrant (approved):
LGTM.

PR #757, opened 2021-07-02
Remove EDSC standard and MemRefBoundsCapture

AlexandreEichenberger:
Next steps, remove EDSC standard references plus MemRefBoundsCapture. 
Unfortunately, it changes a few constant being generated, so I had to update the lit tests for all RNN computations.


chentong319 (approved):
This PR is the application of ImplicitLocOpBuilder, which captures current Loc. I am not sure about the benefit over the increased code complexity (I mean the extra class). Since the previous PR has been approved, this PR itself is OK.

AlexandreEichenberger:
@chentong319 We can look into not using ImplicitLocOpBuilder, it is in the latest MLIR, but it does not mean that we need it. I am open to a cleanup after some discussion once the latest MLIR is integrated

PR #758, opened 2021-07-05
Add an option shapeInformation to set static shapes for the inputs

tungld:
This patch introduces an option, `shapeInformation`, to command `onnx-mlir`, which is to set static dimensions for the inputs of a model.

The motivation is that compiling a model with dynamic dimensions may introduce performance overhead. If we know the exact shapes of the inputs to run the model inference, we should give the compiler such static shapes so that more optimizations can be done.

 
### DIFF
--- a/src/MainUtils.cpp
+++ b/src/MainUtils.cpp
@@ -84,9 +84,15 @@ llvm::cl::opt<bool> useOnnxModelTypes("useOnnxModelTypes",
     llvm::cl::init(false), llvm::cl::cat(OnnxMlirOptions));
 
 llvm::cl::opt<string> shapeInformation("shapeInformation",
-    llvm::cl::desc("Custom shapes for the inputs of the ONNX model"),
-    llvm::cl::value_desc("<input_id:dim,dim,dim|input_id:dim,dim,dim>"),
-    llvm::cl::cat(OnnxMlirOptions));
+    llvm::cl::desc(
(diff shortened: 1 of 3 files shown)


src/MainUtils.cpp
```diff
@@ -83,6 +83,11 @@ llvm::cl::opt<bool> useOnnxModelTypes("useOnnxModelTypes",
     llvm::cl::desc("use types and shapes from ONNX model"),
     llvm::cl::init(false), llvm::cl::cat(OnnxMlirOptions));
 
+llvm::cl::opt<string> shapeInformation("shapeInformation",
+    llvm::cl::desc("Custom shapes for the inputs of the ONNX model"),
+    llvm::cl::value_desc("<input_id:dim,dim,dim|input_id:dim,dim,dim>"),
+    llvm::cl::cat(OnnxMlirOptions));
```
AlexandreEichenberger, line 89: please add a sentence to explain what the input_id is, whether dim are only positive numbers (zero allowed???)

AlexandreEichenberger (approved):
LGTM, conditionally approved after adding some info in the docs... and maybe a sentence more to the online doc.

Do you know that LLVM has an embedded Json reader? You could also use it if you though there are advantages to the formatting.

src/MainUtils.cpp
tungld: Done. Thanks for your comments!

PR #761, opened 2021-07-07
Migrate llvm

AlexandreEichenberger:
Migrate LLVM to current version (July 6th)


imaihal:
@AlexandreEichenberger The error on Jenkins Linux s390x should be coming from `-fno-semantic-interposition` option when building MLIR using GCC. I created a patch to disable it https://reviews.llvm.org/D105453 , , but it is not merged it yet.

AlexandreEichenberger:
> @AlexandreEichenberger The error on Jenkins Linux s390x should be coming from `-fno-semantic-interposition` option when building MLIR using GCC. I created a patch to disable it https://reviews.llvm.org/D105453 , but it is not merged yet. Can you wait for a while?

No problems, I will test manually to see that the errors I see here are indeed fixed by your patch.

AlexandreEichenberger:
Confirming that https://reviews.llvm.org/D105453 fixes the Linux s390. @imaihal please try to respond to the reviewer's request to enable that PR as fast as possible, thanks.

imaihal:
@AlexandreEichenberger The patch was merged! Please use newer commit than `62cc3cdda64aaaca5bfeba4b131ab6b5f50b4e0c`. Thanks!

sstamenova:
@AlexandreEichenberger : We're really excited about this change, thanks for going through all the hard work of removing EDSC and getting here!

src/Conversion/KrnlToLLVM/CMakeLists.txt
```diff
@@ -6,9 +6,11 @@ add_onnx_mlir_library(OMKrnlToLLVM
   LINK_LIBS PUBLIC
   OMSupport
   MLIRAffineToStandard
+  MLIRMemRefToLLVM
```
sstamenova, line 9: When I imported the mlir.cmake file, I tried to limit the libraries that things get linked against to the fewest possible. One thing we could do if we want to avoid some of the churn here is link against *all* conversion libraries. I think it's better to keep a limited list, but it is more work.

src/Dialect/ONNX/MLIRDialectBuilder.hpp
```diff
@@ -0,0 +1,54 @@
+//===- MLIRDialectBuilder.hpp - Support migration to MLIR without EDSC --===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+
+#ifndef TMP_MLIR_DIALECT_STANDARDOPS_UTILS_UTILS_H
```
sstamenova, line 8: Nit: this still has TMP in the name

src/Dialect/ONNX/MLIRDialectBuilder.hpp
```diff
@@ -0,0 +1,54 @@
...
+  OpBuilder &getBuilder() { return b; }
+  Location getLoc() { return loc; }
+
+protected:
+  OpBuilder &b;
+  Location loc;
+};
+
+//===----------------------------------------------------------------------===//
+// from Utils.h
```
sstamenova, line 34: MLIR Utils.h?

doru1004 (approved):
LGTM pending comment resolution.

src/Conversion/KrnlToLLVM/CMakeLists.txt
AlexandreEichenberger: @sstamenova I added the lib here because MemRef was split away from Standard, it now exists as its own lib. Without it here, I get a missing code error when linking. I don't have issue with keeping a limited list at this time. Hopefully these "splitting the libs" will remain rare events. However, if you are inclined to make a link against all conversion libs, I have no problem with that, but in a subsequent PR.

sstamenova: No, I like the short list and I don't think linking against all the conversion libs is necessary. But since I saw that you had to add a couple of them, I thought I'd offer it as an option if you are so inclined.

src/Dialect/ONNX/MLIRDialectBuilder.hpp
AlexandreEichenberger: Tx, fixed it.

src/Dialect/ONNX/MLIRDialectBuilder.hpp
AlexandreEichenberger: Updated the comment.

sstamenova (approved):

PR #762, opened 2021-07-08
Improve the performance of LSTM and GRU

tungld:
This patch is to use`X * WTranspose` instead of `W * XTranspose` in LSTM for better performance because WTranspose can be done statically during compile time.

The same thing is applied to GRU in linearBeforeReset mode so that the number of matmuls is reduced from 6 to 2.


tungld:
@AlexandreEichenberger do you have any idea on why this happened?

AlexandreEichenberger:
@tungld could you check if it also occurs with simd off? (which you can disable in MatMul.cpp). There might be a problem when the result size is not a multiple of 4 along the simonized dimension.

tungld:
@AlexandreEichenberger Both LSTM and MatMul failed if I turned off simd in matmul.cpp.

Error with LSTM:
```
attempt 4 with direction -1, S 1, B 8, I 5, H 8, isDynamicS 0, isDynamicB 0
loc("onnx.MatMul"): error: failed to legalize operation 'krnl.load'
Shape mismatch 1,1,9,9 != 1,1,8,8
```
TestMatMul2D failed also:
```
Shared library ./TestMatmul_main_graph.so has been compiled.
attempt 352 with i 4, j 8, k 4
loc("onnx.MatMul"): error: failed to legalize operation 'krnl.load'
a[0,0] = 0.158778 != b[0,0] = 0.0876529
a[0,1] = -0.33527 != b[0,1] = 0.336544
a[0,2] = -1.13575 != b[0,2] = 1.10128
a[0,3] = -0.469264 != b[0,3] = 0.488915
a[0,4] = -0.343288 != b[0,4] = 0.350792
a[0,5] = -0.613047 != b[0,5] = 0.589918
a[0,6] = 0.783186 != b[0,6] = 0.752333
a[0,7] = 0.376558 != b[0,7] = 0.319272
a[1,0] = -0.606309 != b[1,0] = 0.339673
a[1,1] = 0.58774 != b[1,1] = 0.978273
a[1,2] = 1.34535 != b[1,2] = 0.154687
a[1,3] = 1.33281 != b[1,3] = 1.42537
a[1,4] = -0.813417 != b[1,4] = 1.52815
a[1,5] = 0.191417 != b[1,5] = 0.115867
a[1,6] = -0.0449755 != b[1,6] = 0.52419
a[1,7] = 0.389184 != b[1,7] = 0.548371
a[2,0] = -0.0470706 != b[2,0] = 1.75203
a[2,1] = -0.215459 != b[2,1] = 0.853235
a[2,2] = -0.689007 != b[2,2] = 0.84091
a[2,3] = -0.145059 != b[2,3] = 0.0804008
a[2,4] = -0.45686 != b[2,4] = 0.0487449
a[2,5] = -0.521973 != b[2,5] = 0.487904
a[2,6] = 0.69908 != b[2,6] = 0.473509
a[2,7] = 0.448288 != b[2,7] = 0.691782
a[3,0] = 1.04053 != b[3,0] = 0.154918
a[3,1] = 0.230348 != b[3,1] = 0.104694
a[3,2] = -1.46278 != b[3,2] = 0.614773
a[3,3] = -0.44938 != b[3,3] = 0.0484673
a[3,4] = -0.290006 != b[3,4] = 0.638153
a[3,5] = -0.169058 != b[3,5] = 0.474646
a[3,6] = 0.30455 != b[3,6] = 0.677399
a[3,7] = -0.2883 != b[3,7] = 0.355181
```

tungld:
> loc("onnx.MatMul"): error: failed to legalize operation 'krnl.load'

This was because of unrollAndJam with krnl.load/store. It should be fixed by #769. 

> @tungld could you check if it also occurs with simd off? (which you can disable in MatMul.cpp)

I confirmed that it did not occurred if simd off and unrollAndJam off, but it occurred if simd on and unrollAndJam off.
So, it seemed simd caused the issue.

AlexandreEichenberger:
checking again now that a few new patches went in.

AlexandreEichenberger:
Error is still present. Strange that we have a failure only for Z, wonder what may be happening.

tungld:
@AlexandreEichenberger increasing the threshold for the `krnl.global` lowering resolved the issue. It seems `opt` triggered some optimization for constants like what we saw before.

Anyway, the patch is ready for review/merge now.

AlexandreEichenberger:
> increasing the threshold for the krnl.global lowering resolved the issue. It seems opt triggered some optimization for constants like what we saw before.

@tungld Is it something that we can have a fix to avoid such issues in the future?

src/Conversion/ONNXToKrnl/RNN/GRU.cpp
```diff
@@ -171,6 +170,10 @@ getWeightPack<ONNXGRUOp, GruWeightPack>(
   Value R = op->R();
   // direction
   StringRef direction = op->direction();
+  // linear_before_reset.
+  bool linearBeforeReset = true;
```
AlexandreEichenberger, line 174: Isn't this just
```
linearBeforeReset = (op->linear_before_reset() != 0)

tungld: Just the personal style to read the code smoothly. The short form somehow takes me a few seconds to understand.

tungld:
> @tungld Is it something that we can have a fix to avoid such issues in the future?

Perhaps, we should investigate the threshold `opt` is using. At the same time, we want to have systematic tests for constant inputs like what we have done with dynamic inputs. Currently, only LSTM/GRU/RNN ops are tested with constant inputs. It's better to do with other ops also.

AlexandreEichenberger (approved):
LGTM

PR #763, opened 2021-07-08
Basic support for c and jni runtimes on z/OS

seroyer:
Use __MVS__ standard macro to identify when building on z/OS.  Only the
cruntime and jniruntime need to be built on z/OS, the rest of onnx-mlir
is not needed and would be non-trivial to get working.

Also depends on https://github.com/onnx/onnx-mlir/pull/744

There is no CI support for building this subset on z/OS, and cmake is not supported


caoimhinuibrian:

caoimhinuibrian (approved):
LGTM

PR #765, opened 2021-07-12
Script to run onnx-mlir from docker container

gongsu832:
- Convenient python script to run onnx-mlir from docker container
- Check freopen failure in MainUtils.cpp:outputCode
- Ignore invalid -o option values in main.cpp
- Update README.md about the convenient python script

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,13 +11,23 @@ The Open Neural Network Exchange implementation in MLIR (http://onnx.ai/onnx-mli
 | amd64-Windows | [![Build Status](https://dev.azure.com/onnx-pipelines/onnx/_apis/build/status/MLIR-Windows-CI?branchName=master)](https://dev.azure.com/onnx-pipelines/onnx/_build/latest?definitionId=9&branchName=master)             |
 | amd64-macOS   | [![Build Status](https://github.com/onnx/onnx-mlir/workflows/Build%20x86%20onnx-mlir%20on%20macOS/badge.svg)](https://github.com/onnx/onnx-mlir/actions?query=workflow%3A%22Build+x86+onnx-mlir+on+macOS%22)             |
 
-## Prebuilt Container
+## Prebuilt Containers
 An easy way to get started with ONNX-MLIR is to use a prebuilt docker image.
 These images are created as a result of a successful merge build on the trunk.
 This means that the latest image represents the tip of the trunk.
 Currently there are both Release and Debug mode images for `amd64`, `ppc64le` and `s390x` saved in Docker Hub as, respectively, [onnxmlirczar/onnx-mlir](https://hub.docker.com/r/onnxmlirczar/onnx-mlir) and [onnxmlirczar/onnx-mlir-dev](https://hub.docker.com/r/onnxmlirczar/onnx-mlir-dev).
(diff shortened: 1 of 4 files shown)


caoimhinuibrian (approved):
LGTM

caoimhinuibrian:
@sstamenova there were some Windows changes and the Windows CI is failing, could you take a look please?

gongsu832:
I'm guessing the problem is that `stderr` is a constant on Windows. Not sure how to reopen `stderr` on Windows in this case.

sstamenova:
> I'm guessing the problem is that `stderr` is a constant on Windows. Not sure how to reopen `stderr` on Windows in this case.

That is indeed the case: https://docs.microsoft.com/en-us/cpp/c-runtime-library/stdin-stdout-stderr?view=msvc-160

```
D:\a\1\onnx-mlir\src\MainUtils.cpp(562,39): error C2106: '=': left operand must be l-value [D:\a\1\onnx-mlir\build\src\MainUtils.vcxproj]
D:\a\1\onnx-mlir\src\MainUtils.cpp(573,37): error C2106: '=': left operand must be l-value [D:\a\1\onnx-mlir\build\src\MainUtils.vcxproj]
```

I am not sure if it is possible to re-open stderr or stdout on Windows, but I was wondering - why do we need to do any redirection for stderr at all? As far as I know, the ```print()``` function can take any ```raw_ostream```, so to print the output to a file, we just need to pass it to print instead of ```llvm::errs```. Am I missing a use case?

gongsu832:
I didn't write that code originally but did feel it's a bit strange with the redirection. Perhaps @tjingrant can offer some insights?

sstamenova:
> I didn't write that code originally but did feel it's a bit strange with the redirection. Perhaps @tjingrant can offer some insights?

@tjingrant @gongsu832 : I'd like to clean this bit of code (using stderr for file output) unless there's a reason to keep it as is. If you have any objections, please let me know, otherwise I'll follow up with a PR some time next week.

AlexandreEichenberger:
> @tjingrant @gongsu832 : I'd like to clean this bit of code (using stderr for file output) unless there's a reason to keep it as is. If you have any objections, please let me know, otherwise I'll follow up with a PR some time next week.

@sstamenova  What is the status of this. Should we go with this PR as is, and you propose your cleanup in a different PR? That would appear the easiest way to go about it. Let me know, thanks

sstamenova:
> > @tjingrant @gongsu832 : I'd like to clean this bit of code (using stderr for file output) unless there's a reason to keep it as is. If you have any objections, please let me know, otherwise I'll follow up with a PR some time next week.
> 
> @sstamenova What is the status of this. Should we go with this PR as is, and you propose your cleanup in a different PR? That would appear the easiest way to go about it. Let me know, thanks

This can't be checked in as is because it will fail the Windows build.

sstamenova:
https://github.com/onnx/onnx-mlir/pull/784 makes the necessary changes to write directly to the output file

AlexandreEichenberger:
@gongsu832 I pushed #784 to the master branch, there are now conflict with this one as you forecasted. I let you fix them, thanks

gongsu832:
@AlexandreEichenberger The conflict has been resolved and it's ready to go.

PR #769, opened 2021-07-12
Enabling unroll and jam in matmul

AlexandreEichenberger:
see related issue (#770)


AlexandreEichenberger:
Found a way to make it work, will push solution soon.

AlexandreEichenberger:
Solution was to record which loop to unroll during the lowering from krnl to affine, and once we have affine-only code in that pass, we look into the list of loop to unroll to perform that operation.

Because I need some global state (record loop to unroll during the pattern matching, perform the unrolling in the pass once all patterns have matched), I had to add locks to protect adding a new list in the map <function -> list>. 

I would rather be able for the pattern to access a "current list" that is added to the state of the patterns, but I did not know how to do that. If anyone knows, I would love to hear about it.

AlexandreEichenberger:
I also had to transform the scalar reduction variable into a private array (of size of the unrolling factor) to legally perform the unroll and jam.

tungld (approved):
LGTM.

tungld:
@AlexandreEichenberger is it ready to merge so that we can enjoy the optimization for GEMM?

PR #775, opened 2021-07-14
Run onnx lib timing

AlexandreEichenberger:
Added the ability to make measurements:
```
    -m NUM | --meas NUM
         Measure the kernel execution time NUM times.
         Min 5 iters, shortest/longest point dropped.
```

And for example:
```
(onnx-mlir) alexe@pancetta simd % onnx-mlir --EmitLib matmul256.mlir && run-onnx-lib -m 30  matmul256.so
Shared library matmul256.so has been compiled.
Load model file matmul256.so with entry point run_main_graph
Finish computing 30 iterations

Measurements (eliminating the two extreme points)
  timer med, 1944, avg, 1930, std, 67, min, 1810, max, 2571, sample, 28
```

Where the median, avg, standard dev, min and max are provided.

PS: this was build on top of #769. Only the changes in RunONNXLib.cpp are relevant to this PR


AlexandreEichenberger:
Updated the interface and the printout of the results. Verified the times by corroborating the execution times of the entire command divided by the number of the measurements.

```
run-onnx-lib -m 18 gemm1-1000-2048.so
Shared library gemm1-1000-2048.so has been compiled.
Load model file gemm1-1000-2048.so with entry point run_main_graph
  Did not find model, try in current dir ./gemm1-1000-2048.so
Start computing 18 iterations
  computed 10 iterations
Finish computing 18 iterations
@time, milli-second, med, 4.4, avg, 4.4, std, 0.0, min, 4.4, max, 4.525000, sample, 16
@time, micro-second, med, 4431.0, avg, 4436.0, std, 24.9, min, 4416.0, max, 4525.000000, sample, 16
```

caoimhinuibrian (approved):
LGTM

chentong319 (commented):
1) Since the driver may be used for a model, scale in second may be needed.
2) The min and max in output is actually the second min/max. This info is not in the print out message. When the num of execution is large, the second min/max are not that much different from the first min/max.  If num is small, we can print out all the result.

AlexandreEichenberger:
>Since the driver may be used for a model, scale in second may be needed.

It is already included, when the time can be measured in seconds.

>The min and max in output is actually the second min/max. This info is not in the print out message. When the num of execution is large, the second min/max are not that much different from the first min/max. If num is small, we can print out all the result.

Good point. Now report min/max/avg/std using all the points.

PR #777, opened 2021-07-14
Fix instrument error on windows

chentong319:
The OM runtime library used libraries not available on windows. This temporary fix added `#ifdef _WIN32' with `TO BE FIXED` comment.
When instrumentation part becomes mature, we will add the corresponding part for windows.

### DIFF
--- a/src/Runtime/OMInstrument.inc
+++ b/src/Runtime/OMInstrument.inc
@@ -33,18 +33,22 @@
 
 #include <stdio.h>
 #include <string.h>
+#include <unistd.h>
 
 #include "onnx-mlir/Runtime/OMInstrument.h"
 
+#ifdef _WIN32
+  // TO BE FIXED 
+#else
 #include <sys/time.h>
-#include <string.h>
-#include <unistd.h>
 #include <sys/types.h>
 
 static struct timeval globalTimeVal, initTimeVal;
+static pid_t mypid;
+#endif
+
 static bool timeEnabled = true;
 static bool virtualMemEnabled = true;
-static pid_t mypid;
 static char memCommand[200];
 
 void OMInstrumentInit() {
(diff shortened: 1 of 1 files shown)


src/Runtime/OMInstrument.inc
```diff
@@ -55,6 +59,9 @@ void OMInstrumentInit() {
 	  virtualMemEnabled = !virtualMemEnabled;
   }
 
+#ifdef _WIN32
```
sstamenova, line 62: Rather than having ifdefs in the functions, can you create a couple of helper functions and/or macros to encapsulate the functionality that will be different and then just call them? That will make the code easier to read and maintain.

sstamenova (changes requested):

src/Runtime/OMInstrument.inc
chentong319: Made the change. I changed the semantics of reporting memory in case threads are forked along the execution.

AlexandreEichenberger (approved):
LGTM, thanks for the quick turnaround as the error blocks all other PRs

sstamenova (approved):

AlexandreEichenberger:
@chentong319 

new error, apparently:
```
:\a\1\onnx-mlir\src\Runtime\OMInstrument.inc(36,10): fatal error C1083: Cannot open include file: 'unistd.h': No such file or directory (compiling source file D:\a\1\onnx-mlir\src\Runtime\OMInstrument.c) [D:\a\1\onnx-mlir\build\src\Runtime\cruntime.vcxproj]
```

chentong319:
Fixed that one, hopefully. We need a windows env.

AlexandreEichenberger:
@chentong319 When did you push your latest changes? It is 9:15 am July 15th, and Windows CI shows 1h20 of computations.

PR #779, opened 2021-07-15
Reshape/Squeeze/Unsqueeze without data copy

tungld:
`memref.reinterpret_cast` op in the MemRef dialect is to create a new view of a memref without copying the data buffer of the memref.
This patch lowers ONNX Reshape/Squeeze/Unsqueeze to `memref.reinterpret_cast` to avoid data copy.

- [x] Rewrite dimension computation using IndexExpr
- [x] Use `reinterpret_cast` instead of `krnl.memcpy`


AlexandreEichenberger (approved):
LGTM, thank

PR #780, opened 2021-07-15
Add control on instrument

chentong319:
This patch add a pass to insert instrument op. The pass is controlled by command line options. Details can be found in docs/Instrumentation.md.
Other change is on the control and format of instrument runtime.
Known issue: How to organize files in onnx-mlir to make it easier to define options.  
1. Currently, our option definition should use OptionCategory, `OnnxMlirOptions`, which is defined in MainUtils.cpp.  Therefore I should define the option in MainUtils.cpp and pass it through pass pointer. It is not convenient. Can OnnxMlirOptions be put in a separate include file? 
2. I used cl::bits type options, which is built on a enum type. This type should not be exposed to MainUtils.cpp.
3. However, this enum type is also used by runtime. I could put it in a separate include file. But where in our directory structure? The runtime is c and onyx-mlir is c++. In this patch, I just duplicated the enum type.

### DIFF
--- a/src/Transform/ONNX/InstrumentONNXPass.cpp
+++ b/src/Transform/ONNX/InstrumentONNXPass.cpp
@@ -23,6 +23,7 @@
 #include "llvm/Support/raw_ostream.h"
 
 #include "src/Dialect/ONNX/ONNXOps.hpp"
+#include "src/Dialect/Krnl/KrnlOps.hpp"
 #include "src/Interface/ShapeInferenceOpInterface.hpp"
 #include "src/Pass/Passes.hpp"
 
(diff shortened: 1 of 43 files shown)


caoimhinuibrian (approved):
LGTM

src/Transform/ONNX/InstrumentONNXPass.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+using namespace mlir;
+
+namespace {
+
+/*!
+ * This pass insert KrnlInstrumentOp before and after each ONNX ops
+ */
+
+// TO FIX: this option should be put in the ONNXMlirOptions category
```
sstamenova, line 38: I was actually just looking at this for some options that we'd like to add as well without adding the dependency of various libs on MainUtils. I think clang has a CommonOptions library which defines a variety of command line options that can then be re-used elsewhere by linking against this specific library which is just used for options. How do you feel about doing something similar here? Then we could add all of the options to a single library and link each of the onnx mlir libs that need to know about the options against that?

caoimhinuibrian: @sstamenova that seems reasonable to me, @chentong319 what do you think?

chentong319: I just found out that mlir pass manager provides mechanism to define the option inside the pass. I should try that.

chentong319: @sstamenova @caoimhinuibrian  MLIR provides some example of defining options in a pass. It seems to me that options for each pass can be defined within the pass. In this way, options are closed to their consumer code. The patter used by clang, which put the option together in a separate file, is good for options shared by different passes, such as options for target architecture, optimization level, debugging status.  I will do further investigation.

AlexandreEichenberger:
@chentong319 great new features. For future PR, generating comma separated  strings makes it easier to parse results. 
Also, learning how to make options into passes is a great way along the customizing of optimizations, such as tiling factors... for others. Thanks for learning how to do this

PR #782, opened 2021-07-19
Gemm k unrolled

AlexandreEichenberger:
Unroll the innermost simonized loop of GEMM by I and K loops, resulting in a kernel with 32 load, 16 load splat, 64 FMAs and 0 store. Unroll K by a factor of 4 turned out almost as good as unrolling of 64, prefer to keep the smaller unroll factor to increase applicability.


doru1004 (approved):
LGTM

PR #784, opened 2021-07-20
Use the mlir file utilities for opening input and output files

sstamenova:
1. Rather than directly using memorybuffer to open input files, use the mlir openInputFile function
2. Instead of writing the output to std::err after redirecting std::err to the output file, use the mlir openOutputFile function to open the file and then write directly to it
3. Cleanup the includes in MainUtils.cpp


caoimhinuibrian:

AlexandreEichenberger (approved):
LGTM

PR #788, opened 2021-07-21
Doc readme links

AlexandreEichenberger:
Decided to use the README in the Doc as a repo of links. Instead of replicating the README, just linked to the top one. Every other md file in the doc directory is now accessible by a relative link.

Think this should address our concerns. There is still an issue with the http://onnx.ai/onnx-mlir/ website, one thing at a time.

You may see the file in action here: https://github.com/AlexandreEichenberger/onnx-mlir/tree/doc-readme-links/docs


caoimhinuibrian (approved):
LGTM

AlexandreEichenberger:
Close for now. It turns out that the onnx.ai/onnx-mlir is built from a custom (very old) branch that has nothing in common with master branch, so it appears to be irrelevant for that web site of what the current content of doc is.

We still need to find a way to update that web site. Close this one for now.

Issue #785, opened 2021-07-20
-mtriple, TARGET_TRIPLE & LLVM_HOST_TRIPLE


sstamenova:
llvm keeps track of a couple of triples in cmake for each compilation: `TARGET_TRIPLE` (which is set by using `LLVM_DEFAULT_TARGET_TRIPLE`, for example) & `LLVM_HOST_TRIPLE` (which is set by the llvm cmake based on the host). When TARGET_TRIPLE exists, several of the tools such as `llc` will default to using that triple when invoked and that can be overwritten by passing `-mtriple` to the tool. Since some of the llvm tests only work correctly for some triples, they can either be disabled for other triples/platforms/targets/etc. or they explicitly pass a triple to the tools, so that they can be run whenever that triple is supported.

`onnx-mlir` provides the option to pass `-mtriple` which is then used for each of the tools it invokes (such as `llc`) and when this is not present, defaults to whatever the tools default to by not providing an option. The tests themselves do not specify a triple either, so they always run for the default. (I realize that some of the tests such as the backend tests have the option to specify a test triple in the environment, but they themselves do not do so).

This is all well and good in theory for `onnx-mlir`, but it means that the tests will (often) fail if the llvm compilation is done when specifying a default target triple that is not the host triple. Ideally, the tests will run for the host triple (or whatever the "correct" triple is) regardless of whether the build of llvm specified a default target triple.

This can be done in a couple of different ways:
1) Modify the cmake targets for the tests to specify a default target triple that is expected to pass
2) Modify `onnx-mlir` to default to the `LLVM_HOST_TRIPLE` when `-mtriple` is not specified.

I think 1) is probably the correct approach here, but 2) would also allow the tests to pass and provide a deterministic default.

AlexandreEichenberger:
Currently, when using cmake backend, it eventually goes to test.py which uses a TEST_MTRIPLE and TEST_MCPU to use a default triple. I like your suggestion for 2) better than the current solution that works only for the backend test.

I don't have a preference for 1 vs 2 above. My only suggestion is not to use the `LLVM_HOST_TRIPLE` name because that would imply it also work for clang,... I would suggest `ONNX_MLIR_TRIPLE` and `ONNX_MLIR_MCPU` (which also need to be set). 

I don't have the experience to do much CMAKE engineering, can work on 2 if the consensus is to go along with that.

sstamenova:
[...]

PR #793, opened 2021-07-22
Use the LLVM_HOST_TRIPLE as the default for MainUtils if the triple exists

sstamenova:
This allows every executable that links against MainUtils (such as onnx-mlir, tests, etc.) to default to the host triple. That way if the build of llvm used a different default triple (that is often not supported by onnx mlir), it would not be the default for onnx mlir and the tests (allowing the tests to pass on the platform where onnx mlir was built). If -mtriple is specified, it still takes precedence to the default.


sstamenova:
This addresses #785. One thing that occurred to me was that we could add a cmake variable to control this behavior - something like `ONNX_MLIR_USE_HOST_TRIPLE` or `ONNX_MLIR_USE_DEFAULT_TRIPLE` or `ONNX_MLIR_USE_TRIPLE=host` in order to specify whether the host triple should be used as the default or the llvm default triple should be used for the default or neither.

src/MainUtils.cpp
```diff
@@ -288,9 +288,11 @@ string getTargetCpuOption() {
 
 string getTargetTripleOption() {
   string targetOptions = "";
+  // Comand cannot tolerate extra spaces. Add only when needed.
   if (mtriple != "")
     targetOptions = "--mtriple=" + mtriple;
-  // Comand cannot tolerate extra spaces. Add only when needed.
+  if (mtriple == "" && kHostTriple != "")
```
AlexandreEichenberger, line 294: NIT:
Isn't that just a 'else if (kHostTriple != "")`

sstamenova: Oh, yeah. I had somewhat more complicated logic when I was originally working on options for the change and missed this during the clean up.

caoimhinuibrian:

caoimhinuibrian:
@AlexandreEichenberger looks like a test failure (wrong result) on Z ...

AlexandreEichenberger:
Rebuilding, and if failing again on z390, will try to see if I can reproduce the error locally.

AlexandreEichenberger:
@sstamenova:  in the meanwhile, I wonder if you could not add something in the readme to indicate how to set the triple.

Also, why adding this stuff for the triple and not also for them cpu? In my experience, they are often both needed to compile for the right architecture. Could you look into that as well? That would be great.

sstamenova:
@AlexandreEichenberger : Do you mean adding something about `-mtriple` to the Readme? Or about the host triple that automatically gets set up now?

The triple contains the target, cpu and operating system, so it is very descriptive of what to compile for. I'm not sure what other CPU settings you might be thinking of, can you elaborate?

AlexandreEichenberger:
I meant as a way to help the user: as of now, we have several ways to pass it:
* directly to onnx-mlir using the -mtriple
* your new way, which is either: LLVM_HOST_TRIPLE or ONNX_MLIR_HOST_TRIPLE
* using TEST_MTRIPLE for the test.py...
maybe this is just a few too many ways... I can change the test.py (or feel free to do it) to also use the ONNX_MLIR_HOST_TRIPLE.

sstamenova:
> I meant as a way to help the user: as of now, we have several ways to pass it:
> 
> * directly to onnx-mlir using the -mtriple
> * your new way, which is either: LLVM_HOST_TRIPLE or ONNX_MLIR_HOST_TRIPLE
> * using TEST_MTRIPLE for the test.py...
>   maybe this is just a few too many ways... I can change the test.py (or feel free to do it) to also use the ONNX_MLIR_HOST_TRIPLE.

Ok, that makes sense. I'll add some information to the readme and I'll see about test.py.

I was looking at the options for onnx-mlir and I saw that there is an `mcpu` option as well, but not `march`. `mtriple` specifies both `mcpu` and `march` for clang/opt, but sometimes `mcpu` and `march` are used instead. I couldn't find any place where `mcpu` is specifically needed (perhaps some of the tests?), but it seems to me that maybe we should remove `mcpu` entirely OR add `march` as well. What do you think?

test/backend/test.py
```diff
@@ -60,10 +56,6 @@
     result_dir = tempdir.name+"/"
 print("Test info:")
 print("  temporary results are in dir:"+result_dir)
-if args.mcpu:
```
AlexandreEichenberger, line 63: @sstamenova please remove cpu only if you also added an env var to set the cpu

test/backend/test.py
```diff
@@ -905,10 +897,6 @@ def prepare(cls, model, device='CPU', **kwargs):
         print(name)
         # Command
         command_list = [TEST_DRIVER]
-        if args.mcpu:
```
AlexandreEichenberger, line 908: same here.

MLIR.cmake
```diff
@@ -34,6 +34,29 @@ message(STATUS "BUILD_SHARED_LIBS       : " ${BUILD_SHARED_LIBS})
 set(LLVM_REQUIRES_EH ON)
 message(STATUS "LLVM_REQUIRES_EH        : " ${LLVM_REQUIRES_EH})
 
+# LLVM_HOST_TRIPLE is not properly exported as part of the llvm config. It is actually
+# needed by some of the exported cmake files, so there's work in progress to add it to the
+# LLVM config.
+# If LLVM_HOST_TRIPLE is not available, try to include GetHostTriple.cmake to get the triple.
+# This file is available when MLIR_DIR points to a build (as opposed to an install) directory.
+# When all else fails, default to an empty string which is the old default behavior of onnx-mlir.
+if (DEFINED LLVM_HOST_TRIPLE)
```
AlexandreEichenberger, line 43: In your comment below:

> \>**ONNX_MLIR_DEFAULT_TRIPLE**:STRING
   The default triple for which onnx-mlir will generate code. 

But this does not appear to be the case here. `ONNX_MLIR_HOST_TRIPLE` is reset here regardless of if it was defined before or not. 

My understanding is that this entire section here should be predicated on `ONNX_MLIR_HOST_TRIPLE` being undefined.

AlexandreEichenberger (commented):
@sstamenova This patch seems to "force" the mtriple on Z Linux. At the very least, we should be able to override it in case we want to do some cross compilation. I know of folks needing this. I added a comment above on how to fix this.

Also, in light of the fact that we should be able to change it, the new env var should be named `ONNX_MLIR_TRIPLE` as it may not necessarily be the one of the host.

I think that would still serve your purpose that it be automatically asserted if ONNX_MLIR_HOST_TRIPLE is not asserted to begin with.

Asserting the s390x-ibm-linux caused an error for one test. Will open an issue to initiate the search for a fix to that bug.

AlexandreEichenberger:
#796 was created to follow up on the s390x issue uncovered by this PR.

MLIR.cmake
sstamenova: `ONNX_MLIR_HOST_TRIPLE` is calculated each time cmake is invoked, however, `ONNX_MLIR_DEFAULT_TRIPLE` is not reset if it was defined by the user since it's a cache variable. So if I called cmake with: `-DONNX_MLIR_DEFAULT_TRIPLE=my-triple`, `ONNX_MLIR_DEFAULT_TRIPLE` will have the value of `my-triple` even after `set` is called, because the value I gave it was cached, so `set` will not overwrite it.

sstamenova:
@AlexandreEichenberger : It sounds like we want to keep TEST_MTRIPLE and TEST_MCPU for testing purposes, so I reverted the changes to test.py.

I changed the name of the new cmake variable to `ONNX_MLIR_DEFAULT_TRIPLE` since it will act as the default for `onnx-mlir` unless overwritten with the `-mtriple` option. `ONNX_MLIR_DEFAULT_TRIPLE` is a pure cmake variable in line with how llvm uses cmake variables, so it does not read its value from the environment, should it also check if it's set in the environment?

caoimhinuibrian:

sstamenova:
> #796 was created to follow up on the s390x issue uncovered by this PR.

@AlexandreEichenberger : Do we want to wait until #796 is fixed to make the triple change? Or add a workaround in the tests, so that this particular failure is hidden again?

I'd like to wait to commit the change until we update llvm to a newer commit from main because my change to export the host triple is now in main llvm and I will be able to remove the workaround in mlir.cmake. Hopefully that won't be too long.

sstamenova:
@AlexandreEichenberger : I made the update to this change now that we have access to the triple from llvm. This should be ready to go in unless the failure from #796 still exists.

caoimhinuibrian:

caoimhinuibrian (approved):
LGTM

PR #795, opened 2021-07-23
Fix shape inference for unsqueeze when axes are nagative

### LINKED ISSUE #794: Bug with the Unsqueezed Code
@tungld With the new PR, I got the following error when compiling yolo model.
````
onnx-mlir: /home/chentong/onnx-mlir/src/Dialect/ONNX/ONNXShapeHelper.cpp:1215: mlir::LogicalResult ONNXUnsqueezeOpShapeHelper::Compute(mlir::ONNXUnsqueezeOpAdaptor): Assertion `axis >= 0 && "Invalid axis"' failed.
```

tungld:
resolves #794 


src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2158,35 +2158,44 @@ LogicalResult ONNXUnsqueezeOp::inferShapes(
...
     return emitError("Axes attribute is required");
 
-  SmallVector<int64_t, 4> dims;
-  for (int i = 0, j = 0; i < outRank || j < inRank; ++i) {
-    if (std::find(axes.begin(), axes.end(), i) != axes.end()) {
-      dims.emplace_back(1);
-    } else {
-      dims.emplace_back(operandTy.getShape()[j++]);
+  SmallVector<int64_t, 4> axes;
+  bool hasNegativeAxis = false;
```
AlexandreEichenberger, line 2169: NIT: a better name for this bool is "updateAxis" as technically the axis should not be negative once the rank is added...

AlexandreEichenberger (approved):
LGTM

chentong319 (approved):
LGTM!

src/Dialect/ONNX/ONNXOps.cpp
chentong319: Comment on changing the negative value of axes:
1. It is not part of shape inference
2. What benefit does the change bring about? Since new op might be generated along rewriting, we always have to handle the possible negative value
3. In future, the axes will be a tensor. The IndexExpr to handle the negative value has to be there anyway.
My conclusion: no need to make the change.

AlexandreEichenberger: @chentong319 The way we have handled it before was to have such processing in Onnx Shape helper, which is then used both for shape inference and lowering. That is where we can handle the cases where the axis is known only at runtime.

Since this is to fix a bug, I suggest that we prioritize speed and we can generalize later as we see fit.

PR #798, opened 2021-07-25
Organize options

chentong319:
This PR puts the options common to onnx-mlir and onnx-mlir-opt into one independent file so that they can be used by both onnx-mlir and onnx-mlir-opt.  The special options for onnx-mlir only are still left in MainUtils.cpp as `OnnxMlirOptions` group.
The common options are put into `src/Support/OMOptions.hpp` and `src/Support/OMOptions.cpp` and are added into lib OMSupport. I created one options category as an example and we can have more categories in future.
An options can be defined in OMOptions.* so that it can used in multiple files. However, if it is used only in one file, an options can be defined in that file locally.
I deliberately defined the two options for instrument with different approach for illustration.  Though it requires a little more work, I think it is better to define all the options in OMOptions because conflict or redundancy can be easily identified.
How to define and use the options can be found in docs/options.md.
Other open issues:
* how should the option be uniformly named
* What are the possible groups
* Should we forbid passing option value to a pass and just check the option inside the pass so that a pass in onnx-mlir and onnx-mlir-opt can behave the same way?


caoimhinuibrian (approved):
LGTM

PR #801, opened 2021-07-26
Windows instrumentation support

### LINKED ISSUE #783: Adding "gettimeofday" like timing for Windows
We have added functionality to measure time in two places, and it would be great to have this functionality for Windows as well. 

It is needed in OMInstrument.inc and RunONNXLib.cpp Currently # defined out for Windows, as requested in prior PRs.

There are example code already in the third_party libraries, found by just searching gettimeofday. Or paste in your preferred solution.

@NathanielMcVicar would be great if you could add this functionality, as we have limited development capability on Windows.

Thanks

NathanielMcVicar:
Adds Windows support for instrumentation (time and memory) and changes Linux VMem reporting to a pipe so output can be formatted.

Resolves #783


NathanielMcVicar:
@chentong319 Please let me know how you feel about the error reporting. It is a bit of a one-off, but I don't see any other code to normalize it too. If we are ok with potentially cryptic failures I can just remove it entirely as well. Thanks!

NathanielMcVicar:
Another general question, is it worth adding a second ctest to check the output. This is a bit clunky, but since we can't use lit tests to compile safely across platforms, it would be the best option for avoiding regressions in instrumentation, I think.

NathanielMcVicar:
@chentong319 One final question, would it make more sense to report resident memory, instead of virtual, here?

caoimhinuibrian:

docs/Instrumentation.md
```diff
@@ -30,19 +30,13 @@ For example, a model, mymodel.onnx, is compiled with `Debug/bin/onnx-mlir  --ins
...
-#  5) after  op=    Tanh Time elapsed: 0.115314 accumulated: 0.676507
-156608
-#  6) after  op=     Mul Time elapsed: 0.022779 accumulated: 0.699286
+#  0) after  op= Transpo Time elapsed: 0.000766 accumulated: 0.000766 VMem: 156608
+#  1) after  op= Constan Time elapsed: 0.005398 accumulated: 0.006164 VMem: 156608
+#  2) after  op= Constan Time elapsed: 0.004225 accumulated: 0.010389 VMem: 156608
+#  3) after  op=    Conv Time elapsed: 0.360213 accumulated: 0.370602 VMem: 156608
+#  4) after  op= Softplu Time elapsed: 0.190591 accumulated: 0.561193 VMem: 156608
+#  5) after  op=    Tanh Time elapsed: 0.115314 accumulated: 0.676507 VMem: 156608
+#  6) after  op=     Mul Time elapsed: 0.022779 accumulated: 0.699286 VMem: 156608
```
AlexandreEichenberger, line 39: That is a big improvement. I have an additional request: if by any chances we want to analyze the results, its nice to have an easy format to analyze. Here is what I have used in similar circumstances

```
@instr, op id, 0, after op, Transpo, time elapsed, 0.000766, accumulated, 0.000766, VMem, 156608
```
 namely, an easy and unique pattern to scan for `@inst` and then comma separated fields that make it easy to analyze one or more numerical fields. Still easy to read because of the text, easy to process because of the commas.

src/Runtime/OMInstrument.inc
```diff
@@ -63,29 +72,70 @@ void TimeInit() {
 #endif
 
 #ifdef _WIN32
-void ReportTime() {}
+inline void WinTimerSub(LARGE_INTEGER newTime, LARGE_INTEGER prevTime,
```
AlexandreEichenberger, line 75: There is a similar need for times in the utils/RunONNXLib.cpp. My fault in that I did not try to unify the interface, here is what I used there
```
#ifdef _WIN32
inline void processStartTime() {}
#else
inline void processStartTime() {
  if (!measureExecTime)
    return;
  gettimeofday(&startTime, NULL);
}
#endif

// Process stop time, store result in log.
#ifdef _WIN32
inline void processStopTime() {}
#else
inline void processStopTime() {
  if (!measureExecTime)
    return;
  gettimeofday(&stopTime, NULL);
  timersub(&stopTime, &startTime, &result);
  uint64_t time =
      (uint64_t)result.tv_sec * 1000000ull + (uint64_t)result.tv_usec;
  timeLogInMicroSec.emplace_back(time);
}
#endif
```
 Maybe I could try to unify the interface here and in that file, so that we may simply reuse the time between the two places. It looks like the commonality could be expressed as a simple function whose template is this

```
unsigned long long timer_interval_since_last() {
  static bool first = true;
  // pseudo code
  static XXX lastTime;
  static XXX currTime;
  if (first) {
    first = false;
    lastTime = XXX_time()
    return;
  }
 currTme = XXX_time()
 intervalInMicroSec = XXX_time_diff(lastTime, currTime);
 lastTime = currTime;
 return intervalInMicroSec;
}
```

We could then use this in both places. It simply restrict us to a single timer... probably fine, otherwise we can bring an abstract struct to handle it too.

AlexandreEichenberger:
>Another general question, is it worth adding a second ctest to check the output. This is a bit clunky, but since we can't use lit tests to compile safely across platforms, it would be the best option for avoiding regressions in instrumentation, I think.

I think it is. MLIR has an infrastructure where they test output after running mlir.

chentong319:
> Another general question, is it worth adding a second ctest to check the output. This is a bit clunky, but since we can't use lit tests to compile safely across platforms, it would be the best option for avoiding regressions in instrumentation, I think.

I am looking to enabling instrument in backend test so that the runtime can be tested.

chentong319:
We definitely want to unify the utility used in instrument and RunONNXLib.cpp after this patch.

AlexandreEichenberger:
> We definitely want to unify the utility used in instrument and RunONNXLib.cpp after this patch.

It has to be in C because the RT has to work in both, or can we add a C++ class? Class would be the easiest to encapsulate data and code.

src/Runtime/OMInstrument.inc
```diff
@@ -63,29 +72,70 @@ void TimeInit() {
...
+}
 #else
 void ReportMemory() {
   char memCommand[200];
+  char memOutput[200];
+  FILE* memPipe;
   mypid = getpid();
   sprintf(memCommand, "ps -o vsz --noheader -p %d", mypid);
-  system(memCommand);
+  memPipe = popen(memCommand, "r");
```
chentong319, line 119: This change is excellent!

chentong319 (approved):
LGTM!

chentong319:
For the output format, it is not human friendly to always separate with comma. As long as there is a fixed format, I use shell script to parse the output.  My view is that we will print out for human when VERBOSE is turned on, while the utility should return/print the result for program to handle.

NathanielMcVicar:
> For the output format, it is not human friendly to always separate with comma. As long as there is a fixed format, I use shell script to parse the output. My view is that we will print out for human when VERBOSE is turned on, while the utility should return/print the result for program to handle.

Because the printing and instrumentation itself are comingled, we don't really have the option of different levels of output here. One idea would be to leave the instrumentation itself in C, but create two cross-platform wrappers (C and CPP) that can also handle the different printing requirements of this and RunONNXLib. If so, I would suggest we do that in a subsequent PR.

NathanielMcVicar:
> I am looking to enabling instrument in backend test so that the runtime can be tested.

@chentong319 Does that mean I should hold off on adding additional tests here for now?

chentong319:
@NathanielMcVicar To my understand, the test you added is to test compile and link. That's necessary. What I will do is to enable instrument when the backend test is executed. It may not easy to check the output of instrument in the backend test framework. At least, the output tensor will be correct.

NathanielMcVicar:
I will add an additional ctest then, to check the output. This will make sure the instrumentation doesn't break functionally but continue to compile. After that, I think we can move formatting details and the combination of the two time-keeping frameworks to a subsequent PR.

AlexandreEichenberger:
@NathanielMcVicar let us know explicitly when you believe the PR is ready to merge using the "ready to merge" label, or a message here. We will then do a quick final review and merge.

NathanielMcVicar:
Let's merge it now then, I think adding a functional test is going to have to be a subsequent change. I tried a few things, but it actually requires a lot of very clunky cmake to redirect output, or adding support to the test cpp file itself to redirect the output. My new idea, which will require bigger changes to the code base and break many outstanding PRs, is to move the mlir lit tests to one level down the directory hierarchy and allow for other kinds of lit tests, since lit is much more powerful (and that seems like the correct direction to go in general to me). But I'm sure that will be controversial so the discussion shouldn't be tied to this change.

AlexandreEichenberger:
@chentong319 you are the last one that commented on this one. If you are fine, please merge the PR. Thanks

PR #802, opened 2021-07-26
Conv and pool cannot handle dynamic shapes with SAME_* padding policy

AlexandreEichenberger:
While looking at understanding the conv code again, I discovered that we don't handle properly the cases where the spacial dimensions are runtime with the SAME_UPPER or SAME_LOWER, as we issue "compile time computations that simply take the -1 (dynamic dimensions) as the actual size during shape inference.

While we can later extend it to handle these padding case properly, it seems that this scheme may not be used too often. Can add if the need arises.

So I added error messages to properly flag the cases that we don't handle (instead of generating erroneous code) and disabled the runtime dimensions on the spacial dimensions when SAME_* policy is used in our backend test.

There is some minor infrastructure changes for handling multiple code generation schemes for CONV, with no functionality change at this time.


tungld (approved):
LGTM!

Just would like to confirm whether you want to check the expected failure for Conv pads using FileCheck? E.g.
```mlir
// RUN: onnx-mlir-opt --shape-inference %s -split-input-file -verify-diagnostics

...
// expected-error @+1 {{Conv Pads defined as SAME_UPPER or SAME_LOWER requires compile time X sizes}}
...
```

More info about `-verify-diagnostics`: https://clang.llvm.org/doxygen/classclang_1_1VerifyDiagnosticConsumer.html

I know we haven't done this before in onnx-mlir. Perhaps, another PR for such a failure check so that we know explicitly which op configurations are not supported.

AlexandreEichenberger:
@tungld good suggestion, we have indeed not made test to detect error messages. Something to pick up in a later PR.

AlexandreEichenberger:
@NathanielMcVicar @sstamenova Windows CI has not completed in more than 15h. Should I restart the process, is the machine unhappy? Tx

chentong319:
I think we'd better move all the shape related code into the ONNXShapeHelper.cpp. If this case can will be handled later, the code should be in ONNXShapeHelper.cpp, far from this assertion. In ShapeInference, we just check whether the output IndexExpr has value or not. 
By the way, when I went over ONNXOp.cpp, I found this line  questionable. 
https://github.com/AlexandreEichenberger/onnx-mlir/blob/f0ce9b444f66559146032af0b44c4941f2c784d0/src/Dialect/ONNX/ONNXOps.cpp#L1626
In the ONNXOps.td.inc, the group is a DefaultValueAttr. I do not think we need to give default value in the code. I guess it was an obsolete code. And different default value?

NathanielMcVicar:
> Should I restart the process, is the machine unhappy?
It appears to be, I would restart and then we can check the log of the failure (I can't access currently).

AlexandreEichenberger:
> I think we'd better move all the shape related code into the ONNXShapeHelper.cpp. If this case will be handled later, the code should be in ONNXShapeHelper.cpp, far from this assertion. In ShapeInference, we just check whether the output IndexExpr has value or not.

I agree with you, this is indeed very old and this functionality should be moved into shape-inference. Unfortunately, the same code is reused by many operations (other ops also using padding), so it is not super easy.

At the present time, I am trying to rework the lowering to Krnl. I may take that task here next.

AlexandreEichenberger:
>By the way, when I went over ONNXOp.cpp, I found this line questionable.
https://github.com/AlexandreEichenberger/onnx-mlir/blob/f0ce9b444f66559146032af0b44c4941f2c784d0/src/Dialect/ONNX/ONNXOps.cpp#L1626
In the ONNXOps.td.inc, the group is a DefaultValueAttr. I do not think we need to give default value in the code. I guess it was an obsolete code. And different default value?

I think the code is ok. Its basically saying "if the attribute is default, set it to default so that we don't have to construct an attribute each time". It would work without it too.

PR #804, opened 2021-07-28
Prepare to upgrade squeeze to OpSet 13

chentong319:
The version 11 and 13 for squeeze/unsqueeze are not compatible. This PR creates two versions for them and changes the existing code for v11 to use ONNXSqueezeV11Op. Now ONNXSqueezeOp is for v13, the corresponding code, such as shape inference, lowering, and rewrite, for v13 is not there yet. @NathanielMcVicar may put his local fix for upset 13 implementation in on top of this PR.

Details in changes:
1. utils/gen_onnx_mlir.py: use opName, which is a combination of schema.name and version, to provide fine control in ONNXOp.td.inc generation.
2. All the usage of ONNXSqueezeOP are changed to ONNXSqueezeV11Op. In fact, Squeeze/Unqueeze are used a lot in rewriting and lowering.


AlexandreEichenberger (approved):
LGTM, still would prefer `ONNXUnsqueezeOp11` instead of `ONNXUnsqueezeV11Op` because I think it read so much better. I recall you liked the current version much better, so go for it unless you changed your mind.

PR #805, opened 2021-07-29
Test expected errors for shape inference

tungld:
This patch adds tests for expected errors encountered during shape inference. The tests are mostly for unsupported op settings.

It is recommended to add tests for unsupported settings when implementing an op, so that we can quickly check them and keep them under control.


src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2031,7 +2031,7 @@ LogicalResult ONNXMaxPoolSingleOutOp::inferShapes(
   // Storage order.
   auto storageOrder = storage_order();
   if (storageOrder != 0)
-    return emitError("column major storage order not supported at this time");
+    return emitError("column major storage order not implemented yet");
```
chentong319, line 2034: Column

chentong319 (approved):
LGTM except one small neglect.

src/Dialect/ONNX/ONNXOps.cpp
tungld: Fixed this. Thanks!

PR #810, opened 2021-07-30
readme to be kept in sync

AlexandreEichenberger:
looks like on window, the softlink is transformed in something else. So now have 2 distinct files, with a big warning to keep them in sync.

If someone tell me how to do the following, happy to add it.

`diff README.md docs/README.md` result in an empty line in the makefile... that would be great. I don't know how to do that in a cmake, sorry. 

Ideally that command would be added to the `make check-doc` command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,1 +1,373 @@
-docs/README.md
+<!--- SPDX-License-Identifier: Apache-2.0 -->
+
+<!--- 
+###########################################################
+###########################################################
+###########################################################
+
+KEEP THE TOP DIR README.md and docs/README.md IN SYNC.
+Top readme: used for github users.
+Docs readme: used as template for onnx.ai/onnx-mlir website.
+
+###########################################################
+###########################################################
+###########################################################
+--->
+
+# ONNX MLIR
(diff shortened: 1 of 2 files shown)


tungld:
I seems to me using symlink is better. The problem was perhaps because of Windows environment which somehow used  `core.symlinks = false`, so the symlink became a real file containing the link instead of a link when cloning.

I found some useful discussions: 
- Symlink issue on Windows: https://stackoverflow.com/questions/58587005/why-does-git-clone-override-global-core-symlinks-on-windows => looks like we can avoid by enabling Developer Mode.
- How does git handle symlink: https://newbedev.com/how-does-git-handle-symbolic-links

Hope it helps.

AlexandreEichenberger:
I just doubt that we will get every window user to enable the right mode. Is there a way to have a makefile check the status of a file? If so, we can fail a CI if there is no simlink?

tungld:
If we have to duplicate README.md, is it possible to use DocCheck to make sure they are in sync? We used it to check the .md file for ONNX ops.

tungld (approved):
LGTM!

I created a PR #811 for CI to check whether they are in sync or not.

PR #811, opened 2021-08-02
Check whether README.md is up-to-date

tungld:
This patch adds a check to Jenkins to make sure that `README.md` files in the top level and `docs` folder are in sync. 

This is how the check failed if the two files were different: https://yktpandb.watson.ibm.com/jenkins/job/ONNX-MLIR-Pipeline-Docker-Build/1627/console


utils/check_doc_readme.sh
```diff
@@ -0,0 +1,22 @@
...
+
+ROOT_DIR=$1
+DOC_DIR=$1/docs
+
+README="README.md"
+
+echo "Checking whether docs/README.md is up-to-date or not ..."
+
+diff ${ROOT_DIR}/${README} ${DOC_DIR}/${REAMDE} > readme.patch
+
```
chentong319, line 11: Just curious how about using the following script without the temporary file readme.patch:
```
diff ${ROOT_DIR}/${README} ${DOC_DIR}/${REAMDE}
if [ $? == 0 ] 
then
...
else
...
fi

gongsu832:
There is no need for the check. `docs/README.md` is for generating the web page for http://onnx.ai/onnx-mlir and really should have its own contents. @tjingrant and I were talking about this. `docs/README.md` should really be called `docs/ONNXAI.md` to avoid the confusion that these two need be the same.

utils/check_doc_readme.sh
tungld: It looks working well. Thank you for the suggestion! I did not know about this.

tungld:
@gongsu832 it makes a lot of sense. ONNXAI.md is independent of README.md. Given that the current README.md contains useful information about installation, do we want to put it somewhere in http://onnx.ai/onnx-mlir or we expect that users will find the information in the GitHub repos?

gongsu832:
@tungld Once we rename `docs/README.md` to `docs/ONNXAI.md`, we also need to update some script that generates the webpage. I'm talking to Tian to find out where the script is. As for the content of `ONNXAI.md`, I think it should be just some high level overview of what onnx-mlir is and then just points users to the github for installation instructions.

tungld:
@gongsu832 thanks for the information! I will close this.

PR #814, opened 2021-08-05
Update LLVM commit to 829616c

imaihal:
This PR updates LLVM commit. Current commit is `829616c24119f000132913616a7cd4ad930ca84a`.

- This PR reflected following changes in LLVM.
  -  Change LLVM::DialectCastOp to UnrealizedConversionCastOp. ([D105880](https://reviews.llvm.org/D105880))
  -  ~~Change llvmVoidTy to TypeRange() to meet result type verification in llvm.call [D106937](https://reviews.llvm.org/D106937)~~ ==> This is not needed after merging PR #812.
  -  Use math-to-llvm newly introduced in [D105702](https://reviews.llvm.org/D105702)
  -  Fix type mismatch after verifier is added in [D106398](https://reviews.llvm.org/D106398)
  -  Avoid cyclic dependency in shared library build
     - After [D105903](https://reviews.llvm.org/D105903) in LLVM, shared library build failed in link time. This was because  [D105903](https://reviews.llvm.org/D105903) forces the compiler to emit a strong definition for a class. To avoid this, `OMKrnlOps` needed to be added in LINK_LIBS of `OMSimplifyKrnl`. However, this introduces cyclic dependency(`OMSimplifyKrnl` -> `OMKrnlOps` -> `OMSimplifyKrnl` ). To avoid the cyclic dependency, `OMSimplifyKrnl` were merged into `OMKrnlOps`.


caoimhinuibrian (commented):
@gongsu832 can you take a look at this please?

gongsu832:
Something is wrong with the compilation. All the backend tests failed with something like this:
```
----------------------------- Captured stdout call -----------------------------
test_abs
Failed /workdir/onnx-mlir/build/Debug/bin/onnx-mlir: test_abs
error: 'llvm.insertvalue' op Type mismatch: cannot insert '!llvm.ptr<i64>' into '!llvm.struct<(ptr<f32>, ptr<f32>, i64, array<3 x i64>, array<3 x i64>)>'
```

imaihal:
@gongsu832 

> error: 'llvm.insertvalue' op Type mismatch: cannot insert '!llvm.ptr<i64>' into '!llvm.struct<(ptr<f32>, ptr<f32>, i64, array<3 x i64>, array<3 x i64>)>'

Backend test for vgg19 still get an error, but this was fixed by [ed49bad](https://github.com/onnx/onnx-mlir/pull/814/commits/ed49bad905ac8070efe392d7b67cf464ab55644c). Thanks!

imaihal:
Currently this PR got errors in `test_vgg19_cpu` (ci/circleci build) and MLIR installation on Windows (MLIR-Windows-CI). 

@gongsu832 I can't reproduce the error about `test_vgg19_cpu` on my local environment. Can you get more info (error messages etc.) on cicleci or your environment?

gongsu832:
@imaihal We don't care about circleci since it's going away eventually. It's being broken for a long time now because, if I remember correctly, compiling vgg19 fails due to increased memory requirement after we made some changes to handle constants.

imaihal:
@gongsu832 The rest of error we need to solve is in MLIR-Windows-CI. It seems it fails in MLIR installation on Windows. I'm not sure MLIR on Windows itself has problem or configuration in Windows-CI has problem. Do you know anyone who can help with this problem?

gongsu832:
@sstamenova should be the person who can help.

imaihal:
Hi, @sstamenova , I'm trying to update MLIR commit in this PR, but it seems Windows CI fails because of MLIR installation failure. One of the lit test `IR/invalid-unregistered.mlir` fails. Can you help with this issue?

sstamenova:
@imaihal Did the Windows CI fail more than once?

It looks like the same commit passed on the Windows mlir buildbot (https://lab.llvm.org/buildbot/#/builders/13/builds/10886) and the error emitted by the test is expected, but was not recognized as such. I would be curious to know whether a re-run passes and I will try building the same commit with the same settings locally to see if I can reproduce it since the Windows mlir buildbot uses slightly different options than the onnx-mlir version.

sstamenova:
I can confirm that this fails locally for me. I am not sure what the difference is between the buildbots and the local builds (other than using RTTI which is required for onnx-mlir), so I am investigating. I believe the test will require a fix in the llvm-project repository.

sstamenova:
@imaihal @gongsu832 : I spent some more time trying to figure out why I see the failure locally and it happens in the Windows CI but not on the buildbots and it looks like building directly with Visual Studio (as opposed to using ninja as the generator, but using the Visual Studio build tools) reproduces the failure, but ninja does not.

There are a couple of options here: wait until there's a fix in the main branch of llvm-project and then merge, or change how the Windows CI for onnx-mlir builds, so that it also uses ninja for llvm (which means ninja should be used for onnx-mlir as well). I can help with the change to the Windows CI if you choose to go that route.

imaihal:
@sstamenova Thanks for the investigation!
I think it's better for the Windows CI for onnx-mlir and the buildbots to show the same behavior to avoid future installation failure. However,  I think this is not the case if there are special requirements on Windows CI for onnx-mlir.
Do we need a big fix to use ninja? 　

@gongsu832 @caoimhinuibrian Do you have any thought on this?

s390xlinux:
onnx-mlir build might move to ninja in the future if necessary but this should not prevent the Windows CI from moving to ninja now. It's certainly desirable to have all platforms behave as similarly as possible but I'd expect there will always be cases where we need special case for one platform or another.

sstamenova:
I think I could have the Windows CI (which includes the mlir AND onnx-mlir builds) both moved to ninja some time this week. Note that we could mix build generators across the mlir and onnx-mlir builds but it usually creates more problems, so they should both move to using ninja as the generator.

s390xlinux:

sstamenova:
@imaihal @gongsu832 : I've started a PR here to switch the Windows CI to ninja: https://github.com/onnx/onnx-mlir/pull/827/files

imaihal:
@sstamenova Now the Window CI passed. Thanks for your help!

gongsu832 (approved):
LGTM

PR #819, opened 2021-08-11
Add the data_layout attribute to ModuleOp to denote endianness

### LINKED ISSUE #816: Wrong output for a simple function that returns a constant - happen on ZLinux
While checking the PR #815 I found that a simple program failed on ZLinux:

A test program that just prints out a constant tensor:
```mlir
module  {
  func @main_graph() -> tensor<2xf32> {
    %0 = "onnx.Constant"() {value = dense<[3.000000e+00, 2.000000e+00]> : tensor<2xf32>} : () -> tensor<2xf32>
    return %0 : tensor<2xf32>
  }
  "onnx.EntryPoint"() {func = @main_graph, numInputs = 0 : i32, numOutputs = 1 : i32, signature = "[ \0A]\00@[   { \22type\22 : \22f32\22 , \22dims\22 :
 [2] , \22name\22 : \22z\22 }\0A\0A]\00"} : () -> ()
}
```
Compile the code using `onnx-mlir` and run using Python or C, output is `[2., 3.]` (it should be `[3., 2.]`).

Checking the constant in the generated LLVM IR, it is correctly generated:
```
llvm.mlir.global internal constant @constant_0(dense<[3.000000e+00, 2.000000e+00]> : tensor<2xf32>) : !llvm.array<2 x f32>
```

However, using `-O0` instead of `-O3` in the llvm `opt` command produces the correct result.

tungld:
Resolves #816

Since LLVM optimizations use module's DataLayout to know whether constant data is in little or big endian. We should annotate ModuleOp with correct endianness so that LLVM optimizations correctly read constant data.

This patch adds `llvm.data_layout` to ModuleOp whose value is `e` or `E` for little or big endian, respectively.

By this patch, all constant-related end-to-end tests, except ones our builder does not support, are passed.

### DIFF
--- a/src/Conversion/KrnlToLLVM/KrnlToLLVM.cpp
+++ b/src/Conversion/KrnlToLLVM/KrnlToLLVM.cpp
@@ -31,6 +31,7 @@
 #include "mlir/Target/LLVMIR/ModuleTranslation.h"
 #include "mlir/Transforms/DialectConversion.h"
 #include "llvm/ADT/Sequence.h"
+#include "llvm/Support/Endian.h"
 
 #include "onnx/onnx_pb.h"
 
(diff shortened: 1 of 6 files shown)


AlexandreEichenberger (approved):
LGTM, See that you added the flags for the KRNL to LLVM pass. Do you mind just confirming that the same thing is not needed for the earlier passes from KRNL to affine, for example.

tungld:
> Do you mind just confirming that the same thing is not needed for the earlier passes from KRNL to affine, for example.

It's not needed when we are inside MLIR since MLIR Attribute are used for constants and we can get constant values correctly from Attribute. However, LLVMIR is an exit point of MLIR, we should annotate it with endianness so that external tools can correctly interpret constant values.

PR #821, opened 2021-08-12
Choose the correct version of op

chentong319:
There is an error in previous code that support multiple versions of an op. The error is exposed for yolo model after multiple version of Squeeze and Unsqueezed was added.
There are two versions, 13 and 11 for Squeeze. The yolo model have version 12. We should choose 11. We need the new version only 13 and higher. 
It's a small fix and the document is updated.


AlexandreEichenberger (approved):
LGTM

NathanielMcVicar:
Is this fix required before v13 unsqueeze?

chentong319:
> Is this fix required before v13 unsqueeze?

no, unless your model's op_set is 12.

AlexandreEichenberger (approved):
LGTM

PR #822, opened 2021-08-12
Control memory optimization passes

chentong319:
Our own memory optimization passes, such as memory pooling, bundling, and optimization, sometimes caused significant slowdown in compile time. Alternatively, we can use buffer management passes provided by MLIR. This PR introduces a command line option, `--memory-bundling-enabled`, to control the behavior of onnx-mlir. By default, the option is true and onnx-mlir works the same way as previously. If bundling is disabled, BufferDeallocate pass from MLIR will be invoked and our own memory related passes will not be used.
One detailed change to use BufferDeallocate pass: the memref.allocate will not be moved and deallocate will not be generated when onnx is lowered to krnl. 
I checked output of EmitMILR and found the allocate and deallocate seems to be at the right place.  The instrument output also showed different memory footprint for yolo (much smaller):
```
#  0) after  op= Transpo VMem:285116
#  1) after  op= Constan VMem:285116
#  2) after  op= Constan VMem:285116
#  3) after  op=    Conv VMem:304720
#  4) after  op= Softplu VMem:326356
#  5) after  op=    Tanh VMem:326356
#  6) after  op=     Mul VMem:304796
#  7) after  op= Constan VMem:304796
#  8) after  op= Constan VMem:304796
#  9) after  op=    Conv VMem:315612
# 10) after  op= Softplu VMem:315612
# 11) after  op=    Tanh VMem:326428
# 12) after  op=     Mul VMem:326428
```  
Future work:
1. More buffer passes may be added, for example, promote to alloca.
2. Check the ops with control flow.


tungld:
@chentong319 just thinking about a slightly different way. 

In this patch, you switch between MLIR BufferDeallocate and our onnx-mlir memory pool (OM memory pool) passes.
If MLIR BufferDeallocate works well even for dynamic memref, I think we can do this:
- No need to emit `memref.deallocate` at all in onnx-to-krnl pass
- Always call MLIR BufferDeallocate pass
- Then call OM memory pool passes if enabled.

It means, in onnx-to-krnl pass, we only need to emit `memref.allocate` which I hope can be automated too once we have ONNXShapeHelper for all lowering ops, e.g. just calling `insertAllocationSimple(ArrayRef<IndexExpr>, ....)`

chentong319:
@tungld This PR does what you described when onnx-mlir is invoked with `--memory-bundling-enabled=false`. I did not call insertAllocationSimple but modified insertAllocationAndDeallocation to not move memref.allocate and not generate memref.deallocate.

tungld:
Yes, they are same when `--memory-bundling-enabled=false` but different when `--memory-bundling-enabled=true`. 

I mean even when `--memory-bundling-enabled=true` we still call MLIR BufferDeallocation before OM memory pool passes, so that `memref.deallocate` is automatically emitted instead of manually emitted inside onnx-to-krnl. Thus, the code to emit `memref.deallocate` inside onnx-to-krnl can be totally removed.

chentong319:
@tungld Yes, that can be the next step. Currently I want to keep the existing memory bundling part untouched and try the MLIR pass as an independent path.

src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
```diff
@@ -78,6 +79,12 @@ Value insertAllocAndDealloc(MemRefType type, Location loc,
 
   // Make sure to allocate at the beginning of the block if
   // all dimensions are known.
+  // Don't do these if memory bundling is not enabled
+  // Allow buffer management in MLIR to do the job
+
+  if (!memoryBundlingEnabled)
```
tungld, line 85: I am a bit concerned about the use of an option of the `onnx-mlir` command directly inside a pass, in this case, onnx-to-krnl pass. I think that onnx-to-krnl pass should be self-contained and not depend on the `onnx-mlir` command, which is easier for debugging.

How about adding a local option, e.g. `no-deallocation`, to the `onnx-to-krnl` pass in order to enable/disable deallocation? and users of `onnx-to-krnl` can set the option if they want, e.g., 
- with `onnx-mlir-opt`: `onnx-mlir-opt --convert-onnx-to-krnl='no-deallocation'`
- with `onnx-mlir`: `mlir::createLowerToKrnlPass(/*no-deallocation=*/ true)`

chentong319: @Tungld I discussed my thinking of options in previous PR or issue about command line options for instrument. I did not find an easy way to define option with field Option in definition of pass, Similar situation to propagate options to a pass: either we have initialize the options explicitly or the default value (not the option) has to be set. With the default value set, I did not experiment how onnx-mlir-opt to use that.  Did you ever try that?
In my view, our pass may depend on some shared context info, such as target triple, and optimization/debugging level, to get consistent logic across passes. I am not too worried about using option inside a pass. On the contrary, I feel it is a better way to make onnx-mlir and onnx-mlir-opt work in the same way. The drawback is that it makes the pass in onnx mlir compiler less reusable.  I think that's the concern from MLIR, which provides basic passes. We can discuss the topic more and it may take some times for our compiler to evolve.

tungld: I just checked the use of an option in `onnx-mlir-opt`: https://github.com/tungld/onnx-mlir/pull/1

`onnx-mlir-opt --convert-onnx-to-krnl` will insert deallocation as normal, and `onnx-mlir-opt --convert-onnx-to-krnl='no-dealloc'` will not insert deallocation.

chentong319: @tungld Thanks for trying this. Can I use the no-dealloc option for onnx-mlir or other way to turn on/off no-dealloc?
By the way, when I tried option of instrument, I found I can not use the enum OptionList for onnx-mlir-opt, while I can for onnx-mlir. Also I had difficult in giving multiple options. The document about option in pass management is not enough. Anything else?

tungld: > Can I use the no-dealloc option for onnx-mlir or other way to turn on/off no-dealloc?

Unfortunately, we cannot use the option directly. But, we can set it via constructor. e.g. (I have not really tested this)
```
FrontendToKrnlLoweringPass(bool noDealloc) { this->noDealloc = noDealloc}
```
and then
```
std::unique_ptr<Pass> mlir::createLowerToKrnlPass(bool noDealloc) {
  return std::make_unique<FrontendToKrnlLoweringPass>(noDealloc);
}
```

The thing is we should keep the default constructor (with no arguments) to make options work. It looks like a limitation.

> When I tried option of instrument, I found I can not use the enum OptionList for onnx-mlir-opt, while I can for onnx-mlir. Also I had difficult in giving multiple options. The document about option in pass management is not enough. Anything else?
By the way, I will be on vacation next week.

I am sorry I haven't yet used OptionList. Have a nice vacation :) !

tungld:
Updated the patch so that `onnx-to-krnl` pass has its own options, which makes the pass self-contained.

tungld (approved):
LGTM

AlexandreEichenberger:
@tungld thanks, let's merge it as soon as it passes the test.

AlexandreEichenberger:
@tungld Running `onnx-mlir --help` I see this
```
  --memory-bundling-enabled                         - enable memory bundling related optimizationseveral passes work together to bundle buffersif disabled, buffer management by mlir is used
```

Is it implying that memory bundling is by default off, and it can be turned on by setting this flag? Thanks for clarifying.

tungld:
Tong sets it to true by default so that there is no change to the current onnx-mlir. It seems the help does not display the default value. 

We turn on/off by `--memory-bundling-enabled=true/false`.

PR #825, opened 2021-08-17
Update methods for deallocation insertion when using `ReinterpretCastOp`

imaihal:
This PR updates methods for deallocation insertion when using `ReinterpretCastOp`. 

`ReinterpretCastOp` just changes the view of input memref. So, the input memref should not be deallocated if it is returned value in a function. However,  current method for determining deallocation (`checkInsertDealloc()` and `checkOpResultIsReturned()`) ignores this case.

- An example result of current deallocation (`checkInsertDealloc()`)
%0 and %1 is the same memory address, but %0 is deallocated before returned. This PR avoids to deallocate %0.
```
%0 = memref.alloc() : memref<1x10xf32>
  :
%1  = memref.reinterpret_cast %0 to offset: [0], sizes: [2, 5], strides: [5, 1] : memref<1x10xf32> to memref<2x5xf32>
  :
memref.dealloc %0 : memref<1x10xf32>
return %1 : memref<2x5xf32>
```

This PR checks if the result value of reinterpret_cast op is returned in the function. If it does then dealloc for input memref of reinterpret_cast op is not inserted. `ONNXReshapeOp`,  `ONNXSqueezeV11Op`, and `ONNXUnsqueezeV11Op` are lowered to `ReinterpretCastOp`. So, the result values of them are also checked.


src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
```diff
@@ -139,21 +139,49 @@ Value insertAllocAndDeallocSimple(PatternRewriter &rewriter, Operation *op,
...
+// current op or the result value of reinterpret_cast op whose
+// operand is the result value of current op. If it does then
+// dealloc should not be inserted.
 bool checkInsertDealloc(Operation *currentOp, int resultIndex) {
   auto parentBlock = currentOp->getBlock();
-
   bool insertDealloc = true;
-  parentBlock->walk([&insertDealloc, currentOp, resultIndex](ReturnOp op) {
-    // If there is at least one result to investigate.
-    if (currentOp->getNumResults() > 0) {
```
AlexandreEichenberger, line 150: Commenting here in the old code, since this code was used to generate the new version.

Isn't this test `currentOp->getNumResults() > 0` invariant to the walk? Meaning the currentOp is constant. If so, why is it not around the entire walk...
```
if (currentOp->getNumResults() > 0) {
  parentBlock->Walk(...)
}
```

AlexandreEichenberger: Second comment: Isn't it true that some operations can have multiple results? Is it enough to focus only on the first result here? Again, I understand that you probably did not write this code originally, but reading it now make me wonder if it is correct.

AlexandreEichenberger (commented):
In your example:

```
%0 = memref.alloc() : memref<1x10xf32>
  :
%1  = memref.reinterpret_cast %0 to offset: [0], sizes: [2, 5], strides: [5, 1] : memref<1x10xf32> to memref<2x5xf32>
  :
memref.dealloc %0 : memref<1x10xf32>
return %1 : memref<2x5xf32>
```

It would appear that deleting %0 is correct since %1 has now the info about %0. Is it because some later optimization will remove totally %1 that we don't want to delete %0? It is not clear to me. 

Your PR would be a lot easier to understand if you had also the result of your transformation, namely what do you want the above code to look like.

Also, maybe a bigger question is how does our memory management interface with MLIR? I think I remember that @chentong319 tried to have MLIR automatically insert the delete. If that were the case, maybe we should just not free the values? Or not doing so would interfere with @doru1004 's memory management? I think it is worthwhile thinking a bit about these issues, especially in light of the questions about the validity of the original code.

imaihal:
@AlexandreEichenberger Thanks for your comments!

I explain my understanding using full example.
- input (test.mlir)
```
func private @test_unsqueeze_dealloc(%arg0 : tensor<10x20xf32>) -> tensor<*xf32> {
  %0 = "onnx.Transpose"(%arg0) : (tensor<10x20xf32>) -> tensor<*xf32>
  %1 = "onnx.UnsqueezeV11"(%0) {axes=[0,3]} : (tensor<*xf32>) -> tensor<*xf32>
  "std.return"(%1) : (tensor<*xf32>) -> ()
}
```
Without this PR, this mlir is lowered as follows by `onnx-mlir-opt --shape-inference --convert-onnx-to-krnl test.mlir`. 

```
module  {
  func private @test_unsqueeze_dealloc(%arg0: memref<10x20xf32>) -> memref<1x20x10x1xf32> {
    %0 = memref.alloc() : memref<20x10xf32>
    %c20 = constant 20 : index
    %c10 = constant 10 : index
    %1:2 = krnl.define_loops 2
    krnl.iterate(%1#0, %1#1) with (%1#0 -> %arg1 = 0 to 10, %1#1 -> %arg2 = 0 to 20) {
      %3 = krnl.load %arg0[%arg1, %arg2] : memref<10x20xf32>
      krnl.store %3, %0[%arg2, %arg1] : memref<20x10xf32>
    }
    %c1 = constant 1 : index
    %c20_0 = constant 20 : index
    %c10_1 = constant 10 : index
    %c1_2 = constant 1 : index
    %c1_3 = constant 1 : index
    %c1_4 = constant 1 : index
    %c10_5 = constant 10 : index
    %c200 = constant 200 : index
    %2 = memref.reinterpret_cast %0 to offset: [0], sizes: [1, 20, 10, 1], strides: [200, 10, 1, 1] : memref<20x10xf32> to memref<1x20x10x1xf32>
    memref.dealloc %0 : memref<20x10xf32>
    return %2 : memref<1x20x10x1xf32>
  }
}
```
The `reinterpret_cast` just changes the view of `%0`. So in my understanding, `%2` and`%0` point the same memory allocated by `%0 = memref.alloc() : memref<20x10xf32>` . So, if %0 is deallocated, %2 is also deallocated. Is this not correct?

My expected results is as follows. `memref.dealloc %0 : memref<20x10xf32>` is not inserted 
```
module  {
  func private @test_unsqueeze_dealloc(%arg0: memref<10x20xf32>) -> memref<1x20x10x1xf32> {
    %0 = memref.alloc() : memref<20x10xf32>
    %c20 = constant 20 : index
    %c10 = constant 10 : index
    %1:2 = krnl.define_loops 2
    krnl.iterate(%1#0, %1#1) with (%1#0 -> %arg1 = 0 to 10, %1#1 -> %arg2 = 0 to 20) {
      %3 = krnl.load %arg0[%arg1, %arg2] : memref<10x20xf32>
      krnl.store %3, %0[%arg2, %arg1] : memref<20x10xf32>
    }
    %c1 = constant 1 : index
    %c20_0 = constant 20 : index
    %c10_1 = constant 10 : index
    %c1_2 = constant 1 : index
    %c1_3 = constant 1 : index
    %c1_4 = constant 1 : index
    %c10_5 = constant 10 : index
    %c200 = constant 200 : index
    %2 = memref.reinterpret_cast %0 to offset: [0], sizes: [1, 20, 10, 1], strides: [200, 10, 1, 1] : memref<20x10xf32> to memref<1x20x10x1xf32>
    return %2 : memref<1x20x10x1xf32>
  }
}
```

> how does our memory management interface with MLIR? I think I remember that @chentong319 tried to have MLIR automatically insert the delete. If that were the case, maybe we should just not free the values? Or not doing so would interfere with @doru1004 's memory management? I think it is worthwhile thinking a bit about these issues, especially in light of the questions about the validity of the original code.

Do you mean [checkInsertDealloc()](https://github.com/onnx/onnx-mlir/blob/master/src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp#L144-L159) (maybe developed by @chentong319 ) and [checkOpResultIsReturned](https://github.com/onnx/onnx-mlir/blob/master/src/Transform/EnableMemoryPool.cpp#L30-L39)() (maybe developed by @doru1004 )? These functions determine which value should be freed. When the value is input of return op, the value is not freed. This PR updates these functions.

AlexandreEichenberger:
@imaihal Thanks for your explanations. I get it now, %2 and %0 are aliases, but if we just look into the use of %0 it is not returned and thus appears to be dead.

My questions still remain about the code, but since you used the same pattern as before, I will accept it. But someone should look into this. Do you mind opening an issue about it, if you believe that it is indeed a potential problem (e.g. with operations generating more than one output, e.g. LSTM?) @tungld Any opinion on the code that appears to look only at the first of potentially many results? Tx

AlexandreEichenberger (approved):
LGTM

tungld:
> Do you mind opening an issue about it, if you believe that it is indeed a potential problem (e.g. with operations generating more than one output, e.g. LSTM?) 

+1 for opening an issue.

I think it's better to insert dealloc with a global view of the whole code, that's why we will use MLIR BufferDeallocation soon to inside dealloc automatically. Currently, inserting dealloc is a kind of local solution for each op.

> @tungld Any opinion on the code that appears to look only at the first of potentially many results? Tx

The comment is obsoleted. This function checks for a specific result `checkInsertDealloc(Operation *currentOp, int resultIndex)`. In case of an op having multiple results, we should iterate and check them on by one. E.g. We do so for LSMT/GRU/RNN since they have multiple results.

src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
imaihal: @AlexandreEichenberger Thanks for your comments! I reflected the first comments.
Regarding second comment,  ReinterpretCastOp, ONNXReshapeOp, ONNXSqueezeV11Op, and ONNXUnsqueezeV11Op are processed here. Since they have only one result, I get only the first result.

PR #826, opened 2021-08-17
Upgrade conv to use shape helper that handle runtime image and weights

AlexandreEichenberger:
This version supports runtime everything in CONV, and I have upgraded the TestConv to also test convolutions with auto pad VALID, SAME_LOWER, and SAME_UPPER and random strides, dilation, image and filter size. Test forces runtime checks as these are more constraining than the compile time sizes.

I moved away from using the CONV parameters as now both the Kernel dimensions & padding may be runtime and thus are preferably stored in the CONV shape helpers as lists of Index Expressions.

This patch also enabled the use of verifier for ONNX operations, and I migrated all of the CONV testing of validity in that operations. I find this approach much better as it remove the constant worry about having the right state away from code that transform the representation. Code there has to initially tolerate non-fully defined tensors, but they will be eventually defined, and testing will be fully activated prior to lowering to Krnl. 

Ideally this would have been done in smaller patches, but given the current issue with Jenkins, though it would be easier to proceed with one big patch.


src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -1580,6 +1580,178 @@ LogicalResult ONNXReduceSumV11Op::inferShapes(
...
+  } else {
+    for (int i = 0; i < 2 * spatialRank; ++i)
+      if (ArrayAttrIntVal(pads, i) != 0)
+        return op->emitError("Bad pad value: nonzero pad value only allowed "
+                             "with NOTSET option");
+  }
+  return success();
+}
+
+static LogicalResult verify(ONNXConvOp op) {
```
tungld, line 1686: As I understand this verify will be called when an ONNXConvOp is built. Have you checked that it works if all the inputs of the op are MemRef instead of Tensor?

AlexandreEichenberger:
@tungld as requested, I converted a test of mine after shape inference:

``` mlir
  func @main_graph(%arg0: memref<1x1x5x5xf32>, %arg1: memref<1x1x2x2xf32>) -> memref<1x1x5x5xf32> attributes {input_names = ["x", "W"], output_names = ["y"]} {
    %cst = constant unit
    %0 = "onnx.Conv"(%arg0, %arg1, %cst) {auto_pad = "SAME_LOWER", kernel_shape = [2, 2], strides = [1, 1]} : (memref<1x1x5x5xf32>, memref<1x1x2x2xf32>, none) -> memref<1x1x5x5xf32>
    return %0 : memref<1x1x5x5xf32>
  }
```

and I ran lowering to krnl without any problems. I also ran another test where the image dimensions are dynamic:

``` mlir
  func @main_graph(%arg0: memref<1x1x?x?xf32>, %arg1: memref<1x1x2x2xf32>) -> memref<1x1x?x?xf32> attributes {input_names = ["x", "W"], output_names = ["y"]} {
    %cst = constant unit
    %0 = "onnx.Conv"(%arg0, %arg1, %cst) {auto_pad = "SAME_LOWER", kernel_shape = [2, 2], strides = [1, 1]} : (memref<1x1x?x?xf32>, memref<1x1x2x2xf32>, none) -> memref<1x1x?x?xf32>
    return %0 : memref<1x1x?x?xf32>
  }
```
and it also ran lowering to krnl without any problems.

tungld:
Thank you for testing with MemRef! It's what I wanted to know.

tungld (approved):
LGTM!

PR #828, opened 2021-08-19
Fix IndexExpr bug for min max of literals.

AlexandreEichenberger:
Recently discovered that the `IndexExpr::min()` did not work when all the arguments are literals. This PR fixes it. The previous CONV code avoided this bug by using select, which is a less "affine" operation than min or max. WIth this bug fixed, CONV now use max instead of a select.

Separately, I also discovered that TestConv has too many RapidTest random variables, which made the last one essentially constant. This PR also reduces the number of variables by explicitly iterating over 2 values for strides and dilation (namely 1, default, and 2, a small non default value) and 2 values for static/dynamic testing.

It added a bit more tests, 12 runs of 100 tests, but much better coverage of possible issues. So I think it's worth the tradeoffs. We could potentially reduce it once CONV is very stable.


tungld (approved):
LGTM with pending DCO fix.

PR #829, opened 2021-08-20
Update support for Unsqueeze-13 and Squeeze-13

kwu91:
This PR adds shape inferencing, rewriting, and krnl lowering for Unsqueeze-13 and Squeeze-13. Most of the changes leverage existing implementation for opset 11 but with support for the new interface for opset 13. Unit tests have also been updated, and a test for unknown dimensions is added for Unsqueeze. 

There is an existing draft PR here #691 but the codebase has changed a lot since.

@NathanielMcVicar


AlexandreEichenberger:

tungld:
@kwu91 Thank you for the patch!

As I understand, this patch only deals with the case where `axis` is generated by a constant op. In this case, how about writing DRR rules to convert ONNXSqueeze to ONNXSqueezeV11Op where constant axis values in ONNXSqueeze are converted to attributes in ONNXSqueezeV11Op? So, we don't need to duplicate code between two versions.

This location, https://github.com/onnx/onnx-mlir/blob/master/src/Dialect/ONNX/Rewrite.td, is perhaps suitable for such rules.

kwu91:
@tungld 

Thanks! That's a good point! Just in case I'm not overlooking anything, there currently isn't a helper function (defined in ONNXOpsHelper.cpp or otherwise) that can convert a ConstantOp to an attribute, right?

tungld:
Yes, as fas as I know. Please write a helper function for your purpose. Thanks!

AlexandreEichenberger:
@kwu91 Second @tungld 's opinion, better reuse the code we have for constant axis. If you eventually want to generate the code for the more general case with dynamic axis, I can help you. The overall approach would be as follow:

1.  Write a shape helper function that uses IndexExpr. These are polymorphic types that work as follow: during shape inference, were we cannot write generic code, it will give you an integer answer (if can be computed at compile time) or a '?' stating that its a runtime dimensions.
2. That same shape help (using IndexExpr) will generate the proper runtime code to compute the dimensions during the lowering from ONNX to KRNL. At that time, its ok to write affine/std dialect ops.
3. Will need to rewrite the algorithm to take the axis as a runtime dimension.

And for 1 & 2, it should be close to writing code as of now. You can see Conv for example, you basically create DimIndexExpr for all dimensions, and then can add/sub/mult/div, select, clip,... these values.

kwu91:
@AlexandreEichenberger

Thanks for your valuable input! I think supporting dynamic axes for Squeeze/Unsqueeze will ultimately be the goal. Is there already a work in progress and if so a timeline?

tungld:

tungld:
@kwu91 please let me know when it is ready for review. Thanks!

kwu91:
@tungld It should be ready. Please take a look. Thank you!

src/Dialect/ONNX/Rewrite.td
```diff
@@ -329,4 +333,14 @@ def ConstantOpNormalizationPattern6: Pat<
    [(AttributeIsNotNull:$stringsAttr)]
 >;
 
+def ONNXSqueezeOpLowering: Pat<
+   (ONNXSqueezeOp $data, $axes),
+   (ONNXSqueezeV11Op $data, (createArrayAttrFromConstantOp $axes))
+>;
```
tungld, line 339: Please add a condition to check whether `$axes` is produced by ONNXConstantOp, e.g.
```
def ONNXSqueezeOpLowering: Pat<
    (ONNXSqueezeOp $data, $axes),
    (ONNXSqueezeV11Op $data, (createArrayAttrFromConstantOp $axes)),
    [(IsDenseONNXConstant $axes)]
 >;
```

We have a function helper `isDenseONNXConstant`, please write a wrapper function `IsDenseONNXConstant`. Thanks!

src/Dialect/ONNX/Rewrite.td
```diff
@@ -329,4 +333,14 @@ def ConstantOpNormalizationPattern6: Pat<
...
+
+def ONNXUnsqueezeOpLowering: Pat<
+   (ONNXUnsqueezeOp $data, $axes),
+   (ONNXUnsqueezeV11Op $data, (createArrayAttrFromConstantOp $axes))
+>;
```
tungld, line 344: Please add a condition for `$axes` as above.

test/mlir/onnx/onnx_canonicalization.mlir
```diff
@@ -473,3 +473,27 @@ func @test_rewrite_batchnormtestmode_1d(%arg0 : tensor<64xf32>) -> tensor<64xf32
...
+
+func @test_rewrite_unsqueeze(%arg0 : tensor<10x10xf32>) -> tensor<*xf32> {
+  %axes = "onnx.Constant"() {value = dense<[0, 3]> : tensor<2xi64>} : () -> tensor<2xi64>
+  %0 = "onnx.Unsqueeze"(%arg0, %axes) : (tensor<10x10xf32>, tensor<2xi64>) -> tensor<*xf32>
+  "std.return"(%0) : (tensor<*xf32>) -> ()
+
+  // CHECK-LABEL: test_rewrite_unsqueeze
+  // CHECK: [[UNSQUEEZE:%.*]] = "onnx.UnsqueezeV11"(%arg0) {axes = [0, 3]} : (tensor<10x10xf32>) -> tensor<*xf32>
+  // CHECK: return [[UNSQUEEZE]] : tensor<*xf32>
+}
```
tungld, line 499: Could you add two more tests in case that `%axes` is not constant? In this case, onnx.Squeeze/Unsqueeze will be unchanged.

tungld (commented):
@kwu91, thanks for updating the patch! 

We should add conditions for each DRR rules to make the rules illegal without errors.

Beside lit tests, please enable end-to-end tests in `test/backend/test.py`. Since we only support the constant case, the tests should look like:
```
"test_squeeze_cpu": {CONSTANT_INPUT:{1}},
 "test_squeeze_negative_axes_cpu": {CONSTANT_INPUT:{1}},
```

`1` means the second input is constant.

test/mlir/onnx/onnx_canonicalization.mlir
NathanielMcVicar: @kwu91 Thanks so much for doing this Kevin!
@chentong319 This will replace the v13 unsqueeze we discussed earlier.

NathanielMcVicar:
@kwu91 Thanks so much for doing this Kevin!
@chentong319 This will replace the v13 unsqueeze we discussed earlier.

tungld:

tungld:

tungld:
@kwu91 backend tests failed because there were no shape inference for Squeeze and Unsqueeze. Moving the rules from `rewrite.td` to [decompose.td](https://github.com/onnx/onnx-mlir/blob/master/src/Transform/ONNX/Decompose.td) will fix the tests.

PR #830, opened 2021-08-23
Run-onnnx-lib for models with dynamic/runtime dimensions

AlexandreEichenberger:
Added support to run a given .so where the user can provide the runtime dimensions of a model by a line argument. For example a model with the following signature
```
builtin.func @main_graph(%arg0: tensor<?x?x204xf32>) -> tensor<?x?x1xf32> { ... }
```
can be run with after being compiled with the first 2 "?" being replaced by 7 and 2018, respectively.
```
run-onnx-lib model.so -d "[7, 2048]" -m 8
```

### DIFF
--- a/utils/RunONNXLib.cpp
+++ b/utils/RunONNXLib.cpp
@@ -98,7 +98,7 @@ void (*dll_omTensorListDestroy)(OMTensorList *);
 #define OM_TENSOR_CREATE omTensorCreate
 #define OM_TENSOR_LIST_CREATE omTensorListCreate
 #define OM_TENSOR_LIST_DESTROY omTensorListDestroy
-#define OPTIONS "hn:m:v"
+#define OPTIONS "hn:m:vd:"
 #else
 #define RUN_MAIN_GRAPH dll_run_main_graph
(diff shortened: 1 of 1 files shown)


tungld:
@AlexandreEichenberger once the model was compiled with the first 2 "?" being replaced by 7 and 2048, I though `model.so`'s signature contained `7` and `2048`, and `run-onnx-lib` could run without additional information. Could you explain more why we should provide `-d "[7, 2048]"` to `run-onnx-lib`?

AlexandreEichenberger:
@tungld The idea is that the model is compiled with the "?", which is also reflected in the signature with the "?". Then with that command, it let us generate the tensors with the runtime dimensions where the "?" are substituted (in the data) by the proposed dimensions.

In summary: compile once with the "?", then run multiple times with any runtime values of interest.

tungld:
> In summary: compile once with the "?", then run multiple times with any runtime values of interest.

I see. It makes sense. Thanks! I though `?` was replaced while compiling.

I seems the last commit was not signed off so DCO failed.

tungld (approved):
LGTM.

PR #831, opened 2021-08-24
Change the option memory-bundling-enabled to disable-memory-bundling

tungld:
Change the option `memory-bundling-enabled` to `disable-memory-bundling` to make it easier to understand.

By default, memory bundling is enabled, passing `--disable-memory-bundling` to `onnx-mlir` to disable.

Help is now better to understand:
```
--disable-memory-bundling                 - disable memory bundling related optimizations
                                            where several passes work together to bundle buffers.
                                            Buffer management by MLIR is used instead.
                                            Try this if you experience a significant compile time.

```

### DIFF
--- a/src/MainUtils.cpp
+++ b/src/MainUtils.cpp
@@ -481,21 +481,19 @@ void addONNXToKrnlPasses(mlir::PassManager &pm) {
   // Only emit memref.dealloc if memory bundling is enabled.
   // Otherwise, memref.dealloc will be emitted by buffer-deallocation pass.
   pm.addPass(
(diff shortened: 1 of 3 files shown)


AlexandreEichenberger (approved):
Think its much better, I did not like the previous name, this one is much better.

AlexandreEichenberger:
@tungld , if the PR has just started, would it make sense to add somewhere in the doc a list of options and what they do?

tungld:
> would it make sense to add somewhere in the doc a list of options and what they do?

It seems to me `onnx-mlir --help` can provide enough information. However, we can have one page in http://onnx.ai/onnx-mlir for the usage of `onnx-mlir`. Let me add the page later in another PR that polishes other docs also.

PR #838, opened 2021-08-27
Matmul small tile opt

AlexandreEichenberger:
When the dimensions of a matrix are very small, (say 1,2,3,4) the tiling was still 4,84. 
With this PR, I reduce the tiling to the specific values of constant dimensions, when they are very small. The result is that the inner kernel becomes basically inlined for the actual sizes of the small matrix.

This resulted in small matrices to be simdized, and this in turn highlighted the fact that we did not enforce a default alignment of the alloc to be SIMD friendly. Generally, machines forces malloc of sizes >8 bytes to be aligned to the SIMD width of the machine, but memory bundling was not aware of this. To make it more flexible, I edited all of the memref:alloc* in the code to add a default alignment when none was specified. At the present time, that default alignment is set to 16; it might have to be increased for specific machines, and can be done by adding code that looks at the `-mcpu` value and change it accordingly. That value is `defaultAllocAlign` defined in ONNXToKrnlCommon.hpp.

Once this was done, I had to fix nearly all onnx lowering tests to ignore the added `{alignment = 16 : i64}`. The pattern tested is ignored, as we don't want the lit tests to become dependent on the value of `defaultAllocAlign`.
I changed mlir2FileCheck to also generate patterns where alignment is ignored for alloc

Finally, I added the option to not reuse data when performing measurements with run-onnx-lib as some discrepancy with measurements in other settings may arise from reusing the same input tensor over and over. An option was added to that effect.


AlexandreEichenberger:

tungld:
@AlexandreEichenberger since we need to prepare an attribute for `gDefaultAllocAlign` every time we emit `memref::Alloc`, perhaps creating a helper function for emitting `memref::Alloc` with default alignment would help reduce a lot of redundant code.

For Jenkins errors, it looks like it's a linking problem when building share libs.

AlexandreEichenberger:
> since we need to prepare an attribute for gDefaultAllocAlign every time we emit memref::Alloc, perhaps creating a helper function for emitting memref::Alloc with default alignment would help reduce a lot of redundant code.

@tungld, agreed. I added a MemRefBuilder where we can hide this. I also added the delete (for which we may have different policies if static or dynamic, with respect to bundling) and dim ops (for which we were doing folding sometimes, modified it to be folding all the time).

I moved the `gDefaultAllocAlign` global to the OMOptions.h file as eventually it should be determined as a function of -mcpu/-mtriple option.

AlexandreEichenberger:
@tungld it now compiles, still waiting for the Window CI. If possible, it would be great if you have another look at the code and see if you can approve it today. Thanks

tungld (approved):
LGTM. Thanks for many changes!

Issue #835, opened 2021-08-25
Test suites


sstamenova:
Right now there are several different types of tests in onnx-mlir:

1) The lit tests in test/mlir
2) The numerical tests in test/numerical
3) The backend tests in test/backend
4) The unit tests in test/unit and test/onnx2mlir which do not require an onnx install
5) The one unit test in test/unit/Runtime/DocExampleTest that does require an onnx install
6) The doc tests in doc/doc-check

Some of these tests have their own targets (such as check-doc, check-onnx-lit), though some are executed by directly invoking the test target which runs all ctests. This makes it difficult to have a clean and consistent way to work cross-platform and several of the tests still fail on some platforms (most notably Windows).

The tests can be separated into several categories:
- long-running vs. not
- requiring onnx vs. not
- ctests vs. not
- lit tests vs. not
- etc.

I'd like to propose that we re-organize the tests a little bit so that they at least have their own targets (see https://github.com/onnx/onnx-mlir/issues/563 for a request for a separate target for the long-running numerical tests). We could also introduce other improvements such as using gtest so that we can leverage lit for more of the tests.

I would like to start with a couple of changes:

1. Add check-docs (or some other name of our choosing):
- Move the Runtime/DocExampleTest which requires onnx install to doc/doc-example
- Exclude the building of the DocExampleTest binary from ALL - since it requires an onnx install it creates problems for the ALL target (see https://github.com/onnx/onnx-mlir/issues/561)
- Add a check-docs target which runs both the existing check-doc test and the doc example test
This makes sense because both the DocExampleTest and check-doc are intended to make sure that the docs are up-to-date and provide examples. Both also require an onnx install.

2. Add check-unittest (or some other name of our choosing):
- Add all remaining unit tests (that are not the numerical tests) to a single target that can execute them by leveraging a ctest label
This makes sense too because all of these tests require a simple compilation and then execution of the binary that was created, they all run in 30 seconds or less and do not require onnx.

3. Modify all of the CI flow to use these new targets (along with the already existing check-numerical target which is used on Windows but not in the other CI flows)
This would allow us to run all of the tests based on their targets only after the required dependencies are installed and it will test that the new targets work correctly

I think for 2 (check-unittest), it would also make sense to make those test gtests and use lit to run them. I am not sure whether the numerical tests can similarly by run using gtest and lit, but that would make reporting of the results on a run cleaner as well.

@caoimhinuibrian @gongsu832  @AlexandreEichenberger What do you think?

Issue #561, opened 2021-03-26
test/unit/Runtime/DocExampleTest does not respect cmake configuration


stephenneuendorffer:
This test attempts to install onnx through the setup.py mechanism before executing, which results in an internal (duplicate?) cmake-based compilation of onnx.  However, this cmake does not inherit any environment from the outer cmake, which has been properly configured.  This fails spectacularly in our environment because we need to configure the location of CC, pybind11, and other dependencies.  It's unclear to me how to best fix this, or even what the goal of this test is. At least, maybe it could be made optional?

sstamenova:
I also have a question about this test (as well as a couple of others such as check-doc) that depend on an onnx installation. Does it make sense to move all such tests out of build all and into a separate check-* that explicitly requires an onnx installation? That way we could (maybe, eventually) get to a point where we can build all successfully on a variety of platforms and any additional steps required for the tests will not be part of the build.

caoimhinuibrian:
@sstamenova that seems like a reasonable suggestion to me

PR #839, opened 2021-08-27
Add a target to run all of the document tests together and make it work on Windows

sstamenova:
The DocExampleTest is meant as an example to be referenced on how to use the tools, so it makes sense for it to live with the documentation and be tested together. It also requires an installation of onnx (which the other doc test check-doc) does as well. It is the only part of the ALL target that requires an installation of onnx currently and makes it difficult to build ALL cleanly across platform.  See https://github.com/onnx/onnx-mlir/issues/561.

This change adds a new target to run both document tests together successfully (including on Windows) and makes the `ALL` target not require onnx installation.

Several things are needed to make this happen:

1. Move the document example test from under `test/unit/Runtime/DocExampleTest` to `docs/doc_example`
2. Add a new target `check-doc-example` which only executes the doc example test. This new target builds the test and executes it.
3. Add a new target `check-docs` which runs both the `check-doc` and `check-doc-example` targets allowing both to be run together
4. Mark the doc-example binary, so that it is not built as part of the `ALL` target. This means that none of the steps needed to build it (such as installing onnx, calling onnx-mlir, etc.) will be executed when the `ALL` target is built. These steps are executed when `check-doc-example` is run
5. Remove the explicit install of onnx through pip in the doc example test. The documentation states that this is needed for a variety of things to work correctly and all of the pipelines do it, so it is not needed in the cmake file.
6. Update the pipelines to use `check-docs` instead of `check-doc` - this will execute both tests together.
7. Several different scripts, as well as onnx-mlir, have been updated to work correctly on Windows

Notes:
- Temporarily (until there's a new clean target for the unit tests (see https://github.com/onnx/onnx-mlir/issues/835) the doc example test will be executed twice in the pipelines - once when `check-docs` is called, and once when `make test` is called. Since the test only takes a few seconds, I think this is acceptable.
- The new target is called `check-docs` and the document check target retains the name `check-doc`. I'm not super keen on either name, so if you have better suggestions, please let me know.
- This changes the name of the directory to `doc_example` to match other dirs in the same location. If you prefer a different naming convention, let me know
- The Windows-CI follows the convention of calling a script which contains the actual `cmake` steps. I've followed that here, but as far as I can tell the only reason to do so is to be able to reference the script in the documentation. If that's a valid reason, then I'll continue to do that for any future changes to the Windows-CI, but if it's not a requirement to have the scripts from the CI back the `README` files, I would prefer to not do that since it just complicates things.
- This change excludes the doc example binary from the `ALL` target, so calling `make test` without building the binary will now fail. I think this is fine since the goal is to have a `check-unittest` target to take over the rest of the unit tests and `make test` will no longer be necessary. If it's desired for `make test` to continue executing all available ctests, I can add a cmake option to build the test in the `ALL` target similar to llvm's `LLVM_BUILD_TESTS` 


caoimhinuibrian:

AlexandreEichenberger:
@gongsu832 this changes looks good to me, since you have the most experience with our scripts, do you mind to give this patch your review? Thanks

sstamenova:
@AlexandreEichenberger : I figured out what the issue was with the Windows-CI here - the machines where the tests run have multiple python installations. I fixed that, so the Windows-CI should be green with this change and execute the doc tests.

@gongsu832 : I'd particularly like your opinion on whether I should add a CMake variable to build the test executables.

gongsu832:

docs/doc_example/CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
...
+  LINK_LIBS PRIVATE
+  OMRuntimeTestModelLibrary
+  cruntime
+  )
+set_output_directory(OMRuntimeTest BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR} LIBRARY_DIR ${CMAKE_CURRENT_BINARY_DIR})
+
+add_test(NAME OMRuntimeTest COMMAND OMRuntimeTest WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
+set_property(TARGET OMRuntimeTest PROPERTY FOLDER "Docs")
+set_tests_properties(OMRuntimeTest PROPERTIES LABELS doc-example)
+
```
gongsu832, line 78: The expectation is that `make test` should not fail after `make`. So yes as long as we still have `make test`, either we should make `OMRuntimeTest` work with `make test`, or we should remove it from `make test`.

gongsu832:
Apart from the `OMRuntimeTest` issue, the rest look fine to me.

AlexandreEichenberger:

gongsu832:

docs/doc_example/CMakeLists.txt
caoimhinuibrian: @sstamenova @gongsu832 Does something need to be done here?

sstamenova: I made the change to make `make test` continue to work as before by adding the `ONNX_MLIR_BUILD_TESTS` option. This allows `make test` to stay as it is unless `ONNX_MLIR_BUILD_TESTS` is set to `OFF` at configuration time.

gongsu832 (approved):
LGTM

AlexandreEichenberger:
@sstamenova please look into the WindowCI error. Ready to merge once this is fixed. Thanks

sstamenova:
> @sstamenova please look into the WindowCI error. Ready to merge once this is fixed. Thanks

@AlexandreEichenberger : The failure in the Windows CI seems like an intermittent infrastructure issue:

```
D:\a\2\llvm-project\llvm\lib\Transforms\InstCombine\InstCombineCasts.cpp : fatal error C1085: Cannot write compiler generated file: 'D:\a\2\llvm-project\build\lib\Transforms\InstCombine\CMakeFiles\LLVMInstCombine.dir\InstCombineCasts.cpp.obj': No space left on device
```

I'll merge with master again and I expect that the Windows CI will pass.

gongsu832:

PR #841, opened 2021-08-31
Upgrade MaxPool and AveragePool to handle SAME_LOWER/UPPER with dynamic sizes

AlexandreEichenberger:
Upgraded the shape inference of the MaxPoolSingleOutOp and AveragePoolOp to use the same shape helper as convolutions, merging a slightly different set of features 
* Conv has an optional kernel_shape, the others do not
* Pool have ceil_mode, Conv does not
* dilations is optional for Conv and MaxPool, but not available for AveragePool

Just like for Conv, the operations itself does not carry the non-default attribues as some (kernel size and pad) may now be dynamic values, carried as arrays of IndedExpr.

Verifiers are now also used for Max & Average Pool operations.

Because of this new support, all of the backend tests now use '-1:{-1}' pattern.

Note that the code generation scheme remains practically identical as the previous one. I did not upgrade it to the newer scheme. Just used SymbolIndexExpr for kernelShape and pads.


AlexandreEichenberger:

tungld (approved):
LGTM. Thanks!

AlexandreEichenberger:

PR #844, opened 2021-09-05
New support for resize op

chentong319:
In the new version of yolo model, yolov4, the resize is specified by input sizes, instead of scales. This PR added the support of using sizes input.
Accordingly, test cases for resize using sizes are enabled in our backend test.


tungld (approved):
LGTM. Thanks!

PR #845, opened 2021-09-06
Fuse convolution and addition

tungld:
This patch is to fuse `ONNXAddOp(ONNXConvOp %X, %Y)` into a single ONNXConvOp if `%Y` is a constant, e.g. 

```mlir
%0 = "onnx.Conv"(%input, %weight, %bias) 
%1 = "onnx.Constant"()
%2 = "onnx.Add"(%0, %1)
```

will become 

```mlir
%new_bias = "onnx.Constant"()
%0 = "onnx.Conv"(%input, %weight, %new_bias) 
```
where `new_bias` is computed as `%bias + onnx.squeeze(%1)` by using constant propagation.

Constant propagation for ONNXSqueezeV11 is implemented to support the above fusion.

This patch also moves all canonicaliatione patterns from `combine.td` into `rewrite.td`. This is to reduce duplication code between `combine.td` and 'rewrite.td` where both have the same purpose that is canonicalization.


tungld:

chentong319:
Could you add how new_bias is generated (related to %bias and %1) in the comment?

tungld:
> Could you add how new_bias is generated (related to %bias and %1) in the comment?

I updated the comment.

PR #846, opened 2021-09-06
Update onnx.EntryPoint op definition to allow operation generation by tablegen.

negiyas:
Add the following argument part in the onnx.EntryPoint op definition to enable operation generation by tablegen.
let arguments = (ins SymbolRefAttr:$func,
           I32Attr:$numInputs,
           I32Attr:$numOutputs,
           StrAttr:$signature);

Otherwise, tablegen causes an error with the "error: op 'onnx.EntryPoint' argument number mismatch: 4 in pattern vs. 0 in definition ..." message.


tungld (approved):
LGTM. Thanks!

PR #847, opened 2021-09-07
Move some verification in shape inference into a separate pass

chentong319:
In the shape-inference pass, certain conditions are checked, and if failed, the shape-inference failed. However, some of these conditions may become true if transformation and shape inference are invoked several times. This is the case for yolov4: shape inference and constant propagation caused by Shape op. 
This patch has the following changes:
1.  simply return success in shape inference if the input shape is not desired (unranked) 
2. Add a separate pass call ONNXPreKrnlVerifyPass to be invoked before lowering onnx to krnl.
3. Modify and add the related lit test
4. Add an option to call shape inference and constant propagation repeatedly.

The PR can compile static yolov4 model (with constant batch input).


caoimhinuibrian (commented):
@chentong319 This looks fine, but it's failing the format check right now.

AlexandreEichenberger:
@chentong319 I would like to better understand the problem that arise in Yolo. Could you show the sequence of ops that fail the shape inference and requires repeated calls to it? Thanks

chentong319:
1. Shape inference pass is needed to generate the shape of input for ONNXShapeOp
2. ONNXShapeOp will generate a constant by canonicalization when  input shape is static
3. This result is used as input of SliceOp and further constant propagation can be done and help further shape inference.

This issue was discussed previously.

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2722,7 +2723,7 @@ LogicalResult ONNXDequantizeLinearOp::inferShapes(
     std::function<void(mlir::Region &)> doShapeInference) {
   auto inTy = x().getType().dyn_cast<RankedTensorType>();
   if (!inTy || !inTy.hasStaticShape()) {
-    return emitOpError("Input is not a statically-shaped type");
+    return success();
```
AlexandreEichenberger, line 2726: Question: Some emit an error if not ranked, other if not statically shaped, like here. How do you keep the distinction later on?

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2883,16 +2884,16 @@ LogicalResult ONNXTileOp::inferShapes(
...
   // Read 'repeats' value.
   if (!repeats().getType().isa<RankedTensorType>())
-    return emitError("Repeats tensor not ranked");
+    return success();
 
   // 'repeats' tensor is an 1D tensor.
   auto repeatsTensorTy = repeats().getType().cast<RankedTensorType>();
   if (repeatsTensorTy.getShape().size() != 1)
-    return emitError("Repeats tensor must have rank one");
+    return success();
```
AlexandreEichenberger, line 2896: Same here, must have a rank of 1. How do we keep track of that?

It would seems that we should rather have a 

```
tolerateOrEmitError(phase, message)
```

in which you can switch between generating errors or accepting the current "violation"

src/Transform/ONNX/ONNXPreKrnlVerifyPass.cpp
```diff
@@ -0,0 +1,75 @@
...
+        if (failed(verifyRanked(op)))
+          signalPassFailure();
+      }
+    }
+  }
+
+private:
+  static LogicalResult verifyRanked(Operation &op) {
+    for (auto ty : op.getOperandTypes()) {
+      if (!ty.isa<RankedTensorType>() && !ty.isa<NoneType>()) {
```
AlexandreEichenberger, line 60: Here is where all the previous errors are reduced to "ranked or not". But several errors where of a different nature.

AlexandreEichenberger (commented):
@chentong319  Current approach "loose" some of the shape errors we were detecting. This need to be addressed, either by migrating these into a "verify" code or by selectively turning on/off the error reporting inside the ShapeInference.

src/Dialect/ONNX/ONNXOps.cpp
chentong319: It's my fault that I should separate them here. In long term, we may define a  trait for op that requires special handling.

AlexandreEichenberger: @chentong319 so what is the proposed solution to address this. If it is not done in this PR, we will loose these conditions and it will be much harder to add them later.

chentong319: I checked the code and found the extra condition is not really needed. I'm deleting them now.

src/Dialect/ONNX/ONNXOps.cpp
chentong319: I put the check back.

src/Transform/ONNX/ONNXPreKrnlVerifyPass.cpp
chentong319: In shape inference, we did some semantic check with assumption of all tensors are statically shaped. This is approach may cause trouble. I only deleted the check for concatOp, which caused trouble for yolov4. I can not fix all them. To me it is not important to do the check and we can assume (or use onnx check model tool) the model is correct.

src/Dialect/ONNX/ONNXOps.cpp
AlexandreEichenberger: @chentong319 so you went through all conditions that are not "must have rank" and none of these conditions are needed?

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2381,7 +2381,7 @@ LogicalResult ONNXCastOp::inferShapes(
     std::function<void(mlir::Region &)> doShapeInference) {
   ShapedType inputType = input().getType().dyn_cast<ShapedType>();
   if (!inputType) {
-    return emitError("Non-shaped input type");
+    return success();
```
AlexandreEichenberger, line 2384: Nit: in most places we test for `.getType().isa<RankedTensorType>()` but here its for `getType().dyn_cast<ShapedType>()`. 

It seems to be the same thing, but I actually don't know the difference between ShapedType vs RankedTensorType. 

The same test for ShapedType occurs elsewhere too

src/Transform/ONNX/ONNXPreKrnlVerifyPass.cpp
AlexandreEichenberger: Agreed, if other checks are more of the type "verify", it should be done in a verifier.

AlexandreEichenberger (commented):
If you could cleanup the difference between shape tests vs rank tests, the code would be more consistent. Once this is done, it looks good to me.

chentong319:
1. ShapedType: ShapedType is a super class of RankedTensorType. So far, input of all the op we can handle should be RankedTensorType. I changed all the check. Some ops, such as ZipMap in onnx-ml, require map type or seq type. We may have customized check for some op in future.
2. The check for static shape of input is to  make sure that the shape propagation can really improve the shape info. Since the output shape is checked before propagation, this static shape check is not really needed.

AlexandreEichenberger (approved):
LGTM, thanks for the changes

PR #849, opened 2021-09-07
Update type-map to support nested scopes

gramalingam:
This PR has two simple changes:

(a) Change the simple-map that was used to track the ONNX types of ONNX tensor names to use "SymbolMapping" (which supports nested scopes).

(b) Remove the few lines of #ifdef ONNX_FUNCTION_TYPE_CONTEXT that existed only for temporary support until the ONNX version was upgraded to the one that has type-context (for function-op body-builder).


chentong319:
@gramalingam Is this PR ready for review?

caoimhinuibrian (approved):
LGTM

gramalingam:
@chentong319 : yes, it is ready for review. I added the label. I assume that the CircleCI failure is unrelated? Thanks!

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -1088,6 +1086,9 @@ class FrontendGenImpl {
     // Restore caller context
     frontend_symbols_.popScope(func_name_prefix);
     frontend_symbols_ = std::move(callerScope);
+    onnx_type_map.popScope(func_name_prefix);
+    onnx_type_map = std::move(callerTypeMap);
```
chentong319, line 1090: I did understand why callerTypeMap is needed. I think that push and pop scope should serve the purpose of saving and restoring the map. Could you please explain?

gramalingam: @chentong319 : a nested-scope implies all the names in all the scopes are visible in the inner scopes. This works for handling the nesting of subgraphs in control-flow ops like If/Loop/Scan. However, functions do not represent a nested inner scope in this sense. The function-body cannot reference a name defined in the caller. The current treatment will handle this correctly. (Using push/pop scope will work fine for correct onnx models, but not for incorrect models when there are such name-collisions.)

PR #851, opened 2021-09-09
Better format for shapeInformation option

tungld:
This patch improves the readability of `shapeInformation` option in `onnx-mlir` command, by changing from, say, "0:512,1024|1:32,64" to "0:512x1024,1:32x64".


AlexandreEichenberger (approved):
LGTM. By the way, there is also a JSON parser in LLVM which I used for run-onnx-lib, I definitely like the 2x4x6 format too.

PR #852, opened 2021-09-09
Update the front-page of http://onnx.ai/onnx-mlir

tungld:
This patch is to use a different ONNXAI.md file for the front page of http:/onnx.ai/onnx-mlir, instead of use README.md from the GitHub page, which is to avoid confusion because these two files have two different purposes.


AlexandreEichenberger:
@tungld Have you checked that https://github.com/onnx/onnx-mlir/settings/pages which generates the website does take the ONNXAI.md as the main landing page?

My understanding is that it does not, it just build everything using the README.md. So you would need to rename the ONNXAI.md under the doc  as README.md.

Its ok by me to have a different text in the landing page of GitHub vs onnxai

tungld:
> Have you checked that https://github.com/onnx/onnx-mlir/settings/pages which generates the website does take the ONNXAI.md as the main landing page?

I cannot access https://github.com/onnx/onnx-mlir/settings/pages to see configurations. I checked with my local Jekyll server and it works well with github-pages. The point is README.md is used by default to generate index.html. We can change to use ONNXAI.md to generate index.html. Could we merge to see if it works online, if not we can revert the patch?

AlexandreEichenberger (approved):
Let's try it. Can revert if it does not work. We have to wait a bit after the PR is merged as it takes some time to be propagated

AlexandreEichenberger:
Looks like it worked. Thanks

PR #853, opened 2021-09-09
Add constant split support to Split-13

kwu91:
Similarly to the Unsqueeze and Squeeze Ops, this is an attempt enable Split-13 with support for constant `split` input by rewriting into Split-11. If the `split` argument is a constant, then it is turned into a constant attribute and passed to Split-11. If the `split` argument is a None type, then the `split` attribute for Split-11 will be empty.


chentong319:

kwu91:
This is related to my previous PR https://github.com/onnx/onnx-mlir/pull/829, but implementing these Ops with decompose rules will fail if I want to use BF16 types which v13 supports but v11 does not. This just occurred to me now, that I didn't realize before. Would it make more sense to have the lowering implementation be done in `src/Dialect/ONNX/ONNXOps.cpp` and ` onnx-mlir/src/Conversion/ONNXToKrnl/Tensor/Split.cpp` even if the code is redundant?

chentong319:
You are right on the issue of type.  I really want to avoid duplicating code. Can we define a lowering function to lower this op (V11 and V13) when the split is a constant? Similar to constant propagation. 
Another workaround is to convert or check the type when v13 is rewritten to v11. When BF16 is used, we either have extra cost for type conversion or we just lose some optimization opportunity.

kwu91:
@chentong319 Thanks for your suggestions!

If the input argument is not a constant (even though the split is), then lowering the Op like constant propagation won't work right?
And if the BF16 type needs to be retained throughout the lowering process, then type conversion might not be suitable either?

chentong319:
@kwu91 yes, we will have problem. However,  I think we will finally add the implementation to lower SplitOp with non constant split input. So if we do not rewrite split to splitV11 when the input type is BF16, this case will be handled later, less optimized. It is not a perfect solution.
If you can reorganize the code for lowering to have a utility function that lowers SplitOp with Values and Attributes as input (not the op), in lowering SplitOp, you can call this utility if split input is a constant. We can do similar thing for constant propagation. A little bit more coding.

kwu91:
Thanks @chentong319 

Please let me know if my understanding is incorrect; so if the `split` input is a constant for SplitV13, then the lowering will share a lot of code with the lowering for SplitV11. This shared code can be essentially be lifted to an utility function that both SplitV13 and SplitV11 lowering can call, so that we can avoid as much duplicate code as possible.

chentong319:
@kwu91 Yes, that's correct.

kwu91:
I moved the common code between V11 and V13 to separate functions. I also added tests for V13 where applicable. Let me know what you think. Thank you.

AlexandreEichenberger:
@kwu91 Updated to the latest; @chentong319 please verify that the latest changes satisfy your request. Thanks

chentong319:
This PR accomplished the original goal: handle constant split for SplitOp V13 and SplitOp V11. The code is great in using IndexExpr. I am wondering whether handling dynamic split can be integrated in this framework? I am thinking that to use IndexExpr for third parameter for ONNXSplitOpShapeHelperCommon. So we can handle constant and dynamic value uniformly.
 It's OK to me to first approve this PR if you want to implement for dynamic split with another PR.

src/Dialect/ONNX/ONNXShapeHelper.cpp
```diff
@@ -743,18 +743,20 @@ LogicalResult ONNXSplitOpShapeHelper::Compute(
...
-    auto splitAttribute = createArrayAttrFromConstantOp(builder, splitConstOp);
-    optionalAttr.emplace(splitAttribute);
+    Operation *genericOp = reinterpret_cast<Operation *>(op);
+    ArrayValueIndexCapture splitCapture(
+        genericOp, split, fGetDenseVal, fLoadVal);
+    auto splitRank =
+        splitConstOp.valueAttr().dyn_cast_or_null<DenseElementsAttr>().size();
+    splitCapture.getSymbolList(splitRank, indexExprArray);
   } else if (!split.getType().template isa<NoneType>()) {
     llvm_unreachable("dynamic split not yet supported");
```
chentong319, line 756: I think we can handle the dynamic split now.

kwu91: @chentong319 I think handling dynamic split might require either a separate set of patches or additional effort beyond the scope of this PR? If that's the case, should I keep the error message for now? Let me know what you think. Thanks

chentong319: Yes, we can have a separate PR.

kwu91: Great! Let me know if there are other suggestions. I currently don't have a timeline for implementing the dynamic split. Others can take over if they desire. There are be plans to do the same for Unsqueeze/Squeeze op to allow for BF16 support first.

chentong319:

chentong319:
LGTM!

chentong319 (approved):

PR #854, opened 2021-09-10
Regression to to introduction of alignment constraints for scalar alloc/free

AlexandreEichenberger:
By introducing explicit alignment, we noticed that some benchmarks with alloc/free became slower with the align constraints compared to the same benchmarks without them. This PR simply remove alignment that is guaranteed not to be required for SIMD as they apply to the alloc of scalar


tungld (approved):
LGTM, pending a fix for lit tests.

PR #855, opened 2021-09-10
Disable dynamic memory bundling and fix static bundling

doru1004:
Disable dynamic memory bundling.


AlexandreEichenberger (commented):
@doru1004 does it still introduce the getrefs?

doru1004:
> @doru1004 does it still introduce the getrefs?

Only for the static case which is optimized, otherwise no.

AlexandreEichenberger (approved):
LGTM, but mid-term, we should remote the printout from a default run, and have it as an option on the optimization. @tungld knows how to do that, it should be replicated here.

PR #857, opened 2021-09-13
Replacing the remaining old code in LSTM/RNN/GRU by IndexExpr

tungld:
There are some remaining code in LSTM/RNN/GRU still using the old coding style. Replace them with IndexExpr.

Since we use IndexExpr, lit tests are modified to use canonicalization.


AlexandreEichenberger (approved):
LGTM, thanks for the upgrade.

PR #858, opened 2021-09-13
Support axes with dynamic value for ReduceSum when keepdims() is true

chentong319:
In op_set 13, axes of ReduceSumOp is an input tensor, not an attribute. onnx-mlir had support for ReduceSum only when axes is defined by a Constant. This PR implement support when axes is unknown at compile time. 
I focused on the case when keepdims() is true. Otherwise, I think we can rewrite `ReduceSum(data, axes)[keepdims=false]` to ` Squeeze(ReduceSum(data, axes)[keepdims=true], axes)`.

1. Shape Inference: the output tensor should have the same rank as the input tensor, but known size for every dimension.
2. Convert to Krnl: 
-  Introduce a temporary array for runtime, mask, to set mask[axes[i]] = true for the dimension to be reduced   
-  Alloc output tensor: size of dimension i `mask[i]? 1 :  input Shape [i]`
- Reduction: the output index is `mask[i]? 0 : index of loop i`,  Other code should be the same.
3. Test:  4 test cases for V13 ReduceSum were enabled. lit test was updated.


tungld:
@chentong319 could you please add lit tests for the conversion to krnl when axes are dynamic?

chentong319:
> @chentong319 could you please add lit tests for the conversion to krnl when axes are dynamic?

Test added.
In general, I prefer backend test to lit test. The backend test can tell whether the transformation is correct or not for sure. I feel that output of lowering is too complicated for us to really check the correctness. The lit test may impose extra effort when small changes are made.

tungld:
> In general, I prefer backend test to lit test. The backend test can tell whether the transformation is correct or not for sure. I feel that output of lowering is too complicated for us to really check the correctness. The lit test may impose extra effort when small changes are made.

Yes, agreed. But it is useful to see if changes in a PR affect the other lowerings or not. Perhaps, testing functionalities makes the lit tests more resistant, e.g. one test for the ordering of generated krnl.iterate, one test for primitive computation, one test for dynamic-handling code, etc. instead of checking the whole generated code.

AlexandreEichenberger:
@chentong319 It would be great if you could add a comment in the code that describes which schemes are currently supported. Recalling a bit what @tjingrant was saying about dynamic cases, I looked at this case and believe we can come up with a general scheme. Consider the template below. In My example, I assume rank 2, but it can easily be extended to higher ranks. Below is the code that would be generated, so you need to generate if-then-else using IndexExpr, which can simplify nicely. For the moment, the code for if-then-else is in the handling of MatMul, maybe it could be added to some support function.

``` c++
// setup of bounds
if (axis includes rank 0)  
  UB_red0 = UB0 // we will reduce
  UB_iter = 1          // we will not iterate
else 
  UB_red0 = 1        // we will not reduce
  UB_iter0 = UB0 // we will iterate 

// same for Rank 1, defining UB_red1 and UB_iter1

// code
for i in 0..UB_iter0, j in 0.. UB_iter1 // loop over iterations
   red = 0
   for ii in 0 .. UB_red0, jj in 0.. UB_red1 // loop over reductions
      red += IN[i + ii, j+jj] // When iterate over rank 0, ii=0 and thus does not contribute anything.
                                         // When reduce over rank 0, i=0 and thus contributes nothing too. 
   OUT[i, j] = red
```
Handle the shrinking conditions by squeeze.

The nice thing about this approach is that if you use IndexExpr, the tests will be known at compile time, and the for x=0 to 1 will be eliminated, so you can really only have one generation case. But if you want to have 2 cases, you can still emit the static vs the dynamic axis code template.

I would recommend using the new KrnlBuilder methods with lambda, that entire code can become quite readable and simple.

chentong319:
@AlexandreEichenberger I added the code on top of existing code. I believe that using IndexExpr can simplify some of them. I need to catch up with the change of IndexExpr. IndexExprs are specialized for different cases. Which document is most up-to-date one?
As for separating the reduced and not-reduced loop (as shown in your example):
1.  Not sure whether we should do this transform here or leave it to general loop optimization.
2.  When the axes is unknown at compile time, the number of loops and which loops to be reduced is unknown at compile time. It is equivalent to create krnl loop based on unranked tensor. Do we have such support?  Basically, we can not iterate on ranks in compiler. Similar situation in my code is loop over elements of axes. When dim of axes is dynamic, I generate a krnl loop. Otherwise, I can have fully unroll the loop at compile time. For this case, it might be unnecessary for performance to handle differently. On the other hand, I am wondering whether we can have a similar support as IndexExpr to handle these two case uniformly. A loop level IndexExpr. 
3. For using lambda, could you give me one code example which is the best example you think. I may learn for there.

chentong319:
@gongsu832 My previous version got error in backend-constant test.
```
[libprotobuf FATAL /usr/local/include/google/protobuf/repeated_field.h:1323] CHECK failed: (index) < (current_size_): 
terminate called after throwing an instance of 'google::protobuf::FatalException'
  what():  CHECK failed: (index) < (current_size_): 
```
I ran into such error before I upgrade my protobuf and I did not get this error locally. The test case I added may fit the condition that triggered this error: how missing optional input is encoded. Currently I just disabled the backend-constant for that test case: `test_reduce_sum_negative_axes_keepdims_random_cpu`

AlexandreEichenberger:
> I added the code on top of existing code. I believe that using IndexExpr can simplify some of them. I need to catch up with the change of IndexExpr. IndexExprs are specialized for different cases. Which document is most up-to-date one?

I would look at Gemm.cpp. Basically, you define one IndexExprScope per "level" where you introduce IndexExpr. So for example:

``` C++
  // A: before loop index expression
  for(...) {
    for(...) {
       // B: index expressions
       for() {
         //C: index expressions
      }
      // D: index expressions
}}
```

You introduce a IndexExpressionScope at A, B, C (D will reuse the one done in B).
At each level, you decide if a Value that you want to use as an IndexExpr is a Dim or a Symbol. Typically Indices are a Dim in their immediate loop body, but then further down they can be considered as symbols as they are constant over the duration of the inner bodies.
Whenever you are using a IndexExpr from an enclosing scope, you need to redefine it. Say `UB` was defined as a DimIndexExpr at level A, and you are using it in B, you have to redeclare it, e.g. `SymbolIndexExpr innerUB(UB)`. Because in general, it is best not to reuse an index expression from an outer scope inside an inner scope. Literals are the exceptions, and non dim/symbols may work too.

> As for separating the reduced and not-reduced loop (as shown in your example): Not sure whether we should do this transform here or leave it to general loop optimization.

It is part of the lowering... 

> When the axes is unknown at compile time, the number of loops and which loops to be reduced is unknown at compile time. It is equivalent to create krnl loop based on unranked tensor. Do we have such support? Basically, we can not iterate on ranks in compiler. Similar situation in my code is loop over elements of axes. When dim of axes is dynamic, I generate a krnl loop. Otherwise, I can have fully unroll the loop at compile time. For this case, it might be unnecessary for performance to handle differently. 

If you use the "keepDim" then no ranks are lost.  You can leave the reducing of the dimensionality to a squeeze op. The result will be an unranked tensor.. so we may not support this at this time.

> On the other hand, I am wondering whether we can have a similar support as IndexExpr to handle these two case uniformly. A loop level IndexExpr.

The code template above does not care which loop is a reduction and which is an iteration over data points. We disable loops by setting their range to (0..1( So when the axis are known at compile time, the if-conditions evaluate at compile time and the compiler may be able to remove these useless loops. When axis are dynamic (in the number of loops and which ones), the code just work.

> For using lambda, could you give me one code example which is the best example you think. I may learn for there.

See again Gemm and KrnlBuilder data structure.

tungld:
```C++
// setup of bounds
if (axis includes rank 0)  
  UB_red0 = UB0 // we will reduce
  UB_iter = 1          // we will not iterate
else 
  UB_red0 = 1        // we will not reduce
  UB_iter0 = UB0 // we will iterate 

// same for Rank 1, defining UB_red1 and UB_iter1

// code
for i in 0..UB_iter0, j in 0.. UB_iter1 // loop over iterations
   red = 0
   for ii in 0 .. UB_red0, jj in 0.. UB_red1 // loop over reductions
      red += IN[i + ii, j+jj] // When iterate over rank 0, ii=0 and thus does not contribute anything.
                                         // When reduce over rank 0, i=0 and thus contributes nothing too. 
   OUT[i, j] = red
```

I love the beauty of this code, it extracts all the dim check out of the loops that is suffered by Tong's code. In case of static, redundant loops can be automatically eliminated . Only concern about the dynamic case: bounds are not affines, so it is difficult to be, say, fused, with other loops, while we often use ReduceSum for ReduceL1, ReduceL2, ReduceSumSquare, etc.

AlexandreEichenberger:
> I love the beauty of this code

:-)

@tungld can we assert at compile time if the bounds are not affine? I was under the impressions that we always have square matrices, have not seen such cases, so please feel free to show me a combination that leads to it. 

It might still be better to have one version that handle the constant case, as you said, it might be easier to fuse with other loops. Of we can put a thin layer that "eliminate" loops with 1 iteration only (possibly in our krnl infrastructure)

tungld:
> @tungld can we assert at compile time if the bounds are not affine?

It looks possible with this `remainsLegalAfterInline`: https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/Affine/IR/AffineOps.cpp#L47

We have once used `isValidDim` to check affine-compatibility when lowing krnl.load/store to std.load/store and affine.load/store.

chentong319:
I agree that the code can be totally rewrite in this way. I will close this PR.

chentong319:
I found that the shape inference part of the PR is still needed. Just ignore how the lowering is implemented. At least it works and can be re-implemented in the future.

tungld (approved):
LGTM. Thanks!

Considering that changing to the new approach might take time to rewrite the whole code, it is still worth landing this first with backend tests verified.

PR #861, opened 2021-09-17
Do not emit dealloc and use MLIR buffer-deallocation by default

tungld:
This patch turns off `emitDealloc` in `convert-onnx-to-krnl` pass, and use MLIR `buffer-deallocation` pass to emit dealloc by default.

Once this works well, we will have another patch to completely remove dealloc code in `convert-onnx-to-krnl`.

Removed `test/mlir/onnx/onnx_lowering_with_dealloc.mlir` since it is not an useful test and no one has added a new test to it for a year or so.


AlexandreEichenberger (commented):
@tungld there are plenty of places where alloc/dealloc are used directly without the `insertAllocAndDeallocSimple` call.

We could also switch off in 'MemRefBuilder::dealloc', but even there I did not catch all the `create<memref::DeallocOp>`

What do you think?

tungld:
We can switch off all of them except Scan/RNN by using global gEmitDealloc in checkInsertDeallocation. We need to call checkInsertDealloction whenever we want to emit dealloc for an output buffer.

Scan and RNN ops dealloc for their intermediate buffers, so don't use checkInsertDeallocation. I removed dealloc in RNN ops, and will do the same for Scan.

tungld:
@AlexandreEichenberger I switched off deallocation in Scan op. Also make `memref:Dealloc` an illegal op when lowering onnx to krnl, so that we don't expect that users will emit dealloc at this level.

AlexandreEichenberger (approved):
LGTM

PR #864, opened 2021-09-17
Nested function call translation

gramalingam:
Extend the translation of ONNX functions to MLIR functions to support nested function calls. Generating the ONNX function-body (for a particular call) requires context information, which includes the onnx-types of all parameters. Add type-inference support to generate this information. This allows us to translate ONNX functions that contain calls to more ONNX functions.

Also fix some minor issues (such as checking whether the onnx-type is known or not).


AlexandreEichenberger (commented):
@chentong319 could you review? It is in the frontend.

@gramalingam would you be able to add a lit test and/or backend tests. Best to have new functionality tested, tx.

chentong319 (approved):
LGTM!
@AlexandreEichenberger `test/onnx2mlir/CustomFnTest.cpp` is used to test the implementation. The check-onnx-lit can  not test the frontend.

PR #869, opened 2021-09-21
Make identical symbol/dim index expression reuse the same id

AlexandreEichenberger:
The goal of this PR is to simplify the code in presence of deeply nested loops. For example, an index expression may be first define in an outer loop and its outer loop index expression scope, and then reused in an innermost loop. In the current approach, we had to redefine a variable
```mlir

   krnl.iterate(n ... ) {
     IndexExprScope outerScope;
    DimIndexExpr n(...)

   krnl.iterate () {
        IndexExprScope innerScope;

      SymbolIndexExpr n1(n);
    }
  }
```

We have to reactualizes the outer index expression `n` in the inner scope, and indicating that it is now a Symbol. Due to lambda issues, a new name has to be given, which is not ideal as forgetting to use `n1` in the inner scope will result in an assert at runtime.

So an easier pattern is to use `SymbolIndexExpr(n)` wherever `n` is needed in the inner scope. Before this PR, it would have created a new symbol and thus cluttered the affine expressions. With this PR, only one symbol (or dim) index expression is created per scope for a given value.


tungld (approved):
LGTM.

PR #870, opened 2021-09-21
Add krnl lowering with constant axes for Unsqueeze-13 and Squeeze-13

kwu91:
This PR follows in the same vein as #853. The previous PR (#829), while having less code changes, won't work with the new type additions to V13, namely BF16 support. Like the previous PR, these changes only support constant axes. Support for dynamic axes has been suggested [here](https://github.com/onnx/onnx-mlir/pull/829#issuecomment-904675618), but I currently do not have a timeline on adding that support. Feel free to take up that task if anyone desires.

Signed-off-by: Kevin Wu <6334443+kwu91@users.noreply.github.com>


src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2285,112 +2285,190 @@ LogicalResult ONNXPadOp::inferShapes(
 // Unsqueeze
 //===----------------------------------------------------------------------===//
 
-LogicalResult ONNXUnsqueezeOp::inferShapes(
-    std::function<void(mlir::Region &)> doShapeInference) {
-  return emitError("Not implemented");
+// Update axes attribute so that it contains only positive values.
+// Helper functions for both Unsqueeze and Squeeze Ops
+template <typename Op>
+void updateNegativeAxis(Op *op, ArrayRef<int64_t> axes) {
```
chentong319, line 2291: Do we really need to change the IR for axes if the shape info is stored in Index Expr?

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2285,112 +2285,190 @@ LogicalResult ONNXPadOp::inferShapes(
...
+  OpBuilder builder(op->getContext());
+  if (auto axesConstOp = getONNXConstantOp(op->axes())) {
+    auto tensorType = axesConstOp.getType().template cast<RankedTensorType>();
```
chentong319, line 2294: What's the difference in `template cast` and just `cast`? Should we change all the usage?

src/Dialect/ONNX/ONNXOpsHelper.cpp
```diff
@@ -342,6 +342,21 @@ bool AreTheSameAxisArray(int64_t rank, ArrayAttr lhsAttr, ArrayAttr rhsAttr) {
   return true;
 }
 
+/// Convert ConstantOp to ArrayAttr and test if they have the same values
+bool AreTheSameConstantOpDenseAttr(
+    Builder &builder, int64_t rank, Value lhsOp, Value rhsOp) {
+  ONNXConstantOp lhsConstOp = dyn_cast<ONNXConstantOp>(lhsOp.getDefiningOp());
+  ONNXConstantOp rhsConstOp = dyn_cast<ONNXConstantOp>(rhsOp.getDefiningOp());
+  if ((lhsConstOp && rhsConstOp) &&
+      (!isFromNone(lhsConstOp) && !isFromNone(rhsConstOp))) {
```
chentong319, line 351: Why 'isFromNone' is needed here? Should this function return true if both are `isFromNone`?

src/Dialect/ONNX/ONNXOps.cpp
kwu91: The original ShapeHelper functions for both Squeeze and Unsqueeze expect the axis values to be normalized during Shape inference such that there are no negative values. I lifted the block of code that updates the negative axis such that it's compatible with both ArrayAttr from V11 ops, and constant dense element attr from V13 ops.

src/Dialect/ONNX/ONNXOps.cpp
kwu91: `template cast` is needed here to resolve a compiler error because `updateNegativeAxis` is templatized.

src/Dialect/ONNX/ONNXOpsHelper.cpp
kwu91: Sorry, I was originally testing if `%cst = constant unit` was considered a ConstantOp and if so, test if it's also None. But the cast returns a null anyway, so I forgot to take that out. I will do so in my next push.

AlexandreEichenberger:
Clean run achieved.

src/Dialect/ONNX/ONNXOps.cpp
chentong319: For constant, negative value can be handled in ShapeHelper. For dynamic value, the compare and select is anyway needed in ShapeHelper.

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2285,112 +2285,190 @@ LogicalResult ONNXPadOp::inferShapes(
...
 }
 
+LogicalResult ONNXSqueezeOp::inferShapes(
+    std::function<void(mlir::Region &)> doShapeInference) {
+  auto builder = mlir::Builder(getContext());
+  llvm::Optional<ArrayAttr> optionalAttr;
+  if (auto axesConstOp = getONNXConstantOp(axes())) {
+    auto axesAttr = createArrayAttrFromConstantOp(builder, axesConstOp);
+    optionalAttr.emplace(axesAttr);
+  } else if (!axes().getType().isa<NoneType>()) {
```
chentong319, line 2457: For V13, the empty axes means to reduce every dim with size 1. onnx model generated with newer protobuf, empty axes is represented with Constant of 0 size. You can use `isFromNone(axes())` to check the condition.

src/Dialect/ONNX/ONNXOps.cpp
kwu91: I noticed this thread is marked as resolved, but do you want the handling of negative axes to be moved to ShapeHelper, or leave it as is at this stage? Thanks

chentong319: Just leave it as is.

src/Dialect/ONNX/ONNXOps.cpp
kwu91: I pushed a commit to handle empty axes for Squeeze. Let me know if it meets your expectations. Thanks!

tungld:

PR #871, opened 2021-09-23
Krnl builder cleanup

AlexandreEichenberger:
Simplified the interface for krnl.for, added check inside the handling of krnl.for that resulted in cryptic error otherwise, and added support for krnl if then else.

The old interface
```mlir
createKrnl.iterateIE({jj, kk}, {jj1, kk1}, {zero, zero}, {J, K}, {},
          [&](KrnlBuilder &createKrnl, ValueRange args) {
            ValueRange j1_k1_indices =
                createKrnl.getInductionVarValue({jj1, kk1});
          ...
  }
```
 and the new interface where the indices are calculated in the iterate and passed as parameters:

``` mlir
createKrnl.iterateIE({jj, kk}, {jj1, kk1}, {zero, zero}, {J, K},
          [&](KrnlBuilder &createKrnl, ValueRange j1_k1_indices) {   
   ...
}
```

This make the code less repetitive and remove a chance of errors (by passing the wrong loop indices to the `getInductionVarValues`.


tungld (approved):
LGTM. I guess we will use ifThenElse more in the future, we perhaps have a new place for it then.

AlexandreEichenberger:
Good point, indeed it uses only affine constructs, so maybe better there. Easy to migrate if need in the future

PR #873, opened 2021-09-25
Fix doxygen htmls for onnx.ai/onnx-mlir site

gongsu832:
- add doxygen action to publish docs and doxygen htmls
- clean up deprecated actions
- onnx.ai/onnx-mlir is published from the root of gh-pages branch


gongsu832:
@AlexandreEichenberger @caoimhinuibrian Under `Settings -> Pages` for the onnx-mlir repo, the `Source` needs to be changed to `Branch: gh-pages` and `/ (root)`, like the following,

image

AlexandreEichenberger:
@gongsu832 

Is there any way that we could build from the master branch? Info is more likely to get stale if it is not from master. Ideally, we would also add the dialects (KRNL, ONNX) where MLIR has a way to auto gen doc from the *.td files.

gongsu832:
The gh-pages branch is generated for the latest `docs` directory from the master. Because we are also dynamically generating htmls with doxygen (through the Travis CI before and GitHub Action now), we don't want to generate them into the master branch. Since all modifications to the master should be contained in the PR itself. The CI build should not modify master beyond what's in the PR. That's why by default GitHub Pages are served out of a branch such as gh-pages.

AlexandreEichenberger:
@gongsu832 thanks for the rational. Is there a way to automatically merge the latest master into that branch then?

gongsu832:
That's what the GitHub Action `.github/workflows/doxygen-docs-publish.yml` does, push master docs and doxygen generated htmls into gh-pages branch, which gets published to [https://onnx.ai/onnx-mlir](https://onnx.ai/onnx-mlir) automatically by GitHub.

AlexandreEichenberger (approved):
LGTM

AlexandreEichenberger:
@gongsu832 changed the settings/pages to gh-pages and root as requested. In my experience, it takes about 1h for the changes to be reflected on the website.

gongsu832:
@AlexandreEichenberger It was pretty quick during my tests, usually just a couple of minutes. I just tried, the site is already updated.

AlexandreEichenberger:
@gongsu832 I see it, nice. Would it be possible to also add a page for krnl ir?

AlexandreEichenberger:
no idea, but MLIR folks do it, so maybe some LLVM td to html translator?

gongsu832:
@AlexandreEichenberger I have no idea what `.td` files are but I observed that they get translated into `.cpp.inc` and `.hpp.inc` files. So doxygen can work on the .hpp.inc file I imagine. However, to produce meaningful documentation the generated `.hpp.inc` needs to have proper comment blocks like those in `include/onnx-mlir/Runtime/OMTensor.h`.

PR #877, opened 2021-09-29
Update RunONNXModel.py to run a model from its generated .so file

tungld:
This patch updates `RunONNXModel.py` to allow running from the generated .so file.
Two additional flags `-o` and `-s` are added.

We first compile the model and save the .so file by using flag `-o`:
`RunONNXModel.py model.onnx -o ./shared_lib.so`

Then run inference without re-compiling the model by using flag `-s`:
`RunONNXModel.py model.onnx -s ./shared_lib.so`

### DIFF
--- a/docs/DebuggingNumericalError.md
+++ b/docs/DebuggingNumericalError.md
@@ -32,15 +32,17 @@ reference inputs and outputs in protobuf.
 
 ```bash
 $ python ../utils/RunONNXModel.py  --help
-usage: RunONNXModel.py [-h] [--print_input] [--print_output] [--compile_args COMPILE_ARGS] [--shape_info SHAPE_INFO] [--verify {onnxruntime,ref}]
-                       [--ref_folder REF_FOLDER] [--rtol RTOL] [--atol ATOL]
+usage: RunONNXModel.py [-h] [-o PATH | -s PATH] [--print_input] [--print_output] [--compile_args COMPILE_ARGS] [--shape_info SHAPE_INFO]
+                       [--verify {onnxruntime,ref}] [--ref_folder REF_FOLDER] [--rtol RTOL] [--atol ATOL]
                        model_path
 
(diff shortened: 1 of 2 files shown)


AlexandreEichenberger (approved):
LGTM

chentong319:
Now I remember the other requirement. I have the input for a model and want to compile the model with the shape of that model. What I did is 1) add print shape of input tensor in the script 2) feed the shape with `--shape_info` to the script. 
It would be good to add a flag like '--use-input-shape' to do my manual work automatically. Or at least print out the shape of input tensor. Our compiled model provides function to return the input and output tensor info. Can we use that interface in python to make sure that compatible input is fed to the model?

tungld:
> It would be good to add a flag like '--use-input-shape' to do my manual work automatically. Or at least print out the shape of input tensor.

I see. got it. I'll update to do it automatically.

> Can we use that interface in python to make sure that compatible input is fed to the model?

I think, we can do this inside ExecutionSession so that all Python/C users benefit from that.

tungld:
@chentong319 

I just added a flag `--compile_using_input_shape` as you suggested.

Also, I added a flag `--save-data` to save the inputs and outputs in protobuf format, and then we can use them as reference data. It is useful for debugging by using fixed inputs for multiple runs.

```
# Run the model with random inputs, and save the random inputs as well as the outputs.
RunONNXModel.py model.onnx --save-data=./test_data

# Run the model with the saved data, no need to generate inputs again.
RunONNXModel.py model.onnx --data_folder=./test_data
```

AlexandreEichenberger:
@chentong319 If you are happy, please merge, otherwise state what is still missing, tx

tungld:
@chentong319 please let me know if you have any request. Thanks!

I have recently used this script when implementing a new op, and find it quite convenient to check the correctness.

PR #876, opened 2021-09-28
Add location for onnx ops 

chentong319:
Signed-off-by: Tong Chen <chentong@us.ibm.com>
This is the initial step to add location info for onnx ops.
We can add location info at different stage of compilation: after imported from onnx, after ONNX level optimization, after ONNX to Krnl and etc. Such flexibility can help us debugging and performance tracking for different purpose.
In this PR, I only implemented the method to add Location info after onnx model is imported.  I reused the `.input.mlir` file saved by keepFiles for "source code file". The line number is estimated and may be inaccurate.  It can be more precise to get the line number for interpreting the source code file (for future work). 
Usage: compile with flag `--BuilderLoc`.  For example, `Debug/bin/onnx-mlir --BuilderLoc test_add.onnx`. The result can be found in test_add.input.mlir.  You can also use flag `--preserveLocations` for onnx-mlir to dump with location info. For example,  compile with command`Debug/bin/onnx-mlir --BuilderLoc --preserveLocations test_add.onnx` and the test_add.input.mlir is list below:
```
#loc1 = loc("test_add.input.mlir":2:0)
builtin.module  {
  builtin.func @main_graph(%arg0: tensor<3x4x5xf32> loc("test_add.input.mlir":2:0), %arg1: tensor<3x4x5xf32> loc("test_add.input.mlir":2:0)) -> tensor<3x4x5xf32> attributes {input_names = ["x", "y"], output_names = ["sum"]} {
    %0 = "onnx.Add"(%arg0, %arg1) : (tensor<3x4x5xf32>, tensor<3x4x5xf32>) -> tensor<3x4x5xf32> loc(#loc2)
    return %0 : tensor<3x4x5xf32> loc(#loc3)
  } loc(#loc1)
  "onnx.EntryPoint"() {func = @main_graph, numInputs = 2 : i32, numOutputs = 1 : i32, signature = "[    { \22type\22 : \22f32\22 , \22dims\22 : [3 , 4 , 5] , \22name\22 : \22x\22 }\0A ,    { \22type\22 : \22f32\22 , \22dims\22 : [3 , 4 , 5] , \22name\22 : \22y\22 }\0A\0A]\00@[   { \22type\22 : \22f32\22 , \22dims\22 : [3 , 4 , 5] , \22name\22 : \22sum\22 }\0A\0A]\00"} : () -> () loc(#loc4)
} loc(#loc0)
#loc0 = loc(unknown)
#loc2 = loc("test_add.input.mlir":3:0)
[...]

PR #879, opened 2021-09-29
Revert "Add location for onnx ops  (#876)"

chentong319:
This reverts commit 968c360802309b51b3b6d27d91123dbf9a1ee7b5.

A better solution is found. We can change our main.c drive to use mlir parser to attach precise location info. Let's revert the previous PR first.


AlexandreEichenberger (approved):
Will trust you on this one.

PR #880, opened 2021-09-30
Added Instance norm

AlexandreEichenberger:
Added Instance Norm, including lit tests and found 2 previously unlisted backend tests.

I also updated the support for translating affine mlir files into C++ files, which let me add printf comments and compare values with python implementations listed in the backend test. That makes it really easy to detect errors. In this case, variance is computed as `(sum over all x of (x - mean)^2) )/ n` instead of `(....)/(n-1)` as seen in some places. Found out by just printing the variance values and comparing it with the backend tests verification code.

``` python
def instancenorm_test_mode(x, s, bias, epsilon=1e-5):  # type: ignore
    dims_x = len(x.shape)
    axis = tuple(range(2, dims_x))
    mean = np.mean(x, axis=axis, keepdims=True)
    print("mean is", mean)
    var = np.var(x, axis=axis, keepdims=True)
    print("var is", var)
    dim_ones = (1,) * (dims_x - 2)
    s = s.reshape(-1, *dim_ones)
    bias = bias.reshape(-1, *dim_ones)
    return s * (x - mean) / np.sqrt(var + epsilon) + bias
```


tungld (approved):
LGTM.

I would upgrade the code for BatchNorm using IndexExpr some time in the future to make the code style in file consistent.

AlexandreEichenberger:
> I would upgrade the code for BatchNorm using IndexExpr some time in the future to make the code style in file consistent.

@tungld good point.

PR #881, opened 2021-09-30
Implement ONNX NonZeroOp

tungld:
This patch implements NonZeroOp. 

This op's output shape is quite interesting, it depends on the values of the input, not the input shape. In particular, the second dim of the output shape is the number of nonzero values in the input.
So, the output shape would be dynamic and we should emit code traversing the input before allocating the output memref.

Fix a bug related to endian in `test.py`. When input is of boolean (`i1`), endianness is not relevant but `test.py` wrongly checks that it is in big endian and set output type to big endian, so tests failed on little endian machines x86, ppcle but not Z.

### DIFF
--- a/src/Conversion/ONNXToKrnl/Tensor/NonZero.cpp
+++ b/src/Conversion/ONNXToKrnl/Tensor/NonZero.cpp
@@ -48,7 +48,7 @@ struct ONNXNonZeroOpLowering : public ConversionPattern {
     // Bounds for iterating the input X.
     MemRefBoundsIndexCapture xBounds(X);
     SmallVector<IndexExpr, 4> lbs, ubs;
-    for (decltype(xRank) i = 0; i < xRank; ++i) {
+    for (int64_t i = 0; i < xRank; ++i) {
       lbs.emplace_back(litZero);
       ubs.emplace_back(xBounds.getDim(i));
     }
@@ -104,7 +104,7 @@ struct ONNXNonZeroOpLowering : public ConversionPattern {
           Value val = createKrnl.load(X, loopInd);
           Value eqCond = createMath.eq(val, zero);
           Value zeroOrOne = createMath.select(eqCond, iZero, iOne);
-          for (decltype(xRank) i = 0; i < xRank; ++i) {
+          for (int64_t i = 0; i < xRank; ++i) {
             SmallVector<IndexExpr, 1> posInd;
             posInd.emplace_back(LiteralIndexExpr(i));
             // Load the current position along dimension i.
--- a/src/Conversion/ONNXToKrnl/Tensor/NonZero.cpp
+++ b/src/Conversion/ONNXToKrnl/Tensor/NonZero.cpp
@@ -1,0 +1,138 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//===------------------- NonZero.cpp - Lowering NonZero Op ----------------===//
+//
+// Copyright 2019 The IBM Research Authors.
+//
+// =============================================================================
+//
+// This file lowers the ONNX NonZero Operator to Krnl dialect.
+//
+//===----------------------------------------------------------------------===//
+
+#include "src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.hpp"
+#include "src/Dialect/ONNX/ONNXShapeHelper.hpp"
+
+using namespace mlir;
+
+struct ONNXNonZeroOpLowering : public ConversionPattern {
+  ONNXNonZeroOpLowering(MLIRContext *ctx)
+      : ConversionPattern(mlir::ONNXNonZeroOp::getOperationName(), 1, ctx) {}
+
+  LogicalResult matchAndRewrite(Operation *op, ArrayRef<Value> operands,
+      ConversionPatternRewriter &rewriter) const final {
+    ONNXNonZeroOpAdaptor operandAdaptor(operands);
+    auto loc = op->getLoc();
+
+    // Builder helper.
+    KrnlBuilder createKrnl(rewriter, loc);
+    IndexExprScope outerScope(rewriter, loc);
+
+    // Common information.
+    Value X = operandAdaptor.X();
+    MemRefType xMemRefType = X.getType().cast<MemRefType>();
+    MemRefType resMemRefType = convertToMemRefType(*op->result_type_begin());
+    Type xElementType = xMemRefType.getElementType();
+    Type resElementType = resMemRefType.getElementType();
+    int64_t xRank = xMemRefType.getRank();
+
+    // Constant values.
+    Value iZero = emitConstantOp(rewriter, loc, rewriter.getIndexType(), 0);
+    Value iOne = emitConstantOp(rewriter, loc, rewriter.getIndexType(), 1);
+    Value zero = emitConstantOp(rewriter, loc, xElementType, 0);
+    LiteralIndexExpr litZero = LiteralIndexExpr(0);
+    LiteralIndexExpr litRank = LiteralIndexExpr(xRank);
+
+    // Bounds for iterating the input X.
+    MemRefBoundsIndexCapture xBounds(X);
+    SmallVector<IndexExpr, 4> lbs, ubs;
+    for (decltype(xRank) i = 0; i < xRank; ++i) {
+      lbs.emplace_back(litZero);
+      ubs.emplace_back(xBounds.getDim(i));
+    }
+
+    // Compute the number of nonzero values.
+    MemRefType i64MemRefType = MemRefType::get({}, rewriter.getIndexType());
+    Value nonzeroCount = insertAllocAndDealloc(
+        i64MemRefType, loc, rewriter, /*insertDealloc=*/true);
+    createKrnl.store(iZero, nonzeroCount, {});
+
+    // The result's first dimension size is the input's rank, so it is known.
+    // The result's second dimension size is the number of nonzero values in the
+    // input, and it is unknown at compile time. Thus, we will emit a
+    // krnl.iterate to count the number of nonzero values.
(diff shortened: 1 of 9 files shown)


src/Conversion/ONNXToKrnl/Tensor/NonZero.cpp
```diff
@@ -0,0 +1,138 @@
...
+    MemRefType resMemRefType = convertToMemRefType(*op->result_type_begin());
+    Type xElementType = xMemRefType.getElementType();
+    Type resElementType = resMemRefType.getElementType();
+    int64_t xRank = xMemRefType.getRank();
+
+    // Constant values.
+    Value iZero = emitConstantOp(rewriter, loc, rewriter.getIndexType(), 0);
+    Value iOne = emitConstantOp(rewriter, loc, rewriter.getIndexType(), 1);
+    Value zero = emitConstantOp(rewriter, loc, xElementType, 0);
+    LiteralIndexExpr litZero = LiteralIndexExpr(0);
```
AlexandreEichenberger, line 45: Nit: but this can simply be written as
```
LiteralIndexExpr litZero(0)
LiteralIndexExpr litOne(1)
```
 Optionally, iZero is also litZero.getValue() since they are of type index... but your code for that one is probably better.

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -3678,7 +3678,18 @@ LogicalResult ONNXNonMaxSuppressionOp::inferShapes(
 
 LogicalResult ONNXNonZeroOp::inferShapes(
     std::function<void(mlir::Region &)> doShapeInference) {
-  return emitError(NOT_IMPLEMENTED_MESSAGE);
+  auto builder = mlir::Builder(getContext());
+  Type inputType = getOperand().getType();
+  if (!inputType.isa<RankedTensorType>())
+    return success();
+  SmallVector<int64_t, 2> dims;
+  // The first dimension size is the rank of the input.
```
AlexandreEichenberger, line 3686: That seems to be the reverse: 1st dim is the number of nonzero... aka -1 here
2nd dim is the rank

``` python
condition = np.array([[1, 0], [1, 1]], dtype=bool)
result = np.array(np.nonzero(condition), dtype=np.int64)  # expected output [[0, 1, 1], [0, 0, 1]]
print(result, result.shape)
```
```
[[0 1 1]
 [0 0 1]] (2, 3)
```

src/Conversion/ONNXToKrnl/Tensor/NonZero.cpp
```diff
@@ -0,0 +1,138 @@
...
+            // Load the output at the current position along dimension i.
+            SmallVector<IndexExpr, 2> resInd;
+            resInd.emplace_back(LiteralIndexExpr(i));
+            resInd.emplace_back(DimIndexExpr(currentPos));
+            // If value is nonzero, update the output with the value's index.
+            // Otherwise, keep the output unchanged.
+            Value oldIndex = createKrnl.loadIE(resMemRef, resInd);
+            Value newIndex =
+                rewriter.create<IndexCastOp>(loc, loopInd[i], resElementType);
+            newIndex = createMath.select(eqCond, oldIndex, newIndex);
```
AlexandreEichenberger, line 121: Pretty smart, but could we possibly have an unconditional write after the last reserved position?
It might be a cause to use an affine if-then... I have a copy that moved it from affine (non-exportable) to krnl... listed here if useful to you.

KrnlHelper.hpp and cpp

``` cpp
  void ifThenElse(IndexExprScope &scope, SmallVectorImpl<IndexExpr> &conditions,
      function_ref<void(KrnlBuilder &)> thenFn,
      function_ref<void(KrnlBuilder &)> elseFn);
      
  private:
    void appendToBlock(Block *block, function_ref<void(ValueRange)> builderFn);
```

``` cpp
void KrnlBuilder::ifThenElse(IndexExprScope &scope,
    SmallVectorImpl<IndexExpr> &conditions,
    function_ref<void(KrnlBuilder &)> thenFn,
    function_ref<void(KrnlBuilder &)> elseFn) {
  int64_t rank = conditions.size();
  SmallVector<AffineExpr, 4> affineCond;
  bool allTrue = true;
  bool allFalse = true;
  for (IndexExpr c : conditions) {
    assert(c.isAffine() && "conditions expected to be affine");
    affineCond.emplace_back(c.getAffineExpr());
    if (c.isLiteral()) {
      // Condition can be evaluated at compile time, as it is a literal.
      if (c.getLiteral() < 0) // Inequality is expr >= 0, test if false.
        allTrue = false;
      if (c.getLiteral() >= 0) // Inequality is expr >= 0, test if true.
        allFalse = false;
    } else {
      // Condition cannot be evaluated at compile time
      allTrue = allFalse = false;
    }
  }
  SmallVector<bool, 4> isEq(rank, false);
  auto inset = IntegerSet::get(
      scope.getNumDims(), scope.getNumSymbols(), affineCond, isEq);
  SmallVector<Value, 8> dimAndSymbolList;
  scope.getDimAndSymbolList(dimAndSymbolList);
  auto ifOp = b.create<AffineIfOp>(loc, inset, dimAndSymbolList, true);
  Block *thenBlock = ifOp.getThenBlock();
  Block *elseBlock = ifOp.getElseBlock();
  if (!allFalse) {
    appendToBlock(thenBlock, [&](ValueRange args) {
      KrnlBuilder createKrnl(b, loc);
      thenFn(createKrnl);
    });
  }
  if (!allTrue) {
    appendToBlock(elseBlock, [&](ValueRange args) {
      KrnlBuilder createKrnl(b, loc);
      elseFn(createKrnl);
    });
  }
}

void KrnlBuilder::appendToBlock(
    Block *block, function_ref<void(ValueRange)> builderFn) {
  OpBuilder::InsertionGuard guard(b);
  if (block->empty() ||
      !block->back().mightHaveTrait<OpTrait::IsTerminator>()) {
    b.setInsertionPointToEnd(block);
  } else
    b.setInsertionPoint(&block->back());
  builderFn(block->getArguments());
}
```

AlexandreEichenberger:
@tungld In fact, I am pretty sure that the current algo read&write out of bound, which is a problem if we use parallelism (race condition) or we are unlucky (end of segment/page)

src/Dialect/ONNX/ONNXOps.cpp
tungld: No it's not the reverse, as you see (2, 3), 2 is the rank of inputs (2 dimensions) and 3 is the number of nonzeros

tungld:
@AlexandreEichenberger Just come up with a better algorithm:
- Given an input
```
2, 1
0, 2
0, 1
```
- Normalize it first to a 0-1 tensor where 1 means we have nonzero
```
1, 1
0, 1
0, 1
```

- ReduceSum for each dimension
```
dim0 = ReduceSum(axis = 1) = [2, 1, 1]
dim1 = ReduceSum(axis = 0) = [1, 3]
```

- Now we can compute output without any if-then-else, e.g. for dimension 0:
```python
k = 0
for i range(len(dim0)):
   for j in range(dim0[i]):
       out[0][k+j] = i
   k += d
```

Similar for the other dimensions.

What do you think?

src/Dialect/ONNX/ONNXOps.cpp
AlexandreEichenberger: I lifted the python code from their backend tests... But I agree, it is 3 entries of 2 dim each.

AlexandreEichenberger:
Pretty smart, that would work. I  would avoid computing the binary matrix (more alloc/free) by simply creating the dim0 & dim1 reduction sum directly from the a[i,j] != 0 test that you would have used to store 0 or 1 in the binary matrix. In addition, if you keep the running sum of dim0 too, that will give you the size of the output result to allocate as well.

src/Conversion/ONNXToKrnl/Tensor/NonZero.cpp
tungld: Yes, `LiteralIndexExpr litZero(0)` is definitely better.

tungld:
> I would avoid computing the binary matrix (more alloc/free) by simply creating the dim0 & dim1 reduction sum directly from the a[i,j] != 0 test that you would have used to store 0 or 1 in the binary matrix. In addition, if you keep the running sum of dim0 too, that will give you the size of the output result to allocate as well.

Yes, all these are computed in a single nested loop traversing the input only once.

```python
k = 0
for i range(len(rsum0)):
   d = rsum0[i]
   for j in range(d):
       out[0][k+j] = i
   k += d
```
In face the above loop is not 'affine.for' friendly. The inner iterate's bound is from the outer iterate's operand, and it is invalid for affine.for as stated here: https://github.com/llvm/llvm-project/blob/main/mlir/test/Dialect/Affine/invalid.mlir

So, I have to change to this loop iterating over the output's 2nd dimension (nonzerocount):
```python
for i in range(nonzerocount):
    s = 0, p = -1
    for j in range(len(rsum0)):
        s += rsum0[j]
        p = (i < s and p == -1) ? j : p
    out[0][i] = p

```

src/Dialect/ONNX/MLIRDialectBuilder.cpp
```diff
@@ -65,6 +65,12 @@ Value MathBuilder::slt(Value lhs, Value rhs) {
     return b.create<CmpIOp>(loc, CmpIPredicate::slt, lhs, rhs);
   return b.create<CmpFOp>(loc, CmpFPredicate::OLT, lhs, rhs);
 }
+Value MathBuilder::eq(Value lhs, Value rhs) {
```
AlexandreEichenberger, line 68: Thanks for adding this one

AlexandreEichenberger (approved):
LGTM, nice algorithm, fun to think outside the box from time to time.

PR #882, opened 2021-09-30
create source code for binary

chentong319:
The onnx model comes without user friendly source file and our importer assign `UnknownLoc` to all the operations. As a result, we do not have Location info in our transformation and binary. 
This PR will use the output of preserveMLIR as source file and get Location info by rebuilding operations from this file.
I just reused the option `--preserveMLIR`.
Document is added in `docs/Testing.md`.


AlexandreEichenberger (approved):
LGTM

PR #884, opened 2021-10-01
Memset

AlexandreEichenberger:
Added memset krnl operation, used it in GEMM and Matmul, will let us eventually optimize this op, possibly with library call to optimized versions.

This PR also add a convenient `createAffine.forIE()` method where we can pass lists of loops. It makes them perfectly nested.

### DIFF
--- a/src/Conversion/KrnlToAffine/KrnlToAffine.cpp
+++ b/src/Conversion/KrnlToAffine/KrnlToAffine.cpp
@@ -290,6 +290,15 @@ struct AffineBuilder : DialectBuilder {
         });
   }
 
+  void forIE(SmallVectorImpl<IndexExpr> &lbs, SmallVectorImpl<IndexExpr> &ubs,
+      SmallVectorImpl<int64_t> &steps,
+      function_ref<void(AffineBuilder &, ValueRange)> builderFn) {
+    assert(lbs.size() == ubs.size() && "expected identical sizes");
+    assert(lbs.size() == steps.size() && "expected identical sizes");
+    SmallVector<Value> loopIndices;
+    recursionForIE(lbs, ubs, steps, loopIndices, builderFn);
+  }
+
   // This if then else construct has no arguments to the blocks.
   void ifThenElse(IndexExprScope &scope, SmallVectorImpl<IndexExpr> &conditions,
       function_ref<void(AffineBuilder &createAffine)> thenFn,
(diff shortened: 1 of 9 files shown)


src/Conversion/KrnlToAffine/KrnlToAffine.cpp
```diff
@@ -335,6 +344,27 @@ struct AffineBuilder : DialectBuilder {
...
+  // Support for multiple forIE loops.
+  void recursionForIE(SmallVectorImpl<IndexExpr> &lbs,
+      SmallVectorImpl<IndexExpr> &ubs, SmallVectorImpl<int64_t> &steps,
+      SmallVectorImpl<Value> &loopIndices,
+      function_ref<void(AffineBuilder &, ValueRange)> builderFn) {
+    int d = loopIndices.size();
+    if (d < (int)lbs.size()) {
+      // Issue a loop and recurse again.
+      forIE(
+          lbs[d], ubs[d], steps[d], [&](AffineBuilder &createAffine, Value i) {
```
tungld, line 356: We should make sure the sizes of `lbs`, `usb`, `step and `loopIndices` are all equal in advance. Just in case one of the values is missing.

tungld (changes requested):

src/Conversion/KrnlToAffine/KrnlToAffine.cpp
AlexandreEichenberger: @tungld The check is done above on line 296 & 297, a function that corresponds to the public interface. This one is called once the check passes, and was needed to handle the recursion (arbitrary number of loop depth). So I think it should be all good

tungld: I see. You checked them outside. Thanks!

tungld (approved):
LGTM. 

`krnl.memset` is quite useful.

AlexandreEichenberger:
> krnl.memset is quite useful.

Thanks, and my hope is that we can connect it to an efficient implementation in LLVM, at least for the value=0 case.

chentong319 (commented):
Basically, I am not familiar with our design of Krnl with Affine dialect. To my understanding, this PR contains two components: ForIE and KrnlMemsetOp. 
1. Is the recursive function for ForIE is really needed? Can a loop work?
2. Do we have some utility to iterate memref with ForIE when a constraint or map is provided?
3. To me, the intention of the PR  could be provide some utility (optionally in krnl) to easily construct memref operations. for example: (memref<10x20xf32>: with constraint  2, 4 <= I,j < 6, 7 ) =  memref<4x3xf32>. 
4. The name memset somehow made me think of c library memset. Is this intentional? Should we check the element type of memref and the type of Value? Not clear about our rule about common practice.

AlexandreEichenberger:
@chentong319 thanks for your feedback.

> Is the recursive function for ForIE is really needed? Can a loop work?

We are working with lambda functions, and I am building on that support. This forced me to do recursion. Potentially, we could build a loop with a lambda function just at the end if I was not reusing existing infrastructure based on lambda. Took the modular approach to reuse existing infrastructure.

> To me, the intention of the PR could be provide some utility (optionally in krnl) to easily construct memref operations. for example: (memref<10x20xf32>: with constraint 2, 4 <= I,j < 6, 7 ) = memref<4x3xf32>.

At this stage, the LBs/UBs on the loop drive the iterations. So you could simply say
```C++
createAffine.forIE({2,4}, {6, 7}, {1, 1}, [&](AffineBuilder &createAffine, ValueRange iter) {
  createAffine.store(val, memref, iter)'
}
```

> The name memset somehow made me think of c library memset. Is this intentional? Should we check the element type of memref and the type of Value? Not clear about our rule about common practice

yes, the idea is to eventually lower memset to an optimized LLVM call. Did not want to risk adding additional dependences (think it comes from string.h). Would ideally have an implementation that is embedded in LLVM or stdlib

PR #885, opened 2021-10-02
Add OneHot operation shape inference and lowering to Krnl

shizy818:
Signed-off-by: ZHEN YU SHI <shizy@cn.ibm.com>


AlexandreEichenberger:
FYI, the error is

```
Debug/test.py:1014: RuntimeError
----------------------------- Captured stdout call -----------------------------
test_onehot_negative_indices
Failed /workdir/onnx-mlir/build/Debug/bin/onnx-mlir: test_onehot_negative_indices
depth shape must be 0 or if 1, size must be 1
UNREACHABLE executed at /workdir/onnx-mlir/src/Conversion/ONNXToKrnl/Tensor/OneHot.cpp:55!
______________ OnnxBackendNodeModelTest.test_onehot_with_axis_cpu ______________
[gw2] linux -- Python 3.8.10 /usr/bin/python3.8
test_self = <test.OnnxBackendNodeModelTest testMethod=test_onehot_with_axis_cpu>
device = 'CPU'
```

It passes on window and mac, but not on Linux/x86 as well as power and z. I would be looking at what's up with the Linux x86, which can be run in a docker container if you don't have access to one such machine. See instructions in the readme. Easiest might be to pick a working version, apply your patch, and see what's up.

AlexandreEichenberger:
@shizy818 Please do a manual merge as there appears to be conflicts.

src/Conversion/ONNXToKrnl/Tensor/OneHot.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+    Value indices = operandAdaptor.indices();
+    Value depth = operandAdaptor.depth();
+    Value values = operandAdaptor.values();
+
+    auto depthShape = depth.getType().cast<MemRefType>().getShape();
+    auto indicesShape = indices.getType().cast<MemRefType>().getShape();
+    int64_t indicesRank = indicesShape.size();
+
+    // Axis is a required attribute and should have default value of -1.
```
chentong319, line 30: More precisely, axis is an attribute with default value. You are right to use axis() to get the integer value, instead of using aixsAttr() to get the attribute.

src/Conversion/ONNXToKrnl/Tensor/OneHot.cpp
```diff
@@ -0,0 +1,189 @@
...
+    // Allocate result.
+    Value alloc;
+    Value zero = emitConstantOp(rewriter, loc, rewriter.getIndexType(), 0);
+    Value one = emitConstantOp(rewriter, loc, rewriter.getIndexType(), 1);
+
+    // Load values depending on shape.
+    Value loadedDepth;
+    if (depthShape.size() == 0)
+      loadedDepth = rewriter.create<KrnlLoadOp>(loc, depth, ArrayRef<Value>{});
+    else if (depthShape.size() == 1 && depthShape[0] == 1)
```
chentong319, line 52: When depth is a scalar as specified, depthShape.size() should be 0. Some previous code may not be correct.

src/Conversion/ONNXToKrnl/Tensor/OneHot.cpp
```diff
@@ -0,0 +1,189 @@
...
+      loadedDepth = rewriter.create<KrnlLoadOp>(loc, depth, zero);
+    else
+      llvm_unreachable("depth shape must be 0 or if 1, size must be 1");
+
+    // convert float type to int64 for depth value
+    Type depthElmentType =
+        depth.getType().dyn_cast<MemRefType>().getElementType();
+
+    TypeSwitch<Type>(depthElmentType)
```
chentong319, line 61: The code for depth is correct but could be improved. The IE is supposed to be able to optimize the case when depth is a constant. Since there is a type cast here, I am not sure. Best we can get precise shape when depth is constant. In ONNXShapeHelper:1510 line, aixDim is given a ? without any attempt to optimize. Is that right?
We do not have to optimize it if it is too difficult.

shizy818:
Thank you all. I'm going to check. It was Chinese National vacation and I was out for a couple of days.

src/Conversion/ONNXToKrnl/Tensor/OneHot.cpp
shizy818: OK. So I remove `depthShape.size() == 1 && depthShape[0] == 1` condition. I copy this code from Range. I follow the rule to accept both scalar value or a tensor of length 1.

src/Conversion/ONNXToKrnl/Tensor/OneHot.cpp
shizy818: In ONNXShapeHelper, aixDim is given a `?` first and then assign depth value if it is constant. So we can infer shape when depth is constant.
```
%depth = "onnx.Constant"() {value = dense<10> : tensor<i64>} : () -> tensor<i64>
%0 = "onnx.OneHot"(%arg0, %depth, %arg1) : (tensor<2x2xi64>, tensor<i64>, tensor<2xf32>) -> tensor<*xf32>
  return %0 : tensor<*xf32>
.....
// CHECK: {{.*}} = "onnx.OneHot"(%arg0, [[R0]], %arg1) {axis = 2 : si64} : (tensor<2x2xi64>, tensor<i64>, tensor<2xf32>) -> tensor<2x2x10xf32>
```

And we get an unknown dimension when depth needs a input
```
%0 = "onnx.OneHot"(%arg0, %arg1, %arg2) : (tensor<2x2xi64>, tensor<i64>, tensor<2xf32>) -> tensor<*xf32>
  return %0 : tensor<*xf32>
......
// CHECK: {{.*}} = "onnx.OneHot"(%arg0, %arg1, %arg2)  {axis = 2 : si64} : (tensor<2x2xi64>, tensor<i64>, tensor<2xf32>) -> tensor<2x2x?xf32>
```

Do you mean that during ONNXOneHotOp lowering, we do not need type cast here if depth is already a constant?

shizy818: I see your following comment...

src/Conversion/ONNXToKrnl/Tensor/OneHot.cpp
chentong319: You may change the comment "required attribute". Not a big deal.

shizy818:
@AlexandreEichenberger I merge latest update. Thanks!

PR #886, opened 2021-10-04
Support ONNX ModOp and MeanOp

tungld:
This patch implements ONNX ModOp and MeanOp.

For ModOp, since MLIR integers are signless, so it looks impossible to support signed and unsigned integer remainders as required by ONNX.

### DIFF
--- a/src/Dialect/ONNX/ONNXOps.cpp
+++ b/src/Dialect/ONNX/ONNXOps.cpp
@@ -3653,7 +3653,17 @@ LogicalResult ONNXMaxUnpoolOp::inferShapes(
 
 LogicalResult ONNXMeanOp::inferShapes(
     std::function<void(mlir::Region &)> doShapeInference) {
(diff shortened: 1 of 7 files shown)


src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -3653,17 +3653,48 @@ LogicalResult ONNXMaxUnpoolOp::inferShapes(
...
+  getResult().setType(resultTy);
+  return success();
 }
 
 LogicalResult ONNXMeanVarianceNormalizationOp::inferShapes(
     std::function<void(mlir::Region &)> doShapeInference) {
   return emitError(NOT_IMPLEMENTED_MESSAGE);
 }
 
+static LogicalResult verify(ONNXModOp op) {
```
AlexandreEichenberger, line 3674: Thanks for using the verify... it will make our code stronger in the long term

AlexandreEichenberger (approved):
LGTM

PR #887, opened 2021-10-04
Rewrite ONNX Upsample into ONNX Resize

tungld:
ONNX Upsample is deprecated and can be replaced by ONNX Resize, so we rewrite Upsample into Resize. 

This patch supports all Upsample versions 10, 9, 7.


AlexandreEichenberger (approved):
LGTM

PR #889, opened 2021-10-04
Modernization of Shape Op

AlexandreEichenberger:
Using shape inference, because it will be useful to ExpandOp. 

There is also a refinement for `ONNXOpBroadcastedShapeHelper::Compute` which is currently only used when lowering from ONNX to Krnl, a phase during which there is no more `question mark` type for index expressions (they are only present in shape inference where we cannot gen code). By using `! isLiteral` instead, we now get what was the true intent of the code. 

In addition, there is a new parameter for this broadcast compute to accommodate what is needed for ExpandOp.

### DIFF
--- a/docs/DebuggingNumericalError.md
+++ b/docs/DebuggingNumericalError.md
@@ -32,25 +32,45 @@ reference inputs and outputs in protobuf.
 
 ```bash
 $ python ../utils/RunONNXModel.py  --help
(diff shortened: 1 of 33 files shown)


AlexandreEichenberger:
Keep in mind that this PR is build on top of the memset #884, so you see its changes here too. This will disappear once memset is merged in (which should be soon).

tungld (approved):
LGTM.

PR #890, opened 2021-10-05
Rewrite Pad v2 into the latest Pad version

tungld:
Some models in the ONNX model zoo use Pad version 2. We rewrite Pad v2 into the latest Pad version. 

Edited get_onnx_mIir.py so that the special handler function is applied to the latest Pad version only. Before this patch, it applied to every version of an op.

Re-formated rules for Upsample in Decompose.td without any code change.


AlexandreEichenberger (approved):
LGTM, thanks

PR #892, opened 2021-10-05
Support WhereOp

tungld:
This patch implements ONNX WhereOp.

It also fixed an endianness issue in `test.py` happened with WhereOp where there was a mix of boolean and non-boolean types in inputs.


AlexandreEichenberger (approved):
LGTM

PR #893, opened 2021-10-05
Add scalar support for GatherOp

chentong319:
onnx-mlir failed to compile yolov5 because there is GatherOp with scalar output. This PR adds a support for the special case. But not a general solution for other ops. General solution will be discussed in issue#872.


chentong319:
replaced with a general solution

PR #894, opened 2021-10-05
Create a shared lib with C API

MikeHolman:
This change adds a new shared library om with a C API to use as an alternative to the executable.

This is motivated by the desire to avoid having to serialize onnx graphs to disk before calling onnx-mlir.

The library currently only has two methods: CompileFromFile and CompileFromArray, where CompileFromFile acts in roughly the same way as onnx-mlir.exe does today, compiling from a file containing MLIR or an ONNX protobuf. CompileFromArray compiles from an in-memory ONNX protobuf array.


AlexandreEichenberger:
I asked @gongsu832 to review it as he has experience with our interfaces, and @chentong319 because he does a lot of our front-end work.

I have a high level question: are you planning to have a test of this new interface? If we are adding new functionality, it would be great to ensure that it works through our CIs.

gongsu832:
A few comments.

- I'm not sure we need another lib for this. Can't the functions be part of `MainUtils`?
- None of the other sources uses dynamically generated export headers so using it for just this lib doesn't really buy much and makes the mixed-style source harder to maintain. I'm also in general wary of dynamically generated code as they typically contain a lot of dead code that never get exercised and are typically not for human to read.
- The naming for the targets (e.g., `om`), libraries (e.g., `OMLibrary`), etc. is rather arbitrary. They need to be more meaningful and specific. They should also follow existing naming style. For example, for files under the src directory, we don't prefix them with `OM`.

MikeHolman:
The reason to have a shared library is because we have a consumer team who uses onnx-mlir.exe and adding onnx-mlir as a build dependency is prohibitively expensive. With a shared library we can simply give them the dll and not require them to add onnx-mlir and all its dependencies (in particular LLVM/MLIR) to their build.

I've removed the dynamic export header.

I didn't think the names were arbitrary, this is meant as a general purpose alternative to onnx-mlir, so I kept the name general. But in any case I've renamed and moved the files to be more clear that this is specifically a compiler API. Let me know what you think of the changes.

gongsu832:

caoimhinuibrian:

gongsu832:
@MikeHolman Yes I think this looks much better. We can keep them separate for now and make further improvements and decide whether they should be merged or not according to how things evolve. Thanks.

gongsu832 (approved):
LGTM

src/Compiler/OnnxMlirCompiler.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include "CompilerUtils.hpp"
+
+extern "C" {
+namespace onnx_mlir {
+ONNX_MLIR_EXPORT int CompileFromFile(const char *inputFilename,
+    const char *outputBaseName, EmissionTargetType emissionTarget,
+    const char *mcpu, const char *mtriple) {
+  mlir::OwningModuleRef module;
+  mlir::MLIRContext context;
+  std::string mcpuString;
```
AlexandreEichenberger, line 15: Suggestion: could we have the code that is in common between the two CompileFromFile and CompileFromArray in a common function, say SetCompileContext, handling them cpu, mtriple, dialect registration?

include/OnnxMlirCompiler.h
```diff
@@ -0,0 +1,69 @@
...
+/*!
+ *  Compile an onnx model from a file containing MLIR or ONNX protobuf
+ *  @param inputFilename File name pointing onnx model protobuf or MLIR
+ *  @param outputBaseName File name without extension to write output
+ *  @param emissionTarget Target format to compile to
+ *  @param mcpu Optional target CPU
+ *  @param mtriple Optional target architecture
+ *  @return 0 on success or non-zero error code on failure
+ */
+ONNX_MLIR_EXPORT int CompileFromFile(const char *inputFilename,
```
AlexandreEichenberger, line 45: All of the names in the ONNX MLIR runtime are prefaced with "om". I suggest we do the same here.

src/Compiler/OnnxMlirCompiler.cpp
```diff
@@ -0,0 +1,45 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "OnnxMlirCompiler.h"
+#include "CompilerUtils.hpp"
+
+void SetCompileContext(
```
AlexandreEichenberger, line 8: You can check elsewhere, but we are also trying to use small cap for the function, and we like them to be verbs, so `setCompileContext` would be in the spirit of things

AlexandreEichenberger:

PR #900, opened 2021-10-07
[NFI]: Fix clang warnings.

etiotto:
Fix several clang warnings (mark 'unused' functions that aren't yet used in the code, add some const correctness, remove assignments to dead variables, use constexpr for static variables, use const_cast where appropriate,...).


src/Support/Common.hpp
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+#if defined(__GNUC__) || defined(__clang__)
+#define ATTRIBUTE(x) __attribute__((x))
+#else
+#define ATTRIBUTE(x)
```
caoimhinuibrian, line 6: is there something equivalent to __attribute__ in Microsoft C++?

PR #901, opened 2021-10-08
Support ReverseSequence

chentong319:
1. Shape inference: the output tensor has the same shape of the input()
2. Lowering to Krnl. I chose a simple loop structure. I do not like to generate complicated loops for performance, especially for non-critical operations. The optimization should be done by compiler automatically or at least guided by directives. Here is the details
```
      The semantic of Reversequence can be expressed in loop as:
      for (vector of dim I on the shape of output tensor) {
        vector of dim Iinput = I;
        Iinput[time_axis()]  = I[time_axis()] < seqence_lens[I[batch_axis()]]?
              sequence_lens[I[batch_axis()]]-I[time_axis()]-1:I[time_axis()];
        output[I] = input[Iinput]
      }
```
The condition check (affine expression) can be moved out of some loops and perform loop split on time_axis() loop. The result should have better performance on itself, but may hinder loop fusion. Loop fission or loop interchanging can be applied here.
```
     I is vector of dim on the shape of output tensor;
      for ( I except I[batch_axis()]) {
        // Reverse 
        for (I[time_axis() :  0 to sequence_lens[I[batch_axis]]) {
            Iinput = I;
            Iinput[time_axis()] = sequence_lens[I[batch_axis()]]-I[time_axis()]-1
            output[I] = input[Iinput];
          }
        // Copy
        for (I[time_axis() from sequence_lens[I[batch_axis]] to end) {
              output[I] = input[I];
        }
    }
```
3. Add test cases in lit test.
4. Enable the two related backend test cases.


src/Conversion/ONNXToKrnl/Tensor/ReverseSequence.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+      This transformation should improve performance on this loop nest itself,
+      but may hinder loop fusion with other loop nestes.
+      Also further loop fission or loop interchanging can be applied here.
+      I chose the simple loop structure
+      and believe optimization should be left to compiler, at least for
+      non-critical ops.
+    */
+
+    // Define loops and iteration trip counts (equivalent to size of output)
```
AlexandreEichenberger, line 89: While there is nothing wrong with using the old interface, the new lambda based version is so much easier to read, I would strongly suggest to use it from now on. Ideally we would also update the old versions, but if we could generate the new ones using the new scheme, that would be great. If you believe something is preventing you, I would be happy to add the necessary support.

chentong319: Any documentation for the new lambda approach?

PR #902, opened 2021-10-08
Shape Inference code for the SpaceToDepth operator

etiotto:
The SpaceToDepth operator accepts an input tensor and has a blocksize attribute. The operator reshapes the input tensor according to the blocksize attribute. Specifically, given an input tensor of dimension [N,C,H,W] (where N is the batch axis, C is the channel or depth, H is the height and W is the width) and a blocksize, the output tensor will have shape [N, C * blocksize * blocksize,H / blocksize,W / blocksize].

This PR implements the shape inference code for the operator.

### DIFF
--- a/src/Dialect/ONNX/ONNXShapeHelper.cpp
+++ b/src/Dialect/ONNX/ONNXShapeHelper.cpp
@@ -1371,10 +1371,6 @@ LogicalResult ONNXReshapeOpShapeHelper::Compute(
 template <typename ShapeHelper, typename OperandAdaptor>
 LogicalResult ONNXSqueezeOpShapeHelperCommon(ShapeHelper *shapeHelper,
     OperandAdaptor operandAdaptor, ArrayRef<IndexExpr> indexExprArray) {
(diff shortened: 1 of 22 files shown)


src/Dialect/ONNX/ONNXShapeHelper.cpp
```diff
@@ -36,6 +36,7 @@ ONNXOpShapeHelper<OP>::ONNXOpShapeHelper(OP *newOp)
     : op(newOp), scope(nullptr, newOp->getLoc()),
       fGetDenseVal(getDenseElementAttributeFromONNXValue), fLoadVal(nullptr),
       outputsDims() {
+  assert(op && "Expecting a valid pointer");
```
chentong319, line 39: Do we need to check whether the pointer is null? A safer programming practice?

etiotto: My understanding is that `op` should never be `nullptr`.

PR #904, opened 2021-10-11
Set splat for DenseElementAttr when all dimensions are 1

tungld:
This patch fixes a bug in constant propagation when creating DenseElementAttr for tensors whose all dimension sizes are 1.


AlexandreEichenberger (approved):
LGTM

PR #908, opened 2021-10-12
Changed verify methods to use shaped type with known rank

AlexandreEichenberger:
Previous version only handled ranked tensor types, but members are not included in it, as indicated by @tungld . New verifier methods now use shaped type that have a known rank.


tungld (approved):
LGTM. Thanks!

PR #909, opened 2021-10-12
[NFC]: Modernize code to follow C++11 clang-tidy recommendations

etiotto:
I have installed and run `clang-tidy` with the following configuration:

```
    "clang-tidy.checks": [
        "cppcoreguidelines-*",
        "modernize-*",
        "performance-*",
        "-modernize-use-trailing-return-type",
        "-cppcoreguidelines-init-variables",
        "-cppcoreguidelines-avoid-non-const-global-variables",
        "-cppcoreguidelines-non-private-member-variables-in-classes",
        "-cppcoreguidelines-pro-type-member-init"
    ],
```

The tool highlights several problems with the code base which this patch is fixing. Namely:
- uninitialized variables
- use of typedefs instead of alias declaration (via `using` keyword)
- classes that have declared a custom constructor but have not declared move constructor and move `operator=` ...
- use of `define` to declare names associated with a constant (should use `static constexpr` instead)
- etc...

This patch also introduces the use of the standard LLVM facility for code tracing (i.e. the `LLVM_DEBUG` macro)

### DIFF
--- a/src/Builder/SymbolTable.hpp
+++ b/src/Builder/SymbolTable.hpp
@@ -22,8 +22,8 @@ struct VariableScope {
    * Create a variable scope.
    * @param identifier name of the variable scope.
    */
-  explicit VariableScope<T>(std::string identifier)
-      : identifier(std::move(identifier)){};
+  explicit VariableScope<T>(std::string _identifier)
+      : identifier(std::move(_identifier)), nameToValue(){};
 
   /*!
    * Record a symbol name value correspondence.
@@ -47,16 +47,18 @@ struct VariableScope {
    */
   bool contain(const std::string &name) const;
 
+  const std::string &getIdentifier() const { return identifier; }
+
+private:
   /*!
    * Identifier of the current scope, used for debugging and sanity check.
    */
   const std::string identifier;
 
-private:
   /*!
    * A mapping between symbol name and symbol value.
    */
-  std::map<std::string, T> _nameToValue;
+  std::map<std::string, T> nameToValue;
 };
 
 /*!
@@ -64,13 +66,14 @@ private:
  */
 template <typename T>
 struct SymbolMapping {
+  SymbolMapping() = default;
 
   /*!
    *  Get MLIR tensor by onnx tensor name.
    *  @param name onnx tensor name.
    *  @return onnx mlir tensor corresponding to `name`.
    */
-  T GetTensorByOnnxName(const std::string &name);
+  T GetTensorByOnnxName(const std::string &name) const;
 
   /*!
    *  Add a new mapping from onnx tensor name to MLIR symbol.
@@ -84,7 +87,7 @@ struct SymbolMapping {
    * @param name symbol name.
    * @return whether a symbol with the sepcified name exists.
    */
-  bool ContainKey(const std::string &name);
+  bool ContainKey(const std::string &name) const;
 
   /*!
    * Push a new variable scope with a specified identifier to symbol table.
@@ -104,7 +107,7 @@ private:
   /*!
    *  A list of variable scope, ordered from outermost to innermost.
    */
-  std::vector<VariableScope<T>> _scopes;
+  std::vector<VariableScope<T>> scopes;
 };
 
 /**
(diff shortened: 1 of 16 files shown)


src/Builder/SymbolTable.hpp
```diff
@@ -47,30 +47,33 @@ struct VariableScope {
...
    * Identifier of the current scope, used for debugging and sanity check.
    */
   const std::string identifier;
 
-private:
   /*!
    * A mapping between symbol name and symbol value.
    */
-  std::map<std::string, T> _nameToValue;
+  std::map<std::string, T> nameToValue;
```
AlexandreEichenberger, line 61: Can you clarify on the proper usage of "_", I see that you added in some places, and removed in others.

src/Conversion/ONNXToKrnl/ConvertONNXToKrnl.cpp
```diff
@@ -55,7 +55,7 @@ struct FrontendToKrnlLoweringPass
   // Make sure that we have a valid default constructor and copy
   // constructor to make sure that the options are initialized properly.
   FrontendToKrnlLoweringPass() = default;
-  FrontendToKrnlLoweringPass(const FrontendToKrnlLoweringPass &pass) {}
+  FrontendToKrnlLoweringPass(const FrontendToKrnlLoweringPass &pass){};
```
AlexandreEichenberger, line 58: That one is strange to me, though the ";" was needed for declarations, and was not needed when giving the code.

src/Dialect/Krnl/KrnlHelper.hpp
```diff
@@ -154,12 +154,19 @@ class BuildKrnlLoop {
...
+  BuildKrnlLoop(BuildKrnlLoop &&) = delete;
+  BuildKrnlLoop &operator=(const BuildKrnlLoop &) = delete;
+  BuildKrnlLoop &operator=(BuildKrnlLoop &&) = delete;
 
   // Create kernel loop builder for a loop nest of depth equal to the
   // dimensionality of the operand. An operand of MemRef type is requied.
   BuildKrnlLoop(
       ConversionPatternRewriter &rewriter, Location loc, Value memRefOperand);
-  ~BuildKrnlLoop();
+  virtual ~BuildKrnlLoop() {
```
AlexandreEichenberger, line 166: Confused by the virtual, esp since there appears to be no subclasses.

src/Dialect/ONNX/IndexExpr.hpp
```diff
@@ -399,10 +402,15 @@ class IndexExpr {
 public:
   // Default and shallow copy constructors. Index expressions are usually built
   // using the subclasses listed as friend above.
-  IndexExpr() : indexExprObj(nullptr) {} // Undefined index expression.
+  IndexExpr() = default;
```
AlexandreEichenberger, line 405: I really need `indexExprObj` to be set to nullptr. Is that the default?

etiotto: Yes it is because at line 557  I initialized it explicitly to `nullptr`.

src/Builder/SymbolTable.hpp
AlexandreEichenberger: got it. I personally use the same name as it works (constructor parameter and class/struct variable) but I can see that using a "_" may be clearer.

src/Dialect/Krnl/KrnlHelper.hpp
AlexandreEichenberger: I have been trained to shiver at the sight of a `virtual`... esp for cases where execution time is paramount. It is less important here, I would go with the final then.

src/Dialect/ONNX/IndexExpr.hpp
AlexandreEichenberger: Oh, so you prefer putting the default value in the declaration than the constructor? Is that because when it is set on the declaration, it then apply by default to all of the constructors? I appreciate your input as I would like to understand the suggested changes... which are for the best if that is what is actually happening.

AlexandreEichenberger (approved):
LGTM, please continue to respond to questions as it will help us develop better coding habits.

PR #910, opened 2021-10-12
Decompose DepthToSpace operator.

etiotto:
This PR decomposes the ONNX DepthToSpace operator into  reshape+transpose+reshape as documented by https://github.com/onnx/onnx/blob/master/docs/Operators.md#DepthToSpace


src/Dialect/ONNX/ONNXOps.td.inc
```diff
@@ -821,7 +821,7 @@ def ONNXConvTransposeOp:ONNX_Op<"ConvTranspose",
   "output_shape can also be explicitly specified in which case pads values are auto generated using these equations:"
   ""
   "  total_padding[i] = stride[i] * (input_size[i] - 1) + output_padding[i] + ((kernel_shape[i] - 1) * dilations[i] + 1) - output_shape[i]"
-  "  If (auto_pads != SAME_UPPER): pads[start_i] = total_padding[i]/2; pads[end_i] = total_padding[i] - (total_padding[i]/2)"
+  "  If (auto_pads == SAME_UPPER): pads[start_i] = total_padding[i]/2; pads[end_i] = total_padding[i] - (total_padding[i]/2)"
```
AlexandreEichenberger, line 824: Could you comment on the source of all these text changes in the ONNXOPS.td.inc?

utils/gen_onnx_mlir.py
```diff
@@ -353,7 +353,7 @@
 OpsWithCanonicalizer = ['Add', 'Constant', 'Identity', 'Cast', 'Transpose',
                         'Dropout', 'Shape', 'Size', 'GlobalAveragePool',
                         'GlobalMaxPool', 'SqueezeV11', 'UnsqueezeV11',
-                        'Squeeze', 'Unsqueeze', 'Reshape']
+                        'Squeeze', 'Unsqueeze', 'Reshape', 'DepthToSpace']
```
AlexandreEichenberger, line 356: Nit, if you put down new code, would be good to alphabetize this list.

etiotto:
Currently we lower `DepthToSpace` in `src/Conversions/ONNXToKernel/Tensor/DepthToSpace.cpp` into ONNXReshape -> ONNXTranspose -> ONNXReshape. There are 2 cases to consider, when all dimensions of the shape for `ONNX.Reshape` are known at compile time or when they aren't. In the former we generate a `onnx.Constant` to encode the shape, in the later we have to generate code to compute the new shape. 
I think the current approach is reasonable, it allows reuse of the code we already have to lower `onnx.Reshape` and `onnx.Transpose` so that, should we have a bug ior need to optimize codegen for those operations, we can leverage the effort.  Also, the implementation of SpaceToDepth can leverage the same approach.

PR #912, opened 2021-10-14
Update onnxai

tungld:
This patch adds some links to some .md documents to show them on the onnxai webpage.

It also makes more space for contents by reducing redundant spaces on the left and right sides.
As a result the webpage will look like:
Screen Shot 2021-10-14 at 12 54 16


AlexandreEichenberger:
@sstamenova thanks for your thoughtful analysis. @gongsu832 would you be able to provide input on the options. I am not knowledgeable enough to have an opinion on such matters, sorry.

PR #913, opened 2021-10-14
Builder docs

AlexandreEichenberger:
Wrote document on how to use the new builder infrastructure for Math, MemRef, Krnl, and Affine

### DIFF
--- a/docs/LoweringCode.md
+++ b/docs/LoweringCode.md
@@ -1,0 +1,63 @@
+<!--- SPDX-License-Identifier: Apache-2.0 -->
+
+# Lowering Code
+
+## Generating Standard or MemRef code
+
+The traditional way to generate code in MLIR is to use the `create` methods, which internally correspond `builder` methods associated with each MLIR operations. For example, creating an addition of two values is done as shown below.
+``` C++
+  // Declaration for the input values, to be filled accordingly
+  Value firstIntVal, secondIntVal;
+  Value firstFloatVal, secondFloatVal;
+  OpBuilder rewriter; // Typically inherited from a caller context.
+  Location loc; // Typically derived from an operation.
+
+  Value intRes = rewriter.create<AddIOp>(loc, firstIntVal, secondIntVal);
+  Value floatRes = rewriter.create<AddFOp>(loc, firstFloatVal, secondFloatVal);
+```
+
+In the above code, we need to distinguish between int and float type operations, and the code need to repetitively pass the location.
+
+A newer approach suggested by the MLIR community is to create a math builder, described below. The same code can be generated using the following.
+``` C++
+  // Using hte same declaration as above for values, rewriter, and location.
+  MathBuilder createMath(rewriter, loc);
+  Value intRes = createMath.add(firstIntVal, secondIntVal);
+  Value floatRes = createMath.add(firstFloatVal, secondFloatVal);
+```
+
+MLIR recommended this approach as it reads better, namely "we are creating a math add of two values", and the rewriter and location is now "hidden" inside the lightweight `createMath` object.
+
+In general, this and all other builders can be created as follows.
+  ``` C++
+    // Constructors in class declaration.
+    MathBuilder(OpBuilder &b, Location loc);
+    MathBuilder(DialectBuilder &db);
+
+    // Usage
+    MathBuilder createMath(rewriter, loc); // Use original info.
+    MathBuilder createMath(createKrnl);    // Use info stored in another builder.
+  ```
+
+The Math builder we have been looking at currently contains the following operations. Up to date info should be looked from the `MLIRDialectBuilder.cpp` file. Most operations are self explanatory. They handle both integer and float operations, and will generate an assert when a specific operation is not supported for a specific type.
+
+```C++
+struct MathBuilder : DialectBuilder {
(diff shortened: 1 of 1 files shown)


docs/LoweringCode.md
```diff
@@ -0,0 +1,265 @@
...
+
+The first method, `defineLoops` creates a set of loop descriptors that characterizes a loop iteration space. Initially, a set of loop descriptors characterizes the original loop iteration space, shortly, one such modified set can also be used to characterize an optimized iteration spaces, for example to represent a loop tiled iteration space after applying loop blocking and loop permutation.
+
+The second method above, `iterate` is used to create a set of loops and its corresponding loop body. Until we optimize loops, both the `originalLoops` and the `optimizedLoops` are set to the output of a `defineLoops` method call. These sets describe the iteration space and its dimensionality. The next two parameters are used to describe the lower and the upper bounds of the loop. The last parameter defines a lambda function that implements the body of the loop. This labmda function is invoked with two parameters: an object to create further Krnl operations within the loop body and a list of the current loop index values.
+
+The usage of this builder will become clearer with our same example, setting an array to value zero. This is the same example as in the prior section.
```
etiotto, line 148: our same --> our ?

docs/LoweringCode.md
```diff
@@ -0,0 +1,265 @@
...
+// Create the loop.
+createKrnl.iterate(loopDef, loopDef, {zero, zero}, {ub0, ub1},
+  [&](KrnlBuilder &createKrnl, ValueRange loopInd){
+    // Loop body.
+    createKrnl.store(zero, array, loopInd);
+  });
+```
+**Code example 2: Zeroing an array using the new builder interface**
+
+Using this new scheme, we first define the 2D loop iteration space and then create the loop iteration structure using the `iterate` method. Since the loop is unoptimized, the same `loopDef` value range is passed as the first 2 parameters. The bounds are passed as 2 set of ordered values.
```
etiotto, line 168: "as 2 set" --> "as 2 sets"

docs/LoweringCode.md
```diff
@@ -0,0 +1,265 @@
...
+
+Note that the lambda function create an `createKrnl` builder that is similar to that of the external environment (outside the loop), but customized for inside the loop. Se we can continue to use this overloaded builder to continue constructing krnl operations. In our case, we simply use the `loopInd` (2nd parameter of the lambda function), which are the current loop induction values, to define the element of the array that is set to zero.
```
etiotto, line 170: create -> creates

etiotto: Se we can -> So we can

docs/LoweringCode.md
```diff
@@ -0,0 +1,265 @@
...
+  void memset(Value dest, Value val);
+};
+```
+** Additional Krnl ops supported by the Krnl builder interface. **
+
+Above, both the load and store operations are used to create Krnl memory load and store operations. They should be used instead of the MLIR Affine or Standard dialect operations.
+
+The `block` method takes one loop definition (one value extracted from the output of a `defineLoop` operation) and will split that loop definition into 2, where the first one iterates over blocks of the given side, and the second one iterates inside of a given block. The two loop definitions are returned by the `block` method as a value range containing the two split loops described above.
+
+The `permute` method takes a list of loop definitions are ensure that the iterate will generate code reflecting the permuted order.
```
etiotto, line 196: are ensure that the iterate will generate -> and ensures that the loops will iterate according to the permuted order ?

docs/LoweringCode.md
```diff
@@ -0,0 +1,265 @@
...
+
+The `memcopy` method results in the array given by `dest` to be overwritten by `size` values from the array given by `src`. The `memset` method set the entire array given by `dest` to the value passed in `val`, typically zero.
```
etiotto, line 198: set the entire ->  sets the entire

docs/LoweringCode.md
AlexandreEichenberger: corrected

docs/LoweringCode.md
AlexandreEichenberger: yes, done

AlexandreEichenberger:
@etiotto @tungld thanks for the comments. Edited the text to reflect your changes.

docs/LoweringCode.md
```diff
@@ -0,0 +1,265 @@
...
+// Create the loop.
+createKrnl.iterate(loopDef, loopDef, {zero, zero}, {ub0, ub1},
+  [&](KrnlBuilder &createKrnl, ValueRange loopInd){
+    // Loop body.
+    createKrnl.store(zero, array, loopInd);
+  });
+```
+***Code: Zeroing an array using the new builder interface***
+
+Using this new scheme, we first define the 2D loop iteration space and then create the loop iteration structure using the `iterate` method. Since the loop is unoptimized, the same `loopDef` value range is passed as the first 2 parameters. The bounds are passed as 2 sets of ordered values.
```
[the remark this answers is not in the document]
AlexandreEichenberger: yes, will add a clarification to make it easier to understand

tungld (approved):
LGTM. Thanks for writing the doc!

I would be more helpful if we could add a link to this doc from onnxai website. Just two lines in this section: https://github.com/onnx/onnx-mlir/blob/master/docs/_data/navigation.yml#L32

PR #914, opened 2021-10-14
Support CumSum

tungld:
This patch implements CumSum based on prefix-sum algorithm proposed by Daniel and Steele [1].
The algorithm is visualized for 1D array like this, where `d = log2(n)` is the number of steps.

[1] Hillis, W. Daniel, and Guy L. Steele, Jr. 1986. "Data Parallel Algorithms." Communications of the ACM 29(12), pp. 1170–1183.

Fixed one issue related to `ArrayValueIndexCapture` where `loadDenseElementArrayValueAtIndex` did not support scalar tensors, e.g `memref<i32>`.

### DIFF
--- a/utils/pre-onnx-mlir.py
+++ b/utils/pre-onnx-mlir.py
@@ -1,0 +1,63 @@
+"""
+This script is intended to display, verify or convert an onnx model.
+With option -v or --VERBOSE, the model will be displayed on output
+with option --save, the converted model will be saved to file too.
+onnx package is required
+Example of usage:
+To convert a model, add.onnx, to the opset currently supported by 
+onnx-mlir (e.g. 13) , use command:
+ python pre-onnx-mlir add.onnx --save
+The converted model will be saved into file add-opset-13.onnx
+To display a model, add.onnx, use command:
+  python pre-onnx-mlir add.onnx -v --no_convert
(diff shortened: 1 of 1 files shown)


src/Dialect/ONNX/MLIRDialectBuilder.cpp
```diff
@@ -48,28 +54,71 @@ Value MathBuilder::exp(Value val) {
...
       lhs.getType().isa<IndexType>())
     return b.create<CmpIOp>(loc, CmpIPredicate::eq, lhs, rhs);
   return b.create<CmpFOp>(loc, CmpFPredicate::OEQ, lhs, rhs);
 }
+
 Value MathBuilder::select(Value cmp, Value lhs, Value rhs) {
   return b.create<SelectOp>(loc, cmp, lhs, rhs);
 }
 
+Value MathBuilder::constant(Type type, double val) {
```
AlexandreEichenberger, line 99: Thanks, I like that. I wonder if it might not make sense to make 2 calls, one for int and one for double, since they can be overloaded... Loved the call, but always feel a bit uneasy about casting an int to a double and expecting the right value... basically asking if `i == (int64_t)((double) i)` is true for any `i`.

PR #917, opened 2021-10-14
[NFC] Suppress warnings in third_party subprojects.

etiotto:
This PR adds clang & gcc options to suppress warnings when C/C++ code is compiled in the `third_party` subprojects. 
In addition it fixes an issues in the Runtime onnx-mlir code.  Warnings continue to be emitted for C/C++ code in the onnx-mlir project proper.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,11 +95,8 @@ elseif ((ONNX_USE_PROTOBUF_SHARED_LIBS AND Protobuf_USE_STATIC_LIBS)
   message(FATAL_ERROR
     "ONNX_USE_PROTOBUF_SHARED_LIBS and Protobuf_USE_STATIC_LIBS must be opposites of each other.")
 endif()
-add_subdirectory(third_party/onnx)
 
-add_subdirectory(third_party/pybind11)
-add_subdirectory(third_party/rapidcheck)
-
+add_subdirectory(third_party)
 add_subdirectory(utils)
 add_subdirectory(include)
 add_subdirectory(src)
--- a/third_party/CMakeLists.txt
+++ b/third_party/CMakeLists.txt
@@ -1,0 +1,8 @@
+if (CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
+  # Suppress warnings in third party subprojects.
+  add_compile_options(-Wno-everything)
+endif()
+
+add_subdirectory(onnx)
+add_subdirectory(pybind11)
+add_subdirectory(rapidcheck)
--- a/src/Runtime/jni/jnilog.c
+++ b/src/Runtime/jni/jnilog.c
@@ -164,7 +164,7 @@ void log_printf(
     sprintf(buf, "[-]");
 
   /* Output thread ID, log level, file name, function number, and line number */
-  snprintf(buf + strlen(buf), LOG_MAX_LEN - strlen(buf), "[%lx][%s]%s:%s:%d ",
+  snprintf(buf + strlen(buf), LOG_MAX_LEN - strlen(buf), "[%p][%s]%s:%s:%d ",
       get_threadid(), log_level_name[level], get_filename(file), func, line);
 
   /* Output actual log data */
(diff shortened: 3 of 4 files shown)


CMakeLists.txt
```diff
@@ -89,11 +89,8 @@ elseif ((ONNX_USE_PROTOBUF_SHARED_LIBS AND Protobuf_USE_STATIC_LIBS)
   message(FATAL_ERROR
     "ONNX_USE_PROTOBUF_SHARED_LIBS and Protobuf_USE_STATIC_LIBS must be opposites of each other.")
 endif()
-add_subdirectory(third_party/onnx)
-
-add_subdirectory(third_party/pybind11)
-add_subdirectory(third_party/rapidcheck)
 
+add_subdirectory(third_party)
```
sstamenova, line 93: In my experience, the `third_party` directory should only contain the submodules and NOT any custom `cmake` files. Adding a `cmake` file at the `third_party` directory level makes this less readable and harder to maintain.

third_party/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
```
sstamenova, line 1: Nit: space after if

third_party/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
...
+  # Suppress warnings in third party subprojects.
+  add_compile_options(-Wno-everything)
```
sstamenova, line 3: I would prefer to see all impactful settings in the top-level cmake.

CMakeLists.txt
etiotto: I do not understand the reason the `third_party` directory should not contain a `cmake` file.  Other directories already contain `cmake` files (e.g. src/CMakefile.txt) and I do not find that any harder to maintain.

third_party/CMakeLists.txt
etiotto: fixed.

third_party/CMakeLists.txt
etiotto: Is not easily done because the `add_compile_options(-Wno-everything)` would have to be added like this: 
```
if (CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
  # Suppress warnings in third party subprojects.
  add_compile_options(-Wno-everything)
endif()

add_subdirectory(third_party/onnx)
add_subdirectory(third_party/pybind11)
add_subdirectory(third_party/rapidcheck)

add_subdirectory(utils)
add_subdirectory(include)
...
```
Doing that would  affect anything that follows the `add_compile_options` and therefore it would disable warnings not just for the `third_party` code but all for the `onnx-mlir` code.

Moving
```
add_subdirectory(third_party/onnx)
add_subdirectory(third_party/pybind11)
add_subdirectory(third_party/rapidcheck)
```
to the end of the file is not possible because `cmake` will then fail to produce the makefiles.

CMakeLists.txt
AlexandreEichenberger: > I'm not sure what "no warning policy" you are thinking. Once again, the proper way to fix warnings is to fix your source code, not to sweep the dirt under the carpet.

@gongsu832 I don't see it as our direct responsibility to fix warnings of the third_party directories. However, as they generate a lot of spurious code, it pollutes the output that shows up on our shells as we compile onnx-mlir. A results is that we pay less attentions to our own warnings, thus they tend to increase.

Thus I see this as a good effort to ensure we are more disciplined in our efforts of avoiding warnings.

third_party/CMakeLists.txt
sstamenova: Also, since you're suppressing the warnings for gcc/clang, I'd like to see them suppressed for MSVC which uses a different flag and using `cxx_flags` will make that possible.

gongsu832:
Passing CI is necessary but not sufficient condition for merging, particularly for potentially controversial changes. When a PR generates a lot of discussions, it should only be merged after concerns from all parties are reasonably addressed, and alternatives suggested by others have been adequately considered.

PR #919, opened 2021-10-15
Adding a high level CONTRIBUTE.md that lists our individual help pages in context

AlexandreEichenberger:
Adding a README.md to the docs sub-directory to make it easier to find documentations.

I know that the same directory is also used to drive the `onnx.ai/onnx-mlir` directory, based on a custom branch.

This is the action that I found related to this
```
    - name: copy docs into gh-pages
      run: |
        # Delete everything in gh-pages that are not in docs but keep .git*
        rsync -ahvP --delete --filter 'protect .git*' \
              $GITHUB_WORKSPACE/onnx-mlir/docs/ $GITHUB_WORKSPACE/gh-pages/
```

If this README.md is negatively impacting the `onnx.ai/onnx-mlir`, I suggest that we add a line to delete the docs/README.md from the files copied. Please let me know how to add that. Thanks

### DIFF
--- a/src/Dialect/ONNX/ONNXOps.cpp
+++ b/src/Dialect/ONNX/ONNXOps.cpp
@@ -1610,12 +1610,15 @@ LogicalResult ONNXReduceSumV11Op::inferShapes(
 template <class T>
 static LogicalResult verifyKernelShape(T *op, Value filterOperand,
     Optional<ArrayAttr> kernelShapeOpt, int64_t spatialRank) {
+  if (filterOperand && !hasShapeAndRank(filterOperand)) {
+    // Won't be able to do any checking at this stage.
+    return success();
+  }
   // 1) Get shape of filter. Shape is not guaranteed to be compile time
   // constant.
   ArrayRef<int64_t> filterShape =
-      filterOperand
-          ? filterOperand.getType().cast<RankedTensorType>().getShape()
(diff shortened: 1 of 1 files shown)


gongsu832:
Sorry I don't quite get it why we have `README.md` and `docs/README.md`? Also, `docs` is specifically for `onnx.ai` site so mixing things is probably not a good idea.

AlexandreEichenberger:
I personally do not look for info on the onnx.ai website, always go to our GitHub docs dir. Find searching for the actual file unfriendly. Could add some of that material to a contributing page in the top dir, if there is objection to having another readme in docs.

docs/Workflow.md
```diff
@@ -0,0 +1,151 @@
...
+
+Commit changes made in response to review comments to the same branch on your
+fork.
+
+Very small PRs are easy to review. Very large PRs are very difficult to
+review.
+
+## Code style
+
+The coding style suggested by Google is used in onnx-mlir. See the [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html) for details.
```
etiotto, line 151: This one is strange to me. Given that MLIR is a LLVM project and we use code/infrastructure from MLIR (and LLVM proper) it would make more sense to me to follow the coding style used by LLVM (https://llvm.org/docs/CodingStandards.html).

PR #920, opened 2021-10-15
On Windows, pick python from PATH rather than the registry

### LINKED ISSUE #918: CI issue on Windows, related to python versions
From an unrelated PR (#912)

From @sstamenova 

>> @sstamenova I have seen the MLIR Window CI failing in the doc check, spuriously as it passed the next time I pushed some changes. Could it be a time-out issue (Running out of time?)

>@AlexandreEichenberger No, it is not a time-out issue. It looks like it's a python issue (there's now 3.10 on the agents and it's not happy with onnx). I can do one of two things:
>Change the CMakeLists.txt file for onnx-mlir to explicitly not search the windows registry ever for installed python and to rely on the environment. This will work for the Windows CI and it will work for any developers who've added the path to python to their environment variables (e.g. PATH) (anyone who runs into an error will have to add python to their environment once)
Change the CMakeLists.txt file for onnx-mlir to optionally not search the windows registry for installed python. This will work for developers who haven't added python to their environment and the Windows CI can pass the variable to not search the registry.
Both have pros/cons. I would personally choose 1) because I'd expect that everyone already has the version of python they want to use added to their PATH, but maybe not?

From @gongsu832 

>@sstamenova I don't completely follow but if I understand you correctly you are saying that there are actually two versions of python on windows, one is in the registry and the other in the environment? As long as the current behavior on other platforms, which is to require a minimal python version and to fail cmake if not found, is not altered, however you fix Windows CI should be fine.

>But in general, other platforms have to deal with the same problem happening here, i.e., a newer python version breaks the build or test process. This is typically caused by different parts of the dependencies requiring python going out of sync due to caching (unless of course it's due to some incompatible change in python which is rare but can happen).

>For example, in this particular case the problem seems to be that your onnx installation is built by 3.9 (or whatever previous version) but onnx-mlir is being built with 3.10 (the newly detected version). We don't have this problem on other platforms because we always rebuild onnx whenever we build onnx-mlir since building onnx doesn't take too long. In other words, onnx and onnx-mlir are always built by the same python version. That might be another option to fix your problem if it's feasible on Windows.

Thanks for looking into this

sstamenova:
On Windows, only search the environment for Python (instead of also checking the registry). This allows us to pick up the correct version of Python instead of always picking the latest installed on the machine.

Without this change, cmake will always pick the latest available version which is not necessarily compatible with onnx-mlir/onnx/etc.

This fixes #918.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,6 +41,12 @@ include(CTest)
 include(MLIR.cmake)
 
 # MLIR.cmake calls find_package(MLIR) which sets LLVM_MINIMUM_PYTHON_VERSION
+if (WIN32)
+  # On Windows, only search the environment for Python (instead of also checking the registry).
+  # This allows us to pick up the correct version of Python instead of always picking the latest
+  # installed on the machine.
+  set(Python3_FIND_REGISTRY NEVER)
+endif()
 find_package(Python3 ${LLVM_MINIMUM_PYTHON_VERSION} REQUIRED COMPONENTS Interpreter Development)
 
 # Don't require AWT to allow headless JDK to reduce docker image size


AlexandreEichenberger (approved):
LGTM, provided it does not break the other CIs

gongsu832 (approved):
LGTM

etiotto:
I see the CI are clean. Merging it in so that other PR can go through.

PR #921, opened 2021-10-16
SpaceToDepth lowering to krnl.

etiotto:
This PR adds the ONNX to Krnl lowering transformation for the `SpaceToDepth `operator. The operator is lowered by virtue of a reshape operation followed by a transpose and a final reshape.

### DIFF
--- a/src/Conversion/ONNXToKrnl/CMakeLists.txt
+++ b/src/Conversion/ONNXToKrnl/CMakeLists.txt
@@ -1,6 +1,8 @@
 # SPDX-License-Identifier: Apache-2.0
 
+# Please keep in alphabetical order.
 add_onnx_mlir_library(OMONNXToKrnl
+  ConvertONNXToKrnl.cpp
   ONNXToKrnlCommon.cpp
   ControlFlow/Loop.cpp
   ControlFlow/Scan.cpp
@@ -23,33 +25,34 @@ add_onnx_mlir_library(OMONNXToKrnl
   RNN/GRU.cpp
   RNN/LSTM.cpp
   RNN/RNN.cpp
+  RNN/RNNBase.cpp
   Tensor/ArgMax.cpp
-  Tensor/Identity.cpp
-  Tensor/Reshape.cpp
-  Tensor/Pad.cpp
-  Tensor/Transpose.cpp
-  Tensor/Squeeze.cpp
-  Tensor/Unsqueeze.cpp
+  Tensor/Concat.cpp
+  Tensor/Compress.cpp  
   Tensor/Constant.cpp
   Tensor/ConstantOfShape.cpp
-  Tensor/Concat.cpp
   Tensor/DepthToSpace.cpp
   Tensor/Expand.cpp
-  Tensor/Split.cpp
-  Tensor/Gather.cpp
-  Tensor/Shape.cpp
-  Tensor/Slice.cpp
-  Tensor/Size.cpp
   Tensor/Flatten.cpp
+  Tensor/Gather.cpp 
+  Tensor/Identity.cpp
+  Tensor/NonZero.cpp
+  Tensor/OneHot.cpp  
+  Tensor/Pad.cpp
+  Tensor/Squeeze.cpp
+  Tensor/Split.cpp
+  Tensor/Shape.cpp
+  Tensor/Size.cpp 
+  Tensor/Slice.cpp
+  Tensor/SpaceToDepth.cpp
+  Tensor/Reshape.cpp
   Tensor/Tile.cpp
+  Tensor/Transpose.cpp
+  Tensor/Unsqueeze.cpp
   Tensor/Range.cpp
   Tensor/Resize.cpp
-  Tensor/NonZero.cpp
   Tensor/ReverseSequence.cpp
-  Tensor/OneHot.cpp
-  Tensor/Compress.cpp
   ConvertONNXToKrnl.cpp
-
   LINK_LIBS PUBLIC
   OMConstPropHelper
   OMONNXOps
--- a/src/Dialect/ONNX/ONNXOpsHelper.cpp
+++ b/src/Dialect/ONNX/ONNXOpsHelper.cpp
@@ -220,8 +220,8 @@ ONNXConstantOp getONNXConstantOp(Value value) {
 }
 
 Value createONNXConstantOpWithDenseAttr(
-    PatternRewriter &rewriter, Location loc, Attribute dense) {
-  return rewriter.create<ONNXConstantOp>(loc, Attribute(), dense);
+    OpBuilder &builder, Location loc, Attribute dense) {
+  return builder.create<ONNXConstantOp>(loc, Attribute(), dense);
 }
 
 // Returns true if the Value is defined by none constant
--- a/src/Conversion/ONNXToKrnl/ConvertONNXToKrnl.cpp
+++ b/src/Conversion/ONNXToKrnl/ConvertONNXToKrnl.cpp
@@ -204,6 +204,7 @@ void FrontendToKrnlLoweringPass::runOnOperation() {
   populateLoweringONNXConstantOpPattern(patterns, &getContext());
   populateLoweringONNXConcatOpPattern(patterns, &getContext());
   populateLoweringONNXDepthToSpaceOpPattern(patterns, &getContext());
+  populateLoweringONNXSpaceToDepthOpPattern(patterns, &getContext());
   populateLoweringONNXShapeOpPattern(patterns, &getContext());
   populateLoweringONNXSliceOpPattern(patterns, &getContext());
   populateLoweringONNXSqueezeOpPattern(patterns, &getContext());
--- a/src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
+++ b/src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
@@ -17,6 +17,86 @@
 
 bool gEmitDealloc = true;
 
+Value OnnxToKrnlBuilder::reshape(
+    const Value input, const ArrayRef<DimIndexExpr> shapeDims) const {
+  assert(!shapeDims.empty() && "Shape dimensions should not be empty");
+
+  ShapedType inputType = input.getType().cast<ShapedType>();
+  Type elementType = inputType.getElementType();
+
+  // If the output dimensions are all literals the 'onnx/Reshape' operation
+  // can take the new shape via an 'onnx.Constant'.
+  if (llvm::all_of(
+          shapeDims, [](const DimIndexExpr &dim) { return dim.isLiteral(); })) {
+    SmallVector<int64_t, 6> shape;
+    for (const IndexExpr &dim : shapeDims)
+      shape.push_back(dim.getLiteral());
+
+    auto constantOp =
+        createONNXConstantOpWithDenseAttr(b, loc, b.getI64TensorAttr(shape));
+
+    Value reshapeRes = b.create<ONNXReshapeOp>(
+        loc, MemRefType::get(shape, elementType), input, constantOp);
+
+    return reshapeRes;
+  }
+
+  MemRefBuilder memRefBuilder(b, loc);
+  KrnlBuilder krnlBuilder(memRefBuilder);
+
+  // When the output dimensions aren't all literals we need to generate code
+  // to compute the shape. Allocate a buffer and store the putput dimension
+  // into it.
+  IndexType indexTy = b.getIndexType();
+  int64_t length = shapeDims.size();
+  memref::AllocOp alloc =
+      memRefBuilder.alignedAlloc(MemRefType::get({length}, indexTy), 16);
+
+  for (int64_t i = 0; i < length; ++i) {
+    Value index = emitConstantOp(b, loc, indexTy, i);
+    Value data = shapeDims[i].getValue();
+    krnlBuilder.store(data, alloc, index);
+  }
+
+  // Now create the 'onnx.Reshape' operation. Because the shape is not a
+  // compile time constant it is effectively unknown.
+  SmallVector<int64_t> shape(length, -1);
+  Value reshapeRes = b.create<ONNXReshapeOp>(
+      loc, MemRefType::get(shape, elementType), input, alloc);
+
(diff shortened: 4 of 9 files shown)


src/Conversion/ONNXToKrnl/CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
 # SPDX-License-Identifier: Apache-2.0
 
+# Please keep in alphabetical order.
```
etiotto, line 3: Reordering to be in alphabetical order.

src/Dialect/ONNX/IndexExpr.hpp
```diff
@@ -311,7 +311,7 @@ enum class IndexExprKind {
 // Data structure to hold all the IndexExpr in a given scope. A scope define
 // a scope during which each of the dynamic dimensions are defined and all of
 // the symbols hold constant in that scope.
-class IndexExprScope {
+class IndexExprScope final {
```
etiotto, line 314: Wanted to get rid of `virtual` on the destructor, so declaring this class `final`.

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -4045,29 +4045,29 @@ static LogicalResult verify(ONNXSpaceToDepthOp op) {
   ONNXSpaceToDepthOpAdaptor operandAdaptor(op);
 
   // Check input.
-  auto input = operandAdaptor.input();
+  const Value &input = operandAdaptor.input();
```
etiotto, line 4048: Should avoid using `auto` when the type is not obvious from the RHS of the assignment. Putting the type explicitly makes the code easier to read as it conveys the static type to the reader immediately (I know modern IDE can display the type ... but some ppl might still want to use bare bone development environments :-))

src/Conversion/ONNXToKrnl/Tensor/Tensor.cpp
```diff
@@ -0,0 +1,110 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//====- Tensor.cpp - ONNX dialects to Krnl lowering for tensor operations -===//
+//
+// Copyright 2021 The IBM Research Authors.
+//
+// =============================================================================
+//
+// This file contains common code shared by ONNX to KRNL conversion operators.
```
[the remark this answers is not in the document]
etiotto: I have added removed `Tensor.cpp` and `Tensor.hpp` and added the reshape+transpose API in `OnnxBuilder`. 
Thanks for the comment and suggestion. I think is cleaner now.

src/Dialect/ONNX/ONNXOpsHelper.cpp
```diff
@@ -42,13 +44,84 @@ Value OnnxBuilder::matmul(Type Y, Value A, Value B) const {
   return b.create<ONNXMatMulOp>(loc, Y, A, B);
 }
 
-Value OnnxBuilder::reshape(Type outputType, Value input, Value shape) const {
-  return b.create<ONNXReshapeOp>(loc, outputType, input, shape);
+Value OnnxBuilder::reshape(
```
tungld, line 47: Could we keep the original `reshape`? We still want to use it where `shape` is an output of another op, and don't want to construct a DinIndexExpr array.

src/Dialect/ONNX/ONNXOpsHelper.cpp
```diff
@@ -42,13 +44,84 @@ Value OnnxBuilder::matmul(Type Y, Value A, Value B) const {
...
+    Value index = emitConstantOp(b, loc, indexTy, i);
+    Value data = shapeDims[i].getValue();
+    krnlBuilder.store(data, alloc, index);
+  }
+
+  // Now create the 'onnx.Reshape' operation. Because the shape is not a
+  // compile time constant it is effectively unknown.
+  SmallVector<int64_t> shape(length, -1);
+  Value reshapeRes = b.create<ONNXReshapeOp>(
+      loc, MemRefType::get(shape, elementType), input, alloc);
```
tungld, line 92: Instead of preparing `shape(length, -1)` and calling MemRefCast, could we replace `shape` by `castOutputShape` here? So there is no need to call MemRefCast.

tungld (changes requested):
Thanks for your update! Yes, it is cleaner now.

src/Dialect/ONNX/ONNXOpsHelper.cpp
etiotto: Certainly, I am adding it back.

src/Dialect/ONNX/ONNXOpsHelper.cpp
etiotto: I gave that a try before and got the following error: 

```
 TEST_CASE_BY_USER=test_depthtospace_example_cpu  TEST_DYNAMIC=true /usr/local/bin/python3.9 /Users/etiotto@ca.ibm.com/Source/onnx/onnx-mlir/build/test/backend/Debug/test.py 
 ...
 error: expected result type with size = -1 instead of 2 in dim = 1
// -----// IR Dump After (anonymous namespace)::FrontendToKrnlLoweringPass Failed //----- //
```

I believe the reason for the error is that the new shape is a variable and so the Reshape output type cannot have any constant extents. It expects the output type to be memref<?x?x?x?x?x?xf32> rather than memref<?x2x2x?x?x?xf32>:
```
%14 = memref.alloc() {alignment = 16 : i64} : memref<6xindex>
. . .
krnl.store %10, %14[%c5] : memref<6xindex>
%15 = "onnx.Reshape"(%arg0, %14) : (memref<?x?x?x?xf32>, memref<6xindex>) -> memref<?x2x2x?x?x?xf32>
```
 
That is the reason I ended up injecting a cast aster the Reshape.

etiotto:
Addressed the comments received so far. 

Now `libOMONNXOps.so` uses `emitConstantOp` and `KrnlBuilder::store` from `libOMSupport.so` and `libOMKrnlOps.so` respectively. Therefore `libOMONNXOps.so` now has a dependency on those 2 shared libraries. 
This creates a circular dependency because:
  -  `libOMKrnlOps.so` *already* depends on `libOMONNXOps.so` 
  -  `libOMSupport.so` *already* depends on  `libOMKrnlOps.so`
To break the circular dependency I reorganized:
  - the Support library so that it no longer depends on any other onnx-mlir library, and
  - the KrnlOps library so that it no longer depends on the ONNXOps library


I haad to rework the structure of the shared libraries because in `ONNXOpsHelper.cpp` now uses functionality from both the Support library (libOMSupport.so) and the Krnl library (libOMKrnlOps.so). That created a circular dependency between libOMSupport.so (dependent on libOMKrnlOps.so), libOMKrnlOps.so (dependent on lib

src/Dialect/ONNX/ONNXOpsHelper.cpp
tungld: I see. Thanks for checking!

tungld (approved):
LGTM. Thanks!

src/Conversion/ONNXToKrnl/Tensor/DepthToSpace.cpp
```diff
@@ -73,119 +71,28 @@ struct ONNXDepthToSpaceOpLowering : public ConversionPattern {
       perm = {0, 1, 4, 2, 5, 3};
     }
 
+    OnnxBuilder onnxBuilder(rewriter, loc);
```
AlexandreEichenberger, line 74: nit: since it will need to be rebased, MLIR suggested that the builder's name be named `create` so that it then read `create.reshape(...)`. This convention is currently used throughout the code, might be nice to adopt it here too.

AlexandreEichenberger:
> Now libOMONNXOps.so uses emitConstantOp and KrnlBuilder::store from libOMSupport.so and libOMKrnlOps.so respectively. Therefore libOMONNXOps.so now has a dependency on those 2 shared libraries.
This creates a circular dependency because:

Just requesting confirmation: we went at considerable length to make sure "ONNX" dialects can be used free of "Krnl" constructs, because some folks out there may want to use only the "ONNX" part of the project to lower it to their own/other MLIR dialects. This is why ONNXShapeHelper class use lambda functions `fGetDenseVal` and `fLoadVal` to extract info. During shape inferences, these functions refer only to ONNX operations (get from ONNX constant for the first one, undefined for the second one). During lowering to Krnl, these functions refer to ONNX or KRNL operations, which is fine as this is employed while lowering to Krnl, so clearly using Krnl operations there is fine.

etiotto:
> > Now libOMONNXOps.so uses emitConstantOp and KrnlBuilder::store from libOMSupport.so and libOMKrnlOps.so respectively. Therefore libOMONNXOps.so now has a dependency on those 2 shared libraries.
> > This creates a circular dependency because:
> 
> Just requesting confirmation @etiotto.
> 
> We went at considerable length to make sure "ONNX" dialects can be used free of "Krnl" constructs, because some folks out there may want to use only the "ONNX" part of the project to lower it to their own/other MLIR dialects. This is why ONNXShapeHelper class use lambda functions `fGetDenseVal` and `fLoadVal` to extract info. During shape inferences, these functions refer only to ONNX operations (get from ONNX constant for the first one, undefined for the second one). During lowering to Krnl, these functions refer to ONNX or KRNL operations, which is fine as this is employed while lowering to Krnl, so clearly using Krnl operations there is fine.

With this change the `libOMONNXOps.so` shared library will contain undefined references to symbols in `libOMKrnlOps.so`. To avoid this dependency the member function `Value OnnxBuilder::transpose(
    Type outputType, Value input, ArrayAttr perm) const`  I added to the `OnnxBuilder would not be allowed to generate calls to `onnx.store`...  I can think of a way to solve the problem:  move OnnxBuilder somewhere in the ONNXToKrnl conversion library. Would that be OK @AlexandreEichenberger ? Any other suggestion?

src/Dialect/ONNX/ONNXOpsHelper.cpp
```diff
@@ -12,11 +12,13 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include "src/Dialect/ONNX/ONNXOpsHelper.hpp"
 #include "mlir/IR/TypeUtilities.h"
-
-#include "src/Dialect/ONNX/IndexExpr.hpp"
+#include "src/Dialect/Krnl/KrnlHelper.hpp"
```
AlexandreEichenberger, line 17: That is what you don't want. We "must" keep ONNX Dialect free of anything KRNL related. No refs to any header files. This will forces users of ONNX Dialects that are only interested in ONNX to have krnl, which they have requested not to have.

AlexandreEichenberger (changes requested):
We need to remove dependences to Krnl in ONNX. I am sorry, I may have led you to this and was not clear enough on the requirements. I can see one of two solutions. One would be to create an ONNXBuilder subclass in Krnl that add functionality to ONNXBuilder that is specifically requesting Krnl. 

What I did for the ONNXShapeHelper, where we clearly need to call it first in an ONNX only/no Krnl code gen mode and then in an Krnl code gen mode was to:
1. define the functions that could encapsulate minimum operations requiring the Krnl support
2. add lambda function to the base class that allows the instantiation with the ONNX only functionality first, and then later with the ONNX+Krnl support.
While not amazingly satisfying, this was the simpler way to have a common ONNX only class that could eventually be used in a context where KRNL could also be used.

The functions where generic enough: one is to get an 1D attribute of constant (fetching it from ONNX Constant or the Krnl equivalent named KrnlGlobal), and the other reading from a 1D memref (null instantiation for ONNX only as we then say "runtime" by using -1 encoding of shapes, instantiation to a krnl load when used in the Krnl context).

src/Dialect/ONNX/ONNXOpsHelper.cpp
etiotto: OK I will think about how to reorg the code to minimize dependencies. Might have questions later.

etiotto: I have changed the patch to remove any change to the shared libraries structure. To do that I have created a class  in ONNXToKrnlCommon containing the member functions I need to lower both `SpaceToDepth` and `DepthToSpace` operators. The actual code in those member functions is unchanged. Please take another look @AlexandreEichenberger.

src/Conversion/ONNXToKrnl/Tensor/DepthToSpace.cpp
etiotto: done

tungld:
Not sure this is really a problem or not. Why can't pass different `fGetDenseVal` and `fLoadVal` to different instances of ONNXShapeHelper class?

During shape inference for ONNX, we pass functions operating on ONNXConstant only, and when lowing onnx to krnl, we pass ones operating on ONNXConstant or KrnlGlobal.

src/Dialect/ONNX/ONNXOpsHelper.cpp
AlexandreEichenberger: LGTM, thanks for preserving the independence of ONNX vs KRNL, much appreciated

AlexandreEichenberger (approved):

etiotto:
> Not sure this is really a problem or not. Why can't we pass different `fGetDenseVal` and `fLoadVal` to different instances of ONNXShapeHelper class?
> 
> During shape inference for ONNX, we pass functions operating on ONNXConstant only, and when lowing onnx to krnl, we pass ones operating on ONNXConstant or KrnlGlobal.

I think this questions relates to the new class `OnnxToKrnlBuilder` I have created. That class extends `OnnxBuilder` to add APIs that can use Krnl operations. Those operations will be able to reuse the functionality of the base class where it makes sense. This approach provides the advantage that the base class is completely unaware of its derived class and so, people that want to use `OnnxBuilder`, need not be aware of the Krnl dialect at all.

The alternative of passing code that 'reshape' operation (for example) need to use via a lambda function (or functor) is another possible approach. That approach is sensible if we wanted to use a common algorithm (say for example an algo to visit nodes in a graph) and we needed to customize the action taken by the algorithm depending on context (e.g.  upon visiting a graph node use the lambda function to collect properties in one context or modify the node in another context). 

In this case the alternative is not so attractive (to me) because we do not really have a common functionality we want to customize depending on context.  I might be missing something, and I am happy to reconsider alternatives if that si the case.

PR #924, opened 2021-10-19
Added ONNX-MLIR logo

AlexandreEichenberger:
Added logo that is not a derivative of other logos. It symbolize the transformation of ONNX into MLIR from left to right.


AlexandreEichenberger:
Obviously, anyone is welcome to suggest an alternative. I can also provide the original ppt file to anyone requesting it. The constraint  we have is that our logo cannot be a derivative logo from ONNX, probably the same for MLIR.

PR #933, opened 2021-10-20
[NFI]: Refactor ONNXShapeHelper code.

etiotto:
Split the ONNXShapeHelper code (shape inference) into files corresponding to each `onnx-mlir` operator. 
This will reduce future merge conflicts.


src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -18,23 +18,26 @@
...
-#include <type_traits>
-
-#include "mlir/IR/BuiltinOps.h"
-#include "onnx/defs/schema.h"
-#include "llvm/ADT/TypeSwitch.h"
+#include "FrontendDialectTransformer.hpp"
+#include "src/Interface/HasOnnxSubgraphOpInterface.hpp"
+#include "src/Interface/ResultTypeInferenceOpInterface.hpp"
 
+#pragma clang diagnostic ignored "-Weverything"
```
AlexandreEichenberger, line 25: Is that going to fix the warning problem without adding things to the cmake for third_party? if so, that is great!

AlexandreEichenberger (approved):
LGTM, thanks for the clarification

PR #938, opened 2021-10-20
Moved AffineBuilder to MLIR Builders

AlexandreEichenberger:
Cleaned up the AffineBuilder, which was "hiding" inside a .cpp file. I also eliminated the ImplicitLocOpBuilder which almost never used and just added clutter to the system. Added a `createMath.constantIndex(val)` which create an index literal, which was used very often but needed extra code to get it going.


src/Dialect/ONNX/MLIRDialectBuilder.hpp.inc
```diff
@@ -0,0 +1,135 @@
+//===- MLIRDialectBuilder.inc - Support migration to MLIR without EDSC --===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
```
sstamenova, line 3: Did this code actually come from llvm?

I'm curious why this is a `.inc` file rather than just a `.hpp` or `.cpp` file.

AlexandreEichenberger: I have it as an include file so as to not "pollute" the header file with the implementation, which is needed as this is a template.

AlexandreEichenberger: The Dialect concept is lifted from MLIR, but modified here to use a base class, additional methods. So its a derivative work.

src/Dialect/ONNX/MLIRDialectBuilder.hpp.inc
```diff
@@ -0,0 +1,135 @@
...
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+
+// Implementation of GenericAffineBuilder
+
+template <class LOAD_OP, class STORE_OP>
+Value GenericAffineBuilder<LOAD_OP, STORE_OP>::load(
+    Value memref, ValueRange indices) {
+  // Use Krnl load, see above.
```
tungld, line 13: Nit. The comment is obsoleted.

src/Dialect/ONNX/MLIRDialectBuilder.hpp.inc
```diff
@@ -0,0 +1,135 @@
...
+  return b.create<LOAD_OP>(loc, memref, indices);
+}
+
+template <class LOAD_OP, class STORE_OP>
+inline void GenericAffineBuilder<LOAD_OP, STORE_OP>::store(
+    Value val, Value memref, ValueRange indices) {
+  // Use Krnl store, see above.
```
tungld, line 20: Nit. The comment is obsoleted.

AlexandreEichenberger:

PR #942, opened 2021-10-21
DepthToSpace operator lowering to krnl dialect.

etiotto:
This PR adds the ONNX to Krnl lowering transformation for the `DepthToSpace `operator. The operator is lowered by virtue of a reshape operation followed by a transpose and a final reshape.


src/Conversion/ONNXToKrnl/Tensor/DepthToSpace.cpp
```diff
@@ -0,0 +1,185 @@
...
+    assert(bounds.getRank() == 4 && "Input tensor should have rank equal to 4");
+
+    DimIndexExpr B(bounds.getDim(0));
+    DimIndexExpr C(bounds.getDim(1));
+    DimIndexExpr H(bounds.getDim(2));
+    DimIndexExpr W(bounds.getDim(3));
+    DimIndexExpr newC = create<SignedFloorDivIOp>(C, (bs * bs), rewriter, loc);
+    DimIndexExpr newH = create<MulIOp>(H, bs, rewriter, loc);
+    DimIndexExpr newW = create<MulIOp>(W, bs, rewriter, loc);
+
```
tungld, line 60: IndexExpr supports Mul, FloorDiv, so we can write, say, `newC = C.floorDiv(bs*bs)`, `newH = H * bs`, `newW = W*bs` for brevity.

src/Conversion/ONNXToKrnl/Tensor/DepthToSpace.cpp
```diff
@@ -0,0 +1,185 @@
...
+    if (std::is_same<Op, SignedFloorDivIOp>::value)
+      return LiteralIndexExpr(indexExpr.getLiteral() / constant.getLiteral());
+    else if (std::is_same<Op, MulIOp>::value)
+      return LiteralIndexExpr(indexExpr.getLiteral() * constant.getLiteral());
+
+    llvm_unreachable("Unexpected operator type");
+  }
+
+  // Reshape the 'input' tensor to the shape prodided by 'outputDims'.
+  Value reshape(const Value &input, const DimsExpr &outputDims,
```
tungld, line 113: Could we call ONNXReshape instead of manually doing the lowering? We want to avoid the same lowering at multiple places.

src/Conversion/ONNXToKrnl/Tensor/DepthToSpace.cpp
```diff
@@ -0,0 +1,185 @@
...
+    ShapedType inputType = input.getType().cast<ShapedType>();
+    Type elementType = inputType.getElementType();
+    MemRefType memRefType = MemRefType::get(shape, elementType);
+
+    return emitMemRefReinterpretCastOp(
+        rewriter, loc, input, memRefType, outputDims);
+  }
+
+  // Transpose the 'input' tensor.
+  Value transpose(Value &input, const DimsExpr &outputDims,
```
tungld, line 130: Similarly, I would use ONNXTranspose instead.

src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
```diff
@@ -84,7 +84,7 @@ Value insertAllocAndDealloc(MemRefType type, Location loc,
 // dimensions, it uses the index expressions to retrieve the corresponding
 // values.
 Value insertAllocAndDeallocSimple(PatternRewriter &rewriter, Operation *op,
-    MemRefType type, Location loc, SmallVectorImpl<IndexExpr> &outputDims,
+    MemRefType type, Location loc, const SmallVectorImpl<IndexExpr> &outputDims,
```
tungld, line 87: Thanks for adding this!

test/mlir/onnx/onnx_lowering.mlir
```diff
@@ -1998,6 +1998,47 @@ func private @test_clip_default_min(%arg0: tensor<3xf32>, %arg1: tensor<f32>, %a
 
 // -----
 
+// Test whether the lowering is correct in the presence of dynamic dimensions.
```
tungld, line 2001: If you use ONNXReshape and ONNXTranspose, you may find this helpful: https://github.com/onnx/onnx-mlir/blob/master/src/Conversion/ONNXToKrnl/ConvertONNXToKrnl.cpp#L126 for DepthToSpace. It stops the lowering at a middle state, then the lit test output would be simpler, containing ONNXReshape and ONNXTranspose.

tungld (changes requested):
@etiotto In general, I would call existing ONNX ops as many as possible, instead of duplicating the lowering.

PR #945, opened 2021-10-22
Update LLVM version 10/21/2021

MikeHolman:
Updates code and tests to be compatible with LLVM revision 0bf230d4220660af8b2667506f8905df2f716bdf

The vast majority of changes are due to this upstream change https://reviews.llvm.org/rG8c08f21b6041 which moved arithmetic ops into a new arith dialect out of the std dialect.

Updated code and tests to use the arithmetic operators that were moved from standard dialect.
Added ReconcileUnrealizedCastsPatterns to KrnlToLLVM pass, which was refactored into a new pass upstream.
Added appropriate conversion patterns from arithmetic to LLVM.
Updated tablegen files and gen_onnx_mlir.py to change from DefaultValuedAttr<StrAttr,...> to DefaultValuedStrAttr<StrAttr,...>, due to change in default values being escaped.
Fixed bug in lowering of getMemRef that was exposed by upstream changes where a Mul was using the wrong version of an operand. We were using the source off of the getMemRef op directly instead of from the operands array parameter that contains converted version of the operand.
Updated code for a few other minor interface refactors.


AlexandreEichenberger:
@MikeHolman big thanks, much appreciated

> it was definitely an adventure to get through.

PR #946, opened 2021-10-23
Use ninja as the default build system

mshr-h:
Just like building LLVM.


AlexandreEichenberger:

PR #956, opened 2021-10-29
Conversion support

AlexandreEichenberger:
Because we need some support for conventions in other places than ONNX Cast, I factored the conversion into ArithBuilder. I also added support for boolean and unsigned conversions. The one conversion that is missing is between signed and unsigned at this time. If someone has an algo, that would be great. The other one is from/to String needed in Cast, I am not sure if it is needed anywhere, so let's wait on this one.

This also uncovered the strange way that MLIR handle signed and unsigned. The arithmetic operations only work on signless, and we treat signed bits as signless. Unsigned int do not appear to be handled directly, so there is a special cast that moves the unsigned to signless, apply the op, and then recast to unsigned. There is a later pass that removes redundant special casts (UnrealizedConversionCastOp).  Unfortunately, there is no backend test for unsigned values, I tested on a custom test that all the unrealized conversions are going away before final llvm is emitted.


AlexandreEichenberger:
@tungld I updated the code to also do cast to/from IndexType, and that turned out to be useful as there were places in the code where there were conversion to/from Index from/to float, for example. Now all these cases are handled directly by a `createMath.cast`.

As I went through the code to update the IndexCastOp, I also cleaned up a few constant generation methods that are much cleaner also with the math builder.

I generally did only update files that use the new scheme; code that used the old KrnlLoop approach are too different.

I also found an issue with the handling of signless integer in MathBuilder; by testing `if (value.isSignedInteger())`, that forced signless (such as IndexTypes as well as all of the ONNX Signed integer that are currently mapped as signless) to use the unsigned operations. Clearly that is wrong. I thus flipped the code to test `if (value.isUnsignedInteger())`. It ran the backend test fine, and only had to flip a few opcode in lit tests. 

So it turned out to be way more involved that I had hoped for, but I think the code is now in a better shape for it. Thanks for the nudge to do it, @tungld .

tungld (approved):
LGTM. Thanks! Our MathBuilder is getting more and more convenient.

Issue #937, opened 2021-10-20
Build the onnx submodule via `setup.py` on Windows to match other platforms' CIs.


sstamenova:
See https://github.com/onnx/onnx-mlir/issues/918 for more details.

Right now the Wincows CI calls `pip install` on onnx rather than building it from the submodule. Other CIs (not CircleCI which is obsolete) build the submodule instead. The recommendation is to do the same on Windows assuming it is possible.

I'm opening this so that we can keep track of the issue and I might have time to look at it in the next couple of weeks.

PR #957, opened 2021-10-29
Several small improvements including adding onnx-mlir-doc to ALL target and building onnx from the submodule

sstamenova:
1. Add onnx-mlir-doc to the ALL target. This way it will be run with the default target
2. On Windows instead of installing onnx from pip, build it from source to match other CIs
3. Remove unnecessary include in Compiler

This addresses https://github.com/onnx/onnx-mlir/issues/937


gongsu832 (approved):
LGTM

PR #958, opened 2021-10-31
Parallelize test sets

gongsu832:
We currenly run a few test sets, e.g., check-onnx-backend, check-onnx-backend-dynamic,
    etc., in serial. While we do parallelize within each test set, we don't get much
    benefit from it. Because within each test set, most of the time was spent on one
    or two long running tests. While waiting for these long running tests, we could have
    started the next test sets.

    This patch parallelizes across tests sets by starting them all at once so the long
    running tests from different test sets can run in parallel.

    CompilerLib tests are disabled because a bug has been uncovered that it was actually
    still using onnx-mlir instead of CompilerLibTest to compile the models. Fix the bug
    to use CompilerLibTest and it fails to compile the models.

### DIFF
--- a/test/numerical/CMakeLists.txt
+++ b/test/numerical/CMakeLists.txt
@@ -3,15 +3,15 @@
 add_custom_target(numerical)
 set_target_properties(numerical PROPERTIES FOLDER "Tests")
 
-add_custom_target(check-numerical
+add_custom_target(check-onnx-numerical
   COMMENT "Running the ONNX MLIR numerical regression tests"
   COMMAND "${CMAKE_CTEST_COMMAND}" -L numerical --output-on-failure -C $<CONFIG> --force-new-ctest-process
   USES_TERMINAL
   DEPENDS numerical
   )
-set_target_properties(check-numerical PROPERTIES FOLDER "Tests")
+set_target_properties(check-onnx-numerical PROPERTIES FOLDER "Tests")
 # Exclude the target from the default VS build
-set_target_properties(check-numerical PROPERTIES EXCLUDE_FROM_DEFAULT_BUILD ON)
+set_target_properties(check-onnx-numerical PROPERTIES EXCLUDE_FROM_DEFAULT_BUILD ON)
 
 # add_numerical_unittest(test_name sources... options...
 #   This function (generally) has the same semantic as add_onnx_mlir_executable.
--- a/.azure-pipelines/Windows-CI.yml
+++ b/.azure-pipelines/Windows-CI.yml
@@ -63,7 +63,7 @@ jobs:
 
   - script: |
       call "%ProgramFiles(x86)%\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvarsall.bat" x64
-      call onnx-mlir\utils\check-numerical.cmd
+      call onnx-mlir\utils\check-onnx-numerical.cmd
     displayName: Run onnx-mlir numerical tests
     workingDirectory: $(Agent.BuildDirectory)
     env:
--- a/README.md
+++ b/README.md
@@ -279,9 +279,9 @@ call cmake --build . --config Release --target check-onnx-lit
 
 To run the numerical ONNX MLIR tests, use the following command:
 
-[same-as-file]: <> ({"ref": "utils/check-numerical.cmd", "skip-ref": 1})
+[same-as-file]: <> ({"ref": "utils/check-onnx-numerical.cmd", "skip-ref": 1})
 ```shell
-call cmake --build . --config Release --target check-numerical
+call cmake --build . --config Release --target check-onnx-numerical
 ```
 
 To run the doc ONNX MLIR tests, use the following command after installing third_party ONNX:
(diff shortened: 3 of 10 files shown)


gongsu832:
@MikeHolman The `check-compilerlib-backend` target you created in #894 turns out to be non-functional. While you set `TEST_DRIVER_PATH` in `test/backend/test_config_compilerlib.py.in`, `test.py` is still importing `test_config` so it's just repeating the `check-onnx-backend` tests.

I have fixed the bug so the (slightly renamed for consistency) `check-onnx-backend-compilerlib` will correctly invoke `CompilerLibTest` to compile the models. However, it fails to compile any of the models due to not being able to find libcruntime. So for now I have disabled the test in the dockerfiles. Could you please fix the test so we can enable it again in the dockerfiles? Thanks.

AlexandreEichenberger (approved):
LGTM

test/numerical/CMakeLists.txt
```diff
@@ -3,15 +3,15 @@
 add_custom_target(numerical)
 set_target_properties(numerical PROPERTIES FOLDER "Tests")
 
-add_custom_target(check-numerical
+add_custom_target(check-onnx-numerical
   COMMENT "Running the ONNX MLIR numerical regression tests"
-  COMMAND "${CMAKE_CTEST_COMMAND}" -L numerical --output-on-failure -C $<CONFIG> --force-new-ctest-process
+  COMMAND "${CMAKE_CTEST_COMMAND}" -j $$\{NPROC:-$$\(nproc\)\} -L numerical --output-on-failure -C $<CONFIG> --force-new-ctest-process
```
sstamenova, line 8: This change needs to be reflected in the Windows CI as well - since it does not set NPROC, the tests are no longer running in parallel there.

sstamenova:
I think it would be a good idea to add a parameter to cmake for the parallelization level of the tests. Right now several cmake files depend on the `NPROC` variable being set in the environment but that is both repetitive (needed for every build command) and also not very portable to Windows.

If we added something like `ONNX_MLIR_PARALLEL_TESTS`, we could let it have several options:
1) No parallelization
2) All possible threads
3) A set number of threads

Then all the `check-onnx` targets can follow the same rules.

test/numerical/CMakeLists.txt
gongsu832: So why it didn't take the $(nproc) fallback? $(nproc) does not work on Windows? In that case, perhaps a more Windows compatible change is to make the fallback to be `${CTEST_PARALLEL_LEVEL}`.

gongsu832:
@sstamenova I think we discussed this before. The reason for the NPROC is because it's not a simple constant to be set. We compute it in the `.buildbot/jenkins-build-[llvm-]project.py` based on both CPU numbers and memory size. Parallelism for all Linux Jenkins CI tests are set according to NPROC, which is passed into the Dockerfiles.

sstamenova:
> @sstamenova I think we discussed this before. The reason for the NPROC is because it's not a simple constant to be set. We compute it in the `.buildbot/jenkins-build-[llvm-]project.py` based on both CPU numbers and memory size. Parallelism for all Linux Jenkins CI tests are set according to NPROC, which is passed into the Dockerfiles.

The docket files then invoke cmake, right? So NPROC can be passed into them and they can pass it to cmake setting the variable.

The previous discussion was to always parallelize to the maximum level possible. The current suggestion is to allow cmake to set the variable once based on whatever logic is necessary and then not have to pass it to each make command. I believe that would work for the docker scenario, right?

test/numerical/CMakeLists.txt
sstamenova: I am not sure what $(nproc) fallback you are referring to. The windows ci does not set an nproc variable for the numerical tests anywhere. The change that you made parallelizes the tests based on a variable that ends up having no set value in that case.

gongsu832:
Yes the docker file invokes cmake to generate the Makefiles but parallelism is ultimately controlled by the program that actually runs the tests. Whatever you set with cmake, you still need to pass it to the test program. For example, all the backend tests are run through `pytest`. It has its own way of parallelizing things via the `--forked` option. So it's either passing NPROC or something else set by cmake. Of course, anything that makes it more compatible with Windows we can certainly make those changes.

test/numerical/CMakeLists.txt
gongsu832: `${NPROC:-$(nproc)}` means that if `NPROC` is not set, use `$(nproc)`.

sstamenova: I thought you were saying that there was a fallback to `$(nproc)`, this makes more sense. There's no `$(nproc)` on Windows.

sstamenova:
Would you like me to send a change for PR that would use a cmake variable for this?

test/numerical/CMakeLists.txt
gongsu832: > Would you like me to send a change for PR that would use a cmake variable for this?

No need. I modified the dockerfiles to set CTEST_PARALLEL_LEVEL for the numerical tests so no change required on Windows.

gongsu832:
> Would you like me to send a change for PR that would use a cmake variable for this?

No need. I modified the dockerfiles to set CTEST_PARALLEL_LEVEL for the numerical tests so no change required on Windows.

sstamenova:
LGTM

PR #960, opened 2021-11-01
CategoryMapper Onnx to Krnl lowering.

etiotto:
In this PR I have implemented the lowering code from the Onnx dialect to the Krnl dialect for the `CategoryMapper` operation.  The codegen supports `CategoryMapper` operators that have either a tensor of strings or a tensor of int64_t as their input. I have introduced a `krnl.string` custom type which is used to lower the existing `onnx.String` type. 

### DIFF
--- a/src/Conversion/ONNXToKrnl/ML/CategoryMapper.cpp
+++ b/src/Conversion/ONNXToKrnl/ML/CategoryMapper.cpp
@@ -1,0 +1,297 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//===------------ CategoryMapper.cpp - Lowering CategoryMapper Op ---------===//
+//
+// Copyright 2021 The IBM Research Authors.
+//
+// =============================================================================
+//
+// This file lowers the ONNX CategoryMapper Operator to Krnl dialect.
+//
+//===----------------------------------------------------------------------===//
+
+#include "src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.hpp"
+#include "src/Conversion/ONNXToKrnl/PerfectHash.hpp"
+#include "src/Dialect/ONNX/ShapeInference/ONNXShapeHelper.hpp"
+
+#include "mlir/Dialect/SCF/SCF.h"
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/ADT/StringRef.h"
+#include "llvm/ADT/TypeSwitch.h"
+#include "llvm/Support/Debug.h"
+
+using namespace mlir;
+using llvm::dbgs;
+
+#define DEBUG_TYPE "category_mapper_onnx_to_krnl"
+
+struct ONNXCategoryMapperOpLowering : public ConversionPattern {
+  using PerfectHashTable = struct {
+    Value G;
+    Value V;
+    Value len;
+  };
+
+  ONNXCategoryMapperOpLowering(MLIRContext *ctx)
+      : ConversionPattern(ONNXCategoryMapperOp::getOperationName(), 1, ctx) {}
+
+  LogicalResult matchAndRewrite(Operation *op, ArrayRef<Value> operands,
+      ConversionPatternRewriter &rewriter) const final {
+    auto categoryMapperOp = cast<ONNXCategoryMapperOp>(op);
+    ONNXCategoryMapperOpAdaptor operandAdaptor(operands);
+
+    ONNXCategoryMapperOpShapeHelper shapeHelper(&categoryMapperOp, &rewriter,
+        getDenseElementAttributeFromKrnlValue,
+        loadDenseElementArrayValueAtIndex);
+    LogicalResult shapeComputed = shapeHelper.computeShape(operandAdaptor);
+    (void)shapeComputed;
+    assert(succeeded(shapeComputed) && "Could not compute output shape");
+
+    // Operands and attributes.
+    Location loc = categoryMapperOp.getLoc();
+    Value X = operandAdaptor.X();
+    ArrayAttr cats_int64sAttr = categoryMapperOp.cats_int64sAttr();
+    ArrayAttr cats_stringsAttr = categoryMapperOp.cats_stringsAttr();
+
+    DenseElementsAttr cats_int64s = mlir::DenseElementsAttr::get(
+        RankedTensorType::get(
+            cats_int64sAttr.size(), rewriter.getIntegerType(64)),
+        cats_int64sAttr.getValue());
+    DenseElementsAttr cats_strings = mlir::DenseElementsAttr::get(
+        RankedTensorType::get(
+            cats_stringsAttr.size(), StringType::get(rewriter.getContext())),
+        cats_stringsAttr.getValue());
+
+    IntegerAttr default_int64 = categoryMapperOp.default_int64Attr();
+    DenseElementsAttr default_string =
+        (categoryMapperOp.default_stringAttr())
+            ? mlir::DenseElementsAttr::get(
+                  RankedTensorType::get(
+                      {}, StringType::get(rewriter.getContext())),
+                  categoryMapperOp.default_stringAttr().getValue())
+            : nullptr;
+
+    // Basic information.
+    auto memRefType = convertToMemRefType(*op->result_type_begin());
+    int64_t rank = memRefType.getShape().size();
+    ShapedType inputType = X.getType().cast<ShapedType>();
+    Type elementType = inputType.getElementType();
+
+    // Insert an allocation and deallocation for the result of this operation.
+    Value alloc = insertAllocAndDeallocSimple(
+        rewriter, op, memRefType, loc, shapeHelper.dimsForOutput(0));
+
+    MultiDialectBuilder<KrnlBuilder, MathBuilder> create(
+        rewriter, op->getLoc());
+
+    // Generate a perfect hash table. The hash table will be used to lookup the
+    // index of the input values.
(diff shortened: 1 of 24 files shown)


src/Dialect/Krnl/KrnlHelper.cpp
```diff
@@ -469,14 +469,6 @@ void KrnlBuilder::iterateIE(ValueRange originalLoops, ValueRange optimizedLoops,
       });
 }
 
-void KrnlBuilder::memcpy(Value dest, Value src, Value size) {
```
etiotto, line 472: moved to line 522

src/Dialect/Krnl/KrnlHelper.hpp
```diff
@@ -302,9 +302,6 @@ struct KrnlBuilder : public DialectBuilder {
       function_ref<void(KrnlBuilder &createKrnl, ValueRange indices)>
           bodyBuilderFn);
 
-  void memcpy(Value dest, Value src, Value size);
```
etiotto, line 305: moved to line 361

src/Dialect/Krnl/KrnlOps.hpp
```diff
@@ -31,28 +31,21 @@
...
+
 class KrnlOpsDialect : public Dialect {
 public:
   KrnlOpsDialect(MLIRContext *context);
+  KrnlOpsDialect() = delete;
+
   static StringRef getDialectNamespace() { return "krnl"; }
 
   /// Parse a type registered to this dialect.
-  Type parseType(DialectAsmParser &parser) const override {
```
etiotto, line 40: moved to .cpp file

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -33,7 +34,6 @@
 
 using namespace mlir;
 using namespace mlir::OpTrait::util;
-using namespace mlir::onnxmlir;
```
etiotto, line 36: Prefer to explicitly use onnxmlir::StringType to make it clear we are using that "StringType" rather than the "StringType" defined in the krnl dialect.

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -613,23 +613,23 @@ mlir::Type ONNXOpsDialect::parseType(mlir::DialectAsmParser &parser) const {
 
     if (parser.parseGreater())
       return Type();
-    return SeqType::get(elementTypes);
-  } else {
-    llvm_unreachable("Unexpected onnxmlir keyword");
```
etiotto, line 618: else with llvm_unreachable has been removed in favour of emitting a more informative parse error via `parser.emitError`.

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -613,23 +613,23 @@ mlir::Type ONNXOpsDialect::parseType(mlir::DialectAsmParser &parser) const {
...
+    return onnxmlir::SeqType::get(elementTypes);
   }
+
+  parser.emitError(parser.getNameLoc(), "unknown onnx type: " + keyword);
+  return Type();
 }
 
 void ONNXOpsDialect::printType(
-    mlir::Type type, mlir::DialectAsmPrinter &printer) const {
```
etiotto, line 623: Changes to use a TypeSwitch because I think is cleaner.

test/mlir/onnx/onnx_lowering.mlir
```diff
@@ -729,7 +729,7 @@ func private @test_reducesum(%arg0 : tensor<3x2x2xf32>) -> tensor<*xf32> {
   "std.return"(%0) : (tensor<*xf32>) -> ()
 
   // CHECK-LABEL: test_reducesum
-  // CHECK: [[GLOBAL:%.+]] = "krnl.global"() {name = "constant_0", shape = [1], value = dense<1> : tensor<1xi64>} : () -> memref<1xi64>
```
etiotto, line 732: we should not care about the constant name. Changed to make the tests  more resilient.

src/Conversion/ONNXToKrnl/ML/CategoryMapper.cpp
```diff
@@ -0,0 +1,294 @@
...
+            ? createKrnl.constant(
+                  MemRefType::get({}, StringType::get(rewriter.getContext())),
+                  "default_string", default_string)
+            : nullptr;
+
+    // Lookup the index in the perfect hash table corresponding to
+    // each input value.
+    BuildKrnlLoop inputLoops(rewriter, loc, rank);
+    inputLoops.createDefineAndIterateOp(X);
+    rewriter.setInsertionPointToStart(inputLoops.getIterateBlock());
```
tungld, line 118: This style of building `krnl.iterate` is quite old. Could we change to the new style that uses lambda function: https://github.com/onnx/onnx-mlir/blob/master/docs/LoweringCode.md?

tungld (commented):
@etiotto I have a general question: can we use int64_t instead of int32_t for G and V?CategoryMapper's input type is int64_t.

src/Conversion/ONNXToKrnl/PerfectHash.cpp
```diff
@@ -0,0 +1,211 @@
...
+    if (bucket.size() <= 1)
+      break;
+
+    int32_t hval = 1;
+    size_t item = 0;
+    SmallVector<uint32_t> slots;
+
+    // Repeatedly try different hash values until we find a hash function that
+    // places all items in the bucket into free slots.
+    while (item < bucket.size()) {
```
AlexandreEichenberger, line 160: Will you have a escape if it repeat for ever? E.g. after iterating for X, escape and increase the size of the hash?

src/Conversion/ONNXToKrnl/ML/CategoryMapper.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+    // Create constants to hold the arrays 'G' and 'V'.
+    auto createConstants = [&](const SmallVector<int32_t> &G,
+                               const SmallVector<int32_t> &V) {
+      assert(V.size() == G.size() && "V and G should have the same size");
+
+      MemRefType type = MemRefType::get(
+          {static_cast<int64_t>(V.size())}, builder.getIntegerType(32));
+      res.G = createKrnl.constant(type, "G", builder.getI32VectorAttr(G));
+      res.V = createKrnl.constant(type, "V", builder.getI32VectorAttr(V));
```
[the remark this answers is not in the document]
etiotto: @tungld goot catch. Will fix.

src/Conversion/ONNXToKrnl/ML/CategoryMapper.cpp
etiotto: Done.

etiotto (commented):
Addressed code review comments.

etiotto:
> @etiotto I have a general question: can we use int64_t instead of int32_t for G and V?CategoryMapper's input type is int64_t.

We can but it should not be necessary. G and V are indexes into the 2 arrays that encode the perfect hash table, so they need not have the same data type as the input tensor.

src/Dialect/Krnl/KrnlHelper.cpp
```diff
@@ -519,6 +511,35 @@ void KrnlBuilder::matmul(Value A, ValueRange aStart, Value B, ValueRange bStart,
       globalUBs[1], globalUBs[2], simdize, unroll, overcompute);
 }
 
+Value KrnlBuilder::constant(
+    MemRefType type, StringRef name, DenseElementsAttr value) const {
+  static int32_t constantID = 0;
+  return b.create<KrnlGlobalOp>(loc, type, b.getI64ArrayAttr(type.getShape()),
+      b.getStringAttr(name + std::to_string(constantID++)), value,
+      /*offset=*/nullptr,
+      /*alignment=*/nullptr);
```
tungld, line 520: @etiotto Thanks for mangling the global name! I hope we will often use this builder. It would be great if you add an optional `alignment` argument to the function too. I sometimes need to set alignment to 4K. Thanks!

src/Conversion/ONNXToKrnl/ML/CategoryMapper.cpp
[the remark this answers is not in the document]
AlexandreEichenberger: > I have a general question: can we use int64_t instead of int32_t for G and V?CategoryMapper's input type is int64_t

Since the map size can be large, it would make sense to me to have the smaller data type for the one we create (G if I recall correctly)

src/Conversion/ONNXToKrnl/PerfectHash.cpp
etiotto: The issue of potentially ending with an infinite loop here will be tracked via a separate PR. I have open a work item to track (https://github.com/onnx/onnx-mlir/issues/1010)

src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
```diff
@@ -112,11 +112,13 @@ MemRefType convertToMemRefType(Type type) {
   auto tensorType = type.dyn_cast<TensorType>();
```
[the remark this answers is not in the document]
etiotto: Done.

src/Dialect/Krnl/KrnlHelper.cpp
etiotto: No problem @tungld. I have modified the implementation of this member function to allow optionally passing both offset and alignment.

AlexandreEichenberger (approved):
LGTM

PR #962, opened 2021-11-01
Fix the compiler lib builds for multiconfig generators

sstamenova:
When the code was separated into its own library, the generator expression for the file configuration was dropped. This is needed because file generation happens at cmake config time and for multiconfig generators such as VS or Xcode the build configuration is not known.

This change brings back the generator expression for file generation.

### DIFF
--- a/src/Compiler/CMakeLists.txt
+++ b/src/Compiler/CMakeLists.txt
@@ -2,6 +2,17 @@
 
 include_directories(${PROJECT_BINARY_DIR}/src)
 
+configure_file(
+  ${CMAKE_CURRENT_SOURCE_DIR}/ExternalUtil.hpp.in
+  ${CMAKE_CURRENT_BINARY_DIR}/ExternalUtil.hpp.cfg
+  @ONLY
+  )
+
+file(GENERATE
+  OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/$<CONFIG>/ExternalUtil.hpp
+  INPUT ${CMAKE_CURRENT_BINARY_DIR}/ExternalUtil.hpp.cfg
+  )
+
 # CMAKE_CFG_INTDIR is . for single-config generators such as make, and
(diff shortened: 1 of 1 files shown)


src/Compiler/CMakeLists.txt
```diff
@@ -1,25 +1,25 @@
...
+# CMAKE_CFG_INTDIR is . for single-config generators such as make, and
+# it has a value (e.g. $(Configuration)) otherwise, so we can use it to
+# determine whether we are dealing with a multi-config generator.
+if (NOT "${CMAKE_CFG_INTDIR}" STREQUAL ".")
+  set(FILE_GENERATE_DIR ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_CFG_INTDIR})
+else()
+  set(FILE_GENERATE_DIR ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_BUILD_TYPE})
+endif()
+
 add_custom_target(ExternalUtil DEPENDS ${FILE_GENERATE_DIR}/ExternalUtil.hpp)
```
gongsu832, line 24: This should also be changed to `${CMAKE_CURRENT_BINARY_DIR}/$<CONFIG>/ExternalUtil.hpp`?

sstamenova: No, the targets get evaluated at build time when the configuration is known, so it shouldn't use a generator expression. The `file` call gets evaluated at cmake configuration time, so it needs to use the generator expression.

gongsu832: Oh right. You set `FILE_GENERATOR_DIR` above. Sorry, for some reason I can never read the github diff on the web page properly.

AlexandreEichenberger:
@gongsu832 will defer to you on accepting this patch

gongsu832 (approved):
LGTM

PR #964, opened 2021-11-02
fix the mistake of shape in doc-example

DavidChan0519:
Signed-off-by: david.cheng <david.cheng@enflame-tech.com>


tungld:

tungld (approved):
LGTM. Thank you for the patch!

PR #967, opened 2021-11-03
Onehot update

AlexandreEichenberger:
Update the support for one hot. It now use the modern infrastructure, it use the MathBuilder::castToIndex for the various conversions that are needed for the operation. OneHot has now a verifier.

I test for out of bound indices values, which result in the output value to be set to the offValue for all positions along the axis values. 

I also updated a bit how constant are read. The previous "one hot" did support only reading the constant coming from the ONNX dialect, which was kind of ok since it was not reusing the shape helper in the lowering to krnl. Now we can read on One Hot the constants in ONNX as well as in Krnl using the interface provided by the ShapeInference top class.

All inputs now handle runtime dimensions.

Minor nit, ONNXOpHelper.cpp had lots of `mlir` even though it was working in the `mlir` name space. I removed the superfluous ones.

Another nit was that `getScalarValue` was casting everything to double. So integer values were casted to double, before being casted back to int. The current version use template to cast directly to the proper final type.

A final nit was that `getONNXConstantOpFromDenseAttr` was actually creating an op, not just getting info. It is now rename to `createONNXConstantOpWithDenseAttr` to better match what we do elsewhere.


tungld (approved):
LGTM. Thanks!

PR #972, opened 2021-11-03
Updating docs to move into the right place

AlexandreEichenberger:
Just like `make OMONNXOpsIncTranslation` moves the updated ONNX spec files into the source files, now the  `make onnx-mlir-docs` also move the updated description of the ONNX and Krnl dialect into the `docs/Dialects` subdirectory.

Added reference to this command in the documentation files.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -70,9 +70,9 @@ jobs:
           name: Ensure tablegen documentation is up-to-date
           command: |
             cd onnx-mlir/build
-            cmake --build . --target onnx-mlir-doc
+            cmake --build . --target onnx-mlir-docs
             # Check whether dialect documentation is up-to-date.
-            diff docs/Dialects ../docs/Dialects
+            # diff docs/Dialects ../docs/Dialects
       - run:
           name: Print the Current Time
           command: date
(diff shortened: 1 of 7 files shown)


MLIR.cmake
```diff
@@ -63,12 +63,12 @@ function(add_onnx_mlir_dialect_doc dialect dialect_tablegen_file)
           OUTPUT ${GEN_DOC_FILE}
```
sstamenova, line 63: This needs to change too now because the output has changed.

sstamenova: I would recommend just changing GEN_DOC_FILE to point to the source rather than the bin directory if that's the goal then the rest of the function does not need to change.

sstamenova (changes requested):

.circleci/config.yml
```diff
@@ -70,9 +70,9 @@ jobs:
           name: Ensure tablegen documentation is up-to-date
           command: |
             cd onnx-mlir/build
-            cmake --build . --target onnx-mlir-doc
+            cmake --build . --target onnx-mlir-docs
```
sstamenova, line 73: Can we just remove the circleci since it's obsolete?

MLIR.cmake
AlexandreEichenberger: Thanks, that makes a lot of sense. I am really not an expert at cmake files, learning every days.

.circleci/config.yml
AlexandreEichenberger: will do in a separate PR, we did disable it recently.

sstamenova (approved):

PR #975, opened 2021-11-04
Handle constant outputs of a model

### LINKED ISSUE #929: Output tensor generated for the constant input test models cannot be freed
I ran into this when trying to extend backend tests to JNI.

Most of the `TEST_CONSTANT=true` tests failed for JNI because they get an error
```
free(): invalid pointer
```
in the `jniwrapper.c` when the code is calling `omTensorListDestroy` to free the output tensors from the native code after making a copy to construct the corresponding Java object. The problem didn't manifest in the C/C++ tests because `PyExecutionSession::pyRun` never tried to free the output tensors.

I picked one of the tests `test_reshape_zero_and_negative_dim` to reproduce the error (ONNX model and driver code attached). To compile and run,
```
# pwd
/root/Downloads/onnx-mlir/build/test_reshape_zero_and_negative_dim
# ../Release/bin/onnx-mlir --EmitLib test_reshape_zero_and_negative_dim.onnx
Shared library test_reshape_zero_and_negative_dim.so has been compiled.
# gcc -I../../include driver.c test_reshape_zero_and_negative_dim.so -L../Release/lib -lcruntime
# LD_LIBRARY_PATH=. ./a.out
input  signature: [
]
output signature: [   { "type" : "f32" , "dims" : [2 , 3 , 1 , 4] , "name" : "reshaped" }

]
Tensor[0]: owning: 1
Tensor[0]: elems: 24
Tensor[0]: data[0] 0.548814
Tensor[0]: data[1] 0.715189
Tensor[0]: data[2] 0.602763
Tensor[0]: data[3] 0.544883
Tensor[0]: data[4] 0.423655
Tensor[0]: data[5] 0.645894
Tensor[0]: data[6] 0.437587
Tensor[0]: data[7] 0.891773
Tensor[0]: data[8] 0.963663
Tensor[0]: data[9] 0.383442
Tensor[0]: data[10] 0.791725
Tensor[0]: data[11] 0.528895
Tensor[0]: data[12] 0.568045
Tensor[0]: data[13] 0.925597
Tensor[0]: data[14] 0.071036
Tensor[0]: data[15] 0.087129
Tensor[0]: data[16] 0.020218
Tensor[0]: data[17] 0.832620
Tensor[0]: data[18] 0.778157
Tensor[0]: data[19] 0.870012
Tensor[0]: data[20] 0.978618
Tensor[0]: data[21] 0.799159
Tensor[0]: data[22] 0.461479
Tensor[0]: data[23] 0.780529
free(): invalid pointer
Aborted
```
As you can see, the `owning` flag for the tensor is true, which means that the model runtime generated the tensor memory (as opposed to the memory being user supplied). So when the user calls `omTensorListDestroy`, the runtime is going to free the tensor memory. In this case, I'm guessing that the tensor memory is some sort of static memory that's not malloc-ed? If that's the case, then the correct thing to do is to set the `owning` flag to false so `omTensorDestroy` (called by `omTensorListDestroy`) won't try to free it.

[invalid_pointer.tar.gz](https://github.com/onnx/onnx-mlir/files/7374470/invalid_pointer.tar.gz)

tungld:
This patch handles output OMTensors which hold outputs in a model inference.

Before this patch, an output OMTensor always owns an output pointer (`owning = true`), so when `omTensorListDestroy` is called by users, the OMTensor will free the output buffer. However if an output is a constant which is read-only, it would fail to free the output.

This patch checks for each output, if the output is a constant, set `owning = false` in OMTensor. Otherwise, set `owning = true`. By that way, only OMTensors with `owning=true` will free data when `omTensorListDestroy` is called.

In case of that a constant is not directly returned, but it is consumed by a sequence of ViewOp, we can detect and set `owning = false` too.

The lowering of ONNXConstantOp to KrnlGlobalOp creates a new buffer if ONNXConstantOp's result is returned. Now I remove the creation of a new buffer and directly return the constant result.

This patch also changes all API calls to use int64 instead of int32 to avoid potential integer-related issues on Z.

resolves #929


gongsu832 (commented):
LGTM. Thanks for the fix.

src/Conversion/KrnlToLLVM/KrnlToLLVM.cpp
```diff
@@ -776,6 +779,12 @@ class KrnlUnaryMathOpLowering : public ConversionPattern {
 class KrnlEntryPointOpLowering : public OpRewritePattern<KrnlEntryPointOp> {
 public:
   using OpRewritePattern<KrnlEntryPointOp>::OpRewritePattern;
+  ArrayRef<bool> constantOutputs;
+
+  KrnlEntryPointOpLowering(MLIRContext *ctx, ArrayRef<bool> constantOutputs)
+      : OpRewritePattern<KrnlEntryPointOp>(ctx) {
```
etiotto, line 785: Suggest to initialize the member variable in the constructor init list:
```
  : OpRewritePattern<KrnlEntryPointOp>(ctx), constantOutputs(constantOutputs) {
```

src/Conversion/KrnlToLLVM/KrnlToLLVM.cpp
```diff
@@ -1486,6 +1501,81 @@ void mlir::populateAffineAndKrnlToLLVMConversion(RewritePatternSet &patterns,
...
 }
 
+void mlir::checkConstantOutputs(
+    ModuleOp &module, SmallVectorImpl<bool> &constantOutputs) {
+  Operation *entryPointOp;
+  auto walkResult = module->walk([&](mlir::Operation *op) -> WalkResult {
+    if (llvm::dyn_cast<KrnlEntryPointOp>(op)) {
+      entryPointOp = op;
+      return WalkResult::interrupt();
+    } else
```
etiotto, line 1511: [nit]: The else is not necessary because the `then` part will return, suggest: 
```
if (llvm::dyn_cast<KrnlEntryPointOp>(op))i {
      entryPointOp = op;
      return WalkResult::interrupt();
}

return  WalkResult::advance();

File: src/Conversion/KrnlToLLVM/KrnlToLLVM.cpp [hunk 3]
```diff
@@ -1486,6 +1501,81 @@ void mlir::populateAffineAndKrnlToLLVMConversion(RewritePatternSet &patterns,
...
+          .getLeafReference()
+          .getValue();
+
+  // Get entry function op.
+  Operation *entryFunc;
+  module->walk([&](FuncOp op) -> WalkResult {
+    if (SymbolRefAttr::get(op).getValue() == entryPointFuncName) {
+      entryFunc = op;
+      return WalkResult::interrupt();
+    } else
```
etiotto, line 1533: Same comment as at line 1511.

src/Conversion/KrnlToLLVM/KrnlToLLVM.cpp
```diff
@@ -1486,6 +1501,81 @@ void mlir::populateAffineAndKrnlToLLVMConversion(RewritePatternSet &patterns,
...
+  });
+  assert(entryFunc && "Entry function not found");
+
+  // Get ReturnOp of the entry function op.
+  Operation *returnOp;
+  entryFunc->walk([&](Operation *op) -> WalkResult {
+    if (llvm::dyn_cast<ReturnOp>(op)) {
+      returnOp = op;
+      return WalkResult::interrupt();
+    } else
```
etiotto, line 1544: Ditto as line 1511

AlexandreEichenberger (approved):
LGTM, please consider the suggestions from @etiotto, tend to prefer the if-return-no-else pattern also.

tungld:
@etiotto I changed to use if-return-no-else as you suggested. Thanks!

PR #977, opened 2021-11-05
Print out debug info about opset of importing onnx ops

tungld:
This patch prints out opset of ONNX operations in a given ONNX model. It also prints out the opset version onnx-mlir will use to import an ONNX op.

E.g.
```
$ ./Debug/bin/onnx-mlir --EmitONNXBasic ssd_mobilenet_v1_10.onnx --debug 2>&1 | grep frontend
...
frontend_dialect_transformer: Importing ONNX Cast, opset: 10
frontend_dialect_transformer: Importing ONNX Greater, opset: 10
frontend_dialect_transformer: Importing ONNX If, opset: 10
frontend_dialect_transformer: Importing ONNX Gather, opset: 10
frontend_dialect_transformer: Importing ONNX Shape, opset: 10
frontend_dialect_transformer: Importing ONNX Cast, opset: 10
frontend_dialect_transformer: Importing ONNX Slice, opset: 10
frontend_dialect_transformer: Importing ONNX Squeeze, opset: 10
frontend_dialect_transformer:   - use opset 11
...
```

We can see that all ONNX ops here have opset 10, but ONNX Squeeze will be imported as opset 11 in onnx-mlir.
 
A warning will be printed out if a compiling model contains old ops ( opset < 6).

### DIFF
--- a/src/Builder/FrontendDialectTransformer.cpp
+++ b/src/Builder/FrontendDialectTransformer.cpp
@@ -981,12 +981,13 @@ private:
       return std::string("");
     }
     auto current_opset = opset_map_.find(node.domain())->second;
-    LLVM_DEBUG(llvm::dbgs()
-               << DEBUG_TYPE << ": Importing ONNX " << node.op_type()
-               << ", opset: " << current_opset << "\n");
+
     if (current_opset < OPSET_THRESHOLD)
       llvm::outs() << "Warning: ONNX " << node.op_type() << " opset "
                    << current_opset << " is quite old\n";
+    LLVM_DEBUG(llvm::dbgs()
+               << DEBUG_TYPE << ": Importing ONNX " << node.op_type()
+               << ", opset: " << current_opset << "\n");
 
     // Custom ops may not be present in op_dialect_version_map_. If no version
     // info is found, treat as unversioned (no renaming).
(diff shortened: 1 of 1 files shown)


src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -38,6 +39,12 @@ SUPPRESS_WARNINGS_POP
 #include <map>
 #include <type_traits>
 
+#define DEBUG_TYPE "frontend_dialect_transformer"
+
+/// We consider opset < 6 is old. Users will see a warning if their model
+/// contains ops of old opset.
+#define OPSET_THRESHOLD 6
```
etiotto, line 46: [optional]:  could use `static constexpr int32_t OpsetThreshould=6;`

etiotto (approved):
LGTM

AlexandreEichenberger:
@tungld Maybe for another PR, I would suggest that we have a "minimum level supported" global constant, and that we issue a warning if we find an op below that level. I believe that would help making more meaningful response to users.

And when we get an error in the frontend, and that the op is below this  minimum, instead of generating a cryptic message, we could first say "error parsing op XXX of deprecated Opset X, please consider regenerating and/or up converting your model".

I will open an Issue.

src/Builder/FrontendDialectTransformer.cpp
tungld: Done. Thanks!

tungld:
@AlexandreEichenberger yes, I will do it at multiple levels in another PR. This PR nows prints out a message when importing a model only.

Issue #688, opened 2021-05-14
Gemm inside a loop-if


AlexandreEichenberger:
With the new Keras models translated to TF and then to ONNX, we have the following pattern:

``` 
  onnx.loop (...) {
     ...
     onnx.gemm(...)
   '''
 }
```

Simplified to the extreme, I isolated this pattern that cause problem once we are in the krnl dialect:

``` mlir
    %1001 = krnl.define_loops 1
    krnl.iterate(%1001) with (%1001 -> %1002 = 0 to 35) {
      %1010 = krnl.load %1000[] : memref<i1>
      scf.if %1010 { // comment out to make it work

        // pattern generated when lowering Gemm
        // enf of pattern generated for Gemm
        
      } // if
    } // iter
```

A pattern that has only the "if" or only the "iterate" works just fine. But the one with both results in the following error when further lowering to affine.

```
(onnx-mlir) alexe@pancetta i117-gemm-in-loop % onnx-mlir-opt --convert-krnl-to-affine --canonicalize onnx-gemm-191-krnl-loopif.mlir
Assertion failed: (isPerfectlyNested(input) && "input not perfectly nested"), function permuteLoops, file /Users/alexe/Onnxcode/llvm-project/mlir/lib/Transforms/Utils/LoopUtils.cpp, line 1548.
PLEASE submit a bug report to https://bugs.llvm.org/ and include the crash backtrace.
Stack dump:
0.	Program arguments: onnx-mlir-opt --convert-krnl-to-affine --canonicalize onnx-gemm-191-krnl-loopif.mlir
1.	Stack dump without symbol names (ensure you have llvm-symbolizer in your PATH or set the environment var `LLVM_SYMBOLIZER_PATH` to point to it):
0  onnx-mlir-opt            0x00000001094b05bb llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) + 43
1  onnx-mlir-opt            0x00000001094af238 llvm::sys::RunSignalHandlers() + 248
2  onnx-mlir-opt            0x00000001094b0fb7 SignalHandler(int) + 295
3  libsystem_platform.dylib 0x00007fff2049dd7d _sigtramp + 29
4  libsystem_platform.dylib 0x0000000000018b00 _sigtramp + 18446603339974552992
5  libsystem_c.dylib        0x00007fff203ad411 abort + 120
6  libsystem_c.dylib        0x00007fff203ac7e8 err + 0
[...]

PR #981, opened 2021-11-09
Krnl new pattern (all loops lb/ub defined at once)

AlexandreEichenberger:
Change code pattern generation for Gemm/Matmul to follow the pattern suggested in issue #688


AlexandreEichenberger:
@sstamenova The Window CI fails. Its probably an ordering thing, not sure. Is there a way to have a terminal in a Window machine with the right environment to test directly what is going on? I don't have a window machine that is recent. Thanks

sstamenova:
> @sstamenova The Window CI fails. Its probably an ordering thing, not sure. Is there a way to have a terminal in a Window machine with the right environment to test directly what is going on? I don't have a window machine that is recent. Thanks

There's no way to connect to the machine, unfortunately. Since the failure is in the lit tests, I'm surprised it is only on Windows. Did the line endings change? Or maybe some characters that are not recognized on Windows?

tungld:
Maybe the failure on Windows was because lit tests cannot split a lit test file correctly by recognizing `// -----`.

Correct way to use `// -----` is that it must be surrounded by empty lines.

I saw that some functions around `test_matmul2` did not have an empty line after `// -----`. @AlexandreEichenberger Could you add empty lines and try again?

AlexandreEichenberger:
@tungld sadly that was not it.

sstamenova:
> @tungld sadly that was not it.

I'll try locally and see if I can reproduce it. I already have a build so it should be fairly quick.

sstamenova:
@AlexandreEichenberger I pulled your branch locally and the test failed in release (same as the pipeline). I noticed it had a crash stack, so I reran it in debug and it passed. Debug builds on Windows do some extra initialization, so this can make sense - especially since the error is an AV (C0000005). I was able to get the stack trace from a RelWithDebInfo build:

[fail.txt](https://github.com/onnx/onnx-mlir/files/7529747/fail.txt)

AlexandreEichenberger:
@sstamenova So, just to double check: you are saying that onnx-mlir-opt crashed on the matmul1 example. So its not that the pattern generated not quite match, it is that onnx-mlir-opt crashed on that input and in release mode. 

Thanks for running the example, much appreciated

sstamenova:
> @sstamenova So, just to double check: you are saying that onnx-mlir-opt crashed on the matmul1 example. So its not that the pattern generated not quite match, it is that onnx-mlir-opt crashed on that input and in release mode.
> 
> `AV (C0000005)` corresponds to a segfault?
> 
> Thanks for running the example, much appreciated.

Yes, onnx-mlir-opt crashed with the attached callstack (which is why the pattern didn't match). It crashes in release and relwithdebinfo but not debug. C0000005 is an access violation.

AlexandreEichenberger:
@sstamenova I would have had a very hard time finding the issue without your error log. Big thanks. I was using a `ValueRange` when I should have used a SmallVector. Looks like it works now

AlexandreEichenberger:
@tungld thanks also for your suggestions, much appreciated

tungld (approved):
LGTM.

PR #982, opened 2021-11-09
Fix some rewriting rules in rewrite.td

tungld:
Some rules in `rewrite.td`, such as those for Squeeze, Unsqueeze, and Reshape, require tensors having shape and rank. This patch adds conditions for such rules.

With this patch, we are able to compile yolov3 in the onnx model zoo.


src/Dialect/ONNX/ONNXOpsHelper.hpp
```diff
@@ -142,6 +142,9 @@ bool HasSpecifiedConstantShape(mlir::Value value, mlir::Value shape);
 bool AreTheSameConstantOpDenseAttr(
     mlir::Builder &builder, int64_t rank, mlir::Value lhsOp, mlir::Value rhsOp);
 
+/// Test if 'val' has shape and rank or not.
```
AlexandreEichenberger, line 145: Thanks for moving that op to the helper header file.

AlexandreEichenberger (approved):
The code LGTM.

I will defer to @chentong319 as I know he has a bigger picture view of these frontend optimizations.

chentong319:
A possible small trick: when the rank or shape is unknown but only used to handle negative axis value, we can assume it to be 0. As a result, we can correctly compare when the two axis values are both positive or negative. We just conservatively return not the same when one is positive and one is negative.

tungld:
> A possible small trick: when the rank or shape is unknown but only used to handle negative axis value, we can assume it to be 0. As a result, we can correctly compare when the two axis values are both positive or negative. We just conservatively return not the same when one is positive and one is negative.

@chentong319 done. Thanks for the suggestion.

chentong319 (approved):
LGTM!

PR #985, opened 2021-11-10
JNI Backend Tests

gongsu832:
- add JNI backend tests
- handle ONNX BOOL type, fix getNumElems() not checking empty shape[]
- OMTensor.getRank returns long to be consistent with C/C++ API
- add omTensorSetOwning and expose omTensorListCreateWithOwnership
- more flexible success condition check for various calls in jniwrapper.c
- fix OMT_DEBUG printing shape and strides with data content when rank=0
  since tmp wasn't cleared
- fix EmitJNI to honor output path
- fix get_threadid() snprintf type
- fix Jekyll front matter for doxyen API html title

### DIFF
--- a/include/onnx-mlir/Runtime/OMTensor.h
+++ b/include/onnx-mlir/Runtime/OMTensor.h
@@ -278,6 +278,11 @@ int64_t omTensorGetNumElems(OMTensor *tensor);
  */
 int64_t omTensorGetOwning(OMTensor *tensor);
 
+/**
+ * \brief OMTensor owning flag setter
+ */
+void omTensorSetOwning(OMTensor *tensor, int64_t owning);
+
 #ifdef __cplusplus
 }
 #endif
--- a/src/Runtime/jni/CMakeLists.txt
+++ b/src/Runtime/jni/CMakeLists.txt
@@ -1,20 +1,37 @@
 # SPDX-License-Identifier: Apache-2.0
 
 if (ONNX_MLIR_ENABLE_JNI)
+  # Setup Jsoniter jar
+  include(Jsoniter.cmake)
+  setup_jsoniter(0.9.23)
+
+  # Build javaruntime jar
   add_jar(javaruntime
     src/com/ibm/onnxmlir/OMModel.java
     src/com/ibm/onnxmlir/OMLogger.java
+    src/com/ibm/onnxmlir/OMRunner.java
     src/com/ibm/onnxmlir/OMTensorList.java
     src/com/ibm/onnxmlir/OMTensor.java
+
+    INCLUDE_JARS ${JSONITER_JAR}
     )
+  if (TARGET jsoniter)
+    add_dependencies(javaruntime jsoniter)
+  endif()
   install_jar(javaruntime DESTINATION lib)
 
-  # ONNX_MLIR_LIBRARY_PATH is a generator expression which is not supported by add_jar as the output
-  # directory. Instead, we let add_jar place the jar file in the default location and copy it to the
-  # library path after it is built.
+  # ONNX_MLIR_LIBRARY_PATH is a generator expression which is not supported by
+  # add_jar as the output directory. Instead, we let add_jar place the jar file
+  # in the default location and copy it to the library path after it is built.
+  #
+  # Note target must be ${ONNX_MLIR_LIBRARY_PATH}/javaruntime.jar instead of
+  # just ${ONNX_MLIR_LIBRARY_PATH} for Windows CI.
   add_custom_command(TARGET javaruntime POST_BUILD
-    COMMAND ${CMAKE_COMMAND} -E copy_if_different $<TARGET_PROPERTY:javaruntime,JAR_FILE> ${ONNX_MLIR_LIBRARY_PATH}/javaruntime.jar)
+    COMMAND ${CMAKE_COMMAND} -E copy_if_different
+            $<TARGET_PROPERTY:javaruntime,JAR_FILE>
+            ${ONNX_MLIR_LIBRARY_PATH}/javaruntime.jar)
 
+  # Build jniruntime lib
   add_onnx_mlir_library(jniruntime STATIC
     jniwrapper.c
     jnilog.c
@@ -32,4 +49,5 @@ if (ONNX_MLIR_ENABLE_JNI)
     PROPERTIES
     POSITION_INDEPENDENT_CODE TRUE
     )
+
 endif()
--- a/src/Runtime/jni/src/com/ibm/onnxmlir/OMRunner.java
+++ b/src/Runtime/jni/src/com/ibm/onnxmlir/OMRunner.java
@@ -1,0 +1,224 @@
+package com.ibm.onnxmlir;
+
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.InputStreamReader;
+import java.io.OutputStreamWriter;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.nio.DoubleBuffer;
+import java.nio.FloatBuffer;
+import java.nio.IntBuffer;
+import java.nio.LongBuffer;
+import java.nio.ShortBuffer;
+import java.util.ArrayList;
+import java.util.Base64;
+import java.util.HashMap;
+
+import com.jsoniter.JsonIterator;
+import com.jsoniter.output.JsonStream;
+
+/*import com.fasterxml.jackson.databind.ObjectMapper;*/
+
+public class OMRunner
+{
+    private static class Data {
+	/* For Jsoniter, field names match the keys in the JSON */
+	String buffer;
+	String dtype;
+	long[] shape;
+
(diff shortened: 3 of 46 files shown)


src/Runtime/jni/CMakeLists.txt
```diff
@@ -1,20 +1,57 @@
...
+
+  # Build Jsoniter if an installed version is not found
+  unset(JSONITER_JAR CACHE)
+  find_jar(JSONITER_JAR jsoniter-${JSONITER_VER})
+  if (${JSONITER_JAR} STREQUAL "JSONITER_JAR-NOTFOUND")
+    set(JSONITER_JAR ${JSONITER_REPO_JAR} CACHE STRING "" FORCE)
+
+    add_custom_command(
+      OUTPUT ${JSONITER_JAR}
+      COMMAND git clone -b ${JSONITER_VER} ${JSONITER_REPO_URL} ${JSONITER_REPO_DIR}
```
sstamenova, line 21: Does it make sense to add this as a submodule?

src/Runtime/jni/CMakeLists.txt
```diff
@@ -1,20 +1,57 @@
...
-  # ONNX_MLIR_LIBRARY_PATH is a generator expression which is not supported by add_jar as the output
-  # directory. Instead, we let add_jar place the jar file in the default location and copy it to the
-  # library path after it is built.
+  # ONNX_MLIR_LIBRARY_PATH is a generator expression which is not supported by
+  # add_jar as the output directory. Instead, we let add_jar place the jar file
+  # in the default location and copy it to the library path after it is built.
   add_custom_command(TARGET javaruntime POST_BUILD
-    COMMAND ${CMAKE_COMMAND} -E copy_if_different $<TARGET_PROPERTY:javaruntime,JAR_FILE> ${ONNX_MLIR_LIBRARY_PATH}/javaruntime.jar)
+    COMMAND ${CMAKE_COMMAND} -E copy_if_different
+            $<TARGET_PROPERTY:javaruntime,JAR_FILE> ${ONNX_MLIR_LIBRARY_PATH})
```
sstamenova, line 52: If I remember correctly, we had to setup the full path in the copy command because otherwise the copy did not work correctly. After your change, is it still copying it to the correct file name?

src/Runtime/jni/CMakeLists.txt
gongsu832: Moved Jsoniter setup to submodule to keep the mainline CMakeLists.txt tidy.

src/Runtime/jni/CMakeLists.txt
gongsu832: Yes the file is copied correctly.

gongsu832:

include/onnx-mlir/Runtime/OMTensor.h
```diff
@@ -278,6 +278,11 @@ int64_t omTensorGetNumElems(OMTensor *tensor);
  */
 int64_t omTensorGetOwning(OMTensor *tensor);
 
+/**
```
AlexandreEichenberger, line 281: Quick question, is the doc automatically updated? If not, please regenerate it to also include this new function

src/Runtime/jni/src/com/ibm/onnxmlir/OMTensor.java
```diff
@@ -15,25 +15,25 @@
...
-    final int ONNX_TYPE_FLOAT16    = 10;
-    final int ONNX_TYPE_DOUBLE     = 11;
-    final int ONNX_TYPE_UINT32     = 12;
-    final int ONNX_TYPE_UINT64     = 13;
-    final int ONNX_TYPE_COMPLEX64  = 14;
-    final int ONNX_TYPE_COMPLEX128 = 15;
-    final int ONNX_TYPE_BFLOAT16   = 16;
-
-    final int[] ONNX_TYPE_SIZE = new int[] {
+    final static int ONNX_TYPE_UNDEFINED  = 0;
```
AlexandreEichenberger, line 18: Nit: could you add in the comment with which list this "enum" needs to be kept consistent with?

AlexandreEichenberger (commented):
I would recommend to add a brief md page that indicates how to add a new function to the RT, if it needs to show up in the python/C/C++/Java interface. Does not need to be long, but would be useful for others to get the bigger picture. Thanks.

include/onnx-mlir/Runtime/OMTensor.h
gongsu832: I have no idea how to generate the doc. Since check-docs didn't fail, I assume this part of the code is not covered by check-docs.

src/Runtime/jni/CMakeLists.txt
```diff
@@ -1,20 +1,33 @@
...
   add_jar(javaruntime
     src/com/ibm/onnxmlir/OMModel.java
     src/com/ibm/onnxmlir/OMLogger.java
+    src/com/ibm/onnxmlir/OMRunner.java
     src/com/ibm/onnxmlir/OMTensorList.java
     src/com/ibm/onnxmlir/OMTensor.java
+
+    INCLUDE_JARS ${JSONITER_JAR}
     )
+  if(TARGET jsoniter)
```
sstamenova, line 18: Nit: space after if

src/Runtime/jni/CMakeLists.txt
sstamenova: As far as I can tell, the cmake file still has a git clone command. I think it makes sense instead of invoking git clone from cmake to instead add jsoniter as one of the third_party submodules of the project. That way the dependency will be kept track of in the same way as the rest of the dependencies.

gongsu832: I specifically do not want to add it as a third_party submodule because it's only used for JNI tests. And the JNI related targets in turn are only available when JDK is installed. In other words, I don't want to stick thing optional under the third_party.

Ideally, if the Jsoniter jar is available from some reputed website then it's just the same as downloading the onnx model when doing the tests. But unfortunately I couldn't find such a website to download it so have to resort to build it locally. But really, this is meant to be confined within the backend test subtree.

sstamenova: That makes sense. My concern is that it makes it hard to track what all the dependencies are. But if this is an uncommon scenario, then perhaps it won't be an issue.

gongsu832:

src/Runtime/jni/CMakeLists.txt
sstamenova: Ok, I tested this and on Windows, this will copy the jar file to ${ONNX_MLIR_LIBRARY_PATH} making that the filename if the directory does not exist already (a.k.a. the filename will be lib instead of being in the lib directory). Please keep the name in the copy command.

gongsu832:

gongsu832:

gongsu832:

gongsu832:

gongsu832:

gongsu832:

gongsu832:

AlexandreEichenberger:
@gongsu832 This PR seems to have other PR's code as well, including Kevin's.

src/Dialect/ONNX/AdditionalONNXOps.td
```diff
@@ -21,3 +24,77 @@ def ONNXCustomOp:ONNX_Op<"Custom",
     }
   }];
 }
+
+def ONNXCallOp : ONNX_Op<"ONNX_Call",
```
AlexandreEichenberger, line 28: This seems to be code from #987 Why is it here?

gongsu832:
That's fine. This is an issue with the PR not recognizing that Kevin's PR already got merged. I was going to raise the issue during our next meeting. It didn't use to do this. I remember we were talking about changing the way things get merged to keep the commit log clean. I don't know if that was done and was causing the trouble.

src/Dialect/ONNX/AdditionalONNXOps.td
gongsu832: That's fine. This is an issue with the PR not recognizing that Kevin's PR already got merged. I was going to raise the issue during our next meeting. It didn't use to do this. I remember we were talking about changing the way things get merged to keep the commit log clean. I don't know if that was done and was causing the trouble.

Basically, now whenever I rebase my PR after new commits got merged into master, it doesn't recognize that and shows those new commits onto which I just rebased as being my commits. Very annoying indeed.

src/Runtime/jni/src/com/ibm/onnxmlir/OMRunner.java
```diff
@@ -0,0 +1,224 @@
+package com.ibm.onnxmlir;
```
AlexandreEichenberger, line 1: Please add the copyright info to this as well as all the other .java files. Looks like none of them have them.

AlexandreEichenberger (approved):
LGTM, please add copyright to all of the Java files before merging

AlexandreEichenberger:
Since it is needed in another PR, will merge it now and add the copyright in a subsequent PR

PR #986, opened 2021-11-10
Resize v11

chentong319:
Solve the issue for Resize V11 when resize v10 is added. 
In onnx-mlir there is no rule for which version to pick if a higher and a lower version is supported. The existing  code assumes the lower version with the observation that a new version is introduced only when the interface of the op is changed.  However, for Resize Op, we implemented the recent version (V13) first and added the lower version (V10) later.  The V11 has the same interface with V13.
I do not want to change the basic rule, but just added V11 with a simple rewrite rule.
I also moved V10 code to decompose.td, too.


tungld (approved):
LGTM. Thank you for fixing the issue!

PR #988, opened 2021-11-11
* fix bug for input signature

shizy818:
* add input signature test cases

Signed-off-by: ZHEN YU SHI <shizy@cn.ibm.com>


shizy818:
Now unsigned integer case is not correctly handled. One example of orignal input signature for unsigned int8 is:
```
[    { "type" : ui8 , "dims" : [2 , 2 , 1] , "name" : "x" }

]
```

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -1288,25 +1288,25 @@ class FrontendGenImpl {
...
-    std::string const si32 = std::string("i32");
-    std::string const si64 = std::string("i64");
-    std::string const si16 = std::string("i16");
-    std::string const si8 = std::string("i8");
-    std::string const si1 = std::string("i1");
-    std::string const sui32 = std::string("ui32");
-    std::string const sui64 = std::string("ui64");
-    std::string const sui16 = std::string("ui16");
-    std::string const sui8 = std::string("ui8");
+    std::string const sf32 = std::string(" f32 ");
```
AlexandreEichenberger, line 1291: Can you offer some insight why adding the spaces matters here, and if its not easier to add them where the string is used?

test/backend/test_signature.py
```diff
@@ -0,0 +1,181 @@
+from __future__ import absolute_import
```
AlexandreEichenberger, line 1: Thanks for adding a signature test, this is very helpful. Please add the copyright when adding a new file, and I would urge you to add a paragraph of comments in the file describing the test, to make it easier for us to expand its functionality in the future if required.

gongsu832: Is it possible to incorporate this into test.py? Most of the code here is copied from test.py. Maintaining two copies would be error prone. If test.py gets too large, we can consider modularize it but we don't want to maintain two copies of similar code.

BTW, ID shiyz818 has been added to the whitelist to automatically trigger Jenkins builds.

gongsu832: Hi Zhen Yu, the current pattern is that different tests can be selected by either specifying an option or set an env var. Yes I do realize that signature tests do not need to go through onnx backend. But I think we can still extend test.py with for example --signature and TEST_SIGNATURE to branch to your code.

IMO test.py is a bit disorganized in its current form. So if you can take this opportunity to modularize and organize it a bit that would be very helpful. Thanks.

shizy818: I see. Let me try it.

gongsu832: Thanks Zhen Yu. Just be aware that in #985 I made some changes to test.py to add the JNI backend tests. So you probably want to make your changes on top of that instead.

shizy818: Hi @gongsu832. I incorporate signature test into test.py framework. Here is main changes:
* global variable initialization in varaibles.py
* common function `compile_model` in common.py
* Class SignatureBackendTest is inherited from onnx.backend.test.BackendTest for signature testing. It overwrites `_add_model_test` and `assert_similar_outputs`

Would you help to review? I will wait #985 be merged and then add JNI part. Thanks!

gongsu832: @shizy818 Thanks Zhen Yu for the changes. From a high level I think the pieces look good! There are errors in the code (e.g., ExecutionSession now has 3 params but still called with two) but I think that's only because you are still waiting for #985. This should make the backend tests a lot more maintainable. Thanks!

AlexandreEichenberger:
@shizy818 there are conflicts, you will need to manually merge the code

shizy818:
@AlexandreEichenberger Yes I will do some changes based on #985.

gongsu832:
@shizy818 remember to change the target branch of your PR to `main`. `master` will be deleted this Friday Nov. 26th.

shizy818:
@gongsu832 Thanks for reminding. Jenkins job fails and I'm going to check why.

test/backend/test_signature.py
caoimhinuibrian: @shizy818 @gongsu832 seems that #985 is merged now, so is this one good to go?

gongsu832: I think @shizy818 is still in the process of finishing up the modularization and cleanup of test.py so let's wait for him to answer.

shizy818: Hi @gongsu832 I merged JniExecutionSession from #985 into inference_backend.py last week. Logically I guess it is what it looks now. I double check code cleanup today. Is it Ok for you?

gongsu832: Hi @shizy818 yes it looks good. Thanks for fixing up test.py.

shizy818:
@AlexandreEichenberger @gongsu832  Shall we merge this PR?

gongsu832:
It's failing Windows CI. We need @sstamenova to take a look. It appeared to fail on numerical tests which this patch shouldn't have touched.

AlexandreEichenberger:
Rebasing to double check if we get CI errors for Windows in the numerical test again.

sstamenova:
> It's failing Windows CI. We need @sstamenova to take a look. It appeared to fail on numerical tests which this patch shouldn't have touched.

It looks to me like the patch is touching the numerical tests because it is touching `ExecutionSession` which is used in the tests. Am I misremembering how these tests work?

gongsu832:
Ahh, forgot it touched ExecutionSession.cpp. That certainly could break numerical tests. Are there more specific error messages on the Windows CI? Looking at the patch, I'm guessing the code that most likely to fail would be https://github.com/onnx/onnx-mlir/pull/988/files#diff-5d21b0ab24fd38cd6cb1208aa5cf9ad9138892a016cc88b07f292815e636019aR45-R46

sstamenova:
> Ahh, forgot it touched ExecutionSession.cpp. That certainly could break numerical tests. Are there more specific error messages on the Windows CI? Looking at the patch, I'm guessing the code that most likely to fail would be https://github.com/onnx/onnx-mlir/pull/988/files#diff-5d21b0ab24fd38cd6cb1208aa5cf9ad9138892a016cc88b07f292815e636019aR45-R46

Unfortunately, the messages from the test failures are not particularly helpful. It looks like the tests fail repeatedly getting into a place where they are trying to compile the same library and failing to do so because the library hasn't been cleaned up yet. I remember seeing this when I was getting the tests working - because of the nature of the tests they recompile the library for each iteration, so any failure can cause them to get into this state. Unlike other platforms, Windows can be particular about overwriting existing files which is what is causing this particular error message to dominate the log.

My guess is that the signature for the function on those two lines is wrong or the function doesn't exist at all. If I recall correctly, the functions that will be queried need to be added to the exports file to be properly found (`numerical.def`).

src/Runtime/ExecutionSession.cpp
```diff
@@ -41,6 +41,14 @@ ExecutionSession::ExecutionSession(
     errStr << "Cannot load symbol: '" << entryPointName << "'" << std::endl;
     throw std::runtime_error(errStr.str());
   }
+
+  _inputSignatureFunc = reinterpret_cast<inputSignatureType>(
+      _sharedLibraryHandle.getAddressOfSymbol(inputSignatureName.c_str()));
+  if (!_inputSignatureFunc) {
+    std::stringstream errStr;
+    errStr << "Cannot load symbol: 'omInputSignature'" << std::endl;
```
sstamenova, line 49: A couple of things:
1) If this is not needed everywhere (e.g. the numerical tests don't need this, but JNI does), don't try to load the function always, but only if the name is set.
2) Since the name CAN be set, it is not always `omInputSignature`, so this error needs to reflect the variable not a hard-coded value.

sstamenova (changes requested):

src/Runtime/ExecutionSession.cpp
gongsu832: Thanks @sstamenova for the suggestions. @shizy818 can you see if we can make some changes according to the suggestions and make it work for Windows CI? Thanks.

shizy818: @gongsu832 @sstamenova I see. Thanks. Trying...

shizy818: @gongsu832 @AlexandreEichenberger It passes the Windows CI.

gongsu832 (approved):
LGTM

src/Runtime/ExecutionSession.hpp
```diff
@@ -42,5 +43,8 @@ class ExecutionSession {
 
   // Entry point function.
   entryPointFuncType _entryPointFunc = nullptr;
+
+  // Input signature function
+  inputSignatureType _inputSignatureFunc = nullptr;
```
sstamenova, line 48: Since this is never used in the base class, it should not be here.

src/Runtime/ExecutionSession.hpp
```diff
@@ -25,6 +25,7 @@
 namespace onnx_mlir {
 
 typedef OMTensorList *(*entryPointFuncType)(OMTensorList *);
+typedef const char *(*inputSignatureType)();
```
sstamenova, line 28: Since this is never used in the base class, it should not be here.

sstamenova (changes requested):

src/Runtime/ExecutionSession.hpp
shizy818: @sstamenova I use `_inputSignatureFunc` in function `PyExecutionSession::pyInputSig`. It's just delay initializing only when it is necessary. 

```c++
py::str PyExecutionSession::pyInputSig() {
  std::string inputSignatureName = "omInputSignature";
  _inputSignatureFunc = reinterpret_cast<inputSignatureType>(
      _sharedLibraryHandle.getAddressOfSymbol(inputSignatureName.c_str()));
  assert(_inputSignatureFunc && "Input Signature func not loaded.");

  const char *outputs = _inputSignatureFunc();
  return outputs ? py::str(outputs) : "";
}
```

Since `inputSignatureType` definition is diffrent from `entryPointFuncType`, I cannot reuse `entryPointFuncType` in signature testcase.

src/Runtime/ExecutionSession.cpp
AlexandreEichenberger: > Since the name CAN be set, it is not always omInputSignature, so this error needs to reflect the variable not a hard-coded value.

Good point, it is set as a static constant, so it cannot be arbitrarily set during execution, but you are correct, it would be better to reuse the variable in the error message.

AlexandreEichenberger: > If this is not needed everywhere (e.g. the numerical tests don't need this, but JNI does), don't try to load the function always, but only if the name is set.

Is the solution you prefer is to set the pointer to null by default, and then before using it, we first check if its null and when it is, we load it then? Easy to do if that is what you are requesting @sstamenova

AlexandreEichenberger:
Ok, the confusion here is that many of the changes here are already in the Runtime/ExecutionSession. Apologies if I pushed these changes ahead of your PR, I was simply updating our examples in the doc (to have a end-to-end example using both the C++ and the Python interface), and though that the Python interface could use the addition of a get-signature functionality. 

If you were to prefer your Python interface names:
```
      .def("inputSig", &onnx_mlir::PyExecutionSession::pyInputSig);
```
instead of the one I picked (and currently in the main branch)
```
      .def("input_signature", &onnx_mlir::PyExecutionSession::pyInputSignature)
      .def("output_signature",
          &onnx_mlir::PyExecutionSession::pyOutputSignature);
```
please also update the example (which is not run in the CIs) here: (from docs/mninst_example/mninst.py)
```
print("input signature in json", session.input_signature())
print("output signature in json",session.output_signature())
```

Note that the Python convention seems to prefer "_" names (https://www.python.org/dev/peps/pep-0008/)

src/Runtime/ExecutionSession.hpp
sstamenova: Your usage is not in the base class, so the declaration does not need to be in the base class. You should move the declaration to `PyExecutionSession`.

src/Runtime/ExecutionSession.cpp
sstamenova: It depends on what the goal is. If the expectation is that this function (two, in the other PR) will only be used sometimes, then there should be a different mechanism to find them. For example, instead of initializing them immediately, initializing them in a separate function that gets invoked when needed.

I think, though, that this PR is obsolete with your other change, is it not?

shizy818:
> Ok, the confusion here is that many of the changes here are already in the Runtime/ExecutionSession. Apologies if I pushed these changes ahead of your PR, I was simply updating our examples in the doc (to have a end-to-end example using both the C++ and the Python interface), and though that the Python interface could use the addition of a get-signature functionality.
> 
> If you were to prefer your Python interface names:
> 
> ```
>       .def("inputSig", &onnx_mlir::PyExecutionSession::pyInputSig);
> ```
> 
> instead of the one I picked (and currently in the main branch)
> 
> ```
>       .def("input_signature", &onnx_mlir::PyExecutionSession::pyInputSignature)
>       .def("output_signature",
>           &onnx_mlir::PyExecutionSession::pyOutputSignature);
> ```
> 
> please also update the example (which is not run in the CIs) here: (from docs/mninst_example/mninst.py)
> 
> ```
> print("input signature in json", session.input_signature())
> print("output signature in json",session.output_signature())
> ```
> 
> Note that the Python convention seems to prefer "_" names (https://www.python.org/dev/peps/pep-0008/)

Hi @AlexandreEichenberger, I will fix the conflict using your interface name. I'm sorry this PR lasts a long time. I think I do almost same change like PR https://github.com/onnx/onnx-mlir/pull/1037/files, but It's failing Windows CI. Anyway, I will make sure Windows test run clean tonight at home.

AlexandreEichenberger (changes requested):
LGTM, significant changes to test the signature, I wonder if we could have architected the addition of these tests without as much changes to the existing infra, but since it is done, its ok. 

Could you clarify if the commands to do the backend test are changing? Right now we have 
* check_onnx_backend
* check_onnx_backend_dynamic
* check_onnx_backend_constant
where we can also add `_jni` to test the JNI version of it.

How does the invocations of these new signatures test work with that? Is it a separate target?
Also, right now we can use the `TEST_CASE_BY_USER=selected_test_name` to force a specific test. How does that work (if same make target is used).

So my request is to add documentation if there are changes on how the test are run from a user's perspective. docs/Testing.md would be where to add this additional info.

gongsu832:
The large changes are not just for the signature tests. I specifically asked Zhen Yu to clean up test.py since it was quite disorganized.

shizy818:
@AlexandreEichenberg, following tests should work the same as I run them locally.
- check_onnx_backend
- check_onnx_backend_dynamic
- check_onnx_backend_constant
- check_onnx_backend_xxx_jni

And I add input signature test.
- check_onnx_backend_signature

I do not try `TEST_CASE_BY_USER=selected_test_name` locally. Maybe can someone help to confirm it? 

In fact I do not change much of the logic except that I try to modulize it:
- varaibles.py: all global variables
- common.py: some functions used by xxxExecuteSession
- inference_backend.py: class InferenceBack &EndiannessAwareExecutionSession used for inference testing
- signature_backend.py: class SignatureBackendTest  (Inherit BackendTest for signature testing) & SignatureBackend & SignatureExecutionSession
- test.py: the main process

AlexandreEichenberger:
@shizy818 @gongsu832 great, thanks for the explanation, I would suggest to copy this comment to the top of the main test file for future references

> In fact I do not change much of the logic except that I try to modulize it:
> varaibles.py: all global variables
> common.py: some functions used by xxxExecuteSession
> inference_backend.py: class InferenceBack &EndiannessAwareExecutionSession used for inference testing
> signature_backend.py: class SignatureBackendTest (Inherit BackendTest for signature testing) & SignatureBackend &  SignatureExecutionSession
> test.py: the main process

Add this to the new Testing.md file please

> And I add input signature test.
> check_onnx_backend_signature

test/backend/CMakeLists.txt
```diff
@@ -10,6 +10,26 @@ file(GENERATE
   INPUT ${CMAKE_CURRENT_SOURCE_DIR}/test.py
   )
```
sstamenova, line 12: Is `conftest.py` used anywhere at this point?

shizy818: Actually I do not know how `conftest.py` is used. @gongsu832 Can you explain a bit about this?

shizy818:
@AlexandreEichenberger I add comment in py.test and `check_onnx_backend_signature` Testing.md file

test/backend/CMakeLists.txt
gongsu832: It's used by pytest. In this particular case, it's used to add a pytest option to remove the test progress report that prints a lot of hash, dot, etc. that clutter up the outputs.

AlexandreEichenberger (approved):
LGTM, thanks for the additional doc

test/backend/CMakeLists.txt
shizy818: @sstamenova Based on gong's explaination, I think we should not remove conftest.py here. You still think we need changes here?

AlexandreEichenberger:
@sstamenova, is the PR good for you now?

test/backend/CMakeLists.txt
sstamenova: No, thanks for looking into it.

sstamenova:
LGTM.

sstamenova (approved):

AlexandreEichenberger:
@sstamenova thanks

shizy818:
@caoimhinuibrian Can you do a favor to review the code? Thanks!

PR #994, opened 2021-11-12
support for multiple dialect builder object

AlexandreEichenberger:
Instead of creating multiple builders, e.g.

```C++
  MathBuilder createMath(createKrnl);
  MemRefBuilder createMemRef(createKrnl);
```
and then using them like this

```C++
  createKrnl.defineLoop(1);
  createMath.add(i1, i2);
  createMemRef.alloca(type);
```

we can create a single builder composed of multiple types and then as follows.

```C++
  MultiDialectBuilder<KrnlBuilder, MathBuilder, MemRefBuilder>
    create(rewriter, loc);

  create.krnl.defineLoop(1);
  create.math.add(i1, i2);
  create.mem.alloca(type);
```

Changed usage where it made sense in the code, and added doc to describe new feature.

### DIFF
--- a/src/Conversion/ONNXToKrnl/Math/Gemm.cpp
+++ b/src/Conversion/ONNXToKrnl/Math/Gemm.cpp
@@ -47,13 +47,12 @@ struct ONNXGemmOpLowering : public ConversionPattern {
         [&](KrnlBuilder &createKrnl, ValueRange outerIndices) {
           // Create temp and set to zero, single scalar, no need for default
           // alignment.
-          MemRefBuilder createMemRef(createKrnl);
-          MathBuilder createMath(createKrnl);
-          Value red = createMemRef.alloca(MemRefType::get({}, elementType));
+          MultiDialectBuilder<MemRefBuilder, MathBuilder> create(createKrnl);
+          Value red = create.mem.alloca(MemRefType::get({}, elementType));
           createKrnl.store(zeroVal, red);
           // Inner loop
           ValueRange innerLoop = createKrnl.defineLoops(1);
-          Value innerLb = createMath.constantIndex(0);
+          Value innerLb = create.math.constantIndex(0);
           Value innerUb = shapeHelper.aDims[1].getValue();
           createKrnl.iterate(innerLoop, innerLoop, {innerLb}, {innerUb},
               [&](KrnlBuilder &createKrnl, ValueRange innerIndex) {
@@ -79,7 +78,7 @@ struct ONNXGemmOpLowering : public ConversionPattern {
           // Handle alpha/beta coefficients.
           // new scope
           IndexExprScope innerScope(createKrnl, shapeHelper.scope);
-          Value res = createMath.mul(alphaVal, createKrnl.load(red));
+          Value res = create.math.mul(alphaVal, createKrnl.load(red));
           if (shapeHelper.hasBias) {
             SmallVector<Value, 2> cAccess;
             for (int x = 2 - shapeHelper.cRank; x < 2; ++x) {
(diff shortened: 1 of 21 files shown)


src/Conversion/ONNXToKrnl/Math/Gemm.cpp
```diff
@@ -47,13 +47,12 @@ struct ONNXGemmOpLowering : public ConversionPattern {
         [&](KrnlBuilder &createKrnl, ValueRange outerIndices) {
           // Create temp and set to zero, single scalar, no need for default
           // alignment.
-          MemRefBuilder createMemRef(createKrnl);
-          MathBuilder createMath(createKrnl);
-          Value red = createMemRef.alloca(MemRefType::get({}, elementType));
+          MultiDialectBuilder<MemRefBuilder, MathBuilder> create(createKrnl);
+          Value red = create.mem.alloca(MemRefType::get({}, elementType));
```
tungld, line 51: In this case, I am thinking it looks natural to support `create.krnl` because `create` was created from `createKrnl`. It's helpful so we support it.

tungld (commented):
MultiDialectBuilder is quite useful. Just have one comment to make it more convenient. Thanks!

src/Conversion/ONNXToKrnl/Math/Gemm.cpp
AlexandreEichenberger: That is exactly what I wanted to do, and I think I know how to do it :-)

AlexandreEichenberger: @tungld After further reflection, I don't think its possible. The entire scheme works on template specialization, meaning that between the
```
template<Ts...>
struct MultiDialectBuilder<MathBuilder, Ts>
```

and 
```
template<Ts...>
struct MultiDialectBuilder <MemRefBuilder, Ts>
```

I can generate a different specialization of `MultiDialectBuilder`, one that includes respectively the `math` object member and the other one the `men` object member.

However, the type used in the constructor cannot be used for specialization, its essentially a type of one of the function. This is not part of the signature type of the overall class, and thus I cannot specialize the type for it. 

Another difficulty is that all subclasses, such as `ONNXBuilder` and `KrnlBuilder` are not defined in the MLIRDialectBuilder; specialization can be added though, and it works, but I cannot "add" all of the variants upfront in the MultDialectBuilder.

So, we I believe we  must add each of the desired builder explicitly. 

I listed below a small program that I used to test the scheme. Feel free to experiment
```C++
#include <stdio.h>

class Top {};

class A: public Top {
public:
  A(int i) : val(i) {}
  A(A &aa) : val(aa.val) {}

  int getVal() { return val; }

private:
  int val;
};

class B: public Top {
public:
  B(int i) : val(10 + i)  {}
  B(B &bb) : val(bb.val) {}
  int getVal() { return val; }

private:
  int val;
};

class C: public Top {
public:
  C(int i) : val(100 + i) {}
  C(C &cc) : val(cc.val) {}
  int getVal() { return val; }

private:
  int val;
};

template <class... Ts>
class Builder {
public:
  Builder(int i) {}
};

template <class... Ts>
class Builder<A, Ts...> : public Builder<Ts...> {
public:
  Builder(int i) : Builder<Ts...>(i), a(i) {}
  A a;
};

template <class... Ts>
class Builder<B, Ts...> : public Builder<Ts...> {
public:
  Builder(int i) : Builder<Ts...>(i), b(i) {}
  B b;
};

template <class... Ts>
class Builder<C, Ts...> : public Builder<Ts...> {
public:
  Builder(int i) : Builder<Ts...>(i), c(i) {}
  C c;
};

int main() {
#if 1
  Builder<B, A, C> builder(1);
#else
  Builder<A, B, C> builder(1);
#endif

  int aa = builder.a.getVal();
  printf("aa is %d\n", aa);
  int bb = builder.b.getVal();
  printf("bb is %d\n", bb);
  int cc = builder.c.getVal();
  printf("cc is %d\n", cc);
  return 1;
}
```

AlexandreEichenberger: @etiotto maybe you have an idea if @tungld idea is feasible. Happy to hear ideas

etiotto (approved):
LGTM

tungld (approved):
LGTM.

PR #1000, opened 2021-11-16
Added MemRefsNormalizable trants to krnl ops whose input or output has memref type

imaihal:
I want to add `MemRefsNormalizable` trait in `krnl.matmul` to normalize memrefs in my use case. We usually add the trait every time we encounter a situation where it is needed.  I did this before in these PRs,  krnl.memcpy: https://github.com/onnx/onnx-mlir/pull/322,  krnl.getref: https://github.com/onnx/onnx-mlir/pull/327 .

However, this approach is ad-hoc and may introduce potential bugs (I created this PR because I have a use case where normalization didn't work). To avoid this, this PR creates base class for krnl ops and set `MemRefsNormalizable` trait by default in krnl ops.

Is this acceptable? If not, I will add the trait only in `krnl.matmul`.

### DIFF
--- a/src/Dialect/Krnl/KrnlOps.td
+++ b/src/Dialect/Krnl/KrnlOps.td
@@ -26,15 +26,10 @@ def Krnl_Dialect : Dialect {
   let cppNamespace = "::mlir";
 }
 
-// Base class for Krnl dialect ops.
-class Krnl_Op<string mnemonic, list<OpTrait> traits = []> :
-    Op<Krnl_Dialect, mnemonic, !listconcat(traits,[MemRefsNormalizable])> {
-}
-
 // Require regions to have krnl.terminate terminator operation.
 def ImplicitKrnlTerminator : SingleBlockImplicitTerminator<"KrnlTerminatorOp">;
 
-def KrnlDefineLoopsOp : Krnl_Op<"define_loops"> {
+def KrnlDefineLoopsOp : Op<Krnl_Dialect, "define_loops"> {
   let summary = "define_loops operation";
   let description = [{
     The "krnl.define_loops" operation is used to define input loops,
@@ -63,7 +58,7 @@ def KrnlDefineLoopsOp : Krnl_Op<"define_loops"> {
 }];
 }
 
-def KrnlIterateOp : Krnl_Op<"iterate", [ImplicitKrnlTerminator,
+def KrnlIterateOp : Op<Krnl_Dialect, "iterate", [ImplicitKrnlTerminator,
     DeclareOpInterfaceMethods<LoopLikeOpInterface>]> {
   let summary = "iterate operation";
   let description = [{
@@ -134,7 +129,7 @@ def KrnlIterateOp : Krnl_Op<"iterate", [ImplicitKrnlTerminator,
     let verifier = [{ return ::verify(*this); }];
 }
 
-def KrnlTerminatorOp : Krnl_Op<"terminate", [Terminator]> {
+def KrnlTerminatorOp : Op<Krnl_Dialect, "terminate", [Terminator]> {
   let summary = "Krnl terminator operation";
   let description = [{
     Krnl terminator is a special terminator operation for blocks inside krnl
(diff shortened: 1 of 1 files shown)


src/Dialect/Krnl/KrnlOps.td
```diff
@@ -26,10 +26,15 @@ def Krnl_Dialect : Dialect {
   let cppNamespace = "::mlir";
 }
 
+// Base class for Krnl dialect ops.
+class Krnl_Op<string mnemonic, list<OpTrait> traits = []> :
+    Op<Krnl_Dialect, mnemonic, !listconcat(traits,[MemRefsNormalizable])> {
+}
+
```
AlexandreEichenberger, line 33: Are we sure that all the memref are always normalizable in every of the the Krnl dialect operations?

imaihal: @AlexandreEichenberger Actually I don't know of any examples where memrefs cannot be normalized. I think you or Uday added `MemRefsNormalizable` trait in MLIR. Do you know what kinds of the example can't be normalized?

AlexandreEichenberger: I know I added it :-) but I don't recall precisely if everything can be normalized. I re-read the definition I wrote, quoted below.

----------------------

### MemRefsNormalizable

*   `OpTrait::MemRefsNormalizable` -- `MemRefsNormalizable`

This trait is used to flag operations that consume or produce values of `MemRef`
type where those references can be 'normalized'. In cases where an associated
`MemRef` has a non-identity memory-layout specification, such normalizable
operations can be modified so that the `MemRef` has an identity layout
specification. This can be implemented by associating the operation with its own
index expression that can express the equivalent of the memory-layout
specification of the MemRef type. See [the -normalize-memrefs pass].
(https://mlir.llvm.org/docs/Passes/#-normalize-memrefs-normalize-memrefs)

--------------------

So I think as long as the memory maps are ok, we should be fine. I was just surprised that we would assign this flag to all operations, but it might be ok. For example, what does it mean for a `krnl.define_loop` to be flagged as memref-normalizable when they don't have memrefs. But again, that might be ok.

Maybe someone else has an opinion on this, @tungld @doru1004 ?

tungld: It looks like this way is still adhoc. What if we have ops not from krnl dialect, but some dialects in MLIR and they don't have MemRefNormalize trait too. It means every ops we use must have this trait, which is troublesome. 

For long-term, if memory-layout is simple (e.g. only tiling), we can directly generate normalized MemRef types for ops of interest only. Thus, don't need to normalize any MemRefs. It is perhaps easy if we use IndexExpr.

imaihal: Thanks for your comments.

I thought the traits was added in all ops in AffineDialect using base class, but I realized the base class was not for all ops. It was added only in  [AffineLoadOpBase](https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Dialect/Affine/IR/AffineOps.td#L482) and [AffineStoreOpBase](https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Dialect/Affine/IR/AffineOps.td#L841) and some other ops. 
So, would it be better to add only Krnl dialect ops whose input is memref as before?

AlexandreEichenberger:
@imaihal Please let me know if you still want to go ahead with this PR. The feedback from @tungld and I is that it might be problematic to set all Krnl operations to be memrefnormalizable. Would you consider doing it only for the ops that you currently need? Or do you think that indeed, all Krnl ops should be memrefnormalizable? Thanks for your input on this.

imaihal:
@AlexandreEichenberger Thanks for your comments.

I also think adding the traits in all krnl ops is not appropriate. So, I added only ops whose input or output has memref type.

AlexandreEichenberger (approved):
LGTM, glad we could come to a solution acceptable to all

Issue #993, opened 2021-11-12
Doc for constants


AlexandreEichenberger:
> There is no document about the infrastructure for constant propagation at this time. Please look at the comment at the beginning of ConstProp.cpp and follow one of the examples.

Please add, even if short at this time

tungld:
> Please add, even if short at this time

Will have it soon, early next week. BTW, this is a duplication of #992 :)

PR #1004, opened 2021-11-17
Document about constant propagation

### LINKED ISSUE #992: Document about the infrastructure for Constant Propagation and how-to
Writing a guideline about constant propagation and how to support a new op.

tungld:
This patch adds a document about constant propagation and how to support a new ONNX op.

Resolves #992, #993


docs/ConstPropagationPass.md
```diff
@@ -0,0 +1,266 @@
...
+   - get the buffer from the buffer pool if it was created.
+2. `createConstantOpAndStoreBufferPtr(..., char *buffer)`
+   - create a new ONNXConstantOp using the given buffer, and
+   - add the buffer to the buffer pool.
+3. `allocateBufferFor(Value value, bool useMaxSize = false)`
+   - create a new buffer whose size is obtained from the type of 'value'.
+
+Note that:
+  - The buffers in the buffer pool will be automatically freed. Users don't
+    need to take care about that.
```
AlexandreEichenberger, line 66: Can you add a quick explanation of how the freeing is occurring? It is freed if its associated ONNXConstantOp is deleted?

docs/ConstPropagationPass.md
```diff
@@ -0,0 +1,266 @@
...
+DenseElementsAttr for the final results of constant propagation.
+
+In particular, we maintain a buffer pool internally. When an ONNXConstantOp
+is reached at the first time, we read its DenseElementsAttr and store
+data to an array buffer in the pool. A unique buffer ID is used to map the
+ONNXConstantOp to the buffer. All constant computations are then done on array
+buffers, which is to avoid creating DenseElementsAttr for intermediate
+ONNXConstantOps. Buffers are automatically freed if they are not used.
+
+We provide three helper functions to use when working with buffers:
```
AlexandreEichenberger, line 53: 1 seems to translate an array attribute into a buffer.
3 seems to create a new buffer
2 seems to associate a created buffer with an op.

I would reorder 2 & 3 as it seems more logical progression. What happen if we create a new buffer but don't associated it with an op? Does it get lost? (see request to have a bit more info for how the freeing of the buffers are done).

Is there ever a need to have an explicit freeing of a buffer? E.g. one that is never associated with an op?

docs/ConstPropagationPass.md
```diff
@@ -0,0 +1,266 @@
...
+by a new constant by adding the inputs at compile time. To check if an input is
+a constant, using ONNXConstantOp is not enough since the constant tensor can be
+sparse and we now support dense constant tensors only. We need additionallly
+check a dense constant tensor by using `IsFromDenseONNXConstantOp`.
+
+In the result pattern, to produce a ONNXConstantOp, we will add `lhs`
+and `rhs` at compile time, and emit an ONNXConstantOp. To minimize the
+memory footprint, **this ONNXConstant does not have a DenseElementsAttr**, but
+refers to an internal buffer where the real data is stored. DenseElementsAttrs
+will be added to only **the final ONNXConstantOps of the whold pass**,
```
AlexandreEichenberger, line 121: typo "whole"

docs/ConstPropagationPass.md
```diff
@@ -0,0 +1,266 @@
...
+   list<dag> additionalConstraints = [],
+   dag benefitsAdded = (addBenefit 0)
+>;
+```
+
+More information about DRR can be found [here](https://mlir.llvm.org/docs/DeclarativeRewrites/).
+
+There is a limitation in writing DRRs for `--constprop-onnx` pass so that the
+memory footprint is minimized, that is:
+- Never use ONNXConstant in the result patterns of a DRR, because this
```
AlexandreEichenberger, line 87: you mean ONNXConsantOp?

Could we not to a custom builder that takes a buffer as input and then use one in the rules?

docs/ConstPropagationPass.md
```diff
@@ -0,0 +1,266 @@
...
+    // Use int64_t to avoid the precision loss during computation.
+    IterateConstPropElementwiseBinary<ElementwiseBinaryOp, int64_t>(
+        lhsArray, rhsArray, lhsShape, rhsShape, resArray, outputShape);
+  } else
+    llvm_unreachable("Unknown data type");
+
+  // Construct a new ONNXConstantOp for the result array buffer.
+  // This ONNXConstantOp contains a buffer ID instead of a DenseElementsAttr.
+  ONNXConstantOp res =
+      createConstantOpAndStoreBufferPtr(rewriter, replacingValue, resArray);
```
AlexandreEichenberger, line 172: Could we have a custom builder instead of this custom function? I think it should work and then its better integrated?

docs/ConstPropagationPass.md
```diff
@@ -0,0 +1,266 @@
...
+  
+  // Return the constant.
+  return res;
+}
+```
```
AlexandreEichenberger, line 177: What if we had an FMA, and we decided to create first a buffer for multiplying, and then a second buffer for adding, and then only associate the second buffer with the constant op. Would the first buffer "escape" the freeing?

AlexandreEichenberger (commented):
Can you add this page also in the "CONTRIBUTING.md" page? Thanks

docs/ConstPropagationPass.md
tungld: > It is freed if its associated ONNXConstantOp is deleted?

Yes. I added an explanation.

docs/ConstPropagationPass.md
tungld: Thanks! It is better to reorder. Done.

> Is there ever a need to have an explicit freeing of a buffer? E.g. one that is never associated with an op?

Good question. I added an explanation for this situation. Yes, users need to take care of it and free it explicitly.

docs/ConstPropagationPass.md
tungld: Done. Thanks!

docs/ConstPropagationPass.md
tungld: Could we not to a custom builder that takes a buffer as input and then use one in the rules?

Good idea. I need to think more about how to associate it with the buffer pool. The buffer pool is not global now. It's a big change to make it global across multiple passes, but a huge benefit overall.

docs/ConstPropagationPass.md
tungld: Yes. I think so.

docs/ConstPropagationPass.md
tungld: We need to free the first buffer manually. I added an explanation about this in the Remark section.

tungld:
> Can you add this page also in the "CONTRIBUTING.md" page?

Thanks so much! I tried to find this page but could not remember where it it. I know it exists.

docs/ConstPropagationPass.md
AlexandreEichenberger: tx

docs/ConstPropagationPass.md
AlexandreEichenberger: tx

docs/ConstPropagationPass.md
AlexandreEichenberger: np, it works as is, just trying to see if we could fit it in the scheme of ops.

AlexandreEichenberger:
It exists, and it is in the top dir
https://github.com/onnx/onnx-mlir/blob/master/CONTRIBUTING.md

AlexandreEichenberger (approved):
LGTM after adding link to new page to the CONTRIBUTING page in the top dir.

Issue #1001, opened 2021-11-16
ONNX-MLIR Installation issue on macOS Monterey


Diksha-G:
I am trying to install ONNX-MLIR on macOS Monterey. I followed the steps for Installation on UNIX. I get the following error while building ONNX-MLIR:

[106/300] Building onnx.md...
FAILED: src/Dialect/ONNX/onnx.md /Users/dikshagupta/llvm-project/build/onnx-mlir/build/src/Dialect/ONNX/onnx.md 
cd /Users/dikshagupta/llvm-project/build/onnx-mlir/build && /usr/local/opt/llvm/bin/mlir-tblgen -gen-op-doc -I/Users/dikshagupta/llvm-project/build/onnx-mlir -I /Users/dikshagupta/llvm-project/build/onnx-mlir/src/Dialect/ONNX -I/usr/local/opt/llvm/include -I/usr/local/opt/llvm/include -I/Users/dikshagupta/llvm-project/build/onnx-mlir/build/src/Dialect/ONNX -I/Users/dikshagupta/llvm-project/build/onnx-mlir/build/src/Dialect/ONNX /Users/dikshagupta/llvm-project/build/onnx-mlir/src/Dialect/ONNX/ONNXOps.td --write-if-changed -o src/Dialect/ONNX/onnx.md -d src/Dialect/ONNX/onnx.md.d
Included from /Users/dikshagupta/llvm-project/build/onnx-mlir/src/Dialect/ONNX/ONNXOps.td:78:
/Users/dikshagupta/llvm-project/build/onnx-mlir/src/Dialect/ONNX/ONNXOps.td.inc:365:5: error: Expected a class name, got 'DefaultValuedStrAttr'
  DefaultValuedStrAttr<StrAttr, "NOTSET">:$auto_pad,
  ^
[123/300] Building CXX object third_pa...ir/onnx/version_converter/convert.cc.o
ninja: build stopped: subcommand failed.

I would greatly appreciate any help.

AlexandreEichenberger:
I am running on Monterey and have not had any new difficulties. For BigSur, I had to do this (see in the doc)

>If you are running on OSX Big Sur, you need to add -DCMAKE_CXX_COMPILER=/usr/bin/c++ to the cmake .. command due to changes in the compilers. After the above commands succeed, an onnx-mlir executable should appear in the bin directory.

Also grep the README for "mac" as there are sometimes a few things to do that are different.

[...]

PR #1008, opened 2021-11-17
Removed default building of onnx-mlir docs

AlexandreEichenberger:
Right now, the onnx-mlir docs (generating the dialect md files for onnx and krnl) is generated by default. But this may fail for users that did not install onnx third party tools. In any case, this target `onnx-mlir-docs` should only be called when changes were made to either Krnl or ONNX dialect. This is generally a rare event. So definitely does not belong to the default build process.

This will address issue #1001


doru1004 (approved):
LGTM

PR #1009, opened 2021-11-17
fixes frontend errors found in sequential.onnx

chentong319:

There are two issues caused by LoopOp:
1. The tensors in output list of graph may appear in the input list. Add the code to allow that.
2. The none constant was added to a wrong place. Previous code tried to generate one `none` for the whole function. But the `none` is generated at Op when it is first used. There are two LoopOps in sequential.onnx. The op in the second loop can not access `none` generated in the first loop body. Even though the `none` could be generated at the beginning of the function, I'd like to keep the code simple by generating none for each usage. The optimization will anyway optimize them out.

### DIFF
--- a/src/Builder/FrontendDialectTransformer.cpp
+++ b/src/Builder/FrontendDialectTransformer.cpp
@@ -102,8 +102,6 @@ private:
   ModuleOp module_;
   OpBuilder builder_;
 
-  Value none_;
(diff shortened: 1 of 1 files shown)


AlexandreEichenberger (approved):
LGTM, thanks for the very quick turn around, much appreciated.

PR #1015, opened 2021-11-19
Use polynomial approximation for tanh, sin, cos, log and exp

tungld:
Before this patch, `math.tanh` is expanded into `math.exp`, `math.sub`, `math.add`, and `math.div` by calling `populateExpandTanhPattern`, which is not efficient.

This patch calls `populateMathPolynomialApproximationPatterns` that does polynomial approximation, which leads to better inference performance.

Another difference is that `populateMathPolynomialApproximationPatterns` works with vector type while `populateExpandTanhPattern` does not.

LSTM and GRU use `math.tanh` as an activation, so they would become a little faster.

`populateMathPolynomialApproximationPatterns` also supports sin, cos, log, and exp.

### DIFF
--- a/src/Conversion/KrnlToLLVM/CMakeLists.txt
+++ b/src/Conversion/KrnlToLLVM/CMakeLists.txt
@@ -7,6 +7,7 @@ add_onnx_mlir_library(OMKrnlToLLVM
   OMSupport
   MLIRAffineToStandard
   MLIRMathToLLVM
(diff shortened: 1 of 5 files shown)


AlexandreEichenberger (approved):
LGTM

AlexandreEichenberger:
@tungld Please don't forget that we are now merging against main, not master anymore.

PR #1021, opened 2021-11-22
Generate datalayout in module.

etiotto:
This PR sets the datalayout for the generate LLVM IR module. The datalayout is set according to the effective value of the `triple`. used to compile the input MLIR file. With this PR:

```
 % ./Debug/bin/onnx-mlir ~/tmp/test.mlir  --preserveBitcode
Shared library /Users/etiotto@ca.ibm.com/tmp/test.so has been compiled.
% ~/Source/onnx/llvm-project/build/bin/llvm-dis  ~/tmp/test.bc
% head  ~/tmp/test.ll 
; ModuleID = '/Users/etiotto@ca.ibm.com/tmp/test.bc'
source_filename = "LLVMDialectModule"
target datalayout = "e-m:o-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
target triple = "x86_64-apple-darwin20.6.0"

@c = internal constant [6 x i32] [i32 1, i32 2, i32 3, i32 4, i32 5, i32 6]

declare i8* @malloc(i64)

declare void @free(i8*)
```

### DIFF
--- a/MLIR.cmake
+++ b/MLIR.cmake
@@ -110,6 +110,7 @@ endfunction()
 #     Same semantics as target_link_libraries().
 #   LINK_COMPONENTS llvm_components...
 #     Link the specified LLVM components.
+#     Note: only one linkage mode can be specified.
 #   )
 function(add_onnx_mlir_library name)
   cmake_parse_arguments(ARG
--- a/MLIR.cmake
+++ b/MLIR.cmake
@@ -92,7 +92,7 @@ endfunction()
 
 # add_onnx_mlir_library(name sources...
 #   This function (generally) has the same semantic as add_library. In
-#   addition is supports the arguments below and it does the following
+#   addition it supports the arguments below and it does the following
 #   by default (unless an argument overrides this):
 #   1. Add the library
 #   2. Add the default target_include_directories
@@ -108,12 +108,14 @@ endfunction()
 #     Same semantics as target_include_directories().
 #   LINK_LIBS lib_targets...
 #     Same semantics as target_link_libraries().
+#   LINK_COMPONENTS llvm_components...
+#     Link the specified LLVM components.
 #   )
 function(add_onnx_mlir_library name)
   cmake_parse_arguments(ARG
     "EXCLUDE_FROM_OM_LIBS;NO_INSTALL"
     ""
-    "DEPENDS;INCLUDE_DIRS;LINK_LIBS"
+    "DEPENDS;INCLUDE_DIRS;LINK_LIBS;LINK_COMPONENTS"
     ${ARGN}
     )
 
@@ -140,6 +142,21 @@ function(add_onnx_mlir_library name)
 
   if (ARG_LINK_LIBS)
     target_link_libraries(${name} ${ARG_LINK_LIBS})
+  endif()
+
+  if (ARG_LINK_COMPONENTS)
+    set(LinkageMode)
+    if (ARG_LINK_COMPONENTS MATCHES "^(PUBLIC|PRIVATE|INTERFACE)")
+      list(POP_FRONT ARG_LINK_COMPONENTS LinkageMode)
+    endif()
+
+    llvm_map_components_to_libnames(COMPONENT_LIBS ${ARG_LINK_COMPONENTS})
+
+    if (LinkageMode)
+      target_link_libraries(${name} ${LinkageMode} ${COMPONENT_LIBS})
+    else()
+      target_link_libraries(${name} PRIVATE ${COMPONENT_LIBS})
+    endif()
   endif()
 
   if (NOT ARG_NO_INSTALL)
--- a/src/Compiler/CMakeLists.txt
+++ b/src/Compiler/CMakeLists.txt
@@ -43,6 +43,14 @@ add_onnx_mlir_library(CompilerUtils
   MLIRAffineTransforms
   MLIRLinalgTransforms
   MLIRLLVMToLLVMIRTranslation
+
+  # Link LLVM libraries necessary to query which target architectures are configured.
+  LINK_COMPONENTS PRIVATE
+  AllTargetsDescs 
+  AllTargetsInfos 
+  AllTargetsCodeGens 
+  AllTargetsAsmParsers 
+  MC
   )
 
 # CompilerUtils does not require cruntime or jniruntime to build, however, they are
--- a/src/Compiler/CMakeLists.txt
+++ b/src/Compiler/CMakeLists.txt
@@ -46,10 +46,10 @@ add_onnx_mlir_library(CompilerUtils
 
   # Link LLVM libraries necessary to query which target architectures are configured.
   LINK_COMPONENTS PRIVATE
-  AllTargetsDescs 
-  AllTargetsInfos 
-  AllTargetsCodeGens 
-  AllTargetsAsmParsers 
+  AllTargetsAsmParsers
+  AllTargetsCodeGens
+  AllTargetsDescs
+  AllTargetsInfos
   MC
   )
 
--- a/src/Compiler/CompilerUtils.cpp
+++ b/src/Compiler/CompilerUtils.cpp
@@ -2,7 +2,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-//===--------------------------- MainUtils.cpp ---------------------------===//
+//===-------------------------- CompilerUtils.cpp -------------------------===//
 //
 // Copyright 2019-2020 The IBM Research Authors.
 //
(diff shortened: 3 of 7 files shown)


etiotto:

etiotto:

src/Compiler/CompilerUtils.cpp
```diff
@@ -713,8 +715,50 @@ void emitOutputFiles(string outputBaseName, EmissionTargetType emissionTarget,
...
+static std::string getDataLayout(const Location &loc) {
+  const std::string targetTriple =
+      (mtriple != "") ? mtriple.getValue() : kDefaultTriple;
+  const std::string targetCpu = (mcpu != "") ? mcpu.getValue() : "";
+
+  std::string error;
+  const llvm::Target *LLVMTarget =
+      llvm::TargetRegistry::lookupTarget(targetTriple, error);
+  if (!LLVMTarget) {
+    emitError(loc, Twine("failed to lookup target: ") + error);
```
AlexandreEichenberger, line 729: This look like an error message for compiler writer. Could we have one that is more user friendly, like "this machine is unknown... or something like that?"

AlexandreEichenberger (changes requested):

src/Compiler/CMakeLists.txt
```diff
@@ -24,6 +24,14 @@ add_custom_target(ExternalUtil DEPENDS ${FILE_GENERATE_DIR}/ExternalUtil.hpp)
 
 get_property(OMLibs GLOBAL PROPERTY ONNX_MLIR_LIBS)
 
+# Link LLVM libraries necessary to query which target architectures are configured.
+set(ExtraLLVMLibs)
+list(APPEND ExtraLLVMLibs "LLVMMC")
+llvm_expand_pseudo_components(TARGET_LIBS AllTargetsDescs AllTargetsInfos AllTargetsCodeGens AllTargetsAsmParsers)
```
sstamenova, line 30: Per the llvm cmake files, the function that should be used here is `llvm_map_components_to_libnames` as it is intended for the final user instead of `llvm_expand_pseudo_components`

src/Compiler/CMakeLists.txt
```diff
@@ -24,6 +24,14 @@ add_custom_target(ExternalUtil DEPENDS ${FILE_GENERATE_DIR}/ExternalUtil.hpp)
 
 get_property(OMLibs GLOBAL PROPERTY ONNX_MLIR_LIBS)
 
+# Link LLVM libraries necessary to query which target architectures are configured.
+set(ExtraLLVMLibs)
+list(APPEND ExtraLLVMLibs "LLVMMC")
```
sstamenova, line 29: MC can also be resolved to LLVMMC by using the same component function as below.

I would suggest adding an option to `add_onnx_mlir_library` to make the resolve for all of these happen automatically - this way it could be used in any of the libs and not just here.

sstamenova: `LLVMSupport` is an example of another llvm library that is used multiple places and could be specified as just `Support` with the proposed change.

src/Compiler/CompilerUtils.cpp
```diff
@@ -713,8 +715,50 @@ void emitOutputFiles(string outputBaseName, EmissionTargetType emissionTarget,
...
+
 int compileModule(mlir::OwningModuleRef &module, mlir::MLIRContext &context,
     std::string outputBaseName, EmissionTargetType emissionTarget) {
+  // Initialize the targets support for all targets LLVM was configured for.
+  llvm::InitializeAllTargets();
+  llvm::InitializeAllTargetMCs();
+  llvm::InitializeAllAsmPrinters();
+  llvm::InitializeAllAsmParsers();
+
+  // Set the module datalayout.
```
tungld, line 756: I wonder if we can postpone the insertion of data layout until generating LLVM IR, instead of do it at the beginning of the lowering pipeline.

For example, if our intermediate lowering, e.g. ONNX to Krnl, Krnl to Affine does not depend on data layout, then we don't need it at that time but until LLVMIR. This is useful when, e.g., we generate an intermediate IR in one machine for debugging, and then use it in other machines.

src/Compiler/CMakeLists.txt
etiotto: I added a new option to the `add_onnx_mlir_library` function as suggested.

AlexandreEichenberger (approved):

MLIR.cmake
```diff
@@ -108,12 +108,15 @@ endfunction()
 #     Same semantics as target_include_directories().
 #   LINK_LIBS lib_targets...
 #     Same semantics as target_link_libraries().
+#   LINK_TARGET_ARCH_LIBS
```
sstamenova, line 111: LLVM calls it LINK_COMPONENTS. If we want to keep the same nomenclature.

MLIR.cmake
```diff
@@ -139,7 +142,14 @@ function(add_onnx_mlir_library name)
     )
 
   if (ARG_LINK_LIBS)
-    target_link_libraries(${name} ${ARG_LINK_LIBS})
+    if (ARG_LINK_TARGET_ARCH_LIBS)
```
sstamenova, line 145: I was thinking that this option would take a list of libraries similar to LINK_LIBS and then map them as opposed to hardcoding all of the options in the function.

MLIR.cmake
```diff
@@ -139,7 +142,14 @@ function(add_onnx_mlir_library name)
...
+      llvm_map_components_to_libnames(TARGET_ARCH_LIBS AllTargetsDescs AllTargetsInfos AllTargetsCodeGens AllTargetsAsmParsers MC)
```
sstamenova, line 146: Does this not set `TARGET_ARCH_LIBS` in such a way that it can just be passed to `target_link_libraries` below?

sstamenova (changes requested):

MLIR.cmake
```diff
@@ -139,7 +142,14 @@ function(add_onnx_mlir_library name)
...
+      set(TargetArchLibs)      
+      foreach(t ${TARGET_ARCH_LIBS})
+        list(APPEND TargetArchLibs "${t}")
+      endforeach(t)
+    endif()
+    target_link_libraries(${name} ${ARG_LINK_LIBS} ${TargetArchLibs})
```
sstamenova, line 152: Also, ARG_LINK_LIBS might contain keywords such as private/public. By mixing the call and adding TargetArchLibs here, you're defaulting their visibility to the last keyword from ARG_LINK_LIBS. You should leave the target_link_libraries call for ARG_LINK_LIBS separate.

MLIR.cmake
```diff
@@ -139,7 +142,14 @@ function(add_onnx_mlir_library name)
     )
 
   if (ARG_LINK_LIBS)
```
sstamenova, line 144: You might need ARG_LINK_TARGET_ARCH_LIBS but not ARG_LINK_LIBS. Please separate them.

MLIR.cmake
etiotto: ok let me update it

sstamenova:
Btw, are you amending the commit each time you push? I'd love to be able to see the history and compare, but there's always only a single commit.

etiotto (commented):
Addressed code review comments from @sstamenova

src/Compiler/CompilerUtils.cpp
tungld: Got it. Thanks!

tungld (approved):
LGTM.

MLIR.cmake
etiotto: yes you are right, fixed. Good catch !

etiotto:
@sstamenova how does this PR looks now?

MLIR.cmake
```diff
@@ -142,6 +144,21 @@ function(add_onnx_mlir_library name)
     target_link_libraries(${name} ${ARG_LINK_LIBS})
   endif()
 
+  if (ARG_LINK_COMPONENTS)
+    set(LinkageMode)
+    if (ARG_LINK_COMPONENTS MATCHES "^(PUBLIC|PRIVATE|INTERFACE)")
```
sstamenova, line 149: I would add a comment in the description of the parameter that only one linkage mode can be specified.

src/Compiler/CMakeLists.txt
```diff
@@ -43,6 +43,14 @@ add_onnx_mlir_library(CompilerUtils
   MLIRAffineTransforms
   MLIRLinalgTransforms
   MLIRLLVMToLLVMIRTranslation
+
+  # Link LLVM libraries necessary to query which target architectures are configured.
+  LINK_COMPONENTS PRIVATE
+  AllTargetsDescs
```
sstamenova, line 49: Nit: can these be alphabetised?

sstamenova (approved):

sstamenova (approved):

MLIR.cmake
etiotto: Done.

src/Compiler/CMakeLists.txt
etiotto: Done.

PR #1022, opened 2021-11-24
Add optimization level to onnx-mlir

etiotto:
Add an `onnx-mlir` option to select the optimization level to use when compiling input files. Use the same optimization level to invoke  LLVM's `opt` and `llc`. The default optimization level, if unspecified, is -O0. 

### DIFF
--- a/src/Compiler/CompilerUtils.cpp
+++ b/src/Compiler/CompilerUtils.cpp
@@ -4,7 +4,7 @@
 
 //===-------------------------- CompilerUtils.cpp -------------------------===//
 //
-// Copyright 2019-2020 The IBM Research Authors.
+// Copyright 2019-2021 The IBM Research Authors.
 //
 // =============================================================================
 //
@@ -38,7 +38,7 @@ llvm::cl::OptionCategory OnnxMlirOptions(
 
 namespace {
 
-llvm::Optional<std::string> getEnvVar(std::string name) {
+static llvm::Optional<std::string> getEnvVar(std::string name) {
   if (const char *envVerbose = std::getenv(name.c_str()))
     return std::string(envVerbose);
   return llvm::None;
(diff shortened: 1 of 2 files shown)


chentong319 (commented):
What's the plan for the optimization level option? Will this option be used to control pass selection or  to be used inside onnx-mlir passes?

etiotto:
> What's the plan for the optimization level option? Will this option be used to control pass selection or to be used inside onnx-mlir passes?

Thinking we can used it in `onnx-mlir` to control which optimizations are run. For example with -O0 we would not run any optimizations, and with -O3 all optimizations. Also we might want to control the optimizations that are run by LLVM's `opt` and `llc` via this option.

chentong319:
If used inside a pass, the option has to be passed into the pass. There are two approaches: 
1. Declare and define the option in src/Support/OMOptions.* and use them directly in pass. But the option can not be used by onnx-mlir-opt. This is more of llvm way.
2. Declare option for the pass, and pass and construct the options. This approach is more mlir way, but requires more coding. 
Currently, we are using both and may choose only one in future.

AlexandreEichenberger:
I think this PR is very useful. Could we default to -O3 for now and have a later discussion whether to lower it to -O0?

AlexandreEichenberger:

AlexandreEichenberger:
FYI, While debugging another issue, I force the -O3 flag to be -O0 and it started failing some of the backend tests. I did not remove `opt`.... so that might be the problem, not sure.

AlexandreEichenberger:
@etiotto we discussed this issue during today's meeting and Kevin believe that since the default clang behavior is -O0, we should have the same policy for onnx-mlir.

So my suggestion is to go along with your default -O0, but change all the CIs to use the -O3 option. At some later time, we need to add an additional test run with -O0 as we don't want users trying our project without options and having it to bomb. This second testing can be added as a separate PR.

We can revisit this if you or others disagree...

etiotto:
@AlexandreEichenberger I agree in principle with the decision of making the default match what clang does (so -O0). 

I have also noticed that using -O0 some test fails so I agree that in this PR we can change the tests to explicitly use -O3. 
Longer term we should have lit tests that exercise other optimization levels.

AlexandreEichenberger:
@etiotto Is the `-O3` also used for the numerical benchmarks? I see the change for the backend tests but none for numerical

PR #1023, opened 2021-11-24
Add -v option to onnx-mlir

etiotto:
Add an option (-v) to print out the commands the `onnx-mlir` driver invokes.

### DIFF
--- a/src/Compiler/CompilerUtils.cpp
+++ b/src/Compiler/CompilerUtils.cpp
@@ -111,6 +111,10 @@ static llvm::cl::opt<OptLevel> OptimizationLevel(
         clEnumVal(O3, "Optimization level 3.")),
     llvm::cl::init(O0), llvm::cl::cat(OnnxMlirOptions));
 
+static llvm::cl::opt<bool> VerboseOutput("v",
+    llvm::cl::desc("Use verbose output"), llvm::cl::init(false),
+    llvm::cl::cat(OnnxMlirOptions));
+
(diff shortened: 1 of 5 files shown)


AlexandreEichenberger (approved):
LGTM

test/mlir/onnx/onnx_check_verbosity.mlir
```diff
@@ -0,0 +1,6 @@
+// REQUIRES: system-linux
```
sstamenova, line 1: Is there a reason this can't run on Windows?

etiotto: I can try to add the support. I do not have a Windows system to try on so might take a couple of runs on the CI to get it right (painful given it still takes too long to run CI on Win).

sstamenova: I see that you split the test into a Windows version and a Unix version. Does it make sense to update the checks to work for both (even though we'd lost some flags)? Alternatively, can we use separate check labels for Windows/Unix? I forget if that works.

etiotto: I don't think it makes sense to update the checks so that they work across Windows/Linux because there are too many differences, even the command names are different between the architectures. 
Still working on the test, I do not know how to use difference CHECK-LABELS if the CHECK depends on the system architecture.

sstamenova:
LGTM

PR #1025, opened 2021-11-26
Fix missleading error message when bad path to onnx model is given

negiyas:
Fix missleading error message when bad path to onnx model is given.

Currently, if the given .onnx file can't be found the emitted error could be improved. For example: (note model should be resnet50-v1-7.onnx but I'm intentionally making a typo)
"onnx-mlir resnet50-v1.onnx"
Gives
"Onnx Model Parsing Failed on resnet50-v1.onnx"
This current error message is misleading, because it implies the model file was found but is bad or incompatible.

This PR changes the error message to "unable to open or access resnet50-v1.onnx".

### DIFF
--- a/src/Builder/FrontendDialectTransformer.cpp
+++ b/src/Builder/FrontendDialectTransformer.cpp
@@ -1431,16 +1431,13 @@ void ImportFrontendModelArray(const void *onnxBuffer, int size,
 
 void ImportFrontendModelFile(std::string model_fname, MLIRContext &context,
     OwningModuleRef &module, std::string *errorMessage, ImportOptions options) {
-  // try to open the input file to check if it can be opened in advance.
-  std::fstream test;
-  test.open(model_fname);
-  if (!test) {
+  onnx::ModelProto model;
+  std::fstream input(model_fname, std::ios::in | std::ios::binary);
+  // check if the input file is opened
+  if (!input.is_open()) {
     *errorMessage = "Unable to open or access " + model_fname;
     return;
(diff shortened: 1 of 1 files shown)


AlexandreEichenberger (approved):
LGTM

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -1396,6 +1396,14 @@ void ImportFrontendModelArray(const void *onnxBuffer, int size,
 
 void ImportFrontendModelFile(std::string model_fname, MLIRContext &context,
     OwningModuleRef &module, std::string *errorMessage, ImportOptions options) {
+  // try to open the input file to check if it can be opened in advance.
+  std::fstream test;
+  test.open(model_fname);
+  if (!test) {
```
etiotto, line 1402: I think this should be:
```
if (!test.is_open()) {
```

Thinking more about this ... @negiyas it would be nice if we could avoid opening the file twice. Can you try to see if this would work:

```
  std::fstream input(model_fname, std::ios::in | std::ios::binary);
  if (!input.is_open()) {
    *errorMessage = "Unable to open or access " + model_fname;
    return;
  }
  ```

negiyas: @etiotto It works. Thank you very much for useful comments!

etiotto: @negiyas you are very welcome! Thanks for the fix.

PR #1026, opened 2021-11-29
import seq

chentong319:
This PR added the support to import seq type from onnx model.
1. Extend importTensorType to importType, which calls importTensorType and importSeqType according to the type.
2. Change function signature to allow seq type. The consumer of the signature need to be updated, too.


AlexandreEichenberger (approved):
LGTM

PR #1027, opened 2021-11-30
Make onnx-to-krnl patterns re-usable

tungld:
This patch makes onnx-to-krnl patterns re-usable by grouping all patterns into a single function. Hence, the patterns can be called from a lowering pass in another project.

This patch also fixes a typo in a link in the onnxai webpage.


AlexandreEichenberger (approved):
LGTM, thanks

etiotto (approved):
LGTM. Thank you!

PR #1028, opened 2021-11-30
Set krnl.global alignment using module datalayout if present

etiotto:
If `krnl.global` does not have an explicit alignment attribute sets the alignment based on the module's datalayout attribute (if present).


etiotto:
@AlexandreEichenberger let me know if you have comments.  I would like to address comments (if any) before rebasing.

tungld (approved):
LGTM! Thanks!

PR #1029, opened 2021-12-01
Index value capture for arrays of 0D

AlexandreEichenberger:
Case is needed as sometimes instead of a 1D array, we just get a single value. Now the index capture of such 0D array is enabled.


chentong319 (approved):
LGTM!

gongsu832:

PR #1033, opened 2021-12-02
Support shape inference for sequence ops

chentong319:
1. Allow shape inference to pass. This PR does not do aggressive shape inference on seq type. Future work.
2. Add verifier for SequenceInsert
3. Decompose SequenceConstruct to SequenceEmpty and SequenceInsert.


src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -867,6 +867,61 @@ LogicalResult ONNXSeluOp::inferShapes(
...
+  return success();
+}
+
+static LogicalResult verify(ONNXSequenceInsertOp op) {
+  ONNXSequenceInsertOpAdaptor operandAdaptor = ONNXSequenceInsertOpAdaptor(op);
+
+  // These cast should be guaranteed by default verifier
+  auto seqElementType = operandAdaptor.input_sequence()
+                            .getType()
+                            .dyn_cast<mlir::onnxmlir::SeqType>()
```
AlexandreEichenberger, line 881: I recommend that we do a "dyn_cast_or_null" and provide an error, or is it already guaranteed by the constraints on the types in the onnx*.td file?

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -867,6 +867,61 @@ LogicalResult ONNXSeluOp::inferShapes(
...
+                            .getElementType();
+  auto elementType1 = seqElementType.dyn_cast<ShapedType>().getElementType();
+  auto insertType = operandAdaptor.tensor().getType().dyn_cast<ShapedType>();
```
AlexandreEichenberger, line 884: Same here, don't we have to test if we don't have a shape type? See the patterns in the other verifiers

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -867,6 +867,61 @@ LogicalResult ONNXSeluOp::inferShapes(
...
+}
+
+LogicalResult ONNXSequenceAtOp::inferShapes(
+    std::function<void(mlir::Region &)> doShapeInference) {
+  return success();
+}
+
+LogicalResult ONNXSequenceConstructOp::inferShapes(
+    std::function<void(mlir::Region &)> doShapeInference) {
+  return success();
```
AlexandreEichenberger, line 906: Do you mind just adding a rational why we don't need to do any shape inference. You can copy and paste the same explanation for each... Helpful so that we remember that it is implemented and perform the desired action.

AlexandreEichenberger (commented):
Thanks for adding the verifier. Requesting a few minor changes.

src/Dialect/ONNX/ONNXOps.cpp
chentong319: Yes, it is guaranteed by the constraints in ONNXOps.td.inc.

src/Dialect/ONNX/ONNXOps.cpp
chentong319:
```
ShapedType is a common base class between Vector, UnrankedTensor, RankedTensor, and MemRef types because they share behavior and semantics around shape, rank, and fixed element type
```
All the input and output of operations in onnx-mlir dialect should be shapedType except for seq type. The default verifier should guarantee  that. This may not be true after lowering.

utils/gen_onnx_mlir.py
```diff
@@ -295,6 +295,7 @@
     'ReverseSequence',
     'ScatterND',
     'SpaceToDepth',
+    'SequenceInsert',
```
etiotto, line 298: Add in alphabetical order.

test/mlir/onnx/onnx_decompose.mlir
```diff
@@ -297,3 +297,15 @@ func @test_padv2(%arg0: tensor<1x3x224x224xf32>) -> tensor<*xf32> {
...
+    %0 = "onnx.SequenceConstruct"(%arg0, %arg1) : (tensor<*xf32>, tensor<*xf32>) -> !onnx.Seq<tensor<*xf32>>
+    return %0 : !onnx.Seq<tensor<*xf32>>
+// CHECK-LABEL:  func @test_seqence_construct_1
+// CHECK-SAME:   ([[PARAM_0_:%.+]]: tensor<*xf32>, [[PARAM_1_:%.+]]: tensor<*xf32>) -> !onnx.Seq<tensor<*xf32>> {
+// CHECK-DAG:       [[VAR_0_:%.+]] = "onnx.SequenceEmpty"() : () -> !onnx.Seq<tensor<*xf32>>
+// CHECK-DAG:       [[VAR_cst_:%.+]] = constant unit
+// CHECK:           [[VAR_1_:%.+]] = "onnx.SequenceInsert"([[VAR_0_]], [[PARAM_0_]], [[VAR_cst_]]) : (!onnx.Seq<tensor<*xf32>>, tensor<*xf32>, none) -> !onnx.Seq<tensor<*xf32>>
+// CHECK:           [[VAR_2_:%.+]] = "onnx.SequenceInsert"([[VAR_1_]], [[PARAM_1_]], [[VAR_cst_]]) : (!onnx.Seq<tensor<*xf32>>, tensor<*xf32>, none) -> !onnx.Seq<tensor<*xf32>>
+// CHECK:           return [[VAR_2_]] : !onnx.Seq<tensor<*xf32>>
+}
```
etiotto, line 310: nit: align on same column as `func`

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -867,6 +867,61 @@ LogicalResult ONNXSeluOp::inferShapes(
...
+}
+
+LogicalResult ONNXSequenceEraseOp::inferShapes(
+    std::function<void(mlir::Region &)> doShapeInference) {
+  return success();
+}
+
+// Output should an integer64
```
etiotto, line 919: typo: should an -> should be an

I do not understand the comment. The returned  type is `LogicalResult` what is the output in this context?

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -867,6 +867,61 @@ LogicalResult ONNXSeluOp::inferShapes(
...
+
+static LogicalResult verify(ONNXSequenceInsertOp op) {
+  ONNXSequenceInsertOpAdaptor operandAdaptor = ONNXSequenceInsertOpAdaptor(op);
+
+  // These cast should be guaranteed by default verifier
+  auto seqElementType = operandAdaptor.input_sequence()
+                            .getType()
+                            .dyn_cast<mlir::onnxmlir::SeqType>()
+                            .getElementType();
+  auto elementType1 = seqElementType.dyn_cast<ShapedType>().getElementType();
```
etiotto, line 883: Suggest to use  `Type` rather than auto.

src/Transform/ONNX/Decompose.cpp
```diff
@@ -84,6 +84,18 @@ DenseElementsAttr createDenseArrayAttrOrEmpty(
   }
 }
 
+Value createSequenceConstructOp(
+    PatternRewriter &rewriter, mlir::Value seq, mlir::OperandRange inputs) {
+  auto resType = seq.getType();
```
etiotto, line 89: Suggest to use the correct type (`Type`, `Location`, etc...) rather than using `auto`. The reader will know immediately what is the type of the result, while using `auto` one is left to wonder.

utils/gen_onnx_mlir.py
AlexandreEichenberger: VSCode let you highlight code segment and then right click to order.. useful to me :-)

src/Dialect/ONNX/ONNXOps.cpp
AlexandreEichenberger: I agree, using types when its not crazy complicated (e.g. lambda) is best, think its LLVM policy anyway.

src/Dialect/ONNX/ONNXOps.cpp
chentong319: Comment removed. I should add a verifier for this op.

src/Dialect/ONNX/ONNXOps.cpp
chentong319: Comment added.

src/Dialect/ONNX/ONNXOps.cpp
chentong319: Fixed

test/mlir/onnx/onnx_decompose.mlir
chentong319: fixed

utils/gen_onnx_mlir.py
chentong319: Fixed

src/Transform/ONNX/Decompose.cpp
chentong319: fixed

src/Dialect/ONNX/ONNXOps.cpp
AlexandreEichenberger: I don't see them, maybe they were not pushed?

src/Transform/ONNX/Decompose.cpp
AlexandreEichenberger: same here, I don't see it.

AlexandreEichenberger (commented):
I see the comments stating that changes were made, but I don't see it. Do you mind checking that you pushed the changes? Apologies if I am not looking at the final version.

src/Dialect/ONNX/ONNXOps.cpp
chentong319: Comment (line 870-886) is quite long and valid for all the sequence related operations. I did not copy.

chentong319:
New version of Decompose.cpp is checked in.

AlexandreEichenberger (approved):
LGTM

AlexandreEichenberger:

PR #1034, opened 2021-12-02
Add optional wdir to exec()

gongsu832:
Add optional wdir to exec() to allow running a command
in a specified work directory. This is used to get around
the problem that old ar command does not support the
--output option.

### DIFF
--- a/src/Compiler/CompilerUtils.cpp
+++ b/src/Compiler/CompilerUtils.cpp
@@ -264,8 +264,8 @@ struct Command {
 
     // If in verbose mode, print out command before execution.
     if (verbose)
-      cout << "[" << StringRef(new_wdir).str() << "]"
-	   << _path << ": " << llvm::join(argsRef, " ") << endl;
+      cout << "[" << StringRef(new_wdir).str() << "]" << _path << ": "
+           << llvm::join(argsRef, " ") << endl;
 
     std::string errMsg;
     int rc = llvm::sys::ExecuteAndWait(_path, llvm::makeArrayRef(argsRef),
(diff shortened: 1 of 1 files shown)


src/Compiler/CompilerUtils.cpp
```diff
@@ -239,29 +239,49 @@ struct Command {
...
 
+    // If a work directory is specified, save the current work directory
+    // and switch into it. Note that if wdir is empty, new_wdir will be
+    // cur_wdir.
+    SmallString<8> cur_wdir;
+    SmallString<8> new_wdir(wdir);
+    llvm::sys::fs::current_path(cur_wdir);
+    llvm::sys::fs::make_absolute(cur_wdir, new_wdir);
+    if (std::error_code ec = llvm::sys::fs::set_current_path(new_wdir)) {
+      cerr << StringRef(new_wdir).str() << ": " << ec.message() << endl;
```
etiotto, line 261: Suggest to use `llvm::errs()`, I see is used to emit other error messages..

src/Compiler/CompilerUtils.cpp
```diff
@@ -239,29 +239,49 @@ struct Command {
...
     int rc = llvm::sys::ExecuteAndWait(_path, llvm::makeArrayRef(argsRef),
         /*Env=*/None, /*Redirects=*/None,
         /*SecondsToWait=*/0, /*MemoryLimit=*/0, &errMsg);
 
     if (rc != 0) {
-      fprintf(stderr, "%s\n", llvm::join(argsRef, " ").c_str());
-      fprintf(stderr, "Error message: %s\n", errMsg.c_str());
-      fprintf(stderr, "Program path: %s\n", _path.c_str());
-      fprintf(stderr, "Command execution failed.");
+      cerr << llvm::join(argsRef, " ") << endl
```
etiotto, line 276: Suggest to use `llvm::errs()`, I see is used to emit other error messages..

src/Compiler/CompilerUtils.cpp
AlexandreEichenberger: done, thanks Gong.

src/Compiler/CompilerUtils.cpp
AlexandreEichenberger: same here

AlexandreEichenberger (approved):
LGTM, @etiotto I believe all your requests are addressed.

etiotto:
Yup thanks @AlexandreEichenberger LGTM too.

PR #1037, opened 2021-12-03
Update py runtime

AlexandreEichenberger:
Added the ability to query the signature of a model in the python interface. 

Updated the doc to reflect the newer python interface, and augmented the MNIST example to also include the python interface.


AlexandreEichenberger:
@sstamenova fails on Window, either compiler crashes and/or dll not found. Strangely, it worked for the first test example of matmul but not the second one. Appears to be the same patterns for each of the numerical tests.

```
- Matmul implementation correctness
attempt 1 with i 1, j 1, k 1
Microsoft (R) C/C++ Optimizing Compiler Version 19.29.30137 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

Microsoft (R) Incremental Linker Version 14.29.30137.0
Copyright (C) Microsoft Corporation.  All rights reserved.

/out:./TestMatmul2D_main_graph.dll 
/dll 
/implib:./TestMatmul2D_main_graph.lib 
/def:./TestMatmul2D_main_graph.def 
"/libpath:D:\a\1\onnx-mlir\build\Release\lib" 
cruntime.lib 
./TestMatmul2D_main_graph.obj 
   Creating library ./TestMatmul2D_main_graph.lib and object ./TestMatmul2D_main_graph.exp
Shared library ./TestMatmul2D_main_graph.dll has been compiled.
attempt 2 with i 1, j 1, k 1
Microsoft (R) C/C++ Optimizing Compiler Version 19.29.30137 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

Microsoft (R) Incremental Linker Version 14.29.30137.0
Copyright (C) Microsoft Corporation.  All rights reserved.

/out:./TestMatmul2D_main_graph.dll 
/dll 
/implib:./TestMatmul2D_main_graph.lib 
/def:./TestMatmul2D_main_graph.def 
"/libpath:D:\a\1\onnx-mlir\build\Release\lib" 
cruntime.lib 
./TestMatmul2D_main_graph.obj 
LINK : fatal error LNK1104: cannot open file '.\TestMatmul2D_main_graph.dll'
cl.exe ./TestMatmul2D_main_graph.obj /Fe:./TestMatmul2D_main_graph.dll /LD /link /def:./TestMatmul2D_main_graph.def /libpath:"D:\a\1\onnx-mlir\build\Release\lib" cruntime.lib
Error message: 
Program path: C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/VC/Tools/MSVC/14.29.30133/bin/Hostx64/x64/cl.exe
Command execution failed.
```

Will restart to see if that was a transient problem.

src/Runtime/ExecutionSession.cpp
```diff
@@ -41,6 +41,25 @@ ExecutionSession::ExecutionSession(
     errStr << "Cannot load symbol: '" << entryPointName << "'" << std::endl;
     throw std::runtime_error(errStr.str());
   }
+
+  std::string inputSignatureName("omInputSignature");
```
sstamenova, line 45: My guess is your failure is coming from the code to load these two functions. Are they always needed now? If so, I think you need to add them to the .def file for the tests.

sstamenova, line 45: Also, I think it would be better to have their names as const members of the class, for example, rather than hardcoded here.

AlexandreEichenberger: done

AlexandreEichenberger: added, thanks for the suggestions. I could make it required only when asked for.

AlexandreEichenberger: @sstamenova you were right on the money, adding the functions names to the .def files was what was needed. It now passes the window tests, thanks for the help.

doru1004 (approved):
LGTM

src/Runtime/ExecutionSession.cpp
sstamenova: Great! I wish there was a good way to automate this, but I'm glad we figured it out quickly

PR #1039, opened 2021-12-05
PyRuntime shared lib protobuf error fix

gongsu832:
When running on ubi8 image, shared lib backend tests fail with the following error:

[libprotobuf ERROR google/protobuf/descriptor_database.cc:641] File already exists in database: onnx/onnx-ml.proto
[libprotobuf FATAL google/protobuf/descriptor.cc:1371] CHECK failed: GeneratedDatabase()->Add(encoded_file_descriptor, size):
terminate called after throwing an instance of 'google::protobuf::FatalException'
  what():  CHECK failed: GeneratedDatabase()->Add(encoded_file_descriptor, size):
Aborted (core dumped)

This is because test.py loads (among others) the following two .so shared libs:

- onnx_cpp2py_export.cpython-39-s390x-linux-gnu.so (import onnx)
- PyRuntime.cpython-39-s390x-linux-gnu.so (from PyRuntime import ExecutionSession)

Both libs share the same libprotobuf.so when loaded by test.py. However, they were
both built with the same onnx-ml.pb.cc generated from onnx-ml.proto and the protobuf
runtime requires all compiled-in .proto files have unique names. Hence the error.

PyRuntime doesn't really need onnx beyond the onnx::TensorProto::* types so
we remove onnx from its target_link_libraries. But that also removes some of the
compile definitions and include directories which we add back through
target_compile_definitions and target_include_directories.

I don't know yet why this didn't fail on Ubuntu Focal.

### DIFF
--- a/src/Compiler/CMakeLists.txt
+++ b/src/Compiler/CMakeLists.txt
@@ -60,7 +60,7 @@ if (ONNX_MLIR_ENABLE_JNI)
   add_dependencies(CompilerUtils jniruntime)
 endif()
 
-add_onnx_mlir_library(OnnxMlirCompiler SHARED
+add_onnx_mlir_library(OnnxMlirCompiler
   OnnxMlirCompiler.cpp
 
   EXCLUDE_FROM_OM_LIBS
--- a/src/Runtime/CMakeLists.txt
+++ b/src/Runtime/CMakeLists.txt
@@ -58,8 +58,16 @@ set_target_properties(ExecutionSession
   )
 
 pybind11_add_module(PyRuntime PyExecutionSession.cpp)
+target_compile_definitions(PyRuntime
+  PRIVATE
+  -DONNX_ML=1 -DONNX_NAMESPACE=onnx
+  )
+target_include_directories(PyRuntime
+  PRIVATE
+  $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/third_party/onnx>
+  $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}/third_party/onnx>
+  )
 target_link_libraries(PyRuntime
   PRIVATE
   ExecutionSession
-  onnx
   )
--- a/src/Runtime/PyExecutionSession.cpp
+++ b/src/Runtime/PyExecutionSession.cpp
@@ -14,7 +14,6 @@
 //===----------------------------------------------------------------------===//
 
 #include "onnx/onnx_pb.h"
-#include <third_party/onnx/onnx/onnx_pb.h>
 
 #include "PyExecutionSession.hpp"
 


src/Compiler/CMakeLists.txt
```diff
@@ -60,7 +60,7 @@ if (ONNX_MLIR_ENABLE_JNI)
   add_dependencies(CompilerUtils jniruntime)
 endif()
 
-add_onnx_mlir_library(OnnxMlirCompiler SHARED
+add_onnx_mlir_library(OnnxMlirCompiler
```
sstamenova, line 63: Why is this no longer a shared library? This was intentional when the library was created, so that it can be used by other (external) components as a shared library even when the rest of the libraries are built statically

src/Runtime/CMakeLists.txt
```diff
@@ -58,8 +58,16 @@ set_target_properties(ExecutionSession
   )
 
 pybind11_add_module(PyRuntime PyExecutionSession.cpp)
+target_compile_definitions(PyRuntime
```
sstamenova, line 61: Can you add some of the explanation of why these are added manually here? That way people looking at the cmake will know that they compile definitions are needed and why.

src/Compiler/CMakeLists.txt
gongsu832: A library doesn't have to be shared in order to be used by other components. Either a .a or a .so can be used by other components. Being shared simply means the symbols get resolved at runtime by the loader, while being static means the symbols get resolved at compile time by the linker.

LLVM builds all its libraries shared or static, depending on whether BUILD_SHARED_LIBS is ON or OFF. That's what ONNX-MLIR does too, i.e., all libraries should honor the BUILD_SHARED_LIBS option. The exception is libcruntime.a. It's always built as static because its symbols need to be embedded into the model.so.

sstamenova: Shared libraries and static libraries are used differently, especially DLLs on Windows. This library was added explicitly to be a shared library on all platforms, including Windows.

LLVM builds *most* of its libraries as either shared or static, but there are several similar libraries that are always either static or shared because they are special. This was the PR: https://github.com/onnx/onnx-mlir/pull/894

gongsu832: Yes I know it was added by PR#984. I just don't see any particular reason why it has to be shared if all you want is  other components can use it. "shared" is a somewhat misleading word. A more appropriate word is "dynamic". Static libraries are also shared. They are just shared at compile time rather than at load time.

Those special LLVM libraries that are always built static or shared must have reasons beyond simply being callable from other components. Like our libcruntime.a I mentioned, it must be always built static because its symbols need to be embedded into the model.so at compile time to make model.so self-contained.

The original words from @MikeHolman are "With a shared library we can simply give them the dll and not require them to add onnx-mlir and all its dependencies (in particular LLVM/MLIR) to their build." 

Perhaps a DLL is required on Windows but on other platforms a static lib would work fine as far as I can tell. If that's the case, we can special case for Windows. But on other platforms, OnnxMlirCompiler should honor BUILD_SHARED_LIBS option.

src/Runtime/CMakeLists.txt
gongsu832: I forgot to put the comments in the commit message here. I will add them in the next PR.

PR #1040, opened 2021-12-06
Fix a wrong file name for LRN shape inference

tungld:
The file for shape inference for LRN was wrongly named `RNN.cpp`. This patch changed it to `LRN.cpp`.


AlexandreEichenberger (approved):
LGTM

PR #1046, opened 2021-12-07
Build fix for mac

AlexandreEichenberger:
Mac have issues building the new runtime due to the inclusion of a new file, `google/protobuf/port_def.inc`. This issue appears to be specific to mac.

Fix below add the new path in the cmake file.


etiotto:
I add the same build issue on Mac today. I can confirm that the proposed fix solves the build issue on my Mac environment as well.

etiotto (approved):
LGTM

PR #1051, opened 2021-12-09
Now that O0 is default for models, force numerical to use O3 opt level

AlexandreEichenberger:


etiotto (approved):
LGTM

PR #1052, opened 2021-12-10
Refine Seq shape inference

chentong319:
Try to refine static shape info for seq.
1. The definition of SeqType is modified for static shape inference. It's very likely another field (memref) will be added for dynamic shape and lowering
2. Static shape starts from empty seq and the newly inserted tensor is merged with the existing tensor type of the seq.

### DIFF
--- a/src/Dialect/ONNX/ONNXOps.cpp
+++ b/src/Dialect/ONNX/ONNXOps.cpp
@@ -592,19 +592,13 @@ mlir::Type ONNXOpsDialect::parseType(mlir::DialectAsmParser &parser) const {
       return Type();
 
     SmallVector<mlir::Type, 1> elementTypes;
-    do {
-      // llvm::SMLoc typeLoc = parser.getCurrentLocation();
-      mlir::Type elementType;
-      if (parser.parseType(elementType))
-        return Type();
-
-      // TOFIX: type limitation for Seq? similar but different shape??
-      elementTypes.push_back(elementType);
-    } while (succeeded(parser.parseOptionalComma()));
+    mlir::Type elementType;
+    if (parser.parseType(elementType))
+      return Type();
 
     if (parser.parseGreater())
       return Type();
-    return onnxmlir::SeqType::get(elementTypes);
+    return onnxmlir::SeqType::get(elementType);
   }
 
   parser.emitError(parser.getNameLoc(), "unknown onnx type: " + keyword);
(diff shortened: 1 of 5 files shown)


src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -887,6 +881,35 @@ LogicalResult ONNXSeluOp::inferShapes(
...
+
+  auto seqShape = seqType.getElementType().cast<ShapedType>().getShape();
+  auto seqRank = seqType.getElementType().cast<ShapedType>().getRank();
+  if (seqRank == -1)
+    return success();
+
+  auto tensorShape = tensorType.getShape();
+  auto tensorRank = tensorType.getRank();
+  if (tensorRank != seqRank)
+    return success();
```
AlexandreEichenberger, line 900: Can you just add one comment to explain this. For example
```
// Because shape inference propagation can only be performed on tensors of identical ranks, we return success here, leaving the output type as undefined.
```

I also wonder if you don't have to put an exception when you merge an "empty" list with a first insert... in that case, would you not directly take the first insert's shape?

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -887,6 +881,35 @@ LogicalResult ONNXSeluOp::inferShapes(
...
+  SmallVector<int64_t, 4> dims;
+  for (auto i = 0; i < tensorRank; i++) {
+    if (seqShape[i] != tensorShape[i]) {
```
AlexandreEichenberger, line 903: You could use the `test ? val1 : val2` to make the code more compact.
The approach is smart, I like it. It basically force possibly 2 different literal values to become "fake" runtime. Personally, I would like a comment explaining what you are doing here.

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -887,6 +881,35 @@ LogicalResult ONNXSeluOp::inferShapes(
 
 LogicalResult ONNXSequenceInsertOp::inferShapes(
```
AlexandreEichenberger, line 882: Did you add verifiers to make sure the elementary types are identical?

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -4784,47 +4818,52 @@ namespace mlir {
 namespace onnxmlir {
 namespace detail {
 struct SeqTypeStorage : public mlir::TypeStorage {
-  using KeyTy = llvm::ArrayRef<mlir::Type>;
+  // std::tuple is used because the storage for the elements may be added later
```
AlexandreEichenberger, line 4821: I would like to see a comment explaining the role of this struct.

src/Dialect/ONNX/ONNXOps.cpp
chentong319: At line 887: direct take the insert's shape.

src/Dialect/ONNX/ONNXOps.cpp
chentong319: Yes. Verifier was added in a previous PR.

AlexandreEichenberger (approved):
LGTM

PR #1055, opened 2021-12-10
Instrumentation into blocks

chentong319:

1. Use walk() function to traverse operations recursively.
2. Exclude onnx.Return from instrumentation

### DIFF
--- a/src/Transform/ONNX/InstrumentONNXPass.cpp
+++ b/src/Transform/ONNX/InstrumentONNXPass.cpp
@@ -105,8 +105,8 @@ public:
           opBuilder.create<mlir::KrnlInstrumentOp>(loc, op, tag);
         }
 
-        // Can not insert after Op (e.g. ONNXReturnOP) with IsTerminator
-	if (InstrumentControlBits.isSet(InstrumentAfterOp) &&
+        // Can not insert after Op (e.g. ONNXReturnOP) with IsTerminator Trait
+        if (InstrumentControlBits.isSet(InstrumentAfterOp) &&
             !op->hasTrait<OpTrait::IsTerminator>()) {
           opBuilder.setInsertionPointAfter(op);
           uint64_t tag =
--- a/src/Transform/ONNX/InstrumentONNXPass.cpp
+++ b/src/Transform/ONNX/InstrumentONNXPass.cpp
@@ -97,10 +97,6 @@ public:
         if (!allOpsAllowed && allowedOps.find(opName) == allowedOps.end())
           return;
 
-        // Do not instrument ONNXReturn, which has terminate Trait
-        if (strcmp(opName, "Return") == 0)
-          return;
-
         Location loc = op->getLoc();
         OpBuilder opBuilder(op);
         if (InstrumentControlBits.isSet(InstrumentBeforeOp)) {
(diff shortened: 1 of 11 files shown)


src/Transform/ONNX/InstrumentONNXPass.cpp
```diff
@@ -88,32 +88,34 @@ class InstrumentONNXPass
...
-        const char *opName = op.getName().getStringRef().data() + 5;
+        const char *opName = op->getName().getStringRef().data() + 5;
         if (!allOpsAllowed && allowedOps.find(opName) == allowedOps.end())
-          continue;
+          return;
 
-        Location loc = op.getLoc();
-        OpBuilder opBuilder(&op);
+        // Do not instrument ONNXReturn, which has terminate Trait
+        if (strcmp(opName, "Return") == 0)
```
etiotto, line 101: Suggest to use strncmp.

src/Transform/ONNX/InstrumentONNXPass.cpp
```diff
@@ -88,32 +88,34 @@ class InstrumentONNXPass
     if (instrumentONNXOps == "" || instrumentONNXOps == "NONE")
       return;
     init(instrumentONNXOps);
-    auto function = getFunction();
-    auto &funcBody = function.getBody();
 
-    // Iterate on the operations
-    for (Operation &op : funcBody.getOps()) {
-      if (isa<mlir::ONNXOpsDialect>(op.getDialect())) {
+    // Iterate on the operations nested in this function
```
AlexandreEichenberger, line 92: Question: does the walk works recursively over structured constructs like onnx.loop and onnx.if? Also, it walks in a tree and at the same time add operation. Out of curiosity, are they encountered in the walk (as added ops are also part of the function), but ignored as only ONNX ops are instrumented, and the added ops are KRNL ops. Can you confirm, thanks.

Nit, Doru likes period at the end of sentences.

src/Transform/ONNX/InstrumentONNXPass.cpp
AlexandreEichenberger: Would it be not better to  use an isa<xxxReturn> instead of comparing a name?

src/Transform/ONNX/InstrumentONNXPass.cpp
chentong319: Walk() traverses the mlir IR structure recursively, according to the document. I tested and found the inserted operations were traversed no matter they were insert before or after the current node. The order seems to be pre-order, not specified in document.

src/Transform/ONNX/InstrumentONNXPass.cpp
chentong319: That's true. I will use isa.

chentong319: Actually, my new version checks the Trait instead of the Op. This method can handle any op that has to be at the end of block.

AlexandreEichenberger: Even better

src/Transform/ONNX/InstrumentONNXPass.cpp
AlexandreEichenberger: Thanks, looks like a very useful and robust approach to traversal

AlexandreEichenberger (approved):
LGTM

PR #1057, opened 2021-12-13
Update README.md

YW-Ethan:
Update prerequisites section, including more info for Installation on UNIX.

To install on UNIX, I need to install ONNX model in Python before running these command:
```
mkdir onnx-mlir/build && cd onnx-mlir/build
cmake -G Ninja ..
cmake --build .
```

Also, there is an issue that I had before. When I use `pip install onnx`, it will install ONNX under `/usr/local/bin`; however, `cmake --build .` command will invoke python from `/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions`, and there is no ONNX package under this path. I solved it by deleting all Python under this path, so it then use Python from `/usr/local/bin`. I don't think this is a common issue, so I only document it here FYI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,7 @@ gcc >= 6.4
 libprotoc >= 3.11.0
 cmake >= 3.15.4
 ninja >= 1.10.2
+onnx >= 1.10.2 
 ```
 GCC can be found [here](https://gcc.gnu.org/install/), or if you have [Homebrew](https://docs.brew.sh/Installation), you can use `brew install gcc`. To check what version of gcc you have installed, run `gcc --version`.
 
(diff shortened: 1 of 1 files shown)


AlexandreEichenberger (commented):
The installation of ONNX in Python does not quite work for us as our project is based on a specific back level. This is why it needs to be installed from git. Maybe we need to transform this as a warning, that we should not install ONNX from other sources but via git. Specifically, ONNX  is loaded to us as a third_party module for which we load the current version on which ONNX-MLIR works. Hope this helps.

Netron is a tool that many of us use to look at onnx input files.

Issue #1016, opened 2021-11-19
make error with jni


chentong319:
On the recent master branch, new errors:
1.  complaint about `/home/chentong/onnx-mlir/build/src/Runtime/jni/jsoniter` is already there. I guess its a make script issue. I can delete that directory and continue.
2. I do not understand the following error
```
Cloning into '/home/chentong/onnx-mlir/build/src/Runtime/jni/jsoniter'...
remote: Enumerating objects: 7670, done.
remote: Total 7670 (delta 0), reused 0 (delta 0), pack-reused 7670
Receiving objects: 100% (7670/7670), 1.25 MiB | 17.49 MiB/s, done.
Resolving deltas: 100% (3747/3747), done.
Note: switching to '42e8df1ae1b553d070a6667f49cf63c6904ead5f'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

/bin/sh: 1: mvn: not found
make[2]: *** [src/Runtime/jni/CMakeFiles/jsoniter.dir/build.make:74: src/Runtime/jni/jsoniter/target/jsoniter-0.9.23.jar] Error 127
make[1]: *** [CMakeFiles/Makefile2:3349: src/Runtime/jni/CMakeFiles/jsoniter.dir/all] Error 2
make: *** [Makefile:146: all] Error 2
```
@gongsu832, I assume that these issues are caused by your PR.

gongsu832:
It's a known problem on MacOS. Workaround before it's fixed: `brew install maven` and run `alias nproc="sysctl -n hw.logicalcpu"` in your shell.

PR #1060, opened 2021-12-13
Adding suggested workaround that some experience on MacOS

AlexandreEichenberger:
Adding info listed in issue #1016 as solution to a make error on MacOS


etiotto:
Question: this is just a temporary workaround correct?  I think that the problem should be fixed so that Mac users will not need these in the long term.

AlexandreEichenberger:
I merely lifted the text from Gong in his issue response. I think it appears that something is broken on mac with respect to java, probably a non-standard use. I am ensure that it would get fixed

etiotto (approved):
LGTM.

PR #1063, opened 2021-12-14
add quick start script to install things and run demo all in one go

renxida:
When I first approached this project, it took me quite a while to figure out how things worked in the `utils` folder.

I think it might be a good idea to have one script in the project root that would help a newcomer get started.

I don't really know much about docker or github workflows, and would love it if one of you could help me set up a test for this script in the cloud. It worked on my Ubuntu 20.04, almost fresh install.

I love this project and it has saved me the pain of having to deal with tensorflow and its dependencies, and I want it to be easier for people to approach it.


AlexandreEichenberger:

AlexandreEichenberger:
Thanks for your contribution. I asked @gongsu832  to give a review.

gongsu832:
The main reason a new comer might have trouble building is because the project has a lot of dependencies. So a simple script like this has very little chance to work in various environments without first checking all the prereqs are met. That's why we did NOT provide such a script and instead provided an `onnx-mlir-dev` docker image where you can easily use for dev work without having to go through the trouble of setting up your host environment.

renxida:
> The main reason a new comer might have trouble building is because the project has a lot of dependencies. So a simple script like this has very little chance to work in various environments without first checking all the prereqs are met. That's why we did NOT provide such a script and instead provided an `onnx-mlir-dev` docker image where you can easily use for dev work without having to go through the trouble of setting up your host environment.

Hi gongsu832!

The dockerfile has indeed been very useful. You have a point that maintaining such a script might be a pain.

I had two things that I had wanted to address by this pull request, and I would love it if you could take a look and suggest other solutions:

1. When I just started, I saw the instructions for how to spin up the docker container, but how I could compile and run an onnx model was not immediately available. Perhaps it would be nice to include a link to `https://github.com/onnx/onnx-mlir/tree/main/docs/mnist_example` in the overall `README.md`?

2. I needed to run onnx-mlir outside a container, and there might other people like me. In my usecase, I'm trying to use `perf` and Intel's `sde64` to profile onnx models, neither of which would work in docker containers. I'm thinking this can best be addressed in a separate gist rather than the overall repo?

Best,
Xida

gongsu832:
I agree our REAME.md is rather intimidating and needs to be simplified and better organized.

I don't know about sde64 but perf certainly runs inside containers. Just google "running perf inside container".

renxida:
> I agree our REAME.md is rather intimidating and needs to be simplified and better organized.
Neat! I'll go ahead and see if i can make it just a little easier


> I don't know about sde64 but perf certainly runs inside containers. Just google "running perf inside container".
I tried a number of things from stackoverflow, but it didn't work due to various permissions / capabilities issues. Eventually i just gave up and wrote the script i submitted here : P

PR #1068, opened 2021-12-16
Add ViewLikeOpInterface to krnl.vector_type_cast

tungld:
`krnl.vector_type_cast` op is a view of the source memref, so this patch adds ViewLikeOpInterface to it. 

Though the current onnx-mlir has no problem since `krnl.vector_type_cast` is only used inside MatMul, this change helps buffer-related passes correctly analyze buffer's lifetime when `krnl.vector_type_cast` is used more often.

### DIFF
--- a/src/Dialect/Krnl/KrnlOps.td
+++ b/src/Dialect/Krnl/KrnlOps.td
@@ -21,6 +21,7 @@ include "mlir/IR/OpBase.td"
 include "mlir/Dialect/Shape/IR/ShapeBase.td"
 include "mlir/Interfaces/LoopLikeInterface.td"
 include "mlir/Interfaces/SideEffectInterfaces.td"
+include "mlir/Interfaces/ViewLikeInterface.td"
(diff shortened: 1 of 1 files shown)


etiotto (approved):
LGTM - thanks!

etiotto:
@tungld  I saw that the `Krnl` dialect also has another op that is cast like `KrnlDummyCastOp`.  At first glance that op should also implement the `ViewLikeOpInterface`, what do you think?

tungld:
> @tungld I saw that the Krnl dialect also has another op that is cast like KrnlDummyCastOp. At first glance that op should also implement the ViewLikeOpInterface, what do you think?

Thanks for pointing it out! I just looked at KrnlDummyCastOp. It is used internally as a supporting op in lowering ONNX Loop and Scan, then it disappears by canonicalization. I think we have no problem with it.

AlexandreEichenberger:
@tungld Thanks. Out of curiosity, what does adding this interface buy us?

PR #1069, opened 2021-12-16
Add check-onnx-backend to Mac CI.

etiotto:
Add check-onnx-backend to Mac CI.


utils/check-onnx-numerical.cmd
```diff
@@ -1,2 +1,2 @@
-cd onnx-mlir\build
+cd onnx-mlir/build
```
sstamenova, line 1: Why is this changing?

CMakeLists.txt
```diff
@@ -48,7 +48,13 @@ if (WIN32)
   # This allows us to pick up the correct version of Python instead of always picking the latest
   # installed on the machine.
   set(Python3_FIND_REGISTRY NEVER)
+  endif()
+
+# Use the Python version specified by pythonLocation.
+if(NOT DEFINED Python_ROOT_DIR AND DEFINED ENV{pythonLocation})
```
sstamenova, line 54: Nit: space after if

CMakeLists.txt
```diff
@@ -48,7 +48,13 @@ if (WIN32)
   # This allows us to pick up the correct version of Python instead of always picking the latest
   # installed on the machine.
   set(Python3_FIND_REGISTRY NEVER)
+  endif()
```
sstamenova, line 51: Nit: spacing

CMakeLists.txt
```diff
@@ -48,7 +48,13 @@ if (WIN32)
...
+
+# Use the Python version specified by pythonLocation.
+if(NOT DEFINED Python_ROOT_DIR AND DEFINED ENV{pythonLocation})
+  set(Python3_ROOT_DIR "$ENV{pythonLocation}")
```
sstamenova, line 55: Why is this needed? find_package below should find it and set the root dir, right?

sstamenova (changes requested):

utils/check-onnx-numerical.cmd
etiotto: fixed

CMakeLists.txt
etiotto: fixed

CMakeLists.txt
etiotto: fixed

CMakeLists.txt
sstamenova: So this change is needed just for the buildbot? Wouldn't it be better if we passed in the suggestion as a cmake variable?

e.g. `-DPython3_ROOT_DIR=...`

etiotto: That should work as well. The current code now allows for either passing `-DPython3_ROOT_DIR=...` via cmake or, if that is not defined, leveraging the `pythonLocation ` environment variable.  This is consistent with what is done by the third_party tools which has similar code: 

```
./onnx/third_party/pybind11/tools/pybind11NewTools.cmake:  if(NOT DEFINED Python_ROOT_DIR AND DEFINED ENV{pythonLocation})
./pybind11/tools/pybind11NewTools.cmake:  if(NOT DEFINED Python_ROOT_DIR AND DEFINED ENV{pythonLocation})
```

sstamenova: Reading environment variables like this is generally frowned upon in cmake. Unless there is a strong reason to do it, it is better to rely on the cmake options.

etiotto: ok done

sstamenova (approved):
Thanks!

AlexandreEichenberger (approved):
LGTM

AlexandreEichenberger:
@etiotto Thanks for adding instructions in Readme. Ideally this one should go before #1084 as this info is moved around.

PR #1070, opened 2021-12-16
Update wording

cjvolzka:
Update some terms to avoid use of "master" (where possible) and "sanity" based on https://www.ibm.com/blogs/think/2020/08/words-matter-driving-thoughtful-change-toward-inclusive-language-in-technology/.

Deleted utils/clone-mlir-master.sh as it appears to be longer used. The existing utils/clone-mlir.sh appears to have replaced it but the original file remained.

### Note: Links related to the onnx repo still use "master" as that repo has not been switched to using "main" as the primary branch name.


AlexandreEichenberger (approved):
LGTM

PR #1071, opened 2021-12-17
Allow unary element-wise ops to work with scalar type

tungld:
This patch allows unary element-wise ops to work with scalar type. This is good for LSTM/GRU/RNN where these element-wise ops are used as activations, because we can avoid allocating/deallocating a MemRef of a single element.


src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -863,6 +863,17 @@ struct ONNXElementwiseUnaryOpLowering : public ConversionPattern {
     auto loc = ONNXLoc<ElementwiseUnaryOp>(op);
     auto X = operands[0];
 
+    // If type is scalar, there need not allocate a buffer. Just call scalar
```
etiotto, line 866: If type is scalar, there need not allocate a buffer
-> If the type is a scalar, there is no need to allocate a buffer

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -863,6 +863,17 @@ struct ONNXElementwiseUnaryOpLowering : public ConversionPattern {
...
+    // computation and return the result.
+    // This is efficient when elementwise ops are used as activations for ops
+    // like LSTM/GRU/RNN.
+    if (!X.getType().isa<TensorType>() && !X.getType().isa<MemRefType>()) {
```
etiotto, line 870: I am wondering whether this could be written as: 
```
bool isScalar = !X.getType().isa<ShapedType>();
if (!isScalar) {
```

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
tungld: Done. Thanks!

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
tungld: ShapedType covers VectorType. That's why I didn't use it. I updated the comment to mention vector type. Thanks!

src/Conversion/ONNXToKrnl/RNN/RNNBase.cpp
```diff
@@ -247,12 +247,6 @@ Value applyActivation(OpBuilder &rewriter, Location loc,
   bool isScalar = !operand.getType().isa<ShapedType>();
```
AlexandreEichenberger, line 247: Out of curiosity, before you were including vectors in scalar. So you are intending to not include them here? I think its ok since we do not vectorize before ONNX to KRNL, but I just wonder why it was an issue in the above code but not here.

AlexandreEichenberger (approved):
LGTM, let's wait for @etiotto to also indicates if its ok by him

AlexandreEichenberger (approved):

src/Conversion/ONNXToKrnl/RNN/RNNBase.cpp
tungld: @AlexandreEichenberger Yes, you are right, they are redundant. I just removed them. Thanks!

PR #1075, opened 2021-12-30
Add ScatterElementsOp and RoiAlignOp inferShapes.

PengQi220:
Add ScatterElementsOp and RoiAlignOp inferShapes.

## ScatterElements

image

> after inferShapes

```MLIR
%29 = "onnx.ScatterElements"(%28, %26, %27) {axis = 1 : si64, onnx_node_name = "ScatterElements_28"} : (tensor<64x2560000xf32>, tensor<64x55000xi64>, tensor<64x55000xf32>) -> tensor<64x2560000xf32>
```

### RoiAlign

image

> after inferShapes

```MLIR
%439 = "onnx.RoiAlign"(%133, %438, %436) {onnx_node_name = "RoiAlign_421", output_height = 7 : si64, output_width = 7 : si64, sampling_ratio = 2 : si64, spatial_scale = 1.000000e+00 : f32} : (tensor<1x256    x44x120xf32>, tensor<?x4xf32>, tensor<?xi64>) -> tensor<?x256x7x7xf32>
```


src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -4203,7 +4203,22 @@ LogicalResult ONNXReduceSumSquareOp::inferShapes(
 
 LogicalResult ONNXRoiAlignOp::inferShapes(
     std::function<void(mlir::Region &)> doShapeInference) {
-  return emitError(NOT_IMPLEMENTED_MESSAGE);
+  int64_t height = output_height();
+  int64_t width  = output_width();
+
+  auto xTy = X().getType().cast<RankedTensorType>();
```
chentong319, line 4209: The input tensor may not be ShapedType in this pass of shape inference. You should check the condition before cast. If a tensor is not a ShapedType, simply `return success()`.

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -4203,7 +4203,22 @@ LogicalResult ONNXReduceSumSquareOp::inferShapes(
...
+  auto xShape = xTy.getShape();
+  auto biTy = batch_indices().getType().cast<RankedTensorType>();
+  auto biShape = biTy.getShape();
+
+  SmallVector<int64_t, 4> outputDims = { biShape[0], xShape[1], height, width};
```
chentong319, line 4214: Be aware that your code can only handle tensor dimension with constant size. When a dimension size is only known dynamically (which is represented by -1), the relationship between the input and output (e.g. biShape[0] is the first dimension) is not represented in the code. We have other utilities to represent such relationship. Since this is your first patch, you can add the code for dynamic size later.

chentong319 (commented):
I add two comments with your code.
For DCO failure, remember use `git commit --signoff`

chentong319 (commented):
The new changes fixed the issues in my previous comments.
I saw that you use a forced merge, which should be avoided. Any reason that you have to use forced merge?
Sorry for another requirement: we usually need to add a test case for the added shape inference code. I guess you already has one.  You can add it to test/mlir/onnx/onnx_shape_inference.mlir using the tool util/mlir2FileCheck.py.  I just realized there is no document on how to do that. If you need help, either contact me or wait for the document.

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -4323,7 +4329,8 @@ LogicalResult ONNXScatterOp::inferShapes(
 
 LogicalResult ONNXScatterElementsOp::inferShapes(
     std::function<void(mlir::Region &)> doShapeInference) {
```
AlexandreEichenberger, line 4331: Above code is correct, it might be wise to only set the output result type when `data` is ranked. Reason being that changes in the ops trigger a new wave of shape inference in all the consumer ops, I believe.

src/Dialect/ONNX/ShapeInference/RoiAlign.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+LogicalResult ONNXRoiAlignOpShapeHelper::computeShape(
+    ONNXRoiAlignOpAdaptor operandAdaptor) {
+  // Shape inference indicated by passing a null rewriter pointer.
+  // get input info.
+  Value X = operandAdaptor.X();
+  Value batch_indices = operandAdaptor.batch_indices();
+
+  // Test ranks.
+  if (X.getType().cast<ShapedType>().getShape().size() != 4)
```
AlexandreEichenberger, line 35: The preferred way is to have a verifier that isolate all of the testing, and a shape inference that does  the shape inferencing. See HowToAddAnOperation.md for help on how to do it

AlexandreEichenberger (commented):
Thanks for the contribution. I suggest that you also add a literal test for each of the ops. Please let us know if you need help on this.

PengQi220:
> The new changes fixed the issues in my previous comments. I saw that you use a forced merge, which should be avoided. Any reason that you have to use forced merge? Sorry for another requirement: we usually need to add a test case for the added shape inference code. I guess you already has one. You can add it to test/mlir/onnx/onnx_shape_inference.mlir using the tool util/mlir2FileCheck.py. I just realized there is no document on how to do that. If you need help, either contact me or wait for the document.

1. For DCO failure, and its `Details` is written like this

    ```
    git commit --amend --no-edit --signoff
    git push --force-with-lease origin main
    ```

    I will avoid using `forced merge` later.

2. I will try to add a test case.

src/Dialect/ONNX/ONNXOps.cpp
PengQi220: Thanks for pointing out the problem.

src/Dialect/ONNX/ShapeInference/ONNXShapeHelper.hpp
```diff
@@ -525,3 +525,15 @@ struct ONNXCategoryMapperOpShapeHelper
...
+struct ONNXRoiAlignOpShapeHelper : public ONNXOpShapeHelper<ONNXRoiAlignOp> {
+  ONNXRoiAlignOpShapeHelper(ONNXRoiAlignOp *newOp);
+  ONNXRoiAlignOpShapeHelper(ONNXRoiAlignOp *newOp, OpBuilder *rewriter,
+      ArrayValueIndexCapture::GetDenseVal fGetDenseVal,
+      ArrayValueIndexCapture::LoadVal fLoadVal);
+  LogicalResult computeShape(ONNXRoiAlignOpAdaptor operandAdaptor);
+  // Additional data for RoiAlignOp.
+  SmallVector<IndexExpr, 4> xDims;            // Dim of X.
+  SmallVector<IndexExpr, 1> batchIndicesDims; // Dim of batch_indices.
+};
```
etiotto, line 539: nit: add a newline

src/Dialect/ONNX/ShapeInference/RoiAlign.cpp
```diff
@@ -0,0 +1,56 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//===-------------- RoiAlign.cpp - Shape Inference for RoiAlign Op
```
etiotto, line 5: Reduce the width to 80 characters.

src/Dialect/ONNX/ShapeInference/RoiAlign.cpp
```diff
@@ -0,0 +1,56 @@
...
+//--------------===//
+//
+// This file implements shape inference for the ONNX Shape Operator.
```
etiotto, line 8: Update the comment to reflect this is the implementation of the Region of Interest (RoI) align operation.

src/Dialect/ONNX/ShapeInference/RoiAlign.cpp
PengQi220: A simple test case is added.

src/Dialect/ONNX/ShapeInference/RoiAlign.cpp
PengQi220: Thanks，fixed now.

src/Dialect/ONNX/ONNXOps.cpp
AlexandreEichenberger: thanks

src/Dialect/ONNX/ShapeInference/RoiAlign.cpp
AlexandreEichenberger: Thanks for adding the example. Are you willing to move the "testing for error" into a separate verifier? That is the preferred solution. Easiest to do it while adding the code, though it can be migrated later if you don't have the time now.

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -4323,7 +4329,11 @@ LogicalResult ONNXScatterOp::inferShapes(
...
-  return emitError(NOT_IMPLEMENTED_MESSAGE);
+  if (!data().getType().isa<mlir::RankedTensorType>())
+    return success();
+
+  getResult().setType(data().getType());
+  return success();
```
AlexandreEichenberger, line 4336: Scatter Elements operation also has some requirements that are not tested at this time:
```
nputs

data (differentiable) : T
Tensor of rank r >= 1.
indices (non-differentiable) : Tind
Tensor of int32/int64 indices, of r >= 1 (same rank as input). All index values are expected to be within bounds [-s, s-1] along axis of size s. It is an error if any of the index values are out of bounds.
updates (differentiable) : T
Tensor of rank r >=1 (same rank and shape as indices)
```

Easiest is to add these constraints while doing the shape inference, and add them in a separate verifier

AlexandreEichenberger (commented):
Almost there, thanks for the contributions. Ping me if you want help for verifiers.

src/Dialect/ONNX/ONNXOps.td.inc
```diff
@@ -4803,6 +4803,7 @@ def ONNXRoiAlignOp:ONNX_Op<"RoiAlign",
       return {20};
     }
   }];
```
AlexandreEichenberger, line 4805: We typically don't modify these machine generated files by hand. 

The file that drive the generation of the .inc file is this one: utils/gen_onnx_mlir.py . Search for `OpsWithVerifier` in that file and add the two ops there. Explanation here: https://github.com/onnx/onnx-mlir/blob/main/docs/HowToAddAnOperation.md#generate-the-proper-onnxtdinc

The inc files can be regenerated using the command is `make OMONNXOpsIncTranslation `. If you work in a Docker env, it will work, otherwise look for installation step needed for ONNX here. https://github.com/onnx/onnx-mlir/blob/main/docs/ImportONNXDefs.md#how-to-use-the-script

In your case, since you manually edited the inc file, you can probably skip that step

AlexandreEichenberger (commented):
Thanks a bunch for adding the verifier code. Just one small step (a 2 liner) to insert the code that will generate the `.td.inc` files so that your manual change will not be lost next time we regenerate the code.

src/Dialect/ONNX/ONNXOps.td.inc
PengQi220: Thanks for the wise suggestion, revised.

utils/gen_onnx_mlir.py
```diff
@@ -297,6 +297,8 @@
     'SequenceInsert',
     'SpaceToDepth',
     'TopK',
+    "RoiAlign",
```
AlexandreEichenberger, line 300: We typically add them in alphabetical order... can be fixed later if you want to drop it now

AlexandreEichenberger (approved):
LGTM, thanks for the contribution, much appreciated.

AlexandreEichenberger:

AlexandreEichenberger:
@PengQi220 I downloaded your changes as a patch to investigate what is going on. It turns out that we check that the equivalence between the automatically generated files in git with these that we would generate if we ran the tool to gen them. The two are not in sync. Short answer is that you need to run

``` shell
make OMONNXOpsIncTranslation
```

and commit the modified files. You should not have trouble running that command in the docker image.

PengQi220:
> ```shell
> check-docs
> ```

The master branch has restricted the onnx version, and my new commit has been adapted to the 1.9.0 restricted version.

PengQi220:
@AlexandreEichenberger @chentong319 @etiotto Hi, do I need to do anything more to make this PR acceptable

AlexandreEichenberger:
>  Hi, do I need to do anything more to make this PR acceptable

you did great. Once it passes the test, we will merge it in

AlexandreEichenberger:
Congrats, and now that you have gotten the hand of it, hopefully the next one will look like a piece of cake. Thanks for the contribution, much appreciated.

PR #1082, opened 2022-01-06
Mnist example minor revision

AlexandreEichenberger:
Minor revision to make the mninst example easier to follow. Had to add "./" path qualifier to locate the .so located in the example directory.


tungld (approved):
LGTM!

A minor comment: should we use the same input data for both C and Python example, so that we will see the same output? The output of the C example is now quite ideal :)

AlexandreEichenberger:
@tungld introduced the same image in the python interface example as the one used in the C++ interface, as requested.

PR #1083, opened 2022-01-06
Make the doc example obey ONNX_MLIR_BUILD_TESTS

sstamenova:
Currently, ONNX_MLIR_BUILD_TESTS controls EXCLUDE_FROM_ALL, however add_custom_command will run always unless associated with a target. The second issue is that add_executable will always build as well, so we follow the llvm pattern and explicitly set EXCLUDE_FROM_ALL in the add_onnx_mlir_executable function if it is set for the directory.


gongsu832:

hpmorgan:
> add_custom_command will run always unless associated with a target. 

This shouldn't be the case. If that is happening, the dependency chain is not set up correctly (output(s) chained to target that is included in ALL).

sstamenova:
> > add_custom_command will run always unless associated with a target.
> 
> This shouldn't be the case. If that is happening, the dependency chain is not set up correctly (output(s) chained to target that is included in ALL).

Interesting, so perhaps because the `add_executable` target depends on the imported shared lib which in turns depends on a lib that requires the output, that's causing the `add_custom_command` to always run. Let me try if just fixing the `add_executable` target is sufficient.

AlexandreEichenberger:

sstamenova:
@gongsu832 @etiotto : Can you have a look? I'd like to get this checked in.

gongsu832 (approved):
LGTM

PR #1085, opened 2022-01-07
Make jsoniter work with ninja

sstamenova:
As the title says, the jsoniter external project target doesn't work with ninja. This is an issue with the way external project targets and jar targets are created. We can work around it, by adding a custom intermediate target that copies the jar file after it is downloaded and using the byproducts. This should maintain compatibility with make and other build systems while also enabling ninja support.

This also removes a few tabs from the cmake files.


gongsu832:

gongsu832 (approved):
LGTM

PR #1087, opened 2022-01-08
Support filtering out lit tests based on targets

sstamenova:
Currently we ignore what targets llvm was built for in the lit tests, but recent changes to onnx-mlir explicitly initialize the available targets.
This makes the corresponding change to the lit configuration, so that we can filter out the lit tests based on the available targets.


AlexandreEichenberger:

etiotto (approved):
LGTM

PR #1090, opened 2022-01-11
onnx-mlir.py warning about binary output

gongsu832:
onnx-mlir.py warning about binary output (.so and .jar) not directly usable if host is not Linux


AlexandreEichenberger (approved):
LGTM, thanks

PR #1094, opened 2022-01-11
Switch URLs to use main instead of master

cjvolzka:
Now that [onnx switched from master to main](https://github.com/onnx/onnx/issues/3901), update URLs in onnx-mlir to also reflect the change.

Also remove some extraneous whitespace that my editor caught in the updated files.


AlexandreEichenberger (approved):
LGTM, thanks for the update

AlexandreEichenberger:

PR #1081, opened 2022-01-06
Update LLVM to 700997aef8c1

etiotto:
Updating LLVM to SHA `700997aef8c1` causes a number of build issues due to the following MLIR interface changes:
- https://reviews.llvm.org/D115728
- https://reviews.llvm.org/D113229
- https://reviews.llvm.org/D113881
- https://reviews.llvm.org/D116648
- https://reviews.llvm.org/D113956
- https://reviews.llvm.org/D114698

This patch modifies `onnx-mlir` to use LLVM  `700997aef8c1`.

Signed-off-by: Ettore Tiotto <etiotto@ca.ibm.com>

PR #1095, opened 2022-01-11
Update LLVM level

etiotto:
Continue https://github.com/onnx/onnx-mlir/pull/1081 which was closed unintentionally. 

Updating LLVM to SHA 015192c63415ae828c3a9bc8b56d0667abcca8ae  causes a number of build issues

This patch attempts to modify onnx-mlir to use LLVM 015192c63415ae828c3a9bc8b56d0667abcca8ae.


etiotto:
The patch fixes the build issues. However a number of LIT test fails:
```
Failed Tests (16):
  Open Neural Network Frontend :: conversion/krnl_to_affine/matmul.mlir
  Open Neural Network Frontend :: krnl/krnl_category_mapper.mlir
  Open Neural Network Frontend :: krnl/reshape.mlir
  Open Neural Network Frontend :: onnx/onnx_bundle_memory_pool.mlir
  Open Neural Network Frontend :: onnx/onnx_location.mlir
  Open Neural Network Frontend :: onnx/onnx_lowering.mlir
  Open Neural Network Frontend :: onnx/onnx_lowering_category_mapper.mlir
  Open Neural Network Frontend :: onnx/onnx_lowering_depth_to_space_op.mlir
  Open Neural Network Frontend :: onnx/onnx_lowering_gru_op.mlir
  Open Neural Network Frontend :: onnx/onnx_lowering_lstm_op.mlir
  Open Neural Network Frontend :: onnx/onnx_lowering_nonmaxsuppression_op.mlir
  Open Neural Network Frontend :: onnx/onnx_lowering_range_op.mlir
  Open Neural Network Frontend :: onnx/onnx_lowering_rnn_op.mlir
  Open Neural Network Frontend :: onnx/onnx_lowering_space_to_depth.mlir
  Open Neural Network Frontend :: onnx/onnx_lowering_with_canonicalize.mlir
  Open Neural Network Frontend :: onnx/onnx_math_functions_lowering.mlir


Testing Time: 2.06s
  Unsupported:  3
  Passed     : 41
  Failed     : 16
```
 
 Looking at one of the failure:
 ```
 % /Users/etiotto@ca.ibm.com/Source/tmp/onnx/onnx-mlir/build/Debug/bin/onnx-mlir-opt --shape-inference --convert-onnx-to-krnl --convert-krnl-to-affine --convert-krnl-to-llvm /Users/etiotto@ca.ibm.com/Source/tmp/onnx/onnx-mlir/test/mlir/krnl/reshape.mlir -split-input-file
module  {
}

within split at /Users/etiotto@ca.ibm.com/Source/tmp/onnx/onnx-mlir/test/mlir/krnl/reshape.mlir:3 offset :4:8: error: failed to materialize conversion for result #0 of operation 'onnx.Reshape' that remained live after conversion
  %0 = "onnx.Reshape"(%arg0, %arg1) : (tensor<?x10xf32>, tensor<4xi64>) -> tensor<*xf32>
       ^
within split at /Users/etiotto@ca.ibm.com/Source/tmp/onnx/onnx-mlir/test/mlir/krnl/reshape.mlir:3 offset :4:8: note: see current operation: %67 = "onnx.Reshape"(<<UNKNOWN SSA VALUE>>, <<UNKNOWN SSA VALUE>>) : (tensor<?x10xf32>, tensor<4xi64>) -> tensor<?x?x?x?xf32>
within split at /Users/etiotto@ca.ibm.com/Source/tmp/onnx/onnx-mlir/test/mlir/krnl/reshape.mlir:3 offset :5:3: note: see existing live user here: return %42 : tensor<?x?x?x?xf32>
  ^
  ```

etiotto:
The last LLVM commit that works is 01b55f163a40 [NFC] Tidy up DialectConversion.cpp, the next commit 015192c63415 [mlir:DialectConversion] Restructure how argument/target materializations get invoked exposes the problem.

etiotto:
https://reviews.llvm.org/D111620 changes the type conversion infrastructure in MLIR quite significantly and exposes several issues in our lit tests. Here is an example, consider the following test: 
```
func @test_reshape(%arg0 : tensor<?x10xf32>, %arg1 : tensor<4xi64>) -> tensor<*xf32> {
  %0 = "onnx.Reshape"(%arg0, %arg1) : (tensor<?x10xf32>, tensor<4xi64>) -> tensor<*xf32>
```
During onnx to krnl dialect conversion `onnx.Reshape` is eventually replaced by a `memref.reinterpret_cast`:
```
Legalizing operation : 'onnx.Reshape'(0xf0b8d2b1a60) {
  %0 = "onnx.Reshape"(<<UNKNOWN SSA VALUE>>, <<UNKNOWN SSA VALUE>>) : (tensor<?x10xf32>, tensor<4xi64>) -> tensor<?x?x?x?xf32>

  * Fold {
  } -> FAILURE : unable to fold

  * Pattern : 'onnx.Reshape -> ()' {
    ** Insert  : 'arith.constant'(0xf0b8d290280)
   . . .
    ** Insert  : 'memref.reinterpret_cast'(0xf0b8d2137e0)
    ** Replace : 'onnx.Reshape'(0xf0b8d2b1a60)
```
Each operation inserted is then legalized, when it comes to `memref.reinterpret_cast` we have:
```
    Legalizing operation : 'memref.reinterpret_cast'(0xf0b8d2137e0) {
      %66 = "memref.reinterpret_cast"(%arg0, %50, %54, %58, %62, %65, %64, %62) {operand_segment_sizes = dense<[1, 0, 4, 3]> : vector<4xi32>, static_offsets = [0], static_sizes = [-1, -1, -1, -1], static_strides = [-9223372036854775808, -9223372036854775808, -9223372036854775808, 1]} : (memref<?x10xf32>, index, index, index, index, index, index, index) -> memref<?x?x?x?xf32>

    } -> SUCCESS : operation marked legal by the target
```
Note here that the result type of `memref.reinterpret_cast` is a memref. That result will feed the `std.return` operation. 
However that operation is not legalized because is already considered legal, no matter what is the type of its input (note that the std.return input type is a tensor):
```
//===-------------------------------------------===//
Legalizing operation : 'std.return'(0xf0b8d2da5e0) {
  "std.return"(%67) : (tensor<?x?x?x?xf32>) -> ()

} -> SUCCESS : operation marked legal by the target
//===-------------------------------------------===//
```
Because the result produced by `memref.reinterpret_cast` is a memref and the result consumed by `std.return` is a tensor we have a type mismatch, which is now detected with the following error message:
```
within split at /home/etiotto/Source/onnx/onnx-mlir/test/mlir/krnl/reshape.mlir:3 offset :4:8: error: failed to materialize conversion for result #0 of operation 'onnx.Reshape' that remained live after conversion
  %0 = "onnx.Reshape"(%arg0, %arg1) : (tensor<?x10xf32>, tensor<4xi64>) -> tensor<*xf32>
       ^
within split at /home/etiotto/Source/onnx/onnx-mlir/test/mlir/krnl/reshape.mlir:3 offset :4:8: note: see current operation: %67 = "onnx.Reshape"(<<UNKNOWN SSA VALUE>>, <<UNKNOWN SSA VALUE>>) : (tensor<?x10xf32>, tensor<4xi64>) -> tensor<?x?x?x?xf32>
within split at /home/etiotto/Source/onnx/onnx-mlir/test/mlir/krnl/reshape.mlir:3 offset :5:3: note: see existing live user here: return %42 : tensor<?x?x?x?xf32>
```
This terse error msg is MLIR way to say that, after lowering `onnx.Reshape`, its result feeds an operation (a live user) that requires an input with a different type. 

The way to fix this problem is to consider `std.return` legal only if its input is not a tensor (fix inspired by changes to mlir/docs/Tutorials/Toy/Ch-5.md in https://reviews.llvm.org/D111620).

After adding this code (in FrontendToKrnlLoweringPass::runOnOperation()):
```
  target.addDynamicallyLegalOp<mlir::ReturnOp>([&](Operation *op) {
+    return llvm::none_of(op->getOperandTypes(),
+                         [](Type type) { return type.isa<TensorType>(); });
+  });
```

The test produces: 
```
    //===-------------------------------------------===//
    Legalizing operation : 'memref.reinterpret_cast'(0xfff1edb37e0) {
      %66 = "memref.reinterpret_cast"(%arg0, %50, %54, %58, %62, %65, %64, %62) {operand_segment_sizes = dense<[1, 0, 4, 3]> : vector<4xi32>, static_offsets = [0], static_sizes = [-1, -1, -1, -1], static_strides =
 [-9223372036854775808, -9223372036854775808, -9223372036854775808, 1]} : (memref<?x10xf32>, index, index, index, index, index, index, index) -> memref<?x?x?x?xf32>

    } -> SUCCESS : operation marked legal by the target
    //===-------------------------------------------===//
  } -> SUCCESS : pattern applied successfully
} -> SUCCESS
//===-------------------------------------------===//

//===-------------------------------------------===//
Legalizing operation : 'std.return'(0xfff1ee7d200) {
  "std.return"(%67) : (tensor<?x?x?x?xf32>) -> ()

  * Fold {
  } -> FAILURE : unable to fold

  * Pattern : 'std.return -> ()' {

    //===-------------------------------------------===//
    Legalizing operation : 'std.return'(0xfff1ee7d200) {
      "std.return"(%66) : (memref<?x?x?x?xf32>) -> ()

    } -> SUCCESS : operation marked legal by the target
    //===-------------------------------------------===//
  } -> SUCCESS : pattern applied successfully
} -> SUCCESS
```
We now can 'see' that `std.return` is now consider illegal (because its input has tensor type). Legalization now has a chance to change its input and use the memref result produced by `memref.reinterpret_cast`. With this change we go from 16 lit test failing to 3 (after adjusting some lit tests to account for the injections of unrealized_conversion_cast by the new MLIR type conversion framework)
```
Failed Tests (3):
  Open Neural Network Frontend :: onnx/onnx_lowering.mlir
  Open Neural Network Frontend :: onnx/onnx_lowering_depth_to_space_op.mlir
  Open Neural Network Frontend :: onnx/onnx_lowering_space_to_depth.mlir
```

etiotto:
These 2 failures:
 ```
   Open Neural Network Frontend :: onnx/onnx_lowering_depth_to_space_op.mlir
   Open Neural Network Frontend :: onnx/onnx_lowering_space_to_depth.mlir
  ```
are caused by the generation of a `memref.dimOp` where the input value has tensor type. This is incorrect because that operation expects the input to have a MemRefType or an UnrankedMemRefType (see https://mlir.llvm.org/docs/Dialects/MemRef/#memrefdim-mlirmemrefdimop).  

The problem stems from the code in `[ONNXSpaceToDepthOpLowering|ONNXDepthToSpaceOpLowering]::matchAndRewrite` which incorrectly obtains the operator input directly from the operator itself rather than from the appropriate operator adaptor.

etiotto:
The last lit failure (onnx/onnx_lowering.mlir) can be reduced to the following code:
```
func private @t2(%arg0: tensor<i64>, %arg1: tensor<i1>, %arg2: tensor<1xi64>) -> tensor<1xi64> {
    ^bb0(%body_arg0: tensor<i64>, %body_arg1: tensor<i1>, %body_arg2: tensor<1xi64>):
    %0 = "onnx.Identity"(%body_arg1) : (tensor<i1>) -> tensor<i1>
    %1 = "onnx.Identity"(%body_arg2) : (tensor<1xi64>) -> tensor<1xi64>
    onnx.Return %0, %1 : tensor<i1>, tensor<1xi64>
  }) : (tensor<i64>, tensor<i1>, tensor<1xi64>) -> tensor<1xi64>
  return %0 : tensor<1xi64>
}
```
The `-debug` output shows (onnx-mlir-opt --shape-inference --convert-onnx-to-krnl t2.mlir  -debug  > new 2>&1):
```
. . . 
//===-------------------------------------------===//
Legalizing operation : 'std.return'(0x54ab7c46130) {
  "std.return"(%7) : (tensor<1xi64>) -> ()

  * Fold {
  } -> FAILURE : unable to fold

  * Pattern : 'std.return -> ()' {
Trying to match "{anonymous}::ReturnOpTypeConversion"
"{anonymous}::ReturnOpTypeConversion" result 1

    //===-------------------------------------------===//
    Legalizing operation : 'std.return'(0x54ab7c46130) {
      "std.return"(%0) : (memref<1xi64>) -> ()

    } -> SUCCESS : operation marked legal by the target
    //===-------------------------------------------===//
  } -> SUCCESS : pattern applied successfully
// *** IR Dump After Pattern Application ***
func private @t2(%arg0: memref<i64>, %arg1: memref<i1>, %arg2: memref<1xi64>) -> memref<1xi64> {
  %0 = memref.alloc() {alignment = 16 : i64} : memref<1xi64>
  %1 = krnl.define_loops 1
  krnl.iterate(%1) with (%1 -> %arg3 = 0 to 1) {
    %8 = krnl.load %arg2[%arg3] : memref<1xi64>
    krnl.store %8, %0[%arg3] : memref<1xi64>
  }
  %2 = memref.alloc() : memref<i1>
  %3 = krnl.load %arg1[] : memref<i1>
  krnl.store %3, %2[] : memref<i1>
  %4 = krnl.define_loops 1
  %5 = krnl.load %arg0[] : memref<i64>
  %6 = arith.index_cast %5 : i64 to index
  krnl.iterate(%4) with (%4 -> %arg3 = 0 to %6) {
    %8 = krnl.load %2[] : memref<i1>
    scf.if %8 {
      %9 = arith.index_cast %arg3 : index to i64
      %10 = memref.alloc() : memref<i64>
      krnl.store %9, %10[] : memref<i64>
      %11 = "onnx.Identity"(%arg1) : (memref<i1>) -> tensor<i1>
      %12 = "onnx.Identity"(%0) : (memref<1xi64>) -> tensor<1xi64>
      onnx.Return %11, %12 : tensor<i1>, tensor<1xi64>
      %13 = krnl.dummy_cast %11 : (tensor<i1>) -> memref<i1>
      %14 = krnl.dummy_cast %12 : (tensor<1xi64>) -> memref<1xi64>
      %15 = krnl.load %13[] : memref<i1>
      krnl.store %15, %2[] : memref<i1>
      %16 = krnl.define_loops 1
      krnl.iterate(%16) with (%16 -> %arg4 = 0 to 1) {
        %17 = krnl.load %14[%arg4] : memref<1xi64>
        krnl.store %17, %0[%arg4] : memref<1xi64>
      }
    }
  }
  %7 = "onnx.Loop"(<<UNKNOWN SSA VALUE>>, <<UNKNOWN SSA VALUE>>, <<UNKNOWN SSA VALUE>>) ( {
  }) : (tensor<i64>, tensor<i1>, tensor<1xi64>) -> tensor<1xi64>
  return %0 : memref<1xi64>
}

} -> SUCCESS
//===-------------------------------------------===//
t2.mlir:6:10: error: failed to materialize conversion for result #0 of operation 'onnx.Identity' that remained live after conversion
    %0 = "onnx.Identity"(%body_arg1) : (tensor<i1>) -> tensor<i1>
         ^
t2.mlir:6:10: note: see current operation: %11 = "onnx.Identity"(%arg1) : (memref<i1>) -> tensor<i1>
t2.mlir:4:8: note: see existing live user here: %13 = krnl.dummy_cast %11 : (tensor<i1>) -> memref<i1>
```
The error msg indicated that there is a `onnx.Identity` operation producing %11 of type tensor<i1> which is live because it is used by a `krnl.dummy_cast` which expects also a `tensor<i1>`.  I am not sure what `conversion` need to be generated in this situation because there isn't a type mismatch between the produce and consumer operations. 

For comparison, the output when using the current version of onnx-mlir-opt is:
```
//===-------------------------------------------===//
Legalizing operation : 'onnx.Identity'(0x24a61cdd5b0) {
  %12 = "onnx.Identity"(%0) : (memref<1xi64>) -> tensor<1xi64>

  * Fold {
  } -> FAILURE : unable to fold

  * Pattern : 'onnx.Identity -> ()' {
Trying to match "ONNXIdentityOpLowering"
    ** Replace : 'onnx.Identity'(0x24a61cdd5b0)
"ONNXIdentityOpLowering" result 1
  } -> SUCCESS : pattern applied successfully
// *** IR Dump After Pattern Application ***
func private @t2(%arg0: memref<i64>, %arg1: memref<i1>, %arg2: memref<1xi64>) -> memref<1xi64> {
  %0 = memref.alloc() {alignment = 16 : i64} : memref<1xi64>
  %1 = krnl.define_loops 1
  krnl.iterate(%1) with (%1 -> %arg3 = 0 to 1) {
    %8 = krnl.load %arg2[%arg3] : memref<1xi64>
    krnl.store %8, %0[%arg3] : memref<1xi64>
  }
  %2 = memref.alloc() : memref<i1>
  %3 = krnl.load %arg1[] : memref<i1>
  krnl.store %3, %2[] : memref<i1>
  %4 = krnl.define_loops 1
  %5 = krnl.load %arg0[] : memref<i64>
  %6 = arith.index_cast %5 : i64 to index
  krnl.iterate(%4) with (%4 -> %arg3 = 0 to %6) {
    %8 = krnl.load %2[] : memref<i1>
    scf.if %8 {
      %9 = arith.index_cast %arg3 : index to i64
      %10 = memref.alloc() : memref<i64>
      krnl.store %9, %10[] : memref<i64>
      %11 = "onnx.Identity"(%arg1) : (memref<i1>) -> tensor<i1>
      %12 = "onnx.Identity"(%0) : (memref<1xi64>) -> tensor<1xi64>
      onnx.Return %11, %12 : tensor<i1>, tensor<1xi64>
      %13 = krnl.dummy_cast %11 : (tensor<i1>) -> memref<i1>
      %14 = krnl.dummy_cast %12 : (tensor<1xi64>) -> memref<1xi64>
      %15 = krnl.load %13[] : memref<i1>
      krnl.store %15, %2[] : memref<i1>
      %16 = krnl.define_loops 1
      krnl.iterate(%16) with (%16 -> %arg4 = 0 to 1) {
        %17 = krnl.load %14[%arg4] : memref<1xi64>
        krnl.store %17, %0[%arg4] : memref<1xi64>
      }
    }
  }
  %7 = "onnx.Loop"(<<UNKNOWN SSA VALUE>>, <<UNKNOWN SSA VALUE>>, <<UNKNOWN SSA VALUE>>) ( {
  }) : (tensor<i64>, tensor<i1>, tensor<1xi64>) -> tensor<1xi64>
  return %7 : tensor<1xi64>
}

} -> SUCCESS 
```
Note that even with the current compiler we have exactly the same situation:
```
      %11 = "onnx.Identity"(%arg1) : (memref<i1>) -> tensor<i1>
      . . .
      %13 = krnl.dummy_cast %11 : (tensor<i1>) -> memref<i1>
```
However instead of receiving the error msg from MLIR things continue smoothly and  the final IR looks like:
```
module  {
  func private @t2(%arg0: memref<i64>, %arg1: memref<i1>, %arg2: memref<1xi64>) -> memref<1xi64> {
    %0 = memref.alloc() {alignment = 16 : i64} : memref<1xi64>
    %1 = krnl.define_loops 1
    krnl.iterate(%1) with (%1 -> %arg3 = 0 to 1) {
      %7 = krnl.load %arg2[%arg3] : memref<1xi64>
      krnl.store %7, %0[%arg3] : memref<1xi64>
    }
    %2 = memref.alloc() : memref<i1>
    %3 = krnl.load %arg1[] : memref<i1>
    krnl.store %3, %2[] : memref<i1>
    %4 = krnl.define_loops 1
    %5 = krnl.load %arg0[] : memref<i64>
    %6 = arith.index_cast %5 : i64 to index
    krnl.iterate(%4) with (%4 -> %arg3 = 0 to %6) {
      %7 = krnl.load %2[] : memref<i1>
      scf.if %7 {
        %8 = arith.index_cast %arg3 : index to i64
        %9 = memref.alloc() : memref<i64>
        krnl.store %8, %9[] : memref<i64>
        %10 = krnl.dummy_cast %arg1 : (memref<i1>) -> memref<i1>
        %11 = krnl.dummy_cast %0 : (memref<1xi64>) -> memref<1xi64>
        %12 = krnl.load %10[] : memref<i1>
        krnl.store %12, %2[] : memref<i1>
        %13 = krnl.define_loops 1
        krnl.iterate(%13) with (%13 -> %arg4 = 0 to 1) {
          %14 = krnl.load %11[%arg4] : memref<1xi64>
          krnl.store %14, %0[%arg4] : memref<1xi64>
        }
      }
    }
    return %0 : memref<1xi64>
  }
}
```
I am not sure precisely how the `onnx.Identity` operations and the `onnx.Return` are elided... it seems to me that:
  - the input of the `onnx.Identity` is copy propagated to the `krnl.dummy_cast`
  - the remaining user `onnx.Return` is not live and so can be removed
  - removing `onnx.Return` renders `onnx.Identity` operation also dead so they are removed
 - the remaining code no longer contain any tensor and is therefore valid in the end

This is probably something the older version of MLIR we currently use does under the cover. It seems to me that now MLIR expects the code to be completely bufferized (all tensors changed into memrefs) after all the operations have been converted.

etiotto:
Found out what exact changes cause the last regression described at https://github.com/onnx/onnx-mlir/pull/1095#issuecomment-1010402014:
```
void ArgConverter::applyRewrites() {
   // Apply all of the rewrites replacements requested during conversion.
   for (auto &repl : replacements)
     for (OpResult result : repl.first->getResults())
     if (Value newValue = mapping.lookupOrNull(result, )) { <<<<<< changed to: if (Value newValue = mapping.lookupOrNull(result, result.getType()))
         result.replaceAllUsesWith(newValue);
```
```
LogicalResult OperationConverter::legalizeChangedResultType(
!  if (!replConverter) {
!    llvm::errs() << "BINGO, op: " << *op << "\n";
!    //    return success();
     return emitConversionError();
+  }

**Comment by etiotto** [comment]
@sstamenova the Windows CI failed with "Artifact cannot be uploaded because max quantity has been exceeded or the payment instrument is invalid. https://aka.ms/artbilling for details." can you please take a look ?

File: src/Dialect/ONNX/MLIRDialectBuilder.cpp [hunk 1]
```diff
@@ -392,6 +392,12 @@ memref::CastOp MemRefBuilder::cast(Value input, MemRefType outputType) const {
 }
 
 Value MemRefBuilder::dim(Value val, int64_t index) const {
+  assert((val.getType().isa<MemRefType>() ||
+             val.getType().isa<UnrankedMemRefType>()) &&
```
tungld, line 396: Can we get dim from an UnrankMemRefType?

src/Dialect/ONNX/MLIRDialectBuilder.cpp
```diff
@@ -392,6 +392,12 @@ memref::CastOp MemRefBuilder::cast(Value input, MemRefType outputType) const {
...
+         "memref::DimOp expects input operand to have MemRefType or "
+         "UnrankedMemRefType");
+  assert(index >= 0 && "Expecting a vaid index");
```
tungld, line 399: `vaid` => `valid'

test/mlir/onnx/onnx_lowering_gru_op.mlir
```diff
@@ -16,23 +16,26 @@ func private @test_gru_forward_mode(%arg0: tensor<7x2x3xf32>, %arg1: tensor<1x12
 // CHECK-DAG:       [[VAR_c3_:%.+]] = arith.constant 3 : index
 // CHECK-DAG:       [[VAR_c2_:%.+]] = arith.constant 2 : index
 // CHECK-DAG:       [[VAR_c0_:%.+]] = arith.constant 0 : index
+// CHECK-DAG:       [[UCC_PARAM_3:%.+]] = builtin.unrealized_conversion_cast %arg3 : memref<1x24xf32> to tensor<1x24xf32>
```
tungld, line 19: Just my curious to learn who generated this  `unrealized_conversion_cast`. Could you please elaborate a bit? thanks!

tungld (commented):
It's great that you found the root cause. Should we add `typeConverter` to all lowering patterns? I see you just added it to element-wise ops and Identity op.

test/mlir/onnx/onnx_lowering_gru_op.mlir
etiotto: The `unrealized_conversion_cast` is injected by MLIR after lowering the operations to reconcile types. It was introduced in MLIR around Jan 2021, this is the RFC https://llvm.discourse.group/t/open-meeting-1-14-dialect-conversion-and-type-conversion-the-question-of-cast-operations/2566. Essentially an `unrealized_conversion_cast` is a promise that the source and destination types will converge eventually. It is used to allow temporary intermixing of type systems during partial lowering.

Iy serve a similar purpose to our own `krnl.dummy_cast` AFAIK, and in this PR I also had to make the lowering of `onnx.Loop` and `onnx.Scan` use the new cast opcode rather than dummy casts.  I wanted to keep the "surface" of this PR as small as possible, so I plan to completely remove dummy_cast in an upcoming PR.

src/Dialect/ONNX/MLIRDialectBuilder.cpp
etiotto: done

src/Dialect/ONNX/MLIRDialectBuilder.cpp
etiotto: Yes, according to https://mlir.llvm.org/docs/Dialects/MemRef/#memrefdim-mlirmemrefdimop the source operand can have type "unranked.memref of any type values or memref of any type values"

etiotto:
The Windows CI has failed again with the following "error": 
```
Check failure on line 72 in Build log 

@azure-pipelinesazure-pipelines
/ MLIR-Windows-CI
Build log #L72
Artifact cannot be uploaded because max quantity has been exceeded or the payment instrument is invalid. https://aka.ms/artbilling for details."
```

@sstamenova can you kindly take a look?  If this is going to take time to get resolved I would ask for permission to merge the PR given it is large and we are in urgent need to upgrade the LLVM version we use in the project @AlexandreEichenberger  ?

AlexandreEichenberger:
@sstamenova I concur with @etiotto that it is important for us to be back near the head of the LLVM compiler as we will need to integrate some recent fixes. So looking into what is happening with the Window CI is very important for forward progress.

I don't have access to the Azure CI pages, but a quick search resulted in this (https://stackoverflow.com/questions/64800434/azure-artifact-cannot-be-uploaded-because-max-quantity-has-been-exceeded-or-the)

> Azure Artifacts includes a free usage tier of 2 GB. Any usage below this level isn't billed to your subscription. Above this limit, we charge you for your actual usage. 

Is it possible that the new LLVM get's us over the 2GB quota of free usage?

@caoimhinuibrian any suggestions if that were the case

AlexandreEichenberger:
@etiotto thanks for this amazing investigation and fixes of bugs to integrate the latest LLVM, impressive.

src/Dialect/ONNX/MLIRDialectBuilder.cpp
AlexandreEichenberger, line 399: Typo: vaid -> valid

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -3652,18 +3652,17 @@ static LogicalResult verify(ONNXDepthToSpaceOp op) {
     return op.emitError("Input should have a rank of four");
 
   // Check blocksize.
-  APInt blocksize = operandAdaptor.blocksize().getValue();
-  if (blocksize.isNegative())
+  int64_t blocksize = operandAdaptor.blocksize();
```
AlexandreEichenberger, line 3655: So they changed the convention for the named parameters, they return now the int? Good to know

AlexandreEichenberger (approved):
LGTM

src/Dialect/ONNX/MLIRDialectBuilder.cpp
etiotto: > @etiotto thanks for this amazing investigation and fixes of bugs to integrate the latest LLVM, impressive.

You are welcome, it was a good learning experience.

src/Dialect/ONNX/ONNXOps.cpp
etiotto: yes, one of the several API changes that went into MLIR.

AlexandreEichenberger:
@sstamenova @stephenneuendorffer 

We build mlir successfully, but when we are about to "save it" (not familiar with the terminology, it may be referred to as an "artifact"?) We most likely run out of free space.

I don't have access to the administrative account, but:
1) do we maybe have space for 1 (the currently cached version) but when we add a new one (because we are testing an upgrade, it in effect need space for 2 versions, and then we run out of space)
2) we can have multiple cached versions, but the new one just happen to be too big.
3) we just have accumulated many many versions of MLIR

If it is 1), then maybe there is a way to forget about the cached copy, accept the PR, then delete the old copy, and the future PR using the new one will be ok. (This is bad because folks may still run CIs based on older PRs that use the old MLIR...)

If its 2), maybe there is a way to remove the .os and other files that are not needed, and only keep the libraries... Basically to only have the minimum info needed.

Obviously, if there is a way to reduce the size (solution 2), then that may also help the first case above.

The error log may provide some hints... it say `Number of Chunks Uploaded: 6,148` and  `Total Number of Chunks: 80,316`

So if one new version of MLIR is 6K chunks but we already have 80K chunks, maybe it is just that we have many versions of MLIR and we only need to clean all but the last one or two? So maybe we are in case 3 above.

from the error log
```
Content upload statistics:
Total Content: 3,055.1 MB
Physical Content Uploaded: 193.0 MB
Logical Content Uploaded: 551.5 MB
Compression Saved: 358.5 MB
Deduplication Saved: 2,503.6 MB
Number of Chunks Uploaded: 6,148
Total Number of Chunks: 80,316

ApplicationInsightsTelemetrySender correlated 2 events with X-TFS-Session bc4bd192-3ed2-4f0b-9f13-883ae6457d69
##[error]Artifact cannot be uploaded because max quantity has been exceeded or the payment instrument is invalid. https://aka.ms/artbilling for details.
Finishing: Publish mlir artifact
```

sstamenova:
@NathanielMcVicar would be the right person to have a look :)

NathanielMcVicar:
@AlexandreEichenberger I believe it is 3 (too many old builds), and I thought this issue was resolved, but clearly, it's not. I will check with the ONNX team again.

etiotto:
> @AlexandreEichenberger I believe it is 3 (too many old builds), and I thought this issue was resolved, but clearly, it's not. I will check with the ONNX team again.

I want to highlight that, in this case, we are using a newer level of the LLVM compiler than the one we used so far. This means that you will need to cache the LLVM build because the cache will not contain an image for this LLVM level. So you will temporarily need space for both the current LLVM level and the new one. Once we move to the new one (this PR lands) then the old images can be nuked.

AlexandreEichenberger:
@NathanielMcVicar 

> I believe it is 3 (too many old builds), and I thought this issue was resolved, but clearly, it's not. I will check with the ONNX team again.

Great, that is the best case scenario. We need at most 2 version cached, if we use a bit of discipline, namely never try to  have 2 PR doing upgrades of LLVM simultaneously. 

We can manage them manually (e.g. we can drop the old ones by hand), or maybe we can have a system that delete all but the two most recent ones.

NathanielMcVicar:
I've kicked off another Windows build, which should pass. Currently there is enough storage in the onnx-pipelines Azure org for 3 builds. This would typically be fine, since we would be able to update the LLVM version twice a month. Unfortunately, there is currently a timing issue where multiple builds all get kicked off within the same ~3 hour window and all store to the pipeline. To avoid having to manually delete redundant builds in the future I will need to update the script to check again before publishing. @jcwchen for visibility.

etiotto:
Looks like the Windows CI passed, merging main and doing another run.

AlexandreEichenberger:
> where multiple builds all get kicked off within the same ~3 hour window and all store to the pipeline. 

@NathanielMcVicar this is only when we have a new LLVM trying to be cached, is that correct? When we are stable, all PRs running on the same LLVM version, then there is no race condition, is that correct?

AlexandreEichenberger:
@NathanielMcVicar thanks for the quick turnaround.

etiotto:
@NathanielMcVicar Thank you!

NathanielMcVicar:
> When we are stable, all PRs running on the same LLVM version, then there is no race condition, is that correct?

@AlexandreEichenberger Unfortunately there is still a race when the previous artifact expires. We have a couple weeks, but I will address this before it comes up again.

PR #1097, opened 2022-01-12
Emit print statements only when the verbose option is used

etiotto:
Currently running the backend tests (check-onnx-backend) causes unnecessary verbose output. This PR prevents the verbose output to be emitted (by default) so that any test case failure stands out.  The verbose output can be obtained using the -v option.


AlexandreEichenberger (approved):
LGTM

PR #1099, opened 2022-01-13
Add verifier for concat op

YW-Ethan:
First task:
Add verifier for `concat` op, and move verify steps from inferShapes to verifier.


AlexandreEichenberger (approved):
Outstanding.

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2687,6 +2687,49 @@ LogicalResult ONNXConstantOp::inferShapes(
 // Concat
 //===----------------------------------------------------------------------===//
 
+static LogicalResult verify(ONNXConcatOp op){
```
etiotto, line 2690: clang-format will reject this code unfort.
We use clang-format to format the code for us, you can either invoke the command on files you changed from the shell. OR add integration to the editor you are using. I use vscode and there are packages to support clang-format

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2687,6 +2687,49 @@ LogicalResult ONNXConstantOp::inferShapes(
...
+  ONNXConcatOpAdaptor operandAdaptor =
```
etiotto, line 2691: Suggest:

```
ONNXConcatOpAdaptor operandAdaptor(op);
```

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2687,6 +2687,49 @@ LogicalResult ONNXConstantOp::inferShapes(
...
+    ONNXConcatOpAdaptor(op);
+
+  int inputNum = op.getNumOperands();
+  // Check all inputs.
+  for (int i = 0; i < inputNum; ++i) {
```
etiotto, line 2696: We try to follow LLVM coding guidelines, those guidelines indicate it is prefered to use for-range rather than a C-like for loop  (see https://llvm.org/docs/CodingStandards.html#use-range-based-for-loops-wherever-possible) although old code might not follow these rules (if you see it feel free to fix it).

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2687,6 +2687,49 @@ LogicalResult ONNXConstantOp::inferShapes(
...
+    if (!hasShapeAndRank(operandAdaptor.getOperands()[i])) {
+      // Won't be able to do any checking at this stage.
+      return success();
+    }
+  }
+  // Checking value of axis parameter.
+  auto commonType = operandAdaptor.getOperands()[0].getType().cast<RankedTensorType>();
```
etiotto, line 2703: Is there a better way to get the axis parameter. Getting operand zero is not ideal because the reader does not know that op zero is the axis parameter.

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2687,6 +2687,49 @@ LogicalResult ONNXConstantOp::inferShapes(
...
+  auto commonShape = commonType.getShape();
```
etiotto, line 2704: I personally prefer to not use `auto` where is not immediately clear what is the type of the variable.  In this case what is the type returned by getShape() ?  Use that type to declare `commonShape`. This is also LLVM policy (see https://llvm.org/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable)

P.S.: on line 2703 is fine to use `auto` because is clear that the type is `RankedTensorType` given there is a cast.

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2687,6 +2687,49 @@ LogicalResult ONNXConstantOp::inferShapes(
...
+  int64_t commonRank = commonShape.size();
+  int64_t axisIndex = op.axis();
+  if (axisIndex < 0)
+    axisIndex = commonRank + axisIndex;
+  if (axisIndex >= commonRank)
+    return op->emitError("Concat axis value out of bound");
+
+  for (int i = 1; i < inputNum; ++i) {
+    auto currShape =
```
etiotto, line 2713: no `auto` here pls

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2687,6 +2687,49 @@ LogicalResult ONNXConstantOp::inferShapes(
...
+        operandAdaptor.getOperands()[i].getType().cast<RankedTensorType>().getShape();
+    if ((int64_t)currShape.size() != commonRank)
+      return op->emitError("Concat input must all have the same rank");
+    for (int j = 0; j < commonRank; ++j) {
+      if (j == axisIndex) {
```
etiotto, line 2718: Suggest to use ` continue` here rather than an empty if block:
```
if (j == axisIndex) 
  continue;
  
if (currShape[j] != -1 && commonShape[j] != -1 &&
                 currShape[j] != commonShape[j]) {
        return op->emitError(...);
```

In general consider using early-exits in loops, they make the code more readable.   
One more nit :-)  Avoid using `{}` for single statements (https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements)

etiotto (changes requested):
Good first pull request. I have added some comments on style. Some of the project code does not conform to LLVM coding style and, as we change things and improve existing code we should modify the code to make it follow those guidelines (also good to get into the habit so that if you contribute to LLVM it makes life easier :-))

Can you double check we have a lit test to check verification of the concat operator (we should but best to check). Thanks for your first contribution!

src/Dialect/ONNX/ONNXOps.cpp
YW-Ethan: Hi Ettore,
could you explain a little bit more about this? I think the axis parameter is getting from attribute using `op.axis();`, and this `commonType` and `commonShape` are used to compare with other operands (make sure all operands have the same rank and shape).

etiotto: Mi main issue is with using `operandAdaptor.getOperands()[0]` in that line.  I would prefer `operandAdaptor.getOperands().first`

AlexandreEichenberger:

src/Dialect/ONNX/ONNXOps.cpp
AlexandreEichenberger: @etiotto Please explain why the `first` is better. I am probably old school, I tend to use array notation. Always ready to learn new C++ conventions

etiotto: @AlexandreEichenberger actually the name of the member function is not `first()` but `front()`. The advantage of using `front()` rather than subscripting is that member function checks for an empty container:
```
 ReferenceT front() const {
    assert(!empty() && "expected non-empty range");
    return (*this)[0];
  }
```

AlexandreEichenberger: @etiotto are your requests satisfied? If so, can you approve? Tx

YW-Ethan:
Hi @etiotto,
I have added 3 lit tests for concat verifier under `invalid.mlir`. Please let me know if it looks good to you. Thank you!

test/mlir/onnx/invalid.mlir
```diff
@@ -16,3 +16,27 @@ func @test_depth_to_space_default(%arg0 : tensor<1x256x8x16xf32>) -> tensor<1x16
...
+  "std.return"(%1) : (tensor<*xf32>) -> ()
+}
+
+// -----
+
+func @test_concat_verifier_3(%arg0 : tensor<5x5x1x32xf32>, %arg1 : tensor<5x5x3x32xf32>, %arg2 : tensor<5x5x5x32xf32>) -> tensor<*xf32> {
+  // expected-error @+1 {{Concat input dimensions must be all identical, except for dimension on the axis of the concatenation. Expected something compatible with: 'tensor<5x5x1x32xf32>' but got 'tensor<5x5x3x32xf32>' instead.}}  
+  %1 = "onnx.Concat"(%arg0, %arg1, %arg2) { axis = 1 : si64} : (tensor<5x5x1x32xf32>, tensor<5x5x3x32xf32>, tensor<5x5x5x32xf32>)  -> tensor<*xf32>
+  "std.return"(%1) : (tensor<*xf32>) -> ()
+}
```
etiotto, line 42: add a newline here.

etiotto (approved):
LGTM - Thanks!

P.S.: you can address the minor comments I left while rebasing.

PR #1102, opened 2022-01-18
Pass type converter to all ONNX operations

etiotto:
This PR finishes the type conversion related changes started in #1095. Specifically this pull request modifies all ONNX operators to pass a type converter to their base class.


AlexandreEichenberger (approved):
LGTM

PR #1103, opened 2022-01-18
Nuke KrnlDummyCastOp now that we use MLIR's UnrealizedConversionCastOp

etiotto:
In PR  #1095 we have removed all uses of KrnlDummyCastOp, this PR finish the job and removes all remaining traces of that operation from the code base. 


src/Dialect/Krnl/CMakeLists.txt
```diff
@@ -13,7 +13,6 @@ add_onnx_mlir_library(OMKrnlOps
   KrnlOps.cpp
   KrnlTypes.cpp
   KrnlHelper.cpp
-  SimplifyKrnl.cpp
```
sstamenova, line 16: I feel like there were some additional dependencies needed here because of SimplifyKrnl.cpp and perhaps some of the depends/libs below can be removed now? It was a while back when I touched this, so I am not sure if it's still pertinent.

sstamenova: Also, yay!

etiotto: Thanks I was able  to remove one of the dependency.

PR #1104, opened 2022-01-18
Add an option to stop compilation after objects are created.

etiotto:
Add the `--emitObj` option to `onnx-mlir`. This option stops the compiler after an object is created (skips the link step).

### DIFF
--- a/src/Compiler/CompilerUtils.cpp
+++ b/src/Compiler/CompilerUtils.cpp
@@ -4,7 +4,7 @@
 
 //===-------------------------- CompilerUtils.cpp -------------------------===//
 //
-// Copyright 2019-2021 The IBM Research Authors.
+// Copyright 2019-2022 The IBM Research Authors.
 //
 // =============================================================================
 //
@@ -302,7 +302,7 @@ struct Command {
 void setTargetCPU(const std::string &cpu) { mcpu = cpu; }
 void setTargetTriple(const std::string &triple) { mtriple = triple; }
 
-void LoadMLIR(string inputFilename, mlir::MLIRContext &context,
+static void LoadMLIR(string inputFilename, mlir::MLIRContext &context,
     mlir::OwningModuleRef &module) {
   // Handle '.mlir' input to the ONNX-MLIR frontend.
   // The mlir format indicates that one or more of the supported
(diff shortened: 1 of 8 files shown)


src/Compiler/CompilerUtils.cpp
```diff
@@ -4,7 +4,7 @@
 
 //===-------------------------- CompilerUtils.cpp -------------------------===//
 //
-// Copyright 2019-2021 The IBM Research Authors.
+// Copyright 2019-2022 The IBM Research Authors.
```
AlexandreEichenberger, line 7: can we say `2019-now`

src/Compiler/CompilerUtils.cpp
```diff
@@ -504,13 +508,7 @@ std::string compileModuleToSharedLibrary(
...
 void compileModuleToJniJar(
     const mlir::OwningModuleRef &module, std::string outputBaseName) {
-
-  string bitcodePath = outputBaseName + ".bc";
-  genLLVMBitcode(module, bitcodePath, outputBaseName);
-  llvm::FileRemover bitcodeRemover(
-      bitcodePath, !keepFiles(KeepFilesOfType::Bitcode));
-
-  string modelObjPath = genModelObject(bitcodePath, outputBaseName);
+  string modelObjPath = compileModuleToObject(module, outputBaseName);
```
AlexandreEichenberger, line 511: nice reuse

src/Compiler/CompilerUtils.hpp
```diff
@@ -37,47 +37,24 @@ extern llvm::cl::opt<std::string> instrumentONNXOps;
...
-
 void addKrnlToLLVMPasses(mlir::OpPassManager &pm);
 
 void processInputFile(std::string inputFilename, mlir::MLIRContext &context,
     mlir::OwningModuleRef &module, std::string *errorMessage);
-
 void processInputArray(const void *onnxBuffer, int bufferSize,
     mlir::MLIRContext &context, mlir::OwningModuleRef &module);
 
-onnx_mlir::InputIRLevelType determineInputIRLevel(
```
AlexandreEichenberger, line 65: You removed these because you made them static, only used in the file, right?

AlexandreEichenberger (approved):
LGTM.

src/Compiler/CompilerUtils.cpp
etiotto: thanks

src/Compiler/CompilerUtils.hpp
etiotto: Yes. I made a few function static were they aren't referenced from other compilation units. I think that, in general, is good to limit external functions to just the one that are strictly required.

src/Compiler/CompilerUtils.cpp
etiotto: I do not think we can... AFAIK copyright needs to be updated when we change the content of the file.

src/Compiler/CompilerUtils.hpp
tungld: @etiotto @AlexandreEichenberger making these function static means they are not sharable anymore. Other projects using these functions would fail. This is regression.

PR #1105, opened 2022-01-19
Add one rule for swapping ReshapeOp and MatMulOp

tungld:
This patch adds one rule to swap ReshapeOp and MatMulOp, which puts all ReshapeOp together and the existing fusion rule for ReshapeOp is automatically applied to remove ReshapeOp.


src/Dialect/ONNX/Rewrite.td
```diff
@@ -312,6 +317,22 @@ def RemoveIdentityReshapePattern:  Pat<
...
+ // produces a type consistent with the ones in A and B.
+ // This pattern sometimes happens to models converted by tf2onnx.
+ (ONNXReshapeOp:$res2 (ONNXMatMulOp (ONNXReshapeOp:$res1 $A, $_), $B), $_),
+ // Just multiply A by B.
+ [(ONNXMatMulOp $A, $B)],
+ [(HasRankOf<3> $A), (HasRankOf<3> $res2),
+  (HasRankOf<2> $B), (HasRankOf<2> $res1),
+  (SameDimAt<0, 0> $A, $res2), // A and the final result has the same 1st dim. 
+  (SameDimAt<1, 1> $A, $res2), // A and the final result has the same 2nd dim. 
+  (SameDimAt<1, 2> $B, $res2),  // B and the final result has the same last dim.
```
etiotto, line 332: nit: extra space before comment '//'.

src/Dialect/ONNX/Rewrite.td
```diff
@@ -312,6 +317,22 @@ def RemoveIdentityReshapePattern:  Pat<
   // Check that val has the specified shape.
   [(HasSpecifiedConstantShape $val, $shape)]>;
 
+def RemoveReshapeBeforeMatmul : Pattern<
```
etiotto, line 320: Suggests to modify the name of this pattern to `RemoveReshapeBracketingMatMul` because the pattern expects 2 Reshape operations 'around' a matrix multiply operation (rather than just one before it).

src/Dialect/ONNX/Rewrite.td
```diff
@@ -312,6 +317,22 @@ def RemoveIdentityReshapePattern:  Pat<
...
+ // Reshape is redudant if A and B can be multiplied directly.
+ // E.g. the 1st reshape keep the last dim of A unchanged and the 2nd reshape
+ // produces a type consistent with the ones in A and B.
+ // This pattern sometimes happens to models converted by tf2onnx.
+ (ONNXReshapeOp:$res2 (ONNXMatMulOp (ONNXReshapeOp:$res1 $A, $_), $B), $_),
+ // Just multiply A by B.
+ [(ONNXMatMulOp $A, $B)],
+ [(HasRankOf<3> $A), (HasRankOf<3> $res2),
```
etiotto, line 328: It is unfortunate the pattern is specific to tensors with rank equal to 3. I was wondering if there is a way to generalize the pattern, but probably this is OK for now.

AlexandreEichenberger: I concur, it is really a flattening of all but the innermost dim.

AlexandreEichenberger (commented):
Does the job for the example; technically, it should be extended for n dimensions. 
Also, if the results are fed exclusively to "activation" type operators, that should work too. That is probably best left for a later fix.

tungld:
@AlexandreEichenberger @etiotto I simplify the rule. Now the rule is only to swap Reshape and MatMul, which puts all Reshape together, then the existing fusion rule for reshape is automatically applied to remove reshapes.

AlexandreEichenberger:
@tungld You read my mind. I was thinking about it that indeed, instead of checking for the "sandwich" of matmul by 2 reshapes, it is really that it is ok to migrate the reshape through the mat mull. then we have 2 reshape with nothing in-between,  namely reshape a -> b followed by reshape b -> a.

In fact, probably many of the activation / add / ... ops can forward reshape through them.

Now interestingly, it might be that on a CPU, it is much more efficient to execute a <1000 x a> * <a x b>  matmul rather than a <10 x 100 x a> * <a x b> matmul, because tiling prefer large dimensions... So it may be beneficial in some cases to have reshapes. Something to investigate separately.

AlexandreEichenberger (approved):
LGTM

test/mlir/onnx/onnx_canonicalization.mlir
```diff
@@ -180,6 +180,41 @@ func @test_reshape_removal(%arg0: tensor<10x11x12x13xf32>) -> tensor<10x11x12x13
...
+  return %2 : tensor<3x5x10x1xf32>
+  // CHECK-NEXT: "onnx.MatMul"(%arg0, %arg1) : (tensor<3x5x10x20xf32>, tensor<20x1xf32>) -> tensor<3x5x10x1xf32>
+  // CHECK-NOT: "onnx.Reshape"
+}
+
+// -----
+
+// Check the removal of reshapes that are used for matmul's input and output.
+// The 1st reshape is moved down then fused with the 2nd reshape to become a single reshape.
+// a single reshape.
```
etiotto, line 201: Extra comment line left over.

src/Dialect/ONNX/Rewrite.td
etiotto: no longer relevant given the new implementation.

src/Dialect/ONNX/Rewrite.td
etiotto: no longer relevant given the new implementation.

etiotto (approved):
LGTM - just one minor comment to address before merging.

Thanks @tungld !

test/mlir/onnx/onnx_canonicalization.mlir
tungld: Good catch! Thanks!

tungld:
>Now interestingly, it might be that on a CPU, it is much more efficient to execute a <1000 x a> * matmul rather than a <10 x 100 x a> * matmul, because tiling prefer large dimensions... So it may be beneficial in some cases to have reshapes. Something to investigate separately.

I think so.

AlexandreEichenberger:
@tungld it's ready

Issue #797, opened 2021-07-23
mtriple, mcpu and march


sstamenova:
onnx-mlir right now provides `mtriple` and `mcpu` as options to both the onnx-mlir executable as well as the backend tests, however, it does not provide `march`. For tools such as `opt` or `clang` generally `mtriple` defines both `march` and `mcpu`, but for more granular control (if `mtriple` is not desirable), both `mcpu` and `march` exist as options.

Something to consider is whether `mcpu` is needed for onnx-mlir (since `mtriple` already exists as an option) and if `march` should be added as well.

AlexandreEichenberger:
@sstamenova can we close this?

sstamenova:
I would like to see an answer to the question before we close it. That way if it comes up again for someone, they can find it.

sstamenova:
@gongsu832 or @etiotto : since you've touched these options in the past, are both `mcpu` and `mtriple` needed? Should `march` be added for completeness?

gongsu832:
I don't think I did anything with these options. If I ever touched them, it was probably only because I need to preserve them when I made changes to other parts.

etiotto:
From LLVM `llc` (similar for `opt`) documentation:
```
-mtriple=<target triple>
Override the target triple specified in the input file with the specified string.

-march=<arch>
Specify the architecture for which to generate assembly, overriding the target encoded in the input file. See the output of llc -help for a list of valid architectures. By default this is inferred from the target triple or autodetected to the current architecture.

-mcpu=<cpuname>
Specify a specific chip in the current architecture to generate code for. By default this is inferred from the target triple and autodetected to the current architecture.
``` 

So `-march` and `-mcpu` are used to provide finer level of control than `-mtriple`.  

Currently `onnx-mlir` has `-mtriple` and `-mcpu`, however it lacks `-march`.  I think, for completeness and flexibility we should provide `-march` as well.

PR #1107, opened 2022-01-19
Add support for -march in onnx-mlir

etiotto:
Currently onnx-mlir has -mtriple and -mcpu, however it lacks -march. For completeness and flexibility this PR adds -march as well.

This addresses issue https://github.com/onnx/onnx-mlir/issues/797


sstamenova:
One thing to consider is that the triple is usually a combination of the arch and cpu, so if someone is passing triple + arch/cpu, it's probably a mistake. Are there cases today where they are both passed?

AlexandreEichenberger:
> One thing to consider is that the triple is usually a combination of the arch and cpu, so if someone is passing triple + arch/cpu, it's probably a mistake. Are there cases today where they are both passed?

@sstamenova on our z machines, we need to pass `-mtriple=s390x-ibm-loz -mcpu=z15` where the first one say the machine family/company/os and the second say the specific version of the machine. So for Z machines, both options are needed if we want to generate for an ISA of a particular type (say z15). The triple would be different if we were targeting another os on the same machine, say zOS.

AlexandreEichenberger:
Can you comment on the use of cpu vs arch, and how we should use it (eventually) in ONNX MLIR

etiotto:
This is how I understand it (from llc perspective):
  - the `-march` flag is used to specify the target architecture, it allows the compiler to generate instructions that are part of the ISA of the target architecture. For example `-march=x86-64` would instruct the compiler to generate instruction that run correctly on any processor that implements the x86-64 microarchitecture.
  - the `-mcpu` flag  is used to specify a particular processor, for example `--mcpu=pentiumpro` and would allow the compiler to schedule instructions specifically for that processor.
  
Within `onnx-mlir` we can use -march and -mcpu to determine whether the target machine support SIMD instructions (for example).

AlexandreEichenberger (approved):
I agree with that interpretation. Would you mind pasting it in the code so that we may use as a reference for the future, thanks

etiotto:
> I agree with that interpretation. Would you mind pasting it in the code so that we may use as a reference for the future, thanks

For that we need to first set the target triple in the MLIR module.  Fortunately MLIR allows annotation of a module operation with attributes (in fact that is how we already set the data layout).  The relevant info is here https://mlir.llvm.org/docs/Dialects/LLVM/#data-layout-and-triple

Once a module has the triple annotation we can then use it to retrieve the target architecture. This is how we can do it:
```
      auto targetTripleAttr =
          module->getAttr(LLVM::LLVMDialect::getTargetTripleAttrName());
      StringRef targetTriple = targetTripleAttr.cast<StringAttr>().strref();
      llvm::Triple triple(targetTriple);
      llvm::Triple::ArchType targetArch = triple.getArch();
```

The LLVM Triple object allows other queries (we can retrieve the target OS and/or the vendor name if necessary).
Having the target arch the lowering code of any operation can then determine whether it can generate certain instructions (e.g. SIMD instructions).

AlexandreEichenberger:
> The LLVM Triple object allows other queries (we can retrieve the target OS and/or the vendor name if necessary).
Having the target arch the lowering code of any operation can then determine whether it can generate certain instructions (e.g. SIMD instructions).

As I said before, triple is not sufficient to identify an arch for some targets. E.g. Z's arch is `loz`, and we need the `-mcpu` on Z to identify the actual ISA (e.g. z13, vs a14,..)

etiotto:
I have opened PR 

> > The LLVM Triple object allows other queries (we can retrieve the target OS and/or the vendor name if necessary).
> > Having the target arch the lowering code of any operation can then determine whether it can generate certain instructions (e.g. SIMD instructions).
> 
> As I said before, triple is not sufficient to identify an arch for some targets. E.g. Z's arch is `loz`, and we need the `-mcpu` on Z to identify the actual ISA (e.g. z13, vs a14,..)

That is unfortunate. At least in principle an architecture should match an ISA. Knowing the precise target ISA would be sufficient to determine whether SIMD instructions are part of the ISA. 
AFAIK the MLIR module can accept only the triple (as an attribute) ... so we will have to find another mechanism to encode the setting of `mcpu` in the module or pass a flag to the lowering pass to communicate the `mcpu` value.

Issue #1031, opened 2021-12-01
Pull Requests should always be updated such that they preserve history


sstamenova:
Recently I've noticed people are updating their PRs by force pushing in such a way that the previous commits disappear. This means a reviewer doesn't have an easy way to compare what was changed between the two versions of the PR making it really difficult to track updates. In general, I would expect that PRs should always preserve history so that changes can be compared easily and things don't get missed. Is there a different paradigm expected for onnx-mlir now?

gongsu832:
This is likely because starting recently (about two or three weeks ago) there seemed to be some problem with the PRs not keeping track of changes in the target master (now main) branch.

For example, I submitted a PR1 at some point against main. Then Kevin submitted another PR2 afterwards and PR2 got merged into main first. So I rebased my PR1 to the new main which now includes commits from PR2. However, my PR1 doesn't recognize that PR2 is now part of the main. Instead, it thinks all the commits of PR2 are part of my PR1. That completely screws up the history of my PR1. So I had to force push to get rid of the commits of PR2 to clean things up.

There are also times where git simply screwed up the commit history of a PR after a lot of merging (of other PRs) and rebasing. So I don't think people are intentionally using forced push. Most likely it's due to git somehow screwing up the PR commit history.

etiotto:
My issue with merging rather than rebasing is that the PR will should several partial commits that are meaningless in the history, in addition it will also pollute the history with spurious informations from merging base. I prefer to rebase all my PRs.

AlexandreEichenberger:
I know that there is a recommendation to use rebase, but it seems to me that the "merge" approach works well because when we finally push the changes of a PR to the main branch, we use a "squash and merge" that appears to put all the PR changes after any of the current one on main.

I feel that removing the superfluous history commit (e.g. "I am fixing for the nth time a formatting issue") could be removed using the `rebase -i`. I can try it on my next PR when only merge are used in the history. Think it works, better to double check.

I am not a sophisticated git user, I am happy to follow the recommendations of experts.

AlexandreEichenberger:
This PR is a continuation of #955

etiotto:
@AlexandreEichenberger looking at recent commits I see that a "squashed and merge" commit has a commit msg that contains the list of msgs that belonged to the individual commits before they were squashed, example (not pointing fingers - just took a random recent commit as example here): 
```
commit 0d483afdcd4696c7d14c562bbf02c170b802bb29
[...]

PR #1108, opened 2022-01-19
Document workflow as agreed in issue #1031

etiotto:
In the discussion in #1031 we agreed that the workflow for submitting pull requests for review should be:

1. create new PRs as drafts, allow rebasing commits until the PR is ready for code review
2. once the PR is ready for code review, change it to a regular PR. Reviewers are free to add comments at this point
3. when the PR is in review refrain from squashing commits, this should preserve history while reviews are in progress
4. when the PR has been approved, allow authors to again squash commits and edit the commit msgs
5. in its final approved form the PR should contain a single commit, with a single meaningful message

### DIFF
--- a/docs/Workflow.md
+++ b/docs/Workflow.md
@@ -99,8 +99,13 @@ Commit your changes, always using the `-s` flag in order to sign your commits.
 git commit -s
 ```
 
-Likely you'll go back and edit/build/test some more than `commit --amend -s`
-in a few cycles.
+Likely you'll go back and edit/build/test some more than `commit --amend -s` in a few cycles. To avoid polluting the history you should squash all your feature commits together by using the following command:
+
+```sh
+git commit -i rebase HEAD~n (where n is the number of commits you want to squash)
+```
+
+An editor will pop up, and you should mark all commits except the first one with an 's' to indicate that those are the commits you want to squash with the first one. At this point you will be able to edit the commit message (remove all individual commit messages except for the first one). After rebasing, the `git log` should contain only one commit for your feature.
 
 ### Step 6: Keep your branch in sync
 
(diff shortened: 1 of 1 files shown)


gongsu832:
I'm not sure discussion in https://github.com/onnx/onnx-mlir/issues/1031 warrants such a "restrictive" PR policy. Everyone has different style of writing code and submitting PRs. As long as the process don't create undue management overhead, I don't think we should mandate a particular way of creating PRs. For example, I never create draft PRs because all my PRs are already tested locally and they are ready for review when submitted (with occasional need for additional commits to fix format, etc.), whereas other people like to start with a draft PR and push every single change to the PR (which IMO is actually the source of polluting the commit history).

So I think the root problem is more of a matter of people need to be more disciplined with the commits in the PRs. Mandating all PRs starting from a draft isn't going to fix that.

AlexandreEichenberger:
Maybe we can add a few "encouraged" instead of should.

In general, would it be acceptable to not bother about `git commit -i` until the end? 

Personally, I find that clicking the "Update Branch" on the website is important. Sometimes there are many PRs waiting to be squashed to main, and it is a sequential process, so I like clicking on that. 

Other question: say a PR is finally ready to go, up to date with current main, reviews, but history needs updating. So the recommendation is that the author of the PR does a final `git commit -i` followed by a push. Will that re-trigger the CIs?

etiotto:
> I'm not sure discussion in #1031 warrants such a "restrictive" PR policy...
I lifted the exact agreed steps from that discussion (https://github.com/onnx/onnx-mlir/issues/1031#issuecomment-996093889). Nevertheless, I agree we need to support different styles. The ultimate goals of that discussion was to end up with:
  - committed PRs that contain a clean log message
  - allow for incremental code reviews where comments from reviewers are addressed in specific commits
  
 I can attempt to modify the document changes to "relax" the requirements of starting from a draft PR.
 
 > In general, would it be acceptable to not bother about git commit -i until the end?
 IMO yes, as long as in the end the log has a clear concise message explaining what the patch does.
 
 > Personally, I find that clicking the "Update Branch" on the website is important. Sometimes there are many PRs waiting to be squashed to main, and it is a sequential process, so I like clicking on that.
 I think that is fine, as long as the commit msg is edited to avoid having the individual commits mgs that are being squashed as the final commit msg. 
 
> Other question: say a PR is finally ready to go, up to date with current main, reviews, but history needs updating. So the recommendation is that the author of the PR does a final git commit -i followed by a push. Will that re-trigger the CIs?

Yup. Any push action will trigger CI again.

etiotto:
@gongsu832 @AlexandreEichenberger I attempted to address the concerns expressed above. Please take another look.

AlexandreEichenberger:
By the way, I found an incredible easy way to achieve the stated goals from the GitHub interface.

Only do git merge during development and review... full history is preserved.
Hit "merge" button from interface as many times as needed.

When you are ready to squash and merge, the interface actually list in a text box the actual text that will be put in the log. By default, it is the current list of commit, but one can edit it to whatever one wants. So remove all but the first author/date and last signed-off by... and write your harts content in the  middle.

```
Author: Alexandre Eichenberger <alexe@us.ibm.com>
Date:   Wed Jan 19 15:33:14 2022 +0000

<< edit here whatever you want>>
    
```

Thus, we have the nice text, and the "squash" of the "squash and merge" button already present the changes as only one "event" in the history.

So no need to do any git commit, merge, rebase, if one is content to use the GitHub interface to finalize the text on Github.

AlexandreEichenberger:
@etiotto if that approach above works, I would suggest to split the new approach into two: one though the git commands and one through the GitHub website. Folks can choose what they feel more comfortable with.

doru1004:
> When you are ready to squash and merge, the interface actually list in a text box the actual text that will be put in the log. By default, it is the current list of commit, but one can edit it to whatever one wants. So remove all but the first author/date and last signed-off by... and write your harts content in the middle.

This is what I meant by being able to adjust the message when you do the actual merge.

doru1004:
Rebasing is important for projects that do not do squashed commits and attempt to capture the full commit history. For our case, squashed commits go to the top of the main branch and the commit blurb is editable prior to merging but after the squashing has occurred. By default the message contains the commit messages of all squashed commits but all that can be edited out before clicking the merge button. So actually the only thing we need to be careful to do is to edit out the part containing the squashed commits prior to the final merge.

AlexandreEichenberger:
> So actually the only thing we need to be careful to do is to edit out the part containing the squashed commits prior to the final merge.

exactly. We can still provide the `git` commands for folks that prefer them. I find that using merge all the way through, relying on the squash that is fool-proof, and simply editing the text in the browser works best for me. So I am requesting that we add this in our doc

doru1004:
> So I am requesting that we add this in our doc

Something like: "The commit history will be replaced by the patch name before the merge is final."

sstamenova:
I think the github options are more straightforward also, I think documenting the expectation to use them and provide a meaningful commit message is the best approach - instead of having a second option through git commands. The flow with the git commands is not as straightforward and likely to confuse people.

AlexandreEichenberger:
@etiotto do you want to update the PR?

etiotto:
Agreed with @AlexandreEichenberger that this PR is no longer necessary. We can start a new one and document the workflow in a simpler way, specifically how to use the GUI to create a clean final commit message upon merging a PR.

PR #1110, opened 2022-01-20
Add support for random normal like operation

doru1004:
Add support for random normal like operation.


AlexandreEichenberger (approved):
LGTM, pending passing the CIs.

etiotto:
Hi @doru1004, as part of the work for upgrading the LLVM level I modified the way we create ONNX operators. Now we need to pass a typeConverter like:
```
populateLoweringONNXRandomNormalLikeOpPattern(patterns, typeConverter, ctx);
```

src/Conversion/ONNXToKrnl/Math/RandomNormalLike.cpp
```diff
@@ -0,0 +1,89 @@
...
+#include "src/Dialect/ONNX/ShapeInference/ONNXShapeHelper.hpp"
+
+#include <ctime>
+
+using namespace mlir;
+
+struct ONNXRandomNormalLikeOpLowering : public ConversionPattern {
+  ONNXRandomNormalLikeOpLowering(MLIRContext *ctx)
+      : ConversionPattern(
+            mlir::ONNXRandomNormalLikeOp::getOperationName(), 1, ctx) {}
```
etiotto, line 28: Now, after upgrading to latest LLVM level we should pass typeConverter to `ConversionPattern`:
```
 : ConversionPattern(
      typeConverter, mlir::ONNXRandomNormalLikeOp::getOperationName(), 1, ctx) {}

File: test/mlir/krnl/krnl_random_normal_lowering.mlir [hunk 2]
```diff
@@ -54,3 +54,74 @@ func @test_random_normal_lowering() -> memref<3x4x5xf32> {
...
+  %c3_1 = arith.constant 3 : index
+  %5 = memref.dim %arg0, %c3_1 : memref<3x4x?x?xf32>
+  %6 = arith.muli %4, %5 : index
+  %cst = arith.constant 0.000000e+00 : f32
+  %cst_2 = arith.constant 1.000000e+00 : f32
+  %cst_3 = arith.constant 2.000000e+00 : f32
+  "krnl.random_normal"(%2, %6, %cst, %cst_2, %cst_3) : (memref<3x4x?x?xf32>, index, f32, f32, f32) -> ()
+  return %2 : memref<3x4x?x?xf32>
+
+  // CHECK: llvm.func @get_random_normal_value_f32(!llvm.ptr<f32>, i64, f32, f32, f32)
```
etiotto, line 79: Suggest use of CHECK-LABEL to check the function definition.

test/mlir/krnl/krnl_random_normal_lowering.mlir
```diff
@@ -54,3 +54,74 @@ func @test_random_normal_lowering() -> memref<3x4x5xf32> {
...
+  // CHECK: llvm.func @malloc(i64) -> !llvm.ptr<i8>
```
etiotto, line 80: Suggest use of CHECK-SAME

test/mlir/krnl/krnl_random_normal_lowering.mlir
```diff
@@ -54,3 +54,74 @@ func @test_random_normal_lowering() -> memref<3x4x5xf32> {
...
+  // CHECK: llvm.func @test_random_normal_dynamic_lowering(%arg0: !llvm.ptr<f32>, %arg1: !llvm.ptr<f32>, %arg2: i64, %arg3: i64, %arg4: i64, %arg5: i64, %arg6: i64, %arg7: i64, %arg8: i64, %arg9: i64, %arg10: i64) -> !llvm.struct<(ptr<f32>, ptr<f32>, i64, array<4 x i64>, array<4 x i64>)> {
```
etiotto, line 81: CHECK-LABEL

test/mlir/krnl/krnl_random_normal_lowering.mlir
```diff
@@ -54,3 +54,74 @@ func @test_random_normal_lowering() -> memref<3x4x5xf32> {
...
+  // CHECK: [[C3:%.+]] = llvm.mlir.constant(3 : index) : i64
+  // CHECK: [[C4:%.+]] = llvm.mlir.constant(4 : index) : i64
+  // CHECK: llvm.mlir.constant(1 : index) : i64
```
etiotto, line 84: missing assignment to a variable. In general is it possible to reduce the number of CHECKs to the minimum required to ensure the test contain correct lowering ?

test/mlir/krnl/krnl_random_normal_lowering.mlir
```diff
@@ -54,3 +54,74 @@ func @test_random_normal_lowering() -> memref<3x4x5xf32> {
...
+  // CHECK: [[MUL1:%.+]] = llvm.mul %arg6, %arg5  : i64
+  // CHECK: [[MUL2:%.+]] = llvm.mul [[MUL1]], [[C4]]  : i64
+  // CHECK: %[[MUL3:.+]] = llvm.mul [[MUL2]], [[C3]]  : i64
+
+  /// Allocate aligned tensor:
+  // CHECK: [[POINTER:%.+]] = llvm.mlir.null : !llvm.ptr<f32>
+  // CHECK: llvm.getelementptr [[POINTER]][%[[MUL3]]]
+  // CHECK: llvm.ptrtoint
+  // CHECK: llvm.mlir.constant(16 : index) : i64
+  // CHECK: llvm.add
```
etiotto, line 94: Can the CHECKS that do not assign their results to a variable be removed?

test/mlir/onnx/onnx_lowering.mlir
```diff
@@ -2330,3 +2330,69 @@ func @test_resize1(%arg0 : tensor<3x4xf32>) -> tensor<*xf32> {
...
+    %0 = "onnx.RandomNormalLike"(%arg0) {dtype = 2 : si64, mean = 0.0 :f32, scale = 1.0 : f32, seed = 2.0 : f32} : (tensor<3x4x5xf32>) -> tensor<*xf32>
+    "std.return"(%0) : (tensor<*xf32>) -> ()
+// CHECK-LABEL:  @test_random_normal_like2
+// CHECK-DAG:       [[ALLOC:%.+]] = memref.alloc() {alignment = 16 : i64} : memref<3x4x5xf64>
+// CHECK-DAG:       [[ALL_VALUES:%.+]] = arith.constant 60 : index
+// CHECK-DAG:       [[MEAN:%.+]] = arith.constant 0.000000e+00 : f64
+// CHECK-DAG:       [[SCALE:%.+]] = arith.constant 1.000000e+00 : f64
+// CHECK-DAG:       [[SEED:%.+]] = arith.constant 2.000000e+00 : f64
+// CHECK-DAG:       "krnl.random_normal"([[ALLOC]], [[ALL_VALUES]], [[MEAN]], [[SCALE]], [[SEED]]) : (memref<3x4x5xf64>, index, f64, f64, f64) -> ()
+// CHECK:           return [[ALLOC]] : memref<3x4x5xf64>
```
etiotto, line 2359: This appears strange to me, the lowered code returns a memref with element type `f64`, however the function return (prior to lowering) returns a vector with element type `f32`.  I do not understand how an operator in the function body can cause the element type of the function return type to change.

test/mlir/onnx/onnx_lowering.mlir
```diff
@@ -2330,3 +2330,69 @@ func @test_resize1(%arg0 : tensor<3x4xf32>) -> tensor<*xf32> {
...
+// CHECK-DAG:       [[MUL1:%.+]] = arith.muli [[C12]], [[DIM2]] : index
+// CHECK-DAG:       [[C3:%.+]] = arith.constant 3 : index
+// CHECK-DAG:       [[DIM3:%.+]] = memref.dim %arg0, [[C3]] : memref<3x4x?x?xf32>
+// CHECK-DAG:       [[ALL_VALUES:%.+]] = arith.muli [[MUL1]], [[DIM3]] : index
+// CHECK-DAG:       [[MEAN:%.+]] = arith.constant 0.000000e+00 : f32
+// CHECK-DAG:       [[SCALE:%.+]] = arith.constant 1.000000e+00 : f32
+// CHECK-DAG:       [[SEED:%.+]] = arith.constant 2.000000e+00 : f32
+// CHECK-DAG:       "krnl.random_normal"([[DYN_ALLOC]], [[ALL_VALUES]], [[MEAN]], [[SCALE]], [[SEED]]) : (memref<3x4x?x?xf32>, index, f32, f32, f32) -> ()
+// CHECK:           return [[DYN_ALLOC]] : memref<3x4x?x?xf32>
+}
```
etiotto, line 2398: Can you pls add a test (or change one of the test cases) where `dtype` is not specified (The ONNX docs mention it is optional).

test/mlir/onnx/onnx_shape_inference.mlir
```diff
@@ -2188,6 +2188,64 @@ func @test_random_normal_static_f64() -> tensor<*xf32> {
...
 }
 
+//===----------------------------------------------------------------------===//
+
+// -----
+
+// Test RandomNormalLike static
+
+func @test_random_normal_like_static_f16(%arg0: tensor<1x1x28x28xf32>) -> tensor<*xf32> {
+  %0 = "onnx.RandomNormalLike"(%arg0) {dtype = 0 : si64, mean = 0.0 :f32, scale = 1.0 : f32, seed = 2.0 : f32} : (tensor<1x1x28x28xf32>) -> tensor<*xf32>
```
etiotto, line 2198: `dtype = 0` means the operator will produce a result tensor with element type `f16`, therefore (maybe I am missing something) the output tensor of this operator should be `tensor<*xf16>` ?

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -4047,7 +4047,21 @@ LogicalResult ONNXRandomNormalOp::inferShapes(
 
 LogicalResult ONNXRandomNormalLikeOp::inferShapes(
     std::function<void(mlir::Region &)> doShapeInference) {
-  return emitError(NOT_IMPLEMENTED_MESSAGE);
+  if (!input().getType().isa<RankedTensorType>())
+    return success();
+  auto outputShape = input().getType().cast<RankedTensorType>().getShape();
+  auto elementTypeID = dtype().getValue();
+
+  RankedTensorType outputTensorType =
```
etiotto, line 4055: The documentation (https://github.com/onnx/onnx/blob/main/docs/Operators.md#RandomNormalLike) says that the attribute `dtype` is optional and that when missing the output tensor element type is the same as the element type of the input tensor. Here it seems we would set it to `f32` in that situation.

How many values can `dtype` have? Is 2 the max value?

etiotto (changes requested):
Can you pls also add the implementation of the verify method for the ONNX operator (unless you want to do it in a separate PR).

test/mlir/onnx/onnx_lowering.mlir
doru1004: The new type comes from the `dtype` attribute.

test/mlir/onnx/onnx_shape_inference.mlir
doru1004: Since this is the user input code it doesn't matter. The type is f16 where it matters which is in the generated code. This test shows that even if the user supplies the wrong type to start with, shape inference will enforce the `dtype` type.

doru1004:
> Can you pls also add the implementation of the verify method for the ONNX operator (unless you want to do it in a separate PR).

What would we be verifying that is not already verified by default?

test/mlir/onnx/onnx_lowering.mlir
etiotto: My understanding is that the `onnx.RandomNormalLike` operator output type is determined by the value of the `dtype` attribute (if present). In this case the test case has: 

````
%0 = "onnx.RandomNormalLike"(%arg0) {dtype = 2 : ....} : (tensor<3x4x5xf32>) -> tensor<*xf32>
````

I submit this code illegal because:
  - dtype = 2 means the output tensor element type will be `f64`
  - the operation return type is `tensor<*xf32>`
  - the declared return type has element type `f32` which not match the value of `dtype` (`f64`)
  
I think it should be:

```
 func @test_random_normal_like2(%arg0: tensor<3x4x5xf32>) -> tensor<*xf64> {
    %0 = "onnx.RandomNormalLike"(%arg0) {dtype = 2 : si64, mean = 0.0 :f32, scale = 1.0 : f32, seed = 2.0 : f32} : (tensor<3x4x5xf32>) -> tensor<*xf64>
    "std.return"(%0) : (tensor<*xf64>) -> ()
 ```

etiotto: The verification code for this operator should check that the declared output type is consistent with the value of `dtype`, and emit an error if there is a mismatch. Would be good actually to add diagnostic tests along those lines...

etiotto (changes requested):

etiotto:
> What would we be verifying that is not already verified by default?

See also inline comments. I believe the following to be illegal because the declared return type `tensor<*xf32>` has element type that does not match the output element type implied by the `dtype` value (`f64`):
```
%0 = "onnx.RandomNormalLike"(%arg0) {dtype = 2 : si64, mean = 0.0 :f32, scale = 1.0 : f32, seed = 2.0 : f32} : (tensor<3x4x5xf32>) -> tensor<*xf32>
```
The verification code for this operator should diagnose the mismatch as an error.

doru1004:
> > What would we be verifying that is not already verified by default?
> 
> See also inline comments. I believe the following to be illegal because the declared return type `tensor<*xf32>` has element type that does not match the output element type implied by the `dtype` value (`f64`):
> 
> ```
> %0 = "onnx.RandomNormalLike"(%arg0) {dtype = 2 : si64, mean = 0.0 :f32, scale = 1.0 : f32, seed = 2.0 : f32} : (tensor<3x4x5xf32>) -> tensor<*xf32>
> ```
> 
> The verification code for this operator should diagnose the mismatch as an error.

This is user written code which may or may not respect this rule. This is not generated code. For the sake of generality the test includes input with matching types and input with mismatching types.

The verifier cannot verify unranked tensors and doesn't actually verify anything until after the shape inference step.

test/mlir/onnx/onnx_lowering.mlir
doru1004: I understand the point you're trying to make but in our case, unranked tensors are treated as wildcard/placeholder tensors, the type inside them is irrelevant. it's like saying tensor<*>. No checks should be performed on unranked tensors because their just placeholders with no binding type information.

doru1004:
> See also inline comments. I believe the following to be illegal because the declared return type `tensor<*xf32>` has element type that does not match the output element type implied by the `dtype` value (`f64`):

This rule can only be enforced shape inference. Before that all bets are off.

etiotto (approved):
LGTM - I did not understand that `tensor<*xf32>` means that not only the rank and extents are not known at compile time but **also** that the element type is not known as well (f32 is just a placeholder for the real type).

etiotto:
Actually I am still unsure about `tensor<*xf32>` should mean that "nothing is known about the tensor, not even the element type". 

According to the documentation in MLIR:  https://mlir.llvm.org/docs/Dialects/Builtin/#unrankedtensortype
```
tensor-type ::= `tensor` `<` `*` `x` type `>`
“An unranked tensor is a type of tensor in which the set of dimensions have unknown rank.”
```
It does not state that the element type is unknown. It *only* says that the tensor has no rank and that the extents of the dimension(s) are not known. 

I do not thing there is a way in MLIR to declare a tensor with unknown element type.

etiotto (commented):
We should have a discussion on the semantics of unranked tensors before merging the PR.

doru1004:
> Actually I am still unsure about `tensor<*xf32>` should mean that "nothing is known about the tensor, not even the element type".
> 
> According to the documentation in MLIR: https://mlir.llvm.org/docs/Dialects/Builtin/#unrankedtensortype
> 
> ```
> tensor-type ::= `tensor` `<` `*` `x` type `>`
> “An unranked tensor is a type of tensor in which the set of dimensions have unknown rank.”
> ```
> 
> It does not state that the element type is unknown. It _only_ says that the tensor has no rank and that the extents of the dimension(s) are not known.
> 
> I do not thing there is a way in MLIR to declare a tensor with unknown element type.

`*xf32` is the default value that the ONNX to ONNX Dialect translator uses as a placeholder for unknown tensors. The test you are referring to is a shape inference test which works on the output of the ONNX to ONNX Dialect translator.

AlexandreEichenberger:
@doru1004 

I don't find the sentence below ambiguous.

> An unranked tensor is a type of tensor in which the set of dimensions have unknown rank.

A tensor type has 2 component: its rank/dimensions and its element type (which may in turn be another ranked type, like for vectors, for example).

It feels that if that operation is given a output type, and that type disagree with the `dtype` parameter, when given, or with the input type, otherwise, we have an error.

I don't know that we have flagged error in the past when the element type of the output is not what we expect it to be, and maybe that is something that we could strengthen. But I would certainly expect the output element type to be of `dtype` when given, and input types otherwise.

AlexandreEichenberger:

etiotto:
After offline discussion I believe we can summarize as follow:
  - When translating a ONNX model to the onnx mlir dialect (our own input dialect) the translator will attempt to generate correct types for each operator inputs and also the operator output type
  - when the translator cannot determine what the output type is it currently uses `tensor<*xf32>`
  - at this stage `tensor<*xf32>` is a placeholder for any tensor type (including tensors with different element type`
  - this is at odds with the MLIR documentation at https://mlir.llvm.org/docs/Dialects/Builtin/#unrankedtensortype
  
The `placeholder` choice of `tensor<*xf32>` does not create major issues because one of the first things `onnx-mlir` does is shape inference which will replace the `placeholder` with the most appropriate type

We can argue whether the current choice for a placeholder type is ideal or not... but, practically, it doesn't make much difference. So I am perfectly fine with this PR as it stands, now that I understand how we got here :-)

doru1004:
> @doru1004
> 
> I don't find the sentence below ambiguous.
> 
> > An unranked tensor is a type of tensor in which the set of dimensions have unknown rank.
> 
> A tensor type has 2 component: its rank/dimensions and its element type (which may in turn be another ranked type, like for vectors, for example).
> 
> It feels that if that operation is given a output type, and that type disagree with the `dtype` parameter, when given, or with the input type, otherwise, we have an error.
> 
> I don't know that we have flagged error in the past when the element type of the output is not what we expect it to be, and maybe that is something that we could strengthen. But I would certainly expect the output element type to be of `dtype` when given, and input types otherwise.

This code is the input to the shape inference and therefore it is code that has not been generated by an MLIR pass, it was generated by the ONNX to ONNX Dialect translator which has its limitations. I have had lengthy conversations with @etiotto about this. He has summarized some of our points in the other post.

doru1004:
> * this is at odds with the MLIR documentation at https://mlir.llvm.org/docs/Dialects/Builtin/#unrankedtensortype

The true solution to this problem is to not care at all about the intermediate types and generate `tensor<*>` but afaik it's not supported.

gongsu832:
@jenkins-droid publish this please

PR #1111, opened 2022-01-20
Pass mtriple string into the MLIR module

etiotto:
MLIR allows for setting both the datalayout string and the triple string in the module (https://mlir.llvm.org/docs/Dialects/LLVM/#data-layout-and-triple). We already set the data layout string. 
This PR sets the triple as well.


caoimhinuibrian (approved):
LGTM

PR #1112, opened 2022-01-21
Reverse some static functions to make them sharable

tungld:
Some functions to be shared among drivers based on onnx-mlir was made static by this PR  https://github.com/onnx/onnx-mlir/pull/1104.
This patch reverses them back so that other drivers can call them.


AlexandreEichenberger (approved):
LGTM

etiotto (commented):
LGTM

PR #1113, opened 2022-01-21
Add -O3 to Numerical tests

AlexandreEichenberger:
Add support to compile at a given optimization level (in addition to which target machine) and force all the numerical tests to compile their graphs at -O3


test/compilerlib/CompilerLibTest.cpp
```diff
@@ -49,13 +57,28 @@ int main(int argc, char *argv[]) {
...
+
   int retVal = 0;
   if (compileFromFile) {
     const char *errorMessage = NULL;
     retVal = omCompileFromFile(testFileName.c_str(), outputBaseName.c_str(),
         onnx_mlir::EmitLib, mcpu.empty() ? nullptr : mcpu.c_str(),
         march.empty() ? nullptr : march.c_str(),
-        mtriple.empty() ? nullptr : mtriple.c_str(), &errorMessage);
+        mtriple.empty() ? nullptr : mtriple.c_str(),
+        hasOptLevel ? &optLevel : nullptr, &errorMessage);
```
etiotto, line 81: [minor]: we could just pass the optLevel here given that the default is `OptLevel::O0`. Doing that would avoid the need to define `hasOptLevel`.

etiotto (approved):
LGTM - one minor comment.

AlexandreEichenberger:
@etiotto, thanks for approving this PR. I am actually changing the interface because 
```
ONNX_MLIR_EXPORT int omCompileFromFile(const char *inputFilename, ...
```

is part of our interface, and I doubt folks will be happy if we change the public interface each time we add an option. And I am sure we may ask for more options in the future.

So I am retooling the interface to have a <key, value> list, so that we may add new keys without changing the interface. Will come in a different PR so reviewers can compare

AlexandreEichenberger:
I suggest that we implement #1114 as it will be easier to extend in the future.

AlexandreEichenberger:
Closed as #1114 is better.

PR #1116, opened 2022-01-24
Update LLVM to commit 676bfb2a226.

etiotto:
Update LLVM to commit 676bfb2a226.

### DIFF
--- a/src/Builder/OpBuildTable.inc.dc
+++ b/src/Builder/OpBuildTable.inc.dc
@@ -1,1 +1,1 @@
-file-same-as-stdout({"file": "src/Builder/OpBuildTable.inc", "cmd": ["python", "utils/gen_onnx_mlir.py", "--dry-run-op-build-table"]})
+file-same-as-stdout({"file": "src/Builder/OpBuildTable.inc", "cmd": ["python3", "utils/gen_onnx_mlir.py", "--dry-run-op-build-table"]})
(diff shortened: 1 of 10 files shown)


tungld (approved):
LGTM! Thanks!

The change of `python` to `python3` seems not related to updating LLVM. Please feel free to undo the change to keep the PR clean.

AlexandreEichenberger (approved):
LGTM

src/Builder/OpBuildTable.inc.dc
```diff
@@ -1 +1 @@
-file-same-as-stdout({"file": "src/Builder/OpBuildTable.inc", "cmd": ["python", "utils/gen_onnx_mlir.py", "--dry-run-op-build-table"]})
+file-same-as-stdout({"file": "src/Builder/OpBuildTable.inc", "cmd": ["python3", "utils/gen_onnx_mlir.py", "--dry-run-op-build-table"]})
```
sstamenova, line 1: This change and the equivalent change to the other `.dc` file breaks people/builds that don't have python 3 setup under `python3`. At this point we can expect that python 3 is the default version installed and that `python` points to python 3 and not python 2, so it is better to not prefer to `python3` as an environment variable.

gongsu832: > At this point we can expect that python 3 is the default version installed and that `python` points to python 3 and not python 2

This is not necessarily true. There are systems, notably MacOS, where the default python is still python2 because the OS has leftover baggages still depending on python2. And on most Linux distros, `/usr/bin/python` is "reserved" for python2, i.e., even if you install python3, `/usr/bin/python` will not be created. Instead, only `/usr/bin/python3` is created (of course you can create the python symlink to python3 yourself). So between the two failure scenarios,

1. Use python3 and fail to find it because python3 was not installed
2. Use python and (erroneously) find python2 and fail later when some python3 script is run

I would say #1 is actually an easier failure to deal with.

PR #1118, opened 2022-01-25
Remove BinaryDecoder

tungld:
BinaryDecoder was used to decode constant values when we stored constants in files.
Since we have moved to embedding constants in MLIR IR, this decoder is not used anymore.


AlexandreEichenberger (approved):
LGTM

PR #1119, opened 2022-01-25
changes to make cppcheck happy

AlexandreEichenberger:

cppcheck report errors in jniinit.c as some functions are called by jnidummy with env pointer being NULL. This scenario should never happen, nevertheless adding an `assert(env)` is not a bad thing and resolve this warning.

No other errors were reported (spurious warning about LLVM_DEBUG macro not being defined... code there is just to print debug, so that should be fine).


caoimhinuibrian (approved):
LGTM

PR #1121, opened 2022-01-25
Revert gen onnx mlir script

doru1004:
Fix gen onnx mlir script which was messed up by a previous auto format.


utils/gen_onnx_mlir.py
```diff
@@ -35,229 +35,228 @@
...
+
 # check the version of onnx package being used
-if current_onnx_version != onnx.__version__:
-    print("version of expected onnx is {}, ".format(current_onnx_version) +
+if current_onnx_version != onnx.__version__ :
+    print("version of expected onnx is {}, ".format(current_onnx_version)+
           "while onnx package being used is {}".format(onnx.__version__))
     quit()
 
 current_onnx_version = "1.9.0"
```
sstamenova, line 53: This version string is in 3 separate places, is that normal? It seems wrong.

utils/gen_onnx_mlir.py
```diff
@@ -35,229 +35,228 @@
...
 # check the version of onnx package being used
-if current_onnx_version != onnx.__version__:
-    print("version of expected onnx is {}, ".format(current_onnx_version) +
+if current_onnx_version != onnx.__version__ :
```
sstamenova, line 55: This check is also here 2? 3? times. This seems like another merge issue.

utils/gen_onnx_mlir.py
```diff
@@ -809,48 +805,44 @@ def get_type_inference_func(s, indent, type_inference_code):
...
+        'uint64' : 'UI64',
+        'int8' : 'I8',
+        'int16' : 'I16',
+        'int32' : 'I32',
+        'int64' : 'I64',
+        'float16' : 'F16',
+        'bfloat16' : 'BF16',
+        'float' : 'F32',
+        'double' : 'F64',
+        'unkown' : 'BF16',
```
sstamenova, line 833: Nit: unknown

doru1004:
@sstamenova I am not changing the script in any way, just reverting it since my previous commit accidentally changed the formatting. The fixing of the problems in the script should be done in a separate patch.

AlexandreEichenberger (approved):
LGTM, I don't mind if you want to revert the formatting. However, I would not mind to clangFormat python scripts too, even if originally due to an accident.

sstamenova:
> @sstamenova I am not changing the script in any way, just reverting it since my previous commit accidentally changed the formatting. The fixing of the problems in the script should be done in a separate patch.

Thanks for changing the title, it's clearer now.

doru1004:

PR #1122, opened 2022-01-26
Print model loading time in RunONNXModel.py

tungld:
This patch splits the running time in RunONNXModel.py into loading time (time to load a compiled model) and inference time, which is more accurate.


etiotto (approved):
LGTM

AlexandreEichenberger (approved):
LGTM, have folks complained about load time?

tungld:
> have folks complained about load time?

No, so far. But it is me running a small test and would like to know the inference time only.

Issue #1117, opened 2022-01-25
SIMD for MatMul when M=1


tungld:
I have the following MatMul of NxK and KxM where N and K are large and M is 1:
```mlir
func private @test_matmul1(%arg0 : tensor<100x200xf32>, %arg1 : tensor<200x1xf32>) -> tensor<*xf32> {
   %0 ="onnx.MatMul"(%arg0, %arg1) : (tensor<100x200xf32>, tensor<200x1xf32>) -> tensor<*xf32>
   "std.return"(%0) : (tensor<*xf32>) -> ()
}
```

When lowering it to affine by using `onnx-mlir-opt --shape-inference --convert-onnx-to-krnl --convert-krnl-to-affine --canonicalize`, we get the follows:
```mlir
#set = affine_set<(d0) : (1 >= 0, -d0 - 7 >= 0, 1 >= 0)>
affine.for %arg2 = 0 to 100 step 4 {
   affine.for %arg3 = 0 to 1 step 8 {
      affine.for %arg4 = 0 to 200 step 8 {
         affine.if #set(%arg3) {
             // SIMD code
          } else {
             // None-SIMD code
          }
       }
    }
}
````

Looking at `%arg3` and `#set(%arg3)`, it seemed that the condition in `affine.if` would return false and SIMD code would be not used.

@AlexandreEichenberger any thoughts on this?

AlexandreEichenberger:
The affine condition does not know that `d0` can only have the value `0`. I can add code to ensure that if the M dimensions is known to be smaller than the SIMD with, to disable simd for that operation.

tungld:
Can we simdize MatMul in this case? Given that N and K are large, computation is slow.

AlexandreEichenberger:
True, there may be a special case for J = 1 dim. Let me think about it some more. The #1123 PR simply disable SIMD.

@tungld do you have an idea of how frequent this situation is?

AlexandreEichenberger:
Keeping it open as we really want to generate simd code for that.

AlexandreEichenberger:
now done, may need to further tune the code but closing the issue

PR #1123, opened 2022-01-26
Matmul narrow no simd

AlexandreEichenberger:
Disable SIMD when the dimension is known at compile time and is smaller than the simd width.

This PR responds to issue #1117


AlexandreEichenberger:
Also removed #def for debug and use LLVM_DEBUG now

tungld (approved):
LGTM! Thanks for the change!

PR #1130, opened 2022-01-27
Category mapper: codegen

etiotto:
Finish the code generation for the `CategoryMapper` operator.  A end-to-end test for the operator is in the separate PR https://github.com/onnx/onnx-mlir/pull/1142.


etiotto:
Reviewers in case the PR is too big I can split it into 2 PRss if it is desired. Just let me know.

tungld:
I am quite concerned about compilation performance when we unfold a tensor of strings into individual strings and emit code to get the address of each string, given that the number of categories is large. Have you tried to compile a practical model, e.g. BIDAF to see if compilation time is acceptable?

Still not so clear about why we need a new test infrastructure. It seems that the listed scenarios are what the numerical tests were originally designed for.

etiotto:
> I am quite concerned about compilation performance when we unfold a tensor of strings into individual strings and emit code to get the address of each string, given that the number of categories is large. Have you tried to compile a practical model, e.g. BIDAF to see if compilation time is acceptable?

In MLIR the way to initialize a global constant with strings is by using an initializer (https://mlir.llvm.org/docs/Dialects/LLVM/#llvmmlirglobal-mlirllvmglobalop). So for example given:
```
"krnl.global"() {name = "cats_strings", shape = [3], value = dense<["cat", "human"]> : tensor<2x!krnl.string>} : () -> memref<2x!krnl.string>
```
We need to generate an initializer region like this:
```
module attributes {}  {
 llvm.mlir.global private constant @cat("cat") : !llvm.array<3 x i8>
 llvm.mlir.global private constant @human("human") : !llvm.array<5 x i8>
 llvm.mlir.global internal constant @cats_strings() : !llvm.array<2 x !llvm.ptr<i8>> {
   %0 = llvm.mlir.addressof @cat : !llvm.ptr<!llvm.array<3 x i8>>
   %1 = llvm.mlir.addressof @human : !llvm.ptr<!llvm.array<5 x i8>>
   %c0 = llvm.mlir.constant(0 : i32) : i32
   %pcat = llvm.getelementptr %0[%c0, %c0] : (!llvm.ptr<!llvm.array<3 x i8>>, i32, i32) -> !llvm.ptr<i8>
   %phuman = llvm.getelementptr %1[%c0, %c0] : (!llvm.ptr<!llvm.array<5 x i8>>, i32, i32) -> !llvm.ptr<i8>
   %2 = llvm.mlir.undef : !llvm.array<2 x !llvm.ptr<i8>>
   %3 = llvm.insertvalue %pcat, %2[0] : !llvm.array<2 x !llvm.ptr<i8>>
   %4 = llvm.insertvalue %phuman, %3[1] : !llvm.array<2 x !llvm.ptr<i8>>
   llvm.return %4: !llvm.array<2 x !llvm.ptr<i8>>
 }
}
```
The LLVM IR generated for this MLIR code is:
```
; ModuleID = '/Users/etiotto@ca.ibm.com/tmp/good_constant.bc'
source_filename = "LLVMDialectModule"
target datalayout = "e-m:o-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
target triple = "x86_64-apple-darwin20.6.0"

@cat = private constant [3 x i8] c"cat"
@human = private constant [5 x i8] c"human"
@cats_strings = internal constant [2 x i8*] [i8* getelementptr inbounds ([3 x i8], [3 x i8]* @cat, i32 0, i32 0), i8* getelementptr inbounds ([5 x i8], [5 x i8]* @human, i32 0, i32 0)]
```
which is the standard way to initialize a constant with string literals in LLVM. That is: we are not worse than the LLVM compiler itself.  Note also that I have code to avoid regenerating a string literal redundantly (within the same module). So there should be no duplication of identical strings.  The runtime impact of this code should be minimal, because the strings are initialized statically. 

As for compile time, I do not see any other way to generate the code we need. Of course suggestions are welcome!

etiotto:
> Still not so clear about why we need a new test infrastructure. It seems that the listed scenarios are what the numerical tests were originally designed for.

Not quite, the new tests do not use the 'rapidcheck' infrastructure at all.  The numerical tests take quite long to run, given that they test the same code with different inputs to ascertain the numerical accuracy  of the tests.  That said, would you prefer that we keep the new C++ based tests in the same directory (or a subdirectory) of the numerical tests ? I think we should have a separate make target because (a) the new tests do not check for numerical accuracy and (b) the new tests run a lot faster.

The new tests encapsulate a number of common functionality in the `TestHelper` class. I can see how we could restructure the numerical tests to use the `TestHelper` class, however I did not want to do that in this PR before you guys have a chance to give feedback.  It is certainly possible to common things up in a subsequent PR.

etiotto:
> It's helpful if you could show some rough numbers about compile time.

Timed the `compileModule` function while compiling a test containing a `CategoryMapper` operation where the number of strings in the attribute `cats_strings` is shown by `array_size` below: 

```
array size = 1
duration = 1.557470e+05ms

array size = 10,000
duration = 2.350200e+05ms

array size = 100,000
duration = 8.128280e+05ms

array size = 1,000,000
duration = 7.376579e+06ms
```

Note: timing was done on my Mac, without binding the executable to a processor (machine was quite). Just to give a rough idea of the impact of increasing number of strings...

tungld:
> array size = 1
> duration = 1.557470e+05ms
> 
> array size = 10,000
> duration = 2.350200e+05ms
> 
> array size = 100,000
> duration = 8.128280e+05ms
> 
> array size = 1,000,000
> duration = 7.376579e+06ms

Thank you for the information! Does `ms` stand for `milliseconds` or `microseconds`?

AlexandreEichenberger:
> Thank you for the information! Does ms stand for milliseconds or microseconds?

Typically ms is mili, us is micro. So @etiotto the compilation time varies from 15s to 736s (12 min) ?

etiotto:
Decided to separate the 'extra' test infrastructure in a separate pull request.

tungld (approved):
LGTM!

tungld:
> % time   ./Debug/bin/TestCategoryMapper
> array size = 1000000
> duration = 7.880116e+06us
> ./Debug/bin/TestCategoryMapper  7.99s user 0.74s system 97% cpu 8.989 total

This looks good. Thanks! I am sorry for many requests.

PR #1138, opened 2022-01-29
Benchmarking: first step, reusing code in numerical for building/compiling models

AlexandreEichenberger:
Currently, each numerical benchmark create and compile its own model. The same models will be key to get performance measurements. In order to reuse the code, this PR is pulling the "create and compile" of the numerical in its own library, which will then be used both by the numerical and the future benchmarking (for time) suite.

I am new to CMake engineering, please let me know if there is something that I do that is not recommended.


test/modellib/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+# SPDX-License-Identifier: Apache-2.0
+
+add_onnx_mlir_library(ModelLib
```
sstamenova, line 3: This should be excluded from the list of all libs, by default add_onnx_mlir_library will add this to the OMLibs

test/numerical/CMakeLists.txt
```diff
@@ -57,7 +57,7 @@ set(TEST_LINK_LIBS rapidcheck CompilerUtils ExecutionSession)
 
 add_numerical_unittest(TestConv
   TestConv.cpp
-  LINK_LIBS PRIVATE ${TEST_LINK_LIBS}
+  LINK_LIBS PRIVATE ${TEST_LINK_LIBS} ModelLib
```
sstamenova, line 60: Since this line is getting long, I'd put things on a new line

sstamenova: Also, if ModelLib will be used by all of the tests, you can add it to TEST_LINK_LIBS instead

test/modellib/CMakeLists.txt
AlexandreEichenberger: I think this is done by adding this line
```
  EXCLUDE_FROM_OM_LIBS
```

first time ever editing/creating cmake files, started with a tutorial this morning :-)
Let me know if that is not what you had in mind

test/numerical/CMakeLists.txt
AlexandreEichenberger: sure

gongsu832:
The PR only did "lib-conversion" for TestConv. I assume you will do the same for the other tests such as TestGemm, etc. as well. Is that the idea?

AlexandreEichenberger:
Yes, started to push early to get feedback as I am a novice CMake person... Just added Gemm and Matmul, will now do the RNNs

src/Compiler/CompilerUtils.hpp
```diff
@@ -42,9 +42,10 @@ void setTargetArch(const std::string &arch);
 void setTargetTriple(const std::string &triple);
 void setOptLevel(const onnx_mlir::OptLevel level);
 // Set compile context according to the (key, value) fields passed.
-void setCompileContext(mlir::MLIRContext &context,
-    const llvm::SmallVector<std::pair<onnx_mlir::OptionKind, std::string>, 4>
-        options);
+using CompilerOptionList =
+    llvm::SmallVector<std::pair<onnx_mlir::OptionKind, std::string>, 4>;
+void setCompileContext(
```
etiotto, line 47: nit: I would put this using declaration at the top of the header to give it more "visibility".

test/modellib/ConvModel.cpp
```diff
@@ -0,0 +1,145 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//==============-- ConvModel.cpp - Building Conv Models for tests -===========//
+//
+// Copyright 2022-2022 The IBM Research Authors.
```
etiotto, line 7: 2022 is sufficient

test/modellib/ConvModel.cpp
```diff
@@ -0,0 +1,145 @@
...
+//
+// =============================================================================
+//
+// This file contains a function that build a convolution model and compiles it.
```
etiotto, line 11: typo: builds

test/modellib/MatMulModel.cpp
```diff
@@ -0,0 +1,84 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//==============-- MatMulModel.cpp - Building Conv Models for tests -=========//
+//
+// Copyright 2022-2022 The IBM Research Authors.
```
etiotto, line 7: 2022 only

test/modellib/MatMulModel.cpp
```diff
@@ -0,0 +1,84 @@
...
+//
+// =============================================================================
+//
+// This file contains a function that build a MatMul model and compiles it.
```
etiotto, line 11: typo: build -> builds

test/modellib/ModelLib.hpp
```diff
@@ -0,0 +1,59 @@
...
+//===----------------------------------------------------------------------===//
+
+#include <string>
+
+#include "llvm/ADT/SmallVector.h"
+
+#include "src/Compiler/CompilerUtils.hpp"
+
+// Padding schemes
+#define AUTO_PAD_NOTSET 0
```
etiotto, line 23: [Suggestion]: Use `enum class` to replace the #defines
```
enum class PaddingScheme {
  AutoPadNotSet,
  AutoPadValid, 
  etc...
};

File: test/modellib/ModelLib.hpp [hunk 9]
```diff
@@ -0,0 +1,59 @@
...
+#define AUTO_PAD_VALID 1
+#define AUTO_PAD_LOWER 2
+#define AUTO_PAD_UPPER 3
+#define AUTO_PAD_UB 4
+const std::string getAutoPadName(const int autoPad);
+
+// Conv2D
+bool genConv2DModelAndCompile(
```
etiotto, line 31: A couple of suggestions:
  - to avoid 'polluting' the global namespace I suggest to use a C++ namespace 
  - consider if it make sense to place these functions in a C++ class

test/modellib/ModelLib.hpp
```diff
@@ -0,0 +1,59 @@
...
+    const int kW, const int autoPad, const int pHBegin, const int pHEnd,
+    const int pWBegin, const int pWEnd, const int stride, const int dilation,
+    const int isDynamic,
+    /* conv param out */
+    int &NOut, int &COut, int &HOut, int &WOut);
+
+// GEMM
+bool genGemmAndCompileModel(
+    /* compile option */
+    const std::string modelName, const CompilerOptionList &options,
```
etiotto, line 45: modelName is immutable in this function (const), pass it by reference?

test/modellib/ModelLib.hpp
```diff
@@ -0,0 +1,59 @@
...
+    /* GEMM param in*/
+    const int I, const int J, const int K, const int aTrans, const int bTrans,
+    const int cRank, const float alphaVal, const float betaVal,
+    /* GEMM param out*/
+    llvm::SmallVector<int64_t, 2> &aShape,
```
etiotto, line 50: [Suggestion]: `llvm::SmallVector<int64_t, 2> &` -->  `llvm::SmallVectorImpl<int64_t> &`

I recall that in LLVM using SmallVectorImpl is encouraged (so that the caller can pass any SmallVector 'size'). Unless this function must receive only size == 2, in which case declaring the parameter as it is now enforces that.

test/modellib/ModelLib.hpp
```diff
@@ -0,0 +1,59 @@
...
+    llvm::SmallVector<int64_t, 2> &bShape,
+    llvm::SmallVector<int64_t, 2> &cShape);
+
+// MatMul
+bool genMatMul2DModelAndCompile(
+    /* compile option */
+    const std::string modelName, const CompilerOptionList &options,
```
etiotto, line 57: [Suggestion]:  const std::string -> `const std::string &`

test/numerical/TestGemm.cpp
```diff
@@ -10,13 +10,12 @@
...
-#include "mlir/IR/BuiltinOps.h"
 #include "llvm/Support/FileSystem.h"
 
 #include "src/Compiler/CompilerUtils.hpp"
-#include "src/Dialect/ONNX/ONNXOps.hpp"
 #include "src/Runtime/ExecutionSession.hpp"
 #include "src/Runtime/OMTensorHelper.h"
+#include "test/modellib/ModelLib.hpp"
 
 #define SHARED_LIB_BASE string("./TestGemm_main_graph")
```
etiotto, line 20: [Suggestion]:  
``` 
`#define SHARED_LIB_BASE string("./TestGemm_main_graph")` --> static const llvm::StringRef("./TestGemm_main_graph");`

**Review by etiotto — COMMENTED** [review]
I left some inline comments, mostly stype related. 

PR https://github.com/onnx/onnx-mlir/pull/1130 is, partially, related to this work in the sense that in that PR I have tried to factor out some common functionality used to build a test, compile it, and execute it in a separate class `TestHelper`.  Perhaps we can take some ideas and create a common test infrastructure that serves the need of both the numerical tests and other C++ based tests we might need to create. What do you think @AlexandreEichenberger ?

File: test/modellib/CMakeLists.txt (hunk 1 above)
Comment (sstamenova) [reply test/modellib/CMakeLists.txt:3]: Yes, `EXCLUDE_FROM_OM_LIBS` is the correct option.

File: test/numerical/CMakeLists.txt [hunk 14]
```diff
@@ -53,11 +53,11 @@ endfunction()
 # of the flags for exceptions (among others) are not set correctly.
 llvm_update_compile_flags(rapidcheck)
 
-set(TEST_LINK_LIBS rapidcheck CompilerUtils ExecutionSession)
+set(TEST_LINK_LIBS rapidcheck CompilerUtils ExecutionSession ModelLib)
 
 add_numerical_unittest(TestConv
   TestConv.cpp
-  LINK_LIBS PRIVATE ${TEST_LINK_LIBS}
+  LINK_LIBS PRIVATE ${TEST_LINK_LIBS}
```
sstamenova, line 60: Nit: Remove the extra whitespace that was added

test/numerical/CMakeLists.txt
```diff
@@ -53,11 +53,11 @@ endfunction()
 # of the flags for exceptions (among others) are not set correctly.
 llvm_update_compile_flags(rapidcheck)
 
-set(TEST_LINK_LIBS rapidcheck CompilerUtils ExecutionSession)
+set(TEST_LINK_LIBS rapidcheck CompilerUtils ExecutionSession ModelLib)
```
sstamenova, line 56: Since ModelLib requires ExecutionSession and CompilerUtils, you can drop them from the list.

test/numerical/TestGemm.cpp
sstamenova: This could even be computed based on the file name. I considered doing that when I made the tests work for Windows, but I decided against it at the time to keep the change as small as possible.

test/modellib/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+  ConvModel.cpp
+  GemmModel.cpp
+  GRUModel.cpp
+  MatMulModel.cpp
+  ModelLib.hpp
+  ModelHelper.cpp
+  ModelHelper.hpp
```
sstamenova, line 10: Don't include header files in the list of sources - they should be included in the source files instead.

src/Compiler/CompilerUtils.hpp
AlexandreEichenberger: make sense

src/Compiler/CompilerUtils.hpp
```diff
@@ -42,9 +42,10 @@ void setTargetArch(const std::string &arch);
...
+    mlir::MLIRContext &context, const CompilerOptionList &options);
 // Set compile context, legacy C array and string representation.
 void setCompileContext(mlir::MLIRContext &context,
```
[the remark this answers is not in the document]
AlexandreEichenberger: If you are referring to this call
```
void setCompileContext(mlir::MLIRContext &context,
    const onnx_mlir::OptionKind *key, const char **val, const int64_t num);
```

it is needed because we want to support a C interface as well.

test/modellib/ConvModel.cpp
AlexandreEichenberger: ok

test/modellib/ConvModel.cpp
AlexandreEichenberger: thanks

test/modellib/MatMulModel.cpp
AlexandreEichenberger: changed everywhere, tx

test/modellib/ModelLib.hpp
AlexandreEichenberger: So you would like to add a name space for the lib? I though if you don't need this (and almost no-one needs this), they just don't include the header/lib in their code.

A class does not make sense, but if you still think that a name space is needed, then what names do you think would work?

test/modellib/ModelLib.hpp
AlexandreEichenberger: make sense, done everywhere

test/modellib/ModelLib.hpp
AlexandreEichenberger: I agree with you in principle, but in practice, that prevent us to then initialize the arrays in a neat way.
```
/workdir/onnx-mlir/test/modellib/RNNModel.cpp:54:20: error: no match for 'operator=' (operand types are 'llvm::SmallVectorImpl<long int>' and '<brace-enclosed initializer list>')
   54 |   xShape = {S, B, I};
```

I prefer preserving the nice init properties.

test/modellib/ModelLib.hpp
AlexandreEichenberger: thanks, done everywhere

test/numerical/TestGemm.cpp
AlexandreEichenberger: change the code to use the static const construct

test/numerical/CMakeLists.txt
AlexandreEichenberger: sure, did it, but is there any issues with having redundant dependences, to make is super clear?

test/modellib/CMakeLists.txt
AlexandreEichenberger: I assume from your comment that there should just be removed from the list.

test/numerical/CMakeLists.txt
AlexandreEichenberger: Had to reintroduce it because it was causing errors:

```
In file included from /workdir/onnx-mlir/test/numerical/TestMatMul2D.cpp:15:
/workdir/onnx-mlir/src/Compiler/CompilerUtils.hpp:29:10: fatal error: onnx-mlir/Compiler/OMCompilerTypes.h: No such file or directory
   29 | #include "onnx-mlir/Compiler/OMCompilerTypes.h"
```

src/Compiler/CompilerUtils.hpp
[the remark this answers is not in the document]
etiotto: ok good.

test/modellib/GRUModel.cpp
```diff
@@ -0,0 +1,140 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//==============-- GRUModel.cpp - Building GRU Models for tests
```
etiotto, line 5: Line too long

test/modellib/GemmModel.cpp
```diff
@@ -0,0 +1,107 @@
...
+      /*numInputs=*/3,
+      /*numOutputs=*/1,
+      /*signature*/ signature);
+  module.push_back(entryPoint);
+
+  OwningModuleRef moduleRef(module);
+
+  compileModule(moduleRef, ctx, modelName, onnx_mlir::EmitLib);
+  return true;
+}
```
etiotto, line 107: add newline

test/modellib/MatMulModel.cpp
```diff
@@ -0,0 +1,83 @@
...
+      /*numInputs=*/2,
+      /*numOutputs=*/1,
+      /*signature*/ signature);
+  module.push_back(entryPoint);
+
+  OwningModuleRef moduleRef(module);
+  compileModule(moduleRef, ctx, modelName, onnx_mlir::EmitLib);
+
+  return true;
+}
```
etiotto, line 83: add newline

test/modellib/ModelLib.hpp
etiotto: OK I see. I am fine with that.

test/modellib/RNNModel.cpp
```diff
@@ -0,0 +1,135 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//==============-- RNNModel.cpp - Building RNN Models for tests -======+======//
```
etiotto, line 5: nit: extra '+' in comment

test/modellib/ModelLib.hpp
etiotto: Understand, I am fine without namespace.

etiotto (approved):
LGTM, @sstamenova are you ok too?
I just put some minor comment that can be addressed when merging.

test/numerical/CMakeLists.txt
sstamenova: There's generally no issue with redundant dependencies -- but by avoiding it, we make sure that going forward we don't end up including unneeded libs as dependencies which slow down builds and increase the size of the produced binaries.

It looks like you marked `CompilerUtils` and `ExecutionSession` as `private` libs for the linking of `ModelLib`. It would be better if you marked them as `public` instead and removed them form here.

sstamenova:
> LGTM, @sstamenova are you ok too? I just put some minor comment that can be addressed when merging.

Yes, I would still prefer to avoid the duplicate dependencies, but it's not a big issue.

test/modellib/GRUModel.cpp
AlexandreEichenberger: addressed

test/modellib/MatMulModel.cpp
AlexandreEichenberger: add a new line at the end of the file, if that is what you asked.

test/modellib/RNNModel.cpp
AlexandreEichenberger: Changed to =

AlexandreEichenberger:
>Yes, I would still prefer to avoid the duplicate dependencies, but it's not a big issue.

I tried but it lead to a linking failure. Maybe a better cmake person can give it a shot. I added a todo in the file.

test/modellib/MatMulModel.cpp
etiotto: yep

PR #1140, opened 2022-01-31
Split inferShapes and added verify to Flatten operator

adriansion:
Changes to ONNXOps.cpp resulted in error-free build, though warnings that the verify functions were unused did show during the build.

I wrote a lit test for Flatten but was not fully sure about my correctness, so I have not written the three needed tests for Reshape.


test/mlir/onnx/invalid.mlir
```diff
@@ -40,3 +40,13 @@ func @test_concat_verifier_3(%arg0 : tensor<5x5x1x32xf32>, %arg1 : tensor<5x5x3x
   %1 = "onnx.Concat"(%arg0, %arg1, %arg2) { axis = 1 : si64} : (tensor<5x5x1x32xf32>, tensor<5x5x3x32xf32>, tensor<5x5x5x32xf32>)  -> tensor<*xf32>
   "std.return"(%1) : (tensor<*xf32>) -> ()
 }
+
+// -----
+
+func @test_flatten_verifier_1(%arg0 : tensor<5x5x1x32xf32>) -> tensor<*xf32> {
+  // expected-error @+1 {{Flatten axis value out of range}}
+  %1 = "onnx.Flatten"(%arg0) { axis = 4 : si64} : (tensor<5x5x1x32xf32>) -> tensor<*xf32>
+  "std.return"(%1) : tensor<*xf32>
```
etiotto, line 49: I think should be `"std.return"(%1) : (tensor<*xf32>) -> ()`

AlexandreEichenberger:
Quick ones. You need to complete the DCO certification, did you sign the license agreement? Please state so, and if there is still a DCO issue, then it is because the `-s` flag was omitted in the commit. 

Fo the clang format, you need to format the code. Check that you have the `clang-format` program. You can run it on the files individually, or use a VSCode extension.

I believe the tests fails on this
```
FAIL: Open Neural Network Frontend :: krnl/reshape.mlir (34 of 61)
******************** TEST 'Open Neural Network Frontend :: krnl/reshape.mlir' FAILED ********************
Script:
--
: 'RUN: at line 1';   /workdir/onnx-mlir/build/Debug/bin/onnx-mlir-opt -O3 --shape-inference --convert-onnx-to-krnl --convert-krnl-to-affine --convert-krnl-to-llvm /workdir/onnx-mlir/test/mlir/krnl/reshape.mlir -split-input-file | /workdir/llvm-project/build/bin/FileCheck /workdir/onnx-mlir/test/mlir/krnl/reshape.mlir
--
```

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -1554,31 +1554,38 @@ LogicalResult ONNXBatchNormalizationInferenceModeOp::inferShapes(
 // Reshape
 //===----------------------------------------------------------------------===//
 
-LogicalResult ONNXReshapeOp::inferShapes(
-    std::function<void(mlir::Region &)> doShapeInference) {
+static LogicalResult verify(ONNXReshapeOp op) {
   // Cannot infer shape if no shape tensor is specified.
-  if (!data().getType().isa<RankedTensorType>())
+  if (!op.data().getType().isa<RankedTensorType>())
```
AlexandreEichenberger, line 1559: This call is a bit more precise.

`if (! hasShapeAndRank(op.data())`

please convert elsewhere in the verifiers that you added

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -1554,31 +1554,38 @@ LogicalResult ONNXBatchNormalizationInferenceModeOp::inferShapes(
...
   if (failed(shapeHelper.computeShape(operandAdaptor)))
-    return emitError("Failed to scan Reshape parameters successfully");
+    return op.emitError("Failed to scan Reshape parameters successfully");
+
+  return success();
+}
+
+LogicalResult ONNXReshapeOp::inferShapes(
+    std::function<void(mlir::Region &)> doShapeInference) {
+
```
AlexandreEichenberger, line 1585: you still have to keep the original code below... which was removed. Check a prior PR to see the code that was changed if you are not sure.
 
```
if (!op.shape().getType().isa<RankedTensorType>())
     return success();
```

AlexandreEichenberger (commented):
Commented on the first ONNX Op, comments apply to the second one as well.

src/Dialect/ONNX/ONNXOps.cpp
adriansion: Done

src/Dialect/ONNX/ONNXOps.cpp
adriansion: Done

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2835,6 +2835,24 @@ LogicalResult ONNXSplitV11Op::inferShapes(
 // Flatten
 //===----------------------------------------------------------------------===//
 
+static LogicalResult verify(ONNXFlattenOp op) {
+
+  auto inTy = op.input().getType().dyn_cast<RankedTensorType>();
```
AlexandreEichenberger, line 2840: That will not work when we use memref. In verify, we have to accommodate them as well. That is why we test with 
```
  if (!hasShapeAndRank(op.input())) {
```

then, if you need the type, you should use
```
  auto inTy = op.input().getType().dyn_cast<ShapedTensorType>();
```

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2835,6 +2835,24 @@ LogicalResult ONNXSplitV11Op::inferShapes(
...
+  if (axisValue < -1 * inputRank || axisValue > inputRank) {
+    return op.emitError("ONNXFlattenOP: axis() value is out of range");
+  }
+
+  return success();
+}
+
 LogicalResult ONNXFlattenOp::inferShapes(
     std::function<void(mlir::Region &)> doShapeInference) {
   auto inTy = input().getType().dyn_cast<RankedTensorType>();
```
AlexandreEichenberger, line 2858: Here, auto inTy = input().getType().dyn_cast<RankedTensorType>(); could fail. So you can either use the 'ISA' trick and then you are guaranteed to have this work (when ISA fails, return success)
Or you can use a "dyn_cast_or_null" and test if the result is not null.

AlexandreEichenberger (commented):
See that you removed the reshape. If you follow my recommendations that I sent you via slack, it should work too. I can publish my branch if that helps.

AlexandreEichenberger (approved):
LGTM, thanks on your first contribution

PR #1141, opened 2022-01-31
Generate dialect and type for Dialect/ONNX with Tablegen

chentong319:
Previously, Tablegen was used to generate only the operations for onnx. We directly defined the classes for dialect and types for onnx dialect.
This PR utilizes the Tablegen for dialect and types and deleted the explicitly defined classes.
Some details:
1. The dialect class is changed from ONNXOpsDialect to ONNXDialect, which matches the directory `Dialect/ONNX`.
2. I did not put ONNX dialect into a separate namespace, which I think we should. 
3. To avoid the type name conflict with the type in Krnl, I change StringType in onnx into ONNXStringType.
4. Use the hooks provided by Tablegen for parsing/print and initialization.


MLIR.cmake
```diff
@@ -77,6 +77,17 @@ function(add_onnx_mlir_dialect dialect)
   add_public_tablegen_target(OM${dialect}IncGen)
 endfunction()
 
+function(add_onnx_mlir_dialect_new tdfile dialect_namespace)
```
sstamenova, line 80: Can we have a better name for this function? `_new` is not very descriptive.

I would recommend instead expanding the functionality of the original function to optionally take in the namespace and if the namespace exists, invoking the additional functionality.

sstamenova (changes requested):

MLIR.cmake
chentong319: The namespace here means the name before the types or ops  in .mlir file, such as !onnx.StringType and onnx.Add. It is required for generating classes for Dialect and Types.
This function can replace the existing one after we change all the usages (for Krnl). How about change the old function into add_onnx_mlir_dialect_ops_only?

sstamenova: My suggestion is to have a single function that optionally sets up the namespace if provided. That way the functionality will not change for any existing usage, but new usage that provides the namespace will make the additional invocations.

tungld:
This would be great. It's more aligned with dialects in MLIR.

We still use ONNXOpsDialect, ONNXOpsTypes, how about changing to ONNXDialect, ONNXType, and ONNXOps for Dialect, Type and Ops, respectively? In the future, we may have ONNXAttribute for attributes.

MLIR.cmake
AlexandreEichenberger: > This function can replace the existing one after we change all the usages (for Krnl).

If it is really only temporary, then its ok. You may just add a TODO chentong near it, or call it `add_onnx_mlir_dialect_temporary_name`

AlexandreEichenberger:
> We still use ONNXOpsDialect, ONNXOpsTypes, how about changing to ONNXDialect, ONNXType, and ONNXOps for Dialect, Type and Ops, respectively? In the future, we may have ONNXAttribute for attributes.

AlexandreEichenberger (commented):
@chentong319 Can you add a comment like the list you had on what this PR does (which is type/op tablegen), and then add what subsequent PR will do? That way, we have a roadmap for the future PRs and that way we can see this PR in the scope of the entire effort. Looks like we are all excited about the changes that you are about to push.

src/Dialect/ONNX/ONNXOps.hpp
```diff
@@ -24,73 +24,20 @@
...
-  static StringRef getDialectNamespace() { return "onnx"; }
-};
 } // end namespace mlir
 
 /// Include the auto-generated header file containing the declarations of the
 /// ONNX operations.
 #define GET_OP_CLASSES
 #include "src/Dialect/ONNX/ONNXOps.hpp.inc"
-
-namespace mlir {
```
etiotto, line 58: I think, for consistency with other MLIR dialects (LLVM dialect, SPRIV dialect, ...) the ONNX MLIR types should be in a header file (ONNXTypes.h).  I see that in those dialect they naming convention is to nest types in the dialect namespace, so alternatives are:
```
mlir::onnxmlir::ONNXStringType, or
mlir::onnx::ONNXStringType  <--- prob. more consistent
```

etiotto (commented):
@chentong319 thanks for spearheading this restructuring of the code! 

I am assuming that we want to use naming conventions as close as possible to other MLIR dialects (for consistency and to prepare for possible inclusion in the MLIR codebase). 

With that in mind, here is a summary of suggestions I put inline: 
  - suggest that we declare custom ONNX types in a header file rather that in a .td file (similar to SPRIV and LLVM dialects, examples in SPIRVTypes.h, LLVMTypes.h) 
  - suggest that custom type declarations and classes/functions declaration should be nested in our namespace to avoid "invading" the mlir namespace.
  - suggest that the ONNX MLIR dialect namespace should be nested in the mlir namespace. I would name it `onnx` rather that `onnxmlir` so that we would have (e.g mlir::onnx::<our types, classes,...>)
 
Similar consideration would apply to the Krnl dialect (once we get to restructuring that one).

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -1272,7 +1274,7 @@ class FrontendGenImpl {
     std::string comma = std::string("");
 
     TypeSwitch<Type>(argType)
-        .Case<mlir::onnxmlir::SeqType>([&](mlir::onnxmlir::SeqType seqTy) {
+        .Case<mlir::SeqType>([&](mlir::SeqType seqTy) {
```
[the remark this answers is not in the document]
chentong319: @etiotto Previously, I put only types (StringType and SeqType) in onnxmlir space. Since I am using table gen to define types now, I can not do that. So I use mlir name space for types as well as Ops temporarily. 
With table gen tool, we can put all the definitions of types and ops in ::mlir::onnx with one line of code. But our existing code are using the namespace explicitly with Ops (like mlir::ONNXConstantOp), it is a tedious but huge change. My plan is to leave that change to another PR.

src/Builder/FrontendDialectHelper.cpp
```diff
@@ -264,7 +264,7 @@ mlir::Type convertONNXTypeToMLIRType(
   case onnx::TensorProto_DataType::TensorProto_DataType_BOOL:
     return builder_.getI1Type();
   case onnx::TensorProto_DataType::TensorProto_DataType_STRING:
-    return mlir::onnxmlir::StringType::get(builder_.getContext());
+    return mlir::ONNXStringType::get(builder_.getContext());
```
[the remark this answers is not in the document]
chentong319: agree.

chentong319:
> @chentong319 thanks for spearheading this restructuring of the code!
> 
> I am assuming that we want to use naming conventions as close as possible to other MLIR dialects (for consistency and to prepare for possible inclusion in the MLIR codebase).
> 
> With that in mind, here is a summary of suggestions I put inline:
> 
> * suggest that we declare custom ONNX types in a header file rather that in a .td file (similar to SPRIV and LLVM dialects, examples in SPIRVTypes.h, LLVMTypes.h)
Yes, I plan to do that.
> * suggest that custom type declarations and classes/functions declaration should be nested in our namespace to avoid "invading" the mlir namespace (I agree with your comment in the description "I did not put ONNX dialect into a separate namespace, which I think we should")
The namespace for types and ops is controlled with the `cppNamespace` field of the dialect definition.
> * suggest that the ONNX MLIR dialect namespace should be nested in the mlir namespace. I would name it `onnx` rather that `onnxmlir` so that we would have (e.g mlir::onnx::<our types, classes,...>)
agree.
> 
> Similar consideration would apply to the Krnl dialect (once we get to restructuring that one).

src/Dialect/ONNX/ONNXOps.hpp
chentong319: With separate namespace, we will use mlir::onnx::StringType. The current ONNXStringType is just a hack to make the code compilable.

AlexandreEichenberger:
@chentong319 Did you address all of the comments? If you did, please ping the reviewers so that this PR can go soon as it is pretty large. Thanks

MLIR.cmake
```diff
@@ -70,11 +70,20 @@ function(add_onnx_mlir_dialect_doc dialect dialect_tablegen_file)
...
-  mlir_tablegen(${dialect}.hpp.inc -gen-op-decls "-I${ONNX_MLIR_SRC_ROOT}")
-  mlir_tablegen(${dialect}.cpp.inc -gen-op-defs "-I${ONNX_MLIR_SRC_ROOT}")
-  add_public_tablegen_target(OM${dialect}IncGen)
+# If an extra parameter, the dialect name, is provided,
+# this function will generate dialect and type from the td file
+function(add_onnx_mlir_dialect tdfile)
+  set(LLVM_TARGET_DEFINITIONS ${tdfile}.td)
+  mlir_tablegen(${tdfile}Ops.hpp.inc -gen-op-decls "-I${ONNX_MLIR_SRC_ROOT}")
+  mlir_tablegen(${tdfile}Ops.cpp.inc -gen-op-defs "-I${ONNX_MLIR_SRC_ROOT}")
+  if (${ARGC} EQUAL 2)
```
sstamenova, line 79: Rather than just checking by number of arguments, you can have "proper" named arguments and let cmake parse them for you. It's a lot less prone to error.

MLIR.cmake
```diff
@@ -70,11 +70,20 @@ function(add_onnx_mlir_dialect_doc dialect dialect_tablegen_file)
...
 add_custom_target(onnx-mlir-docs)
 
-function(add_onnx_mlir_dialect dialect)
-  set(LLVM_TARGET_DEFINITIONS ${dialect}.td)
-  mlir_tablegen(${dialect}.hpp.inc -gen-op-decls "-I${ONNX_MLIR_SRC_ROOT}")
-  mlir_tablegen(${dialect}.cpp.inc -gen-op-defs "-I${ONNX_MLIR_SRC_ROOT}")
-  add_public_tablegen_target(OM${dialect}IncGen)
+# If an extra parameter, the dialect name, is provided,
+# this function will generate dialect and type from the td file
+function(add_onnx_mlir_dialect tdfile)
```
sstamenova, line 75: Nit: I like the name `dialect` as the name of the input better.

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -568,53 +568,41 @@ static void insertConvTransposeSpatialDim(SmallVectorImpl<int64_t> &outputDims,
...
-    if (parser.parseGreater())
-      return Type();
-    return onnxmlir::SeqType::get(elementType);
-  }
-
-  parser.emitError(parser.getNameLoc(), "unknown onnx type: " + keyword);
+  Type type;
+  if (generatedTypeParser(parser, keyword, type).hasValue())
+    return type;
+  parser.emitError(parser.getNameLoc(), "invalid 'onnx' type:`")
```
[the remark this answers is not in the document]
chentong319: yes. It's a typo.

MLIR.cmake
chentong319: Changed

src/Dialect/ONNX/CMakeLists.txt
```diff
@@ -6,12 +6,14 @@ if (${CMAKE_SYSTEM_PROCESSOR} STREQUAL "s390x")
...
-add_onnx_mlir_dialect(ONNXOps)
-add_onnx_mlir_dialect_doc(onnx ONNXOps.td)
+add_onnx_mlir_dialect(ONNX onnx)
+add_onnx_mlir_dialect_doc(onnx ONNX.td)
 
 add_onnx_mlir_rewriter(Rewrite)
 
 add_onnx_mlir_library(OMONNXOps
+  ONNXDialect.cpp
+  ONNXTypes.cpp
```
sstamenova, line 16: Nit: sources are mostly alphabetic

sstamenova (approved):
LGTM

src/Builder/FrontendDialectTransformer.cpp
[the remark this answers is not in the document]
etiotto: That is fine. We agree that the ONNX and Krnl types eventually should belong to in their appropriate namespace inside mlir::,  I totally understand that we do this in steps.

src/Dialect/ONNX/ONNXOps.hpp
etiotto: ok. Whether we end up calling it  `mlir::onnx::ONNXStringType` or `mlir::onnx::ONNXStringType` doesn't make much of a difference. For now what you have is a good first step.

PR #1142, opened 2022-02-01
Regression Test for CategoryMapper Operator 

etiotto:
This PR depends on #1130.

This PR  adds an end-to-end test for the `CategoryMapper` operator (custom). That is it adds some infrastructure to create tests for operators using the C API for driving the compiler. This is similar to the `numerical` tests, but without using `rapidchecks`. 

The test infrastructure will be useful in a couple of scenarios:
   - in case a python based test is missing in ONNX (like it is for CategoryMapper)
   - in case we want to add additional end-to-end tests to supplement the python tests

Note: the new test infrastructure has some commonality with the `numerical` tests, however is also quite different because it does not use `rapidcheck` as it is not designed to ascertain the numerical accuracy of the result produced by an operator. For that reason, the tests runs a lot faster and so I thought it would be good to separate. them and create a new cmake target.


test/CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ add_subdirectory(backend)
 # The lit and unit tests do NOT require ONNX package installation
 add_subdirectory(mlir)
 add_subdirectory(numerical)
+add_subdirectory(extra)
```
sstamenova, line 9: I really don't like the name `extra` since it's not descriptive. Is there something more descriptive that we could use instead?

etiotto: I don't really like it much either :-) Open to suggestions.  I can think of:
  - regression (fo regression tests - in the sense of preventing regressions)
  - unittest

sstamenova: `unittest` is probably best, though `backend-cpp` is not bad either.

AlexandreEichenberger: fine with either

test/extra/TestCategoryMapper.cpp
```diff
@@ -0,0 +1,142 @@
...
+  CategoryMapperTester categoryMapperTester(ctx);
+  const CategoryMapperTester::CMAttributes attributes = {{1, 2, 3, 4, 5},
+      {"cat", "dog", "human", "tiger", "beaver"}, -1, "unknown"};
+
+  if (!categoryMapperTester.testInt64ToStr(attributes, {1, 2, 3, 4, 5},
+          {"cat", "dog", "human", "tiger", "beaver"}))
+    return 1;
+
+  return 0;
+}
```
etiotto, line 142: add newline

test/extra/TestHelper.hpp
```diff
@@ -0,0 +1,83 @@
...
+#endif
+  }
+
+private:
+  ModuleOp createEmptyModule() const;
+};
+
+} // namespace Tests
+
+#endif // ONNX_MLIR_TEST_HELPER_H
```
etiotto, line 83: add newline

test/CMakeLists.txt
etiotto: will use `backend-cpp`

test/backend-cpp/TestCategoryMapper.cpp
```diff
@@ -0,0 +1,141 @@
...
+    if (!testHelper.compileTest(SharedLibBase)) {
+      llvm::errs() << "Failed to compile test case\n";
+      return false;
+    }
+
+    // Run the test and verify the result.
+    using OMTensorPtr = unique_ptr<OMTensor, decltype(&omTensorDestroy)>;
+
+    std::vector<OMTensorPtr> inputOMTs, expectedOutputOMTs;
+    auto inputOMT = OMTensorPtr(omTensorCreate((void *)input.data(), inputShape,
```
sstamenova, line 66: I just noticed this - isn't the default to use c++ style casts?

etiotto: This being a C++ source file I would say yes! In general I agree that we should sue use a static_cast rather than a C style cast. Fixed.

sstamenova: Great, thanks! I think there are a couple more remaining ones.

etiotto: Changed them.

test/backend-cpp/TestHelper.cpp
```diff
@@ -0,0 +1,119 @@
...
+      return false;
+    }
+    if (omTensorGetNumElems(out) != omTensorGetNumElems(expected)) {
+      llvm::errs() << "Output tensor has " << omTensorGetNumElems(out)
+                   << "elements, expecting " << omTensorGetNumElems(expected)
+                   << "\n";
+      return false;
+    }
+
+    // Verify that the output tensor contains the expected result.
```
AlexandreEichenberger, line 89: Comparing here implies the type char, which may work for the very specific case you have. Templeting this and making it work for at least chart and float, and for float we use the support for "nearly the same"

test/backend-cpp/TestHelper.hpp
```diff
@@ -0,0 +1,86 @@
...
+#include <vector>
+
+using namespace std;
+using namespace mlir;
+
+namespace BackendCppTests {
+
+// Helper class containing useful function for creating, compiling and running a
+// test.
+class TestHelper {
```
AlexandreEichenberger, line 35: I like the idea of this, maybe the interface could be simplified a bit.  This is what I "implemented" informally in the model library, but without classes

* Build and compile
* Run
* Compare.

and since the build and compile contains all the loc, context, module. builder, there was no need to expose them.

There is a certain elegance in having these var in the class, as then you can have support operations. 
I can see the following process:
1. Build the graph (using helper function). One issue is that for some operations, we need to fill in weights (which are randomly selected numbers). These "constants" needs then to be passed to the computations for the expected values. But these constants can be added to the class, so that is a neat way to do it. So you have here an "abstract builder", and each subclass implements the actual graph. This builder functions would have many parameters specialized for the actual operations, e.g. matmul needs I,J,K, convolution needs N C H W ...
2. compile, that can be done in the parent class
3. compute expected: that would be specialized by each of the subclass, and result in the vector of expected value
4. run: again that can be done by the superclass
5. compare results: you can have implementations of comparison for each type of results, currently float or string

So I like this idea, I would rename this "ModelBuilder" and combine the ones from the numerical and this one into a single library of models we can build, and then have very lightweight process to do the rapid check (lots of tests) vs backend-cpp which is a few tests.

etiotto: > and since the build and compile contains all the loc, context, module. builder, there was no need to expose them.

Several tests would need the same functionality to create a test case. So I have provided member functions to facilitate creating the skeleton for testing an operator. Users of `TestHelper` would then just need to provide the functionality to create the operator that needs to be tested, and rely on the TestHelper class to deal with the skeleton (create a module, create a function in the module, push a return operator in the function, etc..)

test/backend-cpp/TestHelper.cpp
etiotto: I am going to pass the verification code as a function object so that the caller of `runAndVerifyTest` can customize the verification code appropriately for the particular test scenario.

test/backend-cpp/TestHelper.hpp
etiotto: > compare results: you can have implementations of comparison for each type of results, currently float or string

I can address this point by allowing the user of `ModelBuilder` to provide a function object where the verification code is "encoded". That way the consumer can customize the verification code appropriately for the particular scenario it wants to test.

AlexandreEichenberger (approved):
LGTM, really like the changes. Will try to apply the same scheme to numerical tests.

PR #1144, opened 2022-02-02
Make OptLevel option sharable

tungld:
This patch allows other drivers to use the OptLevel option.


etiotto (approved):
LGTM

AlexandreEichenberger (changes requested):
I like it, I also think that I may have used the variable directly here and there. Do you mind searching for these usages and replacing with the interface you are suggesting? Thanks

tungld:
> Do you mind searching for these usages and replacing with the interface you are suggesting? 

Yes, sure. I found two direct usages and changed them.

AlexandreEichenberger (approved):
LGTM, thanks for the changes

PR #1147, opened 2022-02-02
Small dependency optimization for ModelLib

sstamenova:
`ModelLib` already has a dependency on `ExecutionSession` and `CompilerUtils`. By making the dependency public, it can propagate to other targets simplifying the dependency chain.


AlexandreEichenberger (approved):
LGTM, thanks

PR #1149, opened 2022-02-03
Accelerator enablement

caoimhinuibrian:
preliminary framework to allow for integration of accelerator targets


src/Support/OMOptions.cpp
```diff
@@ -40,3 +40,8 @@ llvm::cl::opt<int> onnxOpTransformThreshold("onnx-op-transform-threshold",
 llvm::cl::opt<bool> onnxOpTransformReport("onnx-op-transform-report",
     llvm::cl::desc("Report diagnostic info for op transform passes."),
     llvm::cl::init(false), llvm::cl::cat(OMPassOptions));
+
```
AlexandreEichenberger, line 43: I would suggest to also have an infrastructure to add options from inside the Accelerator, where maybe we record at a very high level the fact that we have one accelerator, and then in the options, we call each recorded accelerators custom options?

AlexandreEichenberger (commented):
It would help to have a general description of the structure and classes that you are seen, how we can insert optimization passes in our driver.

src/Accelerators/Accelerator.cpp
```diff
@@ -0,0 +1,37 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//===-------------------------- OMAccelerator.hpp -------------------------===//
```
etiotto, line 5: [typo]:  "OMAccelerator.hpp" --> "Accelerator.cpp"

src/Accelerators/Accelerator.cpp
```diff
@@ -0,0 +1,37 @@
...
+//
+// Copyright 2022 The IBM Research Authors.
+//
+// =============================================================================
+//
+// OMAccelerator base class
```
etiotto, line 11: OMAccelerator --> Accelerator

src/Accelerators/NNPA/NNPAAccelerator.hpp
```diff
@@ -0,0 +1,30 @@
...
+//
+// Accelerator class for NNPA
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+#include "src/Accelerators/Accelerator.hpp"
+
+namespace mlir {
+class NNPAAccelerator : public Accelerator {
```
etiotto, line 20: [Suggestion]: Make class `final` to prevent deriving it (assuming that NNPA accelerator will not need to be derived from at this point at least).

src/Accelerators/Accelerator.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include "src/Accelerators/NNPA/NNPAAccelerator.hpp"
+#include <iostream>
+#include <vector>
+
+namespace mlir {
+extern NNPAAccelerator nnpaAccelerator;
+std::vector<Accelerator *> *Accelerator::acceleratorTargets;
+
+Accelerator::Accelerator() {
+  std::cout << "creating Accelerator" << std::endl;
```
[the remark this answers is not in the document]
caoimhinuibrian: yep. Will do.

src/Accelerators/Accelerator.hpp
```diff
@@ -0,0 +1,27 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//===-------------------------- Accelerator.hpp -------------------------===//
+//
+// Copyright 2022 The IBM Research Authors.
+//
+// =============================================================================
+//
+// OMAccelerator base class
```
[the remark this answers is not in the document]
caoimhinuibrian: done

src/Accelerators/Accelerator.cpp
```diff
@@ -0,0 +1,37 @@
...
+    acceleratorTargets->push_back(&nnpaAccelerator);
+  }
+}
+
+std::vector<Accelerator *> *Accelerator::getAcceleratorList() {
+  std::cout << "getting accelerator targets" << std::endl;
+  return acceleratorTargets;
+}
+
+} // namespace mlir
```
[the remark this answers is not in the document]
caoimhinuibrian: k, will do later

src/Accelerators/NNPA/NNPAAccelerator.hpp
caoimhinuibrian: Done

caoimhinuibrian:
@tungld could you take a look at this please? I'd like to get it merged soon.

tungld:
@caoimhinuibrian I did some quick fixes to make Jenkins CI passed. I am off today, so will take a closer look tomorrow.

src/Accelerators/Accelerator.cpp
```diff
@@ -0,0 +1,37 @@
...
+// OMAccelerator base class
+//
+//===----------------------------------------------------------------------===//
+#include "src/Accelerators/Accelerator.hpp"
+#include "src/Accelerators/NNPA/NNPAAccelerator.hpp"
+#include <iostream>
+#include <vector>
+
+namespace mlir {
+extern NNPAAccelerator nnpaAccelerator;
```
AlexandreEichenberger, line 20: Is it necessary to have `nnpa` anything here? (include, variable)

src/Accelerators/Accelerator.hpp
```diff
@@ -0,0 +1,33 @@
...
+  static std::vector<Accelerator *> *getAcceleratorList();
+  virtual bool isActive() = 0;
+  virtual void prepareAccelerator(mlir::OwningOpRef<ModuleOp> &module,
+      mlir::MLIRContext &context, mlir::PassManager &pm,
+      onnx_mlir::EmissionTargetType emissionTarget) = 0;
+
+private:
+  static std::vector<Accelerator *> *acceleratorTargets;
+};
+} // namespace mlir
```
AlexandreEichenberger, line 33: nit, add a new empty line at end of file, not sure why it complains about it, folks typically add one.

src/Accelerators/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+  foreach(t ${ACCELERATORS_TO_BUILD})
+    message(STATUS "Targeting ${t}")
+    add_subdirectory(${t})
+    endforeach(t)
+endif( ACCELERATORS_TO_BUILD )
+
+
+
+
+  add_onnx_mlir_library(NoAccel
```
AlexandreEichenberger, line 13: If you  believe it can be removed later, just add a "todo, remove"

src/Accelerators/NNPA/NNPAAccelerator.hpp
```diff
@@ -0,0 +1,33 @@
...
+public:
+  NNPAAccelerator();
+
+  void prepareAccelerator(mlir::OwningOpRef<ModuleOp> &module,
+      mlir::MLIRContext &context, mlir::PassManager &pm,
+      onnx_mlir::EmissionTargetType emissionTarget) override;
+  bool isActive() override;
+};
+
+} // namespace mlir
```
AlexandreEichenberger, line 33: nit end of line

src/Accelerators/NNPA/src/Compiler/DLCompilerUtils.cpp
```diff
@@ -0,0 +1,159 @@
+#include "mlir/Conversion/AffineToStandard/AffineToStandard.h"
```
AlexandreEichenberger, line 1: I assume you will do a grep for all ".cpp", ".hpp" files that don't have a copyright. I can do it once this patch is in, if you prefer.

AlexandreEichenberger (approved):
LGTM, minor changes suggested but nothing that cannot be done later as this will be somewhat iterative. Probably the biggest next step is integrating the tests, so that the CI folks can start changing the scripts... which should also be sensitive to the ACCELERATORS_TO_BUILD variable you added.

src/Accelerators/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+# SPDX-License-Identifier: Apache-2.0
+
+if( ACCELERATORS_TO_BUILD )
+  foreach(t ${ACCELERATORS_TO_BUILD})
+    message(STATUS "Targeting ${t}")
```
sstamenova, line 5: I would make this message either verbose or debug, so that it can be turned on easily, but not clutter the log otherwise.

src/Accelerators/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+    add_subdirectory(${t})
+    endforeach(t)
+endif( ACCELERATORS_TO_BUILD )
+
+
```
sstamenova, line 10: Nit: lots of blank lines that should go away

src/Accelerators/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+
+
+  add_onnx_mlir_library(NoAccel
+  NoAccelerator.cpp
+
+  DEPENDS
```
sstamenova, line 16: If there are no dependencies, you can skip the keyword altogether

src/Accelerators/NNPA/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+set(DLC_BIN_ROOT "${CMAKE_CURRENT_BINARY_DIR}")
+
+set(DLC_LIBRARY_PATH ${CMAKE_LIBRARY_OUTPUT_DIRECTORY})
+set(DLC_RUNTIME_PATH ${CMAKE_RUNTIME_OUTPUT_DIRECTORY})
+
+set(DLC_ONNX_MLIR_SRC_ROOT ${ONNX_MLIR_SRC_ROOT})
+set(DLC_ONNX_MLIR_BIN_ROOT ${ONNX_MLIR_BIN_ROOT})
+
+# The path to find the zDNN library.
+if(DEFINED ENV{ZDNN_LIBRARY_DIR} AND EXISTS $ENV{ZDNN_LIBRARY_DIR})
```
sstamenova, line 13: Nit: space after if

sstamenova, line 13: This is looking for the library dir in the environment. In general, cmake files try to avoid using the environment as much as possible. Can you use something like find_package to find the library instead?

src/Accelerators/NNPA/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+  set(ZDNN_LIBRARY_ENABLED 1)
+else()
+  message(STATUS "ZDNN_LIBRARY_DIR not found. zDNN-related tests will be turned off.")
+  set(ZDNN_LIBRARY_ENABLED 0)
+endif()
+
+add_subdirectory(src)
+
+
+message(STATUS "in NNPA: DLC_ONNX_MLIR_SRC_ROOT=${DLC_ONNX_MLIR_SRC_ROOT}")
```
sstamenova, line 25: All these messages will be better of as debug or verbose.

src/Accelerators/NNPA/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+  DEPENDS
+  DLCompilerUtils
+  
+  INCLUDE_DIRS PUBLIC
+  ${ONNX_MLIR_SRC_ROOT}/include
+  ${ONNX_MLIR_SRC_ROOT}
+  ${DLC_ONNX_MLIR_SRC_ROOT}
+  ${DLC_SRC_ROOT}
+  ${DLC_BIN_ROOT}
+  /workdir/onnx-mlir/third_party/onnx
```
sstamenova, line 41: If you need to use onnx, as onnx as a dependency and the includes will automatically be taken care of. Or at least should.

src/Accelerators/NNPA/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
+# SPDX-License-Identifier: Apache-2.0
+
+set(DLC_SRC_ROOT "${CMAKE_CURRENT_SOURCE_DIR}")
```
sstamenova, line 3: Do all the `DLC_` properties need to be set? If not, they are not really needed.

src/Accelerators/NNPA/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+  ${DLC_ONNX_MLIR_SRC_ROOT}
+  ${DLC_SRC_ROOT}
+  ${DLC_BIN_ROOT}
+  /workdir/onnx-mlir/third_party/onnx
+
+  LINK_LIBS PUBLIC
+  onnx
+  )
+
+  target_link_libraries(NNPAAccel PUBLIC DLCompilerUtils)
```
sstamenova, line 47: These should be under `LINK_LIBS` above.

src/Accelerators/NNPA/NNPAAccelerator.hpp
```diff
@@ -0,0 +1,33 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//===-------------------------- NNPAAccelerator.hpp
```
sstamenova, line 5: Nit: fix this header

src/Accelerators/NNPA/src/CMakeLists.txt
```diff
@@ -0,0 +1,110 @@
+
+# add_dlc_library(name sources...
```
sstamenova, line 2: This looks like a copy of the `add_library` function from MLIR.cmake. Is the intention to add a new global library set?

src/Compiler/CMakeLists.txt
```diff
@@ -54,13 +54,32 @@ add_onnx_mlir_library(CompilerUtils
...
   )
 
+if( ACCELERATORS_TO_BUILD )
+  add_dependencies(CompilerUtils NNPAAccel)
+else ()
+  add_dependencies(CompilerUtils NoAccel)
+endif(ACCELERATORS_TO_BUILD)
+
+
+if( ACCELERATORS_TO_BUILD )
```
sstamenova, line 64: Same comments about spacing here. I'm sure I have other comments on the cmake files, but I'll give you a chance to clean them up a bit more first.

src/Compiler/CMakeLists.txt
```diff
@@ -54,13 +54,32 @@ add_onnx_mlir_library(CompilerUtils
   MC
   )
 
+if( ACCELERATORS_TO_BUILD )
+  add_dependencies(CompilerUtils NNPAAccel)
```
sstamenova, line 58: This would be better resolved as a list of libraries that just gets added above. I saw you have a function for adding the accelerator libraries - once that's cleaned up, it can automatically hold either NoAccel or the list of accelerator libraries to be added as one of the libs under `DEPENDS` for CompilerUtils.

sstamenova (changes requested):

src/Accelerators/Accelerator.cpp
caoimhinuibrian: Yep. It needs to be here to get the accelerator included.

src/Accelerators/Accelerator.hpp
caoimhinuibrian: done

src/Accelerators/CMakeLists.txt
caoimhinuibrian: let me think about it

src/Accelerators/NNPA/src/Compiler/DLCompilerUtils.cpp
caoimhinuibrian: will add license in another PR

src/Accelerators/NNPA/CMakeLists.txt
caoimhinuibrian: the zdnn library doesn't have any cmake files in it, so I don't believe I can

src/Accelerators/NNPA/CMakeLists.txt
caoimhinuibrian: I believe they do need to be set, but I will check

src/Accelerators/NNPA/src/CMakeLists.txt
caoimhinuibrian: Yeah, this is a mess .... I will tidy up in a follow-on PR if that's ok?

src/Accelerators/NNPA/src/Conversion/ZHighToZLow/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+set(LLVM_TARGET_DEFINITIONS ZHighToZLow.td)
+mlir_tablegen(ZHighToZLow.inc -gen-rewriters "-I${DLC_ONNX_MLIR_SRC_ROOT}" "-I${DLC_SRC_ROOT}")
```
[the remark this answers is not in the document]
caoimhinuibrian: Not sure what you are asking here?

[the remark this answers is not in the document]
sstamenova: There is a function that invokes `mlir_tablegen` for you in MLIR.cmake. Can you use it here instead of directly invoking `mlir_tablegen` and having to do all the steps around it yourself?

src/Accelerators/NNPA/src/CMakeLists.txt
sstamenova: I think it would be better to do it now or in a separate PR that goes in before this change since things that get postponed often get postponed indefinitely.

Also just noticed that there's a second `src` directory under the accelerators path. Is it needed? It seems redundant.

tungld (approved):
LGTM as the very first step for the enablement.

However, I highly recommend to change license headers in the stickify folder.

src/Accelerators/NNPA/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
...
+set(DLC_BIN_ROOT "${CMAKE_CURRENT_BINARY_DIR}")
+
+set(DLC_LIBRARY_PATH ${CMAKE_LIBRARY_OUTPUT_DIRECTORY})
+set(DLC_RUNTIME_PATH ${CMAKE_RUNTIME_OUTPUT_DIRECTORY})
+
+set(DLC_ONNX_MLIR_SRC_ROOT ${ONNX_MLIR_SRC_ROOT})
+set(DLC_ONNX_MLIR_BIN_ROOT ${ONNX_MLIR_BIN_ROOT})
+
+# The path to find the zDNN library.
+if (DEFINED ENV{ZDNN_LIBRARY_DIR} AND EXISTS $ENV{ZDNN_LIBRARY_DIR})
```
sstamenova, line 13: How is the library shipped? Even if it doesn't have cmake files, there might be ways for cmake to find it.

src/Accelerators/NNPA/src/CMakeLists.txt
caoimhinuibrian: removed the extra src
still working on the add_dlc_library

src/Accelerators/NNPA/CMakeLists.txt
caoimhinuibrian: This PR adds it to third_party. It seems to vave a regular makefile, but I haven't looked at it.

caoimhinuibrian:
@sstamenova could you take a look and let me know if I've answered your concerns, please?

src/CMakeLists.txt
```diff
@@ -10,6 +11,9 @@ add_subdirectory(Runtime)
 add_subdirectory(Tools)
 add_subdirectory(Compiler)
 
+
```
sstamenova, line 14: Nit: remove extra blank lines

src/Accelerators/NNPA/CMakeLists.txt
sstamenova: I see, so the source is added to `third_party`, I can't seem to figure out how it is built and these environment paths are set?

MLIR.cmake
```diff
@@ -141,6 +141,11 @@ function(add_onnx_mlir_library name)
     target_include_directories(${name} ${ARG_INCLUDE_DIRS})
   endif()
 
+  if (ARG_ACCEL_INCLUDE_DIRS)
```
sstamenova, line 144: So now these libraries will be added to both global properties - is that the goal?

I think it would be better to make the function optionally have a global property to add to and just pass in a different property for both usages (or default to the current one).

sstamenova (changes requested):
In general, this is a very large change, if you could break out pieces, I think it would be easier to commit code that is ready and doesn't need any post-commit changes (such as the headers, etc.). I would think that making all the necessary changes pre-commit would be better for long-term maintenance.

caoimhinuibrian:
> In general, this is a very large change, if you could break out pieces, I think it would be easier to commit code that is ready and doesn't need any post-commit changes (such as the headers, etc.). I would think that making all the necessary changes pre-commit would be better for long-term maintenance.

Yes, I realize that, and normally would try not to make such a big change. However, this is a copy of some code that was previously internal to IBM that we are trying to open, and so getting it merged will allow others to work on it. As long as it is pending, we have an issue with maintaining both pieces. It's also the case that it has almost no impact on the current code. The only point of contact is the NoAccelerator library and the cmake changes to support adding this and any future accelerators.

src/Accelerators/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+  foreach(t ${ACCELERATORS_TO_BUILD})
+    message(DEBUG "Targeting ${t}")
+    add_subdirectory(${t})
+    list(APPEND ACCEL_LIST ${t}"Accel")
+    endforeach(t)
+else (ACCELERATORS_TO_BUILD)
+  set (ACCEL_LIST "NoAccel")    
+endif (ACCELERATORS_TO_BUILD)
+
+add_onnx_mlir_library(NoAccel
```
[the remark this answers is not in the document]
caoimhinuibrian: only if there are no accelerators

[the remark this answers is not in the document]
sstamenova: In that case, can you change the logic to do so?

sstamenova:
> > In general, this is a very large change, if you could break out pieces, I think it would be easier to commit code that is ready and doesn't need any post-commit changes (such as the headers, etc.). I would think that making all the necessary changes pre-commit would be better for long-term maintenance.
> 
> Yes, I realize that, and normally would try not to make such a big change. However, this is a copy of some code that was previously internal to IBM that we are trying to open, and so getting it merged will allow others to work on it. As long as it is pending, we have an issue with maintaining both pieces. It's also the case that it has almost no impact on the current code. The only point of contact is the NoAccelerator library and the cmake changes to support adding this and any future accelerators.

Can you break that specific part out? Maybe a PR that adds the NoAccelerator library and the necessary cmake changes?

src/Accelerators/NNPA/Compiler/CMakeLists.txt
```diff
@@ -13,7 +13,7 @@ add_onnx_mlir_library(DLCompilerUtils
 
   LINK_LIBS PUBLIC
   CompilerUtils
-  ${DLCOMLibs}
+  ${OMLibs}
```
sstamenova, line 16: Isn't this too much? You are now passing all the OMLibs? Or is it OK?

MLIR.cmake
sstamenova: With your latest change, the DLC_LIBS are no longer being set, but all the accelerator libs are added to the OMLibs. Is that intentional?

MLIR.cmake
```diff
@@ -143,7 +143,6 @@ function(add_onnx_mlir_library name)
 
   if (ARG_ACCEL_INCLUDE_DIRS)
     target_include_directories(${name} ${ARG_ACCEL_INCLUDE_DIRS})
```
sstamenova, line 145: Also, the includes don't need a special include property any more.

PR #1153, opened 2022-02-03
Fix a bug in LoopOp lowering where scan outputs are wrongly updated

### LINKED ISSUE #1152: Accuracy issues in yolo models
Reposted an issue reported by @cjvolzka 

For this issue I'm focusing on yolov3 (yolov3.onnx from onnx-model zoo) as an example but we're seeing similar results in other models we tried like ssd-10, tiny-yolo-v3-11, and yolov4. 

For the yolov3, there's three output tensors for each inference. The first output tensor values have accuracy issues compared to the expected results. The second tensor appears OK. The third has -nan results. All the output tensors have different shapes so the issue doesn't appear to be that that we have the first and third tensor mismatched.

tungld:
There is a bug in LoopOp lowering when scan outputs depend on loop carried dependency, but loop carried dependency is updated before returning outputs so the outputs are shifted by one iteration.

This patch updates loop carried dependency at the end of an iteration, so it does not affect outputs in the current iteration.

Resolves #1152 


AlexandreEichenberger:
@chentong319 Added you as a reviewer as you have recently worked on the LOOP code, thanks

src/Conversion/ONNXToKrnl/ControlFlow/Loop.cpp
```diff
@@ -175,6 +169,12 @@ struct ONNXLoopOpLowering : public ConversionPattern {
             std::get<1>(scanIntermediateToFinal),
             /*writePrefix=*/{origIV});
 
+      // Copy intermediate values of loop carried dependencies to MemRef outside
+      // the iteration scope so next iteration can have use them as init value.
```
etiotto, line 173: "can have use" -> "can use"

tungld: Thanks! Fixed it.

chentong319 (approved):
LGTM!

PR #1156, opened 2022-02-03
Introduce onnx-mlir-reduce - a tool to reduce ONNX MLIR (failing) tests.

etiotto:
This pull request introduces a new tool called `onnx-mlir-reduce` which can be used to reduce failing models (or simpler mlir tests). The tool uses the infrastructure in MLIR for reduction of tests (https://mlir.llvm.org/docs/Tools/mlir-reduce), but is aware of the "onnx-mlir" and "krnl" dialects.

The tool attempts to remove lines in the input file to reduce it to a smaller "minimal" one that can still be used to reproduce the original problem. 

To drive the tool one needs:
   - an input test case (this is the original failing test case)
   - a script to drive the compilation process
   
As an example say `model.onnx.mlir` is the input test, and the compilation script (`compile.sh`) is:
```
#!/bin/bash

# Tests for the keyword "error" in the stderr of the compiler output
onnx-mlir-opt --shape-inference -convert-onnx-to-krnl --convert-krnl-to-affine $1 > /tmp/stdout.$$ 2>/tmp/stderr.$$

if [ $? -ne 0 ] && grep 'error' /tmp/stderr.$$; then
  exit 1
else
  exit 0
fi
``` 

To use the tool one would write:
```
onnx-mlir-reduce model.onnx.mlir -reduction-tree='test=compile.sh' -o model.onnx.reduced.mlir
```

I tested it on an input file containing 1009 lines, the tool was able to reduce the test to 819 lines:
```
% wc -l model.onnx.mlir
    1009 model.onnx.mlir
% wc -l model.onnx.reduced.mlir
     819 model.onnx.reduced.mlir
```     


src/onnx-mlir.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+  llvm::cl::opt<string> inputFilename(llvm::cl::Positional,
+      llvm::cl::desc("<input file>"), llvm::cl::init("-"),
+      llvm::cl::cat(OnnxMlirOptions));
+
+  llvm::cl::opt<string> outputBaseName("o",
+      llvm::cl::desc("Base path for output files, extensions will be added."),
+      llvm::cl::value_desc("path"), llvm::cl::cat(OnnxMlirOptions),
+      llvm::cl::ValueRequired);
+  llvm::cl::opt<EmissionTargetType> emissionTarget(
```
sstamenova, line 34: Nit: new line between the options

sstamenova: I assume this file is showing up as new, but is just main.cpp. It would still be nice to add the new line.

etiotto: Yes will do that. Is not a new file, I renamed it from `main.cpp` to `onnx-mlir.cpp` because I think the name is more descriptive. Also renamed the file we used as the main file in for `onnx-mlir-opt` to `onnx-mlir-opt.cpp` because is more descriptive and is more conforming to the naming convention used by the similar MLIR tool `mlir-opt`.  Hope you guys do not mind the renaming.

etiotto: done

sstamenova (approved):

PR #1161, opened 2022-02-04
Update llvm to 853e0aa424e4

etiotto:
Changes:
  - SelectOp, ConstantOp moved to arith dialect from. builtin dialect.
  - OwningModuleRef is now a template (OwningOpRef<ModuleOp>)
  - populateStdExpandOpsPatterns is now part of the memref namespace 
  - ConstantOp in std dialect used to allow a "unit" constant. This is no longer allowed (https://reviews.llvm.org/D118654).

The change that is causing the most trouble is the one in https://reviews.llvm.org/D118654.  We can no longer create a unit constant (e.g. `%0 = constant unit`). We a few 'lit' tests that rely on the unit constant. Here is an example:

```
func @test_upsample(%arg0: tensor<1x1x2x2xf32>, %arg1: tensor<4xf32>) -> tensor<1x1x4x6xf32> {
  %0 = "onnx.Upsample"(%arg0, %arg1) {mode = "nearest"} : (tensor<1x1x2x2xf32>, tensor<4xf32>) -> tensor<1x1x4x6xf32>
  return %0 : tensor<1x1x4x6xf32>
}
```

Which gets transformed into: 
```
func @test_upsample(%arg0: tensor<1x1x2x2xf32>, %arg1: tensor<4xf32>) -> tensor<1x1x4x6xf32> {
  %cst = arith.constant unit
  %cst_0 = arith.constant unit
  %0 = "onnx.Resize"(%arg0, %cst, %arg1, %cst_0) {mode = "nearest"} : (tensor<1x1x2x2xf32>, none, tensor<4xf32>, none) -> tensor<1x1x4x6xf32>
   ...
}
```

Here the 2 unit constants are used to pass a `None` type to the `onnx.Resize` argument `roi` and `sizes`.  That is, they represent the absence of a value (I think). Given that we no longer can use the `constant unit` from the standard dialect, I think we will need to create a custom unit constant in the ONNX Dialect.


etiotto:
@AlexandreEichenberger @tungld @chentong319 given that `constant unit` has been removed in MLIR (see description for details) we need to determine how to represent a `None` type, which, AFAIK  is used to represent the absence of a value. Some ONNX Operators (like onnx.Resize) have argument that can take a `None` type. Again AFAIK we used to pass a `constant unit` (which has None type) to those arguments when a concrete value is not available. 

How can we represent the absence of a value ?  Should we create a custom ONNX operator for it ?

tungld:
Look at [Tensorflow-lite](https://github.com/tensorflow/tensorflow/commit/260cf9c59a104a252ffe4a299666a9efe437789a), I see that they define a new op which is similar to `constant unit`, e.g.
```
%[[CST:.*]] = "tfl.no_value"() {value} : () -> none
```
Should we do the same thing?

etiotto:
> Look at how [Tensorflow-lite](https://github.com/tensorflow/tensorflow/commit/260cf9c59a104a252ffe4a299666a9efe437789a) responded to the constant issue, I see that they define a new op which is similar to `constant unit`, e.g.
> 
> ```
> %[[CST:.*]] = "tfl.no_value"() {value} : () -> none
> ```
> 
> Should we do the same thing? It is just a special op that returns None, no code is required.

My opinion is that we should introduce a similar operator.

AlexandreEichenberger:
agreed, looked like they removed it from the lower dialect because it should be a "higher dialect only" issue.

tungld:
> My opinion is that we should introduce a similar operator.

Agreed.

etiotto:
Opened https://github.com/onnx/onnx-mlir/pull/1169 to introduce a custom operation to replace the std ConstantOp with unit value. Once that PR lands this one should go through easily.

etiotto:
Now that https://github.com/onnx/onnx-mlir/pull/1169 has landed I am going to rebase this PR and continue.

tungld (approved):
LGTM. Thanks!

PR #1163, opened 2022-02-07
ExecutionSession constructor without entry function name

tungld:
Currently, Python users need to pass the entry function name "run_main_graph" to ExecutionSession, e.g.
```python
sess = ExecutionSession(compiled_model_path, "run_main_graph")
outs = sess.run(inputs)
```
which is inconvenient because users need to remember the entry function name.

To simplify ExecutionSession, this patch creates a new constructor without entry function name. An entry function name will be internally obtained via a newly defined function `omEntryPointName`. With this patch, users can write the following python code instead:
```python
sess = ExecutionSession(compiled_model_path)
outs = sess.run(inputs)
```

There is no change about `ExecutionSession(compiled_model_path, "run_main_graph")` so users can still pass an entry point name if they know the name.


etiotto:
@tungld I am wondering whether we even need to allow users to set the entry point name of the main graph. Are there user scenarios or internal scenarios that requires a custom entry point name ?  If not I think it would be OK to remove that functionality.

AlexandreEichenberger:
@etiotto I believe the desire to let users named the entry point function was to enable multiple "networks" within the same ONNX model file. If that feature is not needed, obviously we could use only a single fixed name, but that works only if all the framework generate the same "main_graph" entry point.

@tungld Are you trying to preserve the multiple entry feature, or are you enabling that whatever the ONNX model named the entry point, it can be retrieved? 

Additional nit, I believe the ".so" is used for LINUX/OSX whereas the ".a" is used for Window. It might be better to remove that from the interface too, so that the same python code works on both systems.

chentong319 (commented):
Could there be more than one entry point? For example, multiple disconnected computation graph in a model?

AlexandreEichenberger:
@chentong319 exactly, that is why I recall we have "named" entry points

tungld:
I am sorry for all confusion. In this patch we will keep all the features of ExecutionSession, e.g. allowing to pass an entry point name `ExecusionSession(lib_path, "run_main_graph"), nothing changes except a new constructor `ExecutionSession(lib_path)` which additionally calls function `omEntryPointName` to get the entry point name.

So users can still pass an entry point name to ExecutionSession if they want.

Regarding multiple entry point names, yes, I understand the intention so nothing changes here. However, I think we are not ready for that, for example, when we have multiple entry points, `omInputSignature` and `omOutputSignature` also need to be updated, say, to get signature of a given entry point. At that time, `omEntryPointName` should be updated also.

> Additional nit, I believe the ".so" is used for LINUX/OSX whereas the ".a" is used for Window. It might be better to remove that from the interface too, so that the same python code works on both systems.

Yes, I should not use `.so` in the PR comment. The change here works for Windows as we can see Windows CI passed.

tungld:
> I am wondering whether we even need to allow users to set the entry point name of the main graph. Are there user scenarios or internal scenarios that requires a custom entry point name ?

@etiotto, Entry point name is generated by our onnx importer, so I think users cannot change it. But, if we want to test an MLIR program, we can set it freely, for example, I often do this:
```mlir
module {
  func @resize(%arg0 : tensor<?x256x13x13xf32>, %arg1 : tensor<4xi64>) -> tensor<?x?x?x?xf32> {
    %cst = constant unit
    %0= "onnx.Resize"(%arg0, %cst, %cst, %arg1) {coordinate_transformation_mode = "half_pixel", cubic_coeff_a = -7.500000e-01 : f32, exclude_outside = 0 : si64, extrapolation_value = 0.000000e+00 : f32, mode = "nearest", nearest_mode = "floor"} : (tensor<?x256x13x13xf32>, none, none, tensor<4xi64>) -> tensor<?x?x?x?xf32>
    return %0 : tensor<?x?x?x?xf32>
  }
  "onnx.EntryPoint"() {func = @resize, numInputs = 2 : i32, numOutputs = 1 : i32, signature = ""} : () -> ()
}
```

In this case entry point name would be `run_resize`. With this patch, we don't need to care about the entry point name. We store `run_resize` as a global constant and provide a function `omEntryPointName` to read the constant.

gongsu832:
I remember having a conversation with @tjingrant regarding the entry point name quite a while ago at the beginning of the transition from DLC Classic to ONNX-MLIR. You may have noticed that in the public header `OnnxMlirRuntime.h` there is no prototype declaration for `run_main_graph`. Instead the user needs to write the declaration him/herself.
```
OMTensorList* run_main_graph(OMTensorList*);
```
I think the rationale was that there might be cases where the users want to chose the entry point name themselves. @tungld seemed to have just provided an example for it. And as @tungld mentioned we don't yet have a way for users to do this today but back then it was envisioned that we might provide an option for users to do this at some point.

etiotto:
OK thanks guys for the explanation.

AlexandreEichenberger:
@tungld  @gongsu832  thanks for the clarification and historical view point.

> Regarding multiple entry point names, yes, I understand the intention so nothing changes here. However, I think we are not ready for that, for example, when we have multiple entry points, omInputSignature and omOutputSignature

So it look that we are currently in an inconsistent state, which we should probably address. There seems to be 3 options:
1. fixed name only for one entry point. In this case, we can forcibly rename entry point whatever name to run_main_graph
2. user defined name for only one entry point. In this case, our current solution works, and I like Tung's addition. But really, option 1 works just as fine if we don't give the user the choice of the name.
3. user defined name for possibly multiple entry points. Then we need to add the "function" name to the get signature functions, just like we prepend "run_" to the graph execution entry point. We can then chose to have Tung's approach to find out the name **when there is only one** entry point. We must offer the signature getter with an additional parameter.

Personally, I find 1 and 3 consistent, not sure the benefit of 2 but I have no strong objections to option 2 either.

src/Runtime/PyExecutionSession.hpp
```diff
@@ -25,6 +25,8 @@ namespace onnx_mlir {
 
 class PyExecutionSession : public onnx_mlir::ExecutionSession {
 public:
+  PyExecutionSession(std::string sharedLibPath)
+      : onnx_mlir::ExecutionSession(sharedLibPath){};
```
etiotto, line 29: [nit]: remove trailing semicolon (also the one at line 31 pls).

src/Runtime/ExecutionSession.hpp
```diff
@@ -47,6 +49,7 @@ class ExecutionSession {
   llvm::sys::DynamicLibrary _sharedLibraryHandle;
 
   // Entry point function.
+  static const std::string _entryPointNameFuncName;
```
etiotto, line 52: [Suggestion]:  _entryPointNameFuncName -> _entryPointFuncName

src/Runtime/ExecutionSession.cpp
```diff
@@ -36,6 +42,21 @@ ExecutionSession::ExecutionSession(
     throw std::runtime_error(errStr.str());
   }
 
+  // Get entry point name if it is not given.
+  if (entryPointName.empty()) {
+    entryPointNameFuncType _entryPointNameFunc =
```
etiotto, line 47: [nit]: no need to prefix variable with an underscore for a local variable.



```

**Comment by tungld** [comment]
It looks like we get a consensus on supporting multiple entry points.

Though I haven't seen onnx models that have multiple disconnected subgraphs, one practical case I could imagine is to compile a model for accelerators where the compiled model could include two entry points: one for CPU and the other for accelerator. Then, we can set device for a session, e.g. 
```
sess = ExecutionSession(compiled_model_path)
sess.setDevice("CPU")
```

> 3. user defined name for possibly multiple entry points. Then we need to add the "function" name to the get signature functions, just like we prepend "run_" to the graph execution entry point. We can then chose to have Tung's approach to find out the name when there is only one entry point. We must offer the signature getter with an additional parameter.

I agreed. If there is no objection I will continue this patch for one entry point, then will propose a new patch to fully support multiple entry points.

src/Runtime/ExecutionSession.hpp
tungld: Yeah, it looks a bit confusing but it is actually the name of the function that returns entry point name. Let me keep this name since `entryPointFuncName` may be confusing with the name of the entry point function.

tungld:
@AlexandreEichenberger now we fix the entry point to "run_main_graph" when there is only a single entry point. Multiple entry points will be left for a new PR.

PR #1165, opened 2022-02-07
Fix ResizeOp lowering 

chentong319:
The error was found in yolov4 model. The previous lowering of ResizeOp did not pad the input tensor.  The mode of padding is "edge". This PR implement this padding (for read only) by assigning the index to the edge index (0 or dim-1) if the computed value is out of bound.
Other changes:

1. Add support to all the nearest mode.
2. Restructure the code.
3. Add more backend test


src/Conversion/ONNXToKrnl/Tensor/Resize.cpp
```diff
@@ -147,65 +137,83 @@ struct ONNXResizeOpLowering : public ConversionPattern {
     outputLoops.createDefineAndIterateOp(alloc);
     rewriter.setInsertionPointToStart(outputLoops.getIterateBlock());
 
+    // Constants used in the loop body
+    Value zero = emitConstantOp(rewriter, loc, rewriter.getIntegerType(64), 0);
```
etiotto, line 141: I saw we have `MathBuilder::constant` and I am under the impression we should use that function (by building a `createMath` builder) rather than  `emitConstantOp`. Can you update the code pls?

AlexandreEichenberger: Yes, I recommend the using of the builder, that also isolate us a bit from constant MLIR changes.

chentong319: I replaced the old code with create.math and create.krnl utilities. Lots of changes, particularly for this file. Not recommend such action for old code unless absolutely needed.

src/Conversion/ONNXToKrnl/Tensor/Resize.cpp
```diff
@@ -147,65 +137,83 @@ struct ONNXResizeOpLowering : public ConversionPattern {
...
+            emitConstantOp(rewriter, loc, rewriter.getF32Type(), 0.499999);
+        inIndexFloat =
+            rewriter.create<arith::AddFOp>(loc, inIndexFloat, deltaConstant);
+      } else if (resizeOp.nearest_mode() == "round_prefer_ceil") {
+        Value deltaConstant =
+            emitConstantOp(rewriter, loc, rewriter.getF32Type(), 0.5);
+        inIndexFloat =
+            rewriter.create<arith::AddFOp>(loc, inIndexFloat, deltaConstant);
+      } else if (resizeOp.nearest_mode() == "floor") {
+        inIndexFloat = rewriter.create<math::FloorOp>(loc, inIndexFloat);
```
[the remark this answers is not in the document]
chentong319: That's the limitation of such wrapper.

tungld (approved):
LGTM! Thank you for the fix!

tungld:
@AlexandreEichenberger @etiotto is this ready for merge? Just in case, if you have further comments, please let us know. Thanks!

PR #1166, opened 2022-02-07
Introduce onn-mlir options -Xopt, -Xllc, and -mllvm

etiotto:
There are scenarios where a user of `onnx-mlir` might need to tune the behaviour of the LLVM `opt` and/or `llc` components.  This PR introduces three new options:
- -Xopt <valid opt's flag> -- the flag will be passed to `opt`
- -Xllc <valid llc's flag> -- the flag will be passed to `llc`
- -mllvm <valid opt and llc flag> -- the flag will be passed to both `opt` and `llc`

### DIFF
--- a/include/onnx-mlir/Compiler/OMCompilerTypes.h
+++ b/include/onnx-mlir/Compiler/OMCompilerTypes.h
@@ -5,10 +5,12 @@
 #ifndef ONNX_MLIR_OMCOMPILERTYPES_H
 #define ONNX_MLIR_OMCOMPILERTYPES_H
 
+#ifdef __cplusplus
 namespace onnx_mlir {
+#endif
 
 /* Type of compiler emission targets */
-enum EmissionTargetType {
+typedef enum {
   EmitONNXBasic,
   EmitONNXIR,
   EmitMLIR,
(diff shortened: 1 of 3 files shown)


include/onnx-mlir/Compiler/OMCompilerTypes.h
```diff
@@ -5,42 +5,44 @@
 #ifndef ONNX_MLIR_OMCOMPILERTYPES_H
 #define ONNX_MLIR_OMCOMPILERTYPES_H
 
+#ifdef __cplusplus
 namespace onnx_mlir {
+#endif
 
 /* Type of compiler emission targets */
-enum EmissionTargetType {
+typedef enum {
```
AlexandreEichenberger, line 13: Is that the preferred way? In the OnnxMlirRuntime, they use this:
```
#ifndef __cplusplus
typedef struct OMTensor OMTensor;
#endif
```
namely they use the typedef only for C. I am fine with either versions, want a clarification on which one is better, if there is any differences between the two.

AlexandreEichenberger (approved):
LGTM, pending answer to question on C/C++ style

include/onnx-mlir/Compiler/OMCompilerTypes.h
etiotto: Yes. In C the typedef declaration allows the user to write `EmissionTargetType tgtType;` while the way it was before would force C users to write `enum EmissionTargetType tgtType;`. For C++ users there is not difference between the 2 declarations.

src/Compiler/CompilerUtils.cpp
```diff
@@ -121,6 +121,22 @@ static llvm::cl::opt<bool> VerboseOutput("v",
...
+    llvm::cl::desc("Arguments to forward to LLVM's 'opt' option processing"),
+    llvm::cl::value_desc("A valid LLVM's 'opt' option"),
+    llvm::cl::cat(OnnxMlirOptions), llvm::cl::Hidden, llvm::cl::ValueRequired);
+
+static llvm::cl::opt<std::string> Xllc("Xllc",
+    llvm::cl::desc("Arguments to forward to LLVM's 'llc' option processing"),
+    llvm::cl::value_desc("A valid LLVM's 'llc' option"),
+    llvm::cl::cat(OnnxMlirOptions), llvm::cl::Hidden, llvm::cl::ValueRequired);
+
+static llvm::cl::opt<std::string> mllvm("mllvm",
```
sstamenova, line 134: Nit: I would put all of these options together in this file, so they can be found more easily (the other `-m` options are a bit higher)

src/Compiler/CompilerUtils.cpp
```diff
@@ -411,6 +432,18 @@ static std::string getOptimizationLevelOption() {
   return "";
 }
 
+static std::string getXoptOption() {
+  return (Xopt != "") ? Xopt : std::string();
```
sstamenova, line 436: So for the rest of the options, we return `--march=arch`, for example, but for these we just return the option. Is that going to get passed correctly to the underlying tools?

src/Compiler/CompilerUtils.cpp
```diff
@@ -309,26 +325,37 @@ void setTargetArch(const std::string &arch) { march = arch; }
 void setTargetTriple(const std::string &triple) { mtriple = triple; }
 void setOptLevel(const OptLevel level) { OptimizationLevel = level; }
 OptLevel getOptLevel() { return OptimizationLevel; }
```
sstamenova, line 327: OptLevel is the only one that has just a plain get. Is it needed?

src/Compiler/CompilerUtils.cpp
etiotto: yes that is intentional, given -Xopt --llvm_option_name  we will pass `--llvm_option_name" to the tool.

src/Compiler/CompilerUtils.cpp
etiotto: Yeah, we should avoid the "asymmetry".  Best to leave these untouched in this PR though because  @AlexandreEichenberger is restructuring the option processing via PR https://github.com/onnx/onnx-mlir/pull/1167

PR #1168, opened 2022-02-08
Support BF16 in Frontend

chentong319:

Add the case for BF16 type in frontend helper.


AlexandreEichenberger (approved):
LGTM, can you just add a comment if that result us in supporting this new type, or if not, what are the next steps. Thanks

chentong319:
> A simple test to cover the new option would be nice.

We do not have an easy way to add frontend test yet. We use onnx/backend to test onnx-mlir frontend, but we do want our test cases into onnx. That's a work item for us.
I tested a simple model. onnx-mlir passed the llvm IR generation but got an error after that. I will open an issue.

PR #1169, opened 2022-02-08
Introduce ONNXUnitConstantOp to replace std ConstantOp with unit value

etiotto:
ConstantOp in std dialect used to allow a "unit" constant. This is no longer allowed (https://reviews.llvm.org/D118654).
In this PR we are addressing that change in MLIR by introducing a custom operation in the ONNX dialect. 

### DIFF
--- a/src/Conversion/ONNXToKrnl/Math/Reduction.cpp
+++ b/src/Conversion/ONNXToKrnl/Math/Reduction.cpp
@@ -413,7 +413,7 @@ struct ONNXReduceSumOpLowering : public ConversionPattern {
 
     Value axesValue = llvm::dyn_cast<ONNXReduceSumOp>(op).axes();
     // Dynamic axes
-    if (!isFromNone(axesValue) && !getONNXConstantOp(axesValue)) {
+    if (!isDefinedByUnitConstant(axesValue) && !getONNXConstantOp(axesValue)) {
       dynamicAxes = true;
       // Handle only when keepdims == true
       if (!isKeepdims) {
@@ -444,7 +444,8 @@ struct ONNXReduceSumOpLowering : public ConversionPattern {
                 axesBounds.getDim(0).getValue(), zeroIndex);
         initVal = rewriter.create<SelectOp>(loc, cond, trueVal, falseVal);
       } else {
-        // When axesDim is known, it can not be 0 due to !isFromNone
+        // When axesDim is known, it can not be 0 due to
+        // !isDefinedByUnitConstant
         initVal = falseVal;
       }
       for (auto i = 0; i < inRank; i++) {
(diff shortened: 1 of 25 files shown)


tungld:
Thank you for proposing this op!

It seems to me this op is for create a None value only so could we simplify it. e.g.:

1. Declare a custom builder so that we can use `create<ONNXUnitConstantOp>(UnknownLoc::get(&ctx))` instead of `create<ONNXUnitConstantOp>(UnknownLoc::get(&ctx), builder.getNoneType(), builder.getUnitAttr())`
2. Suggest to use a more descriptive name, e.g. ONNXNoneOp. Honestly, I saw some users were confused on the usage of `constant unit` for none value.

etiotto:
> Declare a custom builder so that we can use create<ONNXUnitConstantOp>(UnknownLoc::get(&ctx)) instead of create<ONNXUnitConstantOp>(UnknownLoc::get(&ctx), builder.getNoneType(), builder.getUnitAttr())

Good idea, incidentally I was thinking about it last night. Make sense to simplify the op that way. 

> Suggest to use a more descriptive name, e.g. ONNXNoneOp. Honestly, I saw some users were confused on the usage of constant unit for none value.

This is subjective of course. I was debating about `ONNXNoneOp` and `ONNXConstantUnitOp`. I settled on the later because it I was replacing a ConstantOp with unit value.  I can see how some ppl might get confused by the term "constant unit` though.

chentong319:
What's the usage of the unit constant? std::constant unit is used to represent the optional tensor in onnx-mlir. In the newer version of onnx model, tensor with 0 length, such as tensor<0xi32> and tensor<0xf32>, is used.

etiotto:
> What's the usage of the unit constant? std::constant unit is used to represent the optional tensor in onnx-mlir...

A 'unit constant' according to this discussion (https://discourse.llvm.org/t/standard-dialect-the-final-chapter/6061) has no well defined semantics in MLIR (hence the removal). We currently use it in cases to provide a 'None' value to arguments of operators that are optional arguments. For example: 

```
  %cst = constant unit
  %0, %1 = "onnx.Split"(%arg0, %cst) { axis = 0 : si64} : (tensor<16x32x64xf32>, none) -> (tensor<*xf32>, tensor<*xf32>)
```

Here the second argument to `onnx.Split` can take a value with "none" type. The unit constant provides that value. 

> In the newer version of onnx model, tensor with 0 length, such as tensor<0xi32> and tensor<0xf32>, is used.
We'd better to have a wrapper to handle both cases.

What are the semantics of tensor with zero elements ?

manbearian:
> > What's the usage of the unit constant? std::constant unit is used to represent the optional tensor in onnx-mlir...
> 
> A 'unit constant' according to this discussion (https://discourse.llvm.org/t/standard-dialect-the-final-chapter/6061) has no well defined semantics in MLIR (hence the removal). We currently use to provide a 'None' value to arguments of operators that are optional arguments. For example:
> 
> ```
>   %cst = constant unit
>   %0, %1 = "onnx.Split"(%arg0, %cst) { axis = 0 : si64} : (tensor<16x32x64xf32>, none) -> (tensor<*xf32>, tensor<*xf32>)
> ```
> 
> Here the second argument to `onnx.Split` can take a value with "none" type. The unit constant provides that value.
> 
> > In the newer version of onnx model, tensor with 0 length, such as tensor<0xi32> and tensor<0xf32>, is used.
> > We'd better to have a wrapper to handle both cases.
> 
> What are the semantics of tensor with zero elements ?

joining the discussion as i'm interested in seeing how this gets resovled. FWIW i took @chentong319 's comment and quickly replaced constant-unit with a 0-length i64 tensor along one path through onnx-mlir and was able to get a simple test case working

etiotto:
Addressed comments from @tungld

src/Dialect/ONNX/ONNX.td
```diff
@@ -192,18 +184,28 @@ def ONNXBatchNormalizationInferenceModeOp: ONNX_Op<"BatchNormalizationInferenceM
...
+    static int getNumberOfResults() { return 1; }
+    static std::vector<int> getTypeMap() { return {20}; }
   }];
 }
 
+def ONNXNoneOp : ONNX_Op<"NoValue", [ConstantLike, NoSideEffect]> {
+  let summary = "An operation representing the absence of a value.";
+  let description = [{
+    This operation can be used to represent the absence of a value. It is typically 
+    used as an argument to operators that have optional parameters.
```
tungld, line 197: Would be helpful if we add an example here, say
```mlir
```

so that uses know `onnx.NoValue` is created by ONNXNoneOp.

src/Conversion/ONNXToKrnl/Math/Reduction.cpp
```diff
@@ -444,7 +444,8 @@ struct ONNXReduceSumOpLowering : public ConversionPattern {
                 axesBounds.getDim(0).getValue(), zeroIndex);
         initVal = rewriter.create<SelectOp>(loc, cond, trueVal, falseVal);
       } else {
-        // When axesDim is known, it can not be 0 due to !isFromNone
+        // When axesDim is known, it can not be 0 due to
+        // !isFromNone
```
tungld, line 448: Looks like an unexpected formatting.

tungld (approved):
LGTM!

I have some minor comments, please feel free to address them. Thanks!

src/Conversion/ONNXToKrnl/Math/Reduction.cpp
etiotto: fixed

etiotto: done

PR #1172, opened 2022-02-09
Generalize CategoryMapper test.

etiotto:
In this PR I generalized the test infrastructure for the CategoryMapper operator by providing templated functions to cut down on code duplication. As a side effect I modified the runtime interfaces for querying OMTensor properties to pass input OMTensor(s) with `const`.

### DIFF
--- a/src/Runtime/OMTensorHelper.h
+++ b/src/Runtime/OMTensorHelper.h
@@ -108,7 +108,7 @@ OMTensor *omTensorCreateWithRandomData(
  * @return typed element by reference at the offset computed by the index array.
  */
 template <typename T>
-T &omTensorGetElem(OMTensor *omt, std::vector<int64_t> indexes);
+T &omTensorGetElem(const OMTensor *omt, std::vector<int64_t> indexes);
 
 /**
  * OMTensor data element getter by index
@@ -118,7 +118,7 @@ T &omTensorGetElem(OMTensor *omt, std::vector<int64_t> indexes);
  * @return typed element by reference at the linear offset.
  */
 template <typename T>
-T &omTensorGetElemByOffset(OMTensor *omt, int64_t index);
+T &omTensorGetElemByOffset(const OMTensor *omt, int64_t index);
 
 /**
  * OMTensor strides computation
@@ -126,7 +126,7 @@ T &omTensorGetElemByOffset(OMTensor *omt, int64_t index);
  * @param omt, pointer to the OMTensor
  * @return data strides of the OMTensor computed from the data sizes.
  */
-std::vector<int64_t> omTensorComputeStridesFromShape(OMTensor *omt);
+std::vector<int64_t> omTensorComputeStridesFromShape(const OMTensor *omt);
 
 /**
  * OMTensor linear offset computation
@@ -135,7 +135,8 @@ std::vector<int64_t> omTensorComputeStridesFromShape(OMTensor *omt);
  * @param indexes, multi-dimensional index array
  * @return linear offset.
  */
-int64_t omTensorComputeElemOffset(OMTensor *omt, std::vector<int64_t> &indexes);
+int64_t omTensorComputeElemOffset(
+    const OMTensor *omt, std::vector<int64_t> &indexes);
 
 /**
  * OMTensor index set computation
@@ -145,7 +146,7 @@ int64_t omTensorComputeElemOffset(OMTensor *omt, std::vector<int64_t> &indexes);
  *         that can be used to access this OMTensor's constituent elements)
  *         for the whole OMTensor.
  */
-std::vector<std::vector<int64_t>> omTensorComputeIndexSet(OMTensor *omt);
+std::vector<std::vector<int64_t>> omTensorComputeIndexSet(const OMTensor *omt);
 
 /**
  * OMTensor "distance" computation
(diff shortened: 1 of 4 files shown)


etiotto:
@NathanielMcVicar @sstamenova I saw this error on the Win CI: 
```
##[error]No hosted parallelism has been purchased or granted. To request a free parallelism grant, please fill out the following form https://aka.ms/azpipelines-parallelism-request
```

Maybe is transitory?

NathanielMcVicar:
> Maybe is transitory?

Unfortunately, it doesn't seem to be. I'm going to fill out the request form since this seems to meet the definition of a public project. I suspect when we moved over to an ONNX build subscription to host more artifacts we lost our 2 free parallel jobs.

NathanielMcVicar:
This appears to be a wide issue effecting AzDO pipelines across organizations. The ONNX team is aware and working to rootcause the issue. See for example https://github.com/conda-forge/conda-forge.github.io/issues/1604. Thanks @jcwchen!

etiotto:
> This appears to be a wide issue effecting AzDO pipelines across organizations. The ONNX team is aware and working to rootcause the issue. See for example https://github.com/conda-forge/conda-forge.github.io/issues/1604. Thanks @jcwchen!

Thank you very much @NathanielMcVicar and @jcwchen for the quick response and investigation.

test/backend-cpp/TestCategoryMapper.cpp
```diff
@@ -123,59 +156,106 @@ class CategoryMapperTester {
...
+    if (!verifyRank(*out, omTensorGetRank(expected)))
+      return false;
+    if (!verifyNumElements(*out, omTensorGetNumElems(expected)))
+      return false;
 
     // Verify that the output tensor contains the expected result.
-    auto outDataPtr = (const char **)omTensorGetDataPtr(out);
-    auto expectedDataPtr = (const char **)(omTensorGetDataPtr(expected));
+    const auto *outDataPtr = static_cast<T *>(omTensorGetDataPtr(out));
+    const auto *expDataPtr = static_cast<T *>(omTensorGetDataPtr(expected));
```
AlexandreEichenberger, line 195: Can you comment on if that cast is safer/ok compared to what was done in numerical, namely
```
  for (int64_t i = 0; i < I; ++i) {
    for (int64_t j = 0; j < J; ++j) {
      omTensorGetElem<float>(ref, {i, j}) = 0;
      for (int64_t k = 0; k < K; k++) {
        float aVal, bVal;
        if (aTrans == 0)
          aVal = omTensorGetElem<float>(a.get(), {i, k});
        else
          aVal = omTensorGetElem<float>(a.get(), {k, i});
        if (bTrans == 0)
          bVal = omTensorGetElem<float>(b.get(), {k, j});
        else
          bVal = omTensorGetElem<float>(b.get(), {j, k});
        omTensorGetElem<float>(ref, {i, j}) += aVal * bVal;
      }
    }
  }
```

I agree that the quoted code above is dreadfully slow by creating a vector for each accesses

src/Runtime/OMTensorHelper.h
```diff
@@ -108,7 +108,7 @@ OMTensor *omTensorCreateWithRandomData(
  * @return typed element by reference at the offset computed by the index array.
  */
 template <typename T>
-T &omTensorGetElem(OMTensor *omt, std::vector<int64_t> indexes);
+T &omTensorGetElem(const OMTensor *omt, std::vector<int64_t> indexes);
```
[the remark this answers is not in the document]
etiotto: Hi @AlexandreEichenberger. The proposed declaration prevents `*omt` to be assigned to. That is, prevents the function to modify the memory location pointed to by `omt`. That memory location can only be read. 

Here is a simple example that clarifies the semantics:
```
int foo(const int *pi) {
  *pi = 3; // wrong: cannot assing the memory pointed to by pi.
  pi++; //ok: can change where pi points to (need "const int * const pi") to prevent this.
  return *pi;
}
```

test/backend-cpp/TestCategoryMapper.cpp
etiotto: The implementation of `omTensorGetElem` is: 
```
template <typename T>
T &omTensorGetElem(const OMTensor *omt, std::vector<int64_t> indexes) {
  int64_t elemOffset = omTensorComputeElemOffset(omt, indexes);
  return ((T *)omt->_alignedPtr)[elemOffset];
}
```

The type cast on the return stmt is in C style. A C-style cast `(T *)` is **not** equivalent to a C++ `static_cast<T*>()` because a C-style cast is not type safe, while a `static_cast ` allows the compiler to report errors. For example: 

```
void foo(char *p) {
  int *pi = (int *)p;   // no compiler error here
  *pi = 3;              // ooops: just overwrote 3 extra bytes in mem. pointed to by p!!!
}
```

If a `static_cast<int*>(p)` was used the compiler would catch the potential runtime problem at compile time ==> safer!

If the intention was to "reinterpret" the memory pointed to by `p` then one should use a C++ `reintepret_cast`.

Of course all of this implies the code section is compiled by a C++ compiler.

etiotto: We prob. should modify the runtime implementation of `omTensorGetElem` and use C++ style casts. But that is another PR.

AlexandreEichenberger (approved):
LGTM

test/backend-cpp/TestCategoryMapper.cpp
etiotto: ok

etiotto:

PR #1173, opened 2022-02-10
Exec session updates

AlexandreEichenberger:
@etiotto recently introduce an alias for `unique_ptr<OMTensor, decltype(&omTensorDestroy)>`, and I used (a slightly renamed version of it) everywhere in the code.

Also, our execution session was only supporting vectors of unique ptr of OMTensor, but not the native OMTensorList. I added a function to also support this.

This PR is short, a larger discussion is whether we should add also a C interface in our runtime for execution session, which essentially does dynamic linking of our entry points.


src/Runtime/ExecutionSession.cpp
```diff
@@ -63,9 +63,8 @@ ExecutionSession::ExecutionSession(
   }
 }
 
-std::vector<std::unique_ptr<OMTensor, decltype(&omTensorDestroy)>>
-ExecutionSession::run(
-    std::vector<std::unique_ptr<OMTensor, decltype(&omTensorDestroy)>> ins) {
+std::vector<OMTensorUniquePtr> ExecutionSession::run(
+    std::vector<OMTensorUniquePtr> ins) {
```
etiotto, line 67: I like the name `OMTensorUniquePtr`, will use it as well.

etiotto (approved):
LGTM

src/Runtime/ExecutionSession.cpp
AlexandreEichenberger: I changed it in your recent code as well.

PR #1175, opened 2022-02-11
Expose ONNX type, size, and name constants in OMTensor

gongsu832:


AlexandreEichenberger (approved):
LGTM once you add the list of files  where this info is redundantly defined to all the places it is defined. I.e. `/* If you update this list, do it also in files x, y, and z */

Thanks

gongsu832:
This is the only file in Java that has the list. C/C++ has its own list but that is completely unrelated to the Java list.

gongsu832:
No the two lists are independent of each other so they don't have to be in sync. However, since they are both from the standard ONNX types they naturally tend to be the same.

PR #1176, opened 2022-02-11
Build benchmark with new class structure

AlexandreEichenberger:
Use new class structure to build models. In its simplest form, here is the "template" for a new test.
```C++
class GemmLibBuilder : public ModelLibBuilder {
public:
  GemmLibBuilder(const std::string &modelName, const int I, const int J,
      const int K, const int aTrans, const int bTrans, const int cRank,
      const float alphaVal, const float betaVal);
  bool build();
  bool prepareInputs();
  bool verifyOutputs();

private:
  // Data that defines model.
  const int I, J, K, aTrans, bTrans, cRank;
  const float alphaVal, betaVal;
  // Derived data that defines model.
  llvm::SmallVector<int64_t, 2> aShape, bShape, cShape;
};
```

where the user adds the variables private to the model, and defines 3 functions to build, compute inputs, and verify outputs.

Test is called as follows
```C++
  GemmLibBuilder gemm(
      SHARED_LIB_BASE.str(), I, J, K, aTrans, bTrans, cRank, alphaVal, betaVal);
  return gemm.build() && gemm.compileAndLoad() && gemm.prepareInputs() &&
         gemm.run() && gemm.verifyOutputs();
```
Where the object is created with all its specific parameters, and then executed using the build, compile, prepare, run, and verify sequence. For timed run: build, compile, and prepare can be done before timing, and only run is done in the timed loop. Verify is then optional.


test/modellib/ModelLib.hpp
```diff
@@ -97,3 +99,67 @@ bool genLSTMModelAndCompile(
...
+  bool build() { llvm_unreachable("subclass must implement build."); }
+  // Compile model from the model and ctx variables. The output is an executable
+  // dynamic library.
+  bool compileAndLoad();
+  // Prepare inputs for running model.
+  bool prepareInputs() { llvm_unreachable("subclass must implement prepare."); }
+  // Run model using prepared inputs, resulting in outputs.
+  bool run();
+  // Verify outputs with reference data.
+  bool verifyOutputs() { llvm_unreachable("subclass must implement verify."); }
```
etiotto, line 123: The member functions that should be implemented by derived classes can be declared as pure virtual functions as:
```
virtual bool verifyOutputs() = 0;
```

That way the compiler will know that derived classes must override the pure virtual function at compile time, that is it will not let you instantiate a derived class object unless all the member functions that are pure virtual are defined in the derived class.

test/modellib/ModelLib.hpp
```diff
@@ -97,3 +99,67 @@ bool genLSTMModelAndCompile(
...
     llvm::SmallVector<int64_t, 3> &cShape, OMTensor *&wOmt, OMTensor *&rOmt,
     OMTensor *&bOmt, OMTensor *&pOmt);
+
+class ModelLibBuilder {
+public:
+  // Define the model. Subclass should add to the builder all of the specific
+  // parameters that uniquely define the model.
+  ModelLibBuilder(const std::string &sharedLibBaseName);
+  // Destructor needed to free the inputs/outputs data structures.
+  ~ModelLibBuilder();
```
etiotto, line 109: This class acts as an abstract class defining an interface that derived classes must implement. The destructor should be made virtual, that way if a `ModelLibBuilder*` is deleted, and the pointer points to a derived class the destructor of the derived class will be run, giving the derived class a chance to clean up resources (if needed).

test/modellib/ModelLib.cpp
```diff
@@ -0,0 +1,100 @@
...
+  std::string fullName(sharedLibBaseName + ".dll");
+#else
+  std::string fullName(sharedLibBaseName + ".so");
+#endif
+  exec = new ExecutionSession(fullName);
+  return exec != nullptr;
+}
+
+bool ModelLibBuilder::run() {
+  assert(exec && "expected successful compile and load");
```
etiotto, line 57: perhaps add `assert(inputs && "expected inputs tensors to be provided)`

test/modellib/ModelLib.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+void ModelLibBuilder::createEntryPoint(FuncOp &funcOp) {
+  FunctionType funcType = funcOp.getType();
+  auto entryPoint = ONNXEntryPointOp::create(
+      loc, funcOp, funcType.getNumInputs(), funcType.getNumResults(), "");
+  module.push_back(entryPoint);
+}
+
+ONNXConstantOp ModelLibBuilder::buildONNXConstantOp(
+    OMTensor *omt, RankedTensorType resultType) {
```
etiotto, line 90: suggest `const OMTensor *`

AlexandreEichenberger: make sense

test/modellib/ModelHelper.hpp
```diff
@@ -12,13 +12,19 @@
 //
 //===----------------------------------------------------------------------===//
 
+#pragma once
```
etiotto, line 15: Need to add a header guard

test/modellib/ModelLib.hpp
```diff
@@ -15,85 +15,212 @@
 //
 //===----------------------------------------------------------------------===//
 
+#pragma once
```
etiotto, line 18: Remove by accident ?

test/modellib/ModelLib.hpp
etiotto: The disadvantage of virtual member functions is the runtime cost for virtual dispatch. In this case I think that is not important given that this class is for internal testing of models, not something that will penalize the runtime of the `onnx-mlir` compiler.

AlexandreEichenberger: I retooled the constructors to make it easier to understand that it is not to be used in a "virtual" way

AlexandreEichenberger: ok, fully addressed by adding virtual

test/modellib/ModelLib.hpp
AlexandreEichenberger: ok, fully addressed by adding virtual

test/modellib/ModelHelper.hpp
AlexandreEichenberger: file was removed because not in use anymore

test/modellib/ModelLib.hpp
AlexandreEichenberger: checked, it is now there

test/modellib/ModelLib.hpp
```diff
@@ -67,26 +67,30 @@ const std::string getAutoPadName(const int autoPad);
...
 public:
-  // Define the model. Subclass should add to the builder all of the specific
+  // Define the model. Subclasses should add to the builder all of the specific
   // parameters that uniquely define the model.
   ModelLibBuilder(const std::string &sharedLibBaseName);
   // Destructor needed to free the inputs/outputs data structures.
-  ~ModelLibBuilder();
+  virtual ~ModelLibBuilder();
+  // Default constructor removed as only subclasses may construct a
+  // ModelLibBuilder using the protected constructors.
```
etiotto, line 78: Adjust comment because now there are no protected constructors.

etiotto (approved):
LGTM - minor comment added inline

PR #1179, opened 2022-02-15
[NFI]: Introduce the RuntimeAPI and RuntimeAPIRegistry classes

etiotto:
This PR consolidates code used to create a calls to Runtime API into its own class `RuntimeAPI`. The new class `RuntimeAPIRegistry` (a singleton) drives the creation of a registry containing the runtime API the compiler can call. 

This PR does not change or introduce new functionality. It simply reorganizes code in a more modular way. The intention is to make it easier to add new runtime API in the future.


tungld (approved):
LGTM!

Thanks for re-organizing! KrnlToLLVM looks less busy now 👍

PR #1180, opened 2022-02-15
Add OMTensorPrint and KrnlPrintTensorOp

etiotto:
This PR introduces the ability to generate a call to a runtime function `omTensorPrint` while generating code. This function allows printing the content of the supplied tensor. This is intended as an aid for debugging runtime failures.

To use it write:
``` 
LLVM_DEBUG(llvm::dbgs() << "Input tensor:\n"; create.krnl.printTensor(X););
```

In addition is sometime useful to print a primitive value at some program point in the code generated. For that purpose I have also introduced the `krnlPrintOp` Krnl dialect operator. This operator is lowered to a call to `printf` with the appropriate format specifier for the value to be printed. 

To use it write:
```
create.krnl.printf("inputElem: ", inputElem, elementType);
```

where `inputElem` is the value to be printed, and `elementType` is the type of the value to be printed.

### DIFF
--- a/docs/DebuggingNumericalError.md
+++ b/docs/DebuggingNumericalError.md
@@ -82,3 +82,21 @@ optional arguments:
   --rtol RTOL           Relative tolerance for verification
   --atol ATOL           Absolute tolerance for verification
 ```
+
+## Debugging the Code Generated for an Operator.
+
+If you know, or suspect, that a particular ONNX MLIR operator produces an incorrect result, and want to narrow down the problem, we provide a couple of useful Krnl operators that allow printing (at runtime) the value of a tensor, or a value that has a primitive data type. 
+
+To print out the value of a tensor at a particular program point, inject the following code (where `X` is the tensor to be printed):
+
+```code
+create.krnl.printTensor("Tensor X: ", X);
+```
+
+Note: currently the content of the tensor is printed only when the tensor rank is less than four.
+
+To print a message followed by one value, inject the following code (where `val` is the value to be printed and `valType` is its type):
+
+```code
+create.krnl.printf("inputElem: ", val, valType);
+```
--- a/include/onnx-mlir/Runtime/OMTensor.h
+++ b/include/onnx-mlir/Runtime/OMTensor.h
@@ -284,6 +284,14 @@ int64_t omTensorGetOwning(const OMTensor *tensor);
  */
 void omTensorSetOwning(OMTensor *tensor, int64_t owning);
 
+/**
+ * Print an OMTensor to stdout.
+ *
+ * @param msg, pointer to descriptive string
+ * @param tensor, pointer to the OMTensor to print
+ */
+void omTensorPrint(const char *msg, const OMTensor *tensor);
+
 #ifdef __cplusplus
 }
 #endif
--- a/include/onnx-mlir/Runtime/OnnxDataTypeMetaData.inc
+++ b/include/onnx-mlir/Runtime/OnnxDataTypeMetaData.inc
@@ -3,28 +3,28 @@
  */
 
 #if defined(OM_TYPE_METADATA_DEF)
+// clang-format off
 // Data type metadata declared in the following format:
-// OM_TYPE_METADATA_DEF( dtype enum name, dtype enum value, dtype size)
+// OM_TYPE_METADATA_DEF(dtype enum name, dtype enum value, dtype size, dtype name) 
 // dtype enum values are standard ONNX data types defined in
 // https://github.com/onnx/onnx/blob/main/onnx/onnx.proto#L484
-// clang-format off
-OM_TYPE_METADATA_DEF(ONNX_TYPE_UNDEFINED,   0,  0)
-OM_TYPE_METADATA_DEF(ONNX_TYPE_FLOAT,       1,  sizeof(float))
-OM_TYPE_METADATA_DEF(ONNX_TYPE_UINT8,       2,  sizeof(uint8_t))
-OM_TYPE_METADATA_DEF(ONNX_TYPE_INT8,        3,  sizeof(int8_t))
-OM_TYPE_METADATA_DEF(ONNX_TYPE_UINT16,      4,  sizeof(uint16_t))
-OM_TYPE_METADATA_DEF(ONNX_TYPE_INT16,       5,  sizeof(int16_t))
-OM_TYPE_METADATA_DEF(ONNX_TYPE_INT32,       6,  sizeof(int32_t))
-OM_TYPE_METADATA_DEF(ONNX_TYPE_INT64,       7,  sizeof(int64_t))
-OM_TYPE_METADATA_DEF(ONNX_TYPE_STRING,      8,  0)
-OM_TYPE_METADATA_DEF(ONNX_TYPE_BOOL,        9,  sizeof(bool))
-OM_TYPE_METADATA_DEF(ONNX_TYPE_FLOAT16,     10, 2)
-OM_TYPE_METADATA_DEF(ONNX_TYPE_DOUBLE,      11, sizeof(double))
-OM_TYPE_METADATA_DEF(ONNX_TYPE_UINT32,      12, sizeof(uint32_t))
-OM_TYPE_METADATA_DEF(ONNX_TYPE_UINT64,      13, sizeof(uint64_t))
-OM_TYPE_METADATA_DEF(ONNX_TYPE_COMPLEX64,   14, 8)
-OM_TYPE_METADATA_DEF(ONNX_TYPE_COMPLEX128,  15, 16)
-OM_TYPE_METADATA_DEF(ONNX_TYPE_BFLOAT16,    16, 2)
+OM_TYPE_METADATA_DEF(ONNX_TYPE_UNDEFINED,   0,  0,                "undefined")
+OM_TYPE_METADATA_DEF(ONNX_TYPE_FLOAT,       1,  sizeof(float),    "float")
+OM_TYPE_METADATA_DEF(ONNX_TYPE_UINT8,       2,  sizeof(uint8_t),  "uint8_t")
+OM_TYPE_METADATA_DEF(ONNX_TYPE_INT8,        3,  sizeof(int8_t),   "int8_t")
+OM_TYPE_METADATA_DEF(ONNX_TYPE_UINT16,      4,  sizeof(uint16_t), "uint16_t")
+OM_TYPE_METADATA_DEF(ONNX_TYPE_INT16,       5,  sizeof(int16_t),  "int16_t")
+OM_TYPE_METADATA_DEF(ONNX_TYPE_INT32,       6,  sizeof(int32_t),  "int32_t")
+OM_TYPE_METADATA_DEF(ONNX_TYPE_INT64,       7,  sizeof(int64_t),  "int64_t")
+OM_TYPE_METADATA_DEF(ONNX_TYPE_STRING,      8,  0,                "const char *")
+OM_TYPE_METADATA_DEF(ONNX_TYPE_BOOL,        9,  sizeof(bool),     "_Bool")
+OM_TYPE_METADATA_DEF(ONNX_TYPE_FLOAT16,     10, 2,                "")
+OM_TYPE_METADATA_DEF(ONNX_TYPE_DOUBLE,      11, sizeof(double),   "double")
+OM_TYPE_METADATA_DEF(ONNX_TYPE_UINT32,      12, sizeof(uint32_t), "uint32_t")
+OM_TYPE_METADATA_DEF(ONNX_TYPE_UINT64,      13, sizeof(uint64_t), "uint64_t")
+OM_TYPE_METADATA_DEF(ONNX_TYPE_COMPLEX64,   14, 8,                "_Complex float")
+OM_TYPE_METADATA_DEF(ONNX_TYPE_COMPLEX128,  15, 16,               "_Complex double")
+OM_TYPE_METADATA_DEF(ONNX_TYPE_BFLOAT16,    16, 2,                "")
 // clang-format on
 #else
 #error "Must define OM_TYPE_METADATA_DEF macro."
--- a/src/Conversion/KrnlToLLVM/KrnlPrint.cpp
+++ b/src/Conversion/KrnlToLLVM/KrnlPrint.cpp
@@ -1,0 +1,84 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//===------ KrnlPrint.cpp - Lower KrnlPrintOp -----------------------------===//
+//
+// Copyright 2022 The IBM Research Authors.
+//
+// =============================================================================
+//
+// This file lowers the KrnlPrintOp operator.
+//
+//===----------------------------------------------------------------------===//
+
+#include "mlir/Conversion/LLVMCommon/Pattern.h"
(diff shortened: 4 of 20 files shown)


etiotto:
TODO: document these operators in http://onnx.ai/onnx-mlir/DebuggingNumericalError.html

tungld:
Though it looks useful, I wonder if it is a common practice in MLIR/LLVM to have such utility functions in a dialect? Have a quick look at MLIR dialects, I don't find any similar functions. It's fine to have `print` for our purpose , but it makes me think again.

For a long term, we should think about whether we should mix ops for debugging and ops for "real run" in a dialect or not, given that there are multiple ways for debugging, for example, to print out a tensor, we simply edit the generated code to return a tensor as output.

chentong319:
I suggest to use the generate KrnlCallOp to handle the print.
We just need to pass  a memref to the runtime library, possibly with some print controls. Type of element is one of controls.
It is a surprise to me that OMTensor does not contain element type.

etiotto:
> Though it looks useful, I wonder if it is a common practice in MLIR/LLVM to have such utility functions in a dialect? Have a quick look at MLIR dialects, I don't find any similar functions. It's fine to have print for our purpose , but it makes me think again.

Yes there are no precedents, but I do not think in and for itself that prevents us to have this functionality. 

> For a long term, we should think about whether we should mix ops for debugging and ops for "real run" in a dialect or not, given that there are multiple ways for debugging, for example, to print out a tensor, we simply edit the generated code to return a tensor as output.

Good point. If we ever upstream the krnl dialect we might have to consolidate the operations that are the core of that dialect. That can be done in the future, note that there are other operations in the dialect that might be not part of the core.

etiotto:
> I suggest to use the general KrnlCallOp to handle the print.

That is possible, however we need to wait for that PR to land. I would not make this PR dependent on that one at this point.  There are other opcodes in the krnl dialect that will need to be replaced.  

> We just need to pass a memref to the runtime library, possibly with some print controls. If we can construct OMTensor at runtime library, we can print with shape and elementType.

This PR adds a new runtime API to print an OMTensor (`void omTensorPrint(const OMTensor *tensor);`), note that this new API is usable in the runtime itself, because it can be used by  other runtime APIs which often have a `OMTensor` available. A `memref` is not a key data structure in the runtime, that is the reason the omTensorPrint is built to receive a OMTensor pointer, which is the central data structure at the runtime level. 

> Current KrnlPrint passes only data pointer. It's just a optimization. We can pass OMTensor as a general practice to save coding, with a little bit extra overhead. With OMTensor, we can print the data along its shape.

I think there is some misunderstanding, but I think we fundamentally agree here because that is what this PR allows us to do. `KrnlPrintTensor` is lowered to a call to `omTensorPrint`, and passes a pointer to a `OMTensor` data structure built by using existing code `i.e. fillOMTensorWithMemRef`. 

The other opcode `KrnlPrintOp` is directly lowered to a printf call, and can be used to print any primitive value (e.g. a float, int, etc..). I found it very useful to help me trace what happens `inside` an operator at runtime.

etiotto:
@tungld @chentong319 I replied to your comments. Please take a look.

chentong319:
I misunderstood the purpose of the `FormatStr` in the PrintOMTensor. That string is used to uniquely tag the output, not the format string for data type of the OMTensor. Right?
I will reuse your code for type conversion (for memref and string) in the lowering of KrnlCallOp.

src/Dialect/Krnl/Krnl.td
```diff
@@ -1036,3 +1036,23 @@ def KrnlFindIndexOp : Op<Krnl_Dialect, "find_index",
...
+
+def KrnlPrintOp : Op<Krnl_Dialect, "print", [MemRefsNormalizable]> {
+  let summary = "Print a value.";
+  let description = [{
+    This operation can be used to print the input value. The user needs to provide a 
+    format string (à la printf) to specify how to print the input value. 
+    If the input value is not specified the operator will print the format string.
+  }];
+
+  let arguments = (ins StrAttr:$format, Optional<AnyType>:$input);
```
chentong319, line 1057: I feel that input can only be primitive types, not Memref, not struct or list. Is it better to pass the type and let the runtime to generate the format string if the format string is platform dependent (such 32 vs. 64).

src/Runtime/OMTensor.inc
```diff
@@ -344,6 +346,76 @@ void *omTensorGetAllocatedPtr(const OMTensor *tensor) {
...
+  OM_DATA_TYPE dataType = omTensorGetDataType(tensor);
+  int64_t rank = omTensorGetRank(tensor);
+  const int64_t *shape = omTensorGetShape(tensor);
+  const int64_t *strides = omTensorGetStrides(tensor);
+  void *dataPtr = omTensorGetDataPtr(tensor);
+
+  printf("\trank = %lld\n", (long long)rank);
+  printf("\tdataType = %s\n", getDataTypeName(dataType));
+
+  if (rank == 1) {
```
AlexandreEichenberger, line 359: would it be possible to add at least a 2D tensor too? If the format is Numpy arrays, that makes for easy integration to Jupyter notebook testing.

src/Conversion/KrnlToLLVM/KrnlPrint.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+#include "llvm/Support/Debug.h"
+
+#define DEBUG_TYPE "krnl_to_llvm"
+
+using namespace mlir;
+
+namespace onnx_mlir {
+
+static int64_t count = 0;
```
chentong319, line 34: Is a good practice to use a mutable static local variable in a pass? It is OK since this is not critical code.

include/onnx-mlir/Runtime/OnnxDataTypeMetaData.inc
```diff
@@ -3,28 +3,28 @@
...
+OM_TYPE_METADATA_DEF(ONNX_TYPE_INT64,       7,  sizeof(int64_t),  "int64_t")
+OM_TYPE_METADATA_DEF(ONNX_TYPE_STRING,      8,  0,                "const char *")
+OM_TYPE_METADATA_DEF(ONNX_TYPE_BOOL,        9,  sizeof(bool),     "_Bool")
+OM_TYPE_METADATA_DEF(ONNX_TYPE_FLOAT16,     10, 2,                "")
+OM_TYPE_METADATA_DEF(ONNX_TYPE_DOUBLE,      11, sizeof(double),   "double")
+OM_TYPE_METADATA_DEF(ONNX_TYPE_UINT32,      12, sizeof(uint32_t), "uint32_t")
+OM_TYPE_METADATA_DEF(ONNX_TYPE_UINT64,      13, sizeof(uint64_t), "uint64_t")
+OM_TYPE_METADATA_DEF(ONNX_TYPE_COMPLEX64,   14, 8,                "_Complex float")
+OM_TYPE_METADATA_DEF(ONNX_TYPE_COMPLEX128,  15, 16,               "_Complex double")
+OM_TYPE_METADATA_DEF(ONNX_TYPE_BFLOAT16,    16, 2,                "")
```
chentong319, line 27: You may consider to add the format string in this table.

src/Runtime/OMTensor.inc
chentong319, line 359: Why only rank == 1? I think we have the shape info for any rank.
Make sure that the output is in cvs format.

chentong319: Yes, Numpy for multiple dimension array.

src/Conversion/KrnlToLLVM/KrnlPrint.cpp
```diff
@@ -0,0 +1,113 @@
+/*
```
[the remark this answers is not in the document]
etiotto: I would rather have each operator in its own file (in fact I think we should split KrnlToLLVM.cpp) for every operator, similarly to what we have done for the ONNX-MLIR operators.  The advantages are:
  - easier to maintain because smaller code footprint can help reduce merge conflicts between PR in-flight
  - make/ninja can parallelize the compilation of the individual files

src/Conversion/KrnlToLLVM/KrnlPrint.cpp
etiotto: It is not for critical code because it might force serialization and global side effects. In this particular case I need to give a unique name to the format specifier used by the printf call and so I think this use is justifiable. Happy to change this if there is a suggestion to help avoid the static global var.

src/Dialect/Krnl/Krnl.td
etiotto: We lower the KrnlPrintOp to a printf call rather than a call to an API in the onnx-mlir runtime library.  If you want I can create the format string during lowering of the operator.  Some ppl might want to print a long long as a pointer though (so with %p rather than %lld) and the current operator would allow for that scenario.

src/Runtime/OMTensor.inc
etiotto: Fair request. I'll work on that.

etiotto: Fair question. It is a current arbitrary limitation because when I wrote this code I was debugging a case with a 1D tensor. I am thinking to extend it to support 2D tensors in this PR, and if we need a solution for any dimensions extend the code at that point?  Acceptable ?

docs/DebuggingNumericalError.md
```diff
@@ -82,3 +82,19 @@ optional arguments:
...
   --atol ATOL           Absolute tolerance for verification
 ```
+
+## Debugging the Code Generated for an Operator.
+
+If you know, or suspect, that a particular ONNX MLIR operator produces an incorrect result, and want to narrow down the problem, we provide a couple of useful Krnl operators that allow printing (at runtime) the value of a tensor, or a value that has a primitive data type. 
+
+To print out the value of a tensor at a particular program point, inject the following code (where `X` is the tensor to be printed):
+
+```code
```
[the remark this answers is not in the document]
AlexandreEichenberger: Best would be to enable at leas 0, 1, and 2 D. In all cases, specify what fully works here.

include/onnx-mlir/Runtime/OMTensor.h
```diff
@@ -248,6 +248,11 @@ static inline int64_t getDataTypeSize(OM_DATA_TYPE dataType) {
   return OM_DATA_TYPE_SIZE[dataType];
 }
 
+/* Helper function to get the ONNX data type C name string */
```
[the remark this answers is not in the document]
AlexandreEichenberger: Not very satisfying, but right now, we put all implementation in the .inc that is used in the C/C++ version, and the header in the OMTensorHelper.h

include/onnx-mlir/Runtime/OMTensor.h
```diff
@@ -284,6 +289,11 @@ int64_t omTensorGetOwning(const OMTensor *tensor);
  */
 void omTensorSetOwning(OMTensor *tensor, int64_t owning);
 
+/**
+ * \brief Print a OMTensor
+ */
```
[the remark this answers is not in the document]
AlexandreEichenberger: I think that would be good, I would want a label for all of my tensor prints.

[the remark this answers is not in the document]
AlexandreEichenberger: I also want to say that there is already support for vector.print
https://mlir.llvm.org/docs/Dialects/Vector/#vectorprint-mlirvectorprintop
but what you have is good too.

src/Runtime/OMTensor.inc
AlexandreEichenberger: sure.

src/Conversion/KrnlToLLVM/KrnlPrint.cpp
etiotto: I ended up finding a way to remove the global static variable.

include/onnx-mlir/Runtime/OMTensor.h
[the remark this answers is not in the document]
etiotto: ok that is good and should assuage @tungld concern that there are no precedents for print like operation in MLIR.

docs/DebuggingNumericalError.md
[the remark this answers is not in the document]
etiotto: done

include/onnx-mlir/Runtime/OnnxDataTypeMetaData.inc
etiotto: the format is not longer needed, I now use std::cout to print the `OMTensor` data.

src/Runtime/OMTensor.inc
etiotto: Done, we can now print `OMTensor` up to rank <= 3.

etiotto: Done, we can now print OMTensor up to rank <= 3.

etiotto:
I believe I have addressed the comments received so far. Please take another look.

src/Runtime/OMTensor.inc
etiotto: The data is printed out like this: 

```
Test 1D
        rank = 1
        dataType = int64_t
        numElems = 3
        strides: [1]
        data: ([1, 2, 3])
Test 2D
        rank = 2
        dataType = float
        numElems = 6
        strides: [2][1]
        data: ([[1, 2], [3, 4], [5, 6]])
Test 3D
        rank = 3
        dataType = float
        numElems = 6
        strides: [2][1][1]
        data: ([[[1], [2]], [[3], [4]], [[5], [6]]])

Comment (AlexandreEichenberger) [reply src/Runtime/OMTensor.inc]: nice, we can then just substitute `data:` for `np.array` and the "()" are already there (which is very very convenient when the arrays has 10k elements...)

File: src/Conversion/KrnlToLLVM/KrnlPrint.cpp (hunk 5 above)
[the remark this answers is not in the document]
Comment (AlexandreEichenberger) [reply src/Conversion/KrnlToLLVM/KrnlPrint.cpp:1]: ok, thanks for the justification

File: docs/DebuggingNumericalError.md [hunk 9]
```diff
@@ -82,3 +82,21 @@ optional arguments:
...
+create.krnl.printTensor("Tensor: ", X)
```
AlexandreEichenberger, line 93: Nit `printTensor("Tensor X: ", X)`

src/Runtime/OMTensorHelper.h
```diff
@@ -162,4 +211,12 @@ template <typename T>
 bool omTensorAreTwoOmtsClose(
     const OMTensor *a, const OMTensor *b, float rtol = 1e-5, float atol = 1e-5);
 
+/**
```
AlexandreEichenberger, line 214: Maybe I was confusing, I would suggest to move this one to the official interface as I think it will provide a convenient add-on to the services we currently have.  I did not care for some of the support ops, this one is potentially useful to users of the interface too.

You should add in the description if the printing goes to stdout or stderr.

AlexandreEichenberger (approved):
LGTM pending small suggested changes

src/Runtime/OMTensorHelper.h
etiotto: done

include/onnx-mlir/Runtime/OnnxDataTypeMetaData.inc
chentong319: Even for c interface?

chentong319 (approved):
LGTM!

PR #1181, opened 2022-02-15
Build Performance testing

AlexandreEichenberger:
Added support to do performance testing. A new target `make check-onnx-perf` was added, which run several short performance tests and report on their performances. They write their results in files labeled `<benchmark-name>.perf.txt`. They can also be individually run:
``` shell
root@1250b049ac2b:~/onnx-mlir/build# Debug/bin/GemmPerf 
2022-02-15T22:03:43+00:00
Running Debug/bin/GemmPerf
Run on (14 X 2300 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x14)
  L1 Instruction 32 KiB (x14)
  L2 Unified 256 KiB (x14)
  L3 Unified 16384 KiB (x14)
Load Average: 0.50, 1.01, 0.64
***WARNING*** Library was built as DEBUG. Timings may be affected.
---------------------------------------------------------------
Benchmark                     Time             CPU   Iterations
---------------------------------------------------------------
BM_MatmulSquare/16        0.001 ms        0.001 ms      1077157
BM_MatmulSquare/32        0.003 ms        0.003 ms       245783
BM_MatmulSquare/64        0.022 ms        0.022 ms        31946
BM_MatmulSquare/128       0.174 ms        0.174 ms         3734
BM_MatmulSquare/256        2.13 ms         2.13 ms          341
BM_MatmulSquare/512        20.5 ms         20.4 ms           31
BM_MatmulSquare/1024        178 ms          178 ms            4
BM_MatmulSquare_BigO       0.17 N^3        0.17 N^3  
BM_MatmulSquare_RMS           3 %             3 %    
BM_GemmSquare/16          0.001 ms        0.001 ms       923459
BM_GemmSquare/32          0.003 ms        0.003 ms       202752
BM_GemmSquare/64          0.022 ms        0.022 ms        31291
BM_GemmSquare/128         0.189 ms        0.189 ms         3950
BM_GemmSquare/256          1.45 ms         1.45 ms          467
BM_GemmSquare/512          11.6 ms         11.6 ms           58
BM_GemmSquare/1024         96.8 ms         96.7 ms            7
BM_GemmSquare_BigO         0.09 N^3        0.09 N^3  
BM_GemmSquare_RMS             1 %             1 %    
```
This is not going to be the final version of the tests, but the first attempt to lay down the foundations of performance testing. We will need additional support to compare the performance to recorded performances for the appropriate machine. There are scripts available for comparions. Additionally, we need to add more perf testing for RNNs as well as customizing the tests to include important sizes.

In general, adding a new tests is trivial using the infrastructure of the model library in `test/modlib`. For example, below is all of the code needed to test for several matrix multiplication sizes:
```c++
static void BM_MatmulSquare(benchmark::State &state) {
  int IJK = state.range(0);
  MatMul2DLibBuilder model(modelName, IJK, IJK, IJK);
  assert(model.build() && model.compileAndLoad(opts) && model.prepareInputs() &&
         "failed matmul");
  for (auto _ : state)
    model.run();
  state.SetComplexityN(IJK);
}
BENCHMARK(BM_MatmulSquare)
    ->RangeMultiplier(2)
    ->Range(16, 1024)
    ->Unit(benchmark::kMillisecond)
    ->Complexity();
```
More documentation is found here https://github.com/google/benchmark


tungld:
> ***WARNING*** Library was built as DEBUG. Timings may be affected.

Looks like DEBUG here is for building onnx-mlir and it does not affect the generated code. In other words, it does not affect the inference time, does it?

test/perf/PerfConv.cpp
```diff
@@ -0,0 +1,47 @@
...
+//
+// Copyright 2022 The IBM Research Authors.
+//
+// =============================================================================
+//
+// This file contains tests for simple test cases,  for an arbitrary small set
+// of parameters.
+//===----------------------------------------------------------------------===//
+
+#include <assert.h>
```
etiotto, line 15: nit: use `cassert`

test/perf/PerfGemm.cpp
```diff
@@ -0,0 +1,56 @@
...
+//
+// Copyright 2022 The IBM Research Authors.
+//
+// =============================================================================
+//
+// This file contains tests for simple test cases, like for an arbitrary small
+// set of parameters.
+//===----------------------------------------------------------------------===//
+
+#include <assert.h>
```
etiotto, line 15: nit: use cassert

etiotto (approved):
LGTM - a couple of nits (these are optional changes if you have to make other changes).

test/perf/PerfConv.cpp
```diff
@@ -0,0 +1,47 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//====-- PerfConv.hpp - Simple Conv performance tests -===//
```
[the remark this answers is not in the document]
AlexandreEichenberger: done

test/perf/PerfGemm.cpp
```diff
@@ -0,0 +1,56 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//====-- PerfGemm.hpp - Simple performance tests -===//
```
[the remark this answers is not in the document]
AlexandreEichenberger: done

test/perf/PerfConv.cpp
AlexandreEichenberger: cool, like it

test/perf/PerfGemm.cpp
AlexandreEichenberger: done too, tx

AlexandreEichenberger:
> Looks like DEBUG here is for building onnx-mlir and it does not affect the generated code. In other words, it does not affect the inference time, does it?

I think it is because we build our own (mine) work repo in the Debug mode, as opposed to the Release mode. We should be able to eventually invoke the `make check-onnx-perf` in our CIs only after building the Release version of our compiler... though my understanding at this time is that while it may change the performance of the compiler, the code generated is the same.... So I now agree that it does not matter for us, unless we are timing the compiler (which we may want to have a benchmark test doing that too)

AlexandreEichenberger:
Maybe someone on the Window/MS can try to help me determine why the building of the numerical benchmark now fails on windows:
```
C:\PROGRA~2\MICROS~1\2019\ENTERP~1\VC\Tools\MSVC\1429~1.301\bin\Hostx64\x64\cl.exe  /nologo /TP -DONNX_ML=1 -DONNX_NAMESPACE=onnx -DUNICODE -D_CRT_NONSTDC_NO_DEPRECATE -D_CRT_NONSTDC_NO_WARNINGS -D_CRT_SECURE_NO_DEPRECATE -D_CRT_SECURE_NO_WARNINGS -D_SCL_SECURE_NO_DEPRECATE -D_SCL_SECURE_NO_WARNINGS -D_UNICODE -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -ID:\a\1\llvm-project\build\include -ID:\a\1\onnx-mlir -ID:\a\1\onnx-mlir\build -ID:\a\1\onnx-mlir\include -ID:\a\1\onnx-mlir\third_party\onnx -ID:\a\1\onnx-mlir\build\third_party\onnx -external:ID:\a\1\protobuf_install\include -external:W0 /DWIN32 /D_WINDOWS   -wd4125 /Zc:inline /Zc:__cplusplus /Zc:strictStrings /Oi /Zc:rvalueCast /bigobj /W4 -wd4141 -wd4146 -wd4244 -wd4267 -wd4291 -wd4351 -wd4456 -wd4457 -wd4458 -wd4459 -wd4503 -wd4624 -wd4722 -wd4100 -wd4127 -wd4512 -wd4505 -wd4610 -wd4510 -wd4702 -wd4245 -wd4706 -wd4310 -wd4701 -wd4703 -wd4389 -wd4611 -wd4805 -wd4204 -wd4577 -wd4091 -wd4592 -wd4319 -wd4709 -wd4324 -w14062 -we4238 /Gw -DSUPPRESS_THIRD_PARTY_WARNINGS /MD /O2 /Ob2   -D_CRT_SECURE_NO_DEPRECATE -D_CRT_SECURE_NO_WARNINGS -D_CRT_NONSTDC_NO_DEPRECATE -D_CRT_NONSTDC_NO_WARNINGS -D_SCL_SECURE_NO_DEPRECATE -D_SCL_SECURE_NO_WARNINGS -DUNICODE -D_UNICODE -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS  /EHsc /GR -UNDEBUG -std:c++14 /showIncludes /Fosrc\Compiler\CMakeFiles\OnnxMlirCompiler.dir\OnnxMlirCompiler.cpp.obj /Fdsrc\Compiler\CMakeFiles\OnnxMlirCompiler.dir\OnnxMlirCompiler.pdb /FS -c D:\a\1\onnx-mlir\src\Compiler\OnnxMlirCompiler.cpp
D:\a\1\onnx-mlir\src\Compiler\OnnxMlirCompiler.cpp(11): error C2491: 'onnx_mlir::omSetCompilerOptionsFromEnv': definition of dllimport function not allowed
D:\a\1\onnx-mlir\src\Compiler\OnnxMlirCompiler.cpp(23): error C2491: 'onnx_mlir::omSetCompilerOptionsFromArgs': definition of dllimport function not allowed
D:\a\1\onnx-mlir\src\Compiler\OnnxMlirCompiler.cpp(30): error C2491: 'onnx_mlir::omSetCompilerOptionsFromArgsAndEnv': definition of dllimport function not allowed
D:\a\1\onnx-mlir\src\Compiler\OnnxMlirCompiler.cpp(38): error C2491: 'onnx_mlir::omSetCompilerOptions': definition of dllimport function not allowed
D:\a\1\onnx-mlir\src\Compiler\OnnxMlirCompiler.cpp(42): error C2491: 'onnx_mlir::omGetCompilerOption': definition of dllimport function not allowed
D:\a\1\onnx-mlir\src\Compiler\OnnxMlirCompiler.cpp(49): error C2491: 'onnx_mlir::omCompileFromFile': definition of dllimport function not allowed
D:\a\1\onnx-mlir\src\Compiler\OnnxMlirCompiler.cpp(65): error C2491: 'onnx_mlir::omCompileFromArray': definition of dllimport function not allowed
```

It is probably something simple that changed, I will investigate, but if @sstamenova or other MS folk know this error well, feel free to give me a hint, thanks

gongsu832:
You probably need a file similar to `test/numerical/numerical.def`

AlexandreEichenberger:
I have a def in test/perf which is the same as in test/numerical

I edited my original message to show when the error occurs, it is while building the `src/Compiler/CMakeFiles/OnnxMlirCompiler.dir/OnnxMlirCompiler.cpp.obj` object, namely much earlier.

gongsu832:
On Windows, the "install onnx-mlir" part only builds onnx-mlir binary. Therefore `libOnnxMlirCompiler.a` was never built since it's not a prereq for onnx-mlir binary. On Linux, we do `make` instead of `make onnx-mlir` so dependencies for all tests, including `libOnnxMlirCompiler.a`, were built.

sstamenova:
```
D:\a\1\onnx-mlir\src\Compiler\OnnxMlirCompiler.cpp(11): error C2491: 'onnx_mlir::omSetCompilerOptionsFromEnv': definition of dllimport function not allowed
```

I have seen this error when we try to build the `OnnxMlirCompiler` library as a static library. We use it as a shared library, so in our repo we've modified it to still say `SHARED` the way it was originally setup here (https://github.com/onnx/onnx-mlir/pull/894). The change that @gongsu832 made here (https://github.com/onnx/onnx-mlir/pull/1039/files) results in this particular failure on Windows.

My recommendation is two prong:
1) Change the library to always be `SHARED`, perhaps just on Windows, but perhaps always. The library is mostly useful as `SHARED` on Windows and not static.
2) Change the Windows builds to do the equivalent of `make all`. I believe at this point we've fixed all (or almost all) the Windows issues that were preventing that from happening

1) should be pretty straightforward to do, I can look at 2) next week.

AlexandreEichenberger:
@sstamenova I added the building of everything and that did not change the error, as far as I am concerned the dependences work so when building for a specific target, all is built for it anyway.

Can I ask you to look at this? You implied it is something easy, I am sure its a setting somewhere, or a `#define` to add to a makefile...
Thanks

gongsu832:
According to `include/OnnxMlirCompiler.h`, in order to compile `src/Compiler/OnnxMlirCompiler.cpp` as part of `libOnnxMlirCompiler.a` on Windows, you need to define `OnnxMlirCompiler_EXPORTS` so you get `__declspec(dllexport)` instead of `__declspec(dllimport)`. So I guess `src/Compiler/CMakeLists.txt` needs to include a definition for `OnnxMlirCompiler_EXPORTS` in the `add_onnx_mlir_library(OnnxMlirCompiler ...)` clause.

AlexandreEichenberger:
@sstamenova I manually copied your suggested changes from #1188 into this PR and it appears to fail in the doc example. Wonder if it is because the python script needs ONNX.

``` shell
FAILED: docs/doc_example/add.dll docs/doc_example/add.lib D:/a/1/onnx-mlir/build/docs/doc_example/add.dll D:/a/1/onnx-mlir/build/docs/doc_example/add.lib 
cmd.exe /C "cd /D D:\a\1\onnx-mlir\build\docs\doc_example && C:\hostedtoolcache\windows\Python\3.9.9\x64\python3.exe D:/a/1/onnx-mlir/build/docs/doc_example/gen_add_onnx.py && D:\a\1\onnx-mlir\build\Release\bin\onnx-mlir.exe add.onnx"
  File "D:\a\1\onnx-mlir\build\docs\doc_example\gen_add_onnx.py", line 1, in <module>
    import onnx
ModuleNotFoundError: No module named 'onnx'
```

Will try to see if I build less it will get me to succeed.

sstamenova:
> @sstamenova I manually copied your suggested changes from #1188 into this PR and it appears to fail in the doc example. Wonder if it is because the python script needs ONNX.
> 
> ```shell
> FAILED: docs/doc_example/add.dll docs/doc_example/add.lib D:/a/1/onnx-mlir/build/docs/doc_example/add.dll D:/a/1/onnx-mlir/build/docs/doc_example/add.lib 
> cmd.exe /C "cd /D D:\a\1\onnx-mlir\build\docs\doc_example && C:\hostedtoolcache\windows\Python\3.9.9\x64\python3.exe D:/a/1/onnx-mlir/build/docs/doc_example/gen_add_onnx.py && D:\a\1\onnx-mlir\build\Release\bin\onnx-mlir.exe add.onnx"
> Traceback (most recent call last):
>   File "D:\a\1\onnx-mlir\build\docs\doc_example\gen_add_onnx.py", line 1, in <module>
>     import onnx
> ModuleNotFoundError: No module named 'onnx'
> ```
> 
> Will try to see if I build less it will get me to succeed.

That's because I forgot to make sure the last options is on the same "line". It should work with the update.

AlexandreEichenberger:

PR #1182, opened 2022-02-16
added pow op verifier

thomas-m1:
Signed-off-by: root <thomas_michalski@hotmail.com>

added verifier for pow op


src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -1140,23 +1140,34 @@ LogicalResult ONNXErfOp::inferShapes(
...
-  if (!getOperand(0).getType().isa<RankedTensorType>() ||
-      !getOperand(1).getType().isa<RankedTensorType>())
-    return success();
-  RankedTensorType lhsTy = getOperand(0).getType().cast<RankedTensorType>();
-  RankedTensorType rhsTy = getOperand(1).getType().cast<RankedTensorType>();
+static LogicalResult verify(ONNXPowOp op) {
+  ONNXPowOp operandAdaptor = ONNXPowOp(op);
+  
+  RankedTensorType lhsTy = operandAdaptor.X().getType().cast<RankedTensorType>();
+  RankedTensorType rhsTy = operandAdaptor.Y().getType().cast<RankedTensorType>();
```
chentong319, line 1147: The type for input X and Y could be tensor or memref. The type is tensor when shape inference is called. When you move the check out to verifier, the memref type has to be taken into consideration.
RankedTensorType and MemRef are subclass of ShapedType. You should cast the type to ShapedType and then getElementType() to handle both.

chentong319 (commented):
For format error, you can either use clang-format or change your file according to the build result.

src/Dialect/ONNX/ONNXOps.td.inc
```diff
@@ -3812,6 +3812,7 @@ def ONNXPowOp:ONNX_Op<"Pow",
         return {20};
       }
     }];
+    let verifier = [{ return ::verify(*this); }];
```
sstamenova, line 3815: You should use the `hasVerifier` version instead. This usage of `verifier` will be obsolete soon.

sstamenova (changes requested):

Issue #1183, opened 2022-02-16
Support multiple entry points


tungld:
Currently, onnx-mlir has not yet fully supported multiple entry points in a model, e.g. We assume there is only a single entry point, that is `run_main_graph` and provide `omInputSignature()` and `omOutputSignature()` functions for querying the input/output of the entry point. However, we would like to extend these signature functions in order to support multiple entry points.

Example of a multi-entry-point code:
```mlir
module {
  func private @test_add(%arg0 : tensor<10xf32>, %arg1 : tensor<10xf32>) -> tensor<*xf32> {
    %0 ="onnx.Add"(%arg0, %arg1) : (tensor<10xf32>, tensor<10xf32>) -> tensor<*xf32>
    "std.return"(%0) : (tensor<*xf32>) -> ()
  }
  func private @test_relu(%arg0 : tensor<10xf32>) -> tensor<*xf32> {
    %0 ="onnx.Relu"(%arg0) : (tensor<10xf32>) -> tensor<*xf32>
    "std.return"(%0) : (tensor<*xf32>) -> ()
  }
  "onnx.EntryPoint"() {func = @test_add, numInputs = 2 : i32, numOutputs = 1 : i32, signature = ""} : () -> ()
  "onnx.EntryPoint"() {func = @test_relu, numInputs = 1 : i32, numOutputs = 1 : i32, signature = ""} : () -> ()
}
```

There are two approaches to supporting multiple entry points:
1. Reuse function names `omInputSignature` and `omOutputSignature`, and allow users to pass an entry point name to query its signature, e.g. `omInputSignature("run_test_add")`.
2. For each entry point name, e,g. `run_test_add`, create two functions `run_test_add_in_signature()` and `run_test_add_out_signature()` to query its signatures.

Though approach 1 reuses function names, we cannot reuse the function content and we need to generate complicated code, e.g. generating blocks and condition branches at LLVM level for searching the signature of the given entry point name.

Meanwhile, approach 2 does not reuse the function names but it is straightforward to implement. We simply use a dynamic name, e.g `run_test_add_in_signature` depending on the entry point name, instead of a fixed name `omInputSignature`. There is no change to the function content and no searching needed.

I prefer approach 2. However, I am a bit concerned about changing the function names since onnx-mlir users are using those names.

gongsu832:
[...]

PR #1184, opened 2022-02-16
Support multiple entry points

tungld:
This patch adds support for multiple entry points by adding the following C functions:
-  `const char *omInputSignature(entry_point)` and `const char *omOutputSignature(entry_point)` to query the input/output signature of the given entry point, respectively.
- `const char **omQueryEntryPoints()` to query all entry point names in a model. It returns an array of strings, e.g. `["run_add", "run_sub", NULL]`. NULL is added as a terminator.

Example with C++ ExecutionSession:
```C
// Model has a single entry point. Entry point defaults to `run_main_graph`.
ExecutionSession sess = ExecutionSession(sharedLib);
std::string inSig = sess.inputSignature();
sess.run(...)

// Model has multiple entry points.
ExecutionSession sess = ExecutionSession(sharedLib, /*defaultEntryPoint=*/false);
std::string* entryPoints = sess.queryEntryPoints();
//
sess.setEntryPoint(entryPoints[0]);
std::string inSig = sess.inputSignature(); // query the input signature of the current entry point.
sess.run(...)
//
sess.setEntryPoint(entryPoints[1]);
std::string inSig = sess.inputSignature(); // query the input signature of the current entry point.
sess.run(...)
```

Python ExecutionSession is similar to C++ ExecutionSession.

This patch does not add a JNI wrapper for `omQueryEntryPoints`.

### DIFF
--- a/include/onnx-mlir/Runtime/OMEntryPoint.h
+++ b/include/onnx-mlir/Runtime/OMEntryPoint.h
@@ -1,0 +1,31 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//===----------- OMEntryPoint.h - OMEntryPoint Declaration header ---------===//
+//
+// Copyright 2022 The IBM Research Authors.
+//
+// =============================================================================
+//
+// This file contains declaration of OMEntryPoint API functions.
+//
+//===----------------------------------------------------------------------===//
+
+#ifdef __cplusplus
+#pragma once
+
+extern "C" {
+#endif
+
+/**
+ * \brief Return all entry point names in a model.
+ *
+ * @return an array of strings. The array ends with NULL. For example:
+ * ["run_add", "run_sub", NULL].
+ */
+const char **omQueryEntryPoints();
+
+#ifdef __cplusplus
+}
+#endif
--- a/include/onnx-mlir/Runtime/OMEntryPoint.h
+++ b/include/onnx-mlir/Runtime/OMEntryPoint.h
@@ -19,7 +19,12 @@ extern "C" {
 #endif
 
 /**
- * \brief Return all entry point names in a model.
+ * \brief Return all entry point names in a model. These entry point names are
+ * the symbols of the inference functions in the model. Users use them to run
+ * inference, e.g. by calling `entry_point_name(OMTensorList).
+ *
+ * An entry point name can be passed to functions `omInputSignature` and
+ * `omOutputSignature` to query its input and output signatures, respectively.
  *
  * @return an array of strings. The array ends with NULL. For example:
  * ["run_add", "run_sub", NULL].
--- a/include/onnx-mlir/Runtime/OMEntryPoint.h
+++ b/include/onnx-mlir/Runtime/OMEntryPoint.h
@@ -24,7 +24,7 @@ extern "C" {
  * @return an array of strings. The array ends with NULL. For example:
  * ["run_add", "run_sub", NULL].
  */
-const char **omQueryEntryPoints();
+const char *const *omQueryEntryPoints();
 
 #ifdef __cplusplus
 }
--- a/include/onnx-mlir/Runtime/OMSignature.h
+++ b/include/onnx-mlir/Runtime/OMSignature.h
@@ -19,22 +19,20 @@ extern "C" {
 #endif
 
 /**
- * \brief Return the model's input signature as a JSON string for the given
- * entry point name.
+ * \brief Return the input signature of the given entry point as a JSON string.
  *
  * @param entry point name
  * @return pointer to input signature JSON string
  */
-const char *omInputSignature(const char *);
+const char *omInputSignature(const char *entryPointName);
 
 /**
- * \brief Return the model's output signature as a JSON string for the given
- * entry point name.
+ * \brief Return the output signature of the given entry point as a JSON string.
  *
  * @param entry point name
  * @return pointer to output signature JSON string
  */
-const char *omOutputSignature(const char *);
+const char *omOutputSignature(const char *entryPointName);
 
 #ifdef __cplusplus
 }
--- a/include/onnx-mlir/Runtime/OMSignature.h
+++ b/include/onnx-mlir/Runtime/OMSignature.h
@@ -23,14 +23,14 @@ extern "C" {
  *
  * @return pointer to input signature JSON string
  */
-const char *omInputSignature();
+const char *omInputSignature(const char *);
 
 /**
  * \brief Return the model's output signature as a JSON string.
  *
  * @return pointer to output signature JSON string
  */
-const char *omOutputSignature();
+const char *omOutputSignature(const char *);
 
 #ifdef __cplusplus
 }
--- a/include/onnx-mlir/Runtime/OMSignature.h
+++ b/include/onnx-mlir/Runtime/OMSignature.h
@@ -19,15 +19,19 @@ extern "C" {
 #endif
 
 /**
- * \brief Return the model's input signature as a JSON string.
+ * \brief Return the model's input signature as a JSON string for the given
+ * entry point name.
  *
+ * @param entry point name
  * @return pointer to input signature JSON string
  */
 const char *omInputSignature(const char *);
 
 /**
- * \brief Return the model's output signature as a JSON string.
+ * \brief Return the model's output signature as a JSON string for the given
+ * entry point name.
  *
+ * @param entry point name
  * @return pointer to output signature JSON string
  */
 const char *omOutputSignature(const char *);
--- a/src/Conversion/KrnlToLLVM/KrnlToLLVM.cpp
+++ b/src/Conversion/KrnlToLLVM/KrnlToLLVM.cpp
@@ -1886,8 +1886,9 @@ void mlir::recordEntryPointSignatures(ModuleOp &module,
 }
 
 void mlir::genSignatureFunction(ModuleOp module,
-    ArrayRef<std::string> entryPointNames, ArrayRef<std::string> inSignatures,
-    ArrayRef<std::string> outSignatures) {
+    const ArrayRef<std::string> entryPointNames,
+    const ArrayRef<std::string> inSignatures,
+    const ArrayRef<std::string> outSignatures) {
   MLIRContext *context = module.getContext();
   Location loc = module.getLoc();
   OpBuilder b(context);
@@ -1934,7 +1935,7 @@ void mlir::genSignatureFunction(ModuleOp module,
         /*isConstant=*/true, LLVM::Linkage::External, inSigVarName, inSigAttr);
     inSigOps.emplace_back(inSigOp);
 
-    // Global constants for input signatures.
+    // Global constants for output signatures.
     mlir::StringAttr outSigAttr =
         mlir::StringAttr::get(context, outSignatures[i]);
     Type outSigArrayType =
--- a/src/Conversion/KrnlToLLVM/KrnlToLLVM.cpp
+++ b/src/Conversion/KrnlToLLVM/KrnlToLLVM.cpp
@@ -43,6 +43,7 @@
 #include "onnx/onnx_pb.h"
 
 #include "src/Conversion/KrnlToLLVM/KrnlToLLVM.hpp"
+#include "src/Conversion/KrnlToLLVM/RuntimeAPI.hpp"
 #include "src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.hpp"
 #include "src/Dialect/Krnl/KrnlOps.hpp"
 #include "src/Pass/Passes.hpp"
@@ -107,21 +108,6 @@ static onnx::TensorProto::DataType mlirTypeToOnnxType(mlir::Type elemType) {
   }
 
   return onnxType;
-}
-
-static FlatSymbolRefAttr getOrInsertExternFunc(StringRef funcName,
-    ModuleOp module, mlir::Type funcType, PatternRewriter &rewriter) {
-  auto *context = module.getContext();
-  if (auto sym = module.lookupSymbol<LLVM::LLVMFuncOp>(funcName)) {
-    assert(sym.getType() == funcType && "wrong symbol type");
-    return SymbolRefAttr::get(context, funcName);
-  }
-
-  // Insert the function into the body of the parent module.
-  PatternRewriter::InsertionGuard insertGuard(rewriter);
-  rewriter.setInsertionPointToStart(module.getBody());
-  rewriter.create<LLVM::LLVMFuncOp>(module.getLoc(), funcName, funcType);
-  return SymbolRefAttr::get(context, funcName);
 }
 
 static int64_t getRankFromMemRefType(LLVM::LLVMStructType memRefTy) {
@@ -1091,36 +1077,6 @@ public:
       : OpRewritePattern<KrnlEntryPointOp>(ctx),
         constantOutputs(constantOutputs) {}
 
-  enum class API {
-    CREATE_OMTENSOR_LIST,
-    CREATE_OMTENSOR,
-    GET_DATA,
-    SET_DATA,
-    GET_DATA_SHAPE,
-    GET_DATA_STRIDES,
-    SET_DATA_TYPE,
-    GET_DATA_TYPE,
-    GET_OMT_ARRAY,
-  };
-
-  struct ApiSpec {
-    API id;
-    std::string name;
-    FlatSymbolRefAttr symbolRef;
-    Type outputTy;
-    SmallVector<Type, 4> inputTys;
-
-    ApiSpec(
-        API id, const std::string &name, Type outputTy, ArrayRef<Type> inputTys)
-        : id(id), name(name), outputTy(outputTy),
-          inputTys(inputTys.begin(), inputTys.end()) {}
-
-    Type funcTy() {
-      return LLVM::LLVMFunctionType::get(outputTy, inputTys,
-          /*isVarArg=*/false);
-    }
-  };
-
   static void genSignatureFunction(PatternRewriter &rewriter,
       MLIRContext *context, std::string funcName, LLVM::GlobalOp sigvar,
       Location loc) {
@@ -1149,7 +1105,8 @@ public:
 
     auto module = op->getParentOfType<ModuleOp>();
     auto *context = module.getContext();
-    auto apiRegistry = RegisterAllApis(module, rewriter);
+    const RuntimeAPIRegistry &apiRegistry =
+        RuntimeAPIRegistry::build(module, rewriter);
     auto loc = op.getLoc();
     auto numOutputs = op->getAttrOfType<IntegerAttr>(
                             KrnlEntryPointOp::getNumOutputsAttrName())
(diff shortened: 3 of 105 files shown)


gongsu832:
@tungld With multiple entries, should we have an API call to query the list of entry names? e.g., something like
```
const char **omQueryEntry();
```
which returns an array of strings terminated by NULL, e.g., something like
```
{ "run_main_graph", "test_add", "test_relu", NULL };
```

tungld:
> @tungld With multiple entries, should we have an API call to query the list of entry names?

Good idea. I will implement that API. Just thinking whether we should return an array of strings or a JSON string. JSON is using for signatures, so it looks more consistent.

src/Conversion/KrnlToLLVM/KrnlToLLVM.cpp
```diff
@@ -1895,6 +1849,202 @@ void mlir::checkConstantOutputs(
...
+  // When there is only a single entry point function in a model, use
+  // DEFAULT_DYN_ENTRY_POINT.
+  if (entryPointNames.size() == 1) {
+    entryPointNames[0] = DEFAULT_DYN_ENTRY_POINT;
+    entryPointNames[0].push_back('\0'); // null terminate the string.
+  }
+}
+
+void mlir::genSignatureFunction(ModuleOp module,
+    ArrayRef<std::string> entryPointNames, ArrayRef<std::string> inSignatures,
```
etiotto, line 1889: `entryPointNames`, `inSignatures ` and `outSignatures ` should not be modified by this function, suggest to make them const.

src/Conversion/KrnlToLLVM/KrnlToLLVM.cpp
```diff
@@ -1895,6 +1849,202 @@ void mlir::checkConstantOutputs(
...
+    mlir::StringAttr inSigAttr =
+        mlir::StringAttr::get(context, inSignatures[i]);
+    Type inSigArrayType =
+        LLVM::LLVMArrayType::get(i8Type, inSignatures[i].size());
+    std::string inSigVarName = entryVarName + "_in_sig";
+    LLVM::GlobalOp inSigOp = b.create<LLVM::GlobalOp>(loc, inSigArrayType,
+        /*isConstant=*/true, LLVM::Linkage::External, inSigVarName, inSigAttr);
+    inSigOps.emplace_back(inSigOp);
+
+    // Global constants for input signatures.
```
etiotto, line 1937: "input signatures" -> "output signatures"

src/Conversion/KrnlToLLVM/KrnlToLLVM.cpp
```diff
@@ -1895,6 +1849,202 @@ void mlir::checkConstantOutputs(
...
+    mlir::StringAttr inSigAttr =
+        mlir::StringAttr::get(context, inSignatures[i]);
+    Type inSigArrayType =
+        LLVM::LLVMArrayType::get(i8Type, inSignatures[i].size());
+    std::string inSigVarName = entryVarName + "_in_sig";
+    LLVM::GlobalOp inSigOp = b.create<LLVM::GlobalOp>(loc, inSigArrayType,
+        /*isConstant=*/true, LLVM::Linkage::External, inSigVarName, inSigAttr);
+    inSigOps.emplace_back(inSigOp);
```
etiotto, line 1935: This code section looks similar to the one above (lines 1918-1925) and below (lines 1938-1946). Suggest to create a lambda or a templated function to avoid code duplication.

src/Runtime/ExecutionSession.cpp
```diff
@@ -96,10 +92,12 @@ OMTensorList *ExecutionSession::run(OMTensorList *input) {
   return _entryPointFunc(input);
 }
 
-std::string ExecutionSession::inputSignature() { return _inputSignatureFunc(); }
+std::string ExecutionSession::inputSignature() {
```
etiotto, line 95: ExecutionSession::inputSignature() --> ExecutionSession::inputSignature() const. Similar comment for `ExecutionSession::outputSignature()`

src/Runtime/ExecutionSession.hpp
```diff
@@ -25,7 +25,7 @@
 namespace onnx_mlir {
 
 typedef OMTensorList *(*entryPointFuncType)(OMTensorList *);
-typedef const char *(*signatureFuncType)();
+typedef const char *(*signatureFuncType)(const char *);
```
etiotto, line 28: Suggest to use a using declaration rather than typedef, given this is C++ code.

src/Runtime/PyExecutionSession.cpp
```diff
@@ -162,6 +162,24 @@ std::vector<py::array> PyExecutionSession::pyRun(
   return outputPyArrays;
 }
 
+void PyExecutionSession::pySetEntryPoint(std::string entryPointName) {
+  setEntryPoint(entryPointName);
+  return;
+}
+
+std::vector<std::string> PyExecutionSession::pyQueryEntryPoint() {
```
gongsu832, line 170: `ExecutionSession.{c|h}pp` use `queryEntryPoints()` so this should be `pyQueryEntryPoints()` to be consistent.

tungld: Yes. Thanks!

tungld:
Addressed all comments from Gong and Ettore. Thanks!

gongsu832:
> @gongsu832 I do not write a JNI wrapper for omQueryEntryPoints. I think you are the best person to do that. Thanks!

@tungld Yes I will add the call for Java API. Thanks.

src/Runtime/PyExecutionSession.cpp
```diff
@@ -162,6 +162,24 @@ std::vector<py::array> PyExecutionSession::pyRun(
   return outputPyArrays;
 }
 
+void PyExecutionSession::pySetEntryPoint(std::string entryPointName) {
+  setEntryPoint(entryPointName);
+  return;
```
etiotto, line 167: no need to return here

include/onnx-mlir/Runtime/OMEntryPoint.h
```diff
@@ -0,0 +1,31 @@
...
+extern "C" {
+#endif
+
+/**
+ * \brief Return all entry point names in a model.
+ *
+ * @return an array of strings. The array ends with NULL. For example:
+ * ["run_add", "run_sub", NULL].
+ */
+const char **omQueryEntryPoints();
```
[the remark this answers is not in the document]
gongsu832: NULL terminated string array is fairly common. Unless there are many use cases where you only want the number of entry points but not the entry points themselves, this would be the simplest API.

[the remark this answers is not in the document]
AlexandreEichenberger: It is fine, but the type may need to be more `const char * const *` as you have pointer to a constant array of pointers, each of which points to a constant char string.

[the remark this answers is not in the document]
tungld: Done. I changed to `const char * const *`. Keep using NULL-terminated arrays. Thanks!

src/Runtime/PyExecutionSession.cpp
tungld: Done. Thanks!

AlexandreEichenberger:
I added some high level comments to the related issue #1183

include/onnx-mlir/Runtime/OMSignature.h
```diff
@@ -19,18 +19,22 @@ extern "C" {
...
 /**
- * \brief Return the model's input signature as a JSON string.
+ * \brief Return the model's input signature as a JSON string for the given
+ * entry point name.
  *
+ * @param entry point name
  * @return pointer to input signature JSON string
  */
-const char *omInputSignature();
+const char *omInputSignature(const char *);
```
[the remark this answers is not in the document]
tungld: Done for both omInputSignature and omOutputSignature. Thanks!

src/Runtime/ExecutionSession.hpp
```diff
@@ -41,10 +41,17 @@ class ExecutionSession {
   // tensor lists.
   OMTensorList *run(OMTensorList *input);
 
+  // Set entry point for this session.
+  void setEntryPoint(std::string);
```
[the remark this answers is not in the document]
tungld: Fixed this.

AlexandreEichenberger:
Copied here the question I posted on the related issue.

@tungld,

I think we must be very clear on what the "entry point" are. It may be self evident to us now, but this is important for the user as they are exposed to these names. Is it correct that these "entry point" are actually the symbol of the compute function? If so, they are exposed to the user as if you have a statically linked compute function, like in many of our examples, we expect the user to define and then call entry_point(inputTensorList), and if the user is working with dynamic libraries, the user is expected to do a `dlsym(entry_point).

I agree with @gongsu832 that we should have functions to get the signatures that passes the entry point name. But this starts to feel like we have a data structure for each entry point.

struct EntryPointStruct {
  char *entryPointName;
  char *inputSignature;
  char *outputSignature;
}
That feel preferable to having hardwired functions to get the signatures, as if we have multiple entry point, we must start to enumerate/generate to get the right function for a solution 1

if (entry_point_name == "test_add")
    run_test_add_in_signature();
else if (entry_point_name == "test_relu")
    run_test_relu_in_signature();
(since the name are constants, probably test the address of the constant is enough).

gongsu832:
@AlexandreEichenberger The main reason the signature APIs are done in the current way as oppose to using a `struct EntryPointStruct` is to maintain backward compatibility.

AlexandreEichenberger:
@gongsu832 we have no signature in our current design, and it feel that a struct would be a very efficient way to implement one

gongsu832:
We do. WMLz folks have been using the Java signature APIs, which is a direct JNI call to the C signature APIs, for quite a while.

src/Runtime/ExecutionSession.cpp
```diff
@@ -96,10 +97,27 @@ OMTensorList *ExecutionSession::run(OMTensorList *input) {
   return _entryPointFunc(input);
 }
 
-std::string ExecutionSession::inputSignature() { return _inputSignatureFunc(); }
+void ExecutionSession::setEntryPoint(const std::string &entryPointName) {
```
AlexandreEichenberger, line 100: If a user does not want the default entry point, and does not call `setEntryPoint`, then the `_entryPointFunc` remains null, and is unconditionally called in the `run` function. I suggest to add something similar to the code below inside the `run` function.
```C
if (!_entryPointFunc) {
     std::stringstream errStr;
     errStr << "Must set the entry point before calling run function"<< std::endl;
   }
```
Would do the same for the other signature functions too.

src/Runtime/ExecutionSession.hpp
```diff
@@ -41,10 +41,17 @@ class ExecutionSession {
   // tensor lists.
   OMTensorList *run(OMTensorList *input);
 
+  // Set entry point for this session.
```
AlexandreEichenberger, line 44: I would add that this call is required before calling run/input/output signatures when default entry point is not selected. To that effect, I would move this def above the definition of run, so that the sequence is clearer.

src/Runtime/ExecutionSession.cpp
```diff
@@ -96,10 +97,27 @@ OMTensorList *ExecutionSession::run(OMTensorList *input) {
...
+  }
+  _entryPointName = entryPointName;
+}
+
+const std::string *ExecutionSession::queryEntryPoints() const {
+  return (const std::string *)_queryEntryPointsFunc();
+}
+
+const std::string ExecutionSession::inputSignature() const {
+  return _inputSignatureFunc(_entryPointName.c_str());
```
AlexandreEichenberger, line 116: if the entry point was not set, then what happens in the `_inputSignatureFunc`? Should we test this here?

include/onnx-mlir/Runtime/OMSignature.h
```diff
@@ -19,18 +19,20 @@ extern "C" {
...
 
 /**
- * \brief Return the model's input signature as a JSON string.
+ * \brief Return the input signature of the given entry point as a JSON string.
  *
+ * @param entry point name
  * @return pointer to input signature JSON string
  */
-const char *omInputSignature();
+const char *omInputSignature(const char *entryPointName);
```
AlexandreEichenberger, line 27: Can you briefly explain how the entryPointName is used to locate the right signature? I tried to read the LLVM changes but I could not really get the big picture.

test/backend-cpp/ModelBuilder.cpp
```diff
@@ -63,8 +63,8 @@ bool ModelBuilder::runAndVerifyTest(std::vector<OMTensorUniquePtr> &inputs,
   assert(!inputs.empty() && "Expecting valid inputs");
 
   // Run the test code.
-  onnx_mlir::ExecutionSession execSession(
-      getSharedLibName(sharedLibBaseName), "run_main_graph");
+  onnx_mlir::ExecutionSession execSession(getSharedLibName(sharedLibBaseName));
+  execSession.setEntryPoint("run_main_graph");
```
AlexandreEichenberger, line 67: Can you confirm that this call is redundant as the exec session uses by default the "run_main_graph" name.

test/numerical/numerical.def
```diff
@@ -1,5 +1,5 @@
-EXPORTS 
-run_main_graph
-omInputSignature
-omOutputSignature
-
+EXPORTS
```
AlexandreEichenberger, line 1: Can you double check that these changes are made to all of the ".def" files? Thanks

src/Runtime/ExecutionSession.hpp
```diff
@@ -24,15 +24,15 @@
...
+using entryPointFuncType = OMTensorList *(*)(OMTensorList *);
+using queryEntryPointsFuncType = const char **(*)();
+using signatureFuncType = const char *(*)(const char *);
 using OMTensorUniquePtr = std::unique_ptr<OMTensor, decltype(&omTensorDestroy)>;
 
 class ExecutionSession {
 public:
   ExecutionSession(std::string sharedLibPath);
-  ExecutionSession(std::string sharedLibPath, std::string entryPointName);
+  ExecutionSession(std::string sharedLibPath, bool defaultEntryPoint);
```
AlexandreEichenberger, line 35: nit, why not just one `ExecutionSession(std::string sharedLibPath, bool defaultEntryPoint=true);`

AlexandreEichenberger:
My point above was on whether the get signatures were efficiently implemented or not (and maybe having a struct that is discovered by a set entry point could be useful). In any case, signatures are not really on the critical path anyway. So it is not that important. I did leave a comment asking for a broad sketch of how it works. 

Otherwise, I left low level comments on the implementation of ExecutionSession.

src/Runtime/ExecutionSession.cpp
tungld: Done for this and other functions too. Thanks!

src/Runtime/ExecutionSession.hpp
tungld: Yes, I changed the order.

src/Runtime/ExecutionSession.cpp
tungld: Done.

test/backend-cpp/ModelBuilder.cpp
tungld: Yes, it's redundant. Done.

include/onnx-mlir/Runtime/OMSignature.h
tungld: LLVM code is equivalent to the following pseudo code:
```C
entry_point_0 = "run_add";
entry_point_0_in_sig = "[....]"; // JSON string.
entry_point_1 = "run_sub";
entry_point_1_in_sig = "[....]"; // JSON string.
const char* omInputSignature(entry_point) {
    if (entry_point == entry_point_1)
      return entry_point_0_in_sig;
    if (entry_point == entry_point_1)
      return entry_point_1_in_sig;
    return undefined;
}
```

We do not see if-then-else but branches at LLVM level.

test/numerical/numerical.def
tungld: Done for other .def files. Thanks!

src/Runtime/ExecutionSession.hpp
tungld: Done. Now we use a single constructor.

include/onnx-mlir/Runtime/OMSignature.h
AlexandreEichenberger: Cool, and the compare `entry_point == entry_point_0` simply compare the addresses of the strings?

test/backend-cpp/ModelBuilder.cpp
AlexandreEichenberger: Its fine if you wanted to test the functionality, but if you do this, just add a comment as otherwise folks may think that it is still needed.

include/onnx-mlir/Runtime/OMSignature.h
tungld: Yes, the address of the first char.

src/Runtime/ExecutionSession.hpp
```diff
@@ -24,15 +24,21 @@
...
 public:
-  ExecutionSession(std::string sharedLibPath);
-  ExecutionSession(std::string sharedLibPath, std::string entryPointName);
+  ExecutionSession(std::string sharedLibPath, bool defaultEntryPoint = true);
+
+  // Get a NULL-terminated array of entry point names.
+  // For example {"run_addition, "run_substraction", NULL}
+  const std::string *queryEntryPoints() const;
+
+  // Set entry point for this session.
```
AlexandreEichenberger, line 40: Please just add in the comments that this call is necessary when the defaultEntryPoint is not used, or if it was used but the user wants now to use a different entry point.  

I have a question related to the string passed to the setEntryPoint. I know that for the "run" part, a dlsym is used, so the string chars are used to locate the symbol. But for the signatures, is it the address of the string that needs to be the same as recorded, or is it the content of the string?

In other words, are either of these calls working?

``` C++
setEntryPoint(queryEntryPoints()[0]); // use string directly
setEntryPoint(string(queryEntryPoints()[0])); // use a copy of the string
```

AlexandreEichenberger (commented):
Thanks for the code changes. Still have a question pending that has an impact on how the interface works. Once that is resolved, it will be all good for me.

src/Runtime/ExecutionSession.hpp
tungld: > setEntryPoint(queryEntryPoints()[0]); // use string directly
> setEntryPoint(string(queryEntryPoints()[0])); // use a copy of the string
> is it the content of the string?

Yes, it is the string content. so both work.

tungld: > Please just add in the comments that this call is necessary when the defaultEntryPoint is not used, or if it was used but the user wants now to use a different entry point.

Done this. Thanks!

include/onnx-mlir/Runtime/OMSignature.h
AlexandreEichenberger: This seems at odd with your comment that the comparison is done on the string content. Can you run a test to make sure that works? Thanks

AlexandreEichenberger (approved):
LGTM, thanks for all the updates and answers to questions.

PR #1185, opened 2022-02-17
Build user image with ubi8-minimal base image

gongsu832:
- build user image with ubi8-minimal base image
- install onnx into lib instead of lib64
- fix time zone setup in container
- download jsoniter jar from maven repo
- run check-mlir on s390x but ignore error for now
- run check-onnx-backend-signature
- unshallow and rename origin to upstream for onnx-mlir-dev
  image to be published

### DIFF
--- a/.buildbot/jenkins-build-llvm-project.py
+++ b/.buildbot/jenkins-build-llvm-project.py
@@ -56,6 +56,8 @@ LLVM_PROJECT_SHA1_FILE      = 'utils/clone-mlir.sh'
 LLVM_PROJECT_SHA1_REGEX     = 'git checkout ([0-9a-f]+)'
 LLVM_PROJECT_DOCKERFILE     = 'docker/Dockerfile.llvm-project'
 LLVM_PROJECT_GITHUB_URL     = 'https://api.github.com/repos/llvm/llvm-project'
+BASE_IMAGE                  = { 'static': 'ubuntu:focal',
+                                'shared': 'registry.access.redhat.com/ubi8-minimal:latest' }
 LLVM_PROJECT_IMAGE          = { 'static': docker_static_image_name,
                                 'shared': docker_shared_image_name }
 BUILD_SHARED_LIBS           = { 'static': 'off',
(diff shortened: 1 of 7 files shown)


gongsu832:
ubi8 base image replaces the ubuntu base image for the user docker image on all platforms. There is no external visible change to users. Step #5 in `docs/docker-example/Dockerfile` shouldn't be necessary any more. But I'm deferring the change until the PR is merged and the pushed dev image is indeed as expected.

AlexandreEichenberger:
Thanks for the update. Step 5 copied below:
```
# 5) Fix git by reattaching head and making git see other branches than main.
WORKDIR /workdir/onnx-mlir
RUN git remote rename origin upstream
RUN git checkout main
RUN git fetch --unshallow
# Add optional personal fork and disable pushing to upstream (best practice).
# RUN git remote add origin https://github.com/<user>/onnx-mlir.git
# RUN git remote set-url --push upstream no_push
```

So you are installing the repo in the docker image under upstream? If not, we will still need some of the steps are many of us are developing in containers using our own fork as "origin"

gongsu832:
With this PR, when the dev image is published, the onnx-mlir repo remote will be renamed to upstream and unshallowed so it's not necessary to do it yourself as described in `docs/docker-example/Dockerfile`. But as I mentioned, I will defer updating `docs/docker-example/Dockerfile` to another PR.

sstamenova:
@gongsu832 This is an old PR, but I'm looking to see which pip packages should be added to a requirements.txt file that can be shared among the builds and dev setup (and pinned to a common version there). It looks like pytest & friends were added to the docker image for llvm-project here, but I didn't think they were needed for llvm-project. Can you explain a little why they were needed in that particular image?

gongsu832:
Because llvm-project image is only rebuilt when we update LLVM to a new commit hash. So pytest and other prereqs for onnx-mlir are installed in llvm-project image so they don't get needlessly reinstalled every time we build the onnx-mlir image.

PR #1188, opened 2022-02-17
Fix the static build of OnnxMlirCompiler lib on Windows and modify the pipeline to build all

sstamenova:
The OnnxMlirCompiler static build was treating the export functions incorrectly - this change sets up the build to properly determine when they should be exported or not based on whether the libraries are built as static or shared libs.

It also now builds the `all` target in the Windows-CI.


AlexandreEichenberger:
Closing this PR as I used it to fix #1181

PR #1191, opened 2022-02-21
Matmul vector

AlexandreEichenberger:
The new `vector::load store` now work really well, so I don't need to "transform" arrays of float into arrays of vectors of floats. This now enable us to be less restrictive on the cases we can simdize, and the code is clearer because we operate in the original data representation.

I also cleaned up the code by using a VectorBuilder.


src/Conversion/KrnlToAffine/KrnlToAffine.cpp
```diff
@@ -763,6 +763,7 @@ class KrnlMatmulLowering : public OpRewritePattern<KrnlMatMulOp> {
       simdize = false;
       LLVM_DEBUG(llvm::dbgs() << "Matmul: No simd due to vl not a literal\n");
     }
+#if 0
```
etiotto, line 766: Given this is not needed I think it would be cleaner to remove the code all together (this is usually what is done in LLVM).

src/Conversion/KrnlToAffine/KrnlToAffine.cpp
```diff
@@ -989,8 +990,8 @@ class KrnlMatmulLowering : public OpRewritePattern<KrnlMatMulOp> {
     // Have to privatize CTmpType by unroll factor (1 if none).
     MemRefType CTmpType = MemRefType::get({unrollFactor}, vecType);
     KrnlBuilder createKrnl(rewriter, loc);
-    Value vecB = createKrnl.vectorTypeCast(B, VL);
-    Value vecC = createKrnl.vectorTypeCast(C, VL);
+    // Value vecB = createKrnl.vectorTypeCast(B, VL);
```
etiotto, line 993: Remove

src/Conversion/KrnlToAffine/KrnlToAffine.cpp
```diff
@@ -975,6 +973,9 @@ class KrnlMatmulLowering : public OpRewritePattern<KrnlMatMulOp> {
     // can simdize only if K is compile time
     assert(J.isLiteral() &&
            "can only simdize with compile time blocking factor on simd axis");
+    MultiDialectBuilder<AffineBuilderKrnlMem, MemRefBuilder, KrnlBuilder>
```
etiotto, line 976: Is the new builder used in the code ? Could not find it.

src/Conversion/KrnlToAffine/KrnlToAffine.cpp
AlexandreEichenberger: yes, noticed it, removed already.

src/Conversion/KrnlToAffine/KrnlToAffine.cpp
AlexandreEichenberger: your right, I updated the code to support MultiDialectBuilder fo VectorBuilder. Code is uploaded already.

src/Conversion/KrnlToAffine/KrnlToAffine.cpp
AlexandreEichenberger: I will remove it after a round of successful test.

AlexandreEichenberger:
Code ran successfully, removed unused tests in matmul lowering. Code is ready to be reviewed.

etiotto (approved):
LGTM

PR #1193, opened 2022-02-22
Force check-doc to use Python3_EXECUTABLE

hpmorgan:
In our team's subtree-merge of onnx-mlir, we configure a venv to give us precise control over the python configuration without messing with the system/user python install. The check-doc target fails in this mode, because it defaults to the python in the path. There are a couple of ways to resolve this: one would be to run the tests with 'cmake -E env' and inject the python dir at the head of the PATH. This has insidious potential side effects, as the test will always run with the value of PATH as it was at configure time. Subsequent changes to the PATH in a shell window will have no effect without rebuilding the cmake cache, and then _those_ changes will persist until the next cache rebuild.

The other approach, which is what I've implemented here, is to render the doc-check file as cmake templates via configure_file. While slightly more complex from an implementation point of view, it is more robust and the results more predictable.

### DIFF
--- a/docs/doc_check/CMakeLists.txt
+++ b/docs/doc_check/CMakeLists.txt
@@ -1,13 +1,16 @@
 # SPDX-License-Identifier: Apache-2.0
 
+set(check_cmd ${Python3_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/check.py)
+
 add_custom_target(check-doc
-  COMMAND ${Python3_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/check.py
+  COMMAND ${check_cmd}
     ${ONNX_MLIR_SRC_ROOT}
     --exclude_dirs
       third_party
       docs/doc_check/test
       docs/_site
-      )
+  COMMAND ${check_cmd} ${CMAKE_BINARY_DIR}
+  )
 
 set_target_properties(check-doc PROPERTIES FOLDER "Docs")
 # Exclude the target from the default VS build
--- a/src/Builder/CMakeLists.txt
+++ b/src/Builder/CMakeLists.txt
@@ -14,3 +14,8 @@ add_onnx_mlir_library(OMBuilder
   OMResultTypeInferenceOpInterface
   onnx
   )
+
+configure_file(OpBuildTable.inc.dc.in
+  ${CMAKE_CURRENT_BINARY_DIR}/OpBuildTable.inc.dc
+  @ONLY
+  )
--- a/src/Builder/OpBuildTable.inc.dc
+++ b/src/Builder/OpBuildTable.inc.dc
@@ -1,1 +1,0 @@
-file-same-as-stdout({"file": "src/Builder/OpBuildTable.inc", "cmd": ["python3", "utils/gen_onnx_mlir.py", "--dry-run-op-build-table"]})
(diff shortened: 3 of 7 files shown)


sstamenova:

docs/doc_check/CMakeLists.txt
```diff
@@ -1,13 +1,16 @@
 # SPDX-License-Identifier: Apache-2.0
 
+set(check_cmd ${Python3_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/check.py)
+
 add_custom_target(check-doc
-  COMMAND ${Python3_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/check.py
+  COMMAND ${check_cmd}
     ${ONNX_MLIR_SRC_ROOT}
```
sstamenova, line 7: At this point is there anything left in the src tree to check?

AlexandreEichenberger: check doc is used for various different script we list in our docs, so yes, it does verify things.

hpmorgan: IIRC, it checks .md files under the src root, in addition to the .dc files.

sstamenova: > check doc is used for various different script we list in our docs, so yes, it does verify things.

@AlexandreEichenberger: I meant - are there files left in the source tree that are being verified after the two `.dc` files move to the bin root?

It sounds like the `.md` files are also verified separately?

hpmorgan: From check.py:

`for doc_file in chain(Path(root_dir).rglob('*.md'), Path(root_dir).rglob('*.dc'))`

sstamenova: I think all of those are under `${ONNX_MLIR_SRC_ROOT}/docs`. Perhaps this could change to checking the `docs` dir and the binary dir?

hpmorgan: There are .md files in the root, though perhaps not ones we need to check. Is there a problem with continuing to glob from the src root?

sstamenova: My thinking was if all the files that need checking live in only a couple of locations, they could be passed directly to simplify the expression and make it easier to track. It's not necessary, but it could be nice.

gongsu832:
I'm trying to understand why in your venv `python3` doesn't work but `@Python3_EXECUTABLE@` does. Since both will be resolved according to the PATH of your venv. I just tried your patch inside a venv where `@Python3_EXECUTABLE@` resolves to `/root/myenv/bin/python3`, which is exactly where `python3` will resolve to. So can you elaborate?

hpmorgan:
> Since both will be resolved according to the PATH of your venv. I just tried your patch inside a venv where `@Python3_EXECUTABLE@` resolves to `/root/myenv/bin/python3`, which is exactly where `python3` will resolve to. So can you elaborate?

Python3_EXECUTABLE is resolved at configure time and is never (automatically) changed, so build commands that use it aren't sensitive to the PATH set in the shell in which it's running. What this means in practice is that one can configure one's bin dir in one shell and then run make/ninja in another shell and it will just work (without having to update the PATH or activate a venv).

sstamenova:
@gongsu832 : Henry's solution also addresses the problem that we talked about earlier - if python is not called `python3` in the environment, the `check-doc` targets fail currently. I think not depending on the name of the python executable and letting cmake handle the paths is much cleaner and supports scenarios that are currently broken (e.g. in my environment, python is called `python` not `python3` and while I could change it, now I wouldn't have to).

gongsu832:
@hpmorgan Yes I see now that you want the resolved python3 path at configuration time to "stick", regardless of the env change later on.
@sstamenova Python3_EXECUTABLE specifically looks for python3[.x] so it does indeed disambiguate between python and python3.
I was trying to think if there is any way to achieve this without going through an additional cmake configure_file but looks like there is none (that I can think of). So I guess it's a small price we'll have to pay.

gongsu832 (approved):
LGTM

PR #1195, opened 2022-02-23
Remove explicit verifier usage as it is about to be obsolete

sstamenova:
Several properties (such as verifier, printer & parser) are going to be obsolete in main mlir soon. This change removes the explicit usage of verifier and instead uses the `hasVerifier` option instead.

Also, some minor whitespace cleanup.


etiotto (approved):
LGTM - thanks!

PR #1197, opened 2022-02-23
Update LLVM to commit d7f0083d and remove explicit usage of parser and printer

sstamenova:
This updates LLVM to commit d7f0083d (just about a day ahead from the existing commit) when parser and printer were made obsolete. Then we leverage the new `hasCustomAssemblyFormat` property to replace parser and printer.


etiotto (approved):
LGTM - thanks!

PR #1199, opened 2022-02-25
[NFI] Split krnlToLLVM.cpp file

etiotto:
No new functionality is added by this PR.  This PR decomposes `krnlToLLVM.cpp` info smaller compilation units (similar to what we have done for the ONNX dialect). 

Also, all Krnl dialect operation conversions are now in a dedicated `krnl` namespace nested inside the `onnx_mlir` namespace.  This allows containment of our `krnl` dialect conversions in its namespace rather than being in the `mlir` namespace (which belongs to official MLIR dialects). 


AlexandreEichenberger:
@etiotto I laud this effort. Unrelated, if you wanted to do something similar for ONNX to Krnl, that would be great.

etiotto:
@AlexandreEichenberger thank you. LEt's review this one first and then I'll put that request on the queue. I'd like to go through review for this one asap because is time consuming to keep this PR up to date with other changes in `KrnlToLLVM.cpp`.

src/Conversion/KrnlToLLVM/ConvertKrnlToLLVM.hpp
```diff
@@ -0,0 +1,85 @@
...
+#pragma once
+
+#include "src/Conversion/KrnlToLLVM/ConvertKrnlToLLVM.hpp"
+#include "src/Dialect/Krnl/KrnlOps.hpp"
+#include "src/Pass/Passes.hpp"
+#include "src/Support/Common.hpp"
+
+const std::string DEFAULT_DYN_ENTRY_POINT = "run_main_graph";
+
+using namespace mlir;
```
sstamenova, line 25: It's generally a good idea to avoid `using` statements for namespaces in headers.

etiotto: Sorry @sstamenova I merged the PR already before seeing this comment. I will post another PR and remove the using declaration in the header file.

etiotto: PR posted.

PR #1203, opened 2022-02-26
pow op verifier

thomas-m1:
re-fixed pow op verifier
Signed-off-by: root <thomas_michalski@hotmail.com>


src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -1141,19 +1141,25 @@ LogicalResult ONNXErfOp::inferShapes(
 // PowOp
 //===----------------------------------------------------------------------===//
 
+static LogicalResult verify(ONNXPowOp op) {
```
chentong319, line 1144: The verifier interface is changed:
`LogicalResult ONNXPowOp::verify(){`

chentong319 (approved):
Congratulations for your first PR!

PR #1204, opened 2022-02-26
update llvm to a7ac120a9ad7

etiotto:
Ths PR updates the LLVM commit we use to `a7ac120a9ad784998a5527fc0a71b2d0fd55eccb`.

### DIFF
--- a/src/Dialect/ONNX/ONNX.td
+++ b/src/Dialect/ONNX/ONNX.td
@@ -244,7 +244,7 @@ def ONNX_SeqType : ONNX_Type<"Seq", "Seq"> {
     $_printer << "<" << getImpl()->ElementType << ">";
   }];
   let parser = [{
-    if (parser.parseLess())
+    if ($_parser.parseLess())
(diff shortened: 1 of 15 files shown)


etiotto:

src/Dialect/ONNX/ONNX.td
```diff
@@ -244,7 +244,7 @@ def ONNX_SeqType : ONNX_Type<"Seq", "Seq"> {
     $_printer << "<" << getImpl()->ElementType << ">";
   }];
   let parser = [{
```
sstamenova, line 246: Looks like I missed one! We should convert this to not user parser anymore either.

sstamenova (approved):
LGTM. It would be nice to remove the last remaining parser, but that can be done separately.

src/Dialect/ONNX/ONNX.td
etiotto: OK we can do that in another PR.

PR #1205, opened 2022-02-28
Encode signatures and entry point names in EBCDIC codepage for z/OS

tungld:
Strings for signatures and entry point name should be encoded in EBCDIC if compiled for z/OS. This patch adds two things to make the strings compatible on z/OS:
- Add a global metadata to generated LLVM IR, something like `!3 = !{i32 1, !"zos_le_char_mode", !"ebcdic"}`
- Encode signatures and entry point names in EBCDIC instead of ASCII when `zos` is specified in `--mtriple`.


src/Compiler/CompilerUtils.cpp
```diff
@@ -521,6 +521,18 @@ static void genLLVMBitcode(const mlir::OwningOpRef<ModuleOp> &module,
     llvm::errs() << "Failed to translate module to LLVMIR.\n";
     exit(1);
   }
+
+  // Emit metadata "zos_le_char_mode" for z/OS. Use EBCDIC codepage by default.
+  bool zOS = StringRef(getTargetTripleOption()).endswith_insensitive("zos");
+  if (zOS) {
+    StringRef charModeKey = "zos_le_char_mode";
+    StringRef charModeValue = "ebcdic";
```
etiotto, line 529: [nit]: you could sink this line inside the if stmt

src/Conversion/KrnlToLLVM/ConvertKrnlToLLVM.cpp
```diff
@@ -178,6 +178,13 @@ void recordEntryPointSignatures(ModuleOp &module,
     SmallVectorImpl<std::string> &entryPointNames,
     SmallVectorImpl<std::string> &inSignatures,
     SmallVectorImpl<std::string> &outSignatures) {
+
+  Attribute mtripleAttr =
+      module->getAttrOfType<::mlir::Attribute>("llvm.target_triple");
+  bool zOS =
```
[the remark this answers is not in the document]
tungld: Done. Thanks for you suggestion! It's useful.

src/Compiler/CompilerUtils.cpp
tungld: Done.

src/Compiler/CompilerUtils.cpp
```diff
@@ -521,6 +521,18 @@ static void genLLVMBitcode(const mlir::OwningOpRef<ModuleOp> &module,
     llvm::errs() << "Failed to translate module to LLVMIR.\n";
     exit(1);
   }
+
+  // Emit metadata "zos_le_char_mode" for z/OS. Use EBCDIC codepage by default.
+  bool zOS = StringRef(getTargetTripleOption()).endswith_insensitive("zos");
```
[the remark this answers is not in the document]
tungld: Done.

src/Compiler/CompilerUtils.cpp
etiotto: hmmm I still see it in at line 528.  Forgot to push ?

etiotto (approved):
LGTM. One minor comment left you can do before merging (optional).

src/Compiler/CompilerUtils.cpp
tungld: Done. I though it was the above if.

PR #1208, opened 2022-03-01
initial support for (optional) accelerators

caoimhinuibrian:


caoimhinuibrian:
@sstamenova asked me to split this out from the NNPA specific stuff. This is just the framework for adding one or more optional accelerators. I will put the NNPA code into a separate PR once this lands. @tungld @etiotto please take a look. The code is just the same as it was in #1149 .

src/Accelerators/Accelerator.cpp
```diff
@@ -0,0 +1,38 @@
...
+  }
+}
+
+std::vector<Accelerator *> *Accelerator::getAcceleratorList() {
+  std::cout << "getting accelerator targets" << std::endl;
+  return acceleratorTargets;
+}
+
+} // namespace mlir
+© 2022 GitHub, Inc.Terms
```
tungld, line 38: This looks like a redundant line?

tungld (changes requested):

src/Accelerators/Accelerator.cpp
caoimhinuibrian: whoops!

tungld (approved):
LGTM!

src/Accelerators/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+# SPDX-License-Identifier: Apache-2.0
+
+if (ACCELERATORS_TO_BUILD)
+  set (ACCEL_LIST "") 
+  foreach(t ${ACCELERATORS_TO_BUILD})
```
gongsu832, line 5: You mean `foreach(t ${ACCEL_LIST})`?

src/Support/OMOptions.cpp
```diff
@@ -40,3 +40,8 @@ llvm::cl::opt<int> onnxOpTransformThreshold("onnx-op-transform-threshold",
 llvm::cl::opt<bool> onnxOpTransformReport("onnx-op-transform-report",
     llvm::cl::desc("Report diagnostic info for op transform passes."),
     llvm::cl::init(false), llvm::cl::cat(OMPassOptions));
+
+llvm::cl::opt<std::string> acceleratorTarget("accel",
```
etiotto, line 44: I think we will eventually want to use an enum to group legal accelerators. Doing that would then allow us to list available accellerator, similar to what LLVM does for this option:  
```
 --code-model=<value>                                                  - Choose code model
    =tiny                                                               -   Tiny code model
    =small                                                              -   Small code model
    =kernel                                                             -   Kernel code model
    =medium                                                             -   Medium code model
    =large                                                              -   Large code model
```

The code to get the listing above would be: 

```
  static cl::opt<CodeModel::Model> CodeModel(
      "code-model", cl::desc("Choose code model"),
      cl::values(clEnumValN(CodeModel::Tiny, "tiny", "Tiny code model"),
                 clEnumValN(CodeModel::Small, "small", "Small code model"),
                 clEnumValN(CodeModel::Kernel, "kernel", "Kernel code model"),
                 clEnumValN(CodeModel::Medium, "medium", "Medium code model"),
                 clEnumValN(CodeModel::Large, "large", "Large code model")));

Comment (etiotto) [reply src/Support/OMOptions.cpp:44]: I am OK with doing this in a subsequent PR given that we want to upload the support for the Telus accellerator asap.

File: src/Accelerators/CMakeLists.txt [hunk 4]
```diff
@@ -0,0 +1,20 @@
...
+    message(DEBUG "Targeting ${t}")
+    add_subdirectory(${t})
+    list(APPEND ACCEL_LIST ${t}"Accel")
+  endforeach(t)
+else (ACCELERATORS_TO_BUILD)
+  set (ACCEL_LIST "NoAccel")    
+endif (ACCELERATORS_TO_BUILD)
+
+add_onnx_mlir_library(NoAccel
```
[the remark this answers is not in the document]
sstamenova: Also, it can be called NoAccelerator like the .cpp file.

src/Compiler/CMakeLists.txt
```diff
@@ -65,6 +65,9 @@ add_onnx_mlir_library(OnnxMlirCompiler
   OnnxMlirCompiler.cpp
 
   EXCLUDE_FROM_OM_LIBS
+  
+  DEPENDS
+  ${ACCEL_LIST}
```
sstamenova, line 70: Is this a dependency or should it be linked?

src/Accelerators/CMakeLists.txt
sstamenova, line 14: By default this function is going to add it to the OMLibs property. Do we want the accelerator libs linked in everywhere the OMLibs are?

sstamenova (changes requested):

src/Accelerators/CMakeLists.txt
caoimhinuibrian: something has to be provided. That's what the NoAccelerator thing was ... I removed it and just use the Accelerator.cpp to make a lib.

caoimhinuibrian:
@sstamenova how does this look now?

src/Accelerators/CMakeLists.txt
sstamenova: Just to make sure I understand correctly, the accelerators will need to be added every place where the OMLibs are used for linking. So in essence, the accelerators will be needed everywhere?

If so, this is ok, otherwise, we should make sure that they don't get added to the OMLibs property.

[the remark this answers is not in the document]
sstamenova: The add for the `NoAccel` library should happen in the else case.

src/Accelerators/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+
+if (ACCELERATORS_TO_BUILD)
+  set (ACCEL_LIST "") 
+  foreach(t ${ACCELERATORS_TO_BUILD})
+    message(DEBUG "Targeting ${t}")
+    add_subdirectory(${t})
+    list(APPEND ACCEL_LIST ${t}"Accel")
+  endforeach(t)
+else (ACCELERATORS_TO_BUILD)
+  set (ACCEL_LIST "NoAccel")
```
sstamenova: Is the `ACCEL_LIST` used anywhere?

sstamenova:
> @sstamenova how does this look now?

There are a couple of things that haven't been addressed it - more specifically, I want to make sure the libs are added to OMLibs only if necessary and that the NoAccel lib is only built if there are no other accelerators..

caoimhinuibrian:
> There are a couple of things that haven't been addressed it - more specifically, I want to make sure the libs are added to OMLibs only if necessary and that the NoAccel lib is only built if there are no other accelerators..

Only the Accelerator lib, which is a stub, is added to OMLibs. That is necessary in order to provide the static method "getAcceleratorList". The NoAccel lib is gone now.

src/Accelerators/CMakeLists.txt
caoimhinuibrian: Not anymore. I have removed it.

src/Accelerators/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+# SPDX-License-Identifier: Apache-2.0
+
+if (ACCELERATORS_TO_BUILD)
+  foreach(t ${ACCELERATORS_TO_BUILD})
+    message(DEBUG "Targeting ${t}")
+    add_subdirectory(${t})
+  endforeach(t)
+else (ACCELERATORS_TO_BUILD)
```
sstamenova, line 8: Nit: you don't need this `else` clause

sstamenova (approved):
LGTM

PR #1212, opened 2022-03-02
[NFI]: Use dialect builder rather than creating operators directly

etiotto:
Rather than constructing [krnl,arith,math,memref] dialect operations directly (e.g. `rewriter.create<KrnlLoadOp>(...)`) we should use the appropriate dialect builder facilities (e.g. `create.krnl.load(....)`).  Using the dialect builder(s) will isolate the code from potential future changes in MLIR (because it will allow us to modify just the builder rather than changing the code more pervasively).

This PR does makes the above changes for a few files. Subsequent PRs will continue this effort. 
 

etiotto:
@gongsu832 could you pls take a look at the amd64 JEnkins, it failes with:
```
Cloning repository https://github.com/onnx/onnx-mlir.git
 > git init /var/lib/jenkins/workspace/ONNX-MLIR-Pipeline-Docker-Build@script/pr_1212 # timeout=10
Fetching upstream changes from https://github.com/onnx/onnx-mlir.git
 > git --version # timeout=10
 > git --version # 'git version 2.25.1'
 > git fetch --tags --force --progress --depth=1 -- https://github.com/onnx/onnx-mlir.git +refs/heads/*:refs/remotes/origin/* # timeout=10
ERROR: Error cloning remote repo 'origin'
hudson.plugins.git.GitException: Command "git fetch --tags --force --progress --depth=1 -- https://github.com/onnx/onnx-mlir.git +refs/heads/*:refs/remotes/origin/*" returned status code 128:
stdout: 
stderr: fatal: the remote end hung up unexpectedly
```

gongsu832:

src/Conversion/ONNXToKrnl/ControlFlow/Loop.cpp
```diff
@@ -68,7 +69,7 @@ struct ONNXLoopOpLowering : public ConversionPattern {
     {
       OpBuilder::InsertionGuard insertGuard(rewriter);
 
-      auto condReg = rewriter.create<KrnlLoadOp>(loc, cond).getResult();
+      Value condReg = createKrnl.load(cond);
```
AlexandreEichenberger, line 72: Interesting, the builders were intended to work with the new interface, and here the loop code is still manually changing insertion points. But this works as onnx-mlir builders only keep a reference to the original `rewriter` as shown below
```c++
struct DialectBuilder {
  DialectBuilder(OpBuilder &b, Location loc) : b(b), loc(loc) {}
  DialectBuilder(DialectBuilder &db) : b(db.b), loc(db.loc) {}
  virtual ~DialectBuilder() {}
  DialectBuilder(DialectBuilder &&) = delete;
  DialectBuilder &operator=(const DialectBuilder &) = delete;
  DialectBuilder &&operator=(const DialectBuilder &&) = delete;

  OpBuilder &getBuilder() const { return b; }
  Location getLoc() const { return loc; }

protected:
  OpBuilder &b;
  Location loc;
};
```

src/Conversion/ONNXToKrnl/NN/Normalization.cpp
```diff
@@ -219,15 +214,14 @@ struct ONNXInstanceNormalizationOpLowering : public ConversionPattern {
...
-                MathBuilder createMath(createKrnl);
                 SmallVector<Value, 6> inputAccessFct = {
                     n.getValue(), c.getValue()};
                 for (int d = 0; d < rank - 2; ++d)
                   inputAccessFct.emplace_back(spatial_loopInd[d]);
                 // tmp += input[n,c, spatial dims]
-                Value oldSum = createKrnl.load(tmpMemRef, {});
-                Value val = createKrnl.load(inputMemRef, inputAccessFct);
-                Value newSum = createMath.add(oldSum, val);
+                Value oldSum = create.krnl.load(tmpMemRef, {});
```
AlexandreEichenberger, line 222: I know that it might work as is, but the model for using builders is to recreate them in each context. Here the create.krml.iterateIE has a lambda functions that provides a KrnlBuilder, and inside that lambda function, that builder should be use to create other builders... So referring here to the external `create.krnl` actually bypass that inner `createKrnl` constructor. 

Please change everywhere accordingly.

AlexandreEichenberger (changes requested):
Builder should not span scopes where new builders are passed by the lambda functions. I showed one case where this pattern is used. Changes should be made in all other cases as well.

The idea to migrate everywhere to the new builders to isolate us from MLIR is great, thanks for starting this.

etiotto (commented):
I looked for cases where a lambda captures a multi dialect builder and instead of using it in the lambda body I changed the code to construct a local multi dialect builder from the `createKrnl` object passed to the lambda.

etiotto:
@AlexandreEichenberger ready for another review round.

src/Conversion/ONNXToKrnl/Math/Reduction.cpp
```diff
@@ -438,19 +412,16 @@ struct ONNXReduceSumOpLowering : public ConversionPattern {
           !llvm::dyn_cast<ONNXReduceSumOp>(op).noop_with_empty_axes()) {
         IndexExprScope axesloopContex(&rewriter, loc);
         MemRefBoundsIndexCapture axesBounds(axesVal);
-        auto zeroIndex = rewriter.create<arith::ConstantIndexOp>(loc, 0);
-        auto cond =
-            rewriter.create<arith::CmpIOp>(loc, arith::CmpIPredicate::eq,
-                axesBounds.getDim(0).getValue(), zeroIndex);
-        initVal =
-            rewriter.create<arith::SelectOp>(loc, cond, trueVal, falseVal);
+        Value zeroIndex = rewriter.create<arith::ConstantIndexOp>(loc, 0);
```
AlexandreEichenberger, line 415: please use this one here and elsewhere (grepping for `ConstantIndexOp`) 

```
  struct MathBuilder {...
  Value constantIndex(int64_t val) const;
  }
```

src/Conversion/ONNXToKrnl/Math/Reduction.cpp
```diff
@@ -664,26 +625,20 @@ struct ONNXReduceSumOpLowering : public ConversionPattern {
         divisor = rewriter.create<arith::IndexCastOp>(
```
AlexandreEichenberger, line 625: There is a cast that gets index to float

src/Conversion/ONNXToKrnl/NN/Normalization.cpp
AlexandreEichenberger: Great! thanks

src/Conversion/ONNXToKrnl/NN/Pooling.cpp
```diff
@@ -132,17 +133,15 @@ void postProcessPoolingWindow<ONNXAveragePoolOp>(
   } else {
     denominator = poolDimValues[0];
     for (unsigned int i = 1; i < poolDimValues.size(); ++i)
-      denominator =
-          rewriter.create<arith::MulIOp>(loc, denominator, poolDimValues[i]);
+      denominator = create.math.mul(denominator, poolDimValues[i]);
     denominator = rewriter.create<arith::IndexCastOp>(
```
AlexandreEichenberger, line 137: use math cast here also. A grep will find them all, so I will stop comment on this

src/Conversion/ONNXToKrnl/Math/Reduction.cpp
etiotto: done

src/Conversion/ONNXToKrnl/Math/Reduction.cpp
etiotto: done

AlexandreEichenberger (approved):
LGTM, big thanks for actualizing the code to the latest builders, much more compact and less repetitive conversions.

PR #1213, opened 2022-03-02
Make checkConstantOutputs and populateAffineAndKrnlToLLVMConversion sharable for other drivers

tungld:


etiotto:
@gongsu832 something is wrong with the Jenkins build on amd64, can you take a loop please when you have time?

gongsu832:
> @gongsu832 something is wrong with the Jenkins build on amd64, can you take a loop please when you have time?

The Jenkins CIs only have 64GB memory. Even with a max of 4 concurrent jobs they can get OOM and Jenkins get killed. Since each job spawns a bunch of compiler and test processes.

PR #1217, opened 2022-03-04
Added Benchmark Parsing Script

adriansion:
Working version, parses Benchmark output for any number of specified ops and accumulates results in array data structure.


test/perf/ParseBenchmarks.py
```diff
@@ -0,0 +1,44 @@
+import subprocess
+import sys
+
+# Add compatible ops here
+OpsWithPerformanceBenchmarks = [
+    'Gemm',
+    'Conv'
+]
+
+ExecutablePath = "/workdir/onnx-mlir/build/Debug/bin/"
```
sstamenova, line 10: This should be determined based on the configuration coming from cmake, so that it works regardless of where the script is run.

AlexandreEichenberger: Check code in RunONNXModel.py line 93. Maybe that can work here too? Or can be used as a basis for a solution

AlexandreEichenberger (commented):
Eventually, we are interested to compare the output of a run with a prior version. Is it possible to have an additional argument to a reference file and report the difference (absolute/percent) for a given metric?

test/perf/ParseBenchmarks.py
```diff
@@ -0,0 +1,44 @@
...
+        LineContentResults.append("cpu=" + tokens[3] + "ms")
+        LineContentResults.append("iterations=" + tokens[5])
+        LineContentResults.append(tokens[6])
+        LineContentResults.append(tokens[7])
+
+        CumulativeResults.append(LineContentResults)
+
+# Working output -- should be updated according to specific needs
+for i in CumulativeResults:
+    print(i)
```
etiotto, line 44: add newline at the end of the file pls

test/perf/ParseBenchmarks.py
```diff
@@ -0,0 +1,44 @@
+import subprocess
```
etiotto, line 1: Add the license and the copyright

etiotto: so this for the license:
```
```
and this for copyright:
```
##################### common.py ################################################
#
# Copyright 2022 The IBM Research Authors.
#
################################################################################

File: test/perf/ParseBenchmarks.py (hunk 1 above)
Comment (AlexandreEichenberger) [reply test/perf/ParseBenchmarks.py:10]: Let me know if you need help

File: test/perf/ParseBenchmarks.py [hunk 4]
```diff
@@ -1,44 +1,216 @@
...
+#
+# --run <Op>: Compute performance benchmarks for specified Op, and write to output file
+#
+# --runall: Compute performance benchmarks for all Ops in OpsWithPerformanceBenchmarks array
+#
+# --readrun <File Op>: Compute performance benchmarks for specified Op, and compare with benchmarks
+# already written to specifiedfile (File must contain same Op)
+#
+# --compare <Op1 Op2>: Compare performance benchmarks written to file for both specified files (each
+# should contain same Op)
```
AlexandreEichenberger, line 14: Great, going in the right direction! Can you add an option that create an error if relative error is above X (where X is given by the option). Thanks

adriansion: Sure, on this now. Should this be a multiplier to be applied on each metric? Or should it be for a specific metric (CPU time, etc.)?

test/perf/ParseBenchmarks.py
```diff
@@ -0,0 +1,264 @@
...
+######################################################################################################
+
+
+
+import os
+import subprocess
+import sys
+import time
+
+
```
AlexandreEichenberger, line 24: Can you add a `print_usage(error_message)` which print the custom error message, followed by the  usage for the script, where all the main options and their subsequent parameters are listed? Kind of what you have on line 2?

We typically use `getopt`, which is a lot easier to manage and does a lot of the checking for you.  You could see in `utils/Mlir2FileCheck.py` how it is done. But don't change your code if its working and it would take you a lot of time to change.

```
        opts, args = getopt.getopt(
            argv, "hdca:n:m:", ["help", "debug", "check", "args=", "names=", "model="])
```

test/perf/ParseBenchmarks.py
```diff
@@ -0,0 +1,264 @@
...
+OpsWithPerformanceBenchmarks = [
+    'Gemm',
+    'Conv'
+]
+
+
+RuntimeEpoch = int(time.time())
+
+
+
```
AlexandreEichenberger, line 50: Please add a short comment to detail the info held by the result. You could also define a constant dictionary that maps the name of the metrics to the number used in the Metrics list.

AlexandreEichenberger (approved):
LGTM

adriansion:
> @adriansion ready to go, or there is more changes you want to make? Leave the integration of the script to a subsequent PR please

All ready, successfully added max relative slowdown yesterday. Will leave CMake integration to new PR, and will merge this one as soon as automated tests after branch update complete.

PR #1219, opened 2022-03-04
Update NNPA code to the latest version and enable lit-tests for NNPA

tungld:
This patch updates NNPA code to the latest version and add all lit tests for NNPA excepts those for stickfication. Lit tests for NNPA is enabled to run on Jenkins Linux s390x, which is to avoid regression for further refactoring NNPA code in the future.

With this patch, all NNPA related passes such as `onnx-to-zhigh`, `zhigh-to-zlow`, `zlow-to-llvm` as well as optimization passes are available in `onnx-mlir-opt`.


src/Accelerators/NNPA/Conversion/ONNXToZHigh/NNPALimit.h
```diff
@@ -0,0 +1,27 @@
...
+// =============================================================================
+//
+// The NNPA maximum supported dimension index size value by using
+// zdnn_get_nnpa_max_dim_idx_size() This value depends on HW.
+//
+//===----------------------------------------------------------------------===//
+
+// The NNPA maximum supported dimension index size value by using
+// zdnn_get_nnpa_max_dim_idx_size() This value depends on HW.
+#define DLCPP_MAXIMUM_DIMENSION_INDEX_SIZE 32768
```
etiotto, line 18: Instead of using C macos to define a constant I suggest we use a proper `static constexpr` variables. Macros are hard to debug.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
```diff
@@ -13,10 +13,12 @@
 //
 //===----------------------------------------------------------------------===//
 
-#include "Conversion/ONNXToZHigh/ONNXLegalityCheck.hpp"
+#include "src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.hpp"
 #include "src/Conversion/ONNXToKrnl/RNN/RNNBase.hpp"
 #include "src/Dialect/ONNX/ShapeInference/ONNXShapeHelper.hpp"
 
+#include "NNPALimit.h"
```
etiotto, line 20: Could you please use the full path to the header file (relative to the base directory of onnx-mlir).

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHighCommon.hpp
```diff
@@ -12,8 +12,10 @@
 //
 //===----------------------------------------------------------------------===//
 
-#include "Conversion/ONNXToZHigh/ONNXLegalityCheck.hpp"
-#include "Support/LayoutHelper.hpp"
+#include "src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.hpp"
+#include "src/Accelerators/NNPA/Support/LayoutHelper.hpp"
+
+#include "NNPALimit.h"
```
etiotto, line 18: Use relative full path to header ?

src/Accelerators/NNPA/Transform/ZLow/ZLowRewrite.cpp
```diff
@@ -0,0 +1,119 @@
+//===-------- ZLowRewrite.cpp - ZLow Rewrite Patterns ---------------------===//
```
etiotto, line 1: Please add license:
```
/*
 */

File: src/Accelerators/NNPA/Transform/ZLow/ZLowRewrite.cpp [hunk 5]
```diff
@@ -0,0 +1,119 @@
...
+//
+// Copyright 2022 The IBM Research Authors.
+//
+// =============================================================================
+//
+// This pass implements optimizations for for ZLow operations.
```
etiotto, line 7: for for -> for

src/Accelerators/NNPA/Transform/ZLow/ZLowRewrite.cpp
```diff
@@ -0,0 +1,119 @@
...
+/// ```
+/// by removing `zlow.stick` and replacing `%res` by `%input`, which is
+/// constrained by that `%input` and `%res` have the same static shape.
+/// This pattern potentially removes `zlow.unstick` and `viewOp` if they are
+/// dangling.
+///
+/// `viewOp` can be any op that inherits ViewLikeOpInterface, e.g.
+/// memref.reinterpret_cast, memref.collapse_shape, memref.expand_shape.
+//
+class StickViewUnstickRemovalPattern : public OpRewritePattern<ZLowStickOp> {
```
etiotto, line 35: Suggest to derive from `ConversionPattern`, are there any advantages using  an OpRewritePattern ?

etiotto, line 35: we should have a namespace for the conversion patterns of the ZLow and ZHigh dialects. Similar to what we now have for the krnl dialect.  Suggest to place ZLow in namespace `onnx_mlir::zlow`, similar for zhigh.

src/Accelerators/NNPA/Transform/ZLow/ZLowRewrite.cpp
```diff
@@ -0,0 +1,119 @@
...
+ */
+class ZLowRewritePass
+    : public PassWrapper<ZLowRewritePass, OperationPass<FuncOp>> {
+public:
+  StringRef getArgument() const override { return "zlow-rewrite"; }
+
+  StringRef getDescription() const override { return "Rewrite ZLow Ops."; }
+
+  void runOnOperation() override {
+    auto function = getOperation();
```
etiotto, line 107: Avoid uses of auto when the static type is not immediately "visible".  https://llvm.org/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable

src/Accelerators/NNPA/Transform/ZLow/ZLowRewrite.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+    ConversionTarget target(getContext());
+    RewritePatternSet patterns(&getContext());
+    patterns.insert<StickViewUnstickRemovalPattern>(&getContext());
+
+    (void)applyPatternsAndFoldGreedily(function, std::move(patterns));
```
etiotto, line 113: Should check whether the transformation was OK. Suggest: 
```
    if (failed(applyPatternsAndFoldGreedily(function, std::move(patterns))))
      return signalPassFailure();

File: src/InitOMPasses.hpp [hunk 9]
```diff
@@ -5,6 +5,10 @@
 #include "mlir/Pass/Pass.h"
 #include "src/Pass/Passes.hpp"
 
+#ifdef __NNPA__
+#include "src/Accelerators/NNPA/Pass/DLCPasses.hpp"
+#endif
+
 using namespace onnx_mlir;
```
etiotto, line 12: Best to avoid using namespace declaration in an header file. (https://stackoverflow.com/questions/5849457/using-namespace-in-c-headers)

src/InitOMPasses.hpp
```diff
@@ -5,6 +5,10 @@
 #include "mlir/Pass/Pass.h"
 #include "src/Pass/Passes.hpp"
 
+#ifdef __NNPA__
```
etiotto, line 8: This header is missing the copyright notice. Can you add it pls ?

src/InitOMPasses.hpp
```diff
@@ -75,5 +79,40 @@ void initOMPasses(int optLevel) {
   mlir::registerPass([]() -> std::unique_ptr<mlir::Pass> {
     return createLowerKrnlShapePass();
   });
+
+#ifdef __NNPA__
+  mlir::registerPass([]() -> std::unique_ptr<mlir::Pass> {
+    return mlir::createONNXToZHighPass();
```
etiotto, line 85: I think we should not place the `mlir` namespace with our conversion passes.  Suggest to place those declaration in namespace `onnx_mlir` ?

src/Tools/onnx-mlir-opt/onnx-mlir-opt.cpp
```diff
@@ -104,6 +109,10 @@ int main(int argc, char **argv) {
 
   registry.insert<mlir::ONNXDialect>();
   registry.insert<mlir::KrnlOpsDialect>();
+#ifdef __NNPA__
```
etiotto, line 112: Could we place these dialect insertions into the registry inside  a function (suggest `registerNNPAPasses()`). Ideally we should avoid the use preprocessor macros all over the place and have something like:

```
   if (NNPAAccellerator.isActive()) 
      registerNNPAPasses();
 ```
 
 I am concerned that over time, as more accelerators are added, we would end up with a bunch of macro checks, making the code less readable.

File: src/Accelerators/NNPA/Dialect/ZLow/ZLowOps.hpp [hunk 13]
```diff
@@ -39,6 +39,6 @@ class ZLowDialect : public Dialect {
 /// Include the auto-generated header file containing the declarations of the
 /// ONNX operations.
 #define GET_OP_CLASSES
-#include "Dialect/ZLow/ZLowOps.hpp.inc"
+#include "src/Accelerators/NNPA/Dialect/ZLow/ZLowOps.hpp.inc"
 
 namespace onnx_mlir {}
```
etiotto, line 44: remove ?

src/Accelerators/NNPA/Pass/DLCPasses.hpp
```diff
@@ -20,9 +20,6 @@
 namespace mlir {
```
etiotto, line 20: I think we should avoid placing our not official dialects in the `mlir` namespace. Suggest to use `onnx_mlir` namespace at minimum, and better to place each dialect conversion passes in its own namespace nested inside onnx_mlir ? 

See the krnl dialect conversions for a template.

etiotto:
Done a first pass of code review @tungld

src/Accelerators/NNPA/Transform/ZLow/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+  ZLowRewrite.cpp
+
+  LINK_LIBS PUBLIC
+  MLIRRewrite
+  MLIRTransformUtils
+  MLIRViewLikeInterface
+  OMZLowOps
+  )
+
+# add_onnx_mlir_library(OMZHighLayoutPropagation
```
sstamenova, line 11: Don't include commented out code

test/mlir/accelerators/nnpa/module_op_be/lit.local.cfg
```diff
@@ -0,0 +1,7 @@
+root = config.root
+# enable when onnx-mlir driver works well with nnpa.
+config.unsupported = True
+# if sys.byteorder == "little":
```
sstamenova, line 4: Same here and in the other cfg files - it's better not to include commented out code

test/mlir/accelerators/nnpa/module_op_be/lit.local.cfg
```diff
@@ -0,0 +1,7 @@
+root = config.root
+# enable when onnx-mlir driver works well with nnpa.
+config.unsupported = True
```
sstamenova, line 3: Should this be true sometimes?

sstamenova:
This change looks like it does a lot more than simply add lit-tests for NPPA. Can you amend the title?

src/InitOMPasses.hpp
caoimhinuibrian, line 8: I don't think we should pollute the high level space with accelerator stuff. Each accelerator should take care of registering its own passes, down inside its own source tree.

src/Tools/onnx-mlir-opt/onnx-mlir-opt.cpp
caoimhinuibrian: yes, see my comment above.

CMakeLists.txt
```diff
@@ -36,6 +36,17 @@ if (MSVC)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -wd4125")
 endif()
 
+# Define preprocessor for accelerators.
+set(NNPA_ENABLED 0)
```
caoimhinuibrian, line 40: I commented later in the code to init the passes, but the same applies here. We shouldn't be polluting the top level of onnx-mlir with knowledge about the accelerators.

caoimhinuibrian:
> This change looks like it does a lot more than simply add lit-tests for NPPA. Can you amend the title?

Yeah, I agree. Maybe split into two PRs?

test/mlir/accelerators/nnpa/module_op_be/lit.local.cfg
tungld: Yes, once `onnx-mlir` command works with NNPAAccelerator. Currently NNPAAccelerator cannot be self-register.

src/Accelerators/NNPA/Transform/ZLow/ZLowRewrite.cpp
tungld: I believe this issue is addressed by #1221 and #1222.

src/InitOMPasses.hpp
tungld: Should we address the issue of the namespace declaration in a header file with a separate PR? As you see, this PR does not modify Line 12 and is not related.

src/InitOMPasses.hpp
tungld: Done. We should have a separate PR for this, I see some other files missing the copyright.

src/InitOMPasses.hpp
tungld: Removed `mlir::`.

src/Accelerators/NNPA/Pass/DLCPasses.hpp
tungld: Agreed. Will have a separate PR for this. Anyway, this part is not code in this patch.

tungld:
@caoimhinuibrian @etiotto @sstamenova, thanks for your comments. I addressed them in the latest commit.

Regarding the following comment:
> Could we place these dialect insertions into the registry inside a function (suggest registerNNPAPasses()). Ideally we should avoid the use preprocessor macros all over the place and have something like:
> 
>    if (NNPAAccellerator.isActive()) 
>       registerNNPAPasses();
> I am concerned that over time, as more accelerators are added, we would end up with a bunch of macro checks, making the code less readable.

I agreed and tried to use NNPAAccelerator. However, it seems NNPAAccelerator cannot be self-register now. If we run `onnx-mlir --accel=NNPA`, we will get segfault. Thus, I had to use preprocessor. It looks like @caoimhinuibrian is trying to fix it with PR #1223, so we can decide whether we wait for NNPAAccelerator working or land this patch first. If this patch is landed, we can have multiple PRs to do a massive refactoring including replacing preprocessors by accelerator self-registration and make sure that the refactoring does not break lit tests for NNPA.

src/Accelerators/NNPA/Transform/ZLow/ZLowRewrite.cpp
tungld: By using OpRewritePattern we can use a derived operation instead of raw operation as in ConversionPattern. As you see, we have ZLowStickOp here instead of Operation, so we don't need to create an operator adaptor to access ZLowStickOp's members.

etiotto:
> @caoimhinuibrian @etiotto @sstamenova, thanks for your comments. I addressed them in the latest commit.
> 
> Regarding the following comment:
> 
> > Could we place these dialect insertions into the registry inside a function (suggest registerNNPAPasses()). Ideally we should avoid the use preprocessor macros all over the place and have something like:
> > if (NNPAAccellerator.isActive())
> > registerNNPAPasses();
> > I am concerned that over time, as more accelerators are added, we would end up with a bunch of macro checks, making the code less readable.
> 
> I agreed and tried to use NNPAAccelerator. However, it seems NNPAAccelerator cannot be self-registered now. If we run `onnx-mlir --accel=NNPA`, we will get segfault. Thus, I had to use preprocessor. It looks like @caoimhinuibrian is trying to fix it with PR #1223, so we can decide whether we wait for NNPAAccelerator working or land this patch first. If this patch is landed, we can have multiple PRs to do a massive refactoring including replacing preprocessors by accelerator self-registration and make sure that the refactoring does not break lit tests for NNPA.

My vote is to land this PR so that we can land https://github.com/onnx/onnx-mlir/pull/1221 and https://github.com/onnx/onnx-mlir/pull/1222 which will move us a step forward in making the new dialects self contained.

CMakeLists.txt
etiotto: Agree. I think we can make the accelerator code functionally work in this PR, and then address this point in a different PR. I am saying so because I think landing this one will allow us to make progress in parallel later.

src/InitOMPasses.hpp
etiotto: Yes.  I put the comment because I saw the code and I understand is best to land this one and then address the comment in another PR.

etiotto (approved):
This PR is a solid step forward. We have more restructuring work to do, but this PR provides the foundation for that work. LGTM!

PR #1223, opened 2022-03-04
make NNPA target "live"

caoimhinuibrian:

### DIFF
--- a/src/Accelerators/Accelerator.cpp
+++ b/src/Accelerators/Accelerator.cpp
@@ -10,17 +10,22 @@
 //
 // Accelerator base class
 //
+// To enable a new accelerator, add the header include, an extern of the
+// subclass and pushback that subclass variable onto acceleratorTargets.
 //===----------------------------------------------------------------------===//
 #include "src/Accelerators/Accelerator.hpp"
+#include "src/Accelerators/NNPA/NNPAAccelerator.hpp"
 #include <iostream>
 #include <vector>
 
 namespace mlir {
+extern NNPAAccelerator nnpaAccelerator;
 std::vector<Accelerator *> *Accelerator::acceleratorTargets;
 
 Accelerator::Accelerator() {
   if (acceleratorTargets == NULL) {
     acceleratorTargets = new std::vector<Accelerator *>();
+    acceleratorTargets->push_back(&nnpaAccelerator);
   }
 }
 
--- a/src/Accelerators/CMakeLists.txt
+++ b/src/Accelerators/CMakeLists.txt
@@ -1,9 +1,10 @@
 # SPDX-License-Identifier: Apache-2.0
-
+set(ACCEL_LIST "")
 if (ACCELERATORS_TO_BUILD)
   foreach(t ${ACCELERATORS_TO_BUILD})
     message(DEBUG "Targeting ${t}")
     add_subdirectory(${t})
+    list(APPEND ACCEL_LIST "${t}Accel")
   endforeach(t)
 else (ACCELERATORS_TO_BUILD)
 endif (ACCELERATORS_TO_BUILD)
@@ -14,4 +15,5 @@ add_onnx_mlir_library(Accelerator
       ${ONNX_MLIR_SRC_ROOT}/include
     LINK_LIBS PUBLIC
       LLVMSupport
+      ${ACCEL_LIST}
   ) 


caoimhinuibrian:
I'm trying to make it so that the accelerators self-register themselves, so that we don't have to pollute the top level code with potentially dozens of them. But, the way static linking works the libs do not get included unless I reference something from above. Fiddlinng with link options, but can't find the correct incantation. Maybe @tungld or @etiotto know how to solve this?

gongsu832:
Your problem may not be exactly the same but take a look at `src/Runtime/jni/jnidummy.c`. Basically you can make a dummy function call those routines that you want to force the linker editor to load.

caoimhinuibrian:
@alexe @etiotto please take a look. This now initializes the accelerators based on the cmake env variable, so the exxposing of accelerators in the base code is not required.

src/Compiler/CompilerUtils.cpp
```diff
@@ -1050,11 +1050,13 @@ void emitOutput(mlir::OwningOpRef<ModuleOp> &module, mlir::MLIRContext &context,
 int compileModule(mlir::OwningOpRef<ModuleOp> &module,
     mlir::MLIRContext &context, std::string outputBaseName,
     EmissionTargetType emissionTarget) {
+  extern void InitAccelerators();    
   setupModule(module, context, outputBaseName);
 
   mlir::PassManager pm(&context, mlir::OpPassManager::Nesting::Implicit);
   // Initialize accelerator if required
   if (acceleratorTarget.compare("") != 0) {
```
etiotto, line 1058: `acceleratorTarget` should be defined in this file so that it shows up in the correct section in the output of `onnx-mlir --help`.  

Propose to use `-maccel` as the option to use to select the accelerator target, as in: 
```
static llvm::cl::list<accel::Accelerator::Kind> maccel("maccel",
llvm::cl::desc("Specify an accelerator to generate code for"),
    llvm::cl::values(clEnumValN(
        accel::Accelerator::Kind::NNPA, "NNPA", "IBM Telum processor")),
```
The `Accelerator` class would have to define a strongly typed enum as in: 
```
namespace onnx_mlir {
namespace accel {

class Accelerator {
public:
  /// Kinds of accelerators.
  enum class Kind {
    NNPA, // IBM Telum coprocessor
  };
  . . .
```
Doing so would help avoid the string comparison and instead of: 
```
if (acceleratorTarget.compare("") != 0)
```
we can then write:
```
if (!maccel.empty()) {
```

src/Accelerators/NNPA/NNPAAccelerator.cpp
```diff
@@ -25,15 +25,19 @@ extern llvm::cl::OptionCategory OMDLCPassOptions;
 extern llvm::cl::opt<DLCEmissionTargetType> dlcEmissionTarget;
 extern llvm::cl::list<std::string> execNodesOnCpu;
 
+mlir::NNPAAccelerator* pnnpa;
+
+void createNNPA() {
```
etiotto, line 30: [Suggestion]: Propose to use a factory method rather than a global function.

In `Accelerator.hpp`:
```
class Accelerator {
public:
  /// Kinds of accelerators.
  enum class Kind {
    NNPA, // IBM Telum coprocessor
  };

  virtual ~Accelerator();

  /// Create a new accelerator based on \p kind, initialize it, and add it to
  /// the list of available accelerators.
  static void create(Accelerator::Kind kind,
      mlir::OwningOpRef<mlir::ModuleOp> &module, mlir::MLIRContext &context,
      mlir::PassManager &pm, onnx_mlir::EmissionTargetType emissionTarget);
...
```

In `Accelerator.cpp`:
```
void Accelerator::create(Accelerator::Kind kind,
    mlir::OwningOpRef<mlir::ModuleOp> &module, mlir::MLIRContext &context,
    mlir::PassManager &pm, onnx_mlir::EmissionTargetType emissionTarget) {
  Accelerator *accel = nullptr;
  switch (kind) {
  case Kind::NNPA:
    accel = new mlir::NNPAAccelerator(); 
    if (!accel)
      llvm::errs() << "NNPA accelerator not supported\n";
    break;
  }
  assert(accel && "Accelerator not initialized correctly");

  // Initialize the new accelerator and add it to the list of available ones.
  accel->prepare(module, context, pm, emissionTarget);
  accelerators.insert(accel);
}
```

CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ endif()
 if (POLICY CMP0116)
   cmake_policy(SET CMP0116 OLD)
 endif()
-
+set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -u nnpa_keep_alive")
```
etiotto, line 32: I think this is no longer needed ?

src/Accelerators/Accelerator.cpp
```diff
@@ -10,6 +10,8 @@
 //
 // Accelerator base class
 //
+// To enable a new accelerator, add the header include, an extern of the
+// subclass and pushback that subclass variable onto acceleratorTargets.
 //===----------------------------------------------------------------------===//
 #include "src/Accelerators/Accelerator.hpp"
 #include <iostream>
```
etiotto, line 17: at line 21 we have:
```
std::vector<Accelerator *> *Accelerator::acceleratorTargets;
```
which is an initialize pointer to a vector.  Suggest to declare a SmallPtrSet (LLVM data type) instead as a private member variable of the `Accelerator class`, that way the constructor need not `new` a new vector:

```
class Accelerator {
public:
  . . .
private:
  /// Accelerators available.
  static llvm::SmallPtrSet<Accelerator *, 2> accelerators;
};
```

The in `Accelerator.cpp`:
```
namespace onnx_mlir {
namespace accel {

llvm::SmallPtrSet<Accelerator *, 2> Accelerator::accelerators;
```

```

File: src/Compiler/CMakeLists.txt [hunk 5]
```diff
@@ -11,6 +11,21 @@ file(GENERATE
   INPUT ${CMAKE_CURRENT_BINARY_DIR}/ExternalUtil.hpp.cfg
   )
 
+file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/InitAccelerators.cpp "void InitAccelerators() {\n")
```
etiotto, line 14: A possible alternative is implemented in PR https://github.com/onnx/onnx-mlir/pull/1227. In Accelerator.cpp there is a factory method called `Accelerator::create` which builds an accelerator according to the value of a enum passed to it.  If does so by invoking `nnpa::NNPAAccelerator::getInstance()` which is defined (strongly) in `NNPAAcclerator.cpp`.  When NNPAAcclerator.cpp is __not__ built (not configured) a weak definition is provided in `Accelerator.cpp`. When NNPAAcclerator.cpp is built, it provides a strong definition for that member function.

That approach would not need the generation via cmake of `InitAccelerators.cpp`

etiotto:
Hi @caoimhinuibrian the build when shared libraries are being generated fails with the following error (circular dependency between shared libraries):
```
[91mCMake Error: The inter-target dependency graph contains the following strongly connected component (cycle):
  "Accelerator" of type SHARED_LIBRARY
    depends on "NNPAAccel" (weak)
    depends on "DLCompilerUtils" (weak)
    depends on "CompilerUtils" (weak)
  "NNPAAccel" of type SHARED_LIBRARY
    depends on "Accelerator" (weak)
    depends on "DLCompilerUtils" (weak)
    depends on "CompilerUtils" (weak)
  "DLCompilerUtils" of type SHARED_LIBRARY
    depends on "CompilerUtils" (weak)
    depends on "Accelerator" (weak)
    depends on "NNPAAccel" (weak)
  "CompilerUtils" of type SHARED_LIBRARY
    depends on "Accelerator" (weak)
    depends on "NNPAAccel" (weak)
    depends on "DLCompilerUtils" (weak)
```

I saw this error and I can offer a solution. Essentially some libraries need to be static libraries (STATIC keyword on `add_onnx_mlir_library`):

src/Accelerator/CMakefile.txt:
```
add_onnx_mlir_library(Accelerator STATIC
  Accelerator.cpp

  INCLUDE_DIRS PUBLIC
    ${ONNX_MLIR_SRC_ROOT}/include

  LINK_LIBS PUBLIC
  LLVMSupport
  ) 
  . . . 
```

src/Accelerators/NNPA/CMakeLists.txt
```
add_onnx_mlir_library(NNPAAccelerator STATIC
  NNPAAccelerator.cpp
  ...
```

src/Accelerators/NNPA/Compiler/CMakeList.txt
```
add_onnx_mlir_library(NNPACompilerUtils STATIC
  NNPACompilerUtils.cpp

...
```

src/Accelerators/NNPA/NNPAAccelerator.cpp
caoimhinuibrian: > [Suggestion]: Propose to use a factory method rather than a global function.
> 
> In `Accelerator.hpp`:
> 
> ```
> class Accelerator {
> public:
>   /// Kinds of accelerators.
>   enum class Kind {
>     NNPA, // IBM Telum coprocessor
>   };
> 
>   virtual ~Accelerator();
> 
>   /// Create a new accelerator based on \p kind, initialize it, and add it to
>   /// the list of available accelerators.
>   static void create(Accelerator::Kind kind,
>       mlir::OwningOpRef<mlir::ModuleOp> &module, mlir::MLIRContext &context,
>       mlir::PassManager &pm, onnx_mlir::EmissionTargetType emissionTarget);
> ...
> ```
> 
> In `Accelerator.cpp`:
> 
> ```
> void Accelerator::create(Accelerator::Kind kind,
>     mlir::OwningOpRef<mlir::ModuleOp> &module, mlir::MLIRContext &context,
>     mlir::PassManager &pm, onnx_mlir::EmissionTargetType emissionTarget) {
>   Accelerator *accel = nullptr;
>   switch (kind) {
>   case Kind::NNPA:
>     accel = new mlir::NNPAAccelerator(); 
>     if (!accel)
>       llvm::errs() << "NNPA accelerator not supported\n";
>     break;
>   }
>   assert(accel && "Accelerator not initialized correctly");
> 
>   // Initialize the new accelerator and add it to the list of available ones.
>   accel->prepare(module, context, pm, emissionTarget);
>   accelerators.insert(accel);
> }
> ```
So, this is what I want to avoid ... this requires modifying the Accelerator class to add a new accelerator. I'm trying to completely contain knowledge about each accelerator inside its own subdirectory.

CMakeLists.txt
caoimhinuibrian: yes, I'll remove it

src/Accelerators/Accelerator.cpp
caoimhinuibrian: @etiotto So, basically thats what I have, except that it isn't a SmallPtrSet. What is the advantage of changing to that?

The new in the constructor was an accident of debugging and I will remove that.

src/Compiler/CMakeLists.txt
caoimhinuibrian: @etiotto I addressed this suggestion in a previous response.

caoimhinuibrian:
> Hi @caoimhinuibrian the build when shared libraries are being generated fails with the following error (circular dependency between shared libraries):
> 
> ```
> �[91mCMake Error: The inter-target dependency graph contains the following strongly connected component (cycle):
>   "Accelerator" of type SHARED_LIBRARY
>     depends on "NNPAAccel" (weak)
>     depends on "DLCompilerUtils" (weak)
>     depends on "CompilerUtils" (weak)
>   "NNPAAccel" of type SHARED_LIBRARY
>     depends on "Accelerator" (weak)
>     depends on "DLCompilerUtils" (weak)
>     depends on "CompilerUtils" (weak)
>   "DLCompilerUtils" of type SHARED_LIBRARY
>     depends on "CompilerUtils" (weak)
>     depends on "Accelerator" (weak)
>     depends on "NNPAAccel" (weak)
>   "CompilerUtils" of type SHARED_LIBRARY
>     depends on "Accelerator" (weak)
>     depends on "NNPAAccel" (weak)
>     depends on "DLCompilerUtils" (weak)
> ```
> 
> I saw this error and I can offer a solution. Essentially some libraries need to be static libraries (STATIC keyword on `add_onnx_mlir_library`):
> 
> src/Accelerator/CMakefile.txt:
> 
> ```
> add_onnx_mlir_library(Accelerator STATIC
>   Accelerator.cpp
> 
>   INCLUDE_DIRS PUBLIC
>     ${ONNX_MLIR_SRC_ROOT}/include
> 
>   LINK_LIBS PUBLIC
>   LLVMSupport
>   ) 
>   . . . 
> ```
> 
> src/Accelerators/NNPA/CMakeLists.txt
> 
> ```
> add_onnx_mlir_library(NNPAAccelerator STATIC
>   NNPAAccelerator.cpp
>   ...
> ```
> 
> src/Accelerators/NNPA/Compiler/CMakeList.txt
> 
> ```
> add_onnx_mlir_library(NNPACompilerUtils STATIC
>   NNPACompilerUtils.cpp
> 
> ...
> ```

@etiotto Thanks for this. I have made that change. Just wondering though .... if someone wants an exclusively shared library implementation, how does that play? I will continue looking for the source of the circularity.

etiotto:
> @etiotto Thanks for this. I have made that change. Just wondering though .... if someone wants an exclusively shared library implementation, how does that play? I will continue looking for the source of the circularity.

@caoimhinuibrian I do not know how to elegantly solve circular dependencies between shared libs. This post (https://cmake.org/pipermail/cmake/2007-October/016753.html) indicates a way ... I haven't tried it.  However, like you, I also think the best way would be to restructure the project a bit to avoid the circular dependency all together.  

At this point, is prob. OK to force some libs to be static. These aren't the only ones in the project either, for example I see that  `cruntime`, `jniruntime` are forced to be static libraries.

src/Accelerators/Accelerator.cpp
etiotto: LLVM's SmallPtrSet etc.. "handles small sets extremely efficiently, but gracefully handles extremely large sets without loss of efficiency." So, generally speaking, these are better than the STL containers and I have seen LLVM STL 'like' containers been used quite a bit in LLVM codebase. Here is more info https://llvm.org/docs/ProgrammersManual.html#dss-smallset

src/Accelerators/NNPA/NNPAAccelerator.cpp
etiotto: OK, I see.  Using a factory method in the base class is pretty idiomatic but it does have the side effect of injecting some knowledge about the specific accelerators in the base Accelerator base class.  On way to limit such knowledge is by encapsulating it into a function:

```
void Accelerator::create(Accelerator::Kind kind,
    mlir::OwningOpRef<mlir::ModuleOp> &module, mlir::MLIRContext &context,
    mlir::PassManager &pm, onnx_mlir::EmissionTargetType emissionTarget) {
  Accelerator *accel = nullptr;
  switch (kind) {
  case Kind::ID:                                                               
    accel = nnpa::NNPAAccelerator::getInstance();                
    if (!accel)                                                                
      llvm::errs() << #ID " accelerator not supported\n";       
   }
```

where `getInstance()` is defined as a weak definition in `Accelerator.cpp` (returns nullptr) and defined to be a strong definition in `NNPAAccelerator.cpp`.  When no accelerator is built the linker uses the weak def., but when an accelerator is built the linker overwrites the weak definition with the strong one. 

In any case, I am not dead set on a particular solution. Limiting the need for the base class to "know" about derived classes is what is important.

src/Compiler/CMakeLists.txt
etiotto: Yep saw it.

src/Accelerators/NNPA/NNPAAccelerator.cpp
etiotto: Also,  using __attribute__((weak))  is not ideal either because is not standard C++. Compilers I am familiar with (IBM XL, clang, gcc) all supports the attribute extension, and are source compatible... however looks like MSVC does not accept it (it has  __declspec(selectany) which does not have the exact semantics we need because it instructs the linker to select any def. available, non necessarily the strong definition).

caoimhinuibrian:
@etiotto I changed the option to -maccel. I think this can land .... we can do the enumeration for option in a follow on PR. Could you approve it for me?

etiotto (approved):
LGTM

PR #1226, opened 2022-03-07
Fix SwapReshapeMatMulPattern rule

tungld:
The rule `SwapReshapeMatMulPattern` only works for static dimensions, so this patch adds a constraint to make sure dimensions are static. Without that, bertsquad model produces wrong results.

Supporting dynamic dimensions is left for another PR since it needs more work to compute dynamic dimensions at runtime.


etiotto (approved):
LGTM

NathanielMcVicar:
@tungld Is adding dynamic support for this rewrite still on your radar? It's become relevant to us, but I haven't grasped the level of effort involved yet.

PR #1232, opened 2022-03-09
Doxygen Java API and fixes for JNI signature wrapper

gongsu832:
- Doxygen Java API calls
- Fix onnx.ai/onnx-mlir page frame
- Fix JNI wrapper input/output signature functions


AlexandreEichenberger (approved):
LGTM, thanks for adding the doc, that is really useful. Saw that you updated the navigation files, which means that this will be avail on the onnx.ai/onnx-mlir.

Can you confirm that this will build the doc automatically? Meaning, we don't need to call a custom `make` target to rebuild the doc after any changes to the library.

gongsu832:
Yes all Java docs are generated automatically whenever you change the relevant comments, just like the C/C++ docs.

PR #1234, opened 2022-03-09
removed src/Conversion/KrnlToAffine/KrnlToAffine.cpp file

AlexandreEichenberger:
Removed file that is not needed anymore


etiotto (approved):
LGTM

etiotto:
I thought that file was removed already. In any case, it is not needed because it has been split into smaller compilation units.

PR #1237, opened 2022-03-09
fix leak when failing to allocate data

AlexandreEichenberger:
Fix leak and also remove "this is not part of the interface" comment, as it is currently part of the interface.

### DIFF
--- a/src/Runtime/OMTensor.inc
+++ b/src/Runtime/OMTensor.inc
@@ -276,8 +276,11 @@ OMTensor *omTensorCreateEmpty(
     return NULL;
 
   void *dataPtr = malloc(omTensorGetNumElems(tensor) * getDataTypeSize(dtype));
-  if (!dataPtr)
+  if (!dataPtr) {
+    // It is safe to free a null pointer (_allocatedPtr is null).
+    omTensorDestroy(tensor);
     return NULL;
+  }
 
   tensor->_alignedPtr = dataPtr;
   tensor->_allocatedPtr = dataPtr;
--- a/include/onnx-mlir/Runtime/OMTensor.h
+++ b/include/onnx-mlir/Runtime/OMTensor.h
@@ -106,9 +106,6 @@ OMTensor *omTensorCreateWithOwnership(void *data_ptr, int64_t *shape,
 /**
  * Create an OMTensor with the specified shape, rank and element type,
  * allocate uninitialized data for the specified shape.
- * This function is intentionally left out from the header because it is only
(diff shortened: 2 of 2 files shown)


src/Runtime/OMTensor.inc
```diff
@@ -276,8 +276,11 @@ OMTensor *omTensorCreateEmpty(
     return NULL;
 
   void *dataPtr = malloc(omTensorGetNumElems(tensor) * getDataTypeSize(dtype));
-  if (!dataPtr)
+  if (!dataPtr) {
+    // It is safe to free a null pointer (_allocatedPtr is null).
```
etiotto, line 280: The comment is out of context because you are not freeing a null ptr here, just calling `omTensorDestroy` (which might free `_allocatedPtr` even if NULL.  I would remove the comment.

AlexandreEichenberger: More complicated, the OMTensor is created with ownership, which means that it will try to deallocate `tensor->_allocatedPtr` which is set to null. I can remove the comment if that is too much info

tungld: I guest what you really want to say here is to free `shape` and `strides` which are already allocated for the tensor, so that we can avoid memory leak when `malloc` fails.
 
If so, we could change the order of statements a bit to avoid this situation, e.g. do malloc for `dataPtr` first and then create a tensor. Something like:
```C
  int64_t numElem = 1;
  for (int64_t i = 0; i < rank; i++)
    numElem *= shape[i];

  void *dataPtr = malloc(numElem * getDataTypeSize(dtype));
  if (!dataPtr)
    return NULL;

  OMTensor *tensor = omTensorCreateWithOwnership(dataPtr, shape, rank, dtype, /*owning=*/true);
  if (!tensor) {
    free(dataPtr);
    return NULL;
  }

  return tensor;
```

gongsu832: More concisely,
```
void *dataPtr;
OMTensor *tensor;
if ((dataPtr = malloc(getNumElems(shape, rank) * getDataTypeSize(dtype))) == NULL ||
    (tensor = omTensorCreateWithOwnership(dataPtr, shape, rank, dtype, /*owning=*/true)) == NULL) {
    free(dataPtr); // OK if dataPtr == NULL
    return NULL;
}
```

AlexandreEichenberger: Elegant, will replace with your suggestion.

tungld (approved):
LGTM!

PR #1238, opened 2022-03-10
Accumulate dynamic tensor in LoopOp

chentong319:
Tensors computed in loop body may be accumulated in a tensor with an extra dimension. For example, a loop body may return a tensor<d1,...,dnXT> and the loop will generate a tensor<IC, d1, ...,dn XT> to accumulate the result of every iteration. IC will be equal to the iteration count, either statically or dynamically. In current implementation, it is assumed that IC is known, at least dynamically, before the loop. The straightforward implementation is to allocate the tensor for accumulation before the loop body. However, this allocation can not be done when any d1, ..., dn is not statically known. The local tensor has not been computed yet and dimOf can not be used to get the dynamic size.

Current solution details:
- use memref<ICxmemref<d1, ...dnxT>> to represent the accumulate tensor.
- allocate it before the loop with IC
- In loop, the accumulation is to store the memref computed in the loop body as a whole.
- After the loop, the memref<ICxmemref<d1, ..., dn xT>> to memref<IC, d1, ..., dnxT>. For the unknown di (di == -1), we can use DimOf on any memref element loaded for the accumulated memref. The LoopOp semantics guarantees that all the local tensor should have the same shape.

Due to issue#1236, I can not test this PR with a model yet.

Future work:

- A generic solution is use SeqType if the trip count is unknown. The current solution is an optimization to avoid adding loop-carried dependence by sequence and extra copying of data.
- Another possible solution is to allocate an empty memref before loop and then re-allocate the memref at the end of first iteration. This solution will also introduce a loop-carried dependence.


src/Conversion/ONNXToKrnl/ControlFlow/Loop.cpp
```diff
@@ -189,7 +204,56 @@ struct ONNXLoopOpLowering : public ConversionPattern {
...
+        // need convertion
+        // ToDo: need a IF statement to handle output is empty
+        // we can safely give 0 to the dynamic dim for alloc
+        // Here loop is assumed to be executed at least once.
+        auto firstElement =
+            create.krnl.load(output, create.math.constantIndex(0));
+        SmallVector<mlir::Value, 4> allocParams;
+        SmallVector<int64_t, 4> dims;
+        dims.emplace_back(output.getType().cast<MemRefType>().getShape()[0]);
+        if (output.getType().cast<MemRefType>().getShape()[0] == -1)
```
tungld, line 225: How about `if (dims[0] == -1)`?

src/Conversion/ONNXToKrnl/ControlFlow/Loop.cpp
```diff
@@ -238,27 +302,51 @@ struct ONNXLoopOpLowering : public ConversionPattern {
         MultiDialectBuilder<KrnlBuilder, MemRefBuilder> create(rewriter, loc);
         auto rankedScanOutTy = memRefType;
         SmallVector<mlir::Value, 4> allocParams;
-        for (int i = 0; i < rankedScanOutTy.getRank(); i++) {
+
+        // Check the loop accumulation dimemsion
```
tungld, line 306: `dimemsion` => `dimension`

src/Conversion/ONNXToKrnl/ControlFlow/Loop.cpp
```diff
@@ -165,10 +168,22 @@ struct ONNXLoopOpLowering : public ConversionPattern {
...
-            std::get<1>(scanIntermediateToFinal),
-            /*writePrefix=*/{origIV});
+          llvm::zip(scanIntermediate, scanOutputs)) {
+        auto elementType = std::get<1>(scanIntermediateToFinal)
+                               .getType()
+                               .cast<MemRefType>()
+                               .getElementType();
+        if (elementType.dyn_cast<MemRefType>()) {
+          // accumulate dynamic tensor
+          // ToDo: May need to copy the tensor
```
tungld, line 178: I think we often use `TODO` instead of `ToDo`.

src/Conversion/ONNXToKrnl/ControlFlow/Loop.cpp
```diff
@@ -189,7 +204,56 @@ struct ONNXLoopOpLowering : public ConversionPattern {
...
+
+    rewriter.restoreInsertionPoint(afterLoop);
+    // accumulate dynamic tensor
+    // Convert the memref<?xmemref<>> to a memref
+    SmallVector<Value, 4> newOutputs;
+    for (auto output : outputs) {
+      auto seqElementType =
+          output.getType().cast<MemRefType>().getElementType();
+      if (seqElementType.isa<MemRefType>()) {
+        // need convertion
```
tungld, line 216: `convertion` => `conversion`

src/Conversion/ONNXToKrnl/ControlFlow/Loop.cpp
```diff
@@ -189,7 +204,56 @@ struct ONNXLoopOpLowering : public ConversionPattern {
...
+              firstElement.getType().cast<MemRefType>().getShape()[i]);
+          if (firstElement.getType().cast<MemRefType>().getShape()[i] == -1)
+            allocParams.emplace_back(create.mem.dim(firstElement, i));
+        }
+        ArrayRef<int64_t> shape(dims.data(), dims.size());
+        auto flatType = MemRefType::get(
+            shape, firstElement.getType().cast<MemRefType>().getElementType());
+        auto alloc = create.mem.alignedAlloc(flatType, allocParams);
+        // copy the value
+        BuildKrnlLoop loop(rewriter, loc, 1);
```
AlexandreEichenberger, line 239: I would suggest to use the new krnl loop creations createKrnl.iterate (here and above), if not too disruptive.

AlexandreEichenberger (approved):
LGTM pending changes by @tungld.

src/Conversion/ONNXToKrnl/ControlFlow/Loop.cpp
```diff
@@ -238,27 +302,51 @@ struct ONNXLoopOpLowering : public ConversionPattern {
...
+          auto elementType = rankedScanOutTy.getElementType();
+          SmallVector<int64_t, 1> dims1(rankedScanOutTy.getShape().begin() + 1,
+              rankedScanOutTy.getShape().end());
+          ArrayRef<int64_t> shape1(dims1.data(), dims1.size());
+          auto seqElementType = MemRefType::get(shape1, elementType);
+          SmallVector<int64_t, 1> dims;
+          dims.emplace_back(rankedScanOutTy.getShape()[0]);
+          ArrayRef<int64_t> shape(dims.data(), dims.size());
+          auto seqType = MemRefType::get(shape, seqElementType);
+          alloc = createMemRef.alignedAlloc(seqType, allocParams);
```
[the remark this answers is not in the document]
chentong319: The type memref<d2, ...., dnxT> is just a pointer and no need for the dynamic size.

src/Conversion/ONNXToKrnl/ControlFlow/Loop.cpp
chentong319: If you are asking dims[0] for the `output`, the dynamic value is given when the memref is allocated and we can use dim op to get it.
If you are asking dims[0] of the firstElement, the dynamic value can be retrieved with dim op when the memref is loaded from memref<*xmemref<>>. The dynamic dim size is part of the runtime data structure.

src/Conversion/ONNXToKrnl/ControlFlow/Loop.cpp
chentong319: This is a copy of old code, which uses an utility function, emitCopy, to generate code.  Update this code when emitCopy is updated to use builder instead of rewriter.

chentong319:
@tungld Thanks for the suggestions. I fixed them.

PR #1239, opened 2022-03-10
explained OM_DATA_TYPE in interface doc

AlexandreEichenberger:
We had a user of our interface which had trouble finding out the definition for OM_DATA_TYPE, which is hidden in a #defined expanded enum that is hard to read.

We currently export only 2 files for the documentation of the C interface, so I just added a short explanation of the type where it is used in the interface.


include/onnx-mlir/Runtime/OMTensor.h
```diff
@@ -86,10 +90,13 @@ OMTensor *omTensorCreate(
  * \brief Create an OMTensor with specified data pointer, shape, rank and
  * element type, manually setting data ptr ownership.
  *
- * Using this constructor, users can
- * specify whether OMTensor owns the data, which subsequently determines whether
- * the memory space underlying the data will be freed or not when OMTensor gets
- * destroyed.
+ * Using this constructor, users can specify whether OMTensor owns the data,
+ * which subsequently determines whether the memory space underlying the data
```
chentong319, line 94: Better to clarify what the memory space is, pointed by the pointers passed into this function or allocated by this function?

AlexandreEichenberger: Thanks @chentong319, you are raising a good point. Will clarify.

AlexandreEichenberger:
@chentong319 changed the doc explanation (comments in the code for doxygen extraction) trying to clarify the explanations for what you requested. I did find the use of "data" somewhat confusing as for me, the shape, rank is just as much "data" as the numerical values of the tensor. So to distinguish, I refer to "numerical data" the values that defines the actual tensor values. I added explicit statements to indicate when arrays passed as parameters (such as shape, stride) are to be managed by the user as they are actually copied internally, whereas the numerical data passed as parameter is not copied internally. Hopefully that helped, otherwise I can take another pass at it.

chentong319 (approved):
LGTM!

PR #1240, opened 2022-03-11
[NNPA] Add stickify and enable constant propagation for zhigh ops

tungld:
This patch makes stickify code open-sourced and enable constant propagation for zhigh ops including stick/unstick ops. This is the last piece of code we would like to open to make sure that all optimization and lowering passes functionally work well in onnx-mlir. 

The old `convert.c` that converts fp32 to dlf16 is replaced by a new code in `DLFf16Conversion.hpp` that supports both fp32-to-dlf16 and dlf16-to-fp32.


chentong319:
To my understanding, we added a separate pass for zhigh ops. Can we just add the rules to our constant propagation? The benefits are 1) no need to iterate over the passes if the folding interferes each other 2) avoid redundant code is utility functions.

tungld:
> To my understanding, we added a separate pass for zhigh ops. Can we just add the rules to our constant propagation? The benefits are 1) no need to iterate over the passes if the folding interferes each other 2) avoid redundant code is utility functions.

Yes, I think it would be our ultimate goal to support constant propagation for accelerators in an optimal way. However, it is still debating how to mix code among onnx-mlir and accelerators given that accelerator code is only loaded/compiled when necessary.

This patch is only to enable our existing constant propagation for zhigh as it is to make sure it functionally works in onnx-mlir.

src/Accelerators/NNPA/Transform/ZHigh/Stickify/Convert.cpp
```diff
@@ -0,0 +1,41 @@
...
+  return num_fields;
+}
+
+uint64_t fp32_to_dlf16(
+    float *fp32_data, uint16_t *dflt16_data, uint64_t num_fields) {
+  return fp32_to_dlf16_in_stride(fp32_data, dflt16_data, num_fields, 1);
+}
+
+uint64_t dlf16_to_fp32_in_stride(uint16_t *dflt16_data, float *fp32_data,
+    uint64_t num_fields, uint32_t input_stride) {
```
chentong319, line 32: can we use default value of input_stride=1 for these two function?

chentong319 (approved):
LGTM!

src/Accelerators/NNPA/Transform/ZHigh/Stickify/Convert.cpp
tungld: Done. I removed this function and put its code inside dlf16_to_fp32 directly since we only use input_stride = 1. Similarly for fp32_to_dlf16.

PR #1243, opened 2022-03-15
Correctly build OnnxMlirCompiler as a static lib

sstamenova:
When OnnxMlirCompiler is a static lib, the functions should not be marked for dllexport or dllimport.


sstamenova:
@etiotto : This is a small change that would be nice to commit, so the lib can be created with the correct dll tags on Windows (a.k.a. none for static libs).

PR #1252, opened 2022-03-22
Optionally return the number of entry points in omQueryEntryPoints

### LINKED ISSUE #1249: Optionally return a count for omQueryEntryPoints()
@tungld I think we talked about the need for a count for the `omQueryEntryPoints()` call. Since most of the time, one typically needs to go through the list of entry points anyway, so a separate call to get the count is left out since one can always just iterate the NULL-terminated list to get the count.

As it turns out, there might be situations where we need to know the count **upfront**, even if we are going to go through the list later. As in the case of the JNI wrapper, the count needs to be known before we start iterating the entry points in order to create a Java String array object upfront to hold the entry points. Of course we can always iterate the list to get the count but iterating twice feels wasteful. So there is indeed a need for the count separately, even if we plan to iterate the NULL-terminated list.

But instead of creating another API call, it might be simpler to just augment `omQueryEntryPoints()` to take an "optional" `count` parameter, i.e.,
```
const char *const *omQueryEntryPoints(int *count);
```
Basically, if a non-NULL `count` is passed in, the count is returned. Regardless, the list is always NULL-terminated.

Comments/suggestions?

tungld:
Resolve #1249 

This patch changes the function signature of `omQueryEntryPoints` from `**i8()` to `**i8(*i64)` to allow returning the number of entry points via its argument.

This patch also fixes an issue in functions `inputSignature(entryPointName)` and `outputSignature(entryPointName)` where NULL will be returned if the entry point name is not found.

### DIFF
--- a/src/Conversion/KrnlToLLVM/ConvertKrnlToLLVM.cpp
+++ b/src/Conversion/KrnlToLLVM/ConvertKrnlToLLVM.cpp
@@ -228,8 +228,8 @@ void recordEntryPointSignatures(ModuleOp &module,
 
 /// This function emits three functions: omQueryEntryPoints, omInputSignature
 /// and omOutputSignature.
-/// - omQueryEntryPoints has type of `**i8 ()` to query an array of entry point
-/// names.
+/// - omQueryEntryPoints has type of `**i8 (*i64)` to query an array of entry
+/// point names.
 /// - omInputSignature and omOutputSignature have type of type `*i8 (*i8)` to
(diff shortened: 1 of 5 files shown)


gongsu832:
Also need to update `include/onnx-mlir/Runtime/OMEntryPoint.h`

tungld:
> Also need to update include/onnx-mlir/Runtime/OMEntryPoint.h

Thanks! I forgot it. Now it is updated.

PR #1253, opened 2022-03-22
Make NNPA-related options work

tungld:
Options for NNPA in command `onnx-mlir` like `--EmitZHighIR`, `--EmitZLowIR` do not work because `onnx_mlir::EmissionTargetType` is not modified to stop compilation when NNPA options exist. This patch fixes that issue and makes the options work.


caoimhinuibrian (approved):
LGTM

PR #1255, opened 2022-03-23
add queryEntryPoints Java API

gongsu832:
- add queryEntryPoints Java API
- use GENERATE_NATIVE_HEADERS option of add_jar (require cmake 3.11+) to generate JNI header since javah was deprecated since Java 8


tungld (approved):
LGTM!

PR #1256, opened 2022-03-23
Do not set ownership for an output OMTensor that is also a block argument

tungld:
This patch solves the following problem.

Given a program where the output is also the input:
```mlir
func @main_graph(%arg0: tensor<1x1x2x2xf32>) -> tensor<1x1x2x2xf32> {
    return %arg0 : tensor<1x1x2x2xf32>
}
```

After doing inference, users may call `omTensorListDestroy` on the input and output lists to destroy them, which causes a memory issue of `double free`. This happens because both input and output refer to and own the same buffer and the buffer is then freed twice.

This patch solves the problem by setting the ownership of the output OMTensor according to whether the output comes from the input (block argument) or not. (We have a way for handling constant outputs, now it is extended to handle block arguments).


src/Conversion/KrnlToLLVM/ConvertKrnlToLLVM.cpp
```diff
@@ -106,13 +106,14 @@ void checkConstantOutputs(
   });
 
   // Check, for each output, if it was transitively produced by a constant or
-  // not.
+  // a block arugment.
```
etiotto, line 109: [typo]: arugment -> argument

etiotto (approved):
LGTM - pls fix typo before merging.

src/Conversion/KrnlToLLVM/ConvertKrnlToLLVM.cpp
tungld: Done. Thanks!

PR #1257, opened 2022-03-23
Fix computation of abs(b) for accuracy evaluation 

imaihal:
I found a bug in computation of abs(b) in the equation for accuracy evaluation. [This part](https://github.com/onnx/onnx-mlir/blob/main/src/Runtime/OMTensor.inc#L632-L633) replaces original b with abs(b), but not correct in following computation. This PR includes clean up of `omTensorAreTwoOmtsClose`, but the line fixes this bug..

--
Without this PR, when we enlarge range of input data in numerical tests, we get strange results (sign is reversed). For example in Gemm, when using the range from -10.0 to 10.0, the numerical tests failed with following errors. (default range is -1.0 to 1.0) This error can be removed by this PR.

```
a[0,0] = -1.17319 != b[0,0] = 1.17318
a[0,3] = -81.5976 != b[0,3] = 81.5976
a[0,4] = -168.018 != b[0,4] = 168.018
 :
 :
Falsifiable after 47 tests and 1 shrink

int:
14

int:
23

int:
19

int:
0

int:
0

int:
1

int:
0

int:
0

no exceptions thrown
Some of your RapidCheck properties had failures. To reproduce these, run with:
RC_PARAMS="reproduce=B8xRl1WbgkWbwxWZtVmb0FGdp9mbgM2byJXZjRnblN3c4Tbqe36xiEC+0mq3tesIhgPtp6drHLSI4Tbqe36xiECgACIgACIEA8iLAAAABMA"
```


etiotto (dismissed):
Can you add a lit test to check this change please ?

imaihal:
@etiotto This PR fixes the equation used in `omTensorAreTwoOmtsClose` to check the results of computation. In numerical tests, `omTensorAreTwoOmtsClose` is used to compare computation results with references. 
Without this fix, we may do wrong comparison. I got wrong comparison in my use case, but I don't know general case to reproduce it.  Do you know good way to check this?

etiotto:
> @etiotto This PR fixes the equation used in `omTensorAreTwoOmtsClose` to check the results of computation. In numerical tests, `omTensorAreTwoOmtsClose` is used to compare computation results with references. Without this fix, we may do wrong comparison. I got wrong comparison in my use case, but I don't know general case to reproduce it. Do you know good way to check this?

OK, so if I understand what you wrote  you do not actually have a test or model that fails to run correctly. Is that right ?
Given that the functionality this PR is changing is used by  `omTensorAreTwoOmtsClose`, which is used in numerical tests, one way to trigger the error would be to add additional inputs to the numerical test to make it fail **without** this fix.  Perhaps you can look into that direction.

etiotto (approved):
LGTM - numerical tests will exercise this code so we have coverage already.

imaihal:
@etiotto Thanks for your comments. 

> one way to trigger the error would be to add additional inputs to the numerical test to make it fail without this fix.

When we enlarge range of input data in numerical tests, we can reproduce the errors. For example in Gemm, when I used the range from -10.0 to 10.0, the numerical tests failed with following errors.  (default range is -1.0 to 1.0) This error can be removed by this PR.
So, if needed, I will update the range. Just confirming here is ok?  

- Change the range
```
diff --git a/test/modellib/GemmModel.cpp b/test/modellib/GemmModel.cpp
index 1825689f..87b1d159 100644
--- a/test/modellib/GemmModel.cpp
+++ b/test/modellib/GemmModel.cpp
@@ -83,9 +83,9 @@ bool GemmLibBuilder::prepareInputs() {
   OMTensor **list = (OMTensor **)malloc(num * sizeof(OMTensor *));
   if (!list)
     return false;
-  list[0] = omTensorCreateWithRandomData<float>(llvm::makeArrayRef(aShape));
-  list[1] = omTensorCreateWithRandomData<float>(llvm::makeArrayRef(bShape));
-  list[2] = omTensorCreateWithRandomData<float>(llvm::makeArrayRef(cShape));
+  list[0] = omTensorCreateWithRandomData<float>(llvm::makeArrayRef(aShape), -10.0, 10.0);
+  list[1] = omTensorCreateWithRandomData<float>(llvm::makeArrayRef(bShape), -10.0, 10.0);
+  list[2] = omTensorCreateWithRandomData<float>(llvm::makeArrayRef(cShape), -10.0, 10.0);
   inputs = omTensorListCreateWithOwnership(list, num, true);
   return inputs && list[0] && list[1] && list[2];
 }
```
- Error 
```
a[0,0] = -1.17319 != b[0,0] = 1.17318
a[0,3] = -81.5976 != b[0,3] = 81.5976
a[0,4] = -168.018 != b[0,4] = 168.018
 :
 :
Falsifiable after 47 tests and 1 shrink

int:
14

int:
23

int:
19

int:
0

int:
0

int:
1

int:
0

int:
0

no exceptions thrown
Some of your RapidCheck properties had failures. To reproduce these, run with:
RC_PARAMS="reproduce=B8xRl1WbgkWbwxWZtVmb0FGdp9mbgM2byJXZjRnblN3c4Tbqe36xiEC+0mq3tesIhgPtp6drHLSI4Tbqe36xiECgACIgACIEA8iLAAAABMA"

```

etiotto:
With reference to this comment: https://github.com/onnx/onnx-mlir/pull/1257#issuecomment-1078772792

@imaihal that is good. Glad you were able to trigger the error. I am fine with the proposed change, hopefully it will not cause the time required to run the numerical test too much.

imaihal:
@etiotto 

When using from -10.0 to 10.0 as the range of input data, I found that we sometimes got errors in numerical test in Gemm and MatMul.  Following is an example in Gemm. The computed result(a) and reference result(b)sometimes did not meet the equation `rtol * abs(b) + atol - abs(a-b) >= 0`. atol and rtol are 1e-5 (This equations is the same with [numpy.allclose](https://numpy.org/doc/stable/reference/generated/numpy.allclose.html). ).

 ```
attempt 66 with i 25, j 21, k 24, aTrans, bTrans, cRank 2, alpha   1.200, beta   1.000
a[4,2] = 0.553863 != b[4,2] = 0.553889
attempt 67 with i 25, j 21, k 24, aTrans, bTrans, cRank 2, alpha   1.200, beta   1.000
a[0,7] = -0.060719 != b[0,7] = -0.0607299
``` 

So, I reverted the input range to original range (-1.0 to 1.0).  Is it acceptable to merge without updating the input range?
We may need to tune atol and rtol to use different range.

imaihal:

AlexandreEichenberger:
@imaihal the PR is approved, please merge latest changes in and merge into the main branch, thanks

imaihal:
I created an issue for this. https://github.com/onnx/onnx-mlir/issues/1286

> When using from -10.0 to 10.0 as the range of input data, I found that we sometimes got errors in numerical test in Gemm and MatMul. Following is an example in Gemm. The computed result(a) and reference result(b)sometimes did not meet the equation `rtol * abs(b) + atol - abs(a-b) >= 0`. atol and rtol are 1e-5 (This equations is the same with [numpy.allclose](https://numpy.org/doc/stable/reference/generated/numpy.allclose.html). ).
> 
> ```
> attempt 66 with i 25, j 21, k 24, aTrans, bTrans, cRank 2, alpha   1.200, beta   1.000
> a[4,2] = 0.553863 != b[4,2] = 0.553889
> attempt 67 with i 25, j 21, k 24, aTrans, bTrans, cRank 2, alpha   1.200, beta   1.000
> a[0,7] = -0.060719 != b[0,7] = -0.0607299
> ```

PR #1258, opened 2022-03-23
Import ONNX constant whose value attribute is empty

tungld:
Currently onnx-mlir does not support importing an onnx constant op whose value attribute is empty. For example, it failed to import the following onnx constant node:
```
node {
    output: "337"
    name: "Constant_184"
    op_type: "Constant"
    attribute {
      name: "value"
      t {
        dims: 0
        data_type: 1
        raw_data: ""
      }
      type: TENSOR
    }
  }
```

This patch supports importing such an empty constant as
```mlir
"onnx.Constant"() {onnx_node_name = "Constant_184", value = dense<> : tensor<0xf32>} : () -> tensor<0xf32>
```


tungld:
> what are the semantics of an empty constant ?

@etiotto In ONNX, it is a tensor without raw_data as you see in onnx:
```
      t {
        dims: 0
        data_type: 1
        raw_data: ""
      }
      type: TENSOR
```
So `tensor<0xdata_type>` in MLIR is the closest representation for such an onnx tensor, and it would be semantically NoneType.

I think `"onnx.Constant"() {onnx_node_name = "Constant_184", value = dense<> : tensor<0xf32>} : () -> tensor<0xf32>` can be replaced by our `onnx.NoValue` (which is for None) by a simple rewriting rule.

In onnx-mlir, both `tensor<0xdata_type>` and `NoneType` refer to none. See https://github.com/onnx/onnx-mlir/blob/main/src/Dialect/ONNX/ONNXOpsHelper.cpp#L248

etiotto:
> > what are the semantics of an empty constant ?
> 
> @etiotto In ONNX, it is a tensor without raw_data as you see in onnx:
> 
> ```
>       t {
>         dims: 0
>         data_type: 1
>         raw_data: ""
>       }
>       type: TENSOR
> ```
> 
> So `tensor<0xdata_type>` in MLIR is the closest representation for such an onnx tensor, and it would be semantically NoneType.
> 
> I think `"onnx.Constant"() {onnx_node_name = "Constant_184", value = dense<> : tensor<0xf32>} : () -> tensor<0xf32>` can be replaced by our `onnx.NoValue` (which is for None) by a simple rewriting rule.
> 
> In onnx-mlir, both `tensor<0xdata_type>` and `NoneType` refer to none. See https://github.com/onnx/onnx-mlir/blob/main/src/Dialect/ONNX/ONNXOpsHelper.cpp#L248
> 
> > Is the model correct?
> 
> Yes, it's correct. I checked it with `onnx.checker.check_model(model)`

Ok thanks for the clarification on the semantics @tungld, then a Constant with no data is a way to materialize `NoneType`.

chentong319:
I guess that this case is introduced by converter. To my memory, when first the tensor<0xT> is introduced, the initializer.raw_data().size() is 0.  But it could be presented in a different way. It is not fully specified how to use protobuf to represent onnx ops.
By the way, the NoneType we introduced is not exactly the same with tensor<0xT>: tensor<0xT> has the type info of T.

chentong319 (approved):
LGTM!

Issue #1259, opened 2022-03-23
Cryptic error when --maccel is set but accelerator is not valid.


cjvolzka:
# Description

Error from `onnx-mlir` is very cryptic when `--maccel=X` is specified but is not valid for the build.

# Steps to reproduce:
1. Build onnx-mlir with `-DACCELERATORS_TO_BUILD=NNPA`
2. Attempt to compile a model using the accelerator. In this example, the case is wrong when compiling. (model from https://github.com/onnx/models/tree/main/vision/classification/mnist/model)
```
onnx-mlir --O3 --EmitLib --mtriple=s390x-ibm-loz --maccel=nnpa --mcpu=z14 mnist-8.onnx
```

# Actual results:
```
error: cannot be converted to LLVM IR: missing `LLVMTranslationDialectInterface` registration for dialect for op: builtin.func
Failed to translate module to LLVMIR.
```

# Expected results
```
error: <maccel_value> is not a supported accelerator. Check maccel value or recompile <cmd_name> with the required DACCELERATORS_TO_BUILD enabled.
```
Wording is open to debate but something like that should help for either typos or the accelerator option wasn't enabled when building onnx-mlir.

# Additional Note:
onnx-mlir help has:
```
  --maccel=<string>                                 - Specify an accelerator to generate code for
                                                      "NONE" or "" for no accelerator
```
Using `--maccel=` or `--maccel=""` works fine. Using `--maccel=NONE` or `--maccel="NONE"` returns the cryptic error message again.

PR #1261, opened 2022-03-23
[maccel]: Change --maccel option from a string option to a list of enums

etiotto:
This PR resolve issue https://github.com/onnx/onnx-mlir/issues/1259 by changing the type for the `--maccel` option.
Previously the `--maccel` option would accept a string as its argument and fail to diagnose illegal/unsupported accelerators. With this PR we accept only valid/supported arguments: 

```
 % onnx-mlir --help |grep -A2 maccel                                                           
  --maccel=<value>                                   - Specify an accelerator to generate code for
    =NNPA                                            -   IBM Telum processor
 ```

When an invalid argument is supplied it is diagnosed during option processing with an hard error:
```
% onnx-mlir --O3 --EmitLib  --maccel="" simplemm.onnx                                     
onnx-mlir: for the --maccel option: Cannot find option named ''!
```

### CODE REVIEW

File: src/Accelerators/Accelerator.hpp [hunk 1]
```diff
@@ -24,8 +24,17 @@ namespace accel {
 
 class Accelerator {
 public:
-  Accelerator();
+  /// Kinds of accelerators.
+  enum class Kind {
+    NNPA, // IBM Telum coprocessor
```
caoimhinuibrian, line 29: This exposes the accelerator names outside of their subdirectory. My idea for this is to have CMake generate the enum values in an include and so on ...

etiotto: This is just an enumerator so I do not understand the reason we do not want to "expose it".  It does not force the accelerator class type to be exposed.

caoimhinuibrian: Yes, but in my opinion, it is simpler to contain everything in the specific accelerator subdirectory and have CMake generate the "glue" based on the values specified in the environment variable. I can do it if you want, but if you feel strongly that this way is better I guess I can let it slide.

src/Compiler/CompilerUtils.cpp
```diff
@@ -126,6 +124,12 @@ static llvm::cl::opt<std::string> march("march",
     llvm::cl::value_desc("Target a specific architecture type"),
     llvm::cl::cat(OnnxMlirOptions), llvm::cl::ValueRequired);
 
+llvm::cl::list<accel::Accelerator::Kind> maccel("maccel",
+    llvm::cl::desc("Specify an accelerator to generate code for"),
+    llvm::cl::values(clEnumValN(
+        accel::Accelerator::Kind::NNPA, "NNPA", "IBM Telum processor")),
+    llvm::cl::cat(OnnxMlirOptions), llvm::cl::ValueRequired);
```
tungld, line 131: Can we make this optional? For example, it appears only when we build with some accelerators, i.e. `ACCELERATORS_TO_BUILD` is specified. 

If it is impossible, we need to let users know the current driver is not built with accelerator when they specify `--maccel=NNPA`

caoimhinuibrian: @tungld @etiotto that's what my approach would do

src/Accelerators/Accelerator.hpp
etiotto: @caoimhinuibrian do you have the cmake code to generate the include file you mentioned, happy to add it to the patch.

caoimhinuibrian (changes requested):
I'd like to have a meeting to talk about this

tungld (approved):
LGTM

PR #1262, opened 2022-03-24
Add a default exports file to be used for model generation

sstamenova:
On Windows generating a shared library (DLL) requires either the usage of `__declspec(dllexport)` or an exports file (`.def`). Since there's no way to add  `__declspec(dllexport)` to the generated object files, onnx-mlir currently depends on an exports file to produce DLLs.

The `.def` is assumed to be named the same as the model being compiled, but that requires any user who wants to compile a model to add a `.def` file. Since we know (mostly) what the exported functions are, we can add a default exports file and leverage it. This way the users no longer have to worry about creating the file or errors if they forget.


sstamenova:
@etiotto @AlexandreEichenberger : This change makes the usage of onnx-mlir simpler on Windows. A couple of things that I ran into that I would like your input on:
1) I placed the `.def` file together with `CompilerUtils` in the source tree. This made sense to me because this is the library that needs it. On the other hand, the headers that specify the names of some of these functions live in `include` or `include\onnx-mlir`, so either location would work. What do you think?
2) I noticed that while `OnnxMlirCompiler.h` and `OnnxMlirRuntime.h` are installed, the headers that they include are not. Perhaps they need to be installed also?
3) I made the build install `OnnxMlirExports.def` only on Windows to the `lib` directory. This makes sense because it is only needed on Windows and `onnx-mlir` and `CompilerUtils` already assume they can find the `lib` directory and use it. We could actually place the file in `bin` or even `include` instead. Thoughts?
4) It wasn't clear to me when/if the names of the exports could be different. As far as I can tell, there are 4 main exported functions right now and there names are defined in various different files. We could try to combine these somehow, so that they don't exist as definitions in multiple locations, but for the moment I just added the `.def` file without doing so. Should we figure out a way to combine them? Did I miss any for the `.def` file? Are there scenarios when they are different?

etiotto:
@sstamenova this is nice, it cleans up the code quite a bit and improves end user experience on Windows.

>I placed the .def file together with CompilerUtils in the source tree. This made sense to me because this is the library that needs it. On the other hand, the headers that specify the names of some of these functions live in include or include\onnx-mlir, so either location would work. What do you think?

I think we could colocate the .def generation with the library that provides the definitions for those functions.   The structure of `CompilerUtils` is likely to change and is probably more future proof to place the .def with library that has the definitions for those functions. 

> I noticed that while OnnxMlirCompiler.h and OnnxMlirRuntime.h are installed, the headers that they include are not. Perhaps they need to be installed also?

I think so, or else when a program include one of those header files it would not find the other headers.

> I made the build install OnnxMlirExports.def only on Windows to the lib directory. This makes sense because it is only needed on Windows and onnx-mlir and CompilerUtils already assume they can find the lib directory and use it. We could actually place the file in bin or even include instead. Thoughts?

I am not familiar with the convention on Windows but, to me, placing it in the lib directory makes sense.

> It wasn't clear to me when/if the names of the exports could be different. As far as I can tell, there are 4 main exported functions right now and there names are defined in various different files. We could try to combine these somehow, so that they don't exist as definitions in multiple locations, but for the moment I just added the .def file without doing so. Should we figure out a way to combine them? Did I miss any for the .def file? Are there scenarios when they are different?

@gongsu832 what is your opinion ?

sstamenova:
> I think we could colocate the .def generation with the library that provides the definitions for those functions. The structure of CompilerUtils is likely to change and is probably more future proof to place the .def with library that has the definitions for those functions.

@etiotto : Right now the headers that define these functions are in the `onnx-mlir/include` directory, but the library is under `onnx-mlir/src/Conversion/KrnlToLLVM`. I could move the `.def` file along with the library, but in that case perhaps the headers should move too?

gongsu832:
The .def file is logically part of the .dll so they should always go together. On z/OS we have something similar. A .dll is always accompanied by a .x file with the exported symbols.

I'm not sure about the issue with header since that should have no direct relation to where the .def is located.

gongsu832:
So you need the .def in order to produce the .dll I guess? Wouldn't it be possible to have the `OMKrnlToLLVM` code that generates the `run_main_graph`, etc. to also generate the .def on Windows?

sstamenova:
> So you need the .def in order to produce the .dll I guess? Wouldn't it be possible to have the `OMKrnlToLLVM` code that generates the `run_main_graph`, etc. to also generate the .def on Windows?

I suppose, but `OMKrnlToLLVM` doesn't directly write any files currently. They are all written by `CompilerUtils`. The header files that have the external definitions also aren't automatically generated, but they are committed directly in the repo.

One potential (more complex) solution would be to have the `.def` file be autogenerated by cmake when `OMKrnlToLLVM` is compiled based on a header (for example) that is also used by `OMKrnlToLLVM` when generating the code. This header would then be leveraged by the external header as well - combining all of the definitions in one place. If we think that's a solution we'd like eventually, I would start with the current change with the `.def` file moved alongside `OMKrnlToLLVM` and a follow up to do the autogeneration and refactor.

gongsu832:
I don't know the details of `OMKrnlToLLVM` but it certainly generates "states" that can be used later. For example, it generates the input/output signatures. So would it be possible to generate the .def and store it somewhere and then provide a function to look it up, just like the signatures?

sstamenova:
> I don't know the details of `OMKrnlToLLVM` but it certainly generates "states" that can be used later. For example, it generates the input/output signatures. So would it be possible to generate the .def and store it somewhere and then provide a function to look it up, just like the signatures?

I don't think they are equivalent because they are needed at different times.

gongsu832:
I'm just using the signatures as an example that `OMKrnlToLLVM` does save states that can be used later. Regardless, to me the most logical thing to do is whichever code that generates the .dll should also generate the .def, if possible.

sstamenova:
> I'm just using the signatures as an example that `OMKrnlToLLVM` does save states that can be used later. Regardless, to me the most logical thing to do is whichever code that generates the .dll should also generate the .def, if possible.

So my understanding is as follows:
- `OMKrnlToLLVM` relies on `mlir` to compile the module. This happens through the various passes and part of this logic adds the functions that will eventually be exported from the shared lib (such as `run_main_graph`)
- `CompilerUtils` is where the shared library is generated and written to disk by making a call to the compiler. This is the call where the `.def` file is needed
- `CompilerUtils` does not make any calls directly to `OMKrnlToLLVM`. It instead relies on the `mlir` infrastructure and the passes that are registered to compile the module and then write the necessary files to disk with the majority of steps relying on tools such as the compiler, opt, etc.

So, as far as I can tell, there's no clean way for `CompilerUtils` to query `OMKrnlToLLVM` for the exported functions and then write them to a `.def` file before compiling the shared library. At the same time, `OMKrnlToLLVM` does not write any files to disk, so that is not the correct solution either.

My current approach is to add the `.def` file along with `OMKrnlToLLVM` since that's the library that ultimately owns which exported functions are generated as @etiotto pointed out. A next step (to make maintenance simpler) would be to rely on `cmake` to generate the `.def` file as well as headers that `OMKrnlToLLVM` could use. I think it makes sense to start with the change to add the `.def` file to the source tree to simplify the Windows usage and then at a later point implement the `cmake` improvement.

gongsu832:
> So, as far as I can tell, there's no clean way for `CompilerUtils` to query `OMKrnlToLLVM` for the exported functions and then write them to a `.def` file before compiling the shared library

So I was thinking why we can't create such a function in OMKrnlToLLVM to query the exported functions, like we have for entry points and signatures? Maybe it's not as simple as I thought

@tungld ?

tungld:
> So I was thinking why we can't create such a function in OMKrnlToLLVM to query the exported functions, like we have for entry points and signatures? 

Entry points and signatures are generated when we compile a model (not onnx-mlir), which looks a bit different here where CompileUtils needs a .def file during the onnx-mlir build.

I feel that the current solution in this patch looks reasonable.

gongsu832:
> Entry points and signatures are generated when we compile a model (not onnx-mlir), which looks a bit different here where CompileUtils needs a .def file during the onnx-mlir build.
> 
> I feel that the current solution in this patch looks reasonable.

OK so the slight complication comes from the fact that the list of exported symbols contains two sets,

- the 1st set, `run_main_graph` (+ other entry points), `omQueryEntryPoints`, `omInputSignature`, and `omOutputSignature`, comes from compiling the model, which is logically associated with `OMKrnlToLLVM`
- the 2nd set, `omTensorCreate`, `omTensorGetDataPtr`, and `omTensorListCreate`, etc, comes from `libcruntime.a`

So the .def file can be put in one of the two places,

- if we put it where the exported symbols are utilized, it would be under `src/Compiler`
- if we put it where the exported symbols are provided, it would be under `src/Runtime`

Either place is reasonable. But I think it would be better not have to drag along a .def file at runtime at all. Basically, the content of the .def file should be embedded into `CompilerUtils.cpp`, either hard coded or somehow generated and included. This is more or less inline with what @sstamenova suggested as the followup autogen solution. Although for now, I wouldn't bother with a .def file, I would just hard code the exported symbols in a header file and have `CompilerUtils.cpp` generate the .def file at runtime on Windows.

AlexandreEichenberger:
> Although for now, I wouldn't bother with a .def file, I would just hard code the exported symbols in a header file and have CompilerUtils.cpp generate the .def file at runtime on Windows.

That seems easy enough: when we compile for Window, the compiler driver found in CompilerUtils.cpp also generate a .def file in the directory where the object are going with "all" of the current entry points. If that is the solution used, we just need to add some warnings in the interface files that if we add a new dll entry point, one should also add it in the right place in the CompilerUtils. Alternatively, once could define a list of strings to be processed in the compiler and the runtime utils, and scan these lists when generating the .def file.

We have support for multiple entry point in a single onnx model file (@tungld added support for this recently), so we need a bit of extra work when multiple entry points are used. We have access to this info when generating the signatures, so a similar mechanism has to be used here.

gongsu832:
> We have support for multiple entry point in a single onnx model file (@tungld added support for this recently), so we need a bit of extra work when multiple entry points are used. We have access to this info when generating the signatures, so a similar mechanism has to be used here.

`CompilerUtils.cpp` should be able to call `omQueryEntryPoints`.

sstamenova:
> omQueryEntryPoints

No, `CompilerUtils.cpp` can't call `omQueryEntryPoints` because that's one of the functions that hasn't yet been compiled or exported (`omQueryEntryPoints` belongs in the `.def` file).

I can add the list of exported functions to `CompilerUtils.cpp` as a hard-coded list and generate the `.def` file that way if that's preferred, but I still think it's better to do it in two stages:
1) Remove all the `.def` files that are currently added for the tests in various places and replace them with a single `.def` file that lives with the library
2) Autogenerate the `.def` file as well as the external functions list when compiling the model, so that they can stay in sync.

sstamenova:
> * the 2nd set, `omTensorCreate`, `omTensorGetDataPtr`, and `omTensorListCreate`, etc, comes from `libcruntime.a`
> 
> So the .def file can be put in one of the two places,

I think the two lists are fundamentally different and don't need to be exposed the same way. The set of functions exported by `libcruntime.a` are statically linked into the executable itself, while the set of functions exported by the model as dynamically retrieved from the shared library with the model. I wasn't able to find example (at least not once that are currently being compiled in the tree) of a scenario where the functions exported by `libcruntime.a` are dynamically linked from the model library instead of `libcruntime.a`. I am not as familiar with this aspect of the libraries, so I could have missed something.

gongsu832:
> No, `CompilerUtils.cpp` can't call `omQueryEntryPoints` because that's one of the functions that hasn't yet been compiled or exported (`omQueryEntryPoints` belongs in the `.def` file).
> 
> I can add the list of exported functions to `CompilerUtils.cpp` as a hard-coded list and generate the `.def` file that way if that's preferred, but I still think it's better to do it in two stages:
> 
>     1. Remove all the `.def` files that are currently added for the tests in various places and replace them with a single `.def` file that lives with the library
> 
>     2. Autogenerate the `.def` file as well as the external functions list when compiling the model, so that they can stay in sync.

OK I guess there is a catch 22 problem on Windows. `omQueryEntryPoints` is generated from the bitcode (I think) and kind of "external" to `CompilerUtils.cpp`. So it needs a .def to call `omQueryEntryPoints` to begin with. Hmm...needs a bit more thinking.

gongsu832:
So on the path of thinking whether it's possible to make this work on Windows without a .def file.

Obviously for the 2nd set of symbols, it's not a problem since we can annotate all the functions with `__declspec(dllexport)` on Windows. For the 1st set, it appears that LLVM does actually support annotating IR according to this https://reviews.llvm.org/D1099 ? Obviously we need our compiler experts to see if this is feasible or not. @tungld ?

tungld:
@gongsu832 good finding!

It looks like we can embed `__declspec(dllexport)` into bitcode when `mtriple` is windows. These are examples for clang: https://github.com/llvm/llvm-project/blob/main/clang/test/CodeGenCXX/dllexport-members.cpp#L29. So theoretically, we can do the same for onnx-mlir. Will look at this in detail tomorrow.

@sstamenova could you give a sample C++ code for symbols we want to export if we use `__declspec(dllexport)`? Thanks!

tungld:
@sstamenova @gongsu832 @AlexandreEichenberger see PR #1307. We can annotate exported functions in  the generated .dll on Windows so that there is no need of .def files.

sstamenova:
Closing because #1307 supersedes this.

PR #1263, opened 2022-03-24
[Accelerator] Do not use NNPA preprocessor to avoid exposing accelerator code

tungld:
This patch is to remove NNPA preprocessor so that accelerator code is not exposed outside its folder.

`InitAccelerators` function is modified to return `bool` instead of `void`. It returns True if there are accelerators initialized. Otherwise, False.

`prepareAccelerator` function is separated into two functions `getOrLoadDialects` and `addPasses` to exactly reflect the function's content.

`OptimizationLevel` option is moved to a common place, that is `OMOption`, to be shared between `onnx-mlir` and `onnx-mlir-opt`, which avoids duplicated declaration.


gongsu832:

gongsu832:

src/Compiler/CompilerUtils.cpp
```diff
@@ -126,14 +126,6 @@ static llvm::cl::opt<std::string> march("march",
     llvm::cl::value_desc("Target a specific architecture type"),
     llvm::cl::cat(OnnxMlirOptions), llvm::cl::ValueRequired);
 
-static llvm::cl::opt<OptLevel> OptimizationLevel(
```
etiotto, line 129: I think we should keep all options that are exposed to the user together. `onnx --help` should show them all under the same category (`OnnxMlirOptions`).  Is my understanding that in `OMOptions.cpp` we should have options that  "provide fine control on passes".  Perhaps you can make this option non-static and expose it to the accelerator via an extern declaration?

src/Support/OMOptions.cpp
```diff
@@ -45,3 +47,11 @@ llvm::cl::opt<std::string> acceleratorTarget("maccel",
     llvm::cl::desc("Specify an accelerator to generate code for\n"
                    "\"NONE\" or \"\" for no accelerator\n"),
     llvm::cl::init(""), llvm::cl::cat(OMPassOptions));
+
+llvm::cl::opt<OptLevel> OptimizationLevel(
```
etiotto, line 51: See my other comment. I think we should keep user facing options together and and use extern declarations to expose it to the accelerator code.  Options that are user facing should show in the --help output under the same category.

src/Tools/onnx-mlir-opt/onnx-mlir-opt.cpp
```diff
@@ -86,13 +75,16 @@ void scanAndSetOptLevel(int argc, char **argv) {
...
-      OptimizationLevel = (OptLevel)num;
+      OptimizationLevel = (onnx_mlir::OptLevel)num;
       return;
     }
   }
 }
 
 int main(int argc, char **argv) {
+  // Initialize accelerators if they exist.
+  bool hasAccelerator = InitAccelerators();
```
etiotto, line 86: [nit]: hasAccelerator -> hasAccelerators. (there could be more than one in the future).

src/Support/OMOptions.cpp
```diff
@@ -14,6 +14,8 @@
 
 #include "src/Support/OMOptions.hpp"
 
+using namespace onnx_mlir;
```
etiotto, line 17: Reminder to update copyright year as well from 2021 to 2022.

src/Compiler/CompilerUtils.cpp
```diff
@@ -1071,7 +1062,8 @@ int compileModule(mlir::OwningOpRef<ModuleOp> &module,
     // for (auto accel : *accTargets) {
```
etiotto, line 1062: Should prob. cleanup commented out code.

etiotto (commented):
Is great we are removing the need to #ifdef  out the accelerator code. So I like the direction very much. 

We have an issue with the organization of compiler options. We need to expose compiler options to the main driver `onnx-mlir` as well as the tool `onnx-mlir-opt` and some of the accelerator code.  Would be nice to keep all options in a common library, I think we might need to pull the compiler options declaration into a separate header file, and "package" their definitions into a thin library that can be then linked by both `onnx-mlir` and `onnx-mlir-opt` (and perhaps by NNPACompilerUtils as well).

src/Compiler/CompilerUtils.cpp
tungld: I agree, options are now quite scattered. I think it requires more work to reorganize options, so I have no plan to do it in this patch.

src/Tools/onnx-mlir-opt/onnx-mlir-opt.cpp
tungld: Done. Thanks!

src/Support/OMOptions.cpp
tungld: Done. Thanks!

caoimhinuibrian (changes requested):
I'd like to have a meeting to talk about this

caoimhinuibrian (approved):
LGTM .... sorry, I misunderstood what you were doing

PR #1264, opened 2022-03-24
[backend_cpp]: Use ModelLib to create CategoryMapper cpp tests.

etiotto:
This PR modifies `TestCategoryMapper.cpp` to make it use `ModelLibBuilder` like the numeric and performance tests.  The old ModelBuilder.hpp and ModelBuilder.cpp files are no longer needed with this PR.


test/backend-cpp/CMakeLists.txt
```diff
@@ -45,10 +45,9 @@ function(add_backend_cpp_unittest test_name)
   endif()
 endfunction()
 
-set(TEST_LINK_LIBS CompilerUtils ExecutionSession)
+set(TEST_LINK_LIBS ModelLib)
```
sstamenova, line 48: You could just reference the lib itself below since there's just the one left.

etiotto: done

test/modellib/ModelLib.cpp
```diff
@@ -20,11 +20,12 @@
 #include "src/Runtime/OMTensorHelper.h"
 #include "test/modellib/ModelLib.hpp"
 
-using namespace std;
 using namespace mlir;
-using namespace onnx_mlir;
 
-ModelLibBuilder::ModelLibBuilder(const string &name)
+namespace onnx_mlir {
+namespace test {
```
AlexandreEichenberger, line 26: Curiosity: why do we use `using namespace mlir` but `namespace onnx_mlir { }`

test/modellib/ConvModel.cpp
```diff
@@ -21,9 +21,10 @@
 #include "src/Runtime/OMTensorHelper.h"
 #include "test/modellib/ModelLib.hpp"
 
-using namespace std;
 using namespace mlir;
-using namespace onnx_mlir;
+
+namespace onnx_mlir {
+namespace test {
```
AlexandreEichenberger, line 27: Thanks for using the `test` namespace.

test/modellib/ConvModel.cpp
```diff
@@ -37,8 +38,8 @@ Conv2DLibBuilder::Conv2DLibBuilder(const std::string &modelName, const int N,
       autoPad(autoPad), pHBegin(pHBegin), pHEnd(pHEnd), pWBegin(pWBegin),
       pWEnd(pWEnd), stride(stride), dilation(dilation), isDynamic(isDynamic) {}
 
-const string Conv2DLibBuilder::getAutoPadName(const ConvAutoPad autoPad) {
-  static const string autoPadName[] = {
+const std::string Conv2DLibBuilder::getAutoPadName(const ConvAutoPad autoPad) {
```
AlexandreEichenberger, line 41: Again, just trying to understand the style here. Did it not compile without the `std` qualifier?

test/modellib/ModelLib.hpp
```diff
@@ -120,6 +124,63 @@ class ModelLibBuilder {
...
+  // Verify that the output tensor has the expected rank.
+  bool verifyRank(const OMTensor &out, int64_t rank) const;
+
+  // Verify that the output tensor has the expected number of elements.
+  bool verifyNumElements(const OMTensor &out, int64_t numElems) const;
+
+  // Verify that the output tensor contains the expected result.
+  bool verifyResults(const OMTensor *out, const OMTensor *expected) const;
+
+private:
```
AlexandreEichenberger, line 178: nit: repeated `private`

test/numerical/TestLSTM.cpp
```diff
@@ -31,7 +30,13 @@ bool isOMLSTMTheSameAsNaiveImplFor(const int direction, const int S,
          lstm.run() && lstm.verifyOutputs();
 }
 
+} // namespace test
+} // namespace onnx_mlir
+
 int main(int argc, char *argv[]) {
+  using namespace onnx_mlir;
+  using namespace onnx_mlir::test;
+
```
AlexandreEichenberger, line 39: question for my own understanding: is it a problem if main is inside of the test and onnx_mlir namespace?

test/perf/PerfConv.cpp
```diff
@@ -16,18 +16,12 @@
 //
 //===----------------------------------------------------------------------===//
 
-#include <cassert>
-#include <iostream>
-#include <string>
```
AlexandreEichenberger, line 21: Question: is it not good form to include things that are needed in a file, even if they may be included implicitly in other header files? Or is it preferred to aim for conciseness.

AlexandreEichenberger (approved):
LGTM, mostly asked questions about style, to better understand some of the style changes.

test/modellib/ModelLib.cpp
etiotto: I prefer 'wrapping' declarations with `namespace { ... }` rather than adding a `using declaration` because, if a function is declared in the file, and is enclosed in `namespace { ... }` it is declared **in** the namespace. Conversely, is there was just a using declaration, the function would be declared in the global namespace, which is something we should avoid polluting.

test/modellib/ConvModel.cpp
etiotto: yw

test/perf/PerfConv.cpp
etiotto: This is arguable. Some argue that is better to include all headers that are required to compile the `.cpp` file, and avoid relying on transitive inclusion.  In this case <string> is already included in `ModelLib.hpp` which is the header file that is most closely associated with this `.cpp` file so I went for conciseness. And the <iostream> header is not required for this .cpp file.

test/modellib/ModelLib.hpp
etiotto: I added it to separate the declaration of private member functions from the declaration of private member variables. Is not required, but personally it helps me read the code.

Issue #1254, opened 2022-03-22
Memory issue with memref<memref<>>


chentong319:
memref<memref<>> is used to implement onnx sequence. I found that the Bufferization::Deallocate pass can not handle such memref correctly. The underlying reason is that a memref is "escaped" when it is stored into memref<memref<>>. In detail, there three cases:

1.  Deallocate the escape memref.  
      ```
      memref.store (%12, %2, %arg3) : (memref<1xf32>, memref<?xmemref<1xf32>>, index) -> ()
      memref.dealloc %12 : memref<1xf32>
      ```
2. Failed to deallocate the memref extracted from memref<memref<>>
3. No deep deallocation when memref<memref<>> is deallocated

Solution
1. Introduce sequence op which will be lowered to memref ops after deallocation
2. SeqStore for writing to a sequence. The input tensor will be copied inside SeqStore.
3. SeqExtract for extracting an element tensor from the sequence. This op is given AllocationOpInterface and will be treated as allocation by deallocation pass
4. SeqCreate for creating a sequence. This op is also give AllocationOpInterface and the deep deallocation can be defined in its buildDealloc() function

Implementation details:

- We need a separate conversion pass, which is after Deallocation and before KrnlToLLVM. The reason is that the type converter in KrnlToLLVM will make it difficult to lower the seq ops at llvm level.
- The Krnl Ops are lowered in different stages. Should they be put into different groups/dialects?

etiotto:
@chentong319 I tried to reproduce the behaviour described above with this simple MLIR test:
```
func @allocation(%arg0: index) {
  %0 = memref.alloc() : memref<1xf32> 
  %1 = memref.alloc() : memref<2xmemref<1xf32>> 
  memref.store %0, %1[%arg0] :  memref<2xmemref<1xf32>> 
  return
 } 
 ```
 
The test allocates a memref of one element (%0) and stores it into another memref (%1). 
If I use `mlir-opt` and run the buffer deallocation pass `-buffer-deallocation` I see the pass injecting the 2 deallocations:

% ~/Source/onnx/llvm-project/build/bin/mlir-opt memreftest.mlir -buffer-deallocation --print-ir-before-all  --print-ir-after-all 
```
// -----// IR Dump Before BufferDeallocation //----- //
func @allocation(%arg0: index) {
  %0 = memref.alloc() : memref<1xf32>
  %1 = memref.alloc() : memref<2xmemref<1xf32>>
[...]

PR #1265, opened 2022-03-24
Lower krnlSeq ops to memref

chentong319:
The ops for sequence are introduced due to the memref deallocation issue discussed  in #1254 
This PR includes
1. definition of the Ops in Krnl

2. Lowering these Ops to memref after bufferization::Deallocation
3.  lit test cases

Only two ops for sequence are defined in this PR. These Ops are used to support to accumulate tensor with dynamic shape for LoopOp. 
New sequence related Ops may be introduced to support sequence related Ops in ONNX.


caoimhinuibrian (approved):
LGTM

PR #1267, opened 2022-03-25
Add the implementation of shape inference for max_roi_pool op

DavidChan0519:
Signed-off-by: david.cheng <david.cheng@enflame-tech.com>


tungld:

tungld:
We are not going to write shape inference code directly in ONNXOps.cpp but put them into folder `src/Dialect/ONNX/ShapeInference`. Also, we should use IndexExpr to compute shape inference because it will be used to allocate memory in the op lowering.

You can take RoiAlign as an example: https://github.com/onnx/onnx-mlir/blob/main/src/Dialect/ONNX/ShapeInference/RoiAlign.cpp

DavidChan0519:
> We are not going to write shape inference code directly in ONNXOps.cpp but put them into folder `src/Dialect/ONNX/ShapeInference`. Also, we should use IndexExpr to compute shape inference because it will be used to allocate memory in the op lowering.
> 
> You can take RoiAlign as an example: [https://github.com/onnx/onnx-mlir/blob/main/src/Dialect/ONNX/ShapeInference/RoiAlign.cpp](https://github.com/onnx/onnx-mlir/blob/main/src/Dialect/ONNX/ShapeInference/RoiAlign.cpp?rgh-link-date=2022-03-25T04%3A20%3A52Z)

got it,  I will modify it later

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -3960,7 +3960,41 @@ LogicalResult ONNXMaxPoolOp::inferShapes(
...
+
+  // 2d tensor: (num_rois, 5)
+  auto roi_shape = rois().getType().cast<RankedTensorType>().getShape();
+  int64_t num_rois = roi_shape[0];
+  SmallVector<int64_t, 2> pooled_dims;
+
+  auto pooled_shape_array_attr = pooled_shape();
+  for (auto pooled_shape_attr : pooled_shape_array_attr) {
+    auto pooled_shape_int_attr = pooled_shape_attr.dyn_cast<IntegerAttr>();
+    if (!pooled_shape_int_attr)
```
chentong319, line 3983: Is this case an error?

chentong319 (approved):
LGTM!

AlexandreEichenberger:
@DavidChan0519 are you able and willing to manually merge this PR?

DavidChan0519:
> @DavidChan0519 are you able and willing to manually merge this PR?

I will resolve the conflict and then submit it again.

tungld:

tungld:
@DavidChan0519 could you please sign DCO so that we can merge this? Thanks!

tungld:

tungld:
@DavidChan0519 this is the suggested instruction to fix DCO: https://github.com/onnx/onnx-mlir/pull/1267/checks?check_run_id=8105622451

PR #1268, opened 2022-03-25
[NNPA][WIP] Enable and support bidirectional LSTM/GRU operations on NNPA.

negiyas:
Closing this PR, since it has been replaced by PR#1471.

Enable and support bidirectional LSTM/GRU operations on NNPA.

TODO
- [X] Set the prev_layer_flag for zlow.lstm/gru ops according to the zlow-level investigation.
- [X] Add conversion test cases about prev_layer setting.
- [ ] Enable the numerical tests on NNPA after PR#1335 is merged.
- [ ] Execute the numerical test on NNPA, and confirm the results.


src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
```diff
@@ -612,6 +614,11 @@ bool isSuitableForZDNN<ONNXGRUOp>(ONNXGRUOp op) {
   Value R = op.R();
   Value B = op.B();
 
+  // Check direction.
+  if ((direction != FORWARD) && (direction != REVERSE) &&
```
etiotto, line 618: A few general comments:
 
 ```auto direction = op.direction():  ```
 
according to LLVM coding rules, which we strive to follow, we should not use 'auto' everywhere. Specifically auto should not be used when the type of the RHS expression is not clear from the context. Using the actual type helps the reader of the code,  `auto` should be used when either the RHS type is obvious from the context, or the type of the RHS expression is 'complex' (like when declaring a lambda function).  

This is not specific to this PR, but I suggest we progressively review the code (especially the accelerator code because is new in this project) and, whenever possible, fixup this issue so that, over time, we can cleanup the code. 

In this case the type of `direction` is StringRef, which surprised me because I would expect it to be an `enum class` given that there is only a limited set of values a 'direction' can have.  Would it be possible to change it to an enum ?  Using an enum is cleaner IMO and also has a performance benefit because it is less expensive to do comparisons against enum values that it is to compare strings.

I also suggest to change the `if` into an assert because `direction` can only be one of FORWARD, REVERSE, or BIDIRECTIONAL, so any other value should not be possible. What do you think?

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.cpp
```diff
@@ -126,28 +126,53 @@ Value getLSTMGRUGetYh(Location loc, PatternRewriter &rewriter, Value val,
...
+  Type elementType = resYh.getType().cast<ShapedType>().getElementType();
   Value axis = zero;
   Value step = one;
-  ONNXSliceOp sliceOp =
-      rewriter.create<ONNXSliceOp>(loc, sliceType, val, start, end, axis, step);
-  ONNXSqueezeV11Op squeezeOp = rewriter.create<ONNXSqueezeV11Op>(
-      loc, resYh.getType(), sliceOp.getResult(), rewriter.getI64ArrayAttr(0));
-  return squeezeOp.getResult();
+  Value ret;
+  if (directionStr.equals_insensitive("forward") ||
```
etiotto, line 139: This string comparisons can be avoided by making the type of direction a C++11 `enum class`.  The code here can then be rewritten as: 

```
switch (direction) {
case Direction::reverse:
  ...
  break;
case Direction::forward:
   ... 
   break;
   etc...
}
```

src/Accelerators/NNPA/Conversion/ZHighToZLow/ZHighToZLow.cpp
```diff
@@ -410,6 +415,9 @@ ZMemRefType convertZTensorToMemRefType(OpBuilder b, Type type) {
         h = (b.getAffineDimExpr(e1) +
              pad_size * (b.getAffineDimExpr(e1).floorDiv(constExprS)))
                 .floorDiv(constExpr64);
+#if 1
```
etiotto, line 418: Remove

src/Accelerators/NNPA/Conversion/ZLowToLLVM/ZLowToLLVM.cpp
```diff
@@ -411,20 +411,24 @@ class ZLowLSTMLowering : public ConvertToLLVMPattern {
     // Direction input.
     Value direction;
     StringRef directionStr = dyn_cast_or_null<ZLowLSTMOp>(op).direction();
-    if (directionStr.equals_insensitive("forward"))
+    if (directionStr.equals_insensitive("forward")) {
```
etiotto, line 414: See my other comments, please use a switch statement

src/Accelerators/NNPA/Conversion/ZHighToZLow/ZHighToZLow.cpp
negiyas: It was temporal fix for WIP draft. This part has been fixed by more logical way. Thank you.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.cpp
negiyas: I am thinking to create another PR to use enum class for the direction values in zHigh and zlow directives as same to the above.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
negiyas: I removed "auto" from the declaration. Thank you.

As for use of `enum class`, do you mean that we should use `enum class` in ZHighOps.td and ZLowOps.td,
or we should use `enum class` only in C++ programs?
To use `enum class` in ZHighOps.td and ZLowOps.td, large code modification is required, because the current ZHighOps.td and ZLowOps.td use `StringAttr` for several places (e.g. layout, padding_type, direction and act_func), so that I will create another PR. Can you give me comments or suggestions?

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.cpp
negiyas: Yes. I will use switch statement after using `enum class` for direction. Thank you.

src/Accelerators/NNPA/Conversion/ZLowToLLVM/ZLowToLLVM.cpp
negiyas: Sure.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
etiotto: > I removed "auto" from the declaration. Thank you.

Thanks @negiyas. 

> As for use of enum class, do you mean that we should use enum class in ZHighOps.td and ZLowOps.td,
or we should use enum class only in C++ programs?
To use enum class in ZHighOps.td and ZLowOps.td, large code modification is required, because the current ZHighOps.td and ZLowOps.td use StringAttr for several places (e.g. layout, padding_type, direction and act_func), so that I will create another PR. Can you give me comments or suggestions?

As I mentioned in this thread I find unintuitive that `direction` is a string, because it has a limited number of possible values, which could be modeled using an enumeration (either a pre C++11 `enum` or, preferably, a C++11 enum class). Using enumerators would allow us to avoid string comparisons which are more costly.

I checked the ONNX specification for LSTM (other operators like GRU are similar). The `direction` attribute is defined to be a string:

```
direction : string (default is forward)
Specify if the RNN is forward, reverse, or bidirectional. Must be one of forward (default), reverse, or bidirectional.
```

So the current code "mirrors" what is in the ONNX specification closely. Internally we could still model the direction as an enum, but I realize this would be a pervasive change. 

In summary, I feel we should review the internal implementation for the direction attribute, but I agree at this point we should do that in a future pull request.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.cpp
etiotto: I now think we should do that in a future PR for the reasons I outlined in this [thread](https://github.com/onnx/onnx-mlir/pull/1268/files#r835268155).

src/Accelerators/NNPA/Conversion/ZHighToZLow/ZHighToZLow.cpp
```diff
@@ -407,8 +412,9 @@ ZMemRefType convertZTensorToMemRefType(OpBuilder b, Type type) {
           llvm_unreachable("Unsupported rank in ZDNN_FICO layout");
         }
         n = constExpr0;
-        h = (b.getAffineDimExpr(e1) +
-             pad_size * (b.getAffineDimExpr(e1).floorDiv(constExprS)))
+        h = (shape[0] *
```
tungld, line 415: Is there a case where `shape[0]` is unknown (`shape[0] == -1`)? Please add a check for `shape[0]` and handle it.

src/Accelerators/NNPA/Conversion/ZHighToZLow/ZHighToZLow.cpp
```diff
@@ -331,18 +331,23 @@ ZMemRefType convertZTensorToMemRefType(OpBuilder b, Type type) {
...
+        // -> (e4, ceil((2 * PADDED(e1))/64), e3, ceil(e2/32), 32, 64)
         e4 = 0;
         e3 = 1;
         e2 = 2;
         e1 = 3;
         n = b.getAffineDimExpr(e4);
-        h = b.getAffineDimExpr(e1).floorDiv(constExpr64);
+        if (shape[1] == 1) {
+          h = b.getAffineDimExpr(e1).floorDiv(constExpr64);
+        } else {
```
tungld, line 346: Does this `else` can handle `shape[1] == -1`?

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.td
```diff
@@ -616,15 +619,16 @@ def replaceONNXLSTMPattern1 : Pattern<
...
+      (GetStrAttr<"none">)),                             // prev_layer
 
    //
    // Post-process to generate three return values of ONNXLSTMOp
    // from two return values of ZHighLSTMOp(=$zHighLSTM).
    //
    (ZHighUnstickOp:$unstick_HnOutput $zhighLSTM__0, (returnType $res__0)),
    (GetLSTMGRUGetY $unstick_HnOutput, $res__0),
-   (GetLSTMGRUGetYh $unstick_HnOutput, $res__0, $res__1, $direction),
+   (GetLSTMGRUGetYh $unstick_HnOutput, $res__0, $res__1, (GetShape $x), $direction),
```
tungld, line 631: I would pass `$x` instead of `GetShape $x` to be consistent with the other arguments. The same for other places.

test/mlir/accelerators/nnpa/conversion/onnx-to-zhigh/lstm.mlir
```diff
@@ -71,6 +71,45 @@ func @test_onnx_to_zhigh_ccfd0_reverse(%X: tensor<7x2000x204xf32>, %W: tensor<1x
...
+// CHECK-DAG:       [[VAR_18_:%.+]] = "onnx.Concat"([[VAR_16_]], [[VAR_17_]]) {axis = 1 : si64} : (tensor<1x1x2000x200xf32>, tensor<1x1x2000x200xf32>) -> tensor<1x2x2000x200xf32>
+// CHECK-DAG:       [[VAR_19_:%.+]] = "onnx.SqueezeV11"([[VAR_18_]]) {axes = [0]} : (tensor<1x2x2000x200xf32>) -> tensor<2x2000x200xf32>
+// CHECK-DAG:       [[VAR_20_:%.+]] = "zhigh.Unstick"([[VAR_cf_output_]]) : (tensor<*xf32>) -> tensor<*xf32>
+// CHECK-DAG:       [[VAR_21_:%.+]] = "onnx.SqueezeV11"([[VAR_20_]]) {axes = [0]} : (tensor<*xf32>) -> tensor<2x2000x200xf32>
+// CHECK:           return [[VAR_10_]], [[VAR_19_]], [[VAR_21_]] : tensor<7x2x2000x200xf32>, tensor<2x2000x200xf32>, tensor<2x2000x200xf32>
+// CHECK:         }
+}
+
+// -----
+
```
tungld, line 112: Could you please another tests for bidirectional LSTM  with unknown dimensions? The same for GRU and the same for zhigh-to-zlow tests for LSTM/GRU. Thanks!

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.cpp
```diff
@@ -126,28 +126,53 @@ Value getLSTMGRUGetYh(Location loc, PatternRewriter &rewriter, Value val,
...
+        loc, sliceType, val, start, end, axis, step);
+    ret = rewriter.create<ONNXSqueezeV11Op>(
+        loc, resYh.getType(), sliceOp.getResult(), rewriter.getI64ArrayAttr(0));
+  } else if (directionStr.equals_insensitive("bidirectional")) {
+    SmallVector<int64_t> splitShape({T, 1, B, H});
+    Type splitType = RankedTensorType::get(splitShape, elementType);
+    SmallVector<Type> splitTypes = {splitType, splitType};
+    int64_t splitAxis = 1;
+    ONNXSplitV11Op splitOp = rewriter.create<ONNXSplitV11Op>(
+        loc, splitTypes, val, splitAxis, nullptr);
```
tungld, line 156: splitAxis has only a single use. Please directly use `/*splitAxis=*/1`.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.cpp
```diff
@@ -126,28 +126,53 @@ Value getLSTMGRUGetYh(Location loc, PatternRewriter &rewriter, Value val,
...
+    Type sliceType = RankedTensorType::get(sliceShape, elementType);
+    Value fwdLastSlice = rewriter.create<ONNXSliceOp>(
+        loc, sliceType, splitOp.getResults()[0], minusOne, intMax, axis, step);
+    Value bkwFirstSlice = rewriter.create<ONNXSliceOp>(
+        loc, sliceType, splitOp.getResults()[1], zero, one, axis, step);
+    SmallVector<int64_t> concatShape({1, D, B, H});
+    Type concatType = RankedTensorType::get(concatShape, elementType);
+    int64_t concatAxis = 1;
+    Value concatOp = rewriter.create<ONNXConcatOp>(
+        loc, concatType, ValueRange({fwdLastSlice, bkwFirstSlice}), concatAxis);
```
tungld, line 167: concatAxis has only a single use. Please directly use `/*concatAxis=*/1`.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.cpp
```diff
@@ -126,28 +126,53 @@ Value getLSTMGRUGetYh(Location loc, PatternRewriter &rewriter, Value val,
...
+    ONNXSplitV11Op splitOp = rewriter.create<ONNXSplitV11Op>(
+        loc, splitTypes, val, splitAxis, nullptr);
+    SmallVector<int64_t> sliceShape({1, 1, B, H});
+    Type sliceType = RankedTensorType::get(sliceShape, elementType);
+    Value fwdLastSlice = rewriter.create<ONNXSliceOp>(
+        loc, sliceType, splitOp.getResults()[0], minusOne, intMax, axis, step);
+    Value bkwFirstSlice = rewriter.create<ONNXSliceOp>(
+        loc, sliceType, splitOp.getResults()[1], zero, one, axis, step);
+    SmallVector<int64_t> concatShape({1, D, B, H});
+    Type concatType = RankedTensorType::get(concatShape, elementType);
```
tungld, line 164: Please directly use `Type concatType = RankedTensorType::get({1, D, B, H}, elementType);` to avoid creating a new vector.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.cpp
```diff
@@ -126,28 +126,53 @@ Value getLSTMGRUGetYh(Location loc, PatternRewriter &rewriter, Value val,
...
+        loc, resYh.getType(), sliceOp.getResult(), rewriter.getI64ArrayAttr(0));
+  } else if (directionStr.equals_insensitive("bidirectional")) {
+    SmallVector<int64_t> splitShape({T, 1, B, H});
+    Type splitType = RankedTensorType::get(splitShape, elementType);
+    SmallVector<Type> splitTypes = {splitType, splitType};
+    int64_t splitAxis = 1;
+    ONNXSplitV11Op splitOp = rewriter.create<ONNXSplitV11Op>(
+        loc, splitTypes, val, splitAxis, nullptr);
+    SmallVector<int64_t> sliceShape({1, 1, B, H});
+    Type sliceType = RankedTensorType::get(sliceShape, elementType);
```
tungld, line 158: Please directly use `Type sliceType = RankedTensorType::get({1, 1, B, H}, elementType);` to avoid creating a new vector.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.cpp
```diff
@@ -126,28 +126,53 @@ Value getLSTMGRUGetYh(Location loc, PatternRewriter &rewriter, Value val,
...
+
+    SmallVector<int64_t> sliceShape({1, D, B, H});
+    Type sliceType = RankedTensorType::get(sliceShape, elementType);
+    ONNXSliceOp sliceOp = rewriter.create<ONNXSliceOp>(
+        loc, sliceType, val, start, end, axis, step);
+    ret = rewriter.create<ONNXSqueezeV11Op>(
+        loc, resYh.getType(), sliceOp.getResult(), rewriter.getI64ArrayAttr(0));
+  } else if (directionStr.equals_insensitive("bidirectional")) {
+    SmallVector<int64_t> splitShape({T, 1, B, H});
+    Type splitType = RankedTensorType::get(splitShape, elementType);
```
tungld, line 152: Please directly use `Type splitType = RankedTensorType::get({T, 1, B, H}, elementType);` to avoid creating a new vector.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.cpp
```diff
@@ -126,28 +126,53 @@ Value getLSTMGRUGetYh(Location loc, PatternRewriter &rewriter, Value val,
...
-      loc, resYh.getType(), sliceOp.getResult(), rewriter.getI64ArrayAttr(0));
-  return squeezeOp.getResult();
+  Value ret;
+  if (directionStr.equals_insensitive("forward") ||
+      directionStr.equals_insensitive("reverse")) {
+    Value start = directionStr.equals_insensitive("forward") ? minusOne : zero;
+    Value end = directionStr.equals_insensitive("forward") ? intMax : one;
+
+    SmallVector<int64_t> sliceShape({1, D, B, H});
+    Type sliceType = RankedTensorType::get(sliceShape, elementType);
```
tungld, line 145: Please directly use `Type sliceType = RankedTensorType::get({1, D, B, H}, elementType);` to avoid creating a new vector.

tungld (changes requested):

src/Accelerators/NNPA/Conversion/ZHighToZLow/ZHighToZLow.cpp
negiyas: 4DS layout is used only for output of LSTM/GRU ops, shape[1] specifies the direction dim and should be 1(forward/backward) or 2(bidirectional), and should not be dynamic. It should be managed in a specific way.
I added the limitation at assert statement to check the condition and add the legality check at src/Accelerators/NNPA/C
onversion/ONNXToZHigh/ONNXLegalityCheck.cpp.

src/Accelerators/NNPA/Conversion/ZHighToZLow/ZHighToZLow.cpp
negiyas: In the FICO layout,  shape[0] specifies the direction dim for LSTM/GRU, should be 1(forward/backward) or 2(bidirectional), and should not be dynamic.
I added the limitation at assert statement to check the condition, and add the legality check at src/Accelerators/NNPA/C
onversion/ONNXToZHigh/ONNXLegalityCheck.cpp.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.td
negiyas: Updated. Thank you.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.cpp
negiyas: done. Thank you.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.cpp
negiyas: done. Thank you.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.cpp
negiyas: done. Thank you.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.cpp
negiyas: done. Thank you.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.cpp
negiyas: done. Thank you.

test/mlir/accelerators/nnpa/conversion/onnx-to-zhigh/lstm.mlir
negiyas: Conversion tests for bidirectional LSTM/GRU with unknown dimensions added. Thank you.

negiyas:
I suppose that all comments at this point are reflected.

src/Accelerators/NNPA/Dialect/ZHigh/ZHighOps.td
```diff
@@ -413,20 +414,22 @@ def ZHighLSTMOp:ZHigh_Op<"LSTM", [NoSideEffect,
                        AnyTypeOf<[ZTensor_FICO, NoneType]>:$hidden_bias,
                        SI64Attr:$hidden_size,
                        DefaultValuedStrAttr<StrAttr, "forward">:$direction,
-                       DefaultValuedAttr<SI64Attr, "-1">:$return_all_steps);
+                       DefaultValuedAttr<SI64Attr, "-1">:$return_all_steps,
+                       DefaultValuedStrAttr<StrAttr, "none">:$prev_layer);
```
tungld, line 418: It seems to me it is too early to introduce `prev_layer` at this high level, and it is early to decide the value of `prev_layer`. Operations sandwiched between two LSTMs may be eliminated by optimizations, and the two LSTMs which is not consecutive will become consecutive then.

tungld (changes requested):

src/Accelerators/NNPA/Dialect/ZHigh/ZHighOps.td
negiyas: I am deleting prev_layer in ZHigh dialect definition, and implementing a new pass setting the prev_layer argument in zlow.LSTM and zlow.GRU ops.

tungld:
@negiyas change to draft. Please let us know when it is ready for review.

negiyas:
> @negiyas change to draft. Please let us know when it is ready for review.
@tungld Thank you!

negiyas:

negiyas:
Closing this PR, since it has been replaced by PR#1471.

PR #1271, opened 2022-03-25
Fix category mapper for case when input value yields default string

etiotto:
When an input tensor contains a int64_t value that is not in the CategoryMapper's category attribute, the default string value needs to be returned. Currently there is a bug in the implementation, this patch fixes the problem and adds and end-to-end test. 

Note: this PR is built on top of https://github.com/onnx/onnx-mlir/pull/1264 and requires that one to be merged.


test/backend-cpp/TestCategoryMapper.cpp
```diff
@@ -71,7 +71,7 @@ int main(int argc, char *argv[]) {
       argc, argv, "TestCategoryMapper\n", nullptr, "TEST_ARGS");
 
   bool rc = testInt64ToStr();
-  rc &= testStrToInt64();
+  //  rc &= testStrToInt64();
```
etiotto, line 74: should not be commented out.

AlexandreEichenberger (commented):
Where is the bug fix? 
I can see how the tests now add an input key that has not in the map, and that a default of "unknown" is returned and tested.

etiotto:
https://github.com/onnx/onnx-mlir/pull/1271#pullrequestreview-925316826

> Where is the bug fix? I can see how the tests now add an input key that has not in the map, and that a default of "unknown" is returned and tested.

Yes, that new test did not work and caused a segfault.

etiotto:
@gongsu832 the Linux s380x build failed because of lack of space, from the log: 
```
Exception: Error processing tar file(exit status 1): write /workdir/llvm-project/build/tools/llvm-libtool-darwin/CMakeFiles/llvm-libtool-darwin.dir/llvm-libtool-darwin.cpp.o: no space left on device
```
Could you please take a look at the build bot?

etiotto:
w.r.t. https://github.com/onnx/onnx-mlir/pull/1271#issuecomment-1082454926. @gongsu832 not sure what happened but the latest build passed on s390. So disregard that request pls.

gongsu832:
Disk space on s390x is low due to many outstanding PRs and the image size has been slowly increasing (dev image used to be roughly 12GB and now 15GB). I added 500GB to the lvm.

AlexandreEichenberger (approved):
LGTM

PR #1272, opened 2022-03-25
Fix instrument and posinf warnings

### LINKED ISSUE #1230: g++ warnings in test/unit/TestInstrumentation.cpp
Using g++ on Linux I get the following warnings:

```
/home/etiotto/Source/onnx/onnx-mlir/test/unit/TestInstrumentation.cpp:18:20: warning: array subscript ‘const int64_t {aka const long int}[0]’ is partly outside array bounds of ‘const char [5]’ [-Warray-bounds]
   18 |   OMInstrumentPoint(*(const int64_t*)op3, 4);
      |   ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~
/home/etiotto/Source/onnx/onnx-mlir/test/unit/TestInstrumentation.cpp:12:14: note: while referencing ‘op3’
   12 |   const char op3[]     = "TOp3";
      |              ^~~
/home/etiotto/Source/onnx/onnx-mlir/test/unit/TestInstrumentation.cpp:19:20: warning: array subscript ‘const int64_t {aka const long int}[0]’ is partly outside array bounds of ‘const char [5]’ [-Warray-bounds]
   19 |   OMInstrumentPoint(*(const int64_t*)op4, 9);
      |   ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~
/home/etiotto/Source/onnx/onnx-mlir/test/unit/TestInstrumentation.cpp:13:14: note: while referencing ‘op4’
   13 |   const char op4[]     = "TOp4";
      |              ^~~
/home/etiotto/Source/onnx/onnx-mlir/test/unit/TestInstrumentation.cpp:21:20: warning: array subscript ‘const int64_t {aka const long int}[0]’ is partly outside array bounds of ‘const char [7]’ [-Warray-bounds]
   21 |   OMInstrumentPoint(*(const int64_t*)opfinal, 12);
      |   ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/etiotto/Source/onnx/onnx-mlir/test/unit/TestInstrumentation.cpp:14:14: note: while referencing ‘opfinal’
   14 |   const char opfinal[] = "TOpFin";
```

### LINKED ISSUE #1245: clang warnings on MathBuilder::positiveInf
The code at https://github.com/onnx/onnx-mlir/blob/698f84d2cc9814c0b39fba37e63f5902d82db3a8/src/Dialect/ONNX/MLIRDialectBuilder.cpp#L273 produces warnings due to the cast from int64_t to double. I'm not sure I fully understand the intent of the code here, but it seems like an issue since casting this constant back to an int64_t will result in a value that is != numeric_limits<int64_t>::max().

NathanielMcVicar:
Fix warnings for gcc and clang. Closes #1245 and closes #1230.

### DIFF
--- a/test/unit/TestInstrumentation.cpp
+++ b/test/unit/TestInstrumentation.cpp
@@ -7,11 +7,11 @@
 #include <thread>
 
 int main(int argc, char *argv[]) {
-  const char opstart[] = "TOpStar";
-  const char op2[]     = "TOp2";
-  const char op3[]     = "TOp3";
-  const char op4[]     = "TOp4";
-  const char opfinal[] = "TOpFin";
+  const char opstart[8] = "TOpStar";
+  const char op2[8]     = "TOp2";
+  const char op3[8]     = "TOp3";
+  const char op4[8]     = "TOp4";
+  const char opfinal[8] = "TOpFin";
   OMInstrumentInit();
   OMInstrumentPoint(*(const int64_t*)opstart, 13);
   OMInstrumentPoint(*(const int64_t*)op2, 1);
--- a/src/Dialect/ONNX/MLIRDialectBuilder.cpp
+++ b/src/Dialect/ONNX/MLIRDialectBuilder.cpp
@@ -270,8 +270,8 @@ Value MathBuilder::positiveInf(Type type) const {
                              : std::numeric_limits<uint32_t>::max();
           break;
         case 64:
-          value = (isSigned) ? std::numeric_limits<int64_t>::max()
-                             : std::numeric_limits<uint64_t>::max();
+          value = static_cast<double>((isSigned) ? std::numeric_limits<int64_t>::max()
+                                                 : std::numeric_limits<uint64_t>::max());
           break;
         default:
           llvm_unreachable("unsupported element type");
--- a/src/Dialect/ONNX/MLIRDialectBuilder.cpp
+++ b/src/Dialect/ONNX/MLIRDialectBuilder.cpp
@@ -270,8 +270,9 @@ Value MathBuilder::positiveInf(Type type) const {
                              : std::numeric_limits<uint32_t>::max();
           break;
         case 64:
-          value = static_cast<double>((isSigned) ? std::numeric_limits<int64_t>::max()
-                                                 : std::numeric_limits<uint64_t>::max());
+          value = static_cast<double>(
+              (isSigned) ? std::numeric_limits<int64_t>::max()
+                         : std::numeric_limits<uint64_t>::max());
           break;
         default:
           llvm_unreachable("unsupported element type");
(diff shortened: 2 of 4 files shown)


NathanielMcVicar:
@AlexandreEichenberger I removed the comma after PERF_MAIN() as the smaller delta, if you would prefer I refactor the macro to resolve the warning please let me know.

sstamenova:
> @AlexandreEichenberger I removed the comma after PERF_MAIN() as the smaller delta, if you would prefer I refactor the macro to resolve the warning please let me know.

In general, fixing the macro would be the better approach

sstamenova:

chentong319 (approved):
LGTM!

test/unit/TestInstrumentation.cpp
```diff
@@ -7,11 +7,11 @@
 #include <thread>
 
 int main(int argc, char *argv[]) {
-  const char opstart[] = "TOpStar";
-  const char op2[]     = "TOp2";
-  const char op3[]     = "TOp3";
-  const char op4[]     = "TOp4";
-  const char opfinal[] = "TOpFin";
+  const char opstart[8] = "TOpStar";
+  const char op2[8]     = "TOp2";
```
tungld, line 11: Should this be 5 instead of 8? The same for the following ones.

tungld:

test/unit/TestInstrumentation.cpp
NathanielMcVicar: It is 8 to make sure the allocation is 64-bits to remove the GCC warning in #1230.

tungld: Which warnings did you get if using `const char op2[5]     = "TOp2";`? It is clear that the string has 5 chars including the null terminator.

NathanielMcVicar: `TestInstrumentation.cpp:19:20: warning: array subscript ‘const int64_t {aka const long int}[0]’ is partly outside array bounds of ‘const char [5]’ [-Warray-bounds]
`
I don't think GCC likes the cast from an array smaller than 8 bytes a pointer to an 8-byte type.

tungld: It looks to me `char[] ` or `char[5]` is not the root cause. It is the problem of casting to int64_t at Line 19. Changing to 8 is just make GCC do not complain about array bounds but the last 3 bytes are undefined.

NathanielMcVicar: But there is no problem with undefined bytes at the end of a C string, GCC just seems to care about that the full width of the pointer type used is within bounds. I wonder if this is related to https://gcc.gnu.org/bugzilla/show_bug.cgi?id=98266 or something similar, but I am in no way a GCC expert and regardless changing the test case to avoid the warning seems pretty reasonable. We can't avoid casting, unless we changed the signature of OMInstrumentPoint.

PR #1276, opened 2022-03-29
Remove duplication in generating InitAccelerators.cpp

tungld:
There is duplication in generating InitAccelerators.cpp in CMakeLists.txt. This patch is to remove the duplicated code.


AlexandreEichenberger (commented):
I find the code-generating cmake quite cryptic. Wouldn't it be easier to define a single macro
e.g. like we did for the OM_TYPE
```
OM_TYPE_METADATA_DEF(ONNX_TYPE_FLOAT,       1,  sizeof(float),    "float")
```
and then using normal C++ code to use the macro in specific circumstances?

caoimhinuibrian:
> 

Alex, I don't understand what you are suggesting here? The CMake stuff isn't C++?

tungld:
Closing this since PR #1281 has the same modification.

PR #1281, opened 2022-03-29
Misc. improvements to Accelerator code.

etiotto:
This PR contains several changes/improvements to the Accelerator support code: 
  - use `llvm::SmallVector` rather than `std::vector` for accelerator list 
  - move `initAccelerators()` in the `onnx_mlir::accel` namespace (to avoid polluting the global namespace)
  - add documentation to the public member function in the `Accelerator` class
  - delete the nnpa pointers when `~NNPAAccelerator` is invoked
  - move createNNPA in the `onnx_mlir::accel` namespace
  - initialize accelerator dialects also for `onn-mlir-reduce`


src/Tools/onnx-mlir-reduce/CMakeLists.txt
```diff
@@ -7,6 +7,8 @@ add_onnx_mlir_executable(onnx-mlir-reduce
 
   LINK_LIBS PRIVATE
   ${OMLibs}
+  Accelerator
```
sstamenova, line 10: Can you make sure this lib is excluded from `OMLibs`? I think its cmake currently adds it to the list

src/Accelerators/Accelerator.hpp
```diff
@@ -17,39 +17,53 @@
...
 
   Accelerator(Kind kind) : kind(kind) {}
-  virtual ~Accelerator();
+  virtual ~Accelerator() = default;
 
   /// Getter for the kind of this accelerator.
   Kind getKind() const { return kind; }
 
-  static std::vector<Accelerator *> getAcceleratorList();
+  /// Returns the set of accelerators available.
```
AlexandreEichenberger, line 44: I find this slightly strange that the class `Accelerator` is the base class for each specialized accelerator (which is great), but also has a separate static list of accelerators. I would simply have a vector of accelerator (global symbol) where all valid accelerators are added. If one has the desire to ensure only the Accelerator subclasses are allowed to add to this class (I don't see why that would really be needed) then I would make a separate class `AcceleratorList` where only the `Accelerator` class (via a friend) can add a new accelerator.

src/Accelerators/CMakeLists.txt
```diff
@@ -1,43 +1,47 @@
 # SPDX-License-Identifier: Apache-2.0
+
 set(ACCEL_LIST "")
-file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/AcceleratorKinds.hpp "// Enumeration of accelerators \n")
+
+# Generate the enumerators used to identify the kind of accelerators.
+set(AcceleratorKinds "${CMAKE_CURRENT_BINARY_DIR}/AcceleratorKinds.inc")
+file(WRITE "${AcceleratorKinds}" "// Enumeration of accelerators.\n")
```
AlexandreEichenberger, line 7: As I also wrote in an other PR, I find this code gen in the CMake hard to read. I suggest to simply generate a file of macros, like we do for `OM_TYPE_METADATA_DEF`. We would build here a single file with all of the needed info in a single macro, and then generate the right code using a #define . I understand it is a matter of preference.

etiotto: @AlexandreEichenberger  not sure I follow. Currently we have a typed enum in the Accelerator class that includes a file generated by this cmake file:
```
class Accelerator {
public:
  /// Kinds of accelerators.
  // clang-format off
  enum class Kind {
  // Note: AcceleratorKinds.inc is generated by cmake.
  #include "src/Accelerators/AcceleratorKinds.inc"
  };
  // clang-format on
```
We had discussions with @caoimhinuibrian and settled on generating the enumerators for each accelerators because we did not want to expose the accelerators details directly in the Accelerator header file. 

The key design point was that we wanted a cmake based solution because we do not know how many accelerators (if any) should be "compiled" in. With that design decision in mind, could you explain your proposed change?

src/Accelerators/Accelerator.hpp
etiotto: The reason this member function is static is because of the way we use it:
```
    onnx_mlir::accel::initAccelerators();
    for (auto *accel : onnx_mlir::accel::Accelerator::getAccelerators()) {
      if (!accel->isActive())
        continue;
      accel->getOrLoadDialects(context);
      accel->addPasses(module, pm, emissionTarget);
    }
```
In that code snippet we first initialize the accelerators availables. During initialization, each specific accelerator, adds itself to the list of accelerators maintained by the `Accelerator` base class. After initialization completes we then query the base class for that list/vector.  Now, because we cannot directly instantiate a Accelerator object (that class is virtual), we cannot make the accelerator list non-static.

src/Tools/onnx-mlir-reduce/CMakeLists.txt
etiotto: Done.

src/Compiler/CMakeLists.txt
```diff
@@ -81,6 +81,7 @@ add_onnx_mlir_library(CompilerUtils
   LINK_LIBS PUBLIC
   ${OMLibs}
   CompilerPasses
+  Accelerator
```
tungld, line 84: Do we need to add this? I see that you make InitAcelerator link to Accelerator.

src/Tools/onnx-mlir-opt/CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ add_onnx_mlir_executable(onnx-mlir-opt
   LINK_LIBS PRIVATE
   ${OMLibs}
   CompilerOptions
+  Accelerator
```
tungld, line 11: Do we need to add this? I see that you make InitAcelerator link to Accelerator.

etiotto: I will revert the change in InitAccelerator and avoid linking the Accelerator library because it already link the NNPAAccel.

sstamenova:
LGTM. I think there's value in reconsidering the approach for generating the files from cmake (as in, it can be probably be improved and possibly even removed), but I would prefer to see a separate PR just for that anyway.

sstamenova (approved):

AlexandreEichenberger:
@etiotto 

one approach to have less code generated in the makefile is as follows. Define a file: AccelKind.inc as follows. That would be the only one auto-generated. For example for a NNPA and a SUPERFAST accelerator

```C++
#define ACCEL_LIST(NNPA)
#define ACCEL_LIST(SUPERFAST)
```

and then your enum can be gen as this
```C++
   enum class Kind {
  #define ACCEL_LIST(X) X ## _accel,
  #include "src/Accelerators/AcceleratorKinds.inc"
    NONE_accel
  };
```

would expand to this
```C++
   enum class Kind {
    NNPA_accel,
   SUPERFAST_accel,
    NONE_accel
  };
```

and where you currently use the options (current cmake `clEnumValN(accel::Accelerator::Kind::${t}`, you can do this

```C++
#define ACCEL_LIST(X) clEnumValN(accel::Accelerator::Kind:: ## X
  #include "src/Accelerators/AcceleratorKinds.inc"
```

The advantage of this scheme is that you define only one pattern in the CMAKE file and then you can refine each usage in the code directly. The drawback is that the user cannot peek at the code like the current approach in the build subdirectories.

src/Accelerators/Accelerator.hpp
AlexandreEichenberger: I know, I find just strange that the Accel is both a virtual class to be specialized by each custom accelerators as well as the holder of a list of accelerators. This functionality could be split in a separate accel list variable (or class, depending how fancy you want to be).

src/Compiler/CMakeLists.txt
etiotto: Reverted the change in InitAccelerator and avoid linking the Accelerator library because it already link the NNPAAccel.

AlexandreEichenberger (approved):
LGTM at this stage, and we can retune the approach of the part where we gen code via the cmakefile if necessary in a follow on PR

PR #1285, opened 2022-03-30
[LRN] Use builder based interface to generate Krnl loops

whitneywhtsang:
This PR updates code generation for `LRN.cpp` to use the builder based interfaces for generation of Krnl loops.


src/Conversion/ONNXToKrnl/Math/LRN.cpp
```diff
@@ -55,87 +55,89 @@ struct ONNXLRNOpLowering : public ConversionPattern {
...
+              cIE + 1 + (sizeIE - 1).ceilDiv(LiteralIndexExpr(2)));
+
+          // Initialize sum, single scalar, no need for default alignment.
+          MemRefType scalarMemRefType = MemRefType::get({}, elementType, {}, 0);
+          MultiDialectBuilder<KrnlBuilder, MemRefBuilder, MathBuilder> create(
+              rewriter, loc);
+
+          Value sumAlloc = create.mem.alloc(scalarMemRefType);
+          createKrnl.store(
+              emitConstantOp(rewriter, loc, elementType, 0), sumAlloc);
```
tungld, line 96: `emitConstantOp` can be replaced by `create.math.constant`

src/Conversion/ONNXToKrnl/Math/LRN.cpp
```diff
@@ -55,87 +55,89 @@ struct ONNXLRNOpLowering : public ConversionPattern {
...
+
+          // Create the sum reduction loop
+          auto outputLoopBody = rewriter.saveInsertionPoint();
```
tungld, line 99: Do we still need to save and restore an insertion point? It looks redundant now.

tungld (changes requested):

src/Conversion/ONNXToKrnl/Math/LRN.cpp
```diff
@@ -55,87 +55,86 @@ struct ONNXLRNOpLowering : public ConversionPattern {
...
+          // Insert computation of square_sum.
+          // square_sum[n, c, d1, ..., dk] = sum(X[n, i, d1, ..., dk] ^ 2),
+          // where max(0, c - floor((size - 1) / 2)) <= i
+          // and i<= min(C - 1, c + ceil((size - 1) / 2)).
+
+          // Get a child IndexExpr context.
+          IndexExprScope childScope(&rewriter, shapeHelper.scope);
+
+          // Compute the lower bound and upper bound for square_sum.
+          const int loopIndexForC = 1;
```
etiotto, line 73: [Suggestion]: The RHS is constant so we could use C++11  `constexpr` here.

src/Conversion/ONNXToKrnl/Math/LRN.cpp
```diff
@@ -55,87 +55,86 @@ struct ONNXLRNOpLowering : public ConversionPattern {
...
-                            create.math.mul(alphaDivSizeValue, sumValue)),
-            betaValue);
-    Value resultValue = create.math.div(xValue, tempValue);
-
-    create.krnl.store(resultValue, alloc, storeIndices);
+    KrnlBuilder createKrnl(rewriter, loc);
+    ValueRange outputLoopDef = createKrnl.defineLoops(outputRank);
+    SmallVector<IndexExpr, 4> lbs(outputRank, LiteralIndexExpr(0));
+    createKrnl.iterateIE(outputLoopDef, outputLoopDef, lbs,
+        shapeHelper.dimsForOutput(0),
```
etiotto, line 62: [Suggestion]:  Use ```shapeHelper.dimsForOutput()```,  zero is the default: 
```
DimsExpr &dimsForOutput(int n = 0) { return outputsDims[n]; }
```

src/Conversion/ONNXToKrnl/Math/LRN.cpp
```diff
@@ -55,87 +55,86 @@ struct ONNXLRNOpLowering : public ConversionPattern {
...
+
+          // Get a child IndexExpr context.
+          IndexExprScope childScope(&rewriter, shapeHelper.scope);
+
+          // Compute the lower bound and upper bound for square_sum.
+          const int loopIndexForC = 1;
+          DimIndexExpr cIE(outputLoopInd[loopIndexForC]);
+          MemRefBoundsIndexCapture inputBounds(input);
+          DimIndexExpr CIE(inputBounds.getDim(loopIndexForC));
+          SymbolIndexExpr sizeIE = LiteralIndexExpr(sizeLit);
```
etiotto, line 77: I'm wondering whether we could avoid the copy and write:
```
SymbolIndexExpr sizeIE(sizeLit);

File: src/Conversion/ONNXToKrnl/Math/LRN.cpp [hunk 6]
```diff
@@ -55,87 +55,86 @@ struct ONNXLRNOpLowering : public ConversionPattern {
...
+          lbMaxList.emplace_back(
+              cIE - (sizeIE - 1).floorDiv(LiteralIndexExpr(2)));
+
+          SmallVector<IndexExpr, 2> ubMinList;
+          ubMinList.emplace_back(CIE);
+          ubMinList.emplace_back(
+              cIE + 1 + (sizeIE - 1).ceilDiv(LiteralIndexExpr(2)));
+
+          // Initialize sum, single scalar, no need for default alignment.
+          MemRefType scalarMemRefType = MemRefType::get({}, elementType, {}, 0);
```
etiotto, line 90: What is the zero in `MemRefType::get({}, elementType, {}, 0);` ? 
Suggest to use: 
```
MemRefType scalarMemRefType = MemRefType::get({}, elementType);
```
which should be equivalent.

src/Conversion/ONNXToKrnl/Math/LRN.cpp
```diff
@@ -55,87 +55,86 @@ struct ONNXLRNOpLowering : public ConversionPattern {
...
+
+                sumValue = createKrnl.load(sumAlloc, ArrayRef<Value>{});
+                sumValue = create.math.add(sumValue, squareVal);
+                createKrnl.store(sumValue, sumAlloc, ArrayRef<Value>{});
+              });
+
+          // Compute and store the output
+          // y = x / ((bias + (alpha / nsize) * square_sum) ** beta)
+          SmallVector<Value, 4> storeIndices;
+          for (int i = 0; i < outputRank; ++i) {
```
etiotto, line 125: [nit] no {} for single block

src/Conversion/ONNXToKrnl/Math/LRN.cpp
```diff
@@ -55,87 +55,86 @@ struct ONNXLRNOpLowering : public ConversionPattern {
...
+          MemRefType scalarMemRefType = MemRefType::get({}, elementType, {}, 0);
+          MultiDialectBuilder<KrnlBuilder, MemRefBuilder, MathBuilder> create(
+              rewriter, loc);
+
+          Value sumAlloc = create.mem.alloc(scalarMemRefType);
+          createKrnl.store(create.math.constant(elementType, 0), sumAlloc);
+
+          // Create the sum reduction loop
+          ValueRange sumLoopDef = createKrnl.defineLoops(1);
+          Value sumValue;
```
etiotto, line 99: I think you `sumValue` should be declared in the loop at line 117 given that every loop iteration initializes it and is not live on exit.

whitneywhtsang: It is live out, it is used at line 132.

etiotto: Line 129 redefines it before it is used at line 132:
```
129  sumValue = createKrnl.load(sumAlloc);
```

src/Conversion/ONNXToKrnl/Math/LRN.cpp
whitneywhtsang:
```
LiteralIndexExpr(int64_t const value);
```
`LiteralIndexExpr` has a constructor for `int`, but `SymbolIndexExpr` doesn't.

src/Conversion/ONNXToKrnl/Math/LRN.cpp
whitneywhtsang: `MemRefType::get({}, elementType)` invokes:
```
static MemRefType get(ArrayRef<int64_t> shape, Type elementType, MemRefLayoutAttrInterface layout = {}, Attribute memorySpace = {});

MemRefType MemRefType::get(ArrayRef<int64_t> shape, Type elementType,                                                                      
                           MemRefLayoutAttrInterface layout,                                                                               
                           Attribute memorySpace) {                                                                                        
  // Use default layout for empty attribute.
  if (!layout)
    layout = AffineMapAttr::get(AffineMap::getMultiDimIdentityMap(
        shape.size(), elementType.getContext()));                                                                                          
  
  // Drop default memory space value and replace it with empty attribute.                                                                  
  memorySpace = skipDefaultMemorySpace(memorySpace);                                                                                       
                                                                                                                                           
  return Base::get(elementType.getContext(), shape, elementType, layout,                                                                   
                   memorySpace);
} 
```
`MemRefType::get({}, elementType, {}, 0)` invokes:
```
static MemRefType get(ArrayRef<int64_t> shape, Type elementType, AffineMap map, unsigned memorySpaceInd);

MemRefType MemRefType::get(ArrayRef<int64_t> shape, Type elementType,
                           AffineMap map, unsigned memorySpaceInd) {

  // Use default layout for empty map.
  if (!map)
    map = AffineMap::getMultiDimIdentityMap(shape.size(),
                                            elementType.getContext());

  // Wrap AffineMap into Attribute.
  Attribute layout = AffineMapAttr::get(map);

  // Convert deprecated integer-like memory space to Attribute.
  Attribute memorySpace =
      wrapIntegerMemorySpace(memorySpaceInd, elementType.getContext());

  return Base::get(elementType.getContext(), shape, elementType, layout,
                   memorySpace);
}
```

whitneywhtsang:
Need to think about how to translate `BuildKrnlLoop::pushBounds(SmallVectorImpl<IndexExpr> &lowerBound, SmallVectorImpl<IndexExpr> &upperBound)` to builder based interface.

src/Conversion/ONNXToKrnl/Math/LRN.cpp
etiotto: ok the implementation differs so you cannot simplify the code. Leave it as is.

etiotto (approved):
LGTM

whitneywhtsang:
@tungld Please take another look when you have time, thanks!

src/Conversion/ONNXToKrnl/Math/LRN.cpp
```diff
@@ -100,23 +99,27 @@ struct ONNXLRNOpLowering : public ConversionPattern {
...
-          krnl::BuildKrnlLoop sumLoops(rewriter, loc, 1);
-          sumLoops.createDefineOp();
-          sumLoops.pushBounds(lbMaxList, ubMinList);
-          sumLoops.createIterateOp();
+          std::vector<Value> loop;
+          defineLoops(rewriter, loc, loop, 1);
+          krnl::KrnlIterateOperandPack pack(rewriter, loop);
+          pack.pushIndexExprsBound(lbMaxList);
+          pack.pushIndexExprsBound(ubMinList);
+          KrnlIterateOp iterateOp = createKrnl.iterate(pack);
```
tungld, line 110: @whitneywhtsang is there any reason why we didn't use lamda function to create `krnl.iterate`?

whitneywhtsang: Do you mean why we don't use 
```
KrnlBuilder::iterate(ValueRange originalLoops, ValueRange optimizedLoops,
    ValueRange lbs, ValueRange ubs,
    function_ref<void(KrnlBuilder &createKrnl, ValueRange indices)>
        bodyBuilderFn);
```
?
If so, it is because I cannot find an interface that takes `KrnlIterateOperandPack`.

tungld: Yes, I mean that one. In this case, upper and lower bounds are IndexExpr, so we should use `KrnlBuilder::iterateIE`.

> If so, it is because I cannot find an interface that takes KrnlIterateOperandPack.

`KrnlIterateOperandPack` will be created by the KrnlIterateOp's build by using upper and lower bounds passed to `KrnlBuilder::iterateIE`: https://github.com/onnx/onnx-mlir/blob/main/src/Dialect/Krnl/KrnlOps.cpp#L244

whitneywhtsang: `lbMaxList` and `ubMinList` are both `SmallVector<IndexExpr>`, not `IndexExpr`, so cannot use either `KrnlBuilder::iterate` or `KrnlBuilder::iterateIE`.

tungld: Got it. Thanks!

tungld (approved):
LGTM.

PR #1288, opened 2022-03-30
Optionally use GEMM for OnnxBuilder::matmul

tungld:
Add an option to `OnnxBuilder::matmul` to switch between ONNXGEMM or ONNXMatMul. This will be used later by LSTM to see which one leads to better performance.


src/Dialect/ONNX/ONNXOpsHelper.cpp
```diff
@@ -39,7 +39,23 @@ Value OnnxBuilder::div(Value A, Value B) const {
...
+  // Gemm only supports rank 2.
+  bool canUseGemm = useGemm && A.getType().isa<ShapedType>() &&
+                    A.getType().cast<ShapedType>().hasRank() &&
+                    (A.getType().cast<ShapedType>().getRank() == 2) &&
+                    B.getType().isa<ShapedType>() &&
+                    B.getType().cast<ShapedType>().hasRank() &&
+                    (B.getType().cast<ShapedType>().getRank() == 2);
+  if (canUseGemm)
+    return b.create<ONNXGemmOp>(loc, Y, A, B, b.createOrFold<ONNXNoneOp>(loc),
+        /*alpha=*/b.getF32FloatAttr(1.0), /*beta=*/b.getF32FloatAttr(1.0),
```
etiotto, line 52: Do we need alpha ? I read in the ONNX operator manual it defaults to 1.0, same for beta. 
Also the ONNX operator description also states that transA and transB default to zero (https://github.com/onnx/onnx/blob/main/docs/Operators.md#Gemm).

P.S.: @tungld  I did not check whether we have a generated build method that takes default arguments as per ONNX spec. Would be handy to have one perhaps.

etiotto (approved):
LGTM with one question.

AlexandreEichenberger:
Its fine, my goal would be to unify code gen for 2d matmul and gemm so that I may generate the best code depending on the data size. I am not 100% sure how to do that yet.

src/Dialect/ONNX/ONNXOpsHelper.cpp
tungld: @etiotto we don't have a custom build method with default arguments yet. So, I need to pass them here. I think we can generate build methods with default arguments for every ONNX ops automatically by modifying `gen_onnx_mlir.py`. Need another PR to do that systematically.

etiotto: yup ok with doing that in a separate PR.

PR #1290, opened 2022-03-30
Add messages to assertions that check shape computed successfully

whitneywhtsang:


etiotto (approved):
LGTM - thanks for taking the time to add the assert msgs to all those operators!

Issue #1280, opened 2022-03-29
MemRefBoundsIndexCapture error when converting custom fraud model


cjvolzka:
We're getting an error trying to compile the following model [jan_improved.onnx.zip](https://github.com/onnx/onnx-mlir/files/8372936/jan_improved.onnx.zip) using onnx-mlir commit 35352d126b2af14e40858e7ddfb984a7e6cc53aa. We get the same error if we try and compile for CPU or NNPA, however NNPA is the desired goal.

At a glance, I didn't notice any unsupported Ops. The model was converted using tf2onnx 1.9.3. When I look at it using Netron I don't see anything strange (ie no dynamic loop ops).

```
onnx-mlir --O3 --EmitLib --mtriple=s390x-ibm-loz --mcpu=z14 jan_improved.onnx  -o CPU/jan_improved.onnx-mlir-preview-llvm -v
dlc: /DLCpp/third_party/onnx-mlir/src/Dialect/ONNX/IndexExpr.cpp:1646: mlir::IndexExpr mlir::MemRefBoundsIndexCapture::get(uint64_t) [with INDEX = mlir::DimIndexExpr; uint64_t = long unsigned int]: Assertion `i < memRank && "index out of bound"' failed.
```

etiotto:
The compiler assert during shape inference on the `onnx.Concat` operator that feeds the first LSTM operator. The issue is the value of the `axis` attribute for that Concat operator, the value of axis is `33554432`:

```
% onnx-mlir jan_improved.onnx --EmitONNXBasic ; grep "onnx.Concat" jan_improved.onnx.mlir 

%76 = "onnx.Concat"(%66, %75, %22, %24, %39, %41, %15, %17, %31, %33, %6, %57, %55, %53, %51, %49, %47) {axis = 33554432 : si64, onnx_node_name = "StatefulPartitionedCall/model_4/tf.concat_1/concat"} : (tensor<*xf32>, tensor<*xf32>, tensor<*xf32>, tensor<*xf32>, tensor<*xf32>, tensor<*xf32>, tensor<*xf32>, tensor<*xf32>, tensor<*xf32>, tensor<*xf32>, tensor<*xf32>, tensor<*xf32>, tensor<*xf32>, tensor<*xf32>, tensor<*xf32>, tensor<*xf32>, tensor<*xf32>) -> tensor<*xf32>
...
```    

[...]

PR #1291, opened 2022-03-30
Diagnose Concat, Compress, ArgMax, ArgMin axis attribute out-of-bouds

etiotto:
This PR is addresses the lack of an informative diagnostic message exposed by  issue https://github.com/onnx/onnx-mlir/issues/1280. Although the input model in that issue is invalid, we should improve diagnostics for cases when the `axis` attribute of certain operators (e.g. Concat, Compres, ArgMax, ArgMin) exceeds the valid range. 

This PR introduces a new class named `onnx_mlir::Diagnostic`, which can be used to emit the diagnostic for attribute that are out of bounds. This class allows consistent emission of the same message for several operators and from either the verification code of during shape inference. 


docs/Dialects/krnl.md
```diff
@@ -953,6 +953,51 @@ Traits: MemRefsNormalizable
 | `scale` | floating-point
 | `seed` | floating-point
 
+### `krnl.seqextract` (::mlir::KrnlSeqExtractOp)
```
AlexandreEichenberger, line 956: thanks for regen the doc

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -590,6 +590,111 @@ static void insertConvTransposeSpatialDim(SmallVectorImpl<int64_t> &outputDims,
...
+
+    return emitWarning(op.getLoc(),
+        msg.concat(": " + attrName)
+            .concat(" attribute exceeds the limit of ")
+            .concat(limitStr)
+            .concat(". Please check whether the model was translated to ONNX "
+                    "correctly."));
+  }
+
+  return InFlightDiagnostic();
```
AlexandreEichenberger, line 616: Can you just add a comment on what is returned. Maybe as simple as adding a `\\ success` comment.
In prior places in this code, we return `success()`

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -590,6 +590,111 @@ static void insertConvTransposeSpatialDim(SmallVectorImpl<int64_t> &outputDims,
...
+    Operation &op, const Twine &attrName, int64_t attrVal) {
+  // Global limit for all attributes.
+  constexpr int64_t attrValLimit = 100;
+
+  if (abs(attrVal) > attrValLimit) {
+    Twine msg(op.getName().getStringRef());
+    std::string limitStr =
+        std::to_string((attrVal > 0) ? attrValLimit : -attrValLimit);
+
+    return emitWarning(op.getLoc(),
```
AlexandreEichenberger, line 608: Does the warning return a failure? Maybe a quick comment may help here clarify this.

AlexandreEichenberger (approved):
LGTM, please clarify the code in a few places with comments

gongsu832:
I meant to say this but kept forgetting. I think we have somewhat inconsistent usage of assertions. Assertions are typically used for bugs only, i.e., which are states we produced internally that should never happen. Invalid user inputs are NOT bugs since users can input whatever they want and we need to handle them. So assertions should never be used on user inputs and other "legitimate" failures. For example, we never do `ptr=malloc(...); assert(ptr);` since malloc failure is perfectly legitimate.

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -590,6 +590,111 @@ static void insertConvTransposeSpatialDim(SmallVectorImpl<int64_t> &outputDims,
   }
 }
 
+//===----------------------------------------------------------------------===//
+// Support function to emit a warning if an operator attibute value is greater
+// than a preset limit.
+//===----------------------------------------------------------------------===//
+
+static InFlightDiagnostic verifyAttributeLimit(
```
tungld, line 598: We should be specific here about what kind of attribute. Otherwise, it looks like the limit is for all attributes, which is not true, for example, `value_int` attribute in constant can be any integer constant. Perhaps, change the function name to a more specific one, e.g. verifyAxisAttributeLitmit.

tungld (changes requested):

etiotto:
> I meant to say this but kept forgetting. I think we have somewhat inconsistent usage of assertions. Assertions are typically used for bugs only, i.e., which are states we produced internally that should never happen. Invalid user inputs are NOT bugs since users can input whatever they want and we need to handle them. So assertions should never be used on user inputs and other "legitimate" failures. For example, we never do `ptr=malloc(...); assert(ptr);` since malloc failure is perfectly legitimate.

Yes, agree. In this case the incorrect input was not diagnosed with a proper error/warning in the operator verification code, which lead to an assert being triggered later on.  Undoubtedly we have other situation similar to this one around. I plan to address this particular problem in all operators (where applicable) in subsequent PRs. This one addresses 4 of them.

etiotto:
> We should be specific here about what kind of attribute. Otherwise, it looks like the limit is for all attributes, which is not true, for example, value_int attribute in constant can be any integer constant. Perhaps, change the function name to a more specific one, e.g. verifyAxisAttributeLitmit.

Fair point. Initially I was thinking we should have a generic function to address limits on operator attributes. We could have the caller pass the particular limit, and the function already accepts the name of the arribute and the operator. I think that would make the function more generally usable. I'll do that shortly.

etiotto:
> > We should be specific here about what kind of attribute. Otherwise, it looks like the limit is for all attributes, which is not true, for example, value_int attribute in constant can be any integer constant. Perhaps, change the function name to a more specific one, e.g. verifyAxisAttributeLitmit.
> 
> Fair point. Initially I was thinking we should have a generic function to address limits on operator attributes. We could have the caller pass the particular limit, and the function already accepts the name of the attribute and the operator. I think that would make the function more generally usable. I'll do that shortly.

I added a Diagnostic class that has a templated function accepting arithmetic types for the type of the attribute and its limit. 
The class has also the global limit for the axis attribute. The idea is to generalize the implementation and facilitate diagnostic for several attributes that might have a preset limit. 
@tung let me know if this is acceptable.

tungld:
> I added a Diagnostic class that has a templated function accepting arithmetic types for the type of the attribute and its limit.
The class has also the global limit for the axis attribute. The idea is to generalize the implementation and facilitate diagnostic for several attributes that might have a preset limit.
@tung let me know if this is acceptable.

@etiotto thanks. It's a good idea.

One concern is the class name `Diagnostic`. MLIR also has `Diagnostic` class (https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/IR/Diagnostics.h), so it is a bit confusing.

I have one question: whether the use of Diagnostic here is aligned with MLIR Diagnostic. For example, according to https://mlir.llvm.org/docs/Diagnostics/, we can use `-mlir-print-op-on-diagnostic` for debugging. Please add lit tests to show how it work.

etiotto:
> I have one question: whether the use of Diagnostic here is aligned with MLIR Diagnostic. For example, according to https://mlir.llvm.org/docs/Diagnostics/, we can use -mlir-print-op-on-diagnostic for debugging. Please add lit tests to show how it work.

I will upload a new commit shortly, with that commit both `onnx-mlir` and `onnx-mlir-opt` will accept the MLIR `-mlir-print-op-on-diagnostic` options and print the operator at fault:
```
% onnx-mlir diag.mlir -mlir-print-op-on-diagnostic
diag.mlir:3:10: warning: onnx.Concat: axis attribute exceeds the limit of 100. Please check whether the model was translated to ONNX correctly.
    %0 = "onnx.Concat"(%arg0, %arg1, %arg2) { axis = 101 : si64} : (tensor<1x1xf32>, tensor<1x1xf32>, tensor<1x1xf32>)  -> tensor<*xf32>
         ^
Error can't load file diag.mlir

% onnx-mlir-opt diag.mlir -mlir-print-op-on-diagnostic
diag.mlir:3:10: warning: onnx.Concat: axis attribute exceeds the limit of 100. Please check whether the model was translated to ONNX correctly.
    %0 = "onnx.Concat"(%arg0, %arg1, %arg2) { axis = 101 : si64} : (tensor<1x1xf32>, tensor<1x1xf32>, tensor<1x1xf32>)  -> tensor<*xf32>
```    
    
 > One concern is the class name Diagnostic. MLIR also has Diagnostic class (https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/IR/Diagnostics.h), so it is a bit confusing.
 
OK we can name it more `OnnxMlirDiagnostic` to avoid confusion.

etiotto:
@tungld I ended up moving the Diagnostic class into a new file so that we can use it during ONNX operators verification as well as well as during shape inference.

etiotto:
> It seems the new code does nothing to handle unknown rank as the comment of this patch said, so the comment should be updated.

Sure, I will make certain the PR description (and also quite important) the commit message used to merge the PR is reflective of how this patch has evolved over time.  More to the substance of this patch, the original issue https://github.com/onnx/onnx-mlir/issues/1280 motivating this PR is **not** correctly diagnosed by the current implementation while with the current PR it is:

```
% onnx-mlir jan_improved.onnx -mlir-print-op-on-diagnostic
error: onnx.Concat 'axis' value is 33554432, accepted range is [-3, 2]
...
```

With this PR the `axis` attribute is correctly diagnosed as out of range, moreover the accepted range is reported. Although I have designed the Diagnostic class to allow us to report values of attributes that look out of range (during verification), yesterday I found out that delaying the message allows the compiler to report the correct range which is more informative.
Should we find that is not possible, we can certainly use the implementation (with little or no modifications) to warn about potential large values of attributes.

> The new code will replace op->emitError by onnx_mlir::Diagnostic::attributeOutOfRange, both returning InFlightDiasnostic. I would like to know what is the compelling reason why emitError is not enough.

Perhaps you have a stale copy of the code (in the browser). The `attributeOutOfRange` member function in this patch function now uses `emitError`.

Let me expand a bit on the rationale for this changes. The new `diagnostic` class provides a centralized way to emit error messages. Currently I have used it to diagnose invalid `axis` values (for several operators)  both during verification and shape inference code using the **same** message.  This is important, because diagnostic messages are currently emitted ad-hoc all over the code base, and so it is easy to end up with diagnostic messages that use inconsistent wordings. 

I believe that centralizing messages in the `Diagnostic` class provides us with much more control on the consistency of the messages and allows for better reuse of the same message.  Also, should we wish to provide language translation of messages, having them in a centralized place will facilitate the translation.

So, see this patch as a way to emit messages for ONNX operators that are reusable, and lead to better consistency.

tungld:
> So, see this patch as a way to emit messages for ONNX operators that are reusable, and lead to better consistency.

Totally agree. Thanks for the explanation!

> I found out that delaying the message allows the compiler to report the correct range which is more informative.

Yeah, this is interesting and it's nontrivial to find an optimal solution.

tungld (approved):
LGTM.

AlexandreEichenberger:
@etiotto If we now use a different approach to report errors, please update the documentation, here
https://github.com/onnx/onnx-mlir/blob/main/docs/ErrorHandling.md

PR #1294, opened 2022-03-30
Upstream Groq-side changes to official onnx-mlir repo

bxing-groq:


bzxing:
@AlexandreEichenberger thanks for noticing - yes you're right. We have staged a bunch of changes but not all are applicable to the general population. I decided to put it back discretionarily. Step 1 would be:
https://github.com/onnx/onnx-mlir/pull/1296#issuecomment-1083653425

Could you advise how I could proceed?

PR #1295, opened 2022-03-30
Accelerator cmake improvements

etiotto:
This PR addresses a comment/suggestion made during code review for https://github.com/onnx/onnx-mlir/pull/1281. The cmake based code generation solution for accelerator code is simplified.


sstamenova:
Have you considered taking this a step further and using it for the `init` as well? (Note: I haven't tested this, the macros probably need some tweaking). Something like:

In the generated file (`Accelerators.inc`):
```
#define ACCELERATORS(create_accelerator) \
  create_accelerator(accelerator1)       \
  create_accelerator(accelerator2)       \
etc....
```

In `Accelerators.hpp`:
```
#include "Accelerators.inc"

#define CREATE_ENUM(name) \
  name,

#define DEFINE_INIT_FUNCTION(name) \
  extern void create##name();

#define INVOKE_INIT_FUNCTION(name) \
  create##name();

#define CREATE_CL_ENUM(name) \
  clEnumValN(accel::Accelerator::Kind::##name, #name, #name "accelerator"),

enum class Kind {
    ACCELERATORS(CREATE_ENUM)
    NONE
  };
```

and in `InitAccelerators.cpp`

```
#include "Accelerators.h"

ACCELERATORS(DEFINE_INIT_FUNCTION)

bool initAccelerators() {
#ifdef HAS_ACCELERATORS
  ACCELERATORS(INVOKE_INIT_FUNCTION)
  return true;
#else
  return false;
#endif
}
```

and in `CompilerOptions.cpp`

```
// clang-format off
  ACCELERATORS(CREATE_CL_ENUM)
  clEnumValN(accel::Accelerator::Kind::NONE, "NONE", "No accelerator")),
// clang format on
```

etiotto:
@sstamenova thanks for the suggestion, I think is quite nice. Working on it.

etiotto:
@sstamenova take another look when you have a chance pls.

src/Accelerators/InitAccelerators.cpp
```diff
@@ -0,0 +1,32 @@
...
+//===----------------------------------------------------------------------===//
+
+#include "src/Accelerators/Accelerator.hpp"
+
+namespace onnx_mlir {
+namespace accel {
+
+//
+bool initAccelerators() {
+  //#ifdef HAS_ACCELERATORS
```
sstamenova, line 22: You need to comment these out, right?

src/Accelerators/InitAccelerators.cpp
```diff
@@ -0,0 +1,32 @@
...
+  APPLY_TO_ACCELERATORS(DECLARE_ACCEL_INIT_FUNCTION)
```
sstamenova, line 23: I think the declarations don't have to be inside the function

src/Accelerators/InitAccelerators.cpp
```diff
@@ -0,0 +1,32 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+#include "src/Accelerators/Accelerator.hpp"
+
+namespace onnx_mlir {
+namespace accel {
+
+//
+bool initAccelerators() {
```
sstamenova, line 21: Is there a reason this is not capitalized? I know it was added in a previous change but it seems weird since everything else is capitalized

src/Accelerators/CMakeLists.txt
```diff
@@ -1,47 +1,57 @@
...
-  file(APPEND "${InitAccelerators}" "  return true;\n")
 else (ACCELERATORS_TO_BUILD)
-  file(APPEND "${InitAccelerators}" "  return false;\n")
+  file(WRITE "${Accelerators}" "#define APPLY_TO_ACCELERATORS(MACRO) \\\n")
 endif (ACCELERATORS_TO_BUILD)
-file(APPEND "${InitAccelerators}" "}\n"
-                                  "} // namespace accel\n"
-                                  "} // namespace onnx_mlir\n")
+
+# Generate the enumerators used to identify the kind of accelerators.
```
sstamenova, line 25: There's quite a bit that needs to be removed here

src/Accelerators/InitAccelerators.cpp
etiotto: ah yes forgot

src/Accelerators/InitAccelerators.cpp
etiotto: ok will move it

src/Accelerators/InitAccelerators.cpp
```diff
@@ -0,0 +1,34 @@
...
+// Initialization of accelerators.
+//
+//===----------------------------------------------------------------------===//
+
+#include "src/Accelerators/Accelerator.hpp"
+
+namespace onnx_mlir {
+namespace accel {
+
+#ifdef HAS_ACCELERATORS
```
sstamenova, line 20: You don't need the ifdef here because if it's not defined, the macro below will expand to nothing anyway. Because the init function needs to return a value depending on the existence of accelerators, it needs the ifdef. I haven't looked at the code, but maybe init can return void since if there are no accelerators, init would succeed by default? In that case you could drop the ifdef there as well.

sstamenova: Ok, I looked at the usage and sometimes the output from `initAccelerators` is used to determine if there are any and sometimes it's not. I would recommend changing `initAccelerators` to return void, removing the need for `HAS_ACCELERATORS` and then modifying the code that was using `initAccelerators` to check for accelerators, to just do it the same way as other places and use the output of `getAccelerators`.

etiotto: Done.

src/Accelerators/CMakeLists.txt
```diff
@@ -1,47 +1,57 @@
...
 
-# Generate the function used to initialize the accelerators.
-set(InitAccelerators "${CMAKE_CURRENT_BINARY_DIR}/InitAccelerators.cpp")
-file(WRITE "${InitAccelerators}" "namespace onnx_mlir {\n"
-                                 "namespace accel {\n"
-                                 "bool initAccelerators() {\n")
-if (ACCELERATORS_TO_BUILD)
+# Generate the Accelerators.inc file.
+set(Accelerators "${CMAKE_CURRENT_BINARY_DIR}/Accelerators.inc")
+if (ACCELERATORS_TO_BUILD)
```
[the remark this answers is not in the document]
etiotto: Done

src/Accelerators/CMakeLists.txt
```diff
@@ -1,55 +1,50 @@
...
-file(WRITE "${InitAccelerators}" "namespace onnx_mlir {\n"
-                                 "namespace accel {\n"
-                                 "bool initAccelerators() {\n")
-if (ACCELERATORS_TO_BUILD)
+if (ACCELERATORS_TO_BUILD) 
   foreach(t ${ACCELERATORS_TO_BUILD})
-    file(APPEND "${InitAccelerators}" "  extern void create${t}();\n")
-    file(APPEND "${InitAccelerators}" "  create${t}();\n")
+    list(APPEND accelerators_inc_generate_commands
+      COMMAND echo "  MACRO\\(${t}\\) \\ " >> ${accelerators_inc})
```
sstamenova, line 21: I am not sure if echo is going to work correctly cross-platform. Is there a reason not to duplicate what we've done for the versions?

sstamenova: Also, the echo commands won't be able to detect if there are changes to the file, I think.

etiotto: Sorry didn't realized I use cmake to perform the echo command:

```
list(APPEND accelerators_inc_generate_commands
  COMMAND ${CMAKE_COMMAND} -E echo "\\#define APPLY_TO_ACCELERATORS\\(MACRO\\) \\ " > ${accelerators_inc})
  ```
  
I'll change the PR to see if that works.

Comment (etiotto) [reply src/Accelerators/CMakeLists.txt]: The Win build fails, @sstamenova could you please try it (don't have access to a Win host at the moment) and let me know what is wrong with the cmake -E echo command(s) ?

**Review by AlexandreEichenberger — APPROVED** [review]
LGTM, I would request that you add a few comments in the modified cmakefile that indicates what is being done for each section, like "# generate MACRO_NAME(accelName)" or "#add list of subdirectories for each activated accelerator". It may also be timely to add a comment in that file on how to add an accelerator to the list of active accelerators... sorry if its already done, I just looked at the diff presented here.

File: src/Accelerators/CMakeLists.txt (hunk 7 above)
Comment (sstamenova) [reply src/Accelerators/CMakeLists.txt]: Both `echo` and `cmake -E echo` are not great cross-platform. The `file` interface is much more stable and it allows you to detect whether the file has been changed.  You can do something like:

```
add_custom_command(OUTPUT "${accelerators_inc}"
  ${CMAKE_COMMAND} -P "${CMAKE_CURRENT_SOURCE_DIR}/GenerateAcceleratorsInc.cmake"
  )
```

This will allow you to invoke the `GenerateAcceleratorsInc.cmake` as part of a separate command that will run as one unit and then inside invoke various `file` commands. This is a good idea if you think it is likely to need to build with different generators during the build and want the file generation to happen at build time rather than configuration time.

If you think it is sufficient to generate the file at configuration time (which is often the case), then you can have the same commands you'd have in the `GenerateAcceleratorsInc.cmake` directly in the `CMakeLists.txt` file. Something along the lines of:

```
# Generate the Accelerators.inc file.
set(accelerators_inc "${CMAKE_CURRENT_BINARY_DIR}/Accelerators.inc")

file(APPEND ${accelerators_inc}.tmp "#define APPLY_TO_ACCELERATORS(MACRO) \\\\\n" )

if (ACCELERATORS_TO_BUILD) 
  foreach(t ${ACCELERATORS_TO_BUILD})
    file(APPEND ${accelerators_inc}.tmp "  MACRO(${t}) \\\\\n" )
  endforeach(t)
endif (ACCELERATORS_TO_BUILD)

file(APPEND ${accelerators_inc}.tmp "\n" )

file(COPY_FILE ${accelerators_inc}.tmp ${accelerators_inc} ONLY_IF_DIFFERENT)
file(REMOVE ${accelerators_inc}.tmp)
```

In this case, you don't end up with a target that you can use for dependencies elsewhere but the file gets created at generation time, so it should be available at build time. It is better not to write directly to the `.inc` file because then cmake won't rebuild the targets if the file hasn't changed, but you could skip the `.tmp` file as well.

Comment (etiotto) [reply src/Accelerators/CMakeLists.txt]: Thanks @sstamenova  I have done something like that, testing it and will commit asap.

File: src/Accelerators/CMakeLists.txt [hunk 8]
```diff
@@ -12,17 +12,14 @@ endif (ACCELERATORS_TO_BUILD)
...
-  foreach(t ${ACCELERATORS_TO_BUILD})
-    list(APPEND accelerators_inc_generate_commands
-      COMMAND ${CMAKE_COMMAND} -E echo_append "  MACRO'('${t}')' \\ " >> ${accelerators_inc})
-  endforeach(t)
-  list(APPEND accelerators_inc_generate_commands
-      COMMAND ${CMAKE_COMMAND} -E echo >> ${accelerators_inc})
-endif (ACCELERATORS_TO_BUILD)
+  COMMAND ${CMAKE_COMMAND} "-DINC_FILE=${accelerators_inc}.tmp" "-DACCELERATORS=${ACCELERATORS_TO_BUILD}" 
+    -P "${create_accelerators_inc_script}")
+list(APPEND accelerators_inc_generate_commands
```
sstamenova, line 19: These can be inside the `.cmake` file. For the versions they are not because generation differs based on how we add the llvm version, but here they can all be together and you could pass the name of the `.inc` file you want to create and let the `.cmake` file decide how to create the `.tmp` name and do the copy, etc.

src/Accelerators/CMakeLists.txt
sstamenova: Thanks! Isn't cmake fun?

sstamenova:
LGTM, thanks for cleaning things up!

sstamenova (approved):

sstamenova:
`COPY_FILE` is part of cmake 3.21: https://cmake.org/cmake/help/latest/command/file.html#copy-file. I would suggest upgrade the required `cmake` versions if possible since there are lots of nice new features that can be leveraged, but if you prefer not to change it, you can see how `llvm` handles the copy-delete here:
https://github.com/llvm/llvm-project/blob/16ed79d569524a5ed59a20ba54089d744e851ea0/llvm/cmake/modules/GenerateVersionFromVCS.cmake

src/Accelerators/CMakeLists.txt
etiotto: Thanks @sstamenova this exercise definitely helped me learn a bit more about cmake. Now that I start to understand a bit more about cmake and its capabilities is much more enjoyable :-). Thanks for holding my hand on this, very much appreciated!

gongsu832:
It's better not to require upgrading cmake unless absolutely necessary. Since our Jenkins build is based on Ubuntu Focal and RedHat ubi8 where the latest official cmake versions are 3.16 and 3.20, respectively.

etiotto:
> It's better not to require upgrading cmake unless absolutely necessary. Since our Jenkins build is based on Ubuntu Focal and RedHat ubi8 where the latest official cmake versions are 3.16 and 3.20, respectively.

For this PR I ended up using the LLVM approach that does not require upgrading cmake.

PR #1300, opened 2022-03-31
Fix warnings in onnx-mlir

whitneywhtsang:
```
onnx-mlir/src/Accelerators/NNPA/Transform/ZHigh/Stickify/Stickify.cpp:204:1: warning: control reaches end of non-void function [-Wreturn-type]
onnx-mlir/src/Accelerators/NNPA/Transform/ZHigh/Stickify/Stickify.cpp:219:1: warning: control reaches end of non-void function [-Wreturn-type]
onnx-mlir/src/Runtime/jni/jniwrapper.c:755:24: warning: format ‘%lld’ expects argument of type ‘long long int’, but argument 3 has type ‘int64_t’ {aka ‘long int’} [-Wformat=]
```


etiotto:
@whitneywhtsang we should probably enable the build compilers options that change warnings into errors. I meant to do that after doing several rounds of code cleanup but it went on the backburner ...

etiotto (approved):
LGTM!

PR #1304, opened 2022-03-31
Add support to limit the affine scope

chentong319:
Krnl loop may not are affine wrt the function region. This PR provide the support to limit the affine scope.
1. KrnlRegionOp is introduced. The region of this op has `AffineScope` trait and all the symbol defined at top of this region will be valid symbol for affine.for inside this region.
2. KrnlRegionOp is added to the body of LoopOp and the dynamic accumulation of loop result. Previously, verifier of affine.for failed for those loops.
3. A pass is added to remove KrnlRegionOp after the Affine.for has been converted to scf.for.
4. Models, test_loop11 (added into check-onnx-backend) and ccf model can be successfully compiled with this PR. 

Future work:
1. Modify create.krnl.iterate to add a KrnlRegionOp outside controlled by a parameter or command line option.


tungld:
@chentong319 a recent commit changed `Pass` to `mlir::Pass`, it should be done also for the new pass in this patch.

tungld:
> Future work:
Modify create.krnl.iterate to add a KrnlRegionOp outside controlled by a parameter or command line option.

Whether we can do this automatically, e.g. by checking whether a symbol is valid or not, if not, wrapping the loop inside AffineScope-traited KrnlRegionOp to make it valid. Not prefer to expose KrnlRegionOp to users, because it's not straightforward to know when we should use KrnlRegionOp or not, given that intermediate generated code may be edited somehow. 

Just for my learning: is there a situation where an invalid symbol becomes valid after some optimizations?

chentong319:
@tungld We can analyze whether a symbol is valid or not. It should be the same code of affine.for verifier. I do not know how to reuse that code for our purpose. We may try to apply KrnlToAffine to each krnl.iterate, and if it failed, we add krnl.region and try again, instead of signalPassFailure(). Do you think the approach feasible?
For your second question: yes, optimization may turn an invalid symbol into valid one. For examples:
1. Constant folding turns a symbol into constant
2. Aggressive code motion may move SSA definition out of IF or Loop to become a definition at the top of FuncOp
3. Control flow optimization may remove the outside loop or if.

Therefore, the best solution is what I mentioned previously: add krnl.region when it is really needed in converting krnl.iterate to affine.for (after all the possible optimizations)

chentong319:
I called verifier explicitly after Krnl.iterate is converted to affine.for. The error can be caught. So it is possible that we can add krnl.region there.  I guess we need to handle affine.load too. The current code has some data structure for loop nest, which I need to understand before I can change the code. 
Another task is to intercept the output of emitError coming from the verifier.

test/backend/inference_backend.py
```diff
@@ -380,6 +380,9 @@ def get_test_models():
         "test_logsoftmax_negative_axis_cpu": {STATIC_SHAPE:{}, DYNAMIC_SHAPE:{-1:{-1}}, CONSTANT_INPUT:{-1}},
         "test_logsoftmax_large_number_cpu": {STATIC_SHAPE:{}, DYNAMIC_SHAPE:{-1:{-1}}, CONSTANT_INPUT:{-1}},
 
+        # LoopOp
+        "test_loop11_cpu": {STATIC_SHAPE:{}, DYNAMIC_SHAPE:{}, CONSTANT_INPUT:{}},
```
tungld, line 384: Are you sure `DYNAMIC_SHAPE:{}` will invoke dynamic tests? I remember, for `DYNAMIC_SHAPE` and `CONSTANT_INPUT`, we need to use `DYNAMIC_SHAPE:{-1:{-1}}` to trigger.

tungld:
@chentong319 thank you for answering my questions!

> We may try to apply KrnlToAffine to each krnl.iterate, and if it failed, we add krnl.region and try again, instead of signalPassFailure(). Do you think the approach feasible in MLIR?

It looks possible if we do not return signalPassFailure(), but I am not 100% sure. My concern is that it is quite expensive. Instead, I see some passes in MLIR do a quick analysis to decide what to do next, e.g. affine-loop-fusion builds a graph of affine.for, affine.load and affine.store, and computes costs before fusion.

Anyway, I think this patch is good to go. Please check backend tests again to make sure it actually runs.

test/backend/inference_backend.py
chentong319: The DYNAMIC_SHAPE is NOT really tested here.  Need some investigation for this test case. I will come back on this later.

chentong319:
@tungld We may reuse the code of the analysis to build affine.for nest.

test/backend/inference_backend.py
chentong319: The issue fixed.

etiotto (commented):
I see that a new test has been added to the backend tests, which is good. Can you also add a lit test (or more) to verify the code generation of krnl.Region and that the transformation which is supposed to remove the krnl.Region runs correctly. Thanks.

test/backend/inference_backend.py
tungld: Thanks!

tungld (approved):
LGTM. Please address Ettore's comments.

AlexandreEichenberger (approved):
LGTM, at some times, it would be good to add a method in the KrnlBuilder that build a region where the inner code is passed as a lambda. I understand that the code you modified still use the old frameowork.

Also, it would be good if you could generate the new krnl md file, which is recommended whenever changing onnx or krnl dialects so that our doc is in sync.

```
make onnx-mlir-docs
```

PR #1303, opened 2022-03-31
[ReverseSequence] Use builder based interface to generate Krnl loops

whitneywhtsang:
Signed-off-by: Whitney Tsang <whitneyt@ca.ibm.com>

PR #1305, opened 2022-03-31
Hoist invariant literal outside of loop

whitneywhtsang:

### DIFF
--- a/src/Conversion/ONNXToKrnl/Math/Softmax.cpp
+++ b/src/Conversion/ONNXToKrnl/Math/Softmax.cpp
@@ -126,6 +126,7 @@ static void emitInstForSoftmaxBeforeV13(ConversionPatternRewriter &rewriter,
   KrnlBuilder createKrnl(rewriter, loc);
   IndexExprScope ieScope(createKrnl);
   MemRefBoundsIndexCapture inputBounds(input);
+  LiteralIndexExpr zeroIE(0);
 
   // Coerce the input into a 2-D tensor. `axis` will be the coercing
   // point. This coercing follows the softmax definition in ONNX:
@@ -141,7 +142,7 @@ static void emitInstForSoftmaxBeforeV13(ConversionPatternRewriter &rewriter,
 
     // Common information to create nested loops.
     int64_t numberOfLoops = rank;
-    SmallVector<IndexExpr, 4> Lbs(numberOfLoops, LiteralIndexExpr(0));
+    SmallVector<IndexExpr, 4> Lbs(numberOfLoops, zeroIE);
     SmallVector<IndexExpr, 4> Ubs;
     inputBounds.getDimList(Ubs);
 
@@ -150,7 +151,7 @@ static void emitInstForSoftmaxBeforeV13(ConversionPatternRewriter &rewriter,
   } else {
     // Define outer loops.
     ValueRange outerLoops = createKrnl.defineLoops(axis);
-    SmallVector<IndexExpr, 4> outerLbs(axis, LiteralIndexExpr(0));
+    SmallVector<IndexExpr, 4> outerLbs(axis, zeroIE);
     SmallVector<IndexExpr, 4> outerUbs;
     for (int i = 0; i < axis; ++i)
       outerUbs.emplace_back(inputBounds.getDim(i));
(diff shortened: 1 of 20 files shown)


tungld:
Are these all places we need to hoist invariant literals? Really appreciate if you scan the other lowerings and see if  there are any missing places.

FYI, these invariant literals will be hoisted automatically when lowering krnl to affine: https://github.com/onnx/onnx-mlir/blob/main/src/Conversion/KrnlToAffine/ConvertKrnlToAffine.cpp#L603

tungld:
Are these all places we need to hoist invariant literals? Really appreciate if you scan the other lowerings and see if  there are any missing places.

FYI, these invariant literals will be hoisted automatically when lowering krnl to affine: https://github.com/onnx/onnx-mlir/blob/main/src/Conversion/KrnlToAffine/ConvertKrnlToAffine.cpp#L603

whitneywhtsang:
> Are these all places we need to hoist invariant literals? Really appreciate if you scan the other lowerings and see if there are any missing places.

Probably not, I was trying to address the review comment (https://github.com/onnx/onnx-mlir/pull/1303#discussion_r840044030), and hoisting only the two lowering I worked on. Sure, let me scan the other lowerings and see if there are any missing places.

> FYI, these invariant literals will be hoisted automatically when lowering krnl to affine: https://github.com/onnx/onnx-mlir/blob/main/src/Conversion/KrnlToAffine/ConvertKrnlToAffine.cpp#L603

That's good to know. IMO we should still do this change, as it is simple and we don't need to depend on later on optimizations/lowering.

etiotto (approved):
LGTM

src/Conversion/ONNXToKrnl/Math/Softmax.cpp
```diff
@@ -126,6 +126,7 @@ static void emitInstForSoftmaxBeforeV13(ConversionPatternRewriter &rewriter,
   KrnlBuilder createKrnl(rewriter, loc);
   IndexExprScope ieScope(createKrnl);
   MemRefBoundsIndexCapture inputBounds(input);
+  LiteralIndexExpr zeroIE(0);
```
etiotto, line 129: [nit]: some files in this PR use the name `zero` and others `zeroIE` for the same expr. Optional: make them consistent.

src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
```diff
@@ -520,7 +520,7 @@ Value emitArgSort(ConversionPatternRewriter &rewriter, Location loc,
   ValueRange loopDef = createKrnl.defineLoops(rank);
   createKrnl.iterateIE(loopDef, loopDef, lbs, outerUbs,
       [&](KrnlBuilder &createKrnl, ValueRange iLoopInd) {
-        IndexExpr i1 = DimIndexExpr(iLoopInd[axis]) + LiteralIndexExpr(1);
+        IndexExpr i1 = DimIndexExpr(iLoopInd[axis]) + oneIE;
```
etiotto, line 523: [nit]: some files in this PR use the name `one` and others `oneIE` for the same expr. Optional: make them consistent.

src/Conversion/ONNXToKrnl/Tensor/Range.cpp
```diff
@@ -113,6 +113,7 @@ struct ONNXRangeOpLowering : public ConversionPattern {
     }
 
     // Create a single loop.
+    LiteralIndexExpr zeroIE(0);
```
etiotto, line 116: [nit]: I would prefer to reduce the live range for this expression (declare closer to its use).

etiotto (approved):
LGTM

src/Conversion/ONNXToKrnl/Math/Softmax.cpp
whitneywhtsang: The reason I use `zeroIE` instead of `zero` is `zero` is already used. I can change all `zero` index expr to `zeroIE` instead.

src/Conversion/ONNXToKrnl/Tensor/Range.cpp
whitneywhtsang: I intentionally define it before `BuildKrnlLoop`, note that krnlLoop will be change to use the builder based interface.

etiotto: ok so is future proof. that is fine.

src/Conversion/ONNXToKrnl/Math/Softmax.cpp
etiotto: ok got it, no don't bother.

whitneywhtsang: Changed it anyways.

PR #1307, opened 2022-04-01
Set functions to be accessible from DLL on Windows.

tungld:
This patch is to annotate functions to be exported on Windows without `.def` files where
- `__declspec(dllexport)` is added to functions in cruntime lib, including `omTensorCreate`, `omTensorListCreate`, etc.
- `dllexport` is added to llvm functions generated by KrnlToLLVM, including `run_main_graph`, `omInputSignature`, `omOutputSignature`, `omQueryEntryPoints`.

This patch also introduces `preserveLLVMIR` option to `onnx-mlir` to save the generated LLVM code into a `.ll` file. `preserveLLVMIR` is used to test annotated functions in this patch.


src/Compiler/CompilerUtils.cpp
```diff
@@ -308,6 +308,26 @@ static void genLLVMBitcode(const mlir::OwningOpRef<ModuleOp> &module,
...
+    exportedFuncs.emplace_back(StringRef("run_main_graph"));
+    exportedFuncs.emplace_back(StringRef("omQueryEntryPoints"));
+    exportedFuncs.emplace_back(StringRef("omInputSignature"));
+    exportedFuncs.emplace_back(StringRef("omOutputSignature"));
+    for (StringRef funcName : exportedFuncs) {
+      llvm::GlobalValue *GV = llvmModule->getNamedValue(funcName);
+      GV->setDSOLocal(true);
+      GV->setDLLStorageClass(llvm::GlobalValue::DLLExportStorageClass);
+      auto func = llvmModule->getFunction(funcName);
+      func->print(llvm::outs());
```
tungld, line 327: These two lines are just for debugging purpose. Will be removed.

src/Compiler/CompilerUtils.cpp
```diff
@@ -308,6 +308,26 @@ static void genLLVMBitcode(const mlir::OwningOpRef<ModuleOp> &module,
   llvm::Metadata *identNode[] = {llvm::MDString::get(ctx, OnnxMlirVersion)};
   identMetadata->addOperand(llvm::MDNode::get(ctx, identNode));
 
+  // Set functions to be accesible from DLL on Windows.
+  if (llvm::Triple(getTargetTripleOption()).isOSWindows()) {
```
tungld, line 312: @sstamenova it seems we cannot detect Windows OS from triple. I don't know why it is. I don't have Windows environement, so could you help check what happened here? 

There are several functions that can be called to check different  Windows environments here: https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/ADT/Triple.h#L561, and I chose the general one.

src/Compiler/CompilerUtils.cpp
```diff
@@ -372,10 +392,9 @@ static std::string genSharedLib(string outputBaseName, std::vector<string> opts,
 #ifdef _WIN32
   string sharedLibPath = outputBaseName + ".dll";
   std::vector<string> outputOpt = {"/Fe:" + sharedLibPath};
-  // link has to be before def and libpath since they need to be passed through
-  // to the linker
-  std::vector<string> sharedLibOpts = {
-      "/LD", "/link", "/NOLOGO", "/def:" + outputBaseName + ".def"};
```
sstamenova, line 378: If we remove the .def file, the functions exported from cruntimelib also need to be marked as dll exports because they can no longer be exported in any other way

src/Compiler/CompilerUtils.cpp
sstamenova: @tungld : What do you mean? `isOSWindows` should return `true` when the triple is targeting Windows, what's the case when you saw it fail?

tungld: It looks `llvm::Triple(getTargetTripleOption()).isOSWindows()` returned false so that the code inside if was not executed, and numerical tests failed. If I used `if(true)`, the code was executed and numerical tests passed. I suspect that `llvm::Triple(getTargetTripleOption()).isOSWindows()` returned false. I cannot confirm this from Windows CI logs, could you please check on your local environment? Thanks!

tungld: I switched to use `ifdef __WIN32` instead of triple.

src/Compiler/CompilerUtils.cpp
tungld: Done. Thank you!

src/Compiler/CompilerUtils.cpp
```diff
@@ -308,6 +308,21 @@ static void genLLVMBitcode(const mlir::OwningOpRef<ModuleOp> &module,
...
+  exportedFuncs.emplace_back(StringRef("omQueryEntryPoints"));
+  exportedFuncs.emplace_back(StringRef("omInputSignature"));
+  exportedFuncs.emplace_back(StringRef("omOutputSignature"));
+  for (StringRef funcName : exportedFuncs) {
+    llvm::GlobalValue *GV = llvmModule->getNamedValue(funcName);
+    GV->setDSOLocal(true);
+    GV->setDLLStorageClass(llvm::GlobalValue::DLLExportStorageClass);
+  }
+#endif
+
```
gongsu832, line 325: @tungld Can you explain a bit how this worked? I would have expected that once you annotate the bitcode properly for `run_main_graph` etc. to be exported, the rest are all taken care of by llc and the linker. Why do we still have to do something in `CompilerUtils.cpp` to export the symbols?

tungld: @gongsu832 Firstly we should distinguish "LLVMIR in mlir" which is a dialect in MLIR and "LLVMIR in llvm" which is the original LLVM existed long time ago before MLIR.

"LLVMIR in mlir" is limited and still not equivalent to "LLVMIR in llvm" , for example we cannot annotate a function with dllexport in "LLVMIR in mlir".

Now, our onnx-mlir  or CompilerUtils in particular, invokes KrnlToLLVM to generates "LLVMIR in mlir", then calls `translateModuleToLLVMIR` to translate "LLVMIR in mlir" to "LLVMIR in llvm`, then tranlates to bitcode, and link etc.

What I did here is to annotate functions of the generated code when they are in "LLVMIR in llvm" so that the functions are equivalent to having `__declspec(dllexport)` (e.g. when -emit-llvm for a C++ code with  `__declspec(dllexport)`. The rest will be all taken care of by llc and the linker as you said.

gongsu832: Thanks @tungld ! Can you put this as comments above this block of code? Makes it easier to understand how this works.

gongsu832: BTW, just saw your comment above:

> This patch cannot handle a model with multiple entry points, e.g. entry points are not `run_main_graph`, but something else that must be queried via `omQueryEntryPoints`.

Is this still true? There is no way to annotate entry points other than `run_main_graph`?

include/onnx-mlir/Compiler/CMakeLists.txt
```diff
@@ -1,3 +1,4 @@
 # SPDX-License-Identifier: Apache-2.0
 
 install(FILES OMCompilerTypes.h DESTINATION include/onnx-mlir/Compiler)
+install(FILES OMCompilerMacros.h DESTINATION include/onnx-mlir/Compiler)
```
sstamenova, line 4: Are all the headers being installed now? In the PR I had sent, I had added a few headers that were previously missing to the install. Can you do that too? https://github.com/onnx/onnx-mlir/pull/1262

include/onnx-mlir/Compiler/OMCompilerMacros.h
```diff
@@ -0,0 +1,29 @@
...
+// This file defines several macros, which allows use of compiler-specific
+// features in a way that remains portable. This header can be included from
+// either C or C++.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef ONNX_MLIR_COMPILER_MACROS_H
+#define ONNX_MLIR_COMPILER_MACROS_H
+
+/// OM_EXTERNAL_VISIBILITY - classes, functions, and variables marked with this
```
sstamenova, line 20: ONNX_MLIR_EXPORT is already defined in OnnxMlirCompier.h. I don't think we need two definitions, so please either use ONNX_MLIR_EXPORT everywhere, or change the definition of OM_EXTERNAL_VISIBILITY, so that it can be used instead of ONNX_MLIR_EXPORT

tungld: I tried to use ONNX_MLIR_EXPORT but got this failure: https://dev.azure.com/onnx-pipelines/onnx/_build/results?buildId=22626&view=logs&j=4060f429-1919-5440-1fce-77a91697330b&t=4cbc8f84-8349-586a-8281-73be4740c2c3, and had no Windows environment to debug.

The definition of ONNX_MLIR_EXPORT is more complicated than I expected and I am not totally understand. Really appreciate if someone from your team could help with this issue by a follow-up PR.

gongsu832: @tungld @sstamenova I will create a PR to clean up the include stuff and hopefully make the ONNX_MLIR_EXPORT work for both Compiler and Runtime libraries.

tungld: Thank you @gongsu832!

include/onnx-mlir/Compiler/CMakeLists.txt
tungld: Done. I pulled them from your PR.

src/Compiler/CompilerUtils.cpp
tungld: It's feasible, e.g. entry points are constants in LLVMIR and we can read them. I will address this in another PR. It requires a little more effort.

tungld: > Can you put this as comments above this block of code? Makes it easier to understand how this works.

I created a separate function `tailorLLVMIR` and put all related code there. Hope it make things clearer.

gongsu832: > It's feasible, e.g. entry points are constants in LLVMIR and we can read them. I will address this in another PR. It requires a little more effort.

That would be perfect! With this we would then completely address the symbol export issue for Windows without a .def file.

tungld: @gongsu832 anyway, I just added support for multiple entry points. So this patch fully supports dllexport for the generated code.

sstamenova:
LGTM, thanks!

sstamenova (approved):

gongsu832 (approved):
LGTM. Thanks for the work. Now we support Windows DLL without having to fiddle with .def files, which is great!

PR #1313, opened 2022-04-04
Fix JNI wrapper entry point name encoding on z/OS

gongsu832:
When Java API inputSignature()/outputSignature() calls
input_signature_jni("run_main_graph")/output_signature_jni("run_main_graph"),
"run_main_graph" needs to be converted from ASCII to EBCDIC before being
fed into omInputSignature(epptr)/omOutputSignature(epptr)


JonTriebenbach:
This line also has to be updated:
https://github.com/onnx/onnx-mlir/blob/917c90b2ee885ef554a1973440f5b735282c3d3a/src/Runtime/jni/jniwrapper.c#L768

The model build fails on zOS with this failure:
 IEW2456E 9207 SYMBOL sig_e2a UNRESOLVED.  MEMBER COULD NOT BE INCLUDED FROM THE
          DESIGNATED CALL LIBRARY.                                              
FSUM3065 The LINKEDIT step ended with return code 8.

tungld (approved):
LGTM.

PR #1319, opened 2022-04-06
Use cmake functions add_onnx_mlir_* for NNPA

tungld:
This patch is to use cmake functions `add_onnx_mlir_dialect`, `add_onnx_mlir_rewrite`, etc. for NNPA code.


src/Accelerators/NNPA/Dialect/ZLow/ZLowOps.cpp
```diff
@@ -2,7 +2,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-//===------------------ ZLowOps.hpp - ONNX Operations ---------------------===//
+//===------------------ ZLow.hpp - ONNX Operations ------------------------===//
```
etiotto, line 5: The name here `ZLow.hpp' doesn't match the file name `ZLowOps.cpp`

sstamenova (approved):

PR #1320, opened 2022-04-06
Reproducible random numbers for filling tensors

AlexandreEichenberger:
Currently, our numerical tests generate random numbers in a non-reproducible way.

Using this PR, a new call was added to the runtime helpers to either print the seed that will be used, or read the seed from an env variable. For example:
``` bash
root@1250b049ac2b:~/onnx-mlir/build# Debug/bin/TestMatMul2D 
Model can reuse the current seed by exporting "TEST_SEED=1097851945"
...
```

ran without the TEST_SEED being defined, so a new random seed was selected. If we were to run the same test again but exporting `TEST_SEED=1097851945`, then that seed would be used, thus enabling us to re-run the same experiment with the same random numbers.

The call to handle the random seeds is here.
```C++
**
 * omDefineSeed.
 * When called, the random number generator for omTensorCreateWithRandomData
 * will be seeded exactly once. The seed is randomly generated when ignoreSeed
 * is nonnull; otherwise the input seedValue is used.
 *
 * @param seed input seed.
 * @param ignoreSeed when nonzero, ths function define its own random seed.
 * @return return the seed that was used.
 *
 */
unsigned int omDefineSeed(unsigned int seed, unsigned int ignoreSeed);
```

The handling of an env variable is done by a static function in `ModelLibBuilder` top class:
```C++
  // Set the random number generator seed to the value passed by the environment
  // variable; if not found, use a random seed. Optional call to enable
  // reproducible random numbers.
  static void setRandomNumberGeneratorSeed(const std::string &envVar);
```

### DIFF
--- a/src/Conversion/ONNXToKrnl/Tensor/Transpose.cpp
+++ b/src/Conversion/ONNXToKrnl/Tensor/Transpose.cpp
@@ -46,6 +46,28 @@ struct ONNXTransposeOpLowering : public ConversionPattern {
     (void)shapecomputed;
     assert(succeeded(shapecomputed) && "Could not compute output shape");
 
+    // If the order of the dimensions whose value is not 1 does not change after
+    // transpose, it is safe to lower transpose to a view op.
+    ArrayRef<int64_t> dims = memRefType.getShape();
+    SmallVector<int64_t, 4> originalAxes;
+    for (uint64_t axis = 0; axis < dims.size(); ++axis)
+      if (dims[axis] != 1)
+        originalAxes.emplace_back(axis);
+    SmallVector<int64_t, 4> permutedAxes;
+    for (uint64_t i = 0; i < rank; ++i) {
+      int64_t axis = ArrayAttrIntVal(permAttr, i);
+      if (dims[axis] != 1)
+        permutedAxes.emplace_back(axis);
+    }
+    if (originalAxes == permutedAxes) {
+      // It is safe to lower to a view op.
+      MemRefBuilder createMemRef(rewriter, loc);
+      Value view =
+          createMemRef.reinterpretCast(data, shapeHelper.dimsForOutput());
+      rewriter.replaceOp(op, view);
+      return success();
+    }
+
     // Insert an allocation and deallocation for the result of this operation.
     Value alloc = insertAllocAndDeallocSimple(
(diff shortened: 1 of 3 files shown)


src/Runtime/OMTensor.inc
```diff
@@ -526,14 +527,33 @@ OMTensor *omTensorCreateWithShape(const std::vector<int64_t> &shape) {
   return omt;
 }
 
+// Static variables used by omDefineSeed and omTensorCreateWithRandomData.
+static unsigned int omUseOneSeed = 0;
+static std::mt19937 omRandomGenerator(0);
+
+// When called, a single seed will be used, either from the given seed or from
+// another random number generator.
+unsigned int omDefineSeed(unsigned int seed, unsigned int ignoreSeed) {
```
etiotto, line 536: [Suggestion]: use an Optional to indicate that the `seed` might not be provided ? As in: 
```
unsigned omDefineSeed(Optional<unsigned> optionalSeed) {
  std::random_device rd;
  unsigned seed = optionalSeed.hasValue() ? optionalSeed.getValue() : rd();
  ...
```

test/modellib/GemmModel.cpp
```diff
@@ -91,6 +91,21 @@ bool GemmLibBuilder::prepareInputs() {
   return inputs && list[0] && list[1] && list[2];
 }
 
+bool GemmLibBuilder::prepareInputs(float dataRange) {
+  const int num = 3;
```
etiotto, line 95: const -> constexpr

test/modellib/MatMulModel.cpp
```diff
@@ -67,6 +67,17 @@ bool MatMul2DLibBuilder::prepareInputs() {
   return inputs && list[0] && list[1];
 }
 
+bool MatMul2DLibBuilder::prepareInputs(float dataRange) {
+  const int num = 2;
```
etiotto, line 71: const -> constexpr

src/Runtime/OMTensor.inc
AlexandreEichenberger: While I agree that the optional is much more elegant, currently the file has no dependence to LLVM, so adding `Optional` just for this use seems like a big change. I will change "ignoreSeed" to `hasSeedValue` so that it may feel more like the optional interface.

etiotto (approved):
LGTM

PR #1321, opened 2022-04-06
Use onnx-to-krnl and krnl-to-llvm to handle NNPA lowerings

tungld:
Currently there is code duplication between onnx-to-krnl and zhigh-to-zlow, and between krnl-to-llvm and zlow-to-llvm, which is not good and error-prone.

This patch removes the duplication by just using onnx-to-krnl and krnl-to-llvm and providing hooks for accelerators. This mechanism is consistent with what we are doing to support accelerators in `onnx-mlir` and `onnx-mlir-opt` drivers.

In summary, to support a new accelerator, users are expected to write hooks for 4 components: `onnx-mlir`, `onnx-mlir-opt`, `onnx-to-krnl`, and `krnl-to-llvm` in their own accelerator class derived from the Accelarator class.


src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
```diff
@@ -143,9 +111,8 @@ void addPassesNNPA(mlir::OwningOpRef<mlir::ModuleOp> &module,
       emissionTarget = EmitMLIR;
     else {
       pm.addPass(mlir::createCanonicalizerPass());
-      // Add instrumentation for remaining Onnx Ops
-      if (instrumentZHighOps != "" && instrumentZHighOps != "NONE")
-        pm.addNestedPass<FuncOp>(createInstrumentONNXPass());
+      // Add instrumentation for ZHigh Ops
+      pm.addNestedPass<FuncOp>(onnx_mlir::zhigh::createInstrumentZHighPass());
```
etiotto, line 115: `onnx_mlir::`   shouldn't be required given we are inside that namespace already.

src/Accelerators/NNPA/Conversion/ZLowToLLVM/ZLowToLLVM.hpp
```diff
@@ -0,0 +1,26 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//===---------- ZLowToLLVM.cpp - Lowering from ZLow to LLVM ---------------===//
```
etiotto, line 5: ZLowToLLVM.cpp -> ZLowToLLVM.hpp

src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
tungld: Done. Thanks!

src/Accelerators/NNPA/Conversion/ZLowToLLVM/ZLowToLLVM.hpp
tungld: Done. Thanks!

etiotto (approved):
LGTM

tungld:

tungld:

PR #1323, opened 2022-04-06
Fix issue when onnx outputs are also used by other nodes

### LINKED ISSUE #1376: Importing onnx graph fails if an output is also used by another node
Importing onnx graph fails if an output is also used by another node. This happens because the output ValueInfo will be registered, and then it will throw an error that it already exists when importing internal ValueInfos.

For example, in the below graph, the result of Relu is used as both an output and an input to another node.
bug

MikeHolman:
Importing onnx graph fails if an output is also used by another node. This happens because the output ValueInfo will be registered, and then it will throw an error that it already exists when importing internal ValueInfos.

Solution is to import the internal ValueInfos before importing the output ValueInfos.

Resolves #1376


sstamenova:

sstamenova:

tungld:
@chentong319 could you take a look at this? thanks!

@MikeHolman I am still confused about the issue. Is it possible to create an issue to explain the situation with your onnx model in details, and link it to this PR?

MikeHolman:
@tungld I've opened #1376. Hopefully that makes the problem a bit more clear.

chentong319:
We ran into a similar case previously.
The document say value_info "Used to store the type and shape information of values that are not inputs or outputs." in [doc link](https://github.com/onnx/onnx/blob/main/docs/IR.md)
My interpretation is that the value_info should not be inputs or outputs. Or does the document just mean that all the tensors that are not described in inputs or outputs list have to be in value_info list, and a tensor can appear more than once in these list?

AlexandreEichenberger:
@chentong319 does the picture shown in #1376 clarify the case being solved?

chentong319:
> @chentong319 does the picture shown in #1376 clarify the case being solved?

No, I do not think that those lists are shown in the picture.

chentong319 (approved):
LGTM!
Even though the document is ambiguous to me, this change will not cause trouble if the model is correct.

MikeHolman:
@chentong319 Thanks for the review! In my model, onnx::Gemm_6 tensor is in both value_info and output lists. The graph was created by onnxruntime, so I think it should be correct.

PR #1330, opened 2022-04-08
Call MLIR CSE pass to remove common sub-expressions

tungld:
This patch just calls MLIR CSE pass to remove common sub-expressions:
- calling CSE before lowering ops to Krnl
- calling CSE before lowering ops to LLVM

This sub-expression removal is useful to remove stick ops in NNPA, for example:
```mlir
%y = cpu_op(x)
%s_y1 = stick(y) {layout = "NHWC"}
%z_1 = nnpa_op1(s_y1)
%s_y2 = stick(y) {layout = "NHWC"}
%z_2 = nnpa_op2(s_y2)
%s_y3 = stick(y) {layout = "NHWC"}
%z_3 = nnpa_op3(s_y3)
```
 will become
 ```mlir
%y = cpu_op(x)
%s_y = stick(y) {layout = "NHWC"}
%z_1 = nnpa_op1(s_y)
%z_2 = nnpa_op2(s_y)
%z_3 = nnpa_op3(s_y)
```
 ```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,6 +60,9 @@ set(ONNX_MLIR_BIN_ROOT ${CMAKE_CURRENT_BINARY_DIR})
 set(ONNX_MLIR_LIBRARY_PATH ${CMAKE_LIBRARY_OUTPUT_DIRECTORY})
 set(ONNX_MLIR_RUNTIME_PATH ${CMAKE_RUNTIME_OUTPUT_DIRECTORY})
 
+set(ONNX_MLIR_VENDOR ${PACKAGE_VENDOR} CACHE STRING
+  "Vendor-specific text for showing with version information.")
+
 include(CTest)
 include(ExternalProject)
 include(MLIR.cmake)
--- a/src/Accelerators/Accelerator.hpp
+++ b/src/Accelerators/Accelerator.hpp
@@ -14,11 +14,14 @@
 
 #pragma once
 
-#include "include/onnx-mlir/Compiler/OMCompilerTypes.h"
+#include "mlir/Conversion/LLVMCommon/TypeConverter.h"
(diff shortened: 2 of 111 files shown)


tungld:
> Looks fine, but could you add a lit test to check that common sub expressions are removed please.

There are many tests for common sub-expression in llvm-project: https://github.com/llvm/llvm-project/blob/main/mlir/test/Transforms/cse.mlir. Should we test them again in onnx-mlir?

etiotto:
> > Looks fine, but could you add a lit test to check that common sub expressions are removed please.
> 
> There are many tests for common sub-expression in llvm-project: https://github.com/llvm/llvm-project/blob/main/mlir/test/Transforms/cse.mlir. Should we test them again in onnx-mlir?

I know, I wasn't not asking for extensive testing, I would add maybe one or two lit tests in `onnx-mlir`.

tungld:
> I know, I wasn't not asking for extensive testing, I would add maybe one or two lit tests in onnx-mlir.

@etiotto I added one lit test to show that it works with ONNX operations.

etiotto (approved):
LGTM

AlexandreEichenberger:
@tungld Thanks for the PR. Still had a question of why each of the uses for the Concat generated its own sequence of transpose/stick? We could change a bit the pattern generator that introduces the transpose/stick, which might be cheaper (in terms of compile time) than running CSE. However, I agree that the pattern generated by the naive pattern is guaranteed to be picked by CSE. Let's revisit this only if it slows down the compiler significantly.

tungld:
> Still had a question of why each of the uses for the Concat generated its own sequence of transpose/stick? 

It's because there are Conv ops consuming the Concat's output, and each Conv op when lowered to ZHigh generates transpose/stick for each of its inputs.

PR #1333, opened 2022-04-11
`-DPACKAGE_VENDOR` can be used to show vendor-specific text as version information

whitneywhtsang:
This change adds a `PACKAGE_VENDOR` variable. When set it makes the version output more closely resemble the `onnx-mlir` version output. The idea came from llvm-project: https://github.com/llvm/llvm-project/blob/ceadf6ee619ce610b269459b129996c4ac7173cb/clang/CMakeLists.txt#L340.

For example, when building with `-DPACKAGE_VENDOR="XL for Linux on Power"`, 
then output LLVMIR contains:
```
!llvm.ident = !{!0}
!0 = !{!"XL for Linux on Power, onnx-mlir version 1.0.0"}
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,6 +60,11 @@ set(ONNX_MLIR_BIN_ROOT ${CMAKE_CURRENT_BINARY_DIR})
 set(ONNX_MLIR_LIBRARY_PATH ${CMAKE_LIBRARY_OUTPUT_DIRECTORY})
 set(ONNX_MLIR_RUNTIME_PATH ${CMAKE_RUNTIME_OUTPUT_DIRECTORY})
 
+if (PACKAGE_VENDOR)
+  set(ONNX_MLIR_VENDOR ${PACKAGE_VENDOR} CACHE STRING
+    "Vendor-specific text for showing with version information.")
+endif(PACKAGE_VENDOR)
+
 include(CTest)
 include(ExternalProject)
 include(MLIR.cmake)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,10 +60,8 @@ set(ONNX_MLIR_BIN_ROOT ${CMAKE_CURRENT_BINARY_DIR})
 set(ONNX_MLIR_LIBRARY_PATH ${CMAKE_LIBRARY_OUTPUT_DIRECTORY})
 set(ONNX_MLIR_RUNTIME_PATH ${CMAKE_RUNTIME_OUTPUT_DIRECTORY})
 
-if (PACKAGE_VENDOR)
-  set(ONNX_MLIR_VENDOR ${PACKAGE_VENDOR} CACHE STRING
-    "Vendor-specific text for showing with version information.")
-endif(PACKAGE_VENDOR)
+set(ONNX_MLIR_VENDOR ${PACKAGE_VENDOR} CACHE STRING
+  "Vendor-specific text for showing with version information.")
 
 include(CTest)
 include(ExternalProject)
--- a/src/Compiler/CompilerUtils.cpp
+++ b/src/Compiler/CompilerUtils.cpp
@@ -40,19 +40,22 @@ using namespace mlir;
 using namespace onnx_mlir;
 
 const std::string OnnxMlirEnvOptionName = "ONNX_MLIR_FLAGS";
-static const std::string OnnxMlirVersion = "onnx-mlir version 1.0.0";
-static const std::string OnnxMlirVersionString =
-#ifdef ONNX_MLIR_VENDOR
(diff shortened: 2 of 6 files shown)


sstamenova:
@whitneywhtsang : Can you give an example of how this would be used and what it would produce as well as some background for why it is needed? Ideally this would be in the description of the PR.

whitneywhtsang:
> @whitneywhtsang : Can you give an example of how this would be used and what it would produce as well as some background for why it is needed? Ideally this would be in the description of the PR.

@sstamenova My bad, updated the description of the PR. Please let me know if it is clear.

etiotto (approved):
LGTM. @sstamenova are you ok with this change?

CMakeLists.txt
```diff
@@ -60,6 +60,11 @@ set(ONNX_MLIR_BIN_ROOT ${CMAKE_CURRENT_BINARY_DIR})
 set(ONNX_MLIR_LIBRARY_PATH ${CMAKE_LIBRARY_OUTPUT_DIRECTORY})
 set(ONNX_MLIR_RUNTIME_PATH ${CMAKE_RUNTIME_OUTPUT_DIRECTORY})
 
+if (PACKAGE_VENDOR)
+  set(ONNX_MLIR_VENDOR ${PACKAGE_VENDOR} CACHE STRING
```
sstamenova, line 64: I think you can unconditionally set `ONNX_MLIR_VENDOR` to `PACKAGE_VENDOR`

test/mlir/driver/llvm.ident.mlir
```diff
@@ -3,7 +3,7 @@
 // RUN: cat %t.ll | FileCheck %s
 
 // CHECK: !llvm.ident = !{![[MD:[0-9]*]]}
-// CHECK: ![[MD]] = !{!"onnx-mlir version 1.0.0 ({{.*}}/onnx-mlir{{.*}} {{[a-zA-Z0-9]+}} {{.*}}/llvm-project{{.*}} {{[a-zA-Z0-9]+}})"}
+// CHECK: ![[MD]] = !{!"onnx-mlir version 1.0.0 ({{.*}}/onnx-mlir{{.*}} {{[a-zA-Z0-9]+}}, {{.*}}/llvm-project{{.*}} {{[a-zA-Z0-9]+}})"}
```
sstamenova, line 6: Will this test pass if the vendor is supplied?

src/Compiler/CompilerUtils.cpp
```diff
@@ -40,13 +40,16 @@ using namespace mlir;
 using namespace onnx_mlir;
 
 const std::string OnnxMlirEnvOptionName = "ONNX_MLIR_FLAGS";
-#if defined(ONNX_MLIR_REPOSITORY) && defined(ONNX_MLIR_REVISION) &&            \
+static const std::string OnnxMlirVersion = "onnx-mlir version 1.0.0";
```
sstamenova, line 43: I think at this point it's probably worth moving all of this to a function call similar to how clang does it. Then when we move to having a version file (which we probably really should), it would also make things much simpler.

sstamenova:
> LGTM. @sstamenova are you ok with this change?

Generally yes, I would like to make sure that the test would still pass if a vendor is supplied.

test/mlir/driver/llvm.ident.mlir
whitneywhtsang: That's a very good question, this test would fail if the vendor-specific text is supplied. 
Not sure what to do about that, may have to remove this test case. Any suggestions?

sstamenova: You can do a couple of things: modify the test so that it passed regardless - this could mean removing parts of the verification string, or adding all of the information to the version regardless (e.g. even if the vendor string is supplied maybe we still want the commits). The other thing you could do is include the test based on whether a vendor string was supplied.

whitneywhtsang: When vendor-specific text is supplied, this test case will fail expectedly (XFAIL) now.

src/Compiler/CompilerUtils.cpp
```diff
@@ -40,19 +40,22 @@ using namespace mlir;
 using namespace onnx_mlir;
 
 const std::string OnnxMlirEnvOptionName = "ONNX_MLIR_FLAGS";
-static const std::string OnnxMlirVersion = "onnx-mlir version 1.0.0";
-static const std::string OnnxMlirVersionString =
+
+namespace {
+
+static std::string getOnnxMlirFullVersion() {
```
sstamenova, line 46: This should move below with the other static functions

whitneywhtsang: sure, done.

sstamenova (approved):

PR #1335, opened 2022-04-12
[NNPA] Add numerical test for NNPA

imaihal:
This PR adds numerical tests for NNPA.

Currently tests for MatMul2D, Gemm, LSTM, and GRU are provided and run by using following command. 
```
cmake --build . --config Release --target check-onnx-numerical-nnpa
```
These tests uses the same test code with numerical tests for CPU (`test/modellib` and `test/numerial`), but uses different cmake file(`test/accelerator/NNPA/numerical/CMakeLists.txt`). 
- Gemm
   Since `alpha` and `beta` should be one for Matmul of zDNN library, #ifdef directive `TEST_GEMM_ALPHA_BETA_1` are added in `test/numerical/TestGemm.cpp` and set in the CMakeLists.txt (`test/accelerator/NNPA/numerical/CMakeLists.txt`)
- LSTM
  Since LSTM of zDNN library does not support peephole tensor, #ifdef directive `TEST_LSTM_NONEP_ONLY` are added in `test/numerial/TestLSTM.cpp` and set in the CMakeLists.
  Currently bidirectinal LSTM is not supported in onnx-mlir for NNPA, so, it is disabled by using #ifdef directive `TEST_RNN_NO_BIDIR` 
- GRU
   Since GRU of zDNN library does not support LinearBeforeReset, #ifdef directive `TEST_GRU_L1` are added in `test/numerial/TestGRU.cpp` and set in the CMakeLists.
  Currently bidirectinal LSTM is not supported in onnx-mlir for NNPA, so, it is disabled by using #ifdef directive `TEST_RNN_NO_BIDIR` 

To use these #ifdef directives, this PR updated `add_onnx_mlir_executable()` in MLIR.cmake.

-  ATOL and RTOL for accuracy check
  ATOL and RTOL need to be changed for tests. They are configured by environment variable `TEST_ATOL`  and `TES_RTOL` in CMakeLists.txt.

- Verification that zDNN instruction are generated in shared library
  `checkInstructionFromEnv()` in ModelLibBuilder verifies that the instruction specified the environment variable is included in generated shared library. The instructions such as `zdnn_matmul_op`, `zdnn_lstm`, and `zdnn_gru` are specified  in CMakeLists.txt. 

- Test results on a machine with NNPA (I removed the elapsed time because I don't know if I can publish it here.)
```
Test project /home/trl/imaihal/onnx-mlir/build/test/accelerators/NNPA/numerical
    Start 1: TestMatMul2DNNPA
1/4 Test #1: TestMatMul2DNNPA ................   Passed
    Start 2: TestGemmNNPA
2/4 Test #2: TestGemmNNPA ....................   Passed
    Start 3: TestLSTMNNPA
3/4 Test #3: TestLSTMNNPA ....................   Passed
    Start 4: TestGRUNNPA
4/4 Test #4: TestGRUNNPA .....................   Passed

100% tests passed, 0 tests failed out of 4
```


AlexandreEichenberger (changes requested):
The code changes are good. I would like for you to add a document page on `AccelNNPAHowToUseAndTest.md` and list the explanation that you gave on the conversation. Over time, this page will add more info.

Please make this addition part of this PR. Thanks

imaihal:
I will remove backend test from this PR because we need to discuss how to set ATOL and RTOL for backend test. We need to set them to pass backend test on NNPA.

tungld:
@imaihal change to draft. Please let us know when it is ready for review.

imaihal:
@gongsu832 When I removed `LD_PRELOAD` in newly added numerical tests in this PR, the results of numerical tests were broken. https://github.com/onnx/onnx-mlir/pull/1335/commits/ae1799e3333a94b08330ee4331c82c83a7f9a7a7 (When I used LD_PRELAD, I can get correct results) I guess I need to modify [CMakeLists.txt ](https://github.com/imaihal/onnx-mlir/blob/nnpa_tests/test/accelerators/NNPA/numerical/CMakeLists.txt) to link libzdnn.a correctly. Do you have any suggestions?

gongsu832:
I'm guessing it's not a cmake file problem but rather the zdnn_init() function wasn't getting called since now libzdnn is statically embedded in the model runtime. There are PRs #1357 #1374 being worked on that should generated the call to zdnn_init() automatically in run_main_graph (and other entry points).

imaihal:
@gongsu832 Thanks for great suggestion. It seems to be the cause. I will remove LD_PRELOAD after they are merged. Thanks!

imaihal:
@AlexandreEichenberger @etiotto @tungld Could you review this PR?

imaihal:

docs/AccelNNPAHowToUseAndTest.md
```diff
@@ -0,0 +1,29 @@
...
+Lit tests and numerical tests are provided for NNPA.
+
+- Lit tests
+
+When building onnx-mlir for NNPA, lit tests for NNPA also run with the command for CPU. The lit tests for NNPA are included in `test/mlir/accelerators/nnpa`.
+
+```
+cmake --build . --target check-onnx-lit
+```
+
```
AlexandreEichenberger, line 22: Can you confirm, and add a sentence to the extend that literal tests of the NNPA can be always performed on any platforms, regardless of whether NNPA is actually present or not. 

@gongsu832 can you remind us on how we can add this test on non-z16 machines? (just for the lit test). Thanks

docs/AccelNNPAHowToUseAndTest.md
```diff
@@ -0,0 +1,29 @@
+<!--- SPDX-License-Identifier: Apache-2.0 -->
```
AlexandreEichenberger, line 1: Big big thanks for adding doc on how to use NNPA. 

I would suggest that you add links to that page from the CONTRIBUTING (just add "handling and testing for specialized accelerator, for example) as well as near the top of /docs/Testing.md, for example "Testing of specific accelerators are described [here](AccelNNPAHowToUseAndTest.md)"

test/accelerators/NNPA/numerical/TestGRU-NNPA.cpp
```diff
@@ -0,0 +1,100 @@
+/*
```
AlexandreEichenberger, line 1: Could you explain in a top comment how different the TestGRU-NNPA.cpp file is from the regular TestGRU.cpp? It was not obviously clear to me what the differences are?

please do so for this and the other TestXXX-NNPA files in this dir

test/accelerators/NNPA/numerical/TestGRU-NNPA.cpp
```diff
@@ -0,0 +1,100 @@
...
+// naive implementation of GRU for a specific set of GRU
+// parameters/configuration.
+bool isOMGRUTheSameAsNaiveImplFor(const int direction, const int S, const int B,
+    const int I, const int H, const int linearBeforeReset,
+    bool isDynamicS = false, bool isDynamicB = false) {
+
+  GRULibBuilder gru(SHARED_LIB_BASE.str(), direction, S, B, I, H,
+      linearBeforeReset, isDynamicS, isDynamicB);
+  return gru.build() && gru.compileAndLoad() && gru.prepareInputs() &&
+         gru.run() && gru.verifyOutputs();
```
AlexandreEichenberger, line 30: Do we maybe need to add a RTOL (& ATOL) parameter to verifyOutputs() because I would expect NNPA errors to be larger than CPUs? If so, I would have a default value of the traditional RTOL and then a specific value. You would then need to change the `areCloseFloat` as well to take a RTOL and ATOL parameter

test/accelerators/NNPA/numerical/TestGRU-NNPA.cpp
```diff
@@ -0,0 +1,100 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
```
AlexandreEichenberger, line 4: please add an updated version of this
```c++
/=============-- LSTMModel.cpp - Building LSTM Models for tests -============//
//
// Copyright 2019-2022 The IBM Research Authors.
//
//
// This file contains a function that builds an LSTM model and compiles it.
//
``` 
for this and the other files in this dir.

docs/AccelNNPAHowToUseAndTest.md
imaihal: @AlexandreEichenberger Yes. The lit tests of the NNPA can run on non-z16 machines when onnx-mlir is built with  `-DONNX_MLIR_ACCELERATORS=NNPA`.

docs/AccelNNPAHowToUseAndTest.md
imaihal: Done. Added some sentences in `CONTRIBUTING.md` and `docs/Testing.md`

test/accelerators/NNPA/numerical/TestGRU-NNPA.cpp
imaihal: I added comments in TestLSTM-NNPA.cpp and TestGRU-NNPA.cpp. Following sentence is for TestGRU-NNPA.cpp.

```
// This file contains configurations for testing GRU models on NNPA.
// This file is based on test/numerical/TestGRU.cpp, but includes NNPA-specific
// configurations. For example, bidirectional GRU is not included because it is
// not supported in onnx-mlir. LinierBeforeReset is disabled because the zDNN
// library used in NNPA does not support it.
```

test/accelerators/NNPA/numerical/TestGRU-NNPA.cpp
imaihal: Yes. We need to use different RTOL and ATOL. They can be configured by using environment variables `TEST_ATOL` and `TEST_RTOL`. So, I set them in `test/accelerators/NNPA/numerical/CMakeLists.txt` as follows.

```
set_tests_properties(${test_name} PROPERTIES ENVIRONMENT "TEST_ATOL=0.01;TEST_RTOL=0.05};LD_PRELOAD=${ZDNN_LIB}")
```

test/accelerators/NNPA/numerical/TestGRU-NNPA.cpp
imaihal: Done. Thanks!

imaihal:
@AlexandreEichenberger I reflected your comments. Thanks!

docs/AccelNNPAHowToUseAndTest.md
```diff
@@ -0,0 +1,29 @@
...
+
+# Build and test for Accelerator NNPA
+
+## Build
+
+The following CMake option is required to build onnx-mlir for NNPA. Add this option in [build command](https://github.com/onnx/onnx-mlir/blob/main/docs/BuildOnLinuxOSX.md/#build)
```
tungld, line 7: `. Add this option in` => ` and this option is in`

docs/AccelNNPAHowToUseAndTest.md
tungld: NNPA is built on Z because we need Z to build libzdnn. NNPA lit tests is currently run on Jenkins s390x where some of the tests, e.g. compiled time stickifies, are related to big endian.

docs/AccelNNPAHowToUseAndTest.md
```diff
@@ -0,0 +1,29 @@
...
+
+The lit tests for NNPA are included in `test/mlir/accelerators/nnpa`. When building onnx-mlir for NNPA, these lit tests also run with the following same command with CPU.
+
+```
+cmake --build . --target check-onnx-lit
+```
+
+- Numerical tests
+
+Numerical tests for NNPA are provided in `test/accelerator/NNPA/numrical`. Currently tests for MatMul2D, Gemm, LSTM, and GRU are provided and run by using following command. For MatMul2D and Gemm, the same test code with CPU is used. For LSTM and GRU, different parameter configurations are provided for NNPA-specific tests. In these tests, appropriate ATOL and RTOL are used to pass the tests.
```
tungld, line 25: `numrical` => `numerical`

test/accelerators/NNPA/numerical/CMakeLists.txt
```diff
@@ -0,0 +1,80 @@
...
+    set_property(TARGET ${test_name} PROPERTY FOLDER "${test_suite_folder}")
+  endif ()
+
+  # Optimization level set by ONNX_MLIR_TEST_OPTLEVEL, defaults to 3
+  add_test(NAME ${test_name}
+    COMMAND ${test_name} -O${ONNX_MLIR_TEST_OPTLEVEL} --maccel=NNPA
+    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+    )
+  set_tests_properties(${test_name} PROPERTIES LABELS numerical-nnpa)
+  set_tests_properties(${test_name} PROPERTIES ENVIRONMENT "TEST_ATOL=0.01;TEST_RTOL=0.05};LD_PRELOAD=${ZDNN_LIB}")
```
tungld, line 43: Do we still need `LD_PRELOAD`?

test/accelerators/NNPA/numerical/CMakeLists.txt
```diff
@@ -0,0 +1,80 @@
...
+
+add_custom_target(check-onnx-numerical-nnpa
+  COMMENT "Running the ONNX-MLIR numerical regression tests for NNPA"
+  COMMAND "${CMAKE_CTEST_COMMAND}" -L numerical-nnpa --output-on-failure -C $<CONFIG> --force-new-ctest-process
+  USES_TERMINAL
+  DEPENDS numerical-nnpa
+  )
+set_target_properties(check-onnx-numerical-nnpa PROPERTIES FOLDER "Tests")
+# Exclude the target from the default VS build
+set_target_properties(check-onnx-numerical-nnpa PROPERTIES EXCLUDE_FROM_DEFAULT_BUILD ON)
```
tungld, line 19: Is it for Windows? I am not sure, just looking at the word `VS`.

tungld (commented):
Could we verify that a test is using NNPA not CPU? Without this verification, I am afraid that the test is still passed but potentially running on CPU.

docs/AccelNNPAHowToUseAndTest.md
imaihal: Done. Thanks!

docs/AccelNNPAHowToUseAndTest.md
imaihal: Done. Thanks!

test/accelerators/NNPA/numerical/CMakeLists.txt
AlexandreEichenberger: According to @cjvolzka , he said that his testing on LoZ indicated that preload were not needed anymore and that the new version of including the init within the evaluation of the graphs worked well. So I don't think its needed anymore

imaihal: I think so, too. However, just deleting `LD_PRELOAD` didn't work because zdnn_init was not found. I'm now investigating the reason.

gongsu832: It's difficult to tell what went wrong without seeing some kind of error output. For tests that are driven by ctest, it helps to pick a single test and run it with verbose on to get a better idea what's happening, for example,
```
ctest -R TestGemm-NNPA --verbose
```

test/accelerators/NNPA/numerical/TestGRU-NNPA.cpp
```diff
@@ -0,0 +1,114 @@
...
+  llvm::cl::ParseCommandLineOptions(
+      argc, argv, "TestGRU\n", nullptr, "TEST_ARGS");
+
+  // RapidCheck test case generation.
+  bool success = rc::check("GRU implementation correctness", []() {
+    // The number of directions.
+    // 1: forward, -1: reverse, 2: bidirectional
+    // const auto D = *rc::gen::element(1, -1, 2);
+    // bidirectional not supported now
+    const auto D = *rc::gen::element(1, -1, 2);
```
AlexandreEichenberger, line 68: If that was the only difference, maybe we could create a TEST_RNN_MODES=full or reduced, and use the same main as in the normal test but setting this env var to different value on the CmakeFiles.txt ?

test/accelerators/NNPA/numerical/CMakeLists.txt
imaihal: I narrowed down the situation. It seems that zdnn_init() is not found when all operations run on CPU. For example in a test for single LSTM op, when using peephole, zdnn_init() is not found.  (the LSTM op runs on CPU since zdnn does not support peephole). In this case, libzdnn.a is not linked in generated so file, `TestLSTM-NNPA_main_graph.so`. 

```
 symbol lookup error: ./TestLSTM-NNPA_main_graph.so: undefined symbol: zdnn_init
```

gongsu832: Is the main branch of your PR up to date? After commit `9cd5cbe7f1fd1c729c9392de13d1a027be51a47f`, functions in `libzdnn.a` should all be included in the generated .so if the .so is created with the `--maccel=NNPA` option. You can add the `-v` option to see the commands for how `TestLSTM-NNPA_main_graph.so` is generated.

gongsu832: I just tried your branch and I can see `zdnn_init` in the `TestLSTM-NNPA_main_graph.so`:
```
linux05:/tmp/onnx-mlir/build# nm ./test/accelerators/NNPA/numerical/TestLSTM-NNPA_main_graph.so|grep zdnn_init
0000000000013098 T zdnn_init
000000000000f930 T zdnn_init_pre_transformed_desc
0000000000017b70 T zdnn_init_ztensor
0000000000017c00 T zdnn_init_ztensor_with_malloc
```

imaihal: @gongsu832 Thanks for the test. Sorry, LSTM tests in current PR is no problem since all test configurations generate code for NNPA. It seems the error happens in the configurations which is not supported in zdnn LSTM. Following is such configuration. (`isNoneP` is 0)  In this configuration, we don't need to call zdnn_init, but I think it is ok to call with zdnn library. 

```
diff --git a/test/accelerators/NNPA/numerical/TestLSTM-NNPA.cpp b/test/accelerators/NNPA/numerical/TestLSTM-NNPA.c
pp
index b8bd4125..8dae1c4d 100644
--- a/test/accelerators/NNPA/numerical/TestLSTM-NNPA.cpp
+++ b/test/accelerators/NNPA/numerical/TestLSTM-NNPA.cpp
@@ -87,14 +87,14 @@ int main(int argc, char *argv[]) {
     const auto isNoneC = *rc::gen::element(0, 1);
     // Whether the weight tensor for peepholes(P) is specified.
     // Set none for peephole since zDNN does not support peephole
-    const auto isNoneP = 1;
+    const auto isNoneP = 0;
 
     RC_ASSERT(isOMLSTMTheSameAsNaiveImplFor(D, S, B, I, H, isDynS == 0,
         isDynB == 0, isNoneH == 1, isNoneC == 1, isNoneP == 1));
   });
   if (!success)
     return 1;
-
+  /*  Skip following tests for reducing test cases.
   // Exhaustive test case generation.
   // Set none for peephole
   for (int64_t s = 3; s < 4; s++)
@@ -115,6 +115,6 @@ int main(int argc, char *argv[]) {
                   // assert(isOMLSTMTheSameAsNaiveImplFor(
                   //    2, s, b, i, h, dyns, dynb, noneh, nonec, 1));
                 }
-
+  */
   return 0;
```

- I used `RC_PARAM=max_success=1` for reducing the tests.
```
diff --git a/test/accelerators/NNPA/numerical/CMakeLists.txt b/test/accelerators/NNPA/numerical/CMakeLists.txt
index a24d45d9..5c39eef3 100644
--- a/test/accelerators/NNPA/numerical/CMakeLists.txt
+++ b/test/accelerators/NNPA/numerical/CMakeLists.txt
@@ -36,11 +36,12 @@ function(add_numerical_unittest test_name)
 
   # Optimization level set by ONNX_MLIR_TEST_OPTLEVEL, defaults to 3
   add_test(NAME ${test_name}
-    COMMAND ${test_name} -O${ONNX_MLIR_TEST_OPTLEVEL} --maccel=NNPA --instrument-zhigh-ops=ALL --InstrumentBefore
ZHighOp --InstrumentAfterZHighOp --InstrumentReportTimeZHigh
+    COMMAND ${test_name} -O${ONNX_MLIR_TEST_OPTLEVEL} --maccel=NNPA --instrument-zhigh-ops=ALL --InstrumentBefore
ZHighOp --InstrumentAfterZHighOp --InstrumentReportTimeZHigh -v
     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
     )
   set_tests_properties(${test_name} PROPERTIES LABELS numerical-nnpa)
-  set_tests_properties(${test_name} PROPERTIES ENVIRONMENT "TEST_ATOL=0.01;TEST_RTOL=0.05")
+  set_tests_properties(${test_name} PROPERTIES ENVIRONMENT "TEST_ATOL=0.01;TEST_RTOL=0.05;RC_PARAMS=max_success=1")

```

- I saw following output 

```
1/1 Test #1: TestLSTM-NNPA ....................***Failed    0.36 sec
Model can reuse the current seed by exporting "TEST_SEED=147355436"
Target options: "--maccel=NNPA"
Using configuration: max_success=1 seed=13632536526712692457

- LSTM implementation correctness
[/home/imaihal/work/onnx-mlir/build/test/accelerators/NNPA/numerical/]/home/imaihal/work/llvm-project/build/bin/op
t: opt -O3 --mtriple=s390x-ibm-linux -o ./TestLSTM-NNPA_main_graph.bc ./TestLSTM-NNPA_main_graph.unoptimized.bc
[/home/imaihal/work/onnx-mlir/build/test/accelerators/NNPA/numerical/]/home/imaihal/work/llvm-project/build/bin/ll
c: llc -O3 --mtriple=s390x-ibm-linux -filetype=obj -relocation-model=pic -o ./TestLSTM-NNPA_main_graph.o ./TestLST
M-NNPA_main_graph.bc
[/home/imaihal/work/onnx-mlir/build/test/accelerators/NNPA/numerical/]/usr/bin/c++: c++ ./TestLSTM-NNPA_main_graph.o -o ./TestLSTM-NNPA_main_graph.so -shared -fPIC -L/home/imaihal/docker/imaihal-ubuntu/work/onnx-mlir/build/Debug/lib -lcruntime -lzdnn -lRuntimeNNPA
/home/imaihal/work/onnx-mlir/build/Debug/bin/TestLSTM-NNPA: symbol lookup error: ./TestLSTM-NNPA_main_graph.so: undefined symbol: zdnn_init
```

```
$ nm ./build/test/accelerators/NNPA/numerical/TestLSTM-NNPA_main_graph
.so|grep zdnn
                 U zdnn_init
                 U zdnn_is_version_runnable
```

gongsu832: @imaihal PR https://github.com/onnx/onnx-mlir/pull/1430 should fix this.

AlexandreEichenberger: @imaihal just landed, so update to the latest and you should have that fix

imaihal: @gongsu832 @AlexandreEichenberger Order! I confirmed it works well. Thanks!  I edited ` addCompilerConfig`, but I didn't change the order. Is there any specific reason why the order is affected?

gongsu832: This is how dynamic linking works. libRuntimeNNPA has functions that call `zdnn_init` so libRuntimeNNPA depends on libzdnn and the linkage order must be `-lRuntimeNNPA -lzdnn`.

If you have `-lzdnn -lRuntimeNNPA`, when the linker sees the calls to `zdnn_init` in libRuntimeNNPA, it doesn't have any libraries after it to resolve `zdnn_init` so it will mark it as "U" (Unresolved) as you saw above. If this were part of generating an executable, the linker would have failed and produced an unresolved symbol error immediately. However, in this case we are generating a .so (`TestLSTM-NNPA_main_graph.so`) so the linker tolerates unresolved symbols because you can still resolve them at runtime through LD_PRELOAD or LD_LIBRARY_PATH.

imaihal: I see. It seemed some configurations worked in the original order (`-lRuntimeNNPA -lzdnn`), but did they happen to work even in wrong order?

gongsu832: The original wrong order before the PR fix was `-lzdnn -lRuntimeNNPA`. Some tests still worked probably because some function in the model runtime might have called `zdnn_init`, i.e., the command to create the .so is something like
```
c++ model.o -o model.so -shared -fPIC -L/opt/Downloads/onnx-mlir/build/Release/lib -lzdnn -lRuntimeNNPA -lcruntime
```
So if some function in `model.o` calls `zdnn_init` then it would have been resolved by `-lzdnn` even if `-lRuntimeNNPA` is in the wrong order.

imaihal: I could clearly see. Thank you very much!

imaihal:
@tungld 

> Could we verify that a test is using NNPA not CPU? Without this verification, I am afraid that the test is still passed but potentially running on CPU.

I updated to print profile results of zHigh ops to confirm zHigh ops are really generated.   [check-onnx-numerical-nnpa.txt](https://github.com/onnx/onnx-mlir/files/8705874/check-onnx-numerical-nnpa.txt) Too much output, but I think we can check it. How do you think of this?

tungld:
> I think we can check it. How do you think of this?

Do you mean an automatic check? if so, yes, please go ahead. It could be an additional thing to check for each run.

imaihal:
> Do you mean an automatic check? if so, yes, please go ahead. It could be an additional thing to check for each run.

Oh, this is not automatic. Is your request to add an automatic check here? I don't know a good way to do it, but one way might be to raise assert when `isSuitableForZDNN` in the legality check returns false, assuming all onnx ops need to be lowered to zHigh ops in numerical tests.

tungld:
> Oh, this is not automatic. Is your request to add an automatic check here? I don't know a good way to do it, but one way might be to raise assert when isSuitableForZDNN in the legality check returns false, assuming all onnx ops need to be lowered to zHigh ops in numerical tests.

A naive way is, for example, to test LSTM op, we expect that `zdnn_lstm` must be called. So checking with `nm` command to see whether we have `zdnn_lstm` in the `.so` file or not. So for each op, we specify the expected zdnn functions to call.

tungld:
> one way might be to raise assert when isSuitableForZDNN in the legality check returns false, assuming all onnx ops need to be lowered to zHigh ops in numerical tests.

This may not be true since it can only capture onnx-to-zhigh, but in general we can lower any op at different dialects to zdnn, e.g. krnl.matmul to zdnn.

imaihal:
> A naive way is, for example, to test LSTM op, we expect that zdnn_lstm must be called. So checking with nm command to see whether we have zdnn_lstm in the .so file or not. So for each op, we specify the expected zdnn functions to call.

I see. I will try this. Thanks!

test/accelerators/NNPA/numerical/TestGRU-NNPA.cpp
AlexandreEichenberger: @imaihal let me know when the changes are in, thanks

imaihal: @AlexandreEichenberger Done. I inserted some ifdef directives in TestLSTM, TestGemm, TestMatMul, and TestGRU, and they are set in `test/accelerators/NNPA/numerical/CMakeLists.txt` So, now only CMakeLists.txt is in the directory.

imaihal:
@tungld 

> A naive way is, for example, to test LSTM op, we expect that zdnn_lstm must be called. So checking with nm command to see whether we have zdnn_lstm in the .so file or not. So for each op, we specify the expected zdnn functions to call.

I created ModelLibBuilder::checkSharedLibInstruction() to verify generated library.

test/modellib/ModelLib.cpp
```diff
@@ -120,6 +121,7 @@ bool ModelLibBuilder::checkSharedLibInstruction(
   }
   return true;
 }
+#endif
```
gongsu832, line 124: Checking whether a symbol exists in a shared library can be done more easily with `dlopen` and `dlsym`.

test/modellib/ModelLib.cpp
```diff
@@ -99,6 +102,28 @@ void ModelLibBuilder::setRandomNumberGeneratorSeed(const std::string &envVar) {
...
+  handle = dlopen(sharedLibName.c_str(), RTLD_LAZY);
+  if (handle == NULL) {
+    printf("%s\n", dlerror());
+    return false;
+  }
+  int *dptr;
+  dptr = (int *)dlsym(handle, instructionName.c_str());
+  if (dptr == NULL) {
+    printf("%s\n", dlerror());
+    return false;
```
gongsu832, line 120: Should call dlclose before returning.

imaihal: @gongsu832 Thanks. I remove ifdef _WIN32 in https://github.com/onnx/onnx-mlir/pull/1335/commits/e20853f1e8c86cab47539264e4004990df373f0c, but it seems it didn't work for WIN32. Is something missing?

I also test another way for WIN32, but it still fails.

gongsu832: It's failing to compile `ModelLib.cpp`
https://dev.azure.com/onnx-pipelines/d175d065-ea6c-4983-99d3-ae963744886b/_apis/build/builds/24898/logs/14

imaihal:
Could you review again? 

@AlexandreEichenberger The test for NNPA uses the same code with CPU (test/modellib and test/numerial) by using ifdef directive in CMakeLists.txt(test/accelerator/NNPA/numerical/CMakeLists.txt).

@tungld I created `checkSharedLibInstruction()` to verify instructions in shared library. We can verify that zDNN instructions are included automatically.

test/accelerators/NNPA/numerical/CMakeLists.txt
imaihal: I removed the line. Thanks!

test/modellib/ModelLib.cpp
imaihal: Done. Thanks!

imaihal:

test/numerical/TestGRU.cpp
```diff
@@ -24,10 +24,28 @@ bool isOMGRUTheSameAsNaiveImplFor(const int direction, const int S, const int B,
...
+    return false;
+  }
+  // Verify generated library
+  std::string instructionName =
+      getenv("TEST_CHECK_INSTRUCTION") ? getenv("TEST_CHECK_INSTRUCTION") : "";
+  std::string sharedLibName =
+      ModelLibBuilder::getSharedLibName(SHARED_LIB_BASE.str());
+  if (!ModelLibBuilder::checkSharedLibInstruction(
+          instructionName, sharedLibName))
+    return false;
```
AlexandreEichenberger, line 46: Can you refactor this code to be much more concise.

```
return gru.build() && gru.compileAndLoad() &&  gru.checkInstruction("TEST_CHECK_INSTRUCTION") && 
          gru.prepareInputs() && gru.run() && gru.verifyOutputs();
```

then you can create a `checkInstruction` that read the env var, if empty return true, and if not empty return ` checkSharedLibInstruction( instructionName, sharedLibName))`. Note that the lib name is already stored in the parent class, which you see when you created `gru` object, so no need to pass it.

If you want to do 2 versions from checkInstruction, one that that the instruction, and another named `checkInstructionFromEnv` and that simply check for the env var, and if undefined return true, otherwise call the `checkInstruction` where the actual instruction string is passed, that would be even cleaner.

test/modellib/ModelLib.cpp
AlexandreEichenberger: If you look at the code in ExecutionSession builder, it seems that this can be entirely avoided using the llvm code
```
  _sharedLibraryHandle =
      llvm::sys::DynamicLibrary::getPermanentLibrary(sharedLibPath.c_str());
  if (!_sharedLibraryHandle.isValid())
    throw std::runtime_error(reportLibraryOpeningError(sharedLibPath));
```

There is no special casing whatever for Window or Linux... So it seems a more robust solution as they already deal with it in LLVM.

test/numerical/TestGRU.cpp
```diff
@@ -78,22 +104,32 @@ int main(int argc, char *argv[]) {
     for (int64_t b = 3; b < 4; b++)
       for (int64_t i = 2; i < 5; i++)
         for (int64_t h = 2; h < 5; h++)
+        // clang-format off
+#ifdef TEST_GRU_L1
+          {
+            int64_t l = 1;
+#else
           for (int64_t l = 0; l < 2; l++) {
+#endif
```
AlexandreEichenberger, line 113: Please keep the loop in all cases, but define this before
```
#ifdef TEST_GRU_1
  int l_min = 1
#else 
  int l_min = 0
#endif 
```
and then for (l=l_min, l<2l ++l)
```

File: test/numerical/TestGemm.cpp [hunk 15]
```diff
@@ -61,8 +61,21 @@ static bool isOMGemmTheSameAsNaiveImplFor(const int I, const int J, const int K,
...
+  // Verify generated library
+  std::string instructionName =
+      getenv("TEST_CHECK_INSTRUCTION") ? getenv("TEST_CHECK_INSTRUCTION") : "";
+  std::string sharedLibName =
+      ModelLibBuilder::getSharedLibName(SHARED_LIB_BASE.str());
+  if (!ModelLibBuilder::checkSharedLibInstruction(
+          instructionName, sharedLibName))
+    return false;
+  return successBuild && gemm.prepareInputs() && gemm.run() &&
+         gemm.verifyOutputs();
```
AlexandreEichenberger, line 78: use same interface as described above

test/numerical/TestLSTM.cpp
```diff
@@ -25,10 +25,28 @@ bool isOMLSTMTheSameAsNaiveImplFor(const int direction, const int S,
...
+  // Verify generated library
+  std::string instructionName =
+      getenv("TEST_CHECK_INSTRUCTION") ? getenv("TEST_CHECK_INSTRUCTION") : "";
+  std::string sharedLibName =
+      ModelLibBuilder::getSharedLibName(SHARED_LIB_BASE.str());
+  if (!ModelLibBuilder::checkSharedLibInstruction(
+          instructionName, sharedLibName))
+    return false;
+  return successBuild && lstm.prepareInputs() && lstm.run() &&
+         lstm.verifyOutputs();
```
AlexandreEichenberger, line 49: use same interface as above

test/numerical/TestLSTM.cpp
```diff
@@ -87,17 +113,25 @@ int main(int argc, char *argv[]) {
             for (int64_t dynb = 0; dynb < 2; dynb++)
               for (int64_t noneh = 0; noneh < 2; noneh++)
                 for (int64_t nonec = 0; nonec < 2; nonec++)
+                // clang-format off
+#ifdef TEST_NONEP_ONLY
+                  {
+                    int64_t nonep = 1;
+#else
                   for (int64_t nonep = 0; nonep < 2; nonep++) {
+#endif
```
AlexandreEichenberger, line 122: use the loop in all case, but define outside of the loops the min value to be 0 or 1

test/numerical/TestMatMul2D.cpp
```diff
@@ -23,8 +23,21 @@ static bool isOMMatmulTheSameAsNaiveImplFor(
...
+  // Verify generated library
+  std::string instructionName =
+      getenv("TEST_CHECK_INSTRUCTION") ? getenv("TEST_CHECK_INSTRUCTION") : "";
+  std::string sharedLibName =
+      ModelLibBuilder::getSharedLibName(SHARED_LIB_BASE.str());
+  if (!ModelLibBuilder::checkSharedLibInstruction(
+          instructionName, sharedLibName))
+    return false;
+  return successBuild && matmul.prepareInputs() && matmul.run() &&
+         matmul.verifyOutputs();
```
AlexandreEichenberger, line 40: same here

AlexandreEichenberger (commented):
Its going in the right directions, and the new infrastructure about testing for the ops is very helpful. Thanks

test/numerical/TestGRU.cpp
imaihal: Done. Thanks!

test/numerical/TestLSTM.cpp
imaihal: Done. Thanks!

test/numerical/TestGemm.cpp
imaihal: Thanks for your comments. Refactoring done.

test/numerical/TestMatMul2D.cpp
imaihal: Thanks for your comments. Refactoring done.

AlexandreEichenberger:
@imaihal let me know when you have all your changes in, and I will review again.

imaihal:
@AlexandreEichenberger Yes. Now ready for review again. Thanks!

test/numerical/TestGemm.cpp
```diff
@@ -92,10 +93,15 @@ int main(int argc, char *argv[]) {
       const auto aTrans = *rc::gen::inRange(0, 2);
       const auto bTrans = *rc::gen::inRange(0, 2);
       const auto cRank = *rc::gen::inRange(1, 3);
+#ifdef TEST_ALPHA_BETA_1
```
tungld, line 96: Suggest to use a more specific name, say `TEST_GEMM_ALPHA_BETA_1`.

test/numerical/TestLSTM.cpp
```diff
@@ -70,14 +80,23 @@ int main(int argc, char *argv[]) {
     // Whether initial value of the cell(initial_c) is specified.
     const auto isNoneC = *rc::gen::element(0, 1);
     // Whether the weight tensor for peepholes(P) is specified.
+#ifdef TEST_NONEP_ONLY
```
tungld, line 83: Suggest to use `TEST_LSTM_NO_P` or `TEST_LSTM_NO_PEEPHOLE` to make it consistent with the other environment variables.

test/accelerators/NNPA/numerical/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+#   ${ONNX_NUMERICALTEST_SRC_DIR}/TestConv.cpp
+#   LINK_LIBS PRIVATE ${TEST_LINK_LIBS}
+#   )
+
+add_numerical_unittest(TestMatMul2D-NNPA zdnn_matmul_op
+  ${ONNX_NUMERICALTEST_SRC_DIR}/TestMatMul2D.cpp
+  LINK_LIBS PRIVATE ${TEST_LINK_LIBS}
+  )
+
+add_numerical_unittest(TestGemm-NNPA zdnn_matmul_op
```
tungld, line 65: Could we check multiple zdnn functions here? For example, some GEMM ops will use both` zdnn_matmul_op` and `zdnn_add_op`. It's ok to implement this with another PR.

tungld (commented):
> @tungld I created checkSharedLibInstruction() to verify instructions in shared library. We can verify that zDNN instructions are included automatically by using the function.

Thanks so much! It's useful.

The current code looks good to me with some minor changes to make things consistent. Please wait for Alex's input too.

In the future, we can improve the code with two things:
1. `checkInstructionFromEnv` should support multiple instructions since one ONNX op may call multiple zDNN ops.
2. `gemm.compileAndLoad() && gemm.checkInstructionFromEnv()`. `compileAndLoad` has loaded the shared library, no need to load it again in `checkInstructionFromEnv`. To do that, we should edit ExecutionSession so that we can get the loaded library from it.

test/modellib/ModelLib.cpp
```diff
@@ -54,14 +54,41 @@ bool ModelLibBuilder::compileAndLoad(
   return compileAndLoad();
 }
 
+bool ModelLibBuilder::checkInstructionFromEnv() {
+  std::string instructionName =
+      getenv("TEST_CHECK_INSTRUCTION") ? getenv("TEST_CHECK_INSTRUCTION") : "";
+  if (instructionName.empty())
+    return true;
```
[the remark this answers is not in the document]
AlexandreEichenberger: Please pass the env var name as a parameter. Its easier to read all the param in the specific test file, e.g. TestMatMul2D....

test/modellib/ModelLib.cpp
```diff
@@ -54,14 +54,41 @@ bool ModelLibBuilder::compileAndLoad(
...
+  return checkInstruction(instructionName);
+}
+
+bool ModelLibBuilder::checkInstruction(const std::string instructionName) {
+  if (instructionName.empty())
+    return true;
+  std::string sharedLibName = getSharedLibName(sharedLibBaseName);
+  llvm::sys::DynamicLibrary sharedLibraryHandle =
```
AlexandreEichenberger, line 69: Please extract the library from the ExecutionSession model variable `onnx_mlir::ExecutionSession *exec;`

test/numerical/TestConv.cpp
```diff
@@ -40,8 +40,9 @@ bool isOMConvTheSameAsNaiveImplFor(const int N, const int C, const int H,
 
   Conv2DLibBuilder conv(SHARED_LIB_BASE.str(), N, C, H, W, kH, kW, autoPad,
       pHBegin, pHEnd, pWBegin, pWEnd, stride, dilation, isDynamic);
-  return conv.build() && conv.compileAndLoad() && conv.prepareInputs() &&
-         conv.run() && conv.verifyOutputs();
+  return conv.build() && conv.compileAndLoad() &&
```
AlexandreEichenberger, line 43: It looks SOOO much better, thanks for the changes.

AlexandreEichenberger (approved):
Almost there... please implement the changes suggested by Tung and I. The added functionality of handing a list of instructions to check should probably let for a subsequent PR.

I will approve it now so that you may merge it once the above changes are made. Big thanks.

test/modellib/ModelLib.cpp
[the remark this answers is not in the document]
imaihal: Done. Thanks!

test/numerical/TestGemm.cpp
imaihal: Done. Thanks!

test/numerical/TestLSTM.cpp
imaihal: Done. Thanks!

test/accelerators/NNPA/numerical/CMakeLists.txt
imaihal: Thanks for your comments! 
It is possible to check multiple functions, but currently it is not easy to select one function from those multiple functions in each testcase.
For example in GEMM, `zdnn_matmul_op` and `zdnn_add_op` can be checked in all testcases of GEMM, but it is not possible to select only `zdnn_matmul_op` in some testcases and `zdnn_matmul_op` and `zdnn_add_op` in the other testcases. I think we need another config for each testcase.
So, I would like to set this TODO for now.

test/modellib/ModelLib.cpp
imaihal: Done. Thanks!

PR #1336, opened 2022-04-12
Diagnose axis attribute for Split, SplitToSequence, TopK, Unique

etiotto:
Diagnose axis constraints for Split, SplitToSequence, TopK, Unique. This concludes the diagnostic for all ONNX operators that have an `axis` attribute.


src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2940,6 +2940,26 @@ LogicalResult ONNXGRUOp::inferShapes(
 // Split
 //===----------------------------------------------------------------------===//
 
+LogicalResult ONNXSplitOp::verify() {
+  ONNXSplitOpAdaptor operandAdaptor(*this);
+  for (const auto &operand : operandAdaptor.getOperands())
+    if (!hasShapeAndRank(operand))
+      return success(); // Won't be able to do any checking at this stage.
+
+  auto inputType = operandAdaptor.input().getType().cast<ShapedType>();
```
AlexandreEichenberger, line 2949: nit use of auto

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2940,6 +2940,26 @@ LogicalResult ONNXGRUOp::inferShapes(
...
+    return onnx_mlir::Diagnostic::attributeOutOfRange(*this->getOperation(),
+        "axis", axisIndex,
+        onnx_mlir::Diagnostic::Range<int64_t>(-inputRank, inputRank - 1));
+
+  return success();
+}
+
 LogicalResult ONNXSplitOp::inferShapes(
     std::function<void(mlir::Region &)> doShapeInference) {
   if (!getOperand(0).getType().cast<RankedTensorType>())
```
AlexandreEichenberger, line 2965: do we want to systematically use "has shape and rank"

src/Support/Diagnostic.cpp
```diff
@@ -30,7 +32,24 @@ mlir::LogicalResult Diagnostic::attributeOutOfRange(mlir::Operation &op,
                                     .concat(", ")
                                     .concat(std::to_string(validRange.max))
                                     .concat("]"));
-};
+}
+
+LogicalResult Diagnostic::operandHasUnexpectedRank(Operation &op,
```
AlexandreEichenberger, line 37: I would also change the name to `emitErrorOperandHasUnexpectedRank`

etiotto: yes will do

src/Dialect/ONNX/ONNXOps.cpp
etiotto: yes I think that is cleaner, started to do that in other PRs. Will do here as well

src/Dialect/ONNX/ONNXOps.cpp
etiotto: Right, in this case the type is clear because the RHS has a cast, so using auto if OK

AlexandreEichenberger (approved):
LGTM

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -4669,6 +4718,41 @@ LogicalResult ONNXTopKOp::inferShapes(
...
+  }
+
+  if (optionalSorted.hasValue()) {
+    // optional sorted attribute must be zero or one.
+    int64_t sorted = optionalSorted.getValue();
+    if (sorted < 0 || sorted > 1)
+      return onnx_mlir::Diagnostic::emitAttributeOutOfRangeError(
+          *this->getOperation(), "sorted", sorted,
+          onnx_mlir::Diagnostic::Range<int64_t>(0, 1));
+  }
```
tungld, line 4812: `optionalSorted` does not depend on rank. Verifying this is doable even though `hasShapeAndRank(X) = false`

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -4632,35 +4675,41 @@ LogicalResult ONNXThresholdedReluOp::inferShapes(
...
+    return onnx_mlir::Diagnostic::emitAttributeOutOfRangeError(
+        *this->getOperation(), "axis", axis,
+        onnx_mlir::Diagnostic::Range<int64_t>(-Xrank, Xrank - 1));
+
+  // K's rank must be zero or one.
+  Value K = operandAdaptor.K();
+  int64_t KRank = K.getType().cast<ShapedType>().getRank();
+  if (KRank > 1)
+    return onnx_mlir::Diagnostic::emitOperandHasUnexpectedRankError(
+        *this->getOperation(), K, KRank, "< 2");
```
tungld, line 4764: K's rank does not depend on X's rank. Verifying this is doable without checking X's rank at the beginning of the function.

tungld (changes requested):
It looks like requiring that all operands have to be ranked before verification is too strict.  Some attributes do not require all operands to be ranked for verification.

src/Dialect/ONNX/ONNXOps.cpp
etiotto: moved code 4805-4812 at the start if the function.

src/Dialect/ONNX/ONNXOps.cpp
etiotto: moved lines 4759-4764 above the check for axis.

etiotto:
@tungld take another look, I addressed your comments.

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -4746,23 +4744,23 @@ LogicalResult ONNXTopKOp::verify() {
           [](const Value &op) { return !hasShapeAndRank(op); }))
     return success(); // Won't be able to do any checking at this stage.
 
+  // K's rank must be zero or one.
```
tungld, line 4746: I think we still cannot come to this point if X is unranked because of `llvm::any_of`. I suggest to check K's rank and X's rank separately when needed instead of `llvm::any_of`.

tungld (changes requested):

src/Dialect/ONNX/ONNXOps.cpp
etiotto: Ok, modified this member function implementation as you suggest. Please take another look.

PR #1338, opened 2022-04-12
Don't include CompilerOptions in pass libraries

manbearian:
My team links the onnx-mlir libraries into our own tool and we'd prefer not to pick up the onnx-mlir command-line options en-masse. Recent changes to refactoring the onnx-mlir options require additional changes to continue to allow this to work which i'm committing here.


sstamenova:

manbearian:
Looks like my change doesn't work for shared-library build. I'll think on another way to address this

PR #1343, opened 2022-04-14
Fix legality check of ONNXToZHigh for MaxPool.

imaihal:
I found that ONNXToZHIgh lowering fails in 1D MaxPool. This PR fixes it.

In the 1D case, `shapeHelper.dilations[1]` does not exist, but it was checked in legality check. Newly added lit test failed before.


src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
```diff
@@ -682,7 +682,7 @@ bool isSuitableForZDNN<ONNXMaxPoolSingleOutOp>(ONNXMaxPoolSingleOutOp op) {
          "Failed to scan ONNXMaxPoolSingleOutOp parameters successfully");
 
   // dilations not supported. Only default one is accepted.
-  if (shapeHelper.dilations[0] != 1 || shapeHelper.dilations[1] != 1)
+  if (llvm::any_of(shapeHelper.dilations, [](int64_t val) { return val != 1; }))
```
tungld, line 685: zDNN pooling ops support 4D tensors only, so it always has dilations[1]. You just need to call `checkLegalityPoolOpsCommon` (which checks tensor rank, etc. ) before this check to fix the issue.

tungld (changes requested):

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp

test/mlir/accelerators/nnpa/conversion/onnx-to-zhigh/pool.mlir
```diff
@@ -101,6 +101,19 @@ func @averagepool_should_lower_to_zhigh_same_upper(%arg0: tensor<1x3x32x32xf32>)
 }
 
 // -----
+// CHECK-LABEL:  func @test_pool_not_lowered_1d
```
tungld, line 104: Should the function name be changed to reflect what it is supposed to check?

imaihal: I used this function name since there is a test for 3d (`test_pool_not_lowered_3d`), but not for 1d. And if there is this test for 1d, I thought we could have avoided this bug. 
You mean we should use more specific name for this PR such as `test_pass_pool_1d_dilation` or something?

tungld: `test_pass_pool_1d_dilation` would be better, but again the ONNX op here was not lowered to ZHigh was not because of dilation I think. To test your code about dilation, you could have a ONNX op that satisfies other conditions but not dilation.

Anyway, `test_pool_not_lowered_3d` is not good too. I don't what `3d` is for. If you know it, please correct it too. Thanks!

imaihal: `1d` and `3d` should come from maxpool1d and maxpool3d which are not supported in zDNN. So, I think they are not lowered because input shape is not 4d.

imaihal: @tungld Update the name. Thanks!

imaihal: Also simplify the function.

test/mlir/accelerators/nnpa/conversion/onnx-to-zhigh/pool.mlir
```diff
@@ -101,12 +101,26 @@ func @averagepool_should_lower_to_zhigh_same_upper(%arg0: tensor<1x3x32x32xf32>)
...
-  %1 = "onnx.AveragePool"(%arg0) {auto_pad = "NOTSET", kernel_shape = [2, 2, 2], pads = [0, 0, 0, 0, 0, 0], strides = [1, 1, 1]} : (tensor<1x3x32x32x32xf32>) -> tensor<1x3x31x31x31xf32>
+  %1 = "onnx.AveragePool"(%arg0) {kernel_shape = [2]} : (tensor<1x3x32xf32>) -> tensor<1x3x31xf32>
+  // CHECK: "onnx.AveragePool"
+
+  return %0, %1 : tensor<1x3x31xf32>, tensor<1x3x31xf32>
+}
+
+// -----
+// COM: Pooling in zDNN only support 4D input (MaxPool1D and MaxPool3D not suported)
+// CHECK-LABEL:  func @test_pool_not_lowered_input5d
```
tungld, line 118: The above line said `MaxPool1D and MaxPool3D`, and you had `test_pool_not_lowered_input3d` for 3D, so this should be `test_pool_not_lowered_input1d` instead of `test_pool_not_lowered_input5d`? It seems they are inconsistent.

imaihal: Sorry for confusing. Legality check for MaxPool checks if dimension size of input is 4 since input for MaxPool2D need to have 4d (NCHW). Input is 3d in MaxPool1D (NCL), and 5d in MaxPool3D(NCDHW) .    
So, I named the functions as follows.
- test_pool_not_lowered_input3d:  ==> not lowered when input tensor is 3d (MaxPool1D case)
- test_pool_not_lowered_input5d:  ==> not lowered when input tensor is 5d (MaxPool3D case)
- Only lowered when input tensor is 4d (MaxPool2D case)

However, simply `test_pool_not_lowered_maxpool1d` and `test_pool_not_lowered_maxpool3d` might be better.

tungld: OK. I see. Thanks!

tungld (approved):
LGTM with pending a change to the lit test name.

Issue #1344, opened 2022-04-14
Error caused by PR1326


chentong319:
@whitneywhtsang 
I got error for models with LoopOp recently. I suspect that the new changes to use new interface is the culprit
[loop-test.txt](https://github.com/onnx/onnx-mlir/files/8490759/loop-test.txt)
I attached a simple test case here. It should be named loop-test.mlir.
I tried this several days ago. I remember I got error complaining something like "env for IndexExpr not defined". At that time, I replaced it with an old version of Loop.cpp and it worked.
Today I got the error 
```
Assertion failed: (originalLoops.size() == ubs.size() && "expected same rank"), function iterateIE, file /Users/tongchen/Projects/onnx-mlir/src/Dialect/Krnl/DialectBuilder.cpp, line 98.
Abort trap: 6
```
Same error are seen on the important medical models, which was compiled correctly after my krnl.region PR#1304.

AlexandreEichenberger:
@chentong319 Is the test for this loop case not yet in the current CIs?

whitneywhtsang:
@chentong319 Thanks for reporting, I will find the PR that causes the failure and revert it first.

PR #1350, opened 2022-04-14
[Loop] Fix "expected same rank" assertion

whitneywhtsang:
Fixed assertion failure reported in https://github.com/onnx/onnx-mlir/issues/1344.


AlexandreEichenberger (approved):
LGTM, thanks for adding the test that ended up showing the issue.

chentong319 (commented):
LGTM!

PR #1357, opened 2022-04-15
Accel init shutdown

AlexandreEichenberger:
Current approach add code int he OMRuntime to operate the implicit Init and shutdown in a thread-safe manner.
Missing to the code is filling in the actual code for the init and shutdown (zdnn_init() and {}, for the NNPA).


include/OnnxMlirRuntime.h
```diff
@@ -22,6 +22,7 @@
 #include <stdint.h>
 #endif
 
+#include <onnx-mlir/Runtime/OMAccelerator.h>
```
gongsu832, line 25: Sorry I'm not understanding why we are exposing these calls to the user?

AlexandreEichenberger: They are part of the runtime as they are called by our code. Why exposed? If some system has time to initialized everything ahead of time and then want a very fast response time, then init is good. If some accel reserve a lot of resources once it is initialized, a user might want to free resources by uninitializing the device.

gongsu832: Being part of runtime doesn't mean they have to be exposed to the users. We have lots of functions in the runtime that are not exposed to the users.

The time you save on the init call is all speculation. A typical inference call may take on the order of milliseconds. Do we expect the init call to take milliseconds? I doubt it. And the saving is only for the first call. As for freeing the resources, that should also be managed internally when the last thread exits the entry point like `run_main_graph`, etc.

The danger of exposing calls to users, particularly ones that are potentially unstable (due to the possibility of adding new accelerators), is that it will make changing the calls very difficult, e.g., if we find out later that the init/fini calls need to take a parameter for some accelerators. We don't add user visible API calls unless we have a high confidence that users must be involved. In this case, I'm not convinced that the users must be involved in the init/fini calls.

AlexandreEichenberger: Ok, will remove then the public interface for now; we can always change our minds later if we have a concrete need for it.

src/Accelerators/NNPA/Runtime/OMRuntimeNNPA.c
```diff
@@ -0,0 +1,74 @@
...
+      /* Actual init. */
+      zdnn_init();
+      /* Release inner mutex, and then set accelerator to initialized. */
+      pthread_mutex_unlock(&OMInnerMutexForInitShutdownNNPA);
+      OMIsInitAccelNNPA = 1;
+    } /* Release outer mutex. */
+    pthread_mutex_unlock(&OMOuterMutexForInitShutdownNNPA);
+  }
+}
+
```
gongsu832, line 56: A couple of problems with this code:

- not sure what's the purpose of using two locks as one lock would be enough to do the job
- you are setting `OMIsInitAccelNNPA` after calling `zdnn_init()`, which means it can be called multiple times, this may or may not work
- in general, it's typically assumed that init/fini calls can be called only once. Therefore, instead of a 1/0 switch, you need a reference count to handle multiple threads. This is a fairly common code pattern, something like the following,
```
init() {
    pthread_mutex_lock();
    if (refcnt == 0) {
        /* do your setup */
    }
    refcnt++;
    pthread_mutex_unlock();
}

fini() {
    pthread_mutex_lock();
    /* It's a bug that fini is called more times than init */
    assert(refcnt > 0);
    if (refcnt == 1) {
        /* do your teardown */
    }
    refcnt--;
    pthread_mutex_unlock();
}
```

AlexandreEichenberger: @gongsu832 the code is correct and uses 2 locks as an abundance of caution. Here is the scenario that the second lock is protecting:
1. I need to test again within the lock if the device has not been init (the refcnt++ in your example, for me setting the `OMIsInitAccelNNPA=1`.
2. Without a fence between the `zdnn_setup()` and the  `OMIsInitAccelNNPA=1`, it would be possible for another thread to see `OMIsInitAccelNNPA=1` before all of the stores associated with the setup are seen by that other thread (release consistency).
3. While on machines like POWER and Z, a memory fence would be enough, there is no posix memory fence that I know of. However, a mutex is guaranteed to have a fence inside the unlock(), thus my use of an inner lock.

Your code snipet is also not valid for the targeted use. I need to be able to read `OMIsInitAccelNNPA` outside of a lock, because grabbing a lock to just see that the device is initialized would be expensive. Instead, I set `OMIsInitAccelNNPA=1` once I can guarantee that every other threads witnessing `OMIsInitAccelNNPA==1` guarantee that they will see the initialized state. Thus, the code inserted inside the `run_main_graph` and equivalent entry points can simply be
```C++
if (!OMIsInitAccelNNPA) OMInitAccelNNPA()
```

There is no need to maintain a ref count, because init can be done only once, and we do not expect a fini after any init. The shutdown is currently not needed, unless we have to do something for migration of threads to another machine.

I know multithreading locks are subtle, I am fairly sure I have this one down. Always welcoming questions as we need to collectively agree that we have the right solution.

gongsu832: > Your code snipet is also not valid for the targeted use

How so? It calls init and fini exactly once regardless of the number of threads. Your code, on the other hand, may call init multiple times. Here is how,

- thread 1 and 2 both check `OMIsInitAccelNNPA` to be 0 so both proceed to lock outer mutex, assume thread 1 gets it so thread 2 blocks
- thread 1 goes through the whole process of locking inner mutex, calll zdnn_init, unlock inner mutex, set `OMIsInitAccelNNPA` to 1, unlock outer mutex
- now thread 2 proceeds and does the whole thing again

> because grabbing a lock to just see that the device is initialized would be expensive

This not true. pthread mutex is completely userspace code typically just a few instructions.

> There is no need to maintain a ref count, because init can be done only once, and we do not expect a fini after any init.

Yes there is. The reference count is for making sure whatever shutdown code is called only once. While zdnn doesn't have a shutdown, I'm assuming in general there might be a shutdown for some accelerator. And only the last thread exiting from run_main_graph (or whatever other entry point that has called init) can call shutdown. Your shutdown code, just like your init code, is flawed and does not guarantee the shutdown is called only once.


> I know multithreading locks are subtle, I am fairly sure I have this one down. Always welcoming questions as we need to collectively agree that we have the right solution.

Which is exactly the reason this type of code tries to be as simple as possible, with as few locks as possible.

AlexandreEichenberger: @gongsu832 

There is no need to call a fini for each init. We only need to call init once (replacing the PRELOAD functionality). Calling fini is just pure overhead. And when the count reach zero, there is still nothing to do. The alternative code pattern is perfectly valid when one need to dynamically track the number of active calls. This is just not needed in our use case.

In all but the first (or a few times, if multiple threads are concurrently seen the device as not being initialized), the only test being done is "read, compare to zero, branch around the function call". Regardless of how inexpensive a mutex is, it will be more than that. 

> The reference count is for making sure whatever shutdown code is called only once. 

There is no shutdown code on zdnn. I confirmed with the authors of the zdnn library. The only use case would be in case of live migration to a different node, or a case where they mentioned some custom env variable were to change. They requested that we don't worry about this for now. 

If you prefer, I can remove the shutdown code at this time. I figured that I could leave it because it might be needed in the future. But if it is confusing, I can remove it. Let me know.

> Which is exactly the reason this type of code tries to be as simple as possible, with as few locks as possible.

Low overhead, no mutex that may serialize parallel threads calling distinct `run_main_graph` in parallel.

gongsu832: In typical case, threads are all spawned together so you code will be basically all blocking on the outer mutex. And since nobody can proceed until you release the outer mutex, I fail to see the use of the inner mutex. Perhaps you can give an example of when the inner mutex is useful.

I'm not talking about zdnn. On the contrary, I'm assuming an accelerator in general may need an init and an fini, each to be called exactly once. How would you do it?

AlexandreEichenberger: > In typical case, threads are all spawned together so you code will be basically all blocking on the outer mutex.

correct, there is no way to avoid this, and the code section above is no different. Once the flag is set, though, no threads will ever attempt to grab the mutex.

>  I fail to see the use of the inner mutex. 

Correct, the inner mutex will always succeed. The operation I need is the memory fence in the `pthread_mutex_unlock`. Without a memory fence between the stores in the `zdnn_init()` and the `OMIsInitAccelNNPA=1`, another thread may see the store associated with `OMIsInitAccelNNPA=1` before a store inside the init, and thus see the initialization in an inconsistent state. The memory fence basically enforce the following: "if any thread witness a memory state set AFTER the fence, it is guaranteed to see any of the memory state set before the fence". So it does not guarantee any other threads to see `OMIsInitAccelNNPA==1` (thus forcing them possibly in one more pointless mutex), but if they see `OMIsInitAccelNNPA==1` then they are guaranteed to see all of the stores performed in the init.

I could find out what the men fence intrinsic is for Z machine (not sure if zOS and LoZ use the same...) but out of simplicity, I just decided to use a spurious lock to enforce this.

> On the contrary, I'm assuming an accelerator in general may need an init and an fini, each to be called exactly once. How would you do it? Maybe we have different ideas on how the init/fini are called. I thought init was called on entry to run_main_graph (and other entry points), and fini (if necessary) is called on exit from run_main_graph (and other entry points).

The ref count solution indeed satisfy this description. It is not what is required by the zdnn at this stage. It only requires `zdnn_init()` to be called once prior to any other zdnn calls.

tungld (commented):
I would like to understand a big picture, e.g. when adding a new accelerator.
Each accelerator must define three functions: OMIsInitAccelX, OMInitAccelX, OMShutdownAccelX. How can we ensure that the accelerator has defined these functions (of course, we can have a documentation say that users should do it, but I am talking about source code)?

src/Accelerators/NNPA/Runtime/OMRuntimeNNPA.c
gongsu832: > Correct, the inner mutex will always succeed. The operation I need is the memory fence in the `pthread_mutex_unlock`. Without a memory fence between the stores in the `zdnn_init()` and the `OMIsInitAccelNNPA=1`, another thread may see the store associated with `OMIsInitAccelNNPA=1` before a store inside the init, and thus see the initialization in an inconsistent state. The memory fence basically enforce the following: "if any thread witness a memory state set AFTER the fence, it is guaranteed to see any of the memory state set before the fence". So it does not guarantee any other threads to see `OMIsInitAccelNNPA==1` (thus forcing them possibly in one more pointless mutex), but if they see `OMIsInitAccelNNPA==1` then they are guaranteed to see all of the stores performed in the init.

There is no memory fence issue in this case. Memory fence is only needed when instructions are scheduled/executed out of order. Since zdnn_init is not an inline function, the last instruction of of zdnn_init will be a "return" instruction. No compiler/hardware will schedule/execute this "return" instruction and OMIsInitAccelNNPA=1 out of order. So no memory fence is ever needed between zdnn_init and OMIsInitAccelNNPA=1.



> The ref count solution indeed satisfy this description. It is not what is required by the zdnn at this stage. It only requires `zdnn_init()` to be called once prior to any other zdnn calls.

So you are going to write this code again for every accelerator we add? I was assuming that you would have a general init/fini like the refcnt code. Then for each accelerator you add, you simply do init(zdnn_init)/fini(NULL), or init(newdnn_init)/fini(newdnn_fini).

gongsu832:
> I would like to understand the big picture, e.g. when adding a new accelerator.
> Each accelerator must define three functions: OMIsInitAccelX, OMInitAccelX, OMShutdownAccelX. How can we ensure that the accelerator has defined these functions (of course, we can have a documentation say that users should do it, but I am talking about source code)?

Since these functions are part of the accelerator which are external to onnx-mlir, we can't really enforce anything on them. It's just a rule that the accelerator code writer will have to abide by. If they don't supply those functions, we are just going to fail when we try to call those functions.

src/Accelerators/NNPA/Runtime/OMRuntimeNNPA.c
```diff
@@ -0,0 +1,96 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//===-------------------------- OMNNPA.c ---------------------------===//
```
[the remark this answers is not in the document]
AlexandreEichenberger: thanks

src/Accelerators/NNPA/Runtime/OMRuntimeNNPA.c
```diff
@@ -0,0 +1,96 @@
...
+
+/* Interface for device init and shutdown.
+ *
+ * Testing if the code is initalized or not should be done prior to any zdnn
+ * computations. This test can be performed in the run_main_graph() without
+ * grabbing a lock, as follows:
+ *
+ * if (!OMIsInitAccelNNPA) OMInitAccelNNPA();
+ *
+ * OMInitAccelNNPA() is thread save, and is guaranteed to set
```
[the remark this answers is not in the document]
AlexandreEichenberger: done.

AlexandreEichenberger:
Added some comments describing the high level interface. While other devices are technically not bound to use the same interface, doing so would reduce the work being added to the compiler. Thus using the same interface is strongly encouraged.

src/Accelerators/NNPA/Runtime/OMRuntimeNNPA.c
tungld: I have one example where a refcount is useful. We have two threads running in parallel but not exactly start and finish at the same time (in other words, not coming to the mutex at the same time):
- thread 1 calls init, starts first and runs for a while.
- thread 2 jumps in, calls init while thread 1 is still running. (we have no problem here when using refcount or two locks)
- now thread 2 calls fini and finishes first:
    - if we uses the two locks, thread 2 will call fini and successfully shutdown the accelerator. Thread 1 would fail to use the accelerator then.
    - if we use refcount, fini knows that there is another active use by looking at the refcount. fini only decreases the refcount without shutting down the accelerator.

AlexandreEichenberger:
>I have one example where a refcount is useful. We have two threads running in parallel but not exactly start and finish at the same time (in other words, not coming to the mutex at the same time):
thread 1 calls init, starts first and runs for a while.
thread 2 jumps in, calls init while thread 1 is still running. (we have no problem here when using refcount or two locks)
now thread 2 calls fini and finishes first:
if we uses the two locks, thread 2 will call fini and successfully shutdown the accelerator. Thread 1 would fail to use the accelerator then.
if we use refcount, fini knows that there is another active use by looking at the refcount. fini only decreases the refcount without shutting down the accelerator.

This is a misunderstanding. Init is called only once to initialize the zdnn lib for ever. Shutdown is currently never called. I could remove it. There is a scenario if the process was moved to a different node, but even then the ZDNN folks tell me that there would be no need to shut down. I can remove the call if that makes everyone here more comfortable.

tungld (approved):
LGTM.

We have consensus on the interfaces proposed in this patch though it is still debating on the internal implementation for NNPA. So I recommend to merge this first so that other patches are not blocked.

gongsu832:
> > I have one example where a refcount is useful. We have two threads running in parallel but not exactly start and finish at the same time (in other words, not coming to the mutex at the same time):
> > thread 1 calls init, starts first and runs for a while.
> > thread 2 jumps in, calls init while thread 1 is still running. (we have no problem here when using refcount or two locks)
> > now thread 2 calls fini and finishes first:
> > if we uses the two locks, thread 2 will call fini and successfully shutdown the accelerator. Thread 1 would fail to use the accelerator then.
> > if we use refcount, fini knows that there is another active use by looking at the refcount. fini only decreases the refcount without shutting down the accelerator.
> 
> This is a misunderstanding. Init is called only once to initialize the zdnn lib for ever. Shutdown is currently never called. I could remove it. There is a scenario if the process was moved to a different node, but even then the ZDNN folks tell me that there would be no need to shut down. I can remove the call if that makes everyone here more comfortable.

I don't think the issue here is about what needs to be done specifically for zdnn. I thought we are trying to have a framework that is relatively easy to add different types of accelerators, as evidenced by the "complex" macros that try to automatically generate the function prototypes, etc. So the question is, do we want to write these functions again and again for every new accelerator we add, or is there a common usage pattern for them that can be factored out so we don't have write it again and again.

> LGTM.
> 
> We have consensus on the interfaces proposed in this patch though it is still debating on the internal implementation for NNPA. So I recommend to merge this first so that other patches are not blocked.

I don't think this PR is in a mergeable state yet. Besides the issue above, the memory fence issue is not resolved.

gongsu832:
> Totally agree. I believe the current 2 functions to init/shutdown are pretty generic and should work for other hypothetical devices, though the devil is always in the detail, so it is difficult to be truly sure before having other device examples.

A generic framework where multiple threads will be calling init/shutdown and only the first should call init and the last should call shutdown will most certainly need a reference count. The current code is based on zdnn not requiring shutdown.


> Memory fence works because mutex requires them on relaxed consistency machines. I suggest that I can change the inner lock in the near future if you prefer for me to replace it by an explicit fence. I just need a bit of time to find out what the intrinsic is for both LoZ and zOS.

I'm saying that there is no need for a memory fence at all unless init_zdnn is inlined.

AlexandreEichenberger:
> A generic framework where multiple threads will be calling init/shutdown and only the first should call init and the last should call shutdown will most certainly need a reference count. 

That is just false.  There are no connection between the two. There is no requirements that a shutdown be called once there is no currently-active graph evaluation.

> I'm saying that there is no need for a memory fence at all unless init_zdnn is inlined.

inlining has nothing to do with memory reordering, sorry.

gongsu832:
> That is just false. There are no connection between the two. There is no requirements that a shutdown be called once there is no currently-active graph evaluation.

- Whether a shutdown should be called or not, called once or multiple times is a property of the accelerator. Just because zDNN doesn't require a shutdown doesn't mean others don't.
- How do you know there is no currently-active graph evaluation?



> inlining has nothing to do with memory reordering, sorry.

So please show an instruction sequence from two threads that exhibits the incorrect behavior you claimed without a memory  fence.

AlexandreEichenberger:
race:

```
thread 1: store value X in memory location Y as part of zdnn_init
thread 1: store 1 in OMIsInitAccelX

thread 2: read value 1 in OMIsInitAccel
thread 2: take actions based on the assumption that zdnn_init has completed, but a load from Y can revel the value before X as there is no memory fence between the two stores in thread 1.
```

Recall that this approach is predicated on allowing reads of  `OMIsInitAccelX` outside of a lock.

Second, for the shutdown, I explicitly rely on the framework to have other means to know there is no pending request. 
If that is troubling for you, I will remove the shutdown as it is not needed for NNPA, as I have already offered.

gongsu832:
Let me make it a bit easier to read:
```
thread 1: store Y = X (in zdnn_init)         (1)
thread 1: store OMIsInitAccelX = 1           (2)

thread 2: load OMIsInitAccelX == 1?          (3)
thread 2: load Y                             (4)
```
Your claim is thread 2 in (4) Y might get a value other than X before it was set by thread 1 in (1).

That is only possible if thread 1 executes (1) and (2) out of order, i.e., it executes (2) before (1).

However, even without a memory fence, it's impossible for that to happen. Because (1) is inside the `zdnn_init` function, so there will be at the minimum a couple of instructions to restore registers and return after (1). Let's call all these instructions (5). In other words, in between (1) and (2), there will be (5). No hardware I know will move (2) in front of (1)+(5), with or without a memory fence. Heck, the instruction in front of (2) won't even be (1)+(5); it will be the instruction to call `zdnn_init`. (1)+(5) will be somewhere completely unrelated to (2). The only way for (2) to be moved in front of (1)+(5) without a memory fence is that `zdnn_init` is inlined so (5) does not exist.

As for the shutdown, I had thought this PR will include the framework code that tracks the active threads so we can handle general cases where both init and shutdown are required, not just zDNN. But if that is intended for another PR, that's fine.

AlexandreEichenberger:
@gongsu832 Apologies... I was under the impression that the Z memory subsystem was the same as the Release Consistency model as on POWER. After confirming with experts, that is not the case. So the fence and/or the inner lock are not needed. I will remove them.

AlexandreEichenberger:
@gongsu832 are you ok to merge this PR. Not opposed to further changes now or later. As the interface is only used by our compiler, there is no issue changing it in the near future to settle to a better solution as need arises.

PR #1361, opened 2022-04-18
Fix a bug about normalizing multiple dereferencing uses in a single op and reenable MLIR CSE

tungld:


AlexandreEichenberger (approved):
LGTM, congrats for finding the cause and a temporary fix. Are you going to open an issue with MLIR?

PR #1366, opened 2022-04-19
Relu - Constant Propagation

RobbiePerrone320:
Constant Propagation implementation for Relu Function


tungld:
@RobbiePerrone320 please fix clang-format as follows:
```
--- ./src/Transform/ONNX/ConstProp.cpp	(original)
+++ ./src/Transform/ONNX/ConstProp.cpp	(reformatted)
@@ -3[9](https://github.com/onnx/onnx-mlir/runs/6087279291?check_suite_focus=true#step:4:9)5,7 +395,[13](https://github.com/onnx/onnx-mlir/runs/6087279291?check_suite_focus=true#step:4:13) @@
 
 template <typename T>
 struct ElementWiseUnaryOpImpl<ONNXReluOp, T> {
-  static T impl(T val) { if(val < 0){return 0;} else {return val;} }
+  static T impl(T val) {
+    if (val < 0) {
+      return 0;
+    } else {
+      return val;
+    }
+  }
 };
```

test/mlir/onnx/invalid.mlir
```diff
@@ -166,7 +192,4 @@ func @test_scatterelements_verifier_4(%arg0 : tensor<3xf32>, %arg1 : tensor<3xf3
 
 func @test_sequence_empty() -> none {
   // expected-error @+1 {{SequenceEmpty dtype() does not match the output type}}
-  %1 = "onnx.SequenceEmpty"() : () -> !onnx.Seq<tensor<*xi32>>
-  %2 = "onnx.NoValue"() {value} : () -> none
-  return %2 : none
-}
+  %1 = "onnx.SequenceEmpty"() : () -> !onnx.Seq<tensor<*xi32>>
```
tungld, line 195: Please undo all the changes in this file, they are not related to this PR

tungld (changes requested):
Please remove third_party/zdnn-lib.

gongsu832 (changes requested):
Your PR is trying to add `third_party/zdnn-lib`. We no longer need it since it's now being built separately. You probably have an outdated base repo.

src/Transform/ONNX/ConstProp.cpp
```diff
@@ -392,6 +392,18 @@ template <typename T>
 struct ElementWiseUnaryOpImpl<ONNXSqrtOp, T> {
   static T impl(T val) { return sqrt(val); }
 };
+
+template <typename T>
+ struct ElementWiseUnaryOpImpl<ONNXReluOp, T> {
```
tungld, line 397: Remove spaces before `struct`.

src/Transform/ONNX/ConstProp.cpp
```diff
@@ -392,6 +392,18 @@ template <typename T>
...
+  static T impl(T val) {
+    if (val < 0) {
+      return 0;
+    } else {
+      return val;
+    }
```
tungld, line 403: We could replace this code with:
```C
if (val < 0)
  return 0;
return val;
```

src/Transform/ONNX/ConstProp.cpp
```diff
@@ -392,6 +392,16 @@ template <typename T>
...
 };
+
+template <typename T>
+struct ElementWiseUnaryOpImpl<ONNXReluOp, T> {
+  static T impl(T val) {
+    if (val < 0)
+      return 0;
+    return val;
+  }
+ };
```
tungld, line 403: Remove spaces before `};` too, which caused clang-format check fail.

gongsu832: It's generally a good idea to do
```
clang-format-9 -i filename
```
for any .h .hpp .c .cpp files you modified so that they get formatted according to the clang-format check.

tungld:

tungld:
@RobbiePerrone320 the patch looks good.  Thanks!

Please fix the DCO issue by following its guidance. To avoid this issue in the future, every commit should be signed off by using "git commit --signoff".

tungld (approved):
LGTM!

Congrats on your first contribution to onnx-mlir!

tungld:
@gongsu832 since you requested changes, need your approval for this to be merged.

gongsu832 (approved):
LGTM

PR #1368, opened 2022-04-20
Fix an issue in function areProducedByTransposeOp when input is a block argument

tungld:


chentong319 (approved):
LGTM!

PR #1370, opened 2022-04-20
[ScatterND]: code generation & improved verification code

etiotto:
Implement support for the ONNX `ScatterND` operator:

- [x] verification code (diagnose operator constraints)
- [x] shape inference (trivial)
- [x] codegen support
- [x] added lit tests to verify constraint diagnostics 
- [x] added lit test to verify code generation
- [x] enable end-to-end test (backend test)


src/Conversion/ONNXToKrnl/Tensor/ScatterND.cpp
```diff
@@ -0,0 +1,120 @@
...
+    Value indices = operandAdaptor.indices();
+    auto dataType = data.getType().cast<ShapedType>();
+    auto indicesType = indices.getType().cast<ShapedType>();
+    auto updatesType = updates.getType().cast<ShapedType>();
+    int64_t dataRank = dataType.getRank();
+    int64_t updatesRank = updatesType.getRank();
+    int64_t indicesRank = indicesType.getRank();
+
+    assert(dataRank >= 1 && "The rank of 'data' must be >= 1");
+    assert(indicesRank >= 1 && "The rank of 'indices' must be >= 1");
```
AlexandreEichenberger, line 44: I assume that these input dependent assert where tested in the verifier and are here only for redundancy (which is perfectly fine IMO).
Code below seems to confirm that the verifiers indeed tested these conditions.

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -4462,55 +4461,120 @@ LogicalResult ONNXScatterElementsOp::verify() {
...
+          std::to_string(dataShape[i]));
+  }
 
   return success();
 }
 
 LogicalResult ONNXScatterNDOp::inferShapes(
     std::function<void(mlir::Region &)> doShapeInference) {
-  if (!data().getType().isa<mlir::RankedTensorType>())
+  if (!data().getType().isa<mlir::ShapedType>())
```
AlexandreEichenberger, line 4558: thoughts we converged to use "has shape and rank"

AlexandreEichenberger (approved):
LGTM

src/Conversion/ONNXToKrnl/Tensor/ScatterND.cpp
etiotto: Yes that is correct.  The idea is to diagnose error conditions in the verifier, and in shape inference defensively assert that the "code" is as expected.

src/Dialect/ONNX/ONNXOps.cpp
etiotto: Good point. Going to fix this.

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -107,19 +107,18 @@ static LogicalResult shapeHelperInferMultipleShapes(
 // Handle shape inference for numpy style broadcasting operators.
 template <class OP, class ADAPTOR>
 static LogicalResult inferShapeForBroadcastingOps(
-    OP *op, Type elementType = nullptr) {
-  assert(op && "Expecting a non-null pointer");
-  ADAPTOR operandAdaptor(*op);
+    OP &op, Type elementType = nullptr) {
```
etiotto, line 110: Note: passing `op` as a reference allows us to avoid assertion check that op is a valid pointer.

PR #1374, opened 2022-04-21
Calling accelerator initialization in entry points functions

tungld:
This patch inserts accelerator initialization functions at the beginning of each entry point function.

This patch depends on PR #1357 that defines initialization functions.

### DIFF
--- a/src/Conversion/KrnlToLLVM/KrnlEntryPoint.cpp
+++ b/src/Conversion/KrnlToLLVM/KrnlEntryPoint.cpp
@@ -84,6 +84,23 @@ public:
         createEntryBlock(dynEntryPointFuncTy, dynamicEntryPointFunc, loc);
     rewriter.setInsertionPointToStart(&entryPointEntryBlock);
 
+    // Emit code to initialize accelerators by calling OMInitAccelX where X is
+    // the accelerator name.
+    if (Attribute maccelAttr =
+            module->getAttrOfType<::mlir::Attribute>("onnx-mlir.accels")) {
+      assert(
+          maccelAttr.isa<ArrayAttr>() && "onnx-mlir.accels must be ArrayAttr");
+      ArrayAttr accels = maccelAttr.cast<ArrayAttr>();
+      for (uint64_t i = 0; i < accels.size(); ++i) {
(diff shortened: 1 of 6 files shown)


AlexandreEichenberger (commented):
I see that you did the functionality in a generic way, which is perfectly fine. As long as each device enforce the same protocol (which I recommend) we are all good.

The code sequence you generated is simply
```C++
OMInitAccelX()
```
Which is perfectly fine. We may spare a function call if we added `if(!OMIsInitAccelX)` but this is optional.

Note that I added now a call `OMInitCompatibleAccelNNPA(uint64_t)` and also test the version number.

AlexandreEichenberger (approved):
LGTM, please update to the integrated scheme that also verify the version number, if you can.
I guess my PR needs to go first anyway.

tungld:
> please update to the integrated scheme that also verify the version number, if you can.

Yes, I just switched to `OMInitCompatibleAccelNNPA(uint64_t)'.

src/Conversion/KrnlToLLVM/KrnlEntryPoint.cpp
```diff
@@ -84,6 +84,29 @@ class KrnlEntryPointOpLowering : public OpRewritePattern<KrnlEntryPointOp> {
...
+      for (uint64_t i = 0; i < accels.size(); ++i) {
+        assert(accels[i].isa<StringAttr>() && "Attribute must be StringAttr");
+        StringRef accelStr = accels.getValue()[i].cast<StringAttr>().getValue();
+        std::pair<StringRef, StringRef> NameAndVersion = accelStr.split('-');
+        uint64_t versionNumberInHex =
+            std::stoul(NameAndVersion.second.str(), nullptr, 16);
+        FlatSymbolRefAttr funcRef = getOrInsertOMInitCompatibleAccel(
+            rewriter, module, NameAndVersion.first);
+        LLVM::ConstantOp versionNumberVal = rewriter.create<LLVM::ConstantOp>(
+            loc, int64Ty, rewriter.getI64IntegerAttr(versionNumberInHex));
```
AlexandreEichenberger, line 104: The parameter is actually unsigned, but it probably does not cause any issue here.

Issue #1286, opened 2022-03-30
Accuracy errors of Gemm and MatMul in numerical tests


imaihal:
When using random input range from -10.0 to 10.0, Gemm and MatMul failed in numerical tests. Following is an example difference in Gemm. `a` is computed results and `b` is references. They sometimes does not meet the equation `rtol * abs(b) + atol - abs(a-b) >= 0`. 
No problem when using random input from -1.0 to 1.0 (current default).
 
```
attempt 78 with i 35, j 34, k 16, bTrans, cRank 1, alpha   1.000, beta   1.000
a[16,5] = -0.285781 != b[16,5] = -0.285766
a[16,21] = 1.76696 != b[16,21] = 1.76693
attempt 79 with i 35, j 34, k 16, bTrans, cRank 1, alpha   1.000, beta   1.000
a[14,8] = -0.225075 != b[14,8] = -0.225091
```

imaihal:
It seems this error can be avoided when disabling simdizing.

https://github.com/onnx/onnx-mlir/blob/main/src/Conversion/KrnlToAffine/KrnlMatmul.cpp#L90
I manually set this value to false. (not sure this is correct test)

@AlexandreEichenberger do you have any suggestions?

AlexandreEichenberger:
The problem with simd disabling is that it changes a lot of how the code is generated. 

Are you running on z? If so, you could also omit the -mcpu=z14 and scalar operations will be generated

imaihal:
I usually use z, but the same error issued in Jenkins test for x86 or/and ppc64 (Sorry, I forgot the details since I lost the logs). 

I think generated code depends on tensor shape. Currently I used "I=100, K=100" and "I=100, J=100, K=100". Is there any shape I should test?

AlexandreEichenberger:
I should probably investigate myself, as I  wrote the code. Will try to look at it tomorrow.

tungld:
@imaihal did you use Debug or Release build mode for onnx-mlir? which may affect the C++ code for naive implementation of gemm/matmul. Meanwhile, our generated code is always optimized with simd by using optimization level 3. That would be a subtle difference.

tungld:
> If we can say that this difference is acceptable for simd optimization. I think we can change atol and rtol to pass the test. I felt this difference was a bit large, so I created this issue.

More importantly, we would like to know the root cause before taking any action.

imaihal:
I think so, too. Thanks!

[...]

PR #1378, opened 2022-04-21
Matmul Precision issue due to fused multiply add: fixed by increasing RTOL to pass over 9 orders of magnitude for input ranges.

AlexandreEichenberger:
There was an issue with numerical tests of matmul and gemm with random number beyond the -1.0 to 1.0 range.

Testing with disabling the fused multiply-add succeeded when performing 10,000 tests using the tight relative tolerance of 1e-5. When enabling fused multiply add (FMA), random coefficients in the range below passed the test with the following  in relative tolerances
```
+- 10 passed with rtol 3e-5
+- 1k passed with rtol 5e-5
+- 100k passed with rtol 5e-5
+=100M passed with rtol 3e-5
```

As a result, I did set conservatively the relative tolerance to 5e-5, which passes tests within 9 orders of magnitudes for the input coefficients.

While it is not necessary to set the range of random numbers within large numbers, for sanity I let the TestMatMul2D use values in the -10.0 to +10.0 range.


AlexandreEichenberger:
associated issue: #1286

test/modellib/ModelLib.cpp
```diff
@@ -134,7 +134,11 @@ bool ModelLibBuilder::areCloseFloat(
     const OMTensor *res, const OMTensor *ref) const {
   if (!res || !ref)
     return false;
-  float rtol = getenv("TEST_RTOL") ? atof(getenv("TEST_RTOL")) : 1e-5;
+  // Because of floating point impact of using FMA instead of multiply followed
+  // by an add,the default RTOL is increased from 1e-5 to 5e-5. This bound has
```
etiotto, line 138: [nit] missing space  "add,the" -> "add, the"

etiotto (approved):
LGTM.

test/modellib/ModelLib.cpp
AlexandreEichenberger: tx

imaihal:
@AlexandreEichenberger Thanks for the investigation. This kinds of work should take much time... 

It seems the Jenkins test didn't pass. Should we change RTOL and ATOL?

```
a[10,0] = -3.22827 != b[10,0] = -3.22778
```
I think this difference is bit large, but do you think this difference is reasonable when using FMA ?

AlexandreEichenberger:
> I think this difference is bit large, 

I agree, and the results from the CIs don't seem to agree with the experiments that I did running the tests on Z14 and x86 Linux. Will investigate some more.

PR #1382, opened 2022-04-22
[GatherND]:  Implement verification, shape inference, and code gen.

etiotto:
Implement support for the ONNX GatherND operator:

 - [x] verification code (diagnose operator constraints)
 - [x] shape inference with helper
 - [x] codegen support
 - [x] lit tests to verify constraint diagnostics
 - [x] lit tests to verify shape inference
 - [x] add lit test to verify code generation
 - [x] enable end-to-end test (backend test)


src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
```diff
@@ -453,10 +453,9 @@ Value foldOrEmitONNXTransposeOp(ConversionPatternRewriter &rewriter,
 }
 
 /// Emit MemRef ReinterpretCastOp to create a new view for 'data'.
-/// The new view is created using the given 'memRefType' and 'outputDims'.
+/// The new view is created using the given 'outputDims'.
 Value emitMemRefReinterpretCastOp(ConversionPatternRewriter &rewriter,
-    Location loc, Value data, const MemRefType &memRefType,
-    SmallVectorImpl<IndexExpr> &outputDims) {
+    Location loc, Value data, SmallVectorImpl<IndexExpr> &outputDims) {
```
AlexandreEichenberger, line 458: It feels like the function `emitMemRefReinterpretCastOp` could be simply substituted by `createMemRef.reinterpretCast(data, outputDims);` in the caller code?

src/Conversion/ONNXToKrnl/Tensor/GatherND.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+struct ONNXGatherNDOpLowering : public ConversionPattern {
+  ONNXGatherNDOpLowering(TypeConverter &typeConverter, MLIRContext *ctx)
+      : ConversionPattern(
+            typeConverter, ONNXGatherNDOp::getOperationName(), 1, ctx) {}
+
+  // When true causes injection of print stmts in the generated code.
+  static constexpr bool emitPrintStmts = false;
+
+  static void printIndices(
```
AlexandreEichenberger, line 34: Could you comment what this function is for? Ok its for debugging, just explain.

src/Runtime/OMTensor.inc
```diff
@@ -423,63 +423,71 @@ void omTensorPrint(const char *msg, const OMTensor *tensor) {
...
+  for (int64_t i = 0; i < rank; i++)
+    printf("[%lld]", (long long)shape[i]);
+  printf("\n");
   printf("\tstrides: ");
   for (int64_t i = 0; i < rank; i++)
     printf("[%lld]", (long long)strides[i]);
   printf("\n");
 
-  printf("\tdata: ([");
+#define START_LOOP(IV, LB, UB)                                                 \
```
AlexandreEichenberger, line 435: My personal preference is not to go crazy with macros. It would feel that using a lambda function here would be perfectly appropriate.

If you really want, you can even do recursive lambda functions where you expand a `{I, J, K}, {ilb, jlb, akb}, {iub, jub, kub}` into a 3 level lambda that expires exactly what you want.

etiotto: Understand your dislike of macros. I used macros here because this is a C function, part of libcruntime.a and so can't use C++ here.  I looked and currently the compiler can only generate code to C function defined in the runtime, because `onnx-mlir` links only libcruntme.a which contains exclusively objects with C source code.

AlexandreEichenberger: Got it, that makes sense. For the record, I don't dislike macros (I used to use quite vile recursive macros...) but find C++ constructs more readable. It make sense to use macro in this context.

etiotto: Using variadic macros now.

src/Conversion/ONNXToKrnl/Tensor/GatherND.cpp
etiotto: done

src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
etiotto: done

AlexandreEichenberger (approved):
LGTM

AlexandreEichenberger:
@etiotto there seems to be an issue during compilation
```
ile included from /workdir/onnx-mlir/src/Runtime/OMTensor.cpp:16:
/workdir/onnx-mlir/src/Runtime/OMTensor.inc: In function 'void omTensorPrint(const char*, const OMTensor*)':
/workdir/onnx-mlir/src/Runtime/OMTensor.inc:475:24: error: taking address of temporary array
  475 |     LOOP_1(((int64_t[]){i}), i, shape[0])
      |            ~~~~~~~~~~~~^~~~
```

sstamenova:
> @sstamenova I am seeing MLIR-Windows-CI cancelled. Would you be able to let us know what is happening? I have seen it for many other PRs as well. Thanks

There's most likely an outage that's impacting the builds - the logs don't seem to indicate that anything else is going on. :(. Let me know if it persists and I'll spend some more time looking into it, but if it's an outage, I expect it will go back to normal soon.

etiotto:
> @etiotto there seems to be an issue during compilation
> 
> ```
> ile included from /workdir/onnx-mlir/src/Runtime/OMTensor.cpp:16:
> /workdir/onnx-mlir/src/Runtime/OMTensor.inc: In function 'void omTensorPrint(const char*, const OMTensor*)':
> /workdir/onnx-mlir/src/Runtime/OMTensor.inc:475:24: error: taking address of temporary array
>   475 |     LOOP_1(((int64_t[]){i}), i, shape[0])
>       |            ~~~~~~~~~~~~^~~~
> ```

I am using compound literals (https://en.cppreference.com/w/c/language/compound_literal) which are legal constructs since C99.  Likely the issue here is that `OMTensor.inc` is included in both `OMTensor.c` and `OMTensor.cpp` and when compiling `OMTensor.cpp` the C++ compiler emits a warning because unfort. compound literals aren't part of the C++ standard specification.

AlexandreEichenberger:
It does not seems to be a warning, I see an "Error"

etiotto:
Yes I pushed a fix now.

src/Dialect/Krnl/DialectBuilder.cpp
```diff
@@ -208,37 +240,14 @@ void KrnlBuilder::printf(StringRef msg) const {
...
-      });
-
+  StringRef format = getFormat(inputType);
   std::string concat(msg.str() + format.str());
   StringRef newFormat(concat);
   b.create<KrnlPrintOp>(loc, newFormat, input);
 }
+void KrnlBuilder::printf(Value input, Type inputType) const {
+  StringRef format = getFormat(inputType);
+  b.create<KrnlPrintOp>(loc, format, input);
```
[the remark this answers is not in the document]
etiotto: Yes, because in some cases the user might want multiple prints on the same line and if "\n" is silently added that wouldn't be possible. The user can always emit a `krnl.print("\n")` to force a newline.

etiotto:
@sstamenova I have some code in OMTensor.inc that uses function like macros. The code compiles on all platforms except on Windows.  Looking at the messages in the log did not really help me figuring out what the issue might be.  Could you take a look if you have some time please ?  Any suggestions very much appreciated.

etiotto:
> @sstamenova I have some code in OMTensor.inc that uses function like macros. The code compiles on all platforms except on Windows. Looking at the messages in the log did not really help me figuring out what the issue might be. Could you take a look if you have some time please ? Any suggestions very much appreciated.
> 
> P.S.: I could undo the change in OMTensor.inc as it is not essential for this PR. But if the issue turn out to be easy to fix that would be preferable.

I got a Windows VM and was able to reproduce the issue using a cut down test. It turns out that MSVC, for historical reasons, "interprets" the C specification somewhat differently than gcc & clang.  MSCV provide a compiler option (https://docs.microsoft.com/en-us/cpp/preprocessor/preprocessor-experimental-overview?view=msvc-170) that makes the preprocessor more conforming to the C standard. Using that flag fixed the problem for the reduce test case. I have added it to the CMakefile and triggered a build nwo...

sstamenova:
> > @sstamenova I have some code in OMTensor.inc that uses function like macros. The code compiles on all platforms except on Windows. Looking at the messages in the log did not really help me figuring out what the issue might be. Could you take a look if you have some time please ? Any suggestions very much appreciated.
> > P.S.: I could undo the change in OMTensor.inc as it is not essential for this PR. But if the issue turn out to be easy to fix that would be preferable.
> 
> I got a Windows VM and was able to reproduce the issue using a cut down test. It turns out that MSVC, for historical reasons, "interprets" the C specification somewhat differently than gcc & clang. MSCV provides a compiler option (https://docs.microsoft.com/en-us/cpp/preprocessor/preprocessor-experimental-overview?view=msvc-170) that makes the preprocessor more conforming to the C standard. Using that flag fixed the problem for the reduced test case. I have added it to the CMakefile and triggered a new build ...

@etiotto : Please let me know if it doesn't work and I'll have a look. I was OOF earlier this week, but I'm back in the office now.

PR #1383, opened 2022-04-25
Use TypeConverter instead of a separate function for converting types.

tungld:
Currently the function `convertToMemRefType` does the job of type conversion in a ConversionPattern. While every ConversionPattern has a type converter, there is no reason to use a separate function like `convertToMemRefType`.

This patch will replace `converToMemRefType` by the ConversionPattern's TypeConverter.


src/Conversion/ONNXToKrnl/Math/Clip.cpp
```diff
@@ -32,7 +32,12 @@ struct ONNXClipOpLowering : public ConversionPattern {
       ConversionPatternRewriter &rewriter) const final {
     Location loc = op->getLoc();
     ONNXClipOp clipOp = cast<ONNXClipOp>(op);
-    MemRefType memRefType = convertToMemRefType(*op->result_type_begin());
+
+    // Convert the output type to MemRefType.
+    Type convertedType = typeConverter->convertType(*op->result_type_begin());
+    if (!convertedType || !convertedType.isa<MemRefType>())
```
etiotto, line 38: We expect the type to be converted correctly, so should this be an assert?

tungld: OK. I'll change to use `assert`.

etiotto (approved):
LGTM.

tungld:

tungld:
@etiotto thanks, I addressed your comments.

PR #1384, opened 2022-04-25
Misc cmake updates

gongsu832:
- force setting cached MLIR_DIR to honor command line argument
- unset cached LLVM_DIR so it changes along with MLIR_DIR
- surround third_party with set(CMAKE_MESSAGE_LOG_LEVEL NOTICE) and
  set(CMAKE_MESSAGE_LOG_LEVEL STATUS) to mask out their cmake
  output so we can see more clearly output by onnx-mlir only.
  third_party cmake output can still be turned on by the --log-level
  command line option
- rename cmake option ACCELERATORS_TO_BUILD to ONNX_MLIR_ACCELERATORS
  for consistency
- allow accelerator lit tests even if accelerator code itself
  cannot be built
- fix bug in zdnn.cmake where ar was run even though libzdnn.so was
  already up to date, causing concurrent builds to corrput libzdnn.a

### DIFF
--- a/MLIR.cmake
+++ b/MLIR.cmake
@@ -3,13 +3,9 @@
 # Must unset LLVM_DIR in cache. Otherwise, when MLIR_DIR changes LLVM_DIR
 # won't change accordingly.
 unset(LLVM_DIR CACHE)
-if (DEFINED ENV{MLIR_DIR})
-  set(MLIR_DIR $ENV{MLIR_DIR} CACHE PATH "Path to directory containing MLIRConfig.cmake" FORCE)
-elseif (NOT DEFINED MLIR_DIR)
+if (NOT DEFINED MLIR_DIR)
   message(FATAL_ERROR "MLIR_DIR is not configured but it is required. "
-    "Either set the environmental variable MLIR_DIR, e.g.,\n"
-    "    MLIR_DIR=/path/to/llvm-project/build/lib/cmake/mlir cmake ..\n"
-    "or set the cmake option MLIR_DIR, e.g.,\n"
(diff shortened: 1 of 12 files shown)


MLIR.cmake
```diff
@@ -1,10 +1,17 @@
 # SPDX-License-Identifier: Apache-2.0
 
+# Must unset LLVM_DIR in cache. Otherwise, when MLIR_DIR changes LLVM_DIR
+# won't change accordingly.
+unset(LLVM_DIR CACHE)
 if (DEFINED ENV{MLIR_DIR})
-  set(MLIR_DIR $ENV{MLIR_DIR} CACHE PATH "Path to directory containing MLIRConfig.cmake")
+  set(MLIR_DIR $ENV{MLIR_DIR} CACHE PATH "Path to directory containing MLIRConfig.cmake" FORCE)
```
sstamenova, line 7: I would prefer to drop the environment var option entirely rather than make it take precedence always. I understand this is unlikely to happen, but I think it's a good idea to consider not using environment variables as much as possible since generally cmake expects to work with configuration options (that are then cached, so you never have to pass them more than once).

sstamenova (approved):

MLIR.cmake
gongsu832: Done.

AlexandreEichenberger:
@gongsu832 

Question: is the MLIR_DIR removed? There are lots of references to it, for example in the docker files
```
ENV MLIR_DIR=/workdir/llvm-project/build/lib/cmake/mlir
```

or in our doc.

If we don't use the env var anymore, then all traces of it should be removed.

PR #1385, opened 2022-04-26
Add explicit instantiation of getScalarValue for more types

philass:
I was trying to use the `getScalarValue` helper function for `int16_t`, `int32_t`, and `float`, but wasn't able to.

I have added explicit instantiations for them in this MR.

### DIFF
--- a/src/Dialect/ONNX/ONNXOpsHelper.cpp
+++ b/src/Dialect/ONNX/ONNXOpsHelper.cpp
@@ -555,7 +555,10 @@ RESULT_TYPE getScalarValue(ONNXConstantOp constantOp, Type type) {
 // Template instantiation for getScalarValue
 
 template double getScalarValue<double>(ONNXConstantOp constantOp, Type type);
+template float getScalarValue<float>(ONNXConstantOp constantOp, Type type);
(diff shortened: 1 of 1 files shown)


philass:
@tungld, if you get time could you take a peak at this?

tungld:

tungld:
@philass I guess the reason `getScalarValue` supports `double` and `int64_t` is to avoid precision loss. Would like to have comments from @AlexandreEichenberger on this.

AlexandreEichenberger (commented):
My guess is that there is no harm in adding the template instantiation. 

Maybe we should look into how reduced precision is used in MLIR? @philass would you be able to search llvm-project/mlir to see if there are examples of `(RESULT_TYPE)(*valueIt).cast<IntegerAttr>().getInt();` being casted to lower precision quantities? I would search for `int32.*getInt\(\)`

philass:
I see, I didn't realize that the decision to use Int64, and double as the return types was because they are sufficient.

As in they can capture all lower precision datatype values without loss of precision. I can just cast after the fact. 

Thanks for the explanation @tungld, and @AlexandreEichenberger. I'm happy to close this

AlexandreEichenberger:
I will assume that since its close, there is no more need for testing.

PR #1386, opened 2022-04-26
Added explanations to rerun a numerical test

AlexandreEichenberger:
Doc currently does not indicate how to re-run a numerical test while generating the same configurations as in a prior run. This ability is super useful, and is now properly documented.


etiotto (approved):
LGTM

PR #1393, opened 2022-04-27
Fix output type when CategoryMapper is imported

chentong319:
The type inference for CategoryMapper in model importer is a special case: The element type of the output type is I64 if the input is string type, or string type if the input is I64 type. onnx-mlir did not handle such case. 
The fix is use a specialized importer for CategoryMapper to determine the output type. Pass the result to the type inference for the output tensor with an added parameter. 
With this PR, bidalf-9.onnx can be parsed by onnx-mlir. Now I got error related to shape inference: 
```
error: Invalid axis value
error: shape inference failed
```


tungld (approved):
LGTM.

PR #1394, opened 2022-04-28
Add a script to check models in the model zoo

tungld:
This patch adds a script that can automatically download models from the model zoo (https://github.com/onnx/models), compile, run the models and verify inference outputs.

By default, the script runs for all models in the model zoo. To run with specific models, please directly change `models_to_run` in the script, e.g. `models_to_run = ['mnist-8', 'yolov4', 'resnet50-v2-7']`


chentong319:
Could we put the script into onnx-mlir/test/choose-a-name? It is part of our test, and it is very likely that we will start the test with a cmake target.

utils/CheckONNXModelZoo.py
```diff
@@ -0,0 +1,278 @@
...
+                ['find', tmpdir, '-name', '*.pb', '-printf', '%h\n'])
+            if (len(pb_files) > 0):
+                has_data_sets = True
+                data_set = pb_files.split('\n')[0]
+        if (not has_data_sets):
+            log_l1("Warning: This model does not have test data sets.")
+
+        # compile, run and verify.
+        log_l1("Checking the model {} ...".format(model_name))
+        compile_args = '--compile_args=-O3'
```
etiotto, line 178: Would be nice to allow users of the script to set the onnx-mlir optimization level (and other compiler options) via the script command line interface.  By default, is the users does not overwrite the compiler options, I think is OK to compile at -O3.

tungld:
@chentong319 @etiotto thanks, I addressed your comments and find a new home for the script.

AlexandreEichenberger (commented):
Please add a section in docs/Testing.md on how to use this, thanks

etiotto (approved):
LGTM

AlexandreEichenberger (approved):
LGTM

tungld:
> Please add a section in docs/Testing.md on how to use this, thanks

Done. Thanks!

PR #1395, opened 2022-04-28
Support for maccel in the compiler interface

AlexandreEichenberger:
Added "--maccel" setting through the programatic compiler interface. So far accel could be requested via the env variable or the command line, but not through the setCompilerOption interface or omSetCompilerOption.

Since we can specify multiple accelerators, the setCompilerOption add to the current list, unless the string "RESET" is provided, in which case the list of target accelerator is reset. An alternative option was to reset the list when an empty string "" is provided, but I preferred the explicit option. 

In order to do this cleanly, I had to be able to scan strings to Accelerator::Kind as well as printing it to string. To do so cleanly, I added variadic parameter to the `APPLY_TO_ACCELERATORS` macro. It still expect the `macro` name in first position, but subsequent parameters are forwarded to the macro itself.

For example we can define a macro that uses a 2nd parameter.
```C++
#define ACCEL_CL_ENUM_FROM_STRING(name, str) \
  if (str.compare(std::string(#name))==0) return accel::Accelerator::Kind::name;
```

Which is invoked at follows.

```C++
  APPLY_TO_ACCELERATORS(ACCEL_CL_ENUM_FROM_STRING, str);
```

In other words, we can pass additional parameters when invoking `APPLY_TO_ACCELERATORS` which will be simply passed through to the actual macro, here `ACCEL_CL_ENUM_FROM_STRING`, as 2nd, 3rd,.... param (here only one more).


include/OnnxMlirCompiler.h
```diff
@@ -91,8 +91,10 @@ ONNX_MLIR_EXPORT int64_t omSetCompilerOptionsFromArgsAndEnv(
  *  omSetCompilerOptionsFromEnvAndArgs. Or overwrite the current value
  *  by calling this function after one of the above 3 setter functions.
  *  @param kind Describe which option kind is being set.
- *  @param val Value of the option being set. Null pointer reset the
- *  option.
+ *  @param val Value of the option being set. Empty string reset the
+ *  option. Setting TargetAccel is different. When passing a valid
+ *  accelerator, it is added to the list of target accelerators unless
+ *  val="RESET", in which case the list is cleared.
```
etiotto, line 97: Given that the maccel command line option accept "NONE" to indicate there are no accelerators I wonder whether would make sense to also use "NONE" instead of "RESET" here. What do you think?

src/Accelerators/Accelerator.cpp
```diff
@@ -25,5 +25,25 @@ const llvm::SmallVectorImpl<Accelerator *> &Accelerator::getAccelerators() {
   return acceleratorTargets;
 }
 
+// Help to print accelerator kinds.
+static std::map<Accelerator::Kind, std::string> mapKind2Strings;
```
etiotto, line 29: Should we also `#include <map>` at the top of the file?

src/Accelerators/Accelerator.cpp
```diff
@@ -25,5 +25,25 @@ const llvm::SmallVectorImpl<Accelerator *> &Accelerator::getAccelerators() {
...
+
+std::ostream &operator<<(std::ostream &out, const Accelerator::Kind kind) {
+  if (mapKind2Strings.size() == 0) {
```
etiotto, line 32: [Suggestion]:  `if (mapKind2Strings.empty()) {`

src/Accelerators/Accelerator.cpp
```diff
@@ -25,5 +25,25 @@ const llvm::SmallVectorImpl<Accelerator *> &Accelerator::getAccelerators() {
...
+    APPLY_TO_ACCELERATORS(ACCEL_CL_ENUM_TO_STRING, mapKind2Strings);
+    mapKind2Strings[Accelerator::Kind::NONE] = "NONE";
+  }
+  return out << mapKind2Strings[kind];
+}
+
+llvm::raw_ostream &operator<<(
+    llvm::raw_ostream &out, const Accelerator::Kind kind) {
+  if (mapKind2Strings.size() == 0) {
```
etiotto, line 41: [Suggestion]: if (mapKind2Strings.empty()) {

src/Accelerators/Accelerator.cpp
```diff
@@ -25,5 +25,25 @@ const llvm::SmallVectorImpl<Accelerator *> &Accelerator::getAccelerators() {
...
+
+std::ostream &operator<<(std::ostream &out, const Accelerator::Kind kind) {
+  if (mapKind2Strings.size() == 0) {
+    APPLY_TO_ACCELERATORS(ACCEL_CL_ENUM_TO_STRING, mapKind2Strings);
+    mapKind2Strings[Accelerator::Kind::NONE] = "NONE";
+  }
+  return out << mapKind2Strings[kind];
+}
+
+llvm::raw_ostream &operator<<(
```
etiotto, line 39: Curious to why we need both "operator<<", one taking a sts:ostream and the other a llvm::raw_ostream.

src/Compiler/CompilerOptions.hpp
```diff
@@ -56,6 +56,10 @@ void setTargetArch(const std::string &arch);
 std::string getTargetArchOption();
 void setTargetCPU(const std::string &cpu);
 std::string getTargetCPUOption();
+// SetTargetAccel add given accleator to the list, reset when str="RESET".
```
etiotto, line 59: accleator -> accelerator

src/Compiler/CompilerOptions.hpp
```diff
@@ -73,7 +77,8 @@ using CompilerOptionList =
 extern std::map<std::string, std::vector<std::string>> CompilerConfigMap;
 
 // Return 0 on success. These functions are not thread-safe and should be called
-// by a single program thread.
+// by a single program thread. For OptionKind==TargetAccel, add given accleator
```
etiotto, line 80: accleator -> accelerator

test/numerical/TestMatMul2D.cpp
```diff
@@ -41,8 +40,10 @@ int main(int argc, char *argv[]) {
   setCompilerOption(OptionKind::CompilerOptLevel, "3");
   llvm::cl::ParseCommandLineOptions(
       argc, argv, "TestMatMul2D\n", nullptr, "TEST_ARGS");
-  bool success;
+  std::cout << "Target options: \""
+            << getCompilerOption(OptionKind::TargetAccel) << "\"\n";
 
+  bool success;
   printf("RapidCheck Matrix-Vector test case generation.\n");
   success = rc::check("Matrix-Vector Matmul implementation correctness", []() {
```
etiotto, line 48: I would remove line 46 and here declare and init the variable: 
```
    bool success = ....
```

test/numerical/TestMatMul2D.cpp
```diff
@@ -41,8 +40,10 @@ int main(int argc, char *argv[]) {
   setCompilerOption(OptionKind::CompilerOptLevel, "3");
   llvm::cl::ParseCommandLineOptions(
       argc, argv, "TestMatMul2D\n", nullptr, "TEST_ARGS");
-  bool success;
+  std::cout << "Target options: \""
```
etiotto, line 43: Not a comment on this line.  
What about the other numerical tests? Can you also print the Target options for them pls.

src/Compiler/CompilerOptions.cpp
```diff
@@ -205,6 +205,57 @@ std::string getTargetCPUOption() {
...
+  if (found) {
+    setTargetAccel(accel);
+    return 0;
+  }
+  return 1;
+}
+
+void setTargetAccel(const accel::Accelerator::Kind accel) {
+  LLVM_DEBUG(llvm::dbgs() << DEBUG_TYPE << "Set accel\"" << accel << "\"\n";);
+  // Add accel to maccel.
```
etiotto, line 242: Would remove the comment because the code is clear enough.

src/Compiler/CompilerOptions.cpp
```diff
@@ -205,6 +205,57 @@ std::string getTargetCPUOption() {
...
+  APPLY_TO_ACCELERATORS(ACCEL_CL_ENUM_FROM_STRING, str);
+  // No specific accelerator found, check if we have Kind::NONE
+  if (str.compare(std::string("NONE")) == 0)
+    return accel::Accelerator::Kind::NONE;
+  // No valid accelerator found, report failure.
+  found = false;
+  return accel::Accelerator::Kind::NONE;
+}
+
+// Return 0 on success, nonzero on error.
```
etiotto, line 222: Could this function return a `bool` instead to indicate success/failure ?

src/Compiler/CompilerOptions.hpp
etiotto: Also document the function return semantics (suggest to use a bool rather than an int for the return type).

include/OnnxMlirCompiler.h
AlexandreEichenberger: That make sense, but there is now a slight inconstancy: if on the command line, a user add `--maccel=NNPA --maccel=NONE` the LLVM mechanics does not know that `NONE` invalidates the other ones. And what if a user say `--maccel=NNPA --maccel=NONE --maccel=NNPA2`? Does it process them linearly, and thus have only `NNPA2`, or the `NONE` apply for all?

That is why I preferred avoiding this issue altogether by adding a distinct name, which is only used by the `setCompilerOption`. The reason it is important there to reset, is that a program that compile itself graphs may want to generate two versions, one with and one without accel. So it has to be able to reset as well as add.

One way to avoid this all together is to treat the string passed to  `setCompilerOption` as a set, namely it always reset the previous set and can add multiple, e.g. comma separated accelerators. That would probably be the cleanest as it would match the other options that simply erase the previous option and reset it with the new ones... Would you be ok with comma separated string for describing a set?

src/Accelerators/Accelerator.cpp
AlexandreEichenberger: yes, thanks for noticing it

src/Accelerators/Accelerator.cpp
AlexandreEichenberger: yes, like it better

src/Accelerators/Accelerator.cpp
AlexandreEichenberger: PS: I am usually pretty good at applying suggestions to other places, I appreciate your thoroughness.

src/Accelerators/Accelerator.cpp
AlexandreEichenberger: Indeed, but they don't appear to inherit from one another. Without the raw_ostream, the `<<` was not available inside the DEBUG statements.

src/Compiler/CompilerOptions.hpp
AlexandreEichenberger: tx

test/numerical/TestMatMul2D.cpp
AlexandreEichenberger: sure

test/numerical/TestMatMul2D.cpp
AlexandreEichenberger: sure

src/Compiler/CompilerOptions.cpp
AlexandreEichenberger: that is not the intent, it is to trace what accel was added, to make sure the logic works.

src/Compiler/CompilerOptions.cpp
AlexandreEichenberger: I agree in principle but its a bit complicated because it is implemented in the `  APPLY_TO_ACCELERATORS(ACCEL_CL_ENUM_FROM_STRING, str);` macro. 

If you really want it, I can add an additional variable, the macro can set it, and then return true...

src/Accelerators/Accelerator.cpp
etiotto: ok

src/Compiler/CompilerOptions.cpp
etiotto: ok if it adds more complication then is not worth it. Is fine as is.

AlexandreEichenberger: Actually, the code is much more succinct with your suggestion, implemented it.

include/OnnxMlirCompiler.h
AlexandreEichenberger: Ok, after investigation, the command line accepts multiple `--maccel=xxx` and add them all together. It does not recognize `--maccel=xxx1,xxx2`. So to be as similar as possible to the command line option, we are keeping the current approach of a special keyword, `RESET` to clear the set when passing option using the `setCompilerOption` approach.

AlexandreEichenberger:

AlexandreEichenberger:

src/Compiler/CompilerOptions.cpp
etiotto: ok

src/Compiler/CompilerOptions.hpp
AlexandreEichenberger: Return an int because the function that uses it in the interface follows the "0" fine, "nonzero" is error code.
Now if you want it, I can return a boolean here and then convert the true/false to an zero/nonzero in that interface.

AlexandreEichenberger: Let me know your preference, and will add comments either way.

AlexandreEichenberger (commented):
left a question.

src/Compiler/CompilerOptions.hpp
etiotto: I am fine with following the zero/non-zero interface.

etiotto (approved):
LGTM

tungld (approved):
LGTM.

PR #1396, opened 2022-04-29
Add MemRefNormalizable trait to the newly added KrnlSeqExtractOp

tungld:
Krnl operations whose inputs or outputs are of MemRefType have MemRefNormalizable trait, but the newly added KrnlSeqExtractOp is missing. This patch is to add the trait to KrnlSeqExtractOp.


chentong319 (approved):
LGTM!

Issue #1061, opened 2021-12-14
Segmentation fault when a model was compiled with -O0, -O1


tungld:
I got segmentation fault with many models if they were compiled using `-O0` or `-O1`, e.g. with resnet50:
```
ONNX_MLIR_HOME=/home/tungld/dl/onnx-mlir/build/Debug python ../utils/RunONNXModel.py resnet50.onnx --compile_args="-mcpu=z14 -O0"
Temporary directory has been created at /tmp/tmpjf_k8lqx
Generating random inputs ...
  - 1st input's shape (1, 3, 224, 224)
  done.

Compiling the model ...
Shared library /tmp/tmpjf_k8lqx/model.so has been compiled.
  took  57.782129378058016  seconds.

Running inference ...
Segmentation fault (core dumped)
```

There was no issue when using `-O2` or `-O3`.

AlexandreEichenberger:
Maybe you can quickly determine that the unoptimized LLVM IR we gen at -O3 or -O1 are the same, and if they are, then we know it is an llvm issue.

I wonder if calling `opt` at low level of optimization is the problem. @etiotto do you know? We could try skipping `opt` all together, at the very least just to see what would happen.

etiotto:
IIRC `clag` calls executes `opt` passes even at -O0. Although clang does not actually calls the `opt` tools directly but rather executes opt passes internally. I haven't looked exactly how that happens in `clang` but I suspect it does not matter whether we call opt explicitly or not. 

This is something that we will need to trace in LLVM.

AlexandreEichenberger:
@etiotto any progress on using -O0?

imiiii:
Hi, I get 
6540 segmentation fault (core dumped)  when running ./onnx-mlir --EmitMLIR model.onnx --O1
But I did not get any error when I am running 
./onnx-mlir --EmitMLIR model.onnx --O3
It dumped model.onnx.mlir, I want to know that is this file usable for entry to mlir-opt in MLIR project ??

whitneywhtsang:
I got similar result as @imiiii, where there is a compile time segfault at lower opts.
\> ./onnx-mlir -O3 [resnet50-v1-7.onnx](https://github.com/onnx/models/blob/main/vision/classification/resnet/model/resnet50-v1-7.onnx) --EmitMLIR
(no errors)
[...]

PR #1397, opened 2022-04-30
Create alloca outside of loops

whitneywhtsang:
According to https://llvm.org/docs/LangRef.html#alloca-instruction, the `alloca` instruction allocates memory on the stack frame of the currently executing function, to be automatically released when this function returns to its caller. 

We can avoid over allocing by placing `alloca` instruction at the function scope and outside of any loops. This change is motivated by https://github.com/onnx/onnx-mlir/issues/1061, where models segfault at runtime when compiling at `-O0`, due to run out of stack space.

An alternative is to hoist allocation out of loops after creation with `--buffer-hoisting`: https://mlir.llvm.org/docs/Passes/#-buffer-hoisting-optimizes-placement-of-allocation-operations-by-moving-them-into-common-dominators-and-out-of-nested-regions, as suggested in https://github.com/onnx/onnx-mlir/issues/1061#issuecomment-1113881796. IMO, it is better to create out of loops in the first place instead of relying later on passes to fix it.

### DIFF
--- a/src/Conversion/ONNXToKrnl/Math/Gemm.cpp
+++ b/src/Conversion/ONNXToKrnl/Math/Gemm.cpp
@@ -57,14 +57,14 @@ struct ONNXGemmOpLowering : public ConversionPattern {
     IndexExpr outerUb1 = shapeHelper.dimsForOutput()[1];
     IndexExpr innerUb = shapeHelper.aDims[1];
     SmallVector<IndexExpr, 3> loopUbs{outerUb0, outerUb1, innerUb};
+    // Create temp, single scalar, no need for default alignment.
+    MultiDialectBuilder<KrnlBuilder, MemRefBuilder, MathBuilder> create(
+        createKrnl);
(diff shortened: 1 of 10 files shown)


AlexandreEichenberger (approved):
LGTM, once you add a comment in the alloca definition that strongly suggest adding the alloca outside of innermost loops... That way, we will hopefully also avoid recreating this issue in the future.

whitneywhtsang:
> LGTM, once you add a comment in the alloca definition that strongly suggest adding the alloca outside of innermost loops... That way, we will hopefully also avoid recreating this issue in the future.

Thanks for reviewing @AlexandreEichenberger. Can you please check if I have added the comment at the place you have in mind? BTW, the alloca should not be place inside of a loop, but just innermost loops.

whitneywhtsang:

whitneywhtsang:

gongsu832:

gongsu832:

gongsu832:

gongsu832:

gongsu832:

PR #1404, opened 2022-05-04
Fix image.

doru1004:
Fix image.

### DIFF
--- a/docker/Dockerfile.onnx-mlir
+++ b/docker/Dockerfile.onnx-mlir
@@ -84,6 +84,7 @@ RUN LLVM_PROJECT_ROOT=${WORK_DIR}/llvm-project \
               git libtool make python39 && \
           rm -rf /var/cache/dnf/* /usr/local/bin/ninja; \
        fi \
+    && cp onnx-mlir/build/Release/lib/*.so . \
     && rm -rf /tmp/* /usr/bin/python llvm-project onnx-mlir
 
 # Copy from onnx-mlir-test to squash removed source layer and


docker/Dockerfile.onnx-mlir
```diff
@@ -84,6 +84,7 @@ RUN LLVM_PROJECT_ROOT=${WORK_DIR}/llvm-project \
               git libtool make python39 && \
           rm -rf /var/cache/dnf/* /usr/local/bin/ninja; \
        fi \
+    && cp onnx-mlir/build/Release/lib/*.so . && ls && \
```
gongsu832, line 87: This is not the right way to do it. All the necessary .so should have already been installed into `/usr/local/lib` by the `make install` above. The reason PyRuntime .so was not installed is because it was created by `pybind11_add_module` which does not explicitly install the .so. The correct way is to just add a install clause in the cmake file.
```
diff --git a/src/Runtime/CMakeLists.txt b/src/Runtime/CMakeLists.txt
index 55d01696..34fa8622 100644
--- a/src/Runtime/CMakeLists.txt
+++ b/src/Runtime/CMakeLists.txt
@@ -107,3 +107,6 @@ target_link_libraries(PyRuntime
   PRIVATE
   ExecutionSession
   )
+install(TARGETS PyRuntime
+  DESTINATION lib
+  )
```

chentong319 (approved):
LGTM!

caoimhinuibrian (commented):
LGTM

PR #1407, opened 2022-05-05
Allow setting optimization level for numerical and backend tests

gongsu832:
- Add cmake variable ONNX_MLIR_TEST_OPTLEVEL to allow setting optimization level for numerical tests
- Add -O|--Optlevel option to allow setting optimization level for backend tests
- Prefix env var in test/backend/variables.py with TEST_ to avoid collision
- Run tests in dev image with -O0 to cover this case


gongsu832:
Running tests with -O0 adds about 10 minutes to the build time, which I think is tolerable.

test/CMakeLists.txt
```diff
@@ -1,5 +1,18 @@
 # SPDX-License-Identifier: Apache-2.0
 
+# Valid optimization levels are 0, 1, 2, and 3
+# Default to 3 if ONNX_MLIR_TEST_OPTLEVEL not set or cached
+set(OPTLEVELS 0 1 2 3)
```
sstamenova, line 5: I think it would be better to have the level specified as OX rather than just X. Otherwise it's a little weird.

test/CMakeLists.txt
```diff
@@ -1,5 +1,18 @@
...
+if (NOT DEFINED ONNX_MLIR_TEST_OPTLEVEL)
+  set(ONNX_MLIR_TEST_OPTLEVEL 3 CACHE STRING "" FORCE)
+endif()
+# Fail if specified/cached ONNX_MLIR_TEST_OPTLEVEL not valid
```
sstamenova, line 9: Nit: "_is_ not valid"

test/numerical/CMakeLists.txt
```diff
@@ -29,9 +29,10 @@ function(add_numerical_unittest test_name)
     set_property(TARGET ${test_name} PROPERTY FOLDER "${test_suite_folder}")
   endif ()
 
-  # Force -O3 when invoking the numerical tests, so that the models in the test
-  # are compiled with -O3. TODO: make that a runtime env variable.
-  add_test(NAME ${test_name} COMMAND ${test_name} -O3 WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
+  # Optimization level set by ONNX_MLIR_TEST_OPTLEVEL, defaults to 3
+  add_test(NAME ${test_name}
+    COMMAND ${test_name} -O${ONNX_MLIR_TEST_OPTLEVEL}
+    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
```
sstamenova, line 35: Nit: the closing paren should be on a new line

test/CMakeLists.txt
gongsu832: It's done this way to be consistent with the way the optlevel is set for backend test.py via `-O {0,1,2,3}`.

chentong319 (approved):
LGTM!

sstamenova (approved):

AlexandreEichenberger:
@gongsu832 There are some benchmarks (esp in numerical) that are hard wired to run with -O3. Search for the `"3"` pattern.

gongsu832:
> @gongsu832 There are some benchmarks (esp in numerical) that are hard wired to run with -O3. Search for the `"3"` pattern.

- Add cmake variable ONNX_MLIR_TEST_OPTLEVEL to allow setting optimization level for numerical tests

PR #1408, opened 2022-05-05
Lower krnl.call to llvm

chentong319:
1. Modified the krnl.call definition and builder. 
    The new krnl.call supports that the returned value is a tensor, which is passed as a parameter to krnl.call. The krnl.call itself has no output. The read/write effect for parameters are set.
2. Lower krnl.call to llvm
    Parameters and attributes of krnl.call Op are turned into parameters for LLVM call op. The function type is determined by the types of parameters. In general, parameters with the tensor type or string type are passed by address, and those with scalar types are passed by value. 
3. Test with 6 new test cases that use the krnl.call.
    Since there is no lit test for krnl to llvm, I tested this PR with backend test. Since ResizeOp is quite complicated for deferent configuration, I manually translated the python implementation of ResizeOp in onnx package into c. 6 test cases are added  to backend test.
4. Some document for krnl.region

This PR can be improved in several places. Since it is already quite large, I will leave the improvement for future PRs:
1. Handle optional input tensor with NoneType in a more generic way. Hopefully, null pointer can be used and leave the runtime library to handle it, reducing the number of runtime functions needed.
2. Propagate attributes with default value. Such attribute may not appear in the Op and is lost when the op is translated to krnl.call. Also all the attributes are blindly copied. If an attribute, not from ONNX Op specification, is added, krnl.call will have an extra parameter for that attribute.
3. The Resize runtime is very inefficient.
4. Only some configuration of ResizeOp is supported by runtime library currently.


chentong319:
The failure on Windows is caused by c compiler for runtime library. Dynamic arrays are used in the code, which is not accepted by the windows. The code looks like the following:
```
void foo(int n) {
float a[n];
}
```
Related to this issue, array of dynamic array is not accepted by g++. It seems to me that I have to change the code to avoid that feature.
```
void f(int n, float a[][n]){
    a[k][j] = 0;
}
void g(int n, int m) {
    float b[m][n];
    f(n, b);
}
```

src/Runtime/OMResize.inc
```diff
@@ -0,0 +1,353 @@
...
+#include "onnx-mlir/Runtime/OMTensor.h"
+
+/**
+ * The runtime implementation of ONNXResizeOp is manually translated python
+ * implementation: onnx/onnx/backend/test/case/node/resize.py (onnx v1.9.0)
+ * This implementation can be improved in efficiency.
+ **/
+
+static void linear_coeffs(float ratio, float *coeffs_buffer, int mode) {
+  coeffs_buffer[0] = 1 - ratio;
```
etiotto, line 20: I'm just looking to this code without context and wondering how can we guarantee that `coeffs_buffer` has at least 2 elements. Also `coeffs_buffer` could be null.   You could change the function signature to:
```
static void linear_coeffs(float ratio, float coeffs_buffer[2], int mode) {

Comment (etiotto) [reply src/Runtime/OMResize.inc:20]: Also `mode` is not used, can you remove it pls

File: src/Runtime/OMResize.inc [hunk 2]
```diff
@@ -0,0 +1,353 @@
...
+  coeffs_buffer[1] = ratio;
+}
+
+static void nearest_coeffs(float ratio, float *coeffs_buffer, int mode) {
+  /* integer ratio is handled outside */
+  switch (mode) {
+  case 0: // round_prefer_float
+    coeffs_buffer[0] = ratio <= 0.5;
```
etiotto, line 28: `coeffs_buffer` is an array of float, but the values stored into it are actually of type `_Bool`. Suffest to change its data type to `_Bool` for C or `bool` for C++.

src/Runtime/OMResize.inc
```diff
@@ -0,0 +1,353 @@
...
+  float *points = (float *)malloc(sizeof(float) * coeffs_n);
+
+  get_neighbor(x_ori, n, input_width, (float *)omTensorGetDataPtr(data), points,
+      exclude_outside);
+  float sum = 0.;
+  for (int i = 0; i < n; i++) {
+    sum += coeffs_buffer[i] * points[i];
+  }
+  free(points);
+  // free OMTensor data1
```
etiotto, line 158: remove commented out code

src/Runtime/OMResize.inc
```diff
@@ -0,0 +1,353 @@
...
+  coeffs_buffer[1] = ((A + 2) * ratio - (A + 3)) * ratio * ratio + 1;
+  coeffs_buffer[2] =
+      ((A + 2) * (1 - ratio) - (A + 3)) * (1 - ratio) * (1 - ratio) + 1;
+  coeffs_buffer[3] =
+      ((A * ((1 - ratio) + 1) - 5 * A) * ((1 - ratio) + 1) + 8 * A) *
+          ((1 - ratio) + 1) -
+      4 * A;
+}
+
+static void get_neighbor(float x, int64_t n, int limit, float *data,
```
etiotto, line 61: All the static functions in this file should have a comment describing what they do.

src/Runtime/OMResize.inc
```diff
@@ -0,0 +1,353 @@
...
+
+  start -= pad_width;
+  end -= pad_width;
+
+  for (int i = start; i <= end; i++) {
+    if (i < 0) {
+      if (exclude_outside)
+        points[i - start] = 0;
+      else
+        points[i - start] = data[0];
```
etiotto, line 110: [Suggestion]: `points[i - start] = exclude_outside ? 0 : data[0];`

src/Runtime/OMResize.inc
```diff
@@ -0,0 +1,353 @@
...
+  int64_t input_width = omTensorGetShape(data)[0];
+  float x_ori;
+  switch (coordinate_transformation_mode) {
+  case 0: // half_pixel
+    x_ori = (x + 0.5) / scale_factor - 0.5;
+    break;
+  case 1: // asymmetric
+    x_ori = x / scale_factor;
+    break;
+  }
```
etiotto, line 138: [Suggestion]: could add a default case and assert that coordinate_transformation_mode can never be any value other than zero or one.

src/Runtime/OMResize.inc
```diff
@@ -0,0 +1,353 @@
...
+      /*float * extrapolation_value*/ 0,
+      /*int coordinate_transformation_mode*/ 0,
+      /*exclude */ 0);
+}
+
+void Resize_Size(
+    OMTensor *output, OMTensor *data, OMTensor *size, char *mode_str) {
+  Coeff_Func_t coeffs_f = NULL;
+  int coeffs_n = 0;
+  if (strcmp(mode_str, "nearest") == 0) {
```
etiotto, line 329: See comment at line 299

src/Runtime/OMResize.inc
```diff
@@ -0,0 +1,353 @@
...
+    coeffs_f = nearest_coeffs;
+    coeffs_n = 2;
+  } else if (strncmp(mode_str, "linear", 6) == 0) {
+    coeffs_f = linear_coeffs;
+    coeffs_n = 2;
+  } else if (strcmp(mode_str, "cubic") == 0) {
+    coeffs_f = cubic_coeffs;
+    coeffs_n = 4;
+  } else {
+    assert(0 && "Resize runtime: unsupported mode");
```
etiotto, line 339: See comment at line 309

src/Runtime/OMResize.c
```diff
@@ -0,0 +1,15 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//===--------- OMResize.c - OMResize C Implementation ---------===//
+//
+// Copyright 2019-2021 The IBM Research Authors.
```
etiotto, line 7: Copyright is 2022

src/Dialect/Krnl/Krnl.td
```diff
@@ -36,31 +36,36 @@ def StringType : Type<CPred<"$_self.isa<krnl::StringType>()">, "string type">;
...
+  > {
   let summary = "call operation";
   let description = [{
     The call operation provides a generic way to call an external function
     at Krnl level.  The `funcName` determines which function to call.
-    The `alloc` is the Value to store the function return. Since allocation
-    of the return MemRef involves shape inference usually with IndexExpr.
-    Thus most of time the allocation should stay in compiler, not in runtime library.
+    The `result` is the Value to store the function return. Currently only
+    one output is supported. `result` has to be resultated memref.
```
etiotto, line 47: resultated memref -> a memref

src/Conversion/ONNXToKrnl/Tensor/Resize.cpp
```diff
@@ -129,12 +129,27 @@ struct ONNXResizeOpLowering : public ConversionPattern {
 
     // Call external function when the mode is not "nearest"
     // Create KrnlCallOp and replace the du chain
+    // One of inputs, scales() and size(), has to be None.
```
etiotto, line 132: One of the inputs, 'scales' or 'size', has to be 'None'.

src/Runtime/OMResize.inc
```diff
@@ -0,0 +1,353 @@
...
+      /*OMTensor scales */ scales,
+      /* Coeff_Func_t*/ coeffs_f,
+      /*int coeffs_n*/ coeffs_n,
+      /* roi */ NULL,
+      /*float * extrapolation_value*/ 0,
+      /*int coordinate_transformation_mode*/ 0,
+      /*exclude */ 0);
+}
+
+void Resize_Size(
```
etiotto, line 325: [nit]: I would name this `resize_size` for consistency. Similar for `resize_scales`.

src/Conversion/KrnlToLLVM/KrnlCall.cpp
```diff
@@ -0,0 +1,214 @@
+
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//===------ KrnlCall.cpp - Lower KrnlCallOp -------------------===//
```
etiotto, line 6: increase length

src/Conversion/KrnlToLLVM/KrnlCall.cpp
```diff
@@ -0,0 +1,214 @@
...
+//
+// Copyright 2019-2022 The IBM Research Authors.
```
etiotto, line 8: 2019-2022 --> 2022

src/Conversion/KrnlToLLVM/KrnlCall.cpp
```diff
@@ -0,0 +1,214 @@
...
+  explicit KrnlCallOpLowering(
+      TypeConverter &typeConverter, MLIRContext *context)
+      : ConversionPattern(
+            typeConverter, KrnlCallOp::getOperationName(), 1, context) {}
+
+  LogicalResult matchAndRewrite(Operation *op, ArrayRef<Value> operands,
+      ConversionPatternRewriter &rewriter) const override {
+    KrnlCallOpAdaptor krnlCallAdaptor(operands);
+    auto loc = op->getLoc();
+    KrnlCallOp krnlCallOp = llvm::dyn_cast<KrnlCallOp>(op);
```
etiotto, line 45: I think here we can use `llvm::cast` given we are not checking whether `krnlCallOp` is null

src/Conversion/KrnlToLLVM/KrnlCall.cpp
```diff
@@ -0,0 +1,214 @@
...
+public:
+  explicit KrnlCallOpLowering(
+      TypeConverter &typeConverter, MLIRContext *context)
+      : ConversionPattern(
+            typeConverter, KrnlCallOp::getOperationName(), 1, context) {}
+
+  LogicalResult matchAndRewrite(Operation *op, ArrayRef<Value> operands,
+      ConversionPatternRewriter &rewriter) const override {
+    KrnlCallOpAdaptor krnlCallAdaptor(operands);
+    auto loc = op->getLoc();
```
etiotto, line 44: [nit]: auto -> Location

src/Conversion/KrnlToLLVM/KrnlCall.cpp
```diff
@@ -0,0 +1,214 @@
...
+    rewriter.eraseOp(op);
+    return success();
+  }
+
+private:
+  static void handleOneParameter(PatternRewriter &rewriter, Operation *op,
+      Value parameter, Value original,
+      llvm::SmallVector<Type, 4> &parameterTypeList,
+      llvm::SmallVector<Value, 4> &parameterList) {
+    auto *context = op->getContext();
```
etiotto, line 88: LLVM code guidelines indicate that if the type of the RHS expression is not immediately clear from the context one should use the actual type rather than `auto`.

src/Conversion/KrnlToLLVM/KrnlCall.cpp
```diff
@@ -0,0 +1,214 @@
...
+          krnl::fillOMTensorWithMemRef(convertedConstantGlobal, omTensor,
+              false /*outOwning*/, rewriter, loc, apiRegistry, module);
+          auto int8Ty = IntegerType::get(context, 8);
+          auto opaquePtrTy = LLVM::LLVMPointerType::get(int8Ty);
+          parameterTypeList.emplace_back(opaquePtrTy);
+          parameterList.emplace_back(omTensor);
+        })
+        .Default([&](Attribute attr) {
+          llvm_unreachable(
+              "This type of Attribute used by krnl.call has not implemented");
```
etiotto, line 188: has not -> is not yet

src/Conversion/ONNXToKrnl/Tensor/Resize.cpp
chentong319: Yes, one of them has to be None. In general, the semantics of None is interpreted at the ONNX Op level. We could use tensor<0xT> for None and lower it to memref. So that we can keep None in Krnl and llvm after ONNX Op. We may have another PR for that functionality.

src/Runtime/OMResize.inc
chentong319: The implementation is translated from python code. Some types and parameters could be optimized as you suggested. I may keep some interface as its python counter part before the code is thoroughly tested.

src/Runtime/OMResize.inc
chentong319: The original code takes the conversion of bool to float. I change the code to make it explicit: `coeffs_buffer[0] = ration <= 0.5 ? 1 : 0;`, for all the assignments.

src/Runtime/OMResize.cpp
```diff
@@ -0,0 +1,15 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//===--------- OMResize.cpp - OMResize C++ Implementation ---------===//
```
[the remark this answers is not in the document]
chentong319: Fixed.

src/Runtime/OMResize.inc
```diff
@@ -0,0 +1,353 @@
...
+ * implementation: onnx/onnx/backend/test/case/node/resize.py (onnx v1.9.0)
+ * This implementation can be improved in efficiency.
+ **/
+
+static void linear_coeffs(float ratio, float *coeffs_buffer, int mode) {
+  coeffs_buffer[0] = 1 - ratio;
+  coeffs_buffer[1] = ratio;
+}
+
+static void nearest_coeffs(float ratio, float *coeffs_buffer, int mode) {
```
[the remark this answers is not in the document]
chentong319: Fixed.

src/Runtime/OMResize.inc
chentong319: Resize comes from the ONNX Op name and Size or Scales comes from the attribute name. I can change function name for these two. But I am thinking of constructing the function name automatically.  So can we just leave them?

src/Conversion/KrnlToLLVM/KrnlCall.cpp
chentong319: Fixed

src/Conversion/KrnlToLLVM/KrnlCall.cpp
```diff
@@ -0,0 +1,214 @@
...
+          // onnx to krnl, and krnl to llvm.
+          // In future, the attributes should be converted in krnl.call builder.
+          // This code passed onnx-mlir-opt --convert-krnl-to-llvm test case,
+          // but failed in onnx-milr for the tensor type for the attribute
+          const auto &apiRegistry = RuntimeAPIRegistry::build(module, rewriter);
+          auto tensorTy = denseAttr.getType().cast<TensorType>();
+          tensorTy.dump();
+          auto memRefTy =
+              MemRefType::get(tensorTy.getShape(), tensorTy.getElementType());
+          memRefTy.dump();
```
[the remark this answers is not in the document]
chentong319: deleted

AlexandreEichenberger (approved):
LGTM, pending satisfying Ettore's comments

PR #1409, opened 2022-05-05
Allow passing of product version info

whitneywhtsang:
This PR aims to provide a way for vendors to pass version information.

As an example assume a product has version 1.2.3  and product ID 1234-ABC. The product would build `onnx-mlir` with `-DONNX_MLIR_PRODUCT_VERSION=1.2.3-1234-ABC`: 
```
!1 = !{i32 2, !"Product Major Version", i32 1}
!2 = !{i32 2, !"Product Minor Version", i32 2}
!3 = !{i32 2, !"Product Patchlevel", i32 3}
!4 = !{i32 2, !"Product Id", !"1234-ABC"}
```
The metadata produces to encode the product version and ID will be consumed by LLVM and is expected the LLVM backend (llc for Z/OS) will store the information in the generated object. 


src/Compiler/CMakeLists.txt
```diff
@@ -66,9 +66,21 @@ add_onnx_mlir_library(CompilerPasses
   )
 
 if (ONNX_MLIR_VENDOR)
-  set_source_files_properties(CompilerUtils.cpp
-    PROPERTIES COMPILE_DEFINITIONS "ONNX_MLIR_VENDOR=\"${ONNX_MLIR_VENDOR} \"")
+  list(APPEND DEFINITIONS "ONNX_MLIR_VENDOR=\"${ONNX_MLIR_VENDOR}\"")
 endif()
+if (PRODUCT_VERSION)
```
sstamenova, line 71: I'd like to see a test for this, after this change there's logic that is not being validated anywhere

sstamenova, line 71: I would recommend keeping the naming convention and version specification consistent with LLVM e.g.:

```
if(NOT DEFINED LLVM_VERSION_MAJOR)
  set(LLVM_VERSION_MAJOR 15)
endif()
if(NOT DEFINED LLVM_VERSION_MINOR)
  set(LLVM_VERSION_MINOR 0)
endif()
if(NOT DEFINED LLVM_VERSION_PATCH)
  set(LLVM_VERSION_PATCH 0)
endif()
if(NOT DEFINED LLVM_VERSION_SUFFIX)
  set(LLVM_VERSION_SUFFIX git)
endif()
```

This sets a default version if none is specified, and also uses the major/minor/patch naming which is fairly standard.

etiotto: +1  for try to be consistent with LLBM naming conventions.

Assuming that the names are `PRODUCT_VERSION_MAJOR`, `PRODUCT_VERSION_MINOR`, `PRODUCT_VERSION_PATCH` and `PRODUCT_ID` I think that the defaults should be:
   - PRODUCT_VERSION_MAJOR = 0
   - PRODUCT_VERSION_MINOR = 0
   - PRODUCT_VERSION_PATCH = 0
   - PRODUCT_ID = "NOT_SPECIFIED"
  
Also, there should be some constraints on usage in the `cmake` files. I am thinking about these rules:
  - if  `PRODUCT_ID` is not set via cmake then overwriting `PRODUCT_VERSION_MAJOR` or `PRODUCT_VERSION_MINOR` or `PRODUCT_VERSION_PATCH` is an configuration error to be diagnosed by cmake
  - if `PRODUCT_ID` is set at least ``PRODUCT_VERSION_MAJOR` must be set to a positive integer value different that 0 
  - if `PRODUCT_VERSION_MINOR` is set then `PRODUCT_VERSION_MAJOR` must be set as well 
  - if `PRODUCT_VERSION_PATCH` is set then  `PRODUCT_VERSION_MINOR` must be set as well

whitneywhtsang: @sstamenova Thanks for the comments. I agree with the standard naming.
What's the benefit of having a default version if none is specified? The only one I can think of is so that we can create a test case. The disadvantage is we increase the LLVM file size with not useful metadata.

whitneywhtsang: Added cmake variables constraints as suggested by @etiotto .

whitneywhtsang: This relies on if we decide to have default values or not.

src/Compiler/CMakeLists.txt
```diff
@@ -66,9 +66,36 @@ add_onnx_mlir_library(CompilerPasses
...
+  endif()
+  list(APPEND DEFINITIONS "PRODUCT_VERSION_PATCH=${PRODUCT_VERSION_PATCH}")
+endif()
+if (PRODUCT_ID)
+  if (NOT DEFINED PRODUCT_VERSION_MAJOR)
+    message(FATAL_ERROR "PRODUCT_VERSION_MAJOR must be defined with PRODUCT_ID")
+  endif()
+  list(APPEND DEFINITIONS "PRODUCT_ID=\"${PRODUCT_ID}\"")
+endif()
+set_property(SOURCE CompilerUtils.cpp APPEND PROPERTY COMPILE_DEFINITIONS ${DEFINITIONS})
```
gongsu832, line 98: Using 4 variables seems a bit cumbersome for this. Instead with regex parsing 1 variable such as `-DONNX_MLIR_VERSION=1.2.3-foo` would suffice:
```
# ONNX_MLIR_VERSION is specified/cached.
if (ONNX_MLIR_VERSION)
  string(REGEX MATCH "^([0-9]+)(\\.([0-9]+))?(\\.([0-9]+))?(\\-(.+))?$"
    VERSION_MATCH ${ONNX_MLIR_VERSION})

  # For a version string 1.2.3-foo
  #
  # CMAKE_MATCH_1 = 1
  # CMAKE_MATCH_2 = .2
  # CMAKE_MATCH_3 = 2
  # CMAKE_MATCH_4 = .3
  # CMAKE_MATCH_5 = 3
  # CMAKE_MATCH_6 = -foo
  # CMAKE_MATCH_7 = foo
  #
  # Major is mandatory, minor, patch, and suffix are optional.
  # When not specified, minor, patch, and suffix are empty
  # (or can default to some value)
  set(ONNX_MLIR_VERSION_MAJOR  ${CMAKE_MATCH_1})
  set(ONNX_MLIR_VERSION_MINOR  ${CMAKE_MATCH_3})
  set(ONNX_MLIR_VERSION_PATCH  ${CMAKE_MATCH_5})
  set(ONNX_MLIR_VERSION_SUFFIX ${CMAKE_MATCH_7})

  message(DEBUG "ONNX_MLIR_VERSION_MAJOR =${ONNX_MLIR_VERSION_MAJOR}")
  message(DEBUG "ONNX_MLIR_VERSION_MINOR =${ONNX_MLIR_VERSION_MINOR}")
  message(DEBUG "ONNX_MLIR_VERSION_PATCH =${ONNX_MLIR_VERSION_PATCH}")
  message(DEBUG "ONNX_MLIR_VERSION_SUFFIX=${ONNX_MLIR_VERSION_SUFFIX}")

  # Major must be specified if ONNX_MLIR_VERSION is specified/cached.
  if ("${ONNX_MLIR_VERSION_MAJOR}" STREQUAL "")
    message(STATUS "onnx-mlir version        : invalid version ${ONNX_MLIR_VERSION} ignored")
    unset(ONNX_MLIR_VERSION CACHE)
  else()
    message(STATUS "onnx-mlir version        : ${ONNX_MLIR_VERSION}")
    list(APPEND DEFINITIONS "ONNX_MLIR_VERSION_MAJOR=${ONNX_MLIR_VERSION_MAJOR}")
    if (NOT "${ONNX_MLIR_VERSION_MINOR}" STREQUAL "")
      list(APPEND DEFINITIONS "ONNX_MLIR_VERSION_MINOR=${ONNX_MLIR_VERSION_MINOR}")
    endif()
    if (NOT "${ONNX_MLIR_VERSION_PATCH}" STREQUAL "")
      list(APPEND DEFINITIONS "ONNX_MLIR_VERSION_PATCH=${ONNX_MLIR_VERSION_PATCH}")
    endif()
    if (NOT "${ONNX_MLIR_VERSION_SUFFIX}" STREQUAL "")
      list(APPEND DEFINITIONS "ONNX_MLIR_VERSION_SUFFIX=${ONNX_MLIR_VERSION_SUFFIX}")
    endif()
    set_property(SOURCE CompilerUtils.cpp APPEND PROPERTY COMPILE_DEFINITIONS ${DEFINITIONS})
  endif()
else()
  message(STATUS "onnx-mlir version        : not set")
endif()
```

whitneywhtsang: Good idea, I modified the code with this idea, but not exactly the same.

src/Compiler/CMakeLists.txt
gongsu832: User settable variables should all start with ONNX_MLIR so I'd use ONNX_MLIR_PRODUCT_VERSION or simply ONNX_MLIR_VERSION instead.

src/Compiler/CMakeLists.txt
```diff
@@ -66,10 +66,49 @@ add_onnx_mlir_library(CompilerPasses
...
+  message(DEBUG "PRODUCT_VERSION_MINOR = ${PRODUCT_VERSION_MINOR}")
+  message(DEBUG "PRODUCT_VERSION_PATCH = ${PRODUCT_VERSION_PATCH}")
+  message(DEBUG "PRODUCT_ID = ${PRODUCT_ID}")
+
+  # Major and product ID are mandatory, minor and patch are optional.
+  # Major must be greater than zero.
+  # When not specified, minor and patch are empty.
+  if (("${PRODUCT_VERSION_MAJOR}" STREQUAL "")
+	  OR ("${PRODUCT_ID}" STREQUAL "")
+	  OR (PRODUCT_VERSION_MAJOR LESS_EQUAL 0))
```
gongsu832, line 97: There is no need to check for PRODUCT_VERSION_MAJOR < 0. The regex match does not allow that.

src/Compiler/CMakeLists.txt
whitneywhtsang: Renamed to `ONNX_MLIR_PRODUCT_VERSION`.

src/Compiler/CMakeLists.txt
whitneywhtsang: Changed `PRODUCT_VERSION_MAJOR LESS_EQUAL 0` to `PRODUCT_VERSION_MAJOR EQUAL 0`.

gongsu832: 0 is a valid major I'd think?

whitneywhtsang: I am following the rules in https://github.com/onnx/onnx-mlir/pull/1409#discussion_r867151399:
> if PRODUCT_ID is set at least `PRODUCT_VERSION_MAJOR` must be set to a positive integer value different that 0

@etiotto is 0 a valid major?

sstamenova: It should be, lots of products ship as 0.0.x, or 0.x.y, etc.

etiotto: Yes 0 is a valid major IMO

whitneywhtsang: Changed to allow 0.

src/Compiler/CMakeLists.txt
AlexandreEichenberger: > User settable variables should all start with ONNX_MLIR 

That is very good practice, thanks for enforcing it.

sstamenova:
> @sstamenova @gongsu832 are you happy with the requested changes?

The changes look fine with the caveat that we still need a test for this.

whitneywhtsang:
> The changes look fine with the caveat that we still need a test for this.

Do you have any suggestions on how to create a test case without having default values?

I hesitate to have default values, as it will create 4 not useful metadata only used for testing purposes.

whitneywhtsang:
@sstamenova I have added a test case, please take another look, thanks.

test/mlir/driver/product.version.mlir
```diff
@@ -0,0 +1,16 @@
+// RUN: onnx-mlir --preserveBitcode %s -o %t
+// RUN: llvm-dis %t.bc -o %t.ll
+// RUN: cat %t.ll | FileCheck %s
+// XFAIL: onnx-mlir-product-version
```
sstamenova, line 4: I think this test should pass always. We don't need to verify that the product version is 0, but that it exists. Then you can skip both this XFAIL as well as the changes to the lit cfg files

whitneywhtsang: Good idea, changed.

sstamenova (approved):

PR #1416, opened 2022-05-10
better NNPA error msg

AlexandreEichenberger:
Better error message when NNPA is not compatible and/or running on a machine that does not have an NNPA.
Should add a requirement that NNPA is only active if `mcpu=z16`... but will add this after another PR that manipulates the active accelerator is in (#1414)


cjvolzka:
Overall I think the new message is a big improvement. Some small tweaks though:
```
Model is running on hardware that is not compatible with the zDNN library that this model was compiled for (version num 0x10000). Please check that the model is running on hardware with an integrated accelerator for AI
 (z16+) that supports the required zDNN library version.
```

Also is it possible to get the version to print the human readble string (ie `1.0.0`) vs `0x10000`

AlexandreEichenberger:
yes to both, let me work on it.

AlexandreEichenberger:
Error message when running a model compiled with nnpa and run on a non-nnpa machine
```
Model is running on hardware that is not compatible with the zDNN library that this model was compiled for (version num 1.0.0). Please check that the model is running on hardware with an integrated accelerator for AI (z16 +) that supports the required zDNN library version.
```
with the version num properly decoded.

cjvolzka (approved):
New message looks good to me. Thanks!

PR #1419, opened 2022-05-10
[FUSION]: Fuse ONNXConvOp fed by ONNXMulOp

etiotto:
Fuse an `ONNXMulOp` when one of the input is a the result of a `ONNXConvOp` and the other input is a produced by a `ONNXConstantOp` if the `ONNXConvOp weights is produced by a `ONXConstantOp`. For example: 

```
%0 = "onnx.Constant"() {value = dense<"...."> : tensor<64x3x7x7xf32>} : () -> tensor<64x3x7x7xf32>
%1 = "onnx.NoValue"() {value} : () -> none
%2 = "onnx.Conv"(%arg0, %0, %1) {...} : (tensor<1x3x224x224xf32>, tensor<64x3x7x7xf32>, none) -> tensor<*xf32>
%3 = "onnx.Constant"() {value = dense<[...]> : tensor<64xf32>} : () -> tensor<64xf32>
%4 = "onnx.UnsqueezeV11"(%3) {axes = [1, 2]} : (tensor<64xf32>) -> tensor<*xf32>
%5 = "onnx.Mul"(%2, %4) : (tensor<*xf32>, tensor<*xf32>) -> tensor<*xf32> 
```

Becomes:
```
 %1 = "onnx.Constant"() {value = dense<"NEW_W_VAL"> : tensor<64x3x7x7xf32>} : () -> tensor<64x3x7x7xf32>
%2 = "onnx.Conv"(%arg0, %1, %0) {auto_pad = "NOTSET", group = 1 : si64, kernel_shape = [7, 7], pads = [3, 3, 3, 3], strides = [2, 2]} : (tensor<1x3x224x224xf32>, tensor<64x3x7x7xf32>, none) -> tensor<1x64x112x112xf32>
```

Where `NEW_W_VAL` is computed by multiplying the original weight input of `onnx.Conv` and the constant input of the ONNXMulOp`.


src/Dialect/ONNX/Rewrite.td
```diff
@@ -263,6 +278,55 @@ def FuseAddConvPattern: Pat<
...
+
+def FuseMulConvNullBiasPattern: Pat<
+  (ONNXMulOp:$res
+    (ONNXConvOp
+       $x, $w, $b, $auto_pad, $dilation, $group, $kernel_shape, $pads, $strides),
+    (ONNXConstantOp:$y $_, $denseAttr, $_, $_, $_, $_, $_, $_)),
+  (ONNXConvOp
+     $x,
+     // new_w
+     (ONNXConstantOpFromDenseAttr (ScaleByAxis<1> $w, $denseAttr)),
```
tungld, line 311: I guess that `ScallByAxis` is to do multiplication of two constants. If so, it looks redundant. We can call something like`ONNXMulOp $w, (ONNXSqueeze $y, $axes)` and let `--constprop-onnx` do it job by folding `ONNXMulOp` and `ONNXSqueeze` into ONNXConstantOp, where `$axes` is rank(y) in order to add 1 as the last dimension.

src/Dialect/ONNX/Rewrite.td
```diff
@@ -117,6 +117,12 @@ def HaveSameLastDim: Constraint<
         "[$1.getType().cast<RankedTensorType>().getRank() - 1])">,
   "Two tensors have the same last dimension">;
 
+class HaveSameDim<int dim>: Constraint<
+  CPred<"onnx_mlir::hasShapeAndRank($0) && onnx_mlir::hasShapeAndRank($1) && "
+        "($0.getType().cast<RankedTensorType>().getShape()[" # dim # "] =="
+        "$1.getType().cast<RankedTensorType>().getShape()[" # dim # "])">,
```
tungld, line 123: Do we need to add a check for dynamic dimensions, i.e. `getShape()[dim] != -1`? In case of dynamic dims, we don't know dim sizes are equal or not.

tungld (changes requested):

src/Dialect/ONNX/Rewrite.td
etiotto: Added

src/Dialect/ONNX/Rewrite.td
etiotto: Done

etiotto:
@tungld please take a second look, I addressed your code review comments.

src/Transform/ONNX/ONNXOpTransformPass.cpp
```diff
@@ -135,6 +135,7 @@ void ONNXOpTransformPass::runOnOperation() {
     dynamicPM.addNestedPass<FuncOp>(onnx_mlir::createDecomposeONNXToONNXPass());
     dynamicPM.addPass(onnx_mlir::createShapeInferencePass());
     dynamicPM.addPass(mlir::createCanonicalizerPass());
+    dynamicPM.addPass(onnx_mlir::createShapeInferencePass());
```
whitneywhtsang, line 138: Why need to add shape inference pass here?

src/Dialect/ONNX/Rewrite.td
```diff
@@ -263,6 +280,55 @@ def FuseAddConvPattern: Pat<
...
+     // new_w
+     (ONNXMulOp $w, (ONNXUnsqueezeV11Op $y, (createArrayAttrOf(getRankOf $y)))),
+     // unchanged operands and attributes.
+     $b, $auto_pad, $dilation, $group, $kernel_shape, $pads, $strides),
+  [(HasNoneType $b),
+   (IsDenseElementsAttr:$denseAttr),
+   (IsFromONNXConstantOpWithDenseElementsAttr:$w),    
+   (HaveSameElementType $w, $y),       // multiplier and Conv weight must have the same element type.      
+   (HasRankGT<1> $w),                  // rank of $w must be at least 2.  
+   (RankXMinusRankYIs<1> $w, $y),      // rank($y) must be equal to rank(w)-1.
```
whitneywhtsang, line 321:
```suggestion
   (RankXMinusRankYIs<1> $w, $y),      // rank($y) must be equal to rank($w)-1.
```

src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -116,6 +110,13 @@ void ONNXAddOp::getCanonicalizationPatterns(
   results.insert<FuseAddConvNullBiasPattern>(context);
 }
 
+/// on the ONNXMulOp.
+void ONNXMulOp::getCanonicalizationPatterns(
+    RewritePatternSet &results, MLIRContext *context) {
+  results.insert<NormalizeAddPattern>(context);
```
whitneywhtsang, line 116: Do you mean `NormalizeMulPattern`?

src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -18,23 +18,17 @@
 
 #include "src/Dialect/ONNX/ONNXOps.hpp"
 #include "src/Dialect/ONNX/ONNXOpsHelper.hpp"
+#include <numeric>
```
whitneywhtsang, line 21: Why do we need to include numeric?

tungld (approved):
LGTM. Please address @whitneywhtsang's comments before merging. Thanks!

src/Transform/ONNX/ONNXOpTransformPass.cpp
etiotto: Because canonicalization can (and in this case does) introduce operators that might have unknown shape for their output. In other places in the code where canonicalization is called we also do shape inference before calling it.

src/Dialect/ONNX/Rewrite.cpp
etiotto: yes now that I reverted code that used std::accumulate we do not need it anymore, will remove.

whitneywhtsang (approved):

PR #1422, opened 2022-05-11
Fix Coding style and documentation

ramakota-helprack:
Removed unused variables from below operations(tiny-yolo op's).
Conv
MaxPoolSingleOut
Const
LeakyRelu
Padding


AlexandreEichenberger:
I assume that since this PR is close, there is no need to check anymore. Open a new one if you want to proceed. Thanks

PR #1424, opened 2022-05-12
Upgrade LLVM level to 9778ec057cf4 

whitneywhtsang:
Updating LLVM to https://github.com/llvm/llvm-project/commit/9778ec057cf4214241e4a970f3e764e3cf150181 causes a number of build issues.
This patch attempts to modify `onnx-mlir` to use LLVM at https://github.com/llvm/llvm-project/commit/9778ec057cf4214241e4a970f3e764e3cf150181.

Including the following changes:
- ODS Op verifier/parser/printer code blocks are deprecated.
- must specify either 'assemblyFormat' or 'hasCustomAssemblyFormat' when 'mnemonic' is set
- Rename `OpAsmParser::OperandType` to `OpAsmParser::UnresolvedOperand`
- Rename the "type" attribute to "function_type"
- [Simplify LoopLikeOpInterface](https://github.com/llvm/llvm-project/commit/e51652f9bf2e63497ed1e007c4a8c31c7ec08930)
- When a pass is defined in an anonymous namespace, it now requires an explicit TypeID after https://github.com/llvm/llvm-project/commit/5e50dd048e3a20cde5da5d7a754dfee775ef35d6.
- etc.

### DIFF
--- a/src/Transform/BundleMemoryPools.cpp
+++ b/src/Transform/BundleMemoryPools.cpp
@@ -50,14 +50,14 @@ typedef std::map<Block *, AlignmentToMemPool *> BlockToMemPool;
 //===----------------------------------------------------------------------===//
 
 /// Retrieve function which contains the current operation.
-ATTRIBUTE(unused) FuncOp getContainingFunction(memref::AllocOp op) {
+ATTRIBUTE(unused) func::FuncOp getContainingFunction(memref::AllocOp op) {
   Operation *parentFuncOp = op->getParentOp();
 
   // While parent is not a FuncOp and its cast to a FuncOp is null.
-  while (!llvm::dyn_cast_or_null<FuncOp>(parentFuncOp))
+  while (!llvm::dyn_cast_or_null<func::FuncOp>(parentFuncOp))
     parentFuncOp = parentFuncOp->getParentOp();
 
-  return cast<FuncOp>(parentFuncOp);
+  return cast<func::FuncOp>(parentFuncOp);
 }
 
 // Check if this value is an argument of one of the blocks nested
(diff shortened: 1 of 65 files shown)


src/Transform/BundleMemoryPools.cpp
```diff
@@ -518,13 +518,15 @@ class KrnlMoveConstantsUp : public OpRewritePattern<arith::ConstantOp> {
...
-class KrnlBundleMemoryPoolsPass
-    : public PassWrapper<KrnlBundleMemoryPoolsPass, OperationPass<FuncOp>> {
+class KrnlBundleMemoryPoolsPass : public PassWrapper<KrnlBundleMemoryPoolsPass,
+                                      OperationPass<func::FuncOp>> {
 
   BlockToMemPool blockToStaticPool;
   BlockToMemPool blockToDynamicPool;
 
 public:
+  MLIR_DEFINE_EXPLICIT_INTERNAL_INLINE_TYPE_ID(KrnlBundleMemoryPoolsPass)
```
AlexandreEichenberger, line 528: Could you just leave a comment to state what that macro is doing? I have seen it elsewhere added, might be useful to state how to add it elsewhere in the future when it will be needed

AlexandreEichenberger (approved):
LGTM, might be useful to add a small comments to some changes, otherwise its good, big thanks

src/Transform/BundleMemoryPools.cpp
whitneywhtsang: When a pass is defined in an anonymous namespace, it requires an explicit TypeID.
For example, `MLIR_DEFINE_EXPLICIT_INTERNAL_INLINE_TYPE_ID` should be placed within a public section of the declaration of the class.
Details in https://github.com/llvm/llvm-project/commit/5e50dd048e3a20cde5da5d7a754dfee775ef35d6.

Issue #1417, opened 2022-05-10
Failure codes for run_main_graph


AlexandreEichenberger:
Currently, we have no way to recover from an error in `run_main_graph`. 

So when an application calls `run_main_graph`, the call either succeeds or it abort the program. That may not be the most friendly behavior.

The function `run_main_graph` returns an `OMTensorList`, so it is not easy to add an `int` return code. However, we can easily repurpose some data inside the `OMTensorList` to hold an error code. For example, the `_owning` is declared as an `int64_t` but really need only 1 bit. So one could easily hide in the data structure an error code. We could then add a new call 'OMfailure(OMTensorList *list)` which returns the error code associated with the list. We can also add an "internal" function that sets the error code. 

Does that sounds reasonable? This feature would not break backward compatibility, except for the fact that with this new approach, failures would no longer results in asserts.

Initially, we could have an error for "zdnn mismatch"; midterms we could also have errors for failure to malloc, for example, or input mismatch.

gongsu832:
In the C world for a function that returns a pointer, the common way for error reporting is to return NULL and set the thread-local global `errno`.

AlexandreEichenberger:
Idea is that `run_main_graph` would return NULL (normally the address of the output tensor list) as well as set the thread-private errno. Here are possible uses of existing errno return values (posix). Several may fit the type of errors that we may want to return. The value that seems to fit best is indicated with a star. 

There are discussion if returning ERRNO is appropriate or if a custom variable with custom values is more appropriate or not.
https://stackoverflow.com/questions/11699596/how-to-set-errno-value

run out of memory:

ENOBUFS  No buffer space available (POSIX.1 (XSI STREAMS option)).
* ENOMEM Not enough space/cannot allocate memory (POSIX.1-2001).

libzdnn incompatible (typically mean that the compile version and the hardware version of the zdnn/NNPA are not compatible)

ELIBACC  Cannot access a needed shared library.
ENODEV No such device (POSIX.1-2001).
ENOTSUP  Operation not supported (POSIX.1-2001).
* EPERM  Operation not permitted (POSIX.1-2001).

input parameter mismatch between compiled model and runtime provided parameter list

E2BIG  Argument list too long (POSIX.1-2001).
* EDOM   Mathematics argument out of domain of function (POSIX.1   C99).

AlexandreEichenberger:
I asked for feedback on applicability to zos. How about Window @sstamenova ?

gongsu832:
[...]

PR #1426, opened 2022-05-12
Rt error reporting

AlexandreEichenberger:
Use errno interface for failure encountered during inference evaluation. See Issue #1417 for discussions.
First version only generate an error when there is a mismatch of versions for the NNPA

### DIFF
--- a/src/Accelerators/NNPA/Runtime/OMRuntimeNNPA.c
+++ b/src/Accelerators/NNPA/Runtime/OMRuntimeNNPA.c
@@ -122,14 +122,21 @@ void OMInitAccelNNPA() {
  *  are incompatible, an ERRNO=EPERM (Operation not permitted POSIX.1-2001) is
  *  set. The caller context is responsible to return NULL in the caller function
  *  so that users would know to check ERRNO to learn about why the call failed.
+ *
+ *  The assumption here is that all models are compiled for the same NNPA. If
+ *  the compatibility succeeds once, it is assumed that all subsequent models
+ *  were all compiled for the same NNPA version. However, if the compatibility
+ *  fails during the first invocation of a model, the function will repetitively
+ *  check if new models are compatible, in case the models were recompiled.
+ *
  *  Function name must be OMInitAccelX where X=NNPA.
  *  @param versionNum Version of the library/Architecture level that models
  *  were compiled for.
  *  @return true for compatible versions or false on incompatible versions.
  */
 uint64_t OMInitCompatibleAccelNNPA(uint64_t versionNum) {
+  static int isCompatible = 0;
   if (!OMIsInitAccelNNPA) {
-    int isCompatible = 1;
     /* Grab mutex. */
     pthread_mutex_lock(&OMMutexForInitShutdownNNPA);
     /* Test again in the mutex to see if accelerator is not initialized. */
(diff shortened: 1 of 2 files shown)


AlexandreEichenberger:
@tungld the change in the interface is that `uint64_t OMInitCompatibleAccelNNPA` now returns an int. If true, we are fine, if false, the run_main_graph needs to return NULL... I think this should be fairly straightforward.

tungld:
> @tungld the change in the interface is that uint64_t OMInitCompatibleAccelNNPA now returns an int. If true, we are fine, if false, the run_main_graph needs to return NULL... I think this should be fairly straightforward.

Oh, I see. May I edit this patch directly to just change the return type of OMInitCompatibleAccelNNPA from `void` to `uint64_t`, it should be a few lines. Then this patch can go first. I will create another PR to follow this patch to handle how `run_main_graph` returns NULL.

tungld:
Somehow I cannot push commit to your branch directly though It seems I have permission. So, I push the changes here that update the signature of `OMInitCompatibleAccelX` at LLVM level.
```diff
diff --git a/src/Conversion/KrnlToLLVM/KrnlEntryPoint.cpp b/src/Conversion/KrnlToLLVM/KrnlEntryPoint.cpp
index 25d356c9..671b4e04 100644
--- a/src/Conversion/KrnlToLLVM/KrnlEntryPoint.cpp
+++ b/src/Conversion/KrnlToLLVM/KrnlEntryPoint.cpp
@@ -86,7 +86,7 @@ public:

     // Emit code to initialize accelerators by calling OMInitCompatibleAccelX
     // where X is the accelerator name.
-    // OMInitCompatibleAccelX's signature is `void (i64)`.
+    // OMInitCompatibleAccelX's signature is `i64 (i64)`.
     if (Attribute maccelAttr =
             module->getAttrOfType<::mlir::Attribute>("onnx-mlir.accels")) {
       assert(
@@ -103,7 +103,7 @@ public:
         LLVM::ConstantOp versionNumberVal = rewriter.create<LLVM::ConstantOp>(
             loc, int64Ty, rewriter.getI64IntegerAttr(versionNumberInHex));
         rewriter.create<LLVM::CallOp>(
-            loc, TypeRange(), funcRef, ArrayRef<Value>({versionNumberVal}));
+            loc, int64Ty, funcRef, ArrayRef<Value>({versionNumberVal}));
       }
     }

@@ -380,14 +380,14 @@ private:
   FlatSymbolRefAttr getOrInsertOMInitCompatibleAccel(
       PatternRewriter &rewriter, ModuleOp module, StringRef accelName) const {
     std::string funcName = "OMInitCompatibleAccel" + accelName.str();
-    // OMInitCompatibleAccelX's signature is `void (i64)`.
+    // OMInitCompatibleAccelX's signature is `i64 (i64)`.
     auto func = module.lookupSymbol<LLVM::LLVMFuncOp>(funcName);
     MLIRContext *ctx = rewriter.getContext();
     if (!func) {
       OpBuilder::InsertionGuard guard(rewriter);
       rewriter.setInsertionPointToStart(module.getBody());
       LLVM::LLVMFunctionType funcType =
-          LLVM::LLVMFunctionType::get(LLVM::LLVMVoidType::get(ctx),
+          LLVM::LLVMFunctionType::get(IntegerType::get(ctx, 64),
               ArrayRef<mlir::Type>({IntegerType::get(ctx, 64)}),
               /*isVarArg=*/false);
       func = rewriter.create<LLVM::LLVMFuncOp>(
```

and 
```diff
diff --git a/test/mlir/krnl/entry_point.mlir b/test/mlir/krnl/entry_point.mlir
index adf61936..697c64e3 100644
--- a/test/mlir/krnl/entry_point.mlir
+++ b/test/mlir/krnl/entry_point.mlir
@@ -242,7 +242,7 @@ module attributes {"onnx-mlir.accels" = ["Pseudo-0x10001", "NNPA-0x10000"]} {
 // CHECK: llvm.func @OMInitCompatibleAccelPseudo(i64)
 // CHECK-LABEL: llvm.func @run_main_graph({{.*}}: !llvm.ptr<i8>) -> !llvm.ptr<i8> {
 // CHECK-NEXT: [[VERSION_NUMBER_1:%.+]] = llvm.mlir.constant(65537 : i64) : i64
-// CHECK-NEXT: llvm.call @OMInitCompatibleAccelPseudo([[VERSION_NUMBER_1]]) : (i64) -> ()
+// CHECK-NEXT: llvm.call @OMInitCompatibleAccelPseudo([[VERSION_NUMBER_1]]) : (i64) -> i64
 // CHECK-NEXT: [[VERSION_NUMBER_2:%.+]] = llvm.mlir.constant(65536 : i64) : i64
-// CHECK-NEXT: llvm.call @OMInitCompatibleAccelNNPA([[VERSION_NUMBER_2]]) : (i64) -> ()
+// CHECK-NEXT: llvm.call @OMInitCompatibleAccelNNPA([[VERSION_NUMBER_2]]) : (i64) -> i64
 }
```

AlexandreEichenberger:
migrated to https://github.com/onnx/onnx-mlir/pull/1428

PR #1428, opened 2022-05-13
Edit entry points to return NULL if failed to initialize accelerators

tungld:
This patch edits the generated entry points, e.g. `run_main_graph` to return NULL when initializing an accelerator fails.

This patch based on PR #1426 (most of the changes are from #1426), so #1426 should be merged first.


AlexandreEichenberger (approved):
LGTM, I was planning some further changes in the execution sessions, will build on top of this one.

PR #1430, opened 2022-05-16
Fix RuntimeNNPA and zdnn dependency order

gongsu832:
- fix ordering of RuntimeNNPA and zdnn when calling addCompilerConfig
- don't use set_union since it sorts the vector


AlexandreEichenberger (approved):
LGTM, thanks @gongsu832

PR #1433, opened 2022-05-17
Use wget to download models in the onnx model zoo

tungld:
The current script `CheckONNXModelZoo.py` uses `git lfs` to checkout models in the model zoo, which does not support parallelism.

This patch uses `wget` instead to download models so that we can download and check multiple models at the same time.

Some other improvements in `CheckONNXModelZoo.py`:
- allow checking segfault error.
- exclude integer models since onnx-mlir does not support quantized models.

On my Z machine, it takes about 17 minutes to check 124 models in the model zoo when using 16 processes, e.g. `python CheckONNXModelZoo.py -njob 16 -compile_args="-O3 --mcpu=z14"`. FYI, this is the result: https://github.com/onnx/onnx-mlir/issues/128#issuecomment-1128755672


test/onnx-model-zoo/CheckONNXModelZoo.py
```diff
@@ -126,18 +156,16 @@ def obtain_all_model_paths():
...
-            " models are deprecated (using very old opsets, e.g. <= 3)"
-    log_l1("# There are {} models in the ONNX model zoo {}".format(
-        len(model_paths), deprecated_msg))
-    log_l1("See https://github.com/onnx/models/pull/389",
-           "for a list of deprecated models\n")
+    excluded_msg = ""
+    if (len(excluded_names) != 0):
+        excluded_msg = "where " + \
+            str(len(excluded_names)) + \
+            " models are not checked because of old opsets or quantization"
```
gongsu832, line 166: Instead of line continuation, it's probably better to use parentheses for multi-line expressions:
```
excluded_msg = ("where " +
    str(len(excluded_names)) +
    " models are not checked because of old opsets or quantization")
```

tungld: Thanks for reminding this! It's better.

tungld:
@gongsu832 @chentong319 any comments on this? Thanks!

gongsu832 (approved):
LGTM

PR #1436, opened 2022-05-17
Allow Xopt, Xllc to accept a comma separates list of options 

etiotto:
**Rationale**
A developer might want to pass to `opt` or `llc` more than one options. For example, one might want to pass to `llc` the options `-debug-entry-values` and `--debugger-tune=gdb` in order to instruct the LLVM backend to emit debug info tuned for a particular debugger (e.g. gdb) and enable debug information for the program entry values.  

This PR changes the option processing for -Xopt and -Xllc to allow passing a comma separates list of options to be passed to `opt` or `llc` respectively.


src/Compiler/CompilerOptions.cpp
```diff
@@ -106,12 +106,14 @@ llvm::cl::opt<bool> VerboseOutput("v", llvm::cl::desc("Use verbose output"),
 llvm::cl::opt<std::string> Xopt("Xopt",
     llvm::cl::desc("Arguments to forward to LLVM's 'opt' option processing"),
     llvm::cl::value_desc("A valid LLVM's 'opt' option"),
-    llvm::cl::cat(OnnxMlirOptions), llvm::cl::Hidden, llvm::cl::ValueRequired);
+    llvm::cl::cat(OnnxMlirOptions), llvm::cl::Hidden, llvm::cl::ValueRequired,
+    llvm::cl::OneOrMore);
```
doru1004, line 110: Does one or more mean that at least one is required?

AlexandreEichenberger: Please explain the rational. Does it add to, or replace?

sstamenova:
What would be the scenario when we want to use xopt of xllc more than once?

src/Compiler/CompilerOptions.cpp
whitneywhtsang: It exists:
```
.. _cl::ZeroOrMore:

* The **cl::ZeroOrMore** modifier (which is the default for the `cl::list`_
  class) indicates that your program will allow the option to be specified zero
  or more times.
  ```

**Comment by etiotto** [comment]
> What would be the scenario when we want to use xopt of xllc more than once?

@sstamenova I added the rationale in the description of the PR.

File: src/Compiler/CompilerOptions.cpp [hunk 2]
```diff
@@ -277,14 +279,38 @@ std::string getOptimizationLevelOption() {
 }
 
 // Support for Xopt.
-void setXoptOption(const std::string &flag) { Xopt = flag; }
+void setXoptOption(const std::vector<std::string> &flags) {
```
AlexandreEichenberger, line 282: @etiotto I wonder if you don't need to reset the list when setting the option. Because using the OnnxMlirCompiler interface, you can have a long running process that need to change the options as it goes. For maccel, we have the reset option; we could add a new set of call that reset a specific option (add a new call in the OnnxMlirCompiler interface, and then remove the RESET for maccel), or reset every time in the new list that you did here. I would prefer a "reset" call, it seems cleaner and then we can also remove the RESET option

src/Compiler/CompilerOptions.cpp
```diff
@@ -341,10 +367,26 @@ std::string getCompilerOption(const OptionKind kind) {
     return getTargetAccel();
   case OptionKind::CompilerOptLevel:
     return getOptimizationLevelOption();
-  case OptionKind::OPTFlag:
-    return getXoptOption();
-  case OptionKind::LLCFlag:
-    return getXllcOption();
+  case OptionKind::OPTFlag: {
+    std::vector<std::string> flags = getXoptOption();
+    std::string s;
```
AlexandreEichenberger, line 372: nit: maybe a stream string would be cleaner.

src/Compiler/CompilerOptions.cpp
```diff
@@ -341,10 +367,26 @@ std::string getCompilerOption(const OptionKind kind) {
...
+    for (int i = 0, n = flags.size(); i < n; ++i) {
+      s.append(flags.at(i));
+      if (i != n - 1)
+        s.append(",");
+    }
+    return s;
+  }
+  case OptionKind::LLCFlag: {
```
sstamenova, line 380: There's a bit of code duplication here that I think could be reduced. Something like:

```
case LLC:
case OPT:
    std::vector flags = LLC ? getLLC : get OPT;
    ...
    
```

sstamenova (approved):

etiotto:
@AlexandreEichenberger I added functions to clear each of the options that can be set via the compiler interface. I'd appreciate your review before I merge this in.

src/Compiler/CompilerOptions.cpp
```diff
@@ -329,6 +369,36 @@ int setCompilerOption(const OptionKind kind, const std::string &val) {
...
+    break;
+  case OptionKind::TargetCPU:
+    clearTargetCPU();
+    break;
+  case OptionKind::TargetAccel:
+    clearTargetAccel();
+    break;
+  case OptionKind::CompilerOptLevel: {
+    clearOptLevel();
+  } break;
```
whitneywhtsang, line 388:
```suggestion
  case OptionKind::CompilerOptLevel:
    clearOptLevel();
    break;
```

AlexandreEichenberger (approved):
LGTM, the interface is quite clear and streamlined, thanks for the update, much improved.

AlexandreEichenberger:
@etiotto do you want to make the few nits to put it down? Tx

src/Compiler/CompilerOptions.cpp
etiotto: done

PR #1438, opened 2022-05-17
Move onnx-mlir version to 0.3.0

whitneywhtsang:
Move `onnx-mlir` version to `0.3.0` from `0.2.0`. `1.0.0` was a mistake.

Signed-off-by: Whitney Tsang <whitneyt@ca.ibm.com>

PR #1440, opened 2022-05-18
Add extra version printer with onnx-mlir full version

whitneywhtsang:
The goal of this PR is to add onnx-mlir version to output of `onnx-mlir --version`.
There are two potential ways to achieve this goal. 
1. `llvm::cl::AddExtraVersionPrinter`: Add extra output to the end of existing version output.
2. `llvm::cl::SetVersionPrinter`: Define a customized printer output.

This PR uses method 1.

For example:
Before this PR:
```
> ./Debug/bin/onnx-mlir --version
LLVM (http://llvm.org/):
  LLVM version 15.0.0git
  Optimized build with assertions.
  Default target: powerpc64le-unknown-linux-gnu
  Host CPU: pwr8
```
After this PR:
```
> ./Debug/bin/onnx-mlir --version
LLVM (http://llvm.org/):
  LLVM version 15.0.0git
  Optimized build with assertions.
  Default target: powerpc64le-unknown-linux-gnu
  Host CPU: pwr8

onnx-mlir version 1.0.0 (git@github.com:whitneywhtsang/onnx-mlir.git 7c494e9061fc59aa6fc61550540fd2141431d8eb, https://github.com/llvm/llvm-project.git 9778ec057cf4214241e4a970f3e764e3cf150181)
```


cjvolzka:
Would it be possible to do something like this or something similar? Most are optional except I'd greatly prefer to drop the git URLs.

- Move onnx-mlir to the top
- Make the onnx-mlir section consistent with LLVM section (ie homepage URL, indenting)
- Only print the SHA and not the whole URL
- Put the llvm sha next to the LLVM version
- Only print the short sha rather than the long one

Note: Where did the 1.0.0 come from? Isn't the current onnx-mlir 0.2.0 w/ 0.3.0 to be cut soon?

```
onnx-mlir (http://onnx.ai/onnx-mlir/)
  version 0.3.0 (7c494)
LLVM (http://llvm.org/):
  LLVM version 15.0.0git (9778e)
  Optimized build with assertions.
  Default target: powerpc64le-unknown-linux-gnu
  Host CPU: pwr8
```

whitneywhtsang:
> Where did the 1.0.0 come from? Isn't the current onnx-mlir 0.2.0 w/ 0.3.0 to be cut soon?

https://github.com/onnx/onnx-mlir/pull/1438 is to fix the version number.

src/Compiler/CompilerUtils.cpp
```diff
@@ -251,6 +237,24 @@ struct Command {
 // =============================================================================
 // Methods for compiling and file processing.
 
+std::string getOnnxMlirFullVersion(bool simplify) {
+  const std::string OnnxMlirVersion = "onnx-mlir version 1.0.0";
+#ifdef ONNX_MLIR_VENDOR
+  return ONNX_MLIR_VENDOR ", " + OnnxMlirVersion;
+#elif defined(ONNX_MLIR_REPOSITORY) && defined(ONNX_MLIR_REVISION) &&          \
+    defined(LLVM_REPOSITORY) && defined(LLVM_REVISION)
+  if (simplify) {
```
whitneywhtsang, line 246: Thinking to either
1. create another cmake variable so vendor can decide to always use the simplified version instead of a boolean for only this one callsite. 
2. always use the simplified version, but that become inconsistent with how llvm does it. 

Please let me know how you think.

cjvolzka:
>I addressed some of the suggestions by adding a simplified version of the onnx-mlir version string.
It now only print the short sha rather than the long one and the git URLs are dropped.
>
>If we want to do any other suggestions, then we need to write our own version printer, i.e., using llvm::cl::SetVersionPrinter instead of llvm::cl::AddExtraVersionPrinter. The disadvantage is duplication of code. Let's see what other thinks before I make that change.

Only printing the short SHA w/o the URL was the primary concern. The others were "would be nice if easy" kind of suggestions. I wouldn't go too deep trying to cover those.


> Thinking to either

> 1. create another cmake variable so vendor can decide to always use the simplified version instead of a boolean for only this one callsite.
> 2. always use the simplified version, but that become inconsistent with how llvm does it.
Please let me know what you think.

How does it look mixing the simplified onnx-mlir version w/ the non-simplified LLVM. Unless something really jumps out I'd lean toward 2. My off the cuff thought is no one will likely  use the additional cmake variable so maybe just wait until someone asks for such a feature.

whitneywhtsang:
The output from `clang --version` includes the repo and the long SHA:
```
> clang --version                                                                                         
clang version 15.0.0 (https://github.com/llvm/llvm-project.git 9778ec057cf4214241e4a970f3e764e3cf150181)
Target: powerpc64le-unknown-linux-gnu
Thread model: posix
InstalledDir: <path to installed dir>
```

> How does it look mixing the simplified onnx-mlir version w/ the non-simplified LLVM. Unless something really jumps out I'd lean toward 2. My off the cuff thought is no one will likely use the additional cmake variable so maybe just wait until someone asks for such a feature.

What do you mean by mixing the simplified onnx-mlir version w/ the non-simplified LLVM? 
Something like: `onnx-mlir version 1.0.0 (onnx-mlir fd6a145f7702, https://github.com/llvm/llvm-project.git 9778ec057cf4214241e4a970f3e764e3cf150181)`?
I don't think we should do that. IMO, we should have either 
- `onnx-mlir version 1.0.0 (git@github.com:whitneywhtsang/onnx-mlir.git fd6a145f7702f3a1789cd6923890efd55f67cf64, https://github.com/llvm/llvm-project.git 9778ec057cf4214241e4a970f3e764e3cf150181)`
  - consistent with `clang --version`.
- or `onnx-mlir version 1.0.0 (onnx-mlir fd6a145f7702, llvm-project 9778ec057cf4)`.
  - idea suggested by @cjvolzka

src/Compiler/CompilerUtils.cpp
```diff
@@ -251,6 +237,27 @@ struct Command {
...
 // Methods for compiling and file processing.
 
+std::string getOnnxMlirFullVersion() {
+  const std::string OnnxMlirVersion = "onnx-mlir version 1.0.0";
+  return
+#ifdef ONNX_MLIR_VENDOR
+      ONNX_MLIR_VENDOR ", " + OnnxMlirVersion;
+#elif defined(ONNX_MLIR_REPOSITORY) && defined(ONNX_MLIR_REVISION) &&          \
+    defined(LLVM_REPOSITORY) && defined(LLVM_REVISION)
+  // Intended for code review, and will remove one option before merging.
```
whitneywhtsang, line 247: Let's decide if we want to simplify the version output to 
1. no repository and short revision
2. or, repository and full revision

sstamenova:
Please make sure to update the description of the PR when you make changes to how the version is printed out, so that it matches what's in the code.

src/Compiler/CompilerUtils.cpp
whitneywhtsang: @cjvolzka Do you agree to use `-DPACKAGE_VENDOR` for no repository and short version?

tungld: +1 for `onnx-mlir version 1.0.0 (onnx-mlir fd6a145f7702, llvm-project 9778ec057cf4)` which includes LLVM and is clean and neat.

gongsu832: +2 for option (1)

etiotto: I agree with @sstamenova and I am also in favor with being consistent with what LLVM does (given also we print out the LLVM commit already).

whitneywhtsang: During offline discussion with @gongsu832, we noticed that clang sometimes doesn't output both the repo url and hash, e.g., 
```
clang version 15.0.0
Target: powerpc64le-unknown-linux-gnu
Thread model: posix
InstalledDir: <path to installed dir>
```
It appears to be controlled by `LLVM_APPEND_VC_REV`, I have created PR https://github.com/onnx/onnx-mlir/pull/1446 to add a similar logic. 
@gongsu832 was amenable of using option (2).

tungld (approved):
LGTM.

tungld:
@sstamenova could you please help see what happened with Windows CI? Thanks!

whitneywhtsang:
@sstamenova I am seeing MLIR-Windows-CI failed with `TypeError: Descriptors cannot not be created directly.`, which seems to be unrelated to my change. Would you be able to let us know what is happening? Thanks

sstamenova:
> @sstamenova could you please help see what happened with Windows CI? Thanks!

It looks like the default windows image in the pool now includes a version of protobuf that doesn't work correctly for the rest of the build. I'll have a look and see if I can figure out what the right fix is.

sstamenova:
@whitneywhtsang : Part of the feedback was to move the onnx-mlir version to the top? I think that's the correct way as well. Have you done it? If so, please update the description to reflect that. If not, is there a reason not to?

src/Compiler/CompilerUtils.cpp
```diff
@@ -251,6 +237,20 @@ struct Command {
 // =============================================================================
 // Methods for compiling and file processing.
 
+std::string getOnnxMlirFullVersion() {
+  const std::string OnnxMlirVersion = "onnx-mlir version 0.3.0";
+  return
+#ifdef ONNX_MLIR_VENDOR
```
sstamenova, line 243: This has gotten fairly complicated at this point. I would consider having defaults for everything in order to simplify the logic and be able to test it in all cases.

test/mlir/driver/version.mlir
```diff
@@ -0,0 +1,4 @@
+// RUN: onnx-mlir --version | FileCheck %s
+// XFAIL: onnx-mlir-vendor
+
+// CHECK: onnx-mlir version 0.3.0 ({{.*}}/onnx-mlir{{.*}} {{[a-zA-Z0-9]+}}, {{.*}}/llvm-project{{.*}} {{[a-zA-Z0-9]+}})
```
sstamenova, line 4: Don't hardcode the version here please, it will require the test to be updated every time the version changes.

sstamenova (changes requested):

whitneywhtsang:
> @whitneywhtsang : Part of the feedback was to move the onnx-mlir version to the top? I think that's the correct way as well. Have you done it? If so, please update the description to reflect that. If not, is there a reason not to?

No, as mentioned in https://github.com/onnx/onnx-mlir/pull/1440#issuecomment-1130174408, if we want to do that, then we need to write our own version printer, i.e., using llvm::cl::SetVersionPrinter instead of llvm::cl::AddExtraVersionPrinter. The disadvantage is duplication of code. And I was waiting to see if other reviewers comment on that. 

An other idea is to simply have `onnx-mlir version 0.3.0 (git@github.com:onnx/onnx-mlir.git fd6a145f7702f3a1789cd6923890efd55f67cf64, git@github.com:llvm/llvm-project.git 9778ec057cf4214241e4a970f3e764e3cf150181)` as the whole output from `--version`. What's your opinion on that?

src/Compiler/CompilerUtils.cpp
whitneywhtsang: By default, `getOnnxMlirFullVersion()` returns `onnx-mlir version 0.3.0`.
If `ONNX_MLIR_VENDOR`, then it returns `ONNX_MLIR_VENDOR, onnx-mlir version 0.3.0`.
If not `ONNX_MLIR_VENDOR` and both repo and revision are defined, then it returns e.g., `onnx-mlir version 0.3.0 (git@github.com:onnx/onnx-mlir.git fd6a145f7702f3a1789cd6923890efd55f67cf64, git@github.com:llvm/llvm-project.git 9778ec057cf4214241e4a970f3e764e3cf150181)`.
This output logic is similar to LLVM. 

The logic actually doesn't change in this PR. Do you want to elaborate on your idea about default values? 
If we have a default value for `ONNX_MLIR_VENDOR`, then we will never generate repo and revision.
If we first check if `ONNX_MLIR_VENDOR` is the default value, then the logic is not simplified.
Similarly, if we have default values for repo and revision, then we don't want to output them if they are the default value, unless we want to output e.g., `onnx-mlir version 0.3.0 (unknown onnx-mlir repo, unknown onnx-mlir revision, unknown llvm repo, unknown llvm revision)`, which is first different from LLVM, and doesn't look good.

test/mlir/driver/version.mlir
whitneywhtsang: Done.

whitneywhtsang (commented):
@sstamenova Thanks for your review comments. Please have another look.

AlexandreEichenberger (changes requested):
@whitneywhtsang this is very useful, and I would suggest that also reporting the ONNX standard that onnx-mlir is parsing would be quite useful. Would you mind adding that line of info. Not sure if we can ferret it out of the third_party/onnx or simply add one more var that we can update when we (quite rarely) update the ONNX opset level.

sstamenova:
> > @whitneywhtsang : Part of the feedback was to move the onnx-mlir version to the top? I think that's the correct way as well. Have you done it? If so, please update the description to reflect that. If not, is there a reason not to?
> 
> No, as mentioned in [#1440 (comment)](https://github.com/onnx/onnx-mlir/pull/1440#issuecomment-1130174408), if we want to do that, then we need to write our own version printer, i.e., using `llvm::cl::SetVersionPrinter` instead of `llvm::cl::AddExtraVersionPrinter`. The disadvantage is duplication of code. And I was waiting to see if other reviewers comment on that.
> 
> An other idea is to simply have `onnx-mlir version 0.3.0 (git@github.com:onnx/onnx-mlir.git fd6a145f7702f3a1789cd6923890efd55f67cf64, git@github.com:llvm/llvm-project.git 9778ec057cf4214241e4a970f3e764e3cf150181)` as the whole output from `--version`. What's your opinion on that?

I think it makes sense to have the fully "qualified" version information (such as optimized build etc) as well as a short version option. It's also much cleaner to have the onnx-mlir version first, so I agree with @cjvolzka that it should come first.

src/Compiler/CompilerUtils.cpp
gongsu832: The purpose of a version is for the users to report it for identifying the code involved. Yes a repo and revision would certainly identify the code but these are "internal" matter which the users don't care about and are hard to remember. A short released version number is typically tagged to a commit hash so it WILL uniquely identify the code and at the same time is more user friendly. That's why you typically never see repo and revision reported in the version for released code.

etiotto: What I see is that this function was simply moved from line 156. The only functional change is the removal of the storage specifier `static`, which is to allow other compilation units to call this function. So looks fine from my perspective.

etiotto (approved):
LGTM, considering also PR #1446 addresses the question about having the commit hash or not. 
@sstamenova is this fine ?

sstamenova:
> LGTM, considering also PR #1446 addresses the question about having the commit hash or not. @sstamenova is this fine ?

I don't think either of these PRs should go in. I went back and had a look again and between the two of them, the logic for how to include which version and when is getting fairly complicated. This is confusing and also inconsistent with how the llvm versions work. We don't have to follow the llvm paradigm, but we need a clear solution that makes sense.

I think the correct approach is to recall *both* PRs, collect the feedback on what the version should look like and when (there was lots of good feedback on both PRs), outline a single proposal (as an 'Issue' perhaps) on how to update the version (and corresponding tests) and commit the changes in a single PR. This way we won't end up with inconsistencies and confusion and there will be a single source of truth on what the plan is.

src/Compiler/CompilerUtils.cpp
```diff
@@ -251,6 +237,20 @@ struct Command {
 // =============================================================================
 // Methods for compiling and file processing.
 
+std::string getOnnxMlirFullVersion() {
+  const std::string OnnxMlirVersion = "onnx-mlir version 0.3.0";
```
sstamenova, line 241: If I recall correctly there was a separate change committed recently that added the option to specify the onnx-mlir version. I would not expect to see hardcoded versions here any more, was this missed last time?

test/mlir/driver/version.mlir
```diff
@@ -0,0 +1,4 @@
+// RUN: onnx-mlir --version | FileCheck %s
+// XFAIL: onnx-mlir-vendor
```
sstamenova, line 2: The version is printed out (or *a* version anyway) regardless of whether vendor is specified. There should be a test for that.

src/onnx-mlir.cpp
```diff
@@ -54,6 +54,9 @@ int main(int argc, char *argv[]) {
   mlir::registerPassManagerCLOptions();
   mlir::registerDefaultTimingManagerCLOptions();
 
+  llvm::cl::AddExtraVersionPrinter(
```
sstamenova, line 57: All of the llvm tools specify their versions first before adding llvm. I looked at the code in more detail and the duplication is very little, so I think the right approach is to follow that pattern and have onnx-mlir version than llvm version similar to how clang, lldb, etc. do it.

sstamenova (changes requested):

src/Compiler/CompilerUtils.cpp
whitneywhtsang: I don't recall such change.

src/onnx-mlir.cpp
whitneywhtsang: ok, I can duplicate some code, and have onnx-mlir show up first.

src/Compiler/CompilerUtils.cpp
sstamenova: https://github.com/onnx/onnx-mlir/pull/1409

whitneywhtsang: Product version is different from onnx-mlir version.

sstamenova: How so? I would expect that they are the same - when would they be different?

This is why we need a single proposal - there are too many changes impacting how the version is printed, what the version is, etc. and it is hard to keep track of what they are and what the goal is. A single clear goal will reduce confusion significantly and will let you get buy off from everyone (or at least everyone will know what to expect).

whitneywhtsang: Yes, I am working on that.

onnx-mlir version is the version for the Open Source project onnx-mlir.
Product version is the version of the product a company may want to ship, where onnx-mlir is one of the components.

src/Version/Version.cpp
```diff
@@ -0,0 +1,140 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//===-------------------------- Version.cpp -------------------------------===//
```
whitneywhtsang, line 5: Referenced https://clang.llvm.org/doxygen/Version_8cpp_source.html.

src/Version/Version.hpp
```diff
@@ -0,0 +1,54 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//===-------------------------- Version.cpp -------------------------------===//
```
whitneywhtsang: Referenced https://clang.llvm.org/doxygen/Version_8h_source.html.

src/Version/Version.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+  std::string repo = getOnnxMlirFullRepositoryVersion(ToIncludeLLVM);
+  if (!repo.empty()) {
+    os << " " << repo;
+  }
+
+  return buf;
+}
+
+#if defined(__GNUC__)
```
whitneywhtsang, line 96: Copied from https://llvm.org/doxygen/CommandLine_8cpp_source.html (line 2439).

src/Version/Version.cpp
```diff
@@ -0,0 +1,140 @@
...
+#define LLVM_IS_DEBUG_BUILD 1
+#else
+#define LLVM_IS_DEBUG_BUILD 0
+#endif
+#else
+// Otherwise, for an unknown compiler, assume this is an optimized build.
+#define LLVM_IS_DEBUG_BUILD 0
+#endif
+
+void getVersionPrinter(llvm::raw_ostream &os) {
```
whitneywhtsang, line 118: Referenced https://llvm.org/doxygen/CommandLine_8cpp_source.html (line 2464).

src/Compiler/CompilerUtils.cpp
whitneywhtsang: https://github.com/onnx/onnx-mlir/issues/1460 is created for the discussion. @sstamenova Please let me know if it is more clear.

src/Compiler/CompilerUtils.cpp
whitneywhtsang: I have created a new file Version.cpp to make it more similar to LLVM implementation, and I have added link references to LLVM implementations as PR comments.

src/Version/Version.hpp
```diff
@@ -0,0 +1,54 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//===-------------------------- Version.hpp -------------------------------===//
```
whitneywhtsang, line 5: Referenced https://clang.llvm.org/doxygen/Version_8h_source.html.

whitneywhtsang:

whitneywhtsang:
@sstamenova Any further comments on https://github.com/onnx/onnx-mlir/issues/1460 or this PR? Thanks in advanced.

src/Version/Version.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+/// Retrieves the repository revision number (or identifier) from which
+/// LLVM was built.
+std::string getLLVMRevision();
+
+/// Retrieves the full repository version that is an amalgamation of
+/// the information in getOnnxMlirRepositoryPath() and  getOnnxMlirRevision().
+/// And getLLVMRepositoryPath() and getLLVMRevision() if \p
+/// ToIncludeLLVM.
+std::string getOnnxMlirFullRepositoryVersion(bool ToIncludeLLVM);
```
sstamenova, line 39: Here and elsewhere: parameters in the onnx-mlir source code start with lower case letters. Any reason to do this differently?

src/Version/Version.cpp
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  return buf;
+}
+
+#if defined(__GNUC__)
+// GCC and GCC-compatible compilers define __OPTIMIZE__ when optimizations are
+// enabled.
+#if defined(__OPTIMIZE__)
+#define LLVM_IS_DEBUG_BUILD 0
```
sstamenova, line 100: This should be ONNX_MLIR_IS_DEBUG_BUILD, right?

src/Version/Version.cpp
```diff
@@ -0,0 +1,138 @@
...
+#ifndef NDEBUG
+  os << " with assertions";
+#endif
+  std::string CPU = std::string(llvm::sys::getHostCPUName());
+  if (CPU == "generic")
+    CPU = "(unknown)";
+  os << ".\n";
+  os << "  Default target: " << llvm::sys::getDefaultTargetTriple() << '\n'
+     << "  Host CPU: " << CPU << '\n';
+  os << "  LLVM version " << LLVM_PACKAGE_VERSION << ' '
```
sstamenova, line 134: Can we put LLVM version before the rest of the properties? Right now we have onnx-mlir and onnx version at the top, a few other options and then llvm version.

src/Version/CMakeLists.txt
```diff
@@ -53,4 +53,24 @@ set_source_files_properties("${version_inc}"
              HEADER_FILE_ONLY TRUE
   )
 
-add_custom_target(OMVersion ALL DEPENDS ${version_inc})
+if (NOT DEFINED ONNX_MLIR_VERSION)
+  set(ONNX_MLIR_VERSION "0.3.0")
+endif()
+list(APPEND DEFINITIONS "ONNX_MLIR_VERSION=\"${ONNX_MLIR_VERSION}\"")
+if (NOT DEFINED ONNX_VERSION)
+  set(ONNX_VERSION "1.11.0")
```
sstamenova, line 61: Can't the default be set based on what's in the submodule instead of having to be explicitly specified here? It seems like this would be prone to error.

src/Version/CMakeLists.txt
```diff
@@ -53,4 +53,24 @@ set_source_files_properties("${version_inc}"
              HEADER_FILE_ONLY TRUE
   )
 
-add_custom_target(OMVersion ALL DEPENDS ${version_inc})
+if (NOT DEFINED ONNX_MLIR_VERSION)
+  set(ONNX_MLIR_VERSION "0.3.0")
```
sstamenova, line 57: It would be cleaner if this was defined in a separate file that is easy to track down and modify, but this is still an improvement.

src/Version/CMakeLists.txt
```diff
@@ -53,4 +53,24 @@ set_source_files_properties("${version_inc}"
...
+list(APPEND DEFINITIONS "ONNX_MLIR_VERSION=\"${ONNX_MLIR_VERSION}\"")
+if (NOT DEFINED ONNX_VERSION)
+  set(ONNX_VERSION "1.11.0")
+endif()
+list(APPEND DEFINITIONS "ONNX_VERSION=\"${ONNX_VERSION}\"")
+if (ONNX_MLIR_VENDOR)
+  list(APPEND DEFINITIONS "ONNX_MLIR_VENDOR=\"${ONNX_MLIR_VENDOR }\"")
+endif()
+list(APPEND DEFINITIONS "LLVM_PACKAGE_VERSION=\"${LLVM_PACKAGE_VERSION}\"")
+set_property(SOURCE Version.cpp APPEND PROPERTY COMPILE_DEFINITIONS ${DEFINITIONS})
```
sstamenova, line 68: Why not use `target_compile_definitions`?

whitneywhtsang: Changed to use `target_compile_definitions`.

src/Version/Version.cpp
whitneywhtsang: Sure.

src/Version/CMakeLists.txt
whitneywhtsang: In `build/third_party/onnx/ONNXConfig.cmake`, `set(ONNX_VERSION "1.11.0")`.
I can probably use that, I am not familiar with cmake, so I need to find out how to do that.

gongsu832: This should do what you need.
```
execute_process(
  COMMAND git tag --points-at HEAD
  WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/third_party/onnx
  OUTPUT_VARIABLE ONNX_VERSION
  OUTPUT_STRIP_TRAILING_WHITESPACE
  )
if ("${ONNX_VERSION}" STREQUAL "")
  message(WARNING "third_party/onnx version not detected, you may be on the wrong commit hash")
else()
  # Strip leading "v"
  STRING(REGEX REPLACE "^v" "" ONNX_VERSION ${ONNX_VERSION})
  message(STATUS "Onnx version             : ${ONNX_VERSION}")
endif()
```

whitneywhtsang: Thanks @gongsu832 that save me a lot of time.

whitneywhtsang: On MacOS:
```
CMake Warning at src/Version/CMakeLists.txt:75 (message):
  third_party/onnx version not detected, you may be on the wrong commit hash
```
Will investigate why.

whitneywhtsang: Changed to `cat VERSION_NUMBER`, and that works on all platforms.

src/Version/CMakeLists.txt
whitneywhtsang: added file VERSION_NUMBER.

src/Version/CMakeLists.txt
```diff
@@ -53,4 +53,31 @@ set_source_files_properties("${version_inc}"
...
+    LLVMSupport
+)
+
+file(READ "${CMAKE_SOURCE_DIR}/VERSION_NUMBER" ONNX_MLIR_VERSION)
+string(STRIP "${ONNX_MLIR_VERSION}" ONNX_MLIR_VERSION)
+list(APPEND DEFINITIONS "ONNX_MLIR_VERSION=\"${ONNX_MLIR_VERSION}\"")
+execute_process(
+  COMMAND cat VERSION_NUMBER
+  WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/third_party/onnx
+  OUTPUT_VARIABLE ONNX_VERSION
```
gongsu832, line 70: @whitneywhtsang I think it's OK to use the VERSION_NUMBER file although it would be helpful to figure out why the git command failed on the Mac build. I looked into this a bit and it appears to me that the github action `checkout@v2` is not cloning the tags properly, which appears to be a known problem https://github.com/actions/checkout/issues/206. I would suggest try the `checkout@v3` action and failing that try the proposed fix:
```
diff --git a/.github/workflows/macos-amd64-build.yml b/.github/workflows/macos-amd64-build.yml
index 8e69a516..5a8b9b8b 100644
--- a/.github/workflows/macos-amd64-build.yml
+++ b/.github/workflows/macos-amd64-build.yml
@@ -9,6 +9,8 @@ jobs:
     - uses: actions/checkout@v2
       with:
         submodules: recursive
+    - run: |
+        git fetch --prune --unshallow --tags
     - uses: actions/setup-python@v2
       with:
         python-version: '3.9'
```
And if this fixes the problem, we should update our `macos-amd64-build.yml`.

gongsu832: @whitneywhtsang I see that the command `git fetch --prune --unshallow --tags` failed. Try without the `--prune` option.

whitneywhtsang: For the record, `checkout@v3` doesn't work:
```
--- a/.github/workflows/macos-amd64-build.yml
+++ b/.github/workflows/macos-amd64-build.yml
@@ -6,7 +6,7 @@ jobs:
   build:
     runs-on: macos-latest
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
       with:
         submodules: recursive
```

Removing `--prune` still fails.

gongsu832: Sigh. OK one last try. Let's just do fetch in third_party/onnx. If this still doesn't work, we will just use VERSION_NUMBER.
```
diff --git a/.github/workflows/macos-amd64-build.yml b/.github/workflows/macos-amd64-build.yml
index 8e69a516..0ac3c8bb 100644
--- a/.github/workflows/macos-amd64-build.yml
+++ b/.github/workflows/macos-amd64-build.yml
@@ -30,6 +30,8 @@ jobs:
         sh ~/work/onnx-mlir/onnx-mlir/utils/build-mlir.sh
     - name: install third_party/onnx
       run: |
+        cd ~/work/onnx-mlir/onnx-mlir/third_party/onnx
+        git fetch --prune --unshallow --tags
         cd ~/work/onnx-mlir/onnx-mlir
         pip3 install -e third_party/onnx
     - name: build onnx-mlir
```

whitneywhtsang: `git fetch --prune --unshallow --tags`
or `git fetch --unshallow --tags`
or `git fetch --prune --tags`
or `git fetch --tags`
are failing with:
```
fatal: remote error: upload-pack: not our ref b57c11391d9c749d6ef2b6139a339acac69bbb76
Errors during submodule fetch:
	third_party/benchmark
Error: Process completed with exit code 1.
```

whitneywhtsang: Fetch in third_party/onnx works.

However, I prefer the following change, as it will make getting onnx-mlir and onnx version consistent. 
```
diff --git a/src/Version/CMakeLists.txt b/src/Version/CMakeLists.txt
index 316d0e32..50c28ca6 100644
--- a/src/Version/CMakeLists.txt
+++ b/src/Version/CMakeLists.txt
@@ -64,17 +64,11 @@ add_onnx_mlir_library(OMVersion
 file(READ "${CMAKE_SOURCE_DIR}/VERSION_NUMBER" ONNX_MLIR_VERSION)
 string(STRIP "${ONNX_MLIR_VERSION}" ONNX_MLIR_VERSION)
 list(APPEND DEFINITIONS "ONNX_MLIR_VERSION=\"${ONNX_MLIR_VERSION}\"")
-execute_process(
-  COMMAND git tag --points-at HEAD
-  WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/third_party/onnx
-  OUTPUT_VARIABLE ONNX_VERSION
-  OUTPUT_STRIP_TRAILING_WHITESPACE
-  )
+file(READ "${CMAKE_SOURCE_DIR}/third_party/onnx/VERSION_NUMBER" ONNX_VERSION)
+string(STRIP "${ONNX_VERSION}" ONNX_VERSION)
 if ("${ONNX_VERSION}" STREQUAL "")
   message(WARNING "third_party/onnx version not detected, you may be on the wrong commit hash")
 else()
-  # Strip leading "v"
-  STRING(REGEX REPLACE "^v" "" ONNX_VERSION ${ONNX_VERSION})
   message(STATUS "Onnx version             : ${ONNX_VERSION}")
   list(APPEND DEFINITIONS "ONNX_VERSION=\"${ONNX_VERSION}\"")
```

@gongsu832 What do you think?

gongsu832: @whitneywhtsang This is fine since onnx provides such a version file. Just keep in mind that most code probably won't. Also, let's keep the change to `macos-amd64-build.yml` so tags are fetched. It's a bit surprising that the checkout action doesn't do this by default. Maybe there is some option we have to specify but we can worry about that later. Thanks for trying all the combinations.

whitneywhtsang: Thanks @gongsu832

gongsu832 (approved):
LGTM

sstamenova (approved):
LGTM

whitneywhtsang:

AlexandreEichenberger:
@whitneywhtsang congrats for having pushed through a solution acceptable to all participants... much appreciated

PR #1443, opened 2022-05-18
Update UsingPyRuntime.md to add information about multiple entry points

tungld:
This patch is to update UsingPyRuntime.md by adding some paragraphs about multiple entry points.

### DIFF
--- a/docs/UsingPyRuntime.md
+++ b/docs/UsingPyRuntime.md
@@ -27,20 +27,55 @@ python3
 
 ### Running the PyRuntime interface
 
-Following is an example of how PyRuntime is leveraged in
-an example program:
+An ONNX model is a computation graph and it is often the case that the graph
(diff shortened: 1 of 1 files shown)


docs/UsingPyRuntime.md
```diff
@@ -27,32 +27,68 @@ python3
...
+entry point to do inference. Below is an example of doing inference with multiple
+entry points.
+```python
+import numpy as np
+from PyRuntime import ExecutionSession
+
+model = 'multi-entry-points-model.so'
+
+# Create a session for this model.
+session = ExecutionSession(model, False) # False to manually set an entry point.
```
AlexandreEichenberger, line 63: think Python let you use the syntax `name=value`, might be good to use it here.

AlexandreEichenberger (approved):
LGTM, thanks for updating the doc

PR #1447, opened 2022-05-19
Rt error reporting

AlexandreEichenberger:
Added more systematic error reporting using runtime exception and errno for finer grain control.


src/Runtime/ExecutionSession.hpp
```diff
@@ -29,6 +29,19 @@ using queryEntryPointsFuncType = const char **(*)(int64_t *);
...
+ * When the execution session does not work for known reasons, this class will
+ * throw std::runtime_error errors. Errno info will provide further info about
+ * the specific error that was raised.
+ *
+ * EDOM when it could not load the library or a needed symbol was not found.
+ * EINVAL when it expected an entry point prior to executing a specific
+ * function.
+ * EPERM when the model executed on a machine without a compatible
+ * hardware/specialzed accelerator.
+ */
```
tungld, line 44: I wonder if you would like to write a document about these errors and put it under the `docs` folder.

tungld (approved):
LGTM!

src/Runtime/ExecutionSession.hpp
AlexandreEichenberger: Yes, I see that there is no doc whatsoever of the ExecutionSession, there are examples in readme for examples, but nothing solid.

AlexandreEichenberger: Will do, but let me do a new PR for a UsingC++Runtime.md where we can better show how to use the C++ interface.

gongsu832: EDOM is a fairly uncommen errno to use. EFAULT is probably more common.

src/Runtime/ExecutionSession.cpp
```diff
@@ -79,42 +65,32 @@ const std::string *ExecutionSession::queryEntryPoints(
...
 std::vector<OMTensorUniquePtr> ExecutionSession::run(
     std::vector<OMTensorUniquePtr> ins) {
-  if (!_entryPointFunc) {
-    std::stringstream errStr;
-    errStr << "Must set the entry point before calling run function"
-           << std::endl;
-    throw std::runtime_error(errStr.str());
-  }
+  if (!_entryPointFunc)
+    throw std::runtime_error(reportMissingEntryPointError("run"));
```
sstamenova, line 77: This seems misleading to me - when I see `reportMissingEntryPointError("run")`, I would assume that the missing entry point is `run`, not that there has to be an entry point set before calling `run`.

src/Runtime/ExecutionSession.hpp
```diff
@@ -59,6 +72,14 @@ class ExecutionSession {
   ~ExecutionSession();
 
 protected:
+  // Error reporting processing when throwing runtime errors. Set errno as
+  // appropriate.
+  std::string reportLibraryOpeningError(const std::string &libraryName) const;
```
sstamenova, line 77: Can these be private instead?

src/Runtime/ExecutionSession.cpp
```diff
@@ -34,41 +34,27 @@ ExecutionSession::ExecutionSession(
...
       _sharedLibraryHandle.getAddressOfSymbol(_outputSignatureName.c_str()));
-  if (!_outputSignatureFunc) {
-    std::stringstream errStr;
-    errStr << "Cannot load symbol: '" << _outputSignatureName << "'"
-           << std::endl;
-    throw std::runtime_error(errStr.str());
-  }
+  if (!_outputSignatureFunc)
+    throw std::runtime_error(reportSymbolLoadingError(_outputSignatureName));
+  errno = 0; // No errors.
```
sstamenova, line 57: Do you have to explicitly set `errno` to zero in all these cases?

src/Runtime/ExecutionSession.cpp
etiotto: Same, probably you want to report that "run_main_graph"  (whatever the default entry point name is) was not present.

src/Runtime/ExecutionSession.cpp
```diff
@@ -135,26 +112,21 @@ OMTensorList *ExecutionSession::run(OMTensorList *input) {
...
 
 const std::string ExecutionSession::inputSignature() const {
-  if (!_entryPointFunc) {
-    std::stringstream errStr;
-    errStr << "Must set the entry point before calling signature function"
-           << std::endl;
-    throw std::runtime_error(errStr.str());
-  }
+  if (!_entryPointFunc)
+    throw std::runtime_error(reportMissingEntryPointError("signature"));
```
etiotto, line 121: Is "signature" the actual name of the entry point ?

test/modellib/ModelLib.cpp
```diff
@@ -61,8 +61,14 @@ bool ModelLibBuilder::run() {
     omTensorListDestroy(outputs);
     outputs = nullptr; // Reset in case run has an exception.
   }
-  outputs = exec->run(inputs);
-  return outputs != nullptr;
+  try {
+    outputs =  exec->run(inputs);
+  } catch (const std::runtime_error &error) {
+    std::cout << "error while running: " << error.what() << std::endl;
```
etiotto, line 67: std::cout ==> std:cerr

test/numerical/TestLoop.cpp
```diff
@@ -108,7 +108,13 @@ bool isOMLoopTheSameAsNaiveImplFor(std::string moduleIR,
   omTensorGetElem<int64_t>(yInitTensor.get(), {0}) = yInit;
   inputs.emplace_back(move(yInitTensor));
 
-  auto outputs = sess.run(move(inputs));
+  std::vector<onnx_mlir::OMTensorUniquePtr> outputs;
+  try {
+    outputs = sess.run(move(inputs));
+  } catch (const std::runtime_error &error) {
+    std::cout << "error while running: " << error.what() << std::endl;
```
etiotto, line 115: std::cout ==> std:cerr

src/Runtime/ExecutionSession.hpp
AlexandreEichenberger: changed to EFAULT as suggested

src/Runtime/ExecutionSession.cpp
AlexandreEichenberger: Changed the name to `reportUndefinedEntryPointIn("run")`. Hopefully that reads better.

src/Runtime/ExecutionSession.hpp
AlexandreEichenberger: I made them private at this time, there is a derived class that might use them in the future (python interface), but does not at this time.

src/Runtime/ExecutionSession.cpp
AlexandreEichenberger: No, before running certain commands, the user must define which entry point this is for. If the default is `run_main_graph`, then signatures and run go towards that one. If there are multiple entry points, the user must first indicate which of the many entry point to use BEFORE calling the run/signature calls.

test/modellib/ModelLib.cpp
AlexandreEichenberger: good point

test/numerical/TestLoop.cpp
AlexandreEichenberger: tx

sstamenova (approved):

etiotto (approved):
LGTM

src/Runtime/ExecutionSession.cpp
gongsu832: errno is only meaningful when a function returns error, i.e., return NULL or -1. So there is never a need to set errno when a function succeeds. `if (error) errno = x` is just bad code.

PR #1448, opened 2022-05-20
Strengthened the error reporting for the OnnxMlirCompiler interface

AlexandreEichenberger:
Not all functions that could fail returned a success/failure code that could be used by our OnnxMlir Compiler interface. Now they do.


src/onnx-mlir.cpp
```diff
@@ -61,9 +61,10 @@ int main(int argc, char *argv[]) {
 
   mlir::OwningOpRef<mlir::ModuleOp> module;
   std::string errorMessage;
-  processInputFile(inputFilename, context, module, &errorMessage);
-  if (!errorMessage.empty()) {
-    printf("%s\n", errorMessage.c_str());
+  int rc = processInputFile(inputFilename, context, module, &errorMessage);
+  if (rc != 0) {
+    if (!errorMessage.empty())
+      printf("%s\n", errorMessage.c_str());
```
etiotto, line 67: Suggest to print the error message to stderr rather than stdout - using fprintf(stderr, ....);

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -1445,33 +1445,40 @@ void ImportFrontendModelInternal(onnx::ModelProto &model, MLIRContext &context,
...
 
-void ImportFrontendModelArray(const void *onnxBuffer, int size,
+// Return 0 on success, error otherwise.
+int ImportFrontendModelArray(const void *onnxBuffer, int size,
     MLIRContext &context, OwningOpRef<ModuleOp> &module,
-    ImportOptions options) {
+    std::string *errorMessage, ImportOptions options) {
   onnx::ModelProto model;
 
   auto parse_success = model.ParseFromArray(onnxBuffer, size);
```
etiotto, line 1454: Suggest to use `bool` rather than auto here.

etiotto (approved):
LGTM - left a few comments.

src/onnx-mlir.cpp
gongsu832: For C++ code, the more C++-ish way is probably to use std:cerr/std:cout, etc.

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -1445,33 +1445,40 @@ void ImportFrontendModelInternal(onnx::ModelProto &model, MLIRContext &context,
...
     *errorMessage = "Unable to open or access " + model_fname;
-    return;
+    return 11;
   }
 
   auto parse_success = model.ParseFromIstream(&input);
   if (!parse_success) {
     *errorMessage = "Onnx Model Parsing Failed on " + model_fname;
-    return;
+    return 12;
```
gongsu832, line 1478: Where do the numbers 10, 11, 12 come from? Are we defining our own return codes? If that's the case, they should be defined with symbolic names.

AlexandreEichenberger: Was just creating random numbers... I see your point :-)

src/onnx-mlir.cpp
AlexandreEichenberger: very good, used prior code, will switch to `std::stderr`

src/Builder/FrontendDialectTransformer.cpp
AlexandreEichenberger: you caught me there

PR #1449, opened 2022-05-20
Add a rewriting rule to derive a max trip count for ONNXLoopOp

tungld:
In some ONNX models, the maximum trip count for LoopOp is set to a big value, e.g. LONG_MAX and termination depends on the break condition inside the loop. In the current lowering of LoopOp, the maximum trip count is used to allocate a buffer for all intermediate loop results. Since the actual number of loop iterations may be much smaller than the maximum trip count, it is redundant and error-prone to allocate a large buffer. For example, we may get segfault if the maximum trip count is out of range.

This pattern tries to derive a new maximum trip count for LoopOp by analyzing the break condition. It only handles a special case where the loop is like a for-loop with step, e.g. `for (i = LB, i < UB, i = i + Step)`.


src/Dialect/ONNX/Rewrite.td
```diff
@@ -700,4 +702,14 @@ def ConstantOpNormalizationPattern6: Pat<
...
 
+//===----------------------------------------------------------------------===//
+// Canonicalization for ONNXLessOp
+//===----------------------------------------------------------------------===//
+
+def LessOpSameCastPattern: Pat<
+   (ONNXLessOp (ONNXCastOp $a, $to1), (ONNXCastOp $b, $to2)),
+   (ONNXLessOp $a, $b),
+   [(Equal $to1, $to2)]
+>;
```
tungld, line 713: I see this pattern in a practical model. So add it here.

etiotto: Example: 
```
#include <iostream>
using namespace std;

int main() {
  short s = -1;
  unsigned short us = 1;

  // original
  bool res = (unsigned int)s < (unsigned int)us;
  cout << "res is " << ((res == true) ? "true" : "false") << "\n";

  // eliminate the casts
  bool res1 = s < us;
  cout << "res1 is " << ((res1 == true) ? "true" : "false") << "\n";

  return 0;
}
```
% clang++  cast.cpp; ./a.out
res is false
res1 is true
```

Comment (AlexandreEichenberger) [reply src/Dialect/ONNX/Rewrite.td:713]: @etiotto we would like to have this one merged today. Do you mind either tabling your issue, or enhancing Tung's PR, as he is probably gone for the weekend. Thanks

Comment (etiotto) [reply src/Dialect/ONNX/Rewrite.td:713]: We cannot merge it anyway because the Windows CI is broken...

Comment (etiotto) [reply src/Dialect/ONNX/Rewrite.td:713]: I see @tungld added an additional check `(HaveSameElementType $a, $b)` to the rewrite rule. In its current form the rules will kick in when:
  1. the cast target types are the same, and
  2. the data type of `$a` and `$b` are the same
  
This is not safe either, this program yields different results when the casts are removed:
```
#include <iostream>
using namespace std;

int main() {
  float f1 = -1.0f, f2 = 1.0f;

  // original
  bool res = (unsigned int)f1 < (unsigned int)f2;
  cout << "res is " << ((res == true) ? "true" : "false") << "\n";

  // eliminate the casts
  bool res1 = f1 < f2;
  cout << "res1 is " << ((res1 == true) ? "true" : "false") << "\n";

  return 0;
}
```
% clang++ cast.cpp; ./a.out
res is false
res1 is true

Comment (tungld) [reply src/Dialect/ONNX/Rewrite.td:724]: > That pattern is OK because the cast target types are not unsigned && sizeof(float) == sizeof(int).

@etiotto  Added these constraints. Thanks for your comments!

File: src/Dialect/ONNX/Rewrite.cpp [hunk 2]
```diff
@@ -98,6 +98,197 @@ bool areProducedByTransposeOp(ValueRange values) {
...
+        // 2. is fed with a constant.
+        Value vInput =
+            loopOp->getOperands()[v.cast<BlockArgument>().getArgNumber()];
+        return getOneConstantValue(vInput);
+      }
+      return getOneConstantValue(v);
+    };
+
+    // Only do this rewriting if the maximum trip count is a constant.
+    int64_t maxTripCount = getOneConstantValue(maxTripCountValue);
```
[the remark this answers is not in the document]
tungld: Thanks for the comment! I updated the code to have a separate function `matchOp` to do matching first.

src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -98,6 +98,197 @@ bool areProducedByTransposeOp(ValueRange values) {
...
+            .walk([&](ONNXReturnOp op) -> WalkResult {
+              if (returnOp)
+                return WalkResult::interrupt();
+              returnOp = op;
+              return WalkResult::advance();
+            })
+            .wasInterrupted() ||
+        !returnOp)
+      return failure();
+    Operation *genericReturnOp = returnOp.getOperation();
```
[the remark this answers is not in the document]
tungld: Done. Thanks!

AlexandreEichenberger:
@etiotto do you mind reviewing this one, we really want it down for 0.3.0

etiotto (approved):
LGTM

PR #1451, opened 2022-05-20
Temporarily disable the doc tests in the Windows CI

sstamenova:
The Windows image in the pool now has a version of protobuf installed that does not work correctly with the rest of the build (it's too high).
To work around the issue, temporarily disable the doc tests.

### DIFF
--- a/.azure-pipelines/Windows-CI.yml
+++ b/.azure-pipelines/Windows-CI.yml
@@ -148,8 +148,10 @@ jobs:
     displayName: Install onnx
     workingDirectory: $(Agent.BuildDirectory)
 
-  - script: |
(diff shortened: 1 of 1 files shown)


sstamenova:
Assuming https://github.com/onnx/onnx-mlir/pull/1452 succeeds, this change won't be necessary at this point. Still TBD whether the protobuf change is successful in the online builds though.

etiotto (approved):
LGTM - thanks. Might want to merge https://github.com/onnx/onnx-mlir/pull/1452 first to determine whether this one is still necessary.

sstamenova:
#1452 went in, so this is not needed.

PR #1459, opened 2022-05-25
Use the ceiling value when deriving a max trip count for ONNXLoopOp

tungld:
This patch uses the ceiling value when computing a max trip count for ONNXLoopOp to make sure the number of iterations is NOT decreased compared to the actual trip count.

This patch also removes a redundant function.

### DIFF
--- a/src/Dialect/ONNX/Rewrite.cpp
+++ b/src/Dialect/ONNX/Rewrite.cpp
@@ -15,6 +15,7 @@
 
 #include "mlir/IR/Matchers.h"
 #include "mlir/IR/PatternMatch.h"
+#include <math.h>
 
 #include "src/Dialect/ONNX/ONNXOps.hpp"
 #include "src/Dialect/ONNX/ONNXOpsHelper.hpp"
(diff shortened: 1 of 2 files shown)


src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -338,7 +328,10 @@ class LoopOpRewriteMaxTripCountPattern : public OpRewritePattern<ONNXLoopOp> {
     int64_t lowerBound = getOneIntergerConstant(startValue);
     int64_t upperBound = getOneIntergerConstant(ubValue);
     int64_t step = getOneIntergerConstant(stepValue);
-    int64_t derivedTripCount = (int64_t)((upperBound - lowerBound) / step);
+    if (step == 0)
+      return std::make_pair(false, -1);
+    int64_t derivedTripCount = (upperBound - lowerBound) / step +
+                               ((upperBound - lowerBound) % step != 0);
```
AlexandreEichenberger, line 334: I am afraid this algorithm does not work for negative divisor. Best using what I used in IndexExpr.cpp, namely
```
    int64_t rval = ceil((1.0 * aa.getLiteral()) / (1.0 * bb.getLiteral()));
```

AlexandreEichenberger (approved):
LGTM upon correcting the ceil for negative divisor

src/Dialect/ONNX/Rewrite.cpp
tungld: Done. I changed to use ceil. Thanks!

PR #1462, opened 2022-05-26
Derive a new max trip count for ONNXLoopOp when the upperbound, lowerbound and step are not constants

tungld:
Before this patch, we have a rewriting rule to derive a new max trip count for ONNXLoopOp when the upperbound, lowerbound and step are constants. In such case, the new max trip count is computed during the compile time.

This patch enhances the rule to support the case wherein the upperbound, lowerbound and step are not constants. In this case, we emit code to compute a derived trip count, and the new max trip count will be `min(old_max_trip_count, derived_max_trip_count)`.

This patch also adds several ops into OnnxBuilder and sorts the ops in the alphabet order.


tungld:
@chentong319 this patch is a response to your second suggestion in the previous PR: https://github.com/onnx/onnx-mlir/pull/1449#pullrequestreview-981993732

src/Dialect/ONNX/DialectBuilder.cpp
```diff
@@ -62,17 +72,26 @@ Value OnnxBuilder::matmul(Type Y, Value A, Value B, bool useGemm) const {
   return b.create<ONNXMatMulOp>(loc, Y, A, B);
 }
 
+Value OnnxBuilder::min(ValueRange inputs) const {
+  assert(inputs.size() >= 2 && "Expect at least two inputs");
```
etiotto, line 76: [nit]: could assert that all inputs have the same data type (using all_of or any_of).

src/Dialect/ONNX/DialectBuilder.cpp
```diff
@@ -62,17 +72,26 @@ Value OnnxBuilder::matmul(Type Y, Value A, Value B, bool useGemm) const {
...
+  return b.create<ONNXMinOp>(loc, inputs[0].getType(), inputs);
+}
+
+Value OnnxBuilder::mul(Value A, Value B) const {
+  return b.create<ONNXMulOp>(loc, A, B);
```
etiotto, line 81: I think adding an assertion (here and for other binary operators) to check that the input have the same data type would be useful.

etiotto (approved):
LGTM - added a few suggestions.

src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -312,30 +311,87 @@ class LoopOpRewriteMaxTripCountPattern : public OpRewritePattern<ONNXLoopOp> {
...
+
+    // Case 1: the upper bound, lower bound and step are constants.
+    // - Compute the new max trip count at the compile time.
+    if (isDefinedByIntegerConstantOp(lbValue) &&
+        isDefinedByIntegerConstantOp(ubValue) &&
+        isDefinedByIntegerConstantOp(stepValue)) {
+      int64_t lowerBound = getOneIntergerConstant(lbValue);
+      int64_t upperBound = getOneIntergerConstant(ubValue);
+      int64_t step = getOneIntergerConstant(stepValue);
+      if ((step <= 0) || (upperBound - lowerBound) <= 0)
```
AlexandreEichenberger, line 352: nit `upperBound <= lowerBound` is easier to read. And if the trip count is zero? we could just get rid of the loop :-)

AlexandreEichenberger (approved):
LGTM too

src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -312,30 +311,87 @@ class LoopOpRewriteMaxTripCountPattern : public OpRewritePattern<ONNXLoopOp> {
     Operation *addOp = cast<ONNXAddOp>(newCounterValue.getDefiningOp());
     Value counterValue = addOp->getOperands()[0];
     Value stepValue = addOp->getOperands()[1];
-    // 1. Step is a constant inside or outside the loop.
-    if (isInvariantArgConstant(stepValue, returnOp))
-      stepValue = getFedValue(stepValue, loopOp);
-    if (!isDefinedByIntegerConstantOp(stepValue))
-      return std::make_pair(false, -1);
-    // 2. Counter is an block argument and updated at each iteration.
+    // Counter is an block argument and updated at each iteration.
```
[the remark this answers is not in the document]
tungld: Done. Thanks!

src/Dialect/ONNX/DialectBuilder.cpp
tungld: Done. Added an assertion for all binary operators.

src/Dialect/ONNX/DialectBuilder.cpp
tungld: Done. Thanks!

PR #1464, opened 2022-05-26
Fix warnings that break local builds that have warnings-as-error on

GabLotus:
Fixing warnings that break local builds when we are treating warnings as error.

The fix in Accelerator.hpp is due to the macro emitting a warning when the variadic argument is empty.

The fix in ConvertKrnlToAffine.cpp is due to an unnecessary semicolon.


AlexandreEichenberger:

GabLotus:
@AlexandreEichenberger thanks for the pointer, I think the signing is correct now.

sstamenova:
@GabLotus : Looks like the PR builds are still broken with this change. Can you make sure they are green?

AlexandreEichenberger:
Looks like your fix causes trouble when building for the NNPA. This accelerator target can be built on any of our machines, even obviously it can only run on our Z machines.

Just add `-DONNX_MLIR_ACCELERATORS=NNPA ` to the `cmake ..` command. Here is the script I use when buiding in a docker for x86 linux

```
rm -rf onnx-mlir/build
mkdir onnx-mlir/build && cd onnx-mlir/build
if [[ -z "$pythonLocation" ]]; then
  cmake -G "Unix Makefiles" -DCMAKE_CXX_COMPILER=/usr/bin/c++ -DMLIR_DIR=/workdir/llvm-project/build/lib/cmake/mlir -DONNX_MLIR_ACCELERATORS=NNPA ..
else
  cmake -G "Unix Makefiles" -DCMAKE_CXX_COMPILER=/usr/bin/c++ -DMLIR_DIR=/workdir/llvm-project/build/lib/cmake/mlir -DONNX_MLIR_ACCELERATORS=NNPA -DPython3_ROOT_DIR=$pythonLocation ..
fi
cmake --build .

# Run lit tests:
export LIT_OPTS=-v
cmake --build . --target check-onnx-lit
```

hope this helps

src/Accelerators/Accelerator.hpp
```diff
@@ -51,7 +51,7 @@ class Accelerator {
   enum class Kind {
     // clang-format off
     NONE,
-    APPLY_TO_ACCELERATORS(CREATE_ACCEL_ENUM)
+    APPLY_TO_ACCELERATORS(CREATE_ACCEL_ENUM, dummy)
```
etiotto, line 54: I think the `CREATE_ACCEL_ENUM`  macro accepts a single argument:
```
#define CREATE_ACCEL_ENUM(name) name,
```
So this code should fail to compile.

etiotto (changes requested):

PR #1456, opened 2022-05-24
Avoid assertion failure of some rewriter pattern constraints

llehtahw:
See https://github.com/llvm/llvm-project/blob/64186e9b351a000123f253557f4f607c33405413/mlir/include/mlir/IR/BuiltinTypeInterfaces.td#L137

PR #1466, opened 2022-05-27
[NNPA] Support bias none in ONNX.Gemm

imaihal:
- This PR allows ONNX.Gemm with none bias to generate zhigh op. This will support `test_gemm_default_no_bias` in backend tests for NNPA.

- Use `hasRank()` before checking `getRank()` to follow PR #1456


src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.td
```diff
@@ -36,6 +36,10 @@ class HasRankOf<int rank> : Constraint<
   CPred<"$0.getType().isa<ShapedType>() && $0.getType().cast<ShapedType>().getRank() == " # rank>
 >;
 
+def IsBiasNoneOr1D : Constraint<
+  CPred<"(($_self).getType().isa<NoneType>()) || (($_self).getType().isa<ShapedType>() && ($_self).getType().cast<ShapedType>().getRank() == 1)">
```
tungld, line 40: This line is too long, please split it into multiple short lines, e.g.
```
"($_self.getType().isa<NoneType>() || "
"  ($_self.getType().isa<ShapedType>() && "
"   $_self.getType().cast<ShapedType>().hasRank() &&"
"   ($_self.getType().cast<ShapedType>().getRank() == 1))"
```

There is a bug that we need to check `$_self.getType().cast<ShapedType>().hasRank()` before `getRank`. This PR #1456 addressed that issue. Could you fix the same issue in this file and possibly other files? Thanks!

tungld (changes requested):

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.td
imaihal: Done. Thanks!
I checked NNPA code and fix it to use `hasRank()`.

tungld (approved):
LGTM. Thanks for the patch!

PR #1467, opened 2022-05-27
Simplify ONNXEntryPoint

tungld:
This patch simplifies ONNXEntryPoint from 
```mlir
"onnx.EntryPoint"() {func = @main_graph, numInputs = 2 : i32, numOutputs = 1 : i32, signature = "[    { \22type\22 : \22f32\22 , \22dims\22 : [3 , 4 , 5] , \22name\22 : \22x\22 }\0A ,    { \22type\22 : \22f32\22 , \22dims\22 : [3 , 4 , 5] , \22name\22 : \22y\22 }\0A\0A]\00@[   { \22type\22 : \22f32\22 , \22dims\22 : [3 , 4 , 5] , \22name\22 : \22sum\22 }\0A\0A]\00"}
```
into
```mlir
"onnx.EntryPoint"() {func = @main_graph}
```
where only `func` is required, and `numInputs`, `numOutputs` and `signature` will be automatically generated when lowering `onnx.EntryPoint` to `krnl.EntryPoint`:
```mlir
"krnl.entry_point"() {func = @main_graph, numInputs = 2 : i32, numOutputs = 1 : i32, signature = "[    { \22type\22 : \22f32\22 , \22dims\22 : [3 , 4 , 5] , \22name\22 : \22x\22 }\0A ,    { \22type\22 : \22f32\22 , \22dims\22 : [3 , 4 , 5] , \22name\22 : \22y\22 }\0A\0A]\00@[   { \22type\22 : \22f32\22 , \22dims\22 : [3 , 4 , 5] , \22name\22 : \22sum\22 }\0A\0A]\00"} : () -> ()
```

This is to solve the issue that it is too early to set signature in `onnx.EntryPoint` when importing a model, because at that time the output tensors may not have the correct shape, e.g. shape is still unknown until shape inference is called.

### DIFF
--- a/src/Conversion/ONNXToKrnl/ConvertONNXToKrnl.cpp
+++ b/src/Conversion/ONNXToKrnl/ConvertONNXToKrnl.cpp
@@ -33,16 +33,137 @@ public:
 
   LogicalResult matchAndRewrite(
       ONNXEntryPointOp op, PatternRewriter &rewriter) const override {
-    rewriter.replaceOpWithNewOp<KrnlEntryPointOp>(op,
-        op->getAttrOfType<SymbolRefAttr>(
-            ONNXEntryPointOp::getEntryPointFuncAttrName()),
-        op->getAttrOfType<IntegerAttr>(
-            ONNXEntryPointOp::getNumInputsAttrName()),
-        op->getAttrOfType<IntegerAttr>(
-            ONNXEntryPointOp::getNumOutputsAttrName()),
-        op->getAttrOfType<StringAttr>(
-            ONNXEntryPointOp::getSignatureAttrName()));
+    ModuleOp module = op.getOperation()->getParentOfType<ModuleOp>();
+
+    SymbolRefAttr funcRefAttr = op->getAttrOfType<SymbolRefAttr>(
+        ONNXEntryPointOp::getEntryPointFuncAttrName());
+    StringRef entryPointName = funcRefAttr.getLeafReference().getValue();
+    Operation *entryPointOp = module.lookupSymbol(entryPointName);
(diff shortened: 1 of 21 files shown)


src/Conversion/ONNXToKrnl/ConvertONNXToKrnl.cpp
```diff
@@ -33,17 +33,138 @@ class ONNXEntryPointLowering : public OpRewritePattern<ONNXEntryPointOp> {
...
+    }
+    ArrayAttr outputNames = op->getAttrOfType<ArrayAttr>("output_names");
+    if (!outputNames) {
+      SmallVector<StringRef, 4> names;
+      for (uint64_t i = 0; i < outputs.size(); ++i)
+        names.emplace_back(StringRef("output_" + std::to_string(i)));
+      outputNames = b.getStrArrayAttr(names);
+    }
+
+    std::string const sf32 = std::string(" f32 ");
```
AlexandreEichenberger, line 116: suggestion: each of the sf64... seemed to be used only once. Could you define it directly in the map? Also, you could make the map static, and possibly have a `static bool isMapInit=false` and initialize on first call if its not possible to have the init of the map directly in a constructor.

AlexandreEichenberger (approved):
This was long overdue, it was painful to init the string in test examples. LGTM, made a minor suggestion.

src/Conversion/ONNXToKrnl/ConvertONNXToKrnl.cpp
tungld: Thanks! I puts the strings directly into the map and make the map static.

PR #1469, opened 2022-05-27
Ccache support with ONNX_MLIR_CCACHE_BUILD option

NathanielMcVicar:
Add support for ccache on Windows and Linux. This isn't tested for other platforms, so additional changes may be required. Doesn't support ccache options through CMake, the way LLVM does, because of how platform dependent this approach is.

Signed-off-by: Nathaniel McVicar <Nathaniel.McVicar@microsoft.com>


CMakeLists.txt
```diff
@@ -36,6 +37,23 @@ if (MSVC)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -wd4125")
 endif()
 
+# Enable ccache
+# LLVM_CCACHE_BUILD for consistency, but this currently has no effect
+set(LLVM_CCACHE_BUILD ${ONNX_MLIR_CCACHE_BUILD})
+if(ONNX_MLIR_CCACHE_BUILD)
```
sstamenova, line 43: Nit: spaces after ifs

CMakeLists.txt
```diff
@@ -36,6 +37,23 @@ if (MSVC)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -wd4125")
 endif()
 
+# Enable ccache
+# LLVM_CCACHE_BUILD for consistency, but this currently has no effect
```
sstamenova, line 41: What effect would it have?

NathanielMcVicar: Some LLVM options have cascading effects in files included through MLIR.cmake. I don't really see what this would be for ccache, but it seems better safe than sorry to avoid a mismatch.

CMakeLists.txt
```diff
@@ -36,6 +37,23 @@ if (MSVC)
...
+set(LLVM_CCACHE_BUILD ${ONNX_MLIR_CCACHE_BUILD})
+if(ONNX_MLIR_CCACHE_BUILD)
+  find_program(CCACHE_PROGRAM ccache)
+  if(CCACHE_PROGRAM)
+    if(CMAKE_SYSTEM_NAME MATCHES "Windows")
+      set(CMAKE_C_COMPILER_LAUNCHER ${CCACHE_PROGRAM})
```
sstamenova, line 47: Do the `_LAUCNHER` properties not work for Ubuntu? According to how they are defined, I would expect them to always work.

CMakeLists.txt
```diff
@@ -36,6 +37,23 @@ if (MSVC)
...
+      set(CMAKE_CXX_COMPILER_LAUNCHER ${CCACHE_PROGRAM})     
+    else()
+      set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ${CCACHE_PROGRAM})
```
sstamenova, line 50: This would technically not work for Xcode which is used on a non-Windows platform

CMakeLists.txt
sstamenova: I don't think there would be any impact here. If anything, the option should go the other way around and the onnx-mlir version should be set to the llvm version.

CMakeLists.txt
NathanielMcVicar: They do, but RULE_LAUNCH_COMPILER is a bit more powerful, since ccache has some support for other compilers.

CMakeLists.txt
NathanielMcVicar: I don't have a Mac to try, but some quick searching shows that ccache with ninja and Xcode requires some [non-trivial additional support](https://stackoverflow.com/a/36515503) anyway, which seems out of scope here.

sstamenova: Can we add a comment here saying so?

sstamenova:

gongsu832:
On Linux, at least for the Jenkins CI, we already do our own equivalent of caching since we build everything in docker. We only rebuild the images when necessary, e.g., we only rebuild llvm-project when we change to a new commit hash, etc. So I don't think ccache will buy us much on Linux. Therefore, I prefer not to introduce more complexity. I have no objection to have it on Windows if it provides visible benefits.

NathanielMcVicar:
> we only rebuild llvm-project when we change to a new commit hash, etc.
This is only for the ONNX MLIR build itself, not LLVM (which I admit is relatively fast, but some machines are much slower). The primary target is developer workstations, not CI or other online builds. I'm happy to remove support if you don't think it's appropriate, but I wanted to clarify who the audience is. Please let me know what you think!

gongsu832:
@NathanielMcVicar Ah OK that would be fine, and defaults to OFF is good. Thanks for the clarification.

NathanielMcVicar:
> I'm actually curious on how much benefit it provides. Since most build systems (e.g., make, ninja, etc) only rebuild changed files already, don't they?

@gongsu832 I often have to do clean builds due to changes in our fork. I think the scenario where this is useful is somewhat limited, but hopefully some people will find it helpful. In terms of the actual improvement, it's pretty huge, with clean build time going from ~4 minutes to <1 minute. CMake takes much longer than the build if I have to reconfigure :)

CMakeLists.txt
sstamenova: Can you add a comment for this also?

CMakeLists.txt
```diff
@@ -36,6 +37,24 @@ if (MSVC)
...
+    if (CMAKE_SYSTEM_NAME MATCHES "Windows")
+      # RULE_LAUNCH_COMPILE is more powerful than C_COMPILER_LAUNCHER (ccache
+      # supports additional compilers) but has issues with cmd and MSVC tools
+      set(CMAKE_C_COMPILER_LAUNCHER ${CCACHE_PROGRAM})
+      set(CMAKE_CXX_COMPILER_LAUNCHER ${CCACHE_PROGRAM})     
+    else()
+      set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ${CCACHE_PROGRAM})
+    endif()
+  else()
+    message(FATAL_ERROR "Unable to find the program ccache. Set ONNX_MLIR_CCACHE_BUILD to OFF.")
```
gongsu832, line 54: I wonder whether it's better to just issue a warning and continue? Either way is fine I guess. No strong preference to either.

NathanielMcVicar: I copied the LLVM behavior here for consistency. If you would prefer it changed I can, but my concern would be that the CMake generate for ONNX MLIR is a pretty long chunk of text, and it's easy to miss warnings in there.

gongsu832 (approved):
LGTM

sstamenova:

PR #1472, opened 2022-05-31
Verify the input tensors of the entry point function

### LINKED ISSUE #1125: onnx-mlir does not check types and shapes of inputs
Currently, onnx-mlir does not check the shapes of inputs passed to a model, which would cause wrong results or even segfault.

For example, an input program is:
```mlir
module {
  func private @main_graph(%arg0 : tensor<4x4xf32>, %arg1 : tensor<4x3xf32>) -> tensor<*xf32> {
     %0 ="onnx.MatMul"(%arg0, %arg1) : (tensor<4x4xf32>, tensor<4x3xf32>) -> tensor<*xf32>
  }
  "onnx.EntryPoint"() {func = @main_graph, numInputs = 2 : i32, numOutputs = 1 : i32, signature = ""} : () -> ()
}
```

which expects 4x4 and 4x3 matrices.

A python program for the above program is:
```python
import numpy as np

model = 'test.so'
sess = ExecutionSession(model, "run_main_graph")
a = np.arange(16).astype('float32') # shape is (16,)
b = np.arange(12).astype('float32') # shape is (12,)
y = sess.run([a, b])
print(y[0])
```
This program produces a correct result though the input shapes does not match the model input shapes, that is:
```
[[ 42.  48.  54.]
 [114. 136. 158.]
 [186. 224. 262.]
 [258. 312. 366.]]
```

A real issue happens when `a` has less than 16 elements, say `a = np.arange(10).astype('float32')`, there is no error but a wrong result:
```
[[ 42.    48.    54.  ]
 [114.   136.   158.  ]
 [ 27.    44.    61.  ]
 [ 15.75  17.5   19.25]]
```

tungld:
This patch adds verification of the input tensors that are passed to entry point functions such as run_main_graph. Entry point signature is used to verify tensors. In particular, it verifies:
- The number of input tensors
- Data type of each input tensor
- Rank and shape of each input tensor

When the input tensors are invalid, `run_main_graph` returns NULL and `errno = EINVAL` (Invalid argument) is emitted. 

A new compile option `--verifyInputTensors` is added to enable verification inside the entry point function. It's off by default.

A new backend test is added to verify this verification, that is `make check-onnx-backend-input-verification`.

This is an example of verification when data type is wrong:
```
Wrong data type for the input 0: expect f32
  File "test.py", line 22, in <module>
    y = sess.run(input=[a, b])
RuntimeError: Runtime error during inference returning with ERRNO code 'Invalid argument'.
```

The verification detected an error about numerical tests on Windows CI:
```
Wrong data type for the input 0: expect f32, but got undefined.
```
It looked like the data type of the randomly generated tensors was not set correctly on Windows. This patch fixed this issue also.

Resolves #1125 


tungld:

src/Runtime/OMTensorList.inc
```diff
@@ -120,3 +127,209 @@ OMTensor *omTensorListGetOmtByIndex(OMTensorList *rlist, int64_t index) {
...
+static inline char *createSubStr(
+    char *str, const char *start, const char *end) {
+  char *startPos = strstr(str, start) + strlen(start);
+  char *endPos = strstr(str, end);
+  if (startPos != NULL && endPos != NULL && (endPos > startPos)) {
+    unsigned len = (unsigned)(endPos - startPos);
+    char *subStr = (char *)malloc(len + 1);
+    strncpy(subStr, startPos, len);
+    subStr[len] = '\0';
+    return subStr;
```
gongsu832, line 141: Line 137-141 can be simplified as `return strndup(startPos, endPos - startPos);`

I think the correct code should be,
```
{
  char *startPos = strstr(str, start);
  char *endPos = strstr(str,end);
  char *p;
  return (startPos && endPos && (p = startPos + strlen(start)) < endPos) ?
    strndup(p, endPos - p) : NULL;
}

Comment (tungld) [reply src/Runtime/OMTensorList.inc]: Thanks so much! I updated the code with your suggestion.

File: src/Runtime/OMTensorList.inc [hunk 2]
```diff
@@ -120,3 +127,209 @@ OMTensor *omTensorListGetOmtByIndex(OMTensorList *rlist, int64_t index) {
...
+  }
+  return NULL;
+}
+
+/**
+ * Verify OMTensorList again the given signature
+ * Signature must be in JSON format with keys `type`, `dims` and `name`, e.g.
+ *   [{"type" : "f32" , "dims" : [4, 8, 16] , "name": "t1"},
+ *    {"type" : "i32" , "dims" : [32, 64, 128] , "name": "t2"}]
+ */
```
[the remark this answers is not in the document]
tungld: Thanks so much! It's what I am looking for. It supports multiple platforms and has a good license.

include/onnx-mlir/Runtime/OnnxDataType.h
```diff
@@ -43,6 +43,18 @@ extern const char *OM_DATA_TYPE_NAME[];
 #ifdef __cplusplus
 // Note by design const map has no [] operator since [] creates a default
 // key value mapping when the key is not found which changes the map
+#if defined(_WIN32)
+// clang-format off
+const std::map<std::string, OM_DATA_TYPE> OM_DATA_TYPE_CPP_TO_ONNX = {
```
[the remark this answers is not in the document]
tungld: It's because of `typeid(T).name` returns different value on Windows. We use `typeid` here `omt->_dataType = OM_DATA_TYPE_CPP_TO_ONNX.at(std::string(typeid(T).name()));` in https://github.com/onnx/onnx-mlir/blob/main/src/Runtime/OMTensor.inc#L535

src/Conversion/KrnlToLLVM/ConvertKrnlToLLVM.cpp
```diff
@@ -551,7 +484,8 @@ void ConvertKrnlToLLVMPass::runOnOperation() {
 
   populateAffineAndKrnlToLLVMConversion(patterns, typeConverter, ctx,
       outputOMTensorOwnerships,
-      /*singleEntryPoint=*/entryPointNames.size() == 1);
+      /*singleEntryPoint=*/singleEntryPoint, entryGlobalOps, inSigGlobalOps,
```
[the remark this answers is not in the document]
tungld: Fixed. Thanks!

src/Conversion/KrnlToLLVM/KrnlEntryPoint.cpp
```diff
@@ -426,13 +447,128 @@ class KrnlEntryPointOpLowering : public OpRewritePattern<KrnlEntryPointOp> {
...
+          /*isConstant=*/true, LLVM::Linkage::External, name, valueAttr);
+      return globalOp;
+    };
+
+    bool zOS = false;
+    if (Attribute mtripleAttr =
+            module->getAttrOfType<::mlir::Attribute>("llvm.target_triple"))
+      zOS = llvm::Triple(mtripleAttr.cast<StringAttr>().getValue()).isOSzOS();
+
+    // // NULL terminated entry point name.
```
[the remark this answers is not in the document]
tungld: Fixed. Thanks!

src/Runtime/OMTensorList.inc
```diff
@@ -120,3 +141,205 @@ OMTensor *omTensorListGetOmtByIndex(OMTensorList *rlist, int64_t index) {
...
+        break;
+      case ONNX_TYPE_COMPLEX128:
+        typeIsOK = true; // no verification at this moment.
+        break;
+      case ONNX_TYPE_BFLOAT16:
+        typeIsOK = true; // no verification at this moment.
+        break;
+      case ONNX_TYPE_UNDEFINED:
+        typeIsOK = false;
+        break;
```
chentong319, line 271: Do we have a table in runtime to convert ONNX type enum and string name?

test/backend/input_verification_backend.py
```diff
@@ -0,0 +1,229 @@
+#!/usr/bin/env python3
+
+##################### input_verification_backend.py ############################
+#
+# Copyright 2022 The IBM Research Authors.
+#
+################################################################################
+
```
chentong319, line 8: If onnx can provide a hook for user to provide input for a model in its backend test, it will be easier for us to add the test.

chentong319: Just realized that we can just save the model in backend test and run it with different inputs with our python script without checking the correctness of output.

test/backend/variables.py
```diff
@@ -42,6 +42,7 @@ def get_args_from_env():
     TEST_DYNAMIC = os.getenv("TEST_DYNAMIC")
     TEST_CONSTANT = os.getenv("TEST_CONSTANT")
     TEST_SIGNATURE = os.getenv("TEST_SIGNATURE")
```
chentong319, line 44: Not particular for this PR: can we use argument to the test.py instead of env variable for these options? I assume that options are better than env variable for such purpose.

chentong319 (approved):
1. The PR looks fine to me. Waiting for other's opinion.
2. We better document: the input/output signature is after shape inference; and the krnl::entryPoint adds some dependency in the .mlir file that may not be detected by MLIR. I ran into this problem when I manually modify .mlir file.
3. The new PR to check input outside of our compiled code is good option for performance. Do we need to option to disable the code gen for signature checking?

test/backend/variables.py
gongsu832: Not necessarily. Depending on the situation, options can be tricky to set due to the command string being interpreted multiple times, as you can see from the following example:
```
if (${PYTEST_XDIST_FOUND} EQUAL 0)
  message(STATUS "Parallel backend tests   : ON")
  set(BACKEND_TEST_COMMAND "${Python3_EXECUTABLE}" "-m" "pytest")
  set(BACKEND_TEST_ARGS "--forked" "-n" "$$\{NPROC:-auto\}" "-q" "--silent")
```
Here you need two `$` because cmake interprets the string first and makes that a single `$` for bash to consume. Similarly for the escaped `\{` so that after the cmake pass you are left with the `{` for bash to consume. For situations like this, it's typically a lot simpler to use env vars to set the parameters.

There are also situations where you definitely want to use env vars. For example, there might be a chain of invocation, i.e., script A invokes B, which invokes C. If you want to pass something to C via command line argument, you will have to write code in A and B to pass on the option. But with env var, you just need to set it when invoking A and it gets automatically propagated to all the subprocesses (B and C).

So in general, often used options should be settable by both command line argument and env var.

chentong319: @gongsu832 thanks for the comments.

AlexandreEichenberger:
I would like to have a bit more high level info to understand what and how this PR is doing the checking. Checking is useful to our uses, no doubts.

I am a bit concerned that we use the compiler to create a json string and then use the json string to verify the input. So why not skipping the (probably inefficient) json string and test directly the inputs from what the shape inference and compiler knows.

What is the performance impact of testing? It is a useful service to the user, but does it have to be inline all the time? 

LLVM has a Json string parser, I have used it in our utils; so creating a "service" that check the input tensors with for the JSON string can be relatively easily done in the runtime (e.g. additional call to ExecutionSession).

If we do it inline all the time, it better be very cheap, or we should consider switching it on/off using an option.

src/Compiler/CompilerPasses.hpp
```diff
@@ -20,7 +20,8 @@ void addONNXToMLIRPasses(mlir::PassManager &pm);
 void addONNXToKrnlPasses(
     mlir::PassManager &pm, int optLevel, bool enableCSE = true);
 void addKrnlToAffinePasses(mlir::PassManager &pm);
-void addKrnlToLLVMPasses(mlir::OpPassManager &pm, bool enableCSE = true);
+void addKrnlToLLVMPasses(mlir::OpPassManager &pm, bool enableCSE = true,
+    bool verifyInputTensors = false);
```
AlexandreEichenberger, line 24: nit: it looks like verifyInputTensor is always defined explicitly, so maybe no need to make it with a default value?

src/Conversion/KrnlToLLVM/KrnlEntryPoint.cpp
```diff
@@ -426,13 +458,212 @@ class KrnlEntryPointOpLowering : public OpRewritePattern<KrnlEntryPointOp> {
...
+    rewriter.setInsertionPointToStart(endBlock);
+  }
+
+  void emitVerificationCodeForInputTensors(PatternRewriter &rewriter,
+      Location loc, const RuntimeAPIRegistry &apiRegistry, Value wrappedInput,
+      StringRef inSigJSON) const {
+    Type int64Ty = rewriter.getI64Type();
+    Type opaquePtrTy = LLVM::LLVMPointerType::get(rewriter.getI8Type());
+
+    auto JSONInput = llvm::json::parse(inSigJSON.data());
```
AlexandreEichenberger, line 502: Just requesting a clarification: here we don't have any JSON calls at runtime, correct?  The compiler parse the JSON string at compile time to generate a test that will compare the string from the signature with the expected string found at compile time from JSON?
So it feels like you implemented the efficient test, isn't it?

tungld: Correct. We parse the JSON string at compile time to get the element type, rank and shape, and then generate runtime code to compare them with the ones from users. It is integer comparison, so quite efficient at runtime.

We can avoid parsing the JSON string at compile time also by storing information about input MemRefs before they are unfolded into LLVM struct. But, I feel that it's unncessary, given that the JSON string correctly reflects input signature.

AlexandreEichenberger (approved):
LGTM

AlexandreEichenberger:
> Verification is very lightweight, it is like if (actualRank != referenceRank) then return NULL for each value we want to verify.

Very good. I assume that the errno value is also set and the pseudo code above just omitted this part.

tungld:
> I assume that the errno value is also set and the pseudo code above just omitted this part.

Yes, you are right.

PR #1371, opened 2022-04-20
Upgrade to Onnx v1.11.0

chentong319:
This PR just upgrades the third_party/onnx. The Ops are not upgraded yet.
A side change is the document. Please verify that.

PR #1479, opened 2022-06-02
Update rapidcheck to 8fafda4

cjvolzka:
Update to the latest rapidcheck commit. This includes https://github.com/emil-e/rapidcheck/commit/ce2b602bbe89e9264fb8725cc3f04049951f29cb which updates some rapidcheck terms for more inclusive language. 

Signed-off-by: Charles Volzka <cjvolzka@us.ibm.com>


AlexandreEichenberger (commented):
Isn't there usually a git or makefile file under onnx-mlir (not in 3rd party) that indicates the version/commit?

cjvolzka:
There is for llvm-project but that's not brought in as a git submodule. For things that are submodules, I think this is the way to update them. For example, https://github.com/onnx/onnx-mlir/pull/1371 appears to do the same thing to update the `onnx` used by onnx-mlir.

The only other reference I can find for `rapidcheck.git` (which I think would be required for any type of `git clone` call) is at https://github.com/onnx/onnx-mlir/blob/main/.gitmodules which doesn't specify a branch or commit.

AlexandreEichenberger:
I am sorry, maybe I understand the changes here wrong. Are you changing the flies directly in the third_party/rapid check? If yes, this feels wrong. If false, where do you indicate which version to take? I would have imagined that this is the role of the https://github.com/onnx/onnx-mlir/blob/main/.gitmodules file.

But maybe I am wrong: is that what you did?
```
Step 1: Add the submodule

   git submodule add git://some_repository.git some_repository
Step 2: Fix the submodule to a particular commit

By default the new submodule will be tracking HEAD of the master branch, but it will NOT be updated as you update your primary repository. In order to change the submodule to track a particular commit or different branch, change directory to the submodule folder and switch branches just like you would in a normal repository.

   git checkout -b some_branch origin/some_branch
Now the submodule is fixed on the development branch instead of HEAD of master.
```
From https://stackoverflow.com/questions/10914022/how-do-i-check-out-a-specific-version-of-a-submodule-using-git-submodule ?

There it say one can also do this:
```
It's easier to use git submodule add -b some_branch git://some_repository.git some_repository
```
which works for branches but not she numbers

Please confirm and then I will accept it as it seems the approved way to do this. Maybe folks with more expertise on submodule magic could confirm... I am only going with what I saw by a google search. @gongsu832 @etiotto @doru1004 ?

cjvolzka:
@AlexandreEichenberger The bit starting at ` In order to change the submodule...` is what I did.
```
By default the new submodule will be tracking HEAD of the master branch, but it will NOT be updated as you update your primary repository. In order to change the submodule to track a particular commit or different branch, change directory to the submodule folder and switch branches just like you would in a normal repository.
```
I moved from the HEAD it was tracking to the current HEAD of rapidcheck's `master`.

AlexandreEichenberger (approved):
LGTM

PR #1481, opened 2022-06-03
Use ONNX_MLIR_SRC_ROOT to find the version file

kile0:
If onnx-mlir is being built as a sub-project, `CMAKE_SOURCE_DIR` will point to the parent's top-level directory, rather than onnx-mlir's top-level directory. This can can cause issues finding the `VERSION_NUMBER` file.

### DIFF
--- a/src/Version/CMakeLists.txt
+++ b/src/Version/CMakeLists.txt
@@ -61,10 +61,10 @@ add_onnx_mlir_library(OMVersion
     LLVMSupport
 )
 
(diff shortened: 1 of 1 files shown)


AlexandreEichenberger:

AlexandreEichenberger:
Thanks for the PR. I think that @whitneywhtsang would be the best candidate to review this, thanks

whitneywhtsang (approved):
LGTM.

PR #1437, opened 2022-05-17
Moved a few alloca out of loops

AlexandreEichenberger:
Moved 2 or 3 alloca out of loops.

Signed-off-by: Alexandre Eichenberger <alexe@us.ibm.com>

PR #1485, opened 2022-06-07
Add BufferLoopHoistingPass after LowerAffinePass

whitneywhtsang:
According to https://llvm.org/docs/LangRef.html#alloca-instruction, the alloca instruction allocates memory on the stack frame of the currently executing function, to be automatically released when this function returns to its caller.

We can avoid over allocing by placing alloca instruction at the function scope and outside of any loops. 

https://github.com/onnx/onnx-mlir/pull/1397 and https://github.com/onnx/onnx-mlir/pull/1437 attempted to create the allocas outside of loops directly. However, it is not always possible to create outside of the whole loop nest. For example, when lowering an operation that is already in a loop nest, or when the outer loop is only created after the placement of the alloca. 

This PR adds BufferLoopHoistingPass after LowerAffinePass to hoist allocations out of loop nests to avoid stack overflow.


etiotto (approved):
LGTM

whitneywhtsang:

whitneywhtsang:

PR #1486, opened 2022-06-07
Fix a bug in MathBuilder::positiveInf and MathBuilder::negativeInf

### LINKED ISSUE #1270: ssd-10 model returns tensors with -1 elements
# Description

After calling `run_main_graph()` on the [ssd-10](https://github.com/onnx/models/blob/main/vision/object_detection_segmentation/ssd/model/ssd-10.onnx) model, the expected number of tensors (3) are returned in the output TensorList. However iterating over the returned tensors and calling `omTensorGetNumElems()` returns -1 for each instead of the expected values (800, 200, and 200 respectively).

# Notes: 
* I don't see this behavior when I try on the [mnist-8](https://github.com/onnx/models/blob/main/vision/classification/mnist/model/mnist-8.onnx) model
* We build onnx-mlir as a Release build with NNPA accelerator enabled

tungld:
`MathBuilder::positiveInf` and `MathBuilder::negativeInf` use `type.isSigned` to check an integer type, but the correct one should be `type.isSignlessInteger`.

This patch also adds op name in KrnlInstrumentOp, which helps know from which ONNX op the code is generated.

Resolves #1270 
 
### DIFF
--- a/src/Dialect/Mlir/DialectBuilder.cpp
+++ b/src/Dialect/Mlir/DialectBuilder.cpp
@@ -214,80 +214,101 @@ Value MathBuilder::constantIndex(int64_t val) const {
 }
 
 Value MathBuilder::negativeInf(Type type) const {
-  double value;
+  Value constant = nullptr;
   TypeSwitch<Type>(type)
-      .Case<Float16Type>(
-          [&](Type) { value = -std::numeric_limits<float>::infinity(); })
-      .Case<Float32Type>(
-          [&](Type) { value = -std::numeric_limits<float>::infinity(); })
-      .Case<Float64Type>(
-          [&](Type) { value = -std::numeric_limits<double>::infinity(); })
+      .Case<Float32Type>([&](Type) {
+        constant = b.create<arith::ConstantOp>(
+            loc, b.getF32FloatAttr(-std::numeric_limits<float>::infinity()));
+      })
+      .Case<Float64Type>([&](Type) {
+        constant = b.create<arith::ConstantOp>(
+            loc, b.getF64FloatAttr(-std::numeric_limits<double>::infinity()));
+      })
       .Case<IntegerType>([&](IntegerType type) {
         unsigned width = type.getWidth();
-        bool isSignless = type.isSignlessInteger();
+        bool isSignless = type.isSignless();
+        bool isSigned = type.isSigned();
+        int64_t value;
         switch (width) {
         case 8:
-          value = (isSignless) ? std::numeric_limits<int8_t>::min()
-                               : std::numeric_limits<uint8_t>::min();
+          value = (isSignless || isSigned)
+                      ? std::numeric_limits<int8_t>::min()
+                      : std::numeric_limits<uint8_t>::min();
           break;
         case 16:
-          value = (isSignless) ? std::numeric_limits<int16_t>::min()
-                               : std::numeric_limits<uint16_t>::min();
+          value = (isSignless || isSigned)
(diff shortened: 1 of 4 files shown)


src/Dialect/Mlir/DialectBuilder.cpp
```diff
@@ -224,23 +224,23 @@ Value MathBuilder::negativeInf(Type type) const {
           [&](Type) { value = -std::numeric_limits<double>::infinity(); })
       .Case<IntegerType>([&](IntegerType type) {
         unsigned width = type.getWidth();
-        bool isSigned = type.isSigned();
+        bool isSignless = type.isSignlessInteger();
         switch (width) {
         case 8:
-          value = (isSigned) ? std::numeric_limits<int8_t>::min()
-                             : std::numeric_limits<uint8_t>::min();
+          value = (isSignless) ? std::numeric_limits<int8_t>::min()
```
etiotto, line 230: I agree that a signless integer min value is conservatively `std::numeric_limits<int8_t>::min()`, however if the integer is not signless it might be signer or unsigned. If it is signed the min value should also be `std::numeric_limits<int8_t>::min()`, right ?

Suggests:

```
 value = (isSignless || isSigned) ? std::numeric_limits<int8_t>::min()
                                                      : std::numeric_limits<uint8_t>::min();

**Review by etiotto — COMMENTED** [review]
Can you also add a LIT test?

File: src/Dialect/Mlir/DialectBuilder.cpp (hunk 1 above)
Comment (AlexandreEichenberger) [reply src/Dialect/Mlir/DialectBuilder.cpp:230]: I believe MLIR treat signless as signed...

**Review by AlexandreEichenberger — APPROVED** [review]
LGTM, please make sure @etiotto's comment is properly addressed.

File: src/Dialect/Mlir/DialectBuilder.cpp (hunk 1 above)
Comment (etiotto) [reply src/Dialect/Mlir/DialectBuilder.cpp:230]: Sure, that is fine. My suggestion is to handle singed types. If the type is a signed integer the code change I suggested will use the min value for ont8_t rather than the min value for uint8_t.

Comment (tungld) [reply src/Dialect/Mlir/DialectBuilder.cpp]: [MLIR constant](https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Dialect/Arithmetic/IR/ArithmeticOps.td#L120) only supports signless at this moment. Anyway, I added `(isSignless || isSigned)` just in case signed constant will be supported in the future, but it would fail if users use signed integers at this moment.

File: src/Dialect/Mlir/DialectBuilder.cpp [hunk 2]
```diff
@@ -214,80 +214,101 @@ Value MathBuilder::constantIndex(int64_t val) const {
 }
 
 Value MathBuilder::negativeInf(Type type) const {
-  double value;
+  Value constant = nullptr;
```
tungld, line 217: We cannot cast the int64_t minimum value to double. So we have to create `arith::ConstantOp` for each case instead of a single `arith::ConstantOp` with `double value`.

tungld:
> Can you also add a LIT test?

Yes, I added them for ReduceMin and ReduceMax.

src/Dialect/Mlir/DialectBuilder.cpp
tungld: This message shows MLIR constant only supports signless.
```
within split at test-reduce.mlir:90 offset :4:10: error: 'arith.constant' op integer return type must be signless
  %cst = arith.constant 1: si64
```

tungld:

PR #1423, opened 2022-05-11
Lowering isnan to KNRL

mmoldawsky:
Signed-off-by: mmoldawsky <mattmoldawsky@gmail.com>

Including changes from other pull request because that one has yet to finish. Those changes are needed in order for these changes to work.

PR #1488, opened 2022-06-08
Add backend test for Is Nan

whitneywhtsang:
1. Add lowering for krnl.isnan
2. Add backend test for Is Nan


src/Conversion/KrnlToLLVM/KrnlUnaryMath.cpp
```diff
@@ -146,22 +146,30 @@ class KrnlUnaryMathOpLowering : public ConversionPattern {
...
+      llvmInType = FloatType::getF32(context);
     else if (inType.isF64())
-      llvmType = FloatType::getF64(context);
+      llvmInType = FloatType::getF64(context);
+    if (outType.isInteger(1))
+      llvmOutType = IntegerType::get(context, 1);
+    else if (outType.isF32())
+      llvmOutType = FloatType::getF32(context);
+    else if (outType.isF64())
+      llvmOutType = FloatType::getF64(context);
```
chentong319, line 160: Are those code intended to convert MLIR type to llvm type? Can llvm type converter be used?

hamptonm1:
@chentong319 So I figured out the issue as to why the backend test was failing....the syntax in `src/Conversion/KrnlToLLVM/KrnlUnaryMath.cpp` was wrong and once I made the fix, the test passed. Let me know your thoughts! I also added float16 and bfloat16 as previously discussed :)

src/Conversion/KrnlToLLVM/KrnlUnaryMath.cpp
AlexandreEichenberger: @hamptonm1 I assume that this is code that you got from an earlier PR. 
@chentong319 can you point to code that use the llvm type converter?

hamptonm1: Yeah this is what was previously there

hamptonm1:
@chentong319 @AlexandreEichenberger I got it to work!!! I finally figured this out. Feel free to review :)

src/Conversion/KrnlToLLVM/KrnlUnaryMath.cpp
```diff
@@ -126,7 +126,16 @@ struct MathFunctionName<KrnlAtanhOp> {
 template <>
 struct MathFunctionName<KrnlIsNaNOp> {
   static std::string functionName(mlir::Type type) {
-    if (type.isF32() || type.isF64())
+
+#if (__APPLE__)
+    if (type.isF32())
+      return "__isnanf";
```
AlexandreEichenberger, line 132: @hamptonm1 so basically, you discovered that on mac platforms, there is a `__isnanf` that provides the same functionality as `isnanf` on Linux? 
Perfect, that will be more efficient that converting to double, good job.

AlexandreEichenberger (approved):
LGTM, thanks for the fix, better solution than the one I proposed, congrats.

src/Conversion/KrnlToLLVM/KrnlUnaryMath.cpp
hamptonm1: Yes!

gongsu832: `isnanf` is obsolete. `isnan` replaces it. Also, this is a C99 standard macro so MacOS also supports it so there is no need to use `__isnanf` (macros/functions starting with __ are typically "internal" version that shouldn't be called directly).

hamptonm1: @gongsu832 If we use `isnan` for `f32` the test will fail for Linux but if we use `isnanf` it will pass and give us the correct results. Hence why the test failed before with the following code:

   if (type.isF32())
      return "isnan";
    if (type.isF64())
      return "isnan";
    llvm_unreachable("Unsupported type for isnan");
  }
};

hamptonm1: If we use `isnan` for `f32` on MacOS it will fail......... MacOs will fail with the following error:


`1517
Undefined symbols for architecture x86_64:
1518
  "__isnanf", referenced from:
1519
      _main_graph in test_isnan.o
1520
      __mlir_ciface_main_graph in test_isnan.o
1521
      _run_main_graph in test_isnan.o
1522
ld: symbol(s) not found for architecture x86_64
1523
clang: error: linker command failed with exit code 1 (use -v to see invocation)
1524`

hamptonm1: So `isnanf` is required and from my testing is certainly not obsolete. If we do not use `isnanf` the backend test will give us the following results:





```
E                   AssertionError: 
E                   Not equal to tolerance rtol=0.001, atol=1e-07
E                   
E                   Mismatch: 50%
E                    x: array([False, False, False, False])
E                    y: array([False,  True, False,  True])
```

hamptonm1: I understand your concern about using `__isnanf` but that internal call was recognized by MacOS and gave me back the correct test results which ensures the input matches the output. If you have a better idea or solution please feel free to share. I have been investigating this for awhile and I did not come across any other solution. But essentially if `isnanf` is not present, the test will fail.

hamptonm1: Here is the reference I used for my understanding: https://github.com/AccelerateHS/accelerate/issues/407

gongsu832: Then you must be doing something wrong. On both Linux and MacOS, I can compile and run the following code and both give the correct results:
```
#include <math.h>
#include <stdio.h>

int main() {
    float f = NAN;
    double d = NAN;
    printf("%s\n", isnan(f) ? "yes" : "no");
    printf("%s\n", isnan(d) ? "yes" : "no");
}
```

gongsu832: Maybe you forgot to include `math.h`?

hamptonm1: I did not do anything wrong.... This code already originated from @whitneywhtsang and @etiotto meaning the backend test was failing before I joined the community. So if it works for you, then why were the backend test failing for the all of the checks? The `isnan` test had to be commented out in order for repo to be stable.

hamptonm1: @gongsu832 Please click on my commit `Update inference backend test` and notice how all of these checks failed once I uncommented the test. If they passed, I would have just simply merged the code @whitneywhtsang had in this PR.

gongsu832: Sorry I'm not following you. I thought you added those `isnan` related code and tried to make the originally commented out backend test (now uncommented) to pass but you were having trouble with `isnan` not working on Linux. Or are you saying that those `isnan` related code were added by Whiteney? If so then it just means that she hasn't had the chance to properly debug this before she left. Either way, I'm trying to say that `isnan` should work for both Linux and MacOS and if it's not then something is wrong.

hamptonm1: The actual code for kernel lowering for `isnan` was added in June of last year: https://github.com/onnx/onnx-mlir/pull/1423/files. The code was already merged but the backend test was not added yet. The code that was merged in June used only `isnan` like you suggested. This particular PR (1488) only added the following 1) updating the llvm input and output types and 2) adding the backend test for `isnan`. I was asked to look into this given the backend test fails. So `isnan` does work but only for f64 and not f32. According to the link above and other google searches we are not the only ones who hit this problem. So I am open to hear about different solutions but what I am explaining is `isnan` does not work without also using `isnanf`. For Linux `isnanf` is still recognized according to the man page: https://linux.die.net/man/3/isnanf. If it is better to discuss then making comments on a PR I am open to talking whenever you are available :)

hamptonm1:
I created this draft PR to test the suggestion from @gongsu832 to make sure we try every possibility: https://github.com/onnx/onnx-mlir/pull/1976. So far I am seeing the `test_isnan_cpu` backend test fail. So Gong, I understand `isnan` should work since Linux says `isnan` covers floats and doubles but not sure why it is not working as designed. We can further explore once you get back from vacation. Enjoy!

PR #1490, opened 2022-06-09
Run the onnx model zoo with test_data_set_0 by default

tungld:
This patch adds some small changes to RunONNXModelZoo.py, where:
- Use `test_data_set_0` for each model since other datasets, e.g. `test_data_set_1`, `test_data_set_1` are sometimes ill-formed. For example,`test_data_set_1` of `ssd_mobilenet_v1_10` contains `input_1.pb` instead of `input_0.pb` (The input index should start from 0).
- Print the date time when the HTML output is generated for each model's output.

### DIFF
--- a/utils/RunONNXModelZoo.py
+++ b/utils/RunONNXModelZoo.py
@@ -285,9 +285,13 @@ def check_model(model_path, model_name, compile_args, report_dir):
         has_data_sets = False
         _, data_sets = execute_commands(
             ['find', tmpdir, '-type', 'd', '-name', 'test_data_set*'])
-        if (len(data_sets) > 0):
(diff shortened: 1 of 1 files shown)


utils/RunONNXModelZoo.py
```diff
@@ -315,8 +319,10 @@ def check_model(model_path, model_name, compile_args, report_dir):
         logger.debug("[{}] {}".format(model_name, msg))
 
         if args.Html:
+            from datetime import datetime
```
gongsu832, line 322: Move this to the top in front of joblib import?

tungld: Done.

gongsu832 (approved):
LGTM

tungld:

PR #1497, opened 2022-06-16
[NNPA] Numerical tests for Conv2D on NNPA

imaihal:
This PR adds numerical tests for Conv2D on NNPA. Also, this introduces a environment variable to configure numerical tests for NNPA. 

- Add new environment variable to set data range for input data and weights.
  This PR Introduces new environment variable `{test_name}_DATARANGE` to change default data range for input data and weights.  Currently it can be used only in Conv2D on NNPA. It is provided because the data range need to be changed to avoid accuracy issue in Conv2D on NNPA. Actual range used in the numerical test for Conv2D are written in cmake file.

- Introduce new environment variable to configure test parameters for Conv2D, Gemm, LSTM, and GRU
  To include only test cases supported by NNPA, this PR provides following parameters. They need to be set environment variable, `TEST_CONFIG`
  -  Conv2D: 
     - -dim=static: only static case because dynamic height and weight dimension are not supported
     - -dilaiton=1: only default dilations (= 1) 
     - -padding=valid_upper:  only VALID and SAME_UPPER pad type.
  - Gemm: 
     - -alpha=1: alpha should be one for Matmul of zDNN library
     - -beta=1: beta should be one for Matmul of zDNN library
  - LSTM:
     - -peephole=0: LSTM of zDNN library does not support peephole tensor
  - GRU:
     - -linearBeforeReset=1: GRU of zDNN library only support LinearBeforeReset=1   
 
- Test results 
[numerical-tests.txt](https://github.com/onnx/onnx-mlir/files/9390507/numerical-tests.txt)


tungld:
I feel that we have more and more environment variables for a single test. Could it be better to use a single environment variable, say TEST_CONV_OPTS="-static -d=1"? Similar for the others tests.

imaihal:
@tungld 
> Could it be better to use a single environment variable, say TEST_CONV_OPTS="-static -d=1 -padding=valid_upper"? Similar for the other tests.

I introduced an environment variable to configure numerical tests as written in top comment in this PR.

Now ready for review. Thanks!

docs/AccelNNPAHowToUseAndTest.md
```diff
@@ -22,22 +22,26 @@ cmake --build . --target check-onnx-lit
...
+Numerical tests for NNPA are provided in `test/accelerators/NNPA/numerical`. Currently tests for Conv2D, MatMul2D, Gemm, LSTM, and GRU are provided and run by using following command. These tests check if a zDNN instruction is generated in shared library in adition to accurary check.
 
 ```
 cmake --build . --config Release --target check-onnx-numerical-nnpa
 ```
 
 These tests uses the same test code with numerical tests for CPU (`test/modellib` and `test/numerial`), but uses different cmake file(`test/accelerator/NNPA/numerical/CMakeLists.txt`).
 
+##### Conv2D
+Since Conv2D of zDNN library does not support the case where dilations equal to one, `-dilation=1` option are added in `test/numerical/TestConv.cpp`. Also, since only VALID and SAME_UPPER as pading type are supported, `-padding=valid_upper is prepared to use the pading type. Currently dynamic height and weight dimension are not supported. So `-dim=static` are provided.
```
tungld, line 34: should `does not support` be `only supports`?

test/accelerators/NNPA/numerical/CMakeLists.txt
```diff
@@ -51,31 +63,42 @@ llvm_update_compile_flags(rapidcheck)
...
 set(TEST_LINK_LIBS rapidcheck ModelLib)
 
-# TODO: Support tests for Conv
-# add_numerical_unittest(TestConvNNPA
-#   ${ONNX_NUMERICALTEST_SRC_DIR}/TestConv.cpp
-#   LINK_LIBS PRIVATE ${TEST_LINK_LIBS}
-#   )
+# Conv2D
+set(TestConvNNPA_instruction zdnn_conv2d)
+set(TestConvNNPA_data_range 0.1)
```
tungld, line 68: Could you please explain general settings of `data_range` and `instruction` in `docs/AccelNNPAHowToUseAndTest.md`? Is it possible to use an upperbound and lowerbound for a range instead of a single value, which is more flexible?

docs/AccelNNPAHowToUseAndTest.md
```diff
@@ -22,22 +22,26 @@ cmake --build . --target check-onnx-lit
...
+Since `alpha` and `beta` should be one for Matmul of zDNN library, `-alpha=1` and `-beta=1` options are added in `test/numerical/TestGemm.cpp` and set in the CMakeLists.txt (`test/accelerator/NNPA/numerical/CMakeLists.txt`)
 
 ##### LSTM
-Since LSTM of zDNN library does not support peephole tensor, #ifdef directive `TEST_LSTM_NONEP_ONLY` are added in `test/numerial/TestLSTM.cpp` and set in the CMakeLists. Currently bidirectinal LSTM is not supported in NNPA, so, it is disabled by using #ifdef directive `TEST_RNN_NO_BIDIR`.
+Since LSTM of zDNN library does not support peephole tensor, `-peephole=0` are added in `test/numerial/TestLSTM.cpp` and set in the CMakeLists.
 
 ##### GRU
-Since GRU of zDNN library does not support LinearBeforeReset, #ifdef directive `TEST_GRU_L1` are added in `test/numerial/TestGRU.cpp` and set in the CMakeLists. Currently bidirectinal LSTM is not supported in NNPA, so, it is disabled by using #ifdef directive `TEST_RNN_NO_BIDIR`.
+Since GRU of zDNN library support only LinearBeforeReset=1, `-linerBeforeReset=1` option is added in `test/numerial/TestGRU.cpp` and set in the CMakeLists.
```
AlexandreEichenberger, line 44: > only LinearBeforeReset=1

Can you explain in english what is not supported? Current explanation is totological. Tx

docs/AccelNNPAHowToUseAndTest.md
```diff
@@ -22,22 +22,26 @@ cmake --build . --target check-onnx-lit
...
-Numerical tests for NNPA are provided in `test/accelerators/NNPA/numerical`. Currently tests for MatMul2D, Gemm, LSTM, and GRU are provided and run by using following command. These tests check if a zDNN instruction is generated in shared library in adition to accurary check.
+Numerical tests for NNPA are provided in `test/accelerators/NNPA/numerical`. Currently tests for Conv2D, MatMul2D, Gemm, LSTM, and GRU are provided and run by using following command. These tests check if a zDNN instruction is generated in shared library in adition to accurary check.
 
 ```
 cmake --build . --config Release --target check-onnx-numerical-nnpa
 ```
 
 These tests uses the same test code with numerical tests for CPU (`test/modellib` and `test/numerial`), but uses different cmake file(`test/accelerator/NNPA/numerical/CMakeLists.txt`).
 
+##### Conv2D
```
AlexandreEichenberger, line 33: Side note, have you though of adding an `-maccel=NNPA` option that then would automatically set the current specialized options?

test/modellib/ConvModel.cpp
```diff
@@ -119,6 +119,19 @@ bool Conv2DLibBuilder::build() {
...
+    return false;
+  list[0] =
+      omTensorCreateWithRandomData<float>({N, C, H, W}, -dataRange, dataRange);
+  list[1] = omTensorCreateWithRandomData<float>(
+      {C, C, kH, kW}, -dataRange, dataRange);
+  inputs = omTensorListCreateWithOwnership(list, num, true);
+  return inputs && list[0] && list[1];
+}
+
 bool Conv2DLibBuilder::prepareInputs() {
```
AlexandreEichenberger, line 135: can you redefine this one as calling the new prepareInputs with `dataRange = 1.0` please?

test/modellib/ModelLib.cpp
```diff
@@ -114,6 +114,25 @@ void ModelLibBuilder::setRandomNumberGeneratorSeed(const std::string &envVar) {
   }
 }
 
+std::map<std::string, std::string> ModelLibBuilder::getTestConfigFromEnv(
```
AlexandreEichenberger, line 117: Thanks for adding this functionality in the model.

test/numerical/TestConv.cpp
```diff
@@ -29,20 +27,19 @@ bool isOMConvTheSameAsNaiveImplFor(const int N, const int C, const int H,
...
   static int testNum = 0;
-  if (DEBUG)
-    printf(
-        "attempt %d with N %d, C %d, H %d, W %d, kH %d, kW %d, pHBegin %d, "
-        "pHEnd %d, pWBegin %d, pWEnd %d, autopad %s, isDynamic %d, stride %d, "
-        "dilation %d\n",
-        ++testNum, N, C, H, W, kH, kW, pHBegin, pHEnd, pWBegin, pWEnd,
-        Conv2DLibBuilder::getAutoPadName(autoPad).c_str(), isDynamic, stride,
-        dilation);
+  printf("attempt %d with N %d, C %d, H %d, W %d, kH %d, kW %d, pHBegin %d, "
```
AlexandreEichenberger, line 30: Are you sure you don't want to keep this printf only under DEBUG?

test/numerical/TestConv.cpp
```diff
@@ -61,21 +58,50 @@ int main(int argc, char *argv[]) {
...
+  std::string paddingType;
+  if (opts["-dim"] == "static") {
+    std::cout << "Dimension type from env: \"" << opts["-dim"] << "\"\n";
+    dimType = 1; // only static
+  } else {
+    dimType = 2; // static and dynamic (default)
+  }
+  if (opts["-dilation"] == "1") {
+    std::cout << "Dilation from env: \"" << opts["-dilation"] << "\"\n";
+    maxDilation = 2; // dilation = 1
```
AlexandreEichenberger, line 75: It might be easier to state in the comment that `maxDilation` in an exclusive upper bound.

test/numerical/TestConv.cpp
```diff
@@ -61,21 +58,50 @@ int main(int argc, char *argv[]) {
       argc, argv, "TestConv\n", nullptr, "TEST_ARGS");
   std::cout << "Target options: \""
             << getCompilerOption(OptionKind::TargetAccel) << "\"\n";
-
-  // Had to explicitly iterate over dynamic as otherwise the random algorithm
+  // Get configurations from an environment variable
+  std::map<std::string, std::string> opts =
+      ModelLibBuilder::getTestConfigFromEnv("TEST_CONFIG");
+  // Set configuration for test
+  int dimType, maxDilation;
```
AlexandreEichenberger, line 65: The code would be more compact if you set `dimType` and `maxDilation` to their default value here, and then only redefine them when an `opts` override them. I find this style easier to read.

test/numerical/TestGRU.cpp
```diff
@@ -52,11 +52,22 @@ int main(int argc, char *argv[]) {
       argc, argv, "TestGRU\n", nullptr, "TEST_ARGS");
   std::cout << "Target options: \""
             << getCompilerOption(OptionKind::TargetAccel) << "\"\n";
+  // Get configurations from an environment variable
+  std::map<std::string, std::string> opts =
+      ModelLibBuilder::getTestConfigFromEnv("TEST_CONFIG");
+  // Set configuration for test
+  int minL;
```
AlexandreEichenberger, line 59: please set the default here, and remove the 'else` clauses, tx.

test/numerical/TestGemm.cpp
```diff
@@ -82,26 +82,37 @@ int main(int argc, char *argv[]) {
       argc, argv, "TestGemm\n", nullptr, "TEST_ARGS");
   std::cout << "Target options: \""
             << getCompilerOption(OptionKind::TargetAccel) << "\"\n";
-
+  // Get configurations from an environment variable
+  std::map<std::string, std::string> opts =
+      ModelLibBuilder::getTestConfigFromEnv("TEST_CONFIG");
+  // Set configuration for test
+  int maxHasAlpha, maxHasBeta;
```
AlexandreEichenberger, line 89: same here, set default here, remove `else` clauses.

AlexandreEichenberger (commented):
looking good, please make small changes and it should be good to go.

docs/AccelNNPAHowToUseAndTest.md
imaihal: Do you mean using only `-maccel=NNPA` without using the environment variable? Yes. it should be possible.

test/numerical/TestConv.cpp
imaihal: I removed it for consistency with [other tests](https://github.com/onnx/onnx-mlir/tree/main/test/numerical) The other tests don't have `DEBUG`. Even without `DEBUG`, the print statement is not output when running numerical tests using cmake command. To print it out, we need to add `--verbose` in [this line](https://github.com/onnx/onnx-mlir/blob/main/test/numerical/CMakeLists.txt#L8).

imaihal: I will add `DEBUG` in all tests if requested.

docs/AccelNNPAHowToUseAndTest.md
imaihal: Yes. Updated. Thanks!

test/accelerators/NNPA/numerical/CMakeLists.txt
imaihal: Updated docs.

docs/AccelNNPAHowToUseAndTest.md
imaihal: Done

test/modellib/ConvModel.cpp
imaihal: Done.

test/numerical/TestConv.cpp
imaihal: Done

test/numerical/TestConv.cpp
imaihal: Done. Thanks!

test/numerical/TestGemm.cpp
imaihal: Done

docs/AccelNNPAHowToUseAndTest.md
AlexandreEichenberger: @imaihal 
I leave it up to you. The solution you have is clearly working, and the cmake does add the flags. It is just if we have to test one manually, we have to turn on the right flags. This is why I though that we could just use the `-maccel=NNPA` to automatically set all the options. Again, either way works, whatever you like is good.

test/modellib/ConvModel.cpp
AlexandreEichenberger: great, less redundant code

AlexandreEichenberger (approved):
LGTM, will let you judge if you want to keep the code as is or simply use `-maccel=NNPA` to setup the individual internal flags (which I would definitely keep, since maybe NNPA2 will remove some of the restrictions.

AlexandreEichenberger (approved):
LGTM, thanks for the great work

test/numerical/TestGRU.cpp
imaihal: Done. Thanks!

docs/AccelNNPAHowToUseAndTest.md
imaihal: @AlexandreEichenberger  I agree with you. I updated code to set them automatically when setting `--maccel=NNPA`. I kept original environment variables since they may be useful in the future.

PR #1499, opened 2022-06-17
Add numerical test for onnx.Scan op

negiyas:
Add numerical test for onnx.Scan op
Currently the execution results and reference answers do not match when the batch size != 2.
We need confirm the reference answers are correct with another way.

TODOs.
- [x] Currently the execution results and reference answers do not match, and need confirm the reference answers are correct.
- [x] Fix the reference answer or the onnx-mlir implementation to match the outputs and the reference answers.


test/modellib/ScanModel.cpp
```diff
@@ -0,0 +1,221 @@
...
+bool ScanLibBuilder::prepareInputs() {
+  constexpr int num = 2;
+  OMTensor **list = (OMTensor **)malloc(num * sizeof(OMTensor *));
+  if (!list)
+    return false;
+  list[0] =
+      omTensorCreateWithRandomData<float>(llvm::makeArrayRef(initialShape));
+  list[1] = omTensorCreateWithRandomData<float>(llvm::makeArrayRef(xShape));
+  // Compute reference. Scan with onnx.Add
+  for (int64_t b = 0; b < B; ++b) {
```
AlexandreEichenberger, line 131: Why this code here and not in the prepareInputs with range. Also, you could use a single version of the code. I would suggest to define
```C++
template <typename T>
OMTensor *omTensorCreateWithRandomData(
    const std::vector<int64_t> &dataSizes, T lbound = -1.0, T ubound = 1.0);

```
in OMTensorHelper.

I would suggest to define in that file a static def of the default values,
e.g.
```C++
const static float omDefaultRangeBound = 1.0;
```
and then you could use that value there, namely define prepareInputs(float range = omDefaultRangeBound);

If you really wanted, one could even define a default for float and another default for int... but really 0 representation is the same for float and ints...

test/modellib/ScanModel.cpp
```diff
@@ -0,0 +1,221 @@
...
+    for (int64_t b = 0; b < B; ++b) {
+      refVal = omTensorGetElem<float>(init, {b, i});
+      for (int64_t s = 0; s < S; s++) {
+        refVal += omTensorGetElem<float>(x, {b, s, i});
+        omTensorGetElem<float>(refz, {s, i}) = refVal;
+      }
+    }
+    omTensorGetElem<float>(refy, {i}) = refVal;
+  }
+#if 0
```
AlexandreEichenberger, line 201: please remove in final version, or add a #define with a variable name that can set for debugging.

test/modellib/ScanModel.cpp
```diff
@@ -0,0 +1,221 @@
...
+  xShape = {B, S, I};
+
+  moduleIR = std::regex_replace(
+      testScanIdentityAdd9, std::regex("%B"), std::to_string(B));
+  moduleIR = std::regex_replace(moduleIR, std::regex("%S"), std::to_string(S));
+  moduleIR = std::regex_replace(moduleIR, std::regex("%I"), std::to_string(I));
+  return true;
+}
+
+bool ScanLibBuilder::compileAndLoad() {
```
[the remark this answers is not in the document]
negiyas: Thank you for the comment.
The onnx.Scan is similar to the onnx.Loop. This PR uses a similar to the "onnx.Loop" case in order to define onnx.Scan's body part (in "^bb0" below) consisting of  multiple onnx operations in the simplest way.
I added the comment to explain it. Thank you.
```
    %1:2 = "onnx.Scan"(%arg0, %arg1) ({
    ^bb0(%body_arg0: tensor<%Ixf32>, %body_arg1: tensor<%Ixf32>):
      %2 = "onnx.Add"(%body_arg0, %body_arg1) :
           (tensor<%Ixf32>, tensor<%Ixf32>) -> tensor<%Ixf32>
      %3 = "onnx.Identity"(%2) : (tensor<%Ixf32>) -> tensor<%Ixf32>
      "onnx.Return"(%2, %3) : (tensor<%Ixf32>, tensor<%Ixf32>) -> ()
    }) {num_scan_inputs = 1 : si64} :
        (tensor<%Ixf32>, tensor<%Sx%Ixf32>)
        -> (tensor<%Ixf32>, tensor<%Sx%Ixf32>)
```

test/modellib/ScanModel.cpp
negiyas: I removed the duplicated part. Thank you!

test/modellib/ScanModel.cpp
```diff
@@ -0,0 +1,206 @@
...
+      mlir::parseSourceString<ModuleOp>(moduleIR, &ctx);
+  OwningOpRef<ModuleOp> module(std::move(moduleOp));
+  if (compileModule(module, ctx, sharedLibBaseName, onnx_mlir::EmitLib) != 0)
+    return false;
+  exec = new ExecutionSession(
+      onnx_mlir::getTargetFilename(sharedLibBaseName, onnx_mlir::EmitLib));
+  return exec != nullptr;
+}
+
+bool ScanLibBuilder::compileAndLoad(const onnx_mlir::CompilerOptionList &list) {
```
AlexandreEichenberger, line 106: Why redefine this one here. It appears to be identical to the superclass
```C++
bool ModelLibBuilder::compileAndLoad(
    const onnx_mlir::CompilerOptionList &list) {
  if (setCompilerOptions(list) != CompilerSuccess)
    return false;
  return compileAndLoad();
}
```

test/modellib/ScanModel.cpp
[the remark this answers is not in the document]
AlexandreEichenberger: It appears to be identical to the superclass
```C++
bool ModelLibBuilder::compileAndLoad() {
  OwningOpRef<ModuleOp> moduleRef(module);
  if (compileModule(moduleRef, ctx, sharedLibBaseName, onnx_mlir::EmitLib) !=
      CompilerSuccess)
    return false;
  std::string libFilename =
      getTargetFilename(sharedLibBaseName, onnx_mlir::EmitLib);
  exec = new ExecutionSession(libFilename);
  return exec != nullptr;
}
```
if so, remove code and declaration in subclass.

test/modellib/ScanModel.cpp
```diff
@@ -0,0 +1,206 @@
...
+  if (setCompilerOptions(list) != 0)
+    return false;
+  return compileAndLoad();
+}
+
+const static float omDefaultRangeBound = 1.0;
```
AlexandreEichenberger, line 112: Please define this one in ModelLib.hpp, and use it there as well as here.
That way, if we ever changed the default value, it would also be changed here. Does that make sense? I should have done it sometime ago, now would be a good time to do it. Tx

test/modellib/ScanModel.cpp
negiyas: @AlexandreEichenberger Removed "compileAndLoad(const onnx_mlir::CompilerOptionList &list)." Thank you.

test/modellib/ScanModel.cpp
negiyas: Thank you for the comments.
I removed the ScanLibBuilder::compileAndLoad() code from ScanModel.cpp, and
moved omDefaultRangeBound definition into ModelLib.hpp.

test/modellib/ScanModel.cpp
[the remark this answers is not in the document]
negiyas: I removed the ScanLibBuilder::compileAndLoad() code from ScanModel.cpp. Thank you for the comment.

negiyas:
@AlexandreEichenberger Yes. This PR is ready for review. Thank you.

negiyas:

AlexandreEichenberger (approved):
LGTM, code is now so more compact since you are reusing the super class methods... thanks for doing the update, hope you also appreciate how more compact the code is.

negiyas:

negiyas:

test/modellib/ModelLib.hpp
```diff
@@ -210,6 +212,25 @@ class GemmLibBuilder : public ModelLibBuilder {
   llvm::SmallVector<int64_t, 2> aShape, bShape, cShape;
 };
 
+class ScanLibBuilder : public ModelLibBuilder {
+public:
+  ScanLibBuilder(const std::string &modelName, const int /*seq=*/S,
+      const int /*inner-dim=*/I, const int /*batch=*/B, const bool is_v8);
+  bool build() final;
+  bool prepareInputs() final;
+  bool prepareInputs(float dataRange);
```
AlexandreEichenberger, line 221: you could have only one def, 
```C++
bool prepareInputs(float dataRange = omDefaultRangeBound);
```

negiyas:

negiyas:

negiyas:

PR #1501, opened 2022-06-21
Record all the backend test cases

chentong319:
This PR records all the backend test cases, models and nodes, in file all_test_names.txt. User can use this file to add new backend test cases without going into third_party/onnx code. Also, the file reveals what test cases are not supported by onnx-mlir so far. When onnx is updated to a new version, we can generate a new version of this file and compare against the older one to find out the changes in backend test cases.


gongsu832:
It's a bit confusing to call this a test since all it does is to save the test names in a file. It feels more like a tool belonging to `utils`.

AlexandreEichenberger (commented):
I see that you decided not to change the logic of the tests but provide instead a list of the tests. That works for me.

Would you be able to create a list that only contains the new ones, that is the ones from the current `all_test_names.txt` minus the one that we already cover? Maybe in `all_test_names_new.txt`?

AlexandreEichenberger:
are you planning to do a version where only the missing tests are listed (avail - already here)?

AlexandreEichenberger:
> are you planning to do a version where only the missing tests are listed (avail - already here)?

any response @chentong319 ?

chentong319:
> are you planning to do a version where only the missing tests are listed (avail - already here)?

Not in this PR. Need to find a unified way to specify the test case for cpu and accelerator.

docs/Testing.md
```diff
@@ -19,6 +19,7 @@ cmake --build . --config Release --target check-onnx-backend[-jni]
 ``` 
 Packages, such as third_party/onnx, needs to be installed to run the backend test. JNI test requires the jsoniter jar which is downloaded from its maven repository by default if no installed version is found on the system. If the user turns on the cmake option `ONNX_MLIR_BUILD_JSONITER` when building ONNX-MLIR, the jsoniter jar will be built locally from the source cloned from its github repository. Note that building jsoniter jar locally requires the maven build tool to be installed.
```
AlexandreEichenberger, line 20: Do we have a link on how third_party/onnx is installed? I did not find any help on this. Some unrelated site to us mention on mac to use `pip install -e third_party/onnx` which is what I do too.

I would add "[[In addition,  ]]JNI test requires the jointer jar" to be a bit more explicit that this is in addition to the one before.

How is the `test/backend/all_test_names.txt` file generated. Custom target or each time?

test/backend/all_test_names.txt
```diff
@@ -0,0 +1,874 @@
+from onnx 1.11.0
```
AlexandreEichenberger, line 1: Do you mind to include? If so, would add #automatically generated, do not modify... and maybe list the target that was use to create it?

docs/Testing.md
chentong319: I do not use -e. Here is the explanation: https://stackoverflow.com/questions/41535915/python-pip-install-from-local-dir

AlexandreEichenberger (approved):
LGTM, thanks for all the updates

PR #1505, opened 2022-06-23
Add fix to cmake file for multiple accelerators

Silvan-K:
This PR includes a minor fix to one of the cmake files to support enabling multiple accelerators. Previously, the build system would fail if one added a new accelerator in `src/Accelerators/XXX` and enabled it in addition to `NNPA` via `-DONNX_MLIR_ACCELERATORS="NNPA;XXX"`


caoimhinuibrian (approved):
LGTM

Issue #1506, opened 2022-06-23
Check of runtime input signature


AlexandreEichenberger:
@cjvolzka asked what check we do when the signature reports a runtime var. Right now in the code I see in KrnlEntryPoint, line 501:

```c++
        if (dim == -1)
          continue; // do not verify dynamic dimensions.
```

I think we could reasonably check for a non-negative value, i.e. return an error if the runtime dim is strictly smaller than 0.

Sounds reasonable?

cjvolzka:
> I think we could reasonably check for a non-negative value, i.e. return an error if the runtime dim is strictly smaller than 0.

A couple of follow up thoughts:
* Should we return an error if the values is less than 1 instead of 0? Is there any case where a dimension size of zero makes sense at runtime?
* Do we need any special handling to confirm the shape is integers. This probably isn't an issue for strictly typed languages like C/C++/Java. However Python could define a shape like `[1.1, 2, 3]`. In that scenario would the verifier catch it, something implicitly/explicitly down casts it, or we blow up?

AlexandreEichenberger:
> Do we need any special handling to confirm the shape is integers. This probably isn't an issue for strictly typed languages like C/C++/Java. However Python could define a shape like [1.1, 2, 3]. In that scenario would the verifier catch it, something implicitly/explicitly down casts it, or we blow up?

In the interface with python, I don't think the sizes can be anything but ints. See https://github.com/onnx/onnx-mlir/tree/main/docs/mnist_example.

```python
# Create an input arbitrarily filled of 1.0 values (file has the actual values).
input = np.full((1, 1, 28, 28), 1, np.dtype(np.float32))
# Run the model. It is best to always use the [] around the inputs as the inputs
# are an vector of numpy arrays.
outputs = session.run([input])
```

Since the vector input is created with a np.xxx command, it will have int sizes.

tungld:
> Should we return an error if the values is less than 1 instead of 0? Is there any case where a dimension size of zero makes sense at runtime?

Dimension size of 0 (e.g. tensor<0xf32>) represents a scalar value in both ONNX and MLIR. E.g., in ONNX: https://github.com/onnx/onnx/blob/main/docs/IR.md#tensor-definition.

tungld:
Implemented this check in https://github.com/onnx/onnx-mlir/pull/1508. Closing ..

PR #1498, opened 2022-06-17
Create LLVMBuider to simplify generating LLVM code - Part 1

tungld:
This patch creates LLVMBuilder and uses it in KrnlToLLVM.

Another PR will follow this to use LLVMBuilder in ZLowToLLVM.

PR #1508, opened 2022-06-24
Use LLVMBuider to simplify generating LLVM code - Part 2

tungld:
Thi patch continues PR #1498 and creates a helper function IfThenElse in LLVMBuilder.

This patch also adds a check to verify non-negative dimensions of the input tensors, reported here #1506

### DIFF
--- a/src/Conversion/KrnlToLLVM/ConvertKrnlToLLVM.cpp
+++ b/src/Conversion/KrnlToLLVM/ConvertKrnlToLLVM.cpp
@@ -304,16 +304,11 @@ void genSignatureFunction(ModuleOp &module,
     b.setInsertionPointToStart(entryBlock);
 
     Value zeroI32 = create.llvm.constant(i32Type, (int64_t)0);
-    Value oneI64 = create.llvm.constant(i64Type, (int64_t)1);
 
-    // 2.1 A buffer to keep a pointer pointing to the return signature string.
-    Value ptrToReturnSig =
-        create.llvm._alloca(i8PtrPtrTy, oneI64, /*alignment=*/0);
-
-    // 2.2 The name of the entry point that we want to return its signature.
+    // 2.1 The name of the entry point that we want to return its signature.
     Value input = entryBlock->getArgument(0);
 
-    // 2.3 Emit code to find the signature of the given entry point.
+    // 2.2 Emit code to find the signature of the given entry point.
     // Iterate over the list of the entry points and check string equality.
     for (uint64_t j = 0; j < numOfEntryPoints; ++j) {
       LLVM::GlobalOp globalEntryPoint = entryGlobalOps[j];
@@ -345,9 +340,7 @@ void genSignatureFunction(ModuleOp &module,
           [&](LLVMBuilder &createLLVM) {
             Value sigAddr = createLLVM.addressOf(globalSignature);
             Value sigI8Ptr = createLLVM.bitcastI8Ptr(sigAddr);
-            createLLVM.store(sigI8Ptr, ptrToReturnSig);
-            Value res = createLLVM.load(ptrToReturnSig);
-            createLLVM._return(res);
+            createLLVM._return(sigI8Ptr);
           });
     }
 
--- a/src/Conversion/KrnlToLLVM/ConvertKrnlToLLVM.cpp
+++ b/src/Conversion/KrnlToLLVM/ConvertKrnlToLLVM.cpp
@@ -267,36 +267,21 @@ void genSignatureFunction(ModuleOp &module,
     Value numOfEntryPoints = entryBlock->getArgument(0);
     // If the argument is not NULL, update its value to return the number of
     // entry points.
-    Block *condBlock = b.getInsertionBlock();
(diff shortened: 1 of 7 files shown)


src/Dialect/Mlir/DialectBuilder.hpp
```diff
@@ -336,6 +345,21 @@ struct LLVMBuilder final : DialectBuilder {
...
+  /// llvm.condBr cond, ^thenBlock, ^elseBlock
+  /// ^thenBlock:
+  ///   thenBody
+  /// ^elseBlock:
+  ///   elseBody
+  /// ^mainBlock
+  ///   ...
+  /// ```
+  void ifThenElse(mlir::function_ref<mlir::Value(LLVMBuilder &createLLVM)> cond,
+      mlir::function_ref<void(LLVMBuilder &createLLVM)> thenFn,
```
AlexandreEichenberger, line 361: Maybe you could create a special name for this type? for example `using` like here `  using ApiRegistry = std::map<RuntimeAPI::API, RuntimeAPI>;`

src/Conversion/KrnlToLLVM/ConvertKrnlToLLVM.cpp
```diff
@@ -267,36 +267,21 @@ void genSignatureFunction(ModuleOp &module,
...
-    create.llvm.store(noep, numOfEntryPointsPtr);
-    b.create<LLVM::BrOp>(loc, ValueRange(), endBlock);
-
-    // Emit code for the false block: do nothing.
-    b.setInsertionPointToStart(falseBlock);
-    b.create<LLVM::BrOp>(loc, ValueRange(), endBlock);
-
-    // Emit code for the end block to return the entry point array.
-    b.setInsertionPointToStart(endBlock);
+    create.llvm.ifThenElse(/*cond=*/
```
AlexandreEichenberger, line 270: great saving in number of lines and so much more readable

src/Dialect/Mlir/DialectBuilder.cpp
```diff
@@ -942,4 +953,47 @@ FlatSymbolRefAttr LLVMBuilder::getOrInsertSymbolRef(ModuleOp module,
...
+  // Emit code for the IF block.
+  b.setInsertionPointToEnd(ifBlock);
+  Value condVal = cond(createLLVM);
+
+  // Branch the block into the THEN and ELSE blocks.
+  condBr(condVal, thenBlock, {}, elseBlock, {});
+
+  // Emit code for the THEN block.
+  b.setInsertionPointToStart(thenBlock);
+  thenFn(createLLVM);
```
AlexandreEichenberger, line 983: You should create a new builder with the local builder, see what is done in SCF
```C++
          SCFBuilder scfBuilder(childBuilder, childLoc);
```
It was the recommendation of MLIR to build a local builder that is specific to the loc.

src/Dialect/Mlir/DialectBuilder.cpp
```diff
@@ -942,4 +953,47 @@ FlatSymbolRefAttr LLVMBuilder::getOrInsertSymbolRef(ModuleOp module,
   return SymbolRefAttr::get(b.getContext(), funcName);
 }
 
+void LLVMBuilder::ifThenElse(
+    mlir::function_ref<Value(LLVMBuilder &createLLVM)> cond,
+    mlir::function_ref<void(LLVMBuilder &createLLVM)> thenFn,
+    mlir::function_ref<void(LLVMBuilder &createLLVM)> elseFn) const {
```
AlexandreEichenberger, line 959: Can you create a custom type using a "using"

src/Dialect/Mlir/DialectBuilder.cpp
tungld: SCF is a bit different here. We are using the builder of `scf.IfOp` that uses `function_ref`, e.g. `function_ref<void(OpBuilder &, Location)> thenBuilder`, so we need to use the local location. But, if we use another builder of `scf.IfOp` that does not use `function_ref` but `setInsertionPointToStart `, the local location is not necessary, e.g. https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/SparseTensor/Transforms/Sparsification.cpp#L746.

IfThenElse here for LLVM is similar to the later one, where we just emit a sequence of ops to represent ifThenElse via LLVM::CondBrOp. LLVM dialect does not have ifOp.

src/Dialect/Mlir/DialectBuilder.hpp
tungld: Yes, I created these aliases inside LLVMBuilder:
```C
using voidFuncRef = mlir::function_ref<void(LLVMBuilder &createLLVM)>;
using valueFuncRef = mlir::function_ref<mlir::Value(LLVMBuilder &createLLVM)>;
```
Please let me know if you prefer another name. Thank!

src/Dialect/Mlir/DialectBuilder.cpp
tungld: Done. Thanks!

AlexandreEichenberger (approved):
LGTM, tx for all the changes

PR #1509, opened 2022-06-27
Update t5-encoder model accoring to the latest onnx model zoo

tungld:
[A recent commit in onnx](https://github.com/onnx/models/pull/534) updates some broken test_data_sets including `t5-decoder-with-lm-head-12`, `t5-encoder-12`.

This patch updates the configuration for `t5-decoder-with-lm-head-12`, `t5-encoder-12` to use the provided test_data_set instead of randomly generate input data by passing shape_info.

With this patch, `t5-decoder-with-lm-head-12`, `t5-encoder-12` are now passed.


gongsu832 (approved):
LGTM

PR #1512, opened 2022-06-28
Elide constant operations by setting elideLargeElementsAttrs of OpPrintingFlags

tungld:
When using `--EmitONNXBasic`, `--EmitONNXIR`, `--EmitMLIR`, a `.tmp` file is generated where ONNXConstantOp and KrnlGlobalOp are elided so that the generated file is small and easy to examine. However, the current approach is specific to ONNXConstantOp and KrnlGlobalOp, which is not scalable if, say, an accelerator has a new kind of constant op.

This patch uses another approach that just simply set the flag `elideLargeElementsAttrs` of OpPrintingFlags, so that all Attribute (no matter of which operation) will be elided. It should work for every operation.

### DIFF
--- a/src/Compiler/CompilerUtils.cpp
+++ b/src/Compiler/CompilerUtils.cpp
@@ -652,11 +652,14 @@ int processInputArray(const void *onnxBuffer, int bufferSize,
 }
 
 // Return 0 on success, error code on error.
-int outputCode(
-    mlir::OwningOpRef<ModuleOp> &module, std::string filenameWithExt) {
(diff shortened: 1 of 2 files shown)


AlexandreEichenberger (approved):
LGTM, wonder if having a compile time option for onnx-mlir would not be helpful.

tungld:
> LGTM, wonder if having a compile time option for onnx-mlir would not be helpful.

`--EmitONNXBasic`, `--EmitONNXIR`, `--EmitMLIR`, etc. are points to output intermediate code in onnx-mlir and we support elision with them. So I feel it is enough. We can add an option later when the need is more obvious.

PR #1513, opened 2022-06-28
Lower LoopOp to SCF.While when condition is loop variant

chentong319:
There are two optional control variable for onnx.LoopOp:  induction variable and termination condition. When LoopOp is controlled by induction variable, onnx.LoopOp can be lowered to krnl.iterate. Otherwise, SCF.While has to be lowered to SCF.While.
This PR implemented the lowering ONNX Loop to SCF.While. 
First, condition of LoopOp is checked to determine whether SCF.While is needed. 
Then the transformation is done in the similar way for krnl.iteration, but with duplicated code. There are three major differences in the code generation:
1. Induction variable has to be explicit in SCF.While, while it is implicit in krnl.iterate. This makes the list of block arguments and return is different.
2. The containing region is different.  In krnl.terate, the onnx loop body is inserted in to the then part of a SCF.If, which has a terminator. There is no need to add SCF.If when SCF.While is used.
3. Optional input of LoopOp is handled in this PR. The previous code assumed they are present.

The targeted WhileOp looks like the following:
Assume the argument list, ARGTYPES,  is the type of {LoopOp.M(), LoopOp.v_initial()}
outputs = list of memref created from LoopOp.v_final()

While.Op({ 0, loopOp.v_initial(): ARGTYPES) ({
      BeforeRegion:
      bb0(a0, ..., an: ARGTYPES): {
           %condition = a0 < LoopOp.M() && a1;
           scf.condition(%condition) outputs 
      } ,
     AfterRegion:
      bb0(b0, ..., bn: ARGTYPES) : {
          "krnl.region"() {
                 LoopOp loop body with parameters replaced with b0, ..., bn)
                 copy the generated loop-carried value into outputs
          }
          iv = b0+1
          store iv to ivMemref
          store new condition to condMemref
          scf.yield(ivMemref, condMemref, outputs). // the scan output excluded
       }) : -> (types of outputs)

Test:
Passed the following test case, modified from test_loop11.onnx. The program ran correctly with different value provided to the last input %m. I added condition computation (%11 and %12) based on an input tensor (%m)
```
module attributes {llvm.data_layout = "e-m:o-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128", llvm.target_triple = "x86_64-apple-darwin20.3.0"} {
  func @main_graph(%arg0: tensor<i64>, %arg1: tensor<i1>, %arg2: tensor<1xf32>, %m : tensor<1xf32>) -> tensor<1xf32>  {
    ^bb0(%arg3: tensor<i64>, %arg4: tensor<i1>, %arg5: tensor<1xf32>):
      %1 = "onnx.Identity"(%arg4) : (tensor<i1>) -> tensor<i1>
      %2 = "onnx.Constant"() {value = dense<[1.000000e+00, 2.000000e+00, 3.000000e+00, 4.000000e+00, 5.000000e+00]> : tensor<5xf32>} : () -> tensor<5xf32>
      %3 = "onnx.Constant"() {value = dense<1> : tensor<i64>} : () -> tensor<i64>
      %4 = "onnx.Add"(%arg3, %3) : (tensor<i64>, tensor<i64>) -> tensor<*xi64>
      %5 = "onnx.UnsqueezeV11"(%arg3) {axes = [0]} : (tensor<i64>) -> tensor<*xi64>
      %6 = "onnx.UnsqueezeV11"(%4) {axes = [0]} : (tensor<*xi64>) -> tensor<*xi64>
      %7 = "onnx.NoValue"() {value} : () -> none
      %8 = "onnx.NoValue"() {value} : () -> none
      %9 = "onnx.Slice"(%2, %5, %6, %7, %8) : (tensor<5xf32>, tensor<*xi64>, tensor<*xi64>, none, none) -> tensor<*xf32>
      %10 = "onnx.Add"(%arg5, %9) : (tensor<1xf32>, tensor<*xf32>) -> tensor<1xf32>
      %11 = "onnx.Less"(%10, %m) : (tensor<1xf32>, tensor<1xf32>) -> tensor<1xi1>
      %12 = "onnx.SqueezeV11"(%11) {axes = [0]} : (tensor<1xi1>) -> tensor<i1>
      onnx.Return %12, %10 : tensor<i1>, tensor<1xf32>
    }) {input_names = ["iter_count", "cond_in", "y_in"], output_names = ["cond_out", "y_out", "scan_out"]} : (tensor<i64>, tensor<i1>, tensor<1xf32>) -> tensor<1xf32>
    return %0 : tensor<1xf32>
  }
}
```
Future work:
1. Scan output is not supported in this PR. Since the trip count for while loop is unknown before the execution, previous approach does not work. Plan to use dynamic data, for instant sequence, to handle this issue in another PR.
2. It is possible to reduce the code duplication by define more functions/lambda functions.

### DIFF
--- a/src/Conversion/ONNXToKrnl/ControlFlow/Loop.cpp
+++ b/src/Conversion/ONNXToKrnl/ControlFlow/Loop.cpp
@@ -490,11 +490,18 @@ struct ONNXLoopOpLowering : public ConversionPattern {
       ConversionPatternRewriter &rewriter) const {
     auto loc = ONNXLoc<ONNXLoopOp>(op);
     auto loopOp = dyn_cast<ONNXLoopOp>(op);
-    ONNXLoopOpAdaptor loopOpAdaptor(operands, op->getAttrDictionary());
     MultiDialectBuilder<KrnlBuilder, MemRefBuilder, MathBuilder> create(
         rewriter, loc);
 
-    // Construct inputs for While. They should be (0, cond, v_initial)
+    // Create memref object for induction variable and outputs
+    // As a result, there is no memref object returned from LoopOp,
+    // just the value in these "global" memref objects.
+    Value ivMemRef =
+        create.mem.alloc(MemRefType::get({}, rewriter.getI64Type()));
+    Value cond = create.mem.alloc(MemRefType::get({}, rewriter.getI1Type()));
+    ONNXLoopOpAdaptor loopOpAdaptor(operands, op->getAttrDictionary());
+
+    // Construct inputs for WhileOp, which should be (0, cond, v_initial)
     // The initial value for iteration should be zero
     SmallVector<Value, 4> whileInputValues;
     SmallVector<Type, 4> whileInputTypes;
@@ -593,8 +600,8 @@ struct ONNXLoopOpLowering : public ConversionPattern {
 
       // Contain the ThenRegion with KrnlRegionOp
       // The krnl loop inside may use symbol computed in LoopOp body
-      //KrnlRegionOp regionOp = rewriter.create<KrnlRegionOp>(loc);
-      //rewriter.setInsertionPointToStart(&regionOp.bodyRegion().front());
+      KrnlRegionOp regionOp = rewriter.create<KrnlRegionOp>(loc);
+      rewriter.setInsertionPointToStart(&regionOp.bodyRegion().front());
 
       SmallVector<Value, 4> params;
       int argIndex = 0;
@@ -602,9 +609,6 @@ struct ONNXLoopOpLowering : public ConversionPattern {
       // argument passed to the body graph function.
       if (hasM) {
           Value iv = afterBlock->getArgument(0);
-          MemRefBuilder createMemRef(rewriter, loc);
-          Value ivMemRef =
-              createMemRef.alloc(MemRefType::get({}, rewriter.getI64Type()));
           create.krnl.store(iv, ivMemRef);
           params.emplace_back(ivMemRef);
           argIndex++;
@@ -624,7 +628,8 @@ struct ONNXLoopOpLowering : public ConversionPattern {
             mapper.map(regionArg, params[i]);
       }
 
-      Region &containRegion = whileOp.getAfter();
+      //Region &containRegion = whileOp.getAfter();
+      Region &containRegion = regionOp.bodyRegion();
       Block &firstBlock = containRegion.front();
       assert(loopBody.getBlocks().size() == 1 &&
                  "Currently only support loop body with 1 block.");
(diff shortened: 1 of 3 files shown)


src/Conversion/ONNXToKrnl/ControlFlow/Loop.cpp
```diff
@@ -425,6 +434,334 @@ struct ONNXLoopOpLowering : public ConversionPattern {
...
+      Region &containRegion = regionOp.bodyRegion();
+      Block &firstBlock = containRegion.front();
+      assert(loopBody.getBlocks().size() == 1 &&
+             "Currently only support loop body with 1 block.");
+      containRegion.getBlocks().splice(
+          containRegion.getBlocks().end(), loopBody.getBlocks());
+
+      // auto newBlocks = llvm::make_range(std::next(firstBlock.getIterator()),
+      // postInsertBlock->getIterator()); auto newBlocks =
+      // containRegion.getBlocks();
```
tungld, line 647: Should these comments be removed?

src/Conversion/ONNXToKrnl/ControlFlow/Loop.cpp
```diff
@@ -31,7 +31,11 @@ struct ONNXLoopOpLowering : public ConversionPattern {
       ConversionPatternRewriter &rewriter) const final {
     auto loc = ONNXLoc<ONNXLoopOp>(op);
     auto loopOp = dyn_cast<ONNXLoopOp>(op);
-    ONNXLoopOpAdaptor loopOpAdapter(operands, op->getAttrDictionary());
+    ONNXLoopOpAdaptor loopOpAdaptor(operands, op->getAttrDictionary());
+
+    if (needWhile(op)) {
+      return rewriteWithWhile(op, operands, rewriter);
```
tungld, line 37: Would it be better to use SCFWhile instead of While, e.g `needSCFWhile`, `rewriteWithSCFWhile`?

src/Conversion/ONNXToKrnl/ControlFlow/Loop.cpp
```diff
@@ -425,6 +434,334 @@ struct ONNXLoopOpLowering : public ConversionPattern {
...
+
+    // Construct inputs for WhileOp, which should be (0, cond, v_initial)
+    // The initial value for iteration should be zero
+    SmallVector<Value, 4> whileInputValues;
+    SmallVector<Type, 4> whileInputTypes;
+    SmallVector<Location, 4> locs;
+    bool hasM = false;
+    bool hasCond = false;
+    // Value index0 = create.math.constantIndex(0);
+    // Value index1 = create.math.constantIndex(1);
```
tungld, line 511: Should these comments be removed?

src/Conversion/ONNXToKrnl/ControlFlow/Loop.cpp
```diff
@@ -425,6 +434,334 @@ struct ONNXLoopOpLowering : public ConversionPattern {
...
+    Value c0 = create.math.constant(rewriter.getI64Type(), 0);
+    Value c1 = create.math.constant(rewriter.getI64Type(), 1);
+    Value ubV;
+    if (!isFromNone(loopOp.M())) {
+      hasM = true;
+
```
tungld, line 517: Remove the empty line.

src/Conversion/ONNXToKrnl/ControlFlow/Loop.cpp
```diff
@@ -425,6 +434,334 @@ struct ONNXLoopOpLowering : public ConversionPattern {
...
+        combinedV = create.math.andi(mCond, condV);
+      } else if (hasM) {
+        combinedV = mCond;
+      } else {
+        combinedV = condV;
+      }
+
+      // SmallVector<Value, 4>
+      // returnValues(beforeBlock->getArguments().begin()+argIndex,
+      // beforeBlock->getArguments().end());
```
tungld, line 589: The comments look obsolete ...

tungld (approved):
LGTM with some redundant comments to remove.

The new approach of using scf.While looks good I think. Since this patch does not support scan outputs, I am afraid that we will see regression with the dataport model which needs scan outputs. I recommend to double check this patch with the dataport model before merging it.

chentong319:
@tungld Thanks for review. I will fix the issues.  Where can I  get the "dataprot model"?

chentong319:
> @chentong319 I just emailed you the model.

The PR can compile the model.

Issue #1517, opened 2022-06-29
[NNPA] Accuracy issue on NNPA in Conv2D at C != 1, kH = 1, kW=1


imaihal:
See wrong results on NNPA at C(the number of channels)  != 1,  kernel height = 1, and kernel width = 1
No problem on CPU.

<html xmlns:m="http://schemas.microsoft.com/office/2004/12/omml"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=PowerPoint.Slide>
<meta name=Generator content="Microsoft PowerPoint 15">

</head>

<body>

N | C | H | W | kH | kW | Result
-- | -- | -- | -- | -- | -- | --
1 | 1 | 32 | 32 | 1 | 1 | OK
1 | 1 | 32 | 32 | 2 | 1 | OK
1 | 1 | 32 | 32 | 1 | 2 | OK
1 | 1 | 32 | 32 | 2 | 2 | OK
1 | 2 | 32 | 32 | 1 | 1 | NG
1 | 2 | 32 | 32 | 2 | 1 | OK
1 | 2 | 32 | 32 | 1 | 2 | OK
1 | 2 | 32 | 32 | 2 | 2 | OK
1 | 3 | 32 | 32 | 1 | 1 | NG
1 | 3 | 32 | 32 | 2 | 1 | OK
1 | 3 | 32 | 32 | 1 | 2 | OK
1 | 3 | 32 | 32 | 2 | 2 | OK
1 | 3 | 32 | 32 | 3 | 1 | OK
2 | 2 | 32 | 32 | 1 | 1 | NG

</body>

</html>

imaihal:
This issue should be fixed by PR https://github.com/onnx/onnx-mlir/pull/1576

AlexandreEichenberger:
Congrats!

PR #1518, opened 2022-06-29
[NNPA] Run Conv2D on CPU when input parameters are C != 1, kH = 1, kW=1

imaihal:
As written in  issue #1517 , we get wrong results on NNPA at C(the number of channels) != 1,  kernel height = 1, and kernel width = 1.  This PR avoids this issue by running CPU instead of NNPA when using the parameter. 

### DIFF
--- a/src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
+++ b/src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
@@ -850,7 +850,7 @@ bool isSuitableForZDNN<ONNXConvOp>(ONNXConvOp op) {
     return false;
 
   // Workaround for issue #1517 Accuracy issue  at C != 1, kH = 1, kW=1
-  if (inputShapeC > 1 && kernelShapeH == 1 && kernelShapeW == 1)
+  if (inputShapeC != 1 && kernelShapeH == 1 && kernelShapeW == 1)
     return false;
 
   return true;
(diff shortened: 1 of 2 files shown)


imaihal:
@tungld @AlexandreEichenberger Could you review this PR?  I continue investigating the issue #1517, but it may take time.   So, I created this PR as the workaround.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
```diff
@@ -848,6 +849,10 @@ bool isSuitableForZDNN<ONNXConvOp>(ONNXConvOp op) {
   if (!isWOK)
     return false;
 
+  // Workaround for issue #1517 Accuracy issue  at C != 1, kH = 1, kW=1
+  if (inputShapeC > 1 && kernelShapeH == 1 && kernelShapeW == 1)
```
tungld, line 853: You may exclude C == -1 as well for unknown dimensions.

tungld (approved):
LGTM.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
```diff
@@ -848,6 +849,10 @@ bool isSuitableForZDNN<ONNXConvOp>(ONNXConvOp op) {
   if (!isWOK)
     return false;
 
+  // Workaround for issue #1517 Accuracy issue  at C != 1, kH = 1, kW=1
```
AlexandreEichenberger, line 852: Can you just add a bit of comment in the source file, e.g.
```
// Currently disable the generation of Conv2D when parameters are xxx
// because of current issue #1517.  When fixed, please remove lit test yyy.
```

AlexandreEichenberger (approved):

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp

AlexandreEichenberger:
@imaihal thanks for the changes

PR #1524, opened 2022-07-01
[NNPA] Update CMakeLists.txt for backend tests for NNPA

imaihal:
This PR updates CMakeLists.txt for backend tests for NNPA to catch up with update in PR https://github.com/onnx/onnx-mlir/pull/1472

No need to run input verification test on NNPA, but we needed to copy `input_verification_backend.py` to use `test/backend/test.py` for NNPA.


AlexandreEichenberger (approved):
LGTM

PR #1233, opened 2022-03-09
Add parallel region to onnx for loop

YW-Ethan:
Test file: (loop.mlir)
```
func private @test_loop_simple_main_graph(%arg0: tensor<i64>, %arg1: tensor<i1>, %arg2: tensor<1xi64>) -> tensor<1xi64> {
  %0 = "onnx.Loop"(%arg0, %arg1, %arg2) ({
    ^bb0(%body_arg0: tensor<i64>, %body_arg1: tensor<i1>, %body_arg2: tensor<1xi64>):
    %0 = "onnx.Identity"(%body_arg1) : (tensor<i1>) -> tensor<i1>
    %1 = "onnx.Add"(%body_arg2, %body_arg0) : (tensor<1xi64>, tensor<i64>) -> tensor<1xi64>
    onnx.Return %0, %1 : tensor<i1>, tensor<1xi64>
  }) : (tensor<i64>, tensor<i1>, tensor<1xi64>) -> tensor<1xi64>
  return %0 : tensor<1xi64>
}
```
Procedure to lower onnx down to llvm:
  1.   `./onnx-mlir/build/Debug/bin/onnx-mlir-opt --convert-onnx-to-krnl --convert-krnl-to-affine ~/workspace/loop.mlir > ~/workspace/test.mlir`  This will convert `onnx.loop` to `omp.parallel`, and convert onnx to krnl.
  2.  ` ./llvm-project/build/bin/mlir-opt ~/workspace/test.mlir -lower-affine -convert-arith-to-llvm -convert-scf-to-openmp  -convert-scf-to-cf -convert-openmp-to-llvm   -convert-memref-to-llvm  -convert-std-to-llvm > ~/workspace/test1.mlir `
  3. ` ./llvm-project/build/bin/mlir-opt ~/workspace/test1.mlir -async-parallel-for > ~/workspace/test2.mlir`
  4. ` mlir-translate --mlir-to-llvmir ~/workspace/test2.mlir > ~/workspace/test3.mlir`

Then, the `test3.mlir` will have the llvmir with omp_parallel.

Road Block: I'm trying to add the same process into `ConvertKrnlToLLVM.cpp`, but not sure how to add `   mlir::async::populateAsyncParallelForPatterns(
    RewritePatternSet &patterns, bool asyncDispatch, int32_t numWorkerThreads,
    const AsyncMinTaskSizeComputationFunction &computeMinTaskSize)`.

And please let me know if I'm on the right track.

[...]

PR #1522, opened 2022-06-30
prereq change to  llvm build for OMP from PR 1233

caoimhinuibrian:
Signed-off-by: Kevin O'Brien <12157865+caoimhinuibrian@users.noreply.github.com>

PR #1525, opened 2022-07-01
Initial omp support

caoimhinuibrian:
This is most of what was in PR #1233
The actual attempt to generate OpenMP code for the Loop op is not here. Loop op has changed dramatically in the interim, and anyway arbitrary Loops are not necessarily parallel, but #1233 assumed that.

I will add a use of OpenMP in a future PR.


caoimhinuibrian:
This depends on #1522

PR #1526, opened 2022-07-02
Enable OpenMP

gongsu832:
- enable OpenMP on x86 and power
- fix bug not using parentheses for proper handling of || continuation


gongsu832:
Close for now until OpenMP works on zLinux.

PR #1528, opened 2022-07-04
Constant propagation for ONNXCastOp

tungld:
This patch implements constant propagation for ONNXCastOp. 

CastOp is often seen in convolution models to process shape info, so doing constant propagation for ONNXCastOp would be beneficial to such models.
 

src/Transform/ONNX/ConstPropHelper.cpp
```diff
@@ -140,22 +141,18 @@ DenseElementsAttr createDenseElementsAttrFromArray(char *arr, Type outputType) {
...
-    isSplat = true;
-  else if (llvm::all_of(
-               resType.getShape(), [](int64_t dim) { return dim == 1; }))
-    isSplat = true;
-  else
-    isSplat = false;
+  bool detectedSplat;
+  assert(DenseElementsAttr::isValidRawBuffer(
+             resType, ArrayRef<char>(arr, sizeInBytes), detectedSplat) &&
+         "The raw buffer is invalid for provided type");
```
tungld, line 147: This is the better way to get splat correctly as well as verify the raw buffer.

src/Transform/ONNX/ConstPropHelper.cpp
```diff
@@ -140,22 +141,18 @@ DenseElementsAttr createDenseElementsAttrFromArray(char *arr, Type outputType) {
...
   return DenseElementsAttr::getFromRawBuffer(
-      resType, ArrayRef<char>(arr, sizeInBytes), /*isSplat=*/isSplat);
+      resType, ArrayRef<char>(arr, sizeInBytes), detectedSplat);
 }
 
 /// Create a dense ONNXConstantOp from a byte array.
 ONNXConstantOp createDenseONNXConstantOp(PatternRewriter &rewriter,
```
AlexandreEichenberger, line 153: Should we add this to the builder? And directly put all the cases in it (meaning `convertDoubleInt64ToExactType`)?
It feels like there is a lot of functions calling one that does some of the work, and then another one that does a bit more. If feel like a `create.onnx.DenseConstant(typeFrom, typeTo, *array)` would be useful abstraction

AlexandreEichenberger (commented):
Glad you found one more pattern to optimize.

src/Transform/ONNX/ConstPropHelper.cpp
tungld: Good suggestion! I will move some functions to the builder. Thanks!

PR #1529, opened 2022-07-05
fix(docker): pulling docker images scripts should be linux/amd64

aarnphm:
This address running this file on m1 (arm-based machine) to pull correct image
from docker hub.

Signed-off-by: Aaron Pham <29749331+aarnphm@users.noreply.github.com>


aarnphm:
noted. Thanks for the clarification.

PR #1534, opened 2022-07-07
Handle SeqType in LoopOp

chentong319:
In lowering ONNXLoopOp to krnl IR, a storage(memref) is allocated outside of loop for every Value to be passed to and returned from loop body. The loop-carried value in LoopOp may be SeqType, which is a dynamic type (its size changes with sequence ops) and cannot be allocated as tensor types.
 This PR added the support for SeqType in LoopOp. Instead of allocating a SeqType, a holder for SeqType value is allocated. In details, SeqType is implemented with memref<*xmemref<*xT>>. The size of the memref is dynamic for every iteration. The SeqType value is stored in a Value of type memref<1xmemref<*xmemref<*xT>>. In loop body, argument of SeqType is loaded from this value in the beginning, and the new value is stored into this value at the end.
One issue complicated the implementation: In PR #1238 , static seq type is used in accumulating dynamic shaped tensor. In the code, I cannot not simply check type of memref of memref but have to check whether this is intermediate result (not scan output) and use the original LoopOp type(which is still SeqType, not lowered to memref<memref<>> yet.
This PR can compile and run the modified backend test: test_loop13_seq_cpu with bufferization::Deallocate pass turned off. It is known to us that we have to find a better way to integrate with bufferization. The current solution is to lower seq ops after deallocation.
The reasons to modify the test_loop13_seq_cpu model are:
1. Our shape inference for subgraph seems not correct: the block argument simply inherited the input seq type. Perhaps need special check to SeqType
2. Return of function does not supported value with SeqType
3. In general we do not support unranked tensor. But sequence of unranked tensor might be a special case we should handle.

Those fixes will be in future PRs.


src/Conversion/ONNXToKrnl/ControlFlow/Loop.cpp
```diff
@@ -107,9 +127,16 @@ struct ONNXLoopOpLowering : public ConversionPattern {
...
           // Make the call to loop body function.
           SmallVector<Value, 4> params = {ivMemRef, loopOpAdaptor.cond()};
-          for (auto value : llvm::make_range(outputs.begin(),
-                   outputs.begin() + loopOpAdaptor.v_initial().size()))
-            params.emplace_back(value);
+
+          // For SeqType, load the value for the storage
+          for (unsigned long i = 0; i < loopOp.v_initial().size(); i++) {
+            if (loopOp.v_initial()[i].getType().isa<SeqType>()) {
+              auto seqValue = create.krnl.load(outputs[i], zero);
```
tungld, line 134: Please use a concrete type instead of `auto`.

src/Conversion/ONNXToKrnl/ControlFlow/Loop.cpp
```diff
@@ -171,24 +198,20 @@ struct ONNXLoopOpLowering : public ConversionPattern {
...
-                   "Expecting loop body function output to consist of "
-                   "tensors/memrefs.");
-            auto outputTy = output.getType().cast<ShapedType>();
-            bodyOutputs[i] = rewriter
-                                 .create<UnrealizedConversionCastOp>(loc,
-                                     MemRefType::get(outputTy.getShape(),
-                                         outputTy.getElementType()),
-                                     output)
-                                 .getResult(0);
+            auto outputTy = output.getType();
```
tungld, line 201: Please use a concrete type instead of `auto`.

src/Conversion/ONNXToKrnl/ControlFlow/Loop.cpp
```diff
@@ -239,47 +266,57 @@ struct ONNXLoopOpLowering : public ConversionPattern {
...
-        for (auto i = 0;
-             i < firstElement.getType().cast<MemRefType>().getRank(); i++) {
-          dims.emplace_back(
-              firstElement.getType().cast<MemRefType>().getShape()[i]);
-          if (firstElement.getType().cast<MemRefType>().getShape()[i] == -1)
-            allocParams.emplace_back(create.mem.dim(firstElement, i));
+        // need to distinguish seqType in v_final and scan
+        if (i < loopOp.v_final().size()) {
+          // In v_final
+          auto v = create.krnl.load(output, zero);
```
tungld, line 277: Please use a concrete type instead of `auto`.

src/Conversion/ONNXToKrnl/ControlFlow/Loop.cpp
```diff
@@ -239,47 +266,57 @@ struct ONNXLoopOpLowering : public ConversionPattern {
...
+          newOutputs.emplace_back(v);
+        } else {
+          // scan output
+          // need to convert memref<memrefs<xT>> to memref<xT>
+          // TODO: need a IF statement to handle output is empty
+          // we can safely give 0 to the dynamic dim for alloc
+          // Here loop is assumed to be executed at least once.
+          auto firstElement =
+              create.krnl.load(output, create.math.constantIndex(0));
```
tungld, line 286: Please use a concrete type instead of `auto`.

src/Conversion/ONNXToKrnl/ControlFlow/Loop.cpp
```diff
@@ -239,47 +266,57 @@ struct ONNXLoopOpLowering : public ConversionPattern {
...
+               i < firstElement.getType().cast<MemRefType>().getRank(); i++) {
+            dims.emplace_back(
+                firstElement.getType().cast<MemRefType>().getShape()[i]);
+            if (firstElement.getType().cast<MemRefType>().getShape()[i] == -1)
+              allocParams.emplace_back(create.mem.dim(firstElement, i));
+          }
+          ArrayRef<int64_t> shape(dims.data(), dims.size());
+          auto flatType = MemRefType::get(shape,
+              firstElement.getType().cast<MemRefType>().getElementType());
+          auto alloc = create.mem.alignedAlloc(flatType, allocParams);
```
tungld, line 302: Please use a concrete type instead of `auto`.

src/Conversion/ONNXToKrnl/ControlFlow/Loop.cpp
```diff
@@ -239,47 +266,57 @@ struct ONNXLoopOpLowering : public ConversionPattern {
...
+          // Here loop is assumed to be executed at least once.
+          auto firstElement =
+              create.krnl.load(output, create.math.constantIndex(0));
+          SmallVector<mlir::Value, 4> allocParams;
+          SmallVector<int64_t, 4> dims;
+          dims.emplace_back(output.getType().cast<MemRefType>().getShape()[0]);
+          if (output.getType().cast<MemRefType>().getShape()[0] == -1)
+            allocParams.emplace_back(create.mem.dim(output, 0));
+          for (auto i = 0;
+               i < firstElement.getType().cast<MemRefType>().getRank(); i++) {
```
tungld, line 293: FYI, we recently add some utility functions to get shape and element type in [TypeUtilities.hpp](https://github.com/onnx/onnx-mlir/blob/main/src/Support/TypeUtilities.hpp). So you can call `getShape(firstElement.getType())` which will check and cast the type for us.

chentong319: Useful utility, but not fit this situation: construct shape from two shape along with values for allocation.

src/Conversion/ONNXToKrnl/ControlFlow/Loop.cpp
```diff
@@ -52,10 +67,15 @@ struct ONNXLoopOpLowering : public ConversionPattern {
...
-      emitCopy(rewriter, loc, std::get<0>(vInitAndFinal),
-          std::get<1>(vInitAndFinal));
+    for (unsigned long i = 0; i < loopOpAdaptor.v_initial().size(); i++) {
+      auto origInput = loopOp.v_initial()[i];
+      if (origInput.getType().isa<SeqType>()) {
+        Value zero = create.math.constantIndex(0);
+        create.krnl.store(loopOpAdaptor.v_initial()[i], outputs[i], zero);
+      } else {
+        emitCopy(rewriter, loc, loopOpAdaptor.v_initial()[i], outputs[i]);
+      }
```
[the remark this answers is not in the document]
chentong319: I think the seqtype operation will be replaced with some seq operation to preserve the bufferization semantics in the next PR.

tungld (approved):
LGTM. Thanks!

PR #1538, opened 2022-07-14
Constant propagation for ONNXGatherOp

tungld:
This patch implements constant propagation for ONNXGatherOp


AlexandreEichenberger (commented):
I have a generic questions with some of these ops. I understand that we had a constant propagation phase because some ops (like add) uses commutativity and associativity to propagate constant operators together, e.g. `(a+const0)+const1 -> a+(const0+const1)`.

However, it looks like many of the newly added const propagation are more of the form where all inputs are constant. If so, would it make sense to add them using the fold interface? Or maybe because we have some rules like above, all need to be rules so that constant from such rules (such as additions) would otherwise not trigger a fold? 

Its ok as is, just wondering the motivation of not using the fold interface which seems more generic.

AlexandreEichenberger (approved):
LGTM

Issue #1514, opened 2022-06-28
[RFC] Add Legalization from ONNX-MLIR to MHLO


Connor-XY:
# [RFC] Add Legalization from ONNX-MLIR to MHLO
Hello everyone, we are from the AML (Applied Machine Learning) team at Bytedance. Given this RFC where they want to add [TOSA legalization pass](https://github.com/onnx/onnx-mlir/issues/1194), we have observed similar interest in support for MHLO dialect (Meta HLO Dialect). We would also like to add the feature of conversion from ONNX to MHLO. The introduction and the implementation of MHLO dialect can be found [here](https://github.com/tensorflow/mlir-hlo) and the corresponding operation definition is [here](https://www.tensorflow.org/mlir/hlo_ops). 
## Objective
We want to add the legalization pass from ONNX-MLIR to MHLO dialect. We would add this conversion as an independent pass, which won't affect the other parts of the ONNX-MLIR. This modification could provide an extra path where ONNX-MLIR can be further lowered into MHLO dialect, and thus enrich the applications of ONNX-MLIR and make use of the optimization efforts on MHLO dialect.
## Motivation
HLO (High Level Optimizer) IR provides a set of mostly orthogonal operations which can give rather complete functionalities. MHLO supports a HLO-like pipeline using MLIR and provides a uniform interface to compile and execute these optimized HLO programs with dynamic shape support. Both Tensorflow and PyTorch models can be legalized to MHLO dialect. The same conversion can also be applied to ONNX models. ONNX is one of the most important and widely used portable formats in machine learning models, and there doesn't exist a path to MHLO dialect currrently. Therefore, it is non-trivial to have a pass for this legalization. If we have the conversion pass from ONNX to MHLO, we could directly convert ONNX models to the MHLO dialect. What’s more, we can convert other models with ONNX API calls if we can support this legalization.
## User Benefit
This addition would consolidate the development efforts of ONNX-MLIR. It will enhance the ecosystem of ONNX-MLIR by enriching the applications of ONNX models. This extra path from ONNX-MLIR to MHLO can bridge the two dialects. MHLO is a proper IR since it is still close to the frontend and can represent models of TensorFlow and PyTorch. If this conversion can be supported, then we can represent all models of these three ecosystems in a unified representation, so we can focus on the optimization of the MHLO dialect. This also adds to the flexibility of models since we can convert models to MHLO dialect as long as every operation can be lowered.
## Design Proposal
[...]

PR #1539, opened 2022-07-14
[Mhlo] BatchNorm & Concat

chenchongsong:
This PR serves the following purposes:

- supports the lowering of `ONNXBatchNormalizationInferenceModeOp` and `ONNXConcatOp` from ONNX-MLIR to MHLO.
- adds tests for the two operations.
- The corresponding RFC is here: https://github.com/onnx/onnx-mlir/issues/1514


tungld:

src/Conversion/ONNXToMhlo/Tensor/Concat.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+    if (op->getNumResults() < 1) {
+      op->emitError() << "ONNXConcatOp Has No Output\n";
+      return failure();
+    }
+    RankedTensorType resultType =
+        op->getResult(0).getType().dyn_cast_or_null<RankedTensorType>();
+    if (resultType == nullptr) {
+      op->emitError() << "Concat Output Is Not Ranked\n";
+      return failure();
```
tungld, line 42: Not specific to this PR, I am a bit concerned about failures during rewrite patterns that can cause the whole compilation pipeline terminated. This would work if all ONNX ops can be lowered to MHLO. However, if an op does not satisfy the condition here, there is no a fallback path so that the op will be lowered to other dialects such as Krnl for CPU or ZHigh for NNPA.

One way to solve this is to use `DynamicallyLegalOp` in `--convert--onnx-to-mhlo` pass to selectively add operations into the conversion target. We did this for `--convert-onnx-to-zhigh`: https://github.com/onnx/onnx-mlir/blob/main/src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.cpp#L323

FYI @AlexandreEichenberger

chenchongsong:

src/Conversion/ONNXToMhlo/Tensor/Concat.cpp
chenchongsong: Thanks for the comment! Will discuss with co-workers about the fallback path for mhlo lowering. @yaochengji

yaochengji: @tungld  Currently `-convert-onnx-to-hlo` uses `applyPartialConversion` and it doesn't make ONNX Dialect or any ONNX operations illegal. Therefore it is fine if there're some ONNX ops couldn't be converted to MHLO.

src/Conversion/ONNXToMhlo/Tensor/Concat.cpp
```diff
@@ -0,0 +1,61 @@
...
+      ConversionPatternRewriter &rewriter) const final {
+
+    Location loc = op->getLoc();
+    ONNXConcatOpAdaptor operandAdaptor(operands);
+    ONNXConcatOp concatOp = llvm::cast<ONNXConcatOp>(op);
+
+    assert(op->getNumResults() == 1 && "ONNXConcatOp shoule have 1 result");
+    Type resultType = op->getResult(0).getType();
+    if (!onnx_mlir::isRankedShapedType(resultType)) {
+      op->emitError() << "Concat Output Is Not Ranked\n";
```
yaochengji, line 38: I don't think it's an error. Suggest using llvm::dbgs() and DEBUG_TYPE here.

src/Conversion/ONNXToMhlo/Tensor/Concat.cpp
tungld: @yaochengji I see. Thanks for clarifying that!

AlexandreEichenberger:
@chenchongsong merged with the latest before starting to make a run, hope  its ok with you.

tungld (approved):
LGTM. Thanks!

PR #1546, opened 2022-07-18
Upgrade to opset 14

philass:
Hey @chentong319,

Sorry for going a little silent on the opset 14 upgrade. I now have backing from work to take this on. 

I noticed that `GRU` and `LSTM` only add the layout attribute like `RNN`. I saw that you updated `RNN` to 14. So I felt comfortable taking the same approach here with `GRU` and `LSTM`. The tests needed a few modifications to match the new builders.

`BatchNormalization` didn't need any modifications as `onnx-mlir` only supports inference. The changes in opset 14 to the spec are only specific to `training` mode, which wasn't supported in the first place.

Like `RNN` a new attribute is added to `Reshape`. I updated the `inference_backend` with  ` # ==LIM== allowzero not supported.` to reflect that we don't have support for the attribute.

I also added shape inference for the two new ops `Trilu` and `HardSwish`.


caoimhinuibrian:

chentong319:

chentong319:
Thanks for adding the support for the new version of ops. That part looks good to me. However, in order to handle the old models, we need to keep the old version op in IR and add a rewrite rule to convert it to the new version. You change the test case for reshape op for the new Reshape Op. Similarly, we need to handle the models with older op-set. One possible solution is to use converter to convert it  to the current op-set. But the converter may not work from our experience. Solution within onnx-mlir is to keep the op dialect for the old version and rewrite the older version op to the new version. In other words, we are doing the onnx converter work in onnx-mlir, which should be avoided in future. I will add the instructions in the docs/ImportONNXDefs.md on how to do it.

philass:
@chentong319,

previous `Reshape`'s will be handled correctly. 

The `AllowZero` attribute has a default value, that defaults to the old behavior of opset 13. This means that when a `reshape` from the previous opset comes in, it will be handled as it was before. 

If you look at the changed test, its only the output that needed to be changed. The input to the test could have come from opset 13. This is what happened with addition of the `layout` attribute that was added to `RNN` in opset 14. No additional RNN for opset 13 needed to be defined. I think the ONNX committee updated ops with attributes with default values, whose default values mirror old behavior specifically to help with this issue.

If you disagree with my assessment here, I am happy to make the change you suggested.

AlexandreEichenberger:
FYI, docs/SupportedONNXOps-cpu.md is a generated file. Use `make onnx_mlir_supported_ops` to regenerate it.

AlexandreEichenberger:
Thanks @philass for taking this on, @chentong319 for looking into it satisfies our backward compatibility requirements. Much appreciated

AlexandreEichenberger:
Thanks @philass for taking this on, @chentong319 for looking into it satisfies our backward compatibility requirements. Much appreciated

chentong319:
Yes, it's fine when the added attribute is with default value.

chentong319 (approved):
Lgtm!

philass:
@AlexandreEichenberger glad I could contribute

@chentong319 Thanks for the review

Could one of you guys run the CI? I fixed the merge conflict with `SupportOps-cpu.md`. Once it passes I think this should be ready to merge

AlexandreEichenberger:
@philass restarted the CI updated to the latest version.

AlexandreEichenberger:
@philass thanks for the update. Your help is much appreciated. 
Just merged it now in the actual repo.

PR #1547, opened 2022-07-18
fixed issue that did not list the unsupported ops

AlexandreEichenberger:
As we changed the logic on how to list unsupported ops, we were missing most of them. This is now fixed

No change to onnx-mlir, just limited to the generation of the tables of op supported and logic to generate them.


tungld (approved):
LGTM.

PR #1553, opened 2022-07-19
[TOSA] Update type converter and unary ops

BraunPhilipp:
This PR serves the following purposes:
* Update type converter to legalize valid TOSA types
* Restructure `ONNXToTOSA` pass to align with other conversion passes
* Support for `onnx.Floor` and `onnx.Neg`
* LIT tests for unary ops

Signed-off-by: Philipp Braun <philipp.braun@amd.com>


AlexandreEichenberger:

AlexandreEichenberger:

src/Conversion/ONNXToTOSA/Math/Elementwise.cpp
```diff
@@ -0,0 +1,71 @@
...
+    RewritePatternSet &patterns, TypeConverter &typeConverter,
+    MLIRContext *ctx) {
+  patterns.insert<ONNXReluOpLoweringToTOSA>(typeConverter, ctx);
+
+#define INSERT_UNARY_PATTERN(ONNXOp, TOSAOp)                                   \
+  target.addIllegalOp<ONNXOp>();                                               \
+  patterns.insert<ONNXUnaryOpLoweringToTOSA<ONNXOp, TOSAOp>>(                  \
+      typeConverter, ctx);
+  INSERT_UNARY_PATTERN(ONNXNegOp, tosa::NegateOp)
+  INSERT_UNARY_PATTERN(ONNXFloorOp, tosa::FloorOp)
```
sjarus, line 67: A subtle distinction between Floor and Negate is that Floor is defined in the TOSA spec only for fp types. Negate is defined for both integer and fp. 
https://www.mlplatform.org/tosa/tosa_spec.html#_floor
https://www.mlplatform.org/tosa/tosa_spec.html#_negate

This is currently noted as a bug in the TorchToTosa path as well - I remembered to check since I'd implemented that piece myself.

sjarus (changes requested):
Thanks for this starting contribution @BraunPhilipp! 

The TOSA parts look great, except for the suggestion inline.

src/Conversion/ONNXToTOSA/Math/Elementwise.cpp
KFAFSP: We could either not use the macro mechanism and have a dedicated pattern, or provide an explicit template specialization here. In either case, it seems appropriate to use the same `notifyMatchFailure` approach on integer types to floor for now.

AlexandreEichenberger: We use a few macros here and there, generally we try to avoid them and use templated functions or some other C++ mechanisms.

AlexandreEichenberger (commented):
@chentong319 this pr use type conversions, since you just worked with this, you may want to give a look to this PR too... thanks

src/Conversion/ONNXToTOSA/Math/Elementwise.cpp
BraunPhilipp: Thank you for the quick response! By my understanding `onnx.Floor` will never use an integer type in `onnx-mlir` https://github.com/onnx/onnx/blob/main/docs/Operators.md#Floor. Nevertheless, I have updated the type converter. Hope this is slightly clearer.

AlexandreEichenberger:

BraunPhilipp:
@sjarus @AlexandreEichenberger I tried to address the comments above in the last commit. Not quite sure if it is possible to disable droid messages; they seem a bit disruptive. The current commits are again rebased on top of main; hopefully the right procedure.

sjarus:
> Thanks again @sjarus @AlexandreEichenberger for reviewing! I tried to address the comments (`notifyMatchFailure` + removal of macros) in the last commit. Not quite sure if it is possible to disable droid messages; they seem a bit disruptive. The current commits are again rebased on top of main; hopefully the right procedure. Please let me know in case this PR requires other changes.

The new version looks good functionally! But I think you may find in future that when similar fp-only ops are added - CeilOp for example - you'd be duplicating the specialization that would only differ in the src/dst op names.

AlexandreEichenberger:
@BraunPhilipp please familiarize yourself with DCO. If you state here compliance with the policy, I will override the check for this PR, no need to do a new one.

From the bottom of our README:

ONNX-MLIR requires committers to sign their code using the [Developer Certificate of Origin (DCO)](https://developercertificate.org/). Practically, each git commit needs to be signed, see [here](https://github.com/onnx/onnx-mlir/blob/main/docs/Workflow.md#step-7-commit--push) for specific instructions.

AlexandreEichenberger:
PS: we have some network issues that result sometimes in jenkins failures. I will run it again, your PR is likely fine.

src/Conversion/ONNXToTOSA/Math/Elementwise.cpp
```diff
@@ -0,0 +1,88 @@
...
+#include "mlir/IR/TypeUtilities.h"
+#include "src/Conversion/ONNXToTOSA/ONNXToTOSACommon.hpp"
+
+using namespace mlir;
+
+namespace onnx_mlir {
+
+namespace {
+
+template <typename ONNXOpT, typename TOSAOpT>
```
AlexandreEichenberger, line 23: This template works well and I am not asking you to change the code. 

You could still look at the template pattern we use in src/Conversion/Math/Elementwise.cpp where we use this:

```C++
template <>
struct ScalarOp<ONNXDivOp> {
  using FOp = arith::DivFOp;
  using IOp = arith::DivSIOp;
};
```
to reduce a bit on the redundant code that you have here. Let me know if you want to change it to that pattern. If not, I am fine accepting the PR as is with the understanding that if you add many more element wise op you would at least consider this pattern, if it fits your usage pattern.

AlexandreEichenberger (approved):
LGTM, again let me know if you want to change the template pattern or not.

AlexandreEichenberger:
@BraunPhilipp please let me know if you want to further change the code, or if you want the code to land as is.

In all cases, you  need to confirm here that you abided by the DCO requirements. Once you do so, I can override the DCO failure when merging with main.

src/Conversion/ONNXToTOSA/Math/Elementwise.cpp
BraunPhilipp: > @BraunPhilipp please let me know if you want to further change the code, or if you want the code to land as is.
> 
> In all cases, you need to confirm here that you abided by the DCO requirements. Once you do so, I can override the DCO failure when merging with main.

Sorry for the late reply due to Europe's vacation season. Thank you for the template suggestion! I will update it this week and double check the DCO requirements.

AlexandreEichenberger: From Swizerland myself... vacations are important :-)

BraunPhilipp:
@Eichenberg Sorry for the late reply! This turned out to be legally more challenging than expected. 

> I hereby confirm that I abided by the DCO requirements.

For the PR itself, I just rebased on top of main and tweaked the templating. Let me know if the implementation requires further changes. Further contributions should be smoother.

AlexandreEichenberger:
updated branch to latest and started the CIs

PR #1554, opened 2022-07-19
Update the llvm commit to f2b94bd7eaa83d853dc7568fac87b1f8bf4ddec6

sstamenova:
Update the llvm commit to f2b94bd7eaa83d853dc7568fac87b1f8bf4ddec6 and update the mlir-hlo submodule to a version that also uses llvm f2b94bd7eaa83d853dc7568fac87b1f8bf4ddec6

This applies any necessary changes to move the mlir and mlir-hlo commits forward. Most notably:
1) Trivial updates to lit tests to use `func.func` instead of plain `func`
2) Use the default type converter for the onnx dialect
3) Several headers have changed locations
4) Most libraries now have `Dialect` as part of the name
5) Update the accelerator dialects to use the auto-generated class definitions


src/Dialect/Krnl/KrnlOps.cpp
```diff
@@ -443,7 +444,16 @@ ParseResult KrnlIterateOp::parse(OpAsmParser &parser, OperationState &result) {
...
-  if (parser.parseRegion(*region, inductionVarRefs, inductionVarTypes))
+
+  SmallVector<OpAsmParser::Argument> entryArgs;
+  for (auto it : llvm::zip(inductionVarRefs, inductionVarTypes)) {
+    OpAsmParser::Argument arg;
+    arg.ssaName = std::get<0>(it);
+    arg.type = std::get<1>(it);
+    entryArgs.push_back(arg);
+  }
+
```
AlexandreEichenberger, line 455: It all look fairly straightforward except for this sequence. Do you mind just explaining this section? Is it something that used to be done inside the `parseRegion`? tx

AlexandreEichenberger (approved):
LGTM (got to love these name changes :-)) pending successful run.

src/Dialect/Krnl/KrnlOps.cpp
sstamenova: This is in response to this change: https://reviews.llvm.org/D124649. I can see if there's a neater way to make the change, but this was pretty straightforward without impacting the rest of the code

AlexandreEichenberger: Code is ok as long as it works functionally. Thanks for the reference, will be useful in the future if we need to revisit or apply elsewhere.

sstamenova:
@AlexandreEichenberger : I finally got through all the accelerator tests and updates and I now see a failure in mlir-hlo. It looks like they've updated the llvm commit that they use but not the library dependencies which is an issue for the shared libraries build. I looked at the repo and they've update llvm several times since, but have continued to not update the library dependencies to work for shared library builds. Is the goal to be able to build everything with shared library builds enabled? So possibly making updates to mlir-hlo, so that onnx-mlir can pick up newer versions of llvm?

sstamenova:
I did make a PR into their repo since the necessary change is tiny this time: https://github.com/tensorflow/tensorflow/pull/56844

AlexandreEichenberger:
@sstamenova Thanks for pushing fixes to the TF-mhlo. Can you be a bit more specific on what needs to be happening between onnx-mlir and mhlo? Do they need to be on the same llvm-project commit because they share libs?

sstamenova:
> @sstamenova Thanks for pushing fixes to the TF-mhlo. Can you be a bit more specific on what needs to be happening between onnx-mlir and mhlo? Do they need to be on the same llvm-project commit because they share libs?

They both have a dependency on llvm-project, so when onnx-mlir moves to a newer llvm commit, mhlo needs to be updated to a version that uses the same commit because mhlo is built in the context of onnx-mlir and the llvm libraries and headers might have changed, so mhlo would fail to build. This is the original scenario we ran into on our team when we tried to merge with latest onnx-mlir because we were using a newer llvm commit for onnx-mlir internally.

The second piece is that onnx-mlir builds twice in CI - once with static libraries and one with shared libraries enabled. The fix that I submitted to mhlo is to make sure that the mhlo shared libraries build works. Since mhlo is built with the same library settings (shared vs. not) as onnx-mlir, that portion of the CI was failing because mhlo's shared libraries build was broken. I did add a temporary workaround in my PR into onnx-mlir, but we should remove that next time onnx-mlir's mhlo submodule is updated.

Does that make sense?

AlexandreEichenberger:
Discussion on LLVM upgrades is ongoing with issue #1558 .

gongsu832:
@jenkins-droid publish this please

PR #1556, opened 2022-07-20
Krnl cleanup

AlexandreEichenberger:
added load/store with base index and  offset to simplify the reading of the optimized code.

basically, most of our builder's load and store were of the kind
``` C++
create.*.load(memref, indices)
```

I added a version that does this

``` C++
create.*.load(memref, indices, offsets)
```
where the rank of offsets can be smaller than the rank of the memref & indices. When that is the case, the offsets are added to the least significant dimensions of indices.

So if indices are (i, j, k, l) and offsets are (K, L), the results will be (i, j, k+K, l+L).


tungld (approved):
LGTM.

PR #1559, opened 2022-07-21
Documentation update for onnx version

chentong319:
Add some instructions.

### DIFF
--- a/docs/ImportONNXDefs.md
+++ b/docs/ImportONNXDefs.md
@@ -135,17 +135,25 @@ In ONNX dialect, the op for the top version has no version in the op name, while
 
 When a model is imported, the highest version which is not higher than the next available version is used. For the example of ReduceSum, if the opset is 12, ONNXReduceSumV11Op is chosen.  
 
-## Migrating
+## Upgrade to a new version of ONNX
 To migrate a new version ONNX, first the third_part/onnx should be upgraded and your installation
 of ONNX.
 Then you can run gen_onnx_mlir.py with flag `--check_operation_version`. The top version for all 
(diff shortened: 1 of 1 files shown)


AlexandreEichenberger (commented):
I am reading your modified page, and there are a few statements before your changes that probably need to be clarified. I read

```
ONNX-MLIR project started when ONNX was at version 1.7.0 and does not intended to be backward compatible. 
```

What we probably need to say is that we need backward compatibility for every new release but we don't seek compatibility with anything before 1.7.0 (maybe give the opset associated with this too).

```
To move to a newer version, manually update the version dictionary in the script.
```
maybe give the variable name, `version_dict` if I am not wrong.

```
gen_onnx_mlir.py with flag --check_operation_version. The top version for all operation will be outputted as a new version_dict. 
```

so this update the version_dict? You recommend then to copy and paste it in the python script? You seems to say that it still need to be manually updated, as for ops that changes format, one need to add both. So maybe you want to explain a bit more why we need to list this. To detect which op has a new opset? Be explicit if you don't mind.

Just giving my feedback as I am not as knowledgeable about this, so my questions may help others that are not so intimately familiar with this upgrading

chentong319:
@AlexandreEichenberger I modified the document based on your comments.

AlexandreEichenberger (commented):
line 126:  # Operation Version <a ref="version"></a>

I  find the text below the title still a bit misleading. State that we are intending to be compatible with every new version but anything older than 1.7.0. 

Otherwise, your changes are good, thanks for clarifying the text, much appreciated.

PR #1560, opened 2022-07-22
Small cleanup in ONNXOpTransformPass to use llvm tools for file creation

sstamenova:
Rather than implementing our own logic for temporary file creation, leverage the existing llvm tools. This has the nice side effect of putting the files in the temporary directory on all platforms.


AlexandreEichenberger (approved):
@sstamenova that is great. Do you mind just adding a comment in the code that this is the preferred way to generate a temp file using LLVM that is destroyed on exception errors (if I read the doc correctly). That will be helpful in the future when we need that too for other purposes. I suspect we probably need it elsewhere already.

LGTM after comment (if appropriate)

sstamenova:
> @sstamenova that is great. Do you mind just adding a comment in the code that this is the preferred way to generate a temp file using LLVM that is destroyed on exception errors (if I read the doc correctly). That will be helpful in the future when we need that too for other purposes. I suspect we probably need it elsewhere already.
> 
> LGTM after comment (if appropriate)

I added a comment - this is the right way to generate a temporary file in the tmp directory that is (generally) expected to be removed always). There's another method if you want to create a temporary file somewhere other than the tmp directory and that we might want to keep (`createUniqueFile`).

PR #1562, opened 2022-07-23
MacOS protobuf fix

gongsu832:
- install specific version of protobuf (3.16.0) on macos build to prevent brew installing latest version and breaking the build
- update protobuf version on Linux to 3.16.0 so now Linux, MacOS, and Windows all use the same protobuf version


sstamenova:
What do you think about adding a requirements.txt file and relying on pip for protobuf, onnx and the various pytest dependencies? Similar to how torch-mlir (https://github.com/llvm/torch-mlir/blob/main/requirements.txt) and mlir (https://github.com/llvm/llvm-project/blob/main/mlir/python/requirements.txt) handle the various dependencies.

This would make it easy to just invoke `pip install` with the `requirements.txt` file in any environment and get the right packages with the right version because we could pin protobuf, onnx, etc. to versions as necessary. We've been thinking about moving into that direction for our internal flows because managing all the dependencies across platforms results in issues like the one that this PR is fixing and will likely, at least internally, add a requirements.txt to onnx-mlir.

gongsu832:
We didn't bother with requirements.txt because the dependencies are beyond just pip install. For example, for protobuf, we still need the native libprotobuf.so to be built in addition to pip install protobuf. For onnx, we need to install our own version under third_party. And the pytest dependencies (e.g., pytest-xdist for parallel tests) are optional.

So basically, we don't have a lot of mandatory dependencies to beginning with. And for the ones we have, we need to do things beyond just pip install. So the value of requirements.txt is limited for us.

utils/install-protobuf.sh
```diff
@@ -0,0 +1,9 @@
+PROTOBUF_VERSION=3.16.0
+git clone -b v${PROTOBUF_VERSION} --recursive https://github.com/google/protobuf.git
```
sstamenova, line 2: Would you update the windows version of this script to also use the tag instead of the specific commit

sstamenova (approved):

AlexandreEichenberger (dismissed):
I suggest that we are at a minimum precise in our list on the landing README page.
Currently we have
```
Prerequisites

gcc >= 6.4
libprotoc >= 3.11.0
cmake >= 3.15.4
ninja >= 1.10.2
```

which appears outdated. Note the same text is repeated in 
https://github.com/onnx/onnx-mlir/blob/main/docs/Prerequisite.md

@gongsu832 so please keep these updated.

If you have the will, we could do the trick of having one file, and then use the infamous doc-checks to make sure the md and the master file are in sync?

Some of the issues we are having is that there is one list for using the software (which is what is listed here) and a more comprehensive list that is needed for testing (esp with check-onnx-backend plus -jni suffixes).

utils/install-protobuf.sh
gongsu832: Done

gongsu832:
@AlexandreEichenberger I updated README.md and Prerequisite.md. To me, doc-checks seems more trouble than its worth so I'll leave it to someone else.

PR #1564, opened 2022-07-26
Remove test binaries in dev image

gongsu832:
Dev image is getting too big (>20GB), remove test binaries (~6GB) to reduce its size


caoimhinuibrian (approved):
LGTM

PR #1565, opened 2022-07-27
Added assert for further assurance

AlexandreEichenberger:
No vulnerability with usage of vsnprintf, added doc to demonstrate that it is fine plus one assert to be totally sure there is nothing fishy.


gongsu832 (approved):
LGTM

PR #1566, opened 2022-07-27
Enable warnings as errors by default

sstamenova:
This turns on warnings as errors in onnx-mlir by default. I expect there will be (many) failures and this is mostly to observe where there are failures in the online builds.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,7 +7,7 @@ project(onnx-mlir)
 
 option(ONNX_MLIR_BUILD_TESTS "Build ONNX-MLIR test executables. If OFF, just generate build targets." ON)
 option(ONNX_MLIR_CCACHE_BUILD "Set to ON for a ccache enabled build." OFF)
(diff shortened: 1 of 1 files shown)


sstamenova:
@AlexandreEichenberger : If you want to reduce warnings, I recommend turning warning-as-error ON by default. This would do that, but I expect to see a bunch of failures. Perhaps we can all pitch in and fix them for all the platforms and then enable it everywhere.

AlexandreEichenberger:
@sstamenova I don't advocate to turn warnings as error, too much trouble. I just find that if one look at the error logs/logs on own's machine from time to time to avoid a new creeping of lots of warning, that should be sufficient.

AlexandreEichenberger:
@sstamenova Was this an experiment to show current warning? If so, and you don't intent to push for changing warning into errors, please close.

Tx

sstamenova:
@AlexandreEichenberger : I do plan on enabling them, but a number of higher priority tasks popped up and I haven't had time to finish. I'll close this for the moment and create a new one when I have time to complete it.

PR #1567, opened 2022-07-28
Support parallel

caoimhinuibrian:
use scf ParallelOp for outer loops on Conv op, future work to enable parallelism in other operators


src/Conversion/ONNXToKrnl/NN/Conv.cpp
```diff
@@ -30,6 +31,7 @@ struct ONNXConvOpLowering : public ConversionPattern {
       MemRefType &memRefType, Value alloc) const {
     auto loc = convOp.getLoc();
     KrnlBuilder createKrnl(rewriter, loc);
+    SCFBuilder createScf(rewriter, loc);
```
AlexandreEichenberger, line 34: There are quite a few builder. The preferred way is to do a 
```C++
MultiDialectBuilder<KrnlBuilder, MathBuilder, MemRefBuilder> create(rewriter, loc);
```
where you put the types you need. See Dialect/MLIR/builder.hpp or in the code for examples

src/Conversion/ONNXToKrnl/NN/Conv.cpp
```diff
@@ -12,6 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include "src/Compiler/CompilerOptions.hpp"
```
AlexandreEichenberger, line 15: See Conversion/ONNXToKrnl/Math/Mathmul.cpp for example

```C++
  ONNXMatMulOpLowering(
      TypeConverter &typeConverter, MLIRContext *ctx, bool enableTiling)
      : ConversionPattern(
            typeConverter, mlir::ONNXMatMulOp::getOperationName(), 1, ctx),
        enableTiling(enableTiling) {}
  bool enableTiling;
...
```

We decided that we would rather not have the options directly used in the individual operations, but when we create the pass and add each pattern, we would specify there the optimizations desired.

src/Conversion/ONNXToKrnl/NN/Conv.cpp
```diff
@@ -67,9 +69,18 @@ struct ONNXConvOpLowering : public ConversionPattern {
...
-    ValueRange outerLoops = createKrnl.defineLoops(3);
+    IndexExpr iOne = LiteralIndexExpr(1);
+
+    SmallVector<Value, 3> lbsStorage, ubsStorage, stepsStorage;
     SmallVector<IndexExpr, 3> outerLbs = {iZero, iZero, iZero};
     SmallVector<IndexExpr, 3> outerUbs = {N, G, COPerGroup};
+    SmallVector<IndexExpr, 3> outerSteps = {iOne, iOne, iOne};
+    IndexExpr::getValues(outerLbs, lbsStorage);
+    IndexExpr::getValues(outerUbs, ubsStorage);
+    IndexExpr::getValues(outerSteps, stepsStorage);
```
AlexandreEichenberger, line 80: If you want to create directly set of values, and you don't need the indexExpr, you can also use
```C++
SmallVector<Value, 3> lbsStorage = (create.math.constantIndex(0), create.math.constantIndex(0), create.math.constantIndex(0)}
```
or because the 3 vals are identical:
```C++
SmallVector<Value, 3> lbsStorage(3, create.math.constantIndex(0))
```

src/Conversion/ONNXToKrnl/NN/Conv.cpp
```diff
@@ -81,129 +92,142 @@ struct ONNXConvOpLowering : public ConversionPattern {
...
+            SmallVector<IndexExpr, 4> resAccessFunc;
+            resAccessFunc.emplace_back(SymbolIndexExpr(outerIndices[0]));
+            resAccessFunc.emplace_back(coInOutputSpacial);
+            for (Value o : outputSpatialIndices)
+              resAccessFunc.emplace_back(DimIndexExpr(o));
+            createKrnl.storeIE(result, alloc, resAccessFunc);
+          }); // Output spacial loops.
+    };
+
+    if (enableParallel) {
```
AlexandreEichenberger, line 219: smart to create a lambda and reuse it in two ways...

src/Conversion/ONNXToKrnl/NN/Conv.cpp
```diff
@@ -81,129 +92,142 @@ struct ONNXConvOpLowering : public ConversionPattern {
     MemRefType tmpType = MemRefType::get({}, memRefType.getElementType());
     // Single scalar, no need for default alignment.
     Value reductionVal = createMemRef.alloca(tmpType);
+    // createKrnl.iterateIE(outerLoops, outerLoops, outerLbs, outerUbs,
+    //    [&](KrnlBuilder &createKrnl, ValueRange outerIndices) {
+    // createScf.parallelLoop(parLbs, parUbs, steps,
+    //    [&](DialectBuilder &createScf, ValueRange outerIndices) {
+    //      bodyFunction(outerIndices);
+    //    }
+    auto bodyFunction = [&](ValueRange outerIndices) {
```
AlexandreEichenberger, line 101: Did you do any changes here? Or is it just the indentation. If there are changes, will review in details

AlexandreEichenberger (commented):
Did you run the numerical tests

you can do it this way:
```
make TestGemm
export TEST_ARGS="-O3 -parallel"
Debug/bin/TestGemm
```

caoimhinuibrian:
replaced by #1625

src/Conversion/ONNXToKrnl/NN/Conv.cpp
caoimhinuibrian: the indexExpr's are also used

PR #1574, opened 2022-08-02
Vsprint doc v2

AlexandreEichenberger:
realized that for unsigned, I need to use `a >= b` instead of `a - b >= 0`


AlexandreEichenberger:
@gongsu832 the assert I used in the prior version of this was always true. This revised version has a proper assert. Thanks for approving it if you are fine by it.

gongsu832 (approved):
LGTM

PR #1575, opened 2022-08-03
Print type of ONNX operations's input operand at runtime

AlexandreEichenberger:
Activate instrumentation using
```bash
onnx-mlir mnist.onnx --instrument-onnx-signature
```

and run will results in this output
```bash
run-onnx-lib mnist.so 
Load model file mnist.so with entry point run_main_graph
MaxPoolSingleOut:
  0: 1x1x28x28xfloat

Constant:
  0: 2xint64_t

Reshape:
  0: 1x1x14x14xfloat
  1: 2xint64_t

Constant:
  0: 128x196xfloat

Constant:
  0: 128xfloat

Gemm:
  0: 1x196xfloat
  1: 128x196xfloat
  2: 128xfloat

Relu:
  0: 1x128xfloat

Constant:
  0: 10x128xfloat

Constant:
  0: 10xfloat

Gemm:
  0: 1x128xfloat
  1: 10x128xfloat
  2: 10xfloat

Softmax:
  0: 1x10xfloat
```

Implemented by adding a pass that adds for each ONNX op a `ONNXPrintSignatureOp`, which eventually get lowered to a sequence of `krnl.printf` and `krnl.printTensor`.

I modified  `krnl.printTensor` to take an optional format command  at the beginning of its `msg` input string. The following formats are defined:
-   `%s`: detailed signature (including shape, type, offsets),
-   `%t`: compact type (ala MLIR: 32x16xfloat),
-   `%d`: data values.
    
When no formatting is provided, "%s%d" is used (detailed signature and data) by default.


doru1004 (approved):
LGTM

doru1004:
> 

I think this is better!

PR #1576, opened 2022-08-04
Fix condition where reinterpret_cast is used in lowering of transpose op.

imaihal:
Output MemRefType are used to check the conditions where `reinterpret_cast` can be used, but the MemRefType should be that of input. This is fixed by this PR.

When output MemRefType is used, `reinterpret_cast` is used in following example in onnx-to-krnl lowering, but this is not correct. This issue occurs when the order of dimension where the value is not 1 is changed by transpose.
```
%3 = "onnx.Transpose"(%arg1) {perm = [2, 3, 1, 0]} : (tensor<2x3x1x1xf32>) -> tensor<1x1x3x2xf32>
```

### DIFF
--- a/src/Conversion/ONNXToKrnl/Tensor/Transpose.cpp
+++ b/src/Conversion/ONNXToKrnl/Tensor/Transpose.cpp
@@ -34,12 +34,19 @@ struct ONNXTransposeOpLowering : public ConversionPattern {
     Value data = operandAdaptor.data();
     auto permAttr = transposeOp.perm();
 
-    // Convert the output type to MemRefType.
-    Type convertedType = typeConverter->convertType(*op->result_type_begin());
-    assert(convertedType && convertedType.isa<MemRefType>() &&
+    // Convert the input type to MemRefType.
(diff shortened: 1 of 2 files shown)


imaihal:
This PR should fix issue https://github.com/onnx/onnx-mlir/issues/1517

AlexandreEichenberger (approved):
LGTM, does it fix the CONV bug?

imaihal:
@AlexandreEichenberger 

> LGTM, does it fix the CONV bug?

Yes. This fixed the CONV kernel 1x1 issue。

AlexandreEichenberger:
@imaihal  If it did fix the CONV bug, shouldn't you disable the workaround? I would expect that you cannot test the NNPA code without disabling it first. Maybe in a subsequent PR?

imaihal:
@AlexandreEichenberger Yes. I thought I will create another PR for it. I will do it soon. Thanks!

AlexandreEichenberger:
@imaihal so you disabled the workaround to ensure that this indeed fix the Conv issue?

PR #1577, opened 2022-08-04
Add a rule to combine two transposes besides leaky relu into a sigle transpose

negiyas:
Add a rule to combine two transposes besides leaky relu into a simple transpose.


AlexandreEichenberger (changes requested):
@negiyas I understand that two transpose that follow each others can be combined. The optimization that you are proposing here is that if between 2 transpose ops there is an activation function, then one may perform the activation function first, and then the combined transpose.

So why restrict this operation to leaky relu? 

I think it could be done for any of the src/Conversions/ONNXToKrnl/Math/Elementwise.cpp operations listed as `ONNXElementwiseUnaryOpLowering`. You will see, there are quite a few.

src/Dialect/ONNX/Rewrite.td
```diff
@@ -386,6 +386,13 @@ def FuseTransposePattern:  Pat<
   // Transpose with combined pattern.
   (ONNXTransposeOp $v, (CreateCombinedTransposedPattern $p1, $p2))>;
 
+def FuseTransposeAndLeakyReluPattern:  Pat<
+  // Transpose of LeakyRelu on a transpose.
+  (ONNXTransposeOp (ONNXLeakyReluOp (ONNXTransposeOp $v, $p1), $alpha), $p2),
+  // LeakyRelu on Transpose with combined pattern.
+  (ONNXTransposeOp (ONNXLeakyReluOp $v, $alpha, (returnType $v)),
+                   (CreateCombinedTransposedPattern $p1, $p2))>;
+
```
tungld, line 395: You don't need to create `CombinedTransposedPattern` just swap transpose and leakyrelu, the two transposes will be combined automatically since we have a rule for that. For example,

`Transpose -> UnaryOp -> Transpose` becomes `UnaryOp -> Transpose -> Tranpose`.

AlexandreEichenberger: Thought so too... Initially though it did not hurt to do it here also.  One could imagine that if there are other rules with higher priority that would handle the transposes... it might be better indeed not to do it here.

AlexandreEichenberger:
Technically, if we had

`transpose(binary-op(transpose(a, X), transpose(b, X), Y)` 

where type of a and b are identical, then we can also do the same trick

`transpose(transpose(binary-op(a, b), X), Y)`

isn't it @negiyas @tungld ?

if so, this can be handled in a different PR.

tungld:
> Technically, if we had
transpose(binary-op(transpose(a, X), transpose(b, X), Y)
where type of a and b are identical and binary op is an element-wise operation, then we can also do the same trick
transpose(transpose(binary-op(a, b), X), Y)

Yes, it's true.

src/Dialect/ONNX/Rewrite.td
tungld: Yes, a global optimization is preferred to decide whether propagating transpose is beneficial or not.

AlexandreEichenberger:
Can you comment on whether you will generalize to all the element ops with one input, or two? Just so that we know at a high level what you intend to do in this PR. Its ok if you want to postpone some of the extensions to a future PR

negiyas:
Let me close this PR, because this PR has been replaced by PR  https://github.com/onnx/onnx-mlir/pull/1636 .
Thank you very much for your comments and I am sorry for troubling you.

PR #1578, opened 2022-08-04
Allow Multi-threading safe in onnx-mlir

Jiaqing-ASU:
Add one new option env var named 'menvVarName' in CompilerOptions and adjust the inputs of function 'ParseCommandLineOptions' in onnx-mlir.cpp. A python testing example named 'multi-threading-test.py' is provided under /docs/mnist_example/.


AlexandreEichenberger:

AlexandreEichenberger:
You need to do the DCO stuff, see at the end of the README.md in the top dir.

AlexandreEichenberger:
You also need to format the code using clang format.

AlexandreEichenberger:

AlexandreEichenberger:
FYI, users were kind of needing feedback on the ops being processed by the accelerator. This PR provide this info, which might be useful to your effort of providing more info as well.

I could not have the info written to a file (due to some strange LLVM behavior) so at this stage it is only printed on the stderr output.

#1587

src/onnx-mlir.cpp
```diff
@@ -58,10 +58,10 @@ int main(int argc, char *argv[]) {
 
   llvm::cl::SetVersionPrinter(getVersionPrinter);
 
-  // Parse options from argc/argv and default ONNX_MLIR_FLAG env var.
+  // Parse options from argc/argv and env var.
   llvm::cl::ParseCommandLineOptions(argc, argv,
       "ONNX-MLIR modular optimizer driver\n", nullptr,
-      OnnxMlirEnvOptionName.c_str());
+      menvVarName.c_str());
```
AlexandreEichenberger, line 64: I think there is a "chicken and egg" issue here. The parse command line option scan the options, but it may process the env var first (that was my understanding, but I may be wrong). So check to see the following.

```bash
export ONNX_MLIR_FLAGS="-O3"
export MY_ONNX_MLIR_FLAGS="-O0"

onnx-mlir -menvVarName= MY_ONNX_MLIR_FLAGS bibi.onnx
```

and maybe add a print of the option level just after parsing. Check that that works, namely that you get -O0.

If you get the -O3, no problem, you can always manually scan the args to search for the specific option before this call.

src/Compiler/CompilerOptions.cpp
```diff
@@ -76,6 +76,11 @@ llvm::cl::opt<std::string> shapeInformation("shapeInformation",
         "unknown dimensions)"),
     llvm::cl::value_desc("value"), llvm::cl::cat(OnnxMlirOptions));
 
+llvm::cl::opt<std::string> menvVarName("menvVarName",
```
AlexandreEichenberger, line 79: the `-m` was principally use for "machine", namely `-march` `-mtripple`, `-mcpu`...

Since the original env var is named ONNX_MLIR_FLAGS, maybe `-custom-env-flags`?

AlexandreEichenberger (commented):
I believe that to be thread safe, you have to create a custom process for the compilation.

If you wonder how to do it, I can create a quick PR to do it. Let me know

AlexandreEichenberger:
Why closed? to generate a new one with the DCO?

Jiaqing-ASU:
> FYI, users were kind of needing feedback on the ops being processed by the accelerator. This PR provide this info, which might be useful to your effort of providing more info as well.
> 
> I could not have the info written to a file (due to some strange LLVM behavior) so at this stage it is only printed on the stderr output.
> 
> #1587

Thank you so much for the information. I closed this PR and opened a new one. I think it seems easier for me to use a new branch and then follow clang-format using 'clang-format -i' and add DCO through 'git commit -s -m'. This is my first time to participate in such an open source project. Thank you very much for your valuable time and guidance.

src/Compiler/CompilerOptions.cpp
Jiaqing-ASU: I'm sorry I didn't notice that just now. The new PR is #1588

src/onnx-mlir.cpp
Jiaqing-ASU: If I understand it correctly, the problem you raised is that the 'menvVarName' variable will be given the default value before argv is fully processed. I have tested this and it works well. However, my guess is that there are not many args provided in the current tests. If the argv is very complicated, I wonder if the problem you mentioned will happen? I don't know. So I think manually scan the args to search for the specific option would make this safe. I will fix this then. For the custom process for the compilation, could you please create a quick PR?

AlexandreEichenberger:
NP, a pleasure to help motivated students

PR #1587, opened 2022-08-05
Report stats of ops

AlexandreEichenberger:
Report statistics on the frequency of ONNX ops in the model. This pass is done after the handling of accelerator, so NNPA ops are also listed. Report can be as text (TXT) or as JSON (JSON).

Enabled by option `-onnx-op-stats=TXT` or `-onnx-op-stats=JSON`

For example:
```bash
> onnx-mlir -O3 -maccel=NNPA -onnx-op-stats=TXT bertsquad-12c.mlir 
Operations encountered:
-----------------------
   func.func            , 1
   func.return          , 1
   onnx.Add             , 50
   onnx.Cast            , 1
   onnx.Constant        , 219
   onnx.ConstantOfShape , 1
   onnx.Gather          , 1
   onnx.MatMul          , 24
   onnx.Mul             , 136
   onnx.NoValue         , 1
   onnx.OneHot          , 1
   onnx.Pow             , 12
   onnx.Reciprocal      , 25
   onnx.ReduceMean      , 50
   onnx.Reshape         , 68
   onnx.Softmax         , 12
   onnx.SplitV11        , 1
   onnx.Sqrt            , 25
   onnx.SqueezeV11      , 2
   onnx.Sub             , 62
   onnx.Transpose       , 49
  zhigh.Add             , 62
  zhigh.MatMul          , 74
  zhigh.Mul             , 62
  zhigh.Stick           , 336
  zhigh.Tanh            , 12
  zhigh.Unstick         , 162
```


AlexandreEichenberger:
@tungld I think many of the stick may come from `constant -> stick -> z-op`. They seem to be optimized later. Right now this call is just at the beginning of `addONNXToKrnlPasses`. So the number of stick maybe misleading compared to the actual output. I used the normal pass from LLVM. Maybe we would need to make a custom pass that takes out such patterns in the future, if clients have issues with this inflated number.

tungld:
> @tungld I think many of the stick may come from constant -> stick -> z-op. They seem to be optimized later. Right now this call is just at the beginning of addONNXToKrnlPasses

You are right, `addONNXToKrnlPasses` covers onnx-to-zhigh only and `constant->stick->z-op` optimization is done later since it is specific to NNPA.

tungld (approved):
LGTM

PR #1589, opened 2022-08-08
added OptType and optional ops

sorenlassen:
implemented shape inference for the ops

added lit tests

replaced isOptional logic from gen_onnx_mlir.py

note that this PR doesn't lower OptType (type conversion) and lower optional ops


AlexandreEichenberger:
> TODO: lower OptType (type conversion) and lower optional ops

please use the "draft mode" so that we know when we need to start reviewing. You can always put it on draft and explicitly ask us to review it if you want early feedback.

Tx

sorenlassen:
> > TODO: lower OptType (type conversion) and lower optional ops
> 
> please use the "draft mode" so that we know when we need to start reviewing. You can always put it on draft and explicitly ask us to review it if you want early feedback.

sorry for the confusion but I was hoping we could land this PR as is and save lowering for another round

I reworded "TODO" to "note that this PR doesn't lower" :)

BTW, it's not immediately clear to me how to do the type conversion for OptType to lower it to krnl, I couldn't see anything obvious in KrnlTypeConverter to pattern match against - any suggestions are welcome

AlexandreEichenberger:
go it, thanks for the clarification. On our side, @chentong319 is the one that introduces the types, so I would recommend to exchange info with him. Tong is away for another week or so, he should be able to help once he is back.

utils/gen_onnx_mlir.py
```diff
@@ -485,7 +491,7 @@ def onnx_attr_type_to_mlir_attr_type(t):
         mlir_attr_type = 'StrAttr'
     elif onnx_attr_type == "strings":
         mlir_attr_type = 'StrArrayAttr'
-    elif onnx_attr_type == 'type':
+    elif onnx_attr_type in {'type', 'type_proto'}:
```
AlexandreEichenberger, line 494: Can you just comment on what this new `type_proto` is for us not familiar with the onnx types, tx

AlexandreEichenberger (approved):
LGTM, just had to minor requests, would be great if you could answer them before dropping the code. Thanks for adding the Optional

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -4417,6 +4417,53 @@ LogicalResult ONNXOneHotOp::inferShapes(
...
+}
+
+LogicalResult ONNXOptionalOp::inferShapes(
+    std::function<void(mlir::Region &)> doShapeInference) {
+  Type ty;
+  if (auto typeAttr = type()) {
+    ty = typeAttr.value();
+  } else {
+    ty = input().getType();
+    assert(!ty.isa<NoneType>());
```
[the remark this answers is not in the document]
sorenlassen: done:
```
    // checked in verify()
    assert(!ty.isa<NoneType>() && "type attribute or input value needed");
```

utils/gen_onnx_mlir.py
sorenlassen: added the comment:
```
        # 'type' is the attribute type used in special_attr_types,
        # 'type_proto' is Optional op's type attribute's type
```

PR #1590, opened 2022-08-08
fixes for SeqType and sequence ops

sorenlassen:
As I implemented OptType in PR #1589 I noticed some issues with SeqType and sequence ops that are solved here:
1. SeqType required its element type to be a ShapedType, which conflicts with its use with a map element type in ZipMap
2. the SequenceConstruct op decomposition didn't support most data types
3. we didn't infer shapes for the SequenceConstruct op

Note that Conversion/KrnlToLLVM/KrnlToLLVMHelper now depends on Dialect/ONNX/ONNXOpsHelper for mlirTypeToOnnxType. Please let me know if there's a better way to reuse this logic.

### DIFF
--- a/src/Conversion/KrnlToLLVM/KrnlToLLVMHelper.cpp
+++ b/src/Conversion/KrnlToLLVM/KrnlToLLVMHelper.cpp
@@ -16,6 +16,7 @@
 #include "mlir/Target/LLVMIR/TypeToLLVM.h"
 #include "src/Dialect/Krnl/KrnlOps.hpp"
 #include "src/Dialect/Mlir/DialectBuilder.hpp"
+#include "src/Dialect/ONNX/ONNXOpsHelper.hpp"
 #include "llvm/ADT/TypeSwitch.h"
 
 using namespace mlir;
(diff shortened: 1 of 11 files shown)


sorenlassen:
> @chentong319 is the person on our end that did a lot of the sequences. I really would like him to review it, he is currently away for another 2 weeks, would that be ok with you?

that's ok, I'm happy to wait

src/Conversion/KrnlToLLVM/KrnlToLLVMHelper.cpp
```diff
@@ -115,59 +116,8 @@ int64_t getRankFromMemRefType(LLVM::LLVMStructType memRefTy) {
...
-      .Case<LLVM::LLVMStructType>(
-          [&](LLVM::LLVMStructType) { onnxType = onnx::TensorProto::STRING; });
-
-  if (onnxType == onnx::TensorProto::UNDEFINED) {
-    elemType.dump();
-    llvm_unreachable("MLIR type cannot be converted to ONNX type");
-  }
-
-  return onnxType;
+int64_t mlirTypeToOnnxType(Type elemType) {
```
chentong319, line 119: Can the krnl::mlirTypeToOnnxType be replaced with onnx_mlir::mlirTypeToOnnxType?

src/Dialect/ONNX/ONNXOpsHelper.cpp
```diff
@@ -357,11 +358,13 @@ bool AreTheSameConstantOpDenseAttr(
...
 bool hasShapeAndRank(Value val) {
   Type valType = val.getType();
-  if (!valType.isa<SeqType>())
-    return valType.isa<ShapedType>() && valType.cast<ShapedType>().hasRank();
-
-  // Check the element type for a Sequence.
-  return valType.cast<SeqType>().getElementType().hasRank();
+  ShapedType shapedType;
+  if (auto seqType = valType.dyn_cast<SeqType>()) {
+    shapedType = seqType.getElementType().dyn_cast<ShapedType>();
```
chentong319, line 363: How about Seq(Map(T1, T2))? It's unclear to me what's the usage of the function hasShapeAndRank.

src/Dialect/ONNX/ONNXOpsHelper.cpp
```diff
@@ -622,4 +625,60 @@ mlir::Type convertONNXTypeToMLIRType(
   llvm_unreachable("Unsupported data type encountered.");
 }
 
+// Convert an MLIR type to the correspoding ONNX type.
+int64_t mlirTypeToOnnxType(Type elemType) {
+  onnx::TensorProto::DataType onnxType = onnx::TensorProto::UNDEFINED;
```
chentong319, line 630: Reason to return int64_t instead of enum type onnx::TensorProto::DataType?

src/Conversion/KrnlToLLVM/KrnlToLLVMHelper.cpp
sorenlassen: I got the impression that the src/Dialect/Krnl source code tries to avoid dependencies on ONNX and therefore I put this dependency on ONNX inside the implementation file KrnlToLLVMHelper.cpp, hiding it from the use in KrnlEntryPoint.cpp which just includes the header file KrnlToLLVMHelper.hpp

so this PR creates a half-hidden dependency of Krnl on ONNX which is not ideal, I'm open to suggestions to do it differently

FWIW, AFAICT the RuntimeAPI SET/GET_DATA_TYPE calls in `krnl::fillOMTensorWithMemRef()` and `KrnlEntryPointOpLowering::emitVerificationCodeForInputTensors()` just want any one-to-one mapping of MLIR data types to i64 and doesn't care that the i64 numbers match the onnx::TensorProto enum

src/Dialect/ONNX/ONNXOpsHelper.cpp
sorenlassen: the callers just need an int64_t number so I thought it was most efficient to do it like this and avoid that callers need to include onnx/onnx_pb.h

src/Dialect/ONNX/ONNXOpsHelper.cpp
sorenlassen: if valType is Seq(Map(T1, T2)) then seqType.getElementType() is Map(T1, T2) and dyn_cast<ShapedType>() returns nullptr and hasShapeAndRank() returns false, which I think follows the spirit of the previous implementation and use of hasShapeAndRank

yeah, hasShapeAndRank is a bit strange and the way it's used in many verify() and inferShapes() implementations in ONNXOps.cpp could probably be done more cleanly using functions from src/Support/TypeUtilities

src/Dialect/ONNX/ONNXOpsHelper.cpp
chentong319: Yes, it is good to avoid the include from onnx package. If the enum type is not explicit used, int64_t is fine.

chentong319 (approved):
LGTM!

sorenlassen:
@chentong319 thank you for the review. Please merge it for me.

chentong319:
@sorenlassen Please resolve the conflict. I will try to merge this PR first once it passes the test.

sorenlassen:
> @sorenlassen Please resolve the conflict. I will try to merge this PR first once it passes the test.

@chentong319: done

PR #1593, opened 2022-08-09
Explanation of the difference between the two docker images

AlexandreEichenberger:
Doc change only, no functional change. Describe that to  run .so, user need to fetch the dev version.


gongsu832 (approved):
LGTM

PR #1594, opened 2022-08-09
implemented shape inference for SplitToSequence 

sorenlassen:
also added more checks to ONNXSplitToSequenceOp::verify()

added lit tests for ONNXSplitToSequenceOp::verify() 

added SplitToSequence shape inference lit tests

### DIFF
--- a/src/Dialect/ONNX/ONNXOps.cpp
+++ b/src/Dialect/ONNX/ONNXOps.cpp
@@ -3002,25 +3002,25 @@ LogicalResult ONNXSplitToSequenceOp::verify() {
   auto splitShape = splitType.getShape();
   int64_t splitRank = splitShape.size();
   if (splitRank > 1)
-    return emitError() << "split has rank " << splitRank << " > 1";
+    return emitOpError() << ": split has rank " << splitRank << " > 1";
   if (DenseElementsAttr entries =
           getDenseElementAttributeFromONNXValue(splitValue)) {
     if (splitRank == 0) {
       int64_t scalar = getScalarValue<int64_t>(entries, splitType);
       if (scalar <= 0)
-        return emitError() << "split scalar " << scalar << " <= 0";
+        return emitOpError() << ": split scalar " << scalar << " <= 0";
     } else {
       int64_t sum = 0;
       for (IntegerAttr entry : entries.getValues<IntegerAttr>()) {
         int64_t i = entry.getInt();
         if (i < 0)
-          return emitError() << "split tensor has entry " << i << " < 0";
+          return emitOpError() << ": split tensor has entry " << i << " < 0";
         sum += i;
       }
       int64_t dimSize = inputShape[axisIndex];
       if (dimSize != -1 && dimSize != sum)
-        return emitError() << "split tensor entries sum to " << sum
-                           << " != axis dimension size " << dimSize;
+        return emitOpError() << ": split tensor entries sum to " << sum
+                             << " != axis dimension size " << dimSize;
     }
   }
 
--- a/src/Dialect/ONNX/ONNXOps.cpp
+++ b/src/Dialect/ONNX/ONNXOps.cpp
@@ -2974,8 +2974,8 @@ LogicalResult ONNXSplitToSequenceOp::verify() {
   if (!hasShapeAndRank(inputValue))
     return success(); // Won't be able to do any checking at this stage.
 
-  ShapedType inputType = inputValue.getType().cast<ShapedType>();
-  auto inputShape = inputType.getShape();
+  auto inputType = inputValue.getType().cast<ShapedType>();
+  ArrayRef<int64_t> inputShape = inputType.getShape();
   int64_t inputRank = inputShape.size();
 
   int64_t axisIndex = axis();
@@ -2998,20 +2998,20 @@ LogicalResult ONNXSplitToSequenceOp::verify() {
           onnx_mlir::Diagnostic::Range<int64_t>(0, 1));
     return success();
   }
-  ShapedType splitType = splitValue.getType().cast<ShapedType>();
-  auto splitShape = splitType.getShape();
+  auto splitType = splitValue.getType().cast<ShapedType>();
+  ArrayRef<int64_t> splitShape = splitType.getShape();
   int64_t splitRank = splitShape.size();
   if (splitRank > 1)
     return emitOpError() << ": split has rank " << splitRank << " > 1";
   if (DenseElementsAttr entries =
           getDenseElementAttributeFromONNXValue(splitValue)) {
     if (splitRank == 0) {
-      int64_t scalar = getScalarValue<int64_t>(entries, splitType);
+      auto scalar = getScalarValue<int64_t>(entries, splitType);
       if (scalar <= 0)
         return emitOpError() << ": split scalar " << scalar << " <= 0";
     } else {
       int64_t sum = 0;
-      for (IntegerAttr entry : entries.getValues<IntegerAttr>()) {
+      for (auto entry : entries.getValues<IntegerAttr>()) {
         int64_t i = entry.getInt();
         if (i < 0)
           return emitOpError() << ": split tensor has entry " << i << " < 0";
@@ -3035,8 +3035,8 @@ LogicalResult ONNXSplitToSequenceOp::inferShapes(
 
   // NOTE: all the asserts below are conditions checked in verify()
 
-  ShapedType inputType = inputValue.getType().cast<ShapedType>();
-  auto shape = inputType.getShape();
+  auto inputType = inputValue.getType().cast<ShapedType>();
+  ArrayRef<int64_t> shape = inputType.getShape();
   int64_t rank = shape.size();
   int64_t axisIndex = axis();
   assert((-rank <= axisIndex && axisIndex < rank) && "axis out of range");
@@ -3067,14 +3067,14 @@ LogicalResult ONNXSplitToSequenceOp::inferShapes(
       dims.erase(dims.begin() + axisIndex);
     }
   } else {
-    ShapedType splitType = splitValue.getType().cast<ShapedType>();
-    auto splitShape = splitType.getShape();
+    auto splitType = splitValue.getType().cast<ShapedType>();
+    ArrayRef<int64_t> splitShape = splitType.getShape();
     int64_t splitRank = splitShape.size();
     assert(splitRank <= 1 && "invalid split tensor rank");
     if (DenseElementsAttr entries =
             getDenseElementAttributeFromONNXValue(splitValue)) {
       if (splitRank == 0) {
-        int64_t scalar = getScalarValue<int64_t>(entries, splitType);
+        auto scalar = getScalarValue<int64_t>(entries, splitType);
         assert(scalar > 0 && "invalid split scalar");
         if (dimSize != -1) {
           length = dimSize / scalar;
--- a/src/Dialect/ONNX/ONNXOps.cpp
+++ b/src/Dialect/ONNX/ONNXOps.cpp
@@ -2970,27 +2970,138 @@ LogicalResult ONNXSplitV11Op::inferShapes(
 //===----------------------------------------------------------------------===//
 
 LogicalResult ONNXSplitToSequenceOp::verify() {
-  ONNXSplitToSequenceOpAdaptor operandAdaptor(*this);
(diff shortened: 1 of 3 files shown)


src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2924,27 +2924,138 @@ LogicalResult ONNXSplitV11Op::inferShapes(
 //===----------------------------------------------------------------------===//
 
 LogicalResult ONNXSplitToSequenceOp::verify() {
-  ONNXSplitToSequenceOpAdaptor operandAdaptor(*this);
```
sorenlassen, line 2927: I'm curious about the use of operandAdaptor across the code base: it seems like it's used as an unnecessary detour to access inputs and attributes in places like shape inference where we have direct access to them through the op itself

my interpretation of the MLIR docs
https://mlir.llvm.org/docs/OpDefinitions/#operand-adaptors
and the tutorial example:
https://mlir.llvm.org/docs/Tutorials/Toy/Ch-5/#conversion-patterns
is that operandAdaptor is meant for cases where only the inputs and attributes are available, but not the op, and we want to access inputs and attributes with named methods in another way than through an op

but I'm happy to reintroduce operandAdaptor here if that's what everybody else prefers

sorenlassen: I see how in ConversionPatterns, like in ONNXMatMulOpLowering as an example, operandAdaptor allows you to access the lowered types, but AFAICT this doesn't apply to the shape checking and inference logic in the verify() and most inferShapes() methods in ONNXOps.cpp

I can see why the shape inference logic in the ShapeHelper classes takes operandAdaptor input to the computeShape method, because this is reused (through IndexExpr) across inferShapes() in ONNXOps.cpp and in the matchAndRewrite method in the corresponding ConversionPattern (e.g. ONNXMatMulOpShapeHelper is called from both ONNXMatMulOp::inferShapes and ONNXMatMulOpLowering::matchAndRewrite)

in summary, I believe that operandAdaptor is an unnecessary detour in the case of inferShapes() logic that doesn't use IndexExpr to become reusable for conversion logic, namely in all the inferShapes() methods that are implemented fully in ONNXOps.cpp without calling any ShapeHelpers

AlexandreEichenberger: That make sense. I would wait for @chentong319 to chime in before taking a decision, but your conclusions feel right to me. Maybe @tungld also has an opinion.

tungld: @sorenlassen That makes sense to me too.

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2924,27 +2924,138 @@ LogicalResult ONNXSplitV11Op::inferShapes(
...
-  if (!hasShapeAndRank(input))
+  Value inputValue = input();
+  if (!hasShapeAndRank(inputValue))
     return success(); // Won't be able to do any checking at this stage.
 
-  auto inputType = input.getType().cast<ShapedType>();
-  int64_t inputRank = inputType.getShape().size();
-  int64_t axisIndex = axis();
+  ShapedType inputType = inputValue.getType().cast<ShapedType>();
+  auto inputShape = inputType.getShape();
```
tungld, line 2932: According to [LLVM coding standards](https://llvm.org/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable), we can use `auto` with `cast<type>()` and should avoid `auto` in `auto inputShape = inputType.getShape();` I think `getShape` returns `ArrayRef<int64_t>`

tungld (commented):
I have two comments:
- Why don't we use IndexExpr for shape inference as for other ops?
- In case of unbalanced split, could you still represent the output type with concrete values, say, `onnx.Seq<tensor<?x?x3xf32>`, or need to use `onnx.Seq<tensor<*xf32>`?

Anyway, would like to hear comments from @chentong319, he's our expert in Sequence type.

src/Dialect/ONNX/ONNXOps.cpp
sorenlassen: done

thanks for the explanation

sorenlassen:
> Why don't we use IndexExpr for shape inference as for other ops?

IndexExpr is more complicated so seems unnecessary as long as the logic is only used in inferShapes - I'm happy to move to IndexExpr and move the logic to a ONNXOpShapeHelper subclass if anybody wants to reuse the logic for lowering (or some other purpose), just let me know

sorenlassen:
> In case of unbalanced split, could you still represent the output type with concrete values, say, `onnx.Seq<tensor<?x?x3xf32>`, or need to use `onnx.Seq<tensor<*xf32>`?

I think that's what I tried to do with this implementation - can you please spell out the SplitToSequence inputs (and input shapes) and attributes for the example you have in mind so I can make sure I understand?

tungld:
> I'm happy to move to IndexExpr and move the logic to a ONNXOpShapeHelper subclass if anybody wants to reuse the logic for lowering (or some other purpose), just let me know

Yes, I meant it could be reused for lowering to allocate the output buffer if applicable.

> can you please spell out the SplitToSequence inputs (and input shapes) and attributes for the example you have in mind so I can make sure I understand

This is an example I have in mind where the last chunk would contain only one element along axis. What is the expected output type?

```mlir
%cst = "onnx.Constant"(){value = dense<3> : tensor<i64>} : () -> tensor<i64>
%0 = "onnx.SplitToSequence"(%arg0, %cst) {axis = 0: si64} : (tensor<10xf32>, tensor<i64>) -> !onnx.Seq<tensor<*xf32>>
```

sorenlassen:
> This is an example I have in mind where the last chunk would contain only one element along axis. What is the expected output type?
> 
> ```mlir
> %cst = "onnx.Constant"(){value = dense<3> : tensor<i64>} : () -> tensor<i64>
> %0 = "onnx.SplitToSequence"(%arg0, %cst) {axis = 0: si64} : (tensor<10xf32>, tensor<i64>) -> !onnx.Seq<tensor<*xf32>>
> ```

The output type is !onnx.Seq<tensor<?xf32>>

See test_splittosequence_6 in onnx_shape_inference.mlir:
```
func.func @test_splittosequence_6(%arg0: tensor<4x?x3xf32>) -> !onnx.Seq<tensor<*xf32>> {
  %cst = "onnx.Constant"(){value = dense<3> : tensor<i64>} : () -> tensor<i64>
  %0 = "onnx.SplitToSequence"(%arg0, %cst) : (tensor<4x?x3xf32>, tensor<i64>) -> !onnx.Seq<tensor<*xf32>>
  return %0 : !onnx.Seq<tensor<*xf32>>
// CHECK-LABEL:  func @test_splittosequence_6
// CHECK-SAME:   ([[PARAM_0_:%.+]]: tensor<4x?x3xf32>) -> !onnx.Seq<tensor<?x?x3xf32>> {
// CHECK:           [[VAR_0_:%.+]] = "onnx.Constant"() {value = dense<3> : tensor<i64>} : () -> tensor<i64>
// CHECK:           [[VAR_1_:%.+]] = "onnx.SplitToSequence"([[PARAM_0_]], [[VAR_0_]]) : (tensor<4x?x3xf32>, tensor<i64>) -> !onnx.Seq<tensor<?x?x3xf32>>
// CHECK:           return [[VAR_1_]] : !onnx.Seq<tensor<?x?x3xf32>>
}
```

tungld:
> The output type is !onnx.Seq<tensor<?xf32>>
> See test_splittosequence_6 in onnx_shape_inference.mlir:

OK. I got it. Thanks!

AlexandreEichenberger:
Yes, we used to have the same code written twice, once to identify the constant shapes during shape inference, and once to generate the actual shape computations when the shapes where dynamic. It became very tedious. IndexExpressions used in ShapeInference let you write the code once, and depending on how the ShapeInference helpers are built, it will either default to >0 or -1 if dyn, or generate the code to "instantiate" the computations for the dyn dimensions.

sorenlassen:
> IndexExpressions used in ShapeInference let you write the code once

right, I see the benefits of this approach when it's reused for lowering

I took a hard look at rewriting the shape inference logic with IndexExpr but it proved to be hard because of the many conditionals

given the complexity, I propose that we land the code as is and wait to port it to IndexExpr till it's needed for lowering, which might be quite some time into the future because lowering of sequence types is quite restricted (currently requires statically known sequence length) and is unlikely to work for anything that doesn't have a static shape

the verify() and inferShapes() implementation in this PR seems worthwhile even if it doesn't directly support lowering of dynamic shapes

AlexandreEichenberger:
> I took a hard look at rewriting the shape inference logic with IndexExpr but it proved to be hard because of the many conditionals

@chentong319 as implementor of the many sequence op, do you mind chiming in?

sorenlassen:
@chentong319 what do you think?

src/Dialect/ONNX/ONNXOps.cpp
chentong319: @sorenlassen @AlexandreEichenberger @tungld  I agree that there is no need to use Adaptor because shape inference does not use the pattern rewriting framework. We may clean up the code when we have time.

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2924,27 +2924,140 @@ LogicalResult ONNXSplitV11Op::inferShapes(
...
   // axis attribute must be in the range [-r,r-1], where r = rank(input).
   if (axisIndex < -inputRank || axisIndex >= inputRank)
     return onnx_mlir::Diagnostic::emitAttributeOutOfRangeError(
         *this->getOperation(), "axis", axisIndex,
         onnx_mlir::Diagnostic::Range<int64_t>(-inputRank, inputRank - 1));
+  if (axisIndex < 0)
+    axisIndex += inputRank;
+
+  Value splitValue = split();
+  if (splitValue.getType().isa<NoneType>()) {
```
chentong319, line 2945: Use utility function isFromNone(Value) to check whether the input split is provided or not. The utility can cover different ways of representing None in onnx-mlir.

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2924,27 +2924,140 @@ LogicalResult ONNXSplitV11Op::inferShapes(
...
+
+  // start with length unknown and dims == shape with unknown dimension size
+  // for axis (-1 is ShapedType::kDynamicSize), and edit it as needed below
+  int64_t length = -1;
+  SmallVector<int64_t, 4> dims(shape.begin(), shape.end());
+  dims[axisIndex] = -1;
+
+  Value splitValue = split();
+  if (splitValue.getType().isa<NoneType>()) {
+    // keepdims attribute is ignored unless split input is specified
```
chentong319, line 3009: Document says: If input 'split' is specified, this attribute is ignored.
It's easier for me to understand if the comment is: Since split is not specified, check the keepdims attribute

sorenlassen: done

src/Dialect/ONNX/ONNXOps.cpp
sorenlassen: done

sorenlassen: as an aside, I wonder if it would work to simplify the implementation of isFromNone() to just check if the type is none

in principle, any value with NoneType should mean that the value will be None at runtime

I wonder if the current implementation of isFromNone() wants to deal with a situation where an input has NoneType as a placeholder until shape inference fills in a proper type?

chentong319 (approved):
LGTM!

chentong319:
Additional comments:
1. I do not see the value to use IndexExpr here. Most of code here is to handle the value of constant.  We did not put a general computation for non constant situation, in which the computation is not affine and can not be expressed with IndexExpr.
2. For seq type, the shape inference pass should try to get the precise type for element type as much as possible.  Even though we can alway use !onnx.seq<*xT>. But when an element is loaded out, we will get an unranked tensor at compile time. This is a particular problem for onnx-mlir. This PR infers the rank and size of the tensor for sequence element with the inputs, especially when some of them are constants.

sorenlassen:
@chentong319 thanks for the review, please merge for me

I rebased 1 hour ago and all checks passed

sorenlassen:
@chentong319 please merge this PR for me

PR #1595, opened 2022-08-10
Uplift BatchNormalization, CastLike, Bernoulli to opset 15

philass:
This PR helps move some ops to the latest opset. These are all ops whose latest redefinition are in opset 15, and haven't been updated or added in `onnx-mlir`

Ops introduced in Opset 15
- CastLike
- Bernoulli

Ops changes
- BatchNormalization

Note that BatchNormalization has a more relaxed definition in Opset15. It allows more flexibility in the types of its inputs.


AlexandreEichenberger:
I have not looked at the code in detail yet, our goal is to maintain backward compatibility with older models. So as you upgrade, make sure that the old version also works. Sometimes it is easy, as the old one works by default, sometime there is some minimal work (e.g. when an attribute becomes a input), and sometime we had to rename the old version of the op with an explicit version number (make this one up, `ONNXSqueezeV13Op`), migrate all our current code to that versioned op, and make the new op the default (without version number) with the new support it requires. I believe it is well documented in our docs folder.

AlexandreEichenberger:

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2690,6 +2690,34 @@ LogicalResult ONNXCastOp::inferShapes(
...
 }
 
+//===----------------------------------------------------------------------===//
+// CastLike
+//===----------------------------------------------------------------------===//
+
+LogicalResult ONNXCastLikeOp::inferShapes(
+    std::function<void(mlir::Region &)> doShapeInference) {
+  ShapedType inputType = input().getType().dyn_cast<RankedTensorType>();
+  if (!inputType) {
```
AlexandreEichenberger, line 2700: Any reason that the traditional 
```c++
if (!hasShapeAndRank(input())) {
     return success();
   }
```
is not used here?

Do you want to propagate the `unranked` type? Do you have a specific need for the unranked? My understanding is that onnx-mlir pretty much can't do anything with unranked at this time. But that might be only in our use case (which is lowering to actual code).

AlexandreEichenberger: @chentong319 I see that some shape inference simply propagate the input types to the output, but some other just report `success` when rank and shape are not yet avail. Is that ok?

@philass this question is not about your code, just that it made me realize that we are not entirely consistent. If we wanted to change how we do things, it would come to a different PR, so you are fine.

AlexandreEichenberger (approved):
LGTM

AlexandreEichenberger:
@philass since you don't have rights to push down code (I think), let me know when you are ready to do so.

AlexandreEichenberger (changes requested):
Actually, I have a question before merging. Was there any action that are needed for BatchNorm? You stated that the newer version is more flexible, but I don't see any code change associated to it?

I would venture that if we don't support the lowering of these new options, the lowering will fail? So  shouldn't we do something about it? Minimally generating an error "unsupported" when lowering.

I know your PR is only about shape inference, but it has an impact on the rest of the code.

@chentong319 can you chime in?

src/Dialect/ONNX/ONNXOps.cpp
chentong319: Some really old code did not consider the case of unranked tensor. If the input is unranked, shape inference can do nothing other than return. There is no check when the input type is copied to the output type. The copy may overwrite the shape info coming with the model (such as partial static info).

chentong319:
No change needed for lowering.  The only change of BatchNormalization is that ReduceMean has to be float when the input is float16. That is used only in the train mode. We only lowered BathNormalization for inference mode.

AlexandreEichenberger (approved):
Thanks @chentong319, all good then

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2690,6 +2690,34 @@ LogicalResult ONNXCastOp::inferShapes(
...
+    return success();
+  }
+
+  TensorType targetType = target_type().getType().dyn_cast<TensorType>();
+  if (!inputType) {
+    return success();
+  }
+  auto targetElementType = targetType.getElementType();
+
+  auto getOutputType = [&inputType](Type elementType) -> Type {
```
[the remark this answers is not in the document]
philass: I copied the pattern in Cast https://github.com/onnx/onnx-mlir/blob/7f5c46ded6f07f45d8779b9bc6a550c0f6c21217/src/Dialect/ONNX/ONNXOps.cpp#L2672-L2691

src/Dialect/ONNX/ONNXOps.cpp
philass: I'd be happy to look into improving some of the Shape Inference for better consistency in other PR

philass:
@AlexandreEichenberger I'm ready to have this merged :)

AlexandreEichenberger:
@chentong319 ready to approve? If yes, and @philass is happy with the state of the PR, I can merge it. Tx to all

AlexandreEichenberger:
@chentong319 are you ready to approve this PR, or suggest changes? Tx
@philass there are some conflicts that need manual attention.

philass:
Thanks @AlexandreEichenberger,

I Fixed the conflicts

AlexandreEichenberger:
updating and hopefully the AMD ci will come clean

philass:
Thanks @AlexandreEichenberger

AlexandreEichenberger:

philass:
@chentong319 / @tungld,

Could one of you guys merge it for me?

PR #1596, opened 2022-08-10
Doc fixes

gongsu832:
- add LLVM monitor job result link in DocumentList.md
- fix broken model zoo section in Testing.md

### DIFF
--- a/docs/DocumentList.md
+++ b/docs/DocumentList.md
@@ -25,7 +25,7 @@ at the ONNX operand level.
 * All the passes may be controlled with [options](Options.md).
 * How to handle errors can be found [here](ErrorHandling.md).
 * How to support a new accelerator can be found [here](AddCustomAccelerators).
-* A Jenkins monitor job was setup to help with updating LLVM commit. It locates the next commit we can update to without breaking ONNX-MLIR, as well as the commit that will break ONNX-MLIR. You can see the commit(s) [here](https://www.onnxmlir.xyz/jenkins/job/LLVM-Watch-Docker-Build/LLVM_20Watch_20Report/).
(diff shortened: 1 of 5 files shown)


docs/DocumentList.md
```diff
@@ -25,7 +25,7 @@ at the ONNX operand level.
 * All the passes may be controlled with [options](Options.md).
 * How to handle errors can be found [here](ErrorHandling.md).
 * How to support a new accelerator can be found [here](AddCustomAccelerators).
-* A Jenkins monitor job was setup to help with updating LLVM commit. It locates the next commit we can update to without breaking ONNX-MLIR, as well as the commit that will break ONNX-MLIR. You can see the commit(s) [here](https://www.onnxmlir.xyz/jenkins/job/LLVM-Watch-Docker-Build/LLVM_20Watch_20Report/).
+* A Jenkins monitor job was setup to help with updating LLVM commit. It locates the next commit we can update to without breaking ONNX-MLIR, as well as the commit that will break ONNX-MLIR. You can see the commit(s) here: [s390x](https://www.onnxmlir.xyz/jenkins/job/LLVM-Watch-Docker-Build/LLVM_20Watch_20Report/), [ppc64le](https://www.onnxmlir.xyz/jenkinp/job/LLVM-Watch-Docker-Build/LLVM_20Watch_20Report/), [amd64](https://www.onnxmlir.xyz/jenkinx/job/LLVM-Watch-Docker-Build/LLVM_20Watch_20Report/).
```
AlexandreEichenberger, line 28: You probably need to say the one breaking has the red dot, and the one working has the orange (maybe it could be green as green is typically "good/go").

My preference would be to put this section to https://github.com/onnx/onnx-mlir/blob/main/CONTRIBUTING.md page, if you prefer it here, that is ok too

gongsu832: Green doesn't look good on the tooltip background (platinum) so I used orange. But maybe that's only because I'm color weak. I can change it to green and if it looks OK to you and other people, we can keep it green.

I looked at CONTRIBUTING.md but didn't find an existing section to fit this in. I didn't want to create a whole new section just for this.

AlexandreEichenberger (approved):
LGTM

PR #1602, opened 2022-08-14
Fix llvm monitor job to retrieve llvm-project commit history incrementally

gongsu832:
- Implement incremental retrieval of llvm-project commit history to avoid the 5000/hour github REST API call limit
- Use green (instead of orange) for the successful build color
- Add line number to log output in all build scripts

### DIFF
--- a/.buildbot/LLVM-Watch-Docker-Build.xml
+++ b/.buildbot/LLVM-Watch-Docker-Build.xml
@@ -1,7 +1,7 @@
 <?xml version='1.1' encoding='UTF-8'?>
 <project>
   <actions/>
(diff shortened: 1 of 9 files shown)


AlexandreEichenberger (approved):
LGTM, do you mind just adding in the short text you added on the DocumentList.md that "by zooming on the graph, the green and red dots represent respectively the the last-successful and first-failed builds"

gongsu832:
> LGTM, do you mind just adding in the short text you added on the DocumentList.md that "by zooming on the graph, the green and red dots represent respectively the the last-successful and first-failed builds"

Actually I think it's probably better to do that on the graph itself. Let me try that.

AlexandreEichenberger:
> Actually I think it's probably better to do that on the graph itself.

agreed

PR #1607, opened 2022-08-15
[NNPA] Turn a broadcasting BinaryOp whose one of the inputs is a constant into a non-broadcasting BinaryOp

tungld:
In the bertsquad model, there are many broadcasting BinaryOp(s) whose one of its inputs is a constant, e.g. 
```mlir
%0 = "onnx.Constant"() {value = dense<1.0> : tensor<768xf32>} : () -> tensor<768xf32>
%1 = "onnx.Mul"(%arg0, %189) : (tensor<256x768xf32>, tensor<768xf32>) -> tensor<256x768xf32>
```

Since NNPA does not support broadcasting, these ops will not be executed on NNPA. To overcome this, we expand the constant to have the same shape as the other input, so that there is no broadcasting and the op can be run on NNPA. The above `onnx.Mul` will become:
```mlir
%0 = "onnx.Constant"() {value = dense<1.000000e+00> : tensor<768xf32>} : () -> tensor<768xf32>
%1 = "onnx.Shape"(%arg0) : (tensor<256x768xf32>) -> tensor<2xi64>
%2 = "onnx.Expand"(%0, %1) : (tensor<768xf32>, tensor<2xi64>) -> tensor<256x768xf32>
%3 = "onnx.Add"(%arg0, %2) : (tensor<256x768xf32>, tensor<256x768xf32>) -> tensor<256x768xf32>
```

With `--canonicalize --constprop-onnx`, we get:
```mlir
%0 = "onnx.Constant"() {value = dense<1.000000e+00> : tensor<256x768xf32>} : () -> tensor<256x768xf32>
%1 = "onnx.Add"(%arg0, %0) : (tensor<256x768xf32>, tensor<256x768xf32>) -> tensor<256x768xf32>
```

### DIFF
--- a/utils/mlir2FileCheck.py
+++ b/utils/mlir2FileCheck.py
@@ -199,7 +199,7 @@ def process_line(i, line):
 
     # Process definition of variables.
     # Special handling of function header.
-    if re.match(r'\s+(builtin\.)?func', line) is not None:
+    if re.match(r'\s+(func\.)?func', line) is not None:
         # Have a function: reset dictionary and ref counts
         name_dict = prepare_name_dict.copy()
         refcount_dict.clear()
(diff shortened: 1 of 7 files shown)


utils/mlir2FileCheck.py
```diff
@@ -199,7 +199,7 @@ def process_line(i, line):
 
     # Process definition of variables.
     # Special handling of function header.
-    if re.match(r'\s+(builtin\.)?func', line) is not None:
+    if re.match(r'\s+(func\.)?func', line) is not None:
```
tungld, line 202: This is to fix generating `CHECK-LABEL`. `builtin.func` was changed to `func.func` in the current MLIR.

AlexandreEichenberger: tx

AlexandreEichenberger (approved):
LGTM, will be super helpful

PR #1609, opened 2022-08-15
If

sorenlassen:
Implemented verify() and inferShapes() for If op. Tested with lit tests.

Also reordered the else and then branches so that 
```
  %0 = "onnx.If"(%arg0) ({
    %1 = "onnx.Constant"() {value = dense<1> : tensor<i32>} : () -> tensor<i32>
    onnx.Return %1, %2 : tensor<i32>
  }, {
    %1 = "onnx.Constant"() {value = dense<2> : tensor<i32>} : () -> tensor<i32>
    onnx.Return %1 : tensor<i32>
  }) : (tensor<i1>) -> tensor<i32>
```
has the then branch before the else branch and not the other way around, which I found very counterintuitive. Curiously, the ONNX spec lists the else_branch before the then_branch attribute, so it wasn't enough to not sort the attributes alphabetically and I put in special logic in `gen_onnx_mlir.py` to put then_branch before else_branch, controlled by the `special_attr_order` set.


AlexandreEichenberger:
@chentong319 added you as reviewer since I believe `if` is similar to `loop`, from what I recall from Tian's conversations.

AlexandreEichenberger:
@chentong319 do you mind looking into this one, as it is related to loop? tx

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -4164,11 +4164,128 @@ LogicalResult ONNXHardSwishOp::inferShapes(
   return success();
 }
 
+//===----------------------------------------------------------------------===//
+// ONNXIfOp
+//===----------------------------------------------------------------------===//
+
+namespace {
+bool areIfTypesMergeable(Type ifResultType, Type branchResultType) {
```
chentong319, line 4172: Can this merge function be used for all the joint of control flow, such as Loop? Can you attach a link to MLIR document that specifies the rule?

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -4164,11 +4164,128 @@ LogicalResult ONNXHardSwishOp::inferShapes(
...
+  if (OptType lhsOptType = lhs.dyn_cast<OptType>()) {
+    OptType rhsOptType = rhs.cast<OptType>();
+    return OptType::get(
+        mergeIfTypes(lhsOptType.getElementType(), rhsOptType.getElementType()));
+  }
+  llvm_unreachable("mergeIfTypes called with non tensor/seq/opt type");
+}
+} // namespace
+
+LogicalResult ONNXIfOp::verify() {
```
chentong319, line 4241: I am wondering whether the type of if result should be a super set of type of then branch and else branch. Or in another way, are we determine 1) T1 and T2 are mergable or 2) T1 contains T2?

sorenlassen: I don't think super set would match the the ONNX documentation for the If op which says "The return values in the `then_branch` and `else_branch` must be of the same data type".

src/Dialect/ONNX/ONNXOps.cpp
sorenlassen: I based the implementation on the MLIR documentation of the If op [here](https://github.com/onnx/onnx/blob/main/docs/Operators.md#If) where it says:

> "The return values in the `then_branch` and `else_branch` must be of the same data type. The `then_branch` and `else_branch` may produce tensors with the same element type and different shapes. If corresponding outputs from the then-branch and the else-branch have static shapes S1 and S2, then the shape of the corresponding output variable of the if-node (if present) must be compatible with both S1 and S2 as it represents the union of both possible shapes."

After re-reading this I notice they use the terminology "compatible" and "union", so I renamed `areIfTypesMergeable` to `areCompatibleIfTypes` and renamed `mergeIfTypes` to `unionOfIfTypes`.

I can try to reuse the logic for Loop. If it works, we can rename the methods to something more general and move them to Support/TypeUtilities. How about we leave that for a follow-up PR?

chentong319:
To keep the code simple, I prefer not to change the index for the 'then' and 'else' subgraph. To me, the index is anyway implementation dependent and the string name for the subgraph is always used.  The original order is simply lexicographic order of the name.

sorenlassen:
> To keep the code simple, I prefer not to change the index for the 'then' and 'else' subgraph. To me, the index is anyway implementation dependent and the string name for the subgraph is always used. The original order is simply lexicographic order of the name.

I really think that people are likely to read `"onnx.If(%1) ({A}, {B})` to mean `if %1 then A else B` and therefore it is too confusing if we keep it the other way around. I checked the 'if' constructs in MLIR (scf.if, affine.if, tosa.cond_if) and they all have 'then' before 'else'.

I appreciate that you want to keep the code simple so I refactored the code in `gen_onnx_mlir.py` to stash the complexity into a standalone funtion `order_attr_names` and I polished the implementation and added comments to make it easier to understand. PTAL.

Of course, if I can't change your mind just let me know and I will remove it.

chentong319:
I agree that the output `onnx.If(%1) ({A}, {B})` is confusing to user.  Previously I thought the index is only used in accessing the then/else region in code. It may require more change if we customize the printer and parser for If Op.

chentong319 (approved):
LGTM!

PR #1611, opened 2022-08-15
[Mhlo] More ops and make Mhlo optional

Connor-XY:
This pull request:
1. Add option to the CMakeLists.txt so that Mhlo could be an optional library.
2. Add target as the parameter for decompose pass.
3. Add Mhlo lowering for Pow, Sqrt, Mul, MatMul, Conv, Argmax, Flatten, Shape, Transpose, Identity, Slice, Split, Squeeze, Unsqueeze, and Gather.


src/Conversion/CMakeLists.txt
```diff
@@ -1,7 +1,9 @@
 # SPDX-License-Identifier: Apache-2.0
 
 add_subdirectory(ONNXToKrnl)
-add_subdirectory(ONNXToMhlo)
+if (ONNX_MLIR_ENABLE_MHLO)
+    add_subdirectory(ONNXToMhlo)
```
sstamenova, line 5: Nit: use two spaces for cmake files indentation

CMakeLists.txt
```diff
@@ -9,6 +9,7 @@ option(ONNX_MLIR_BUILD_TESTS "Build ONNX-MLIR test executables. If OFF, just gen
 option(ONNX_MLIR_CCACHE_BUILD "Set to ON for a ccache enabled build." OFF)
 option(ONNX_MLIR_ENABLE_WERROR "Enable warnings as errors." OFF)
 option(ONNX_MLIR_SUPPRESS_THIRD_PARTY_WARNINGS "Suppress warning in third_party code." ON)
+option(ONNX_MLIR_ENABLE_MHLO "Enable MHLO support." ON)
```
sstamenova, line 12: This list is alphabetic

sstamenova:
Making MHLO optional is very exciting. Thank you!

tungld:

src/Conversion/ONNXToMhlo/Math/MatMul.cpp
```diff
@@ -0,0 +1,123 @@
...
+    DimsExpr aDims = shapeHelper.aDims;
+    DimsExpr bDims = shapeHelper.bDims;
+    llvm::BitVector aPadDims = shapeHelper.aPadDims;
+    llvm::BitVector bPadDims = shapeHelper.bPadDims;
+
+    DimsExpr outputDims = shapeHelper.dimsForOutput();
+    llvm::SmallVector<int64_t, 4> aShape;
+    llvm::SmallVector<int64_t, 4> bShape;
+    for (int64_t i = 0; i < paddedRank - 2; i++) {
+      aShape.push_back(outputDims[i].isLiteral() ? outputDims[i].getLiteral()
```
sorenlassen, line 57: suggestion: define a helper function to map an IndexExpr to its literal value or -1

src/Conversion/ONNXToMhlo/Tensor/ArgMax.cpp
```diff
@@ -0,0 +1,160 @@
...
+    // Gather info.
+    Location loc = op->getLoc();
+    ONNXArgMaxOpAdaptor operandAdaptor(operands);
+    ONNXArgMaxOp argMaxOp = llvm::cast<ONNXArgMaxOp>(op);
+
+    // shape helper
+    ONNXArgMaxOpShapeHelper shapeHelper(&argMaxOp);
+    LogicalResult shapecomputed = shapeHelper.computeShape(operandAdaptor);
+    (void)shapecomputed;
+    assert(!failed(shapecomputed) && "expected to succeed");
```
sorenlassen, line 72: message is not very informative, I think "Could not compute output shape" would be better

src/Conversion/ONNXToMhlo/NN/Conv.cpp
```diff
@@ -0,0 +1,133 @@
...
+    bool needPadding = (padding == "NOTSET");
+    for (int64_t i = 0; i < spatialRank; i++) {
+      if (!needPadding) {
+        flattenPaddings.push_back(pads[i].getLiteral());
+        flattenPaddings.push_back(pads[i + spatialRank].getLiteral());
+      } else {
+        int64_t kdTerm = (kernelShape[i].getLiteral() - 1) * dilations[i] + 1;
+        int64_t padFront = pads[i].getLiteral();
+        int64_t padBack =
+            (outputDims[i + spatialOffset].getLiteral() - 1) * strides[i] +
```
sorenlassen, line 91: do we know that all these things we call getLiteral() on are literals?

src/Conversion/ONNXToMhlo/NN/Conv.cpp
```diff
@@ -0,0 +1,133 @@
...
+            kdTerm - inputShape[i + spatialOffset] - padFront;
```
sorenlassen, line 92: do we know that inputShape[i + sparialOffset] is not -1?

sorenlassen (commented):
I’m interested in learning about lowering so I read through the PR and made some comments, but I’m no expert and feel free to ignore if the comments are not helpful.

sorenlassen:
@Connor-XY ^

test/mlir/conversion/onnx_to_mhlo/Math/Reduction.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: onnx-mlir-opt --convert-onnx-to-mhlo %s -split-input-file | FileCheck %s
+// RUN: onnx-mlir-opt --decompose-onnx="target=mhlo" --shape-inference --convert-onnx-to-mhlo %s --canonicalize -split-input-file | FileCheck %s
```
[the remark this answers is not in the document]
Connor-XY: We decompose the softmax operator into primitive operators when the target is mhlo. We add the other two options to do shape inference and canonicalize the result

src/Conversion/ONNXToMhlo/Tensor/ArgMax.cpp
```diff
@@ -0,0 +1,160 @@
...
+        reductionDimensions);
+    BuildArgmaxReductionBody(
+        elementType, indexElementType, &reduction.body(), &rewriter);
+
+    Value result = reduction.getResult(1);
+    if (isKeepdims) {
+      if (outputShapedType.hasStaticShape())
+        result = rewriter.create<mhlo::ReshapeOp>(loc, outputType, result);
+      else {
+        SmallVector<Value> dims;
```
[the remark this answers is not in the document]
Connor-XY: The shape is dynamic, so we still need to use shape dialect to get the dim size of the output.

src/Conversion/ONNXToMhlo/Tensor/Flatten.cpp
```diff
@@ -0,0 +1,75 @@
...
+    ONNXFlattenOpAdaptor operandAdaptor(operands);
+    ONNXFlattenOp flattenOp = llvm::cast<ONNXFlattenOp>(op);
+
+    Value input = operandAdaptor.input();
+    assert(isRankedShapedType(input.getType()) && "Expected Ranked ShapedType");
+    ShapedType inputType = input.getType().cast<ShapedType>();
+    int64_t rank = inputType.getRank();
+    int64_t axis = flattenOp.axis();
+    axis = axis >= 0 ? axis : rank + axis;
+    assert(axis >= -rank && axis <= rank - 1);
```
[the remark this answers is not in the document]
Connor-XY: Agree, I'll fix that.

src/Conversion/ONNXToMhlo/Tensor/ArgMax.cpp
```diff
@@ -0,0 +1,160 @@
...
+    // axis & keepdims attribute
+    int64_t axis = argMaxOp.axis();
+    assert(axis >= -dataRank && axis <= dataRank - 1);
+    axis = axis >= 0 ? axis : (dataRank + axis);
+
+    int64_t keepdims = argMaxOp.keepdims();
+    bool isKeepdims = (keepdims == 1) ? true : false;
+
+    // Get type information
+    llvm::SmallVector<int64_t, 1> axes;
```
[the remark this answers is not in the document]
Connor-XY: Agree, I'll fix that.

src/Conversion/ONNXToMhlo/Math/MatMul.cpp
Connor-XY: I'll add the helper function

src/Conversion/ONNXToMhlo/Tensor/ArgMax.cpp
Connor-XY: will fix that

src/Conversion/ONNXToMhlo/NN/Conv.cpp
Connor-XY: The current conv only supports where padding and kernel size are known. We will add support for dynamic shape.

sorenlassen: I see

I suggest you add assertions to make it clear that pads, kernelShape, outputDims, and inputShape must all be known (unless this is already checked elsewhere)

src/Conversion/ONNXToMhlo/Tensor/Flatten.cpp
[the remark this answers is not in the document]
AlexandreEichenberger: I would  first add the rank when negative, and then the axis should be from 0 inclusive to rank exclusive.

test/mlir/conversion/onnx_to_mhlo/Math/Reduction.mlir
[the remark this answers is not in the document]
AlexandreEichenberger: I know we may not have set the best example, the preferred way is to test only one pass at a time, so that potential errors are easier to detect. I would then do 2 test files, one to test the first pass and one to do the second pass. 

We have often added `canonicalize` as a secondary pass as often the output without canonicalize is very redundant, so using this as a second pass is fine if it makes the test easier.

src/Conversion/ONNXToMhlo/Math/MatMul.cpp
AlexandreEichenberger: This function already exists, check in IndexExpr.hpp
```C++
  // Helpers for list of IndexExpressions
  static void getShape(llvm::SmallVectorImpl<IndexExpr> &indexExprList,
      llvm::SmallVectorImpl<int64_t> &intDimList);
```

src/Conversion/ONNXToMhlo/NN/Conv.cpp
AlexandreEichenberger: My recollection is that we support dynamic images sizes, but expect the filter size to be known at compile time...
I will try to add some dyn size testing if its not already there.

AlexandreEichenberger:

src/Conversion/ONNXToMhlo/NN/Conv.cpp
AlexandreEichenberger: Confirmed, in test/modellib/ConvModel and test/numerical/TestConv we are building/testing 2D convolutions that are dynamic in each of the NCHW dimensions.

AlexandreEichenberger:

src/Conversion/ONNXToMhlo/NN/Conv.cpp
```diff
@@ -0,0 +1,147 @@
...
+          return failure();
+      }
+    }
+
+    for (int64_t i = 0; i < spatialRank; i++) {
+      if (!needPadding) {
+        flattenPaddings.push_back(pads[i].getLiteral());
+        flattenPaddings.push_back(pads[i + spatialRank].getLiteral());
+      } else {
+        int64_t kdTerm = (kernelShape[i].getLiteral() - 1) * dilations[i] + 1;
```
AlexandreEichenberger, line 102: Our shape inference compute left/right top/bottom padding. Is the mhlo padding different than onnx and thus justify recomputing some of it here?

AlexandreEichenberger (commented):
I left some questions and comments in the text. 
I can see that there is a pattern that is starting to add a lot of code: if dimensions are constant, then use them, otherwise introduce a lot of shape ops to compute it.

We used a similar pattern at the beginning (before shape dialect was established I believe). It introduced a lot of redundant code, that was difficult to debug/upgrade as ONNX ops sometime change what is attributes vs what can become a dynamic input (with potentially dynamic dimensions).

This is why we introduced IndexExpr which can (depending on the context) use the same computation to either compute a shape number (nonnegative if known at compile time, -1 if dynamic) OR introduce computations to determine the shape.

I would generally encourage you to do so, possibly updating our IndexExpr to use shape operations if you don't tolerate our computations. I believe we use mostly arithmetic plus DimOp, which fetch the dimension out of a tensor. We have a slight complication in that we have accelerators that use maps for data layout, and we need to keep both the iteration space (original sizes of the tensors) as well as the mapped data size. We had to introduce some ops to handle this, maybe MLIR support has improved since then.

Basically, I am encouraging you to reuse our infra (and possibly upgrade it a bit) so that you don't have to generate a lot of complicated code to recompute the shape when dynamic using redundant code.

src/Conversion/ONNXToMhlo/NN/Conv.cpp
yaochengji: Let me help answer this question. `mhlo.convolution` is a generic N-D conv and its padding is of shape (N, 2), where padding[i] specifies lower and upper padding values of spatial dimension i. You can refer to [this](https://www.tensorflow.org/xla/operation_semantics#convwithgeneralpadding_convolution) for more details.

yaochengji:
@AlexandreEichenberger do you think we need to add some milestone model tests here? Ex. resnet 18, bert-tiny.

AlexandreEichenberger:
> @AlexandreEichenberger do you think we need to add some milestone model tests here? Ex. resnet 18, bert-tiny.

We currently test for a few benchmarks as part of our backend tests (run on all of our mandatory CIs): from inference_backend.py:

```
        "test_densenet121_cpu": {STATIC_SHAPE:{}},
        "test_inception_v1_cpu": {STATIC_SHAPE:{}},
        "test_resnet50_cpu": {STATIC_SHAPE:{}, DYNAMIC_SHAPE:{0:{-1}}},
        "test_shufflenet_cpu": {STATIC_SHAPE:{}},
        "test_squeezenet_cpu": {STATIC_SHAPE:{}},
        "test_vgg19_cpu": {STATIC_SHAPE:{}},
```

I am not aware on how to add tests there, we need to be a bit careful about not taking too long as some of our CIs may timeout.
If you believe that an additional benchmark should be here, I would suggest a distinct PR to add it.

src/Conversion/ONNXToMhlo/Math/Elementwise.cpp
```diff
@@ -159,17 +139,14 @@ struct ONNXElementwiseUnaryOpLoweringToMhlo<ONNXReluOp>
       : ConversionPattern(ONNXReluOp::getOperationName(), 1, ctx) {}
   LogicalResult matchAndRewrite(Operation *op, ArrayRef<Value> operands,
       ConversionPatternRewriter &rewriter) const final {
-    Location loc = NameLoc::get(
-        StringAttr::get(op->getContext(), ONNXReluOp::getOperationName()),
-        op->getLoc());
+    Location loc = op->getLoc();
```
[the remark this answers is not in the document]
Connor-XY: This is reasonable and we will add the comment

src/Conversion/ONNXToMhlo/Math/Elementwise.cpp
```diff
@@ -228,6 +201,32 @@ struct ONNXElementwiseBinaryOpLoweringToMhlo : public ConversionPattern {
   }
 };
 
+template <>
+struct ONNXElementwiseBinaryOpLoweringToMhlo<ONNXPowOp>
```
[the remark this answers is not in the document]
Connor-XY: This is answered above

src/Conversion/ONNXToMhlo/Math/MatMul.cpp
```diff
@@ -0,0 +1,118 @@
...
+    DimsExpr aDims = shapeHelper.aDims;
+    DimsExpr bDims = shapeHelper.bDims;
+    llvm::BitVector aPadDims = shapeHelper.aPadDims;
+    llvm::BitVector bPadDims = shapeHelper.bPadDims;
+
+    DimsExpr outputDims = shapeHelper.dimsForOutput();
+    llvm::SmallVector<int64_t, 4> aShape;
+    llvm::SmallVector<int64_t, 4> bShape;
+    for (int64_t i = 0; i < paddedRank - 2; i++) {
+      aShape.push_back(getLiteralValue(outputDims[i]));
```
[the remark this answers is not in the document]
Connor-XY: Yes, MHLO can handle -1 shapes. It seems that we need to figure out how to reuse IndexExpr and modify your previous implementation for MHLO conversion. We can use the getShape() function, and also add the new call in IndexExpr.

src/Conversion/ONNXToMhlo/NN/Conv.cpp
```diff
@@ -0,0 +1,147 @@
...
+      : ConversionPattern(mlir::ONNXConvOp::getOperationName(), 1, ctx) {}
+
+  LogicalResult matchAndRewrite(Operation *op, ArrayRef<Value> operands,
+      ConversionPatternRewriter &rewriter) const final {
+
+    ONNXConvOpAdaptor operandAdaptor(operands);
+    ONNXConvOp convOp = llvm::dyn_cast<ONNXConvOp>(op);
+    Location loc = op->getLoc();
+
+    ONNXConvOpShapeHelper shapeHelper(&convOp);
```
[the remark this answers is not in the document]
Connor-XY: I remember that when I tried to use shapehelper for sliceOp, fGetDenseVal failed since we only pass the operator pointer to the shapeHelper, but I couldn't remember clearly what the issue is.

src/Conversion/ONNXToMhlo/NN/Conv.cpp
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    // paddings
+    DimsExpr pads = shapeHelper.pads;
+    llvm::StringRef padding = convOp.auto_pad();
+    int64_t spatialRank = rank - spatialOffset;
+    SmallVector<int64_t> flattenPaddings;
+    bool needPadding = (padding == "NOTSET");
+
+    // currently only support static shape
```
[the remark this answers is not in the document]
Connor-XY: Yes, we can use the function to test if all index expressions are literals, and we need to modify IndexExpr so we can reuse the result from the shape helper.

sstamenova:
@Connor-XY How difficult would it be to split out the change that makes mhlo optional? It seems like there's no concern about that particular aspect and it would be great to have sooner rather than later.

yaochengji:
> @Connor-XY How difficult would it be to split out the change that makes mhlo optional? It seems like there's no concern about that particular aspect and it would be great to have sooner rather than later.

Hi @sstamenova ,

@Connor-XY gets a cold these days. I think making mhlo optional is not difficult.

src/Conversion/ONNXToMhlo/NN/Conv.cpp
Connor-XY: Due to ceil mode in ONNX, the padding can be the same even when the output shape differs when auto_pad is NOTSET. Therefore, we need to recalculate the padding. You could refer to @test_default_maxpoolsingleout_strides_nonunifpad and @test_default_maxpoolsingleout_strides_nonunifpad_ceil in both test/mlir/conversion/onnx_to_mhlo/NN/Pooling.mlir and test/mlir/onnx/onnx_shape_inference_maxpool.mlir, and see the difference.

src/Conversion/ONNXToMhlo/NN/Conv.cpp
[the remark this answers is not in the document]
Connor-XY: You could refer to the case below where we need to calculate the padding again. For simplicity we previously just calculate all paddings instead of reuse those in the shape helper.

Connor-XY:
@AlexandreEichenberger I believe it is beneficial if we can reuse IndexExpr, but I currently have some concerns about this. First of all, it seems that in shape inference pass, when dimension is dynamic, all of the IndexExprs are QuestionmarkIndexExpr. Ideally we hope that the final result could be a Value and we generate the code after calling the shape helper so we can reuse the result in the lowering. This is different from what it is currently and I believe this requires substantial changes to the current design. What's more, I currently couldn't figure out a good way to differentiate Mhlo tensor and memref. I think it should be done wrt the scope or the rewriter, but couldn't think of a proper way to do that. We would like to ask for some suggestions about how we should modify IndexExpr so we could reuse your calculation while not affect the proper usage of the structure.

AlexandreEichenberger:
> all of the IndexExprs are QuestionmarkIndexExpr.

and that is fine. When half the dims are constants, and the other are dynamic, the results will be a mixture of literals and question marks.

> final result could be a Value 

when doing shape inference, the only interesting outcome (short of symbolic analysis) is either constant lit or unknown. There is no need for values at this time. When lowering code, then we want the literal and/or the computations, all of which can be values. There are methods that transform index expressions in literal (individual or list of).

> to differentiate Mhlo tensor and memref. 

are MHLO tensors different from the MLIR tensors? I would look at how we use index expression now. During shape inference, for sure we use the tensors. During lowering, we have access to both unconverted inputs (tensors) or converted inputs (members). Not 100% sure which one we use.

I believe we want to avoid redundancy of code during shape inference and lowering. How we get there is less important, though we have invested in IndexExpr and it would be nice to adapt it to everyone's need. We can look together for solutions that works for your needs

yaochengji:
> when doing shape inference, the only interesting outcome (short of symbolic analysis) is either constant lit or unknown.

@Connor-XY , agreed with @AlexandreEichenberger, it is shape inference, not shape reification, we only need shape calculation for the needs of operations like mhlo.dynamic_reshape.


> are MHLO tensors different from the MLIR tensors? 

@Connor-XY I don't think we need to take memref into consideration. All the inputs/outputs of MHLO and ONNX dialect are of type tensor.

yaochengji:
Hi @AlexandreEichenberger , currently @Connor-XY 's internship in Bytedance is over so I'll take over this PR.

I have a question when extracting `Value` from `MemRefBoundsIndexCapture` or `IndexExpr`.
Ex. when converting onnx flattern op to mhlo, currently the code snippet handling the shape computation is as below:
```
Value flattenDimFirst = rewriter.create<arith::ConstantIndexOp>(loc, 1);
Value inputShape = rewriter.create<shape::ShapeOfOp>(loc, input);
for (int64_t i = 0; i < axis; i++) {
  Value dim = rewriter.create<shape::GetExtentOp>(loc, inputShape, i);
  flattenDimFirst =
      rewriter.create<shape::MulOp>(loc, flattenDimFirst, dim);
}
```

After changing to `IndexExpr` utility, I guess it should be as below:
```
MemRefBoundsIndexCapture XBounds(op.X());
IndexExpr mulDim = XBounds.get(0);
for (int64_t i = 1; i < axis; i++) {
  mulDim = mulDim * XBounds.get(1);
}
```

Besides, it seems I should add a correct `IndexExprScope` and modify the `MemRefBoundsIndexCapture::get` implementation code to make it works correct. Therefore, I think in the `Value` extraction scenario `IndexExpr` doesn't simplify the code very much. Could we keep code like this as it is?

yaochengji:
Hi @AlexandreEichenberger, I rebased this branch and fixed some error after it. Could you take a look again?

philass:

src/Conversion/ONNXToMhlo/Tensor/Transpose.cpp
```diff
@@ -0,0 +1,80 @@
...
+    ShapedType outputShapedType = outputType.cast<ShapedType>();
+    int64_t rank = outputShapedType.getShape().size();
+
+    // Attributes
+    llvm::Optional<ArrayAttr> permAttr = transposeOp.perm();
+    DenseIntElementsAttr permAxis;
+    RankedTensorType permAxisType =
+        RankedTensorType::get({rank}, rewriter.getI64Type());
+    SmallVector<int64_t, 4> permAxisList;
+    if (permAttr.hasValue()) {
```
philass, line 50: I think this should be `has_value()`, and thats what causes the windows pipeline to fail.

`hasValue()` was deprecated in MLIR

yaochengji: Fixed, thanks @philass

philass:

AlexandreEichenberger (commented):
Some work for the copyright comments.
This code still has a lot of "if constant do this, if dynamic do that". That is the pattern that our ShapeInference data structures avoid, as we use IndexExpression to compute either the dynamic value, or simplify to integer when all the values involved in the computations are constant.

My concern is that the code with the "if constant/dynamic" pattern is more complex to maintain long term. So I continue to advocate to migrate to that new code pattern.

Maybe you want to put this code down first, and then investigate if you can migrate to using our shape inference helper classes?

src/Conversion/ONNXToMhlo/Math/MatMul.cpp
```diff
@@ -0,0 +1,128 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//===----------------- Matmul.cpp - Lowering Matmul Op --------------------===//
+//
+// Copyright 2019-2022 The IBM Research Authors.
```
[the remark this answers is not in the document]
yaochengji: I only put the date this time. Because I don't know what process to go through to put the company name (ByteDance) here. It may take some time to figure out.

yaochengji:
> My concern is that the code with the "if constant/dynamic" pattern is more complex to maintain long term. So I continue to advocate to migrate to that new code pattern.

If your concern is "if constant/dynamic" pattern makes things complex, we can treat all situation as dynamic in onnx->mhlo conversion and then a canonicalization pass will be added after that to simplify the static situations. BTW, this is how it implemented in TF. For IndexExpr, my concern is that it is not flexible enough to support arbitrary cases.


> Maybe you want to put this code down first, and then investigate if you can migrate to using our shape inference helper classes?

Yes, could we refine the code later.

src/Conversion/ONNXToMhlo/Math/MatMul.cpp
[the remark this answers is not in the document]
AlexandreEichenberger: Then just leave your company out... but it does not seems to be right to add `IBM Authors` when you are not employed by IBM, if you see what I mean.

AlexandreEichenberger:
> Besides, it seems I should add a correct IndexExprScope and modify the MemRefBoundsIndexCapture::get implementation code to make it works correct. Therefore, I think in the Value extraction scenario IndexExpr doesn't simplify the code very much. Could we keep code like this as it is?

The issue is not so much using index expression vs adding code directly, it is more about reusing the shape inference helpers (or of they are not used, maybe migrating the current code into a shape helper so that it may be reused).

Unless the computations are distinct because you are targeting MHLO, our goal should be to reuse shape inference as much as possible and avoid "if constant/dynamic" branches.

src/Conversion/ONNXToMhlo/Math/MatMul.cpp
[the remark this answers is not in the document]
yaochengji: I didn't add `IBM Authros`, either. Only `Copyright 2022` currently.

yaochengji:
fix some conflicts.

@AlexandreEichenberger do you have any other comments? If not, could this PR be merged?

yaochengji:
> fix some conflicts.
> 
> @AlexandreEichenberger do you have any other comments? If not, could this PR be merged?

@tungld could you help take a look at this? (I guess @AlexandreEichenberger is on leave these days)

yaochengji:
@tungld the test passed. Is there any work we should do on this PR?

tungld:
@yaochengji the patch looks good to me. However, still would like to hear the final decision from @AlexandreEichenberger as he has followed the patch since its inception.

AlexandreEichenberger (approved):
LGTM, urging future PR to rely more on existing shape inference framework to reduce redundancy.

yaochengji:
> LGTM, urging future PR to rely more on existing shape inference framework to reduce redundancy.

Sure, this will be our next PR.

yaochengji:
@AlexandreEichenberger As it is already approved, could you merge this PR after CI passes?

AlexandreEichenberger:
> As it is already approved, could you merge this PR after CI passes?

will do

PR #1612, opened 2022-08-16
[NNPA] Rewrite N-D MatMul into 3-D MatMul so that it can run on NNPA

tungld:
This patch is to rewrite N-D MatMul into 3-D MatMul so that it can run on NNPA. For example, the following MatMul:
```mlir
func.func @test_matmul(%arg0: tensor<4x12x256x256xf32>, %arg1: tensor<4x12x256x64xf32>) -> (tensor<4x12x256x64xf32>) {
    %0= "onnx.MatMul"(%arg0, %arg1) : (tensor<4x12x256x256xf32>, tensor<4x12x256x64xf32>) -> tensor<4x12x256x64xf32>
    return %0 : tensor<4x12x256x64xf32>
}
```
will become:
```mlir
func.func @test_matmul(%arg0: tensor<4x12x256x256xf32>, %arg1: tensor<4x12x256x64xf32>) -> tensor<4x12x256x64xf32> {
    %0 = "onnx.Constant"() {value = dense<[-1, 256, 256]> : tensor<3xi64>} : () -> tensor<3xi64>
    %1 = "onnx.Reshape"(%arg0, %0) {allowzero = 0 : si64} : (tensor<4x12x256x256xf32>, tensor<3xi64>) -> tensor<48x256x256xf32>
    %2 = "onnx.Constant"() {value = dense<[-1, 256, 64]> : tensor<3xi64>} : () -> tensor<3xi64>
    %3 = "onnx.Reshape"(%arg1, %2) {allowzero = 0 : si64} : (tensor<4x12x256x64xf32>, tensor<3xi64>) -> tensor<48x256x64xf32>
    %4 = "onnx.MatMul"(%1, %3) : (tensor<48x256x256xf32>, tensor<48x256x64xf32>) -> tensor<48x256x64xf32>
    %5 = "onnx.Constant"() {value = dense<[4, 12, 256, 64]> : tensor<4xi64>} : () -> tensor<4xi64>
    %6 = "onnx.Reshape"(%4, %5) {allowzero = 0 : si64} : (tensor<48x256x64xf32>, tensor<4xi64>) -> tensor<4x12x256x64xf32>
    return %6 : tensor<4x12x256x64xf32>
  }
```

### DIFF
--- a/src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
+++ b/src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
@@ -72,6 +72,7 @@ void addONNXToZHighPasses(
   pm.addPass(onnx_mlir::createShapeInferencePass());
   pm.addPass(mlir::createCanonicalizerPass());
   pm.addNestedPass<func::FuncOp>(onnx_mlir::createConstPropONNXToONNXPass());
+  pm.addPass(onnx_mlir::createShapeInferencePass());
   // Add instrumentation for Onnx Ops in the same way as onnx-mlir.
   if (instrumentZHighOps == "" || instrumentZHighOps == "NONE")
     pm.addNestedPass<func::FuncOp>(onnx_mlir::createInstrumentONNXPass(
--- a/src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHighCommon.td
+++ b/src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHighCommon.td
@@ -57,4 +57,10 @@ def ACT_RELUAttr: NativeCodeCall<"$_builder.getStringAttr(\"ACT_RELU\")">;
 
 def GetTypeOf : NativeCodeCall<"$0.getType()" >;
 
+def GetNullAttr : NativeCodeCall<"Attribute()">;
+
+def GetZeroI64Attr: NativeCodeCall<
+  "IntegerAttr::get($_builder.getIntegerType(64, /*isSigned=*/true), APInt(64, 0, /*isSigned=*/true))"
+>;
+
 #endif // ONNX_TO_ZHIGH_COMMON
--- a/src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
+++ b/src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
@@ -60,6 +60,7 @@ Value getSqrtResultBatchNormA(
 Value reshapeTo3D(PatternRewriter &rewriter, Location loc, Value val) {
   MultiDialectBuilder<OnnxBuilder> create(rewriter, loc);
   int64_t rank = getRank(val.getType());
+  assert(rank > 3 && "Require rank > 3");
   Type elementType = getElementType(val.getType());
   Type shapeType = RankedTensorType::get({rank}, rewriter.getI64Type());
   Type twoElemType = RankedTensorType::get({2}, rewriter.getI64Type());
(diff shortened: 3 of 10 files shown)


tungld:

tungld:

src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
```diff
@@ -72,6 +72,7 @@ void addONNXToZHighPasses(
   pm.addPass(onnx_mlir::createShapeInferencePass());
   pm.addPass(mlir::createCanonicalizerPass());
   pm.addNestedPass<func::FuncOp>(onnx_mlir::createConstPropONNXToONNXPass());
+  pm.addPass(onnx_mlir::createShapeInferencePass());
```
chentong319, line 75: Is this additional pass needed by this change or in general?

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHighCommon.td
```diff
@@ -57,4 +57,10 @@ def ACT_RELUAttr: NativeCodeCall<"$_builder.getStringAttr(\"ACT_RELU\")">;
 
 def GetTypeOf : NativeCodeCall<"$0.getType()" >;
 
+def GetNullAttr : NativeCodeCall<"Attribute()">;
+
+def GetZeroI64Attr: NativeCodeCall<
+  "IntegerAttr::get($_builder.getIntegerType(64, /*isSigned=*/true), APInt(64, 0, /*isSigned=*/true))"
+>;
```
chentong319, line 64: When we have time, we may need have a common.td for all the rewriting.

src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
tungld: It is needed in general like what you did to repeat "constprop->shape inference->...".

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHighCommon.td
tungld: Totally agree, there are many duplicated definitions now.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
```diff
@@ -170,6 +307,48 @@ void RewriteONNXForZHighPass::runOnOperation() {
...
+  target.addDynamicallyLegalOp<ONNXMatMulOp>([](ONNXMatMulOp op) {
+    Type aType = op.A().getType();
+    Type bType = op.B().getType();
+    if (!isRankedShapedType(aType) || !isRankedShapedType(bType))
+      return true;
+
+    // Only support static shape at this moment though the code supports dynamic
+    // shape as well.
+    //
+    // The reason is that lowering (3Dx3D) ONNXMatMul of dynamic shape to NNPA
```
chentong319, line 324: Is the error related to reshaping of 4D to 3D?

tungld: It seems reshaping of 4D to 3D is correct. For example, keeping the code that reshapes 4D to 3D and using 3D ONNXMatMul leads to the correct result. But if we allow lowering 3D ONNXMatMul to NNPA instead of CPU, we got a wrong result. Looks like something is wrong when lowering a 3D ONNXMatMul of dynamic shape to NNPA or when the lowering is combined with other lowerings.

chentong319 (approved):
The PR itself is fine.
Some comments for future:
1. It seems to me that it is easier to fulfill this goal after the tensor is lowered to memref. memref cast can be used directly to perform the reshape. Is it true?
2. The lit test for constant shape has several optimization after this rewriting. Does our driver possibly have phase issue? Will slice and concat op remain for dynamic shape?
3. Best to add such function into zenDNN library.

tungld:
> It seems to me that it is easier to fulfill this goal after the tensor is lowered to memref. memref cast can be used directly to perform the reshape. Is it true?

Once the tensor is lowered to memref, ONNXMatMul has been lowered to krnl loops and it is non-trivial to transform krnl loops to zhigh.MatMul for NNPA. So rewriting N-D ONNXMatMul at an early stage is important.

> Will slice and concat op remain for dynamic shape?

Yes, they will.

chentong319:
@tungld I mean in function ZHighToZLowMatMulOpLowering. The input is already memref type. Add memref.cast when needed and then create ZLowMatMulOp. Not sure about the interaction with stickify.

PR #1614, opened 2022-08-17
Reduce the number of parallel jobs to avoid OOM

gongsu832:
It appears that memory usage for compiling and/or running models are increasing, frequently causing OOM on the Jenkins CI servers. So reduce the number of parallel jobs to avoid OOM.


gongsu832:
Problem appears to be a bad PR. Close for now.

PR #1616, opened 2022-08-18
Try fixing onnx installation on MacOS

gongsu832:
On MacOS, replace "pip3 install -e" with "python3 setup install" when installing onnx


tungld (approved):
LGTM.

PR #1618, opened 2022-08-19
Introduce attribute delayed in krnl.memset and enable N-D Matmul to 3-D MatMul for dynamic shape

tungld:
In PR #1612, we disabled support for dynamic shape when rewriting N-D Matmul into 3-D Matmul due to an unknown issue.

After investigating it, I found that it was caused by `krnl.memset` when the input is **a MemRef with affine_map**. More info about the issue is in [the below comment](https://github.com/onnx/onnx-mlir/pull/1618#issuecomment-1223420437).

This patch fixes the issue by introducing an attribute, named `delayed`, in `krnl.memset` to indicate two modes of `krnl.memset`: 
- `delayed = false`: do memset over **the original iteration space** of the MemRef with affine_map.
- `delayed = true`: do memset over **the extended iteration space** (the space after the MemRef is normalized) of the MemRef with affine_map.

For example, if we want to set original elements to 1.0 and padding elements to 0.0 in a tiled MemRef:
```
   #map_2ds = affine_map<(d0, d1) -> (d0, d1 floordiv 64, 0, 0, 31, d1 mod 64)>
   %2 = memref.alloc(%i0, %i1) {alignment = 4096 : i64} : memref<?x?xf16, #map_2ds>

  // Set all elements including padding elements to 0.0.
  krnl.memset %2, %cst_f0 {delayed = true} : memref<?x?xf16, #map_2ds>

  // Set elements in the original area (non-padding elements) to 1.0
  krnl.memset %2, %cst_f1 {delayed = false} : memref<?x?xf16, #map_2ds>

```

### DIFF
--- a/src/Dialect/Krnl/Krnl.td
+++ b/src/Dialect/Krnl/Krnl.td
@@ -1075,10 +1075,30 @@ def KrnlMemsetOp : Op<Krnl_Dialect, "memset", [MemRefsNormalizable,
                    "$_self.cast<MemRefType>().getElementType()">]> {
   let summary = "Set buffer to a given value.";
   let description = [{
-    Krnl operation that set buffer to a given value.
+    Krnl operation that sets a buffer to a given value.
+    In case that the buffer is a MemRef with affine_map, `delayed` indicates
+    whether we set values along original or extended iteration space.
+    
+    For example, given 
+    - an affine_map `#tile = affine_map < (i)->(i floordiv 4, i mod 4) >`, and
+    - a buffer of type `memref<5xf32, #tile>`
+
+    Original iteration space is along the first axis that has 5 elements.
+
+    If we do normalization, the memref becomes `memref<2x4xf32>`. Now we have
+    an extended iteration space along two axes of sizes 2 and 4, respectively.
+    This extended iteration space has 8 elements in total.
+
+    If `delayed = false`, the original iteration space is used to set values.
+    In the above example, only 5 out of 8 elementes will be set to the given value.
+
+    If `delayed = true`, the extended iteration space is used to set values.
+    In the above example, all 8 elements will be set to the given value.
+
   }];
 
-  let arguments = (ins AnyMemRef:$dest, AnyType: $value); 
+  let arguments = (ins AnyMemRef:$dest, AnyType: $value,
+    DefaultValuedAttr<BoolAttr, "false">:$delayed); 
 
   let assemblyFormat = [{ $dest `,` $value attr-dict `:` type($dest) }];
 }
--- a/docs/SupportedONNXOps-NNPA.md
+++ b/docs/SupportedONNXOps-NNPA.md
@@ -26,7 +26,7 @@ NNPA has hardware limitations in dimension index size and tensor size, which are
 | **LSTM** |14 |- `direction` and `hidden_size` in `W` must have static dimensions.<br>- `R` must have static dimensions.<br>- `B` and `initial_h` have static dimensions if given. `B`'s direction dim must be 1 or 2.<br>- `P`(peepholes), `activation_alpha`, and `activation_beta` are not supported.<br>- `activations` must be `["Sigmoid", "Tanh", "Tanh"]`.<br>- `clip` is not supported.<br>- `input_forget` must be default value(0).<br>- `layout` is not supported. | |
 | **Log** |13 |Input tensor must have 4 dimensions. | |
 | **LogSoftmax** |13 | | |
-| **MatMul** |13 |Ranks of input tensors must be (Rank of A, Rank of B) = (M, N), where M >= 2 and N >= 2. If M or N > 3, only supports static shape at this moment. | |
+| **MatMul** |13 |Ranks of input tensors must be (Rank of A, Rank of B) = (M, N), where M >= 2 and N >= 2. | |
 | **Max** |13 |- Shape of input tensors must be the same since broadcasting is not supported.<br>- Input tensors must have static dimensions. | |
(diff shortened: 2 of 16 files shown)


tungld:
Convert to Draft to investigate more.

chentong319 (approved):
LGTM!
Any insight into why?

AlexandreEichenberger:
Out of curiosity, N-d matmul is just multiple 2D matmul correct? So would there be any reason not to collapse the N-dim to 3D at all times to have fewer outer loops? Or maybe it does not matter? I would think that if we try to parallelize, it would be nicer to have only the 3rd dim to parallelize.

tungld:
> N-d matmul is just multiple 2D matmul correct? So would there be any reason not to collapse the N-dim to 3D at all times to have fewer outer loops?

It's correct. No reason except I just do it for NNPA now. We can move the rules to a common place without any modification. I can do it in this patch too.

tungld:
@AlexandreEichenberger Regarding the issue of wrong results on NNPA when 3D Matmul has dynamic shape, I summary what Tong and I discussed offline and would like to hear your thoughts.

NNPA Matmul's signature is 
```C
zdnn_status zdnn_matmul_op(const zdnn_ztensor *input_a,
                           const zdnn_ztensor *input_b,
                           const zdnn_ztensor *input_c,
                           zdnn_matmul_ops op_type, zdnn_ztensor *output);
```
that computes `input_a * input_b + input_c`. In case of 3D, shapes are:
- input_a: [s, m, n], layout 3DS
- input_b: [s, n, p], layout 3DS
- input_c: [s, p], layout 2DS

When we lower 3D Matmul to NNPA, we need to prepare a zero-ed `input_c`, and **the wrong result came from wrong `input_c` where not all elements are zero-ed.**

This is our code to prepare a zero-ed `input_c`.
```mlir
#map_2ds = affine_map<(d0, d1) -> (d0, d1 floordiv 64, 0, 0, 31, d1 mod 64)>
%input_c = memref.alloc(%i0, %i1) {alignment = 4096 : i64} : memref<?x?xf16, #map_2ds>
krnl.memset %input_c, %cst_f0 : memref<?x?xf16, #map_2ds>
```

There are two situations to lower the above code:
1. lower it to affine, then normalize MemRef: we are doing this in onnx-mlir and it produces a wrong result.
2. normalize MemRef, then lower to affine: this leads to the correct result.

Situation 1 led to a wrong result because only the mapped part has zero values while the other part is untouch. We think this is natural to affine_map where only part of interest is visible. In particular, below is the generated code using `--convert-krnl-to-affine`:
```mlir
    %4 = memref.alloc(%1, %3) {alignment = 4096 : i64} : memref<?x?xf16, #map0>
    affine.for %arg2 = 0 to #map1(%1) {
      affine.for %arg3 = 0 to #map2(%1, %3) {
        affine.store %cst, %4[%arg2, %arg3] : memref<?x?xf16, #map0>
      }
    }
```
then `--normalize-memrefs`:
```mlir
    %5 = memref.alloc(%1, %4) {alignment = 4096 : i64} : memref<?x?x1x1x32x64xf16>
    %6 = memref.cast %5 : memref<?x?x1x1x32x64xf16> to memref<?x?x1x1x32x?xf16>
    affine.for %arg2 = 0 to %1 {
      affine.for %arg3 = 0 to %3 {
        affine.store %cst, %5[%arg2, %arg3 floordiv 64, 0, 0, 31, %arg3 mod 64] : memref<?x?x1x1x32x64xf16>
      }
    }
```
It is clear that not all elements in `memref<?x?x1x1x32x64xf16>` are set to zero.

Now, consider Situation 2, we do `--normalize-memrefs` first and get:
```mlir
    %7 = memref.alloc(%4, %5, %6) {alignment = 4096 : i64} : memref<?x?x1x1x32x?xf16>
    krnl.memset %7, %cst : memref<?x?x1x1x32x?xf16>
```

then `--convert-krnl-to-memref`:
```mlir
   %7 = memref.alloc(%4, %5, %6) {alignment = 4096 : i64} : memref<?x?x1x1x32x?xf16>
    affine.for %arg2 = 0 to #map3(%4) {
      affine.for %arg3 = 0 to #map4(%4, %5) {
        affine.for %arg4 = 0 to 1 {
          affine.for %arg5 = 0 to 1 {
            affine.for %arg6 = 0 to 32 {
              affine.for %arg7 = 0 to #map5(%4, %5, %6) {
                affine.store %cst, %7[%arg2, %arg3, %arg4, %arg5, %arg6, %arg7] : memref<?x?x1x1x32x?xf16>
              }
            }
          }
        }
      }
    }
```
All elements in memref<?x?x1x1x32x?xf16> are set to zero.

In summary:
- If we need to `memset` a zTensor, is it reasonable to use `affine_map` to allocate the zTensor or not? It seems it's not and we should use a raw shape instead, e.g. `memref<?x?x1x1x32x?xf16>`.
- The semantic of `krnl.memset` is currently reasonable or not, meaning it resets only the mapped part.
- Whether this is the issue in `--normalize-memrefs` in MLIR.

tungld:
@AlexandreEichenberger I implemented two modes for `krnl.memset` as your suggestion by introducing an additional attribute in `krnl.memset`. Please take a look. Thanks!

src/Dialect/Krnl/Krnl.td
```diff
@@ -1075,10 +1075,30 @@ def KrnlMemsetOp : Op<Krnl_Dialect, "memset", [MemRefsNormalizable,
                    "$_self.cast<MemRefType>().getElementType()">]> {
   let summary = "Set buffer to a given value.";
   let description = [{
-    Krnl operation that set buffer to a given value.
+    Krnl operation that sets a buffer to a given value.
```
AlexandreEichenberger, line 1078: great doc, thanks

AlexandreEichenberger (approved):
LGTM. thanks for the quick turnaround

PR #1623, opened 2022-08-22
eliminate KrnlDialectOperandParser

sorenlassen:
instead call parser.parse[Optional]Operand() and parser.resolveOperand() directly

avoids the need to work around failure to parse commas in some situations

also fixes these compiler warnings in KrnlOps.cpp:
```
warning: ignoring return value of function declared with 'nodiscard' attribute [-Wunused-result]
```


src/Dialect/Krnl/KrnlOps.cpp
```diff
@@ -327,99 +326,106 @@ ParseResult KrnlIterateOp::parse(OpAsmParser &parser, OperationState &result) {
...
-    // Bound is not an SSA id, then it must be an integer.
-    // Parse an integer constant attribute.
-    // Get the attribute location.
-    llvm::SMLoc attrLoc = parser.getCurrentLocation();
-    Attribute boundAttr;
-    NamedAttrList tempBoundAttrContainer;
-    if (parser.parseAttribute(
-            boundAttr, builder.getIndexType(), "temp", tempBoundAttrContainer))
-      return failure();
+  if (failed(parser.parseOptionalRParen())) {
```
AlexandreEichenberger, line 329: Do you mind explaining why we define and use the `parseBound` only when failing the `parseOptionalRParen`.
Also, because the code is long, do you mind just labeling the end of the lambda function with a comment, as well as the code that follows as the continuation of when `parseOptionalRParen` fails... thanks

sorenlassen: parseBound was only needed when the parenthesis were non-empty

I realize that the code was rather convoluted so I refactored parseBound and parseLoop into a struct LoopParser that I hope makes it easier to follow

AlexandreEichenberger (approved):
I know that the first time you just copied the previous pattern, this is so much more readable, big thanks

sorenlassen:
Thanks for the review. Please merge it for me.

PR #1625, opened 2022-08-22
add parallel support in conv

caoimhinuibrian:
Signed-off-by: Kevin O'Brien <12157865+caoimhinuibrian@users.noreply.github.com>
This is a new version of PR #1567 - closing that.


src/Conversion/ONNXToKrnl/NN/Conv.cpp
```diff
@@ -67,143 +73,164 @@ struct ONNXConvOpLowering : public ConversionPattern {
...
+    ValueRange steps(stepsStorage);
+    ValueRange parUbs(ubsStorage);
     // Iterate over the outer loops
     // for n = 0 .. N:
     //   for g = 0 .. G:
     //     for coPerGroup = 0 .. COPerGroup:
     //       co = g * COPerGroup + coPerGroup;
 
-    MemRefBuilder createMemRef(createKrnl);
+    // MemRefBuilder createMemRef(create);
```
AlexandreEichenberger, line 94: can be removed, as well as most of the commented out code below... unless you need to dematerialize later. But from the look of it, its not needed as you are using a lambda function and instantiate very nicely below for either parallel or sequential.

src/Dialect/Mlir/DialectBuilder.cpp
```diff
@@ -653,6 +654,20 @@ void SCFBuilder::ifThenElse(Value cond,
   }
 }
 
+void SCFBuilder::parallelLoop(ValueRange lowerBounds, ValueRange upperBounds,
```
AlexandreEichenberger, line 657: Out of curiosity, how does it do multiple parallel loops? Does it do an openmp collapse, or simply parallel of parallel ..... of parallel?

AlexandreEichenberger (approved):
LGTM after cleaning up a bit of the commented code. Code looks very clean and readable.

src/Dialect/Mlir/DialectBuilder.cpp
caoimhinuibrian: currently it just lowers to regular branching code

src/Dialect/Mlir/DialectBuilder.cpp
```diff
@@ -653,6 +654,20 @@ void SCFBuilder::ifThenElse(Value cond,
...
+    ValueRange steps,
+    function_ref<void(DialectBuilder &createKrnl, ValueRange)> bodyFn) const {
+  // SmallVectorImpl<Value> ivStorage;
+  b.create<scf::ParallelOp>(loc, lowerBounds, upperBounds, steps,
+      [&](OpBuilder &childBuilder, Location childLoc,
+          ValueRange inductionVars) {
+        KrnlBuilder builder(childBuilder, childLoc);
+        printf("%d\n", inductionVars.size());
```
sorenlassen, line 665: is this leftover "printf debugging" that should be removed? I happened to notice it because it causes the warning:
```
src/Dialect/Mlir/DialectBuilder.cpp:665:24: warning: format specifies type 'int' but the argument has type 'size_t' (aka 'unsigned long') [-Wformat]
        printf("%d\n", inductionVars.size());
```

sorenlassen: I'm removing it in PR #1644

AlexandreEichenberger: Thanks

PR #1626, opened 2022-08-23
Fix some warnings in onnx-mlir (mostly around third party code)

sstamenova:
There are multiple warnings coming from third party code even when we ask for them to be suppressed - this is partially due to the move to C++ 17. This change updates the code to suppress the warnings, so that all of the ones coming from third party code are now suppressed and also updates some of our tests and binaries to properly set the llvm flags to make sure that things build cleanly.

This does *not* fix all existing warnings, there will be others that need to be addressed separately.


AlexandreEichenberger (approved):
LGTM upon passing all the test. Will restart as AMD failure maybe a fluke

AlexandreEichenberger:

PR #1628, opened 2022-08-23
Update llvm to e5d5146323ffaa13eb5185616c6ae5c36b69352d

tungld:
This updates LLVM to a green commit:
```
Week of 8/22/2022:
Green LLVM commit: e5d5146323ffaa13eb5185616c6ae5c36b69352d
Green MHLO commit: ace4030dd55fce2a74e46f71f1937feb61ed1e3f (branch greencommit/2022-08-22-e5d51463)
```

This update is to include a fix for memref normalization where non-normalizable operations with identity map layouts blocked  normalization of the entire function: https://github.com/llvm/llvm-project/commit/183c4a391ef344220664d1d103d43639468bf103

### DIFF
--- a/docs/BuildOnLinuxOSX.md
+++ b/docs/BuildOnLinuxOSX.md
@@ -14,7 +14,7 @@ Firstly, install MLIR (as a part of LLVM-Project):
 ``` bash
 git clone -n https://github.com/llvm/llvm-project.git
 # Check out a specific branch that is known to work with ONNX-MLIR.
-cd llvm-project && git checkout 061e0189a3dab6b1831a80d489ff1b15ad93aafb && cd ..
+cd llvm-project && git checkout e5d5146323ffaa13eb5185616c6ae5c36b69352d && cd ..
 ```
 
 [same-as-file]: <> (utils/build-mlir.sh)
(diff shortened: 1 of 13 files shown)


AlexandreEichenberger (approved):
LGTM
If possible, in the future, would it be possible to keep the llvm related update clean of other unrelated onnx-mlir cleanup/changes, it would make reviewing easier. Tx

tungld:
Changes here were because of the changes in LLVM only, for example LLVM replaced `AbsOp` by `AbsFOp` and added `AbsIOp`, so we had to change code related to AbsOp to make onnx-mlir works. I am sorry for not explaining about AbsOp.

I looks like this LLVM commit id failed to compile on MacOs.

AlexandreEichenberger:
@gongsu832 mentioned during our last meeting that the failure on MacOS was due to a mismatch of python. Is the fix already in? If so, this might be a new failure.

tungld:
I guess the python mismatch issue was solved with this PR #1616. @sorenlassen memtioned that this LLVM commit `59548fe873d8d98e359fb21fbb2a0852fed17ff5` will fix the current MacOS issue, so I tried it with PR #1630, and it worked. It's better to move to `59548fe873d8d98e359fb21fbb2a0852fed17ff5`

PR #1631, opened 2022-08-24
[NNPA] Directly stick/unstick between NCHW and NHWC without transpose

tungld:
ZDNN stickify and unstickify can work directly with NCHW layout, it means we don't need to explicitly transpose from/to NCHW to/from NHWC before stickify/unstickify.

This patch changes ZHigh Stick and Unstick ops to **additionally** support NCHW directly, so that we don't need to generate a transpose op.

- ZHigh.Stick now has two attributes: `fromLayout` and `toLayout` to specify the source and target layouts for stickification. When `fromLayout` is none, everything works as before this patch.
- ZHigh.Unstick now has one attributes: `toLayout` to specify the target layout we want to unstikify to. `fromLayout` is unnecessary since it is specified in the tensor type of the input zTensor. When `toLayout` is none, everything works as before this patch.
- Existing rewriting rules that work with transpose are unchanged. Additionally, this patch adds the same rule but for direct stickify/unstickify.


AlexandreEichenberger:
@tungld this is a big patch, so do you mind explaining in a bit more details at a high level

>ZHigh.Stick now has two attributes: fromLayout and toLayout to specify the source and target layouts for stickification. When fromLayout is none, everything works as before this patch.

The default ONNX is NCHW... so if the `fromLayout` is none, then we still assume NCHW but we generate the transpose followed by the stick. I assume right now that there is a rule that does the "detect a conv, add a transpose, stick before, add unstick transpose after". So at a high level, how was this rule changed? Since we never have NHWC in native onnx, why did we add a to/from layout. Would be great to explain a bit more the overall design and how the solution helps optimization (which I am sure was your goal).

tungld:
> I assume right now that there is a rule that does the "detect a conv, add a transpose, stick before, add unstick transpose after".

Currently, an ONNXConv (X: tensor<1x2x3x4xf32>) will be lowered to :
```mlir
X_NHWC = transpose(X)  {to = NHWC} : tensor<1x2x3x4xf32> -> tensor<1x4x2x3xf32>
X_NHWC_S = stick (X_NHWC) {to = NHWC} : tensor<1x4x2x3xf32> -> tensor<1x4x2x3xf32, "NHWC">
Y_NHWC_S = zhigh.conv(X_NHWC_S) : tensor<1x4x2x3xf32, "NHWC"> -> tensor<1x4x2x3xf32, "NHWC">
Y_NHWC = unstick(Y_NHWC_S) : tensor<1x4x2x3xf32, "NHWC"> -> tensor<1x4x2x3xf32>
Y = transpose (Y_NHWC) {to = NCHW} : tensor<1x4x2x3xf32> -> tensor<1x2x3x4xf32>
```

It becomes the follows with this patch:
```mlir
X_NHWC_S = stick (X) {to = NHWC, from = NCHW} : tensor<1x2x3x4xf32> -> tensor<1x4x2x3xf32, "NHWC">
Y_NHWC_S = zhigh.conv(X_NHWC_S) : tensor<1x4x2x3xf32, "NHWC"> -> tensor<1x4x2x3xf32, "NHWC">
Y = unstick(Y_NHWC_S) {to = NCHW} : tensor<1x4x2x3xf32, "NHWC"> -> tensor<1x2x3x4xf32>
```

or with transpose (use `from = none`):
```mlir
X_NHWC = transpose(X)  {to = NHWC} : tensor<1x2x3x4xf32> -> tensor<1x4x2x3xf32>
X_NHWC_S = stick (X_NHWC, from = none) {to = NHWC} : tensor<1x4x2x3xf32> -> tensor<1x4x2x3xf32, "NHWC">
Y_NHWC_S = zhigh.conv(X_NHWC_S) : tensor<1x4x2x3xf32, "NHWC"> -> tensor<1x4x2x3xf32, "NHWC">
Y_NHWC = unstick(Y_NHWC_S) : tensor<1x4x2x3xf32, "NHWC"> -> tensor<1x4x2x3xf32>
Y = transpose (Y_NHWC) {to = none} : tensor<1x4x2x3xf32> -> tensor<1x2x3x4xf32>
```

So, my intention was to make it backward compatible so we distinguish two cases for NHWC: "stick from NCHW" and "stick from NHWC". But it seems that once we have *direct* stick from NCHW, there is no need for "stick from NHWC" (because our lowering does not transpose anything to NHWC). If it is the case, life is easier, we always stick to NHWC from NCHW.

tungld:
Replaced this patch by #1640

PR #1632, opened 2022-08-24
Matmul broadcast

AlexandreEichenberger:
Enables optimized (simdized/tiled for registers) matmul where A is broadcasted to B (rank(A)>2, rank(B)==2) or B is broadcasted to A (rank(B)>2, rank(A)==2), which will be very useful for 1x1 convolution (stride & dilation of 1, pad of 0).

On x86, a 4x broadcast results in about 4x the runtime of the no-broadcast optimized matmul.
```
BM_MatmulSquare/1024                    154 ms          154 ms            4 FLOP=2.14748G FLOPS=13.9771G/s
BM_MatmulSquareBroadcastB4x/1024        641 ms          641 ms            1 FLOP=8.58993G FLOPS=13.4108G/s
```
where the (null optimized) broadcast version is about 4x slower (due to the fact it computes 4x the regular matmul)

Without the optimization, we get this for the same 1024x1024 matmul
```
BM_MatmulSquareBroadcastB4x/1024       6916 ms         6898 ms            1 FLOP=8.58993G FLOPS=1.24528G/s
```
where the unoptimized (old) matmul is 40x slower (due to the fact it does 4x the work and is not optimized).

So on x86, the speedup of using the new optimized path for matmul with broadcast is about 10x.


src/Conversion/ONNXToKrnl/Math/MatMul.cpp
```diff
@@ -315,10 +398,22 @@ struct ONNXMatMulOpLowering : public ConversionPattern {
...
+    } else if (enableTiling && aRank == 2 && bRank > 2) {
+      // Broadcasting B.
+      assert(cRank == bRank && "expected IxK * *xKxJ = *xIxJ result");
+      replace2x2Matmul2dBroadcasting(matMulOp, operandAdaptor, elementType,
+          shapeHelper, /*broadcasting B*/ true, alloc, zero, rewriter, loc);
+    } else if (enableTiling && aRank > 2 && bRank == 2) {
+      // Broadcasting A.
+      assert(cRank == aRank && "expected IxK * *xKxJ = *xIxJ result");
+      replace2x2Matmul2dBroadcasting(matMulOp, operandAdaptor, elementType,
+          shapeHelper, /*broadcasting B*/ false, alloc, zero, rewriter, loc);
```
tungld, line 416: I think you can cover the case where `aRank == bRank && aRank > 2` with a small extension. It is not broadcasting but a batched matmul, e.g. `(B1xB2xIxK) * (B1xB2xKxJ)` and it does `(IxK) * (K*J)` for each batch index. Do you plan to do it in this patch also?

tungld (approved):
LGTM.

Thanks for the great work! It's a worth extension.

tungld:
@AlexandreEichenberger it looks like you committed the old `third_party/mlir_hlo` that caused Jenkins failed.

src/Conversion/ONNXToKrnl/Math/MatMul.cpp
AlexandreEichenberger: When the B1...Bn are the same values? 

Because I think broadcasting also supports the case where a 1 is matched with a Bx and then there is a broadcasting there. The case you are suggesting is indeed easier, if they are equally matched. Is that what you had in mind?

AlexandreEichenberger: If it is, let me do in a different PR, because I also need to upgrade the ModelLib to generate and test for this new case. It does seems worthwhile.

tungld: >  The case you are suggesting is indeed easier, if they are equally matched. Is that what you had in mind?

Yes, I am thinking about the "equally-matched" case at this moment.

AlexandreEichenberger: and no runtime, as there is no way to know if a `-1` will materialize as a `1` or matching '10' for example

Issue #1584, opened 2022-08-05
Usage of --shapeInformation results in different outcome depending on the input values


AlexandreEichenberger:
Looking at the bertwquad12.onnx model (https://github.com/onnx/models/tree/main/text/machine_comprehension/bert-squad/model)

When I do shape inference with the `?` dynamic values replace by 4, the final model is fully shaped with constant shapes

``` bash
> onnx-mlir -O3 -v -EmitONNXIR bertsquad-12.onnx --shapeInformation="0:4,1:4x256,2:4x256,3:4x256"
> grep "?" bertsquad-12.tmp |wc -l
0
```

But when the `?` values are replaced by 1, not all of the shapes are compile time shapes
```bash
>onnx-mlir -O3 -v -EmitONNXIR bertsquad-12.onnx --shapeInformation="0:1,1:1x256,2:1x256,3:1x256"
> grep "?" bertsquad-12.tmp |wc -l
9
```

To me that does not totally make sense. Would be nice to have someone investigate the discrepancy.

AlexandreEichenberger:
@negiyas any insight on what is happening?

negiyas:
Please look at the following outputs of the both cases.

** In the dynamic value== 1 case,** 

    %1110 = "onnx.Mul"(%1096, %1104) : (tensor<256x768xf32>, tensor<256x1xf32>) -> tensor<256x768xf32>
    %1111 = "onnx.Mul"(%1110, %1105) : (tensor<256x768xf32>, tensor<768xf32>) -> tensor<256x768xf32>
    %1112 = "onnx.Add"(%1111, %1109) {onnx_node_name = "bert/encoder/layer_11/output/LayerNorm/batchnorm/add_1"} : (tensor<256x768xf32>, tensor<256x768xf32>) -> tensor<256x768xf32>
    %1113 = "onnx.Constant"() {value = dense<[256, 768]> : tensor<2xi64>} : () -> tensor<2xi64>
    %1114 = "onnx.Constant"() {value = dense<[1, 256, 2]> : tensor<3xi64>} : () -> tensor<3xi64>
    %1115 = "onnx.Reshape"(%1112, %1113) {allowzero = 0 : si64} : (tensor<256x768xf32>, tensor<2xi64>) -> tensor<?x?xf32>
    %1116 = "onnx.Constant"() {value = dense<[1.25070888E-4, 1.26759609E-4]> : tensor<2xf32>} : () -> tensor<2xf32>
[...]

PR #1633, opened 2022-08-25
Change the default value for the "--onnx-op-transform-threshold" from 3 to 10

negiyas:
Change the default value for the "--onnx-op-transform-threshold" option from 3 to 10 to fix the issue 1584 ( https://github.com/onnx/onnx-mlir/issues/1584 )

It should not be an big issue to change the default value from 3 to 10, because of the following reasons.

(1) At least 4 times repeat of the corresponding passes are necessary to fix this issue.
(2) If the target code is not changed by an iteration, further iterations are stopped at the iteration. 

Confirmed that this issue is solved by the PR.


negiyas:
Confirmed that the issue 1584 is solved by theis PR.

AlexandreEichenberger (approved):
LGTM. Could you just run the resnet50 from the check-onnx-backend and record the time with 4 and 10 in this PR, so that we have a reference of if it slows down at all? That would be great.

Thanks for finding the issue and a quick fix to it, much appreciated

negiyas:
I measured  time to execute "TEST_CASE_BY_USER=test_resnet50_cpu cmake --build . --target check-onnx-backend" by changing the default value from 4 to 10.
We cannot see any slowdown  according to the measurement results.

4: Ran 2026 tests in 156.981s
6: Ran 2026 tests in 157.096s
8: Ran 2026 tests in 156.717s
10: Ran 2026 tests in 157.068s

AlexandreEichenberger:
Fantastic, big thanks

PR #1634, opened 2022-08-25
Allows for model input names to be the same as ONNX node inputs

kernhanda:
Currently, if you try to compile an ONNX model where the input name is the same as the input of a node, you hit the following error (with assertions enabled):

```
onnx-mlir: /code/onnx-mlir/src/Builder/SymbolTable.hpp:129: void onnx_mlir::SymbolMapping<T>::AddMapping(const string&, T) [with T = onnx::TypeProto; std::string = std::__cxx11::basic_string<char>]: Assertion `!scopes.back().contain(name) && "Tensor already exists."' failed.
[1]    1059 IOT instruction  ./onnx-mlir -v --EmitMLIR
```

It is legal to have an ONNX model with this configuration. Attached is one such model. [test_model.zip](https://github.com/onnx/onnx-mlir/files/9421795/test_model.zip)

This PR fixes this bug.


kernhanda:
If this fix is too heavy handed, it could also be fixed by checking that the tensor that already exists matches the tensor being added again.

AlexandreEichenberger:
If you don't mind, I am just going to update it to the latest and that will also start all of the CIs. Let me know if you prefer I don't in the future. Just do a git pull before doing further changes to your PR.

AlexandreEichenberger:
@chentong319 this is a frontend issue, do you mind looking at the implication of it, tx

kernhanda:
ping :)

chentong319:
In the document of [onnx IR](https://github.com/onnx/onnx/blob/main/docs/IR.md), the value_info for graph is `Used to store the type and shape information of values that are not inputs or outputs.`
This model include the input and output tensor in the value_info. Is the model correct or the specification is relaxed?

kernhanda:
> In the document of [onnx IR](https://github.com/onnx/onnx/blob/main/docs/IR.md), the value_info for graph is `Used to store the type and shape information of values that are not inputs or outputs.` This model include the input and output tensor in the value_info. Is the model correct or the specification is relaxed?

TBH, I'm not sure. It's possible that the IR is more well-defined and restrictive than the model specification. The model provided can be loaded just fine by the onnx library and even passes `onnx.checker.check_model`. Netron can also open the model fine.

chentong319:
Other tools may forget to check this condition. For this PR, it is OK to me if you add a comment to the change. Basically, onnx-mlir assumes the model is correct and is not eager to check the correctness of the model.

kernhanda:
> Other tools may forget to check this condition. For this PR, it is OK to me if you add a comment to the change. Basically, onnx-mlir assumes the model is correct and is not eager to check the correctness of the model.

Done! Please merge in the PR when you get a chance. TIA!

AlexandreEichenberger:
updated to the latest, which also restart all of the cis

AlexandreEichenberger:
@chentong319 please approve or provide more suggestions on this PR

chentong319 (approved):
LGTM!

PR #1643, opened 2022-08-26
Matmul broadcast when the broadcast dims of A & B are identical

AlexandreEichenberger:
Extension suggested by @tungld to support more broadcast patterns for matmul, namely when both A &B have the same identical broadcast dimensions (cannot handle dynamic dimension as we cannot rule out that one would be `1` and the other `100`, for example, thus needing an actual broadcast pattern.

This is an example of what this PR adds, where both broadcast dim of arg0 and arg1 (for A and B) are `2x3`.

```mlir
func.func private @test_matmul3(%arg0 : tensor<2x3x10x5xf32>, %arg1 : tensor<2x3x5x10xf32>) -> tensor<*xf32> {
  %0 ="onnx.MatMul"(%arg0, %arg1) : (tensor<2x3x10x5xf32>, tensor<2x3x5x10xf32>) -> tensor<*xf32>
  "func.return"(%0) : (tensor<*xf32>) -> ()
// mlir2FileCheck.py -a'["A", "B"]' -n'{"0": "RES"}'
// CHECK-LABEL:  func.func private @test_matmul3
// CHECK-SAME:   ([[A_:%.+]]: memref<2x3x10x5xf32>, [[B_:%.+]]: memref<2x3x5x10xf32>) -> memref<2x3x10x10xf32> {
// CHECK-DAG:       [[VAR_c0_:%.+]] = arith.constant 0 : index
// CHECK-DAG:       [[VAR_cst_:%.+]] = arith.constant 0.000000e+00 : f32
// CHECK-DAG:       [[VAR_c2_:%.+]] = arith.constant 2 : index
// CHECK-DAG:       [[VAR_c3_:%.+]] = arith.constant 3 : index
// CHECK-DAG:       [[VAR_c10_:%.+]] = arith.constant 10 : index
// CHECK-DAG:       [[VAR_c5_:%.+]] = arith.constant 5 : index
// CHECK-DAG:       [[RES_:%.+]] = memref.alloc() {{.*}}: memref<2x3x10x10xf32>
// CHECK:           krnl.memset [[RES_]], [[VAR_cst_]] {delayed = false} : memref<2x3x10x10xf32>
// CHECK:           [[LOOP_0_:%.+]]:2 = krnl.define_loops 2

// broadcast iterate

[...]

PR #1644, opened 2022-08-27
fix/silence more compiler warnings

sorenlassen:
silenced the rapidcheck deprecation warnings for gcc, clang

also removed a leftover printf which caused a compiler warning about %d not matching type size_t

also fixed a reorder-ctor warning


test/numerical/Common.hpp
```diff
@@ -1,9 +1,12 @@
...
-#include <rapidcheck.h>
-#undef _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS
-
-#include "test/modellib/ModelLib.hpp"
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated"   // gcc, clang
```
sorenlassen, line 6: I only tested this with clang on Mac but according to stackoverflow
https://stackoverflow.com/a/11853015
`#pragma GCC` should work for both gcc and clang

test/numerical/Common.hpp
```diff
@@ -1,9 +1,12 @@
...
+#define _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS // MSVC
```
sorenlassen, line 7: @sstamenova I believe this fix from PR #1626 is MSVC specific, it didn't work for me with clang on Mac

AlexandreEichenberger: So the `#pragma GCC diagnostic ignored "-Wdeprecated"` is the equivalent of  this flag?
And does this works on MSVC? if so, let's have as little redundancy as possible, tx

sstamenova: Since MSVC != GCC, the GCC diagnostics don't impact the MSVC builds. I suggest using an `ifdef` or similar to check which compiler it is and then include the appropriate suppression. For example: `#if defined(__GNUC__)`, `if defined(__clang__)`, `if defined(_MSC_VER)`, etc.

You can always look at the Windows CI build to see whether there are any warnings/errors.

sorenlassen: thanks for the suggestion, I now choose between MSVC and gcc/clang with `#if defined(_MSC_VER)`

I verified that `#pragma GCC` works for clang, which is also explained here: https://stackoverflow.com/a/11853015

src/Dialect/Mlir/DialectBuilder.cpp
```diff
@@ -662,7 +662,6 @@ void SCFBuilder::parallelLoop(ValueRange lowerBounds, ValueRange upperBounds,
       [&](OpBuilder &childBuilder, Location childLoc,
           ValueRange inductionVars) {
         KrnlBuilder builder(childBuilder, childLoc);
-        printf("%d\n", inductionVars.size());
```
sstamenova, line 665: Nice catch

sstamenova (approved):

sorenlassen:
thanks for the review, please merge it for me

sorenlassen:
@AlexandreEichenberger  PTAL because I added a fix for a reorder-ctor warning (from PR #1643)

I hope you don't mind I updated the PR after review

AlexandreEichenberger:
> I hope you don't mind I updated the PR after review

of course not, any feedback is useful. I saw the same issue and fixed it in a follow on, if your PR goes first, that is absolutely fine. My mistake for not looking at the warning to begin with.

AlexandreEichenberger:
@sorenlassen Updating to the latest. Let me know if you want me to merge it to main

PR #1646, opened 2022-08-27
fix lit test zlow-normalize-by-using-dummyop.mlir

### LINKED ISSUE #1645: lit test zlow-normalize-by-using-dummyop.mlir fails
when I run the NNPA lit tests as described [here](https://github.com/onnx/onnx-mlir/blob/main/docs/AccelNNPAHowToUseAndTest.md) `zlow-normalize-by-using-dummyop.mlir` fails
```
.../zlow-normalize-by-using-dummyop.mlir:7:12: error: FAILED: expected string not found in input
// FAILED: "multiple dereferencing uses in a single op not supported"
<stdin>:4:92: note: scanning from here
0. Program arguments: .../onnx-mlir-opt --maccel=NNPA --normalize-memrefs .../zlow-normalize-by-using-dummyop.mlir
```

the lit test RUN command begins with
```
build/Debug/bin/onnx-mlir-opt --maccel=NNPA --normalize-memrefs test/mlir/accelerators/nnpa/transform/zlow-normalize-by-using-dummyop.mlir 2>&1
```
and the output of that is
```
Assertion failed: (false && "multiple dereferencing uses in a single op not supported"), function replaceAllMemRefUsesWith, file Utils.cpp, line 1126.
Stack dump:
0.	Program arguments: build/Debug/bin/onnx-mlir-opt --maccel=NNPA --normalize-memrefs test/mlir/accelerators/nnpa/transform/zlow-normalize-by-using-dummyop.mlir
...
```
but the lit test is looking for
```
// FAILED-LABEL: onnx-mlir-opt 
// FAILED: "multiple dereferencing uses in a single op not supported"
```
so the problem is that the first occurrence of the `onnx-mlir-opt` appears after the "multiple dereferencing..." string

the failure may be platform specific: I use clang on Mac

sorenlassen:
the output starts with the failed assert message so
there is no label we can use with `FAILED-LABEL:`

closes #1645

### DIFF
--- a/test/mlir/accelerators/nnpa/transform/zlow-normalize-by-using-dummyop.mlir
+++ b/test/mlir/accelerators/nnpa/transform/zlow-normalize-by-using-dummyop.mlir
@@ -3,7 +3,6 @@
 // COM: Current MLIR normalize-memres does not support multiple dereferencing uses
 // in a single op, check expected failure emitted by MLIR. 
 
-// FAILED-LABEL: onnx-mlir-opt 
 // FAILED: "multiple dereferencing uses in a single op not supported"
 
 // RUN: onnx-mlir-opt --maccel=NNPA --zlow-dummyop-for-multideref --normalize-memrefs --canonicalize %s | FileCheck --check-prefix=PASSED %s


tungld (approved):
LGTM. Thanks!

PR #1648, opened 2022-08-29
Fix a bug in constant prop for SliceOp

### LINKED ISSUE #1610: Crash on SliceOp during ConstProp
I ran into a problem when I was lowering an ONNX model with the onnx.SliceOp. The end input for the SliceOp is from a constant operator, but the op only contains buffer_id as the attribute instead of the value attr. When calling `allocateBufferFor(replacingValue.getType(), /*useMaxSize=*/true)`, the shape of the type still contains a dynamic dimension (-1). Therefore, it crashes on `llvm-project/mlir/lib/IR/BuiltinTypeInterfaces.cpp:33: static int64_t mlir::ShapedType::getNumElements(llvm::ArrayRef<long int>): Assertion 'num >= 0 && "integer overflow in element count computation"' failed.`

The backtrace:
```
#4  0x00005618b8c607fc in mlir::ShapedType::getNumElements(llvm::ArrayRef<long>) ()
#5  0x00005618b825ff20 in onnx_mlir::getNumberOfElements (ty=...) at ../third_party/onnx-mlir/src/Support/TypeUtilities.cpp:48
#6  0x00005618b7c01434 in getMaxSizeInBytes (ty=...) at ../third_party/onnx-mlir/src/Transform/ONNX/ConstPropHelper.cpp:29
#7  0x00005618b7c01775 in allocateBufferFor (type=..., useMaxSize=true) at ../third_party/onnx-mlir/src/Transform/ONNX/ConstPropHelper.cpp:73
#8  0x00005618b7bc21d2 in (anonymous namespace)::ConstPropSlice (rewriter=..., replacingValue=..., constValue=...) at ../third_party/onnx-mlir/src/Transform/ONNX/ConstProp.cpp:822
#9  0x00005618b7bd0ea7 in (anonymous namespace)::SliceofConst::matchAndRewrite (this=0x5618d2faf2d0, op0=0x5618c76027a0, rewriter=...) at third_party/onnx-mlir/src/Transform/ONNX/ONNXConstProp.inc:1814
#10 0x00005618b8973f0d in mlir::PatternApplicator::matchAndRewrite(mlir::Operation*, mlir::PatternRewriter&, llvm::function_ref<bool (mlir::Pattern const&)>, llvm::function_ref<void (mlir::Pattern const&)>, llvm::function_ref<mlir::LogicalResult (mlir::Pattern const&)>) ()
#11 0x00005618b8953914 in mlir::applyPatternsAndFoldGreedily(llvm::MutableArrayRef<mlir::Region>, mlir::FrozenRewritePatternSet const&, mlir::GreedyRewriteConfig) ()
#12 0x00005618b7bfe777 in mlir::applyPatternsAndFoldGreedily (op=0x5618bdbf3c90, patterns=..., config=...)
    at ../third_party/llvm_build_rtti/include/mlir/Transforms/GreedyPatternRewriteDriver.h:71
#13 0x00005618b7bd53b4 in (anonymous namespace)::ConstPropONNXToONNXPass::runOnOperation (this=0x5618d7e1b2e0) at ../third_party/onnx-mlir/src/Transform/ONNX/ConstProp.cpp:1080
#14 0x00005618b8a36683 in mlir::detail::OpToOpPassAdaptor::run(mlir::Pass*, mlir::Operation*, mlir::AnalysisManager, bool, unsigned int) ()
#15 0x00005618b8a36ae0 in mlir::detail::OpToOpPassAdaptor::runPipeline(mlir::OpPassManager&, mlir::Operation*, mlir::AnalysisManager, bool, unsigned int, mlir::PassInstrumentor*, mlir::PassInstrumentation::PipelineParentInfo const*) ()
#16 0x00005618b8a35cee in mlir::detail::OpToOpPassAdaptor::runOnOperationAsyncImpl(bool) ()
#17 0x00005618b8a3672d in mlir::detail::OpToOpPassAdaptor::run(mlir::Pass*, mlir::Operation*, mlir::AnalysisManager, bool, unsigned int) ()
#18 0x00005618b8a36ae0 in mlir::detail::OpToOpPassAdaptor::runPipeline(mlir::OpPassManager&, mlir::Operation*, mlir::AnalysisManager, bool, unsigned int, mlir::PassInstrumentor*, mlir::PassInstrumentation::PipelineParentInfo const*) ()
#19 0x00005618b8a37568 in mlir::LogicalResult llvm::function_ref<mlir::LogicalResult (mlir::OpPassManager&, mlir::Operation*)>::callback_fn<mlir::detail::OpToOpPassAdaptor::run(mlir::Pass*, mlir::Operation*, mlir::AnalysisManager, bool, unsigned int)::{lambda(mlir::OpPassManager&, mlir::Operation*)#1}>(long, mlir::OpPassManager&, mlir::Operation*) ()
#20 0x00005618b7b71024 in llvm::function_ref<mlir::LogicalResult (mlir::OpPassManager&, mlir::Operation*)>::operator()(mlir::OpPassManager&, mlir::Operation*) const (
    this=0x5618bd7f9ab0, params#0=..., params#1=0x5618bd8212a0) at ../third_party/llvm/build/include/llvm/ADT/STLFunctionalExtras.h:68
#21 0x00005618b7b70f13 in mlir::Pass::runPipeline (this=0x5618bd7f9a40, pipeline=..., op=0x5618bd8212a0) at ../third_party/llvm/build/include/mlir/Pass/Pass.h:195
#22 0x00005618b7b6fe1c in (anonymous namespace)::ONNXOpTransformPass::runOnOperation (this=0x5618bd7f9a40) at ../third_party/onnx-mlir/src/Transform/ONNX/ONNXOpTransformPass.cpp:143
#23 0x00005618b8a36683 in mlir::detail::OpToOpPassAdaptor::run(mlir::Pass*, mlir::Operation*, mlir::AnalysisManager, bool, unsigned int) ()
#24 0x00005618b8a36ae0 in mlir::detail::OpToOpPassAdaptor::runPipeline(mlir::OpPassManager&, mlir::Operation*, mlir::AnalysisManager, bool, unsigned int, mlir::PassInstrumentor*, mlir::PassInstrumentation::PipelineParentInfo const*) ()
#25 0x00005618b8a37091 in mlir::PassManager::run(mlir::Operation*) ()
```

tungld:
Resolves #1610 

When doing constant propagation for SliceOp, shape inference is called and it computes IndexExprs for starts/ends/steps/axes. If starts/ends/steps/axes are constants generated **during**  the constant propagation, they are linked with an internal raw buffer and shape inference fails to read their values from these buffer-linked constants.

This patch fixes the issue by requiring that starts/ends/steps/axes must be a true ONNX constant that not generated **during** the constant propagation. After one call of the constant propagation finishes, buffer-linked constants will become true ONNX constants, then another call of the constant propagation will work as normal.


chentong319:
To my understanding, the added parameter is to mark whether this function is called inside constant propagation pass or not.  It may be less error prone that the constant propagation pass uses utility functions with a different name defined just its .td file with explicit name, not a parameter.

tungld:
>It may be less error prone that the constant propagation pass uses utility functions with a different name defined just its .td file with explicit name, not a parameter.

@chentong319 all functions here are used inside the constant propagation only. Did you suggest that using two C++ functions instead of a single function `isFromDenseONNXConstantOp` with a parameter?

sorenlassen:
the code changes LGTM

 IIUC the idea is that we can place extra restrictions on the SliceOfConst constant propagation pattern and let it fail if its constant inputs are buffer-linked (similarly, we could restrict the pattern to only apply if the output shape has been inferred, which is the next thing that would fail if we converted the buffer-links to dense elements attributes on the fly) because constant propagation and shape inference are run multiple times and these passes can make progress in the next iteration of `ONNXOpTransformPass::runOnOperation()` when constants have been converted to dense elements attributes and shape inference has been run after the constants have propagated to the inputs
 
 (I commented on the problem in issue #1610 but couldn't see a good way to solve it because I thought it was necessary to force the pattern to succeed with the buffer-linked constants and incomplete shape inference it was given)

tungld:
> IIUC the idea is that we can place extra restrictions on the SliceOfConst constant propagation pattern and let it fail if its constant inputs are buffer-linked (similarly, we could restrict the pattern to only apply if the output shape has been inferred, which is the next thing that would fail if we converted the buffer-links to dense elements attributes on the fly) because constant propagation and shape inference are run multiple times and these passes can make progress in the next iteration of ONNXOpTransformPass::runOnOperation() when constants have been converted to dense elements attributes and shape inference has been run after the constants have propagated to the inputs

Yes, exactly.

> we could restrict the pattern to only apply if the output shape has been inferred, which is the next thing that would fail if we converted the buffer-links to dense elements attributes on the fly

Sure, we can impose a restriction on the output type also if its shape is not static during the current constant propagation. Shape inference will infer a static shape, and the next constant propagation will work.

PR #1654, opened 2022-08-30
Only update the output shape if the infered shape is better - unary elementwise ops

### LINKED ISSUE #1480: Incorrect output_signature() for models with dynamic dimensions
# Description
Output signature from dynamic models has changed. As a specific example, the model from https://github.ibm.com/ai-on-z/in-house-onnx-models/blob/main/TF-ONNX/ccf_gru_dyn_tf2onnx_new.onnx as of 09965003137f82d8891676c986ec6403faa9c3cb 
 used to return: (note `1` for the final dim)
```
output signature: [   { "type" : "f32" , "dims" : [-1 , -1 , 1] , "name" : "dense" }

]
```

But as of 91ee7a2ed38e91cf480b06d0361899d2d63ce5ca returns: note `-1` for the final dim)
```
output signature: [   { "type" : "f32" , "dims" : [-1 , -1 , -1] , "name" : "dense" }

]
```
Running the model through netron.app shows the final dimesion in the .onnx file is indeed supposed to be `1`
image

# Reproduction
Model was compiled using flags
```
--O3 --EmitLib --mtriple=s390x-ibm-loz --mcpu=z16 --maccel=NNPA ccf_lstm_dyn_tf2onnx_new.onnx
```

sig_client.py code:
```
import os
import sys

model_lib = sys.argv[1]
if not os.path.exists(model_lib):
    raise FileNotFoundError('model lib "' + model_lib + '" was not found.')
session = ExecutionSession(model_lib)
print("input signature: " + session.input_signature())
print("output signature: " + session.output_signature())
```
which was called using python 3.8:
```
 PYTHONPATH=/code/build/lib python3 sig_client.py NNPA/ccf_lstm_dyn_tf2onnx_new.so
```

tungld:
Currently our shape inference always overwrites the output shape with a new shape. Sometimes, shape inference has not been powerful enough to infer static dimensions.

Meanwhile, in many onnx models, outputs of ONNX ops have static shape or partially static shape. In such cases, we should respect the existing shape and only update the output shape if the inferred shape is better, for example, the inferred shape can infer a static dimension.

This patch adds a function to check whether the inferred shape is better than the existing shape or not, and it applies the function to the shape inference of unary element-wise ops as the first candidates.

Resolves #1480 


AlexandreEichenberger (approved):
LGTM; is there a need to also check for ops beyond unary element-wise ops?

PR #1656, opened 2022-08-30
Fix namespace leak

philass:
Signed-off-by: Philip Lassen <philiplassen+git@gmail.com>


AlexandreEichenberger (approved):
LGTM, you are right, we don't like `using namespace` in header files. Good catch.

PR #1657, opened 2022-08-31
Fix additional warnings on Windows

sstamenova:
Technically, some of them would apply to other platforms as well. This brings down the warnings on Windows to just a few left.


sstamenova:
@sorenlassen : FYI

src/Transform/ONNX/ShapeInferencePass.cpp
```diff
@@ -97,8 +97,9 @@ class ShapeInferencePass
   }
 
   static LogicalResult runShapeInferenceOnRegion(Region &r) {
-    std::function<void(Region &)> doShapeInference =
-        &ShapeInferencePass::runShapeInferenceOnRegion;
+    std::function<void(Region &)> doShapeInference = [&](Region &region) {
```
sorenlassen, line 100: nit: can you omit the `&` in `[&]`? it doesn't seem like anything is captured

test/numerical/Common.hpp
```diff
@@ -2,14 +2,14 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-#if defined(__MSC_VER)
+#if defined(_MSC_VER)
```
sorenlassen, line 5: thanks for fixing this!

sorenlassen (approved):
LGTM

src/Transform/ONNX/ShapeInferencePass.cpp
AlexandreEichenberger: if applicable, make sense to remove the captured context

AlexandreEichenberger (approved):
LGTM after last comment from @sorenlassen is acted upon.

PR #1658, opened 2022-08-31
Fixing ONNXPad building logic when constant value is absent

rdicecco:
This change does two things:
1. Fixes the default behaviour of ONNXPadOp when the constant_value isn't provided such that the pad value is created with the appropriate element type (prior to this change it was unconditionally an FP32 DenseElementsAttr)
2. Adds a verifier to validate that the constant_value type matches the input type.

Signed-off-by: Roberto DiCecco <dicecco.rob@gmail.com>


AlexandreEichenberger:
Please look if this upgrade enable us to run more of the backend tests for PAD... Can help if you are not sure where to look.

AlexandreEichenberger:
Absolutely. Relevant help is here
https://github.com/onnx/onnx-mlir/blob/main/docs/Testing.md

Important paragraph:

> All the test cases provided by onnx package are listed in file test/backend/all_test_names.txt. check-onnx-backend will selectively run some of them. The node and model tests in onnx that will be run by check-onnx-backend is defined by variable test_to_enable in test/backend/test.py. 

So look into `test/backend/all_test_names.txt` for all the avail test. Locate the one for Pad. See if by any chance there is one we don't run yet by looking at the one we run here: `test/backend/inference_backend.py`. You can probably guess how new tests are added by copying a line adding an existing test and modifying the name for the one you are trying to add.

good luck and thanks

AlexandreEichenberger:
Due to our lack of doc, added some in a new PR, so look there too for info #1661

rdicecco:
@AlexandreEichenberger looks like all of the pad tests are currently enabled - it'd probably be beneficial to add a new test for this problem, though it looks like the tests are an external dependence (on the onnx repo), is the procedure to PR to onnx first and then post update to enable the test?

AlexandreEichenberger:
Thanks for checking. First line is to enable the backend, and for some complicated ops, we generate numerical tests (see test/numerical and test/modelLib, but that is pretty involved. So its probably good enough as is.

AlexandreEichenberger:
@chentong319 do you mind reviewing this PR, there are changes to the FrontEnd. Thanks

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -903,17 +909,6 @@ class FrontendGenImpl {
...
-      for (const auto &item : node.input()) {
-        if (const onnx::TensorProto *tensorPtr =
-                initializedTensors.GetByOnnxName(item)) {
-          inputs.push_back(EmitInitializerForInputTensor(
-              UnknownLoc(), builder_, *tensorPtr));
-        } else if (const Value *valuePtr =
-                       frontend_symbols_.GetByOnnxName(item)) {
-          inputs.push_back(*valuePtr);
-        }
-      }
```
chentong319, line 916: Good to replace this code with getNodeInput().

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -891,9 +891,15 @@ class FrontendGenImpl {
...
-      auto elementType = builder_.getF32Type();
+
+      std::vector<Value> inputs;
+      getNodeInputs(node, inputs);
+      auto elementType =
+          inputs[0].getType().cast<TensorType>().getElementType();
+
+      llvm::SmallVector<Attribute, 2> values(
+          1, builder_.getZeroAttr(elementType));
+
```
chentong319, line 902: The original code tried to use tensor<0xT> to represent none. Do you want to use a different way to create the none?

rdicecco: I haven't changed the underlying tensor type aside from the element type, in the original code it would be tensor<1xf32> (with value of 0.0), after the change it becomes tensor<1xT> (with value of (T)(0.0)).

I think there could maybe be an argument for: disallowing the NoneType all together for the input (since effectively it's being handle here), or allowing it and maybe breaking out the handling from this portion, thoughts?

chentong319: You are right that this optional input has a default value.

chentong319 (approved):
LGTM!

AlexandreEichenberger:
Thanks @chentong319 and @rdicecco

rdicecco:
No problem, happy to help :)

PR #1659, opened 2022-08-31
[NNPA] - Only update the output shape if the infered shape is better

tungld:
This patch continues PR #1654 and updates shape inference for ZHigh element-wise ops.

Functions `inferShapeForUnaryElementwiseOps` and `isInferredShapeBetter` are moved from `ONNXOps.cpp` to `ONNXShapeHelper.cpp` so that ZHigh can call them.

### DIFF
--- a/src/Dialect/ONNX/ShapeInference/ONNXShapeHelper.cpp
+++ b/src/Dialect/ONNX/ShapeInference/ONNXShapeHelper.cpp
@@ -399,6 +399,47 @@ LogicalResult ONNXGenericPoolShapeHelper<OP_TYPE, OP_ADAPTOR>::computeShape(
   return success();
 }
 
+/// Check if the given inferredShape is better than the existing shape of val.
+bool isInferredShapeBetter(ArrayRef<int64_t> inferredShape, Value val) {
+  // Val is unranked. Any inferred shape is better.
+  if (!hasShapeAndRank(val))
+    return true;
+
+  ArrayRef<int64_t> existingShape = val.getType().cast<ShapedType>().getShape();
+
+  // Only get the infered shape if unknown dims in existingShape get static
+  // values in inferredShape.
+  for (unsigned i = 0; i < existingShape.size(); ++i) {
+    int64_t inferedDim = inferredShape[i];
+    int64_t existingDim = existingShape[i];
+    // existingDim is static but inferedDim is unknown: not good.
+    if ((existingDim != -1) && (inferedDim == -1))
+      return false;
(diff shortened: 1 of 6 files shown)


AlexandreEichenberger (approved):
LGTM, can you explain why we focus only on unary elementwise ops? Would it not be needed for all?

tungld:
> LGTM, can you explain why we focus only on unary elementwise ops? Would it not be needed for all?

We should do this for all ops. And we start with unary elementwise ops because, e.g. sigmod, softmax are often the last ops in a model that decide the output signature.

Also, if all ops in the user's onnx model has shape (e.g. via calling ONNX shape inference) we also benefit from that.

In case of NNPA, I think the last op should be unstick, but I still support other ops just in case shape becomes dynamic by some rewriting rules.

src/Dialect/ONNX/ShapeInference/ONNXShapeHelper.cpp
```diff
@@ -399,6 +399,47 @@ LogicalResult ONNXGenericPoolShapeHelper<OP_TYPE, OP_ADAPTOR>::computeShape(
...
+/// Handle shape inference for unary element-wise operators.
+LogicalResult inferShapeForUnaryElementwiseOps(Operation *op) {
+  if (!hasShapeAndRank(op->getOperand(0)))
+    return success();
+
+  // When the output has ranked, see if its shape is better than the input's
+  // shape.
+  ArrayRef<int64_t> inputShape =
+      op->getOperand(0).getType().cast<ShapedType>().getShape();
+  if (!isInferredShapeBetter(inputShape, op->getResult(0)))
```
chentong319, line 435: Should we merge the specified type and the inferred type? For example, tensor<1x-1xT> and tensor<?x2xT> should be tensor<1x2xT>.

tungld: Yes, absolutely we can do that. Let me do it with another PR. BTW, do you have any suggestion about applying this shape mixing in a systematic way, e.g. no need to manually call it in each inferShape since developers may forget to use it.

chentong319: Replace setType `op->getResults()[0].setType(yTy);` with another function, say enhanceType  in MLIR. That may not be accepted by llvm. We can define a utility function in our own TypeUtility. There is a PR for IfOp, which contains some type merging. The two are different in semantic, but they can be viewed as type operations. One is `and` op on types, and the other is `or` op on types.

PR #1661, opened 2022-08-31
[DOC only] Add explanation on how to add a backend test.

AlexandreEichenberger:
Added a bit more info on how to add a new backend test using the new list that a recent PR added.


chentong319 (approved):
LGTM!

PR #1663, opened 2022-09-01
Adds option to specify graph entry point name

kernhanda:
This changes the behavior when a single entry point is specified. Prior to this change, the entry point would be ignored if there was only a single entry point and `run_main_graph` would be emitted. After this change, the entry point name is honored, even if there's only one entry point.

### DIFF
--- a/src/Conversion/KrnlToLLVM/KrnlEntryPoint.cpp
+++ b/src/Conversion/KrnlToLLVM/KrnlEntryPoint.cpp
@@ -89,8 +89,6 @@ public:
             .getLeafReference()
             .getValue();
 
-    // When there is only a single entry point function in a model, use
-    // DEFAULT_DYN_ENTRY_POINT.
     std::string dynEntryPointName = "run_" + staticEntryPointFuncName.str();
 
     // Record entry point name, input and output signatures in order to emit
--- a/src/Conversion/KrnlToLLVM/KrnlEntryPoint.cpp
+++ b/src/Conversion/KrnlToLLVM/KrnlEntryPoint.cpp
@@ -92,8 +92,6 @@ public:
     // When there is only a single entry point function in a model, use
     // DEFAULT_DYN_ENTRY_POINT.
     std::string dynEntryPointName = "run_" + staticEntryPointFuncName.str();
-    if (singleEntryPoint)
-      dynEntryPointName = DEFAULT_DYN_ENTRY_POINT;
 
     // Record entry point name, input and output signatures in order to emit
     // signature-related functions later.
--- a/test/mlir/krnl/entry_point.mlir
+++ b/test/mlir/krnl/entry_point.mlir
@@ -1,7 +1,5 @@
 // RUN: onnx-mlir-opt --convert-krnl-to-llvm --canonicalize %s -split-input-file | FileCheck %s
 
-// COM: Generate the default entry point "run_main_graph" since there is only
-// COM: one single point.
 module {
   func.func private @first_entry(%arg0: memref<10xf32>) -> memref<10xf32> {
     return %arg0 : memref<10xf32>
@@ -13,7 +11,7 @@ module {
 // CHECK:         llvm.mlir.global external constant @_entry_point_0_in_sig("[in_sig]\00")
 // CHECK:         llvm.mlir.global external constant @_entry_point_0_out_sig("[out_sig]\00")
 
-// CHECK-LABEL:   llvm.func @run_main_graph
+// CHECK-LABEL:   llvm.func @run_first_entry
 // CHECK:             ([[ARG0:%.+]]: !llvm.ptr<i8>) -> !llvm.ptr<i8>
 // CHECK:           {{.*}} = llvm.call @omTensorListGetOmtArray([[ARG0]]) : (!llvm.ptr<i8>) -> !llvm.ptr<ptr<i8>>
 
--- a/test/mlir/krnl/entry_point.mlir
+++ b/test/mlir/krnl/entry_point.mlir
@@ -1,6 +1,84 @@
 // RUN: onnx-mlir-opt --convert-krnl-to-llvm --canonicalize %s -split-input-file | FileCheck %s
 
 module {
+  func.func private @main_graph(%arg0: memref<10xf32>) -> memref<10xf32> {
+    return %arg0 : memref<10xf32>
+  }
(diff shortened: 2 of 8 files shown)


tungld:

tungld:
@kernhanda thank you for the patch! could you please look at why lit tests failed? You can check it locally by using `make check-onnx-lit`. Thanks!

test/mlir/krnl/entry_point.mlir
```diff
@@ -13,7 +11,7 @@ module {
 // CHECK:         llvm.mlir.global external constant @_entry_point_0_in_sig("[in_sig]\00")
 // CHECK:         llvm.mlir.global external constant @_entry_point_0_out_sig("[out_sig]\00")
 
-// CHECK-LABEL:   llvm.func @run_main_graph
+// CHECK-LABEL:   llvm.func @run_first_entry
```
sstamenova, line 14: Shouldn't the default continue to be generated as `run_main_graph`?

Also, we should have at least one test to validate providing a different name.

sstamenova (changes requested):

test/mlir/krnl/entry_point.mlir
kernhanda: > Shouldn't the default continue to be generated as run_main_graph?

This is open for debate, IMO. I'm fine if the answer is "yes", but I'd like some clarification. 

With this change, if the user provides their own graph name, thus overriding `main_graph`, for an ONNX model, the entry point emitted would still be `run_main_graph`, somewhat defeating the purpose. I could update the behavior such that `run_main_graph` is emitted if the user didn't override the default graph name.

The other scenario is where an existing MLIR file is being compiled, where there is only one kernel entry point specified. In this case, the current behavior is to disregard the name of the entry point and always emit `run_main_graph`. TBH, this doesn't make sense to me, but if there are reasons/dependencies that rely upon this behavior, I can make the change more selective.

> Also, we should have at least one test to validate providing a different name.
I'll add an ONNX model graph test, then.

sstamenova: So in the source, `main_graph` is still the default sometimes. There should be a test for this and if it's never the default, it should not be in the source. It sounds like we need at least 3 tests for the entry points.

kernhanda: Updated the test to ensure the default scenario is tested and then a scenario where the single entry point has a non-default name.

kernhanda:

test/mlir/krnl/entry_point.mlir
sstamenova: So the three tests I would expect to see are:
1. default "main_graph". My understanding is that if no name is specified, this is the name picked. Now both tests are specifying a name (though one of them is specifying "main_graph")
2. entry point with name in the graph
3. Using the command line option to specify a name other than "main_graph"

tungld:

kernhanda:
> I see a big advantage of specifying a custom entry point name: users have two (single entry) onnx models and they want to run them in the same Python/C program. In this case, entry points should be different.

Yes, that's exactly the motivation here.

> I prefer this. Since we say that default graph name is `main_graph` when importing an onnx model, it's clear to users. Also, Java and C interfaces are using `run_main_graph` as the default entry point.

With this change, if you import a model and don't specify the graph name, it does default to `main_graph`. 

> I am not so convince that onnx-mlir users will compile an MLIR file as input. It's mainly for onnx-mlir developers. So it looks fine to use the entry point as it is.

Do you think it's necessary to make the change so that if a MLIR file has a single entry point named something other than `main_graph` that it should still result in an entry point named `run_main_graph`?

^ This is the only question at the moment as it's the only difference in existing functionality. Otherwise, if you don't specify the graph name option, the behavior is the same as it is without this change.

test/mlir/krnl/entry_point.mlir
kernhanda: Are there any existing tests that import a basic ONNX model? I see that the repository has a `mnist.onnx` in the `docs` directory, but AFAICT, there are no tests that act on an ONNX model.

tungld:
> Do you think it's necessary to make the change so that if a MLIR file has a single entry point named something other than main_graph that it should still result in an entry point named run_main_graph?

If users specify an entry point in an MLIR via `onnx.EntryPoint` we should respect that and should not change to `main_graph`. In summary we only allow users to set entry point when importing an onnx model that does not have a concept of entry point. What do you think?

kernhanda:
> > Do you think it's necessary to make the change so that if a MLIR file has a single entry point named something other than main_graph that it should still result in an entry point named run_main_graph?
> 
> If users specify an entry point in an MLIR via `onnx.EntryPoint` we should respect that and should not change to `main_graph`. In summary we only allow users to set an entry point (`main_graph` by default) when importing an onnx model that does not have a concept of entry point. What do you think?

I agree! This PR matches that functionality then 😄 

@sstamenova requested tests that exercise the model import scenario, but I couldn't find any existing tests that do this. Any guidance on how to do this or if it's necessary would be appreciated.

src/Conversion/KrnlToLLVM/KrnlEntryPoint.cpp
```diff
@@ -89,11 +89,7 @@ class KrnlEntryPointOpLowering : public OpRewritePattern<KrnlEntryPointOp> {
             .getLeafReference()
             .getValue();
 
-    // When there is only a single entry point function in a model, use
-    // DEFAULT_DYN_ENTRY_POINT.
     std::string dynEntryPointName = "run_" + staticEntryPointFuncName.str();
-    if (singleEntryPoint)
-      dynEntryPointName = DEFAULT_DYN_ENTRY_POINT;
```
tungld, line 96: @kernhanda because we don't use `singleEntryPoint` anymore, could you please remove it elsewhere, e.g. in function arguments? Thanks!

test/mlir/krnl/entry_point.mlir
tungld: We don't have tests that import an ONNX model directly but @sorenlassen recenlty add tests that create a small ONNX model in JSON and pass it to `onnx-mlir` command line. One example is: https://github.com/onnx/onnx-mlir/blob/main/test/mlir/onnx/parse/test_abs.json. Hope it helps.

tungld:
@kernhanda any update on this?

tungld:
@kernhanda we would like to have this functionality, so do you have time to update this PR? If not I can take it over and create a new PR. thanks!

PR #1664, opened 2022-09-01
Conv1x1 matmul

AlexandreEichenberger:
Transform conv 1x1 into matmul. Detailed scheme is as follow:

```
   from:
     res = CONV(X=<NxCIxHxW>, W=<COxCIx1x1>)
   to:
     XX = reshape(X, <N, CO, H*W>) // flatten the last 2 dims.
     WW = squeeze(W) // get rid of the last 2 1s in the dims.
     MM = matmul(WW, XX) //  <CO, CI> * <N, CI, H*W> = <N, CO, H*W>
     if (has bias) {
        BB = unsqueeze(B, {0,2}) // <CO> -> <1, CO, 1>
        MM = add(MM, BB)
     }
     res = reshape(MM, <N, CO, H, W>)
```

I added a separate pass implemented in ConvOpt.cpp


AlexandreEichenberger:
@tungld I have not looked into calling this optimization only for CPU. Since NNPA has a different optimization sequence for the ONNX to ONNX part, do you mind letting me know if the code is ok as is or if I need a switch in the pass manager to turn it on only for CPU.

Thanks

tungld:
> @tungld I have not looked into calling this optimization only for CPU. Since NNPA has a different optimization sequence for the ONNX to ONNX part, do you mind letting me know if the code is ok as is or if I need a switch in the pass manager to turn it on only for CPU.

NNPA will call `addONNXToMLIRPasses`: https://github.com/onnx/onnx-mlir/blob/main/src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp#L120, so we need to condition it inside `addONNXToMLIRPasses` by either checking `-maccel` (e.g. empty `-maccel` indicates on CPU) or providing an option to `addONNXToMLIRPasses`.

tungld:

src/Dialect/ONNX/DialectBuilder.cpp
```diff
@@ -201,4 +224,67 @@ Value OnnxBuilder::where(
...
+    // NewShapeVal is [-1,M,N] where M & N are the kept vals from the input.
+    newShapeVals =
+        concat(outputShapeType, ValueRange({minusOneVal, keepVals}), 0);
+  else
+    // NewShapeVal is [M,N,-1] where M & N are the kept vals from the input.
+    newShapeVals =
+        concat(outputShapeType, ValueRange({keepVals, minusOneVal}), 0);
+  // Shape inference will infer the correct shape later, thus use -1 for
+  // collapsed dims.
+  std::vector<int64_t> rankedTensorDims;
```
tungld, line 279: Would it be better to use SmallVector?

src/Dialect/ONNX/DialectBuilder.cpp
```diff
@@ -201,4 +224,67 @@ Value OnnxBuilder::where(
...
+//
+// e.g. val has type 2x3x4x5xf32
+// reshape(val, 3, true) -> reshape([-1, 4, 5], val, [-1, 4, 5])
+
+// When  "collapseMostSignificant" is false, then we collapse the R-N+1 least
+// significant dimensions and keep the N-1 most significant dimensions as is.
+//
+// e.g. reshape(val, 3, false) -> reshape([2, 3, -1], val, [2, 3, -1])
+
+Value OnnxBuilder::reshapeToNDim(
```
tungld, line 243: Thanks for the generalization!

src/Transform/ONNX/ConvOpt.cpp
```diff
@@ -0,0 +1,259 @@
...
+        create.onnx.slice(batchCoutShapeType, xShapeVals, 0, 1);
+    Value CoutShapeVal =
+        create.onnx.slice(batchCoutShapeType, wShapeVals, 0, 1);
+    Value spatialShapeVal =
+        create.onnx.slice(spatialShapeType, xShapeVals, 2, rank + 1);
+    // Output shape values: batch, Cout, spatial shape values
+    Value outputShapeVals = create.onnx.concat(
+        shapeType, {batchShapeVal, CoutShapeVal, spatialShapeVal}, 0);
+    // Output type is the same as input, except for Cin becomes Cout.
+    std::vector<int64_t> outputTensorDims;
```
tungld, line 192: Please use SmallVector.

test/numerical/TestConv.cpp
```diff
@@ -87,11 +87,83 @@ int main(int argc, char *argv[]) {
...
+               1, 512, 64, 13, 13, 1, 1, 0, 0, 0, 0, ConvAutoPad::NOTSET) &&
+           "failed test from test_squeezenet_cpu");
+    assert(isOMConvTheSameAsNaiveImplFor(
+               1, 64, 256, 13, 13, 1, 1, 0, 0, 0, 0, ConvAutoPad::NOTSET) &&
+           "failed test from test_squeezenet_cpu");
+    assert(isOMConvTheSameAsNaiveImplFor(
+               1, 512, 1000, 13, 13, 1, 1, 0, 0, 0, 0, ConvAutoPad::NOTSET) &&
+           "failed test from test_squeezenet_cpu");
+    return 0;
+  }
```
tungld, line 161: This is cool. You can query all patterns in the backend benchmarks. Is `if (1)` redundant?

tungld (commented):
The PR is in a good shape I think. Waiting for making it work for CPU only. Thanks!

src/Dialect/ONNX/DialectBuilder.cpp
AlexandreEichenberger: done

src/Dialect/ONNX/DialectBuilder.cpp
AlexandreEichenberger: your welcome

src/Transform/ONNX/ConvOpt.cpp
AlexandreEichenberger: done

test/numerical/TestConv.cpp
AlexandreEichenberger: Its call grep and copy, but I was able to diff debug output to guarantee that it TestConv called the same kernels as in the benchmark. The only difference with the benchmark is that it also use a bias, and our testing does not currently support tests of bias... maybe something to add.

AlexandreEichenberger:
@tungld Done all the requested changes, including calling only for `targetCPU=maccel.empty()`

AlexandreEichenberger:

AlexandreEichenberger:

tungld (approved):
LGTM.

tungld:
> @tungld Done all the requested changes, including calling only for targetCPU=maccel.empty()

Thank you for the changes!

PR #1671, opened 2022-09-05
build: decouple CompilerOptions from ONNXToKrnl pass

ashay:
Prior to this patch, any consumer of the ONNXToKrnl pass (transitively)
depended on the CompilerOptions library.  This dependency caused
conflicts in identically-named command line options between the
CompilerOptions library and the consumers of the ONNXToKrnl pass.

This patch introduces a local option to the ONNXToKrnl pass so that its
dependency on CompilerOptions is removed, thus eliminating the symbol
conflicts.  When its time to pass a command line option to enable
parallelization in the ONNXTOKrnl pass, the call to
`createLowerToKrnlPass()` in CompilerPasses.cpp, which correctly depends
on the CompilerOptions library, will be able to pass the command line
value to the ONNXToKrnl pass.

### DIFF
--- a/src/Compiler/CompilerPasses.cpp
+++ b/src/Compiler/CompilerPasses.cpp
@@ -115,7 +115,7 @@ void addONNXToKrnlPasses(mlir::PassManager &pm, int optLevel, bool enableCSE,
   if (enableInstrumentONNXSignature)
     pm.addNestedPass<func::FuncOp>(
         onnx_mlir::createInstrumentONNXSignaturePass());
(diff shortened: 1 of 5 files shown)


tungld (commented):
Great! You are right. Do you mind to edit the call to `createLowerToKrnlPass` in https://github.com/onnx/onnx-mlir/blob/main/src/Compiler/CompilerPasses.cpp#L118 as well to include `enable-parallel`? Thanks!

tungld:

tungld:

tungld (approved):
LGTM. Thanks!

PR #1672, opened 2022-09-05
avoid re-definition of macro

ashay:
This patch adds a guard statement to only define the
`_SILENCE_ALL_CXX17_DEPRECATION_WARNINGS` macro if it has not previously
been defined.  Similarly, it checks to make sure that the macro has been
defined before un-defining it, but only if it was defined in this file.

This patch is necessary if there is a dependency on ORT or another third
party library that sets the macro.  If there is no such dependency, then
the macro is correctly set and unset in the root CMakeLists.txt, making
this patch a NOP.


tungld:
@jenkins-droid  test this please

sstamenova:
@ashay  : Can you share the steps that produced the error in your build?

tungld:

sstamenova:
@tungld : I'm following up with Ashay offline to see what the issue is, it may be that this change is not necessary because the definition in the cmake file applies to third party code while the definition in the header file applies to the numerical tests.

chentong319 (approved):
LGTM!

sstamenova:
We looked into this some more and the issue in Ashay's case is because we also depend on ORT which defines that flag. However, this might still happen in onnx-mlir in the future, so the change still makes sense here.

@ashay : Can you update the description to add some of these details?

test/numerical/Common.hpp
```diff
@@ -3,14 +3,18 @@
  */
 
 #if defined(_MSC_VER)
+#ifndef _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS
```
sstamenova, line 6: Since we are now saying that this might come from some place other than us, please save the save of the macro and then restore it below - don't just `undef` always since it may have been set by someone other than us.

sstamenova (changes requested):

test/numerical/Common.hpp
sstamenova: Something like:

```
#ifndef _SILENCE
#define _SILENCE
#else 
#define _SILENCE_WAS_DEFINED
```

```
#ifdef _SILENCE && !defined SILENCE_WAS_DEFINED
#undef _SILENCE
```

sstamenova: I would say something like: `if there's a dependency on ORT *or another third party library*...`

sstamenova (approved):

PR #1673, opened 2022-09-06
Fix the deallocation problem for sequence op

chentong319:
Since the sequence op is implemented as store pointers of element in memref, the buffer deallocation pass, which assumes no pointer, can not work correctly with previous sequence op lowering.
In this PR,  the load or store of the pointer of sequence element is encapsulated in krnl sequence ops. These Ops are lowered after bufferization::deallocation pass. These krnl sequence ops defines the interface for bufferazation so that deallocation can be generated correctly.
A lit test to check deallocation result is added.


AlexandreEichenberger:
@chentong319 ready for reviewing?

src/Dialect/Krnl/Krnl.td
```diff
@@ -180,26 +180,52 @@ def KrnlSeqExtractOp : Op<Krnl_Dialect, "seqextract", [MemRefsNormalizable,
...
+    When the attribute 'copy' is 1: the extracted element is copied and then return. 
+    The returned element is regarded as allocated by this Op because the 
+    element is created (copied) when inserted into the sequence.
+    Attribute 'copy' can be 0 when analysis or code generation guarantees 
+    that every element is extracted exactly once from the sequence.
   }];
   let arguments = (ins AnyMemRef:$seq,
-                       Index:$index);
+                       Index:$index,
+                       DefaultValuedAttr<SI64Attr, "0">:$copy);
```
AlexandreEichenberger, line 194: copy feels like a boolean attribute, see
```
    DefaultValuedAttr<BoolAttr, "true">:$simdize,
```
for example

src/Dialect/Krnl/Krnl.td
```diff
@@ -180,26 +180,52 @@ def KrnlSeqExtractOp : Op<Krnl_Dialect, "seqextract", [MemRefsNormalizable,
...
-    sequence is represented with memref<memref<>>.
-    This op loads a tensor for the sequence 'seq' at position 'index',
-    and return the tensor, which will be freed by Bufferization::Deallocation.
-    The element in the sequence will become null.
+    This op loads an element from the input sequence 'seq' at position 'index'.
+    The position value is guaranteed to be positive.
+    When the attribute 'copy' is 0: the extracted element is directly returned
+    When the attribute 'copy' is 1: the extracted element is copied and then return. 
+    The returned element is regarded as allocated by this Op because the 
+    element is created (copied) when inserted into the sequence.
```
AlexandreEichenberger, line 188: sentence above is hard to parse.

src/Dialect/Krnl/Krnl.td
```diff
@@ -180,26 +180,52 @@ def KrnlSeqExtractOp : Op<Krnl_Dialect, "seqextract", [MemRefsNormalizable,
...
+    Attribute 'copy' can be 0 when analysis or code generation guarantees
```
AlexandreEichenberger, line 189: I would expand a bit more on this. "If an optimization pass can determine that each element is only extracted once, then the copy=false can be used"

Is it required that each element be extracted once? If one is never extracted, should it still work? If so, "each element is extracted at most once"

src/Dialect/Krnl/Krnl.td
```diff
@@ -180,26 +180,52 @@ def KrnlSeqExtractOp : Op<Krnl_Dialect, "seqextract", [MemRefsNormalizable,
...
+    The sequence is implemented as memref of memref in onnx-mlir.
+    When an element is stored into a sequence, its pointer is stored.
+    MLIR assumes values for Op, not pointer. As a result, 
+    The memref for the element will be freed after this Op if it is the last
+    use of the element.
+    To avoid dangling pointer, the element is copied before it is inserted 
+    into the sequence. The allocation for the copy will be revealed to graph
+    when the element is extracted from the sequence. Consequently, the 
+    deallocation pass can handle the space correctly.
+  }];
```
AlexandreEichenberger, line 216: Would it make sense to have a bool "copy" which an optimization can also determine to set to false if certain conditions are satisfied?

src/Dialect/Krnl/Krnl.td
```diff
@@ -180,26 +180,52 @@ def KrnlSeqExtractOp : Op<Krnl_Dialect, "seqextract", [MemRefsNormalizable,
...
   let description = [{
-    sequence is represented with memref<memref<>>.
-    This op will copy the tensor to be stored, and cast the type if needed.
-    The motivation to introduce this Op is to help bufferization::deallocation
-    The experiment showed that memref will be freed after the memref is stored.
-    However, the store of memref only write out the pointer for the memref,
-    and its memory cannot be freed.
+    This op assumes that the input seq has the space for the new element and 
+    only need to copy the element to be stored, and cast the type if needed.
+    There is no return of a new seq, different from KrnlSeqInsertOp.
```
AlexandreEichenberger, line 228: same, would an optimization be able to use a copy=false argument?
What happen to the previous element? Is it guaranteed to be undefined? Or are there circumstances where it should be freed first?

AlexandreEichenberger (commented):
Trying to understand the scheme, which looks solid, but left me with a few questions. Will review this until I have a better overall understanding.

src/Dialect/Krnl/Krnl.td
chentong319: No. Since the deallocation pass can not handle the pointer  stored in sequence, the element will be freed by deallocation if there is no other use. We may add traits for one shot optimization to transfer the space and avoid the copy in future.

chentong319: Deallocation will free the input element if there is no other usage. In future, we may use one-shot optimization pass to specify the transferring of content and avoid copying.

src/Dialect/Krnl/Krnl.td
chentong319: `every element is extracted exactly once`

src/Dialect/Krnl/Krnl.td
AlexandreEichenberger: So would it make sense to add the "copy" field now?

src/Dialect/Krnl/Krnl.td
AlexandreEichenberger: Technically, if an element is not used out of a list, why would it disable the `copy=0`. IMO, the correct statement is "at most once"

chentong319: The element needs to be extracted or erased exactly once from the sequence. Otherwise, the memory for the element will be leaked.

src/Dialect/Krnl/Krnl.td
chentong319: No. Deallocation may be added before the element is extracted and used. Since the pointer for the element is stored to and loaded from the sequence. Deallocation pass can not see their u-d relationship.

src/Dialect/Krnl/Krnl.td
AlexandreEichenberger: If an element is not used, you imply that copy cannot be zero which makes no sense either.

AlexandreEichenberger: Additionally, if a value is used exactly twice, would it make sense to have the first use with copy=1 but the second/last use with copy=0?

chentong319: yes

chentong319: > If an element is not used, you imply that copy cannot be zero which makes no sense either.

The seqextract is a use the element.

chentong319: > If an element is not used, you imply that copy cannot be zero which makes no sense either.

An element can be accessed only through KrnlSeqExtract. If there is no KrnlSeqExtract, there is no copy attribute.

AlexandreEichenberger (commented):
much better, suggested you use a method in our krnl create class, otherwise its all good after a spell check

src/Conversion/ONNXToKrnl/ConvertONNXToKrnl.cpp
```diff
@@ -347,8 +347,9 @@ void FrontendToKrnlLoweringPass::runOnOperation() {
 
   // If `emitDealloc` is turned off, make sure we don't have buffer deallocation
   // at this level. Will use MLIR buffer-deallocation for this purpose instead.
-  if (!emitDealloc)
-    target.addIllegalOp<mlir::memref::DeallocOp>();
+  // However, since the SequenceErase needs to emit memref dealloc, the previous
```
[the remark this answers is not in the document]
chentong319: This PR needs to comment out the code because dealloc will be generated by SequenceErase.

src/Conversion/ONNXToKrnl/Sequence/SequenceAt.cpp
```diff
@@ -31,17 +31,21 @@ struct ONNXSequenceAtOpLowering : public ConversionPattern {
...
-    auto correctionIE = positionIE + boundIE;
-    positionIE = IndexExpr::select(positionIE < 0, correctionIE, positionIE);
-    auto outputVal = create.krnl.load(
-        operandAdaptor.input_sequence(), positionIE.getValue());
+    // Handle the negative position
+    IndexExpr condIE = positionIE < 0;
+    IndexExpr fixedPosition = positionIE + boundIE;
+    positionIE = IndexExpr::select(condIE, fixedPosition, positionIE);
+
+    Value outputVal = rewriter.create<KrnlSeqExtractOp>(loc, outputMemRefType,
```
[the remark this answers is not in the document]
chentong319: Not a big fan of the dialect builder, especially when there is nothing abstracted out by builder.

src/Conversion/SeqToMemref/KrnlSeqExtract.cpp
```diff
@@ -51,8 +54,25 @@ class KrnlSeqExtractOpLowering : public ConversionPattern {
     // if the element is read from seq after extracted, or deep deallocation
     // is added when seq is freed
 
-    rewriter.replaceOp(op, output);
-    return success();
+    if (thisOp.copy() == 0) {
+      rewriter.replaceOp(op, output);
+      return success();
+    } else {
+      if (!output.getType().isa<MemRefType>())
```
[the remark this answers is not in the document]
chentong319: In future, we will handle the non MemRefType case. I usually use unreachable for that purpose: not a semantic error but an implementation limitation. Not sure about the convention in llvm.

src/Conversion/SeqToMemref/KrnlSeqInsert.cpp
```diff
@@ -0,0 +1,144 @@
...
+    auto seqElementType =
+        seq.getType().cast<MemRefType>().getElementType().cast<MemRefType>();
+    auto casted = create.mem.cast(alloc, seqElementType);
+
+    // Copy the tensors in the sequence without duplicating the object
+    // ToFix: An analysis pass is needed to determine whether duplicating
+    // is needed.
+
+    // Copy elements before the insertion position
+    rewriter.create<scf::ForOp>(loc, create.math.constantIndex(0),
```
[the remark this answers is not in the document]
chentong319: This is due to the limitation of our current driver: lowering affine is before deallocation pass. I will add a comment.

src/Dialect/Krnl/Krnl.td
```diff
@@ -175,31 +175,83 @@ def KrnlIterateOp : Op<Krnl_Dialect, "iterate", [ImplicitKrnlTerminator,
...
 
+// ONNX sequence type (!onnx.Seq) is implemented with memref of memref.
+// When an element(tensor/memref) is stored into a sequence, it becomes
+// invisible to the bufferization analysis. When onnx-mlir invokes
+// bufferization::Deallocation pass to insert memref::dealloc, the element
+// may be freed. Therefore, element need to be copied when inserted into
+// a sequence. 'copy' here means to allocate a new memref and copy the content 
+// of the memref. The copy is a costly op to maintain the SSA requirement.
+// To reduce the overhead of copy, elements are not copied when
+// a new sequence is created from an exist sequence. Hower, when an element
```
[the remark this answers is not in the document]
chentong319: Thanks. I will try them.

src/Conversion/ONNXToKrnl/Sequence/SequenceAt.cpp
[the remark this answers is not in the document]
AlexandreEichenberger: Well we can abstract the handling of the int attribute, which I personally don't like to remember this awkward sequence... but if you don't want to do it, it is fine.

AlexandreEichenberger (approved):
LGTM pending spelling check

PR #1674, opened 2022-09-06
[DOC only] Onnx ai link broken

AlexandreEichenberger:
one link lead to a badly formatted html file. This fix this.


chentong319 (approved):
LGTM!

PR #1675, opened 2022-09-06
build: fix path to test data directory

ashay:
Prior to this patch, in a repository setup where onnx-mlir is not
present in the top-level directory, the build failed because the
`setup_model_download()` function assumed that the `third_party/onnx`
directory is located at the root of the repository.

This patch fixes the problem by replacing the reference to
`CMAKE_SOURCE_DIR` with `ONNX_MLIR_SRC_ROOT`.


sstamenova:
Thanks! Can you check whether there may be other places like that in the cmake files?

sstamenova (approved):

sstamenova:

PR #1679, opened 2022-09-07
show onnx-mlir-opt CLI options

sorenlassen:
I noticed that `onnx-mlir-opt -h` wasn't displaying `--split-input-file` and some other options

this PR adds the following output to `onnx-mlir-opt -h`:
```
ONNX-MLIR-OPT Options:
These are opt frontend options.

  --allow-unregistered-dialect                         - Allow operation with no registered dialects
  -o <filename>                                        - Output filename
  --split-input-file                                   - Split the input file into pieces and process each chunk independently
  --verify-diagnostics                                 - Check that emitted diagnostics match expected-* lines on the corresponding line
  --verify-each                                        - Run the verifier after each transformation pass
```

### DIFF
--- a/src/Tools/onnx-mlir-opt/onnx-mlir-opt.cpp
+++ b/src/Tools/onnx-mlir-opt/onnx-mlir-opt.cpp
@@ -42,31 +42,36 @@
 using namespace mlir;
 using namespace onnx_mlir;
 
-static llvm::cl::opt<std::string> input_filename(
-    llvm::cl::Positional, llvm::cl::desc("<input file>"), llvm::cl::init("-"));
+// Options for onnx-mlir-opt only.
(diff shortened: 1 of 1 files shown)


sorenlassen:
fwiw, here's a different way to do it:
https://github.com/sorenlassen/onnx-mlir-einsum/commit/b75eb59f722b31b6dfe9386f303c1ae8d94f7a37

I came up with this alternative after looking at the "original" implementation:
https://github.com/llvm/llvm-project/blob/main/mlir/lib/Tools/mlir-opt/MlirOptMain.cpp#L204-L238

AlexandreEichenberger (approved):
LGTM, thanks for doing these "quality of interface for user" fixes, much appreciated

PR #1681, opened 2022-09-07
Combine inferred shape and user-defined shape in ONNXShapeHelper

tungld:
Before this patch, we combine the inferred shape and user-defined shape during shape inference for all operations. However, when lowering an operation, such combination is not applied because the lowering calls ShapeHelper.

In this patch, we move the shape combination into ShapeHelper so that both shape inference and lowering can utilize the combination.


src/Conversion/ONNXToKrnl/Math/Reduction.cpp
```diff
@@ -471,7 +471,7 @@ struct ONNXReduceSumOpLowering : public ConversionPattern {
         KrnlBuilder createKrnl(rewriter, loc);
         ValueRange loopDef = createKrnl.defineLoops(1);
         SmallVector<IndexExpr, 4> lbs(1, LiteralIndexExpr(0));
-        createKrnl.iterateIE(loopDef, loopDef, lbs, shapeHelper.dimsForOutput(),
+        createKrnl.iterateIE(loopDef, loopDef, lbs, axesDims,
```
tungld, line 474: This looks a wrong use of `dimsForOutput` to store dims for axes. So fix it.

src/Dialect/ONNX/ShapeInference/ONNXShapeHelper.cpp
```diff
@@ -65,6 +65,48 @@ ONNXOpShapeHelper<OP>::ONNXOpShapeHelper(OP *newOp, int numResults,
...
 }
 
+template <>
+Value ONNXOpShapeHelper<Operation>::getOutput(int n) {
+  return op->getResult(n);
+}
+
+template <class OP>
+DimsExpr &ONNXOpShapeHelper<OP>::dimsForOutput(int n) {
+  Value output = getOutput(n);
```
tungld, line 75: @AlexandreEichenberger I found that this is good location to insert the combination code, in the sense that it minimizes the changes.

src/Dialect/ONNX/ShapeInference/ONNXShapeHelper.cpp
```diff
@@ -65,6 +65,48 @@ ONNXOpShapeHelper<OP>::ONNXOpShapeHelper(OP *newOp, int numResults,
...
+  // Try to combine the inferred shape and the output's shape if possbile.
+  if (isRankedShapedType(output.getType())) {
+    llvm::ArrayRef<int64_t> existingDims = getShape(output.getType());
+    // Do not handle the case of scalar tensor whose type can be tensor<f32>
+    // or tensor<1xf32>. Just use the inferredShape in this case.
+    if (existingDims.size() >= 1 && outputsDims[n].size() >= 1) {
+      if (existingDims.size() != outputsDims[n].size())
+        llvm::outs() << "Inferred rank (" << outputsDims[n].size()
+                     << ") vs. existing rank (" << existingDims.size() << ")\n";
+      assert((existingDims.size() == outputsDims[n].size()) &&
```
AlexandreEichenberger, line 85: this should be transformed into an assert without the if, or the if with a `llvm_unreachable` with error message. Single assert seems enough.

src/Dialect/ONNX/ShapeInference/ONNXShapeHelper.cpp
```diff
@@ -65,6 +65,48 @@ ONNXOpShapeHelper<OP>::ONNXOpShapeHelper(OP *newOp, int numResults,
...
+
+template <class OP>
+DimsExpr &ONNXOpShapeHelper<OP>::dimsForOutput(int n) {
+  Value output = getOutput(n);
+  // Try to combine the inferred shape and the output's shape if possbile.
+  if (isRankedShapedType(output.getType())) {
+    llvm::ArrayRef<int64_t> existingDims = getShape(output.getType());
+    // Do not handle the case of scalar tensor whose type can be tensor<f32>
+    // or tensor<1xf32>. Just use the inferredShape in this case.
+    if (existingDims.size() >= 1 && outputsDims[n].size() >= 1) {
```
AlexandreEichenberger, line 81: Why predicate all of this for sizes > 0? Size zero means no info?

src/Dialect/ONNX/ShapeInference/ONNXShapeHelper.cpp
```diff
@@ -65,6 +65,48 @@ ONNXOpShapeHelper<OP>::ONNXOpShapeHelper(OP *newOp, int numResults,
...
+      if (existingDims.size() != outputsDims[n].size())
+        llvm::outs() << "Inferred rank (" << outputsDims[n].size()
+                     << ") vs. existing rank (" << existingDims.size() << ")\n";
+      assert((existingDims.size() == outputsDims[n].size()) &&
+             "Inferred shape and existing shape are inconsistent in the number "
+             "of elements");
+      for (unsigned i = 0; i < existingDims.size(); ++i) {
+        // existingDim is static, inferedDim is unknown: update the
+        // inferredDim.
+        if ((existingDims[i] != -1) && outputsDims[n][i].isQuestionmark())
```
AlexandreEichenberger, line 91: I would restructure 
```C++
if (existingDIms[i] != -1) {
  if (question mark) { 
    // we have question mark only during the shape inference
  } else if (!literal) {
     // NEW CASE: we have computation during the lowering... use the constant given by user
   } else if (literal) {
     // issue a warning if found a different value, keep user given value
  } 
}
```

src/Dialect/ONNX/ShapeInference/Gather.cpp
```diff
@@ -39,14 +39,16 @@ LogicalResult ONNXGatherOpShapeHelper::computeShape(
...
     if (i == axisIndex)
       for (IndexExpr d : indicesDims)
-        dimsForOutput().emplace_back(d);
+        outputDims.emplace_back(d);
     else
-      dimsForOutput().emplace_back(dataDims[i]);
+      outputDims.emplace_back(dataDims[i]);
   }
 
+  dimsForOutput() = outputDims;
```
AlexandreEichenberger, line 51: I don't like that... we should disallow it as now `dimsForOutput` calculate stuff, only to be overridden immediately by the `outputDims`.

Please add a setter to only assign the value.

AlexandreEichenberger (changes requested):
please add a setter, and I wonder if you can also use the user given constant during lowering

src/Dialect/ONNX/ShapeInference/ONNXShapeHelper.cpp
tungld: Now using assert only. Thanks!

src/Dialect/ONNX/ShapeInference/ONNXShapeHelper.cpp
tungld: Size zero means `tensor<elementtype>` that is a scalar tensor. `tensor<1xelementype>` whose rank is 1 is also to represent a scalar tensor in onnx-mlir. They are used exchangebly in onnx-mlir. So not sure `tensor<elementtype>` or `tensor<1xelementtype>` is correct. It seems there is some op in yolov3 models that can handle both.

src/Dialect/ONNX/ShapeInference/ONNXShapeHelper.cpp
tungld: Done and added the new case so that the lowering code uses static dim instead. Hope it's what you mean. Thanks!

src/Dialect/ONNX/ShapeInference/Gather.cpp
tungld: Added a setter `setOutputDims()` to use instead.

src/Dialect/ONNX/ShapeInference/ONNXShapeHelper.cpp
AlexandreEichenberger: got it, tx

src/Dialect/ONNX/ShapeInference/ONNXShapeHelper.cpp
AlexandreEichenberger: thanks

AlexandreEichenberger (approved):
LGTM, thanks for addressing our concerns.

AlexandreEichenberger:
@tungld one side effect of this PR (and related improvement in other PRs) is that if the user now give erroneous shapes with the compile options, and since the user shapes are given preference over compiler derived shapes, we can get more errors when verifying ops. A good example is convolution, where some of the parameters must be the same in the input image/X and the input filter/W. 

So maybe we should upgrade our error message to indicate that if a shape was given as an input and we get this error, the user should double check the validity of the shapes given with the compile option.

tungld:
@AlexandreEichenberger The shapes given with the compile option (e.g. `--shapeInformation`) are for operation inputs. Meanwhile, the combination here deals with outputs' shapes, e.g. the output shape of Conv is static inside the onnx model. It looks to me the error in the conv example is not due to this combination.

PR #1682, opened 2022-09-07
build: fix dependencies on MLIR libraries and tools

ashay:
An in-tree build, i.e. when all MLIR targets are not already built and are instead built on demand, exposes a handful of missing dependencies on various MLIR libraries and tools.  This patch adds the necessary dependencies.

In particular:

 - Both onnx-mlir-opt.cpp and onnx-mlir-reduce.cpp include InitAllPasses.h and InitAllDialects.h, thus requiring all dialect and conversion libraries to be linked into those targets.

 - CompilerUtils.cpp shells out to opt and llc tools, thus requiring the two binaries to be built.

 - InitAccelerators.cpp, just like Accelerator.cpp, includes Accelerators.hpp, which includes mlir/IR/BuiltinOps.h, this requring a dependency on MLIRIR, although this can probably be trimmed to be leaner.

### DIFF
--- a/src/Accelerators/CMakeLists.txt
+++ b/src/Accelerators/CMakeLists.txt
@@ -43,6 +43,7 @@ add_onnx_mlir_library(InitAccelerators
 
   DEPENDS 
     AcceleratorsInc
+    MLIRIR
 
   LINK_LIBS PUBLIC
     ${ACCEL_TARGET_LIST}
--- a/src/Compiler/CMakeLists.txt
+++ b/src/Compiler/CMakeLists.txt
@@ -123,6 +123,8 @@ add_onnx_mlir_library(CompilerUtils
   DEPENDS
   ExternalUtil
   MLIRIR
+  llc
+  opt
 
   INCLUDE_DIRS PRIVATE
   ${FILE_GENERATE_DIR}
--- a/src/Tools/onnx-mlir-opt/CMakeLists.txt
+++ b/src/Tools/onnx-mlir-opt/CMakeLists.txt
@@ -1,11 +1,15 @@
 # SPDX-License-Identifier: Apache-2.0
 
 get_property(OMLibs GLOBAL PROPERTY ONNX_MLIR_LIBS)
+get_property(DialectLibs GLOBAL PROPERTY MLIR_DIALECT_LIBS)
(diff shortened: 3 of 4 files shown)


sstamenova:
@AlexandreEichenberger : Do we want onnx-mlir to support in-tree builds where both llvm and onnx-mlir are including in the same parent cmake file?

AlexandreEichenberger:
> @AlexandreEichenberger : Do we want onnx-mlir to support in-tree builds where both llvm and onnx-mlir are including in the same parent cmake file?

I really don't have strong opinions on such issue, and would rely on the recommendation of others on this matter

@caoimhinuibrian @tungld @chentong319 @gongsu832 suggestions?

gongsu832:
> @AlexandreEichenberger : Do we want onnx-mlir to support in-tree builds where both llvm and onnx-mlir are including in the same parent cmake file?

Trying to understand what this means exactly. You mean someone wants to have a parent project that includes both llvm-project and onnx-mlir? Can you elaborate a bit?

gongsu832:
> That said, shouldn't we always add the necessary dependencies, regardless of how the project is built? Targets like onnx-mlir-opt do depend on certain MLIR targets that aren't expressed in the CMakeCache.txt files, don't they?

In theory, yes. And in practice, when you build things statically (as is the case with llvm-project default build), it's also relatively simple, i.e., you can just add the missing dependencies to the list. However, things get far more complicated when you build shared libraries. Because the list is now order sensitive. That is, if `libfoo.so` depends on `libbar.so`, the list must have the order `-lfoo -lbar` whereas the order doesn't matter if you build static `libfoo.a` and `libbar.a`.

So you can see how things quickly go out of hand when you have a lot of shared libraries with complicated dependencies. Yes it's still possible to work out the exact dependencies of every library and list them in the correct order but the maintenance overhead is prohibitive. That's why onnx-mlir build assumes that llvm-project has been fully built and we don't try to work out dependencies on all its individual components.

That said, supporting in-tree build against on-demand MLIR targets is not a bad thing in general. I'm just afraid that to fully support it there would be more work required than this patch and there is associated upkeep cost. But of course, this is an open source project so if you or someone else is willing to implement and maintain it, I certainly wouldn't object.

AlexandreEichenberger:
will ignore the verify patch as the pr is closed

PR #1683, opened 2022-09-08
Simplified interface for compile from programs

AlexandreEichenberger:
New interface allows all of the onnx-mlir options to be used
```
ONNX_MLIR_EXPORT int64_t omCompileFromFileViaCommand(const char *inputFilename,
    const char *flags, const char **outputFilename, const char **errorMessage);
```

and, for example, if you need the output to be named after something special, we can simply add `-o output name` in the flag string. Or if you want to output a JNI, just add `-EmitJNI`.

Also, it is now possible to use the `"-customEnvFlags=MY_ONNX_MLIR_FLAGS"` so that the default is taken from a custom environment variable, not necessarily the default `ONNX_MLIR_FLAGS`. 

Full interface:
```C
/*!
 *  C interface to compile an onnx model from a file via onnx-mlir command.
 *  This interface is thread safe, and does not take any flags from the
 *  current environment. All flags are passed by using the flags parameter,
 *  including the "-o output-file-name" option or the "-EmitXXX" options. All
 *  options that are available to onnx-mlir are also available here.
 *
 *  This call rely on executing onnx-mlir compiler. The user can override its
 *  default location by using the ONNX_MLIR_BIN_PATH environment variable.
 *
 *  When generating libraries or jar files, the compiler will link in
 *  lightweight runtimes / jar files. If these libraries / jar files are not in
 *  the system wide directory (typically /usr/local/lib), the user can override
 *  the default location using the ONNX_MLIR_LIBRARY_PATH environment variable.
 *
 *  @param inputFilename File name pointing onnx model protobuf or MLIR.
 *  Name may include a path, and must include the file name and its extention.
 *
 *  @param outputFilename Output file name of the compiled output for the given
 * emission target. User is responsible for freeing the string.
 *
 *  @param flags A char * contains all the options provided to compile the
 * model.
 *
 *  @param errorMessage Output error message, if any. User is responsible for
 * freeing the string.
 *
 *  @return 0 on success or OnnxMlirCompilerErrorCodes on failure.
 */
ONNX_MLIR_EXPORT int64_t omCompileFromFile(const char *inputFilename,
    const char *flags, const char **outputFilename, const char **errorMessage);
```

### DIFF
--- a/test/perf/PerfHelper.hpp
+++ b/test/perf/PerfHelper.hpp
@@ -34,7 +34,9 @@
     ::benchmark::Initialize(&argc, argv);                                      \
     onnx_mlir::omSetCompilerOption(                                            \
         onnx_mlir::OptionKind::CompilerOptLevel, O3.c_str());                  \
-    if (onnx_mlir::omSetCompilerOptionsFromEnv(envArgName.c_str()) != 0)       \
+    const char *myArgv[] = "perf-algo";                                        \
+    if (!llvm::cl::ParseCommandLineOptions(                                    \
+            1, myArgv, "set options for perf-algo", envArgName.c_str()))       \
       return 2;                                                                \
     if (::benchmark::ReportUnrecognizedArguments(argc, argv))                  \
       return 1;                                                                \
--- a/test/perf/PerfHelper.hpp
+++ b/test/perf/PerfHelper.hpp
@@ -27,16 +27,14 @@
 // Define performance main, with default opt level of 3, and scan PERF_ARGS to
 // override default onnx-mlir compiler options.
 #define PERF_MAIN()                                                            \
-  const std::string envArgName("PERF_ARGS");                                   \
-  const std::string O3("3");                                                   \
-                                                                               \
   int main(int argc, char **argv) {                                            \
     ::benchmark::Initialize(&argc, argv);                                      \
-    onnx_mlir::omSetCompilerOption(                                            \
-        onnx_mlir::OptionKind::CompilerOptLevel, O3.c_str());                  \
-    const char *myArgv[] = "perf-algo";                                        \
-    if (!llvm::cl::ParseCommandLineOptions(                                    \
-            1, myArgv, "set options for perf-algo", envArgName.c_str()))       \
+    int onnxMlirArgc = 2;                                                      \
+    const char *onnxMlirArgv[onnxMlirArgc];                                    \
+    onnxMlirArgv[0] = argv[0];                                                 \
+    onnxMlirArgv[1] = "-O3";                                                   \
+    if (!llvm::cl::ParseCommandLineOptions(onnxMlirArgc, onnxMlirArgv,         \
+            "set options for perf-algo", nullptr, /*env var*/ "PERF_ARGS"))    \
       return 2;                                                                \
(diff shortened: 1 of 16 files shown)


AlexandreEichenberger:
@Jiaqing-ASU FYI, I streamlined the interface, it should be much easier to integrate with python as now we can pass all of the compile options as part of the flag.

tungld:
Thanks so much! It's an important step towards a single python session that does both compilation and running.

Jiaqing-ASU:
> @Jiaqing-ASU FYI, I streamlined the interface, it should be much easier to integrate with python as now we can pass all of the compile options as part of the flag.

Thanks @AlexandreEichenberger for this great simplified interface!

Jiaqing-ASU:
> My understanding is that you should be able to have a compile python interface that has no need for complicated "option" setting methods. It should be as simple as `xxx.compile(input-file, flags)` where the flags are similar to the options one would give to the `onnx-mlir` command.

> If you had time, one thing that is kind of broken is that error messages in the C++ interface don't get anything from the separate process that runs the compilation. If you know how, one could redirect the stderr into a stream that would then be dumped in the "errorMessage" array. If you don't have the time, that is ok too. Would be great if you completed the python interface as you have already invested effort into understanding it.

Totally agree. I will take a look at them and try to fix those stuff as much as I can once this PR has landed and let you know the progress. Many thanks!

test/perf/PerfHelper.hpp
```diff
@@ -27,14 +27,14 @@
...
 #define PERF_MAIN()                                                            \
-  const std::string envArgName("PERF_ARGS");                                   \
-  const std::string O3("3");                                                   \
-                                                                               \
   int main(int argc, char **argv) {                                            \
     ::benchmark::Initialize(&argc, argv);                                      \
-    onnx_mlir::omSetCompilerOption(                                            \
-        onnx_mlir::OptionKind::CompilerOptLevel, O3.c_str());                  \
-    if (onnx_mlir::omSetCompilerOptionsFromEnv(envArgName.c_str()) != 0)       \
+    int onnxMlirArgc = 2;                                                      \
```
sstamenova, line 32: This causes build breaks when compiling on Windows. I noticed we don't build these tests on Windows right now, so I'd like to add them to the CI build to avoid another quiet failure like that, but before that we'll need a fix for the failure as well.

The specific issue is with `onnxMlirArgc`:

```
D:\sources\onnx-mlir-mainline\test\perf\PerfGemm.cpp(124): error C2131: expression did not evaluate to a constant
D:\sources\onnx-mlir-mainline\test\perf\PerfGemm.cpp(124): note: failure was caused by a read of a variable outside its lifetime
D:\sources\onnx-mlir-mainline\test\perf\PerfGemm.cpp(124): note: see usage of 'onnxMlirArgc'
D:\sources\onnx-mlir-mainline\test\perf\PerfGemm.cpp(124): error C3863: array type 'const char *[onnxMlirArgc]' is not assignable
```

but I wonder why `PERF_MAIN` is a macro and not a function call? Same with the other functions in this file.

AlexandreEichenberger:
@sstamenova I created #1692 to remove the inlined code and transformed it into a separate lib. 
It is a better style of code in any cases. Hopefully that will help the error you witnessed too (which looked somewhat cryptic to me).

PR #1684, opened 2022-09-08
Support ingesting Opset 15 Shape Op

philass:
This PR adds support for shape inference for Opset-15 shape op. 

Opset 15 added a default start attribute to the shape op (defaults to 0), and a optional end attribute.

These changes are not breaking changes to the older Shape op. Therefor we do not need to create a new ShapeOp to support the difference in API.


AlexandreEichenberger:
Do you mind adding a short description of the transfer, tx

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.td
```diff
@@ -88,6 +88,17 @@ def replaceONNXBatchNormalizationInferenceModePattern : Pattern<
 // 
 //===----------------------------------------------------------------------===//
 
+// Get a type for a tensor that stores the shape of another tensor.
+def GetStart: NativeCodeCall<
+  //"IntegerType::get($_builder, 64, IntegerType::Signed)"
```
AlexandreEichenberger, line 93: Should remove commented code. Have you though about adding a custom builder of the operation that does not have the start/end and would automatically generate them.... instead of modifying each of the patterns involved with the new shape op?

src/Dialect/ONNX/ONNXOpsHelper.cpp
```diff
@@ -495,12 +495,34 @@ Value normalizeConstantOp(
...
+  }
+
+  // Normalize if start/end are not in (0, ..., rank)
+  if (start < 0) {
+    start = start + rank;
+  }
+  if (end < 0) {
+    end = end + rank;
+  }
+
```
AlexandreEichenberger, line 523: would it make sense to add asserts to make sure that start and end are now in inclusively 0..rank-1 for start, 0..rank for end?

philass: Good catch! This was actually wrong because the behaviour is meant to do be the following

> "Optional attributes start and end can be used to compute a slice of the input tensor's shape."
  "If start axis is omitted, the slice starts from axis 0."
  "The end axis, if specified, is exclusive (and the returned value will not include the size of that axis)."
  "If the end axis is omitted, the axes upto the last one will be included."
  "Negative axes indicate counting back from the last axis."
  "Note that axes will be clipped to the range [0, r-1], where r is the"
  "rank of the input tensor if they are out-of-range (after adding r in the case of"
  "negative axis). Thus, specifying any end value > r is equivalent to specifying an end"
  "value of r, and specifying any start value < -r is equivalent to specifying a start"
  "value of 0."
  
  I've rejigged the logic to use `getDataShapeBounds` which handles the clipping logic. This puts the normalizing behavior in one place, where it is handled correctly

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.td
philass: I did consider adding a custom builder. In all honesty I tried and was having some issues getting it to work.
 
 But you makes good points.  I have added a Native code call to more succinctly create ONNX Shape ops
 ```
 // Create an ONNX Shape Op with type
def CreateShapeOp: NativeCodeCall<
  "$_builder.create<mlir::ONNXShapeOp>($_loc, $0, $1, IntegerAttr(), 0)"
>;
```

And I now call this, to avoid having to use the start and end attributes everywhere.

Comment (AlexandreEichenberger) [reply src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.td]: Awesome, glad you found a way to simplify the code

File: src/Dialect/ONNX/Rewrite.cpp [hunk 3]
```diff
@@ -102,6 +103,36 @@ bool areProducedByTransposeOp(ValueRange values) {
...
+DenseElementsAttr createDenseElementsAttrFromShape(PatternRewriter &rewriter,
+    Value value, int64_t start = 0, llvm::Optional<int64_t> end = llvm::None) {
+
+  auto inType = value.getType().cast<ShapedType>();
+  assert(inType.hasRank() && "inType must be ranked");
+  auto shape = inType.getShape();
+  int64_t rank = inType.getRank();
+
+  int64_t endValue = end.has_value() ? end.value() : rank;
+
```
AlexandreEichenberger, line 116: would it make sense to have an assert here for sizes, or better, it could be added in the verifier (apologies if it is already there)

src/Dialect/ONNX/ShapeInference/Shape.cpp
```diff
@@ -9,60 +9,83 @@
...
+namespace {
+
+// The Shape op spec says:
+//
+// "Note that axes will be clipped to the range [0, r-1], where r is the
+// rank of the input tensor if they are out-of-range (after adding r in the case
+// of negative axis). Thus, specifying any end value > r is equivalent to
+// specifying an end value of r, and specifying any start value < -r is
+// equivalent to specifying a start value of 0."
+int64_t normalize(int64_t axis, int64_t rank) {
```
AlexandreEichenberger, line 28: I see what you do here, it sounds reasonable but is it the standard? Or should we instead report an error? 
In general, we have erred in favor of reporting errors so as to make the user aware of the issue, instead of "fixing" the model the way we "hope" the user wanted.
As a minimum, I would issue a warning.
Note also that if I recall correctly, sometime the clipping is at rank and sometime rank-1... so this method may not be applied everywhere.

@chentong319 do you have an opinion on this issue?

AlexandreEichenberger (commented):
getting very close, thanks for the improvement. Left a question.

src/Dialect/ONNX/ShapeInference/Shape.cpp
philass: Sorry if this was confusing. This was actually lifted for the ONNX specification

> Shape-15
Takes a tensor as input and outputs an 1D int64 tensor containing the shape of the input tensor. Optional attributes start and end can be used to compute a slice of the input tensor's shape. If start axis is omitted, the slice starts from axis 0. The end axis, if specified, is exclusive (and the returned value will not include the size of that axis). If the end axis is omitted, the axes upto the last one will be included. Negative axes indicate counting back from the last axis. Note that axes will be clamped to the range [0, r-1], where r is the rank of the input tensor if they are out-of-range (after adding r in the case of negative axis). Thus, specifying any end value > r is equivalent to specifying an end value of r, and specifying any start value < -r is equivalent to specifying a start value of 0.

It seems weird, but I have based it off the ONNX spec of Shape.

AlexandreEichenberger: Great, I would then just reflect this policy in the function name, so that others may not be tempted to use it when it is not applicable. Add something of the sort "ClampedPerSpec" or something similar.

AlexandreEichenberger (approved):
LGTM

AlexandreEichenberger:
I believe you should now be able to squash PRs that have been approved.

philass:
@AlexandreEichenberger It looks like I can!

Thanks for the review as always! I appreciate you taking the time :)

AlexandreEichenberger:
@philass is it ready to go?

AlexandreEichenberger:
@philass 

there seems to be some compilation error linked with the NNPA. Do you mind looking into it, and if it is something that we need to do, let us know
```
[91m/workdir/onnx-mlir/src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.td:249:1: error: op 'onnx.Shape' argument number mismatch: 2 in pattern vs. 3 in definition
def rewriteSoftmaxNDto2D: Pat<
^
[0m[91mmake[2]: *** [src/Accelerators/NNPA/Conversion/ONNXToZHigh/CMakeFiles/OMONNXRewriteONNXForZHighIncGen.dir/build.make:106: src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXRewriteONNXForZHigh.inc] Error 1
[0m[91mmake[1]: *** [CMakeFiles/Makefile2:10481: src/Accelerators/NNPA/Conversion/ONNXToZHigh/CMakeFiles/OMONNXRewriteONNXForZHighIncGen.dir/all] Error 2
```

philass:
> @philass
> 
> there seems to be some compilation error linked with the NNPA. Do you mind looking into it, and if it is something that we need to do, let us know
> 
> ```
> [91m/workdir/onnx-mlir/src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.td:249:1: error: op 'onnx.Shape' argument number mismatch: 2 in pattern vs. 3 in definition
> def rewriteSoftmaxNDto2D: Pat<
> ^
> �[0m�[91mmake[2]: *** [src/Accelerators/NNPA/Conversion/ONNXToZHigh/CMakeFiles/OMONNXRewriteONNXForZHighIncGen.dir/build.make:106: src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXRewriteONNXForZHigh.inc] Error 1
> �[0m�[91mmake[1]: *** [CMakeFiles/Makefile2:10481: src/Accelerators/NNPA/Conversion/ONNXToZHigh/CMakeFiles/OMONNXRewriteONNXForZHighIncGen.dir/all] Error 2
> ```

Yeah I'll look into this. I won't merge and squash until this is resolved. I'll have some free time this weekend to look into it

PR #1685, opened 2022-09-08
ONNX parser lit tests

sorenlassen:
with the goal of making it easier to identify gaps and verify improvements to the ONNX parser this PR introduces ONNX parser lit tests with ONNX protobuf input expressed textually in JSON

the PR extends onnx-mlir to take JSON input either from a file with .json extension or from STDIN so that the JSON input can be piped in  the RUN command in the lit tests strips `//` comments (which are not valid JSON) with sed and pipes the stripped JSON into onnx-mlir, and then pipes the MLIR output of onnx-mlir to STDOUT with `—printIR`

parse failures are tested using the `FAILED` template from
`test/mlir/accelerators/nnpa/transform/zlow-normalize-by-using-dummyop.mlir`

the script `utils/onnx2json.py` translates ONNX models to JSON textual format


src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -1406,21 +1410,36 @@ int ImportFrontendModelArray(const void *onnxBuffer, int size,
...
-  if (!input.is_open()) {
-    *errorMessage = "Unable to open or access " + model_fname;
-    return InvalidInputFileAccess;
-  }
+  if (model_fname == "-" || model_fname.endswith(".json")) {
+    auto buf = openInputFile(model_fname, errorMessage);
+    if (!buf) {
+      return InvalidInputFileAccess;
+    }
+    google::protobuf::StringPiece json(
```
sorenlassen, line 1422: idea: we could remove `//` comments from buf here before calling JsonStringToMessage, then we wouldn't need to strip the comments with sed in the lit tests

test/mlir/onnx/parse/test_abs.json.mlir
```diff
@@ -0,0 +1,78 @@
+// RUN: sed 's/^[ \t]*\/\/.*//' %s | onnx-mlir --EmitONNXBasic --printIR | FileCheck %s
```
sorenlassen, line 1: the sed command to remove `//` comments is rather cryptic

see comment in `FrontendDialectTransformer.cpp` about changing onnx-mlir to ignore comments

alternatively we could make the RUN line more self explanatory by calling a `strip-comments` script that runs the sed command, but I'm not sure how to put the script in a place where the lit test runner can find it

test/mlir/onnx/parse/test_gridsample.json.mlir
```diff
@@ -0,0 +1,124 @@
...
+  },
+  "opsetImport": [
+    {
+      "version": "16"
+    }
+  ]
+}
+// GridSample was added in ONNX v17 and not yet added to onnx-mlir
+
+// FAILED: GridSample this Op is not supported by onnx-mlir's onnx dialect
```
sorenlassen, line 124: using the `FAILED` template from
`test/mlir/accelerators/nnpa/transform/zlow-normalize-by-using-dummyop.mlir`

test/mlir/onnx/parse/test_abs.json.mlir
sorenlassen, line 1: I couldn't get `cmake --build . --target check-onnx-lit` to run the lit tests without adding the extra `.mlir` extension to the lit tests

(but an added benefit of the `.mlir` extension is that vscode syntax highlighting is nicer, vscode marks up the `//` comments in red because they are invalid in json)

sorenlassen: I figured out how to add '.json' to test/mlir/lit.cfg.py to run .json lit tests and I renamed all the parse tests to .json extension

test/mlir/onnx/parse/test_gridsample.json
```diff
@@ -0,0 +1,124 @@
+// RUN: sed 's/^[ \t]*\/\/.*//' %s | onnx-mlir --EmitONNXBasic --printIR 2> test_gridsample.log; cat test_gridsample.log | FileCheck --check-prefix=FAILED %s
```
sorenlassen, line 1: MLIR-Windows-CI seems to run in powershell and I couldn't come
up with a cross-platform way to pipe from a command with a
non-zero error code directly into FileCheck, so instead piping the
stderror output to a file and then piping that file into FileCheck

initially, I had copied the approach from `test/mlir/accelerators/nnpa/transform/zlow-normalize-by-using-dummyop.mlir` but that didn't work in powershell - I think we never saw that lit test fail the MLIR-Windows-CI because that CI isn't configured with `-DONNX_MLIR_ACCELERATORS=NNPA`

sorenlassen:
this PR is ready for review

the "Jenkins Linux ppc64le" failure looks unrelated to this PR, it seems like that CI pipeline broke this morning

src/Compiler/CompilerUtils.cpp
```diff
@@ -611,31 +611,39 @@ void registerDialects(mlir::MLIRContext &context) {
...
-  if (!inputIsONNX && !inputIsMLIR) {
-    *errorMessage = "Invalid input file '" + inputFilename +
-                    "': Either an ONNX model (.onnx), or an MLIR file (.mlir) "
+  // Decide if the input file is an ONNX model (either ONNX protobuf or JSON) or
+  // a model specified in MLIR. The extension of the file is the decider or, if
+  // the file is "-", then it's STDIN and we treat it as JSON.
+  bool inputIsONNX = false;
+  bool inputIsJSON = false;
+  bool inputIsMLIR = false;
+  if (inputFilename == "-") {
```
AlexandreEichenberger, line 622: Is there a reason that inputs from stdin must be JSON format? That feels a bit arbitrary. Wonder if we don't want an option
`-stdin-format=json|mlir|onnx`
and have onnx being the default.

test/mlir/onnx/parse/test_abs.json
```diff
@@ -0,0 +1,78 @@
+// RUN: sed 's/^[ \t]*\/\/.*//' %s | onnx-mlir --EmitONNXBasic --printIR | FileCheck %s
```
AlexandreEichenberger, line 1: Maybe it was explained elsewhere, would it be possible to hide the sed within the onnx-mlir pipeline? This feels cumbersome to use.

test/mlir/onnx/parse/test_gridsample.json
AlexandreEichenberger: This solution seems to work only when no --split-file option is in.

utils/onnx2json.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python3
```
AlexandreEichenberger, line 1: Is it used systematically in the pipelines, or is it just a tool?
Is the goal to not have .onnx models in the git, and transform them to json instead? Please provide a bit of context to this.

AlexandreEichenberger (commented):
Receptive to this addition, asked for a few changes and/or clarification.
Because the JSON is much more verbose, I would suggest in the lit test that use JSON to add a short comment stating what the graph is building.

test/mlir/onnx/parse/test_abs.json
sorenlassen: yes, we can remove // comments in onnx-mlir, then we don't need to strip the comments with sed in the lit tests, as I mentioned here:
https://github.com/onnx/onnx-mlir/pull/1685#discussion_r966182472
https://github.com/onnx/onnx-mlir/pull/1685#discussion_r966179380

I went ahead and did that, and then we don't need to read input from stdin anymore

utils/onnx2json.py
sorenlassen: it's just a tool, just for the purpose of constructing parse lit tests

I clarified the purpose in the script's block comment

src/Compiler/CompilerUtils.cpp
sorenlassen: it was just to facilitate chaining json parse lit tests with sed, but now that onnx-mlir ignores // comment lines and I removed sed from the scripts there is no longer any need to read from stdin, so I removed the read-from-stdin logic again

test/mlir/onnx/parse/test_gridsample.json
sorenlassen: right, it only works without --split-input-file

I think that's an ok compromise

as it happens, onnx-mlir doesn't support --split-input-file and all the parse lit tests can only do one onnx model per lit test - I don't see an easy way around that and I propose that we live with that limitation

src/Compiler/CompilerUtils.cpp
AlexandreEichenberger: oh, I like that, thanks

test/mlir/onnx/parse/test_abs.json.mlir
AlexandreEichenberger: > see comment in FrontendDialectTransformer.cpp about changing onnx-mlir to ignore comments

I like that. So comments are ignored for the mlir textual format (which it already does, from my understanding and experience) as well as now from the JSON file? 
I am glad someone will not have errors because of a badly formed cryptic sed command (love sed, but it may not be a generally shared feeling :-) )

test/mlir/onnx/parse/test_gridsample.json
AlexandreEichenberger: But if I understand well, now that onnx-mlir JSON input ignores the comment lines, we can again work normally.

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -1406,21 +1411,42 @@ int ImportFrontendModelArray(const void *onnxBuffer, int size,
...
-  if (!input.is_open()) {
-    *errorMessage = "Unable to open or access " + model_fname;
-    return InvalidInputFileAccess;
-  }
+  if (model_fname.endswith(".json")) {
+    auto buf = openInputFile(model_fname, errorMessage);
+    if (!buf) {
+      return InvalidInputFileAccess;
+    }
+    // Remove // comments, which are non-standard json but appear in lit tests
```
AlexandreEichenberger, line 1423: Lines are removed if the start with `//` ? 
Maybe an error can be returned if you detect JSON lines that have a `//` in them (aka comments that terminate the json line).
Ignore if difficult to do, but maybe be explicit in the comments here on which comments we tolerate and which one we do not (if applicable)

AlexandreEichenberger (approved):
LGTM, since it is about the parser, let's have @chentong319 comment on it too.

AlexandreEichenberger:
@sorenlassen just a high level question: was JSON added just to onnx-mlir-opt or to both (aka also onnx-mlir). Fine either way, just try to understand the big picture

src/Builder/FrontendDialectTransformer.cpp
sorenlassen: yes, lines starting with `//` are ignored now

all the relevant RUN/COM/CHECK/FAILED commands in lit tests appear in lines starting with `//`

end-of-line // comments are very rare in lit tests and I suggest that we don't support them in the parse lit tests

I added a warning printed to stderr if there's a // in a line that doesn't begin with //, this way if you put an end-of-line comment in a json file you will see the warning as additional diagnostic info to go with the json parsing error (or, if the // was part of some valid json you can ignore the false-positive warning)

sorenlassen:
> was JSON added just to onnx-mlir-opt or to both (aka also onnx-mlir)

this PR only adds JSON to onnx-mlir (which parses onnx), not onnx-mlir-opt (which only parses .mlir)

the parse lit tests therefore use onnx-mlir while all the other lit tests use onnx-mlir-opt

test/mlir/onnx/parse/test_gridsample.json
sorenlassen: we still need the workaround (pipe to an error log, semicolon, and then cat the error log) to make the RUN command work cross-platform when onnx-mlir returns a non-zero return code, it's needed in the two lit tests that test that parsing fails as expected, namely `test_gridsample.json` and `zipmap.json`

chentong319 (approved):
LGTM!

AlexandreEichenberger:
> the parse lit tests therefore use onnx-mlir while all the other lit tests use onnx-mlir-opt

if it works, that is great. If we have to work through hoops to use `onnx-mlir` in the lit test, let's consider adding the json format to `onnx-mlir-opt` as well in the future.

PR #1686, opened 2022-09-09
Print signature in presence of operands with  none types

AlexandreEichenberger:
Yolo 4 has ops that have operands with none types. When trying to use `--instrument-onnx-signature` to trace at runtime the operand signatures, the verifier complained as the PrintSignature ONNX op did not support NONE type as one of its input types.

This PR remedy this issue.


tungld (approved):
LGTM.

PR #1692, opened 2022-09-13
Perf noinline

AlexandreEichenberger:
In the test/perf, there were functions that were inlined by #def macros.This may contribute to compile issue on Windows.

This PR transformed the inline macros into functions integrated in a new `PerfLib`, so hopefully this will help Windows... and it is a nicer code anyway.

### DIFF
--- a/test/perf/CMakeLists.txt
+++ b/test/perf/CMakeLists.txt
@@ -46,7 +46,7 @@ endfunction()
 
 # The CompilerUtils ExecutionSession are also included in ModelLib,
 # but it did not compile when I removed these two. TODO, figure out why.
-set(TEST_LINK_LIBS PerfLib ModelLib CompilerUtils benchmark)
+set(TEST_LINK_LIBS benchmark CompilerUtils ModelLib PerfLib)
 
 add_perf_unittest(PerfGemm
   PerfGemm.cpp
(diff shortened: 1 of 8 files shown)


sstamenova:
Thanks! Could you change the windows build to build the tests as well as part of this PR? This would make sure that it's fixed with this change (it's the only known failure in the test build):

https://github.com/onnx/onnx-mlir/blob/6448fbaab09bd0c1e5e9c5dc90278c7a7f037d22/utils/build-onnx-mlir.cmd#L11

`-DONNX_MLIR_BUILD_TESTS=OFF` would go away.

sstamenova (approved):

AlexandreEichenberger:
Changed, hopefully this will work, otherwise will ask you for help on the window build. tx

test/perf/CMakeLists.txt
```diff
@@ -1,5 +1,17 @@
...
 
+add_onnx_mlir_library(PerfLib
+  PerfHelper.cpp
+  EXCLUDE_FROM_OM_LIBS
+
+  INCLUDE_DIRS PRIVATE
+  ${ONNX_MLIR_SRC_ROOT}/third_party/benchmark/include
+  ${ONNX_MLIR_BIN_ROOT}/third_party/benchmark/include
+
+  LINK_LIBS PRIVATE
```
sstamenova, line 11: Since PerfLib is meant to be linked by other test libs, I would make the dependencies (and probably the include dirs) PUBLIC. Then you won't have to add them below since PerfLib will carry them.

AlexandreEichenberger: got it, your input on this is appreciated as cmake magic is still eluding me :-)

sstamenova: It looks like it's the shared libraries build that's failing. Let me try locally and see what the issue is. Probably a couple of hours to get the build all done and figured out, so I should have a fix this afternoon.

PR #1697, opened 2022-09-15
[NNPA] Support cpu-side padding to enable to run onnx.Conv with manual padding on NNPA.

negiyas:
Support cpu-side padding to enable to run onnx.Conv with manual padding on NNPA.

[X] Fix/Create  a lit test for this PR.
[X] Add and confirm numerical test cases specific to this PR.


src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.td
```diff
@@ -946,7 +954,55 @@ def replaceONNXConv2DPattern : Pattern<
...
+    (ONNXPadOp $x,
+       (ONNXConstantOpFromDenseAttr (insertZerosForNonPaddedDims<2> $pads)),
+       (ONNXConstantOpFromDenseAttr (FloatAttrOfValue<0> $res)),
+       (StringAttrOfValue<"constant">),
+       (returnType (NativeCodeCall<"CreatePaddedXType($0, $1)"> $x, $pads))),
+
+     $w, $b, (StringAttrOfValue<"VALID">), $dilation, $group, $kernel_shape,
+     (createArrayAttrOfZerosFrom $pads),
+     $strides),
+  [(HasNonZeroInArrayAttr:$pads), (IsNotStringAttrOfValue<"VALID"> $auto_pad)]
```
tungld, line 1005: Since we do not rewrite ONNXConv to ZHighConv here, I recommend to move this transformation to `RewriteONNXForZHigh.td`.

negiyas: Moved this pattern, its related modifications and its lit tests from ONNXToZHigh.td to RewriteONNXForZHigh.td.
Thank you.

negiyas:
The current code causes the following errors in the "ConvertKrnlToAffinePass" pattern at compilation time in numerical test.

I suppose that code generated by "RewriteONNXForZHighPass" includes issues, but I cannot find issues.
Can someone kindly give me comments or suggestions?

**Test script to reproduce the error**
```
cat << EOF > numericaltest-onnx.mlir
module attributes {llvm.data_layout = "E-m:e-i1:8:16-i8:8:16-i64:64-f128:64-v128:64-a:8:16-n32:64", llvm.target_triple = "s390x-ibm-linux", "onnx-mlir.accels" = ["NNPA-0x10000"]} {
  func.func @main_graph(%arg0: tensor<3x64x55x55xf32>, %arg1: tensor<64x64x1x1xf32>) -> tensor<*xf32> {
    %1 = "onnx.Conv"(%arg0, %arg1, %0) {auto_pad = "NOTSET", dilations = [1, 1], group = 1 : si64, kernel_shape = [1, 1], pads = [1, 1, 1, 1], strides = [1, 1]} : (tensor<3x64x55x55xf32>, tensor<64x64x1x1xf32>, none) -> tensor<*xf32>
    return %1 : tensor<*xf32>
  }
}
EOF
./build/Debug/bin/onnx-mlir -mcpu=z14 --maccel=NNPA --EmitLLVMIR numericaltest-onnx.mlir # --mlir-print-ir-after-all
```

**Error**
```
$ ./debug-cpuconvpadding.sh
onnx-mlir: /home1/negishi/src/dlc.git/llvm-project/mlir/lib/Dialect/Affine/IR/AffineOps.cpp:2563: static void mlir::AffineLoadOp::build(mlir::OpBuilder&, mlir::OperationState&, mlir::Value, mlir::AffineMap, mlir::ValueRange): Assertion `map.getNumInputs() == mapOperands.size() && "inconsistent index info"' failed.
./debug-cpuconvpadding.sh: line 12: 4070594 Aborted                 (core dumped) ./build/Debug/bin/onnx-mlir -mcpu=z14 --maccel=NNPA --EmitLLVMIR numericaltest-onnx.mlir

```

** Code generated by RewriteONNXForZHighPass"
```
module attributes {llvm.data_layout = "E-m:e-i1:8:16-i8:8:16-i64:64-f128:64-v128:64-a:8:16-n32:64", llvm.target_triple = "s390x-ibm-linux", "onnx-mlir.accels" = ["NNPA-0x10000"]} {
  func.func @main_graph(%arg0: tensor<3x64x55x55xf32>, %arg1: tensor<64x64x1x1xf32>) -> tensor<3x64x57x57xf32> {
    %1 = "onnx.Constant"() {value = dense<[0, 0, 1, 1, 0, 0, 1, 1]> : tensor<8xi64>} : () -> tensor<8xi64>
    %2 = "onnx.Constant"() {value = dense<0.000000e+00> : tensor<1xf32>} : () -> tensor<1xf32>
    %3 = "onnx.Pad"(%arg0, %1, %2) {mode = "constant"} : (tensor<3x64x55x55xf32>, tensor<8xi64>, tensor<1xf32>) -> tensor<3x64x57x57xf32>
    %4 = "onnx.Conv"(%3, %arg1, %0) {auto_pad = "VALID", dilations = [1, 1], group = 1 : si64, kernel_shape = [1, 1], pads = [0, 0, 0, 0], strides = [1, 1]} : (tensor<3x64x57x57xf32>, tensor<64x64x1x1xf32>, none) -> tensor<3x64x57x57xf32>
    return %4 : tensor<3x64x57x57xf32>
  }
}

```

tungld:
I tried to compile your .mlir code but didn't see the error:
```
tungld@53315fb59bc7:~/dl/onnx-mlir/build$ cat test.mlir
module attributes {llvm.data_layout = "E-m:e-i1:8:16-i8:8:16-i64:64-f128:64-v128:64-a:8:16-n32:64", llvm.target_triple = "s390x-ibm-linux", "onnx-mlir.accels"
= ["NNPA-0x10000"]} {
  func.func @main_graph(%arg0: tensor<3x64x55x55xf32>, %arg1: tensor<64x64x1x1xf32>) -> tensor<*xf32> {
    %1 = "onnx.Conv"(%arg0, %arg1, %0) {auto_pad = "NOTSET", dilations = [1, 1], group = 1 : si64, kernel_shape = [1, 1], pads = [1, 1, 1, 1], strides = [1, 1]} : (tensor<3x64x55x55xf32>, tensor<64x64x1x1xf32>, none) -> tensor<*xf32>
    return %1 : tensor<*xf32>
  }
}
tungld@53315fb59bc7:~/dl/onnx-mlir/build$ ./Debug/bin/onnx-mlir -mcpu=z14 --maccel=NNPA --EmitLLVMIR test.mlir -v
Full MLIR code written to:
        test.onnx.mlir
```

negiyas:
Fixed. Thank you @tungld!
Type of the third argument of onnx.Pad was wrong. It should not be tensor<1xf32> but tensor<f32>.
```
%3 = "onnx.Pad"(%arg0, %1, %2) {mode = "constant"} : (tensor<3x64x55x55xf32>, tensor<8xi64>, tensor<1xf32>) -> tensor<3x64x57x57xf32>
```

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
```diff
@@ -397,6 +473,10 @@ void RewriteONNXForZHighPass::runOnOperation() {
     return true;
   });
 
+  target.addDynamicallyLegalOp<ONNXConvOp>([](ONNXConvOp op) {
+    return !op.pads() || !hasNonZeroInArrayAttr(op.pads().value());
```
tungld, line 477: NNPA supports several cases of non-zero pads. But it looks like this will rewrite ALL conv ops that have non-zero pads no matter these pads are supported by NNPA or not. So, the later pass `convert-onnx-to-zhigh` has no chance to lower supported non-zero padding op directly to ZHigh. Note that `rewrite-onnx-for-zhigh` is called before `convert-onnx-to-zhigh`.

To avoid that, we need to include in here the legalization condition for Conv in ONNXLegalityCheck.cpp: https://github.com/onnx/onnx-mlir/blob/main/src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp#L778, and maybe more to handle dynamic/static cases.

test/numerical/TestConv.cpp
```diff
@@ -156,6 +160,12 @@ int main(int argc, char *argv[]) {
   assert(isOMConvTheSameAsNaiveImplFor(
              1, 512, 1000, 13, 13, 1, 1, 0, 0, 0, 0, ConvAutoPad::NOTSET) &&
          "failed test from test_squeezenet_cpu");
+  assert(isOMConvTheSameAsNaiveImplFor(
+             3, 64, 64, 55, 55, 1, 1, 1, 1, 1, 1, ConvAutoPad::NOTSET) &&
```
tungld, line 164: Looks like the intention here is for conv ops that can be expressed by matmul, while the two new added test are not.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
negiyas: Sure! I am fixing it.

negiyas: Fixed. Thank you!

test/numerical/TestConv.cpp
negiyas: Changed the condition. Thank you.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
```diff
@@ -397,6 +473,11 @@ void RewriteONNXForZHighPass::runOnOperation() {
     return true;
   });
 
+  target.addDynamicallyLegalOp<ONNXConvOp>([](ONNXConvOp op) {
+    return !op.pads() || !hasNonZeroInArrayAttr(op.pads().value()) ||
+           isSuitableForZDNN<ONNXConvOp>(op);
```
tungld, line 478: Is it enough to use only `isSuitableForZDNN<ONNXConvOp>(op)`? 

Looking at this It seems there is the third case where we cannot run directly on NNPA or cannot do padding for NNPA. In summary, it looks to me there are three cases:
- Conv ops that can run on NNPA directly.
- Conv ops that need to pad to run on NNPA.
- Conv ops do not belong to the above 2 cases and they still run on CPU.

I thought we only had the first two cases. Could you confirm that? Thanks!

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.td
```diff
@@ -257,4 +257,61 @@ def rewriteSoftmaxNDto2D: Pat<
...
+def HasNonZeroInArrayAttr: Constraint<CPred<"hasNonZeroInArrayAttr($_self)">,
+                                       "has non-zero elements">;
+
+// Check that a StrAttr does not contain a specific value.
+class IsNotStringAttrOfValue<string val>:
+  Constraint<CPred<"$0.cast<StringAttr>().getValue() != \"" # val # "\"">>;
+
+// Check the a convolution operation is NOT suitable for zDNN
+def IsConvolutionOpNotSuitableForZDNN:
+//  Constraint<CPred<"!isConvolutionOpSuitableForZDNN($_self)">>;
```
tungld, line 293: It seems this is redundant.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.td
```diff
@@ -257,4 +257,61 @@ def rewriteSoftmaxNDto2D: Pat<
...
+    (ONNXPadOp $x,
+       (ONNXConstantOpFromDenseAttr (insertZerosForNonPaddedDims<2> $pads)),
+       (ONNXConstantOpFromDenseAttr (FloatAttrOfValue<0> $res)),
+       (StringAttrOfValue<"constant">),
+       (returnType (NativeCodeCall<"CreatePaddedXType($0, $1)"> $x, $pads))),
+
+     $w, $b, (StringAttrOfValue<"VALID">), $dilation, $group, $kernel_shape,
+     (createArrayAttrOfZerosFrom $pads),
+     $strides),
+  //[(HasNonZeroInArrayAttr:$pads), (IsNotStringAttrOfValue<"VALID"> $auto_pad)]
```
tungld, line 313: Remove comment?

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.td
```diff
@@ -257,4 +257,61 @@ def rewriteSoftmaxNDto2D: Pat<
...
+  [(HasNonZeroInArrayAttr:$pads), (IsConvolutionOpNotSuitableForZDNN:$res)]
```
tungld, line 314: Still wonder if `HasNonZeroInArrayAttr:$pads` is necessary.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
negiyas: @tungld Thank you for your comments.

Let me explain my understanding.

This rule should be applied only when
(Condition A) Conv ops cannot run on NNPA before applying the rule <AND>
(Condition B) Conv ops can run on NNPA after applying the rule

And
(Condition A) The "(IsConvolutionOpNotSuitableForZDNN:$res)"  condition can be used.
(Condition B) After applying the rule, autopad becomes "VALID", which can be executed on NNPA always.

So that the "(IsConvolutionOpNotSuitableForZDNN:$res)"  condition is enough, as you described.

It's fine, but the current "ConvOpPaddingPattern" rule works onnx.Conv with the non-zero pads attribute only,
because it generates onnx.Pad with the specified "pads" attribute.
So that I supposed that the pattern need not be applied when the pads are not specified or zeros, and
the " !op.pads() || !hasNonZeroInArrayAttr(op.pads().value()" condition is used to stop applying the rule.

I am happy if you confirm my understanding. Thank you.

negiyas: Thank you for the useful comments.

Let me check the three cases according to pads attribute.
- "VALID" and "SAME_UPPER" cases can be supported on NNPA directly always.
- "SAME_LOWER" case is not supported on NNPA at all at this point.
- "NOTSET" case can be supported on NNPA on directly or with cpu-side padding.

The "SAME_LOWER" case may be the third case (with the current implementation).

negiyas: Updated the rule to support all onnx.Conv on NNPA, and remove all conditions except "isSuitableForZDNN<ONNXConvOp>(op)" from "addDynamicallyLegalOp<ONNXConvOp>" in RewriteONNXForZHigh.cpp

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.td
negiyas: Fixed. Thank you.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.td
negiyas: Removed the Comment. Thank you.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.td
negiyas: Removed the condition. Thank you.

test/numerical/TestConv.cpp
```diff
@@ -36,8 +36,8 @@ bool isOMConvTheSameAsNaiveImplFor(const int N, const int CIn, const int COut,
     const int H, const int W, const int kH, const int kW, int pHBegin,
     int pHEnd, int pWBegin, int pWEnd, const ConvAutoPad autoPad) {
   static int testNum = 0;
-  printf("attempt %d with N %d, Cin %d, Cout %d, H %d, W %d, kH %d, kW %d, "
-         "pHBegin %d, pHEnd %d, pWBegin %d, pWEnd %d, autopad %s, isDynamic "
+  printf("attempt %d with N %d, Cin %d, Cout %d, input [%d, %d], "
+         "kernel [%d, %d], pads [%d, %d, %d, %d], autopad %s, isDynamic "
```
tungld, line 40: I like the previous one where we know which pad is for H/W and begin/end.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
```diff
@@ -216,6 +217,94 @@ bool CanExpandPowOpToMul(ONNXPowOp op) {
...
+  SmallVector<int64_t, 4> vals(nElements, 0);
+  return rewriter.getI64ArrayAttr(vals);
+}
+
+// Create Type for Padded input
+Type CreatePaddedXType(Value x, ArrayAttr pads) {
+  RankedTensorType inputType = x.getType().cast<RankedTensorType>();
+  ArrayRef<int64_t> inputShape = inputType.getShape();
+  Type elementType = inputType.getElementType();
+  SmallVector<int64_t, 4> paddingShape = {0, 0, 0, 0};
```
tungld, line 295: Use `SmallVector<int64_t, 4> paddingShape(4, 0)` for brevity.

tungld (approved):
LGTM with the some minor changes. Thank you for addressing my comments!

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
```diff
@@ -216,6 +217,94 @@ bool CanExpandPowOpToMul(ONNXPowOp op) {
...
+// This function is used for padding attribute in Conv.
+ArrayAttr getPadsForNNPAConv(PatternRewriter &rewriter, Value ret) {
+  ONNXConvOp op = dyn_cast<ONNXConvOp>(ret.getDefiningOp());
+  ONNXConvOpAdaptor operandAdaptor = ONNXConvOpAdaptor(op);
+  ONNXConvOpShapeHelper shapeHelper(&op);
+  assert(succeeded(shapeHelper.computeShape(operandAdaptor)));
+  SmallVector<int64_t, 4> vals = {0, 0, 0, 0};
+  for (size_t i = 0; i < shapeHelper.pads.size(); i++) {
+    vals[i] = shapeHelper.pads[i].getLiteral();
+  }
```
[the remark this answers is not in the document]
negiyas: Fixed. Thank you!

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
negiyas: Fixed. Thank you!

test/numerical/TestConv.cpp
negiyas: Fixed. Thank you.

AlexandreEichenberger:
@negiyas folks are reporting that this might generate issues with some benchmarks.

PR #1698, opened 2022-09-15
build: add narrow dependencies on MLIR libraries and tools

ashay:
This patch uses the `dependentDialects` field to specify upstream MLIR
dialects that the ONNX and Krnl dialects depend on.  By doing so, we
avoid the need to include InitAllDialects.h (and link against all MLIR
dialect libraries) in onnx-mlir-reduce.cpp.  A similar change can be
applied to onnx-mlir-opt.cpp, but that change is more involved given the
various conversion passes referenced by the `onnx-mlir-opt` binary.

This patch also adds `llc` and `opt` binaries to the dependency list of
the `CompilerUtils` target, since CompilerUtils.cpp shells out to
those two binaries.

Finally, InitAccelerators.cpp, just like Accelerator.cpp, includes
Accelerators.hpp, which includes mlir/IR/BuiltinOps.h, this requring a
dependency on MLIRIR, although this can probably be trimmed to be
leaner.

Thanks to @sstamenova for suggesting the use of the `dependentDialects` field!


AlexandreEichenberger:
updated to start all CIs.

AlexandreEichenberger:
Added @gongsu832 as reviewer as he does a lot of the build crafting for us. Others may also chime in, of course

AlexandreEichenberger:
DCO procedures need to be followed. Please check link to instructions at bottom of README page.

Also, our CIs will probably not really test this as we continue to fully build LLVM before onnx-mlir, which I think @gongsu832 indicated that he wants to continue doing so.

So you may have to be creative to check that this works on your local environment.

sstamenova:

AlexandreEichenberger:
FYI Error seems to be missing object during the linking in Release mode. Release mode is much more finicky about the libraries and the order in which they are listed (compared to Debug mode).

sstamenova:

gongsu832 (approved):
LGTM

PR #1699, opened 2022-09-15
support external data

### LINKED ISSUE #1696: Support for models with External Data
We briefly discussed this issue in the `onnx-mlir` Monday meeting. I wanted to add some more background to the problem, and create an issue so that we can track the progress

Its possible that an ONNX model stores some of its data externally https://github.com/onnx/onnx/blob/main/docs/ExternalData.md

As of right now there is no support for this in `onnx-mlir`.

sorenlassen:
closes #1696

the onnx parser now reads external data if any tensors have data_location=EXTERNAL

the implementation of `onnxTensorProtoToDenseElmAttr` in `src/Builder/FrontendDialectHelper.cpp` was rewritten to add this logic, plus made the following improvements:
- added support for uint16, uint32, uint64 data types
- removed many memory copies, resulting in almost 3x speedup: on my macbook the time to parse the 570MB vgg19 model with `onnx-mlir --EmitONNXBasic` went from 52s to 19s

added `utils/onnxExternalizeData.py` which was used for manual testing: I took all the models from `third_party/onnx/onnx/backend/test/data/real` (e.g. vgg19) and made their data external in different ways:
```
utils/onnxExternalizeData.py vgg19/model.onnx
utils/onnxExternalizeData.py --no_all_tensors_to_one_file   vgg19/model.onnx
utils/onnxExternalizeData.py --no_convert_attribute  vgg19/model.onnx
utils/onnxExternalizeData.py --no_all_tensors_to_one_file --no_convert_attribute  vgg19/model.onnx
```
and for each ran `onnx-mlir --EmitONNXBasic vgg19/model.onnx` and checked that the resulting `vgg19/model.onnx.mlir` is the same

also added three parse lit tests to test raw, nonraw, and external data represenations of tensors

added utils/testing/x_data.py and utils/testing/zipmap.py scripts used to generate parse lit tests


src/Builder/FrontendDialectHelper.cpp
```diff
@@ -11,232 +11,197 @@
...
-        tensorType, llvm::SmallVector<bool, 64>(data.begin(), data.end()));
-    break;
-  }
+  llvm::ArrayRef<int64_t> tensorDims(tp.dims().data(), tp.dims().size());
+  mlir::Type elmType = convertONNXTypeToMLIRType(
+      builder, (onnx::TensorProto_DataType)tp.data_type());
+  auto tensorType = mlir::RankedTensorType::get(tensorDims, elmType);
+  switch (tp.data_type()) {
+  case (onnx::TensorProto::FLOAT):
+    return createDenseElmAttr<float>(externalDataDir, tp, tensorType);
```
AlexandreEichenberger, line 181: Nice use of templates to streamline the code

AlexandreEichenberger (commented):
Asked @tungld to review it too since he did work on constants and @chentong319 since he is doing work on the frontend.

Looks great to me. Would you mind highlighting with a comment where you cut down the copies. Might help better understand where you removed them. Tx

src/Builder/FrontendDialectHelper.cpp
```diff
@@ -11,232 +11,203 @@
...
+      offset = parseOffsetOrLength(entry.value());
+    } else if (entry.key() == "length") {
+      length = parseOffsetOrLength(entry.value());
+    }
+  }
+  assert(!location.empty() && "missing external data location");
+  llvm::SmallVector<char> path(externalDataDir.begin(), externalDataDir.end());
+  llvm::sys::path::append(path, location);
+  auto bufferOrError = llvm::MemoryBuffer::getFileSlice(
+      path, length, offset, /*IsVolatile=*/false);
```
sorenlassen, line 55: llvm will in some cases create a memory buffer that points to the file content by mmap and thereby read without memory copies, just a small heap allocation of the MemoryBuffer metadata

src/Builder/FrontendDialectHelper.cpp
```diff
@@ -11,232 +11,203 @@
...
+  } else {
+    // Not raw, no need to take care of endianness.
+    const auto &data = TransformValueToONNXData<T>::data(tp);
+    // Access data directly via ArrayRef if same size as T,
+    // or copy into correctly typed SmallVector otherwise
+    // because DenseElementsAttr needs argument type of the correct bitwidth.
+    typedef typename std::conditional<sizeof(T) == sizeof(data[0]),
+        llvm::ArrayRef<T>, llvm::SmallVector<T>>::type ArrayRefOrSmallVector;
+    ArrayRefOrSmallVector array(data.begin(), data.end());
+    return denseBuilder(llvm::makeArrayRef(array));
```
sorenlassen, line 156: avoiding a memory copy here in the cases where T has the same size as data's type's value_type, namely for float, double, i32, i64

the cases that have to do a memory copy via SmallVector are bool, i8, u8, i16, u16 from tp.int32_data, and u32 from tp.uint64_data

src/Builder/FrontendDialectHelper.cpp
```diff
@@ -11,232 +11,203 @@
...
-    auto tensorType = mlir::RankedTensorType::get(tensorDims, elmType);
-    denseElmAttr = mlir::DenseElementsAttr::get(
-        tensorType, llvm::SmallVector<bool, 64>(data.begin(), data.end()));
-    break;
-  }
+  llvm::ArrayRef<int64_t> tensorDims(tp.dims().data(), tp.dims().size());
+  mlir::Type elmType = convertONNXTypeToMLIRType(
+      builder, (onnx::TensorProto_DataType)tp.data_type());
+  auto tensorType = mlir::RankedTensorType::get(tensorDims, elmType);
+  auto denseBuilder = [tensorType](auto arrayRef) {
```
sorenlassen, line 182: the use of `auto` is critical here because it's a polymorphic lambda that takes on different types in the different places it is used in the switch statement below

sorenlassen:
> Would you mind highlighting with a comment where you cut down the copies

done

there are more copies that can be removed in FrontendDialectTransformer.cpp, I'll try to address those in a follow-up PR

sorenlassen:
FYI, after this PR lands I plan to rebase PR #1701 to support float16

AlexandreEichenberger (approved):
LGTM, appreciated discovering where we had all the copies... we must use the lesson learned here, for other big buffers.

Lets wait for one other review if you don't mind, looks good from my perspective

tungld:
LGTM without any objection on constants.

Just a small comment, @sorenlassen do you plan to allow users specifying an external data directory or not? It seems ONNX supports that situation.

sorenlassen:
> do you plan to allow users specifying an external data directory or not? It seems ONNX supports that situation.

I'm not planning to but I'll be happy to add that if anyone finds a use for it

philass:
@AlexandreEichenberger can we merge this?

I believe there is a follow up PR for fp16 ingestion that needs this to land first

philass:
I'm happy to roll it back if anyone identifies any issues

PR #1704, opened 2022-09-18
Ingest all of Opset 16 

### LINKED ISSUE #1201: Upgrade to ONNX opset-16
Hi All,

ONNX 1.11.0 was released recently, it has bf16 type supported added to some of the existing ops like Where, GreaterOrEqual, LessOrEqual. I was wondering if there is a plan to take a dependency on the latest ONNX release?

philass:
Closes #1201 

ScatterElements adds a DefaultStrAttribute (therefor it doesn't need the introduction of a new versioned op)

RoiAlign adds a DefaultStrAttribute (which is defaulted to the old behaviour)

And finally GridSample ingestion was added. It still needs a shape inference implementation. I test was deleted that checked that the parsing of gridSample failed. It could be deleted because GridSample is now ingestible


src/Transform/ONNX/Decompose.td
```diff
@@ -49,6 +49,8 @@ def GetNullIntegerAttr : NativeCodeCall<"IntegerAttr()">;
 
 def GetNullStringAttr : NativeCodeCall<"StringAttr()">;
 
+def GetNoneStringAttr : NativeCodeCall<"$_builder.getStringAttr(\"none\")">;
```
chentong319, line 52: Different from GetNullStringAtt?

philass: Good observation!

I've updated the code to use `getNullStringAttr` instead

src/Dialect/ONNX/ONNXOps.td.inc
```diff
@@ -5277,7 +5323,8 @@ def ONNXScatterElementsOp:ONNX_Op<"ScatterElements",
   let arguments = (ins AnyTypeOf<[TensorOf<[UI8]>, TensorOf<[UI16]>, TensorOf<[UI32]>, TensorOf<[UI64]>, TensorOf<[I8]>, TensorOf<[I16]>, TensorOf<[I32]>, TensorOf<[I64]>, TensorOf<[BF16]>, TensorOf<[F16]>, TensorOf<[F32]>, TensorOf<[F64]>, TensorOf<[StringType]>, TensorOf<[I1]>, TensorOf<[Complex<F32>]>, TensorOf<[Complex<F64>]>]>:$data,
     AnyTypeOf<[TensorOf<[I32]>, TensorOf<[I64]>]>:$indices,
     AnyTypeOf<[TensorOf<[UI8]>, TensorOf<[UI16]>, TensorOf<[UI32]>, TensorOf<[UI64]>, TensorOf<[I8]>, TensorOf<[I16]>, TensorOf<[I32]>, TensorOf<[I64]>, TensorOf<[BF16]>, TensorOf<[F16]>, TensorOf<[F32]>, TensorOf<[F64]>, TensorOf<[StringType]>, TensorOf<[I1]>, TensorOf<[Complex<F32>]>, TensorOf<[Complex<F64>]>]>:$updates,
-    DefaultValuedAttr<SI64Attr, "0">:$axis);
+    DefaultValuedAttr<SI64Attr, "0">:$axis,
+    DefaultValuedStrAttr<StrAttr, "none">:$reduction);
```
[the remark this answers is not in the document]
philass: Thanks for the pointer

philass:
@chentong319 Thanks for the review.

I've updated the PR with your suggestions. Please let me know if you would any other changes

chentong319 (changes requested):
Could you add a test case into test/mlir/onnx/onnx_decompose.td?

philass:
@chentong319,

The following test in `test/mlir/onnx/onnx_decompose.mlir` already tests this rule

https://github.com/onnx/onnx-mlir/blob/35b424c84d12e9196da0ec9d3194e67e6a76c18e/test/mlir/onnx/onnx_decompose.mlir#L403-L414

I'm not exactly sure what another test for `ScatterND` would add in terms of coverage

chentong319:
Oh, I forgot that you just updated the rule, not add a new one.

chentong319 (approved):

chentong319 (approved):
LGTM!

philass:
Thanks for the review! Much appreciated

PR #1706, opened 2022-09-19
Fix remaining warnings on Windows and turn on warning-as-error

sstamenova:
This change applies the remaining fixes needed for warnings to be treated as errors on Windows and then turns them on on Windows only. Of particular interest are a couple of changes:

- The `check-docs` target needs an update to properly parse the directive even when it contains parenthesis in the path. For example: `C:/Program Files (x86)/Microsoft Visual Studio/Shared/Python39_64/python.exe`
- The Windows CI build now builds `onnx` first and then builds all of the targets that are enabled in `onnx-mlir`. This will make sure that no targets get missed
- The remaining warnings (4927) in `ConstProp.cpp` are disabled because they will require a comprehensive change to `ConstProp.td`. I'll create an issue to track this.


sstamenova:
@AlexandreEichenberger : All the warnings on Windows are fixed and I've set up the Windows CI to enable werror. I tried to do it for other platforms also, but there are failures on Ubuntu with gcc (I use clang and the clang build is green), so I backed that part of the change out.

tungld:
> The remaining warnings (4927) in ConstProp.cpp are disabled because they will require a comprehensive change to ConstProp.td. I'll create an issue to track this.

I have no idea why it happens, but try to guess. Could you try this change to see if it can remove the warning:
```diff
--- a/src/Transform/ONNX/ConstProp.cpp
+++ b/src/Transform/ONNX/ConstProp.cpp
@@ -351,7 +351,7 @@ void IterateConstPropElementwiseBinary(char *lhs, char *rhs,
 /// Do element-wise binary calculation of 'lhs' and 'rhs' values and create an
 /// ONNXConstantOp for the result.
 template <typename ElementwiseBinaryOp>
-ONNXConstantOp ConstPropElementwiseBinary(
+Value ConstPropElementwiseBinary(
     PatternRewriter &rewriter, Value replacingValue, Value lhs, Value rhs) {
   Type elementType =
       replacingValue.getType().cast<ShapedType>().getElementType();
@@ -383,7 +383,7 @@ ONNXConstantOp ConstPropElementwiseBinary(
   ONNXConstantOp res =
       createConstantOpAndStoreBufferPtr(rewriter, replacingValue, resArray);

-  return res;
+  return res.getResult();
 }
```

sstamenova:
@AlexandreEichenberger: I'd like to get this committed sooner rather than later, so that we don't have to deal with new warnings being introduced in the mean time.

tungld (approved):
LGTM.

Issue #1710, opened 2022-09-20
ShapeOp whose start is negative is failed when doing shape inference and lowering to Krnl


tungld:
I tested this code for ShapeOp whose start is negative:
```mlir
module {
  func.func @test_shape_to_dim_positive_axis(%arg0: tensor<?x256x?xi64>) -> tensor<2xi64> {
    %0 = "onnx.Shape"(%arg0) {end = 0 : si64, start = -1 : si64} : (tensor<?x256x?xi64>) -> tensor<2xi64>
    return %0 : tensor<2xi64>
  }
}
```

It failed when doing shape inference: `./Release/bin/onnx-mlir-opt shape.mlir --shape-inference`
```
onnx-mlir-opt: /home/tungld/dl/onnx-mlir/src/Dialect/Mlir/IndexExpr.cpp:407: static void onnx_mlir::IndexExpr::getShape(llvm::SmallVectorImpl<onnx_mlir::IndexE
xpr>&, llvm::SmallVectorImpl<long int>&): Assertion `val >= 0 && "expected positive values only"' failed.
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
Stack dump:
0.      Program arguments: ./Release/bin/onnx-mlir-opt shape.mlir --shape-inference
 #0 0x000003ffacf9fce0 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) (/home/tungld/dl/llvm-project/build/lib/libLLVMSupport.so.16git+0x21fce0)
 #1 0x000003ffacf9db70 llvm::sys::RunSignalHandlers() (/home/tungld/dl/llvm-project/build/lib/libLLVMSupport.so.16git+0x21db70)
 #2 0x000003ffacf9dd5c SignalHandler(int) Signals.cpp:0:0
 #3 0x000002aa16d1210e
 #4 0x000003ffac8bdf44 raise /build/glibc-6qIgzY/glibc-2.27/signal/../sysdeps/unix/sysv/linux/raise.c:51:0
 #5 0x000003ffac8bf3ca abort /build/glibc-6qIgzY/glibc-2.27/stdlib/abort.c:81:0
 #6 0x000003ffac8b5f44 __assert_fail_base /build/glibc-6qIgzY/glibc-2.27/assert/assert.c:89:0
 #7 0x000003ffac8b5fc4 (/lib/s390x-linux-gnu/libc.so.6+0x35fc4)
[...]

PR #1711, opened 2022-09-20
Add Shape op verifier

philass:
I don't think the attributes provided constituted a valid shape op instance. 

But it should have failed more gracefully. I've added a verifier for shape op. As well as a test for it.

I also have updated the docs to reflect that lowering doesn't support slices. 

Closes https://github.com/onnx/onnx-mlir/issues/1710

Thanks for flagging this for me @tungld


tungld (approved):
LGTM. Thanks for your quick response!

PR #1713, opened 2022-09-20
Onnx additional ops cleanup

AlexandreEichenberger:
Moved all the ONNX ops that we add for convenience into the AdditionalONNXOps.td.
Labeled all the additional ops in the detailed comment as "non-standard, added for onnx-mlir".
Removed unneeded `"` added to the .td description files. Because of that change, every description lines are trivially changed.

Cleaned up a bit the gen_onnx_mlir.py's comments. There is no functional changes there except for removing the unneeded `"` for md descriptions.

There were also updates from the `.td` files that were not reflected in the `.md` files; there were synchronized here. That was the case for krnl and high/zlow dialect. This is a reminder that when we change the `.td` files, we have to run the two commands below
```shell
make OMONNXOpsIncTranslation
make onnx-mlir-docs
```
where the first command recreate the `.td` files, and the second recreate the `.md` files from the `.td` files. 

If someone could add dependences to remake all of these when the td files or `gen_onnx_mlir.py` change, that might be useful (though these commands generate files that are under the git `src` and `docs` directories. Think we did not automatize these command because we want to keep manual control of generated files into the git tree.


AlexandreEichenberger:
Fixed similar formatting issues for the ZHigh and ZLow dialects

chentong319 (approved):
LGTM!

PR #1715, opened 2022-09-20
Refine memory management for SeqType

chentong319:
KrnlSeqAlloc and KrnlSeqDealloc are introduced to support deep deallocation for the sequence type.
The lowering of onnx sequence ops are changed according to the new design, which is described in docs/SequenceType.md.
The new design guarantees there is no memory leak or dangling pointer for operations on onnx sequence type.


docs/SequenceType.md
```diff
@@ -0,0 +1,318 @@
...
+
+### Construct a new sequence from an old sequence
+The sequence ops, SequenceInsert and SequenceErase, construct a new sequence
+with the elements from an old sequence. It is fine to use KrnlSeqExtractOp to get an
+element from the old sequence and use KrnlSeqStoreOp to store it into the new sequence.
+But there are two copy operations for the element. Since it is known that the loaded
+element is only used here while the old sequence is still live, a regular memref.load 
+can be used to avoid one copy. This is a simple optimization for sequence lowering.
+
+## Example
```
AlexandreEichenberger, line 134: Would it make sense to introduce each of the seq op and describe their inputs... I find it a bit harder to follow the examples without knowing the semantic, at a high level.

Also, could you describe at a high level what the function does.... "insert addition to an input sequence at position X...."

docs/SequenceType.md
```diff
@@ -0,0 +1,318 @@
...
+      %22:3 = krnl.get_induction_var_value(%2#0, %2#1, %2#2) : (!krnl.loop, !krnl.loop, !krnl.loop) -> (index, index, index)
+      ...
+      %23 = krnl.load %arg1[%22#0, %22#1, %22#2] : memref<3x4x5xf32>
+      %24 = krnl.load %arg1[%22#0, %22#1, %22#2] : memref<3x4x5xf32>
+      %25 = arith.addf %23, %24 : f32
+      krnl.store %25, %1[%22#0, %22#1, %22#2] : memref<3x4x5xf32>
+    }
+
+    // Sequence Insert
+    %6 = "krnl.seqalloc"(%5) : (index) -> memref<?xmemref<?x4x5xf32>>
```
AlexandreEichenberger, line 168: %4, %5 are not defined here. Think the examples does not really need the add, you could just insert arg1 directly in the list arg0.

docs/SequenceType.md
```diff
@@ -0,0 +1,318 @@
...
+    %c0_8 = arith.constant 0 : index
+    %7 = krnl.define_loops 1
+    krnl.iterate(%7) with (%7 -> %arg2 = 0 to %4){
```
AlexandreEichenberger, line 171: Could you add comments in the code to describe what each loops are doing?

docs/SequenceType.md
```diff
@@ -0,0 +1,318 @@
...
+    krnl.store %19, %17[%c0_16] : memref<3xi64>
+    krnl.store %20, %17[%c1_17] : memref<3xi64>
+    krnl.store %21, %17[%c2] : memref<3xi64>
+
+    return %17 : memref<3xi64>
+  }
+```
+
+After --buffer-deallocation pass
+```
```
AlexandreEichenberger, line 204: Can you add a sentence to indicate what we should be looking at?

AlexandreEichenberger (commented):
@chentong319 added comments first to the md document. Will review the rest later today

docs/SequenceType.md
chentong319: I did not include the whole output.  %4 and %5 is the result of dim of op. The onnx.add is used to show the deallocate of sequence element. If it is an argument of function, no memref.dealloc will be added

chentong319:
@AlexandreEichenberger I made changes according to your comments.

AlexandreEichenberger (approved):
LGTM

PR #1716, opened 2022-09-21
Support other element types rather than float32 in RunONNXModel.py

### LINKED ISSUE #1608: Support different data types rather than float32 in RunONNXModel.py
`RunONNXModel.py` script is able to generate random data input for a model. However, it uses float32 as data type by default. Generating random data should check the input types and use the correct type.

tungld:
Resolves #1608 

`RunONNXModel.py` script is able to generate random data input for a model. However, it uses float32 as data type by default. This patch adds support for other data types.


AlexandreEichenberger (approved):
LGTM. 

Just to clarify, now integer are supported. Same with float, is that only one size?

chentong319 (approved):
LGTM!

tungld:
> Just to clarify, now integer are supported. Same with float, is that only one size?

No, that's for all size. By `np_elem_type = TENSOR_TYPE_TO_NP_TYPE[elem_type]`, ONNX function `TENSOR_TYPE_TO_NP_TYPE` will produce the corresponding numpy type for each ONNX element type.

PR #1720, opened 2022-09-21
Add a debug windows CI build

sstamenova:
I expect this will be too slow for us to add to CI, but this will test out the timing.


sstamenova:
@AlexandreEichenberger : Looks like we can't setup Debug builds right now, because the available agents don't have enough memory: https://dev.azure.com/onnx-pipelines/onnx/_build/results?buildId=30338&view=logs&j=1c607959-0f5a-56cf-c8f0-3ee944da09dc&t=980d2cda-d837-57ef-c98d-579f7383e11b

PR #1723, opened 2022-09-22
[Docs] Recommend ninja to build backend tests(NFC)

xgupta:


docs/BuildONNX.md
```diff
@@ -2,7 +2,7 @@
 
 # Installing `third_party ONNX` for Backend Tests or Rebuilding ONNX Operations
 
-Backend tests are triggered by `make check-onnx-backend` in the build directory and require a few preliminary steps to run successfully. Similarly, rebuilding the ONNX operations in ONNX-MLIR from their ONNX descriptions is triggered by `make OMONNXOpsIncTranslation`.
+Backend tests are triggered by `ninja check-onnx-backend` in the build directory and require a few preliminary steps to run successfully. Similarly, rebuilding the ONNX operations in ONNX-MLIR from their ONNX descriptions is triggered by `ninja OMONNXOpsIncTranslation`.
```
chentong319, line 5: Could we keep both ninja and make in document?

xgupta: I am not sure I understand why we want to keep both. Other documents like BuildOnWindows.md and BuildOnLinuxOSX.md uses only ninja.

gongsu832: Because not everybody is familiar with ninja. And most systems don't come with ninja pre-installed. Other documents should have been updated with both.

xgupta: Sorry I am not sure if the change worth it. I will drop the pr. Thank you for looking into it. Appreciate your time.

PR #1724, opened 2022-09-22
[Docs] Suggest to build in release mode with assertion on Linux

xgupta:


AlexandreEichenberger:
updated to latest and restarted the CIs

AlexandreEichenberger:
Can you just explain the rational, is it also done for other LLVM projects... tx

xgupta:
> Can you just explain the rational, is it also done for other LLVM projects... tx

Oh yeah, I thought again about it and I think It would be fine to keep the build options as it is for onnx-mlir.

Issue #1722, opened 2022-09-22
[NNPA] Yolov3-12 model cannot be compiled for NNPA by the latest main branch.


negiyas:
Yolov3-12 model cannot be compiled for NNPA by the latest main branch.

**Model:**
https://github.com/onnx/models/blob/main/vision/object_detection_segmentation/yolov3/model/yolov3-12.onnx
Downloaded by `wget https://github.com/onnx/models/blob/main/vision/object_detection_segmentation/yolov3/model/yolov3-12.onnx?raw=true -O yolov3-12.onnx`

**onnx-mlir version:**
commit 489a04f5c8f3b350dafdf4372157dcda22e8f8c9 (HEAD -> main, upstream/main, upstream/HEAD)
Date:   Thu Sep 22 03:03:07 2022 +0900

**Command to reproduce this issue**
```
$ ./build/Debug/bin/onnx-mlir -mcpu=z14 --maccel=NNPA --EmitZHighIR  --shapeInformation='0:1x3x416x416,1:1x2' yolov3-12.onnx
error: 'zhigh.Mul' op requires the same layout for all operands and results
```
**onnx.Mul ops in ONNXIR**
```
$ ./build/Debug/bin/onnx-mlir -mcpu=z14 --maccel=NNPA --EmitONNXIR  --shapeInformation='0:1x3x416x416,1:1x2' yolov3-12.onnx
(3.9.1_dlc) negishi@zedctest:/home1/negishi/src/dlc.git/MODELS$ grep onnx.Mul yolov3-12.tmp
    %319 = "onnx.Mul"(%318, %312) {onnx_node_name = "TFNodes/yolo_evaluation_layer_1/mul_6"} : (tensor<?x?x?x?x?xf32>, tensor<?x?x?x?x?xf32>) -> tensor<?x?x?x?x?xf32>
    %329 = "onnx.Mul"(%327, %328) {onnx_node_name = "TFNodes/yolo_evaluation_layer_1/mul_1"} : (tensor<?x?x?x?x?xf32>, tensor<1x1x1x3x2xf32>) -> tensor<?x?x?x3x2xf32>
    %340 = "onnx.Mul"(%1, %339) {onnx_node_name = "TFNodes/yolo_evaluation_layer_1/mul_8"} : (tensor<2xf32>, tensor<f32>) -> tensor<2xf32>
[...]

PR #1726, opened 2022-09-22
Fix issues to compile Yolov3-12

negiyas:
Fix issue https://github.com/onnx/onnx-mlir/issues/1722 (Yolov3-12 model cannot be compiled for NNPA by the latest main branch).

Stop conversion to enable onnx.LeakyRelu on NNPA when the stickified layouts of input and output are not the same.
(Most onnx.LeakyRelu of Yolov3-12 are still enabled on NNPA with this PR.)


tungld (approved):
LGTM. Thanks for a quick fix!

By the way, I am a bit confused when reading `IsPlusConstantFloat`.  Though it looks correct, how about changing to `IsPositiveConstantFloat` to avoid confusion.

PR #1727, opened 2022-09-22
Update onnx-mlir product version

hamptonm1:
I have added the product version number (--DONNX_MLIR_PRODUCT_VERSION flag) for not only the compiler output when doing a `cmake -G Ninja` but also for the actual version output `onnx-mlir --version`. I also updated the syntax and formatting of the output. The output results are shown below:

compiler: 
```
-- Accelerators             : NONE
-- Onnx version             : 1.11.0
-- Product version          : 4.0.0-zdlc
-- Tests optimization level : 3
-- Parallel backend tests   : ON
-- JNI backend tests        : ON
-- JSONITER_JAR             : /usr/share/java/jsoniter-0.9.23.jar
-- Accelerator lit tests    : NONE
-- Configuring done
-- Generating done
```

onnx-mlir --version:

```
root@67b991b7e17f:~/onnx-mlir/build# onnx-mlir --version
onnx-mlir version 0.3.1, onnx version 1.11.0 (https://github.com/hamptonm1/onnx-mlir.git 49f1f1e1db380b69abfdeb9ce455e8b99e0205fb)
onnx-mlir product version 4.0.0-zdlc
LLVM version 16.0.0git (https://github.com/llvm/llvm-project.git 59548fe873d8d98e359fb21fbb2a0852fed17ff5)
DEBUG build with assertions.
Default target: x86_64-unknown-linux-gnu
Host CPU: skylake

```


AlexandreEichenberger:
@hamptonm1 please familiarize yourself with the DCO procedures (main readme, contributing section)

AlexandreEichenberger:

AlexandreEichenberger:
You need to make sure the DCO procedure are followed. See contributing in README.md from onnx.mlir

AlexandreEichenberger:
You also need to follow the formatting from onnx-mlir.

hamptonm1:
> @hamptonm1 please familiarize yourself with the DCO procedures (main readme, contributing section)

Yes I did.... will make the updates for the DCO procedures. Please ignore the notifications.

src/Version/Version.cpp
```diff
@@ -89,8 +89,16 @@ std::string getOnnxMlirFullVersion(bool toIncludeLLVM) {
 #endif
   std::string repo = getOnnxMlirFullRepositoryVersion(toIncludeLLVM);
   if (!repo.empty()) {
-    os << " " << repo;
+    os << " " << repo << '\n';
   }
+#ifdef ONNX_MLIR_PRODUCT_VERSION
+  os << "onnx-mlir product version " ONNX_MLIR_PRODUCT_VERSION;
```
cjvolzka, line 95: Since this is to be used by vendors, we should make come before the rest of the version information. So something like:
```
onnx-mlir product version 4.0.0-zdlc, onnx-mlir version 0.3.1, onnx version 1.11.0 (https://github.com/hamptonm1/onnx-mlir.git 49f1f1e1db380b69abfdeb9ce455e8b99e0205fb)
LLVM version 16.0.0git (https://github.com/llvm/llvm-project.git 59548fe873d8d98e359fb21fbb2a0852fed17ff5)
DEBUG build with assertions.

Comment (cjvolzka) [inline src/Version/Version.cpp:95]: Instead of hardcoding "onnx-mlir product version" this should use the `ONNX_MLIR_VENDOR` (which works out with the earlier comment of moving it up in the order). So maybe something like:
```
#if defined(ONNX_MLIR_VENDOR) || defined(ONNX_MLIR_PRODUCT_VERSION)
os << ONNX_MLIR_VENDOR << " " << ONNX_MLIR_PRODUCT_VERSION << ", "
#endif
```

(opinion) I used an OR because I think if someone specifies one, they'll likely want both in the end. I figured OR will make it easier because if you initially/accidentally only set one, at least they'd see the one they did set. Otherwise we could do an AND but then it wouldn't be obvious (without looking at the code) to understand why their change isn't showing up if they did set just one initially.

I wouldn't be against a more complicated statement that better handles one or the other alone but my gut thought is that might be overkill. I'm open to other thoughts, that's just my initial feelings.

Comment (hamptonm1) [reply src/Version/Version.cpp:95]: Sure! I can move that around. I was just following the output of the compiler which has onnx-mlir version first than the product version.

Comment (hamptonm1) [reply src/Version/Version.cpp:95]: @cjvolzka The only thing is having onnx-mlir version and onnx-version with the GitHub url is extremely long.... I think adding onnx-mlir product version on the same line will render awkwardly. I can test it out and see but I know content will probably be entered on the following line.

Comment (hamptonm1) [reply src/Version/Version.cpp:95]: I see your thought process... I would lean towards using `&&` because we would want the vendor name and the product version to be displayed and not one or the other. It some what forces the user to have to specify both flags because I can see someone indicating one and not the other even by accident.

Comment (cjvolzka) [reply src/Version/Version.cpp:95]: > It some what forces the user to have to specify both flags because I can see someone indicating one and not the other even by accident.

This is where I was waffling myself. On the one hand, it'd make sure both are specified. On the other hand, if someone only specifies one, it won't be obvious why it has no effect unless they dig into the code so it might take a while to debug what is a 5 second fix on their end.

If we go with `&&` is there a way to make it a build error if one is specified and not the other? Then we could give a useful error.

Comment (hamptonm1) [reply src/Version/Version.cpp:95]: Okay this is what it looks like:

```
root@d7f7476a247e:~/onnx-mlir/build# onnx-mlir --version
IBM Z Deep Learning Compiler 4.0.0-zdlc, onnx-mlir version 0.3.1, onnx version 1.11.0 (https://github.com/hamptonm1/onnx-mlir.git 40cdc98e1f4062fea9cf84e6a57c01dbef93f4c2)
 LLVM version 16.0.0git (https://github.com/llvm/llvm-project.git 59548fe873d8d98e359fb21fbb2a0852fed17ff5)
 DEBUG build with assertions.
 Default target: x86_64-unknown-linux-gnu
 Host CPU: skylake
```

As you can see the GitHub url is appended to the next line, so I am thinking what about this instead:

```
root@d7f7476a247e:~/onnx-mlir/build# onnx-mlir --version
IBM Z Deep Learning Compiler 4.0.0-zdlc
onnx-mlir version 0.3.1, onnx version 1.11.0 (https://github.com/hamptonm1/onnx-mlir.git40cdc98e1f4062fea9cf84e6a57c01dbef93f4c2)
LLVM version 16.0.0git (https://github.com/llvm/llvm-project.git 59548fe873d8d98e359fb21fbb2a0852fed17ff5)
DEBUG build with assertions.
Default target: x86_64-unknown-linux-gnu
Host CPU: skylake
```

Comment (cjvolzka) [reply src/Version/Version.cpp:95]: I'd be good with that. I'd actually be fine with putting onnx-mlir and onnx also on their own lines too. 

I'm unsure if that was discussed in the original PR that added VENDOR/VERSION support and rejected or just not brought up then.

Comment (hamptonm1) [reply src/Version/Version.cpp:95]: Hmm I might have an alternative idea but you can let me know your thoughts once I push the code updates :)

File: src/Version/Version.cpp [hunk 2]
```diff
@@ -82,14 +82,23 @@ std::string getOnnxMlirFullVersion(bool toIncludeLLVM) {
...
   os << ONNX_MLIR_VENDOR;
+#else
+  os << "onnx-mlir product version";
+#endif
+#ifdef ONNX_MLIR_PRODUCT_VERSION
+  os << " " << ONNX_MLIR_PRODUCT_VERSION << '\n';
+#else
+  os << " " << PRODUCT_VERSION_MAJOR << '.';
+  os << PRODUCT_VERSION_MINOR << '.';
+  os << PRODUCT_VERSION_PATCH << '-' << PRODUCT_ID << '\n';
```
cjvolzka, line 93: Aren't the variables in this ELSE derived from `ONNX_MLIR_PRODUCT_VERSION`? I'm not sure if it's expected for them to be set independently of that. So what I think would happen is if `ONNX_MLIR_PRODUCT_VERSION` is unset (which it will be for community builds) then you'll always come down this ELSE and print `0.0.0-NOT SPECIFIED` (if I'm reading the flow correctly in https://github.com/onnx/onnx-mlir/blob/main/src/Compiler/CMakeLists.txt)

src/Version/Version.cpp
hamptonm1: Okay sounds good to me!

src/Version/Version.cpp
hamptonm1: I think I resolved everything from our last conversation.... let me know your thoughts now.

hamptonm1: This is what it looks like when the vendor and product number is specified:

```
root@d7f7476a247e:~/onnx-mlir/build# onnx-mlir --version
IBM Z Deep Learning Compiler 4.0.0-zdlc
onnx-mlir version 0.3.1, onnx version 1.11.0 (https://github.com/hamptonm1/onnx-mlir.git 40cdc98e1f4062fea9cf84e6a57c01dbef93f4c2)
LLVM version 16.0.0git (https://github.com/llvm/llvm-project.git 59548fe873d8d98e359fb21fbb2a0852fed17ff5)
DEBUG build with assertions.
Default target: x86_64-unknown-linux-gnu
Host CPU: skylake
```

hamptonm1:
This is without the vendor and product number specified:

```
root@d7f7476a247e:~/onnx-mlir/build# onnx-mlir --version
onnx-mlir version 0.3.1, onnx version 1.11.0 (https://github.com/hamptonm1/onnx-mlir.git 40cdc98e1f4062fea9cf84e6a57c01dbef93f4c2)
LLVM version 16.0.0git (https://github.com/llvm/llvm-project.git 59548fe873d8d98e359fb21fbb2a0852fed17ff5)
DEBUG build with assertions.
Default target: x86_64-unknown-linux-gnu
Host CPU: skylake
```

src/Version/Version.cpp
```diff
@@ -118,13 +115,18 @@ std::string getOnnxMlirFullVersion(bool toIncludeLLVM) {
 #endif
 
 void getVersionPrinter(llvm::raw_ostream &os) {
+#if defined(ONNX_MLIR_VENDOR) && defined(ONNX_MLIR_PRODUCT_VERSION)
+  os << ONNX_MLIR_VENDOR << " " << PRODUCT_VERSION_MAJOR << '.';
+  os << PRODUCT_VERSION_MINOR << '.';
+  os << PRODUCT_VERSION_PATCH << '-' << PRODUCT_ID << '\n';
```
cjvolzka, line 121: Aren't `PRODUCT_VERSION_MAJOR`,  `PRODUCT_VERSION_MINOR`, and `PRODUCT_VERSION_PATCH` and `PRODUCT_ID` built by splitting apart the `ONNX_MLIR_PRODUCT_VERSION` supplied at compile time? If so, is there a reason to not use `ONNX_MLIR_PRODUCT_VERSION` directly instead of rebuilding it from the composite components here?

hamptonm1: Yes there is a reason... if I attempt to use `ONNX_MLIR_PRODUCT_VERSION` it will fail on the `cmake` command with the following error:

```
on.dir/Version.cpp.o -MF src/Version/CMakeFiles/OMVersion.dir/Version.cpp.o.d -o src/Version/CMakeFiles/OMVersion.dir/Version.cpp.o -c ../src/Version/Version.cpp
<command-line>: error: too many decimal points in number
../src/Version/Version.cpp:122:16: note: in expansion of macro 'ONNX_MLIR_PRODUCT_VERSION'
  122 |   os << " " << ONNX_MLIR_PRODUCT_VERSION << '\n';
      |                ^~~~~~~~~~~~~~~~~~~~~~~~~
../src/Version/Version.cpp: In function 'void onnx_mlir::getVersionPrinter(llvm::raw_ostream&)':
<command-line>: error: 'zdlc' was not declared in this scope
<command-line>: note: in definition of macro 'ONNX_MLIR_PRODUCT_VERSION'
[452/509] Building CXX object test/num...CMakeFiles/TestLoop.dir/TestLoop.cpp.o
ninja: build stopped: subcommand failed.

```

I tried 3 times....so adding the code the way I did gets around this problem :)

hamptonm1: Also I am not rebuilding it.... ONNX_MLIR_PRODUCT_VERSION is broken up into these 4 segments by using a `CMAKE_MATCH_1` and then a SET which was already added to the code.

cjvolzka: Ok. I thought that might be the case but I wanted to make sure. PR looks good to me. I'd say give it overnight to see if there's any other comments from other timezones but otherwise it's good to go.

cjvolzka (approved):

PR #1730, opened 2022-09-23
fix einsum decomposition bug

sorenlassen:
ArrayRef was used after the underlying array changed, fixed it by creating a copy of the underlying array and make the ArrayRef point to the copy


philass (approved):
LGTM, thanks for fixing this

Issue #1639, opened 2022-08-25
[RFC] Adding an ONNX to Torch conversion


qedawkins:
Hello everyone, in a recent RFC in torch-mlir a conversion from ONNX dialect to the Torch dialect in torch-mlir was proposed (https://github.com/llvm/torch-mlir/issues/1255). After feedback we are looking at adding that conversion pass to onnx-mlir.

## Proposal

- Add torch-mlir as a submodule: https://github.com/llvm/torch-mlir
- Add a new conversion pipeline from ONNX to Torch (e.g. "--convert-onnx-to-torch-pipeline") to manage both op and type conversion.
- Conversion tests in "test/mlir/conversion/onnx_to_torch/"
- Ideally target the torch-mlir backend contract which comes after shape inference and maximizing value semantics

## Proof of Concept

A proof of concept for the conversion can be found at: https://github.com/nod-ai/onnx-mlir/tree/convert-to-torch
A conversion for ONNXAddOp and ONNXConstantOp is included along with unit tests for each. This includes five passes in the conversion pipeline:

1. **"convert-onnx-to-torch"** handles the op by op lowering to torch dialect
2. **"convert-function-types-to-torch-types"** converts function arguments to torch types (e.g. torch.tensor) that wasn't converted in the previous pass
3. **"finalize-torch-type-conversion"** finalizes the conversion to torch types and appropriately removes any UnrealizedCastConversionOp's
4. **"erase-onnx-entry-point"** right now this entry point op gets removed for compatibility with torch-mlir. If this poses a problem on the onnx-mlir side feedback would be much appreciated.
5. **"refine-torch-func-value-semantics"** changes the types of tensors for public functions without uses to have value semantics. This is a compatibility requirement for torch-mlir but also a reflection of this PoC conversion targeting non-value semantic tensors. If tensors in onnx-mlir implicitly have value semantics then this pass can be removed and the conversions in 1, 2, 3 can be changed to match the type conversion in torch-mlir more closely.

Looking forward to any and all feedback!

cc @powderluv @silvasean @sstamenova @ashay

AlexandreEichenberger:
We are generally receptive, and we already have an effort from onnx to mhlo and onnx to tossa. Have you considered piggybacking on these? Namely from mhlo or toss to torch-mlid? 
[...]

PR #1731, opened 2022-09-23
Initialize conversion passes from ONNX to Torch-MLIR backend contract

qedawkins:
Includes LLVM lit tests for ONNXAddOp and ONNXConstantOp. This PR is a result of the RFC discussed in #1639. The following overview of the PR is collected from there, but see the RFC for a more detailed discussion. This includes four passes in the conversion pipeline:

1. **"convert-onnx-to-torch"** handles the op by op lowering to torch dialect
2. **"convert-function-types-to-torch-types"** converts function arguments to torch types (e.g. torch.vtensor) that wasn't converted in the previous pass
3. **"finalize-torch-type-conversion"** finalizes the conversion to torch types and appropriately removes any UnrealizedCastConversionOp's
4. **"erase-onnx-entry-point"** right now this entry point op gets removed for compatibility with torch-mlir.

As requested there is an option to disable the torch-mlir build with `ONNX_MLIR_ENABLE_TORCH`. One point of awkwardness here though comes from torch-mlir including llvm-project as a submodule. Thus when adding torch-mlir to onnx-mlir as a submodule, cloning recursively will clone the entirety of llvm as well (it doesn't build it as we are only building the parts of torch-mlir that we need). Suggestions on how to manage the submodules in this case, as well as any comments on if/how to break up this PR are welcome.

If there is anything in particular needed to help with reviewing here please let me know.


BraunPhilipp:
Thank you for the great work! You could probably configure a shallow submodule clone.

```
git config -f .gitmodules submodule.third_party/torch-mlir.shallow true
```

qedawkins:
> Thank you for the great work! You could probably configure a shallow submodule clone.

Hmmm as far as I can tell a shallow submodule won't prevent the full llvm-project submodule in Torch-MLIR from being cloned when cloning with
```
git clone --recursive https://github.com/onnx/onnx-mlir.git
```
Correct me if I'm wrong but it didn't work when I tried your suggestion. I opened an issue in Torch-MLIR about formal support for the use case here (https://github.com/llvm/torch-mlir/issues/1411).

tungld:

qedawkins:
I can convert this PR to a draft until I verify that the Windows build works, but style comments/reviews are still welcome at this point.

AlexandreEichenberger:
You need to abide by DCO. See Contributing section of main Readme.md. Every commits need to be signed. It can be overridden on this page (details next to DCO) but this should be used only exceptionally.

AlexandreEichenberger:

AlexandreEichenberger:

qedawkins:
I pushed a new change that changes to a temporary fork of Torch-MLIR that will be used for iterating this PR without needing to create PRs in Torch-MLIR for each small change. Once the reviewing is done here we can switch back to the main branch of Torch-MLIR. With this I was able to build successfully in the Docker image used by the CI with all unit tests passing, but I haven't tested locally on all platforms.

With this, if there are more CI failures, especially as this PR is reviewed, I am wondering if there is a way to set up so that running the CI doesn't require admin approval each time to allow for faster iteration (perhaps on a temporary branch within ONNX-MLIR). @AlexandreEichenberger if you or one of the other admins could help/comment here that would be greatly appreciated! If this isn't possible then no problem, but I figured I would ask.

gongsu832:

AlexandreEichenberger:

PR #1733, opened 2022-09-24
cleanup some include files

sorenlassen:


chentong319:
@sorenlassen Did you use any tool to help you in cleaning up the include files? I noticed this issue when files were copied and modified.  I wonder whether there is any tool or rule could help.

sorenlassen:
> Did you use any tool to help you in cleaning up the include files? I noticed this issue when files were copied and modified. I wonder whether there is any tool or rule could help.

I didn't, I just made this change manually because the includes in `FrontendDialectHelper.hpp` were out of control, but I don't think it's a systemic issue that we need to prioritize in general, I don't think include files are in a particularly bad state in this code base

chentong319 (approved):
LGTM!

PR #1735, opened 2022-09-25
parse string type from getTypeMap()

sorenlassen:
added a parse lit test that passes with this fix


chentong319 (approved):
LGTM!

PR #1736, opened 2022-09-25
fix 2 compiler warnings

sorenlassen:
fixes `'sprintf' is deprecated` warning on Mac and `base class should be explicitly initialized in the copy constructor` warnings on Linux

the solution to the copy constructor warning was copied from `InstrumentONNXPass.cpp`

warning from Apple clang version 14.0.0 (clang-1400.0.29.201):
```
.../src/Conversion/ONNXToKrnl/PerfectHash.cpp:47:5: warning: 'sprintf' is deprecated: This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead. [-Wdeprecated-declarations]
    sprintf(str, "%lld", (long long)val);
/Library/Developer/CommandLineTools/SDKs/MacOSX13.0.sdk/usr/include/stdio.h:188:1: note: 'sprintf' has been explicitly marked deprecated here
__deprecated_msg("This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead.")
```

compiler warning in Linux CI pipelines:
```
.../src/Transform/ONNX/ConvOpt.cpp: In copy constructor '{anonymous}::ConvOptONNXToONNXPass::ConvOptONNXToONNXPass(const {anonymous}::ConvOptONNXToONNXPass&)':
.../src/Transform/ONNX/ConvOpt.cpp:211:3: warning: base class 'class mlir::PassWrapper<{anonymous}::ConvOptONNXToONNXPass, mlir::OperationPass<mlir::func::FuncOp> >' should be explicitly initialized in the copy constructor [-Wextra]
   ConvOptONNXToONNXPass(const ConvOptONNXToONNXPass &pass) {}
```

### DIFF
--- a/src/Conversion/ONNXToKrnl/PerfectHash.cpp
+++ b/src/Conversion/ONNXToKrnl/PerfectHash.cpp
@@ -43,10 +43,7 @@ public:
 
   // Adaptation of 32-bit FNV for int64_t values.
   static inline uint32_t hash(uint32_t hval, int64_t val) {
-    char str[20];
-    sprintf(str, "%lld", (long long)val);
-
-    return hash(hval, std::string(str));
+    return hash(hval, std::to_string(val));
   }
 
   // Extracts the keys of the given map.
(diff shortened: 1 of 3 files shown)


chentong319 (approved):
LGTM!

PR #1739, opened 2022-09-27
remove initializedTensors map from ONNX parser

sorenlassen:
instead insert a constant for each initializer in the frontend_symbols_ symbol map

this is better for several reasons:

1. initializedTensors were incorrectly visible to function bodies (in TryImportFunctionCallNode)

2. nested bindings didn't hide initializedTensors within their scope

3. it's more efficient to create a constant for each initializer once rather than every time an initializer is accessed

4. it's simpler to look up a single symbol mapping


chentong319:
Can nested scopes have different initialized tensor with the same name? Even though the initialized tensors of outer scope are visible in the inner scope, the initialized tensors of the inner scope should not be legal for the outer scope. If the initialized tensors are not push/pop at the boundary of a scope, we cannot detect the error in the model. Right?  Of course, error detection is not on our top priorities.

sorenlassen:
I believe that we get the correct and logical behavior by making initializers follow the same visibility logic as graph inputs, as is done in this PR by putting initializers into frontend_symbols_. An initializer from an outer scope can be hidden in an inner scope by another initializer with the same name or graph input or node output with the same name. I believe this is what we want.

chentong319:
I see. The PR can be described as "merge the InitializedTensors and Frontend_symbols_".

chentong319 (approved):
LGTM!

PR #1740, opened 2022-09-27
factored input shaping into ModelInputShaper

sorenlassen:
this makes the it easier to read and understand the input shaping logic and also makes the rest of the FrontendDialectTransformer implementation shorter and easier to read


AlexandreEichenberger:
Is there any functionality changes, or the code was just moved in a separate class (which I like)

sorenlassen:
there are no functional changes

there are some minor code cleanups but fundamentally just moved the code

src/Builder/ModelInputShaper.cpp
```diff
@@ -0,0 +1,118 @@
...
+        dims.emplace_back(dimSize);
+      }
+      inputs_shape_information_.insert(std::make_pair(inputID, dims));
+    }
+  }
+}
+
+namespace {
+RankedTensorType forceShape(
+    RankedTensorType tensorTy, const std::vector<int> &forcedDims) {
```
tungld, line 80: Any reason that we do not put this inside ModelInputShaper as a private function too?

tungld (commented):
Thanks for reorganizing the code! It looks better.

src/Builder/ModelInputShaper.cpp
sorenlassen: anonymous namespace hides the implementation best, as it doesn't require a method declaration in the header file

it's a pure function from inputs to output and doesn't need access to the class's fields

tungld: Got it. Thanks!

src/Builder/ModelInputShaper.hpp
```diff
@@ -0,0 +1,82 @@
...
+//    and
+//    - change the first and second dimensions in the second input to unknown
+//    dimensions
+class ModelInputShaper {
+public:
+  ModelInputShaper();
+
+  void setShapeInformation(const std::string &shapeInformation);
+
+  mlir::Type reshape(int inputIndex, mlir::Type inputType) const;
```
tungld, line 68: Could you add some comments for `setShapeInformation` and `reshape` so that users know how to use them? Then it's good to go.

tungld (changes requested):

src/Builder/ModelInputShaper.hpp
sorenlassen: good point

done

tungld: Thanks!

sorenlassen:
thanks for the review feedback

please approve and merge if no one objects to this change

PR #1744, opened 2022-09-27
Streamline pointer logic inside of FrontendDialectTransformer's Infer…

messerb5467:
…Types.

This PR introduces logic to streamline pointer logic found towards the end of FrontendDialectTransformer's InferTypes. I was seeing issues with use of out of scope local variables, so I collapsed this down to the smallest possible case that passes our unit tests as well as quiesces other issues I was seeing.

Signed-off-by: Brad Messer <messerb5467@gmail.com>

### DIFF
--- a/src/Builder/FrontendDialectTransformer.cpp
+++ b/src/Builder/FrontendDialectTransformer.cpp
@@ -1125,8 +1125,6 @@ private:
 
   void InferTypes(const onnx::FunctionProto *func,
       std::vector<onnx::TypeProto> &inputTypes) {
-    // types: Used for temporary copies of Types, freed at end of function.
-    std::vector<std::unique_ptr<onnx::TypeProto>> types;
     std::unordered_map<std::string, onnx::TypeProto *> typeMap;
     // Initialize types and values (if available) of function inputs:
     const auto num_inputs =
@@ -1148,10 +1146,7 @@ private:
 
       // Update types:
       for (int i = 0; i < n.output_size(); ++i) {
-        std::unique_ptr<onnx::TypeProto> p =
-            std::make_unique<onnx::TypeProto>(*node_ctx.getOutputType(i));
-        typeMap[n.output(i)] = p.get();
-        types.push_back(std::move(p));
+        typeMap[n.output(i)] = node_ctx.getOutputType(i);
       }
     }
 


messerb5467:
@chentong319 @gramalingam Could you review this? I unfortunately can't seem to alert anyone by assigning reviewers.

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -1125,8 +1125,6 @@ class FrontendGenImpl {
 
   void InferTypes(const onnx::FunctionProto *func,
       std::vector<onnx::TypeProto> &inputTypes) {
-    // types: Used for temporary copies of Types, freed at end of function.
```
sorenlassen, line 1128: from reading the code I think `types` vector solves that the lifetime of the pointer returned from node_ctx.getOutputType() cannot be counted on to extend past the lifetime of node_ctx which is only the body of the `num_inputs` for loop and ends before the pointer is read in the `typeMap` for loop afterwards

therefore I think you need to make an additional change to make it safe: change typeMap to
```
std::unordered_map<std::string, onnx::TypeProto> typeMap
```
and copy the TypeProtos into the map instead of just pointers

sorenlassen: btw, copying the TypeProtos into typeMap will copy the inputTypes, which is less efficient than the existing implementation

another approach, which would make everything simpler is to omit typeMap and just stick things directly into `onnx_type_map` with onnx_type_map.AddMapping instead of typeMap, but you may need to check `ContainsKey` like in 'AddValueInfo()`

messerb5467: I've looked through the IR.md, onnx operators document, and the onnx.in.proto defining the base data types we're working with as part of the FrontendDialectTransformer objects. I don't have enough to see how you would rip out the node context and other bits yet, so I'm wondering if you have any more thoughts of things I should look at. You do seem to think this is possible though and I'm learning a lot which makes continued pursuit of this fix a good idea.

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -1125,15 +1125,12 @@ class FrontendGenImpl {
...
-    std::vector<std::unique_ptr<onnx::TypeProto>> types;
-    std::unordered_map<std::string, onnx::TypeProto *> typeMap;
     // Initialize types and values (if available) of function inputs:
     const auto num_inputs =
         std::min(func->input_size(), static_cast<int>(inputTypes.size()));
     for (int i = 0; i < num_inputs; ++i) {
       const std::string &input_name = func->input(i);
-      typeMap[input_name] = &inputTypes[i];
+      onnx_type_map.AddMapping(input_name, &inputTypes[i]);
```
sorenlassen, line 1133: I believe the argument should just be `inputTypes[i]` without `&`

src/Builder/FrontendDialectTransformer.cpp
sorenlassen: I think inferTypes is improved with this change. I'm not sure there are immediate ways to improve it further but I'm open to ideas.

Otherwise there are plenty of other places in the code base to make improvements. Please keep writing PRs. If you want to run some ideas by people on the project then a good way is to open an issue to propose changes or ask questions.

messerb5467: One thing I wanted to clarify is that this change doesn't really work and I would have to figure out how to factor out the nodeContext and a few other variables to really streamline this which I don't think is possible at this time and why I was hoping for a bit more feedback. If there's plenty of other places in the code though, closing this PR is probably the way to go just because I don't have enough base knowledge yet to be fully helpful.

sorenlassen: oops, I didn't notice that node_ctx needed the `typeMap` argument for its constructor

I think the best solution is to re-introduce the typeMap from before but make it point into onnx_type_map, i.e., all the places you call `onnx_type_map.AddMapping(name, typeproto)` you also put it into typeMap: `typeMap[name] = onnx_type_map.GetByOnnxName(name)`

sorenlassen: I noticed the compile error in the CI logs, looks like typeMap's 2nd type parameter needs to be a const pointer

or, if the node_ctx constructor won't accept that, you can probably work around it with a const_cast somewhere

messerb5467: I double checked the assignment you saw and [maps by default will define const behind the covers](https://stackoverflow.com/questions/50593012/why-does-stdunordered-map-not-work-with-const-stdstring-key). For normal use you can just pass in directly what we need and we can move forward. I'll keep an eye out. The last few are after hours for me, so I'm not checking back as quickly.

sorenlassen: the stackoverflow issue is different

try `std::unordered_map<std::string, const onnx::TypeProto *> typeMap`, I think that will work because that matches 
https://github.com/onnx/onnx/blob/main/onnx/shape_inference/implementation.h#L126

sorenlassen:
the change looks good to me

the const_cast isn't pretty but looks like the right way to go, I looked into InferenceContextImpl in third_party/onnx/onnx/shape_inference/implementation.h and it doesn't modify the type protos in typeMap, it looks like a mistake that it omits const from the pointer type argument to unordered_map (I made a wrong comment about this above where I mistook the 2nd and 3rd arguments)

@chentong319 are you also ok with this change?

to fix the DCO CI failure click on the "Details" link and follow the instructions (and going forward use `git commit -s` as described [here](https://github.com/onnx/onnx-mlir/blob/6ff79af88914bd0e3bcc96a595c79ba6424a6e95/docs/Workflow.md#step-7-commit--push))

to fix the clang format CI failure run clang-format on the source file, see [here](https://github.com/onnx/onnx-mlir/blob/6ff79af88914bd0e3bcc96a595c79ba6424a6e95/docs/Workflow.md#code-style)

messerb5467:
> the change looks good to me
> 
> the const_cast isn't pretty but looks like the right way to go, I looked into InferenceContextImpl in third_party/onnx/onnx/shape_inference/implementation.h and it doesn't modify the type protos in typeMap, it looks like a mistake that it omits const from the pointer type argument to unordered_map (I made a wrong comment about this above where I mistook the 2nd and 3rd arguments)
> 
> @chentong319 are you also ok with this change?
> 
> to fix the DCO CI failure click on the "Details" link and follow the instructions (and going forward use `git commit -s` as described [here](https://github.com/onnx/onnx-mlir/blob/6ff79af88914bd0e3bcc96a595c79ba6424a6e95/docs/Workflow.md#step-7-commit--push))
> 
> to fix the clang format CI failure run clang-format on the source file, see [here](https://github.com/onnx/onnx-mlir/blob/6ff79af88914bd0e3bcc96a595c79ba6424a6e95/docs/Workflow.md#code-style)



> the change looks good to me
> 
> the const_cast isn't pretty but looks like the right way to go, I looked into InferenceContextImpl in third_party/onnx/onnx/shape_inference/implementation.h and it doesn't modify the type protos in typeMap, it looks like a mistake that it omits const from the pointer type argument to unordered_map (I made a wrong comment about this above where I mistook the 2nd and 3rd arguments)
> 
> @chentong319 are you also ok with this change?
> 
> to fix the DCO CI failure click on the "Details" link and follow the instructions (and going forward use `git commit -s` as described [here](https://github.com/onnx/onnx-mlir/blob/6ff79af88914bd0e3bcc96a595c79ba6424a6e95/docs/Workflow.md#step-7-commit--push))
> 
> to fix the clang format CI failure run clang-format on the source file, see [here](https://github.com/onnx/onnx-mlir/blob/6ff79af88914bd0e3bcc96a595c79ba6424a6e95/docs/Workflow.md#code-style)

Definitely. I was more or less just doing some quick testing last night to get this working after hours so I did a less pretty job. I wasn't planning to keep this PR open, but at this point its probably better for record tracking purposes as well as minimizing confusion.

messerb5467:
The signoff amend is going as cleanly. I'm going to close this out and re-upload. Yesterday was wanting to build on the excitement and learning of a new code base, but now its time for the discipline of the normal software process.

PR #1745, opened 2022-09-27
[Doc only] Adding reference to onnx-mlir-serving

AlexandreEichenberger:
Started a list of projects that use/are based on onnx-mlir. Maybe others can add their projects too, if they feels it would be valuable to them.


chentong319:
Are we going to use `Projects` tab in GitHub?

AlexandreEichenberger:
@chentong319 I am not familiar with the Project tab, I think it is an alternative way to have dialogs rather than using issues.

This PR is trying to make a list of external project that have a connection with ONNX-MLIR. If you believe there is a better place to put this info than on the README.md, I have no problem exploring another avenue.

chentong319 (approved):
LGTM!
Previously I thought you wanted to list the projects in onnx-mlir.

PR #1747, opened 2022-09-28
Update protobuf

cjvolzka:
Protobuf 3.16 is no longer being updated by that community. This upgrades to the closest protobuf version to 3.16 that is still supported which includes current security updates.


AlexandreEichenberger:
@cjvolzka think the ppc jenkins failed for other PRs, so it is probably good if it succeeds with s390

sstamenova:
It would be nice to add a comment in the commit message or in the scripts about how/why this particular version was chosen, but looks good otherwise.

cjvolzka:
Updated description.

sstamenova (approved):

chentong319 (approved):
LGTM!

PR #1752, opened 2022-09-29
setup.py install is deprecated, use pip install instead

gongsu832:


AlexandreEichenberger (approved):
LGTM

sstamenova:
Can you elaborate in the description on which `setup.py` has been deprecated and then make the change to how `onnx` is built in the Windows CI as well?

gongsu832:
Direct invocation of setup.py has been deprecated for a while but not all cases give warning about the deprecation. See https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html for more details. I only saw it on MacOS build a couple of days ago. And Linux build didn't give warning so far.

In any case, I just updated the windows CI to use pip to install onnx. So all direct invocations of setup.py have been fixed, except protobuf. Apparently the way its setup.py is written prevents it from being invoked by pip.

PR #1753, opened 2022-09-29
Streamline InferTypes in FrontendDialectTransformer.cpp

messerb5467:
This PR introduces a fix to remove the use of the types and local pointer variables causing errors in cppcheck. That was still a great implementation, but [experimentation got us to a workable solution](https://github.com/onnx/onnx-mlir/pull/1744) without the local pointer trouble. @sorenlassen @chentong319 if you could review, that'd be appreciated.
Signed-off-by: Brad Messer <messerb5467@gmail.com>

### DIFF
--- a/src/Builder/FrontendDialectTransformer.cpp
+++ b/src/Builder/FrontendDialectTransformer.cpp
@@ -960,15 +960,15 @@ private:
 
   void InferTypes(const onnx::FunctionProto *func,
       std::vector<onnx::TypeProto> &inputTypes) {
-    // types: Used for temporary copies of Types, freed at end of function.
-    std::vector<std::unique_ptr<onnx::TypeProto>> types;
     std::unordered_map<std::string, onnx::TypeProto *> typeMap;
     // Initialize types and values (if available) of function inputs:
     const auto num_inputs =
         std::min(func->input_size(), static_cast<int>(inputTypes.size()));
     for (int i = 0; i < num_inputs; ++i) {
       const std::string &input_name = func->input(i);
-      typeMap[input_name] = &inputTypes[i];
+      onnx_type_map.AddMapping(input_name, inputTypes[i]);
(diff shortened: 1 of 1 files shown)


sorenlassen (approved):
LGTM

I'll merge it tomorrow unless @chentong319 requests any changes

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -983,16 +983,12 @@ class FrontendGenImpl {
...
-            std::make_unique<onnx::TypeProto>(*node_ctx.getOutputType(i));
-        typeMap[n.output(i)] = p.get();
-        types.push_back(std::move(p));
+        const std::string &output_name = n.output(i);
+        onnx_type_map.AddMapping(output_name, *node_ctx.getOutputType(i));
+        typeMap[output_name] = const_cast<onnx::TypeProto *>(
+            onnx_type_map.GetByOnnxName(output_name));
       }
     }
-
```
chentong319, line 992: I noticed that the modification moved the AddMapping from the outside of node loop into the node loop. Could one element of typeMap be assigned with different value inside the node loop? I guess not since an output tensor should be defined only once by onnx semantic, and our shape inference updates the type of the output of the current node. I somehow understand why Tian wrote the code in this way.

sorenlassen: in theory the old code would allow an attempted insertion of an output into typeMap to be ignored if the name was already mapped in typeMap, but I don't think we need to support that because if we look at how the same names are entered into frontend_symbols_, they are entered much the same way with calls to AddMapping (via BindOnnxName and ImportNode+buildOutputAndOperation)

in other words, in practice the names added are distinct (because otherwise frontend_symbols_.AddMapping would break somewhere else) so it works to call onnx_type_map.AddMapping directly without going through typeMap

PR #1754, opened 2022-09-30
Fix for GatherND verifier

philass:
The function is trying to verify range of indices when the indices are constant at compile time. That is why it has `if (DenseElementsAttr valueAttribute = getDenseElementAttributeFromONNXValue(indices))` check is moved to the outer level, since if the value is not known at compile time the rest of the checks are meaningless.

Also there is a fundamental mistake in the way it checks. `dataShape[i] for  0 <= i < indicesRank` can be out of range. Since it does not add the check for compile time constant at the beginning and it has this out of range issue, even for non-constant indices (we technically should not check anything) we might reach a seg-fault.

The code also completely reorganizes the way the check is supposed to happen and removed one layer of nested loop.
datashape that should be checked is dataShape[i+b]. However, i is not in range of indicesRank. i should be in range of indices[indicesRank - 1], since the last dimension of indices is the one actually indicates number of gather axes.

Co-authored-by: Hadi Jooybar <mjooybar@groq.com>
Signed-off-by: Philip Lassen <philiplassen+git@gmail.com>


src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -3650,25 +3650,24 @@ LogicalResult ONNXGatherNDOp::verify() {
...
-    if (dataDimAtAxis < 0)
-      continue;
-
-    if (DenseElementsAttr valueAttribute =
-            getDenseElementAttributeFromONNXValue(indices))
-      for (IntegerAttr value : valueAttribute.getValues<IntegerAttr>()) {
-        static int n = 0;
-        int64_t index = value.getInt();
-        if (index < -dataDimAtAxis || index > dataDimAtAxis - 1)
+  //   -data.shape[b + i] <= indices[...,i] <= (data.shape[b + i]-1)].
```
tungld, line 3653: It makes sense. Gathering should start from `b+1` axis. 

So it looks like [the description](https://github.com/onnx/onnx/blob/main/docs/Operators.md#GatherND) in ONNX was wrong `All values in indices are expected to be within bounds [-s, s-1] along axis of size s (i.e.) -data_shape[i] <= indices[...,i] <= data_shape[i] - 1. It is an error if any of the index values are out of bounds.`, doesn't it?

AlexandreEichenberger:
@tungld I let you complete the review once you believe this PR is ready to land

src/Dialect/ONNX/ONNXOps.cpp
philass: Yes I think the ONNX description is wrong unfortunately

tungld (approved):
LGTM. Thanks for the fix!

PR #1695, opened 2022-09-14
Simplify shape-related operations

tungld:
This patch is to simplify shape-related operations that are often generated by onnx converters and complicate our shape inference for unknown dimensions.

Usually shape-related calculation starts with an ONNXShapeOp to get the shape of a tensor, e.g.
```mlir
%0 = "onnx.Shape"(%arg0) : (tensor<?x256x1xf32>) -> tensor<3xi64>
%1 = "onnx.Reshape"(%arg1, %0) : (tensor<?x256xf32>,  tensor<3xi64>) -> tensor<?x?x?xf32>
```
And the problem is that ONNXShapeOp returns a tensor whose values are unknown at compile time, so the value `256` in <?x256> cannot be propagated further after this ONNXShapeOp. Hence, the output shape of Reshape is totally unknown.

The essential idea in this patch to solve the problem is representing ONNXShapeOp by ONNXDimOp (newly introduced in this path) and ONNXConcatOp. By that way, each invidual dimension of a tensor is exposed and easily propagated through other ops. For example, the above "onnx.Shape" will rewritten into:
```mlir
%0 = "onnx.Dim"(%arg0) {axis = 0 : si64} : (tensor<?x256x1xi64>) -> tensor<1xi64>
%1 = "onnx.Constant"() {value = dense<256> : tensor<1xi64>} : () -> tensor<1xi64>
%2 = "onnx.Constant"() {value = dense<1> : tensor<1xi64>} : () -> tensor<1xi64>
%3 = "onnx.Concat"(%0, %1, %2) {axis = 0 : si64} : (tensor<1xi64>, tensor<1xi64>, tensor<1xi64>) -> tensor<3xi64>
%4 = "onnx.Reshape"(%arg0, %3) : (tensor<?x256xf32>,  tensor<3xi64>) -> tensor<?x?x?xf32>
``` 

Now, it's straighforward to update the output shape of Reshape from `<?x?x?xf32>` to `<?x256x1xf32>` by looking at the inputs of Concat.

Applying this patch to an example in bertsquad-12 model (See #1579), the input program:
```mlir
module attributes {llvm.data_layout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128", llvm.target_triple = "x86_64-unknown-linux-gnu", "onnx-mlir.accels" = ["NNPA-0x10000"]} {
[...]

PR #1755, opened 2022-09-30
Simplify GatherOp when its inputs are dimensions

tungld:
This is complementary to #1695 by adding support for GatherOp.


chentong319 (commented):
The PR itself is fine in handling a special case involving two ONNX ops in d-u relationship. In data flow analysis, usually a transfer function is defined for each op so that we can handle all the combination of d-u flow uniformly. However, this PR can be accepted as a quick hack if @AlexandreEichenberger  approves it.

AlexandreEichenberger (approved):
Generally agree with Tong's comment, but let this one go through in the sake of expediency.

PR #1757, opened 2022-10-03
Assign each question mark a unique negative integer value

tungld:
IndexExpr currently returns a fix value, -1, for a dimension of question mark during compile time, e.g. when calling shape inference.

This patch assigns question marks with unique negative integer values, starting from -2. This is useful when a input dimension is passed to the output dimension without any computation. In such case, we can check that the output dimension is the same as the input dimension or not via the unique values. However, such checking requires storing IndexExpr for the input inside ShapeHelpr, which requires modifying ShapeHelper. Another PR will be created for that if this patch is accepted.

An optional argument, `uniqueQuestionMark` is added to `IndexExpr::getShape`, which is to return negative values (-N) for question marks.
```C
static void getShape(llvm::SmallVectorImpl<IndexExpr> &indexExprList,
                     llvm::SmallVectorImpl<int64_t> &intDimList,
                     bool uniqueQuestionMark);
```
By default, `uniqueQuestionMark = false` is used for backward compatibility.


AlexandreEichenberger (approved):
LGTM, the next step is that when we have a computation with a question mark, the output is a new question mark (unless silly computations such as +0, *1, ...). But when the output is the same as an input question mark, then the same question mark id is used. Not 100% sure if/when new ones are created vs passed as reference.

PR #1764, opened 2022-10-05
Handle scalar tensor tensor<type> in SimplifyShapeRelatesOps pass

tungld:
Some ops to process shape return a scalar tensor `tensor<i64>` instead of `tensor<1xi64>`, while our SimplifyShapeRelatesOps pass always use `tensor<1xi64>`. So we need to convert `tensor<1xi64>` to `tensor<i64>` by a squeeze op.

### DIFF
--- a/src/Transform/ONNX/SimplifyShapeRelatedOps.cpp
+++ b/src/Transform/ONNX/SimplifyShapeRelatedOps.cpp
@@ -138,11 +138,20 @@ void getDimsInt64(Value val, SmallVectorImpl<int64_t> &result) {
 }
 
 /// Create a ConcatOp to concat the list of tensors.
-Value emitConcatOpForDims(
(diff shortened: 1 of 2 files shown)


AlexandreEichenberger (approved):
LGTM

AlexandreEichenberger:
> So we need to convert tensor<i64> to tensor<1xi64> by a squeeze op.

is confusing: as going from i64 to 1xi64 is an unsqueezed I think. But from the code, I think you are normalizing all the 1xi64 to i64, isn't it?

tungld:
> is confusing: as going from i64 to 1xi64 is an unsqueezed I think. But from the code, I think you are normalizing all the 1xi64 to i64, isn't it?

Oops, my mistake in writing the comment. We go from 1xi64 to i64. I just corrected my comment. Thanks!

PR #1767, opened 2022-10-06
Add a flag to allow using the inferred shape without refinement in updateType

tungld:
It looks like the shape inference for LoopOp/ScanOp potentially extends the number of dimensions for some output. A comment from LoopOp' shape inference:
```
  // For scan outputs, we set their shape to be the shape of the return
  // values of the loop body function corresponding to scan outputs, but
  // with an extra leading dimension.
```

`updateType` function currently always refine an inferred shape by using static dimensions from the existing shape (if existed), and it requires that the ranks of the inferred and existing shapes must be the same. However, it seems it's not true for LoopOp/ScanOp where the rank is increased by one.

So this patch adds an optional argument, `refineShape`, to `updateType` function, which allows disabling refining an inferred shape. This is applied to LoopOp/ScanOp.

### DIFF
--- a/src/Dialect/ONNX/ONNXOps.cpp
+++ b/src/Dialect/ONNX/ONNXOps.cpp
@@ -5069,7 +5069,8 @@ LogicalResult ONNXScanOp::inferShapes(
     // input to Loop operation, but we need to eliminate the possibility of
     // early termination to be sure.
     updateType(std::get<1>(vScanOutputValToTy), squeezedShape,
-        rankedScanTy.getElementType());
+        rankedScanTy.getElementType(), /*encoding=*/nullptr,
+        /*useInferredShape=*/true);
   }
 
   // Now we have modified loop body function input signatures according to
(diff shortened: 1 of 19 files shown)


chentong319 (approved):
LGTM!

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -5069,7 +5069,8 @@ LogicalResult ONNXScanOp::inferShapes(
     // input to Loop operation, but we need to eliminate the possibility of
     // early termination to be sure.
     updateType(std::get<1>(vScanOutputValToTy), squeezedShape,
-        rankedScanTy.getElementType());
+        rankedScanTy.getElementType(), /*encoding=*/nullptr,
+        /*refineShape=*/false);
```
sorenlassen, line 5073: should the updateType() call below in line 5113 also be called with refineShape=false?

tungld: Good catch. I'll create a PR to update this. Thanks!

PR #1771, opened 2022-10-07
Add more functionality to conv transpose shape inference

philass:

### DIFF
--- a/test/mlir/onnx/onnx_shape_inference.mlir
+++ b/test/mlir/onnx/onnx_shape_inference.mlir
@@ -506,6 +506,58 @@ func.func @test_conv_transpose_2(%arg0 : tensor<1x64x36x48xf32>, %arg1 : tensor<
 }
 
 // -----
+
+func.func @test_conv_transpose_3(%arg0: tensor<1x1x3x3xf32>, %arg1: tensor<1x2x3x3xf32>) -> tensor<*xf32> {
+  %0 = "onnx.NoValue"() {value} : () -> none
+  %1 = "onnx.ConvTranspose"(%arg0, %arg1, %0) {output_padding = [1, 1], strides = [3, 2]} : (tensor<1x1x3x3xf32>, tensor<1x2x3x3xf32>, none) -> tensor<*xf32>
+  return %1 : tensor<*xf32>
+
+// CHECK-LABEL: test_conv_transpose_3
+// CHECK: [[CST:%.+]] = "onnx.NoValue"() {value} : () -> none
+// CHECK-NEXT: [[RES_ATTR:%.+]] = "onnx.ConvTranspose"(%arg0, %arg1, %0) {auto_pad = "NOTSET", dilations = [1, 1], group = 1 : si64, kernel_shape = [3, 3], output_padding = [1, 1], output_shape = [1, 2, 10, 8], pads = [0, 0, 0, 0], strides = [3, 2]} : (tensor<1x1x3x3xf32>, tensor<1x2x3x3xf32>, none) -> tensor<1x2x10x8xf32>
+// CHECK: return [[RES_ATTR]] : tensor<1x2x10x8xf32>
+}
+
+// -----
+
+func.func @test_conv_transpose_4(%arg0: tensor<1x1x3x3xf32>, %arg1: tensor<1x2x3x3xf32>) -> tensor<*xf32> {
+  %0 = "onnx.NoValue"() {value} : () -> none
+  %1 = "onnx.ConvTranspose"(%arg0, %arg1, %0) {pads = [1, 2, 1, 2], strides = [3, 2]} : (tensor<1x1x3x3xf32>, tensor<1x2x3x3xf32>, none) -> tensor<*xf32>
+  return %1 : tensor<*xf32>
+
+// CHECK-LABEL: test_conv_transpose_4
+// CHECK: [[CST:%.+]] = "onnx.NoValue"() {value} : () -> none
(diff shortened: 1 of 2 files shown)


test/mlir/onnx/onnx_shape_inference.mlir
```diff
@@ -478,6 +478,32 @@ func.func @test_conv_transpose_2(%arg0 : tensor<1x64x36x48xf32>, %arg1 : tensor<
...
+  %0 = "onnx.NoValue"() {value} : () -> none
+  %1 = "onnx.ConvTranspose"(%arg0, %arg1, %0) {pads = [1, 2, 1, 2], strides = [3, 2]} : (tensor<1x1x3x3xf32>, tensor<1x2x3x3xf32>, none) -> tensor<*xf32>
+  return %1 : tensor<*xf32>
+
+// CHECK-LABEL: test_conv_transpose_4
+// CHECK: [[CST:%.+]] = "onnx.NoValue"() {value} : () -> none
+// CHECK-NEXT: [[RES_ATTR:%.+]] = "onnx.ConvTranspose"(%arg0, %arg1, %0) {auto_pad = "NOTSET", dilations = [1, 1], group = 1 : si64, kernel_shape = [3, 3], output_shape = [1, 2, 7, 3], pads = [1, 2, 1, 2], strides = [3, 2]} : (tensor<1x1x3x3xf32>, tensor<1x2x3x3xf32>, none) -> tensor<1x2x7x3xf32>
+// CHECK: return [[RES_ATTR]] : tensor<1x2x7x3xf32>
+}
+
```
sorenlassen, line 506: suggestion: add the

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2138,28 +2178,42 @@ LogicalResult ONNXConvTransposeOp::inferShapes(
...
+    // Insert batch size.
+    outputShapeFinal.emplace_back(xShape[0]);
+    // Insert number of filters being applied (number of output channels *
+    // groups).
+    outputShapeFinal.emplace_back(outChannels);
+    // Compute and insert spatial dims.
+    insertConvTransposeSpatialDim(outputShapeFinal, xShape, kernelShape,
+        padsOpt, stridesOpt, outputPads, outputShape, dilationsOpt);
+    output_shapeAttr(builder.getI64ArrayAttr(outputShapeFinal));
+    getResult().setType(
```
sorenlassen, line 2213: nit: both if and else branch end with the same statement, so you can move it outside the conditional

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -2138,28 +2178,42 @@ LogicalResult ONNXConvTransposeOp::inferShapes(
...
-  insertConvTransposeSpatialDim(outputDims, xShape, kernelShape, padsOpt,
-      stridesOpt, outputPads, outputShape, dilationsOpt);
+  llvm::SmallVector<int64_t, 4> outputShapeFinal;
 
-  // Set the output shape if it's not already set
-  if (!outputShape.has_value()) {
-    output_shapeAttr(builder.getI64ArrayAttr(outputDims));
+  if (outputShape.has_value()) {
+    if (xShape[0] != ArrayAttrIntVal(outputShape, 0)) {
+      return emitError("mismatch in batch size");
```
sorenlassen, line 2185: nit: emitOpError is better because it helpfully prints the op name

same in line 2188 below

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -597,6 +597,46 @@ static LogicalResult RNNShapeInference(T *op, int gates) {
   return success();
 }
 
+static void insertConvTransposePads(SmallVectorImpl<int64_t> &inferedPads,
+    const llvm::StringRef autoPad, ArrayRef<int64_t> xShape,
```
sorenlassen, line 601: nit: remove const, it's not idiomatic

sorenlassen (approved):

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -597,6 +597,46 @@ static LogicalResult RNNShapeInference(T *op, int gates) {
...
+    Optional<ArrayAttr> kernelShape, Optional<ArrayAttr> padsOpt,
```
sorenlassen, line 602: 1 more thing: since the code below assumes kernelShape, stridesOpt, outputShapeOpt all satisfy has_value() (otherwise they hit assertion failure in ArrayAttrIntVal()), I think it will be clearer if they are passed into insertConvTransposePads with type ArrayAttr and change the caller to call insertConvTransposePads with .value() of all those args

sorenlassen: or forget about it if that doesn't match the existing logic for insertConvTransposeSpatialDim

test/mlir/onnx/onnx_shape_inference.mlir
philass: I've added the tests from that PR

src/Dialect/ONNX/ONNXOps.cpp
philass: Good catch

src/Dialect/ONNX/ONNXOps.cpp
philass: I didn't know that `emitOpError` was better than `emitError`. Thanks for the info

src/Dialect/ONNX/ONNXOps.cpp
philass: Done

src/Dialect/ONNX/ONNXOps.cpp
philass: Yeah, just copying the patterns in the file.

I'll leave a refactor for things like this for another PR

AlexandreEichenberger: how different is this from the regular conv? Also, could it be packaged into a shape inference helper class... trying to reduce the amount of specific shape inference inside ONNXOps.cpp...

PR #1776, opened 2022-10-11
Add shape inference for Upsample

philass:


src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -5476,7 +5476,58 @@ LogicalResult ONNXUniqueOp::inferShapes(
 
 LogicalResult ONNXUpsampleOp::inferShapes(
     std::function<void(mlir::Region &)> doShapeInference) {
-  return emitError(NOT_IMPLEMENTED_MESSAGE);
+  // Sanity checks on input data argument
+  if (!X().getType().isa<RankedTensorType>()) {
```
chentong319, line 5480: Could the sanity check be put into verifier?

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -5476,7 +5476,58 @@ LogicalResult ONNXUniqueOp::inferShapes(
...
+  if (!scalesConstOp) {
+    return success();
+  }
+  auto valueAttr = scalesConstOp.valueAttr().dyn_cast<DenseElementsAttr>();
+  if (!valueAttr) {
+    return emitError("Scales constant is not a DenseElementsAttr");
+  }
+  int scaleIdx = 0;
+  // Why are the scale values float's?
+  for (auto it = valueAttr.getValues<FloatAttr>().begin();
```
chentong319, line 5512: It is possible that scale is the result of some ratio, which is a float.

chentong319:
I saw upsample beams deprecated. Can we use onnx.resize to express onnx.upsample?

src/Dialect/ONNX/ONNXOps.cpp
philass: Good point. I've added a verifier, and put the checks in there

src/Dialect/ONNX/ONNXOps.cpp
philass: Ahh yes that makes sense. Thanks for the clarification

philass:
@chentong319 thanks for the review,

I've updated the PR to address the comments. Please let me know if you would like any further changes

chentong319 (approved):
LGTM!

PR #1778, opened 2022-10-11
Refine the lowering of onnx.concat

chentong319:
The existing implementation is correct in functionality, but lost some high level semantics in lowering. The generated loop code may be difficult for loop optimization. The semantics of concat is that the dimensions (except the concat.axis()) should be the same for all the input tensor.  The new PR tries to find a constant dim  of an input tensor if there is one. If all input tensors have a dynamic dim, we should use memref.dim on one input (any one is fine), instead of using different input tensors. The resulted loops will clearly have the **SAME**  dynamic bound, more friendly for further loop optimization.
This PR also changes the usage of IndexExpr in lowering. The size of concatenated dimension is accumulated across loops, instead of partially redundantly for each input. The tricky part is how to use IndexExpr in nested scope.


src/Dialect/ONNX/ShapeInference/Concat.cpp
```diff
@@ -37,19 +37,36 @@ LogicalResult ONNXConcatOpShapeHelper::computeShape(
...
+            outputDims[dim] = currInputBounds.getDim(dim);
+            isConstant[dim] = true;
+          } else if (i == numInputs - 1) {
+            // If no constant dimension found for all the inputs, use the
+            // dynamic size of the last input.
+            outputDims[dim] = currInputBounds.getDim(dim);
+          }
+        }
+      }
+    }
```
tungld, line 67: Is it easier to initialize `outputDims` with the first input's dims? Then accumulate the dim along the axis and replace other dims if there is a constant. By that way, we don't need the `isConstant`.

chentong319: Updated.

tungld (approved):
LGTM!

PR #1780, opened 2022-10-11
Derive kernel shape when not provided

philass:
Co-authored-by: Hadi Jooybar <mjooybar@groq.com>


src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -384,6 +384,8 @@ static LogicalResult processConvPadParam(T *op, ArrayRef<int64_t> inputShape,
       // -1)*dilation + 1".
       auto sumOfPad = (outputSize - 1) * strideVal +
                       ((kernelSize - 1) * dilationVal + 1) - inputSize;
+      if (sumOfPad < 0)
+        sumOfPad = 0;
```
tungld, line 388: Just wonder if this would be an assertion? Is it possible that `sumOfPad < 0`?

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -1915,6 +1917,30 @@ LogicalResult ONNXConvOp::inferShapes(
...
+    int32_t spatialRank = xShape.size() - spatialOffset;
+    SmallVector<int64_t, 2> defaultVals;
+    for (int i = 0; i < spatialRank; ++i)
+      defaultVals.emplace_back(weightShape[spatialOffset + i]);
+    // Convert to ArrayRef, then build attribute, then store attribute.
+    ArrayRef<int64_t> defaultRefs(defaultVals);
+    auto builder = mlir::Builder(getContext());
+    kernel_shapeAttr(builder.getI64ArrayAttr(defaultRefs));
+    kernelShape = kernel_shape();
+  }
```
tungld, line 1938: I see this code duplicated at several locations. How about creating a function `processConvKernelShapeParam` for this code, similar to `processConv{Dilation/Stride/Pad}Param`, and call it inside `processConvTypeParam`?

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -1915,6 +1917,30 @@ LogicalResult ONNXConvOp::inferShapes(
...
+
+  // Process strides, dilations, and pads.
+  LogicalResult res = processConvTypeParams<>(this, X());
```
tungld, line 1941: Looking at the comment before ONNXConvOp::inferShapes, it looks like the intension is to not change the operation's attributes during shape inference but the output's shape only. As I see it, it is still debating. Perhaps, you can raise this issue in our next weekly meeting.

philass: I think I will take a shot at rewriting this as a canonicalization

philass:
@tungld,

The goal of the PR has now been changed. 

I've added a `processConvKernelShapeParam` function to take care of the duplicate no kernel provided handling.

I realized that is sufficient for me to simply expose `processConvTypeParams`, and then I can use it to change the ONNX ops in my companies repo, without it effecting everybody else.

If it is a clean solution I can upstream it later.

philass:
I'm not sure how you feel about the implementation and definition being in different base file names. I can  keep the function implementation `ONNXOps.cpp` and put its definition in `ONNXOpsHelper.hpp` if you feel that is cleaner (It would make the Diff far smaller).

src/Dialect/ONNX/ONNXOps.cpp
philass: I've added a comment to explain the case where this happens. 

It can happen in the case when the dimension of the filter size is 1.

example: 

`outputSize = 10, strideVal = 2, kernelSize = 1, dilationVal = 1, inputSize = 20`

then 
```
sumOfPad = (outputSize - 1) * strideVal + ((kernelSize - 1) * dilationVal + 1) - inputSize
                      = (10 - 1) * 2 + ((1 - 1) * 1 + 1) - 20
                      = 18 + 1 - 20
                      = - 1

src/Dialect/ONNX/ONNXOps.cpp
philass: Good observation, I've added a `processConvKernelShapeParam` function

tungld:
> I can keep the function implementation ONNXOps.cpp and put its definition in ONNXOpsHelper.hpp if you feel that is cleaner (It would make the Diff far smaller).

Yes, please make this change. Then it's good to go. Thanks!

src/Dialect/ONNX/ONNXOps.cpp
AlexandreEichenberger: We were initially rewriting attributes, but found it easier to simple compute it into shape inference. 
The main problem is that when you change attributes, its a "compiler constant only" situation. But there are cases where it is not compile time constant (e.g the padding), so I found it simpler to have the pads reside into the shape inference helper object, so that we can always fetch it from there, as opposed to "if it is constant, get it from the attribute, if its run time, get it from elsewhere".

I am happy to have this as a topic of discussion in our weekly meeting.

AlexandreEichenberger: Looking at the code, it looks like we have changed the attributes in the past, so it make sense to continue to do so. So indeed putting it into canonicalization may make sense.

tungld (approved):
LGTM. Thanks for the changes!

PR #1782, opened 2022-10-12
finished DequantizeLinear verify

### LINKED ISSUE #1784: error: onnx.DequantizeLinear: 'axis' value is 1, accepted range is [-1, 0]
happens for resnet50-v1-12-qdq from the model zoo:
https://www.onnxmlir.xyz/jenkins/job/ONNX-MLIR-Pipeline-Docker-Build/Model_20Zoo_20Report/

the problem op is:
```
%469 = "onnx.DequantizeLinear"(%8, %9, %10) {onnx_node_name = "ConvBnFusion_BN_B_resnetv17_batchnorm0_beta_DequantizeLinear"} : (tensor<64xi32>, tensor<1xf32>, tensor<1xi32>) -> tensor<*xf32>
```

sorenlassen:
and simplified its inferShapes implementation

added lit tests

closes #1784

### DIFF
--- a/src/Dialect/ONNX/ONNXOps.cpp
+++ b/src/Dialect/ONNX/ONNXOps.cpp
@@ -3172,60 +3172,75 @@ LogicalResult ONNXQuantizeLinearOp::inferShapes(
 //===----------------------------------------------------------------------===//
 
 LogicalResult ONNXDequantizeLinearOp::verify() {
-  auto xTy = x().getType().cast<ShapedType>();
-  auto scaleTy = x_scale().getType().cast<ShapedType>();
+  // Is tensor known to be a scalar (rank 0 or rank 1 with 1 element)?
+  auto isScalar = [](RankedTensorType t) {
+    return t.getRank() == 0 || (t.getRank() == 1 && t.getDimSize(0) == 1);
+  };
+  // Is tensor known to be a non-scalar 1-D vector (longer than 1)?
+  auto isNonScalar1D = [](RankedTensorType t) {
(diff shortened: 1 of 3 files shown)


src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -3172,41 +3172,83 @@ LogicalResult ONNXQuantizeLinearOp::inferShapes(
...
-
-  auto yTy = y().getType().cast<ShapedType>();
-
-  if (!yTy.hasStaticShape()) {
-    FloatType f32 = FloatType::getF32(getContext());
-    RankedTensorType outType = RankedTensorType::get(inTy.getShape(), f32);
-    y().setType(outType);
-  }
+  if (auto xTy = x().getType().dyn_cast<RankedTensorType>())
+    updateType(y(), xTy.getShape());
```
chentong319, line 3251: No need to add. It seems possible to me that  the shape[0] of zero() or scale() can be used to refine X and Y when X and Y are unranked and zero or scale is 1D. It's a corner case.

chentong319 (approved):
LGTM! 
If you want to add the corner case for shape inference, let me know. I will not merge this PR until you decide.

src/Dialect/ONNX/ONNXOps.cpp
sorenlassen: > It seems possible to me that the shape[0] of zero() or scale() can be used to refine X and Y when X and Y are unranked and zero or scale is 1D.

good idea, I implemented this

I also added one more check in verify, namely x_zero_point non-scale 1-D rank should match input axis dim size

please take another look

PR #1785, opened 2022-10-13
inline function calls during onnx import

sorenlassen:
solves #1787


chentong319:
For inlining, I think that we are going to use the inline support from MLIR, aren't we? The FrontendDialectTransformer.cpp will remain the same, leaving option of inlining or cloning to the later pass. The MLIR tutorial described an inlining example (https://mlir.llvm.org/docs/Tutorials/Toy/Ch-4/#inlining). Should try that with onnx-mlir?

sorenlassen:
> are we going to use the inline support from MLIR?

@chentong319 I have reached the conclusion that MLIR functions are a bad fit, please see my last comment in issue #2185

sorenlassen:
abandoning this PR, this effort has moved to PR #2242

PR #1792, opened 2022-10-18
Clean up gen_onnx_mlir.py

chentong319:
No real change. Just fixed a redefinition of variable. Add some comments. Removed an unneeded comment in generated code.
Not really change the behavior of the compiler


tungld (approved):
LGTM.

PR #1762, opened 2022-10-04
Introduction of ONNX layout for SIMD optimizations

AlexandreEichenberger:
New PR derived from #1709, which add data layout in ONNX that are good for SIMD exploitation. It is still in the original ONNX dialect as I believe it will be beneficial for other applications. However, it will only be activated by selected flags, currently it is off as it is only a partial result.

Example of transformation:
```mlir
    %1 = "onnx.Conv"(%arg0, %arg1, %0) {auto_pad = "NOTSET", kernel_shape = [2, 2], pads = [0, 0, 0, 0]} : (tensor<5x3x32x32xf32>, tensor<?x3x2x2xf32>, none) -> tensor<5x?x31x31xf32>
```

becomes

```mlir
    %1 = "onnx.LayoutTransform"(%arg0) {layout = "UNDEFINED"} : (tensor<5x3x32x32xf32>) -> tensor<5x3x32x32xf32, #onnx.encoding<{dataLayout = "NCHW4C"}>>
    %2 = "onnx.LayoutTransform"(%arg1) {layout = "UNDEFINED"} : (tensor<?x3x2x2xf32>) -> tensor<?x3x2x2xf32, #onnx.encoding<{dataLayout = "KCMN4C4K"}>>
    %3 = "onnx.Conv"(%1, %2, %0) {auto_pad = "NOTSET", group = 1 : si64, kernel_shape = [2, 2], pads = [0, 0, 0, 0]} : (tensor<5x3x32x32xf32, #onnx.encoding<{dataLayout = "NCHW4C"}>>, tensor<?x3x2x2xf32, #onnx.encoding<{dataLayout = "KCMN4C4K"}>>, none) -> tensor<5x?x31x31xf32>
    %4 = "onnx.LayoutTransform"(%3) {layout = "UNDEFINED"} : (tensor<5x?x31x31xf32>) -> tensor<5x?x31x31xf32>
```

Next steps will be to lower the new ops to Krnl dialect. Note that `layout="UNDEFINED"` is the original ONNX layout. Maybe I could use a better name for it. Suggestions welcome.

PR includes a lit test.

PR #1793, opened 2022-10-18
Lowering of ONNX LayoutTransform to Krnl

AlexandreEichenberger:
Lower ONNX Layout transform ops to Krnl IR, generating the proper maps for the data.

E.g.
```mlir
    %0 = "onnx.LayoutTransform"(%arg0) {target_layout = "NCHW4C"} : (tensor<5x3x32x32xf32>) -> tensor<5x3x32x32xf32, #onnx.encoding<{dataLayout = "NCHW4C"}>>
```

becomes
```mlir
#map = affine_map<(d0, d1, d2, d3) -> (d0, d1 floordiv 4, d2, d3, d1 mod 4)>

    %0 = memref.alloc() {alignment = 16 : i64} : memref<5x3x32x32xf32, #map>
    %1:4 = krnl.define_loops 4
    krnl.iterate(%1#0, %1#1, %1#2, %1#3) with (%1#0 -> %arg1 = 0 to 5, %1#1 -> %arg2 = 0 to 3, %1#2 -> %arg3 = 0 to 32, %1#3 -> %arg4 = 0 to 32){
      %4:4 = krnl.get_induction_var_value(%1#0, %1#1, %1#2, %1#3) : (!krnl.loop, !krnl.loop, !krnl.loop, !krnl.loop) -> (index, index, index, index)
      %5 = krnl.load %arg0[%4#0, %4#1, %4#2, %4#3] : memref<5x3x32x32xf32>
      krnl.store %5, %0[%4#0, %4#1, %4#2, %4#3] : memref<5x3x32x32xf32, #map>
    }
```

Corrected also spelling mistakes here and there. 
In addition, there were many header files without `#pragma once` which caused an error, so systematically added the pragma where missing.


src/Conversion/ONNXToKrnl/CMakeLists.txt
```diff
@@ -8,8 +8,10 @@ add_onnx_mlir_library(OMONNXToKrnl
   ControlFlow/If.cpp
   ControlFlow/Loop.cpp
   ControlFlow/Scan.cpp
-  Math/CumSum.cpp
+  ConvertONNXToKrnl.cpp
```
AlexandreEichenberger, line 11: Added one new file, and sorted the old ones that were not in alphabetical order

src/Accelerators/NNPA/Conversion/ZHighToZLow/ZHighToZLow.cpp
```diff
@@ -47,14 +47,17 @@ namespace zhigh {
 /// layout into this map, so that we can obtain the layout for the MemRef later
 /// when lowering other ops.
 llvm::SmallMapVector<mlir::Value, mlir::StringAttr, 4> stickedLayouts;
-mlir::StringAttr readLayout(mlir::Value val) { return stickedLayouts[val]; }
-void storeLayout(mlir::Value val, mlir::StringAttr layout) {
+static mlir::StringAttr readLayout(mlir::Value val) {
```
AlexandreEichenberger, line 50: added static as only used here, and fixed spelling mistakes

src/Dialect/ONNX/AdditionalONNXOps.td
```diff
@@ -195,13 +195,13 @@ def ONNXLayoutTransformOp:ONNX_Op<"LayoutTransform", [NoSideEffect,
     This operation is not part of the standard and was added to assit onnx-mlir.
   }];
   let arguments = (ins AnyTypeOf<[TensorOf<[F32]>, ONNXTensor_NCHW4C,
-      ONNXTensor_KCMN4C4K]>:$In,
+      ONNXTensor_KCMN4C4K]>:$data,
```
AlexandreEichenberger, line 198: Changed "In" to data to be more inline with other operators.

src/Conversion/ONNXToKrnl/Tensor/LayoutTransform.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+    // Operands and attributes.
+    Value data = operandAdaptor.data();
+
+    // Convert the input type to MemRefType.
+    Type inConvertedType = typeConverter->convertType(data.getType());
+    assert(inConvertedType && inConvertedType.isa<MemRefType>() &&
+           "Failed to convert type to MemRefType");
+    MemRefType inMemRefType = inConvertedType.cast<MemRefType>();
+    uint64_t rank = inMemRefType.getShape().size();
```
tungld, line 43: Why don't we get rank from outMemRefType? Looking at the loop below, I guess the input and output must have the same rank.

AlexandreEichenberger: Yes, the LayoutTransform operation by definition have inputs and outputs of the same logical rank. The layout, when not the default one, introduce a map which will increase the dimensionality by 1 or 2 (depending on the current layouts) of the MAPPED data. Let me add a comment in the code to stress that point.

tungld: But it looks like we don't have the mapped data at this stage unless we normalize memrefs before this lowering.
Anyway, we can get rank from dataBounds also.

AlexandreEichenberger: done

tungld (approved):
LGTM.

By the way, I see this message `hi alex: success in converting conv to matmul` when running convolution models. It was added in #1762. Could you remove it in this patch?

PR #1794, opened 2022-10-18
replace buffer pool with resource pool

### LINKED ISSUE #1718: [RFC] Constant propagation and folding using DenseResourceElementsAttr
In onnx-mlir, we have `--constprop-onnx` pass that is used to do constant propagation, e.g. when all inputs of an operation are constants, the pass will replace the operation by a new constant computed during compile time. More info about `--constprop-onnx` can be found here: https://github.com/onnx/onnx-mlir/blob/main/docs/ConstPropagationPass.md.

All constants in onnx-mlir are represented by DenseElementsAttr. It's important to note that DenseElementsAttr are immortal objects living in MLIR context (See [1](https://discourse.llvm.org/t/attribute-storage-memory-consumption/2218) and [2 ](https://discourse.llvm.org/t/mlircontext-doesnt-free-denseelementsattributes-and-bloats-memory/3691/6)for more details). Once we create them, we cannot delete it in the same context.

In `--constprop-onnx`, we maintain a buffer pool to manage all intermediate constants and DenseElementsAttr are only created for the final constant. By this way, we was able to reduce the memory consumption during compile time for resnet152 from 7.1 GB to 0.6GB (See https://github.com/onnx/onnx-mlir/pull/530).

However, there are two issues with the current `--constprop-onnx`:
1. It is painful to work with the buffer pool. We need to mange memory manually. Furthermore, the buffer pool exists in the pass, so it's not sharable across multiple calls of `--constprop-onnx`
2. The code in `--constprop-onnx` is nontrivial to use for the [fold mechanism](https://github.com/llvm/llvm-project/blob/main/mlir/docs/Canonicalization.md#canonicalizing-with-the-fold-method) in MLIR that can be invoked directly anywhere with an `OpBuilder` via `OpBuilder::createOrFold` to produce a constant.

To overcome the issue of immortal attributes, MLIR recently introduces [DenseResourceElementsAttr](https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/IR/BuiltinAttributes.td#L449) that works on [resource blobs](https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/IR/AsmState.h#L83). The important point is that resource blobs supports user-defined deleter that will free the underlying data when the data is not used.

We are using `DenseResourceElementsAttr` in onnx-mlir to store stickified constants for NNPA. See https://github.com/onnx/onnx-mlir/blob/main/src/Accelerators/NNPA/Transform/ZHigh/ZHighConstPropagation.cpp#L100.

I plan to use `DenseResourceElementsAttr` for `--constprop-onnx`. Any comments are welcome and let me know if someone is interested in working on this.

sorenlassen:
resolves #1718

use DenseResourceElementsAttr instead of buffer_id attributes in constant propagation

TODO: complete the replacement (only replaced ConstPropElementwiseUnary/Binary and ConstPropCast so far) and then remove buffer pool

TODO: add commentary to explain the new solution

TODO: replace DenseResourceElementsAttr with another ElementsAttr which prints nicely for lit tests, and then use it beyond constant propagation (ResourcePool and ResourceGarbageCollector are designed for that)


chentong319:
@sorenlassen I found PR#1799 depends on this PR: the special Attribute type is needed to materialize from after a constant op is folded to an attribute. What's your plan for this PR?

sorenlassen:
> What's your plan for this PR?

I'm working on a new version that introduces a custom attribute in place of DenseResourceElementsAttr (which has some important limitations). I hope to be able to share a draft tomorrow.

sorenlassen:
@chentong319 sorry for taking so long but I finally put together a draft PR for using a new attribute DisposabieElementsAttr instead of DenseResourceElementsAttr: https://github.com/sorenlassen/onnx-mlir-einsum/pull/1

discussed here: https://github.com/onnx/onnx-mlir/issues/1718#issuecomment-1322115440

PR #1796, opened 2022-10-19
onnx dialect as LLVM external project

christopherbate:
"onnx dialect as LLVM external project" 

What it means:
- ONNXDialect will be built with LLVM when you build MLIR upstream. Effectively it becomes an extension of the `llvm-project/mlir` source tree. 

Here's what I did:

- Create directory structure matching https://github.com/iree-org/iree/tree/main/llvm-external-projects
- Reorganize `src/Dialects/ONNX/` into the directory structure
- Put `src/Support` and `src/Interface` into the directory structure
- Redo all the CMakeLists files to be compatible with the llvm build. This is actually pretty simple. I didn't even look at the existing ones.
- namespace renaming to put ops in `mlir::onnx_mlir::` instead of `mlir::`

> Naming/namespace usage as-is (putting ops in `::mlir::`) is undesirable for probably many downstream consumers. I see using `mlir::onnx` is probably not good either due to ambiguity with `::onnx`, so I put everything in `::mlir::onnx_mlir`, which is mostly accomplished through a find-and-replace of `namespace onnx_mlir {` to `namespace mlir::onnx_mlir {`. Name sounds  weird, but it's a trade off with doing a more complicated re-name.

- Remove usage of `::onnx::` utils since downstream user might not have onnx library installed. There's only 2-3 such usages for this type of pattern:

```
if (dtypeAttr()) {
    elementType = convertONNXTypeToMLIRType(builder,
        onnx::TensorProto_DataType)dtypeAttr().getValue().getSExtValue());
  } else {
    elementType = builder.getF32Type();
  }
```

> Can you guys remove  this `onnx::` dependency? Then ONNXDialect would be dependency free.

- some modifications for LLVM at `main:HEAD` such as setting ONNXDialect accessor types to non-prefixed.
- some other hacks I can't remember, but it wasn't much.

Incredibly, it builds with LLVM at `main:HEAD`:

```
git clone https://github.com/llvm/llvm-project.git
mkdir my_build
cd onnx-mlir/llvm-external-projects/onnx-mlir-dialects
./build_tools/build_standalone.sh $PWD/my_build $PWD/llvm-project
```

## What this would mean for onnx-mlir build

Obviously I haven't showed how to replace `Dialect/ONNX` in the build as I just copied the files into the new system. 

As suggested above, you would change your build so that you would build the `llvm-external-projects` repo when you build LLVM/MLIR. You could also make it compatible with pre-installed LLVM (I think), but building with LLVM is easiest. Also see how IREE does their integration.

This effectively treats everything in `llvm-external-projects` repo (i.e. ONNXDialect) as a dependency of the rest of the project, so you can't have dependencies flowing back from other parts of the build into `llvm-external-projects`.

This has the following advantages:
1. As I've shown above, any downstream user can co-build your core dialects when they build LLVM as part of their project
2. Upstream build system essentially handles most of the work, no need to build a complicated build system predicating every feature.
3. You've isolated your core components from the rest of the repo, and if a developer introduces a dependency going the wrong way, they will automatically break the build.


AlexandreEichenberger:
Thanks, this explains a bit more concretely what you are trying to achieve. In order to process the protobuff input, we currently need the third_party/onnx and third_party/protobuf sub directories... so that would work with this new scheme?

I can double check on our end the dependences on onnx, but I think it is currently needed. @gongsu832 or @chentong319 do you know more about the depenence on onnx, to parse the inputs (namely the frontend).

christopherbate:
> Thanks, this explains a bit more concretely what you are trying to achieve. In order to process the protobuff input, we currently need the third_party/onnx and third_party/protobuf sub directories... so that would work with this new scheme?
> 
> I can double check on our end the dependences on onnx, but I think it is currently needed. @gongsu832 or @chentong319 do you know more about the depenence on onnx, to parse the inputs (namely the frontend).

It can be supported. I didn't attempt it due to time yesterday. My guess is that most of your users will want the frontend import tool included as well.

christopherbate:
I confirmed that I can build the ONNX dialect and the frontend import tools (src/Builder) on LLVM main head in this manner.

AlexandreEichenberger:
> I confirmed that I can build the ONNX dialect and the frontend import tools (src/Builder) on LLVM main head in this manner.

So the next steps would be to have only one copy of the files, and modify our regular build so that it works also with the new structure.

How does the third_party/onnx works then? Is it added to llvm-project?

christopherbate:
> So the next steps would be to have only one copy of the files, and modify our regular build so that it works also with the new structure.

Yes, that's correct, although I think based on this investigation, I would recommend first making some preparatory changes to make the transition smoother. All the changes below are purely to capture correct dependency relations. Note here that I'm assuming the only things you want to include in `llvm-external-project` folder are the `ONNXDialect` and the frontend import tools under `src/Builder`.

1. Ensure dependencies do not flow from `src/Dialect/` into `src/Support`. This can mostly be fixed by moving almost all files from `src/Dialect/Mlir` into `src/Support` (which seems to make sense to me since `Mlir` is not a dialect). However, there is one method in `src/Support/DialectBuilder` that requires a dependency on `KrnlDialect`. That would need to be fixed/put under `KrnlDialect`.

2. `src/Support/DialectBuilder` has a  lot of dependencies in MLIR  (e.g. VectorDialect, LLVMDialect) that are not actually used when you just build `ONNXDialect` and the frontend import tools to convert `.onnx` files to `.mlir` at the ONNX dialect level. But almost everything depends on `DialectBuilder`, so downstream users will be forced to build these dependencies that are not actually used, increasing binary size, etc.

3. `src/Interfaces` would need to be packaged into the `llvm-external-project` folder as well. You could move it under the ONNX dialect to make that a little clearer, unless there's a dialect that uses any of those interfaces without depending on ONNX dialect.

This whole thing will work **without** making any namespace changes that I mentioned above. I'd still recommend putting the ONNX dialect in a sub-namespace below ::mlir, but maybe that can be a separate debatable issue. 

> How does the third_party/onnx works then? Is it added to llvm-project?

I guess that depends on how you are developing. I think currently you are recommending developers to pre-build and install `llvm-project`? If that's the case, then yes, `third_party/onnx` would get pre-built and installed before hand as well. However, it's possible to add build configurations where your current build method is supported as well as using the items in `llvm-external-projects` separately.

1. "Default": this is your current build method, using pre-installed LLVM project. The items in `llvm-external-project` would be co-built with the rest of your project, not with LLVM. In this case you're not using `llvm-external-project` mechanism, so we could rename that folder to something like `onnx-mlir-core`.
2. "External": your larger project is not built at all, rather someone building `llvm-project` can add everything in `onnx-mlir-core/llvm-external-projects` to their build via the external project mechanism. 
3. "Developer-ease-of-use": Co-build everything with your project, including `llvm-project`, `onnx`, `protobuf` (I think it can be done ;) ), etc. It should make things easier on developers since they won't need to have a multi-stage setup. Keeping in sync with upstream MLIR would also be easier this way.

Option 3 could replace 1 in the long term unless there's a strong argument for option 1?

At this point I have a script that can move/modify all your files automatically to create the "llvm-external-projects" folder. I can share that so we can reliably reproduce the new structure and keep it in sync, as I assume this will take some time to fully plan out.

mgehre-amd:
This looks very useful! Is the work still ongoing?

llvm-external-projects/onnx-mlir/CMakeLists.txt
```diff
@@ -0,0 +1,66 @@
...
+# Setup python.
+################################################################################
+
+if(MLIR_ENABLE_BINDINGS_PYTHON)
+  include(MLIRDetectPythonEnv)
+  mlir_configure_python_dev_packages()
+endif()
+
+add_subdirectory(include)
+add_subdirectory(lib)
```
mgehre-amd, line 60: This failed for me because there is no `lib` directory added in this PR.

AlexandreEichenberger:
I believe the test here was to simply replicate the normal onnx-mlir files and put a copy of them in a place that would be friendly to include with LLVM. 

Obviously, a final solution would not replicate the file but migrate the whole infrastructure in the new place. So work would needs to be done to migrate the current infrastructure to this new way of building onnx-mlir.

If someone is interested in picking up this work, I believe that folks at IBM have no objection as we want this project to work for all. It is not currently a priority item for us as our current deployment scheme does not require this new organization. We see the value of it, though.

If someone were to pickup the work, I would recommend that they restart from scratch from the current main and replicate the same/similar changes as proposed here, as the code has evolved since that test.

Hope this helps.

christopherbate:
I have an alternative to this which requires less invasive changes. I'm going to close this and reopen a separate MR.

PR #1797, opened 2022-10-19
Support compile-only mode in RunONNXModelZoo.py

tungld:
This patch adds an option `--compile-only` to RunONNXModelZoo.py script, which only compiles models in the onnx model zoo.

`--compile-only` can be used to check compilation of the onnx model zoo for NNPA, e.g. 
```bash
python RunONNXModelZoo.py  -c "-O3 --mcpu=z14 --maccel=NNPA" --compile-only
```


utils/RunONNXModel.py
```diff
@@ -90,6 +91,9 @@
                     type=str,
                     default="0.01",
                     help="Absolute tolerance for verification")
+parser.add_argument('--only-compile-model',
```
gongsu832, line 94: Maybe shorter `--compile-only`?

tungld: Yes, sure. Just changed it.

gongsu832 (approved):
LGTM

PR #1799, opened 2022-10-20
Prepare to add constant folding to ONNX Ops

chentong319:
This PR added constant materializer to ONNX dialect and the fold function for ONNXConstantOp.
So we can be ready to add folder function to other Ops in ONNX dialect.


chentong319:
Known error for check-onnx-lit: the order of statements may be different.

chentong319:
@tungld There two lit test failed because empty constant `onnx.constant()` is generated or used in these tests. Why such constant is needed? It is against the semantic to constant. We can not fold other Op if its input come from an empty constant. If we really need such Op, we can define one special Op, like onnx.Data.

tungld:
> There two lit test failed because empty constant onnx.constant() is generated or used in these tests. Why such constant is needed?

In which lit tests did you see empty constants? Did "empty" mean there was no `value` attribute?
Only in constant propagation we can have `buffer_id` instead of `value` I think.

AlexandreEichenberger:
Great to hear you and @sorenlassen  are both trying to help out with better support for constant handling and folding. Can you two @chentong319 and @sorenlassen ensure that both efforts are compatible, thanks.

chentong319:
> > There two lit test failed because empty constant onnx.constant() is generated or used in these tests. Why such constant is needed?
> 
> In which lit tests did you see empty constants? Did "empty" mean there was no `value` attribute? Only in constant propagation we can have `buffer_id` instead of `value` I think.

`buffer_id` is handled in this PR.
In onnx_canonicalization.mlir:
```
func.func @test_conv_batchnormtestmode_fusion_nobias(%arg0 : tensor<1x3x224x224xf32>) -> tensor<1x64x112x112xf32> {
    %0 = **"onnx.Constant"()** : () -> tensor<64x3x7x7xf32>
    %1 = "onnx.Conv"(%arg0, %0, %cst) {auto_pad = "NOTSET", dilations = [1, 1], group = 1 : si64, kernel_shape = [7, 7], pads = [3, 3, 3, 3], strides = [2, 2]} : (tensor<1x3x224x224xf32>, tensor<64x3x7x7xf32>, none) -> tensor<1x64x112x112xf32>
```
onnx_elide_constant.mlir will generate such constant. @AlexandreEichenberger Should we delete ElideConstant related code? I remember it is used when we build a map between constant Op and it value , and that approach is not used anymore. Right?

chentong319:
> Great to hear you and @sorenlassen are both trying to help out with better support for constant handling and folding. Can you two @chentong319 and @sorenlassen ensure that both efforts are compatible, thanks.

This PR is orthogonal to @sorenlassen's PR#1794.  This PR tries to enable the fold mechanism provided by MLIR. PR#1794 changes how Constant is represented in ConstantOp . If PR#1794 is added, this PR has to add code to handle the attribute introduced by PR#1794.  PR#1794 has at least one advantage in the eye of this PR: an unique attribute type is used for the constant representation and there will be no confusion when the constant Op is folded to an attribute.

sorenlassen:
I agree that this PR is orthogonal to PR #1794

chentong319:
I looked into the problem in onnx_canonicalization.mlir further.  I will replace those empty constantOp with function arguments if nobody objects.

chentong319:
I just found that this PR depends on PR#1794 for the special Attribute Type to represent the buffer. Otherwise, it will have trouble for ConstantOp with value_int() attribute.

tungld:
> onnx_elide_constant.mlir will generate such constant.

I believe we no longer use our own constant eliding that does not work with any kind of constant operations, e.g. constant operations in ONNX, Krnl, ZHigh dialects. We have changed to use MLIR attribute eliding where attributes instead of operations are shorten.

sorenlassen:
> > onnx_elide_constant.mlir will generate such constant.
> 
> I believe we no longer use our own constant eliding

I deleted onnx constant eliding in PR #1899

chentong319:
Replaced with PR#2009

PR #1800, opened 2022-10-20
Create unique QuestionMarkIndexExpr by using a hash code of tensor and axis

tungld:
Problem: given a tensor value, to do shape computation we call `MemRefBoundsIndexCapture(val)` e.g.
```c
Value data = operandAdaptor.data();
MemRefBoundsIndexCapture dataBounds(data);
```
Assume that axis 2 is an unknown dimension and now we get the QuestionMarkIndexExpr of it.
```c
QuestionMarkIndexExpr dim = dataBounds.getDim(2);
int64_t questionMarkValue = dim.getQuestionmark();
```
The problem is: everytime we call `dataBounds.getDim(2)`, a new QuestionMarkIndexExpr is created and its value is changed.

To keep the question mark value unchanged for a specific axis of a tensor, this patch uses a hash_code, that is a hash combination of the tensor value and axis, for the question mark value. 

### DIFF
--- a/src/Dialect/Mlir/IndexExpr.cpp
+++ b/src/Dialect/Mlir/IndexExpr.cpp
@@ -1165,11 +1165,12 @@ QuestionmarkIndexExpr::QuestionmarkIndexExpr() : IndexExpr() {
   indexExprObj->initAsQuestionmark();
 }
 
-QuestionmarkIndexExpr::QuestionmarkIndexExpr(Value val, int64_t axis)
+QuestionmarkIndexExpr::QuestionmarkIndexExpr(
+    Value tensorOrMemref, int64_t index)
     : IndexExpr() {
   indexExprObj = new IndexExprImpl();
   assert(indexExprObj && "failed to allocate IndexExpr implementation");
-  indexExprObj->initAsQuestionmark(val, axis);
+  indexExprObj->initAsQuestionmark(tensorOrMemref, index);
 }
 // Don't care about otherIndexExpr as question marks have no real data.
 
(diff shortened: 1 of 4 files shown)


src/Dialect/Mlir/IndexExpr.cpp
```diff
@@ -1165,6 +1165,12 @@ QuestionmarkIndexExpr::QuestionmarkIndexExpr() : IndexExpr() {
   indexExprObj->initAsQuestionmark();
 }
 
+QuestionmarkIndexExpr::QuestionmarkIndexExpr(Value val, int64_t axis)
```
AlexandreEichenberger, line 1168: I think the val here is of type Tensor or Memref, for this to work, correct? Maybe you could rename `val` to be bit more explicit. Maybe `tensorOrMemRefVal` or `arrayVal` and `axis` becoming `index`. It will help comprehension. 

Please change in other similar instances below

src/Dialect/Mlir/IndexExprDetail.cpp
```diff
@@ -55,9 +50,18 @@ void IndexExprImpl::initAsUndefined() {
...
-  const std::lock_guard<std::mutex> lock(indexExprQuestionMarkMutex);
+  // Question mark has value of -1 by default.
+  init(/*isDefined*/ true, /*literal*/ false, IndexExprKind::Questionmark, -1,
+      AffineExpr(nullptr), Value(nullptr));
+}
+
+void IndexExprImpl::initAsQuestionmark(Value val, int64_t axis) {
+  // Each question mark is assigned a unique integer that is obtained
+  // by hashing the tensor value and the target axis/dimension.
+  llvm::hash_code questionValue =
```
AlexandreEichenberger, line 61: Can you just indicate in the code what is the scope of this hash? And where it is declared, tx.

AlexandreEichenberger (approved):
LGTM, please look at name and comment suggestions I made before merging. Tx.

If `MemRefBoundsIndexCapture` is the gateway to get the "fixed" value for question marks, I would suggest also to add comments to that effect in the header file, namely that for symbolic analysis to work, one need to use that call or the specific question mark constructor with value & axis...

PR #1801, opened 2022-10-21
Use IndexExpr to compute shape for unary element-wise ops

tungld:
Though it is quite simple to infer shape for unary element-wise operations, this patch uses IndexExpr for the shape inference to help analyze unknown dimensions in a systematic way in the future.


AlexandreEichenberger (approved):
LGTM

tungld:
> PS: I see the error in Windows is due to a warning treated as an error... would it make sense to also enable this on Linux so that we may detect them while developing the code (on Linux or macOS for many of us)?

Thanks, Alex! I fixed the error. Yes, I think it makes sense to enable it on Linux.

PR #1810, opened 2022-10-26
Change split(' ') to split() in RunONNXModel.py to avoid an error about command-line option

imaihal:
Small update in RunONNXModel.py to avoid an error about command-line option.

I got an error when using "--compile-args -O3&nbsp;&nbsp;--mtriple=s390x-ibm-loz --mcpu=z14".  "--compile-args "-O3&nbsp;--mtriple=s390x-ibm-loz --mcpu=z14" works instead. Looks the same, but the first one has two space between `-O3` and `--mtriple`. 

--compile-args "-O3&nbsp;&nbsp;--mtriple=s390x-ibm-loz --mcpu=z14"  ==> Error
--compile-args "-O3&nbsp;--mtriple=s390x-ibm-loz --mcpu=z14"  ==> No error

To avoid this error, we can use default separator in split().

### DIFF
--- a/utils/RunONNXModel.py
+++ b/utils/RunONNXModel.py
@@ -321,7 +321,7 @@ def main():
             # Prepare compiler arguments.
             command_str = [ONNX_MLIR]
             if args.compile_args:
-                command_str += args.compile_args.split(' ')
+                command_str += args.compile_args.split()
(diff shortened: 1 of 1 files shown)


tungld (approved):
LGTM

PR #1813, opened 2022-10-28
Fix RuntimeAPIRegistry singleton class memory leak.

gongsu832:
The singleton class pointer was set to null between calls to compileModule(), which leads to memory leak. The idea of a singleton is that it should last for the lifetime of the process. However, the fact that it cannot be shared across compileModule() suggests that it's not the right construct to use. In addition, the singleton calss is not thread safe without proper serialization.

This patch changes the singleton to a normal class. It's verified to fix the memory leak reported by valgrind and also makes the code thread safe.


tungld (approved):
LGTM.

chentong319:
You mentioned at meeting that it did not work to add delete before setting to null. Do you know why?
To my understanding, this change will create multiple instances of RuntimeAPIRegistry. Any impact?

gongsu832:
> You mentioned at meeting that it did not work to add delete before setting to null. Do you know why? To my understanding, this change will create multiple instances of RuntimeAPIRegistry. Any impact?

Yes. Because the delete was inside the destructor, so it was deleting itself again and again to infinity.

I don't know what determines the number of instances of RuntimeAPIRegistry to be created. I tried all the numerical tests and also compiling a few models such as inception, lstm, mnist, resnet50, and squeeznet and they all show that one instance was created for a call to compileModule().

tungld:
There would be one instance of RuntimeAPIRegistry for each compilation, which is used to look up and emit external function declarations into LLVM IR, e.g. OMTensor related functions or zDNN functions.

gongsu832:
> There would be one instance of RuntimeAPIRegistry for each compilation, which is used to look up and emit external function declarations into LLVM IR, e.g. OMTensor related functions or zDNN functions.

OK then that means there was no reason to use a singleton at all to begin with. I guess whoever wrote the code intended to share a single instance across multiple compileModule() calls but got into trouble with numerical tests. So he/she just reset the instance pointer to null in the destructor, causing a new singleton to be created each time when compileModule() is called. This effectively made the singleton class a normal class, but also introduced the memory leak.

PR #1814, opened 2022-10-28
Bug fixes for RunONNXModel.py

Silvan-K:
This PR addresses some minor bugs in RunONNXModel.py:

-  The option `--verify-all-ops` did not work in my use case because the intermediate tensors that were added to the models as outputs had no type associated to them which caused onnx-mlir to reject the model when trying to compile it
- The option `--load-so` did not work in combination with `--verify onnxruntime` did not work because the modified onnx model was not saved to the temporary directory


tungld:

utils/RunONNXModel.py
```diff
@@ -147,13 +147,31 @@ def execute_commands(cmds):
...
 def extend_model_output(model, intermediate_outputs):
+
+    # Run shape inference to make sure we have valid tensor value infos for all
+    # intermediate tensors available
+    model = onnx.shape_inference.infer_shapes(model)
+    value_infos = {vi.name: vi for vi in model.graph.value_info}
+    graph_inputs = {vi.name: vi for vi in model.graph.input}
+    graph_outputs = {vi.name: vi for vi in model.graph.output}
+
+    # Retrieven tensor value info for each intermediate output
```
tungld, line 158: typo: Retrieven => Retrieve

utils/RunONNXModel.py
```diff
@@ -271,8 +289,9 @@ def main():
     # Load the onnx model.
     model = onnx.load(args.model_path)
 
-    # Get the output names that we want to verify.
-    # If using onnxruntime for verification, we can verify every operation output.
+    # Get names of all intermediate tensors and modify model such that each of
+    # them will is an output of the model. If using onnxruntime for
```
tungld, line 293: `will is` -> `will be`

tungld (changes requested):
Thank you for the fixes! Please fix some typos and it's good to go.

utils/RunONNXModel.py
silvan-lemurian: Thanks, fixed the typo

utils/RunONNXModel.py
silvan-lemurian: Thanks, fixed the typo

tungld:

tungld (approved):
LGTM. Thanks!

PR #1815, opened 2022-10-31
Generalize layout propagation and clean up some unused functions

tungld:
Before this patch layout propagation is only applied to NHWC layout, e.g.
```
[X_ztensor] -> Unstick (NHWC) -> Stick (4D) -> zhigh.Relu
```
The layout NHWC  is propagated to Relu, and leads to:
```
[X_ztensor] -> zhigh.Relu -> Unstick (NHWC) -> Stick (4D)
```

Since unary and binary operations can work with arbitrary layout, this patch generalizes the layout propagation to apply to arbitrary layout, for example,
```
[X_ztensor] -> Unstick (3DS) -> Stick (3D) -> zhigh.Relu
```
will become
```
[X_ztensor] -> zhigh.Relu -> Unstick (3DS) -> Stick (3D)
```

Unstick-stick pairs will be automatically eliminated by optimizations for unstick and stick operations.


src/Accelerators/NNPA/Transform/ZHigh/ZHighLayoutPropagation.td
```diff
@@ -31,185 +31,221 @@ include "src/Accelerators/NNPA/Dialect/ZHigh/ZHigh.td"
...
+// other rules can be applied to remove unstick-stick pairs.For example,
+// This sequence of operations:
+//   zhigh.Conv -> Unstick (NHWC) -> Stick (4D) -> zhigh.Relu
+// will become:
+//   zhigh.Conv -> zhigh.Relu -> Unstick (NHWC) -> Stick (4D)
+// then, canonicalization of unstick/stick will remove the unstick-stick pair.
+//
+//===----------------------------------------------------------------------===//
+
+// Exp
```
AlexandreEichenberger, line 63: Quick question: is it possible to make a rule that 1) detect if the op has the "layout oblivious" property and if it does, that the stick/unstick ops are moved around? It is true that the set of ops is more limited that the ONNX standard, so there are not that many, but it might still be a nice feature to have.

tungld: Yes, it's possible, just use tablegen for easy tracking since there are not many zHigh ops as you know. If you prefer I can change it to writing C++ instead of tablegen. Perhaps, we would have the same thing for CPU layout in the future.

AlexandreEichenberger (approved):
LGTM

PR #1819, opened 2022-10-31
Fix QuantizeLinear shape inference

philass:
Fix QuantizeLinear shape inference in cases where type needs to be read from zero point attribute

### DIFF
--- a/test/mlir/onnx/onnx_shape_inference.mlir
+++ b/test/mlir/onnx/onnx_shape_inference.mlir
@@ -1555,6 +1555,25 @@ func.func @test_quantize_linear_1(%arg0 : tensor<5x2x3x4xf32>, %arg1 : tensor<f3
   // CHECK: return [[RES]] : tensor<5x2x3x4xi8>
 }
 
+func.func @test_quantize_linear_2(%arg0 : tensor<5x2x3x4xf32>, %arg1: tensor<f32>, %arg2: tensor<ui8>) -> tensor<*xui8> {
+ %0 = "onnx.QuantizeLinear"(%arg0, %arg1, %arg2) {} : (tensor<5x2x3x4xf32>, tensor<f32>, tensor<ui8>) -> tensor<*xui8>
+ "func.return"(%0) {} : (tensor<*xui8>) -> ()
+
+ // CHECK-LABEL: test_quantize_linear_2
+ // CHECK: [[RES:%.+]] = "onnx.QuantizeLinear"(%arg0, %arg1, %arg2) {axis = 1 : si64} : (tensor<5x2x3x4xf32>, tensor<f32>, tensor<ui8>) -> tensor<5x2x3x4xui8>
+ // CHECK: return [[RES]] : tensor<5x2x3x4xui8>
+}
+
+func.func @test_quantize_linear_3(%arg0 : tensor<5x2x3x4xf32>, %arg1: tensor<f32>) -> tensor<*xui8> {
+%none = "onnx.NoValue"() {value} : () -> none
+ %0 = "onnx.QuantizeLinear"(%arg0, %arg1, %none) {} : (tensor<5x2x3x4xf32>, tensor<f32>, none) -> tensor<*xui8>
+ "func.return"(%0) {} : (tensor<*xui8>) -> ()
+
+ // CHECK-LABEL: test_quantize_linear_3
+ // CHECK: [[RES:%.+]] = "onnx.QuantizeLinear"(%arg0, %arg1, %0) {axis = 1 : si64} : (tensor<5x2x3x4xf32>, tensor<f32>, none) -> tensor<5x2x3x4xui8>
+ // CHECK: return [[RES]] : tensor<5x2x3x4xui8>
+}
+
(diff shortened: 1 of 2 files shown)


src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -3267,10 +3267,18 @@ LogicalResult ONNXQuantizeLinearOp::inferShapes(
...
-    // TODO: Unfortunately, we can't tell if this should be signed or
-    // unsigned here...
-    IntegerType i8Type = IntegerType::get(getContext(), 8);
-    RankedTensorType outType = RankedTensorType::get(inTy.getShape(), i8Type);
+    auto zero = y_zero_point();
+    RankedTensorType outType;
+    // If zero point type isn't provided, output type defaults to ui8.
+    if (isFromNone(zero)) {
+      auto builder = mlir::Builder(this->getContext());
+      IntegerType ui8Type = builder.getIntegerType(8, false);
```
sorenlassen, line 3275: nit: I think the most readable and unambiguous way to express this (and gets rid of the builder) is:
```
IntegerType::get(getContext(), 8, IntegerType::Unsigned)
```

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -3267,10 +3267,18 @@ LogicalResult ONNXQuantizeLinearOp::inferShapes(
...
+      outType = RankedTensorType::get(inTy.getShape(), ui8Type);
+    } else {
+      // If zero point is provided, output type is same as zero point type.
+      auto zpTy = zero.getType().cast<ShapedType>().getElementType();
+      outType = RankedTensorType::get(inTy.getShape(), zpTy);
```
sorenlassen, line 3280: nit: to make naming consistent `zpTy` should refer to its tensor type, so you should write:
```
auto zpTy = zero.getType().cast<ShapedType>();
outType = RankedTensorType::get(inTy.getShape(), zpTy.getElementType());
```

test/mlir/onnx/onnx_shape_inference.mlir
```diff
@@ -1555,6 +1555,16 @@ func.func @test_quantize_linear_1(%arg0 : tensor<5x2x3x4xf32>, %arg1 : tensor<f3
   // CHECK: return [[RES]] : tensor<5x2x3x4xi8>
 }
 
+func.func @test_quantize_linear_2(%arg0 : tensor<5x2x3x4xf32>, %arg1: tensor<f32>, %arg2: tensor<ui8>) -> tensor<*xui8> {
```
sorenlassen, line 1558: for completeness you could add one more lit test with y_zero_point `None`

sorenlassen (approved):

src/Dialect/ONNX/ONNXOps.cpp
philass: Thanks! I was struggling to find `IntegerType` in the mlir docs. I saw the function on the builder.

But your suggestion is definitely better.

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -3267,10 +3267,18 @@ LogicalResult ONNXQuantizeLinearOp::inferShapes(
...
+    // If zero point type isn't provided, output type defaults to ui8.
+    if (isFromNone(zero)) {
+      IntegerType ui8Type =
+          IntegerType::get(getContext(), 8, IntegerType::Unsigned);
+      outType = RankedTensorType::get(inTy.getShape(), ui8Type);
+    } else {
+      // If zero point is provided, output type is same as zero point type.
+      auto zpTy = zero.getType().cast<ShapedType>();
+      outType = RankedTensorType::get(inTy.getShape(), zpTy.getElementType());
+    }
```
tungld, line 3281: Just wanted to make this more concise, how about 
```C
Type elementType;
if (isFromNone(zero))
  elementType = IntegerType::get(getContext(), 8, IntegerType::Unsigned);
else
  elementType = zero.getType().cast<ShapedType>().getElementType();
auto outType = RankedTensorType::get(inTy.getShape(), elementType);
```

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -3267,10 +3267,18 @@ LogicalResult ONNXQuantizeLinearOp::inferShapes(
   auto yTy = y().getType().cast<ShapedType>();
 
   if (!yTy.hasStaticShape()) {
-    // TODO: Unfortunately, we can't tell if this should be signed or
-    // unsigned here...
-    IntegerType i8Type = IntegerType::get(getContext(), 8);
-    RankedTensorType outType = RankedTensorType::get(inTy.getShape(), i8Type);
+    auto zero = y_zero_point();
```
tungld, line 3270: We are trying to avoid `auto`, it looks like its type is `Value` here.

src/Dialect/ONNX/ONNXOps.cpp
philass: Thanks for the suggestion. That looks a lot cleaner

PR #1821, opened 2022-10-31
Prelu verify fix

### LINKED ISSUE #1820: Prelu verify is overly aggresive
Prelu doesn't make sure that slope and xShape have shape and rank before doing the verify checks. 

For example verify will be overly aggresive in the following case 

```
%30 = "onnx.PRelu"(%28, %29) {onnx_node_name = "PRelu_14"} : (tensor<*xf32>, tensor<64x1x1xf32>) -> tensor<*xf32>
```

below is the Prelu verify method

```cpp
LogicalResult ONNXPReluOp::verify() {
  ArrayRef<int64_t> xShape = X().getType().cast<ShapedType>().getShape();
  ArrayRef<int64_t> slopeShape =
      slope().getType().cast<ShapedType>().getShape();

  // PRelu supports unidirectional broadcasting, that is slope should be
  // unidirectional broadcastable to input X.
  if (slopeShape.size() > xShape.size())
    return emitError("Slope tensor has a wrong shape");
  return success();
}
```

philass:
Closes #1820


tungld (approved):
LGTM.

philass:
Thanks for the review!

PR #1822, opened 2022-11-01
Introduce DimAnalysis class for unknown dimension analysis and apply it to NNPA

tungld:
This patch introduces a **DimAnalysis** class to analyze unknown dimensions in ONNX IR.
In particular, it classifies unknown dimensions into groups/sets. Dimensions in a group are supposed to be equal at runtime time. Each group has an unique ID.

ShapeHelper is used to explore the equality relationship between input and output unknown dimensions.

A pass `--onnx-dim-analysis` is introduced for the testing purpose where it emits `onnx.DimGroup` to the IR to show group IDs.

This patch only supports a limited number of operations (e.g. operations in ResNet50 and BertSquad). More supported operations will be added in future PRs.

As an example, this patch shows how to apply DimAnalysis to ONNXToZHigh pass where we identify if there is broadcasting in binary operations such as Add/Div/Mul/Sub or not. Applying to ResNet50 whose input is `?x3x224x224`, All 16 ONNXAdd ops are identified as non-broadcasting ops and are lowered to zHigh for NNPA.

The number of operations in ResNet50 when **not** using DimAnalysis:
Screen Shot 2022-11-01 at 12 28 47

The number of operations in ResNet50 when using DimAnalysis:
Screen Shot 2022-11-01 at 21 02 22


src/Dialect/ONNX/ShapeInference/Pad.cpp
```diff
@@ -60,7 +60,12 @@ LogicalResult ONNXPadOpShapeHelper::computeShape(
     DimIndexExpr dimInput(dataBounds.getDim(i));
 
     // Calculation for output size.
-    IndexExpr dimOutputFinal = padBegin + dimInput + padEnd;
+    IndexExpr dimOutputFinal;
+    if (padBegin.isLiteral() && padEnd.isLiteral() &&
```
AlexandreEichenberger, line 64: Is the problem that if they are both zero, and the input dim is question mark, that is the only way to preserve the same question mark? 
I think we should change the propagation rules that `question mark + 0` should preserve the question mark in the index expressions. Then you can do that systematically, `? * 1`, `? / 1`, `select(true, ?, x)`...
Then we don't have this exception here, and it would apply to many many more cases.

AlexandreEichenberger (commented):
First rounds of comments, will read the `ONNXDimAnalysis` files next

src/Transform/ONNX/ONNXDimAnalysis.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+#pragma once
+
+#include "mlir/IR/BuiltinTypes.h"
+
+namespace onnx_mlir {
+
+class DimAnalysis {
+public:
+  using DimT = std::pair<mlir::Value, uint64_t>;
```
AlexandreEichenberger, line 23: Can you indicate for each type what they are used for. That would help the comprehension.

src/Transform/ONNX/ONNXDimAnalysis.hpp
```diff
@@ -0,0 +1,72 @@
...
+  /// Create a new analysis for all values in a module.
+  DimAnalysis(mlir::ModuleOp op);
+
+  /// Analyzes the relationship amongs unknown dimensions.
+  void analyze();
+
+  /// Returns the grouping result of unknown dimensions.
+  DimSetMapT getGroupingResult() const { return dimSetMap; }
+
+  /// Test if two unknown dimensions are the same or not.
```
AlexandreEichenberger, line 40: what are `lhs` and `rhs`? Are they value of a shaped type kind? So `tensorOrMemref`? If so, a more explicit name would facilitate the usage of the interface.

src/Transform/ONNX/ONNXDimAnalysis.cpp
```diff
@@ -0,0 +1,588 @@
...
+
+using namespace mlir;
+
+namespace {
+
+//===----------------------------------------------------------------------===//
+// Helper functions for dimension analysis.
+//===----------------------------------------------------------------------===//
+
+/// Check if two sets are interleaved or not.
```
AlexandreEichenberger, line 41: Interleave, do you mean do the the two sets intersect? A set based name would be clearer, maybe `nonEmptyIntersection`

src/Transform/ONNX/ONNXDimAnalysis.cpp
```diff
@@ -0,0 +1,588 @@
...
+  return false;
+}
+
+/// Given a QuestionMarkIndexExpr representing an unknown dimension, find the
+/// same unknown dimensions in the inputs.
+void findAndAddSameDim(const onnx_mlir::QuestionmarkIndexExpr &qmOuputIE,
+    mlir::ValueRange operands, onnx_mlir::DimAnalysis::DimSetT &sameDims) {
+  int64_t outputQMValue = qmOuputIE.getQuestionmark();
+  // Cannot process if the question mark value is -1 that is a common value for
+  // every question mark.
```
AlexandreEichenberger, line 67: I would abstract here a bit the interface.
```
bool qmOutputIE.specificQuestionmark() 
bool qmOutputIE.sameQuestionmark(iIR)
```
whereas the first one assert if IE is not a question mark, and return true if `!= -1`,
and whereas the second one assert if either is not a question mark, and return true if their values are not `-1` and are the same (negative) value.

src/Transform/ONNX/ONNXDimAnalysis.cpp
```diff
@@ -0,0 +1,588 @@
...
+    onnx_mlir::MemRefBoundsIndexCapture vDims(v);
+    for (int64_t i = 0; i < rank; ++i) {
+      onnx_mlir::IndexExpr iIR = vDims.getDim(i);
+      if (iIR.isQuestionmark()) {
+        if (iIR.getQuestionmark() == outputQMValue)
+          sameDims.insert(onnx_mlir::DimAnalysis::DimT(v, i));
+      }
+    }
+  }
+  return;
```
AlexandreEichenberger, line 84: no need for return

src/Transform/ONNX/ONNXDimAnalysis.cpp
```diff
@@ -0,0 +1,588 @@
...
+    }
+  }
+  return false;
+}
+
+/// Given a QuestionMarkIndexExpr representing an unknown dimension, find the
+/// same unknown dimensions in the inputs.
+void findAndAddSameDim(const onnx_mlir::QuestionmarkIndexExpr &qmOuputIE,
+    mlir::ValueRange operands, onnx_mlir::DimAnalysis::DimSetT &sameDims) {
+  int64_t outputQMValue = qmOuputIE.getQuestionmark();
```
AlexandreEichenberger, line 65: Ideally this call would not be needed, because we only query the index expression directly for the info we need. See below.

src/Dialect/ONNX/ShapeInference/Pad.cpp
tungld: Yes, I am aware of this and it's interesting enough to be addressed with a new PR. So, I just do a quick fix for PadOp that is used in ResNet50.

AlexandreEichenberger: Sure, so add a TODO to remove this handling once the proper arithmetic is in.

src/Transform/ONNX/ONNXDimAnalysis.cpp
tungld: Could you elaborate a bit? I don't see "below" things, maybe you forgot pushing them ...

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
```diff
@@ -37,10 +38,11 @@ bool isValidElementType(Value val) {
 }
 
 /// A function to check whether two tensors have the same shape or not.
-/// In case where they have the same rank but unknown dimensions, we cannot
-/// detect whether the shapes are exactly the same or not. Hence, return false.
+/// In case where they have the same rank but unknown dimensions, we use the
+/// dimension analysis to try inferring the equality at compile.
 /// Also, check the ranks of two tensors, they must be in range of (0, 4].
-bool haveSameStaticShape(Value value1, Value value2) {
+bool haveSameShape(
```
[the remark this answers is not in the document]
tungld: I put a TODO here to find a new home for it. I tried to move it the ONNX Ops helper but it caused cyclic dependency: ONNXOp -> DimAnalysis -> ONNXOps -> ...

src/Dialect/ONNX/AdditionalONNXOps.td
```diff
@@ -158,14 +158,39 @@ def ONNXDimOp : ONNX_Op<"Dim",
...
+    "onnx.DimGroup"(%tensor) {axis = 0 : si64, group_id = 1: si64} : (tensor<?x3x5xf32>) -> ()
+    ```
+
+    `axis` identifies the dimension position in the tensor.
+
+    `group_id` identifies the group id of the dimension. It is non-negative.
+    Value -1 for `group_id` means the dimension does not belong to any group.
+
+    This operation is not part of the standard and was added to assist onnx-mlir.
+  }];
```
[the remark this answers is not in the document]
tungld: Added.

src/Transform/ONNX/ONNXDimAnalysis.cpp
tungld: Done. Thanks!

tungld:
@AlexandreEichenberger thanks for your comments. I addressed most of them except the last one I am a bit confused. I also added a document on how to use DimAnalysis and more comments to code as @chentong319 suggested. It's ready for another round of review.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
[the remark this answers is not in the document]
AlexandreEichenberger: The obvious spot would be inside DimAnalysis.

src/Transform/ONNX/ONNXDimAnalysis.cpp
AlexandreEichenberger: To the extend possible, we should not need the `getQuestionmark()` call.
We should have
```
bool qmOutputIE.specificQuestionmark() 
bool qmOutputIE.sameQuestionmark(iIR)
```
where we ask index expression if an index expression question mark is specific (<-1) or not (==-1), and if they are the same (same specific number).

If we were to change the representation, it would not impact the code here.

src/Transform/ONNX/ONNXDimAnalysis.cpp
AlexandreEichenberger: I believe this comment was not resolved, at least as I see it in the code below

src/Transform/ONNX/ONNXDimAnalysis.cpp
tungld: Thanks for th explaination! I addressed this just now.

src/Transform/ONNX/ONNXDimAnalysis.cpp
tungld: Done. I created the two functions but forgot using them. Thanks!

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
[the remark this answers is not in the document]
tungld: I moved it into DimAnalysis. Thanks!

docs/UnknownDimensionAnalysis.md
```diff
@@ -0,0 +1,30 @@
+It is often the case where we want to know two unknown dimensions are equal or not at compile time. This helps with decision on how to lowering an ONNX operator. For example, given an ONNXAddOp as follows:
```
AlexandreEichenberger, line 1: Awesome, thanks for adding this

src/Dialect/ONNX/AdditionalONNXOps.td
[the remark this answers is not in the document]
AlexandreEichenberger: thanks.

src/Transform/ONNX/ONNXDimAnalysis.hpp
AlexandreEichenberger: tx

AlexandreEichenberger (approved):
LGTM, the text of a comment in IndexExpr needs to be corrected, otherwise all good

src/Dialect/Mlir/IndexExpr.hpp
```diff
@@ -650,6 +650,17 @@ class QuestionmarkIndexExpr : public IndexExpr {
   ~QuestionmarkIndexExpr() = default;
   QuestionmarkIndexExpr &operator=(const QuestionmarkIndexExpr &) = delete;
   QuestionmarkIndexExpr &operator=(QuestionmarkIndexExpr &&) = delete;
+
+  // Query functions.
+  // Check if the question mark is specific so that it can be distinguished from
+  // other question marks.
+  bool specificQuestionmark() const;
+  // Check if two question marks are the same or not. Two question marks are the
+  // same if they are not specific and have the same value.
```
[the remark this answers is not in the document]
tungld: Good catch! Fixed it, thanks!

PR #1807, opened 2022-10-25
Update the instrumentation tool

imaihal:
This PR updates the instrumentation tools
- Created dialect-independent pass for instrumentation (`InstrumentPass`) (https://github.com/onnx/onnx-mlir/issues/1790)
  - Removed zhigh-specific instrumentation pass (`InstrumentZHighPass`) 
- Updated command-line options (https://github.com/onnx/onnx-mlir/issues/1790)
  - Newly added `--instrument-stage=<stage name>` to set stage to be instrumented. Following stages are provided.
    - `before-onnx-to-krnl` : Profile for onnx ops (before lowering to krnl)
    - `nnpa-before-onnx-to-zhigh` : [NNPA] Profile for onnx ops
    - `nnpa-before-onnx-to-krnl` : [NNPA] Profile for onnx and zhigh ops
    - `nnpa-before-krnl-to-llvm` : [NNPA] Profile for zlow ops 
  - Changed `--instrument-onnx-ops` to `--instrument-ops`
- Display instrumented `<dialect name>.<ops name>` in profile output at runtime (https://github.com/onnx/onnx-mlir/issues/1790)

Example output (part of yolov3)
  - --maccel=NNPA --instrument-stage=nnpa-before-onnx-to-krnl --instrument-ops=onnx.,zhigh. --InstrumentBeforeOp --InstrumentAfterOp --InstrumentReportTime 
```
#1082) before    zhigh.StickifiedConstant Time elapsed: 0.000000 accumulated: 1666678513.998812
#1083) after     zhigh.StickifiedConstant Time elapsed: 0.000001 accumulated: 1666678513.998813
#1084) before    zhigh.StickifiedConstant Time elapsed: 0.000000 accumulated: 1666678513.998813
#1085) after     zhigh.StickifiedConstant Time elapsed: 0.000001 accumulated: 1666678513.998814
#1086) before    zhigh.Conv2D Time elapsed: 0.000000 accumulated: 1666678513.998814
#1087) after     zhigh.Conv2D Time elapsed: 0.000094 accumulated: 1666678513.998908
#1088) before    zhigh.Unstick Time elapsed: 0.000000 accumulated: 1666678513.998908
#1089) after     zhigh.Unstick Time elapsed: 0.000081 accumulated: 1666678513.998989
#1090) before     onnx.Transpose Time elapsed: 0.000001 accumulated: 1666678513.998990 (Transpose__472)
[...]

PR #1823, opened 2022-11-02
Update docs and options for instrumentation tool

imaihal:
This PR updates for https://github.com/onnx/onnx-mlir/pull/1807

  - Updated stage name in `--instrument-stage=<stage name>` and used enum for this option.
```
  --instrument-stage=<value>                        - Specify stage to be instrumented:
    =Onnx                                             -   Profile for onnx ops. For NNPA, profile onnx ops before lowering to zhigh.
    =ZHigh                                            -   NNPA profiling for onnx and zhigh ops.
    =ZLow                                             -   NNPA profiling for zlow ops.
```
- Update docs for instrumentation tool
- Refactoring
  - Remove stages from Instrumentation pass
  - Separate files for NNPA options (NNPACompileOptions.cpp and .hpp)
  - Move NNPA word under NNPA folder using macro, `APPLY_TO_ACCELERATORS()`


imaihal:

imaihal:
@chentong319 I updated previous PR about instrumentation tool. This PR uses enum for instrumentStage option. Also I changed stage name. These may be easier to understand for users.

include/onnx-mlir/Compiler/OMCompilerTypes.h
```diff
@@ -53,6 +53,13 @@ typedef enum {
   InstrumentReportMemory
 } InstrumentActions;
 
+typedef enum {
+  afterOnnxToOnnx,
+  nnpaAfterOnnxToOnnx,
+  nnpaAfterOnnxToZhigh,
+  nnpaAfterZhighToZlow
```
tungld, line 60: This does not conform to our policy in supporting accelerator. So `nnpa` should not appear if we do not build the compiler with `--ONNX_MLIR_ACCELERATORS=NNPA`. Everything about `nnpa` should be inside the NNPA accelerator folder and work in a plug-and-play manner.

chentong319: This is a tricky issue. NNPA has its own main and accordingly should have its own options for the main. I remember that it was discussed to separate the options for passes from those for main. Currently, there is no NPAACompilerOptions.*. However, for this case, we need to extend the enum type,  similar to the type extension in MLIR. It's too complicated for our purpose. Perhaps, we just use string type for the option to keep the choices open. It's simply but user will not know what are the right options to provide. I can not think of a better solution unless the option utility provides some support.

chentong319: Following this issue, we can add some explanation and example of instrumentStage when an accelerator is enabled or disabled (using NNPA as an example)

docs/Instrumentation.md
```diff
@@ -6,48 +6,61 @@ Instrumentation is prototyped in onnx-mlir and can be used to debug runtime issu
...
 
 ```
-#  0) after  op= Transpo Time elapsed: 0.000766 accumulated: 0.000766 VMem: 156608 (model/transpose1)
-#  1) after  op= Constan Time elapsed: 0.005398 accumulated: 0.006164 VMem: 156608
-#  2) after  op= Constan Time elapsed: 0.004225 accumulated: 0.010389 VMem: 156608
-#  3) after  op=    Conv Time elapsed: 0.360213 accumulated: 0.370602 VMem: 156608 (model/conv1)
-#  4) after  op= Softplu Time elapsed: 0.190591 accumulated: 0.561193 VMem: 156608 (model/softplus1)
-#  5) after  op=    Tanh Time elapsed: 0.115314 accumulated: 0.676507 VMem: 156608 (model/tanh1)
-#  6) after  op=     Mul Time elapsed: 0.022779 accumulated: 0.699286 VMem: 156608 (model/mul1)
+#  0) after onnx.Transpose Time elapsed: 0.000766 accumulated: 0.000766 VMem: 156608 (model/transpose1)
```
chentong319, line 40: Reduce `Time elapsed` to `Elapsed` and give more space for op name. Does it make sense to print one digit (from 6 to 7) for the time? I forgot what's the precision of the clock tick.

src/Compiler/CompilerOptions.cpp
```diff
@@ -133,15 +133,14 @@ llvm::cl::opt<OptLevel> OptimizationLevel(llvm::cl::desc("Levels:"),
         clEnumVal(O3, "Optimization level 3.")),
     llvm::cl::init(O0), llvm::cl::cat(OnnxMlirCommonOptions));
 
-llvm::cl::opt<std::string> instrumentStage("instrument-stage",
-    llvm::cl::desc(
-        "Specify stage to be instrumented\n"
-        "\"before-onnx-to-krnl\" : Profile for onnx ops (before lowering to "
-        "krnl)\n"
-        "\"nnpa-before-onnx-to-zhigh\" : [NNPA] Profile for onnx ops\n"
```
chentong319, line 141: Here is place to use #ifdef

docs/Instrumentation.md
imaihal: > Does it make sense to print one digit (from 6 to 7) for the time? 

It seems `gettimeofday()` gives microseconds. So, current 6 digits may be appropriate.

include/onnx-mlir/Compiler/OMCompilerTypes.h
tungld: `--maccel` is perhaps a good example about inserting enum values for NNPA while building the compiler: https://github.com/onnx/onnx-mlir/blob/main/src/Compiler/CompilerOptions.cpp#L98

AlexandreEichenberger:
Question on the naming, please chime in if that make sense. The proposed options:
```
--instrument-stage=<value>                        - Specify stage to be instrumented:
    =afterOnnxToOnnx                                -   Profile for onnx ops.
    =nnpaAfterOnnxToOnnx                            -   [NNPA] Profile for onnx ops.
    =nnpaAfterOnnxToZhigh                           -   [NNPA] Profile for onnx and zhigh ops.
    =nnpaAfterZhighToZlow                           -   [NNPA] Profile for zlow ops.
```

Wouldn't it be simpler to simply have instrument stage at ONNX, ZHigh, Zlow? And with ONNX and `maccel=NNPA` you get the proposed `nnpaAfterOnnxToOnnx`, and without it you get the `afterOnnxToOnnx`.

These more generic names allows us to "move" the instrumentation up and down if we were to add a future optimization that we would include. 

For example, say we add a super useful zhigh to zhigh optimization, would we then need to add a new instrument stage `nnpa=AfterZhighToZhigh`? 

I find it easier to ask the user to state at what level he wants the instrumentation, and we decide what is the best relevant point for this instrumentation to go.

include/onnx-mlir/Compiler/OMCompilerTypes.h
chentong319: Since different accelerator may have different dialect and different points for instrumentation,  we need to include the real information from each accelerator directory that is in the accelerator list. Then put them together into an enum type for instrumentation stage. There may be other values needed to be put together for other purpose.  I do not know how to do it in a simple way. I am OK with that this PR may have some flaw on this option if we do not have a good solution.

chentong319: Can we use weak definition in src/Compiler and then redefine instrumentStage in NNPA/Compiler?

chentong319:
> Wouldn't it be simpler to simply have instrument stage at ONNX, ZHigh, Zlow?
At least this approach has the same problem that the details of a particular accelerator is exposed no matter the accelerator is turned on or not.

include/onnx-mlir/Compiler/OMCompilerTypes.h
imaihal: Thanks for your comments!
`APPLY_TO_ACCELERATORS()` used in `--maccel` looks smart solution. I'm trying to implement using it.

> Can we use weak definition in src/Compiler and then redefine instrumentStage in NNPA/Compiler?

This is what I should do. I will try. Thanks

src/Accelerators/Accelerator.hpp
```diff
@@ -42,6 +42,15 @@
...
 
+#define ACCEL_INSTRUMENTSTAGE_ENUM(name)                                       \
+  , nnpaAfterOnnxToOnnx, nnpaAfterOnnxToZhigh, nnpaAfterZhighToZlow
+
+#define ACCEL_INSTRUMENTSTAGE_CL_ENUM(name)                                    \
+  , clEnumVal(nnpaAfterOnnxToOnnx, #name " Profile for onnx ops."),            \
+      clEnumVal(                                                               \
+          nnpaAfterOnnxToZhigh, #name " Profile for onnx and zhigh ops."),     \
+      clEnumVal(nnpaAfterZhighToZlow, #name " Profile for zlow ops.")
+
```
imaihal, line 53: I moved NNPA related option (stage name) to Accelerator files using `APPLY_TO_ACCELERATORS()`, but I'm not sure how I can move them to NNPA files (NNPACompilerUtils.cpp etc.)...

imaihal: Moved NNPA stage name to `NNPACompilerUtils.hpp` using `INSTRUMENTSTAGE_EUM_NNPA` and `INSTRUMENTSTAGE_CL_ENUM_NNPA` by commit https://github.com/onnx/onnx-mlir/pull/1823/commits/4ff1354556d8b19981e318286b58d4562edc9492

imaihal:
I set ready for review, but we still need discussions to reflect @AlexandreEichenberger and @chentong319 comments.

chentong319:
@imaihal Your approach of using #define seems good to me. You do not need to consider my suggestion if the #define works.

chentong319 (approved):
LGTM!

src/Accelerators/Accelerator.hpp
imaihal: @chentong319 I removed stages from instrumentation pass as you suggested. I think this is better because developers can move the pass up and down easily for their needs.

tungld:
> --instrument-stage=<value>                        - Specify stage to be instrumented:
    =Onnx                                             -   Profile for onnx ops. For NNPA, profile onnx ops before lowering to zhigh.
    =ZHigh                                            -   NNPA profiling for onnx and zhigh ops.
    =ZLow                                             -   NNPA profiling for zlow ops.

It seems the instrument-stage is now using dialect name. If so, is it possible to get information from `-instrument-ops=onnx.,zhigh.,.zlow` instead of using new option. If we can utilize `--instrument-ops`, there is no need to handle `--instrument-stage` for cpu and nnpa.

imaihal:
@chentong319 @AlexandreEichenberger I changed the stage name as @AlexandreEichenberger suggested.

When building with `-DONNX_MLIR_ACCELERATORS=NNPA `, stages for ZHigh and ZLow are available for NNPA.
```
  --instrument-stage=<value>           - Specify stage to be instrumented:
    =Onnx                                               -   Profile for onnx ops. For NNPA, profile onnx ops before lowering to zhigh.
    =ZHigh                                              -   NNPA profiling for onnx and zhigh ops.
    =ZLow                                               -   NNPA profiling for zlow ops.
```
When building without`-DONNX_MLIR_ACCELERATORS=NNPA `, only one stage is available.
```
  --instrument-stage=<value>                           - Specify stage to be instrumented:
    =Onnx                                              -   Profile for onnx ops.
```
 When compiling model with `-maccel` and `--instrument-stage=Onnx`, instrumentation before lowering to zhigh are used.

imaihal:
@chentong319 @AlexandreEichenberger We should have defined spec of instrumentation tool more before implementation, but as @tungld suggested, current stages used in assumed usecases can be reproduced from `--maccel` and `--instrument-ops`.
    
1. `--instrument-ops=onnx.` ==> `--instrument-stage=Onnx`
2. `--maccel=NNPA` and `--instrument-ops=onnx.` ==> `--maccel=NNPA` and `--instrument-stage=Onnx`
3. `--maccel=NNPA` and `--instrument-ops=onnx.,zhigh.` ==> `--maccel=NNPA` and `--instrument-stage=ZHigh`
4. `--maccel=NNPA` and `--instrument-ops=zlow.` ==> `--maccel=NNPA` and `--instrument-stage=ZLow`

Do you think we still keep stage option for flexibility? Or should we reproduce the stages from instrumentation option?
Since I have spent a lot of time on this, I want to finish this PR.

chentong319:
> > --instrument-stage=                        - Specify stage to be instrumented:
> > =Onnx                                             -   Profile for onnx ops. For NNPA, profile onnx ops before lowering to zhigh.
> > =ZHigh                                            -   NNPA profiling for onnx and zhigh ops.
> > =ZLow                                             -   NNPA profiling for zlow ops.
> 
> It seems the instrument-stage is now using dialect name. If so, is it possible to get information from `-instrument-ops=onnx.,zhigh.,zlow.` instead of defining a new option `--instrument-stage`. If we can utilize `--instrument-ops`, there is no need to handle `--instrument-stage` for cpu and nnpa.

In general, it is possible that the same instrument-op for different instrumentation stage. For example, we may have more than one instrument stage for ONNX dialect. Logically, instrument-stage specifies when to instrument while instrument-ops specifies what to be instrumented. They are orthogonal to me.

tungld:
> Logically, instrument-stage specifies when to instrument while instrument-ops specifies what to be instrumented.

Yes, I understand this. My concern is that `--instrument-stage` is rather difficult for users to understand because it highly depends on the internal order of calling passes.

chentong319:
> > Logically, instrument-stage specifies when to instrument while instrument-ops specifies what to be instrumented.
> 
> Yes, I understand this. My concern is that `--instrument-stage` is rather difficult for users to understand because it highly depends on the internal order of calling passes.

Instrumentation tools usually require user to understand some details of the system. Let's finish this PR which has been here for a long time, and leave the possible improvement to future PR.

PR #1826, opened 2022-11-03
Verify user docker image

gongsu832:
Add a Jenkins pipeline step to verify the user docker image after it's built. The step will use the image just built to compile a simple mnist-8 model from the model zoo to check if the image works or not.


tungld (approved):
LGTM.

PR #1828, opened 2022-11-03
Onnx ops reorganization (no functional changes)

AlexandreEichenberger:
This massive PR removed the Dialect/ONNX/ShapeInference subdirectories, and split the 6k lines Dialect/ONNX/ONNXOps.cpp into files that contains both the verifiers, shape inference, and sometime additional methods required for that op. Most files have only one operation (or sometime one op and its older versions for backward compatibility). Some files group multiple related ops for max code reuse, e.g. RNNs, Elementwise ops with broadcast, Reductions, Convolutions, Pooling.

I did not reorganize the ONNXShapeHelper.hpp/cpp at this time. It could also be broken down in individual files. Let me know if you feel that this step is important too.

I left the "unimplemented" ops error handling in the main ONNXOps.cpp to better shame us into implementing them.

Because this PR touch so many files, my suggestions is that we merge it soon to avoid conflicts, and I promise I will take your suggestions at hart in a subsequent PR, unless you have strong objections, of course.


src/Dialect/ONNX/ONNXShapeHelper.inc
```diff
@@ -0,0 +1,54 @@
+/*
```
chentong319, line 1: Reason to separate this file out?

chentong319:
Question about the directory structure. I know there are subdirectories in OnnxToKrnl conversion. I feel that it is easier to locate an Op in one directory unless there is an existing or obvious guideline for us to group the Ops. If we keep the subdirectories, should we put them under something like Dialect/ONNX/ONNXOpsImplement? All these comments are not required changes.

src/Dialect/ONNX/ONNXShapeHelper.inc
AlexandreEichenberger: I don't like to have code in the header files, and by having it into an `inc` file, it looks a bit cleaner. I would like to have this templated code in the header files, as otherwise we need to put 50+ template instantiation, and every new op using that code pattern will have to do so also, which is a cause of pain as the errors for not doing so are cryptic.

AlexandreEichenberger:
>  If we keep the subdirectories, should we put them under something like Dialect/ONNX/ONNXOpsImplement?

I like this suggestions. Will do, it should not have a big impact. Thanks

src/Dialect/ONNX/Math/ElementwiseUnary.cpp
```diff
@@ -0,0 +1,778 @@
...
+  int64_t inputRank = inputType.getRank();
+  int64_t axisValue = axis();
+
+  // axis attribute must be in the range [-r,r], where r = rank(input).
+  if (axisValue < -inputRank || axisValue > inputRank)
+    return onnx_mlir::Diagnostic::emitAttributeOutOfRangeError(
+        *this->getOperation(), "axis", axisValue,
+        onnx_mlir::Diagnostic::Range<int64_t>(-inputRank, inputRank - 1));
+
+  return inferShapeForUnaryElementwiseOps(this->getOperation());
```
sorenlassen, line 294: I can see how it makes sense for Hardmax to share shape inference with conventional unary ops, but it's misleading that it is in a source file called "elementwise".

Should we rename the source fil to just "Unary.cpp" or break ops like Hardmax (and Scatter etc) into other files?

src/Dialect/ONNX/Math/ElementwiseUnary.cpp
```diff
@@ -0,0 +1,778 @@
...
+
+  auto inputType = input().getType().cast<ShapedType>();
+  int64_t inputRank = inputType.getRank();
+  int64_t axisValue = axis();
+
+  // axis attribute must be in the range [-r,r], where r = rank(input).
+  if (axisValue < -inputRank || axisValue > inputRank)
+    return onnx_mlir::Diagnostic::emitAttributeOutOfRangeError(
+        *this->getOperation(), "axis", axisValue,
+        onnx_mlir::Diagnostic::Range<int64_t>(-inputRank, inputRank - 1));
```
sorenlassen, line 292: random observation, not related to this PR: seems like we should remove this and just rely on verify() to police rules that are unrelated to shape inference

sorenlassen (approved):
I like this change and I suggest we go ahead and land it as is and then iterate on further improvements.

E.g. I'd like to add extra helper functions to relevant ops and a header file, including the ShapeHelper class definition for each op. Then we can put the declaration of getDataShapeBounds and computeSelectedData in Shape.hpp and we can move EinsumOpHelper.{hpp,cpp} into EinsumOp.{hpp,cpp}.

src/Dialect/ONNX/ONNXOpsHelper.hpp
```diff
@@ -253,8 +271,10 @@ mlir::Type convertONNXTypeToMLIRType(
...
-template <class T>
-mlir::LogicalResult processConvTypeParams(
-    T *op, mlir::Value inputOperand, mlir::Value W);
-
 } // namespace onnx_mlir
+
+#define NOT_IMPLEMENTED_MESSAGE                                                \
+  (getOperationName() +                                                        \
+      ": is not supported at this time. Please open an issue on "              \
+      "https://github.com/onnx/onnx-mlir and/or consider contribute code. "    \
```
sorenlassen, line 279: "contributing"

src/Dialect/ONNX/ONNXOpsHelper.hpp
```diff
@@ -17,14 +17,32 @@
 #include "mlir/Dialect/Traits.h"
 #include "mlir/IR/AffineExpr.h"
 #include "mlir/IR/AffineMap.h"
+#include "mlir/IR/Block.h"
```
sorenlassen, line 20: I would have preferred to have the fewest possible includes in commonly used header files like this one, so .cpp files don't pull in excess indirect includes. Is it an option to make only the .cpp files that need them include Block.h, SetVector.h, etc?

AlexandreEichenberger:
> EinsumOpHelper.{hpp,cpp} into EinsumOp.{hpp,cpp}.

Usually I added a header file when it was shared between two unrelated ops.

> E.g. I'd like to add extra helper functions to relevant ops and a header file, including the ShapeHelper class definition for each op. 

Agreed, that would go under the "break the ShapeHelper" files. But in many ways there is a lot of reuse, so it may end up being header files for many many ops. Reuse is good.

>  I suggest we go ahead and land it as is and then iterate on further improvements.

yes, subsequent changes will be less prone to conflicts, except maybe for "breaking ShapeHelper"

src/Dialect/ONNX/Math/ElementwiseUnary.cpp
AlexandreEichenberger: Let me split them from the elementary

src/Dialect/ONNX/Math/ElementwiseUnary.cpp
AlexandreEichenberger: I fully agree.. It is often convenient when writing code to put the assert immediately where written, MLIR verifiers are very powerful and indeed it would be the right place. Many ops don't have verifiers... we should do a push to add them. Another PR though.

src/Dialect/ONNX/ONNXOpsHelper.hpp
AlexandreEichenberger: done

src/Dialect/ONNX/ONNXOpsHelper.hpp
AlexandreEichenberger: Agreed, this can be done in a later PR as it will be small localized changes.

chentong319 (approved):
LGTM!

tungld (approved):
LGTM.

Issue #1655, opened 2022-08-30
AlexNet, ResNet... fail to compile in the onnxmlirczar/onnx-mlir:s390x container


cfig-xk:
When compiling AlexNet.onnx, ResNet50.onnx models using `onnx-mlir --EmitLib AlexNet.onnx` in the onnxmlirczar/onnx-mlir:s390x container, I got the following error:
`error: reduction dimension must be the same
error: Failed to scan onnx.MatMul parameters successfully
error: shape inference failed`

Notes:

- I saved the AlexNet in 'tf' format based on python3.6 tensorflow2.4.0 Ubuntu 20.04.4 LTS Linux version 5.15.0-41-generic s390x,  here is the code:
`import tensorflow as tf
import pathlib
import random

from keras.layers import Flatten,BatchNormalization

from keras.layers.convolutional import Conv2D
from keras.layers.merge import add
from keras.models import Sequential

from keras.layers import Input, merge, ZeroPadding2D
from keras.layers import Dense, Activation
from keras.layers.core import Dropout
from keras.layers.convolutional import Convolution2D
from keras.layers.pooling import AveragePooling2D, GlobalAveragePooling2D, MaxPooling2D
from keras.layers.normalization import BatchNormalization

import six
from keras.regularizers import l2
from keras.models import Model

import keras.backend as K
K.set_image_data_format('channels_last')

try:
    from keras import initializations
except ImportError:
    from keras import initializers as initializations

os.environ["CUDA_VISIBLE_DEVICES"]="-1"

class Config(object):
    normal_size = 227
    channles = 3
    classNumber = 1000

class MODEL(object):

    def __init__(self,config):
        self.config = config

    # AlexNet
    def AlexNet(self):
        model = Sequential()
        # input_shape = (64,64, self.config.channles)
        input_shape = (self.config.normal_size, self.config.normal_size, self.config.channles)
        model.add(Convolution2D(96, (11, 11), input_shape=input_shape,strides=(4, 4),  padding='valid',activation='relu', kernel_initializer='random_normal'))
        model.add(MaxPooling2D(pool_size=(3, 3), strides=(2, 2)))#26*26
        model.add(Convolution2D(256, (5, 5), strides=(1, 1), padding='same', activation='relu', kernel_initializer='random_normal'))
        model.add(MaxPooling2D(pool_size=(3, 3), strides=(2, 2)))
[...]

PR #1829, opened 2022-11-03
Fix MHLO installed targets

yaochengji:
From [issue](https://github.com/onnx/onnx-mlir/issues/1655)

- add all the targets of high level dialect in mlir-hlo repo

PS: On my side I cannot reproduce `error while loading shared libraries: libChloOps.so.16git: cannot open shared object file` before this patch.


CMakeLists.txt
```diff
@@ -150,7 +150,7 @@ add_subdirectory(third_party/pybind11)
 add_subdirectory(third_party/rapidcheck)
 
 if (ONNX_MLIR_ENABLE_MHLO)
-  add_subdirectory(third_party/mlir-hlo EXCLUDE_FROM_ALL)
```
sstamenova, line 153: Why is it necessary to remove this? My expectation is that this should be fine since we don't want to build it as part of the `all` target and only when needed by other targets.

src/Conversion/ONNXToMhlo/CMakeLists.txt
```diff
@@ -2,11 +2,19 @@
...
-  )
+  MLIRMhloUtils
+  MhloDialect
+  MhloRegisterDialects
+  StablehloBase
+  StablehloOps
+  StablehloRegister
+  StablehloTypeInference
+
+  DESTINATION ${CMAKE_INSTALL_LIBDIR}
```
sstamenova, line 16: Are these intended to be installed where onnx-mlir is installed? If so, you should follow the onnx-mlir pattern for choosing installation paths.

src/Conversion/ONNXToMhlo/CMakeLists.txt
```diff
@@ -45,10 +53,10 @@ add_onnx_mlir_library(OMONNXToMhlo
   OMONNXRewrite
   OMONNXOps
   OMSupport
-  )
```
sstamenova, line 48: Why? We generally follow the pattern of indenting the closing parenthesis, so the change should be the other way around - indent it everywhere.

sstamenova (changes requested):

src/Conversion/ONNXToMhlo/CMakeLists.txt
yaochengji: This is what my auto-formatter do, fixed.

src/Conversion/ONNXToMhlo/CMakeLists.txt
yaochengji: Yes, changed to `lib` instead of `${CMAKE_INSTALL_LIBDIR}`

gongsu832:
> PS: On my side I cannot reproduce `error while loading shared libraries: libChloOps.so.16git: cannot open shared object file` before this patch.

Sorry I'm a bit confused by this sentence. You mean you can use `onnxmlirczar/onnx-mlir` docker image to compile a model successfully? That shouldn't be possible. What did you actually try? Simply running the image should produce the error:
```
# docker run --rm -ti onnxmlirczar/onnx-mlir
onnx-mlir: error while loading shared libraries: libChloOps.so.16git: cannot open shared object file: No such file or directory
```

yaochengji:
> You mean you can use `onnxmlirczar/onnx-mlir` docker image to compile a model successfully?

I used the docker file of onnx-mlir to build my own docker image, but I think it's similar to `onnxmlirczar/onnx-mlir`.
My error is the same as the [issue](https://github.com/onnx/onnx-mlir/issues/1655) submmiter's , not `onnx-mlir: error while loading shared libraries: libChloOps.so.16git: cannot open shared object file: No such file or directory`.

```
onnx-mlir --EmitLib AlexNet.onnx 
error: reduction dimension must be the same
error: Failed to scan onnx.MatMul parameters successfully
error: shape inference failed
```

gongsu832:
> > You mean you can use `onnxmlirczar/onnx-mlir` docker image to compile a model successfully?
> 
> I used the docker file of onnx-mlir to build my own docker image, but I think it's similar to `onnxmlirczar/onnx-mlir`. My error is the same as the [issue](https://github.com/onnx/onnx-mlir/issues/1655) submmiter's , not `onnx-mlir: error while loading shared libraries: libChloOps.so.16git: cannot open shared object file: No such file or directory`.
> 
> ```
> onnx-mlir --EmitLib AlexNet.onnx 
> error: reduction dimension must be the same
> error: Failed to scan onnx.MatMul parameters successfully
> error: shape inference failed
> ```

Then I'm guessing your docker image either managed to install the missing libChloOps.so.16git (and potentially others) or you kept the source tree in your image.

yaochengji:
> Then I'm guessing your docker image either managed to install the missing libChloOps.so.16git (and potentially others) or you kept the source tree in your image.

I tried with a clean docker image, but still could not reproduce it. And I tried to test it directly by building a docker image with `docker/Dockerfile.onnx-mlir` but I don't know how to specify the docker file's parameter.

@gongsu832 could you help test it on your side?

gongsu832:
> > Then I'm guessing your docker image either managed to install the missing libChloOps.so.16git (and potentially others) or you kept the source tree in your image.
> 
> I tried with a clean docker image, but still could not reproduce it. And I tried to test it directly by building a docker image with `docker/Dockerfile.onnx-mlir` but I don't know how to specify the docker file's parameter.
> 
> @gongsu832 could you help test it on your side?

Why don't you just `docker pull onnxmlirczar/onnx-mlir`?

yaochengji:
> Why don't you just `docker pull onnxmlirczar/onnx-mlir`?

Yes, this can reproduce. But the problem is that I cannot start the docker container. (I need to build onnx-mlir in the docker container)

gongsu832:
> > Why don't you just `docker pull onnxmlirczar/onnx-mlir`?
> 
> Yes, this can reproduce. But the problem is that I cannot start the docker container. (I need to build onnx-mlir in the docker container)

Another thing to check, in your own docker image, did you build llvm-project with `-DBUILD_SHARED_LIBS=ON`? By default, llvm-project builds everything statically and then onnx-mlir will also build everything statically, then you won't see the problem since all the libraries will have been embedded into the onnx-mlir binary. You can check the size of your onnx-mlir binary. If it's a few hundred MBs, then you are building statically.

I'm thinking of adding a build option for the user image so that the build tree can be kept for debugging purpose. But before that, you can do the equivalent by:

- build a shared llvm-project image:
   `docker build --build-arg BUILD_SHARED_LIBS=ON -t llvm-project-shared -f docker/Dockerfile.llvm-project`
- build a shared onnx-mlir image:
   `docker build --build-arg BASE_IMAGE=llvm-project-shared -t onnx-mlir-shared -f docker/Dockerfile.onnx-mlir`
   (but remember to remove the line in the dockerfile `&& rm -rf /tmp/* /usr/bin/python llvm-project onnx-mlir`)

gongsu832:
> ~: docker run onnxmlirczar/onnx-mlir bash
> onnx-mlir: error while loading shared libraries: libChloOps.so.16git: cannot open shared object file: No such f

The correct command is
```
docker run --rm -ti --entrypoint=bash onnxmlirczar/onnx-mlir
```
But this won't help too much since all the build sources have been removed so you need to rebuild anyway.

yaochengji:
> But this won't help too much since all the build sources have been removed so you need to rebuild anyway.

I started a new dokcer container by `docker run --rm -ti --entrypoint=bash onnxmlirczar/onnx-mlir-dev`. And basically I followed this [doc](https://github.com/onnx/onnx-mlir/blob/main/docs/BuildOnLinuxOSX.md) to build llvm-project and onnx-mlir ( use`cmake --build . --target install` instead and `rm -rf onnx-mlir/build/*` after building). But still I cannot reproduce the `libChloOps.so` error.

gongsu832:
> > But this won't help too much since all the build sources have been removed so you need to rebuild anyway.
> 
> I started a new dokcer container by `docker run --rm -ti --entrypoint=bash onnxmlirczar/onnx-mlir-dev`. And basically I followed this [doc](https://github.com/onnx/onnx-mlir/blob/main/docs/BuildOnLinuxOSX.md) to build llvm-project and onnx-mlir ( use`cmake --build . --target install` instead and `rm -rf onnx-mlir/build/*` after building). But still I cannot reproduce the `libChloOps.so` error.

As I mentioned above, you need to build everything **shared** to produce the problem. `onnxmlirczar/onnx-mlir-dev` builds everything static, and the doc you followed also builds everything static.

BTW, for `onnxmlirczar/onnx-mlir-dev` image, you don't need to specify `--entrypoint=bash`. But as I said, it builds everything in static. So you will have to rebuild both llvm-project and onnx-mlir in shared anyway, just like what I told you above if you are going to build your own shared image.

src/Conversion/ONNXToMhlo/CMakeLists.txt
```diff
@@ -52,3 +64,7 @@ add_onnx_mlir_library(OMONNXToMhlo
 target_link_libraries(MhloDialect PUBLIC
   MLIRSparseTensorDialect
   )
+
+target_link_libraries(StablehloTypeInference PUBLIC
```
sstamenova, line 68: Why is this needed? The two libs are both linked into the target above

yaochengji:
@gongsu832 Thanks for the reminding of the static/shared diffrence, I rebuilt llvm-project and onnx-mlir in a shared way. Now I can reproduce the `libChloOps.so` error and already tested locally.

src/Conversion/ONNXToMhlo/CMakeLists.txt
yaochengji: But `StablehloTypeInference` links into `StablehloBase`, which is missing in `mlir-hlo` repo.

gongsu832:

src/Conversion/ONNXToMhlo/CMakeLists.txt
sstamenova: Have you let the mhlo folks know, so that it can be fixed there as well?

sstamenova (approved):
LGTM, but please make sure that the mhlo dependency issues gets fixed in mhlo as well - either by letting them know or sending a PR.

src/Conversion/ONNXToMhlo/CMakeLists.txt
yaochengji: It is added. https://github.com/openxla/stablehlo/pull/456.

BTW, I find that the additional linking relationship below is already added in mlir-hlo repo.
```
target_link_libraries(MhloDialect PUBLIC
  MLIRSparseTensorDialect
  )
```

yaochengji:
@gongsu832  @sstamenova , can this PR be merged now? I will delete the additional LINK_LIBS later.

PR #1832, opened 2022-11-04
Get layout from Operation when lowering ZHigh to ZLow

tungld:
A data layout of a zTensor can be obtained directly from an operation's operand whose type is TensorType, so we don't need to store a mapping between a MemRefType and a data layout during the zhigh-to-zlow lowering anymore.


AlexandreEichenberger (approved):
LGTM

PR #1833, opened 2022-11-04
IndexExpr propagation (neutral element handling)

AlexandreEichenberger:
Better handling of the neutral element for Index expression. For example:
`a+0` return `a`, which is important when we have "question mark", that is runtime unknown. It is important that both `a+0` and `a` have the same "symbolic unknown marker" hidden into the question mark representation of the index expression.

Same was done for sub/mul/div. Select true/false already did the right thing.

A slight consequence is that some of the maps are now created in a different order, thus the minor changes in the lit tests. To alleviate this, I moved some offending tests in the `canonicalized` version of the lit tests. Also, some "krnl.global" use monotonically increasing "ids" which is a problem in FileCheck. The `mlir2FileCheck.py` now reg-exp the id in the FileCheck outputs.


src/Dialect/Mlir/IndexExpr.cpp
```diff
@@ -736,14 +743,13 @@ IndexExpr IndexExpr::operator%(IndexExpr const b) const {
...
   F2 valueFct = [](IndexExpr const aa, IndexExpr const bb) -> IndexExpr {
-    if (bb.isLiteral() && bb.getLiteral() == 1) {
-      return aa.deepCopy();
-    }
     return NonAffineIndexExpr(aa.getRewriter().create<arith::RemSIOp>(
         aa.getLoc(), aa.getValue(), bb.getValue()));
   };
   // Index b must be a literal.
-  return binaryOp(b, true, true, litFct, affineExprFct, valueFct);
+  // Neutral value: ignore here that x % x = x.
```
tungld, line 750: `x % x = 0` isn't it?

tungld (approved):
LGTM with pending the comment about the `a mod b` case.

Thank you for adding this handling!

PR #1834, opened 2022-11-07
Update supported opset to 16

philass:

### DIFF
--- a/src/Dialect/ONNX/ONNXOps.hpp
+++ b/src/Dialect/ONNX/ONNXOps.hpp
@@ -25,7 +25,7 @@
 
 namespace mlir {
 // OpSet level supported by onnx-mlir
-static constexpr int CURRENT_ONNX_OPSET = 13;
+static constexpr int CURRENT_ONNX_OPSET = 16;
 } // end namespace mlir
 
 #define GET_OP_CLASSES


src/Dialect/ONNX/ONNXOps.hpp
```diff
@@ -32,7 +32,7 @@
 namespace mlir {
 
 // OpSet level supported by onnx-mlir
-static constexpr int CURRENT_ONNX_OPSET = 13;
+static constexpr int CURRENT_ONNX_OPSET = 16;
```
AlexandreEichenberger, line 35: @hamptonm1 are we not at 17 now?
Could we link by comments all the places where this version opset is updated in the project?

For example in `gen_onnx_mlir.py`
```
# Change this variable only when upgrading the ONNX support within ONNX-MLIR.
current_onnx_version = "1.12.0"
```

Let's write in both places (and additional if there are other vars associated with the level) comments indicating to keep them in sync.

hamptonm1: Yeah since we are on onnx 1.12.0 that corresponds to opset 17 but not all of the opsets have been updated. Some remain on 16 or later according to the onnx documentation. Seems like `onnx` introduced a lot of newer opsets as well with version 17.

AlexandreEichenberger (approved):

PR #1836, opened 2022-11-07
Constant propagation for ONNXReshapeOp

tungld:
ReshapeOp does not change the underlying data, so its constant propagation is simple forwarding the input to the output with a new shape.


AlexandreEichenberger (approved):
LGTM, thanks for thinking of it

sorenlassen (approved):
nice

PR #1837, opened 2022-11-07
Use 700 when loading .so for Java

cjvolzka:
AppScan on Cloud doesn't like other/world permissions set on files.

When running models on Java we extract the .so file from the .jar so we can load it. The file is exacted to a temp directory, loaded and then deleted. For z/OS cases, we have to set the permission to include the execute. However since the file is loaded and immediately deleted, we should be able to restrict the permissions to only the current user to make ASoC happy.


gongsu832 (approved):
LGTM

PR #1842, opened 2022-11-08
[NNPA] Make ONNXOps work with zTensor

tungld:
Summary: this patch makes ONNXOps work with zTensor, e.g., ONNXAddOp can work directly on zTensor so that we don't need to unstickify tensors before passing them to ONNXAddOp.


AlexandreEichenberger (approved):
LGTM, looking forward to see how we can optimize the transpose and such on stickified data.

PR #1845, opened 2022-11-09
[NNPA] Rewrite N-D to 3-D softmax instead of 2-D to avoid out-of-range on a dimension size

tungld:
We did rewrite N-D to 2-D softmax to make it run on NNPA, however it makes the compressed dimension big and sometimes larger than the supported value by NNPA.

NNPA Softmax also supports 3-D in addition to 2-D, so it is better to rewrite to 3-D instead. 


AlexandreEichenberger (approved):
LGTM

PR #1847, opened 2022-11-09
Renaming of onnx-mlir libraries

AlexandreEichenberger:
Some time ago, we renamed our libraries to start with OM prefix, but many of the newly added libs do not follow this pattern. I changed all the ones we use regularly. This PR renames them.

Also, the name of the compile object in the python interface used "Py" as prefix, but really should use "OM", so that was changed too.

There are many libs that are still not using the OM names, notably the NNPA and the HLO ones. I suggest that we adopt the same standard as MLIR (which prefix theirs with MLIR) and also change the prefixes of these.

The ones that are touchy are the "runtime" libs as their names may be already used explicitly, so these we should consider a bit more carefully.

No functionality was changed.


AlexandreEichenberger:
@tungld could I also rename the NNPA prefixed libs, or do you prefer have the accel as prefix only.

tungld:
> @tungld could I also rename the NNPA prefixed libs, or do you prefer have the accel as prefix only.

Yes, please do the same for NNPA. I am thinking of using OMNNPA prefix, but please feel free to make your own decision.

AlexandreEichenberger:
@tungld Added OMNNPA prefix as requested

src/Runtime/CMakeLists.txt
```diff
@@ -112,7 +112,7 @@ target_include_directories(PyRuntime
   )
 target_link_libraries(PyRuntime
   PRIVATE
-  ExecutionSession
+  OMExecutionSession
```
tungld, line 115: I know another place using ExecutionSession that is https://github.com/onnx/onnx-mlir/blob/main/utils/RunONNXModel.py#L121. RunONNXModel.py is used by Jenkins to check the onnx model zoo.

tungld (approved):
LGTM! 

Please make sure to make a change in `utils/RunONNXModel.py` before merging. Otherwise, Jenkins check for the onnx model zoo would fail.

src/Runtime/CMakeLists.txt
AlexandreEichenberger: @tungld 

I have not renamed the module within the python interface because others may use it... I have tested this example

``` python
import numpy as np

# Load the model mnist.so compiled with onnx-mlir.
model = './mnist.so'

# Load the model mnist.so compiled with onnx-mlir.
model = './mnist.so'
session = ExecutionSession(model)
```

which is from mnist_runPyRuntime.py

Is that what the issue is?

Now if you agree that we should also change the module name inside python, to match the compiler interface...

Here is what I have (and like) for compiler module
``` python
import numpy as np
from PyCompileAndRuntime import OMCompileExecutionSession

# Load onnx model and create CompileExecutionSession object.
inputFileName = './mnist.onnx'
# Set the full name of compiled model
sharedLibPath = './mnist.so'
# Set the compile option as "-O3"
session = OMCompileExecutionSession(inputFileName,sharedLibPath,"-O3")

```

AlexandreEichenberger: @tungld let me know, will not merge before I get your input on this

tungld: OK, so `ExecutionSession ` here is not the one in `from PyRuntime import ExecutionSession` . Then  it's OK to merge now. Thanks!

tungld:
```
Here is what I have (and like) for compiler module

import numpy as np
from PyCompileAndRuntime import OMCompileExecutionSession

# Load onnx model and create CompileExecutionSession object.
inputFileName = './mnist.onnx'
# Set the full name of compiled model
sharedLibPath = './mnist.so'
# Set the compile option as "-O3"
session = OMCompileExecutionSession(inputFileName,sharedLibPath,"-O3")
```

I see we can change  the python module from `ExecutionSession` to `OMExecutionSession` to make it consistent with `OMCompileExecutionSession`. Totally agree.

AlexandreEichenberger: Renamed ExecutionSession to OMExecutionSession in the python object. Believe that I made the changes everywhere.

tungld: Thanks so much! LGTM!

tungld:
@AlexandreEichenberger could you resolve conflicts and try to merge this as soon as possible to avoid other conflicts? thanks!

PR #1854, opened 2022-11-14
Avoid cyclic rewriting in simplify-shape-related-ops-onnx pass

tungld:
There is a bug in `simplify-shape-related-ops-onnx` pass that causes cyclic rewriting in which the rewriting pattern for Reshape and ConstantOfShape returns `success` no matter the output shape is updated or not.

This patch is to fix that bug by only return `success` if the output shape is updated.


AlexandreEichenberger (approved):
LGTM

Issue #1856, opened 2022-11-15
Memory leak when running Python models


cjvolzka:
# Steps to reproduce:

1. Compiled [mnist-12 model](https://github.com/onnx/models/blob/main/vision/classification/mnist/model/mnist-12.onnx) using: `--O3 --EmitLib --mtriple=s390x-ibm-loz --mcpu=z14 mnist-12.onnx`
2. Run model using the supplied [leak_client.py.zip](https://github.com/onnx/onnx-mlir/files/10009548/leak_client.py.zip) (`./leak_client.py <path_to_compiled_mnist-12.so)
    * The client prints it's PID when starting
    * The client runs the model in an infinite loop
    *  The client doesn't save the output and uses the same input for each inference
    * The client is hardcoded for mnist-12. `input_shape` must be changed to try other models
4. Use `top` or other tool to monitor memory usage of the PID printed when the client starts 

# Notes:
* I was running using Python 3.9.13
* The leak seems worse for faster models. (ie the more iterations that can be run, the faster the leak grows)
* At one minute the model managed to take 2GB of memory in 340000 iterations. The first iteration starts around 100 mb and grows from there over time. 
* The model.so itself doesn't show a leak when checked with `valgrind` using a C based client

AlexandreEichenberger:
@cjvolzka can you check that it is working for you, and close if appropriate, tx

cjvolzka:
Thanks for the reminder. I tested this on Friday and the fix looks good. Thanks for tracking down the Python issue @tungld!

PR #1857, opened 2022-11-15
Fix memory leak in pyRun

tungld:
This is to fix the memory leak reported in #1856.

We should clean up the input and output tensor lists before `return`, and the return policy will be handled automatically by pybind as follows: https://pybind11-jagerman.readthedocs.io/en/stable/advanced.html#return-value-policies


AlexandreEichenberger (approved):
LGTM

cjvolzka:
I gave this a try and the fix looks good. The client running the mnist-12 model stayed at approximately 39.2 Mb memory usage over all iterations.

PR #1859, opened 2022-11-15
Streamlining Shape Inference (first step)

AlexandreEichenberger:
Shape inference right now has multiple interfaces to generate the shape helpers. One of the first issue was that shape helpers have a different constructor for analysis and code gen phase:
```C++
struct ONNXOpShapeHelper {
  // Constructor for shape inference. Reuse scope if given, otherwise create one
  // now and free it in destructor.
  ONNXOpShapeHelper(OP *newOp, int numResults, IndexExprScope *inScope);
  // Constructor when code can be generated. Reuse scope if given, otherwise
  // create one now and free it in destructor.
  ONNXOpShapeHelper(OP *newOp, int numResults, mlir::OpBuilder *rewriter,
      ArrayValueIndexCapture::GetDenseVal fGetDenseVal,
      ArrayValueIndexCapture::LoadVal fLoadVal, IndexExprScope *inScope);
```

where the second constructor has to further pass low level functions to extract constants and load values (needed as some phase uses Krnl ops, but we don't want to introduce dependences to KRNL in ONNX or Mlir dialects).

So the first step is to clean up this. I am reusing the "builder" infrastructure to generate `IndexExpression` with an interface that works both in Analysis (question mark represents runtime values) and Code Gen/Code lowering (code is generated to produce the needed values).

```C++
struct NewONNXOpShapeHelper {
  // Constructor for shape inference.
  NewONNXOpShapeHelper(OP *op, mlir::ValueRange operands,
      IndexExprBuilder *ieBuilder, IndexExprScope *scope);
```
To that effect: `IndexExprBuilder` is introduced (generic class with the algorithms to process the constant/symbol/dims) and currently two lower classes:
* `IndexExprBuilderForAnalysis` for analysis, relying only on ONNX dialect
* `IndexExprBuilderForKrnl` for lowering to Krnl.

Other dialects (e.g. MHLO...) can simply implement 3 very simple virtual functions and be fully compatible with the scheme.

To demonstrate, I started to reformat the scheme for Elementwise Unary shape inference.

```C++
LogicalResult NewONNXGenericOpUnaryShapeHelper::computeShape() {
  // Output and input have the same shape. Just pass the input shape to the
  // output.
  uint64_t rank = createIE->getShapeRank(operands[0]);
  DimsExpr outputDims;
  for (uint64_t i = 0; i < rank; ++i)
    outputDims.emplace_back(createIE->getShapeAsDim(operands[0], i));
  setOutputDims(outputDims);
  return success();
}
```

And for example the lowering using shape inference
```C++
    MultiDialectBuilder<IndexExprBuilderForKrnl, KrnlBuilder> create(rewriter, loc);
    NewONNXGenericOpUnaryShapeHelper shapeHelper(
        op, operands, (IndexExprBuilder *)&create.krnlIE);
    auto shapeComputed = shapeHelper.computeShape();
    ...
    ValueRange loopDef = create.krnl.defineLoops(memRefType.getRank());
    SmallVector<IndexExpr, 4> lbs(memRefType.getRank(), LiteralIndexExpr(0));
    SmallVector<IndexExpr, 4> ubs;
    create.krnlIE.getShapeAsDims(X, ubs);
    create.krnl.iterateIE(loopDef, loopDef, lbs, ubs,
          [&](KrnlBuilder &createKrnl, ValueRange loopInd) { ... }

```
No more capture of the bounds, all is integrated within the `create.krnlIE` builder to scan shapes of operands.

At this stage, I would like to gather feedback before I retool many of the shape inference methods.

The step after this will be to integrate shape helper within the ONNX class as a trait (just like shape inference) so that it may be deployed using a common interface. This will be in a subsequent PR as it is relatively orthogonal to this.


tungld (approved):
LGTM. 

I like the idea of using IndexExprBuilder to unify multiple uses of ShapeHelper.

gongsu832:
@AlexandreEichenberger This PR breaks MacOS build.

AlexandreEichenberger:
> @AlexandreEichenberger This PR breaks MacOS build.

@gongsu832 #1863 implements the changes that @chentong319 recommended for fixing the mac issue. Thanks to both of you, sorry for the problems.

PR #1860, opened 2022-11-15
Shape Inference for DFT

hamptonm1:
Add the Shape Inference for the new DFT Opset associated with version 17.


hamptonm1:
FYI - @cjvolzka

hamptonm1:
@AlexandreEichenberger I know you are making updates in this department but I went with the old way for now and we can always update it later to utilize the shape helpers you have in place.

docs/SupportedONNXOps-cpu.md
```diff
@@ -49,7 +49,7 @@ Onnx-mlir currently supports ONNX operations targeting up to opset 16. Limitatio
 | **Cos** |7 | | |
 | **Cosh** |9 | | |
 | **CumSum** |14 | | |
-| **DFT** | |unsupported | |
+| **DFT** | |17 | |
```
AlexandreEichenberger, line 52: would hold off on that. This file is for CPU support, so we need the lowering to krnl before stating it is supported.

AlexandreEichenberger (approved):
LGTM, you will need to learn the new interface, but if you squeeze this PR before mine, I will do it for this one.

docs/SupportedONNXOps-cpu.md
hamptonm1: OHH okay let me change that back to unsupported then.

hamptonm1: Fixed!

hamptonm1:
> LGTM, you will need to learn the new interface, but if you squeeze this PR before mine, I will do it for this one.

Yes I would like to learn, so that I can do this in the near future!

src/Dialect/ONNX/ONNXOps/Math/DFT.cpp
```diff
@@ -0,0 +1,81 @@
...
+    if (isOneSided) {
+      outputDims.emplace_back(dataBounds.getDim(i));
+    } else {
+      if (axis + 1 == i) {
+        outputDims.emplace_back(dataBounds.getDim(i).floorDiv(2));
+      } else {
+        outputDims.emplace_back(dataBounds.getDim(i));
+      }
+    }
+    outputDims.emplace_back(LiteralIndexExpr(2));
```
AlexandreEichenberger, line 58: @hamptonm1 I believe the "2" dim should be added only once after the end of the loop. Please confirm.

Will correct in the PR where I merge your changes into mine.

hamptonm1: Yeah....testing it now

PR #1861, opened 2022-11-16
Fuse Concat Ops

chentong319:
This PR performs special Op fusion: Concat+(Shape, Transpose) into ConcatShapeTranspose at the decompose-onnx step. This may not be perfect place for the fusion. But it works for the target model and I just do not want to write another pass just for this. The PR is a followup of PR#1817.

### DIFF
--- a/src/Transform/ONNX/Decompose.cpp
+++ b/src/Transform/ONNX/Decompose.cpp
@@ -212,7 +212,7 @@ static bool isConcatFuseMatched(
     Operation *op, ONNXShapeOp &shapeOp, ONNXTransposeOp &transposeOp) {
   shapeOp = NULL;
   transposeOp = NULL;
-  bool failed;
+  bool failed = false;
   for (Operation *user : op->getUsers()) {
     if (isa<ONNXShapeOp>(user) && !shapeOp)
       shapeOp = cast<ONNXShapeOp>(user);
@@ -330,7 +330,7 @@ void DecomposeONNXToONNXPass::runOnOperation() {
   }
 #endif
 
-  if (failed(applyFullConversion(function, target, std::move(patterns))))
+  if (failed(applyPartialConversion(function, target, std::move(patterns))))
     signalPassFailure();
 }
 
--- a/src/Transform/ONNX/Decompose.cpp
+++ b/src/Transform/ONNX/Decompose.cpp
@@ -199,6 +199,45 @@ void populateDecomposingONNXBeforeMhloPatterns(
 }
 #endif
 
+struct ConcatFusePattern : public ConversionPattern {
+  ConcatFusePattern(MLIRContext *context)
+      : ConversionPattern(ONNXConcatOp::getOperationName(), 4, context) {}
+  LogicalResult matchAndRewrite(Operation *op, ArrayRef<Value> operands,
+      ConversionPatternRewriter &rewriter) const final {
+
+    ONNXConcatOp concatOp = ::llvm::dyn_cast<ONNXConcatOp>(op);
+    ONNXConcatOpAdaptor operatorAdaptor(operands, op->getAttrDictionary());
+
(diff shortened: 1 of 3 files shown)


src/Transform/ONNX/Decompose.cpp
```diff
@@ -199,6 +199,52 @@ void populateDecomposingONNXBeforeMhloPatterns(
...
+    for (Operation *user : op->getUsers()) {
+      if (isa<ONNXShapeOp>(user) && !shapeOp)
+        shapeOp = cast<ONNXShapeOp>(user);
+      else if (isa<ONNXTransposeOp>(user) && !transposeOp)
+        transposeOp = cast<ONNXTransposeOp>(user);
+      else
+        failed = true;
+    }
+    if (!(shapeOp && transposeOp && !failed))
+      return success();
```
tungld, line 231: Just curious, should this be `return failure()` to say the pattern does not match.

src/Transform/ONNX/Decompose.cpp
```diff
@@ -270,7 +331,7 @@ void DecomposeONNXToONNXPass::runOnOperation() {
   }
 #endif
 
-  if (failed(applyPartialConversion(function, target, std::move(patterns))))
+  if (failed(applyFullConversion(function, target, std::move(patterns))))
```
tungld, line 334: Why it is needed? Just for my learning.

tungld (approved):
LGTM. Thanks!

src/Transform/ONNX/Decompose.cpp
chentong319: Yes, it should be failure.

src/Transform/ONNX/Decompose.cpp
chentong319: Accidental change. I am also learning and testing any difference. I think that we should use Op transform instead of Conversion for this pass because it is transform within a dialect. We do not need typeconverter and target. Example llvm-project/mlir/lib/Dialect/SCF/Transforms/ForToWhile.cpp

src/Transform/ONNX/Decompose.cpp
chentong319: Actually I am not sure whether the failure means failure of match or rewrite. Maybe for match. If match and write is defined separately, the rewrite does not have return value.  If there are more than one rule for one operation, the match failure for each rule should not cause the failure the applyxxxxConversion.

chentong319:
@tungld After I changed the code, the lit test failed on linux but passed Mac with applyPartialConversion. Now I am testing applyFullConversion. If it failed, could you check my changes. I can't find the reason.
With debugging tool, I found that the matchAndRewrite will not be invoked on DynamicallyLegalOp.

tungld:
>  If it failed, could you check my changes. 

Yes, will take it over.

src/Transform/ONNX/Decompose.cpp
```diff
@@ -199,6 +199,60 @@ void populateDecomposingONNXBeforeMhloPatterns(
...
+// into a special Op
+//   %2, %3 = ConcatShapeTranspose(inputs, axis, start, end, perm)
+// This fusion is an experimental work for performance
+
+// Helper function: is the ConcatOp matched to the fusion pattern?
+static bool isConcatFuseMatched(
+    Operation *op, ONNXShapeOp &shapeOp, ONNXTransposeOp &transposeOp) {
+  shapeOp = NULL;
+  transposeOp = NULL;
+  bool failed;
```
tungld, line 215: @chentong319, this code caused the failures on Jenkins since `failed` is not intialized. It happens when building onnx-mlir with Release mode. Chaning this to `bool failed = false` will fix the failures.

chentong319: Thanks

PR #1863, opened 2022-11-16
fixes for mac

AlexandreEichenberger:
Fixes for mac build, originally introduced by #1859
* Warning for non virtual destructors of classes with virtual functions
* `template` keyword where the compiler may be confused as templates are used for the class and a function call within the class. See https://www.aerialmantis.co.uk/blog/2017/03/17/template-keywords/ for explanation on the second point. Added a warning comment in the file that needed this fix so that we may better remember the cause of this extra `template` keyword.


hamptonm1 (approved):
LGTM!

PR #1869, opened 2022-11-21
Support faster sort algorithm in onnx.NonMaxSuppression and onnx.TopK

negiyas:
This PR introduces a faster sort algorithm (quick sort and comb sort) **on CPU** in the emitArgSort function.
 The quick sort algorithm is used for onnx.NonMaxSupression and onnx.TopK operations.

[Conditions to use faster sort algorithm]
This PR uses faster sort algorithms if the following two conditions are met. The original bubble sort code is used in other cases.
1. the input rank should be less or equal to 6
2. the axis to be sorted should be the last rank. Otherwise the original bubble sort code is used.
I consider that the conditions are met in all onnx.NonMaxSupression cases and the most practical cases of onnx.TopK.

[How to guarantee stable sort]
This function is expected to provide "stable" sort (e.g. "test_nonmaxsuppression_identical_boxes_cpu" test in backend test), so that it should preserve the input data order, if the values are the same, although most sort algorithms (e.g. quick sort) are not stable.
To guarantee the original order among the same values, we use special compare functions to compare two values at first, then compare the input orders if the values are the same. Luckily the target list is managed by index and data arrays, and the index array is initialized by the input order (1, 2, 3, ...), so that we can compare the input orders as well as their values.
 

gongsu832:
I'm wondering why you are writing your own quick sort instead of using the qsort in the standard C library?

negiyas:
> I'm wondering why you are writing your own quick sort instead of using the qsort in the standard C library?

I agree that the standard qsort library are preferred to be used, but we need a non-standard compare function to compare two elements.

The compare function for the standard qsort library, as you know, is `int (*compare)(const void *arg1, const void *arg2)`, and the standard qsort function swaps contents of arg1 and arg2 if necessary.

However, in our case, the target list to be sort is managed by two arrays, those are (1) data array holding data themselves, and (2) index array holding offsets of the data array. (c.f. The sort function sorts the index array only, and does not change the data array.)  Accordingly, to compare two elements in the target list, we need to access elements in both index and data arrays.

The issue to use the standard compare function is how to pass all necessary information with its two arguments. For example, we can pass two index elements with the two arguments, but cannot pass the base pointer of the data array. There may be some tricks (e.g. pass base pointer of the data array by a static variable, OR reconfigure(copy) the target array before sorting), but they have drawbacks (e.g. the sort function cannot be reentrant, OR additional processing and memory are required).

Your comments or suggestions are welcome.

gongsu832:
Did you look at the `qsort_r` function? It allows a 3rd argument in which you can pass in arbitrary data (e.g., the base pointers to your data and index array). It's basically designed for situations exactly like yours.
```
       void qsort_r(void *base, size_t nmemb, size_t size,
                  int (*compar)(const void *, const void *, void *),
                  void *arg);

       The  qsort_r()  function is identical to qsort() except that the comparison function
       compar takes a third argument.  A pointer is passed to the comparison  function  via
       arg.   In this way, the comparison function does not need to use global variables to
       pass through arbitrary arguments, and is therefore reentrant  and  safe  to  use  in
       threads.
```

negiyas:

negiyas:
@gongsu832 Thank you for your nice suggestion about qsort_r. (I looked at qsort and qsort_r manual at first, but I did not think it is so important at the time.)
I rewrote this PR to use the standard qsort_f function in all environments supporting the function, but 
some environments (e.g. MacOS and Windows) do not support "qsort_r", because Since it is not a POSIX standard API.
A custom implementation of qsort_r is used for such environments.

gongsu832:
MacOS does support `qsort_r`. It just passes the arguments in a different order from that of Linux:
```
     void
     qsort_r(void *base, size_t nel, size_t width, void *thunk,
         int (*compar)(void *, const void *, const void *));

     The qsort_r() function behaves identically to qsort(), except that it takes an
     additional argument, thunk, which is passed unchanged as the first argument to function
     pointed to compar.  This allows the comparison function to access additional data
     without using global variables, and thus qsort_r() is suitable for use in functions
     which must be reentrant.
```
Also as I mentioned above, on Windows the function is called `qsort_s`: https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/qsort-s?view=msvc-170

negiyas:
@gongsu832 Thank you for the info. I am refactoring the code to use qsort_r and qsort_s on MacOS and Windows.
After confirming all tests passed, I will remove the custom quick sort code.

src/Runtime/OMSort.inc
```diff
@@ -0,0 +1,453 @@
...
+  void *dataPtr = omTensorGetDataPtr(inputTensor);
+  int64_t sort_elems = inputShape[axis];
+  // Sorting not necessary for empty array
+  if (sort_elems == 0)
+    return;
+
+  // Choose the appropriate compare function
+  compareFunctionType *compareElements =
+      getCompareFunction(ascending, dataType);
+  uint64_t datasize = getDataSize(dataType);
```
gongsu832, line 388: We already have `OM_DATA_TYPE_SIZE[dataType]` for this in `OnnxDataType.h`.

src/Runtime/OMSort.inc
```diff
@@ -0,0 +1,453 @@
...
+            sortFunc((void *)idx, sort_elems, sizeof(uint64_t), compareElements,
+                data);
+#endif
+          }
+        }
+      }
+    }
+  }
+  return;
+}
```
gongsu832, line 453: The implementation of qsort seems way too complicated than necessary. Maybe this is something you can make use of for `OMSort.inc`:
```
#include <assert.h>
#include <stddef.h>

#ifndef __cplusplus
#define __USE_GNU
#include <stdlib.h>
#endif

#include "onnx-mlir/Runtime/OnnxDataType.h"

/* Compare function for case when there is an index array and a data array,
 * and the array being sorted is the index array but value comparison is
 * using the data array.
 */
#define COMPARE(type)                                                          \
  do {                                                                         \
    int i1 = *((const type *)p1);                                              \
    int i2 = *((const type *)p2);                                              \
    return (((const type *)qarg->data_ptr)[i1] >                               \
               ((const type *)qarg->data_ptr)[i2]) -                           \
           (((const type *)qarg->data_ptr)[i1] <                               \
               ((const type *)qarg->data_ptr)[i2]);                            \
  } while (0)

/* Custom data passed into the qsort_r/s function */
typedef struct qsort_arg {
  void *data_ptr;         /* pointer to data array  */
  OM_DATA_TYPE data_type; /* ONNX data type         */
} qsort_arg_t;

/* Use comp_ascending for ascending sort,
 * use comp_descending for descending sort.
 */
#if defined(__APPLE__) || defined(_MSC_VER)

extern int comp_ascending(void *arg, const void *p1, const void *p2);
extern int compdecending(void *arg, const void *p1, const void *p2);

int comp_decending(void *arg, const void *p1, const void *p2) {
  return comp_ascending(arg, p2, p1);
}

int comp_ascending(void *arg, const void *p1, const void *p2) {

#else /* Linux */

extern int comp_ascending(const void *p1, const void *p2, void *arg);
extern int comp_decending(const void *p1, const void *p2, void *arg);

int comp_decending(const void *p1, const void *p2, void *arg) {
  return comp_ascending(p2, p1, arg);
}

int comp_ascending(const void *p1, const void *p2, void *arg) {

#endif

  qsort_arg_t *qarg = (qsort_arg_t *)arg;

  switch (qarg->data_type) {
  case ONNX_TYPE_BOOL:
    COMPARE(bool);
  case ONNX_TYPE_UINT8:
    COMPARE(uint8_t);
  case ONNX_TYPE_INT8:
    COMPARE(int8_t);
  case ONNX_TYPE_UINT16:
    COMPARE(uint16_t);
  case ONNX_TYPE_INT16:
    COMPARE(int16_t);
  case ONNX_TYPE_UINT32:
    COMPARE(uint32_t);
  case ONNX_TYPE_INT32:
    COMPARE(int32_t);
  case ONNX_TYPE_UINT64:
    COMPARE(uint64_t);
  case ONNX_TYPE_INT64:
    COMPARE(int64_t);
  case ONNX_TYPE_FLOAT:
    COMPARE(float);
  case ONNX_TYPE_DOUBLE:
    COMPARE(double);
  default:
    assert(false && "unexpected data type in compare function");
  }
}

/* quick sort wrapper function to deal with slightly different qsort
 * variations on different platforms.
 *
 * For sorting an index array whose values are in a separate data array,
 * use arg to pass in a pointer to qsort_arg_t struct.
 *
 * For ascending sort, use the compare function comp_ascending.
 * For descending sort, use the compare function comp_descending.
 */
void quick_sort(void *base, size_t num, size_t size,
#if defined(__APPLE__) || defined(_MSC_VER)
    int (*comp)(void *, const void *, const void *),
#else /* Linux */
    int (*comp)(const void *, const void *, void *),
#endif
    void *arg) {
#if defined(__APPLE__)
  qsort_r(base, num, size, arg, comp);

#elif defined(_MSC_VER)
  qsort_s(base, num, size, comp, arg);

#else /* Linux */
  qsort_r(base, num, size, comp, arg);
#endif
}
```

AlexandreEichenberger:
Would it make sense to have some template (zero overhead, or something similar) that can be instantiated for each OS and that would centralize the different calls to capture in one spot the difference among OSs?

The idea would be that the special code would all be contained into one spot, instead of spreading wherever there are calls that need to be OS specific.

Any suggestions @gongsu832 ? There are probably solutions out there that we can adopt to do this cleanly.

gongsu832:
The code suggested above basically does that for qsort. On all platforms, you make the same `quick_sort(base, num, size, comp_ascending, arg)` or `quick_sort(base, num, size, comp_descending, arg)` call.

src/Runtime/OMSort.inc
negiyas: Fixed to use OM_DATA_TYPE[dataType] instead of the custom getDataSize function. Thank you!

src/Runtime/OMSort.inc
negiyas: Thank you for the comment. I agree that your method is simpler.

Let me explain the reason why type-specific sort and compare functions are used.

As background, the sort and compare functions are called in very tight loop so many times (80 * 10646 * 10647/2 times in the Yolov3 case), and only the sort function is called in the loop as you can see the code below.

Type specific sort and compare functions are used for the performance optimization, by using type-specific sort and compare functions, we can move "switch" statements to check the data type called in the loop to the outsize of the loop. And I suppose that it improves the performance.

```
  // Sort the the 6th axis in the outer 5 loops
  for (int dim0 = 0; dim0 < shape[0]; dim0++) {
    for (int dim1 = 0; dim1 < shape[1]; dim1++) {
      for (int dim2 = 0; dim2 < shape[2]; dim2++) {
        for (int dim3 = 0; dim3 < shape[3]; dim3++) {
          for (int dim4 = 0; dim4 < shape[4]; dim4++) {
            uint64_t off = dim0 * strides[0] + dim1 * strides[1] +
                           dim2 * strides[2] + dim3 * strides[3] +
                           dim4 * strides[4];
            void *data = ((char *)dataPtr) + datasize * off;
            uint64_t *idx = order + off;
#if defined(__APPLE__)
            sortFunc((void *)idx, sort_elems, sizeof(uint64_t), data, compareElements);
#else
            sortFunc((void *)idx, sort_elems, sizeof(uint64_t), compareElements, data);
#endif
          }
        }
      }
    }
  }
```

gongsu832: OK thanks for the background. It would be nice to have some comments about all these complexities are for optimizing performance.

negiyas: I appended some descriptions to explain the reason why data-specific functions are used. Thank you for the comments.

negiyas:
We found that the  "test_nonmaxsuppression_identical_boxes_cpu" test in backend test including the identical boxes fails on MacOS.  The test sorts an array including 10 exactly same floating point numbers, and it expects that the original order is not changed, but, as you know, the quick sort and comb sort algorithms are not stable sort algorithms, so that we cannot expect that the input order is preserved.

To avoid this issue, we changed compare functions to compare two values at first and compare the original orders themselves if the values are the same. Fortunately the list is managed by index and data arrays, and the index array is initialized by the input order (1, 2, 3, ...).  

This compare functions guarantee the input order among the same values and makes sorting algorithms stable.

AlexandreEichenberger:
@gongsu832 do you mind approving the PR, or making further suggestions for changes? would be great to have this PR merged. Thanks

src/Runtime/OMSort.inc
```diff
@@ -0,0 +1,460 @@
...
+#define declare_compare_function(fname, typeName, direction, symbol)           \
+  int compare##fname##direction(                                               \
+      const void *idx1p, const void *idx2p, void *dataPtr) {                   \
+    uint64_t idx1 = *((uint64_t *)idx1p);                                      \
+    uint64_t idx2 = *((uint64_t *)idx2p);                                      \
+    typeName *data = (typeName *)dataPtr;                                      \
+    return (data[idx1] symbol data[idx2]) ||                                   \
+           ((data[idx1] == data[idx2]) && (idx1 > idx2));                      \
+  }
+#endif
```
gongsu832, line 80: Windows and Linux cases are missing the fix for making qsort stable.

negiyas: @gongsu832 The "? 1 : -1" part is comes from comparison function specification only for MacOS, the compare functions for Windows and Linux should return 0 if comparison fails.

negiyas:

src/Runtime/OMSort.inc
gongsu832: Not sure where you get that. On all platforms, the comparison function should return < 0, = 0, or > 0 if the 1st element is less than, equal to, or larger than the 2nd element. Of course, to make it stable, you can return > 0 or < 0 when the two elements are equal.

negiyas: @gongsu832 I am sorry. I fixed the issues. Thank you!

negiyas:
Can someone look at the following backend test errors on MacOS?
It seems that numpy is not install on the test Python environment.

```
======================================================================
ERROR: test_abs_cpu (__main__.OnnxBackendNodeModelTest)
----------------------------------------------------------------------
  File "/Users/runner/hostedtoolcache/Python/3.10.9/x64/lib/python3.10/site-packages/onnx/backend/test/runner/__init__.py", line 259, in device_test_func
    return test_func(*args, device=device, **kwargs)
  File "/Users/runner/hostedtoolcache/Python/3.10.9/x64/lib/python3.10/site-packages/onnx/backend/test/runner/__init__.py", line 315, in run
    self.assert_similar_outputs(ref_outputs, outputs,
  File "/Users/runner/work/onnx-mlir/onnx-mlir/build/test/backend/Debug/inference_backend.py", line 1173, in assert_similar_outputs
    super(InferenceBackendTest, cls).assert_similar_outputs(ref_outputs, outputs, rtol, atol)
  File "/Users/runner/hostedtoolcache/Python/3.10.9/x64/lib/python3.10/site-packages/onnx/backend/test/runner/__init__.py", line 181, in assert_similar_outputs
    if ref_outputs[i].dtype == np.object:
  File "/Users/runner/hostedtoolcache/Python/3.10.9/x64/lib/python3.10/site-packages/numpy/__init__.py", line 284, in __getattr__
    raise AttributeError("module {!r} has no attribute "
AttributeError: module 'numpy' has no attribute 'object'

======================================================================
ERROR: test_acos_cpu (__main__.OnnxBackendNodeModelTest)
----------------------------------------------------------------------
  File "/Users/runner/hostedtoolcache/Python/3.10.9/x64/lib/python3.10/site-packages/onnx/backend/test/runner/__init__.py", line 259, in device_test_func
    return test_func(*args, device=device, **kwargs)
  File "/Users/runner/hostedtoolcache/Python/3.10.9/x64/lib/python3.10/site-packages/onnx/backend/test/runner/__init__.py", line 315, in run
    self.assert_similar_outputs(ref_outputs, outputs,
  File "/Users/runner/work/onnx-mlir/onnx-mlir/build/test/backend/Debug/inference_backend.py", line 1173, in assert_similar_outputs
    super(InferenceBackendTest, cls).assert_similar_outputs(ref_outputs, outputs, rtol, atol)
  File "/Users/runner/hostedtoolcache/Python/3.10.9/x64/lib/python3.10/site-packages/onnx/backend/test/runner/__init__.py", line 181, in assert_similar_outputs
    if ref_outputs[i].dtype == np.object:
  File "/Users/runner/hostedtoolcache/Python/3.10.9/x64/lib/python3.10/site-packages/numpy/__init__.py", line 284, in __getattr__
    raise AttributeError("module {!r} has no attribute "
AttributeError: module 'numpy' has no attribute 'object'

```

gongsu832:
@negiyas I assume the intention of keeping the different sorting algorithms is for comparison purpose but ultimately only one is needed and the others will be removed?

negiyas:
@gongsu832  I removed all code for other sorting algorithms except the quick sort, custom code of the quick sort in OMSort.inc, all code for selecting sorting algorithms, and fixed related lit tests. Thank you.

gongsu832 (approved):
LGTM

PR #1871, opened 2022-11-21
Simplify and speed up ONNXOpTransformPass hash

sorenlassen:
raw_sha1_ostream bypasses the creation of a temporary file.

Tested with `onnx-mlir resnet152-v1-7.onnx --mlir-timing` on Mac and saw ONNXOpTransformPass timing reduce 3x from 21s to 7s.

Also added AsmPrinter cloptions to onnx-mlir so we can call with `--mlir-elide-elementsattrs-if-larger=1` which further reduced ONNXOpTransformPass timing to 1.5s.


chentong319 (approved):
LGTM!

PR #1873, opened 2022-11-21
added conversion of add and sub from onnx to tosa

ljfitz:
Where possible, the implementation aligns with the [TensorFlow to Tosa conversion](https://github.com/tensorflow/tensorflow/blob/278edb0ebf089c4c0d26e7bbcbad31eb94423659/tensorflow/compiler/mlir/tosa/transforms/legalize_tfl.cc#L583) for ease of cross-referencing conversions to tosa-mlir.

Signed-off-by: Liam Fitzpatrick <liam.fitzpatrick@amd.com>


test/mlir/conversion/onnx_to_tosa/Math/Elementwise.mlir
```diff
@@ -28,10 +30,49 @@ func.func @test_neg(%arg0: tensor<10x10xf32>) -> tensor<10x10xf32> {
 // CHECK-NEXT:      [[VAR_0_:%.+]] = "tosa.negate"([[PARAM_0_]]) : (tensor<10x10xf32>) -> tensor<10x10xf32>
 }
 
+// -----
```
tungld, line 33: I suppose that we should have an empty line before and after `// -----` to make it work.

src/Conversion/ONNXToTOSA/Math/Elementwise.cpp
```diff
@@ -42,6 +42,38 @@ class ONNXElementwiseUnaryOpLoweringToTOSA
   }
 };
 
+template <typename ONNXOpT, typename TosaOpT>
+class ONNXAddSubOpLoweringToTOSA : public OpConversionPattern<ONNXOpT> {
```
tungld, line 46: This looks good, but please consider to have a more general name like `ONNXBinaryElementwiseOpLoweringToTOSA` if more ops (e.g. Min, Max, ...) fall into this category.

ljfitz: You're right @tungld - we're trying to keep somewhat aligned with the related TFL/torch-to-tosa conversion passes, but this is an example where a generalization is appropriate in ONNX that isn't for TFL.

tungld (approved):
LGTM! Thanks for the patch!

tungld:
@ljfitz could you update your PR with the current base branch? I somehow cannot do that to this PR. Thanks!

eric-k256 (approved):
TOSA-wise, looks good, nice contribution.

PR #1874, opened 2022-11-21
DisposableElementsAttr for const prop with garbage collection

sorenlassen:
Introduce a new attribute DisposableElementsAttr. See the [PDF presentation](https://github.com/onnx/onnx-mlir/wiki/presentations/DisposableElementsAttr202212.pdf) from the [Nov 29th meeting](https://github.com/onnx/onnx-mlir/wiki/Informal-meeting-agenda-and-notes#nov-29th).

It is a little bit like DenseResourceElementsAttr in that it can be backed by heap allocated memory which can be freed at the end of a compiler pass.

It's primary purpose is to replace the buffer pool in onnx-mlir constant propagation. See [onnx-mlir issue 1718](https://github.com/onnx/onnx-mlir/issues/1718) for a description of the problems with the buffer pool, and see [PR 1794](https://github.com/onnx/onnx-mlir/pull/1794) for how to replace it with DenseResourceElementsAttr.

DisposableElementsAttr behaves more like a conventional DenseElementsAttr and is therefore easier for the rest of the code base to interoperate with.

DisposableElementsAttr feature set:
* Abstracts from the representation of the underlying data, which allows:
  * Data can be a "wide" high precision cpp type, e.g. float can be represented by double, uint16_t can be represented by uint64_t,
  * Data is managed by a llvm::MemoryBuffer and memory can owned/unowned or mmapped file,
  * Data is reference counted with std::shared_ptr and can be shared,
  * Data can be "strided" and present a view, like pytorch tensors,
  * Data can apply element-wise transformation on demand when accessed.
* Interoperable:
  * Implements same ElementsAttr interface as DenseElementsAttr
  * API hides implementation complexity from callers
  * Prints the same DenseElementsAttr (no lit tests change, new attr is transparent to mlir assembly)
* Designed for efficiency:
   * Low memory usage (apart from the use of wide cpp types) through shared buffers, strides, garbage collection
   * Avoids most memory copies
   * Inlines functions in inner loops without template instantiation combinatorial explosion
   * Speeds up constant propagation 20x for models with extensive constant propagation (resnet152-v1-7, arcfaceresnet100-8).

A new enum BType and a union WideNum are introduced to manage tensor element types and element representation. They cover all non-complex onnx tensor datatypes, including FLOAT16 and BFLOAT16. `FrontendDialectHelper.cpp` was updated to make use of BType and can now ingest BFLOAT16. WideNum is used extensively in `ConstProp.cpp`.

`DisposableGarbageCollector` garbage collects DisposableElementsAttr MemoryBuffers between compiler passes.

The use of DisposableElementsAttr instead of DenseElementsAttr is an implementation detail and therefore hidden from lit test assembly output. This is done by printing DisposableElementsAttr as DenseElementsAttr in the ONNXConstantOp custom assembly format introduced in PR #1898.

This PR was run on the models resnet152-v1-7, arcfaceresnet100-8 with build type Release and flag --mlir-timing. The ConstPropONNXToONNXPass timing went from 17s to 0.8s on a MacBook. (And const prop is where most of the time was spent in the compiler passes. If we also address the slow hashing in ONNXOpTransformPass we can speed up end to end compilation time for these models 2x from 50-60s to 25s.)

Added constant propagation report controlled by a flag --onnx-const-prop-report:
```
onnx-mlir resnet152-v1-7.onnx --onnx-const-prop-report

constprop report (cumulative), entries: 5, total invocations:1534, total input elements:83885151
  Concat invocations:98 input elements:196
  Unsqueeze invocations:253 input elements:134912
  ElementwiseUnary invocations:210 input elements:91584
  Reshape invocations:98 input elements:24940544
  ElementwiseBinary invocations:875 input elements:58717915
```
```
onnx-mlir arcfaceresnet100-8.onnx --onnx-const-prop-report

constprop report (cumulative), entries: 4, total invocations:1225, total input elements:52607322
  Unsqueeze invocations:203 input elements:45440
  ElementwiseUnary invocations:257 input elements:57473
  Reshape invocations:50 input elements:11136
  ElementwiseBinary invocations:715 input elements:52493273
```

To measure the memory savings I used `libgmalloc` with `MallocStackLogging` and the `heap` tool on Mac (see PR #1951). I tested with the arcfaceresnet100-8 model, where almost all the constant propagation happens in the first const prop pass. WIth the old buffer pool I measured the heap immediately before const prop and then before and after the "clean up" replacement of the buffer_id helper attributes with DenseElementsAttr:
```
Before const prop:
Physical footprint:         275.9M
Physical footprint (peak):  537.5M
All zones: 36965 nodes (271531744 bytes)

   COUNT      BYTES       AVG   CLASS_NAME                                        TYPE    BINARY
   =====      =====       ===   ==========                                        ====    ======
     307  267095552  870018.1   memalign in llvm::function_ref<mlir::StorageUniquer::BaseStorage* (mlir::StorageUniquer::StorageAllocator&)>::callback_fn<mlir::detail::DenseIntOrFPElementsAttrStorage* ...
```
```
After const prop, before cleaning up helper attributes:
Physical footprint:         883.8M
Physical footprint (peak):  883.8M
All zones: 37072 nodes (1122468096 bytes)

   COUNT      BYTES       AVG   CLASS_NAME                                        TYPE    BINARY
   =====      =====       ===   ==========                                        ====    ======
    1891  850679216  449856.8   calloc in onnx_mlir::allocateBufferFor(mlir::Type, bool)
     307  267095552  870018.1   memalign in llvm::function_ref<mlir::StorageUniquer::BaseStorage* (mlir::StorageUniquer::StorageAllocator&)>::callback_fn<mlir::detail::DenseIntOrFPElementsAttrStorage* ...
```
```
After const prop, after cleaning up helper attributes:
Physical footprint:         515.4M
Physical footprint (peak):  1.3G
All zones: 35365 nodes (501215600 bytes)

   COUNT      BYTES       AVG   CLASS_NAME                                        TYPE    BINARY
   =====      =====       ===   ==========                                        ====    ======
     482  496479232  1030039.9   memalign in llvm::function_ref<mlir::StorageUniquer::BaseStorage* (mlir::StorageUniquer::StorageAllocator&)>::callback_fn<mlir::detail::DenseIntOrFPElementsAttrStorage* ...
```
With DisposableElementsAttr I ran with `--mlir-elide-elementsattrs-if-larger=1` to avoid the ONNXOpTransformPass printing the contents of the DisposableElementsAttr, which currently creates a DenseElementsAttr copy and fills up memory (see TODO in `DisposableElementsAttr::printAsDenseElementsAttr()` and PR #1871). I measured immediately before and after the first const prop pass, and then again immediately before and after ScrubDisposablePass which replaces all DisposableElementsAttr with DenseElementsAttr and frees all allocated MemoryBuffers:
```
Before const prop:
Physical footprint:         283.2M
Physical footprint (peak):  520.9M
All zones: 39154 nodes (271726352 bytes)

   COUNT      BYTES       AVG   CLASS_NAME                                        TYPE    BINARY
   =====      =====       ===   ==========                                        ====    ======
     823  267108912  324555.2   (anonymous namespace)::MemoryBufferMem<llvm::WritableMemoryBuffer>
```
```
After const prop:
Physical footprint:         570.3M
Physical footprint (peak):  570.3M
All zones: 41067 nodes (697539904 bytes)

   COUNT      BYTES       AVG   CLASS_NAME                                        TYPE    BINARY
   =====      =====       ===   ==========                                        ====    ======
    1385  692657904  500114.0   (anonymous namespace)::MemoryBufferMem<llvm::WritableMemoryBuffer>
```
```
Before scrub:
Physical footprint:         383.5M
Physical footprint (peak):  570.3M
All zones: 36069 nodes (481175008 bytes)

   COUNT      BYTES       AVG   CLASS_NAME                                        TYPE    BINARY
   =====      =====       ===   ==========                                        ====    ======
     364  476439184  1308898.9   (anonymous namespace)::MemoryBufferMem<llvm::WritableMemoryBuffer>
```
```
After scrub:
Physical footprint:         151.8M
Physical footprint (peak):  570.3M
All zones: 35189 nodes (321206592 bytes)

   COUNT      BYTES       AVG   CLASS_NAME                                        TYPE    BINARY
   =====      =====       ===   ==========                                        ====    ======
     184  316448256  1719827.5   memalign in llvm::function_ref<mlir::StorageUniquer::BaseStorage* (mlir::StorageUniquer::StorageAllocator&)>::callback_fn<mlir::detail::DenseIntOrFPElementsAttrStorage* ...
```
(I'm not sure why the `heap` says physical footprint 151MB much smaller than the 321MB in "All zones". Regardless, the numbers for DIsposableElementsAttr are smaller than for buffer pool.)

So, all in all, memory usage with buffer pool peaked at 1.3GB and ended at 515MB, while DisposableElementsAttr peaked at 570MB and ended at 151MB (or 321MB?).  So DisposableElementsAttr uses less memory and peak memory usage is less than half of buffer pool.

Some follow-ups I will address after this PR lands are:
- Add scatterND, slice, concat, and gather builder methods to ElementsAttrBuilder (see TODOs in ConstProp.cpp).
- Rename "BType" to "DataType".
- Expand the document [here](https://github.com/onnx/onnx-mlir/blob/main/docs/ConstPropagationPass.md) with a guidance for adding OnnxElementsAttrBuilder methods for new ops.


sorenlassen:
@tungld any idea why the MacOS amd64 CI build fails like this?
```
[1/34] cd /Users/runner/work/onnx-mlir/onnx-mlir/build/src/Runtime/jni && /usr/local/Cellar/cmake/3.25.0/bin/cmake -E copy_if_different /Users/runner/work/onnx-mlir/onnx-mlir/build/src/Runtime/jni/jsoniter/maven/jsoniter-0.9.23.jar /Users/runner/work/onnx-mlir/onnx-mlir/build/src/Runtime/jni/jsoniter/jsoniter-0.9.23.jar
[2/31] Running utility command for javaruntime
[3/31] cd /Users/runner/work/onnx-mlir/onnx-mlir/build/test/backend && ONNX_MODELS=/Users/runner/work/onnx-mlir/onnx-mlir/build/test/backend/Debug/models /usr/local/Frameworks/Python.framework/Versions/3.11/bin/python3.11 /Users/runner/work/onnx-mlir/onnx-mlir/build/test/backend/Debug/test.py
FAILED: test/backend/CMakeFiles/check-onnx-backend /Users/runner/work/onnx-mlir/onnx-mlir/build/test/backend/CMakeFiles/check-onnx-backend 
cd /Users/runner/work/onnx-mlir/onnx-mlir/build/test/backend && ONNX_MODELS=/Users/runner/work/onnx-mlir/onnx-mlir/build/test/backend/Debug/models /usr/local/Frameworks/Python.framework/Versions/3.11/bin/python3.11 /Users/runner/work/onnx-mlir/onnx-mlir/build/test/backend/Debug/test.py
  File "/Users/runner/work/onnx-mlir/onnx-mlir/build/test/backend/Debug/test.py", line 30, in <module>
    import onnx
ModuleNotFoundError: No module named 'onnx'
```

tungld:
@sorenlassen somehow it wasn't able to find the installed onnx. We can ignore it at this moment. @gongsu832 fyi.

gongsu832:
Apparently system configuration has changed yet again causing onnx and onnx-mlir finding different python versions. The former finds `/Users/runner/hostedtoolcache/Python/3.10.8/x64/bin/python3` while the latter finds `/usr/local/Frameworks/Python.framework/Versions/3.11/bin/python3.11`. Sigh, I'm starting getting tired of github actions.

gongsu832:
PR https://github.com/onnx/onnx-mlir/pull/1877 should fix the MacOS build.

tungld:
@sorenlassen looking at the speed up, it looks promising. How about memory consumption, e.g. the peak memory consumption during compilation? It seems that you call GC between passes, so all intermediate created in a single constant propagation are piled up, aren't they?

sorenlassen:
> How about memory consumption, e.g. the peak memory consumption during compilation? It seems that you call GC between passes, so all intermediate created in a single constant propagation are piled up, aren't they?

I haven't measured it. Do you have any suggestions for how to measure?

Yes, the all intermediate attributes pile up during a single const prop pass pile up, like the old buffer pool, but in theory it should be better because:
1. When creating a DisposableElementsAttr from a DenseElementsAttr no memory is allocated (other than metadata) because it's constructed with a MemoryBuffer that points into the raw data of the (immortal) DenseElementsAttr.
2. ConstPropTranspose and ConstPropExpand shares the data of their input and only attaches new strides.
3. ConstPropCast, ComputeConstPropElementwiseUnary (and ComputeConstPropElementwiseBinary when one input is splat) also shared the data of their input and only attaches an elementwise transform function.

tungld:
> I haven't measured it. Do you have any suggestions for how to measure?

I simply use `top` command to roughly observe the memory consumption.

> ConstPropTranspose and ConstPropExpand share the data of their input (with shared_ptr) and only attach new strides.

How about the situation in which the input of Transpose has multiple uses? In other words, The Transpose we are doing constant propagation is not the only use of its input.

sorenlassen:
@tungld can you recommend models to test const prop on?

I ran through the model zoo and found that resnet152-v1-7 and arcfaceresnet100-8 did the most const prop and measured performance on these (see PR description above) but I'm interested in trying out more models.

sorenlassen:
> > I haven't measured it. Do you have any suggestions for how to measure?
> 
> I simply use `top` command to roughly observe the memory consumption.

I found `libgmalloc` and the `heap` tool on Mac and measured the memory usage with buffer pool and DisposableElementsAttr. I updated the PR description with my findings. DisposableElementsAttr uses less memory and peak memory usage is less than half of buffer pool.

sorenlassen:
@tungld and @chentong319, I merged the constant-materializer branch from PR #1799, as discussed last week, see [this commit.](https://github.com/onnx/onnx-mlir/pull/1874/commits/a174a692ea87e8f14686e2a0f8e00651b275e46b)

On top of that I implemented `fold` for ONNXTransposeOp in [this commit.](https://github.com/onnx/onnx-mlir/pull/1874/commits/83c89155d94d1a70e0cddf25b7fea60ced192ad8)

I had to remove the onnx_elide_constants.mlir lit test (see PR #1899).

I also had to comment out the lit tests in onnx_canonicalization.mlir where constants ended up with dynamic types (@test_shape1, @test_size1, @test_constant_1). I'm not sure how to address this properly, but I think that's beyond the scope of DisposableElementsAttr.

sorenlassen:
I moved the constant folding logic from this PR to a draft PR #1902

sorenlassen:
fyi, I landed PR #1898 with onnx constant custom assembly format and based on that I plan over the next day or two to rework this PR to modify how DisposableElementsAttr is printed and, of course, resolve the merge conflicts

AlexandreEichenberger:
@sorenlassen 
Could you give a high level view of which PR is doing what with respect to this proposal, just maybe a one-liner per pr so that it be easier to dive in reviewing these.

Just looking at the interfaces, I see that you moved `DenseElementsAttr` to `ElementsAttr`, but you are proposing the use of `DisposableElementsAttr`. Is the goal that it can be switched to the one or the other? A short explainer on this (and where the code is) would be helpful as well.

sorenlassen:
> Could you give a high level view of which PR is doing what with respect to this proposal

Will do.

FYI, I just uploaded the presentation from last week and added a [link](https://github.com/onnx/onnx-mlir/wiki/presentations/DisposableElementsAttr202212.pdf) from the wiki meeting notes.

sorenlassen:
> Could you give a high level view of which PR is doing what with respect to this proposal, just maybe a one-liner per pr so that it be easier to dive in reviewing these.

This PR introduces DisposableElementsAttr and rewrites the constant propagation logic in ConstProp.cpp to replace the buffer pool. This is a standalone contribution which leads to the speedup and memory saving explained in the PR description.

It also opens the door to two further opportunities which is explained in these two draft PRs:
1. Draft PR #1902: Constant folding (see PR #1799 and the discussion in issue #1718).
2. Draft PR #1770: Hybrid ONNX-to-ONNX transformation pass.

> Just looking at the interfaces, I see that you moved `DenseElementsAttr` to `ElementsAttr`, but you are proposing the use of `DisposableElementsAttr`. Is the goal that it can be switched to the one or the other? A short explainer on this (and where the code is) would be helpful as well.

I found that virtually everywhere in the onnx-mlir code base where we use DenseElementsAttr we only use the methods defined in the ElementsAttr interface, so I changed a bunch of places in the existing code to make that explicit. This makes it possible to substitute DisposableElementsAttr (which also implements ElementsAttr) for DenseElementsAttr in many places - in this PR FrontEndDialectHelper imports onnx tensor constants as DisposableElementsAttr and the new constants created during constant propagation use DisposableElementsAttr.

With this PR some onnx constants are created with DisposableElementsAttr (in the places mentioned above) and in some other places existing logic that creates small onnx constants with DenseElementsAttr remain, e.g. in `createDenseElementsAttrFromShape()` in Rewrite.cpp.

tungld:
@sorenlassen thank you again for the great work.

I tried to infer a global picture from the patch. @sorenlassen corrects if there is something wrong.

Both `DenseElementsAttr` and the new `DisposableElementsAttr` are subclasses of ElementsAttr. `DenseElementsAttr` are used explicitly in the IR while `DisposableElementsAttr` is used implicitly. In other words, `DenseElementsAttr` in the IR will be parsed/read as `DisposableElementsAttr` whose underlying buffer will go to a `DisposablePool`, and `DisposableElementsAttr` will be then printed out as `DenseElementsAttr`'s format in the IR. This makes me a bit puzzled  in knowing when to use which one.

`DisposableElementsAttr` is created via ElementsAttrBuilder that will put the attribute into the `DisposablePool`. `DisposablePool` is a dialect interface, living in the dialect context. Thus, the `DisposablePool` would live through multiple passes.  Garbage collection will be called at the end of each pass. This is quite clever. 

At any time, we can scrub a `DisposableElementsAttr` into a `DenseElementsAttr`. But `DenseElementsAttr`'s buffer is immortal, we should do that at a good timing.

Since we change how constant propagation is implemented, could you please update the document (here)[https://github.com/onnx/onnx-mlir/blob/main/docs/ConstPropagationPass.md] about how to write constant propagation for a new op?

src/Dialect/ONNX/ElementsAttrBuilder.hpp
```diff
@@ -0,0 +1,177 @@
...
+
+  static std::atomic<size_t> counter;
+
+  DisposablePool &disposablePool;
+};
+
+//===----------------------------------------------------------------------===//
+// Deferred Method Definitions
+//
+// TODO: move so standalone ElementsAttrBuilder.inc source file
```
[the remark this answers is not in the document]
sorenlassen: I went ahead and moved to .hpp.inc

src/Dialect/ONNX/ElementsAttrBuilder.hpp
```diff
@@ -0,0 +1,177 @@
...
+
+#include "src/Dialect/ONNX/DisposableElementsAttr.hpp"
+#include "src/Dialect/ONNX/DisposablePool.hpp"
+#include "src/Support/WideNum.hpp"
+
+#include <atomic>
+
+namespace onnx_mlir {
+
+class ElementsAttrBuilder {
```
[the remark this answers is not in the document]
sorenlassen: I added comments and reduced the number of public methods to make the interface easier to document

tungld:
@sorenlassen
1. How about postponing the FLOAT16 and BFLOAT16 support to another patch? That makes the patch clean and I believe it's not essential  in this patch.
2. I would hide all `BType`, `WideNum`, `Array`, etc. inside `DisposableElementsAttr` or `ElementsAttrBuilder`. I don't think we will use them outside the purpose of doing constant folding with the biggest precision. Or, we can have a single folder for all stuffs related to `DisposableElementsAttr` and `ElementsAttrBuilder`.
3. With (2), we move all primitive constant computations in ConstProp.cpp into `ElementsAttrBuilder`, similar to what we are doing with DialectBuilder. For example,
```mlir
DisposableElementsAttr lhs = getConstValueAsDisposableElements(elementsBuilder, lhsValue);
DisposableElementsAttr rhs = getConstValueAsDisposableElements(elementsBuilder, rhsValue);
ElementsAttr resultElements = elementsBuilder.binaryOp<ElementwiseBinaryOp>(lhs, rhs, replacingType)
```
where `binaryOp` will be in `ElementsAttrBuilder` and BType, WideNum are hidden.

In other words, users use only `ElementsAttrBuilder` to create `DisposableElementsAttr` and do computations with `DisposableElementsAttr` as inputs and outputs.

4. Please update the document [here](https://github.com/onnx/onnx-mlir/blob/main/docs/ConstPropagationPass.md) about how to use the new interface.

src/Transform/ONNX/ConstProp.cpp
```diff
@@ -243,229 +194,125 @@ ONNXConstantOp createConstantOpAndStoreBufferPtr(
...
+
+  ElementsAttrBuilder elementsBuilder(rewriter.getContext());
+  DisposableElementsAttr lhs =
+      getConstValueAsDisposableElements(elementsBuilder, lhsValue);
+  DisposableElementsAttr rhs =
+      getConstValueAsDisposableElements(elementsBuilder, rhsValue);
+  BType operandsBType = lhs.getBType();
+  assert(operandsBType == rhs.getBType());
+  ElementsAttr resultElements = elementsBuilder.combine(lhs, rhs, replacingType,
+      combinerOfElementwiseBinaryOp<ElementwiseBinaryOp>(operandsBType));
```
tungld, line 251: I would move `combinerOfElementwiseBinaryOp` into `ElementsAttrBuilder` and provide, say `ElementsAttrBuilder.binaryOp<ElementwiseBinaryOp>`, which  is the same as you did for `reshape`, `transpose`. By that way, we don't need to expose `BType` here.

sorenlassen: all your 4 suggestions are good, I'll go to work on all of them and let you know if I run into any obstacles

sorenlassen:
> How about postponing the FLOAT16 and BFLOAT16 support to another patch? That makes the patch clean and I believe it's not essential  in this patch.

If you don't mind, I'd prefer if we land FLOAT16 and BFLOAT16 first. I created a separate PR #1926 with those.

src/Dialect/ONNX/DisposableElementsAttributeStorage.hpp
```diff
@@ -0,0 +1,118 @@
...
+  // In this case the buffer data type and the type's element type must promote
+  // to the same double/i64/u64 widetype
+  // (both are float, or both are signed ints, or both are unsigned ints).
+  //
+  // Garbage collection clears the reader when the DisposableElementsAttr is
+  // disposed.
+  Reader reader;
+};
+
+} // namespace mlir
```
AlexandreEichenberger, line 118: not that it matters a lot, but it looks like the formatter like empty line at the end of the file.

src/Dialect/ONNX/DisposablePool.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+class DisposablePool : public mlir::DialectInterface::Base<DisposablePool> {
+  friend class ElementsAttrBuilder; // allow access to insert()
+public:
+  static DisposablePool &create(mlir::MLIRContext *context);
+
+  static DisposablePool *get(mlir::MLIRContext *context);
+
+  // Disposes every DisposableElementsAttr and in moduleOp replaces each with a
+  // DenseElementsAttr.
```
AlexandreEichenberger, line 38: Can you comment in the comment if that process is reversible, or meant to be used to permanently go toward regular dense element attributes. Tx

src/Dialect/ONNX/DisposablePool.hpp
```diff
@@ -0,0 +1,81 @@
...
+  ~DisposablePool();
+
+  mlir::DisposableElementsAttr lookup(size_t id) const;
+
+  // Disposes every DisposableElementsAttr in the pool which is unreachable
+  // (doesn't appear in moduleOp).
+  void garbageCollectUnreachable(mlir::ModuleOp moduleOp);
+
+  void close() {
+    assert(pool.empty() && "pool must be scrubbed before close");
```
AlexandreEichenberger, line 51: Should there be a `close` in the destructor, to make sure it is not simply lost? Or maybe there is no need as there is one obj for the duration of the compilation unit.

src/Dialect/ONNX/ElementsAttrBuilder.hpp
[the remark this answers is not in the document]
AlexandreEichenberger: thanks

src/Dialect/ONNX/ElementsAttrBuilder.hpp
```diff
@@ -0,0 +1,127 @@
...
+#include <atomic>
+
+namespace onnx_mlir {
+
+// Builds DisposableElementsAttr instances.
+// Every instance is inserted into DisposablePool which garbage collects
+// unreachable instances between compiler passes.
+class ElementsAttrBuilder {
+public:
+  ElementsAttrBuilder(DisposablePool &disposablePool);
```
AlexandreEichenberger, line 27: are the two constructor identical? just different objects, or there is a difference between the two. I would specify either way in a comment

src/Dialect/ONNX/ElementsAttrBuilder.hpp
```diff
@@ -0,0 +1,127 @@
...
+  // applying the given function to the element.
+  template <typename UnaryFunction = std::function<WideNum(WideNum)>>
+  static Transformer functionTransformer(UnaryFunction fun);
+
+  // Returns a DisposableElementsAttr where each element is transformed
+  // by running the given transformer on all the elements.
+  mlir::DisposableElementsAttr transform(mlir::DisposableElementsAttr elms,
+      mlir::Type transformedElementType, Transformer transformer);
+
+  // Returns a DisposableElementsAttr that is the result of applying
```
AlexandreEichenberger, line 72: Can you specify which calls below reuse the original copy of the data and only add an "accessor" method that will compute the actual values when executed? probably all, so you could just state that.

src/Support/Strides.hpp
```diff
@@ -0,0 +1,118 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//===---------------------------- Strides.hpp -----------------------------===//
+//
+// Strides helper functions.
```
AlexandreEichenberger, line 7: Could you just add small description of what stride are supported? N dimensional? does it have to be multiple of the dimension of the original arrays, are they any restrictions...

AlexandreEichenberger (commented):
Asking for a few more explanation/comments in the code. As a developer, it may look obvious, it will probably help a casual user of this new data structure.

On the actual code, I will defer to @tungld and @chentong319 who have more experience on this area of the compiler.

AlexandreEichenberger:
@sorenlassen let us know when you want us to review it.

src/Dialect/ONNX/DisposableElementsAttr.hpp
```diff
@@ -0,0 +1,277 @@
...
+  //
+  // Use value_begin<X>(), value_end(), or getValues<X>() to iterate over the
+  // the elements, where X can be any scalar cpp type, or APFloat (if element
+  // type is floating point) or APInt (if element type is integer).
+  //
+  // Note that iteration is slow because it invokes getReader() for every
+  // element and, furthermore, performs a slow calculation from flat index to
+  // buffer position if the underlying buffer is not contiguous, namely when its
+  // strides are not the default strides for the type shape. It's more efficient
+  // to copy out data in bulk with readWideNums().
```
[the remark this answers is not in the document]
sorenlassen: expanded the comments with an example:
```
//===--------------------------------------------------------------------===//
  // Other access to the elements:
  //
  // getArray<X>() and getWideNums() return the elements as their C++ type X
  // or "widened" (promoted) to type WideNum, as an ArrayBuffer which either has
  // an array reference to the underlying data, if it happens to be of that type
  // and contiguous and not transformed, or owns a copy of the contiguous and
  // transformed data. The caller can call the ref() method on the ArrayBuffer
  // to access the data as an ArrayRef while it is in scope.
  //
  // readArray<X>(dst) and readWideNums(dst) copy the elements into the dst
  // array.
  //
  // For example, if data, indices, and updates are of type
  // DisposableElementsAttr then ScatterNDElementsAttr(data, indices, updates)
  // can be implemented with:
  //
  //   return ElementsAttrBuilder(ctx).fromWideNums(data.getType(),
  //     [&](MutableArrayRef<WideNum> dst) {
  //       data.readWideNums(dst);
  //       ArrayBuffer<WideNum> indicesData = indices.getWideNums();
  //       ArrayBuffer<WideNum> updatesData = updates.getWideNums();
  //
  //       // Write updatesData.get() into dst according to indicesData.get()...
  //
  //       // ArrayBuffers end of lexical scope: any data copies are freed.
  //     });
  //===--------------------------------------------------------------------===//
```

src/Dialect/ONNX/DisposableElementsAttributeStorage.hpp
sorenlassen: done

src/Dialect/ONNX/DisposablePool.hpp
sorenlassen: no need as there is one obj for the duration of the compilation unit

and I don't think it's necessary to insist that all instances are scrubbed or garbage collected, I think there could be valid use cases for that

src/Dialect/ONNX/ElementsAttrBuilder.hpp
sorenlassen: I changed it to have only one constructor

src/Support/Strides.hpp
sorenlassen: added

src/Dialect/ONNX/DisposablePool.hpp
sorenlassen: I added the sentence:
```
This is irreversible and is called when we are done transforming the ONNX dialect, just before we lower it.
```

sorenlassen:
> 1. How about postponing the FLOAT16 and BFLOAT16 support to another patch? That makes the patch clean and I believe it's not essential  in this patch.

This was separated out into PR #1926

> 2. I would hide all `BType`, `WideNum`, `Array`, etc. inside `DisposableElementsAttr` or `ElementsAttrBuilder`. I don't think we will use them outside the purpose of doing constant folding with the biggest precision. Or, we can have a single folder for all stuffs related to `DisposableElementsAttr` and `ElementsAttrBuilder`.

I have addressed this suggestion partially:

I have refactored things so that BType and strides no longer appears in the interface (public methods) of DisposableElementsAttr and ElementsAttrBuilder, and the logic in ConstProp.cpp no longer uses BType and strides.

I also refactored ElementsAttrBuilder so that its methods take ElementsAttr args and return ElementsAttr results, making it transparent that it uses DisposableElementsAttr under the hood.

WideNum is still part of the interface and is used in ConstProp.cpp, because this is how we can work with ElementsAttr elements without writing code for each cpp type (bool, int8, uint8, bfloat16, etc) and without using the Attribute/APFloat/APInt types which are inefficient and hard to work with (e.g. you need separate code paths for float and int).

I like the idea of putting it all in a folder to separate the set of classes that implement all the (Disposable)ElementsAttr logic. I will do that in a follow-up change. I added a TODO in the PR description.

> 3. With (2), we move all primitive constant computations in ConstProp.cpp into `ElementsAttrBuilder`, similar to what we are doing with DialectBuilder. For example,
> 
> ```mlir
> DisposableElementsAttr lhs = getConstValueAsDisposableElements(elementsBuilder, lhsValue);
> DisposableElementsAttr rhs = getConstValueAsDisposableElements(elementsBuilder, rhsValue);
> ElementsAttr resultElements = elementsBuilder.binaryOp<ElementwiseBinaryOp>(lhs, rhs, replacingType)
> ```
> 
> where `binaryOp` will be in `ElementsAttrBuilder` and BType, WideNum are hidden.
> 
> In other words, users use only `ElementsAttrBuilder` to create `DisposableElementsAttr` and call functions with `DisposableElementsAttr` as inputs and outputs.

I like your suggestion.

I should then move the logic in ConstPropSplitPatternCommon, ScatterNDImpl, ConstPropSliceImpl, ConstPropConcatImpl, etc from ConstProp.cpp into new split(), scatterND(), slice(), concat() methods in ElementsAttrBuilder, right?

I will go ahead and do that in a follow-up change.

However, I would like to insulate ElementsAttrBuilder from ONNX ops themselves, and therefore I would like to keep the mapping of elementwise ops to c++ functions with ElementWiseBinaryOpImpl and ElementWiseUnaryOpImpl in ConstProp.cpp and pass the mapped c++ functions to ElementsAttrBuilder. I made a new getWideNumWrappedTemplateFunction helper that concisely packages the ElementWiseBinaryOpImpl and ElementWiseUnaryOpImpl specializations into functions that take WideNum args and returns WIdeNums. That way there's a minimal amount of glue logic between ElementWiseUnary/BinaryOpImpl and ElementsAttrBuilder. I hope you're ok with that approach.

> 4. Please update the document [here](https://github.com/onnx/onnx-mlir/blob/main/docs/ConstPropagationPass.md) about how to use the new interface.

I haven't done this yet but will do this later this week. I added a TODO in the PR description.

sorenlassen:
> @sorenlassen let us know when you want us to review it.

Please give it another look.

I made a number of changes in response to the review comments from last month. It has generated some extra TODOs listed at the end of the PR description, which I will address in the coming days, but it will be helpful if you can do another pass and tell me if you agree on the set of TODOs.

sorenlassen:
@tungld @AlexandreEichenberger I moved DisposableElementsAttr, ElementsAttrBuilder, and supporting source files (DIsposablePool, WideNum, Strides, etc) into a separate folder Dialect/ONNX/ElementsAttr.

It was an interesting puzzle how to avoid a circular dependency between Dialect/ONNX and Dialect/ONNX/ElementsAttr. I solved it by introducing a wrapper class OnnxElementsAttrBuilder (in the spirit of OnnxBuilder) which uses ONNXDialect to retrieve the DisposablePool constructor argument to ElementsAttrBuilder. Moreover, DisposableGarbageCollector and ScrubDisposablePass call DisposablePool with a dictionary of which ops and attributes to scan for DisposableElementsAttr instances in a module. In this way DisposablePool doesn't know the onnx ops (ONNXConstantOp and ONNXConstantOfShapeOp) and therefore has no dependency on Dialect/ONNX. The one-way dependency between Dialect/ONNX and Dialect/ONNX/ElementsAttr should be clear from the make files.

sorenlassen:
@tungld @AlexandreEichenberger @chentong319 I updated the documentation [here](https://github.com/sorenlassen/onnx-mlir-einsum/blob/sorenlassen/disposable/docs/ConstPropagationPass.md).

I propose that we now land the PR in its current form. I'm open to more suggestions, of course, but will prefer to do further changes as follow ups. At the end of the PR description I listed some follow up work I plan to do to address some TODOs that are flagged in the code and documentation.

As you know, I've been testing and refining the implementation for months by now and I believe it's mature and production ready and I would like to put it to use at Groq.

tungld:
>I should then move the logic in ConstPropSplitPatternCommon, ScatterNDImpl, ConstPropSliceImpl, ConstPropConcatImpl, etc from ConstProp.cpp into new split(), scatterND(), slice(), concat() methods in ElementsAttrBuilder, right?
I will go ahead and do that in a follow-up change.

Yes, it's what I meant.

> In this way DisposablePool doesn't know the onnx ops (ONNXConstantOp and ONNXConstantOfShapeOp) and therefore has no dependency on Dialect/ONNX.

Great. I also think `DisposablePool` should work with other ops since it is called after each pass.

tungld (approved):
LGTM! 

My requests were reflected well. Though there are small some remaining items like moving functions from ConstProp.cpp to ElementsBuilder, they can be done in another PR. In general, this patch is big enough as a step towards the new way of constant propagation.
Really appreciate your great work over the months! Thank you very much!

Please wait for responses from @AlexandreEichenberger and @chentong319 before merging this.

sorenlassen:
> Please wait for responses from @AlexandreEichenberger and @chentong319 before merging this.

@AlexandreEichenberger and @chentong319 as discussed in the meeting today, please approve this PR

chentong319 (commented):
Is the DisposableElementsAttr used from Builder? I saw the pass to convert DisposableElementsAttr to DenseElementsAttr. Is there a pass to do the reversal?

sorenlassen:
> Is the DisposableElementsAttr used from Builder?

yes, if you mean the FrontendDialect classes

FrontendDialectHelper.cpp parses tensor constants into DisposableElementsAttr (and external_data is mmapped in DisposableElementsAttr's underlying MemoryBuffer)

> I saw the pass to convert DisposableElementsAttr to DenseElementsAttr. Is there a pass to do the reversal?

no, it's only one way

DisposableElementsAttr is used throughout the onnx passes and then removed before lowering to Krnl or other lowering targets

chentong319 (commented):
1. Is constant folding (with ONNX Dialect) included in this PR?
2. We may have to rethink how to handle the onnx ops constructed while lowering to krnl (not for this PR). The unrealized conversion should be handled by canonicalization. We may have to integrate the lowering to krnl with the ONNX to ONNX transformation. Otherwise, it always may be possible that we miss the optimization opportunity on the newly constructed ops.

sorenlassen:
> Is constant folding (with ONNX Dialect) included in this PR?

no, this will be a follow-up

last month I drafted how they fit together in draft PR #1902

> We may have to rethink how to handle the onnx ops constructed while lowering to krnl (not for this PR). The unrealized conversion should be handled by canonicalization. We may have to integrate the lowering to krnl with the ONNX to ONNX transformation. Otherwise, it always may be possible that we miss the optimization opportunity on the newly constructed ops.

right

fwiw, this was partially addressed in PR #1954 where we "see through" unrealized conversions to do the constant folding in ONNXToKrnlCommon

chentong319 (approved):
LGTM!

sorenlassen:
thank you all for the review and approvals

@AlexandreEichenberger unless you protest I will merge this PR later today

AlexandreEichenberger (approved):
LGTM, thanks

Issue #1879, opened 2022-11-28
Ingesting RandomNormalOps is broken


philass:
For the following program 
```python3
import onnx
from onnx import helper
from onnx import TensorProto
import numpy as np


def make_constant_of_shape_model(input_shape, input_val=None, dtype=np.float32):

    onnx_dtype = onnx.mapping.NP_TYPE_TO_TENSOR_TYPE[np.dtype(dtype)]
    size = input_shape
    X = helper.make_tensor_value_info("X", onnx_dtype, size)
    Z = helper.make_tensor_value_info("Z", onnx_dtype, size)
    random_node = helper.make_node(
        "RandomNormal", [], ["RN"], dtype=1, seed=0.0, shape=input_shape
    )
    add_node = helper.make_node("Add", ["RN", "X"], ["Z"])

    graph_def = helper.make_graph(
        [random_node, add_node],
        "RandomNormalModel",
        [X],
        [Z],
    )
    model = helper.make_model(graph_def, producer_name="onnx")
    model = onnx.shape_inference.infer_shapes(model)
    onnx.checker.check_model(model)

    onnx.save(model, "program.onnx")
    print(model)

def main():
    make_constant_of_shape_model((32, 23), dtype=np.float32)
    # make_constant_of_shape_model((9,), dtype=np.int32)
    # make_constant_of_shape_model(input_shape=(5, 7, 9))  # Test for no optional value


if __name__ == "__main__":

    np.random.RandomState(seed=37)
    main()
```

Running the following
```bash
./Debug/bin/onnx-mlir program.onnx --mlir-print-ir-before-all
```
gives
```
// -----// IR Dump Before (anonymous namespace)::DecomposeONNXToONNXPass (decompose-onnx) //----- //
"func.func"() ({
^bb0(%arg0: tensor<32x23xf32>):
  %0 = "onnx.RandomNormal"() {dtype = 1 : si64, mean = 0.000000e+00 : f32, scale = 1.000000e+00 : f32, seed = 0.000000e+00 : f32, shape = [32, 23]} : () -> none
  %1 = "onnx.Add"(%0, %arg0) : (none, tensor<32x23xf32>) -> tensor<32x23xf32>
  "func.return"(%1) : (tensor<32x23xf32>) -> ()
}) {function_type = (tensor<32x23xf32>) -> tensor<32x23xf32>, input_names = ["X"], output_names = ["Z"], sym_name = "main_graph"} : () -> ()

[...]

PR #1880, opened 2022-11-28
Fix ingestion of RandomNormal

philass:
This addresses https://github.com/onnx/onnx-mlir/issues/1879


philass:
hey @sorenlassen,

Can you help explain to me, the best way to test frontend ingestion (when you have time). I want to use the onnx model generated from the example file in https://github.com/onnx/onnx-mlir/issues/1879.

sorenlassen:
> Can you help explain to me, the best way to test frontend ingestion (when you have time). I want to use the onnx model generated from the example file in #1879.

You can create a .json parse lit test in test/mlir/onnx/parse and add the python script that constructs the .onnx model in utils/testing. One example is test/mlir/onnx/parse/zipmap.json and utils/testing/zipmap.py (the .json test is unusual because it tests failure, you can also look at nonraw_data.json but the associated python script utils/testing/x_data.py is more involved).

chentong319 (approved):
LGTM!

philass:
Thanks for the pointer @sorenlassen!

And thanks for the review @chentong319

utils/gen_onnx_mlir.py
```diff
@@ -421,7 +421,22 @@
...
+      if (dtype() == 0) {
+        resultTypes.push_back(mlir::UnrankedTensorType::get(
+            FloatType::getF16(getContext())));
+      } else if (dtype() == 1) {
+        resultTypes.push_back(mlir::UnrankedTensorType::get(
+            FloatType::getF32(getContext())));
+      } else if (dtype() == 2) {
+        resultTypes.push_back(mlir::UnrankedTensorType::get(
+            FloatType::getF64(getContext())));
+      }
```
sorenlassen, line 436: I suggest you fail if dtype > 2
```
  else llvm_unreachable(...);
```
otherwise

utils/gen_onnx_mlir.py
```diff
@@ -421,7 +421,22 @@
...
+    }
+    resultTypes.push_back(mlir::UnrankedTensorType::get(
```
sorenlassen, line 438: you have to put an `else` before this so you don't push two result types when the dtype attribute is set, right?

sorenlassen (changes requested):

utils/gen_onnx_mlir.py
philass: Good point!

src/Dialect/ONNX/ONNXOps.td.inc
```diff
@@ -4427,6 +4427,27 @@ def ONNXRandomNormalOp:ONNX_Op<"RandomNormal",
     static std::vector<int> getTypeMap() {
       return {-1};
     }
+    std::vector<mlir::Type> resultTypeInference() {
+      std::vector<mlir::Type> resultTypes;
+      if (auto attr = dtypeAttr()) {
```
sorenlassen, line 4432: two nits (sorry I didn't see them when I made the earlier comments):

1. I think it's clearer to say `if (dtype().has_value())` because dtype() returns Optional<int64_t> and dtypeAttr() returns an IntegerAttr, and it's a little bit confusing to introduce the attr variable when it's not used

2. lines 4433-4448 should be indented (it's curious clang-format didn't catch this)

sorenlassen (approved):

src/Dialect/ONNX/ONNXOps.td.inc
tungld: > lines 4433-4448 should be indented (it's curious clang-format didn't catch this)

clang-format does not scan `.inc` files I guess.

sorenlassen: good point

please ignore my comment about indentation, it doesn't matter in generated code

I should have commented on the generating code in gen_onnx_mlir.py (where my 1st nit still applies)

PR #1883, opened 2022-11-29
Shape helper reorg: broadcast, shape, and expand 

AlexandreEichenberger:
Added new ops to the new shape helper scheme. Refined the interface. Now all shape helper are created using a common constructor
```C++
  /* Constructor for shape inference.

   This class and its specialized subclasses are used in one of two situation:
   1) For shape analysis (where no code is generated) during shape inference.
   2) For shape generation (where runtime shapes are computed using generated
   code) during lowering.

   @param op Operation to be analyzed.

   @param operands Operands of the operation to be analyzed. When passing an
   empty list, the operands are taken from the operations. When passing an
   explicit, non-empty list, these operands are used instead of the ones from
   the operation.

   The former option (empty list) is used during shape inference.

   The later option (explicit list) is used during lowering, as typically there
   are two sets of operands, the ones already lowered (explicit list) and the
   original ones (contained within the op). Using shapes during lowering
   typically deals with already-lowered operands.

   @param ieBuilder Class that scans the operands to gather IndexExpr from them.
   Typically used to gather shape and constant values.

   During shape inference, we typically use IndexExprBuilderForAnalysis, which
   uses questionmark for values unkown at compile time.

   During lowering, we typically use and Index Expr Builder that generates code
   for values unknown at compile time. Example of such subclasses are
   IndexExprBuilderForKrnl (generates Krnl ops) or IndexExprBuilderForMhlo
   (generates Shape/MHLO ops).

   However, during lowering, it may be sometime advantageous to perform the
   analysis of the index expressions in the "old" dialect, e.g. in ONNX instead
   of the destination dialect. To enable this, just pass `{}` as operands and
   the original operands associated with the unmodified operation will be used.

   @param scope Index expression scope to be used. If none is provided, a new
   scope is created and stored internally. This scope will then be destructed
   when the current object is destructed.

   */
  NewONNXOpShapeHelper(mlir::Operation *op,
      mlir::ArrayRef<mlir::Value> operands, IndexExprBuilder *ieBuilder,
      IndexExprScope *scope);
```

The class structure was also simplified, and a IndexExpr builder for the MHLO dialect was added.


tungld (approved):
LGTM! Thanks for the great work!

Issue #1876, opened 2022-11-25
Behavior of softmax when axis is set


maxbartel:
Hi!

When converting softmax from onnx to the kernel dialect without setting the `axis` attribute 
```mlir
func.func @test_softmax(%arg0: tensor<13x21x3xf32>) -> tensor<13x21x3xf32> {
  %1 = "onnx.Softmax"(%arg0) : (tensor<13x21x3xf32>) -> tensor<13x21x3xf32>
  func.return %1 : tensor<13x21x3xf32>
}
```
the loops are as expected, so:
```mlir
// Outerloop
krnl.iterate(%0#0, %0#1) with (%0#0 -> %arg1 = 0 to 13, %0#1 -> %arg2 = 0 to 21)
// Innerloops 
krnl.iterate(%2) with (%2 -> %arg3 = 0 to 3)
```

But when setting the `axis` attribute, instead of switching just the axis used for the innerloop against the other axis, all axes after the attribute are moved from the outerloop to the innerloop. For example:
```mlir
func.func @test_softmax(%arg0: tensor<13x21x3xf32>) -> tensor<13x21x3xf32> {
  %1 = "onnx.Softmax"(%arg0) {axis = 0 : si64} : (tensor<13x21x3xf32>) -> tensor<13x21x3xf32>
  func.return %1 : tensor<13x21x3xf32>
}
```
leads to 
```mlir
// Outerloop
// Not there anymore
// Innerloops 
krnl.iterate(%2#0, %2#1, %2#2) with (%2#0 -> %arg1 = 0 to 13, %2#1 -> %arg2 = 0 to 21, %2#2 -> %arg3 = 0 to 3)
```

Another example: 
From
```mlir
func.func @test_softmax(%arg0: tensor<13x21x3xf32>) -> tensor<13x21x3xf32> {
  %2 = "onnx.Softmax"(%arg0) {axis = 1 : si64} : (tensor<13x21x3xf32>) -> tensor<13x21x3xf32>
  func.return %2 : tensor<13x21x3xf32>
}
```
to
```mlir
// Outerloop
krnl.iterate(%0) with (%0 -> %arg1 = 0 to 13)
// Innerloops 
krnl.iterate(%2#0, %2#1) with (%2#0 -> %arg2 = 0 to 21, %2#1 -> %arg3 = 0 to 3)
```

[...]

PR #1884, opened 2022-11-30
feat: split softmax into softmaxV11 and softmax

maxbartel:
@tungld As discussed in https://github.com/onnx/onnx-mlir/issues/1876 this patch splits the `SoftmaxOp` into `SoftmaxOp `and `SoftmaxV11Op`. Doing this also has the benefit that the special handling in the `FrontendDialectTransformer` could be deleted. I also adapted the ONNXToKrnl conversion and the lit test to work with the new implementation. Since this is my first solo contribution:

> I hereby confirm that I abided by the DCO requirements.


tungld:

tungld (approved):
LGTM! Congrats on your first contribution and looking forward to next ones. Thanks!

tungld:

tungld:
@maxbartel it's ready to merge, but the patch is out-of-date with the base branch. Usually, we can merge for you, but somehow I cannot do it with PRs from `Xilinx` github account, perhaps your local repos has some settings that prevent us to do that. So if the situation does not change, we always need to wait for you updating with the base branch, which slows down the merging.

Anyway, please update the patch with the base branch. Then I will merge it if there is no other PR comes in in advance. Otherwise, we need to update again.

tungld:

maxbartel:
@tungld thank you! I will see if it is a setting on our side and if we can change it.

PR #1891, opened 2022-12-02
Experimental: convert onnx to linalg

chentong319:
This PR is intended to experiment how to convert onnx ops to Linalg ops, mixed with krnl ops.
Only onnx.MatMulOp is selected to be converted to Linalg.MatmulOp. Currently, the conversion has to be enabled with command flag `enableLinalg`. 
Since the Linalg uses tensor dialect and bufferization dialect, the change also includes some passes to lower these dialect, as well as Linalg dialect. 
The status of this PR: missing the lowering of bufferization.alloc_tensor.


src/Compiler/CompilerPasses.cpp
```diff
@@ -121,17 +122,33 @@ void addONNXToKrnlPasses(mlir::PassManager &pm, int optLevel, bool enableCSE,
   if (enableInstrumentONNXSignature)
     pm.addNestedPass<func::FuncOp>(
         onnx_mlir::createInstrumentONNXSignaturePass());
+  if (enableLinalg) {
+    pm.addPass(onnx_mlir::createLowerONNXToLinalgPass());
+    // Linalg bufferization can be before or after LowerToKrnlPass
+    // However, sice bufferization::AllocTensorOp is in LowerToKrnl temporarily,
+    // these passes have to be called before LowerToKrnlPass
+    pm.addNestedPass<func::FuncOp>(createLinalgBufferizePass());
```
MikeHolman, line 130: Using `OneShotBufferization` is the recommended way to do bufferization now rather than calling individual bufferize passes.

src/Compiler/CompilerPasses.cpp
```diff
@@ -121,17 +122,33 @@ void addONNXToKrnlPasses(mlir::PassManager &pm, int optLevel, bool enableCSE,
...
+    pm.addNestedPass<func::FuncOp>(
+        bufferization::createEmptyTensorToAllocTensorPass());
+  }
   pm.addPass(onnx_mlir::createLowerToKrnlPass(optLevel, enableParallel));
   // An additional pass of canonicalization is helpful because lowering
   // from ONNX dialect to Standard dialect exposes additional canonicalization
   // opportunities.
+
+  // For Linalg and Krnl mixed IR:
+  // Canonicalization pass will clean up bufferization::to_tensor and to_memref
```
MikeHolman, line 141: We needed to add a new pass to accomplish eliminate type conversions, and looking at your IR output it seems you still have bufferization::to_memref showing up.

Strictly speaking, I don't know if such a pass should be necessary, and in your case it may be because you aren't doing one shot bufferization, but if that doesn't clean things up, you might want to look at adding a pass for cleaning up type conversions.

src/Conversion/ONNXToLinalg/Math/MatMul.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+    // ToFix: dimension size is assumed to be static
+    SmallVector<Value> newDynamicSizes;
+    auto outV = rewriter.create<tensor::EmptyOp>(loc, outputType.getShape(),
+        outputType.getElementType(), newDynamicSizes);
+
+    SmallVector<Value, 1> outputs;
+    outputs.emplace_back(outV);
+    auto newOp =
+        rewriter.create<linalg::MatmulOp>(loc, outputType, operands, outputs);
```
MikeHolman, line 44: You may be aware, but linalg::MatMulOp only works for 2D MatMul, and you will need to call different MatMul ops depending on tensor shape.

MikeHolman (commented):
Overall, change is looking good! I'm excited to see this come together!

src/Compiler/CompilerPasses.cpp
chentong319: In this particular test case, the to_tensor/to_memref are cleaned up by canonicalization pass. Since ONNX has data types other than tensor, such as Sequence and Map type, I think we may have to convert to_tensor/to_memref to UnrealizedConversionCastOp so that they can be cancelled out by canonicalization. The cast Op can handle any type.

src/Compiler/CompilerPasses.cpp
chentong319: I just tried the OneShotBufferization and got the following error:
```
loc("run-matmul/matmul-dyn.mlir":4:8): error: op was not bufferized
// -----// IR Dump After OneShotBufferize Failed (one-shot-bufferize) //----- //
```
Could you tell me how to use the OneShotBufferization? I add the pass before lowering to Krnl with this line of code:
```
pm.addPass(mlir::bufferization::createOneShotBufferizePass());
```

src/Conversion/ONNXToLinalg/Math/MatMul.cpp
chentong319: You are right. I need to check the shape of tensor for linalg.matmul. Other cases can be either lowered to Krnl, or linalg.generic.

PR #1893, opened 2022-12-02
Shapehelper reorg: Pad, OneHot, and RoiAlign

AlexandreEichenberger:
Added these operations (Pad, OneHot, and RoiAlign),

Also, ONNX has input operands that are use in shape computations that can be float, see OneHot depth param for example. So I added support for converting any input types into integer in IndexExprBuilder class to support this. I did not add this support for attributes... hopefully not needed, but will do if I see that pattern too.


tungld (approved):
LGTM.

PR #1894, opened 2022-12-03
Shapehelper reorg: add ArgMin/Max, CategoryMapper, Clip, Compress

AlexandreEichenberger:
next installment of ops being refactored to use the new shape helper scheme


tungld (approved):

tungld:
LGTM.

PR #1900, opened 2022-12-06
normalize value_float/int/string to rank zero scalars

sorenlassen:
The normalization of value_float and value_int onnx constant ops to "rank 1 scalars" (with shape {1}) is causing problems for constant folding (see PR #1799 and PR #1874). I think it was wrong and we should make these rank 0 scalars (with empty shape {}), as implemented in this PR.

### DIFF
--- a/test/mlir/onnx/onnx_canonicalization.mlir
+++ b/test/mlir/onnx/onnx_canonicalization.mlir
@@ -610,7 +610,7 @@ func.func @test_constant_1() -> tensor<i64> {
   %0 = "onnx.Constant"() {value_int = 1 : si64} : () -> tensor<i64>
   return %0 : tensor<i64>
 // CHECK-LABEL:       func @test_constant_1
-// CHECK:           [[VAR_0:%.+]] = "onnx.Constant"() {value = dense<1> : tensor<1xi64>} : () -> tensor<i64>
+// CHECK:           [[VAR_0:%.+]] = "onnx.Constant"() {value = dense<1> : tensor<i64>} : () -> tensor<i64>
 // CHECK:           return [[VAR_0]] : tensor<i64>
 }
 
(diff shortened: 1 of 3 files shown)


test/mlir/onnx/onnx_canonicalization.mlir
```diff
@@ -621,7 +621,7 @@ func.func @test_constant_2() -> tensor<f32> {
   %0 = "onnx.Constant"() {value_float = 2.0 : f32 } : () -> tensor<f32>
   return %0 : tensor<f32>
 // CHECK-LABEL:     func @test_constant_2 
-// CHECK: [[VAR_0:%.+]] = "onnx.Constant"() {value = dense<2.000000e+00> : tensor<1xf32>} : () -> tensor<f32>
+// CHECK: [[VAR_0:%.+]] = "onnx.Constant"() {value = dense<2.000000e+00> : tensor<f32>} : () -> tensor<f32>
```
tungld, line 624: It makes sense to use `tensor<f32>` here. Totally agree. I am just curious is this the following change (especially, replacing `dims` by `{1}`) made the difference?
```
SmallVector<int64_t, 1> dims(1, 1);
SmallVector<float, 1> values(1, attr.getValue().convertToFloat());
auto tensorType = RankedTensorType::get(dims, elementType);
return DenseElementsAttr::get(tensorType, makeArrayRef(values));
```
to
```
auto tensorType = RankedTensorType::get({1}, elementType);
return DenseElementsAttr::get(tensorType, {attr.getValue()});
```

As I understand `auto tensorType = RankedTensorType::get({1}, elementType);` will produce `tensor<1xf32>` and `auto tensorType = RankedTensorType::get({}, elementType);` will produce `tensor<f32>`.

chentong319: The history of the onnxConstantOp. At the beginning, there are only two possible attribute: value and sparse_value. When the other attributes are added, I just converted them into dense attribute so that the other part of compiler does not need to change. Should we do the conversion or just keep keep the attribute?

sorenlassen: > Should we do the conversion or just keep the attribute?

We probably make the assumption in many places that onnx constant ops always have a DenseElementsAttr, so it could be a big change to keep other attributes.

I think it's outside the scope of this PR so I suggest we land this PR without answering that question here.

tungld: Got it. Thanks for pointing it out!

tungld (approved):
LGTM!

PR #1901, opened 2022-12-06
Shapehelper reorg (LRN, Gather*, reshape)

AlexandreEichenberger:
Next installation of revised shape inference


src/Conversion/KrnlToLLVM/ConvertKrnlToLLVM.cpp
```diff
@@ -422,7 +422,7 @@ void ConvertKrnlToLLVMPass::runOnOperation() {
   LLVMTypeConverter typeConverter(ctx, options);
   customizeTypeConverter(typeConverter);
 
-#if 0
+#if 0 // Should that be removed?
```
tungld, line 425: I think so, the above `customizeTypeConverter' covers this.

tungld (approved):
LGTM.

PR #1903, opened 2022-12-07
Shapehelper reorg for ReverseSequence, SpaceToDepth, Split*, Squeeze*

AlexandreEichenberger:
No major changes, significantly rewrote Squeeze for which the logic was a bit convoluted. 

Established a new pattern for code reuse for ops with multiple versions (e.g. Split and Squeeze that have 2 versions, v11 and current). The code pattern that I propose use fewer templated functions and use instead a common ShapeHelper subclass that is templated. Common code is typically `commonComputeShape` which can take many additional parameters to handle the variated numbers of cases opened by the versions. Then  the overloaded `computeShape` can simply gather these additional parameters and then call the common code. This seems to work well.


tungld (approved):
LGTM!

PR #1910, opened 2022-12-09
[NNPA] Use LayoutTransform to transform zTensor

tungld:
This patch makes `onnx.LayoutTransform` work with zTensor so that it can transform a zTensor of a layout to a zTensor of another layout without unstickify and stickify data.

Some important changes:
- the attribute `target_layout` of `onnx.LayoutTransform` was changed from `StringAttr` to generic `Attribute`. By this way `onnx.LayoutTransform` accepts different layout attributes defined by different dialects, e.g. `NCHW4C` in onnx dialect, `4DS` in zhigh dialect.
- Mnemonic `encoding` was change to`layout` to make it easy to understand, e.g. `#onnx.encoding<{dataLayout = "NCHW4C"}>>` becomes `#onnx.layout<{dataLayout = "NCHW4C"}>`.
- a lit test that makes `onnx.LayoutTransform` be a no-op when the same layout is used was added.
- a rewriting rule to replace the sequence of two ops `zhigh.unstick -> zhigh.stick` by `onnx.LayoutTransform` was added.
- `AttrBase.td` was added to a new folder `src/IR` to define the base Attribute for layout. The folder IR is inspired by MLIR where its IR folder contains builtin definitions for Attribute, Type, and Op, those are not specific to a dialect.


AlexandreEichenberger (approved):
LGTM, I like that you started going toward the MLIR structure with the IR subdirectory. We should have a non-functional PR that moves files appropriately for the ONNX and Krnl dialect. Current structure was ok at first but has grown a bit unwieldy.

PR #1911, opened 2022-12-09
Fix importer error for visa model

chentong319:
Two fixes:
1. Import of DenseElementsAttr of string type. Unlike other data type, we can not construct the dense attribute directly from ArrayRef of data. Also in protobuf interface, string_data is RepeatedPtrData not RepeatedData.
2. Empty output tensor assumed to be placeholder in multiple outputs case

Not implemented: DenseElementsAttr of string from external or raw data. Need to read more specification to do the data massage. llvm::sys::SwapByte does not support string type.

@negiyas  With this fix, I reached the error of tensor of string for ONNXEqualOp which you reported. If any tensor is allowed for Equal, the model can be imported to onnx-mlir and then ran into unsupported Op: Unique.


AlexandreEichenberger:
@chentong319 so this is a step toward handling the strings?

chentong319:
> @chentong319 so this is a step toward handling the strings?

This PR added support for string DenseElementAttr. Onnx-mlir already has the support for tensor of string. What is missing is that  the string DenseElementAttr is stored in external data or raw data in onnx model.

AlexandreEichenberger (approved):
LGTM, thanks for the long comment explaining the ONNX string

PR #1904, opened 2022-12-07
Add regex to accept version appendix for compilers

maxbartel:
Depending on the installation method, compilers get a version appendix. This adds support for that to the lit tests.

Signed-off-by: Maximilian Bartel <maximilian.bartel@amd.com>

PR #1912, opened 2022-12-09
Fix -o in onnx-mlir-opt

maxbartel:
If `onnx-mlir-opt` was used with `-o` to create an outputfile, the file was always deleted after the run. This fixes it and also adds a test to make sure a file is created.

The implementation and test are adapted from [upstream MLIR](https://github.com/llvm/llvm-project/blob/05ff7606c9d47135ecf5b69e25b1327634f6fa27/mlir/lib/Tools/mlir-opt/MlirOptMain.cpp#L306-L315).

### DIFF
--- a/src/Tools/onnx-mlir-opt/onnx-mlir-opt.cpp
+++ b/src/Tools/onnx-mlir-opt/onnx-mlir-opt.cpp
@@ -189,8 +189,12 @@ int main(int argc, char **argv) {
   }
 
   // TODO(imaihal): Change preloadDialectsInContext to false.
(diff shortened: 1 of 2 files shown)


AlexandreEichenberger:

AlexandreEichenberger (approved):
LGTM

maxbartel:
@AlexandreEichenberger In https://github.com/onnx/onnx-mlir/pull/1904#issuecomment-1342234688 I asked what one has to do, to get the jenkins pipeline automatically started. Could you explain that, please?

PR #1914, opened 2022-12-12
Rewrite softmax v11 to v13 when axis is the last dimension

tungld:
When `axis` in Softmax version 11 is the last dimension, Softmax version 11 is the same as Softmax version 13. This patch adds a rule for such a case.

This patch also re-organizes the list of canonicalization operations in the alphabetical orders without functional changes. 

### DIFF
--- a/src/Dialect/ONNX/Rewrite.cpp
+++ b/src/Dialect/ONNX/Rewrite.cpp
@@ -545,7 +545,16 @@ public:
 // =============================================================================
 /// Register optimization patterns as "canonicalization" patterns.
 /// Add op to OpsWithCanonicalizer in gen_onnx_mlir.py to activate.
+/// Please keep in alphabetical order.
 // =============================================================================
+
+/// on the ONNXBatchNormalizationInferenceModeOp.
+void ONNXBatchNormalizationInferenceModeOp::getCanonicalizationPatterns(
+    RewritePatternSet &results, MLIRContext *context) {
+  results.insert<FuseBatchNormInferenceModeConvPattern>(context);
+  results.insert<RewriteBatchNormInferenceModeConvPattern1>(context);
+  results.insert<RewriteBatchNormInferenceModeConvPattern2>(context);
+}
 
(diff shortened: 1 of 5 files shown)


src/Dialect/ONNX/Rewrite.td
```diff
@@ -933,4 +940,16 @@ def DimOpToConstantPattern: Pat<
    [(DimAtIndexIsConstant $input, $axis)]
 >;
 
+//===----------------------------------------------------------------------===//
+// Canonicalization for ONNXSoftmaxV11 to the latest version
+//===----------------------------------------------------------------------===//
+
+// When axis is the last dimension of the input tensor, the semantics of V11 is
+// the same as that of V13 (the latest version).
+def SoftmaxV11ToLatestPattern: Pat<
```
chentong319, line 949: Latest may become ambiguous in future. The comment is important.

chentong319 (approved):
LGTM!

src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -557,36 +566,146 @@ void ONNXAddOp::getCanonicalizationPatterns(
   results.insert<FuseAddConvNullBiasPattern>(context);
 }
 
+/// on the ONNXCastOp.
+void ONNXCastOp::getCanonicalizationPatterns(
+    RewritePatternSet &result, MLIRContext *context) {
+  result.insert<CastEliminationPattern>(context);
+  result.insert<FuseCastCastPattern>(context);
```
bxing-groq, line 573: This pattern causes a loss of clipping behaviour if you do f32->i8->f32.

Mind if we add a constraint to ensure the clipping behavior is preserved? Or alternative solutinos? 

For vis: @philass @sorenlassen

sorenlassen: I merged PR #2211 to disable the pattern. Issue #2210 tracks the effort to replace it with a version that fuses only casts that are sound to fuse.

PR #1915, opened 2022-12-12
Add conversion of Softmax from ONNX to TOSA as well as utilities

maxbartel:
Softmax does not have a 1:1 conversion to TOSA. To lower it we have to decompose it. The shape inference builder that TensorFlow and Pytorch are using was also added.


tungld:

src/Conversion/ONNXToTOSA/ONNXToTOSALegalizeUtils.hpp
```diff
@@ -0,0 +1,91 @@
...
+      llvm::dyn_cast<mlir::InferShapedTypeOpInterface>(op.getOperation());
+  if (!shapeInterface)
+    return op;
+
+  llvm::SmallVector<mlir::ShapedTypeComponents> returnedShapes;
+  if (shapeInterface
+          .inferReturnTypeComponents(op.getContext(), op.getLoc(),
+              op->getOperands(), op->getAttrDictionary(), op->getRegions(),
+              returnedShapes)
+          .failed())
```
tungld, line 49: Does `inferReturnTypeComponents` generate runtime code for unknown dimensions or just return -1 if unknown?

@AlexandreEichenberger fyi, `InferShapedTypeOpInterface` may be interesting to you. There is a duplicated effort here to infer shape and update shape.

src/Conversion/ONNXToTOSA/ONNXToTOSALegalizeUtils.hpp
```diff
@@ -0,0 +1,91 @@
...
+    return op;
+
+  // We need to use the element type of the existing result type to generate
+  // the new result shaped type. This is because rescale can include a cast to
+  // different bit-width types and does not have a TypeAttr to define the
+  // target type.
```
tungld, line 55: We have a function [updateType](https://github.com/onnx/onnx-mlir/blob/main/src/Dialect/ONNX/ONNXOps/ShapeHelper.hpp#L149)  doing the same job. E.g, `updateType(result, predictedShape, elementType)` will update the type of `result` by using `predictedShape` whenever feasible. Please consider to use it.

tungld (changes requested):

eric-k256:
I'll comment from the TOSA spec perspective where this change looks good. I think @tungld's comments are helpful, and we can bring applicable versions to the other implementations that convert to TOSA.

src/Conversion/ONNXToTOSA/ONNXToTOSALegalizeUtils.hpp
maxbartel: While you are right, we would like to keep this as is to align as much as possible to [tf-to-tosa](https://github.com/tensorflow/tensorflow/blob/0cf4668f0c6e88ebc5cad993e4956199af2b23de/tensorflow/compiler/mlir/tosa/transforms/legalize_utils.h#L144-L198) and [torch-to-tosa](https://github.com/llvm/torch-mlir/blob/af9e8a5e63e2ca8d498d781d9d56f2c2f8362146/include/torch-mlir/Conversion/TorchToTosa/TosaLegalizeUtils.h#L59-L110).

src/Conversion/ONNXToTOSA/ONNXToTOSALegalizeUtils.hpp
maxbartel: It depends on the specific implementation of the interface method for each operation. Most of the time it returns a dynamic shape if it cannot figure out the right one.

tungld:

src/Conversion/ONNXToTOSA/ONNXToTOSALegalizeUtils.hpp
maxbartel: After talking with @ljfitz who had a conversation with @eric-k256, we decided to adopt `updateType` and Eric thinks of using this in TF, too.

tungld: Cool. Thanks for using `updateType`! I hope you find it convenient.

tungld (approved):
LGTM!

ljfitz:
@tungld - it looks like the failure in the github actions is unrelated (`AttributeError: module 'numpy' has no attribute 'object'`). Can you re-trigger the testing?

tungld:
> it looks like the failure in the github actions is unrelated (AttributeError: module 'numpy' has no attribute 'object'). Can you re-trigger the testing?

Yes, I think so.

PR #1917, opened 2022-12-13
Shapehelper reorg: add ShapeHelper for Einsum, OneHotEncoder, isNan, RandomLike, Reductions

AlexandreEichenberger:
Added shape inference ShapeHelper framework for the ops listed above. They used to compute shape inference without the framework.

In addition, removed the `#define` for Pooling ops, using template instead


src/Dialect/ONNX/ONNXOps/Math/ElementwiseBroadcast.cpp
```diff
@@ -28,10 +28,10 @@ using namespace onnx_mlir;
 namespace {
 
 // Verify shape for numpy style broadcasting operators.
-template <class OP, class ADAPTOR>
+template <class OP_TYPE>
```
tungld, line 31: Had the same thought that `class ADAPTOR` was redundant when I was trying to use this function in NNPA.

src/Dialect/ONNX/ONNXOps/Math/ElementwiseBroadcast.cpp
```diff
@@ -28,10 +28,10 @@ using namespace onnx_mlir;
...
 static LogicalResult verifyShapeForBroadcastingOps(
-    OP &op, Type elementType = nullptr) {
-  ADAPTOR operandAdaptor(op);
+    OP_TYPE &op, Type elementType = nullptr) {
```
tungld, line 33: Could we avoid template by using `Operation *op` instead, and move it to a common place, e.g. `ShapeHelper.cpp`? I want to use this function in NNPA code also. It seems `operandAdator` is used to get operands that can be obtained from `op->getOperands()`.

src/Dialect/ONNX/ONNXOps/Math/ElementwiseBroadcast.cpp
AlexandreEichenberger: Yep, I saw this code from someone here, and decided to start using it more

src/Dialect/ONNX/ONNXOps/Math/ElementwiseBroadcast.cpp
AlexandreEichenberger: did it in the latest version

tungld (approved):
LGTM.

PR #1918, opened 2022-12-13
[NFC] Reorganize ZHigh dialect: one cpp file for each operation

tungld:
No funtional change in this patch.

This patch reorganizes the ZHigh dialect folder so that it is aligned with the ONNX dialect folder. In particular,
- Each file is created for an operation, including code for custom builder, verifier, shape inference, and canonicalization rules.
- A new file `ZHighOps/OpHelper.td` is created to define common tablegen functions used in canonicalization.

Next step will be for using the new shape helper for ZHigh operations.


AlexandreEichenberger (approved):
LGTM, thanks for the reorg

PR #1919, opened 2022-12-13
[NNPA] Use the new ONNXOpShapeHelper for ZHigh operations

tungld:
This patch uses the new ONNXOpShapeHelper for ZHigh operations. It contains code in PR #1918 that is expected to be landed first.


src/Accelerators/NNPA/Conversion/ZHighToZLow/ZHighToZLow.cpp
```diff
@@ -979,22 +975,21 @@ struct ZHighToZLowMatMulOpLowering : public ConversionPattern {
...
     ZHighMatMulOpAdaptor operandAdaptor(operands);
 
     MultiDialectBuilder<KrnlBuilder, MathBuilder> create(rewriter, loc);
     IndexExprScope scope(create.krnl);
 
-    ZHighMatMulOpShapeHelper shapeHelper(&matmulOp, &rewriter);
-    LogicalResult shapecomputed = shapeHelper.computeShape(operandAdaptor);
-    assert(succeeded(shapecomputed) && "Could not compute output shape");
+    IndexExprBuilderForKrnl createKrnlIE(rewriter, loc);
+    ZHighMatMulOpShapeHelper shapeHelper(op, operands, &createKrnlIE);
```
AlexandreEichenberger, line 984: note here: if you create an IndexExprScope outside, you should pass it to the shape helper. Or better, don't create one and use the `shapeHelper.getScope()` to get it. As is, you creating an unused `scope` at the top level which is not used as shapeHelper, when not given any in the constructor, will create its own to save its value. Please check here and elsewhere.

src/Accelerators/NNPA/Dialect/ZHigh/ZHighOps/Conv2D/Conv2D.cpp
```diff
@@ -0,0 +1,149 @@
...
+  // X: [B, HI, WI, CI]
+  Value X = operandAdaptor.input();
+  // W: [KH, KW, CI, CO]
+  Value W = operandAdaptor.input_kernel();
+  // Get attributes.
+  ArrayAttr strides = convOp.strides();
+  StringRef paddingType = convOp.padding_type();
+
+  // Get bounds
+  MemRefBoundsIndexCapture XBounds(X);
```
AlexandreEichenberger, line 39: I know this is a bit of a pain, but I would like to discontinue the "Capture" calls, as they don't really abide by the new scheme. You can look into the new code that I wrote, 
```C++
  MemRefBoundsIndexCapture XBounds(X);
  IndexExpr B = XBounds.getDim(0);
```
should be replaced by
```C++
  createIE->getShapeAsDim(X, 0)
```
which will automatically generate the right code for analysis & lowering as it relies on IndexExprBuilders to do the right thing. In addition, I think it is more readable because the new function name tells exactly what is done.

It's a bit of work, sorry about that. I have changed all occurrence of the old interface in the new methods that I migrated, now there are still a few remaining for ops that did not use shape inference in lowering.. I will work on these, it would be much appreciated if you could work on the accel ones. Thanks

AlexandreEichenberger (changes requested):
Requested a few changes to better match the latest pattern, namely the deprecation of "capture" interface for systematic use of the IndexExprBuilder interface.

src/Accelerators/NNPA/Conversion/ZHighToZLow/ZHighToZLow.cpp
tungld: Thanks! I removed the unused `scope`.

src/Accelerators/NNPA/Dialect/ZHigh/ZHighOps/Conv2D/Conv2D.cpp
tungld: I replaced all `MemRefBoundsIndexCapture` by `getShapeAsDims` as you suggested. Now, there is no `MemRefBoundsIndexCapture` in NNPA code. Thanks!

src/Accelerators/NNPA/Dialect/ZHigh/ZHighOps/Conv2D/Conv2D.cpp
```diff
@@ -19,6 +19,71 @@ using namespace onnx_mlir;
...
+  // W: [KH, KW, CI, CO]
+  Value W = operandAdaptor.input_kernel();
+  // Get attributes.
+  ArrayAttr strides = convOp.strides();
+  StringRef paddingType = convOp.padding_type();
+
+  // Get bounds
+  SmallVector<IndexExpr, 4> XDims, WDims;
+  createIE->getShapeAsDims(X, XDims);
+  createIE->getShapeAsDims(W, WDims);
```
AlexandreEichenberger, line 41: that works too, very nice

AlexandreEichenberger (approved):
LGTM, thanks for the changes

AlexandreEichenberger:
@tungld please look into the mac issue, tx

tungld:
> @tungld please look into the mac issue, tx

Looks like this is a common issue that affects all current PRs. The error said "AttributeError: module 'numpy' has no attribute 'object'", which seems not to be caused by this PR's changes

PR #1921, opened 2022-12-14
added string_data lit test

sorenlassen:
and optimized the string import logic by using StringRef instead of StringAttr
and clarified that ONNX string tensors are never external data or raw data

also cleaned up typos and inconsistencies in the x_data.py lit test script


AlexandreEichenberger:
@chentong319 would be great if you could review it, you have more familiarity with the frontend file as I do. Big thanks

chentong319 (approved):
Thanks to clarify that string is not allowed in external data.
LGTM!

PR #1922, opened 2022-12-14
Improve documentation for creating an OMTensor with or without ownership

hamptonm1:
These documentation updates are a result of the slack conversation and questions that Jan Hofmeier had. Please let me know if I am missing anything or need to add/exclude a statement or two. Thank you!!


include/OnnxMlirRuntime.h
```diff
@@ -114,6 +121,16 @@
  * ```
  * Exactly as it should be.
  *
+ * \subsection freeing-tensor-memory Freeing Tensor Memory
+ *
+ *
+ * When creating an OMTensor or OMTensorList with a specified data pointer, a user has the option to
```
cjvolzka, line 127: We should seperate how OMTensor withOwnership works vs OMTensorList. This basically discribes how OMTensor withOwnership works but what is "owned" by on OMTensorList is much more intricate. (See earlier comment)

AlexandreEichenberger:
Just to clarify:

Owning for tensor:
```C++
void omTensorDestroy(OMTensor *tensor) {
...
  if (tensor->_owning) {
    free(tensor->_allocatedPtr);
  }
...
}
```

namely, if the Tensor is destroyed calling this explicit function, and the tensor is "owning", then the data holding the values will be freed.

For tensor lists
```C++
void omTensorListDestroy(OMTensorList *list) {
...
  for (int64_t i = 0; i < list->_size; i++)
    omTensorDestroy(list->_omts[i]);
  if (list->_owning) {
    free(list->_omts);
  }
  free(list);
}
```

No matter what, every tensor will be destroyed. Their own tensor data will be destroyed if they are owning (see above).

Now if the list "owning", then the list of pointer to the tensor will be freed.

# Case 1

The pattern below is safe only because the lifetime of the vector `tensor_vector` span beyond the `omTensorListDestroy(tensor_list)`.

```C++
{
  vector<OMTensor*> tensor_vector;
  tensor_vector.push_back(OMTensorCreate()) // tensor "A"
  tensor_vector.push_back(OMTensorCreate()) // tensor "B"
  tensor_vector.push_back(OMTensorCreate()) // tensor "C"

  OMTensorList* tensor_list = omTensorListCreateWithOwnership(tensor_vector.data(), tensor_ptrs.size(), false);
  omTensorListDestroy(tensor_list)
} // tensor_vector implicitly feed here.
```
Note that above, the ownership MUST BE false, as otherwise the implicit freeing of `tensor_vector` will free memory that was already freed.

# Case 2

The example below is unsafe
```C++
{
  vector<OMTensor*> tensor_vector;
  tensor_vector.push_back(OMTensorCreate()) // tensor "A"
  tensor_vector.push_back(OMTensorCreate()) // tensor "B"
  tensor_vector.push_back(OMTensorCreate()) // tensor "C"

  OMTensorList* tensor_list = omTensorListCreateWithOwnership(tensor_vector.data(), tensor_ptrs.size(), false);
} // tensor_vector implicitly feed here.
  
omTensorListDestroy(tensor_list) 
```
because the `tensor_vector` is implicitly freed before its last use in `omTensorListDestroy(tensor_list)`.

# Case 3

So unless case 1 applies, the only safe way is the following.
```C++
{
  OMTensor** tensor_vector = (OMTensor**) malloc(3*sizeof(OMTensor*));
  tensor_vector.push_back(OMTensorCreate()) // tensor "A"
  tensor_vector.push_back(OMTensorCreate()) // tensor "B"
  tensor_vector.push_back(OMTensorCreate()) // tensor "C"

  OMTensorList* tensor_list = omTensorListCreateWithOwnership(tensor_vector.data(), tensor_ptrs.size(), true);
  omTensorListDestroy(tensor_list)
}
```

Now we did an explicit malloc, and then that malloced data will be freed exactly when the list is destroyed.

# Personal opinion

This interface is terrible. It may have been designed for unique ptrs and/or for the compiler; for users it is very confusing.

1. List ownership is only for a very small data... namely a couple of pointers. In my opinion, the pointers should always be copied into a list owned by the list, so that it does not matter when the user's list is freed or not. 
2. I find that forcing to free each tensor of the list is also bad. For example, we may reuse an input tensor in 2 lists, or reused in the same list. Not having the option of manually free the tensor is super bad in my opinion.

If I were to redesign this interface, I would "append" a list past the data structure of the OMTensorList so that no extra malloc is needed, the list is always copied, and "owning" would refer to the tensor themselves... meaning we call  the individual tensor destroy for tensors in the list only in presence of the owning field.

AlexandreEichenberger:
@hamptonm1 @cjvolzka feel free to contradict me if you believe my understanding of the interface is not correct.

I will then let you asses if the proposed changes of documentation is compliant with our best understanding of the interface.

As for the proposed change, it would change the interface. If no-one actually complain and/or use it, then it's ok not to change it. If you know folks getting confused, then we should probably contemplate changing it... it would be for the best long term.

hamptonm1:
@AlexandreEichenberger Thanks for your in depth feedback, it will take me a bit to process!

cjvolzka:
> feel free to contradict me if you believe my understanding of the interface is not correct.

Unless there's an additional subtly I'm missing, I think we're in agreement on what currently happens. Your Cases more clearly convey some of the intricacies I was attempting to highlight myself.

> As for the proposed change, it would change the interface. If no-one actually complain and/or use it, then it's ok not to change it. If you know folks getting confused, then we should probably contemplate changing it... it would be for the best long term.

I'm actually all for changing this part of the C interface as it's pretty confusing. We're already bumping the version to 0.4 for the next release because of interface change for the Python runtime (ie next release won't be a pure "drop in replacement" from 0.3.2). So I'd be up for including this proposed changed too in the next release. Then we might avoid a future round of disruptive updates when others run into this. Let's chat on this in our next meeting to confirm to figure out logistics.

Until then we can continue with this PR so we better document what's happening currently.

cjvolzka:
> This interface is terrible. It may have been designed for unique ptrs and/or for the compiler; for users it is very confusing.
> 1. List ownership is only for a very small data... namely a couple of pointers. In my opinion, the pointers should always be copied into a list owned by the list, so that it does not matter when the user's list is freed or not.
> 2. I find that forcing to free each tensor of the list is also bad. For example, we may reuse an input tensor in 2 lists, or reused in the same list. Not having the option of manually free the tensor is super bad in my opinion.

For 1, so far the largest I've seen is 11 input tensors. There might be larger but copying at most a couple of dozen pointers seems like a minimal one time cost (per inference) vs all the confusion of the current interface and the headaches we may save.

For 2, I didn't think about input tensor reuse but that's a good point. Another case might be where a model has multiple output tensors and you want one of those output tensors to be fed into model in a pipeline but don't need to keep the rest of the output tensors. Currently I think you'd have to keep the whole output list just to avoid destroying the one you'd want to keep. 

To be honest, I originally assumed OMTensorListWithOwnership controlled whether the OMTensors in the list were cleaned up when the OMTensorList was destroyed. (In my mind at least that seem liked the closest analog to how OMTensorWithOwnership worked on the data buffers they "contained"). I didn't realize I had it wrong until Jan dug into the onnx-mlir code and pointed out the OMTensors were always getting destroyed with the OMTensorList.

So my opinion based on 1. and 2. above...
1. We keep OMTensorListCreateWithOwnership to stay consistent with OMTensor (but change what it does)
2. Follow @AlexandreEichenberger's recommendation related to his bullet point 1. That the OMTensor mallocs it's own internal list that it control and that list is always cleaned up when the OMTensorList is destroyed. (Let me know Alex if that's not what you meant).
3. Change OMTensorListCreateWithOwnership to control if the OMTensors inside the list are destroyed with the List. If `true`, then on OMTensorList destruction, we loop through and destroy the OMTensors. If `false`, then the OMTensor are _not_ destoryed and it'll be up to the application to manually destroy the ones they no longer want.
    * The question that leaves behind is what should the default ownership state be when plain OMTensorListCreate is used. Should it be `true` or `false`? Similarly what should the ownership setting be on the OMTensorList run_main_graph() returns?

AlexandreEichenberger:
FYI, the compiler may uses these interfaces and may rely on the functions with the old functionality.
It can be changed; for example it is possible (I have not checked) that the compiler create the list on the heap and then rely on the ownership to delete that list. Can you check @hamptonm1 ? 
If that is the case, we can either rename the old functions with a suffix "deprecated" and change in the compiler the functions to use these "deprecated" ones, hide them from the header file, and have in the header file the new behavior

Or we can change how the compiler use them so that we only need one version.

AlexandreEichenberger:
@hamptonm1 this PR has been out there for a while, is it ready for review?

AlexandreEichenberger:
if ready, @cjvolzka would you mind to review it? tx

hamptonm1:
> @hamptonm1 this PR has been out there for a while, is it ready for review?

@AlexandreEichenberger It is not ready as of yet, I have not updated it since the initial review. Will update this by the end of the week. And I will re-request everyones review accordingly. Thanks!

docs/mnist_example/README.md
```diff
@@ -195,10 +195,19 @@ int main() {
...
-  // Create a tensor list.
+
+  // Create a tensor using omTensorCreateWithOwnership (returns a pointer to OMTensor).
+  // When the parameter, owning is set to "true", the OMTensor will release the data
+  // pointer upon destruction. If owning is set to false, the data pointer will not be
+  // freed upon destruction.
+  OMTensor *tensor = omTensorCreateWithOwnership(img_data, shape, rank, ONNX_TYPE_FLOAT, /*owning=*/true);
+
+  // Create a tensor list using omTensorListCreateWithOwnership (returns a pointer to the OMTensorList).
+  // When the parameter, owning is set to "true", the OMTensorList will release the data
```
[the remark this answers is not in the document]
hamptonm1: In situations like this it would be best for me to have a meeting to discuss instead of writing in a PR, I process data better via audio. I spoke with Jan and he stated the following  "I think the word data_ptr adds confusion, because it is used everywhere to describe the C arrays that hold the actual data. For the TensorList, it's just a pointer to an array, which contains pointers to the tensors." So I will update the wording to avoid confusion as Jan mentioned.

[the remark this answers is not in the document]
hamptonm1: Updated...sending PR link to Jan as well to see if this is clearer.

include/OnnxMlirRuntime.h
hamptonm1: Okay I separated them.....let me know if this is better. I try to summarize the words @AlexandreEichenberger stated in his examples below.

docs/mnist_example/README.md
```diff
@@ -198,13 +198,13 @@ int main() {
 
   // Create a tensor using omTensorCreateWithOwnership (returns a pointer to OMTensor).
   // When the parameter, owning is set to "true", the OMTensor will release the data
-  // pointer upon destruction. If owning is set to false, the data pointer will not be
+  // pointer upon destruction. If owning is set to false, the tensor array will not be
```
cjvolzka, line 201: A couple of notes:
1. in the same sentence, one part has "data pointer" and another has "tensor array". These should be the same
1. OMTensor itself has a paramater called "data_ptr" so I think either "data_ptr" or "data pointer" is fine here.

hamptonm1: Thanks!

docs/mnist_example/README.md
```diff
@@ -198,13 +198,13 @@ int main() {
...
   // freed upon destruction.
   OMTensor *tensor = omTensorCreateWithOwnership(img_data, shape, rank, ONNX_TYPE_FLOAT, /*owning=*/true);
 
   // Create a tensor list using omTensorListCreateWithOwnership (returns a pointer to the OMTensorList).
   // When the parameter, owning is set to "true", the OMTensorList will release the data
-  // pointer upon destruction. If owning is set to false, the data pointer will not be
+  // pointer upon destruction. If owning is set to false, the tensor array will not be
```
cjvolzka, line 207: A couple of notes:
1. in the same sentence, one part has "data pointer" and another has "tensor array". These should be the same.
1. I think "tensor" makes more sense than "data pointer" but it could be a little more explcit given the complexity of the mechanics. So something like...
```
// When the parameter, "owning" is set to "true", the OMTensorList will release its array of OMTensor pointers 
// upon destruction. If owning is set to "false", the array of OMTensor pointers will not be released. 
// Regardless of "owning", all of the OMTensor objects themselves are released with the OMTensorList
```

include/OnnxMlirRuntime.h
```diff
@@ -124,12 +124,17 @@
...
- * Otherwise, if the flag is set to "true", then destruction of the tensor (omTensorDestroy or omTensorListDestroy)
+ *
+ * In the case of OMTensor, if the ownership flag is set to "false", then a user is responsible for freeing memory until the last use.
+ * Otherwise, if the flag is set to "true", then the destruction of the tensor (omTensorDestroy)
  * will also free any associated memory and data buffers.
  *
+ * With respects to OMTensorList, no matter the value of the ownership flag, every tensor will be destroyed. 
+ * However, if the ownership flag is set to "true", then the tensor data will be destroyed (omTensorListDestroy) and the 
+ * pointer to the tensor list will be freed. Otherwise, if the ownership flag is false, the tensor array is implicitly freed before its last use.
+ *
```
cjvolzka, line 137: * With respects to OMTensorList, no matter the value of the ownership flag, every OMTensor object in the OMTensorList will be destroyed. 
 * If the OMTensorList ownership flag is set to "true", then the array of OMTensor pointers will be destroyed with the OMTensorList. 
  * Otherwise, if the ownership flag is false, the array of OMTensor pointers will not be destroyed with the OMTensorList and must be released separately.

docs/mnist_example/README.md
hamptonm1: Okay this is fixed

docs/mnist_example/README.md
hamptonm1: Fixed

cjvolzka (approved):
LGTM. Thanks!

PR #1923, opened 2022-12-14
Suppress non-floating complex warning on newer MSVC versions

nhat-nguyen:
MSVC recently started issuing warnings when `std::complex` template is instantiated with types other than `float` and `double`, which caused failures in our builds. More details at https://github.com/microsoft/STL/pull/2759.

Unfortunately, the code that triggered the warning is part of upstream `mlir`, and accommodating this change isn't trivial. I propose that we suppress the warning for the time being.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -125,8 +125,8 @@ if (CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
   # Disable warning for non-standard usages of std::complex template which was
   # introduced in MSVC 14.34.
   # More details at https://github.com/microsoft/STL/pull/2759.
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -D_SILENCE_NONFLOATING_COMPLEX_DEPRECATION_WARNING")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_SILENCE_NONFLOATING_COMPLEX_DEPRECATION_WARNING")
-
 endif()
 
 # Suppress warnings in third party code if requested.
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -117,10 +117,26 @@ else()
...
+  # llvm/include/mlir/IR/BuiltinAttributes.h(353): warning C4996:
+  # 'std::complex<llvm::APFloat>::complex': warning STL4037: The effect of
+  # instantiating the template std::complex for any type other than float,
+  # double, or long double is unspecified.
+  # You can define _SILENCE_NONFLOATING_COMPLEX_DEPRECATION_WARNING to suppress
+  # this warning.
+  #
+  # Since the code is part of upstream mlir, suppress the warning for now.
+
+  add_definitions(-D_SILENCE_NONFLOATING_COMPLEX_DEPRECATION_WARNING)
```
sstamenova, line 139: Rather than adding it to definitions, see below for how other warnings are disabled and follow the same pattern.

CMakeLists.txt
```diff
@@ -117,10 +117,26 @@ else()
...
 endif()
 
-# Use the new MSVC preprocessor to improve standard conformance.
-if (CMAKE_CXX_COMPILER_ID MATCHES "MSVC") 
+if (CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
+  # Use the new MSVC preprocessor to improve standard conformance.
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /Zc:preprocessor")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /Zc:preprocessor")
+
+  # Starting from MSVC 14.34, the following warning was introduced:
```
sstamenova, line 125: I don't think we need quite so much context in the comment. I would recommend removing the example and instead linking to the change that added the warning.

sstamenova (changes requested):

CMakeLists.txt
nhat-nguyen: Thanks for the suggestion. Updated.

CMakeLists.txt
nhat-nguyen: Simplified comment. :)

CMakeLists.txt
```diff
@@ -117,10 +117,16 @@ else()
...
-if (CMAKE_CXX_COMPILER_ID MATCHES "MSVC") 
+if (CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
+  # Use the new MSVC preprocessor to improve standard conformance.
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /Zc:preprocessor")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /Zc:preprocessor")
+
+  # Disable warning for non-standard usages of std::complex template which was
+  # introduced in MSVC 14.34.
+  # More details at https://github.com/microsoft/STL/pull/2759.
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_SILENCE_NONFLOATING_COMPLEX_DEPRECATION_WARNING")
```
sstamenova, line 128: Do we need it in C_FLAGS as well?

CMakeLists.txt
```diff
@@ -117,10 +117,16 @@ else()
...
+
```
sstamenova, line 129: Nit: no need for new line

sstamenova:

sstamenova:

sstamenova (approved):
Thanks!

PR #1924, opened 2022-12-14
Fix issue in Gemm Add fusion

philass:
Currently when `beta=0`. This will incorrectly fuse the add op.

`beta` should be set to 1 in the pattern that it is replaced with.

Co-authored-by: Zachary Cetinic <zcetinic@groq.com>


src/Dialect/ONNX/Rewrite.td
```diff
@@ -254,7 +254,7 @@ def MulAddToGemmOptPattern : Pat<(ONNXAddOp (ONNXMatMulOp:$res $m1, $m2), $m3),
 
 // onnx.add(onnx.Gemm(%X, %Y, None), %Z) = onnx.Gemm(%X, %Y, %Z)
```
sorenlassen, line 255: I suggest that you make beta=1.0 clear here, e.g. by writing it as
```
// onnx.add(onnx.Gemm(%X, %Y, None), %Z) = onnx.Gemm(%X, %Y, %Z) {beta=1.0}
```

sorenlassen (approved):
LGTM

thanks for adding lit test

src/Dialect/ONNX/Rewrite.td
philass: Good point! Thanks for the review

tungld (approved):
LGTM.

PR #1925, opened 2022-12-15
Scrub upsample comment

hamptonm1:
I updated the comment from `sanity` to `safety`. Also I did a search in VSCode to see if there are any lingering references to `master` in the `onnx-mlir` repo. Seems like there are only 3 places where `master` is present. Two instances are `PyTorch` links which will break if we use the word `main` and another section that uses the following link: `https://raw.githubusercontent.com/caldwell/renderjson/master/` . This link is actually broken, I get back a 400- Invalid response so I wonder if we can just remove this from the `onnx-mlir/.buildbot/jenkins-check-model-zoo.py`.

### DIFF
--- a/src/Dialect/ONNX/ONNXOps/Tensor/Upsample.cpp
+++ b/src/Dialect/ONNX/ONNXOps/Tensor/Upsample.cpp
@@ -34,7 +34,7 @@ LogicalResult ONNXUpsampleOp::verify() {
   auto inputTy = X().getType().cast<RankedTensorType>();
   int32_t inputRank = inputTy.getShape().size();
 
-  // Sanity checks on scale argument
+  // Safety checks on scale argument
   auto scalesTy = scales().getType().cast<RankedTensorType>();
   if (scalesTy.getShape().size() != 1) {
     return emitError("Scales tensor must be rank-1");


hamptonm1:
@cjvolzka Side point: There are a lot of references to `master` in `dlc-automation`. Maybe we want to scrub that too :)

cjvolzka:
Yeah, there's not much we can do about pybind11 unless we can get them to switch their primary branch name. That might not be impossible, but that's a task for another time.

The rapidjson is the one I hit at a glance and was curious if there were any more or if we could remove it. I took a closer look, that's an interesting case. https://raw.githubusercontent.com/caldwell/renderjson/master/ itself is a dead link. However, the build code has:
```
    urlretrieve(RENDERJSON_URL + RENDERJSON_JS,
                os.path.join(workspace_reportdir, RENDERJSON_JS))
```
That ultimately generates https://raw.githubusercontent.com/caldwell/renderjson/master/renderjson.js which does work. After some poking around, that is the same thing as https://github.com/caldwell/renderjson/blob/master/renderjson.js.

So we likely can't get rid of this 'master' reference without breaking our builds. However we can at least fix it so the RENDERJSON_URL itself is a valid link by using this instead:
```
RENDERJSON_URL             = 'https://github.com/caldwell/renderjson/blob/master/'
```

hamptonm1:
> Yeah, there's not much we can do about pybind11 unless we can get them to switch their primary branch name. That might not be impossible, but that's a task for another time.
> 
> The rapidjson is the one I hit at a glance and was curious if there were any more or if we could remove it. I took a closer look, that's an interesting case. https://raw.githubusercontent.com/caldwell/renderjson/master/ itself is a dead link. However, the build code has:
> 
> ```
>     urlretrieve(RENDERJSON_URL + RENDERJSON_JS,
>                 os.path.join(workspace_reportdir, RENDERJSON_JS))
> ```
> 
> That ultimately generates https://raw.githubusercontent.com/caldwell/renderjson/master/renderjson.js which does work. After some poking around, that is the same thing as https://github.com/caldwell/renderjson/blob/master/renderjson.js.
> 
> So we likely can't get rid of this 'master' reference without breaking our builds. However we can at least fix it so the RENDERJSON_URL itself is a valid link by using this instead:
> 
> ```
> RENDERJSON_URL             = 'https://github.com/caldwell/renderjson/blob/master/'
> ```

Okay let me update that...sounds good to me!

cjvolzka (approved):

PR #1926, opened 2022-12-15
ingest BFLOAT16

sorenlassen:
introduces float_16 and bfloat_16 data types to represent both FLOAT16 and BFLOAT16 as c++ data types, which makes it possible to treat them like the other scalar types in many ways

this will also make it possible to do constant propagation, see PR #1874

this PR has been factored out from PR #1874

### DIFF
--- a/src/Builder/FrontendDialectHelper.cpp
+++ b/src/Builder/FrontendDialectHelper.cpp
@@ -20,6 +20,7 @@
 
 #include "src/Builder/FrontendDialectHelper.hpp"
 #include "src/Dialect/ONNX/ONNXOps/OpHelper.hpp"
+#include "src/Support/FloatingPoint16.hpp"
 
 namespace {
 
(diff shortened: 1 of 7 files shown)


tungld (approved):
LGTM.

sorenlassen:
@tungld  thanks for the review

fyi, I made a follow-up change to remove denseBuilder which is no longer necessary after denseBuilderU16ToF16 was removed (this in turn made the string type comments out of date, so I edited them accordingly and with a view towards the follow-up PR #1921)

tungld:
> fyi, I made a follow-up change to remove denseBuilder which is no longer necessary after denseBuilderU16ToF16 was removed (this in turn made the string type comments out of date, so I edited them accordingly and with a view towards the follow-up PR https://github.com/onnx/onnx-mlir/pull/1921)

I see. That's ok. Thanks!

PR #1927, opened 2022-12-15
Make finding python version more robust

nhat-nguyen:
Starting from cmake 3.15, the Python_FIND_STRATEGY variable was introduced and can take 2 values:

- `LOCATION`: stops lookup as soon as a suitable python version is found. This is the `NEW` behavior for this policy.
- `VERSION`: finds the most recent python version. This is the `OLD` behaviour or when `CMP0094` is undefined.

Using the `NEW` policy will make our builds more robust because the most recent python version installed isn't necessarily compatible.

Signed-off-by: Nhat Nguyen <honguye@microsoft.com>


AlexandreEichenberger:
If its ok with you, by updating the branch it get the latest and run the full tests (since I am an approved user for the Jenkins). You will have to refresh your local branch.

CMakeLists.txt
```diff
@@ -13,6 +13,14 @@ option(ONNX_MLIR_SUPPRESS_THIRD_PARTY_WARNINGS "Suppress warning in third_party
 
 set(CMAKE_CXX_STANDARD 17)
 
+# New in cmake 3.15:
+# https://cmake.org/cmake/help/latest/policy/CMP0094.html
+# Stops python lookup as soon as a version satisfying version constraints
+# is found.
+if (POLICY CMP0094)
+  cmake_policy(SET CMP0094 NEW)
+endif()
```
gongsu832, line 22: If this requires 3.15, the minimum required cmake version (currently 3.13.4) should be bumped up to 3.15.

AlexandreEichenberger: @gongsu832 would it be fine to bump to 3.15? Or are we help back to lower version because of some dependence?

@nhat-nguyen would you be willing to try moving us up to 3.15 if appropriate? Tx

nhat-nguyen: definitely, i can try moving us up to 3.15

gongsu832: Actually we currently use 3.13.4 because that's what llvm uses. There is a comment in llvm that starting from llvm 17 they will require cmake 20.2 and they suggested to move up to 20.2 directly. This would require we move our Ubuntu base docker image from focal to jammy. I will create a PR to do that.

AlexandreEichenberger: @gongsu832 please link the PR so @nhat-nguyen will know when we may proceed with this one, thanks

sstamenova:
There's now a PR to bump the min version to 3.20. See #2064 . This change could land after it completes.

sstamenova (approved):

AlexandreEichenberger:
@gongsu832 is that something that you would want? Or should we close it? Inactive for 2 years now

gongsu832:
We can close. The impact of the change is minimal.

PR #1932, opened 2022-12-20
Preinstall numpy 1.23.5 since 1.24 appears to break onnx

gongsu832:


tungld (approved):
LGTM. Thanks for the quick fix!

PR #1934, opened 2022-12-20
Shapehelper added conv (all but transpose) dropout eyelike, constantofshape, range, resize, dequantize)

AlexandreEichenberger:
Updated above ops, unfortunately there are still more to come


src/Dialect/ONNX/ONNXOps/Quantize/DequantizeLinear.cpp
```diff
@@ -34,6 +34,50 @@ int64_t nonScalar1DLen(ShapedType ty) {
...
+  if (d != -1) {
+    int64_t r = xTy.getRank();
+    int64_t a = operandAdaptor.axis();
+    // Checked in verify:
+    assert(-r <= a && a < r && "axis out of range");
+    if (a < 0)
+      a += r;
+    if (!outputDims[a].isLiteral()) {
+      outputDims[a] = LiteralIndexExpr(d);
+    } else {
```
tungld, line 65: Looks like this `else` is redundant. `assert` can come without `else`.

tungld (approved):
LGTM.

src/Dialect/ONNX/ONNXOps/Quantize/DequantizeLinear.cpp
AlexandreEichenberger: I see, fine I can remove the else.

PR #1938, opened 2022-12-21
Add new opset 18 ops

philass:


AlexandreEichenberger (approved):
LGTM,

is there some what to record somewhere that ops have a shape inference but not yet lowering? It can be found out by searches, it might be useful to have it recorded explicitly somewhere. Probably for some later PR.

Issue #1936, opened 2022-12-20
Known macOS build issues with Protobuf version 3.19 and 3.20


hamptonm1:
When upgrading protobuf from 3.18.3 to 3.20.3, I noticed the macOS build failed with the following:

```
Undefined symbols for architecture x86_64:
[720](https://github.com/onnx/onnx-mlir/actions/runs/3743651637/jobs/6356066681#step:10:721)
  "google::protobuf::internal::InternalMetadata::~InternalMetadata()", referenced from:
[721](https://github.com/onnx/onnx-mlir/actions/runs/3743651637/jobs/6356066681#step:10:722)
      google::protobuf::MessageLite::~MessageLite() in libonnx_proto.a(onnx-ml.pb.cc.o)
[722](https://github.com/onnx/onnx-mlir/actions/runs/3743651637/jobs/6356066681#step:10:723)
      google::protobuf::MessageLite::~MessageLite() in libonnx_proto.a(onnx-data.pb.cc.o)
[723](https://github.com/onnx/onnx-mlir/actions/runs/3743651637/jobs/6356066681#step:10:724)
ld: symbol(s) not found for architecture x86_64
```

This is actually a known issue in the protobuf community, please see the following issues: 
https://github.com/protocolbuffers/protobuf/issues/9947
https://github.com/facebookincubator/velox/issues/2029
 
 
This problem is fixed in protobuf version 3.21.3, however `onnx` community has not upgraded to that version. Instead they are on version 3.20.3 which is where the failure lies. There is a workaround if we still want to upgrade to version 3.20.3 by adding the following to the CMake.txt file:

`add_definitions(-DNDEBUG)`

hamptonm1:
Should we upgrade to protobuf 3.20.3 and include the workaround or should we wait for `onnx` to upgrade to `3.21.3` which fixes the issue and therefore we do not need to upgrade. @AlexandreEichenberger @gongsu832 @chentong319 @tungld ? Any comments about this topic :)

gongsu832:
In general, only upgrade when there is a specific reason to do so. Stay in sync with onnx is desirable but if the new version has known problem like the one above, I'd wait until things get sorted out.

chentong319:
To my understanding, protobuf 3.20.3 is required by onnx opset 18. protobuf 3.21 is needed for Mac to fix the issue with protobuf 3.20.3 on Mac. Can protobuf 3.21.3 work with other components in onnx-mlir? Open a draft PR to try?

[...]

PR #1941, opened 2022-12-23
Upgrade to protobuf 3.20.3

hamptonm1:
I upgraded Protobuf from 3.18.3 to 3.20.3 which matches what the `onnx` community currently has for version `1.13.0`. However, MacOS fails for version 3.20.3 (see this link for the reason: https://github.com/onnx/onnx-mlir/issues/1936). So what I have done based on @chentong319 suggestion, I upgraded everything else to 3.20.3 and kept Mac build on 3.18.3. Please let me know your thoughts.


docker/Dockerfile.llvm-project
```diff
@@ -22,49 +22,49 @@ WORKDIR ${WORK_DIR}
...
+    make -j${NPROC} install && \
+    cd ${WORK_DIR} && rm -rf ninja; \
+    fi \
     && ln -sf /usr/bin/python3 /usr/bin/python
 
 # Install protobuf
-ARG PROTOBUF_VERSION=3.18.3
-RUN git clone -b v${PROTOBUF_VERSION} --recursive https://github.com/google/protobuf.git \
+ARG PROTOBUF_VERSION=3.20.3
+RUN git clone -b v${PROTOBUF_VERSION} --recursive https://github.com/protocolbuffers/protobuf.git \
```
gongsu832, line 67: This should be the only change to Dockerfile.llvm-project. The other messed up indentation changes should be reverted.

gongsu832: I meant the version and url changes should be the only changes needed.

hamptonm1: > I meant the version and url changes should be the only changes needed.

Sure!! Let me change the format

hamptonm1: @gongsu832 How does this look now? Ready to merge this in :)

gongsu832 (approved):
LGTM

utils/install-protobuf.sh
```diff
@@ -1,6 +1,6 @@
 # Check out protobuf v3.18.3
 PROTOBUF_VERSION=3.18.3
```
sstamenova, line 2: So is this only used for Mac?

hamptonm1: @sstamenova Yes!!! Only because Protobuf 3.20.3 failed for Mac, so kept it at the older version.

hamptonm1: @sstamenova I explain why Mac was failing with Protobuf 3.20.3 here: https://github.com/onnx/onnx-mlir/issues/1936

sstamenova: Thanks!

PR #1943, opened 2022-12-28
Opset 18 Reduce ops

philass:
All reduce ops in opset 18 have been modified to match the specification of `ONNXReduceSum`. Namely they add a `noop` attribute, and axes is changed from an attribute to an input.

The simplest approach is to simply rewrite all older versioned `ReduceOps` into their opset 18 counterparts. I have elected to only take this approach with
- `ReduceL1`
- `ReduceL2`
- `ReduceLog`
- `ReduceLogSum`
- `ReduceLogSumSquare`

However I've elected to keep the implementations of the following ops. And rename them to V13. 
- `ReduceMax`
- `ReduceMin`
- `ReduceMean`
- `ReduceProd`


AlexandreEichenberger (commented):
Could you comment why you choose to version some of the ops?

Also, when adding new functionality (i.e. supporting new ops or a new varient of an op), it would be useful if you could explicitly list which ones need further work to lower to Krnl/Affine/...

While it can be derived from a careful reading of the code, it would be useful to have this info explicit so that all of the folks working on their backends for ONNX would be able to better detect what further work is needed.

Does that make sense? Or do you think it's too much work?

tungld:
>  Namely they add a noop attribute, and axes is changed from an attribute to an input.

It looks to me V13 can be totally converted to V18, e.g. by creating a value from the axes attribute.

> However I've elected to keep the implementations of the following ops. And rename them to V13.

Could you explain more about the reason to keep the implementation? I don't see any good reason to keep the shape inference and lowering of V13. Thanks!

AlexandreEichenberger:
I concur, the preferred option is to migrate the old ops to the new one and ensuring that the same quality of code is generated for the old ops (as typically older version means more things known at compile time). This has the added benefit that the newer op that still use constant (which is just the new way to express the semantic of the older op) will also be optimized to be as good as the older versions of the op...

src/Dialect/ONNX/ONNXOps/Math/Reduction.cpp
```diff
@@ -74,75 +74,87 @@ LogicalResult ONNXGenericReductionOpShapeHelper<OP_TYPE>::customComputeShape(
...
+    // Handle case where reduction op has axes as input.
+    if (isFromNone(operandAdaptor.axes())) {
+      // Default will be used.
+    } else if (getONNXConstantOp(operandAdaptor.axes())) {
+      createIE->getIntFromArrayAsSymbols(operandAdaptor.axes(), axes);
+    } else {
+      // When the axis is dynamic, try to infer the rank of output tensor
+      int64_t dataRank = createIE->getShapedTypeRank(operandAdaptor.data());
+      int64_t axlesSize = createIE->getArraySize(operandAdaptor.axes());
+      if (!operandAdaptor.keepdims() && axlesSize < 0 /*undef shape*/) {
```
chentong319, line 111: Avoid using "<0" for dynamic dim.
Here we return success without improving the shape info, hoping further transformation could provide extra info. If that does not happen, an assertion error for unranked tensor should arise at the pre-lowering check. I tried the following example with main branch(without this PR):
```
func.func private @main_graph(%arg0: tensor<3x2x2xf32>, %arg1: tensor<?xi64>) -> tensor<*xf32> {
  %0 = "onnx.ReduceSum"(%arg0, %arg1) {keepdims = 0 : si64, noop_with_empty_axes = 1 : si64} : (tensor<3x2x2xf32>, tensor<?xi64>) -> tensor<*xf32>
  return %0 : tensor<*xf32>
}
```
I got the following result for shape inference:
```
func.func private @main_graph(%arg0: tensor<3x2x2xf32>, %arg1: tensor<?xi64>) -> tensor<f32> {
  %0 = "onnx.ReduceSum"(%arg0, %arg1) {keepdims = 0 : si64, noop_with_empty_axes = 1 : si64} : (tensor<3x2x2xf32>, tensor<?xi64>) -> tensor<f32>
  return %0 : tensor<f32>
}
```
That's weird. 
Of course, the keep dims == 0 caused assertion error in lowering.
```
loc("reducesum.mlir":3:8): error: not keepdims() not implemented
```
These issues need to be addressed in future PR.

philass:
@AlexandreEichenberger, @tungld, @chentong319

Thanks to all three of you for the feedback.

One thing I would like to mention when considering whether to remove the old lowering for (ReduceMax, ReduceMin, ReduceProd, ReduceMean) is the testing situation. For end to end testing we rely on the ONNX backend tests. Unfortunately these are only for opset 18. It therefor becomes quite difficult to validate that opsets before 18 work correctly (which are the implementation/opsets used in production). There are tests for full models, but its not obvious which ops are called on which models.

I would like to point out I didn't really write any code. Unfortunately the `git diff` doesn't really reflect whats happening. `ReduceSum` already existed with two lowering implementations, and two shape inference implementations. One for `ReduceSum11` and another for `ReduceSum` (introduced in opset 13 with axes inputs). As a result, My code simply templated this implementation to work with all the modified reduction ops, which have lowering implementations.


> Could you explain more about the reason to keep the implementation? I don't see any good reason to keep the shape inference and lowering of V13. Thanks!

The reason for keeping the implementation is that it relax's the pass ordering dependency. Without shape inference implementations for the older versions of the ops, decomposition must be called before shape inference. In general the shape inference already exists for the older version of the op.

I do agree that Rewriting to the latest op is always the cleanest solution. However with the current testing situation, I feel that adding the code as is, reduces the chance of introducing bugs to people who currently use `onnx-mlir` in production.

src/Dialect/ONNX/ONNXOps/Math/Reduction.cpp
philass: Thanks for the review @chentong319,

I actually didn't write that code. I just templated the existing implementation to extend its support to `ReduceProd`, `ReduceMean`, `ReduceMax`, `ReduceMin`.

I do agree with you points. However as you suggested I think addressing them in a future PR makes sense, as they are slightly unrelated with the templating I did. 

The `git diff` of this file doesn't do a good job of showing that only a few line were changed.

AlexandreEichenberger:
@philass thanks for your careful explanation, it makes sense in the light of our discussion at last meeting and your text above.

>  Unfortunately these are only for opset 18. It therefor becomes quite difficult to validate that opsets before 18 work correctly (which are the implementation/opsets used in production). There are tests for full models, but it's not obvious which ops are called on which models.

Is it true that we can preserve some testing if we migrate some of the lit tests before the upgrade to the `vXXX` op names, and then add the same tests for the new opset? That way we can keep some testing.

We can ask @gongsu832 if we should have tests (all the times or sanity from time to time) of backend tests for older versions (not sure how easy/hard it would be from a logistics perspective).

AlexandreEichenberger:
Any blocker on this?

philass:
@AlexandreEichenberger no blockers,

I need to chug a coffee and resolve the merge conflicts. Thanks for the reminder. I'll try and have this done by the meeting tomorrow

src/Conversion/ONNXToKrnl/Math/Reduction.cpp
```diff
@@ -92,6 +132,17 @@ struct ScalarOp<ONNXReduceMeanOp> {
 //===----------------------------------------------------------------------===//
 // Scalar unary ops for lowering ONNXReduceMaxOp
 //===----------------------------------------------------------------------===//
+template <>
```
AlexandreEichenberger, line 135: We could keep common code together.
```C++
static Value emitScalarOpForONNXReduceMaxOps (ConversionPatternRewriter &rewriter,
   MathBuilder createMath(rewriter, loc);
   Value lhs = scalarOperands[0];
   Value rhs = scalarOperands[1];
   Value max = createMath.sgt(lhs, rhs);
   return createMath.select(max, lhs, rhs);
 }
```
and then call this static op in the templated `ONNXReduceMaxOp` and `ONNXReduceMaxV13Op` ops. Same for the other ops that uses the same pattern.

src/Dialect/ONNX/ONNXOps/Math/Reduction.cpp
```diff
@@ -74,80 +74,109 @@ LogicalResult ONNXGenericReductionOpShapeHelper<OP_TYPE>::customComputeShape(
   return success();
 }
 
+template <typename OP>
```
AlexandreEichenberger, line 77: Nice structure, learn something new every day

src/Dialect/ONNX/ONNXOps/Math/Reduction.cpp
```diff
@@ -74,80 +74,109 @@ LogicalResult ONNXGenericReductionOpShapeHelper<OP_TYPE>::customComputeShape(
...
-    // Else set is as questionmarks. Output tensor should have the same rank as
-    // the input. But size of dims is unknown.
-    DimsExpr outputDims(outputRank, QuestionmarkIndexExpr(/*isFloat*/ false));
-    setOutputDims(outputDims);
-    return success();
+    } 
+  } 
+  bool noopWithEmptyAxes = false;
+  if constexpr (isAxesInput<OP_TYPE>) {
+    noopWithEmptyAxes = operandAdaptor.getNoopWithEmptyAxes() != 0;
```
AlexandreEichenberger, line 147: Since the high level description states that this is unsupported, should we not generate an error?

AlexandreEichenberger (approved):
LGTM, left some comments that I let you asses if you want to address with new code or not. Will wait merging #2053 so that you don't have more conflicts to handle.

src/Conversion/ONNXToKrnl/Math/Reduction.cpp
```diff
@@ -380,6 +448,10 @@ struct ONNXReduceSumOpLowering : public OpConversionPattern<ONNXReduceSumOp> {
     int64_t keepdims = adaptor.getKeepdims();
     bool isKeepdims = (keepdims == 1);
 
+    // Noop
+    auto noop = llvm::dyn_cast<ONNXReductionOp>(op).getNoopWithEmptyAxes();
```
tungld, line 452: We don't have to cast here, `reduceOp` can be used directly, so `reduceOp. getNoopWithEmptyAxes()`;

src/Conversion/ONNXToKrnl/Math/Reduction.cpp
```diff
@@ -396,8 +468,7 @@ struct ONNXReduceSumOpLowering : public OpConversionPattern<ONNXReduceSumOp> {
     Value valueOne = nullptr;
     std::map<int64_t, int64_t> outInDimMap;
 
-    // Axes is an optional input, do not use adaptor to get it.
-    Value axesValue = reduceSumOp.getAxes();
+    Value axesValue = llvm::dyn_cast<ONNXReductionOp>(op).getAxes();
```
tungld, line 471: Same here, `reduceOp` can be used directly, so `reduceOp. getAxes()`;

src/Dialect/ONNX/ONNXOps/Math/Reduction.cpp
```diff
@@ -74,80 +74,110 @@ LogicalResult ONNXGenericReductionOpShapeHelper<OP_TYPE>::customComputeShape(
...
+      }
+      // With keep dim, by def the output has the same rank as the input;
+      // without keep dim, we remove 1 rank per value in the 1D axes tensor.
+      int64_t outputRank =
+          operandAdaptor.getKeepdims() ? dataRank : dataRank - axlesSize;
+      assert(outputRank >= 0 && "expected to keep at least one dim");
+
+      // Interesting idea below, namely to reuse info from output, or if not
+      // there, from input putting question mark in there. Not sure if
+      // successful, if it is, it should be generalized to all ops.
```
tungld, line 130: This is done automatically whenever you call `computeShape`: https://github.com/onnx/onnx-mlir/blob/main/src/Dialect/ONNX/ONNXOps/ShapeHelper.cpp#L217. It will compare the computed shape with existing shape of the output, and get information from the existing shape if possible. For example, computed dim is ? but existing dim is 5, then 5 is used.

philass:
Hey @AlexandreEichenberger,

Thanks for the review! I think the  feedback you gave makes sense. I'll make those changes in a following PR, where I handle the noop case. 

I want to make sure I land this tonight before it gets to out of sync with head

AlexandreEichenberger:
@philass thanks for the PR. Your help is much appreciated. Big thanks

PR #1944, opened 2022-12-30
Introduce custom quick sort in src/Runtime/OMSort.inc (No test executed on zOS!!)

negiyas:
This PR introduces a custom quick sort function in src/Runtime/OMSort.inc to try to fix a zos issue (not supporting qsort_r).
**### No test executed on zOS!!**
A bug included in the custom quick sort function was reported on z/OS, was reproduced by using the custom quick sort function on zLinux, and confirmed that is was fixed., but still not tested on z/OS.


cjvolzka:
I gave this a test and all the modelzoo models pass with this fix on z/OS.

PR #1946, opened 2023-01-03
Adding shape inference for RNNs

AlexandreEichenberger:
Note that there was a strange behavior that I fixed, @tungld .

RNNs with layout=1 were changed by a Rewrite pattern, where inputs and outputs were transposed, and fed in the SAME RNN op. That op was then sent to recompute shape. But it had erroneous output shapes (from before the transformation, as the same op was used, not a new one). Thus the regular shape inference, which picks the "best" out of the shape inference and the output types, it was picking the output type. That problem did not exist before as it was not using the shape inference scheme, and thus did not use the code that merge old type and new interested type shapes.

So my fix was to reset the output types of reused op before being sent to shape inference to have -1 only shapes. That was much cleaner than starting to distinguish in shape inference which ops have valid output shapes vs which ones do not.

As mentioned in the code, it might be cleaner to create a new RNN op instead of reusing the previous one... but it felt more intrusive a solution at this time. I let you assess if that would be a simpler fix @tungld .


src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -525,6 +525,7 @@ class RNNOpRewriteLayoutPattern : public OpRewritePattern<ONNXOp> {
           rewriter.getIntegerAttr(
               rewriter.getIntegerType(64, /*isSigned=*/true), 0));
       // Update the output shape.
+      resetTypesShapeToQuestionmarks(onnxOp);
```
tungld, line 528: Could you please put some additional comments here (something similar to the comments for `resetTypesShapeToQuestionmarks`)? Though it may be duplicated, it helps other developers know the reason when scanning the code here?

src/Transform/ONNX/SimplifyShapeRelatedOps.cpp
```diff
@@ -45,7 +45,7 @@ exposed and easily propagated through other ops. For example, the above
 %4 = "onnx.Reshape"(%arg0, %3) : (tensor<?x256xf32>,  tensor<3xi64>) -> tensor<?x?x?xf32>
 ```
 
-Now, it's straighforward to update the output shape of Reshape from
+Now, it's straightforward to update the output shape of Reshape from
```
tungld, line 48: Thanks!

src/Dialect/ONNX/ONNXOps/ShapeHelper.cpp
```diff
@@ -384,6 +384,33 @@ void updateType(Value val, ArrayRef<int64_t> shape, Type elementType,
...
+  SmallVector<int64_t, 4> newShape(shape.size(), -1);
+  // Build result type.
+  RankedTensorType resType;
+  if (encoding)
+    resType = RankedTensorType::get(newShape, elementType, encoding);
+  else
+    resType = RankedTensorType::get(newShape, elementType);
+  // Reset type
+  val.setType(resType);
+}
```
tungld, line 406: I wonder if we can write this function in a more concrete way, say
```C++
// Only deal with ranked tensor types here.
RankedTensorType valType = val.getType().dyn_cast<RankedTensorType>();
if (!valType)
    return;
// Reset any compile time literal to unknown (aka question marks).
SmallVector<int64_t, 4> newShape(valType.getRank(), -1);
auto resType = RankedTensorType::Builder(valType).setShape(newShape);
// Reset type
val.setType(resType);
```

I haven`t tested the code, but hope it works.

src/Dialect/ONNX/ONNXOps/RNN/RNN.cpp
```diff
@@ -22,138 +22,162 @@ using namespace onnx_mlir;
...
-      yTy = RankedTensorType::get(
-          {batchSize, seqLength, numDir, hiddenSize}, elementType);
+      yOutputDims = {batchSize, seqLength, numDir, hiddenSize};
     } else {
-      yTy = RankedTensorType::get(
-          {seqLength, numDir, batchSize, hiddenSize}, elementType);
+      yOutputDims = {seqLength, numDir, batchSize, hiddenSize};
     }
-    op->getResult(0).setType(yTy);
+    IndexExpr::debugPrint("set0", yOutputDims);
```
tungld, line 110: Not sure if you still want to keep this debug in the final code. Same for `set1` and `set2` below.

src/Dialect/ONNX/ONNXOps/ShapeHelper.cpp
```diff
@@ -68,8 +68,8 @@ static void refineDims(DimsExpr &inferredDims, Value output) {
     if (inferredDims[i].isLiteral() &&
         (existingDims[i] != inferredDims[i].getLiteral())) {
       // Warning for users.
-      llvm::outs() << "Warning: [Shape inference] the inferred dim ("
-                   << inferredDims[i].getLiteral()
+      llvm::outs() << "Warning: [Shape inference, dim " << i
+                   << "] the inferred dim (" << inferredDims[i].getLiteral()
```
tungld, line 72: Thanks!

tungld (approved):
LGTM! Thanks for updating RNN ops!

Just some minor comments and please feel free to address them before merging.

src/Dialect/ONNX/Rewrite.cpp
AlexandreEichenberger: Good suggestion. Added comments at your request.

src/Dialect/ONNX/ONNXOps/ShapeHelper.cpp
AlexandreEichenberger: totally make sense.

src/Dialect/ONNX/ONNXOps/RNN/RNN.cpp
AlexandreEichenberger: should have caught it, thanks for spotting it.

PR #1947, opened 2023-01-05
Shapehelper cleanup: removal of the Capture interface

AlexandreEichenberger:
Now that the Shape Helper interface is fully in place, there is no need for the old Capture interface. I had to further upgrade lot of code to fully eliminate their uses; now it is removed,


tungld (approved):
LGTM!

PR #1948, opened 2023-01-06
use DenseElementsAttr::reshape to implement foldOrEmit

sorenlassen:
Simplified the implementation of foldOrEmitONNXSqueezeV11Op, foldOrEmitONNXUnsqueezeV11Op.

Question: Is it ok to change the logic to create an ONNXConstantOp of tensor type and fix the lit test accordingly? In my view, it's a mistake that the old logic created ONNXConstantOps with memref type.

### DIFF
--- a/src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
+++ b/src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
@@ -337,12 +337,7 @@ Value foldOrEmitONNXSqueezeV11Op(ConversionPatternRewriter &rewriter,
         getDenseElementAttributeFromONNXValue(input);
     TensorType tensorType = create.onnx.toTensor(resultType);
     DenseElementsAttr squeezedElements = inputElements.reshape(tensorType);
-    // TODO: Fix lit tests so we can change this to just:
-    // Value constVal = create.onnx.constant(squeezedElements);
-    // The lit tests want the constant to have memref type, so we do this:
-    Value constVal = rewriter.create<ONNXConstantOp>(loc, resultType,
-        Attribute(), squeezedElements, FloatAttr(), ArrayAttr(), IntegerAttr(),
-        ArrayAttr(), StringAttr(), ArrayAttr());
+    Value constVal = create.onnx.constant(squeezedElements);
     return create.onnx.toMemref(constVal);
   } else {
     return create.onnx.toMemref(
--- a/src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
+++ b/src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
@@ -330,19 +330,18 @@ Value getDimOrConstant(ConversionPatternRewriter &rewriter, Location loc,
 /// and return a constant.
 Value foldOrEmitONNXSqueezeV11Op(ConversionPatternRewriter &rewriter,
     Location loc, Type resultType, Value input, int64_t axis) {
-  MultiDialectBuilder<KrnlBuilder, MathBuilder, MemRefBuilder, OnnxBuilder>
-      create(rewriter, loc);
+  MultiDialectBuilder<OnnxBuilder> create(rewriter, loc);
+  TensorType tensorType = create.onnx.toTensor(resultType);
   if (krnl::isKrnlGlobalConstant(input) || isDenseONNXConstant(input)) {
     DenseElementsAttr inputElements =
-        getDenseElementAttributeFromONNXValue(input);
-    TensorType tensorType = create.onnx.toTensor(resultType);
+        input.getDefiningOp()->getAttrOfType<DenseElementsAttr>("value");
     DenseElementsAttr squeezedElements = inputElements.reshape(tensorType);
     Value constVal = create.onnx.constant(squeezedElements);
     return create.onnx.toMemref(constVal);
   } else {
     return create.onnx.toMemref(
         rewriter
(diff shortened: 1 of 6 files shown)


src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
```diff
@@ -333,16 +333,16 @@ Value foldOrEmitONNXSqueezeV11Op(ConversionPatternRewriter &rewriter,
...
-    Value constVal =
-        create.onnx.constantFromRawBuffer(resultType, outputBuffer);
-    free(outputBuffer);
-    free(inputBuffer);
+    DenseElementsAttr inputElements =
+        getDenseElementAttributeFromONNXValue(input);
+    TensorType tensorType = create.onnx.toTensor(resultType);
+    DenseElementsAttr squeezedElements = inputElements.reshape(tensorType);
+    // TODO: Fix lit tests so we can change this to just:
+    // Value constVal = create.onnx.constant(squeezedElements);
```
sorenlassen, line 341: @tungld I'd like to simplify the logic to do this, and change the lit tests accordingly. Please let me know if you agree with this change to the lit tests.

tungld: @sorenlassen yes, at onnx level, we don't need memref any more. 

Only one concern is during the lowering, it looks like we have a mix of krnl constants and onnx constants. onnx constants will be lowered to krnl constants that are working on memref. Note that we have `if (krnl::isKrnlGlobalConstant(input) || isDenseONNXConstant(input)) {` at the beginning.

sorenlassen: this code generates an onnx.Constant (`create.onnx.constantFromRawBuffer()`) which would be followed by a builtin.unrealized_conversion_cast (`create.onnx.toMemref()`) if the constant had tensor type, which seems correct regardless whether input is a krnl constant or onnx constant

I think we can conclude that it was wrong to construct a onnx.constant with memref type (which would cause 'const.toMemref()' to skip the builtin.unrealized_conversion_cast)

therefore I will change this to
```
TensorType tensorType = create.onnx.toTensor(resultType);
DenseElementsAttr squeezedElements = inputElements.reshape(tensorType);
Value constVal = create.onnx.constant(squeezedElements);
return create.onnx.toMemref(constVal);
```
which will generate an onnx.constant with tensor type followed by a builtin.unrealized_conversion_cast to memref type, and then these two will be lowered to a krnl constant elsewhere in the lowering logic

I will make the change and update the lit tests

sorenlassen: done

the lit test changes show that some onnx.Transpose and onnx.SplitV11 ops failed to fold when builtin.unrealized_conversion_cast is inserted after the onnx.Constant when foldOrEmitONNXSqueezeV11Op folds ONNXSqueezeV11Op

alternatively, we could create krnl.global constant:
```
TensorType tensorType = create.onnx.toTensor(resultType);
DenseElementsAttr squeezedElements = inputElements.reshape(tensorType);
return create.krnl.constant(resultType, "constant_", squeezedElements);
```
then the constant triggers constant folding of Transpose and SplitV11 again, but I'm unsure if it's correct to generate Krnl ops here

yet another alternative is to refine foldOrEmitONNXTransposeOp, foldOrEmitONNXSplitOp, and all the constant folding methods to not just check `if (krnl::isKrnlGlobalConstant(input) || isDenseONNXConstant(input))` but also trigger if input is a builtin.unrealized_conversion_cast following a onnx.Constant and then fold that constant

src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
```diff
@@ -333,16 +333,11 @@ Value foldOrEmitONNXSqueezeV11Op(ConversionPatternRewriter &rewriter,
...
-            ->getAttrOfType<::mlir::Attribute>("value")
-            .dyn_cast_or_null<mlir::DenseElementsAttr>());
-    char *outputBuffer = allocateBufferFor(resultType, /*useMaxSize=*/false);
-    convertDoubleInt64ToExactType(resultType, inputBuffer, outputBuffer);
-    Value constVal =
-        create.onnx.constantFromRawBuffer(resultType, outputBuffer);
-    free(outputBuffer);
-    free(inputBuffer);
+    DenseElementsAttr inputElements =
+        getDenseElementAttributeFromONNXValue(input);
```
sorenlassen, line 337: TODO: I need to update this so it also handles the krnl.global constant case

same in lines 357-358 below

sorenlassen: fixed

src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
sorenlassen: > yet another alternative is to refine foldOrEmitONNXTransposeOp, foldOrEmitONNXSplitOp, and all the constant folding methods to not just check `if (krnl::isKrnlGlobalConstant(input) || isDenseONNXConstant(input))` but also trigger if input is a builtin.unrealized_conversion_cast following a onnx.Constant and then fold that constant

I drafted that solution in draft PR #1954 if you want to check that out

sorenlassen:
abandoning this PR, will go with PR #1954 instead

PR #1950, opened 2023-01-06
Shapehelper reorg: Adding of the shape helper interface

AlexandreEichenberger:
Every ONNX ops now has a `getShapeHelper(...)` function that returns the proper shape helper object for the given ONNX operation.

At this time, there remains about 30+ ops that do not have shape helpers. Some are trivial, some not so. A later PR will add the remaining easy ops without shape helpers.

Since the interface is mandatory, operations without shape helpers return a generic `ONNXUnimplementedOpShapeHelper` that fails when calling `computeShape` on it.

The interface has the following comment:
```
     Interface to access a registered method to infer the return shapes for an
    operation that can be used during type inference. The shape inference
    interface method takes an operation and its operands as input, augmented
    by an index expression builder and an index expression scope as optional
    inputs. It returns a specialized shape helper object for that operation,
    casted as a generic ONNXOpShapeHelper.

    For operations that do not support shape helpers at this stage, a
    `ONNXUnimplementedOpShapeHelper` object is returned. This object does not
    compute shapes, and simply return failure when `computeShape` is called
    on it.

    The new object is allocated on the heap and it is the responsability
    of the object user to free the memory after last use.
```

This interface is not yet hooked up to the Analysis or shape inference passes.


src/Interface/ShapeHelperOpInterface.hpp
```diff
@@ -0,0 +1,143 @@
...
+      IndexExprBuilder *ieBuilder, /* Use IndexExprBuilderForAnalysis if null.*/
+      IndexExprScope *scope);      /* Install local scope if null. */
+  virtual ~ONNXOpShapeHelper();
+
+  // Every leaf class is expected to create a computeShape with the following
+  // signature. This method is responsible to compute at a minimum the output
+  // dims.
+  virtual mlir::LogicalResult computeShape() = 0;
+  // Helper function that set n'th output dims from the given value.
+  mlir::LogicalResult computeShapeFromOperand(mlir::Value operand, int n = 0);
```
chentong319, line 93: The purpose of this function is unclear to me: Can the shape be computed with only one operand? Should "ValueRange operands" be used or does this function is just to set the dim?

src/Dialect/ONNX/ONNXOps/Math/Bernoulli.cpp
```diff
@@ -26,6 +26,11 @@ using namespace onnx_mlir;
 // Shape Inference
 //===----------------------------------------------------------------------===//
 
+ONNXOpShapeHelper *ONNXBernoulliOp::getShapeHelper(Operation *op,
+    ArrayRef<mlir::Value> oper, IndexExprBuilder *ieb, IndexExprScope *scope) {
+  return getNewShapeHelper<ONNXUnaryOpShapeHelper>(op, oper, ieb, scope);
+}
```
tungld, line 32: It looks like this code is not automatically generated, doesn't it?

I think the current infrastructure you built is quite good so that we see repeated calling patterns for all operations. Given that, it looks to me, we can define a list of operations that need `ShapeHelperOpInterface` in `gen_onnx_mlir.py` and automatically generate `getShapeHelper` for each operation by using `gen_onnx_mlir.py`.

To do that, each operation need to have a shape helper with the same name, `ONNXBernoulliOpShapeHelper` for `ONNXBernoulliOp`. In such a case, a new class `ONNXBernoulliOpShapeHelper` that inherits `ONNXUnaryOpShapeHelper` is needed. What do you think?

tungld (commented):
Thanks for adding this useful interface! The code is good and I just have comments on how to make `getShapeHelper` code automatically generated.

src/Interface/ShapeHelperOpInterface.hpp
AlexandreEichenberger: @chentong319 some ONNX ops have multiple inputs but the shape is only defined by one operand (meaning identical to that one). This function allow me to reuse the "set output # `i` to be the same as the shape as the type of value `operand`". This only work for some ops, obviously.

src/Dialect/ONNX/ONNXOps/Math/Bernoulli.cpp
AlexandreEichenberger: Yes, that is correct. There are many ops that reuse a shape type at the moment, for example all of the Unary and Binary. 

If you want me to go that route, I will have to put a 
```
using ONNXxxxOpShapeHelper = ONNXUnaryOpShapeHelper<ONNXxxxOp>
```
for each op that does not already have a specific helper for it. 

Let me know if you prefer that pattern, and I can try migrate to that pattern.

src/Interface/ShapeHelperOpInterface.hpp
AlexandreEichenberger: Based on your input, I am clarifying the comments in a subsequent PR, thanks for the feedback on this one.

src/Dialect/ONNX/ONNXOps/Math/Bernoulli.cpp
tungld: Yes, please migrate to that pattern, so that we have a uniform interface for all ops' shape helper. Thanks!

AlexandreEichenberger: will do

tungld: > The only way this could work if there was a way to define in the ODS of the ONNX op code to be only included in the cpp file, which I cannot find at this time.

As I understand, `extraClassDefinition` will add code to the created cpp file and `extraClassDeclaration` will add code to the created hpp file. So if we don't have `extraClassDeclaration`, the code will be only added to the cpp file. Does it still not work because of cyclic dependency?

AlexandreEichenberger: It looks like I can disable the interface for the ONNX Ops, and add the shapeHelper as `extraClassDeclaration` with the code once I remove the dependence from ONNXOps/ShapeHelper.hpp to ONNXOps.hpp, by using the forward declaration of all of the ONNX Ops classes found using GET_OP_FWD_CLASSES define in conjunction with the automatically generated ONNXOps.hpp.inc (which will define only the `class ONNXAddOp;` type fo refs).

Additional ONNX Ops have to be handled separately.

AlexandreEichenberger:
#1955 implements the changes proposed in this PR.

PR #1951, opened 2023-01-09
heap reporting

sorenlassen:
Added HeapReporter PassInstrumentation which dumps a heap log before/after the passes specified in the command line arguments ---report-heap-before and --report-heap-after.

Added script utils/onnx-mlir-report-heap.sh to run onnx-mlir with the necessary setup and execution environment, which means running it in the debugger on MacOS.

Only implemented for MacOS for now. Hopefully, someone will implement it on other OSs in the future.

Was used to measure the memory effects in PR #1874.


AlexandreEichenberger (approved):
LGTM.

AlexandreEichenberger:
@gongsu832 would you be familiar with what might be needed for Linux?

PR #1955, opened 2023-01-12
Shapehelper reorg: Adding of the shape helper interface (v2)

AlexandreEichenberger:
This is a version of the Shape Helper interfaces that improves on #1950 in 2 ways:

* each ONNX op now is mandated to have its own ShapeHelper class. They can be a `using` reference to another one.
* code to `getShapeHelper` is now auto-generated by the `gen_onnx_mlir.py` file.

Otherwise, functionality is the same


src/Interface/ShapeHelperOpInterface.td
```diff
@@ -0,0 +1,53 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//===- ShapeHelperInterface.td - ShapeHelper Interface -*- tablegen -===//
```
tungld, line 3: Please expand it up to 80 characters.

tungld (approved):
Great! Thank you for a quick turnaround!

src/Interface/ShapeHelperOpInterface.td
AlexandreEichenberger: Sure, since there is already a follow up PR where I add the interface to all but the ops that accept Seq and Opt, let me make the change there.

PR #1956, opened 2023-01-13
Shapehelper reorg: shape helper for all ops (but the ones with optional and sequences)

AlexandreEichenberger:
Added many remaining ONNX ops that were using the UNDEFINED shape helper. Now only a few ops are remaining without it, namely ops that use Opt or Seq types. These need some special handling that is not currently avail within the infrastructure.


tungld (approved):
LGTM!

PR #1957, opened 2023-01-14
Fix Resize implementation

philass:
This PR reverts the resize implementation of `shapeInference`. The issue is the new implementation casts `scales[i]` to an integer. This is not the correct behavior. 
 
It is hard to implement Resize with `indexExprs` because of the lack of support for the float type. Since Resize shape inference has a bug for `scales` I've elected to go to the old implementation that way we have it working correctly on `main`. 

I have filed an associated issue https://github.com/onnx/onnx-mlir/issues/1958.

@AlexandreEichenberger this only meant as a temporary roll back until we figure out a proper way to support floating point numbers in the `IndexExpr` framework.

### DIFF
--- a/src/Dialect/ONNX/ONNXOps/Tensor/Resize.cpp
+++ b/src/Dialect/ONNX/ONNXOps/Tensor/Resize.cpp
@@ -76,9 +76,60 @@ LogicalResult ONNXResizeOp::inferShapes(
   if (!X().getType().isa<RankedTensorType>()) {
     return success();
   }
-  Type elementType = X().getType().cast<RankedTensorType>().getElementType();
-  ONNXResizeOpShapeHelper shapeHelper(getOperation(), {});
-  return shapeHelper.computeShapeAndUpdateType(elementType);
+  // TODO : Return to this implementation once floating point scales are handled
+  //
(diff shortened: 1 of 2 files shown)


src/Dialect/ONNX/ONNXOps/Tensor/Resize.cpp
```diff
@@ -76,9 +76,60 @@ LogicalResult ONNXResizeOp::inferShapes(
...
+      if (ShapedType::isDynamic(inputTy.getShape()[i]))
+        newDim = -1;
+      else
+        newDim = inputTy.getShape()[i] * scalesConstant[i];
+      dims.emplace_back(newDim);
+    }
+
+    updateType(getResult(), dims, inputTy.getElementType());
+  } else {
+    ElementsAttr sizesAttrs = getElementAttributeFromONNXValue(sizes());
```
sorenlassen, line 120: let's keep the ONNXResizeOpShapeHelper logic in this case

philass: I've updated the PR with your suggestion

sorenlassen (approved):
LGTM provided the CI agrees

PR #1959, opened 2023-01-16
constant propagation for ONNXWhereOp

sorenlassen:
includes a bug fix in getMemoryBuffer() in ElementsAttrBuilder.cpp, which was exposed by the new const prop for Where

similarly, also added some checks in ElementsAttrHelper.hpp.inc in ElementsAttrHelper.cpp which would have caught some mistakes, and I threw in a refactor to remove code duplication between the get/readElementsWideNums() implementations, plus an optimization when the ElementsAttr is splat

### DIFF
--- a/src/Transform/ONNX/ConstProp.cpp
+++ b/src/Transform/ONNX/ConstProp.cpp
@@ -239,7 +239,7 @@ Value ConstPropWhere(PatternRewriter &rewriter, Value replacingValue,
   Type replacingType = replacingValue.getType().cast<ShapedType>();
 
   ElementsAttr cond = getConstValueElements(condValue);
-  assert(cond.getElementType.isInteger(1) &&
+  assert(cond.getElementType().isInteger(1) &&
          "ONNXWhereOp condition has bool element type");
   ElementsAttr lhs = getConstValueElements(lhsValue);
   ElementsAttr rhs = getConstValueElements(rhsValue);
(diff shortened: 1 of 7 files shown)


src/Transform/ONNX/ConstProp.cpp
```diff
@@ -226,6 +226,33 @@ Value ConstPropElementwiseUnary(
...
+//===----------------------------------------------------------------------===//
+// Code to perform constant propagation for ONNXWhereOp in presence of
+// broadcast.
+//
+/// Does element-wise ternary (cond ? lhs : rhs) with broadcast on all inputs.
+//===----------------------------------------------------------------------===//
+
+Value ConstPropWhere(PatternRewriter &rewriter, Value replacingValue,
+    Value condValue, Value lhsValue, Value rhsValue) {
+  ConstPropCounters::count("ElementwiseBinary", {lhsValue, rhsValue});
```
philass, line 238: I don't think `ElementwiseBinary` is the correct term for this?

philass (approved):
LGTM

It could be nice if `ConstPropWhere` didn't need the result as the first argument. But I see how it greatly simplifies the implementation to not scrape the type off of it.

src/Transform/ONNX/ConstProp.cpp
sorenlassen: good catch, fixed

sorenlassen:
> It could be nice if `ConstPropWhere` didn't need the result as the first argument. But I see how it greatly simplifies the implementation as it allows the type to be scraped off of it as opposed to manually calculating the broadcasted shape if the condition, and the if else branches.

yeah, this is how all the ConstPropX functions have always been implemented

fwiw, replacingValue.getLoc() is also used in the construction of the resulting constant, but that probably doesn't matter

PR #1961, opened 2023-01-16
Cleanup of is<RankedTensorType> in shape inference

AlexandreEichenberger:
Non functional changes

The function `hasShapeAndRank` is more precise an cover more cases than the simple `is<RankedTensorType>()` in shape inference. Replaced all the cases where it made sense.


tungld (approved):
LGTM!

PR #1965, opened 2023-01-18
clean up some old code

chentong319:
Two changes:
1. Detect optional input. The NoneType is originally used to represent the optional input when it is not presented. Later, some onnx model used 0 length tensor for that purpose. We have function isFromNone for both case.  This change replace the NoneType check for some old code with isFromNone. Also isFromNone is changed to check the type, not how the value (definition point) is defined because a Value may not have a definition point in constructing process.
2. Delete unnecessary use of TypeConverter. The TypeConverter should be called automatically on the input by Conversion process, according to the document. Old code did not use type converter and the explicit type conversion was replaced with type converter when type converter is added. Such code is not needed.

### DIFF
--- a/src/Conversion/ONNXToKrnl/RNN/RNNBase.cpp
+++ b/src/Conversion/ONNXToKrnl/RNN/RNNBase.cpp
@@ -19,9 +19,6 @@ using namespace mlir;
 
 namespace onnx_mlir {
 
-// Check a Value's type is none or not.
-bool isNoneType(Value val) { return val.getType().isa<NoneType>(); }
-
 // Get a dimension of the tensor's shape.
 int64_t dimAt(Value val, int index) {
   return val.getType().cast<ShapedType>().getShape()[index];
@@ -35,7 +32,7 @@ Value allocAllHidden(ConversionPatternRewriter &rewriter, Location loc,
   IndexExprBuilderForKrnl createKrnlIE(rewriter, loc);
   IndexExprScope scope(createKrnlIE);
   Value alloc;
-  if (!isNoneType(output)) {
+  if (!isFromNone(output)) {
     SmallVector<IndexExpr, 4> dims;
     // Get seq_length from X.
     dims.emplace_back(createKrnlIE.getShapeAsDim(X, 0));
@@ -113,14 +110,14 @@ void initializeIntermediateStates(ConversionPatternRewriter &rewriter,
           initialIVs.emplace_back(zeroIndex);
           initialIVs.emplace_back(loopInd[0]);
           initialIVs.emplace_back(loopInd[1]);
-          if (isNoneType(initialH))
+          if (isFromNone(initialH))
             createKrnl.store(zero, forwardHt, IVs);
           else {
             Value h = createKrnl.load(initialH, initialIVs);
             createKrnl.store(h, forwardHt, IVs);
           }
           if (!onlyHidden) {
-            if (isNoneType(initialC))
+            if (isFromNone(initialC))
               createKrnl.store(zero, forwardCt, IVs);
             else {
               Value c = createKrnl.load(initialC, initialIVs);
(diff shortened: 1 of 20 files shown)


src/Dialect/ONNX/ONNXOps/OpHelper.cpp
```diff
@@ -327,18 +327,17 @@ Value createNoneFloatConstant(PatternRewriter &rewriter, Location loc) {
...
 
-    if (auto c = dyn_cast<ONNXConstantOp>(op))
-      if (c.value().has_value())
-        if (auto e = c.valueAttr().dyn_cast<ElementsAttr>()) {
-          auto shape = e.getType().getShape();
-          if (shape.size() == 1 && shape[0] == 0)
-            return true;
-        }
+  if (auto ty = v.getType().dyn_cast<ShapedType>()) {
+    auto shape = ty.getShape();
```
tungld, line 338: Do we need to check `hasRank` before obtaining the shape?

src/Conversion/ONNXToKrnl/RNN/RNNBase.cpp
```diff
@@ -19,9 +19,6 @@ using namespace mlir;
 
 namespace onnx_mlir {
 
-// Check a Value's type is none or not.
```
AlexandreEichenberger, line 22: excellent, by removing the function, we can be sure it is not used anymore. I am assuming that you removed all definitions and uses.

src/Dialect/ONNX/ONNXOps/OpHelper.cpp
```diff
@@ -327,18 +327,17 @@ Value createNoneFloatConstant(PatternRewriter &rewriter, Location loc) {
 }
 
 // Returns true if the Value is defined by a unit constant.
+// The unit constant can  be 1. NoneType, or 2. 1D tensor with 0 length
```
AlexandreEichenberger, line 330: The name `isFromNone` maybe technically correct but does not really convey the meaning of the specs. For example, CONV is defined as having an optional input  bias.

Since we are already changing names, I would recommend to use `hasOptionalInputValue(...)`, which would  returns false if "is from none" and true otherwise.

AlexandreEichenberger:
Question: did you also eliminate all the `isa<NoneType>()` ?

src/Dialect/ONNX/ONNXOps/OpHelper.cpp
chentong319: I thought about changing the name. But the new name  containing "Optional" may be confused with the new feature OptionalType. How about leaving it?

src/Dialect/ONNX/ONNXOps/OpHelper.cpp
chentong319: No.  I tested with unranked tensor. UnrankedTensor is not a ShapedType. We need to call hasRank for TensorType.

src/Dialect/ONNX/ONNXOps/OpHelper.cpp
AlexandreEichenberger: Yes, Optional is used in 2 context, that is why I had OptionalInput here... Maybe `IsNone`?

If you want to keep it as is, I am fine too.

sorenlassen:
> The NoneType is originally used to represent the optional input when it is not presented. Later, some onnx model used 0 length tensor for that purpose. We have function isFromNone for both case.  This change replace the NoneType check for some old code with isFromNone. Also isFromNone is changed to check the type, not how the value (definition point) is defined because a Value may not have a definition point in constructing process.

I think it's a really good idea to clean up the inconsistencies about optional inputs and NoneType / isFromNone.

I think the best long term solution is to make all optional inputs NoneType so that `!value.getType().isa<NoneType>` is always the correct and unambiguous way to check if a value has a value. Using an empty tensor seems wrong because empty tensors are valid values in most situations as most tensor operations are well-defined for empty tensors.

This PR may be a good step forward to reduce confusion, but I'm curious if people agree that it's a good long term direction to standardize on NoneType or not as the single way to express if an optional input is absent or present?

chentong319:
It is the implementation detail to use NoneType or zero-sized tensor to represent the optional input. The zero-size tensor has extra info for the element type. The new OptionalType contains further more type info for the tensor. This PR is intended to encapsulate the implementation for possible future evolving.

chentong319:
> Picking up on @sorenlassen message. I understand that we appear have multiple ways to represent optional inputs. The useful revision here provide a method that return the right results no matter how we represent optional inputs.
> 
> Soren is asking if we could normalize to one method only? Are there technical difficulties that forces us to have the one vs the other?

That's doable.

sorenlassen:
I support that we land this PR in its current form and open an issue where we can discuss and try to agree to normalize to one method only, and then one of us can implement anything we agree on in a follow-on PR

AlexandreEichenberger (approved):
LGTM, thanks for the cleanup effort, much appreciated

sorenlassen:
> I think the best long term solution is to make all optional inputs NoneType so that `!value.getType().isa<NoneType>` is always the correct and unambiguous way to check if a value has a value.

I implemented this in PR #2133

PR #1973, opened 2023-01-24
Block copy for ONNXTranspose

tungld:
This patch implements ONNXTranspose by doing block copy if the inner-most dimensions are kept unchanged or not permuted.
For example, below is an ONNXTranspose in the bertsquad model:
```mlir
%0 = "onnx.Transpose"(%arg0) {perm = [0, 2, 1, 3] } : (tensor<?x256x12x64xf32>) -> tensor<?x12x256x64xf32>
```

Some performance results for the above ONNXTranspose with different batch sizes:

- block copy: 
    - bs1: 0.000576365739107132  seconds.
    - bs8: 0.004491869360208511  seconds 
    - bs16: 0.0086372010409832  seconds
    - bs32: 0.02807724103331566  seconds
    - bs64: 0.05093304067850113  seconds
    - bs128: 0.1489739865064621 seconds
    - bs256: 0.31475839018821716  seconds.
 
 - element-wise/scalar copy:
    - bs1: 0.0006000734865665436  seconds
    - bs8: 0.0046231709420681  seconds.
    - bs16: 0.008425723761320114  seconds.
    - bs32: 0.022214848548173904  seconds.
    - bs64: 0.05527787283062935  seconds
    - bs128: 0.21235326677560806 seconds
    - bs256 0.4105483293533325  seconds.

Block copy looks better with large batch sizes (128 and 256), while both show similar performance for small batch sizes.

In this patch, KrnlMemcpy and its lowering to LLVM are updated to support offsets.


src/Conversion/KrnlToLLVM/KrnlMemcpy.cpp
```diff
@@ -39,45 +39,53 @@ class KrnlMemcpyOpLowering : public ConversionPattern {
...
 
     // First operand.
-    Type dstType = operandAdaptor.dest()
-                       .getType()
-                       .cast<LLVM::LLVMStructType>()
-                       .getBody()[1];
-    Value alignedDstMemory =
-        create.llvm.extractValue(dstType, operandAdaptor.dest(), {1});
-    Value alignedInt8PtrDstMemory = create.llvm.bitcastI8Ptr(alignedDstMemory);
+    Value alignedDstMemory = create.llvm.extractValue(dstType, dest, {1});
```
AlexandreEichenberger, line 65: Thanks for adding to the builders :-)

src/Conversion/ONNXToKrnl/Tensor/Transpose.cpp
```diff
@@ -20,83 +20,192 @@ using namespace mlir;
 namespace onnx_mlir {
 
 struct ONNXTransposeOpLowering : public ConversionPattern {
+  using MDBuilder = MultiDialectBuilder<KrnlBuilder, IndexExprBuilderForKrnl,
```
AlexandreEichenberger, line 23: If you wanted to shorten `MultiDialectBuilder` to `MDBuilder` globally, I think that would improve readability as it kind of obvious that it is for multiple ones as we see the list in the template brackets. I understand that you are using it here in a `using`, which is another way to shorten it too.

src/Conversion/ONNXToKrnl/Tensor/Transpose.cpp
```diff
@@ -20,83 +20,192 @@ using namespace mlir;
...
         typeConverter->convertType(*op->result_type_begin());
     assert(outConvertedType && outConvertedType.isa<MemRefType>() &&
            "Failed to convert type to MemRefType");
     MemRefType outMemRefType = outConvertedType.cast<MemRefType>();
-    uint64_t outRank = outMemRefType.getShape().size();
 
     // Get shape.
     ONNXTransposeOpShapeHelper shapeHelper(op, operands, &create.krnlIE);
     shapeHelper.computeShapeAndAssertOnFailure();
+    DimsExpr &outDims = shapeHelper.getOutputDims();
```
AlexandreEichenberger, line 51: I like the overall idea a lot, namely to try to move data in chunks. I wonder if we could not create an infrastructure that would allow this type of optimization to be used elsewhere too, with an interface that would be easy to use. 

krnlOptDataMove(loop indices, vector of LB for loop iteration, vector of UB for loop iteration, memref from, memref to, set of affine map access function from, set of affine map access function to)

if there is an easy way to check which affine function are identical (is there such a check?), and/or which are continuous, then we could automatically detect which ones can be combined.

Alternatively, we could also generate the normal code, and then have an op that gather contiguous loads. That could be run after memref normalization, so it would handle even cases with complicated maps. I would even check if there is not already something like this in MLIR.

I understand that this might be for another PR, just wanted to see if this approach can be applied to all the other memcpy that we experience when lowering ops.

src/Conversion/ONNXToKrnl/Tensor/Transpose.cpp
```diff
@@ -20,83 +20,192 @@ using namespace mlir;
...
+  }
 
-    if (originalAxes == permutedAxes) {
-      // It is safe to lower to a view op.
-      MemRefBuilder createMemRef(rewriter, loc);
-      Value view =
-          createMemRef.reinterpretCast(data, shapeHelper.getOutputDims());
-      rewriter.replaceOp(op, view);
-      return success();
+  // Determine how many consecutive inner-most dimensions are not permuted.
```
AlexandreEichenberger, line 98: Is it true that this only holds if there is no map on the members? If so, I would mention it in the comments to the function.

chentong319 (commented):
1. It is definitely good for performance if the data can be moved in chunks, instead of element by element. The PR is great to demonstrate the performance gain on transposeOp
2. In future, a general approach should be considered. For example, Linalg abstracts out the maps from loops to tensor access for all the tensor accesses. The optimization can be applied in general for different Ops.
3. For this PR, should Index type be used for memcpy offset and length? I feel I64 is at LLVM level. Here the element type of input/output Memref is available. 
4. Alex already asked: How to handle the map in memref?

src/Conversion/ONNXToKrnl/Tensor/Transpose.cpp
tungld: Yes, I agree that it would be great to generalize this for other situations.

tungld:
> 4. Alex already asked: How to handle the map in memref?

The map must be identity at this moment. I need to check this.

src/Conversion/ONNXToKrnl/Tensor/Transpose.cpp
tungld: Yes, it's true. I need to check the map on the memrefs.

tungld:
> 3. For this PR, should Index type be used for memcpy offset and length? I feel I64 is at LLVM level. Here the element type of input/output Memref is available.

@chentong319 did you suggest that `offset` and `length` are in the number of elements instead of bytes? Then, we use the element type to calculate `offset` and `length` in bytes at LLVM level?

chentong319:
> > 5\. For this PR, should Index type be used for memcpy offset and length? I feel I64 is at LLVM level. Here the element type of input/output Memref is available.
> 
> @chentong319 did you suggest that `offset` and `length` are in the number of elements instead of bytes? Then, we use the element type to calculate `offset` and `length` in bytes at LLVM level?

Yes. At least offset which is the index of the starting element. Length can be the number of elements. Type I64 is OK.

src/Conversion/ONNXToKrnl/Tensor/Transpose.cpp
tungld: Done. I added code to verify that there is no map on the memrefs. Mentioned that in the comment too.

src/Conversion/ONNXToKrnl/Tensor/Transpose.cpp
tungld: I leave this as it is at this moment since different levels of lowering need different builders.

tungld:
> Yes. At least offset which is the index of the starting element. Length can be the number of elements. Type I64 is OK.

@chentong319 I changed the code according to your suggestion.

AlexandreEichenberger (approved):
LGTM pending Tong to approve changes as well

tungld:
@chentong319  I changed the code according to your suggestion. Please take a look. Thanks!

chentong319 (approved):
LGTM!

Issue #1958, opened 2023-01-14
IndexExpr support for Floats


philass:
Moving to the new implementation for `Resize` with `IndexExpr` caused a bug.

The calculation for a given rank `i` is `axis[i] = floor(scales[i] * inputDim[i])`. Unfortunately `IndexExprs` don't have support for floats. 

The current implementation doesn't do this rounding correctly as it casts the `scales[i]` to an int. This causes the wrong result. 

`IndexExprs` could add support for floats however this is quite a large change, that would either effect many functions or result in the introduction of many new functions that mirror their Int equivalents.

AlexandreEichenberger:
I saw that too when I was refactoring the code. I have to think a bit about it as indices are naturally about integers. But indeed ONNX has some computations that use float to compute an index, how unfortunate :-)

AlexandreEichenberger:
PR #1974 added support and changed code for resize accordingly.

PR #1974, opened 2023-01-25
Added support for float in index expressions

AlexandreEichenberger:
There are some float computations in index expressions, for example for resize where the new size is defined as
`(int)  ((float) input size * scale)`.

Thus I added support for float expressions in IndexExpr and IndexExprBuilder.

Floats can be found in literals, question marks (just for sanity checks), affine expr (but only as literals), and non-affine (as runtime  values).

Support was added for all relevant IndexExpr, but currently only a subset is used (multiply). Was easier to do it in one go. Every computations verify that there is never an illegal mix of float and int/index/predicate, again to make sure that values are used properly. There are 2 conversion operation in IndexExpr, from float to index integers, and back. Some operations don't support float (for example the ceil/floor divisions), others that don't support int (floor/ceil). It's all labeled in the header files.

There are new methods to read 1D arrays of values as float: `getFloatFromArrayAsNonAffine` and variants found in `IndexExprBuilders`, which automatically detect literal values.

Implement features requested in #1958


tungld (approved):
LGTM!

philass (commented):
Thanks for fixing this Alex! Much appreciated

PR #1976, opened 2023-01-25
Test suggestions

hamptonm1:


gongsu832:
After taking a closer look at the code, it appears that the `isnan` function is NOT called by the compiler itself but rather generated for the model runtime code. In that case, the IR is compiled by llc/opt which does NOT go through the normal C/C++ preprocessing. So macro wrappers like `isnan` should not be used. In this case, it's actually correct to use the internal library routines. On Linux, the routines are `__isnanf` (for F32) and `__isnan` (for F64), and on MacOS, the routines are `__isnanf` (for F32) and `__isnand` (for F64). So a patch like the following should work:
```
diff --git a/src/Conversion/KrnlToLLVM/KrnlUnaryMath.cpp b/src/Conversion/KrnlToLLVM/KrnlUnaryMath.cpp
index e8345dad..3fa75381 100644
--- a/src/Conversion/KrnlToLLVM/KrnlUnaryMath.cpp
+++ b/src/Conversion/KrnlToLLVM/KrnlUnaryMath.cpp
@@ -126,17 +126,16 @@ struct MathFunctionName<KrnlAtanhOp> {
 template <>
 struct MathFunctionName<KrnlIsNaNOp> {
   static std::string functionName(mlir::Type type) {
-
-#if (__APPLE__)
+    // Use internal functions rather than macro isnan since the generated IR
+    // does not go through normal C/C++ preprocessing.
     if (type.isF32())
       return "__isnanf";
+    if (type.isF64())
+#if (__APPLE__)
+      return "__isnand";
 #else
-    if (type.isF32())
-      return "isnanf";
+      return "__isnan";
 #endif
-
-    if (type.isF64())
-      return "isnan";
     llvm_unreachable("Unsupported type for isnan");
   }
 };
```

hamptonm1:
> After taking a closer look at the code, it appears that the `isnan` function is NOT called by the compiler itself but rather generated for the model runtime code. In that case, the IR is compiled by llc/opt which does NOT go through the normal C/C++ preprocessing. So macro wrappers like `isnan` should not be used. In this case, it's actually correct to use the internal library routines. On Linux, the routines are `__isnanf` (for F32) and `__isnan` (for F64), and on MacOS, the routines are `__isnanf` (for F32) and `__isnand` (for F64). So a patch like the following should work:
> 
> ```
> diff --git a/src/Conversion/KrnlToLLVM/KrnlUnaryMath.cpp b/src/Conversion/KrnlToLLVM/KrnlUnaryMath.cpp
> index e8345dad..3fa75381 100644
> --- a/src/Conversion/KrnlToLLVM/KrnlUnaryMath.cpp
> +++ b/src/Conversion/KrnlToLLVM/KrnlUnaryMath.cpp
> @@ -126,17 +126,16 @@ struct MathFunctionName<KrnlAtanhOp> {
>  template <>
>  struct MathFunctionName<KrnlIsNaNOp> {
>    static std::string functionName(mlir::Type type) {
> -
> -#if (__APPLE__)
> +    // Use internal functions rather than macro isnan since the generated IR
> +    // does not go through normal C/C++ preprocessing.
>      if (type.isF32())
>        return "__isnanf";
> +    if (type.isF64())
> +#if (__APPLE__)
> +      return "__isnand";
>  #else
> -    if (type.isF32())
> -      return "isnanf";
> +      return "__isnan";
>  #endif
> -
> -    if (type.isF64())
> -      return "isnan";
>      llvm_unreachable("Unsupported type for isnan");
>    }
>  };
> ```

Thanks for your response and validating my initial use of  `__isnanf`. Technically what I had in the PR that is merge is fine please look at this reference: https://discuss.tvm.apache.org/t/resolved-is-there-a-good-way-to-implement-isnan-op/4050

```
Float32 | Float64
-- | --
Win | _isnanf | _isnan/isnan
OSX | __isnanf | __isnand/isnan
Linux | isnanf | isnan
```

Meaning isnan and `__isnand` produces the same results and isnan and `__isnan` are also equivalent. So what is merged works and is accurate. However, if we want to update `isnan` to `__isnand`, I am perfectly fine with that. However, please understand it will produce the same result and is just as accurate. Thanks for reviewing the PR while on vacation!

gongsu832:
Are you sure `isnan` works for F64? The backend test apparently only tests for the F32 case. You can change the F64 case to any arbitrary function like `foo` and the backend test will still succeed.

hamptonm1:
> Are you sure `isnan` works for F64? The backend test apparently only tests for the F32 case. You can change the F64 case to any arbitrary function like `foo` and the backend test will still succeed.

Hmmm well I know for sure `isnanf` works for Linux for f32 and I would not need to change that to `__isnanf`. Based on the test results. So I can keep my same logic and update the return values to see what I get back based on your feedback.

hamptonm1:
So for `f64` Linux recognizes `__isnan` but not for windows. It fails. So `isnan` was recognized by both linux and windows. I can test f64 for the backend and see what happens.

hamptonm1:
This was just for testing...so closing this out now :)

gongsu832:
Why is it closed? This shows that the changes you made do not completely work for f64.

hamptonm1:
> Why is it closed? This shows that the changes you made do not completely work for f64.

It’s closed because I tested using your suggestions which didn’t work. So I kept what I had initially. I asked if we should discuss and you mentioned it was resolved. How would you like to proceed further?

hamptonm1:
From looking at the commit history the only thing that fails for the way I originally had it was the windows build for f64. So I am happy to discuss with you via a webex, zoom or huddle call when you are available. Feel free to propose a time that works best for you.

gongsu832:
So we need to fix it for f64 on Windows. I don't know much about windows. You can consult @sstamenova for windows related problems.

PR #1981, opened 2023-01-31
onnx-mlir: replace binary invocations of opt and llc with library calls

ashay:
This patch contains five commits, the last of which makes the key change.

Prior to this patch, onnx-mlir and onnx-mlir-opt invoked the opt and llc
binaries, which required these binaries and all LLVM libraries to be
copied to the same directory where onnx-mlir artifacts are installed.
This also created problems in shared library builds, since copying the
binaries and libraries does not update the RPATH embedded in them.

To resolve these problems, this patch replaces invocations of the opt
and llc binaries in CompilerUtils.cpp with calls to LLVM APIs for
running the optimizer and the assembler.  As a minor ancillary benefit,
this change allows us to skip writing the unoptimized bitcode file to
disk, since the same `llvm::Module` object can be carried over to the
optimizer code as well as the assembler code.

A downside of this change is that we can no longer pass arbitrary
options to opt and llc, like we previously could using the `Xopt`,
`Xllc`, and `mllvm` flags to onnx-mlir and onnx-mlir-opt.


sstamenova:

ashay:
Fixed library dependencies.  Also, onnx-mlir now outputs a bitcode file, regardless of the platform, so I added a line to dump the LLVM IR from the bitcode file before checking it against the reference output.

AlexandreEichenberger:

AlexandreEichenberger:
> A downside of this change is that we can no longer pass arbitrary
options to opt and llc, like we previously could using the Xopt,
Xllc, and mllvm flags to onnx-mlir and onnx-mlir-opt.

@ashay are there likely options that we are currently using that are not possible anymore, or some that we may miss? Is it a matter of adding them to the main onnx-mlir driver? Or can we compile in multiple steps to circumvent this limitation? Would be great if you could elaborate a bit, thanks.

src/Compiler/CompilerUtils.cpp
```diff
@@ -373,96 +364,119 @@ std::string getTargetFilename(
   llvm_unreachable("all cases should be handled in switch");
 }
 
+// Apply optimization passes specified by `optPipeline` on the LLVM IR module
+// for the specified target machine and write the resulting bitcode to
+// `outputStream`.
+static int runOptPasses(llvm::Module *llvmModule, llvm::TargetMachine *target,
```
AlexandreEichenberger, line 370: Is there a way to run Opt Passes without having to specify each of the passes here? My concern is that if an additional pass was added in the main Opt, we will not see it here unless we also update the code below accordingly. So isn't there a way to call opt using a single function call (where detailed of the options are then maintained by Opt itself)?

ashay: Unfortunately, there isn't a single function call to replace the binary invocation of opt.  The passes that run in the opt binary are listed in its [500-line long `main()` function](https://github.com/llvm/llvm-project/blob/main/llvm/tools/opt/opt.cpp#L420), and over [multiple files](https://github.com/llvm/llvm-project/tree/main/llvm/tools/opt) in that directory.

So you're right that if a new pass is added to opt, we'd miss it in the function that I added.  Let me quickly check if anyone has expressed interest in refactoring the code for the opt and llc binaries.

ashay:
> are there likely options that we are currently using that are not possible anymore, or some that we may miss?

Based on the tests, it seems like there is only one option (`--data-sections`) that we have passed via the `Xllc`, `Xopt`, and `mllvm` flags, but the test that passes this `--data-sections` option (test/mlir/driver/check_passthrough_options.mlir) doesn't depend on the option for functionality (it only checks whether this option is forwarded from onnx-mlir to opt and llc).

ashay:
Based on some cursory inquiries, it doesn't seem like the code for the opt and llc binaries is going to be refactored soon (so that we could mimic their functionality using a single function call).

It seems I also underestimated the extent of changes necessary to support multiple platforms.  I am closing this PR now, and I will revisit it later if it becomes a pressing issue.

PR #1984, opened 2023-01-31
Add Shape Inference, Lowering and backend test for IsInf

hamptonm1:
`IsInf` is very similar to `IsNaN`, so I was able to add the shape inference, lowering of the kernel, backend test, lit test and onnx-lowering test for this PR. I also updated the documentation and the `gen_onnx_mlir.py`.


hamptonm1:
Everything works except for the `test_isinf_negative_cpu` and `test_isinf_positive_cpu` backend tests are failing. I currently commented these two tests out and I can create a separate issue to further investigate why they are failing but `test_isinf` passes as expected.  Let me know if that is okay with everyone :)

src/Conversion/KrnlToLLVM/KrnlUnaryMath.cpp
```diff
@@ -123,6 +123,24 @@ struct MathFunctionName<KrnlAtanhOp> {
   }
 };
 
+template <>
+struct MathFunctionName<KrnlIsInfOp> {
+  static std::string functionName(mlir::Type type) {
+
+#if (__APPLE__)
```
AlexandreEichenberger, line 130: nit, could you put the #if inside of the `type.isF32()` test? It would read much better, as it is only the string that differs, not the test.

src/Dialect/Krnl/Krnl.td
```diff
@@ -12,7 +12,7 @@
 // When you perform a change to the KRNL dialect, please follow the steps below
 // to update the Krnl dialect documentation.
 //
-// 1) cd onnx-mlir/build 
+// 1) cd onnx-mlir/build
```
AlexandreEichenberger, line 15: I am guilty of the same all the time, so this is a memo to myself too. In the future, it is better to have  non-functional changes like spelling/spaces updates in a separate PR: one for non-functional changes that folks can approve very quickly, and one for the functional changes where reviewers pay more attentions to the actual changes.

src/Conversion/KrnlToLLVM/KrnlUnaryMath.cpp
hamptonm1: Sure let me update it

src/Dialect/Krnl/Krnl.td
hamptonm1: Ohh this was by accident...my visual studio code fixes certain white spaces and indentation once I save the file. Sorry about that :)

src/Conversion/KrnlToLLVM/KrnlUnaryMath.cpp
hamptonm1: It does not look aesthetically pleasing once I switch the order but I understand the need for consistency....so I made the update.

src/Dialect/Krnl/Krnl.td
hamptonm1: But here the whitespace was removed on line 15....so I think we should be fine

hamptonm1:
@chentong319 Any thoughts?

tungld (approved):
LGTM. Thanks!

test/backend/inference_backend.py
```diff
@@ -427,9 +427,12 @@ def get_test_models():
...
 
-        # IsInf
+        # ==OP== IsInf
+        "test_isinf_cpu": {STATIC_SHAPE:{}, DYNAMIC_SHAPE:{-1:{-1}}, CONSTANT_INPUT:{-1}},
+        # "test_isinf_negative_cpu": {STATIC_SHAPE:{}, DYNAMIC_SHAPE:{-1:{-1}}, CONSTANT_INPUT:{-1}},
+        # "test_isinf_positive_cpu": {STATIC_SHAPE:{}, DYNAMIC_SHAPE:{-1:{-1}}, CONSTANT_INPUT:{-1}},
 
-        # ==OP== IsNan
+        # ==OP== IsNaN
         "test_isnan_cpu": {STATIC_SHAPE:{}, DYNAMIC_SHAPE:{-1:{-1}}, CONSTANT_INPUT:{-1}},
```
chentong319, line 436: How about test_isinf_negative_cpu and test_isinf_positive_cpu?

hamptonm1: @chentong319 I have a separate issue open to debug those two backend test. Will keep you posted :)

PR #1977, opened 2023-01-26
(NFC) Upgrade dependencies to "green commits" for week of 01-23-2023

christopherbate:
Upgrades llvm-project and mlir-hlo project dependencies to the commits from 01-23-2022 per https://github.com/llvm/torch-mlir/issues/1178.

LLVM-Project: 9acc2f37bdfce08ca0c2faec03392db10d1bb7a9 
Mlir-hlo: 4a173356bb1291b97046545429d7851cbc771d88

The major changes in the onnx-mlir codebase consist of:

- All generated accessor methods in operations move to prefixed form (e.g. `op.input()` -> `op.getInput()`).
- Hard-coded uses of `-1` to represent a dynamic dimension in the context of a `ShapedType` (e.g. `TensorType`, `MemRefType`, and derivatives) are removed and replaced with `ShapedType::kDynamic`.
- Replace uses of `mlir::None` with `std::nullopt`
- Replace uses of `llvm::None` with `std::nullopt`.
- Replace uses of `llvm::makeArrayRef` with `llvm::ArrayRef`

Warnings not yet fixed:

  - Warnings due to MLIR `fold` API migration

I labeled this as NFC because it's just trying to upgrade the llvm dependency and there's no functional change. But it affects nearly every file due to getter/setter prefix change in upstream that makes MLIR no longer support the non-prefixed form.

PR #1985, opened 2023-01-31
Shapehelper infra used in Dim Analysis

AlexandreEichenberger:
This PR uses the ShapeHelper interface to return the shape helpers used for the analysis. We use to detect each op, and then call that op's shape helper. Now it is done via the interface.

There is another bit of cleanup: we had 2 lists of unsupported ops, one for shape inference and one for shape helper. This replication adds to the work needed to be done when adding new operators. There is now one instance: `src/Dialect/ONNX/ONNXUnsupportedOps.hpp` with a macro to generate the right content in the right context.

I also verified that each ONNX ops that has the shape helper interface did indeed have the code associated with it.


AlexandreEichenberger:
Please do not merge into main until the LLVM PR goes through (#1977).

src/Transform/ONNX/ONNXDimAnalysis.cpp
```diff
@@ -329,56 +311,63 @@ void DimAnalysis::visitDim(
...
+      isa<ONNXModOp>(op) ||
+      isa<ONNXMulOp>(op) ||
+      isa<ONNXOrOp>(op) ||
+      isa<ONNXPowOp>(op) ||
+      isa<ONNXSubOp>(op) ||
+      isa<ONNXSumOp>(op) ||
+      isa<ONNXWhereOp>(op) ||
+      isa<ONNXXorOp>(op)) {
+    // c lang-format on
+    if (exploreSameInputDims(dim, op, sameDims)) {
```
tungld, line 344: It seems the logic is different from the old code.

In the old code, the following code is executed even if `exploreSameInputDims` failed.

src/Transform/ONNX/ONNXDimAnalysis.cpp
```diff
@@ -329,56 +311,63 @@ void DimAnalysis::visitDim(
...
+      isa<ONNXMinOp>(op) ||
+      isa<ONNXModOp>(op) ||
+      isa<ONNXMulOp>(op) ||
+      isa<ONNXOrOp>(op) ||
+      isa<ONNXPowOp>(op) ||
+      isa<ONNXSubOp>(op) ||
+      isa<ONNXSumOp>(op) ||
+      isa<ONNXWhereOp>(op) ||
+      isa<ONNXXorOp>(op)) {
+    // c lang-format on
```
tungld, line 343: Typos: `clang`

src/Transform/ONNX/ONNXDimAnalysis.cpp
AlexandreEichenberger: @tungld so you are telling me the test can be performed regardless of if we successfully ran the check? Note that is is known to work unconditionally for all these ops, its more like unsupported ops that would fail. But I can remove the test if you feel more comfortable with that.

src/Transform/ONNX/ONNXDimAnalysis.cpp
AlexandreEichenberger: tx, fixed in next commit

src/Transform/ONNX/ONNXDimAnalysis.cpp
AlexandreEichenberger: Changed the `if (explore...)` to `assert(explore...)` since they are guaranteed to work for these ops. Prefer checking my assumptions that just hoping it should work.

tungld (approved):
LGTM! Thanks for the patch! Hope that many models benefit from this.

PR #1997, opened 2023-02-03
[non functional changes] Update of meeting calling info.

AlexandreEichenberger:
Right now, the meeting is hardwired in the README page. I moved it to the wiki page, with only a reference here to the wiki page. Easier to maintain.


tungld (approved):
LGTM.

PR #1999, opened 2023-02-03
Stop unknown dimension analysis when a dimension comes from a constant

tungld:
There is a bug in the unknown dimension analysis where it tried to explore a dimension of a ONNXConstantOp.
In such case, a dimension is a constant, so we can stop the analysis.


AlexandreEichenberger (approved):
LGTM

Issue #2001, opened 2023-02-06
Models failing to download - blocking builds & PRs


sstamenova:
It looks like several of the backend tests that rely on model download are failing and it is impacting both local builds as well as the CI for PRs. For example:

https://github.com/onnx/onnx-mlir/pull/1993

```
gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now

gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
make[3]: *** [test/backend/CMakeFiles/download_model_for_test_inception_v1.dir/build.make:61: test/backend/Debug/models/inception_v1.tar.gz] Error 2
make[3]: *** Deleting file 'test/backend/Debug/models/inception_v1.tar.gz'
make[2]: *** [CMakeFiles/Makefile2:17407: test/backend/CMakeFiles/download_model_for_test_inception_v1.dir/all] Error 2
```

I'm not really familiar with this set up, so I am not sure whether we can reach out to the owners and figure out what's wrong or if it would be better to disable some tests temporarily.

sstamenova:
@gongsu832 : Can you advise here?

philass:
https://github.com/onnx/onnx/issues/4857

sstamenova:
@philass : It sounds like we should disable the tests from the onnx-mlir side as well unless we're going to update onnx. Can you do that or should I have a look? Thanks!

philass:
Sorry, just saw your message now.

I reviewed the PR. I think it was/is the correct course of action

sstamenova:
Thank you @gongsu832

sstamenova:
Looks like they have a solution in `onnx` to get models from a new location: https://github.com/onnx/onnx/issues/4857#event-8462803624

gongsu832:
> Looks like they have a solution in `onnx` to get models from a new location: [onnx/onnx#4857 (comment)](https://github.com/onnx/onnx/issues/4857#event-8462803624)

OK that simplify things. Once that PR is merged upgrading `third_party/onnx` should allow the backend model tests to pass again.

sstamenova:
onnx 1.13.1 is now available: https://github.com/onnx/onnx/releases/tag/v1.13.1

gongsu832:
@philass Are you OK with upgrading to 1.13.1?

cjvolzka:
[...]

PR #2003, opened 2023-02-06
Disable failing onnx models from the backend tests

sstamenova:
Based on the information available from `onnx`, access to these has been lost for the moment (possibly forever?). Until there's a better solution, disable them to unblock everyone.

See https://github.com/onnx/onnx-mlir/issues/2001 and https://github.com/onnx/onnx/issues/4857 for more information


philass (approved):
Thank you for looking deeper into this.

I think this is the appropriate action

PR #1902, opened 2022-12-07
DRAFT: constant folding implementation with DisposableElementsAttr

sorenlassen:
This draft demonstrates how to do constant folding with DisposableElementsAttr from PR #1874.

I merged the constant-materializer branch from PR #1799.

On top of that I implemented fold for ONNXTransposeOp.

I had to comment out the lit tests in onnx_canonicalization.mlir where constants ended up with dynamic types (@test_shape1, @test_size1, @test_constant_1). I'm not sure how to address this properly.

The test_if_cpu backend test fails with a malloc error, maybe the same memory is freed twice. Not sure why.
```
cmake --build . --target check-onnx-backend
TEST_VERBOSE=1 TEST_CASE_BY_USER=test_if_cpu python3 test/backend/Debug/test.py
...
Python(94900,0x1e5706500) malloc: *** error for object 0x10559fb50: pointer being freed was not allocated
```

PR #2009, opened 2023-02-08
Add constant fold to onnx dialect

chentong319:
This PR is the new version of #1799 on the current code base to avoid handling the merging conflicts.
This PR enables the Fold support in onnx Dialect. 
1. Dialect materializer support
2. Add fold function to SqueezeOp as an example
3. Update the lit test. A new test case is added. For SqueezeOp, the canonicalization can do what constprop does previously. The existing test is updated because the folding in canonicalization moves all the constants to the beginning of the function, and inconsistent constantOp in previous test cases cannot work anymore.

Details:
1. Fold is called automatically by canonicalization. We can move most of the work in onnx-mlir ConstProp pass, except for restructuring tree for commutative Ops, to fold function. The benefit is to use `OpBuilder::createOrFold` when Ops are generated in transformation.
2. There is some overlapping between our constant propagation and shape inference: we assume that the shape inference is done. constant propagation may use the type of Op output to generate  output constant. This is not true for canonicalization with fold. In canonicalization of ConvOp, the generated SqueezeV11Op has unranked tensor output.
3. Fold function also requires that the attribute of a ConstantOp is consistent with its output.


src/Dialect/ONNX/ONNXOps/Tensor/Squeeze.cpp
```diff
@@ -183,3 +183,39 @@ namespace onnx_mlir {
 template struct ONNXCommonSqueezeOpShapeHelper<ONNXSqueezeOp>;
 template struct ONNXCommonSqueezeOpShapeHelper<ONNXSqueezeV11Op>;
 } // namespace onnx_mlir
+
+//===----------------------------------------------------------------------===//
+// Folder
+//===----------------------------------------------------------------------===//
+OpFoldResult ONNXSqueezeOp::fold(FoldAdaptor adaptor) {
+  OnnxElementsAttrBuilder elementsBuilder(getContext());
```
sorenlassen, line 191: better to move it down just before it's used, so it's only created when used

src/Dialect/ONNX/ONNX.td
```diff
@@ -41,6 +41,8 @@ def ONNX_Dialect : Dialect {
   let useDefaultAttributePrinterParser = 0;
   let useFoldAPI = kEmitFoldAdaptorFolder;
   let dependentDialects = ["func::FuncDialect"];
+  let useFoldAPI = kEmitFoldAdaptorFolder;
```
sorenlassen, line 44: duplicated, already appears two lines above

src/Dialect/ONNX/ONNXOps/Tensor/Squeeze.cpp
```diff
@@ -183,3 +183,39 @@ namespace onnx_mlir {
...
+  if (!adaptor.getData() || !adaptor.getAxes()) {
+    // Use original Op if Data or Axes is not constant
+    return nullptr;
+  }
+
+  // ToFix: copied from constprop.cpp.
```
sorenlassen, line 197: this is the way to do it and there's nothing to fix, right? i.e. we shouldn't have "ToFix" in the comment

if the result shape is not static then we just cannot fold, because the constant folding breaks down if a constant op has non-static shape, as it must derive its type from the attribute, which always has static type

src/Dialect/ONNX/ONNXOps/Tensor/Squeeze.cpp
```diff
@@ -183,3 +183,39 @@ namespace onnx_mlir {
...
+  // Constant propagation pass always follows shape inference,
+  // and assumes the shape is inferenced.
+  // But the canonicalization of ConvOp may generate SqueezeOp with
+  // unranked tensor output. Type checking is needed here.
+  auto ty = getOperation()->getResults()[0].getType().dyn_cast<ShapedType>();
+  if (!(ty && ty.hasStaticShape()))
+    return nullptr;
+
```
sorenlassen, line 205: simpler: `if (!hasStaticShape(getSqueezed().getType()) return nullptr;`

src/Dialect/ONNX/ONNXOps/Tensor/Squeeze.cpp
chentong319: I am thinking of calling Shape inference here to fix it.

sorenlassen: the issue is that if we change the operation's result type then things go hay-wire, especially if it's the last op in a function body then MLIR will complain that the return result (operand of the return op) doesn't match the function's return type, so we will need to propagate the type change to some of the users of the return value, which might have knock-on effects wherever we canonicalize ops

maybe we should do that: change the canonicalize pass etc to do the necessary shape inference and shape propagation to support constant folding, but I can't predict how many obstacles we'll encounter if we try that

chentong319: I tried to add the shape inference. It works for my test case. I will check the last op issue you mentioned.
Now `onnx-mlir-opt --canonicalize` of the following model (in onnx_canonicalization.mlir)
```
func.func @test_fuse_add_conv(%arg0 : tensor<1x1x28x28xf32>, %arg1 : tensor<8x1x5x5xf32>) -> tensor<1x8x28x28xf32> {
    %0 = "onnx.Conv"(%arg0, %arg1, %cst) {auto_pad = "SAME_UPPER", dilations = [1, 1], group = 1 : si64, kernel_shape = [5, 5], onnx_node_name = "Convolution28", strides = [1, 1]} : (tensor<1x1x28x28xf32>, tensor<8x1x5x5xf32>, none) -> tensor<1x8x28x28xf32>
    %1 = onnx.Constant dense<[[[-0.161539719]], [[-0.433835655]], [[0.091641359]], [[-0.0168522168]], [[-0.0650264397]], [[-0.131737873]], [[0.0204175506]], [[-0.121110231]]]> : tensor<8x1x1xf32>
    %2 = "onnx.Add"(%0, %1) : (tensor<1x8x28x28xf32>, tensor<8x1x1xf32>) -> tensor<1x8x28x28xf32> 
    return %2 : tensor<1x8x28x28xf32> 
}
```
Will generate the result 
```
module {
  func.func @test_fuse_add_conv(%arg0: tensor<1x1x28x28xf32>, %arg1: tensor<8x1x5x5xf32>) -> tensor<1x8x28x28xf32> {
    %0 = onnx.Constant dense<[-0.161539719, -0.433835655, 0.091641359, -0.0168522168, -0.0650264397, -0.131737873, 0.0204175506, -0.121110231]> : tensor<8xf32>
    %1 = "onnx.Conv"(%arg0, %arg1, %0) {auto_pad = "SAME_UPPER", dilations = [1, 1], group = 1 : si64, kernel_shape = [5, 5], strides = [1, 1]} : (tensor<1x1x28x28xf32>, tensor<8x1x5x5xf32>, tensor<8xf32>) -> tensor<1x8x28x28xf32>
    return %1 : tensor<1x8x28x28xf32>
  }
}
```
Same as `onnx-mlir-opt --canonicalize --shape-inference --constprop-onnx` with implicit pass ordering removed.

AlexandreEichenberger (commented):
@chentong319 high level questions: should we switch in the code places where we have `create` to `createOrFold`? I am thinking for example the the dialect builder, where this would catch most cases.

As is, the pattern that handled the folding does not seems to have been removed, as you advertised in the conversation header. Is it because there are still `create`? If so, does it mean that unless we replace all `create` to `createOrFold` we should hold on to the folding patterns?

Thanks for clarifying this for others to better understand what will be needed for the other ops.

chentong319:
> @chentong319 high level questions: should we switch in the code places where we have `create` to `createOrFold`? I am thinking for example the the dialect builder, where this would catch most cases.
> 

I  think so. The most needed case is in onnx to krnl lowering.

> As is, the pattern that handled the folding does not seems to have been removed, as you advertised in the conversation header. Is it because there are still `create`? If so, does it mean that unless we replace all `create` to `createOrFold` we should hold on to the folding patterns?

Keep the Constant Propagation for a while. At least it contains the tree restructuring for commutative operations.
> 
> Thanks for clarifying this for others to better understand what will be needed for the other ops.

chentong319:
@sorenlassen Sorry I forgot that you already have the draft PR #1902 for folding. My PR is more of the experiment with SqueezeOp. I will close this  PR.  However, there are several changes you may be useful for your PR.

1. Add Shape Inference  in fold. 
2. Update the const propagation for SizeOp.
3. Some lit test cases are not correct.
4. I did not add any change to the disposableDenseElementAttr.

sorenlassen:
@chentong319 sorry for the confusion, I'm very supportive of your work in this PR and encourage you to continue, I wasn't actively working on constant folding and just closed PR #1902 to make that clear

what I meant to say in today's meeting is that I've been thinking about the underlying issues and I'd like to help, if you share details about the obstacles you encounter

chentong319:
The current issue I have is runtime error in python for test_if_cpu in check-onnx-backend. Other questions I have are:
1. Does createOrFold work lowering onnx to krnl when unrealized cast or bufferization.to_mem/to_tensor is added? If not,  what's the solution?
2. Do we need to.keep all the different attribute for ONNXConstantOP, or just use sparseValue and Value?
3. Any change needed in the disposableDenseElementsAttr for fold as your PR name indicated?

@sorenlassen Can we schedule an online meeting to discuss how to combine the two PRs if you like to? I am available before 4 pm today.

sorenlassen:
> Can we schedule an online meeting to discuss how to combine the two PRs

yes, that sounds good

sorry I didn't respond to your message earlier - which timezone are you in? I'm available today from now until late (5:30pm-9pm PST in California) or tomorrow after 4pm PST

sorenlassen:
> 2\. Do we need to.keep all the different attribute for ONNXConstantOP, or just use sparseValue and Value?

I think it will be ok to just use sparseValue and Value

I think we assume in many places that only Value is used

it might make sense in the future to support sparseValue

the others can hopefully be canonicalized to Value, if we don't already

sorenlassen:
> 3\. Any change needed in the disposableDenseElementsAttr for fold as your PR name indicated?

No.

The reason I wrote PR #1902 was because @tungld asked in the Nov 29 meeting how DisposableElementsAttr would interact with constant folding: https://github.com/onnx/onnx-mlir/wiki/Informal-meeting-agenda-and-notes#nov-29th-2022
The PR was meant to demonstrate that DisposableElementsAttr will work with constant folding (except I ran into the same issues as you with the test_if_cpu backend test and ONNXSizeOp etc).

(See: https://github.com/onnx/onnx-mlir/pull/1874#issuecomment-1339633488)

chentong319:
> > Can we schedule an online meeting to discuss how to combine the two PRs
> 
> yes, that sounds good
> 
> sorry I didn't respond to your message earlier - which timezone are you in? I'm available today from now until late (5:30pm-9pm PST in California) or tomorrow after 4pm PST

I am on east coast. How about 5:30 pm PST today(Thursday) https://ibm.webex.com/meet/chentong90?

sorenlassen:
> How about 5:30 pm PST today(Thursday)

that works, I'll see you then

src/Dialect/ONNX/ONNXOps/Tensor/Constant.cpp
```diff
@@ -70,3 +70,27 @@ LogicalResult ONNXConstantOp::inferShapes(
...
+    return getValueIntAttr();
+  else if (getValueIntsAttr())
+    return getValueIntsAttr();
+  else if (getValueFloatsAttr())
+    return getValueFloatsAttr();
+  else {
+    assert(getValueStringAttr() &&
+           "ONNXConstantOp does not have a valid attribute");
+    return getValueStringAttr();
+  }
```
sorenlassen, line 95: we should also check getValueStringsAttr(), right?

sorry for trying to rewrite the whole method but I'd prefer if we write it as:
```
if (auto sparseValue = getSparseValueAttr()) return sparseValue;
if (auto value = getSparseValueAttr()) return value;
...
if (auto valueStrings = getValueStringsAttr()) return valueStrings;
llvm_unreachable("ONNXConstantOp does not have a valid attribute");
```
so it's consistent and symmetrical for ease of reading, and we don't look up any attribute twice

src/Dialect/ONNX/ONNXOps/Tensor/Squeeze.cpp
```diff
@@ -183,3 +183,46 @@ namespace onnx_mlir {
...
+//===----------------------------------------------------------------------===//
+// Folder
+//===----------------------------------------------------------------------===//
+OpFoldResult ONNXSqueezeOp::fold(FoldAdaptor adaptor) {
+  // Fold type
+  if (failed(inferShapes(nullptr)))
+    return nullptr;
+
+  // Fold value
+
```
sorenlassen, line 196: please remove blank line

src/Dialect/ONNX/ONNXOps/Tensor/Squeeze.cpp
```diff
@@ -183,3 +183,46 @@ namespace onnx_mlir {
...
+  assert(hasStaticShape(getSqueezed().getType()) &&
+         "Shape should be static when the inputs are constant");
+
+  OnnxElementsAttrBuilder elementsBuilder(getContext());
+  return elementsBuilder.reshape(
+      adaptor.getData(), getShape(getSqueezed().getType()));
+}
+
+OpFoldResult ONNXSqueezeV11Op::fold(FoldAdaptor adaptor) {
+
```
sorenlassen, line 211: please remove blank line

test/mlir/onnx/onnx_canonicalization.mlir
```diff
@@ -96,78 +96,50 @@ func.func @cast_elimination(%arg0: tensor<2xf32>) -> tensor<2xf32> {
 
 // -----
 
-func.func @test_conv_batchnormtestmode_fusion_nobias(%arg0 : tensor<1x3x224x224xf32>) -> tensor<1x64x112x112xf32> {
+func.func @test_conv_batchnormtestmode_fusion_nobias(%arg0: tensor<1x3x224x224xf32>, %0: tensor<64x3x7x7xf32>, %2: tensor<64xf32>, %3: tensor<64xf32>, %4: tensor<64xf32>, %5: tensor<64xf32>) -> tensor<1x64x112x112xf32> {
```
sorenlassen, line 99: any reason to break with the convention to name function args `%arg1` etc?

same in line 123 below

test/mlir/onnx/onnx_fold.mlir
```diff
@@ -0,0 +1,32 @@
+// RUN: onnx-mlir-opt --shape-inference --canonicalize %s -split-input-file | FileCheck %s
```
sorenlassen, line 1: what happens if you omit --shape-inference?

src/Accelerators/NNPA/Dialect/ZHigh/ZHigh.td
```diff
@@ -28,6 +28,7 @@ def ZHigh_Dialect : Dialect {
   let summary = "A high-level dialect for the ONNX NNPA accelerator ISA.";
   let cppNamespace = "::onnx_mlir::zhigh";
   let useDefaultAttributePrinterParser = 1;  
+  let useFoldAPI = kEmitFoldAdaptorFolder;
```
sorenlassen, line 31: this change and the corresponding change to ZLow looks unrelated, should they be omitted from this PR?

src/Dialect/ONNX/ONNXOps.cpp
```diff
@@ -211,6 +213,17 @@ ParseResult ONNXConstantOp::parse(OpAsmParser &parser, OperationState &result) {
...
+Operation *ONNXDialect::materializeConstant(
+    OpBuilder &builder, Attribute value, Type type, Location loc) {
+
+  // The atrribute could be DenseElemnemntsAttr, IntAttr, FloatAttr and etc.
+  // onnx builder is used to convert it into value()
+  OnnxBuilder onnx(builder, loc);
+  Value result = onnx.constant(value);
+  return result.getDefiningOp();
+}
+
```
sorenlassen, line 226: please move to the end of the file so that all the print and parse methods are next to each other

and please follow the file convention and add`//===-----...` etc to the comment above the method

sorenlassen (approved):
Nice!

Just some stylistic comments

tungld:
> Update the lit test. A new test case is added. For SqueezeOp, the canonicalization can do what constprop does previously

Should we remove the rule for SqueezeOp in `constprop` since its `fold` does the same thing?

chentong319:
> > Update the lit test. A new test case is added. For SqueezeOp, the canonicalization can do what constprop does previously
> 
> Should we remove the rule for SqueezeOp in `constprop` since its `fold` does the same thing?

We will move most of the constant propagation into fold in future.

PR #2010, opened 2023-02-09
Add a requirements.txt file for PyPi requirements in onnx-mlir

sstamenova:
This adds a `requirements.txt` file at the root of the repository with the current PyPi package requirements and pinned versions for them. This is the equivalent of the mlir change that pinned their package requirements (https://reviews.llvm.org/D142563). 

This change is pinning the requirements to a specific version (or a range) depending on the requirement. A couple of considerations:

* numpy 1.24 deprecates np.object, np.bool, np.float, np.complex, np.str, and np.int which are used heavily in onnx-mlir
* not all versions of each package are available on every platform - to the best of my knowledge, these ranges should work on Ubuntu, CentOS, macOS and Windows

Adding a minimum and maximum version, or pinning to a specific versions where possible, helps with two major goals - security and maintainability. It gives us an opportunity to make sure that the packages being used are not part of a security attack as well as guaranteeing that they support the features that onnx-mlir depends on (see note about numpy deprecation).

This also adds the requirements.txt file to the setup instructions and to the pipelines on Windows and macOS. I've not added it to the docker images but it can be used there as well with a little more effort.

### DIFF
--- a/.github/workflows/macos-amd64-build.yml
+++ b/.github/workflows/macos-amd64-build.yml
@@ -39,8 +39,8 @@ jobs:
     - name: install python requirements
       run: |
         cd ~/work/onnx-mlir/onnx-mlir
-        pip3 install --upgrade wheel
-        pip3 install -r requirements.txt
+        python3 -m pip install --upgrade wheel
+        python3 -m pip install -r requirements.txt
     - name: install third_party/onnx
       run: |
         cd ~/work/onnx-mlir/onnx-mlir/third_party/onnx
--- a/.github/workflows/macos-amd64-build.yml
+++ b/.github/workflows/macos-amd64-build.yml
@@ -36,11 +36,15 @@ jobs:
         cd ~/work/onnx-mlir
         sh ~/work/onnx-mlir/onnx-mlir/utils/clone-mlir.sh
         sh ~/work/onnx-mlir/onnx-mlir/utils/build-mlir.sh
+    - name: install python requirements
+      run: |
+        cd ~/work/onnx-mlir/onnx-mlir
+        pip3 install --upgrade wheel
+        pip3 install -r requirements.txt
     - name: install third_party/onnx
       run: |
         cd ~/work/onnx-mlir/onnx-mlir/third_party/onnx
         git fetch --prune --unshallow --tags
-        pip3 install numpy==1.23.5
         python3 -m pip install -v .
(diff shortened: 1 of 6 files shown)


gongsu832:
I think it's fine to provide `requirements.txt` as an optional way to install the prereqs. But for docker builds on Linux, it's not going to be used, at least for now. Because prereqs like numpy are not installed with pip due to the fact that on s390x these packages are not prebuilt so installing them with pip requires compiling from source therefore increases build time. So on s390x Linux these packages are installed with system package manager to use the prebuilt binaries to save time.

.github/workflows/macos-amd64-build.yml
```diff
@@ -36,11 +36,15 @@ jobs:
         cd ~/work/onnx-mlir
         sh ~/work/onnx-mlir/onnx-mlir/utils/clone-mlir.sh
         sh ~/work/onnx-mlir/onnx-mlir/utils/build-mlir.sh
+    - name: install python requirements
+      run: |
+        cd ~/work/onnx-mlir/onnx-mlir
+        pip3 install --upgrade wheel
```
NathanielMcVicar, line 42: I don't know the details of macOS pip, but it seems odd to use `pip3` here and `python3 -m pip` below.

NathanielMcVicar (approved):

.github/workflows/macos-amd64-build.yml
hamptonm1: I agree, we should be consistent....maybe we should lean towards using `python3 -m pip`. Seems like python3 -m pip ensures that we are using the python interpreter specified. I think the reason why we are using `python3 -m pip install -v .` on line 48 is to install pip for the `third_party/onnx` directory but if line 43 worked previously....I do not see the need for this.

AlexandreEichenberger:
> I think it's fine to provide requirements.txt as an optional way to install the prereqs. But for docker builds on Linux, it's not going to be used, at least for now. Because prereqs like numpy are not installed with pip due to the fact that on s390x these packages are not prebuilt so installing them with pip requires compiling from source therefore increases build time. So on s390x Linux these packages are installed with system package manager to use the prebuilt binaries to save time.

@gongsu832 does that apply to docker builds for Linux for Power and amd64?

gongsu832:
All Linux platforms share the same docker build files. We are not going to special case s390x.

Issue #1940, opened 2022-12-22
[NNPA] Wrong access index for layout 1D and 2DS


tungld:
In NNPA code, generated data access index for zTensor of layout 1D and 2DS was wrong.

For example, [current generated data access index for 1D](https://github.com/onnx/onnx-mlir/blob/main/test/mlir/accelerators/nnpa/conversion/zhigh-to-zlow/test-datalayout.mlir#L7) is:
```
affine_map<(d0) -> (0, d0 floordiv 64, 0, 0, 31, d0 mod 64)>
```

The correct one should be 
```
affine_map<(d0) -> (0, d0 floordiv 64, 0, 0, 0, d0 mod 64)>
```
where all dimension indices except `d0` must be 0.

Same for 2DS, [current generated data access index](https://github.com/onnx/onnx-mlir/blob/main/test/mlir/accelerators/nnpa/conversion/zhigh-to-zlow/test-datalayout.mlir#L37) is:
```
affine_map<(d0, d1) -> (d0, d1 floordiv 64, 0, 0, 31, d1 mod 64)>
```

while, the correct one should be:
```
affine_map<(d0, d1) -> (d0, d1 floordiv 64, 0, 0, 0, d1 mod 64)>
```

This issue will affect the accuracy of two NNPA operations: `Convolution`  (1D is used for bias) and `MatMul` (1D and 2DS are used for bias)

cjvolzka:
Is this resolved by https://github.com/onnx/onnx-mlir/pull/2011 or is there more to be done here?

AlexandreEichenberger:
A possibility is to extend 1D to 2D with size 1 in the first dim, or 2DS to 3DS with size 1 in the second dim. Tried to see if used in a simple "add" example, both 1D and 2DS where sent to CPU at this time... so we probably not missing much.

PR #2011, opened 2023-02-09
[NNPA] Data transformation directly on stickified data

tungld:
This patch is to rewrite `affine.load` and `affine.store` so that they will read/write data directly from/to stickified data.
Thus, there is no need to stickify and unstickify data.

This can handle data transformation such as: transpose, concat, split, etc. or any `affine.for` loop whose `load` and `store` are using affine map to read/write data.

Please see LIT tests for various situations.

# Pattern to rewrite
 ```
 zlow.unstick -> affine.for (affine.load -> affine.store) -> zlow.stick
    |                            |
    |                            '--------> affine.store) -> zlow.stick
    |
    '----------> affine.for (affine.load -> affine.store) -> zlow.stick
                                                              ^
 zlow.unstick -> affine.for (affine.load -> affine.store) ----'
 ```

# Example:

 Consider the following code: 
 ```mlir
 zlow.unstick(%stick, %A) {layout = "2D"}: memref<2x3xf16, #map2D>, memref<2x3xf32>
 affine.for {
   %a = affine.load(%A, %load_indices) : memref<2x3xf32>
   affine.store(%a, %B, %store_indices) : memref<4x5x6xf32>
}
 %res = memref.alloc() : memref<4x5x6xf16, #map3D>
 zlow.stick(%B, %res) {layout = "3D"}: memref<4x5x6xf32>, memref<4x5x6xf16, #map3D>
 ```
 `%stick` memref is unstickified and shuffled by the pair of (affine.load,affine.store),
 then stickified again. It said data are transfered from a stickified memref
 into another stickified memref via a chain of affine transformation.

 The above code can be rewritten into the following code:
 ```mlir
 %res = memref.alloc() : memref<4x5x6xf16, #map3D>
 affine.for {
   %a = affine.load(%stick, %load_indices) : memref<2x3xf16, #map2D>
   affine.store(%a, %res, %store_indices) : memref<4x5x6xf16, #map3D>
}
 ```
 where data will be directly loaded from / stored to stickified memref.

 This pattern is often found in code generated for data transformation such
 as Transpose, Concat, and Split.

 # Why does this rewriting work?
 
 - This rewriting depends on the fact that `zlow.stick` and `zlow.unstick`
 maintain an affine map that maps one element in a memref to an element in
 another memref. Those maps are `#map2D` and `#map3D` in the above example.
 Combined with affine.load and affine.store, one element in a stickified
 memref can be forwarded directly into an element in another stickifired
 memref without `zlow.stick` and `zlow.unstick`.

 - The shape of the input and output memrefs of `zlow.stick`/`zlow.unstick`
 are the same except the case of layout NCHW. In case of NCHW, dimensions are
 permuted, so we handle NCHW as a special case in this rewriting.
 ```mlir
  zlow.stick(%X, %res) {layout = "NCHW"}: memref<1x3x5x7xf32>, memref<1x5x7x3xf16, #mapNHWC>
  ```
  Shape of `%X` is in NCHW while shape of `%res` is in NHWC.

 ```mlir
 zlow.unstick(%X, %res) {layout = "NCHW"}: memref<1x5x7x3xf16, #mapNHWC>, memref<1x3x5x7xf32>
 ```
  Shape of `%X` is in NHWC while shape of `%res` is in NCHW.

# Limitations

 - Unstickified (or CPU-targeted) memrefs (`%A` and `%B`) must have no affine map.
 Theoretically, we could support affine map on unstickified memrefs by
 composing affine-map, but at this moment all unstickified memrefs in onnx-mlir have no affine map.


src/Accelerators/NNPA/Transform/ZLow/ZLowRewrite.cpp
```diff
@@ -100,6 +112,330 @@ class StickViewUnstickRemovalPattern : public OpRewritePattern<ZLowStickOp> {
...
+///
+/// * Pattern to rewrite
+/// ```
+/// zlow.unstick -> affine.for (affine.load -> affine.store) -> zlow.stick
+///    |                            |
+///    |                            '--------> affine.store) -> zlow.stick
+///    |
+///    '----------> affine.for (affine.load -> affine.store) -> zlow.stick
+///                                                              ^
+/// zlow.unstick -> affine.for (affine.load -> affine.store) ----'
```
chentong319, line 125: Can this Op be any Op to operate on the sticky data?

src/Accelerators/NNPA/Transform/ZLow/ZLowRewrite.cpp
```diff
@@ -100,6 +112,330 @@ class StickViewUnstickRemovalPattern : public OpRewritePattern<ZLowStickOp> {
...
+        operandMap.map(NCHWIndices, NHWCIndices);
+      }
+      rewriter.clone(*storeOp.getOperation(), operandMap);
+      rewriter.eraseOp(storeOp);
+    }
+
+    // Remove ZLowUnstickOp.
+    rewriter.eraseOp(unstickOp);
+    // Copy ZLowStickOp to the removableStickOps. We cannot remove it now
+    // because there are potentially other AffineStoreOps connecting to it.
```
chentong319, line 364: connecting -> using? Clearer on use/def relation.

src/Accelerators/NNPA/Transform/ZLow/ZLowRewrite.cpp
```diff
@@ -100,6 +112,330 @@ class StickViewUnstickRemovalPattern : public OpRewritePattern<ZLowStickOp> {
...
+  bool matchAndCollectAffineStore(const SmallVectorImpl<AffineLoadOp> &loadOps,
+      SmallVectorImpl<AffineStoreOp> &storeOps) const {
+    for (AffineLoadOp loadOp : loadOps) {
+      Value loadValue = loadOp.getValue();
+      for (Operation *user : loadValue.getUsers()) {
+        if (user == loadOp.getOperation())
+          continue;
+        if (auto storeOp = llvm::dyn_cast<AffineStoreOp>(user)) {
+          // Store's input must be defined by a memref.alloc.
+          Value storeMemref = storeOp.getMemref();
```
chentong319, line 398: This constraint will not exist if this transformation is done before bufferization.   Right?

tungld: Yes, bufferization would be done in advance.

src/Accelerators/NNPA/Transform/ZLow/ZLowRewrite.cpp
tungld: In general, yes. The purpose here is to remove zlow.unstick/zlow.stick, so I check zlow.unstick/zlow.stick only.

src/Accelerators/NNPA/Transform/ZLow/ZLowRewrite.cpp
tungld: Fixed this. Thanks!

AlexandreEichenberger (commented):
High level question: did you consider removing the stick/unsick operation and directly passing the layout to the memory-move operations (aka  concat/split/transpose/...)

tungld:
> High level question: did you consider removing the stick/unsick operation and directly passing the layout to the memory-move operations (aka concat/split/transpose/...)

Yes, I did it for Concat: https://github.com/onnx/onnx-mlir/blob/main/src/Accelerators/NNPA/Transform/ZHigh/ZHighLayoutPropagation.cpp#L343-L363
```
    // Layout propagation for ONNX Ops. Since DLFLOAT16 is not supported in
    // LLVM, we disable computation operators like Add, Div, etc. and enable
    // only data movement operators like Concat.

    // Add
    // patterns.insert<ONNXBinaryOpLayoutPropPattern<ONNXAddOp>>(&getContext());
    // Div
    // patterns.insert<ONNXBinaryOpLayoutPropPattern<ONNXDivOp>>(&getContext());
    // Mul
    // patterns.insert<ONNXBinaryOpLayoutPropPattern<ONNXMulOp>>(&getContext());
    // Sub
    // patterns.insert<ONNXBinaryOpLayoutPropPattern<ONNXSubOp>>(&getContext());
    // Reciprocal
    // patterns.insert<ONNXUnaryOpLayoutPropPattern<ONNXReciprocalOp>>(
    //     &getContext());
    // Sqrt
    // patterns.insert<ONNXUnaryOpLayoutPropPattern<ONNXSqrtOp>>(&getContext());

    // Concat
    patterns.insert<ONNXConcatLayoutPropagatePattern>(&getContext());
```
Then, I realized that it might cause confusion. By description, ONNX concat/split/transpose/... work with dense tensors and do not consider tensors with layout (or sparse tensors), and in their lowering to krnl we need to care layouts that potentially come from some accelerators. So it complicates optimization decision during the lowering.

tungld:
@AlexandreEichenberger @chentong319 this patch is ready to review. I would like to land it first since it does not contradict high-level optimizations at zhigh level.

src/Accelerators/NNPA/Transform/ZLow/ZLowRewrite.cpp
```diff
@@ -100,6 +112,330 @@ class StickViewUnstickRemovalPattern : public OpRewritePattern<ZLowStickOp> {
...
+      // This DummyOp is used to make the intermediate generated code valid. It
+      // will be removed automatically via canonicalization.
+      Value dummyConverter =
+          rewriter.create<ZLowDummyOp>(loc, stickifiedElementType, storeValue);
+      // Clone storeOp with new Memref, Value, and Indices.
+      IRMapping operandMap;
+      operandMap.map(storeOp.getMemref(), stickMemref);
+      operandMap.map(storeOp.getValue(), dummyConverter);
+      // Permute indices in case of NCHW layout.
+      if (isNCHWLayout) {
```
AlexandreEichenberger, line 347: Don't you have to filter out more layouts initially (only saw the 1D and 2DS being filtered out). Or support more of them here?

Maybe it would be more resilient at the top if you stated the one you support, and all the ones that are not supported would then report failure...

```
if (!(layout supported)) report failure
```

tungld: > Don't you have to filter out more layouts initially (only saw the 1D and 2DS being filtered out). Or support more of them here?

No, only 1D and 2DS are not supported since their affine_maps have issue reported here: https://github.com/onnx/onnx-mlir/issues/1940. The remaining layouts have no problem and we support them all. So I filtered out 1D and 2DS at the beginning at Line 227.

AlexandreEichenberger (approved):
LGTM

PR #2013, opened 2023-02-10
[NNPA] Rewrite Sqrt into zDNN-supported operations: div, exp and log

tungld:
Since zDNN does not support `sqrt`, this patch rewrite `sqrt` into zDNN-supported operations, i.e. div, exp, and log.
For efficiency, the rewriting is done only when `sqrt` is sandwiched by `zhigh.unstick` and `zhigh.stick`, that makes sure `zhigh.unstick` and `zhigh.stick` will be removed.

Formulas for `sqrt`: `sqrt(x) = exp (0.5 * log(x))`


src/Accelerators/NNPA/Dialect/ZHigh/ZHighOps/Stick/ZHighStick.td
```diff
@@ -80,4 +80,43 @@ def ReplaceONNXLeakyReluPattern: Pat<
...
+// In this rewriting pattern we use:
+//   - `exp(log(x)/2)` for `sqrt(x)`, and 
+//   - `div(1,x) for `reciprocal`
+// So, `1/sqrt(X) = div(1, exp(log(x)/2))`
+//
+// The pattern
+//   zhigh.Stick (onnx.Reciprocal (onnx.Sqrt (zhigh.Unstick (%X))))
+// can be replaced by
+//   zhigh.Div (1, zhigh.Exp (zhigh.Div (zhigh.Log(%X), 2))
+//
```
AlexandreEichenberger, line 107: Can we use this equality?

1/(x^n) == x^(-n)

and thus 1/sqrt(n) = 1 / x^(1/2) = x ^ (-1/2).

using this with the log, we would get
1/sqrt(n) = exp( (-1/2) * log(x))

Also, is the `x / 2` as fast as the `0.5 * x` on the zAIU?

tungld: Good suggestion! I will do some experiments to see which one is the best on the zAIU. Thanks!

AlexandreEichenberger: Did it worked?

src/Accelerators/NNPA/Dialect/ZHigh/ZHighOps/Stick/ZHighStick.td
```diff
@@ -80,4 +80,42 @@ def ReplaceONNXLeakyReluPattern: Pat<
...
 
+// Since zDNN does not support Sqrt(X), we calculate it by using zDNN-supported
+// operations, i.e. Exp and Log.
+// Formulas: `sqrt(X) = exp(log(x)/2) = exp(0.5 * log(x))`
+def ReplaceONNXSqrtPattern: Pat<
+  (ZHighStickOp:$stick (ONNXSqrtOp (ZHighUnstickOp $X)), $layout),
+  (ZHighExpOp (ZHighMulOp (ZHighLogOp $X, (returnType $X)),
+                          (ZHighStickOp (GetConstantOfType<"0.5"> $X), $layout),
+                          (returnType $X))),
+  [(IsStaticShapeTensor $X), (SameLayout $X, $stick)]
```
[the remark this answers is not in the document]
tungld: It is because of the second operand of `ZHighMulOp` that is a constant of `0.5`. We create a constant tensor `(GetConstantOfType<"0.5"> $X)` by using X's static shape. Then ZHighStickOp of the constant can be folded at compile time. Otherwise, there is overhead of stickification.

src/Accelerators/NNPA/Dialect/ZHigh/ZHighOps/Stick/ZHighStick.td
tungld: I am sorry that I haven't had chance to do experiments because the accelerator machine is having an issue creating a container. 

So I changed to use `1/sqrt(x)= exp (-0.5 * log(x))` in the code. In general, using `*` would be better than `/`. I think  the patch is ready.

src/Accelerators/NNPA/Dialect/ZHigh/ZHighOps/Stick/ZHighStick.td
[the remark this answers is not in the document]
AlexandreEichenberger: got it, tx.

AlexandreEichenberger (approved):
LGTM

Issue #1975, opened 2023-01-25
Error in loading ConstantOp 


chentong319:
Here is the error I mentioned at the meeting.
I run `onnx-mlir --EmitONNXBasic my_model.onnx' and then `onnx-mlir --EmitONNXIR my_model.onnx.mlir`. I got error message
```
error: unknown attribute `dense_disposable` in dialect `ONNX`
```
This error occurs only on one model, which can not be shared publicly. The constant in my_model.onnx.mlir looks like this
```
%0 = "onnx.Constant"() {value = #onnx.dense_disposable<#1:"0x000000000000000000000000000000000000000000000000"> : tensor<3xi64>} : () -> tensor<3xi64>
```
The constant is not printed as dense format, while the constant of other model (output of --EmitONNXBasic) is in dense format and can be loaded back to onnx-mlir. Can you figure out the reason without the model?

By the way, I'd like to explain the question I asked at meeting. We can import a model from .onnx file and the constant is represented with the disposable attribute for performance. The disposable attribute is converted to dense attribute before onnx to KRNL conversion. However, we may print out the IR, and then run onnx-mlir again. To my understanding, the constant will be print as dense, not the disposable. Right? Can we print out the disposible attribute? When onnx-mlir load the print out of dense attribute again, do we need add a pass to convert the dense Attribute to disposable? Or has it been done automatically?

sorenlassen:
thanks for the error report!

I'll first try to figure out why it was printed out as `#onnx.dense_disposable<..>` and not `dense<..>` and then I'll try to answer the other questions

sorenlassen:
I can't recreate the error you described and I'm puzzled how you got it to print `"onnx.Constant"` with double quotes - it must have bypassed `ONNXConstantOp::print()` from OnnxOps.cpp which I can't see how can happen

studying the implementation I found another corner case where `#onnx.dense_disposable` might get printed, namely if the DisposableElementsAttr type doesn't match the ONNXConstantOp type, which would probably be impossible to parse - I'll try to verify and fix that

chentong319:
The error did not come from ONNXConstantOp::print(OpAsmPrinter &odsPrinter), which is not invoked for the model with error. The attribute print is called by OperationPrinter::printGenericOp. So the DisposableElementsAttr is printed as an Attribute directly from mlir::AsmPrinter::Impl::printNamedAttribute.
I do not know why. What's the Op?  Is the model correct though it passed onnx_check?
[...]

PR #2020, opened 2023-02-13
Custom print and parse ConstantOfShape

sorenlassen:
Solves #1975 

The new custom assembly of onnx.ConstantOfShape is the same as the generic, just without the double quotes around "onnx.ConstantOfShape" and, most importantly, any DisposableElementsAttr is printed as a DenseElementsAttr.

Also fixed the onnx.Constant custom assembly to print any DisposableElementsAttr in the attribute dictionary as a DenseElementsAttr in the rare cases (when there are extra attributes or when the ElementsAttr type doesn't match the op type) where the attribute dictionary is written out as curly brace enclosed comma separated names and values.

I tested that the changes work on bidaf-9.onnx


AlexandreEichenberger (approved):
LGTM

PR #2022, opened 2023-02-14
Introduce onnx.ShapeTransform to optimize data transformation

tungld:
This patch introduces `onnx.ShapeTransform` operation to optimize data transformation on CPU such as Transpose, Reshape.

`onnx.ShapeTransform`  transforms a tensor into another tensor by using an affine map. This is elemement-wise transformation, so each element in the input will be copied to an element in the output via the affine map.
The affine map must be bijective.

For example, the following code is using `onnx.ShapeTransform` to reshape a tensor from 2D to 4D.
```mlir
#reshape = affine_map(d0, d1) -> (d0/32, d0%32, d1/64, d1%64)
%Y       = onnx.ShapeTransform(%X) {index_map = #reshape} :  (tensor<128x128xf32>) -> tensor<4x32x2x64xf32>
```

`onnx.ShapeTransform` will be materialized into an `affine.for` via lowering to `krnl` dialect, e.g.
```mlir
%alloc = memref.alloc() {alignment = 16 : i64} : memref<4x32x2x64xf32>
affine.for %arg1 = 0 to 128 {
  affine.for %arg2 = 0 to 128 {
    %0 = affine.load %arg0[%arg1, %arg2] : memref< 128x128xf32 >
    affine.store %0, %alloc[%arg1 / 32, %arg1 % 32, %arg2 / 64, %arg2 % 64] : memref<4x32x2x64xf32>
  }
}
```

Important property of `onnx.ShapeTransform` is that it is composable. When being canonicalized, ShapeTransform operations are composed into a new ShapeTransform operation by composing their affine maps.

The following two `onnx.ShapeTransform` operations can be composed into a single `onnx.ShapeTransform`, e.g.
```mlir
#transpose = affine_map<(d0, d1, d2, d3) -> (d2, d0, d1, d3)>
#reshape   = affine_map<(d0, d1) -> (d0 / 32, d0 % 32, d1 / 64, d1 % 64)>

%0  = "onnx.ShapeTransform"(%arg0) {index_map = #reshape} : (tensor<128x128xf32>) -> tensor<4x32x2x64xf32>
%1  = "onnx.ShapeTransform"(%0) {index_map = #transpose} : (tensor<4x32x2x64xf32>) -> tensor<2x4x32x64xf32>
```
will become
```mlir
#reshape_transpose = affine_map<(d0, d1) -> (d1 / 64, d0 / 32, d0 % 32, d1 % 64)>
%0 = "onnx.ShapeTransform"(%arg0) {index_map = #reshape_transpose} : (tensor<128x128xf32>) -> tensor<2x4x32x64xf32>
```

In this patch, `onnx.ShapeTransform` is used to optimize a pattern in bertsquad-12 on NNPA, where a sequence of `Reshape -> Transpose -> Reshape`  becomes a single `ShapeTransform`.


chentong319:
Any constraints on the affine map? Will every element of the output tensor be written? Will some output tensor be written more than once?

tungld:
> Any constraints on the affine map? Will every element of the output tensor be written? Will some output tensor be written more than once?

It would be injective, but not necessarily bijective. However, if we use the input type and the affine map to compute the output type, perhaps, it should be bijective.

tungld:
>  Is there any reason that this is a zhigh op? It feel that it could really be useful to ONNX in general.

Sure, I can move it to ONNX.

src/Accelerators/NNPA/Dialect/ZHigh/ZHighOps/OpHelper.cpp
```diff
@@ -209,5 +212,229 @@ bool oneIsOfNHWCLayout(Type t1, Type t2) {
   return true;
 }
 
+/// Check if ONNXReshapeOp is reshaping 2D to 4D by tiling each input dimension.
```
AlexandreEichenberger, line 215: Code is kind of fine as is, it feels like one could use a single version that goes from small to big, and one big to small.

AlexandreEichenberger: It also feel that nothing here is Z-related. If so, it should find a place outside of NNPA subduers

AlexandreEichenberger (approved):
LGTM, thanks for the complete solution, with ONNX shape inference, merging rules in ONNX, and then creation in NNPA for specific dialects.

It feels that we could switch many of the regular ops into ShapeTransform, and then benefit from the merging too, outside of this specific NNPA application. I assume that this might come in a later PR.

tungld:
@sorenlassen FYI.

PR #2024, opened 2023-02-14
Fix castlike shape inference

philass:


src/Dialect/ONNX/ONNXOps/Math/ElementwiseUnary.cpp
```diff
@@ -154,7 +154,8 @@ LogicalResult ONNXCastLikeOp::inferShapes(
   if (!hasShapeAndRank(getInput()))
     return success();
 
-  Type elementType = (*this)->getAttr("to").cast<::TypeAttr>().getValue();
+  Type elementType =
+      getTargetType().getType().dyn_cast<RankedTensorType>().getElementType();
```
sorenlassen, line 158: getElementType() from TypeUtilities is convenient here: `getElementType(getTargetType().getType())` 

otherwise I would nit pick about dyn_cast vs cast and RankedTensorType vs ShapedType :)

sorenlassen (approved):
good test

AlexandreEichenberger (approved):
LGTM

src/Dialect/ONNX/ONNXOps/Math/ElementwiseUnary.cpp
philass: Funny enough... I think you still nit picked :p

PR #2026, opened 2023-02-15
Update krnl.md

sstamenova:
At some point an update of krnl.md was missed - update the file and also change the target to only copy the `.md` files when they've changed.


chentong319 (approved):
LGTM!

PR #2031, opened 2023-02-17
parse DisposableElementsAttr

sorenlassen:
Solves #1975 

I changed the print format a little bit, removing the `#` before the "id", so it's now
```
#onnx.dense_disposable<1:"0x00010000">
```
instead of
```
#onnx.dense_disposable<1:"0x00010000">
```
I made the change because I couldn't figure out how to parse the `#`

I ran it on @tungld's example model bertsquad-12-int8.onnx and I no longer see the error `error: unknown attribute 'dense_disposable' in dialect 'ONNX'`

### DIFF
--- a/src/Dialect/ONNX/ElementsAttr/DisposableElementsAttr.cpp
+++ b/src/Dialect/ONNX/ElementsAttr/DisposableElementsAttr.cpp
@@ -15,6 +15,7 @@
 #include "src/Support/TypeUtilities.hpp"
 
 #include "llvm/ADT/StringExtras.h"
+#include "llvm/Support/Endian.h"
 
 #include <algorithm>
 #include <string>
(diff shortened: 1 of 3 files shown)


tungld:
I see this code in MLIR about parsing "#": https://github.com/llvm/llvm-project/blob/main/mlir/lib/AsmParser/Parser.cpp#L943

Hope it help.

sorenlassen:
> I see this code in MLIR about parsing "#": https://github.com/llvm/llvm-project/blob/main/mlir/lib/AsmParser/Parser.cpp#L943

yes, I saw that and tried to use it but discovered that it's not exposed in the public interface, it's only internal to the implementation in mlir/lib

in any case, I think it's fine to not use `#`

chentong319 (approved):
LGTM!

PR #2036, opened 2023-02-20
NFC: remove redundant or unnecessary semicolons

christopherbate:
This fixes warnings that are present when building with clang-15.


sorenlassen (approved):
LGTM

PR #2037, opened 2023-02-20
Elementwise simd: preparation

AlexandreEichenberger:
Preparing Elementwise for SIMD:
* used the builder functions systematically
* test for types now include handling vectors as well.
This PR does not change the code being generated.

I also cleaned up the code in the MLIR builders to have a common pattern (for a given type, generate the code; if reach end of function: `llvm_unreachable` as it should have been handled by the supported type above.

Added many additional math functions, including unsigned compare, plus a generic compare that perform automatically the signed, unsigned, float compare depending on the input types.

Unsupported ops (like copySign for integer) are now labeled with the type `void` and when instantiating the code for the lowering, the code check that the type of the op is not `void`. Caught a few errors where isNan is only for float, but we checked the type of the output which is `i1`. Now we check the type of the first operand.

Note: it appears that ONNX Mod used the 'copySign" function but from the math (&backend test), it does not seems to be needed. Any idea why it is used?


AlexandreEichenberger:

AlexandreEichenberger:

AlexandreEichenberger:
Subsumed by #2042

PR #2039, opened 2023-02-21
added min versions for torch as suggested by Gong

AlexandreEichenberger:
Sweep was unhappy about unqualified requirement. Updated the docs/mnist_example requirement file to suggestion by @gongsu832

### DIFF
--- a/docs/mnist_example/requirements.txt
+++ b/docs/mnist_example/requirements.txt
@@ -1,2 +1,2 @@
-torch>=1.9.1
-torchvision>=0.10.1
+torch>=1.9.1, <=1.11.0
+torchvision>=0.10.1, <=1.12.0
--- a/docs/mnist_example/requirements.txt
+++ b/docs/mnist_example/requirements.txt
@@ -1,2 +1,2 @@
-torch>=1.9.0
-torchvision>=0.10.0
+torch>=1.9.1
+torchvision>=0.10.1
--- a/docs/mnist_example/requirements.txt
+++ b/docs/mnist_example/requirements.txt
@@ -1,2 +1,2 @@
-torch
-torchvision
+torch>=1.9.0
+torchvision>=0.10.0


docs/mnist_example/requirements.txt
```diff
@@ -1,2 +1,2 @@
-torch
-torchvision
+torch>=1.9.0
```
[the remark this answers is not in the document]
gongsu832: We don't know what exact versions are required. The versions were used in an online article (https://analyticsindiamag.com/guide-to-feed-forward-network-using-pytorch-with-mnist-dataset/) Since we are specifying the minimal versions, lower versions are preferred since they are less restrictive.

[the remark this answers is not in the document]
gongsu832: Sure. torch>=1.9.1 and torchvision>=0.10.1 are fine.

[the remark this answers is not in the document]
AlexandreEichenberger: updating to these values

docs/mnist_example/requirements.txt
```diff
@@ -1,2 +1,2 @@
-torch
-torchvision
+torch>=1.9.1
```
sstamenova, line 1: It's better to have a min and a max in terms of preventing security issues with newly published versions.

AlexandreEichenberger: @sstamenova do you have suggestions on the max?

sstamenova: > @sstamenova do you have suggestions on the max?

Whatever the latest currently is? It's been out long enough that if it had security issues, we'd know

cjvolzka: > It's better to have a min and a max in terms of preventing security issues with newly published versions.

I'm unfamiliar with this. Is it a relatively new recommendation? I'll agree there's the risk of new versions having new zero days. However in having a fixed max, that means someone needs to periodically go through and update these. 

Otherwise say 1.9.2 comes out and it closes a security hole. We'd specifically not install that new versioin until someone goes back and updates this file. That seems much risker because then it's potentially locked into a version with a known vulnerability vs the risk of new vulnerabilities that no one knows about.

sstamenova: Not having limits leaves us open to supply chain attacks as a new version might depend on a compromised package. To prevent supply chain attacks, we (and all the packages we consume) should restrict which versions are supported to avoid the injection of compromised packages.

sstamenova: Also, unrelated to the security issue, leaving the upper bound open leaves the possibility of breaks like the one from numpy where a new version deprecated some features that were necessary and users that installed newer versions were broken

gongsu832: Having or not having a max both will require some manual intervention depending on the situation.

- having a max, a new version higher than the max may be excluded when it should not be (e.g., containing security fixes)
- not having a max, a new version higher than the unspecified max could break the build (e.g., due to incompatible changes)

So in terms of management overhead, they are about equal. Having a max provides an important property: it's deterministic regardless of whatever versions exist on the system. So in practice, if you want the `requirements.txt` to work on any system, you should specify a max.

cjvolzka: On the security side, I think we'll just have to agree to disagree. The short version: I'm comfortable with my current +1 and merging if others agree. If anyone feels strongly enough to -1 they can and I'll go along with it. If we do go that way, I highly recommend (but would not block on) having a full plan first.

The long version:

> Not having limits leaves us open to supply chain attacks as a new version might depend on a compromised package. To prevent supply chain attacks, we (and all the packages we consume) should restrict which versions are supported to avoid the injection of compromised packages.

I'll concede this but I think it works better in theory than practice. For this to work we'd have to watch every dependency throughout the project and only pull in updates after X amount of time. Although there's no guarantee of how long X would need to be. A poisoned supply chain may not be detected immediately. So it's still possible we get poisoned supplies.

In reality, by capping a version, it's possible the upstream pushes a fix and then we don't pull it in because we don't go back and notice it. We'd need a system to sweep through our entire repo and update these caps on some frequent basis.

I could maybe be convinced do this as a dedicated effort. That means we come up with a plan first to do this project wide. Setup rotations for people to sweep for updates on some schedule. Once we have that, then consistently set version caps throughout the project. 

What I'm leary of, is do it as a one off in a few limited spots. In that case I'm guessing no one will come back and look at them in any reasonable timeframe. 


> not having a max, a new version higher than the unspecified max could break the build (e.g., due to incompatible changes)

Yep, this is a risk. However it's got a good knock on effect. We _know_ there's something changing and we might need to refactor something on our end to keep upgrading. With a capped max, if we don't regularly update the cap, we might not know about an incompatible change compounding an already critical issue.

So say a log4j level issue is found in (fake numbers ahead) pytest back to version 3.0.0. They fix it in pytest 3.2.1 and above but we're capped to 3.0.1. What we then don't currently know is that in 3.1 they made a backwards incompatible change. So now we have an unexpected additioinal blocker that prevents us from pulling in a critical fix quickly. 

On the other hand, If we pulled in the lastest pytest by default we'd have known when 3.1 was released that we have a problem. We could then temp cap it to unblock our builds, but at least know about it so we can address it when bandwidth is available. Once fixed, we can remove the cap and therefor be better prepared to quickly pull in 3.2.1. 

Even if we never get the bandwidth, at the cap we should have a comment (like we do for `numpy` in another requirements file) so we will at least have a heads up about the challenges we'll face. 

> So in terms of management overhead, they are about equal. Having a max provides an important property: it's deterministic regardless of whatever versions exist on the system. So in practice, if you want the requirements.txt to work on any system, you should specify a max.

idk, about that. Once we specify a range you are deterministic in that range. But different machines may still have different combinations of packages based on their environment. So it should help reduce but not eliminate random build failures.

gongsu832: I think you are overthinking this. In terms of security, having a max could exclude a newer version with security fixes but not having a max could include a newer version that has security holes. So to me, it's a wash. Either way you have to watch the security related events and patch things up accordingly.

> idk, about that. Once we specify a range you are deterministic in that range. But different machines may still have different combinations of packages based on their environment. So it should help reduce but not eliminate random build failures.

Not sure I follow you. Regardless of the combinations of packages on the system, if `pip3 install -r requirements.txt` succeeds, it means all packages, including their recursive dependencies would have been within the proper range (if not they would have been installed). So the build shouldn't fail. If it failed, then someone didn't get their `requirements.txt` right.

cjvolzka: > So to me, it's a wash. Either way you have to watch the security related events and patch things up accordingly.

Agreed. That's why I wouldn't -1 if we wanted to set a max. Personally though I think not having a max is the better of two imperfect answers on the security front, for the reasons I gave above.

>Not sure I follow you. Regardless of the combinations of packages on the system, if pip3 install -r requirements.txt succeeds, it means all packages, including their recursive dependencies would have been within the proper range (if not they would have been installed). 

A bit of a digression from the security topic, but as an illustration:
Nodes tend not to be exactly the same. Some are added, updated or otherwise manipulated at different times. So it's very possible for different nodes could get different versions of packages all within the "acceptable" range.

So given requirements.txt:
```
apples>=1.1.0, <=1.9.2
pears>=1.5.0, <= 2.0.0
```

And due to various factors around platform, environments, maintenance timing, etc
Node A gets:
apple 1.2.2
pears 1.9.0

Node B gets:
apple 1.2.2
pears 2.0.0

Node C gets:
apple 1.2.3
pears 2.0.0

Now say there's a bug in apples 1.2.2 that makes it incompatible with pears 2.0.0 and it's fixed by apple 1.2.3. Then Nodes A and C will run fine but Node B fails because it has that bad combination. When you start equating in dependencies of dependencies, the chances of something happening goes up. Unless you have a single build node or a tool that guarantees all nodes are the exact same, I've yet to see a build system not periodically have issues specific to a single node due to such things.

sstamenova: One thing we ran into before we started adding max versions to requirements files is that we could not predict when breaks would happen because of new versions. When we ran into the numpy issue, for example, someone had to go investigate and spent a bunch of time trying to figure out what changed where to cause the failures. It was not immediately clear that they were related to one of the PyPi packages, so we wasted time investigating that we would not have otherwise. So from a very pragmatic standpoint, it is cheaper in the long term to have a max because any version changes will be deliberate and failures will be expected.

cjvolzka: I would be ok with capping for that reason. However new versions can often have security fixes which we risk missing unless we have a plan in place. So I'd prefer to have a plan in place before switching to such a model. 

Off the top of my head:
1. Frequency and Rotation length
1. Rotation Expectations 
    * Probably just try latest. Update if success. Comment and open git Issue for failures to be handled later
1. Location identification/consolidation
    * A list so we remember what to check or some tag we can easily search the code for places to check.
1. Might need to implement some mechanism to make sure a node is actually using the new versions.
    * If we increase a cap, but a node already has an old version that's still above the min, then I don't know if it'll actually update to the new max package or not.
    * If it doesn't, then we might not actually be trying the new package version. So later a node will just "randomly" fail when something does cause it to try the new version which leaves us with the original problem. 
3. Consistency.
    * Someone make a pass at finding all our requirements and setting the current max caps.


tl;dr
So I'm not against it, but personally I'd say that would be dedicated effort that's beyond the scope of this specific PR.

gongsu832: > So given requirements.txt:
> 
> ```
> apples>=1.1.0, <=1.9.2
> pears>=1.5.0, <= 2.0.0
> ```

When you write like that, you are basically telling pip that any version of 1.10<=apples<=1.9.2 should work with any version of 1.5.0<=pears<=2.0.0. Of course pip is not going to actually check all the combinations since it's assumed that **you already did that before you wrote down the `requirements.txt`**.

So if you have a node B where apples==1.2.2 does not work with pears==2.0.0, then you wouldn't have written the `requirements.txt` like the above.

AlexandreEichenberger:
Good discussion here, way beyond this poor nminst example. I picked a max from not too recent a version. It's an example, so it should be fine.

https://pytorch.org/get-started/previous-versions/

AlexandreEichenberger:

PR #2042, opened 2023-02-23
Preparation for Elementwise simd (v2): upgrade of the alloc infrastructure.

AlexandreEichenberger:
Non functional PR: upgraded infrastructure / fixed bugs.

Ripped the old "insertAllocAndDealloc" infrastructure and replaced it by `MemRefBuilder.alignedAlloc` infrastructure. Removed the "emitDealloc" flags from the passes. 

Corrected a lot of `<0`, `<=0`, `>0`... references with shape, and replaced it with appropriate `ShapedType::kDynamic` calls.

There were errors with respect to shape inference that cannot determine the shape of a dynamic output (e.g. Compress or NonNull ops). Now these truly dynamic shapes can be represented by `IndexExpr Questionmarks` even during lowering; obviously, the op-specific code lowering is then responsible to introduce code that will generate these actual values. Before, we generated a literal `-1` value... which often would get refined away by the often erroneous static sizes given by a model. This is now fixed.

Fixed spelling mistakes and updated the dates of modified files to 2023.


src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
```diff
@@ -542,7 +542,8 @@ bool isSuitableForZDNN<ONNXReduceMeanOp>(
   }
 
   // Check dimensions.
-  if ((shapeData[2] < 0) || (shapeData[3] < 0) || (shapeData[2] > 1024) ||
+  if ((shapeData[2] == ShapedType::kDynamic) ||
+      (shapeData[3] == ShapedType::kDynamic) || (shapeData[2] > 1024) ||
```
chentong319, line 546: should we define some constant for 1024? Not related to this PR.

src/Conversion/KrnlSeqToMemref/KrnlSeqAlloc.cpp
```diff
@@ -48,9 +48,10 @@ class KrnlSeqAllocOpLowering : public ConversionPattern {
     Value outputSeq = thisOp.getResult();
     auto outputType = outputSeq.getType().cast<MemRefType>();
     Value alloc;
-    if (outputType.isDynamicDim(0))
-      alloc = create.mem.alloc(outputType, operandAdaptor.getLength());
-    else
+    if (outputType.isDynamicDim(0)) {
+      llvm::SmallVector<Value, 4> length(operandAdaptor.getLength());
+      alloc = create.mem.alloc(outputType, length);
```
chentong319, line 53: Should we always use alignedAlloc?

chentong319 (approved):
LGTM!
I also like the change to always use create.xxx.yyy.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
AlexandreEichenberger: 1024 is a HW limit for the NNPA. So you are suggesting that it should be a constant, or a variable that could change in the future for different instances of NNPAs, if that were the case? That make sense.
@tungld if that make sense to you, please create a new PR for Tong's suggestion.

AlexandreEichenberger:
> I also like the change to always use create.xxx.yyy.

Thanks, it's a pattern that @tungld started and I like it too. I did not systematically change it in the code, but will change it when I revise specific codes.

PR #2047, opened 2023-03-02
Compilation log in RunONNXModel.py

tungld:
This patch improves RunONNXModel.py in which compilation log is dumped into a file for debugging.


AlexandreEichenberger (approved):
LGTM.

I am also making a few changes, some benchmarks use booleans and I added support to generate / tests boolean results.

tungld:
> I am also making a few changes, some benchmarks use booleans and I added support to generate / tests boolean results.

Thank you!

tungld:

AlexandreEichenberger (approved):
LGTM, thanks

tungld:

AlexandreEichenberger:
@tungld Can I ask for a change of names in the options?
```
  --save-data PATH      Path to a folder to save the inputs and outputs in protobuf
  --data-folder DATA_FOLDER
                        Path to a folder containing inputs and outputs stored in protobuf. If --verify=ref, inputs and outputs are reference data for verification
```
I would remember these options better if the first one remains `save-folder` and the second one was renamed to `use-folder`.

AlexandreEichenberger:
And that might be too much work within that script (we could do a meta script that would do that), but it might be useful to have something like this:
`-compile-args` and `-reference-compile-args`. It would compile twice, the reference args would be used to define the `save-data` and the normal compile args would be used normally, and the results of the normal compilation would be compared to the reference compilation.

tungld:
> Or since in verify you use ref as the saved data, we could rename it to save-ref and use-ref... that would be most consistent.

We have `--save-so` and `--load-so` for .so file. How about using `--save-ref` and `--load-ref` to make them consistent.

> And that might be too much work within that script (we could do a meta script that would do that), but it might be useful to have something like this:
-compile-args and -reference-compile-args. It would compile twice, the reference args would be used to define the save-data and the normal compile args would be used normally, and the results of the normal compilation would be compared to the reference compilation.

I see. Got what you mean, a meta script looks better.

PR #2049, opened 2023-03-02
Update the verifier in LayoutTransform

hamptonm1:


src/Dialect/ONNX/ONNXOps/Additional/LayoutTransform.cpp
```diff
@@ -43,10 +43,18 @@ LogicalResult ONNXLayoutTransformOp::inferShapes(
 //===----------------------------------------------------------------------===//
 LogicalResult ONNXLayoutTransformOp::verify() {
   if (hasShapeAndRank(getData()) && hasShapeAndRank(getOutput())) {
-    if (getShape(getData().getType()) != getShape(getOutput().getType()))
-      return emitOpError("Input and output tensors must have the same shape");
+    // Get the unknown dimension from data.
+    auto dataType = getData().getType().dyn_cast<RankedTensorType>();
+    auto outputType = getOutput().getType().dyn_cast<RankedTensorType>();
+    // Check if input has a static dimension and output has dynamic dimensions
+    if (!dataType && outputType) {
```
tungld, line 50: This check means `dataType` is unranked (e.g. `tensor<2xf32>`) and `outputType` is ranked (e.g. `tensor<?x3x?xf32>`). I guess what you want to check is `dataType` or `outputType` is unranked, that is `if (!dataType || !outputType)` or `!(dataType && outputType)`. How about this check:

```C
if (auto dataType = getData().getType().dyn_cast<RankedTensorType>())
  if (auto outputType = getOutput().getType().dyn_cast<RankedTensorType>())
     return (getShape(dataType) == getShape(outputType);

return success();
```

hamptonm1: Thanks for your feedback...let me try it out and see my results.

hamptonm1:
@tungld what do you think now? I also want to test this PR against Roberta-base-11 and see what I get back.

src/Dialect/ONNX/ONNXOps/Additional/LayoutTransform.cpp
```diff
@@ -43,8 +43,9 @@ LogicalResult ONNXLayoutTransformOp::inferShapes(
 //===----------------------------------------------------------------------===//
 LogicalResult ONNXLayoutTransformOp::verify() {
   if (hasShapeAndRank(getData()) && hasShapeAndRank(getOutput())) {
```
tungld, line 45: @hamptonm1 this line looks redundant. It checks that both data and output must be shaped and ranked.

hamptonm1: Okay I can remove this line...it was leftover from the previous code. Let me remove then.

src/Dialect/ONNX/ONNXOps/Additional/LayoutTransform.cpp
```diff
@@ -42,10 +42,9 @@ LogicalResult ONNXLayoutTransformOp::inferShapes(
 // Verifier
 //===----------------------------------------------------------------------===//
 LogicalResult ONNXLayoutTransformOp::verify() {
-  if (hasShapeAndRank(getData()) && hasShapeAndRank(getOutput())) {
-    if (getShape(getData().getType()) != getShape(getOutput().getType()))
-      return emitOpError("Input and output tensors must have the same shape");
-  }
+  if (auto dataType = getData().getType().dyn_cast<RankedTensorType>())
+    if (auto outputType = getOutput().getType().dyn_cast<RankedTensorType>())
+      return success(getShape(dataType) == getShape(outputType));
```
tungld, line 47: @hamptonm1 does it fix your issue? or maybe we need to have a fine-grained check for each dimension of the input and output.
For example : input is `tensor<1x2x3xf32>` and output is `tensor<1x?x3xf32>`. In this case we return success because we cannot verify the real value of the output's second dim (that is unknown).

hamptonm1: Let me test this now!

hamptonm1: @tungld Does this update look better?

tungld: @hamptonm1 it looks good. Thanks!

tungld (approved):
LGTM. Thanks!

src/Dialect/ONNX/ONNXOps/Additional/LayoutTransform.cpp
```diff
@@ -42,10 +38,19 @@ LogicalResult ONNXLayoutTransformOp::inferShapes(
...
+    if (auto outputType = getOutput().getType().dyn_cast<RankedTensorType>())
+      for (int64_t i = 0; i < dataType.getRank(); ++i)
+        // Check if there is an unknown dimension in the dataShape and
+        // outputShape. If there is an unknown dimension, we will return true.
+        // If we know the dimension of dataShape and outputShape they should be
+        // equal, if not then we return false.
+        if (dataType.getShape()[i] == ShapedType::kDynamic ||
+            outputType.getShape()[i] == ShapedType::kDynamic)
+          return success();
+        else if (dataType.getShape()[i] != outputType.getShape()[i])
```
sorenlassen, line 51: @hamptonm1 this causes a compiler warning
```
LayoutTransform.cpp:51:9: warning: add explicit braces to avoid dangling else [-Wdangling-else]
        else if (dataType.getShape()[i] != outputType.getShape()[i])
```

sorenlassen: here are LLVM's guidelines for when to (not) omit curly braces:
https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements

PR #2050, opened 2023-03-02
Terminology update

cjvolzka:
Update terms to avoid use of "sanity" based on https://www.ibm.com/blogs/think/2020/08/words-matter-driving-thoughtful-change-toward-inclusive-language-in-technology/.


tungld (approved):
LGTM.

AlexandreEichenberger:
I was the one introducing these tests, for our own safety while we are transitioning to use kDynamic... Thanks for catching them @cjvolzka

AlexandreEichenberger:
If there was a way to add these checks in a CI... they would be fixed during development.

PR #2051, opened 2023-03-03
Switch to use OpConversionPattern instead of ConversionPattern - part 1 - ONNXToKrnl

tungld:
This patch replaces ConversionPattern by OpConversionPattern used in ONNXToKrnl, which is more convenient because OpAdaptor is available to use without explicitly construction.

ONNXLoc is used to annotate more bug information (e.g. operation name) into Location, instead of using the bare location via `op->getLoc`. Now all lowerings from ONNX to Krnl are using ONNXLoc.

Replacement for other lowerings for Krnl, ZHigh, ZLow will come next.


AlexandreEichenberger (approved):
LGTM

PR #2052, opened 2023-03-03
Update Onnx to v1.13.1

gongsu832:
So we can reenable backend model tests.

### DIFF
--- a/test/backend/inference_backend.py
+++ b/test/backend/inference_backend.py
@@ -1051,14 +1051,11 @@ def get_test_models():
         ############################################################
         # Model (alphabetical order)
     variables.model_test_to_enable_dict = {
-        # From https://github.com/onnx/onnx-mlir/issues/2001: Disable the models below - they are failing to download
-        # correctly because access to them has been lost. Until a better fix is available, unblock everyone by simply
-        # not running them.
-        # "test_densenet121_cpu": {STATIC_SHAPE:{}},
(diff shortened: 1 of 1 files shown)


gongsu832:
@cjvolzka

AlexandreEichenberger (approved):
LGTM, glad to hear we are able to reintroduce the benchmarks.

Do we state somewhere on the main README what our current ONNX is? I don't see it in the requirements (maybe it's not a requirement as we install it). But it might be useful to have it explicit in human readable form.

Unrelated, would it be possible to add a "make check-onnx-backend-light" that skips the benchmarks? Sometime in development, we want just a quick run; benchmarks take quite a bit more time.

gongsu832:
> Do we state somewhere on the main README what our current ONNX is? I don't see it in the requirements (maybe it's not a requirement as we install it). But it might be useful to have it explicit in human readable form.

We don't specify the onnx version since you must install whatever version that comes with onnx-mlir.

> Unrelated, would it be possible to add a "make check-onnx-backend-light" that skips the benchmarks? Sometime in development, we want just a quick run; benchmarks take quite a bit more time.

Currently the backend tests include both node and model tests. We can add an option to select either node or model tests (default to both).

AlexandreEichenberger:
> Currently the backend tests include both node and model tests. We can add an option to select either node or model tests (default to both).

If you could do a test that only does the small tests (nodes), that would be helpful to me and possibly others. Tx

PR #2053, opened 2023-03-03
Elementwise SIMD: support for unary ops

AlexandreEichenberger:
Added support for unary op SIMD code gen.

* Every element wise op now check if it is supported for the given types. There are now 3 possibilities: a dialect op used to implement the onnx op, a "unsupported" value, and a "custom" value (in which case a custom version needs to be provided).
* Every element wise operating has a field to indicate if it can be simdized or not. Eventually, we may want to add a "arch" type so that it may be simdized only for ops that are beneficial for the given target.
* Added support to math constant to splat values when type is vector
* Added support to memref reshape op (simple op, plus support to/from flat)
* Added support for subview and collapseShape. They need an additional pass, so if used, this pass needs to be enabled.

Testing:
* added 2 onnx_lowering files; one without -O3, which tests unoptimized code, and one with -O3 to test the new patterns, including the SIMD elementary ops (which are only gen at -O3).
* fixed RunONNXModel.py to better support boolean inputs and outputs.

Notable: llvm does not seems to support vectors of bits well, so ONNX ops that work on bits are not generating SIMD code at this time.


src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -685,27 +916,164 @@ Value emitScalarOpFor<ONNXRoundOp>(ConversionPatternRewriter &rewriter,
...
+    ArrayRef<Value> operands, int64_t alignment, int64_t simdUnroll) {
+  Type outputElementType = outputMemRefType.getElementType();
+
+  // generate SIMD code of VL elements per vector.
+  IndexExprScope allocScope(create.vec, shapeHelper->getScope());
+  int64_t VL =
+      create.vec.getMachineVectorLength(outputElementType) * simdUnroll;
+  // Alloc memory with padding for SIMD.
+  Value alloc = create.mem.alignedAllocWithSimdPadding(
+      outputMemRefType, shapeHelper->getOutputDims(), simdUnroll, alignment);
```
tungld, line 960: Now I understand that you create a padded buffer first then view it as the original shape. That's fine. I though you created a buffer of the original shape and viewed it as a padded buffer when looking at your slide at our meeting.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -685,27 +916,164 @@ Value emitScalarOpFor<ONNXRoundOp>(ConversionPatternRewriter &rewriter,
...
+          VectorType vecType =
+              VectorType::get({VL}, memRefType.getElementType());
+          Value loadedVal = create.vec.load(vecType, flatOper, loopInd);
+          loadedVals.emplace_back(loadedVal);
+        }
+        Value loweredOpResult = emitScalarOpFor<ElementwiseUnaryOp>(
+            rewriter, create.getLoc(), op, vecElementType, loadedVals);
+        // Store result in the resulting array.
+        create.vec.store(loweredOpResult, flatAlloc, loopInd);
+      });
```
tungld, line 999: So, reshapeFromFlat is unnecessary here, isn't it?

src/Dialect/Mlir/DialectBuilder.cpp
```diff
@@ -963,6 +976,59 @@ memref::DeallocOp MemRefBuilder::dealloc(Value val) const {
...
+// Reshape.
+
+memref::ReshapeOp MemRefBuilder::reshape(
+    MemRefType destType, Value valToReshape, Value destShapeStoredInMem) const {
+  return b().create<memref::ReshapeOp>(
+      loc(), destType, valToReshape, destShapeStoredInMem);
+}
+
+memref::ReshapeOp MemRefBuilder::reshapeToFlat(Value valToReshape,
+    llvm::SmallVectorImpl<IndexExpr> &dims, Value &size1D) const {
```
tungld, line 989: It looks to me `dims` is redundant. Are they the dims of `valToReshape`?

tungld (approved):
LGTM.

src/Dialect/Mlir/DialectBuilder.cpp
AlexandreEichenberger: yes, but since they might be already computed, I did not want to insert code to recompute them inside the call. I can make a second call without it if you prefer.

tungld: I see, that's fine as it is. Thanks!

src/Conversion/ONNXToKrnl/ObjectDetection/NonMaxSuppression.cpp
```diff
@@ -216,7 +216,6 @@ struct ONNXNonMaxSuppressionOpLowering
       ConversionPatternRewriter &rewriter) const final {
     Operation *op = nmsOp.getOperation();
     Location loc = ONNXLoc<ONNXNonMaxSuppressionOp>(op);
-    ValueRange operands = adaptor.getOperands();
```
tungld, line 219: Thanks for fixing this!

tungld (approved):
LGTM. Thanks!

PR #2054, opened 2023-03-06
Fix duplicated symbol errors in KrnlToLLVM conversion.

negiyas:
An error occurs with “Redefinition of symbol named free” in KrnlToLLVM conversion pass, because the current onnx-mlir generate each symbol with its corresponding string used in the onnx model. 

This PR prepends “om_”  to symbol names for string constants to avoid the symbol conflict with C function names and symbol names for the string constants.


AlexandreEichenberger (approved):
LGTM, thanks for the fix

PR #2056, opened 2023-03-06
Support onnx.Clip op in an elementwise unary operation.

negiyas:
This PR implements Clip op a an elementwise unary operation in the onnx-krnl conversion pass, and removes onnx.Clip specific conversion code.

This PR also includes fixes for assert errors in shape inference of all elementwise unary ops with unranked inputs and their lit test.


src/Dialect/ONNX/ONNXOps/Math/Clip.cpp
```diff
@@ -41,46 +59,80 @@ LogicalResult ONNXClipOpShapeHelper::computeShape() {
 // Shape Inference
 //===----------------------------------------------------------------------===//
 
-LogicalResult ONNXClipOp::inferShapes(
-    std::function<void(Region &)> doShapeInference) {
+template <typename T>
```
AlexandreEichenberger, line 62: Nit: we typically put all this in one line.

src/Dialect/ONNX/ONNXOps/Math/Clip.cpp
```diff
@@ -41,46 +59,80 @@ LogicalResult ONNXClipOpShapeHelper::computeShape() {
...
+LogicalResult ONNXClipOpGeneralInferShapes(
```
AlexandreEichenberger, line 63: The pattern I have been using is when we have a custom function, I add it to the class. See this example for reference:
```C++
struct ONNXBroadcastOpShapeHelper : public ONNXOpShapeHelper {
  ONNXBroadcastOpShapeHelper(mlir::Operation *op, mlir::ValueRange operands,
      IndexExprBuilder *ieBuilder = nullptr, IndexExprScope *scope = nullptr,
      bool hasUniBroadcasting = false, bool hasNoBroadcasting = false)
      : ONNXOpShapeHelper(op, operands, ieBuilder, scope), inputsDims(),
        outputRank(0), hasUniBroadcasting(hasUniBroadcasting),
        hasNoBroadcasting(hasNoBroadcasting) {}
  virtual ~ONNXBroadcastOpShapeHelper() {}

  // Custom shape compute which takes additional parameters.
  mlir::LogicalResult customComputeShape(
      mlir::ValueRange initialOperands, DimsExpr *additionalOperand);

  // Default shape compute (every operands of the operation and no additional
  // parameters).
  mlir::LogicalResult computeShape() override {
    return customComputeShape(operands, nullptr);
  }
...
```

Here the default `computeShape` uses the specialized `customComputeShape` with default parameters, and then the ones that needs special computations, say
```C++
// Helper for ExpandOp
struct ONNXExpandOpShapeHelper : public ONNXBroadcastOpShapeHelper {
  ONNXExpandOpShapeHelper(mlir::Operation *op, mlir::ValueRange operands,
      IndexExprBuilder *ieBuilder = nullptr, IndexExprScope *scope = nullptr)
      : ONNXBroadcastOpShapeHelper(op, operands, ieBuilder, scope) {}
  virtual ~ONNXExpandOpShapeHelper() {}
  mlir::LogicalResult computeShape() final;
};
```
 
are defined and in their `computeShape()` call we can put the extra processing to generate the specialized parameters that are needed for their parent class `customComputeShape` which will compute the shape with the different input params.

src/Dialect/ONNX/ONNXOps/Math/Clip.cpp
```diff
@@ -41,46 +59,80 @@ LogicalResult ONNXClipOpShapeHelper::computeShape() {
...
+    return success();
+  RankedTensorType inputTy = getInput().getType().cast<RankedTensorType>();
+  Type elementType = inputTy.getElementType();
+  ONNXClipV6OpShapeHelper shapeHelper(getOperation(), {});
   return shapeHelper.computeShapeAndUpdateType(elementType);
 }
 
+LogicalResult ONNXClipV11Op::inferShapes(
+    std::function<void(Region &)> doShapeInference) {
+  return ONNXClipOpGeneralInferShapes<ONNXClipV11Op>(
```
AlexandreEichenberger, line 119: This does not work because we expect each operation to have their `computeShape` to perform the precise computations.

AlexandreEichenberger (changes requested):
Please use the pattern where each op's `computeShape()` generate the precise info. This is needed for analysis.

src/Dialect/ONNX/ONNXOps/Math/Clip.cpp
negiyas: @AlexandreEichenberger Thank you very much for your comments.
Because output of onnx.ClipOp has the same shape and element type of its first input argument, it can be considered as an elementsize unary operation. I updated this PR by using the ONNXUnaryOpShapeHelper for onnx.Clip/V6/V11/V12 ops.

src/Dialect/ONNX/ONNXOps/Math/Clip.cpp
negiyas: The clang-format command divides the line to two lines. The line might be too long??

AlexandreEichenberger (commented):
I think its a great idea to process clip as a unary op (even though it has technically 2 additional scalar inputs).

Would you be willing to migrate the lowering of clip to Conversion/ONNXToKrnl/Math/Elementwise.cpp also?
The advantage is that it would then be simdized for free. 

You just need to migrate the code below 
```
    auto computeResult = [&](LocalDialectBuilder &create,
                             const ValueRange &indices) { // indices={i,j,k}
      Value loadedVal = create.krnl.load(input, indices); // load input[i,j,k]
      Value res = loadedVal;
      if (!isFromNone(min)) {
        Value minVal = create.krnl.load(min);             // load min
        Value lessThanMin = create.math.slt(res, minVal); // (input[i,j,k]<min)
        res = create.math.select(lessThanMin, minVal, res);
      }
      if (!isFromNone(max)) {
        Value maxVal = create.krnl.load(max);
        Value lessThanMax = create.math.slt(res, maxVal);
        res = create.math.select(lessThanMax, res, maxVal);
      }
      create.krnl.store(res, alloc, indices);
    };
```
to the templated function 
```
template <>
Value emitScalarOpFor<ONNXHardSigmoidOp>(ConversionPatternRewriter &rewriter,
```

That would be a big simplification of the code. Thanks

src/Dialect/ONNX/ONNXOps/ShapeHelper.hpp
```diff
@@ -703,7 +707,12 @@ struct ONNXNonSpecificOpShapeHelper : public ONNXOpShapeHelper {
 // clang-format off
 using ONNXBatchNormalizationInferenceModeOpShapeHelper = ONNXNonSpecificOpShapeHelper<mlir::ONNXBatchNormalizationInferenceModeOp>;
 using ONNXCategoryMapperOpShapeHelper = ONNXNonSpecificOpShapeHelper<mlir::ONNXCategoryMapperOp>;
+#if 0
 using ONNXClipOpShapeHelper = ONNXNonSpecificOpShapeHelper<mlir::ONNXClipOp>;
+using ONNXClipV9OpShapeHelper = ONNXNonSpecificOpShapeHelper<mlir::ONNXClipV9Op>;
```
tungld, line 712: Is it V6 or V9?

src/Dialect/ONNX/ONNXOps/ShapeHelper.hpp
```diff
@@ -703,7 +707,12 @@ struct ONNXNonSpecificOpShapeHelper : public ONNXOpShapeHelper {
 // clang-format off
 using ONNXBatchNormalizationInferenceModeOpShapeHelper = ONNXNonSpecificOpShapeHelper<mlir::ONNXBatchNormalizationInferenceModeOp>;
 using ONNXCategoryMapperOpShapeHelper = ONNXNonSpecificOpShapeHelper<mlir::ONNXCategoryMapperOp>;
+#if 0
```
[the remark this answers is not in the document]
negiyas: Fixed. Thank you!

src/Dialect/ONNX/ONNXOps/ShapeHelper.hpp
negiyas: Fixed to V6. Thank you!!

negiyas:
@Alex Thank you for the comments.
I will also migrate its conversion code according to your suggestions.

src/Dialect/ONNX/ONNXOps/Math/Clip.cpp
negiyas: Because onnx.Cilp op is treated as an elementwise unary operation (like LeakyRelu), the logic is integrated into src/Dialect/ONNX/ONNXOps/Math/Elementwise.cpp, and src/Dialect/ONNX/ONNXOps/Math/Clip.cpp is removed.

negiyas:
@AlexandreEichenberger
Updated the code. Thank you for 

```
   auto computeResult = [&](LocalDialectBuilder &create,
                             const ValueRange &indices) { // indices={i,j,k}
      Value loadedVal = create.krnl.load(input, indices); // load input[i,j,k]
      Value res = loadedVal;
      if (!isFromNone(min)) {
        Value minVal = create.krnl.load(min);             // load min
        Value lessThanMin = create.math.slt(res, minVal); // (input[i,j,k]<min)
        res = create.math.select(lessThanMin, minVal, res);
      }
      if (!isFromNone(max)) {
        Value maxVal = create.krnl.load(max);
        Value lessThanMax = create.math.slt(res, maxVal);
        res = create.math.select(lessThanMax, res, maxVal);
      }
      create.krnl.store(res, alloc, indices);
    };
```

For an optimization, krnl.load ops to load min and max are removed from emitScalarOpFor<ONNXClipOp>, and are moved to  ONNXElementwiseUnaryOpLowering to load min and  max in advance out of the loop. 

```
template <>
Value emitScalarOpFor<ONNXClipOp>(ConversionPatternRewriter &rewriter,
  MultiDialectBuilder<KrnlBuilder, MathBuilder> create(rewriter, loc);
  Value res = scalarOperands[0];
  Value min = scalarOperands[1];
  Value max = scalarOperands[2];
  if (!isFromNone(min)) {
    Value lessThanMin = create.math.slt(res, min); // (input[i,j,k]<min)
    res = create.math.select(lessThanMin, min, res);
  }
  if (!isFromNone(max)) {
    Value lessThanMax = create.math.slt(res, max); // (input[i,j,k]>max)
    res = create.math.select(lessThanMax, res, max);
  }
  return res;
}
```

tungld:
> For an optimization, krnl.load ops to load min and max are removed from emitScalarOpFor, and are moved to ONNXElementwiseUnaryOpLowering to load min and max in advance out of the loop.

It seems to me these loads will be moved out automatically when lowering krnl to affine, and this modification here is redundant. Could you try to call `convert-krnl-to-affine` to see if the loads are automatically moved outside the loop?

negiyas:
@tungld Thanks. I reverted modifications for the optimization.

tungld (commented):
High-level question: we rewrite ClipV6, ClipV11 and ClipV12 in Decompose.td: https://github.com/onnx/onnx-mlir/blob/main/src/Transform/ONNX/Decompose.td#L294, so they will not present after calling ``onnx-decompose` pass. So why we need shape inference for them?

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1552,6 +1596,8 @@ void populateLoweringONNXElementwiseOpPattern(RewritePatternSet &patterns,
       typeConverter, ctx, enableSIMD);
   patterns.insert<ONNXElementwiseBinaryOpLowering<mlir::ONNXPReluOp>>(
       typeConverter, ctx, enableSIMD, /*isUniBroadcasting=*/true);
+  patterns.insert<ONNXElementwiseUnaryOpLowering<mlir::ONNXClipOp>>(
+      typeConverter, ctx, /*enableSIMD=*/false, /*numScalarArgs=*/2);
```
tungld, line 1600: It looks like this is not the way to turn off SIMD for an operation. As I understand we use `using SimdEnabled = NoSimdScalarOp;` in ScalarOp template. @AlexandreEichenberger is this correct?

negiyas: Thanks. This part is removed.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1140,6 +1172,10 @@ struct ONNXElementwiseUnaryOpLowering
     Value alloc = create.mem.alignedAlloc(
         memRefType, shapeHelper.getOutputDims(), alignment);
 
+    std::vector<Value> scalarArgs;
```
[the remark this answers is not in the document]
negiyas: Fixed. Thank you.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1149,15 +1185,23 @@ struct ONNXElementwiseUnaryOpLowering
...
           });
     } else {
       Value loadedVal = create.krnl.load(X);
+      SmallVector<Value, 4> args;
+      args.emplace_back(loadedVal);
+      for (int i = 0; i < numScalarArgs; i++)
+        args.emplace_back(scalarArgs[i]);
       auto loweredOpResult = emitScalarOpFor<ElementwiseUnaryOp>(
-          rewriter, loc, op, elementType, {loadedVal});
+          rewriter, loc, op, elementType, args);
```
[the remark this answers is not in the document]
negiyas: @tungld 
Fixed the code to pass an array consisting of the loaded value of operand[0] and all other operands except operand[0] 
as scalarOperands argument of emitScalrOpFor< ElementwizeUnaryOps>.
The "numScalarArgs" variable is removed from the code. 
Thank you!

negiyas:
@tungld 
> High-level question: we rewrite ClipV6, ClipV11 and ClipV12 into the latest ClipOp version in Decompose.td: https://github.com/onnx/onnx-mlir/blob/main/src/Transform/ONNX/Decompose.td#L294, so they will not present after calling ``onnx-decompose` pass. So why do we need shape inference for them?

I see. I removed shape-inference for onnxV6/V11/V12 and their lit tests from this PR.

tungld (approved):
LGTM. Thanks for migrating the code for Clip!

Please update the patch's title and main comment accordingly, since they are not descriptive for the current code.

negiyas:
@tungld Updated the title. Thanks !

tungld:
> @tungld Updated the title. Thanks !

Thanks! please make sure to update the main comment also.

tungld:
@negiyas could you please edit the first comment of this PR at the top of this page to reflect the current code? the comment starts with:

```
onnx.ClipV6/V11/V12 compilation is supported by onnx-mlir, but their shape-inference is not implemented yet.
So that, clipV6 lit test causes the following errors, and the situations about ClipV11/12 are the same. This PR implements their shape-inference.

```

Thanks!

negiyas:
@tungld Updated the first comment of this PR at the top of this page. Thank you!

tungld:
In the comment: 

> This PR also fixes the shape inference errors of onnx.ClipV6/V11/V12 op, too.

I believe this patch does nothing about the shape inference of `ClipV6/V11/V12` ops, doesn't it? and I am not sure what is the problem with the shape inference of `ClipV6/V11/V12` ops. These operations are decomposed into ClipOp, so they should not exist in following generated code.

negiyas:
@tungld Sorry. I forgot to commit fixes for assert errors in shape inference of all elementwise unary ops with unranked inputs and their lit test.
I added the fixes, and updated the first comment of this PR in . Thank you.

src/Dialect/ONNX/ONNXOps/Math/ElementwiseUnary.cpp
```diff
@@ -32,6 +32,8 @@ namespace onnx_mlir {
 /// Handle shape inference for unary element-wise operators.
 LogicalResult inferShapeForUnaryOps(Operation *op) {
   Value input = op->getOperand(0);
+  if (!hasShapeAndRank(input))
+    return success();
```
tungld, line 36: Thanks for the fix!

negiyas:
@AlexandreEichenberger Can you kindly review the change request, again?

> Please use the pattern where each op's `computeShape()` generate the precise info. This is needed for analysis.
The file including the commented part (src/Dialect/ONNX/ONNXOps/Math/Clip.cpp) is removed by PR.

test/mlir/onnx/onnx_shape_inference.mlir
```diff
@@ -43,6 +43,18 @@ func.func @test_default_unary_elementwise_user_shape_3(%arg0 : tensor<?x3x4xf32>
 
 // -----
 
+// COM: Check if unranked shape input can be handled without crashing
```
AlexandreEichenberger, line 46: I assume that the code for clip did not change

AlexandreEichenberger (approved):
LGTM, thanks for the updates

tungld:
@AlexandreEichenberger @negiyas since we will have a new way to define SIMD soon by PR #2127. Could you land this non-SIMD version first? I would like to use Clip in DynamicDequantizeLinear. Thanks!

negiyas:
@tungld Thanks. I defer to enable SIMD onnx.Clip.

AlexandreEichenberger:
Will add support for SIMD clip in a subsequent PR

AlexandreEichenberger:
Support for SIMD and integer (instead of just float) in #2131

Issue #1844, opened 2022-11-08
Unable to compile ResNet50 model from Keras


cjvolzka:
Attempting to compile example ResNet50 model from Keras using `--O3 --EmitLib --mtriple=s390x-ibm-loz --mcpu=z14 ResNet50.onnx --onnx-op-stats TXT -v` fails with:

```
Operations encountered:
-----------------------
  func.func             , 1
  func.return           , 1
  onnx.Add              , 46
  onnx.Concat           , 59
  onnx.Constant         , 122
  onnx.Conv             , 23
  onnx.Dim              , 29
  onnx.Gemm             , 1
  onnx.MatMul           , 30
  onnx.MaxPoolSingleOut , 1
  onnx.Pad              , 1
  onnx.ReduceMean       , 1
  onnx.Relu             , 49
  onnx.Reshape          , 59
  onnx.Softmax          , 1
  onnx.Squeeze          , 1
  onnx.Transpose        , 1
onnx-mlir: /build/llvm-project/mlir/lib/Dialect/Affine/IR/AffineOps.cpp:2723: static void mlir::AffineLoadOp::build(mlir::OpBuilder&, mlir::OperationState&, mlir::Value, mlir::AffineMap, mlir::ValueRange): Assertion `map.getNumInputs() == mapOperands.size() && "inconsistent index info"' failed.
```

At a glance, I'm not sure why it is failing. I used the most recent versions of tensorflow and tf2onnx for Mac (x86). Looking at the netron graphy of the model, I don't see any Ops that jump out as unsupported.

Note: I tried this with `main` and the 0.3.1 and 0.3.0 releases and got the same result.

To create and save the model: (requires tensorflow)
```
import tensorflow
tensorflow.keras.applications.resnet50.ResNet50(weights='imagenet').save("ResNet50")
```
To convert to ONNX: (requires tf2onnx)
```
python -m tf2onnx.convert --saved-model ResNet50/ --output ResNet50.onnx
```

chentong319:
Can you use gdb to locate where that error occurred? Or attach your ResNet50.onnx here.

cjvolzka:
Is there someplace I can upload the model? It's too large to attach to github directly.

xlindo:
Similar to https://github.com/onnx/onnx-mlir/issues/1571

chentong319:
I did some investigation with Mike today. Here is the outcome:
This krnl.load caused the assertion error when krnl is lowered to affine:
```
[...]

PR #2060, opened 2023-03-08
Fix issue compiling ResNet50 model from Keras

mikeessen:
This update fixes issue https://github.com/onnx/onnx-mlir/issues/1844 which is a compile error with ResNet50 model from Keras.

### DIFF
--- a/src/Conversion/ONNXToKrnl/Tensor/Pad.cpp
+++ b/src/Conversion/ONNXToKrnl/Tensor/Pad.cpp
@@ -31,7 +31,6 @@ struct ONNXPadOpLowering : public OpConversionPattern<ONNXPadOp> {
     ValueRange operands = adaptor.getOperands();
     Value data = adaptor.getData();
     Value constantValue = adaptor.getConstantValue();
-
     StringRef padMode = adaptor.getMode();
 
     // Builder helper.
--- a/src/Conversion/ONNXToKrnl/Tensor/Pad.cpp
+++ b/src/Conversion/ONNXToKrnl/Tensor/Pad.cpp
@@ -69,22 +69,25 @@ struct ONNXPadOpLowering : public OpConversionPattern<ONNXPadOp> {
       // This way is to avoid using `select` in computing indices as doing for
       // 'edge' and 'reflect' modes.
       Value cValue;
-      SmallVector<Value, 1> loadIndices;
-      if (MemRefType constantValueType =
-              constantValue.getType().dyn_cast<MemRefType>()) {
-        if (constantValueType.getElementType().isF32() &&
-            constantValueType.getRank() == 1) {
-          // If the constant_value type is 1xf32 do krnl.load with an index of
-          // 0 to avoid an assertion failure in AffineLoadOp::build().
-          Value idx = create.math.constantIndex(0);
-          loadIndices = {idx};
-        }
-      }
       if (constantValue.getType().isa<NoneType>()) {
         // Default to 0 if constant_value is not specified.
         cValue = create.math.constant(resElementType, 0);
-      } else
+      } else {
+        SmallVector<Value, 1> loadIndices;
+        if (MemRefType constantValueType =
+                constantValue.getType().dyn_cast<MemRefType>()) {
+          if (constantValueType.getElementType().isF32() &&
+              constantValueType.getRank() == 1) {
+            // If the constant_value type is 1xf32 do krnl.load with an index of
+            // 0 to avoid an assertion failure in AffineLoadOp::build(). This is
+            // a work around for an incorrect ResNet50 model. The issue this
+            // fixes is issue number 1844.
+            Value idx = create.math.constantIndex(0);
+            loadIndices = {idx};
+          }
+        }
         cValue = create.krnl.load(constantValue, loadIndices);
+      }
 
       // Initialize the result to the constant value.
       create.krnl.memset(resMemRef, cValue);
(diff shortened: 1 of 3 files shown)


src/Conversion/ONNXToKrnl/Tensor/Pad.cpp
```diff
@@ -68,9 +68,23 @@ struct ONNXPadOpLowering : public OpConversionPattern<ONNXPadOp> {
...
+          // 0 to avoid an assertion failure in AffineLoadOp::build().
+          useConstantIndexZero = true;
+        }
+      }
       if (constantValue.getType().isa<NoneType>()) {
         // Default to 0 if constant_value is not specified.
         cValue = create.math.constant(resElementType, 0);
+      } else if (useConstantIndexZero == true) {
+        // Do krnl.load with an index of 0 when constant_value type is 1xf32.
+        auto idx = create.math.constantIndex(0);
```
AlexandreEichenberger, line 86: Nit: only use `auto` for lambda functions or really ugly types. Please replace with `Value`

src/Conversion/ONNXToKrnl/Tensor/Pad.cpp
```diff
@@ -68,9 +68,23 @@ struct ONNXPadOpLowering : public OpConversionPattern<ONNXPadOp> {
       // This way is to avoid using `select` in computing indices as doing for
       // 'edge' and 'reflect' modes.
       Value cValue;
+      bool useConstantIndexZero = false;
```
AlexandreEichenberger, line 71: Code as is is ok, code would be more readable and easier if you did the following:
```C++
  SmallVector<Value, 1> loadIndices;
```
and simply adding `constant zero` to it where you currently set `useConstantIndexZero = true` and then calling the load with that `loadIndices` set that is either empty or has one 0 value.

src/Conversion/ONNXToKrnl/Tensor/Pad.cpp
mikeessen: Update to not use auto.  Thanks!

src/Conversion/ONNXToKrnl/Tensor/Pad.cpp
mikeessen: Good idea!  Updated as you suggested.  Thanks!

mikeessen:

mikeessen:

src/Conversion/ONNXToKrnl/Tensor/Pad.cpp
```diff
@@ -68,11 +69,22 @@ struct ONNXPadOpLowering : public OpConversionPattern<ONNXPadOp> {
...
+      if (MemRefType constantValueType =
+              constantValue.getType().dyn_cast<MemRefType>()) {
+        if (constantValueType.getElementType().isF32() &&
+            constantValueType.getRank() == 1) {
+          // If the constant_value type is 1xf32 do krnl.load with an index of
+          // 0 to avoid an assertion failure in AffineLoadOp::build().
+          Value idx = create.math.constantIndex(0);
+          loadIndices = {idx};
+        }
+      }
```
chentong319, line 82: I would move this section of code into the else part of isa<NoneType> because you need this only when the type isa<NoneType> is false.

src/Conversion/ONNXToKrnl/Tensor/Pad.cpp
```diff
@@ -68,11 +69,22 @@ struct ONNXPadOpLowering : public OpConversionPattern<ONNXPadOp> {
       // This way is to avoid using `select` in computing indices as doing for
       // 'edge' and 'reflect' modes.
       Value cValue;
+      SmallVector<Value, 1> loadIndices;
+      if (MemRefType constantValueType =
+              constantValue.getType().dyn_cast<MemRefType>()) {
+        if (constantValueType.getElementType().isF32() &&
+            constantValueType.getRank() == 1) {
+          // If the constant_value type is 1xf32 do krnl.load with an index of
```
chentong319, line 77: You can add the issue number in the comment.
Please clearly comment that this fix is a workaround for incorrect model. Could we print a warning message here using llvm warning mechanism?

src/Conversion/ONNXToKrnl/Tensor/Pad.cpp
mikeessen: Moved this code into else part of isa.  Thank you.

src/Conversion/ONNXToKrnl/Tensor/Pad.cpp
mikeessen: Updated comment.  Tried the emit warning and it caused onnx-mlir --EmitLib to fail.  Thank you.

src/Conversion/ONNXToKrnl/Tensor/Pad.cpp
```diff
@@ -71,8 +72,22 @@ struct ONNXPadOpLowering : public OpConversionPattern<ONNXPadOp> {
       if (constantValue.getType().isa<NoneType>()) {
         // Default to 0 if constant_value is not specified.
         cValue = create.math.constant(resElementType, 0);
-      } else
-        cValue = create.krnl.load(constantValue, {});
+      } else {
+        SmallVector<Value, 1> loadIndices;
+        if (MemRefType constantValueType =
```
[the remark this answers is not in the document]
mikeessen: Removed the conditional check for MemRefType.  Thanks!

chentong319 (approved):
LGTM!

PR #2062, opened 2023-03-08
OnnxBuilder now use create and infer shape

AlexandreEichenberger:
Added infrastructure for a "create op and infer shapes" and used it in the OnnxBuilder.

This change allows the users of the builder not to have to manually infer the shapes. So potentially, code that invoke the builder could be simplified. I have not performed this simplification at this time. It is just that when adding new ops, this step is now automatically done by the shape inference interface.


tungld (approved):
LGTM

Issue #2063, opened 2023-03-08
Bump minimum cmake version to 3.20


sstamenova:
Right now the minimum required cmake version for onnx-mlir is 3.13.4 to match the previous minimum required version for llvm and mlir. llvm recently bumped the minimum version to 3.20 (https://reviews.llvm.org/D144509), so now seems the right time for onnx-mlir to move to 3.20 as well.

@gongsu832 or anyone else: Do you have any concerns with this? If not, I'll send a PR shortly.

sstamenova:
We do have a bit of a reprieve until all buildbots are updated to 3.20 and the change is re-committed to llvm, but after that happens, we won't be able to build newer versions of llvm on the jenkins machines since they have an older version.

PR #2064, opened 2023-03-08
Bump cmake minimum required version to 3.20.0

sstamenova:
This matches the current minimum required version in llvm (see https://reviews.llvm.org/D144509)

See https://github.com/onnx/onnx-mlir/issues/2063 as well


NathanielMcVicar:
LGTM, although somewhat painful as described in the linked issue.

gongsu832:
Unfortunately Ubuntu Focal, which our dev docker image is based on, is only up to cmake 3.16.3.

sstamenova:
> Unfortunately Ubuntu Focal, which our dev docker image is based on, is only up to cmake 3.16.3.

How are we going to address that once we try to update llvm to a commit that requires 3.20?

gongsu832:
I will have to look into upgrading the base image to Jammy, which will have cmake 3.22.1.

sstamenova:
> I will have to look into upgrading the base image to Jammy, which will have cmake 3.22.1.

@gongsu832 Ok, I will leave this PR up to be completed once the base image supports it. I am not sure when (though I suspect very soon) LLVM will re-commit the new version requirement and one of the next llvm updates will need this.

NathanielMcVicar:
Another option is to add the apt.kitware repo to the docker image setup. That might be lower friction until you are ready to move to jammy.

gongsu832:
I'll try Jammy and see how many things break and proceed with it if not too many. Otherwise I'll look into apt.kitware. Thanks for the suggestion.

gongsu832:
@sstamenova I'm wondering, if we upgrade our base image to Jammy, we will have to also upgrade the LLVM no-warning buildbot to Jammy, right?

sstamenova:
> @sstamenova I'm wondering, if we upgrade our base image to Jammy, we will have to also upgrade the LLVM no-warning buildbot to Jammy, right? Since the buildbot config is supposed to match our build env.

Yes, whichever way you use to upgrade cmake here will need to be applied to the buildbot since it will also require the newer version of cmake.

gongsu832:
> Yes, whichever way you use to upgrade cmake here will need to be applied to the buildbot since it will also require the newer version of cmake.

I tried Jammy and it was eventless. Everything pretty much just worked. So I will put out a PR to upgrade to Jammy. Once that's done, can I just create another Jammy based docker container for the LLVM no-warning buildbot, replicating all the configurations from the Focal based buildbot, and then shutting down the Focal based buildbot?

gongsu832:
> Assuming you connect with the same worker name and password, it should work.

Great. I'll give that a shot once we upgrade to Jammy. Thanks.

MikeHolman:

NathanielMcVicar:
@gongsu832 it looks like the CMake version on the Jenkins builds is still below 3.20. Is there an additional step we should take to update these? Thanks!

gongsu832:
It's being addressed by https://github.com/onnx/onnx-mlir/pull/2234

NathanielMcVicar:
> @NathanielMcVicar Windows CI appears to be stuck not reporting status. Is there any way to trigger it manually?

I see it reporting currently, can you point me to an example? Sorry, I missed this on Friday.

AlexandreEichenberger:
@gongsu832 are the conditions that were holding us to a lower cmake minimum version still holding us at this time?

AlexandreEichenberger:
I will defer this question for @gongsu832

gongsu832:
Dev image is based on Ubuntu Jammy which comes with cmake 3.22.1 and user image is based on Redhat ubi8 which comes with cmake 3.20.2. So bumping up to 3.20.0 shouldn't be a problem.

NathanielMcVicar (approved):

PR #2065, opened 2023-03-08
ssize_t is deprecated as of Python 3.10

NathanielMcVicar:
Python 3.10 and later requires using Py_ssize_t, as documented here https://docs.python.org/3/c-api/arg.html#strings-and-buffers.


sstamenova (approved):

NathanielMcVicar:
> It's strange to me that why the intermediate `const_cast<ssize_t *>` is needed to begin with. Wouldn't `const_cast<int64_t *>` just work?

It seems to work for me locally, but @Jiaqing-ASU please let us know if there is some subtle Python capi issue here, as I am not very familiar.

gongsu832:
```
/Users/runner/work/onnx-mlir/onnx-mlir/src/Runtime/PyExecutionSession.cpp:91:9: error: const_cast from 'const pybind11::ssize_t *' (aka 'const long *') to 'int64_t *' (aka 'long long *') is not allowed
        const_cast<int64_t *>(inputPyArray.shape()),
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
So it looks like the extra cast is to make AppleClang happy. Would be nice to have a comment there.

Jiaqing-ASU:
> > It's strange to me that why the intermediate `const_cast<ssize_t *>` is needed to begin with. Wouldn't `const_cast<int64_t *>` just work?
> 
> It seems to work for me locally, but @Jiaqing-ASU please let us know if there is some subtle Python capi issue here, as I am not very familiar.

Hello @NathanielMcVicar. As the errors shown in the `MacOS amd64 / build`, it will give errors after removing the intermediate `const_cast<ssize_t *>`. I remembered once I met this kind of errors and that is why I put the `const_cast<ssize_t *>` in the middle. But there might be other solutions.

NathanielMcVicar:
@gongsu832 This seems dangerous to me, as it suggests on some platforms the type of elements in the array returned by the Python api is not int64_t, which would be effectively hidden but not fixed by the double cast. Is this possible and we just don't have any tests catching the incorrect behavior?

gongsu832:
@NathanielMcVicar I wonder about that too. The source type is `'const pybind11::ssize_t *' (aka 'const long *')`. On 64-bit Linux,long is 8 bytes so casting it to int64 is fine. But on MacOS (maybe Windows too) long is 4 bytes. So casting it to int64_t will definite break things. We got by so far only because we don't really test that part of the code.

Going back to double cast doesn't really fix the problem. I think we somehow have to make the source type pybind11::some_8_byte_type. Or maybe instead of casting, we need to construct an int64_t array out of a long array. Unfortunately I don't know C++ enough to have a quick try. Do you think that's possible?

NathanielMcVicar:
@Jiaqing-ASU This should resolve any issues with the safety of the casts, but please confirm that I haven't broken the Python bindings in the process. The tests in mnist-example seem to work.

Jiaqing-ASU:
> @Jiaqing-ASU This should resolve any issues with the safety of the casts, but please confirm that I haven't broken the Python bindings in the process. The tests in mnist-example seem to work.

I haven't seen any potential issues from the changed files and hopefully it can pass all the checks.

NathanielMcVicar:
Does anyone have any insight into what's happening here. I don't see how my changes could be causing these warnings, but I also don't see them popping up on the CI builds for other PRs?
```
[ 49%] Built target OMShapeHelperOpInterface
In file included from /workdir/llvm-project/mlir/include/mlir/IR/AttrTypeSubElements.h:20,
                 from /workdir/llvm-project/mlir/include/mlir/IR/StorageUniquerSupport.h:16,
                 from /workdir/llvm-project/mlir/include/mlir/IR/AttributeSupport.h:17,
                 from /workdir/llvm-project/mlir/include/mlir/IR/Attributes.h:12,
                 from /workdir/llvm-project/mlir/include/mlir/IR/BuiltinAttributeInterfaces.h:13,
                 from /workdir/llvm-project/mlir/include/mlir/IR/BuiltinTypes.h:12,
                 from /workdir/llvm-project/mlir/include/mlir/Dialect/Affine/IR/AffineMemoryOpInterfaces.h:17,
                 from /workdir/llvm-project/mlir/include/mlir/Dialect/Affine/IR/AffineOps.h:17,
                 from /workdir/onnx-mlir/src/Accelerators/NNPA/Transform/ZLow/ZLowRewrite.cpp:15:
/workdir/llvm-project/llvm/include/llvm/ADT/DenseMap.h: In member function 'virtual void onnx_mlir::zlow::ZLowRewritePass::runOnOperation()':
/workdir/llvm-project/llvm/include/llvm/ADT/DenseMap.h:1178:18: warning: '*((void*)(& removableStickOps)+16).llvm::SmallDenseMap<onnx_mlir::zlow::ZLowStickOp, llvm::detail::DenseSetEmpty, 4, llvm::DenseMapInfo<onnx_mlir::zlow::ZLowStickOp, void>, llvm::detail::DenseSetPair<onnx_mlir::zlow::ZLowStickOp> >::LargeRep::NumBuckets' may be used uninitialized in this function [-Wmaybe-uninitialized]
     return Small ? InlineBuckets : getLargeRep()->NumBuckets;
            ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

PR #2067, opened 2023-03-09
Fix wrong indices of onnx types

tungld:
`getTypeMap` in ONNX op definition defines element type for each output, e.g. `{1, 7, 1}`. Values, e.g. `1` and `7` are the indices of types in the list `onnx_types` in `gen_onnx_mlir.py`:
```
onnx_types = (
    'bool', 'int8', 'int16', 'int32', 'int64', 'unknown', 'float16',
    'float', 'double', 'complex64', 'complex128', 'string'
)
```

However this `onnx_types` is not reflecting the correct order of onnx types defined in ONNX https://github.com/onnx/onnx/blob/main/onnx/onnx.in.proto#L481, that is
```
enum DataType {
    UNDEFINED = 0;
    // Basic types.
    FLOAT = 1;   // float
    UINT8 = 2;   // uint8_t
    INT8 = 3;    // int8_t
    UINT16 = 4;  // uint16_t
    INT16 = 5;   // int16_t
    INT32 = 6;   // int32_t
    INT64 = 7;   // int64_t
    STRING = 8;  // string
    BOOL = 9;    // bool

    // IEEE754 half-precision floating-point format (16 bits wide).
    // This format has 1 sign bit, 5 exponent bits, and 10 mantissa bits.
    FLOAT16 = 10;

    DOUBLE = 11;
    UINT32 = 12;
    UINT64 = 13;
    COMPLEX64 = 14;     // complex with float32 real and imaginary components
    COMPLEX128 = 15;    // complex with float64 real and imaginary components

    // Non-IEEE floating-point format based on IEEE754 single-precision
    // floating-point number truncated to 16 bits.
    // This format has 1 sign bit, 8 exponent bits, and 7 mantissa bits.
    BFLOAT16 = 16;

    // Future extensions go here.
  }
```

As a result, types are wrong when importing a model. For example, when importing `onnx.DynamicQuantizeLinear` in the quantized bertsquad, its output element types should be `uint8`, `f32`, `unint8` that is `{2, 1, 2}` for `getTypeMap`, but the current `getTypeMap` for `DynamicQuantizeLinear` is `{1, 7, 1}` that is wrong.

This patch is to fix such wrong onnx type issue by using the list of types from ONNX.


chentong319 (approved):
Thanks for the update code and comments.

PR #2068, opened 2023-03-09
[MHLO] DisposableElementsAttr to mhlo,  ConstProp for Equal and ConstantOfShape

chenchongsong:
- Converts ONNX Dialect with DisposableElementsAttr to Mhlo
- Supports ConstProp for Equal and ConstantOfShape Op

@AlexandreEichenberger @tungld @yaochengji Please kindly have a look :)


src/Transform/ONNX/ConstProp.cpp
```diff
@@ -843,6 +848,27 @@ Value ConstPropReshape(
...
+Value ConstPropConstantOfShape(PatternRewriter &rewriter, Value replacingValue,
+    Value shape, Attribute value) {
+  ConstPropCounters::count("ConstantOfShape", {shape});
+  llvm::SmallVector<int64_t, 4> shapeVector;
+  ElementsAttr shapeAttr =
+      getONNXConstantOp(shape).getValueAttr().cast<ElementsAttr>();
+  for (mlir::IntegerAttr dim : shapeAttr.getValues<IntegerAttr>())
+    shapeVector.push_back(dim.getInt());
+
+  ElementsAttr constElements = value.cast<ElementsAttr>();
```
sorenlassen, line 864: please add a comment here or in ConstProp.td explaining that we rely on value always being set, namely by the (questionable) "canonicalization" in ONNXConstantOfShapeOp::inferShapes()

(the casual reader my wonder about this given that the onnx spec says that the value attribute is optional)

src/Transform/ONNX/ConstProp.cpp
```diff
@@ -843,6 +848,27 @@ Value ConstPropReshape(
...
+//===----------------------------------------------------------------------===//
+
+Value ConstPropConstantOfShape(PatternRewriter &rewriter, Value replacingValue,
+    Value shape, Attribute value) {
+  ConstPropCounters::count("ConstantOfShape", {shape});
+  llvm::SmallVector<int64_t, 4> shapeVector;
+  ElementsAttr shapeAttr =
+      getONNXConstantOp(shape).getValueAttr().cast<ElementsAttr>();
+  for (mlir::IntegerAttr dim : shapeAttr.getValues<IntegerAttr>())
+    shapeVector.push_back(dim.getInt());
```
sorenlassen, line 862: I think you can do this simpler:
```
llvm::SmallVector<int64_t, 4> shapeVector(shapeAttr.getValues<int64_t>());
```

src/Transform/ONNX/ConstProp.cpp
```diff
@@ -843,6 +848,27 @@ Value ConstPropReshape(
...
+
+  ElementsAttr constElements = value.cast<ElementsAttr>();
+  OnnxElementsAttrBuilder elementsBuilder(rewriter.getContext());
+  ElementsAttr expandedElements =
+      elementsBuilder.expand(constElements, shapeVector);
```
sorenlassen, line 867: if shapeVector is empty (the result is a scalar) then this stretches the functionality of expand because we "contract" constElements's 1D shape to 0D

looking at the implementation of expand() it's possible that it could work but can you please add a lit test for this? and maybe it's best to special case this with something like:
```
if (shapeVector.empty()) expandedElements = elementsBuilder.reshape(constElements, shapeVector);
```

tungld:
@chenchongsong thanks for adding ConstProp for Equal and ConstantOfShape! Please make changes according to the comments from @sorenlassen.

chenchongsong:
Thanks for the review from @sorenlassen ! These comments are very helpful. Changes have been made accordingly. Could you have a look again? @sorenlassen @tungld

tungld:

src/Conversion/ONNXToMhlo/Tensor/Constant.cpp
```diff
@@ -13,6 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "src/Conversion/ONNXToMhlo/ONNXToMhloCommon.hpp"
+#include "src/Dialect/ONNX/ElementsAttr/DisposableElementsAttr.hpp"
```
tungld, line 16: Do we still need to include this header file?

tungld (approved):
LGTM. Thanks for the patch!

Please wait for @sorenlassen's comments before merging this.

sorenlassen (approved):
LGTM, thanks for making the changes

tungld:

tungld:
Will merge this once all required Jenkins CIs passed. Thanks!

PR #2073, opened 2023-03-10
Introduce onnx.Scan importer to set unranked shape for onnx.Scan outputs to utilize shape inference for setting their shapes.

negiyas:
An extra leading dimension need to be inserted for some shapes of onnx.Scan outputs for some models, such as Bidaf-9 in the model zoo.
This PR introduces onnx.Scan importer to set unranked shape for onnx.Scan outputs to utilize shape inference for setting their shapes.


chentong319:
If possible, encapsulate ScanOp specific code in importScan and do not complicate buildOutputAndOperation. What if another Op requires special handling? Add another parameter? I do not know how to reorganize the code. If you do not see a better way, I will approve this PR.

chentong319:
Is it a bug in the model that extra outputs need to be added?

negiyas:
@chentong319 Thanks for the comments.

> Is it a bug in the model that extra outputs need to be added?

I personally do not suppose that it is a bug in the model and a special handling is necessary for the **the scan_outputs part of onnx.Scan outputs**, because other systems, such as onnxruntime (https://onnxruntime.ai/) or netron (https://netron.app/) inserts the leading dimension (representing the loop iterations) for the while importing the model.  

This PR simply ignores shapes for the **the scan_outputs part of onnx.Scan outputs** at import time, and set them at shape inference time.  This PR changes shapes of the part only, and does not change shapes for other outputs.

```
onnx.Scan

Inputs (1 - ∞)
initial_state_and_scan_inputs (variadic, heterogeneous) : V
    Initial values of the loop's N state variables followed by M scan_inputs
Outputs (1 - ∞)
final_state_and_scan_outputs (variadic, heterogeneous) : V
    Final values of the loop's N state variables followed by K scan_outputs 
```

We can consider another strategy with ignoring shapes of ALL outputs of all operators in original models, because some models include wrong shapes and shape inference can recover them, as you know. This may be a better strategy, but the impacts may be larger.
Your comments are very welcome.

chentong319:
To me, it still could be an error in the onnx model generator. onnxruntime and neutron are just consumer of onnx model. 

I agree on that the shape is tensor for Scan can be determined by out shape inference. But your changes are in the Builder, FrontendDialectTransformer.cpp. Is it possible that the input of Scan Op is unranked tensor when the model is imported and there is no way to insert an extra dimension? 
My previous comment is about the code structure in FrontendDialectTransformer.cpp. If the change is moved shape inference, that concern will disappear.

negiyas:

chentong319:
That makes sense. Can you pass unranked type to buildOutputAndOperation through `givenOutputTypes` parameter? I'd like to avoid the change for scan op inside buildOutputAndOperation.

chentong319 (approved):
LGTM!

PR #2076, opened 2023-03-12
Add targets to select node or model only backend tests

gongsu832:
- Add -t|--type option in test.py to select backend node or model tests
- Add cmake targets for backend node or model tests
      check-onnx-backend-node      (run node  tests only, C/C++)
      check-onnx-backend-node-jni  (run node  tests only, JNI)
      check-onnx-backend-model     (run model tests only, C/C++)
      check-onnx-backend-model-jni (run model tests only, JNI)
- Fix build-run-onnx-lib.sh to deal with symlink in directory name


test/backend/variables.py
```diff
@@ -164,6 +164,14 @@ def get_args_from_env():
         default=os.getenv("TEST_OPTLEVEL", "3"),
         help="set compiler optimization level (default: 3 if TEST_OPTLEVEL env var not set)",
     )
+    parser.add_argument(
+        "-t",
+        "--type",
+        type=str,
+        choices=["node", "model"],
+        default=os.getenv("TEST_TYPE", ""),
```
chentong319, line 172: Is empty string a choice?

chentong319: I see that all test cases are selected if type is not defined.

chentong319 (approved):
LGTM!

test/backend/variables.py
gongsu832: Yes. `if not args.type` will be true if the option is an empty string and all tests will be enabled by default.

PR #2078, opened 2023-03-13
Fix run_man_graph leak after OMTensorList changes

cjvolzka:
* The generated llvm code for run_main_graph was still using malloc to create a temporary list array. This changes that to a stack alloc instead.
* Fixed some warnings
* Fix some leaks in code examples
* Removed unused param from `CREATE_OMTENSOR_LIST`

### DIFF
--- a/docs/mnist_example/mnist.cpp
+++ b/docs/mnist_example/mnist.cpp
@@ -303,7 +303,7 @@ int main() {
   }
 
   // Free the output as it is no longer needed.
-  omTensorListDestroy(tensorListIn);
+  omTensorListDestroy(tensorListOut);
 
   printf("The digit is %d\n", digit);
   return 0;
--- a/docs/mnist_example/mnist.cpp
+++ b/docs/mnist_example/mnist.cpp
@@ -272,7 +272,7 @@ static float img_data[] = {-0.4242129623889923f, -0.4242129623889923f,
 int main() {
   // Create an input tensor list of 1 tensor.
   int inputNum = 1;
-  OMTensor **inputTensors = (OMTensor **)malloc(inputNum * sizeof(OMTensor *));
+  OMTensor *inputTensors[inputNum];
   // The first input is of tensor<1x1x28x28xf32>.
   int64_t rank = 4;
   int64_t shape[] = {1, 1, 28, 28};
--- a/docs/mnist_example/mnist.cpp
+++ b/docs/mnist_example/mnist.cpp
@@ -284,6 +284,9 @@ int main() {
   // Compute outputs.
   OMTensorList *tensorListOut = run_main_graph(tensorListIn);
 
+  // Free the input as it is no longer needed.
+  omTensorListDestroy(tensorListIn);
+
   // Extract the output. The model defines one output of type tensor<1x10xf32>.
   OMTensor *y = omTensorListGetOmtByIndex(tensorListOut, 0);
   float *prediction = (float *)omTensorGetDataPtr(y);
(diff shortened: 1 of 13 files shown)


src/Conversion/KrnlToLLVM/KrnlEntryPoint.cpp
```diff
@@ -236,16 +236,8 @@ class KrnlEntryPointOpLowering : public OpRewritePattern<KrnlEntryPointOp> {
...
-    auto mallocSym = getOrInsertMalloc(rewriter, module);
-    // TODO(tjingrant): get pointer size from data layout.
-    size_t kPtrSize = 8;
-    Value outputOmtPtrsArraySizeInByte = create.llvm.constant(
-        int64Ty, (int64_t)(outMemRefList.size() * kPtrSize));
-    Value outOmtPtrsArr = create.llvm.call(
-        LLVM::LLVMPointerType::get(IntegerType::get(module.getContext(), 8)),
-        mallocSym, ArrayRef<Value>(outputOmtPtrsArraySizeInByte));
-    outOmtPtrsArr = create.llvm.bitcastI8PtrPtr(outOmtPtrsArr);
+    Value outOmtPtrsArr = create.llvm._alloca(
```
AlexandreEichenberger, line 239: I assume that you lifted this pattern from elsewhere. Pointers are typically aligned at a multiple of 8 byes ... I assume that there is something like "aligned at natural data size" in that logic. Meaning that you have seen other pointer arrays being allocated like that.

AlexandreEichenberger (commented):
LGTM, would be good to also get @tungld review as he has done some of this LLVM lowering I believe.

docs/mnist_example/mnist.cpp
```diff
@@ -272,7 +272,7 @@ static float img_data[] = {-0.4242129623889923f, -0.4242129623889923f,
 int main() {
   // Create an input tensor list of 1 tensor.
   int inputNum = 1;
-  OMTensor **inputTensors = (OMTensor **)malloc(inputNum * sizeof(OMTensor *));
+  OMTensor *inputTensors[inputNum];
```
gongsu832, line 275: Since we are changing this file, might as well add the `omTensorListDestroy(tensorListOut)` at the end even though it's not strictly necessary in this case. But it can serve as a reminder to the users that they are responsible for freeing the memory for the output tensor list, deep or shallow.

src/Conversion/KrnlToLLVM/KrnlEntryPoint.cpp
tungld: @cjvolzka As you can see the old code assumes the pointer size is 8, so it would be helpful to add a comment saying "Assume that OMTensor pointer size is 8", because the allocation size is potentially different from the actual size for the tensor list. The comment helps us debug the code in the future.

cjvolzka: @AlexandreEichenberger correct. I lifted the pattern from a few lines above it actually (line 204):
```
      Value ptrToMemRef =
          create.llvm._alloca(memRefPtrTy, one, /*alignment=*/0);
```

tungld (approved):
LGTM. Thanks for the patch!

docs/mnist_example/mnist.cpp
```diff
@@ -299,6 +302,9 @@ int main() {
     }
   }
 
+  // Free the output as it is no longer needed.
+  omTensorListDestroy(tensorListIn);
```
gongsu832, line 306: copy-and-paste error. Should be `tensorListOut`.

cjvolzka: Yes on both counts. Thanks for catching!

PR #2079, opened 2023-03-14
Fix Modelzoo Check

gongsu832:
- add --log-to-file to optionally log output to a file
- allow max 30 minutes for compilation+inference for a model check


tungld (approved):
LGTM. Thank you for the fix!

Issue #2070, opened 2023-03-09
Regression: bertsquad models fail to compile for CPU on s390x


cjvolzka:
The bertsquad models are failing to compile for CPU using the latest onnx-mlir commits. Unfortunately the last commit to post a modelzoo status at https://www.onnxmlir.xyz/jenkins/job/ONNX-MLIR-Pipeline-Docker-Build/Model_20Zoo_20Report/ is a couple of days old. Commit e0a1504c9ae319ac6856ba67a334ab918fa8c193 appeared fine. Somewhere after that is when the issue started. 

For example [bertsquad-10.onnx](https://github.com/onnx/models/blob/main/text/machine_comprehension/bert-squad/model/bertsquad-10.onnx):
```
onnxmlirczar/onnx-mlir --version
onnx-mlir version 0.4.0, onnx version 1.13.1 (https://github.com/onnx/onnx-mlir.git 115161446650842b58ea24c2bd3bf4ef95b0a2c0)

onnxmlirczar/onnx-mlir --O3 --EmitLib --mtriple=s390x-ibm-loz --mcpu=z14 /workdir/bertsquad-10.onnx
onnx-mlir: /workdir/onnx-mlir/src/Dialect/Mlir/IndexExpr.cpp:114: int onnx_mlir::IndexExprScope::addDim(mlir::Value): Assertion `indexInList(symbols, value) == -1 && "Cannot have a dim that is already a symbol"' failed.
```

Other models that failed our regression test:
bertsquad-10, bertsquad-12

tungld:
@cjvolzka @gongsu832 I also saw the same error with `-03` but  bertsquad-12 ran correctly with `-O2`. It is potentially something related to `-03`.

tungld:
@AlexandreEichenberger the lowering of Sqrt caused this model failed. I think it it related to the SIMD code in PR #2053. This is an example to re-produce the issue:
```mlir
func.func private @test_sqrt(%arg0 : tensor<?x1xf32>) -> tensor<*xf32> {
  %0 = "onnx.Sqrt"(%arg0) : (tensor<?x1xf32>) -> tensor<*xf32>
  "func.return"(%0) : (tensor<*xf32>) -> ()
}
```

The issue only happens when the tensor is `<?x1>`. There is no problem with `<?x2>` or the second dim is not 1.

Below is the error in details, when running `onnx-mlir-opt -O3 --shape-inference --convert-onnx-to-krnl`:
```
[...]

PR #2081, opened 2023-03-14
fix for bug with SIMD

AlexandreEichenberger:
Fix bug for SIMD found in #2070


tungld (approved):
LGTM. Thanks for the fix!

PR #2082, opened 2023-03-14
Fix run onnx lib

AlexandreEichenberger:
There was a place where `omTensorListCreate` still used an additional parameter that was removed in the new interface.


AlexandreEichenberger:
This seems to have been fixed in an related PR

PR #2083, opened 2023-03-14
SIMD support: added a disable switch and performance tests

AlexandreEichenberger:
Added a `-disable-simd` switch. SIMD is only activated at `-O3` and now we can have `-O3` without SIMD by using that switch. This should be particularly useful for specific debugging of the new SIMD feature.

In addition, I added initial support for testing the performance of element wise functions, which can highlight the benefits of SIMD. Currently I added support for  binary Add and HardSigmoid, two example of ops that can be simdized. Hard sigmoid performs a FMP plus two clips (for max and min).


src/Dialect/ONNX/ONNXOps/Additional/ShapeTransform.cpp
```diff
@@ -31,16 +31,16 @@ LogicalResult ONNXShapeTransformOpShapeHelper::computeShape() {
   auto inputType = input.getType().cast<ShapedType>();
   Type elementType = inputType.getElementType();
   ArrayRef<int64_t> inputDims = inputType.getShape();
-  uint64_t outputRank = indexMap.getNumResults();
+  int64_t outputRank = indexMap.getNumResults();
```
AlexandreEichenberger, line 34: made some fix to remove warnings

src/Dialect/ONNX/ONNXOps/Math/Reduction.cpp
```diff
@@ -128,7 +127,7 @@ LogicalResult ONNXGenericReductionOpShapeHelper<OP_TYPE>::computeShape() {
       // Interesting idea below, namely to reuse info from output, or if not
       // there, from input putting question mark in there. Not sure if
       // successful, if it is, it should be generalized to all ops.
-
+      OP_TYPE reduceOp = llvm::cast<OP_TYPE>(op);
```
AlexandreEichenberger, line 130: moved statement to remove warnings

test/modellib/ElementwiseModel.cpp
```diff
@@ -0,0 +1,205 @@
...
+  builder.create<func::ReturnOp>(loc, results);
+  module.push_back(funcOp);
+  createEntryPoint(funcOp);
+  return true;
+}
+
+bool Elementwise2DLibBuilder::prepareInputs(
+    float dataRangeLB, float dataRangeUB) {
+  // Create dynamic list (to be freed before end of function).
+  OMTensor **list = (OMTensor **)malloc(inputNum * sizeof(OMTensor *));
```
tungld, line 99: PR #2078  tried to replace this `malloc` by `OMTensor* list[inputNum];`, recommend to use it here.

src/Compiler/CompilerOptions.cpp
```diff
@@ -198,6 +198,11 @@ llvm::cl::opt<bool> enableParallel("parallel",
                    "Set to 'true' if you want to enable parallelization."),
     llvm::cl::init(false), llvm::cl::cat(OnnxMlirOptions));
 
+llvm::cl::opt<bool> disableSimdOption("disable-simd",
+    llvm::cl::desc("Disable SIMD optimizations (default: on at O3, "
+                   "off otherwise). Set to `true` to disable SIMD at O3."),
```
tungld, line 203: `default: on at O3` looks a bit confusing, in the sense that `on` is for this flag `disable-simd` or for `SIMD on`? Usually, `default` refers to default value for the current flag.

src/Compiler/CompilerOptions.cpp
```diff
@@ -198,6 +198,11 @@ llvm::cl::opt<bool> enableParallel("parallel",
...
+    llvm::cl::init(false), llvm::cl::cat(OnnxMlirOptions));
+
 llvm::cl::opt<bool> enableSimdDataLayout("simd-data-layout",
```
tungld, line 206: I happened to see this, so are `simd-data-layout` and `disable-simd` exclusive or not? Looks like `dsiable-simd` does not include `simd-data-layout`.

src/Dialect/ONNX/ONNXOps/Additional/ShapeTransform.cpp
tungld: Thanks!

src/Compiler/CompilerOptions.cpp
AlexandreEichenberger: I can improve the text.

test/modellib/ElementwiseModel.cpp
tungld: I see. So the current code looks fine as you free them properly.

src/Compiler/CompilerOptions.cpp
tungld: > I could combine the two flags for now. Is that what you would like?

Yes, `disable-simd` is used for both unary ops, matmul and gemm I guess, so it's great to include convolution too in it instead of a separate one `simd-data-layout`

test/modellib/ElementwiseModel.cpp
AlexandreEichenberger: Now use stack alloc to size 2, and may only use 1 when inputNum==1. Simpler code.

src/Compiler/CompilerOptions.cpp
AlexandreEichenberger: Made the `disable-simd` also disable `simd-data-layout`

test/modellib/ElementwiseModel.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+bool Elementwise2DLibBuilder::prepareInputs(
+    float dataRangeLB, float dataRangeUB) {
+  constexpr int num = 2;
+  assert(inputNum <= num, "bad constant");
+  OMTensor *list[num];
+
+  // Create elements in the list.
+  if (!list)
+    return false;
```
tungld, line 104: This `if` looks redundant.

tungld (approved):
LGTM.

test/modellib/ElementwiseModel.cpp
AlexandreEichenberger: good catch, will remove.

PR #2085, opened 2023-03-15
Fix memory leak in the lowering of krnl.Call

tungld:
The lowering of `krnl.Call` to LLVM IR prepares OMTensors to wrap its MemRef arguments for calling external functions. However, these OMTensors are not destroyed, which results in memory leak.

This patch emits omTensorDestroy calls to clean up the OMTensors after calling external functions.


cjvolzka (commented):
I'm not qualified to approve the code but I did try the models with these changes. I confirmed it fixed the leaks :-)

```
==2771190== 
==2771190== HEAP SUMMARY:
==2771190==     in use at exit: 0 bytes in 0 blocks
==2771190==   total heap usage: 441,752 allocs, 441,752 frees, 118,477,019 bytes allocated
==2771190== 
==2771190== All heap blocks were freed -- no leaks are possible
==2771190== 
```

chentong319 (approved):
LGTM!
Could you add a comment in Krnl.call definition in krnl.td?

tungld:
> Could you add a comment in Krnl.call definition in krnl.td?

Done. Thanks!

PR #2086, opened 2023-03-15
Fix an error on krnl-to-affine conversion for onnx.CategoryMapper with rank-2+ inputs

negiyas:
An error occurs in compilation of the bidaf-9 model in the model zoo.
The situation can be confirmed by  the following simple test case, 

```
$ cat test_category_mapper_string_to_int64-onnxbasic.mlir
func.func private @test_category_mapper_string_to_int64(%arg0 : tensor<2x2x!onnx.String>) -> tensor<2x2xi64> {
  %0 = "onnx.CategoryMapper"(%arg0) {cats_int64s = [1, 2, 3], cats_strings = ["cat", "dog", "cow"], default_int64 = -1: si64} : (tensor<2x2x!onnx.String>) -> tensor<2x2xi64>
  "func.return"(%0) : (tensor<2x2xi64>) -> ()
}

$ time /home/negishi/src/dlc.git/onnx-mlir.main/build/Debug/bin/onnx-mlir-opt --shape-inference --convert-onnx-to-krnl --canonicalize --convert-krnl-to-affine test_category_mapper_string_to_int64-onnxbasic.mlir
onnx-mlir-opt: /home1/negishi/src/dlc.git/llvm-project/mlir/lib/Dialect/Affine/IR/AffineOps.cpp:2964: static void mlir::AffineLoadOp::build(mlir::OpBuilder&, mlir::OperationState&, mlir::Value, mlir::AffineMap, mlir::ValueRange): Assertion `map.getNumInputs() == mapOperands.size() && "inconsistent index info"' failed.
```
This error comes from the following invalid krnl.load op generated by onnx-to-krnl conversion for onnx.Category with rank-2 inputs.
```
       %7 = krnl.load %6[%5#0, %5#1] : memref<2x!krnl.string>
```

In this case, the krnl.load has two indices, so that the input shape should have rank-2 shape.
```
      %7 = krnl.load %6[%5#0, %5#1] : memref<2x2x!krnl.string>
```

This PR fixes the onnx-to-krnl conversion for onnx.CategoryMapper and its test cases.


AlexandreEichenberger:
Reading the specs:

> Type Constraints
> T1 in ( tensor(int64), tensor(string) ): The input must be a tensor of strings or integers, either [N,C] or [C].
> T2 in ( tensor(int64), tensor(string) ): The output is a tensor of strings or integers. Its shape will be the same as the input shape.

So good catch. You may want to add a check (if there is a verify, in the verify; otherwise in the shape inference) that the rank is 1 or 2.

AlexandreEichenberger (approved):
LGTM, please consider adding a check for rank == 1 or 2.

negiyas:
@AlexandreEichenberger Check if the input rank == 1 or 2 is added at the validation logic. Thanks!

AlexandreEichenberger:
> Check if the input rank == 1 or 2 is added at the validation logic. Thanks!

Thanks @negiyas for the follow up, much appreciated

PR #2087, opened 2023-03-16
Fix memory leak in utils/RunONNXLib.cpp

gongsu832:
dll_omTensorCreateWithOwnership is actaully using omTensorCreate, which doesn't set _owning to true so tensor data was not freed by omTensorListDestroy.


tungld (approved):
LGTM.

PR #2088, opened 2023-03-17
[NNPA] Disable the rewriting of Sqrt into Exp and Log

tungld:
PR #2013 rewrites `sqrt(x)` and `1/sqrt(x)` by using `exp` and `log`. However, `log(0)` is undefined, so it is unsafe to rewrite `sqrt(x)`. But `1/sqrt(x)` can still be rewritten since it implies `x != 0`.

This patch removes the rewrite pattern for `sqrt(x)` only.

This issue is reported by @inouehrs. Thanks, Inoue-san!


tungld:
@cjvolzka FYI.

AlexandreEichenberger (commented):
LGTM, good catch  @inouehrs

tungld:
> I don't recall how much time this actually saves us in a model. Overall it saves some stick/unstick and a CPU sqrt at the cost of calling a series of NNPA ops.

For bertsquad, we rewrote `1/sqrt(x)` not `sqrt(x)`. So disabling `sqrt(x)` rewrite does not effect us much at this moment.

>If the gain is minimal, then yeah we can just drop this. However if it's impactful (ie more that 5% reduction or so), then would it make sense to handle 0 as a special case? (ie if x=0 return 0, else exp(log(x)/2) = exp(0.5 * log(x)).

It might not be efficient to handle 0 here. `x` is a tensor not a scalar, so there will be an overhead checking whether a tensor contains a 0 or not.

cjvolzka (approved):

PR #2091, opened 2023-03-19
StridesIterator

sorenlassen:
StridesIterator provides a faster implementation of traverseStrides() and is also an alternative to traverseStrides(), illustrated by a reimplementation of of Slice constant propagation.

I benchmarked the speedup with
```
onnx-mlir arcfaceresnet100-8.onnx --mlir-elide-elementsattrs-if-larger=1 --mlir-timing
```
where the slowest ConstPropONNXToONNXPass dropped form 0.2583s to 0.1884s on my MacBook M1 Pro.


src/Dialect/ONNX/ElementsAttr/Strides.hpp
```diff
@@ -119,15 +120,122 @@ struct StridedArrayRef : public llvm::ArrayRef<T> {
...
+  }
+
+  // End iterator: ends after the given number of iterations.
+  StridesIterator(size_t iterations) : value{0, iterations} {}
+
+  // End iterator: ends after one traversal of shape.
+  StridesIterator(llvm::ArrayRef<int64_t> shape)
+      : StridesIterator(mlir::ShapedType::getNumElements(shape)) {}
+
+  StridesIterator(const StridesIterator &) = default;
```
philass, line 178: I'd be interested to know if these definitions are needed or if they will be auto generated by the compiler

src/Dialect/ONNX/ElementsAttr/Strides.hpp
```diff
@@ -119,15 +120,122 @@ struct StridedArrayRef : public llvm::ArrayRef<T> {
...
+
+  StridesIterator &operator=(const StridesIterator &) = default;
```
philass, line 180: here too

src/Dialect/ONNX/ElementsAttr/Strides.hpp
```diff
@@ -119,15 +120,122 @@ struct StridedArrayRef : public llvm::ArrayRef<T> {
...
+    return value.flattenedIndex != other.value.flattenedIndex;
+  }
+
+  reference operator*() const { return value; }
+
+  pointer operator->() const { return &value; }
+
+  inline StridesIterator &operator++() {
+    ++(value.flattenedIndex);
+    for (auto axis = shape.size();;) {
```
philass, line 196: Impressive how you managed to avoid all uses of curly braces :p

philass (approved):
LGTM!

src/Dialect/ONNX/ElementsAttr/Strides.hpp
sorenlassen: they are not, I added the comment:
```
  // These declarations are redundant, the compiler generates this constructor
  // and operator automatically, but included to flag that StridesIterator can
  // be copied, which is used in the implementation of operator++(int) below.
```

PR #2092, opened 2023-03-20
Decompose CustomOp that is ONNXRuntime FusedMatMul

tungld:
ONNX models produced by ONNXRuntime, e.g. [bertsquad-int8](https://github.com/onnx/models/blob/main/text/machine_comprehension/bert-squad/README.md#quantization), may have a custom operation, FusedMatMul, that is only supported by ONNXRuntime.

In this patch, we decompose FusedMatMul into standard ONNX operations: ONNXMatMul and ONNXTranspose, so that onnx-mlir can compile such models. This is a short-term solution. For long-term, it would be ideal if ONNX frameworks do not generate such framework-specific operations.


src/Transform/ONNX/Decompose.cpp
```diff
@@ -577,6 +581,177 @@ struct ConcatFusePattern : public ConversionPattern {
...
+// Example of onnx.Custom:
+// ```
+// "onnx.Custom"(%0, %1) {alpha = 1.250000e-01 : f32,
+//                        domain_name = "com.microsoft",
+//                        function_name = "FusedMatMul",
+//                        transA = 0 : si64, transB = 1 : si64} :
+//              (tensor<*xf32>, tensor<*xf32>) -> tensor<*xf32>
+// ```
+
+static bool isCustomOpMatched(ONNXCustomOp customOp, FloatAttr &alphaAttr,
```
chentong319, line 607: Better to use a more specific name for this match because CustomOp may be matched to other pattern in future. I am wondering can whether this match function can be defined in the pattern as the match  function.

tungld: I changed the function to a specific one `isCustomOpFusedMatMulMatched`, and moved this function inside the pattern. However, we use this function in `target.addDynamicallyLegalOp` that happens before any pattern applied, so I don't think we can defined it as the match function of the pattern.

chentong319: That's true.

chentong319 (approved):
LGTM!

PR #2094, opened 2023-03-20
Elementwise simd: disable DIV simd code gen which causes issues for GPT2

AlexandreEichenberger:
Identified one operation that causes GPT2 to diverge from expected results. Will investigate further reason why in a successive PR.

Found the `RunONNXModelZoo.py -m gpt2-10  -l debug -c="-O3"` command very intuitive to debug; though there was no option to preserve the output of the compilation phase (unlike `RunONNXModel.py`). Thus I added the `--log-to-file` option to the `RunONNXModelZoo.py` so that we may now compile a specific model from the zoo while also saving in a log file the output of the compiler. The new command is now `RunONNXModelZoo.py -m gpt2-10  -l debug -c="-O3" --log-to-file` and will save in the default file the log. A specific log file name can also be specified. Use `-h` for full list of options.


cjvolzka:
LGTM. I tested a build with this PR against our client with gpt2-10, gpt2-lm-head-10, and roberta-sequence-classification-9. All three models return within tolerances again.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1016,6 +1021,10 @@ static LogicalResult getVariadicSimdCodeFullyFlattened(
   Type outputElementType = outputMemRefType.getElementType();
   unsigned numArgs = op->getNumOperands();
 
+  if (DEBUG)
+    fprintf(stderr, "hi alex, simdizing variadic op %s\n",
```
tungld, line 1025: Please make sure to remove this "hi, alex" :)

tungld (approved):
LGTM. Thanks for the fix!

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
AlexandreEichenberger: Ohhh, but I want to be famous :-)

PR #1953, opened 2023-01-11
ConvTranspose op support

imaihal:
This PR support ConvTranspose op by rewriting it into ONNXConvOp and other ONNX ops for preparing input and attributes. This rewriting is included in DecomposeONNXToONNXPass.

Input, weight, and some other attributes for ConvTranspose op are rewritten is as follows.
1. Reverse elements of weight tensor using ONNXTransposeOp and ONNXReverseSequenceOp.
2. Insert pads between elements in input data tensor using ONNXSplitOp, ONNXPadOp, and ONNXConcatOp.
3. Use ONNXConv op using converted weight tensor, input data tensor, other  attributes.
Screen Shot 2023-02-04 at 0 36 33

PR #2099, opened 2023-03-22
Support more cases in rewriting of ConvTranspse op

imaihal:
Subsequent PR of https://github.com/onnx/onnx-mlir/pull/1953 to support auto_pad of SAME_UPPER/LOWER for ConvTranspose op.
Also removed redundant code.

- [x] Use ONNXOpBuiler in more ops and remove result type when using it.
- [x] Update shape inference to use latest style.
- [x] Support auto_pad SAME_UPPER/SAME_LOWER.


src/Dialect/ONNX/ONNXOps/NN/Conv.cpp
```diff
@@ -741,138 +670,14 @@ LogicalResult ONNXConvTransposeOp::inferShapes(
...
-  outputShapeFinal.emplace_back(outChannels);
-  // Insert spatial dims.
-  for (int32_t i = 0; i < spatialRank; ++i) {
-    outputShapeFinal.emplace_back(ArrayAttrIntVal(outputShape, i));
-  }
-  getResult().setType(
-      RankedTensorType::get(outputShapeFinal, xTy.getElementType()));
-
-  return success();
+  Type elementType = getX().getType().cast<ShapedType>().getElementType();
```
tungld, line 678: You can use `getElementType(getX().getType())` from `src/Support/TypeUtilities.hpp`.

src/Dialect/ONNX/DialectBuilder.cpp
```diff
@@ -146,18 +146,19 @@ Value OnnxBuilder::mul(Type resultType, Value A, Value B) const {
       resultType, toTensor(A), toTensor(B));
 }
 
-Value OnnxBuilder::pad(Type outputType, Value input, Value pads,
-    Value constantValue, std::string mode) const {
+Value OnnxBuilder::pad(
+    Value input, Value pads, Value constantValue, std::string mode) const {
+  Type elementType = input.getType().cast<ShapedType>().getElementType();
```
tungld, line 151: You can use `getElementType(input.getType())` from `src/Support/TypeUtilities.hpp`.

src/Transform/ONNX/Decompose.cpp
```diff
@@ -270,40 +271,46 @@ Value reverseWeightTensor(
...
-}
-
-ArrayAttr createUnitStrides(PatternRewriter &rewriter, ArrayAttr strides) {
-  SmallVector<int64_t, 2> unitStrides(strides.size(), 1);
-  return rewriter.getI64ArrayAttr(unitStrides);
+bool hasUnitStrides(ONNXConvTransposeOp op) {
+  ONNXConvTransposeOpShapeHelper shapeHelper(op.getOperation(), {});
+  LogicalResult shapecomputed = shapeHelper.computeShape();
+  assert(succeeded(shapecomputed) &&
+         "unexpected inferShapes failuer for ConvTrans op");
```
tungld, line 311: Please use `computeShapeAndAssertOnFailure` instead of `computeShape`, that includes assertion.

tungld: Do we really need to call shapeHelper just for getting the stride attribute?

src/Transform/ONNX/Decompose.cpp
```diff
@@ -397,47 +363,67 @@ Value insertPadAxis(PatternRewriter &rewriter, Location loc, Value input,
...
 Value insertAdditionalPadsConvTranspose(PatternRewriter &rewriter, Location loc,
-    ONNXConvOp op, Value input, ArrayAttr outputShapeAttr) {
-  ONNXConvOpShapeHelper shapeHelper(op.getOperation(), {});
-  shapeHelper.computeShapeAndAssertOnFailure();
-  int inputRank = shapeHelper.getOutputDims().size();
+    ONNXConvOp convOp, Value input, ONNXConvTransposeOp op) {
+  ONNXConvOpShapeHelper convShapeHelper(convOp.getOperation(), {});
+  Type elementType = input.getType().cast<ShapedType>().getElementType();
+  assert(succeeded(convShapeHelper.computeShapeAndUpdateType(elementType)) &&
+         "unexpected inferShapes failure for Conv op");
```
tungld, line 398: `computeShapeAndUpdateType` has assertion. There is no need to assert here.

src/Transform/ONNX/Decompose.cpp
```diff
@@ -397,47 +363,67 @@ Value insertPadAxis(PatternRewriter &rewriter, Location loc, Value input,
...
-                    ? shapeHelper.getOutputDims()[i].getLiteral()
+    int64_t d = convShapeHelper.getOutputDims()[i].isLiteral()
+                    ? convShapeHelper.getOutputDims()[i].getLiteral()
                     : ShapedType::kDynamic;
     inputShape.emplace_back(d);
   }
+  ONNXConvTransposeOpShapeHelper shapeHelper(op.getOperation(), {});
+  LogicalResult shapecomputed = shapeHelper.computeShape();
+  assert(succeeded(shapecomputed) &&
+         "unexpected inferShapes failuer for ConvTrans op");
```
tungld, line 410: Please use `computeShapeAndAssertOnFailure` instead.

tungld (changes requested):
Thanks for refactoring the code for ConvTransposeOp! LGTM in general.

I see that shapeHelper is called again and again in different functions. I guess it's because of the use of tablegen, so you cannot pass it around. So, to avoid calling shapeHelper multiple times, it's perhaps better to write the pattern in C++ instead of tablegen. What do you think?

src/Dialect/ONNX/ONNXOps/NN/Conv.cpp
imaihal: Done. I fixed the same pattern in the file. Thanks!

src/Dialect/ONNX/DialectBuilder.cpp
imaihal: Done. Thanks!

src/Transform/ONNX/Decompose.cpp
```diff
@@ -270,40 +271,46 @@ Value reverseWeightTensor(
...
+ArrayAttr getPadsConvTranspose(
+    PatternRewriter &rewriter, Location loc, ONNXConvTransposeOp op) {
+  // Calculate pads for generated Conv op.
   // new_pads = kernel -  pads - 1
   // Reference: Dumoulin, Vincent, and Francesco Visin. "A guide to convolution
   // arithmetic for deep learning." arXiv preprint arXiv:1603.07285 (2016).
+  ONNXConvTransposeOpShapeHelper shapeHelper(op.getOperation(), {});
+  LogicalResult shapecomputed = shapeHelper.computeShape();
+  assert(succeeded(shapecomputed) &&
+         "unexpected inferShapes failuer for ConvTrans op");
```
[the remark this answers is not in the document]
imaihal: Done. Thanks!

src/Transform/ONNX/Decompose.cpp
imaihal: Removed shapehelper in this function. Thanks.

src/Transform/ONNX/Decompose.cpp
imaihal: Done. Thanks.

src/Transform/ONNX/Decompose.cpp
imaihal: Done. Thanks!

src/Transform/ONNX/Decompose.cpp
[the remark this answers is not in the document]
tungld: @imaihal I still see the old code here. Did you push your change to the PR?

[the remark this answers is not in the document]
imaihal: @tungld Sorry. I may forgot to include the change.

imaihal:
> So, to avoid calling shapeHelper multiple times, it's perhaps better to write the pattern in C++ instead of tablegen. What do you think?

I think so... I used tablegen to use the same approach with other rewriting, but most of code for this rewriting are written in C++ functions.

tungld:
>I think so... I used tablegen to use the same approach with other rewriting, but most of code for this rewriting are written in C++ functions.

Please let us know if you are moving to C++ code or still want to keep tablegen code. Thanks!

imaihal:
@tungld Thanks for letting me know. I want to set it as TODO, but is it acceptable?

tungld (approved):
LGTM. Thank you!

PR #2100, opened 2023-03-22
Lower ONNXDequantizeLinear to krnl

tungld:
This patch supports ONNXDequantizeLinear, where:
- ONNXDequantizeLinear is lowered to Krnl dialect
- Backend and LIT tests for ONNXDequantizeLinear are enables.
- Only support for per-tensor or layer dequantization. Not support for per-axis dequantization at this moment.

Besides, this patch edits `RunONNXModel.py` script to support `ui8` input.

The lowering of ONNXDequantizeLinear borrows `ONNXElementwiseUnaryOpLowering`, so this patch contains some code from #2056. It's better to land #2056 before this patch.

### DIFF
--- a/src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
+++ b/src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
@@ -971,10 +971,10 @@ Value emitScalarOpFor<ONNXDequantizeLinearOp>(
   Type xzeroPointTy = XZeroPoint.getType();
 
   // Only support scalar scale and zero_point.
-  if (!(isRankedShapedType(xscaleTy) && getRank(xscaleTy) == 0))
-    llvm_unreachable("[ONNXDequantizeLinearOp] Only per-tensor dequantization");
-  if (!(isRankedShapedType(xzeroPointTy) && getRank(xzeroPointTy) == 0))
-    llvm_unreachable("[ONNXDequantizeLinearOp] Only per-tensor dequantization");
+  assert((isRankedShapedType(xscaleTy) && getRank(xscaleTy) == 0) &&
+         "[ONNXDequantizeLinearOp] Only support per-tensor dequantization");
+  assert((isRankedShapedType(xzeroPointTy) && getRank(xzeroPointTy) == 0) &&
+         "[ONNXDequantizeLinearOp] Only support per-tensor dequantization");
 
   Value scaleFloat = create.krnl.load(XScale);
   Value zeroPointInt = create.krnl.load(XZeroPoint);
--- a/src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
+++ b/src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
@@ -946,6 +946,46 @@ Value emitScalarOpFor<ONNXRoundOp>(ConversionPatternRewriter &rewriter,
 }
 
 //===----------------------------------------------------------------------===//
+// Scalar unary ops for lowering ONNXDequantizeLinearOp
+//===----------------------------------------------------------------------===//
+template <>
+struct ScalarOp<ONNXDequantizeLinearOp> {
+  using FOp = NotSuportedScalarOp;
+  using IOp = CustomScalarOp;
+  using SimdEnabled = NoSimdScalarOp;
+};
+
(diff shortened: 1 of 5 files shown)


test/backend/inference_backend.py
```diff
@@ -715,7 +718,7 @@ def get_test_models():
         "test_reduce_l2_negative_axes_keep_dims_example_cpu": {STATIC_SHAPE:{}, DYNAMIC_SHAPE:{-1:{-1}}, CONSTANT_INPUT:{-1}},
         "test_reduce_l2_negative_axes_keep_dims_random_cpu": {STATIC_SHAPE:{}, DYNAMIC_SHAPE:{-1:{-1}}, CONSTANT_INPUT:{-1}},
 
-        # ==OP== #ReduceMax
+        # ==OP== ReduceMax
```
tungld, line 721: This caused `make onnx_mlir_supported_ops` failed to generate doccument files for supported ops. Fix it here.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1160,15 +1200,23 @@ struct ONNXElementwiseUnaryOpLowering
       create.krnl.iterateIE(loopDef, loopDef, lbs, ubs,
           [&](KrnlBuilder &createKrnl, ValueRange loopInd) {
             Value loadedVal = createKrnl.load(X, loopInd);
+            SmallVector<Value> args;
+            args.emplace_back(loadedVal);
+            for (uint64_t i = 1; i < operands.size(); i++)
+              args.emplace_back(operands[i]);
```
tungld, line 1206: This code is borrowed from PR #2056.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1160,15 +1200,23 @@ struct ONNXElementwiseUnaryOpLowering
...
+                rewriter, loc, op, elementType, args);
             // Store result in the resulting array.
             createKrnl.store(loweredOpResult, alloc, loopInd);
           });
     } else {
       Value loadedVal = create.krnl.load(X);
+      SmallVector<Value> args;
+      args.emplace_back(loadedVal);
+      for (uint64_t i = 1; i < operands.size(); i++)
+        args.emplace_back(operands[i]);
```
tungld, line 1217: This code is borrowed from PR #2056.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -945,6 +945,46 @@ Value emitScalarOpFor<ONNXRoundOp>(ConversionPatternRewriter &rewriter,
...
+  // y is of type f32.
+  Value XInt = scalarOperands[0];
+  Value XScale = scalarOperands[1];
+  Value XZeroPoint = scalarOperands[2];
+
+  Type xscaleTy = XScale.getType();
+  Type xzeroPointTy = XZeroPoint.getType();
+
+  // Only support scalar scale and zero_point.
+  if (!(isRankedShapedType(xscaleTy) && getRank(xscaleTy) == 0))
```
AlexandreEichenberger, line 974: NIT: this feels like an assert to me

utils/RunONNXModel.py
```diff
@@ -296,15 +297,17 @@ def generate_random_input(input_signature, input_shapes):
         if (np.issubdtype(np_elem_type, np.floating)):
             lb = -1.0
             ub = 1.0
-
-        elif (np.issubdtype(np_elem_type, np.integer)):
-            lb = -10
+        elif (np.issubdtype(np_elem_type, np.uint8)):
```
AlexandreEichenberger, line 300: NIT: is it possible to use a non-sized type (like for integers, its `np.integer`. Maybe there is a `np.unsigned`.. or if not, we could use the sizes 8, 16, 32, 64.

tungld: I seems there is no `np.unsigned`, so I checked them case by case. Also, I added two new options `--lower-bound` and `--upper-bound` to allow users set the range for random inputs.

AlexandreEichenberger (approved):
LGTM

PR #2101, opened 2023-03-22
introduce cmake option ONNX_MLIR_DECOMP_ONNX_CONVTRANSPOSE

sorenlassen:
This is needed for Groq to disable this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,6 +8,7 @@ project(onnx-mlir)
 option(ONNX_MLIR_BUILD_TESTS "Build ONNX-MLIR test executables. If OFF, just generate build targets." ON)
 option(ONNX_MLIR_CCACHE_BUILD "Set to ON for a ccache enabled build." OFF)
 option(ONNX_MLIR_ENABLE_MHLO "Enable MHLO support." ON)
+option(ONNX_MLIR_DECOMP_ONNX_CONVTRANSPOSE "Enable ONNXConvTransposeOp decomposition." ON)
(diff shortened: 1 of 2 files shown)


imaihal:
@sorenlassen Does the Groq have own implementation or rewriting for ConvTranspose op?

sorenlassen:
> @sorenlassen Does the Groq have own implementation or rewriting for ConvTranspose op?

yes, we lower it to Groq's internal dialect

we don't support ONNXReverseSequenceOp at the moment so we can't compare the performance of your decomposition with Groq's lowering

philass (approved):
Thanks for unblocking.

I think this is a relatively concise hack for opting in and out of this decomposition for different downstream users.

We can leave this in while we investigate torch-mlir style opt in and out decompositions that the AMD folks mentioned in the weekly meeting.

tungld:
It's unfortunate to have this PR merged in hurry. It's quite weird to deal with a specific operation using this way, saying using CMake. We should discuss a better solution that uses MLIR infrastructure.

philass:
Hey @tungld,

Sorry for rushing this into the codebase. The AMD folk have kindly opened an issue where we can discuss a robust way of solving the decomposition issue https://github.com/onnx/onnx-mlir/issues/2105.

We were hoping to be able to use a short term fix while we figured out the best way to solve the problem in the long run. Sorry for not giving you enough time to review the PR. 

I sent you a slack message as well if you want to discuss this with me further.

tungld:
@philass thanks for the information! Got it.

PR #2103, opened 2023-03-22
LLVM update: Highlight issues with opaque pointers

gargaroff:
Hey there,

As mentioned briefly in the meeting yesterday, the LLVM dialect in MLIR has switched to opaque pointers per default. This is causing a lot of issues in the Kernel dialect. Unfortunately I cannot refactor the code to use opaque pointers as part of the bump.

With this PR, I wanted to highlight some places that I found where code will need to change once opaque pointers become not only the default, but the only type of pointer there is. The idea is that the people, that understand the needs of the Kernel dialect and are familiar with the affected parts of the code, can start migrating the code use opaque pointers, such that at some point in the future, we can enable opaque pointers in the LLVM dialect again.

For the actual bump, I will disable opaque pointers and open a new PR without these changes here.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -185,7 +185,7 @@ llvm_update_compile_flags(rapidcheck)
 
 if (ONNX_MLIR_ENABLE_MHLO)
   llvm_update_compile_flags(stablehlo-opt)
-  llvm_update_compile_flags(stablehlo-interpreter)
+  llvm_update_compile_flags(stablehlo-translate)
   llvm_update_compile_flags(mlir-hlo-opt)
 endif()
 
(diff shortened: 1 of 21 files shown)


gargaroff:
@AlexandreEichenberger I'm not actually requesting a review from you here. I'm just adding you to let you forward this to people that should be aware of these required changes with regards to opaque pointers.

cjvolzka:
@tungld can you take a look at this PR and offer any guidance on the TODO questions in things like ZLowToLLVM.cpp

tungld:
@gargaroff can we use opaque pointers with the current LLVM that is being used in onnx-mlir? I'll give it a try to find a correct modification.

gargaroff:
> @gargaroff can we use opaque pointers with the current LLVM that is being used in onnx-mlir? I'll give it a try to find a correct modification.

@tungld That should be doable. Opaque pointers need to be turned on the LLVM Dialect. There is a flag that got enabled since the last bump. I don't have the code in front of me know, but I can look for it later when I'm at work

tungld:
@gargaroff thanks, I found it, that is `options.useOpaquePointers = true;` I am looking at how to change our code to opaque pointers.

> For the actual bump, I will disable opaque pointers and open a new PR without these changes here.

Yes, please go ahead with LLVM update, we will enable opaque pointers later.

gargaroff:
I'll close this one now so you can focus on the migration. The actual bump can be found in PR #2107

PR #2106, opened 2023-03-23
Pad opset18 update

philass:
Updated Pad to opset 18


src/Dialect/ONNX/ONNXUnsupportedOps.hpp
```diff
@@ -55,6 +55,7 @@ UNSUPPORTED_OPS(ONNXMomentumOp)
 UNSUPPORTED_OPS(ONNXMultinomialOp)
 UNSUPPORTED_OPS(ONNXNegativeLogLikelihoodLossOp)
 UNSUPPORTED_OPS(ONNXNormalizerOp)
+UNSUPPORTED_OPS(ONNXPadV13Op)
```
AlexandreEichenberger, line 58: Could I ask to add a comment here, to indicate that an op is "unsupported" because we forced it to become a more recent op, as opposed to op we currently don't support because its unimplemented?

This could be another PR, but maybe we could define the following
```C++
#define UNSUPPORTED_OPS_REPLACED_BY(_unsupportedOp, _replacedOp) \
               UNSUPPORTED_OPS(_unsupportedOp)

...

UNSUPPORTED_OPS_REPLACED_BY(ONNXPadV13Op, ONNXPadOp)
```

AlexandreEichenberger (approved):
LGTM,

could you document the restriction on the PAD implementation in `inference_backend.py`

chentong319 (approved):
LGTM!

src/Dialect/ONNX/ONNXUnsupportedOps.hpp
philass: I think I will do this in a follow up PR. Because I think a few other ops match this pattern.

I'll include those ops, when I make the follow up PR.

PR #2108, opened 2023-03-23
Fuse unary element-wise Ops when onnx is converted to krnl  

chentong319:
This PR fuses any number of unary element-wise Ops that follows an element-wise Op. No special fused ONNX Op needs to be introduced. The function is implemented with three utility functions:
1. Check the number of user and the type of the user Op to form a fusible Op list.
2. Use the EmitUnaryElementWiseOpFor<T> to perform the loop body transformation
3.  Replace the last Op in the  fusible Op list with allocated memory (which is the only krnl.store) and erase all the other fused Ops.

Current draft limitation:
1. Not all unary element-wise Ops are added. Need some tedious work.
2. The fusion can start only from variadic element-wise Op now. It is easy to add code for other element-wise Ops.


chentong319:
@tungld and @AlexandreEichenberger Could you provide feedback before I finish the missing part of this PR?

tungld:
@chentong319 the approach looks reasonable. Could we generalize this for other operation such as Conv? so that Relu will be fused into Conv.

chentong319:
> @chentong319 the approach looks reasonable. Could we generalize this for other operation such as Conv? so that Relu will be fused into Conv.

yes. The fusion can start from any Op as long as the Op has only one output. I should move the code of fusion utility into ONNXToKrnlCommon.*.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1088,6 +1088,79 @@ static LogicalResult getVariadicSimdCodeFullyFlattened(
...
+    // Check the users is an unary elementwise op
+    // The right solution, I think, is to define EmitScalarOpFor as
+    // an interface of unary elementwise Ops.
+    // In the draft PR, it is assumed that the candidates are Sqrt and Relu
+    Operation *user;
+    for (Operation *temp : currentProducer->getUsers()) {
+      user = temp;
+      break;
+    }
+    if (isa<ONNXSqrtOp>(user)) {
```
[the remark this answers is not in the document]
chentong319: Thanks for the suggestion. Not familiar with the variadic template.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1376,7 +1461,8 @@ struct ONNXElementwiseVariadicOpLowering
 
             Value finalResult = emitPostProcessingFor<ElementwiseVariadicOp>(
```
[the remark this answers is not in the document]
chentong319: Binary ops are important. Will be considered in future.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1362,6 +1362,114 @@ static LogicalResult getVariadicSimdCodeFullyFlattened(
...
+  }
+}
+
+template <>
+bool enqueueFusedOp(Operation *op, SmallVector<Operation *, 2> &fusibleOps,
+    SmallVector<EmitScalarFunc, 2> &fuseEmitFunctions) {
+  return false;
+}
+
+bool enqueueFusedOp(Operation *op, SmallVector<Operation *, 2> &fusibleOps,
```
AlexandreEichenberger, line 1407: While this recursive template is very crafty, I am not very fond of this approach because if forces us to look into a list that is separate of the function definition.

I would prefer something that uses the declarative type that we already use. Here is what I used to do for SIMD before I changed it to something else.

```
using SimdScalarOp = std::true_type;    // scalar op can be simdized.
using NoSimdScalarOp = std::false_type; // scalar op cannot be simdized.

template <typename Op>
struct ScalarOp {
  using FOp = NotSuportedScalarOp;
  using IOp = NotSuportedScalarOp;
  using SimdEnabled = NoSimdScalarOp;
};
```

So you would use that pattern for `FusionEnabled` with `FusibleOp` and `NonFusibleOp`.

Then you can use a simple if..

```C++
    if constexpr (SimdizableOp<ElementwiseUnaryOp>::value) {
      // SIMD is enabled for this operation, test if desired and feasible

```

This code let you have an "if" that will statically be enabled when the op can be fused.

I prefer this approach because when we look at the function, we can clearly see declaratively if it can be fused. If we add a new operation, we will be forced to define it one way or another, so it is also better for maintainability.

chentong319: I feel the difference is whether to instantiate/specialize the template manually or not. The SIMD/scalar code emission needs to define the llvm op for each ONNX Op, which has to be done manually. Here, it's just true or false based on the list of ONNX Op types specified. 
The best way, to me, is to use traits/interface for dialect.

AlexandreEichenberger: SIMD flag was only defined for the broadcast ops.

chentong319: Anyway, can we have slight different way of programming?

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1621,6 +1729,20 @@ struct ONNXElementwiseVariadicOpLowering
...
+    bool fusibleOpsFound =
+        findFusibleUnaryOps(op, fusibleOps, fuseEmitFunctions);
+    if (fusibleOpsFound) {
+      // After fusion, the only store is for the last Op.
+      // Therefore, the allocation should be the output of the last Op
+      Operation *lastOp = fusibleOps[fusibleOps.size() - 1];
+      outputMemRefType =
+          this->typeConverter->convertType(lastOp->getResults()[0].getType())
+              .template cast<MemRefType>();
+    }
```
tungld, line 1744: This looks redundant. The type of the last op would be similar to the type of the current op, since all fused ops are unary ops. Otherwise, we cannot use shapeHelper of the current op in the `alignedAlloc`. In other words, it would be fine to alloc the output buffer by using the current op's output type.

chentong319: I add a test case for dynamic size with broadcast. The alignedAlloc takes the shapeHelper of the current Op for the dynamic dimensions and the memory ref for the type. Since the element-wise ops do not change the shape of. tensor. alignedAlloc works. We cannot use the shapeHelper of the last Op because the inputs are not available for shape propagation.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
chentong319: simd or all the element-wise emit function is called from the Op with its ONNX Op type known.   The extra work for fusion detection is to check the type of a generic op, which is the user of the current Op.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
tungld: Got it. Thanks!

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
tungld: It looks using traits/interface is better especially when we extend the fusion for other ops like Conv rather than element-wise ops. Currently `struct ScalarOp` is internal to element-wise ops only.
Perhaps, we can take the same approach as `shapeHelper`. Then, check if an op has a trait, like `UnaryOp` then get its `emitScalarFunc` to fuse.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
chentong319: I really ran into an example of cast in the backend test. For test_dynamicquantizelinear_expanded.onnx, part of the list of ops are listed below:
```
   %9 = "onnx.Sub"(%0, %8) : (tensor<f32>, tensor<f32>) -> tensor<f32> 
   %10 = "onnx.Clip"(%9, %0, %1) : (tensor<f32>, tensor<f32>, tensor<f32>) -> tensor<f32>
    %11 = "onnx.Round"(%10) : (tensor<f32>) -> tensor<f32>
    %12 = "onnx.Cast"(%11) {to = ui8} : (tensor<f32>) -> tensor<ui8>
    %13 = "onnx.QuantizeLinear"(%arg0, %7, %12) {axis = 1 : si64} : (tensor<6xf32>, tensor<f32>, tensor<ui8>) -> tensor<6xui8>
    return %13, %7, %12 : tensor<6xui8>, tensor<f32>, tensor<ui8>
```
Clip, Round, and Cast  are fused. I fixed an error for this test case: the Clip should not be fused with Sub. Though Clip is lowered as ElementwiseUnaryOp in our implementation, it has three inputs. I guess the two optional inputs are added in the newer version of onnx. The root of the problem came from the interface: we pass a ValueRange for input but did not specify the expected size for unary and binary Op. The ValueRange is for variadic Ops.
I guess @tungld and @AlexandreEichenberger may be interested in this issue.

tungld: Yes, we treat Clip as a unary op since the last two tensors are scalar. We had been passing the two tensors as they are to emitScalarOpFor<ONNXClipOp>, but recently their loaded values are passed to instead that would be convenient for SIMD I guess.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
chentong319: Yes. The final solution is to use interface. We can define an interface for krnl lowering.
You are right on the struct ScalarOp is to element-wise ops only (with op type known) and we need the interface to check whether an ONNX Op has the interface or not.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1385,6 +1388,122 @@ static LogicalResult getVariadicSimdCodeFullyFlattened(
...
+      ONNXMeanOp, ONNXRoundOp, ONNXDequantizeLinearOp, ONNXSqrtOp>(
+      op, fusibleOps, fuseEmitFunctions);
+}
+
+// This function starts for elementwise operation, op.
+// A successor op (user) is fusible if it's the only user and an unary
+// elementwise Op. The Op and its EmitScalarOpFor<T> are recorded into
+// the vector.
+static bool findFusibleUnaryOps(Operation *op,
+    SmallVector<Operation *, 2> &fusibleOps,
```
tungld, line 1453: I often see the use of `SmallVectorImpl<Operation*>` function argument, instead of `SmallVector<Operation *, 2>`. Same for the other functions in this patch.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1385,6 +1388,122 @@ static LogicalResult getVariadicSimdCodeFullyFlattened(
...
+    Operation *user = *currentProducer->getUsers().begin();
+    if (!enqueueFusedOp(user, fusibleOps, fuseEmitFunctions))
+      break;
+
+    currentProducer = user;
+  }
+
+  if (fusibleOps.size() > 0)
+    LLVM_DEBUG(
+        { llvm::dbgs() << "unary op fused: " << fusibleOps.size() << "\n"; });
```
tungld, line 1471: Uhm, is it seriously checking the condition for LLVM_DEBUG here?

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -12,11 +12,14 @@
...
 //===----------------------------------------------------------------------===//
 
+#include "llvm/Support/Debug.h"
+
 #include "src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.hpp"
 #include "src/Dialect/Krnl/DialectBuilder.hpp"
 #include "src/Dialect/ONNX/ONNXOps/ShapeHelper.hpp"
 
 #define DEBUG 0 /* Log which functions are simdized. */
+#define DEBUG_TYPE "unaryfuse"
```
tungld, line 22: Since this type is for the whole file, would be better to have a more general name, e.g. `element-wise-lowering`.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1385,6 +1388,122 @@ static LogicalResult getVariadicSimdCodeFullyFlattened(
...
+  // Notice: Though ClipOp is classified as unary element op in this file,
+  // ClipOp requires one required input and two optional input
+
+  return enqueueFusedOp<ONNXCastOp, ONNXSinhOp, ONNXCoshOp, ONNXSigmoidOp,
+      ONNXHardSigmoidOp, ONNXEluOp, ONNXReluOp, ONNXLeakyReluOp, ONNXPReluOp,
+      ONNXSeluOp, ONNXReciprocalOp, ONNXSoftplusOp, ONNXSoftsignOp, ONNXSignOp,
+      ONNXMaxOp, ONNXMinOp, ONNXNegOp, ONNXLessOp, ONNXLessOrEqualOp,
+      ONNXGreaterOp, ONNXGreaterOrEqualOp, ONNXEqualOp, ONNXNotOp, ONNXModOp,
+      ONNXMeanOp, ONNXRoundOp, ONNXDequantizeLinearOp, ONNXSqrtOp>(
+      op, fusibleOps, fuseEmitFunctions);
```
tungld, line 1445: How about sorting the list in the alphabetical order?

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1385,6 +1388,122 @@ static LogicalResult getVariadicSimdCodeFullyFlattened(
...
+
+// Variadic template to iterate all the fusible Ops
+template <typename T>
+bool enqueueFusedOpImpl(Operation *op, SmallVector<Operation *, 2> &fusibleOps,
+    SmallVector<EmitScalarFunc, 2> &fuseEmitFunctions) {
+  if (isa<T>(op)) {
+    fusibleOps.emplace_back(op);
+    fuseEmitFunctions.emplace_back(emitScalarOpFor<T>);
+    return true;
+  } else {
```
tungld, line 1408: nit: remove `else` and just return `false`.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1385,6 +1388,122 @@ static LogicalResult getVariadicSimdCodeFullyFlattened(
...
+template <typename T = void, class... Ts>
+bool enqueueFusedOp(Operation *op, SmallVector<Operation *, 2> &fusibleOps,
+    SmallVector<EmitScalarFunc, 2> &fuseEmitFunctions);
+
+template <typename T, class... Ts>
+bool enqueueFusedOp(Operation *op, SmallVector<Operation *, 2> &fusibleOps,
+    SmallVector<EmitScalarFunc, 2> &fuseEmitFunctions) {
+  if (enqueueFusedOpImpl<T>(op, fusibleOps, fuseEmitFunctions)) {
+    return true;
+  } else {
```
tungld, line 1422: nit: remove `else`.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1452,21 +1571,36 @@ struct ONNXElementwiseUnaryOpLowering
...
+    SmallVector<EmitScalarFunc, 2> fuseEmitFunctions;
+    bool fusibleOpsFound =
+        findFusibleUnaryOps(op, fusibleOps, fuseEmitFunctions);
+    if (fusibleOpsFound) {
+      // After fusion, the only store is for the last Op.
+      // Therefore, the allocation should be the output of the last Op
+      Operation *lastOp = fusibleOps[fusibleOps.size() - 1];
+      outputMemRefType =
+          this->typeConverter->convertType(lastOp->getResults()[0].getType())
+              .template cast<MemRefType>();
```
tungld, line 1592: We use these code repeatedly. I think it can be simpler if we have a class, say, `UnaryOpFusionHelper` and hide all `fusibleOps`, `fuseEmitFunctions`, and `outputMemRefType` inside, e.g.
```C
UnaryOpFusionHelper fusionHelper(rewriter, loc);
fusionHelper.fuse(op);
bool fusibleOpsFound = fusionHelper.isFeasible();
outputMemRefType = fusionHelper.getOutputType();
```

and inside `krnl.iterate`:
```C
loweredOpResult = fusionHelper.emitFuseOps(loweredOpResult);
```

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
chentong319: What's the difference?

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
chentong319: Condition is to reduce the debug info printout.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
chentong319: agree

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
tungld: `SmallVector<Operation *, 2>` requires that users of the function must declare a vector with exact type of `SmallVector<Operation *, 2>` to pass in, while with `SmallVectorImpl<Operation*>`, users can declare a vector of `SmallVector<Operation *, 4>` for example. It's more flexible.

Description of `SmallVectorImpl` in LLVM: https://llvm.org/doxygen/classllvm_1_1SmallVectorImpl.html#details
```
This class consists of common code factored out of the SmallVector class to reduce code duplication based on the SmallVector 'N' template parameter.
```

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
tungld: It would be great to do so, but I am afraid that it requires a lot of changes at this time, since it depends on `emitScalarForOp` that should be reorganized too.

chentong319: You are right. The code of template for emitScalarForOp has to be reorganized. I pushed the change that define the OpFusionHelper inside the Elementwise.cpp.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1385,6 +1388,156 @@ static LogicalResult getVariadicSimdCodeFullyFlattened(
...
+  // Therefore, the allocation should be the output of the last Op
+  MemRefType updateOutputType(MemRefType outputType) {
+    if (!fusibleOpsIsEmpty()) {
+      Operation *lastOp = fusibleOps_[fusibleOps_.size() - 1];
+      return MemRefType::get(
+          outputType.getShape(), lastOp->getResults()[0]
+                                     .getType()
+                                     .template cast<ShapedType>()
+                                     .getElementType());
+    } else {
```
tungld, line 1506: Prefer to follow LLVM coding style in which `else` after `return` here can be removed to reduce indentation and to have easier reading: https://llvm.org/docs/CodingStandards.html#don-t-use-else-after-a-return. See the same pattern in several places in this patch.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1385,6 +1388,156 @@ static LogicalResult getVariadicSimdCodeFullyFlattened(
...
+  // After fusion, the only store is for the last Op.
+  // Therefore, the allocation should be the output of the last Op
+  MemRefType updateOutputType(MemRefType outputType) {
+    if (!fusibleOpsIsEmpty()) {
+      Operation *lastOp = fusibleOps_[fusibleOps_.size() - 1];
+      return MemRefType::get(
+          outputType.getShape(), lastOp->getResults()[0]
+                                     .getType()
+                                     .template cast<ShapedType>()
+                                     .getElementType());
```
tungld, line 1505: We have `getElementType(type)` in `src/Support/TypeUtilities.hpp`. Prefer to use it to avoid boilerplate.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1385,6 +1388,156 @@ static LogicalResult getVariadicSimdCodeFullyFlattened(
...
+      auto currentOp = fusibleOps_[i];
+      auto emitScalar = fuseEmitFunctions_[i];
+      // ToFix: use the ONNX location(ONNXLoc) of each Op.
+      // The current obstacle is that no easy way to know the type of Op,
+      // which is needed by ONNXLoc<T>(op).
+      Location loc = currentOp->getLoc();
+      Type currentElementType = currentOp->getResults()[0]
+                                    .getType()
+                                    .cast<ShapedType>()
+                                    .getElementType();
```
tungld, line 1524: We have `getElementType(type)` in `src/Support/TypeUtilities.hpp`. Prefer to use it to avoid boilerplate.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1385,6 +1388,156 @@ static LogicalResult getVariadicSimdCodeFullyFlattened(
...
+
+    if (!fusibleOpsIsEmpty())
+      LLVM_DEBUG({
+        llvm::dbgs() << "unary op fused: " << fusibleOps_.size() << "\n";
+      });
+  }
+
+  // After fusion, the only store is for the last Op.
+  // Therefore, the allocation should be the output of the last Op
+  MemRefType updateOutputType(MemRefType outputType) {
```
tungld, line 1498: We don't always update it. Suggest to use `updateOrGetOutputType`.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1385,6 +1388,156 @@ static LogicalResult getVariadicSimdCodeFullyFlattened(
...
+                                     .getType()
+                                     .template cast<ShapedType>()
+                                     .getElementType());
+    } else {
+      return outputType;
+    }
+  }
+
+  // Emit fusion Ops
+  Value emitFuseOps(Value finalResult) {
```
tungld, line 1512: `emitFuseOps` may not fuse any ops, depending whether `findFusibleOps`. Suggest to use `tryEmitFuseOp` for the intention that it may succeed or fail.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1385,6 +1388,156 @@ static LogicalResult getVariadicSimdCodeFullyFlattened(
...
+                                    .getType()
+                                    .cast<ShapedType>()
+                                    .getElementType();
+      finalResult = emitScalar(
+          rewriter_, loc, currentOp, currentElementType, finalResult);
+    }
+    return finalResult;
+  }
+
+  void replaceFuseOps(Value alloc) {
```
tungld, line 1531: Same here, `tryReplaceFuseOps` would be better.

tungld (approved):
LGTM with some suggestions on function names and coding style.

It looks good as a starting point for op fusion. Many thanks!

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
chentong319: All the operation defined on OpFusionHelper works when the fusible Op list is empty. I would like to regard that empty list is one situation for OpFusionHelper. Alex also pointed out to add explicit if statement for empty fusible Op list. However, I prefer the uniform way to handle fusion list empty and non-empty. Using "try" may lead to the wonder of failure handling. I kept the name of this method and changed the name of the other two methods.

chentong319:
Thanks for the suggestions.
I used the getElementType util.
All the operation defined on OpFusionHelper works when the fusible Op list is empty. Alex also pointed out to add explicit if statement for empty fusible Op list. However, I prefer the uniform way to handle fusion list empty and non-empty. Using "try" may lead to the wonder of failure handling. I changed updateOutputType to getOutputType, kept the name of emitFuseOp, and changed replaceFuseOps to just replaceOrEraseONNXOps because the starting ONNX op and the fused ONNX Ops are all handled.

PR #2109, opened 2023-03-24
[WIP] Reduce time and memory consumption to compile krnl.global op with many string constants

negiyas:
[WIP] Reduce time and memory consumption to compile krnl.global op with many string constants

1. Consolidate all string symbols in one krnl.global op into one constants including the all string constants
2. Reduce the number of operations in llvm.mlir.global op by removing llvm.getelementptr for each string constant.


negiyas:
Closing this PR, since PR #2727 covers this PR.

PR #2110, opened 2023-03-24
Do not replace unstick/stick by LayoutTransform when layout is 1D or 2DS

tungld:
Because of [an issue](https://github.com/onnx/onnx-mlir/issues/1940) about wrong index access for 1D and 2DS layout, this patch disables the replacement of unstic/stick by LayoutTransform in case of 1D and 2DS layout.

### DIFF
--- a/src/Accelerators/NNPA/Dialect/ZHigh/ZHighOps/OpHelper.cpp
+++ b/src/Accelerators/NNPA/Dialect/ZHigh/ZHighOps/OpHelper.cpp
@@ -195,20 +195,19 @@ Value getConstantOfType(
   return create.onnx.constant(denseAttr);
 }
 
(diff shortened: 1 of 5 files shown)


AlexandreEichenberger (approved):
LGTM, can you create an issue that keep track of the original bug and this PR, so that we may undo the changes once the bug is fixed? Or if you think it cannot be fixed, then just state so in this PR.

Tx

tungld:
>can you create an issue that keep track of the original bug and this PR, so that we may undo the changes once the bug is fixed?

Yes, the original bug is being tracked here: https://github.com/onnx/onnx-mlir/issues/1940 that has linked to related PRs.

tungld:

PR #2111, opened 2023-03-24
Fix IndexExprScope in TransposesOp lowering

tungld:
This patch fixes an issue about IndexExprScope in TransposesOp lowering and a function signature in RunONNXLib.cpp

### DIFF
--- a/src/Conversion/ONNXToKrnl/Tensor/Transpose.cpp
+++ b/src/Conversion/ONNXToKrnl/Tensor/Transpose.cpp
@@ -197,6 +197,7 @@ private:
     create->krnl.iterateIE(loopDef, loopDef, lbs, inUBs,
         [&](KrnlBuilder &createKrnl, ValueRange indices) {
           MultiDialectBuilder<MathBuilder, KrnlBuilder> create(createKrnl);
+          IndexExprScope loopScope(createKrnl);
           // Compute destination and source offsets for memcpy.
           IndexExpr destOffsetIE = LiteralIndexExpr(0);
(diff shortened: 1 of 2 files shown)


src/Conversion/ONNXToKrnl/Tensor/Transpose.cpp
```diff
@@ -197,6 +197,7 @@ struct ONNXTransposeOpLowering : public OpConversionPattern<ONNXTransposeOp> {
     create->krnl.iterateIE(loopDef, loopDef, lbs, inUBs,
         [&](KrnlBuilder &createKrnl, ValueRange indices) {
           MultiDialectBuilder<MathBuilder, KrnlBuilder> create(createKrnl);
+          IndexExprScope loopScope(createKrnl);
```
tungld, line 200: Without this, it failed to create `SymbolIndexExpr` because of not the same scope.

AlexandreEichenberger: Sometimes I think the distinction between symbols and dims are really a nuisance, and not sure how key it is anyways... Thought on that?

AlexandreEichenberger (approved):
LGTM

src/Conversion/ONNXToKrnl/Tensor/Transpose.cpp
tungld: I seems there is no difference if we just do calculation over IndexExpr, but it matters if we get an affine map from IndexExpr. If we have `a*b` in the affine map and `a` is a dim, then `b` must be a literal or symbol. If `b` is symbol, we have a semi affine map.

PR #2114, opened 2023-03-27
constant propagate add/sub zeros, mul/div ones

sorenlassen:
Also added ElementsAttr equality comparison with a unit test. It is not used anywhere yet.

### DIFF
--- a/src/Dialect/ONNX/ElementsAttr/ElementsAttrBuilder.cpp
+++ b/src/Dialect/ONNX/ElementsAttr/ElementsAttrBuilder.cpp
@@ -140,15 +140,17 @@ bool ElementsAttrBuilder::equal(ElementsAttr lhs, ElementsAttr rhs) {
 }
 
 /*static*/
-bool ElementsAttrBuilder::allEqual(ElementsAttr elms, WideNum n) {
+bool ElementsAttrBuilder::allEqual(
+    ElementsAttr lhs, WideNum broadcastedRhsValue) {
+  WideNum n = broadcastedRhsValue;
   return dispatchByBType(
-      btypeOfMlirType(elms.getElementType()), [elms, n](auto btype) {
+      btypeOfMlirType(lhs.getElementType()), [lhs, n](auto btype) {
         using cpptype = CppType<btype>;
         auto nEquals = [n](cpptype x) {
           constexpr BType TAG = toBType<cpptype>;
           return n.narrow<TAG>() == x;
         };
-        if (auto disposable = elms.dyn_cast<DisposableElementsAttr>()) {
+        if (auto disposable = lhs.dyn_cast<DisposableElementsAttr>()) {
           if (disposable.isTransformedOrCast()) {
             ArrayBuffer<WideNum> nums = disposable.getBufferAsWideNums();
             return llvm::all_of(nums.get(), [n](WideNum m) {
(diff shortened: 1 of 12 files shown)


src/Dialect/ONNX/ElementsAttr/ElementsAttrBuilder.cpp
```diff
@@ -100,6 +102,74 @@ DenseElementsAttr ElementsAttrBuilder::toDenseElementsAttr(
...
 }
 
+/*static*/
+bool ElementsAttrBuilder::equal(ElementsAttr lhs, ElementsAttr rhs) {
+  auto lhsType = lhs.getType();
+  auto rhsType = rhs.getType();
+  auto elementType = lhsType.getElementType();
+
+  // Equality between signed/unsigned or int/fp not supported.
+  assert(wideBTypeOfBType(btypeOfMlirType(elementType)) ==
```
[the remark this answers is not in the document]
sorenlassen: I believe it would be confusing to work around callers passing incompatible arguments. I added the comment:
```
  // Asserts if these preconditions are violated (doesn't return false as that
  // would hide whether the lhs and rhs are different or incompatible).
```

src/Dialect/ONNX/ElementsAttr/ElementsAttrBuilder.hpp
```diff
@@ -50,6 +50,20 @@ class ElementsAttrBuilder {
   static mlir::DenseElementsAttr toDenseElementsAttr(
       mlir::ElementsAttr elements);
 
+  // Compares contents for equality. Argument shapes must be broadcast
+  // compatible. Element types must the same.
+  // Asserts if these preconditions are violated (doesn't return false as that
+  // would hide whether the lhs and rhs are different or incompatible).
```
sorenlassen, line 56: I discoverd that it didn't work to allow element types with different bit widths, because then you get different results depending on which element type you use for the comparison if one is bool and another is a wider unsigned integer type

I also concluded that it's bad to compare at higher precision, so comparison now happens at the type specified by the element type and I added a unit test to demonstrate that unequal f32s can become equal when cast to f16

@AlexandreEichenberger PTAL

src/Dialect/ONNX/ElementsAttr/ElementsAttrBuilder.hpp
```diff
@@ -50,6 +50,22 @@ class ElementsAttrBuilder {
...
+  // Note that elements may be compared for equality at a higher precision than
+  // the precision of the element types and therefore may return false even if
+  // the elements would be identical when narrowed to the element types.
+  //
+  // TODO: Move this function to a better place, since it doesn't build
+  //       anything, but it's here for now for efficient elements access.
+  static bool equal(mlir::ElementsAttr lhs, mlir::ElementsAttr rhs);
+
+  // More efficient alternative to
+  // equal(elms, DenseElementsAttr::get(elms.getType(), n.toAPFloat/Int())).
```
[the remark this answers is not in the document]
AlexandreEichenberger: fair enough

sorenlassen:

src/Transform/ONNX/ConstProp.td
```diff
@@ -208,6 +220,19 @@ def AddConstProp : Pat<
     // Additional constraints (dense)
     [(IsFromDenseONNXConstantOp:$lhs), (IsFromDenseONNXConstantOp:$rhs)]>;
 
+// TODO: Expand $a to $result's shape instead of requiring ValuesHaveSameType.
```
philass, line 223: `$a` isn't named in the tablegen rule. You probably mean `$x `

philass (approved):
LGTM.

This will allow us to more efficiently do these checks in Groq's codebase.

I wonder if the solution in the long run would be to have fold methods that take care of this.

src/Transform/ONNX/ConstProp.td
sorenlassen: good catch, fixed

sorenlassen:
> I wonder if the solution in the long run would be to have fold methods that take care of this.

yes, I think so

I hope that will be easy with most of the logic in reusable ElementsAttr helpers outside ConstProp.cpp

PR #2115, opened 2023-03-27
Put node name into location when importing onnx model

chentong319:
Node in onnx model may have a name. Currently, the name is added as an attribute to the name. However, this information may be lost at code rewriting. 
The PR uses the node name, if exists, to create a NameLoc for the onnx Op. Therefore, the node name can be propagated automatically with Location mechanism in MLIR.
For example, result from gpt2-10.onnx with flag `--preserveLocations --EmitONNXIR` contains some lines without onnx_node_name, but location info.
```
%919 = "onnx.Reshape"(%910, %918) {allowzero = 0 : si64} : (tensor<?x?x12x64xf32>, tensor<2xi64>) -> tensor<?x?xf32> loc(#loc1433)

#loc616 = loc("Reshape_798")
#loc621 = loc("Reshape_810")
#loc1433 = loc(fused[#loc621, #loc616])
```


AlexandreEichenberger:
For the constants, have you thought of picking up the name of its users?

AlexandreEichenberger (approved):
LGTM

chentong319:
> For the constants, have you thought of picking up the name of its users?

The constants come from Initializer tensor, not node. I added the code to use tensor name as location for ConstantOp.

PR #2120, opened 2023-03-29
Fix an error in constant propagation for ConstantOfShape

tungld:
This patch adds an additional condition for ConstantOfShape's constant propagation that is to check whether the output type is static or not. Constant propagation should happen when the output type is static (for example by shape inference).

### DIFF
--- a/src/Support/TypeUtilities.cpp
+++ b/src/Support/TypeUtilities.cpp
@@ -51,6 +51,7 @@ int64_t getRank(Type ty) {
 
 /// Get the number of elements.
 int64_t getNumberOfElements(Type ty) {
+  assert(ty.cast<ShapedType>().hasStaticShape() && "Has unknown dimensions");
   ArrayRef<int64_t> shape = getShape(ty);
   return ShapedType::getNumElements(shape);
 }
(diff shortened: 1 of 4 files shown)


src/Support/TypeUtilities.cpp
```diff
@@ -51,6 +51,7 @@ int64_t getRank(Type ty) {
 
 /// Get the number of elements.
 int64_t getNumberOfElements(Type ty) {
+  assert(ty.cast<ShapedType>().hasStaticShape() && "Has unknown dimensions");
```
sorenlassen, line 54: fwiw, I just learned that `x.cast<T>()` is officially deprecated in favor of `cast<T>(x)` https://mlir.llvm.org/deprecation/

sorenlassen (approved):

src/Support/TypeUtilities.cpp
tungld: Thanks! It's good to know. We should have a PR to update our whole code.

sorenlassen: actually, if we're going to cast ty to ShapedType anyway, how about we change the function to take ShapedType input argument:
```
int64_t getNumberOfElements(ShapedType ty) {
  assert(ty.hasStaticShape() && ..);
  return ty.getNumberOfElements();
}
```

since ShapedType is a type interface you can pass any Type and it will automatically do the cast to ShapedType

tungld: It's a good idea. Let me change it to `ShapedType`.

tungld:

PR #2121, opened 2023-03-29
Analyze unknown dimensions for ExpandOp when shape is from concatenation of dims

tungld:
This adds a case in DimAnalysis for ExpandOp when its shape is concatenation of unknown dimensions.

It seems this situation will happen with other operations, so we need a systematic way to handle it. However, this patch is for ExpandOp to make DimAnalysis work with the roBERTa model.


tungld:
@AlexandreEichenberger this PR is ready for review.

AlexandreEichenberger (approved):
LGTM

Issue #2123, opened 2023-03-29
ConstPropSliceImpl fails if output shape is 0


maxbartel:
Hi!

I found a bug regarding the flatten operator coming from torch.
If a flatten layer is used like
```python3
x = torch.flatten(x)
```

the raw ONNX IR that gets generated is 
```mlir
module attributes {llvm.data_layout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128", llvm.target_triple = "x86_64-unknown-linux-gnu"} {
  func.func @main_graph(%arg0: tensor<1x1x100x50xf32>) -> tensor<5000xf32> attributes {input_names = ["onnx::Shape_0"], output_names = ["8"]} {
    %0 = "onnx.Shape"(%arg0) {onnx_node_name = "/Shape", start = 0 : si64} : (tensor<1x1x100x50xf32>) -> tensor<*xi64>
    %1 = onnx.Constant dense<0> : tensor<1xi64>
    %2 = onnx.Constant dense<0> : tensor<1xi64>
    %3 = onnx.Constant dense<0> : tensor<1xi64>
    %4 = "onnx.NoValue"() {value} : () -> none
    %5 = "onnx.Slice"(%0, %2, %3, %1, %4) {onnx_node_name = "/Slice"} : (tensor<*xi64>, tensor<1xi64>, tensor<1xi64>, tensor<1xi64>, none) -> tensor<*xi64>
    %6 = onnx.Constant dense<-1> : tensor<1xi64>
    %7 = "onnx.Concat"(%5, %6) {axis = 0 : si64, onnx_node_name = "/Concat"} : (tensor<*xi64>, tensor<1xi64>) -> tensor<*xi64>
    %8 = "onnx.Reshape"(%arg0, %7) {allowzero = 0 : si64, onnx_node_name = "/Reshape"} : (tensor<1x1x100x50xf32>, tensor<*xi64>) -> tensor<5000xf32>
    return %8 : tensor<5000xf32>
  }
  "onnx.EntryPoint"() {func = @main_graph} : () -> ()
}
```
after running `onnx-mlir-opt -shape-inference` we get the following:

```mlir
module attributes {llvm.data_layout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128", llvm.target_triple = "x86_64-unknown-linux-gnu"} {
[...]

PR #2125, opened 2023-03-29
support empty shape in StridesIterator

sorenlassen:
fixes issue #2123


maxbartel (approved):
Thank you for the fix!

PR #2126, opened 2023-03-30
simplify ElementsAttrBuilder::reduce() with StridesIterator

sorenlassen:


philass (approved):
The new implementation looks cleaner.

Thanks!

PR #2127, opened 2023-03-30
Elementwise simd: added high level machine description of simd support in x86 (SSE 4.2, AVX2) and Z16

AlexandreEichenberger:
Up to now, one could only specify if an elementary op could be simdized or not. Now we have a more precise model of what architecture support what operations. 

Since we now need to know for which machine we generate SIMD, the default version of onnx-mlir will not generate SIMD (LLVM backend might). To activate SIMD on z, specify the `-mcpu` to be `z14` to `z16`. On x86, specify the `-march=x86-64` which currently default to a machine supporting SSE4.

On Z16, I measured the performance of a few element wise op (1k x 1k x fp32). `Add` performance was unchanged (backend ace this one); `SigmoidHard` improves from 5 us to 2us (2.5x); `Sigmoid` improves from 10.0 to 7.7 (30%).

Best way to compare is to specify the machine flags to enable simd at `-O3` and then add the `-disable-simd` flag to turn it off.

Below is help on how this model is supported. Main support is in `src/Dialect/MLIR/VectorMachineSupport.hpp/.cpp`

There are broad categories of ops considered for SIMD.
```C++
enum class GenericOps {
  AbsGop,
  ArithmeticGop, /* Simple compute ops: add/sub */
  CeilDivGop,
  CeilGop,
  CompareGop, /* All compare operations, signed/unsigned fixed/float. */
  ConversionGop,
  CopySignGop,
  DivGop,
  ExpGop,
  FloorDivGop,
  FloorGop,
  FmaGop,
  LogGop,
  LogicalGop, /* All logical ops: and, or, xor, not, nor, nand,... */
  MinMaxGop,
  MulGop,
  PowGop,
  RemGop,
  RoundGop,
  SelectGop,
  ShiftGop,   /* Shift operations: logical/arithmetic. */
  ShuffleGop, /* All bit/byte moving operations: shuffle, rotate, shift. */
  SqrtGop,
  SumAcrossGop,      /* Sum across vector. */
  TrigArcGop,        /* Arc trigonometry ops: asin, acos, atan. */
  TrigGop,           /* Trigonometry ops: sin, cos, tan. */
  TrigHyperbolicGop, /* Hyperbolic trig. */
};
```

and the processing an elementary op that can be simdized, we have a templeted function that let us express which operations are used: for example min

```C++
template <>
Value emitScalarOpFor<ONNXMinOp>(ConversionPatternRewriter &rewriter,
  // ONNXMinOp(%X, %Y) = SelectOp(CmpFOp(OLT, %X, %Y),
  //                              %X,
  //                              %Y)
  CheckIfCustomScalarOpIsSupported<ONNXMinOp>(elementType);
  Value lhs = scalarOperands[0];
  Value rhs = scalarOperands[1];
  MultiDialectBuilder<MathBuilder> create(rewriter, loc);
  // could return create.math.min(lhs, rhs);
  Value cond = create.math.lt(lhs, rhs);
  return create.math.select(cond, lhs, rhs);
}
```

uses a compare and a select. This is expressed as this:

```C++
template <>
double analyzeSimdFor<ONNXMinOp>(Type t, int64_t &von, int64_t &son) {
  return simdAnalysis(
      {GenericOps::CompareGop, GenericOps::SelectGop}, {1, 1}, t, von, son);
}
```

The code in the vector modeling hardware will go through each of the ops listed above, and compute the weighted average SIMD width for the operations. If it is above a threshold, it will direct the code to emit SIMD code. The length of the vector will depend on the number of SIMD ops (the more, the shorter) and the number of SIMD registers (the more registers, the longer).

For ops that do not support SIMD, simply omit the  `double analyzeSimdFor<OnnxOp>(...)` template and it will default to no SIMD.


src/Conversion/KrnlToAffine/ConvertKrnlToAffine.cpp
```diff
@@ -617,6 +618,7 @@ void ConvertKrnlToAffinePass::runOnOperation() {
 
   MLIRContext *ctx = &getContext();
   OpBuilder builder(ctx);
+  VectorMachineSupport::setGlobalVectorMachineSupport(march, mcpu, "");
```
AlexandreEichenberger, line 621: Need to define a global with the current vector machine support; free it at end of pass so that we have no men leaks.

src/Dialect/ONNX/ONNXOps/Additional/LayoutTransform.cpp
```diff
@@ -38,8 +38,10 @@ LogicalResult ONNXLayoutTransformOp::inferShapes(
 // Verifier
 //===----------------------------------------------------------------------===//
 LogicalResult ONNXLayoutTransformOp::verify() {
-  if (auto dataType = getData().getType().dyn_cast<RankedTensorType>())
-    if (auto outputType = getOutput().getType().dyn_cast<RankedTensorType>())
+  if (auto dataType = getData().getType().dyn_cast<RankedTensorType>()) {
```
AlexandreEichenberger, line 41: code was giving warnings, this version does not

src/Dialect/Mlir/VectorMachineSupport.cpp
```diff
@@ -0,0 +1,254 @@
...
+      vectorizedOpNum += num;
+    else
+      scalarOpNum += num;
+    totVL += vl * num;
+  }
+  // Compute final values
+  int64_t totNum = vectorizedOpNum + scalarOpNum;
+  scalarOpNum = gopsSize - vectorizedOpNum;
+  if (totNum == 0)
+    return 0.0;
```
tungld, line 93: So `if` here is redundant since the `return` covers it.

src/Dialect/Mlir/VectorMachineSupport.cpp
```diff
@@ -0,0 +1,254 @@
...
+  // lengths.
+  for (int64_t i = 0; i < gopsSize; ++i) {
+    // Even if unsupported, we can still proceed with one value per op.
+    int64_t vl = std::max((int64_t)1, getVectorLength(gops[i], elementType));
+    // If past last value, assume 1; otherwise use actual value.
+    int64_t num = 1;
+    if (i < gopsNumSize)
+      num = gopsNum[i];
+    // Accumulate weighted scalar/vectorized num and vl length.
+    if (vl > 0)
```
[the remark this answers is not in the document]
AlexandreEichenberger: yes, vl == 0 means no simd, so increase the count for scalar op. I now renamed `totVl` to `totProcessedValues`. Processed values are vl when vl>0, and 1 otherwise (for scalar). That should be clearer.

Thanks for catching that.

src/Dialect/Mlir/VectorMachineSupport.cpp
```diff
@@ -0,0 +1,254 @@
...
+  scalarOpNum = 0;
+  // Determine which operations support SIMD and accumulate their vector
+  // lengths.
+  for (int64_t i = 0; i < gopsSize; ++i) {
+    // Even if unsupported, we can still proceed with one value per op.
+    int64_t vl = std::max((int64_t)1, getVectorLength(gops[i], elementType));
+    // If past last value, assume 1; otherwise use actual value.
+    int64_t num = 1;
+    if (i < gopsNumSize)
+      num = gopsNum[i];
```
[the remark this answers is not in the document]
AlexandreEichenberger: yes, and in fact, I don't use that feature. Let me assert the list have the same length.

src/Dialect/Mlir/VectorMachineSupport.cpp
AlexandreEichenberger: yes. thanks

src/Dialect/Mlir/VectorMachineSupport.hpp
```diff
@@ -0,0 +1,156 @@
...
+// Generic ops to determine if operations are supported for lowering to vector
+// SIMD operations. An additional operation's element type will be provided to
+// further refine whether an operation can be effectively vectorized on the
+// given target hardware. This list roughly corresponds to the operations
+// supported by MathDialectBuilder, with some combining of similar operations
+// (e.g. all the compares).
+
+enum class GenericOps {
+  AbsGop,
+  ArithmeticGop, /* Simple compute ops: add/sub */
```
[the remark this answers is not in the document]
AlexandreEichenberger: It includes related ops, like a compare (which is typically `a-b` with detecting > = < 0.
Added a comment.

src/Dialect/Mlir/VectorMachineSupport.cpp
```diff
@@ -0,0 +1,254 @@
+//===-- VectorMachineSupport.cpp - Helper for what SIMD ops are supported -===//
```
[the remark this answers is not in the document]
AlexandreEichenberger: good catch

src/Dialect/Mlir/VectorMachineSupport.hpp
```diff
@@ -0,0 +1,156 @@
+//===-- VectorMachineSupport.hpp - Helper for what SIMD ops are supported -===//
```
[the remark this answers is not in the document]
AlexandreEichenberger: added, tx

AlexandreEichenberger:
@tungld thanks for the comments, believe I responded to all of them.

src/Dialect/Mlir/VectorMachineSupport.cpp
```diff
@@ -65,33 +69,31 @@ int64_t VectorMachineSupport::getVectorLength(Type elementType) {
...
+  int64_t totProcessedValues = 0.0;
   vectorizedOpNum = 0;
   scalarOpNum = 0;
   // Determine which operations support SIMD and accumulate their vector
   // lengths.
   for (int64_t i = 0; i < gopsSize; ++i) {
     // Even if unsupported, we can still proceed with one value per op.
-    int64_t vl = std::max((int64_t)1, getVectorLength(gops[i], elementType));
+    int64_t vl = getVectorLength(gops[i], elementType);
     // If past last value, assume 1; otherwise use actual value.
```
tungld, line 82: This comment is not valid any more.

tungld (approved):
LGTM.

src/Dialect/Mlir/VectorMachineSupport.cpp
AlexandreEichenberger: fixed, tx

PR #2129, opened 2023-03-31
Lowering QuantizeLinear and DynamicQuantizeLinear to Krnl

tungld:
This patch lowers ONNXQuantizeLinear and ONNXDynamicQuantizeLinear to Krnl. It also edits the lowering of EntryPoint to support output unsigned integers, which is necessary for backend tests.


src/Conversion/ONNXToKrnl/Math/Reduction.cpp
```diff
@@ -455,7 +455,13 @@ struct ONNXReductionOpLowering : public OpConversionPattern<ONNXReductionOp> {
...
     IndexExprScope mainScope(&rewriter, loc);
     DimsExpr axesDims;
-    create.krnlIE.getShapeAsDims(axesVal, axesDims);
+    if (isFromNone(axesVal) && !isNoop) {
+      // The default is to reduce over all the dimensions of the input tensor if
+      // 'noop_with_empty_axes' is false
+      for (int64_t i = 0; i < inRank; ++i)
+        axesDims.emplace_back(LiteralIndexExpr(i));
+    } else
+      create.krnlIE.getShapeAsDims(axesVal, axesDims);
```
tungld, line 464: Fix an issue in Reduce when `axes` is not given.

src/Dialect/Mlir/DialectBuilder.cpp
```diff
@@ -635,8 +635,9 @@ Value MathBuilder::cast(Type destType, Value src) const {
     // TosaToLinalg in MLIR uses a fancier algorithm that clamps values to
     // min/max signed/unsigned integer values.
     if (destType.isUnsignedInteger()) {
-      Value cast = castToSignless(src, srcWidth);
-      return b().create<arith::FPToUIOp>(loc(), destType, cast);
+      Type castType = b().getIntegerType(destWidth);
+      Value cast = b().create<arith::FPToUIOp>(loc(), castType, src);
+      return castToUnsigned(cast, destWidth);
```
tungld, line 640: Fix an issue when casting from float to unsigned int. In this case, `src` is float, so we cannot call `castToSignless` on `src`

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1480,8 +1493,17 @@ struct ONNXElementwiseUnaryOpLowering
...
+      for (uint64_t i = 1; i < operands.size(); i++) {
+        if (isNoneValue(operands[i])) {
+          args.emplace_back(operands[i]);
+          continue;
+        }
+        assert(isScalarValue(operands[i]) &&
+               "unary expected scalar additional values");
+        Value loadedVal = create.krnl.load(operands[i]);
+        args.emplace_back(loadedVal);
+      }
```
tungld, line 1506: This is to fix Clip for the scalar tensor case.

src/Conversion/KrnlToLLVM/KrnlEntryPoint.cpp
```diff
@@ -47,17 +47,24 @@ class KrnlEntryPointOpLowering : public OpRewritePattern<KrnlEntryPointOp> {
   SmallVectorImpl<LLVM::GlobalOp> &entryGlobalOps;
   SmallVectorImpl<LLVM::GlobalOp> &inSigGlobalOps;
   SmallVectorImpl<LLVM::GlobalOp> &outSigGlobalOps;
+  std::map<std::string, SmallVector<MemRefType, 4>> &inputMemRefTypes;
+  std::map<std::string, SmallVector<MemRefType, 4>> &outputMemRefTypes;
```
tungld, line 51: Changes in `KrnlEntryPoint.cpp` is to handle unsigned integer outputs. LLVM IR only uses signless integers, so we have to store the original MLIR type (e.g. ui8) so that we can construct output OMTensors with correct types.

tungld:
@AlexandreEichenberger this PR is ready for review. Thanks!

src/Conversion/ONNXToKrnl/Quantization/QuantizeLinear.cpp
```diff
@@ -0,0 +1,127 @@
...
+        typeConverter->convertType(qlOp.getResult().getType()));
+    MemRefType yScaleMemRefType = YScale.getType().cast<MemRefType>();
+
+    // Types
+    Type elementType = xMemRefType.getElementType();
+    Type quantizedElementType = yMemRefType.getElementType();
+    int64_t rank = xMemRefType.getRank();
+
+    // Does not support per-axis and i8.
+    assert(yScaleMemRefType.getRank() == 0 &&
```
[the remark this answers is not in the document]
tungld: No, currently it is for the lowering only. It would be passed the verifier. I will support these soon when there is some spare time.

src/Conversion/ONNXToKrnl/Quantization/QuantizeLinear.cpp
```diff
@@ -0,0 +1,127 @@
...
+      zeroPoint = create.math.constant(elementType, 0.0);
+
+    // Compute y.
+    ValueRange loopDef = create.krnl.defineLoops(rank);
+    SmallVector<IndexExpr> lbs(rank, LiteralIndexExpr(0));
+    create.krnl.iterateIE(loopDef, loopDef, lbs, shapeHelper.getOutputDims(0),
+        [&](KrnlBuilder &createKrnl, ValueRange loopInd) {
+          MultiDialectBuilder<KrnlBuilder, MathBuilder, OnnxBuilder> create(
+              createKrnl);
+          Value x = create.krnl.load(X, loopInd);
```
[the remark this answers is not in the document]
tungld: Yes, I think so, something like `iterateIE` with SIMD flag or two input functions: one for scalar, one for simd, so we can switch between them.

src/Conversion/ONNXToKrnl/Quantization/DynamicQuantizeLinear.cpp
```diff
@@ -0,0 +1,155 @@
...
+    // Compute max(x) and min (x).
+    Value none = create.onnx.none();
+    Value XMax = create.onnx.toMemref(
+        create.onnx.reduceMax(yScaleMemRefType, X, none, false));
+    Value XMin = create.onnx.toMemref(
+        create.onnx.reduceMin(yScaleMemRefType, X, none, false));
+    Value xMax = create.krnl.load(XMax);
+    Value xMin = create.krnl.load(XMin);
+    // Include 0 to max(x) and min(x).
+    // x_min = min(min(x), 0)
```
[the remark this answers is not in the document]
tungld: This is to include 0 as one element in the X tensor when computing `min(X)`. Same to `max`.
I found that this is necessary to get correct result. DynamicQuantizeLinear spec does not mention this, but its python example does: https://github.com/onnx/onnx/blob/main/docs/Operators.md#DynamicQuantizeLinear
```python
X = np.array([0, 2, -3, -2.5, 1.34, 0.5]).astype(np.float32)
x_min = np.minimum(0, np.min(X))
x_max = np.maximum(0, np.max(X))
```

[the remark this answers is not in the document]
cjvolzka: I bumped into this today. As @tungld mentioned, the [DynamicQuantizeLinear Doc](https://onnx.ai/onnx/operators/onnx__DynamicQuantizeLinear.html) page doesn't show this (but I think it's alluded to by `data range is adjusted to include 0.`). The reference example does have it at https://github.com/onnx/onnx/blob/main/onnx/reference/ops/op_dynamic_quantize_linear.py#L15-L16.

I did a little digging, and it looks like PyTorch does the same thing: https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/quantized/QTensor.cpp quantize_per_tensor_dynamic() -> quant_utils::ChooseQuantizationParams -> https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/quantized/cpu/QuantUtils.h#L95-L99
```
  // We extend the [min, max] interval to ensure that it contains 0.
  // Otherwise, we would not meet the requirement that 0 be an exactly
  // representable value.
  min = std::min(min, 0.f);
  max = std::max(max, 0.f);
```

I created https://github.com/onnx/onnx/issues/5597 to hopefully get the doc updated to match the reference example.

src/Conversion/ONNXToKrnl/Quantization/DynamicQuantizeLinear.cpp
```diff
@@ -0,0 +1,149 @@
...
+    Value lessThanZero = create.math.slt(xMin, zero);
+    xMin = create.math.select(lessThanZero, xMin, zero);
+
+    // Compute y_scale.
+    Value scale = create.math.div(
+        create.math.sub(xMax, xMin), create.math.sub(qMax, qMin));
+    create.krnl.store(scale, YScale);
+
+    // Compute y_zero_point.
+    Value interZeroPoint = create.math.div(create.math.sub(qMin, xMin), scale);
```
cjvolzka, line 108: @tungld Is this correct? If I'm reading this right, I think it calculates `interZeroPoint =  (qMin - xMin) / (scale)`. However it shouldn't it be `interZeroPoint = qMin - (xMin / scale)`?

Also does this correctly handle the case where the range was expanded to up/to_the_right to include zero? 

For example, say the inputs were `x = np.array([-1.0, -2.1, -1.3, -2.5, -3.34, -4.0])` For the reasons mentioned earlier in the thread, the `y_scale` is calculated so that it includes zero even though it's not actually in the range (ie `np.maximum(0,np.max(x))`). In this scenario (where `x.max() < 0`, then the zeroPoint should be qMax, so 255 for unsigned int8. 

Sadly this is another case where the [ONNX documentation](https://onnx.ai/onnx/operators/onnx__DynamicQuantizeLinear.html) shows one thing `intermediate_zero_point = qmin - min(x)/y_scale` and the [reference example](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/quantized/cpu/QuantUtils.h#L149-L158) does another
```
  double zero_point_from_min = qmin - min / static_cast<double>(scale);
  double zero_point_from_max = qmax - max / static_cast<double>(scale);
  double zero_point_from_min_error =
      std::abs(qmin) - std::abs(min / static_cast<double>(scale));
  double zero_point_from_max_error =
      std::abs(qmax) - std::abs(max / static_cast<double>(scale));
  double initial_zero_point =
      zero_point_from_min_error < zero_point_from_max_error
      ? zero_point_from_min
      : zero_point_from_max;
 ````

Comment (tungld) [reply src/Conversion/ONNXToKrnl/Quantization/DynamicQuantizeLinear.cpp:108]: Good catch! as I remember I used `interZeroPoint =  (qMin - xMin) / (scale)` because I saw it in their examples: https://github.com/onnx/onnx/blob/main/docs/Operators.md#examples-45
```
Y_ZeroPoint = np.clip(round((0 - x_min) / Y_Scale), 0, 255).astype(np.uint8)
```

It happened to be true when qMin = 0.

PR #2130, opened 2023-03-31
fix issue with if statement nesting in ONNXLayoutTransformOp::verify

MikeHolman:
The else if statement is now binding to the if outside the for loop, which is not the intended behavior. I've added some additional curly braces to fix this.


sorenlassen (approved):
LGTM

AlexandreEichenberger:
thanks for fixing this dangling else issue.

PR #2132, opened 2023-04-01
MatMulInteger const prop

sorenlassen:
Note that it is implemented with a reusable helper method ElementsAttrBuilder::matMul() which is designed to also work for the MatMul op, so it will be easy to add const prop for MatMul afterwards.

### DIFF
--- a/src/Dialect/ONNX/ElementsAttr/ElementsAttrBuilder.cpp
+++ b/src/Dialect/ONNX/ElementsAttr/ElementsAttrBuilder.cpp
@@ -512,14 +512,16 @@ ElementsAttr ElementsAttrBuilder::matMul(ElementsAttr lhs, ElementsAttr rhs) {
          "matMul() requires identical element types");
 
   ArrayRef<int64_t> lhsShape = lhsType.getShape();
-  size_t lhsRank = lhsShape.size();
+  unsigned lhsRank = lhsShape.size();
 
   ArrayRef<int64_t> rhsShape = rhsType.getShape();
-  size_t rhsRank = rhsShape.size();
+  unsigned rhsRank = rhsShape.size();
 
   // TODO: Handle care where lhs and/or rhs have rank 1.
   assert(lhsRank >= 2);
   assert(rhsRank >= 2);
+  int64_t lhsDimSize = lhsShape[lhsRank - 2];
+  int64_t rhsDimSize = rhsShape[rhsRank - 1];
   int64_t reductionDimSize = lhsShape[lhsRank - 1];
   assert(reductionDimSize == rhsShape[rhsRank - 2]);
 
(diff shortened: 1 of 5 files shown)


src/Dialect/ONNX/ElementsAttr/ElementsAttrBuilder.cpp
```diff
@@ -517,6 +517,120 @@ ElementsAttr ElementsAttrBuilder::reduce(ElementsAttr elms,
...
 
+ElementsAttr ElementsAttrBuilder::matMul(ElementsAttr lhs, ElementsAttr rhs) {
+  ShapedType lhsType = lhs.getType();
+  ShapedType rhsType = rhs.getType();
+  Type elementType = lhsType.getElementType();
+  assert(elementType == rhsType.getElementType() &&
+         "matMul() requires identical element types");
+
+  ArrayRef<int64_t> lhsShape = lhsType.getShape();
+  unsigned lhsRank = lhsShape.size();
```
philass, line 528: Do we use `unsigned` in other places in the codebase?

philass (approved):
Left two small comments.

Otherwise looks good.

src/Transform/ONNX/ConstProp.cpp
```diff
@@ -404,6 +412,131 @@ Value ConstPropReduce(PatternRewriter &rewriter, Value replacingValue,
...
+    ArrayRef<int64_t> matrixShape, ElementsAttr zeroPoint) {
+  ShapedType zeroPointType = zeroPoint.getType();
+  auto zeroPointShape = zeroPointType.getShape();
+  auto zeroPointRank = zeroPointShape.size();
+  if (zeroPointRank == 0 || (zeroPointRank == 1 && zeroPointShape[0] == 1)) {
+    // Scalar case is easy: zeroPoint trivially broadcasts to matrix's shape.
+    // Scalars can be represented as singleton tensors with rank 0 or 1.
+  } else if (zeroPointRank == 1) {
+    // Vector with zero point scalar per row. Same shape as a matrix column.
+    assert(zeroPointRank == 1);
```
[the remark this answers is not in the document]
sorenlassen: removed

src/Dialect/ONNX/ElementsAttr/ElementsAttrBuilder.cpp
sorenlassen: MLIR uses unsigned in many places, e.g. for indexes into shapes and operands and results lists

maybe it's a little bit unusual to use it for rank? I changed it to size_t

src/Transform/ONNX/ConstProp.cpp
```diff
@@ -109,6 +110,13 @@ bool isVariadicOperandFromDenseONNXConstantOp(ValueRange operands) {
   return llvm::all_of(operands, [](Value v) { return isDenseONNXConstant(v); });
 }
 
+Value ConstZeroTensor(
+    PatternRewriter &rewriter, Location loc, ShapedType type) {
+  return createONNXConstantOpWithDenseAttr(rewriter, loc,
+      DenseElementsAttr::get(
+          type, rewriter.getZeroAttr(type.getElementType())));
+}
```
tungld, line 118: Since we are moving to using DialectBuilder, It would be better to use OnnxBuilder for ONNXConstantOp here.

src/Transform/ONNX/ConstProp.cpp
```diff
@@ -404,6 +412,129 @@ Value ConstPropReduce(PatternRewriter &rewriter, Value replacingValue,
...
+// Code to perform constant propagation for matrix multiplication.
+//===----------------------------------------------------------------------===//
+
+// Takes the matrix shape and zero point for the LHS argument to MatMulInteger
+// and returns the zero point if it broadcasts to the matrix shape or else
+// returns the zero point reshaped so it broadcasts to the matrix shape.
+ElementsAttr reshapeMatMulIntegerLhsZero(
+    ArrayRef<int64_t> matrixShape, ElementsAttr zeroPoint) {
+  ShapedType zeroPointType = zeroPoint.getType();
+  auto zeroPointShape = zeroPointType.getShape();
```
tungld, line 425: I think we try to not overuse `auto` as LLVM coding standard: https://llvm.org/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable. In this case, I guest type is `ArrayRef<int64_t>`

src/Transform/ONNX/ConstProp.cpp
sorenlassen: ok, changed in PR #2140

src/Transform/ONNX/ConstProp.cpp
sorenlassen: ok, changed in PR #2140

PR #2133, opened 2023-04-01
use the same 'none' of NoneType everywhere

sorenlassen:
Replaced the use of empty tensors to represent 'none' in ONNXResizeOp decompositions and simplified isFromNone() to
```
// Test if the value is none. Since none is a unit value it never makes a
// difference whether it's a constant (the result of ONNXNoneOp) or the
// optional result of some other op (e.g. ONNXDropoutOp mask result).
// Note: It's ok to inline the isa<NoneType> test and not call this function.
inline bool isFromNone(mlir::Value value) {
  return llvm::isa<mlir::NoneType>(value.getType());
}
```

I inlined the implementation to emphasize that it's fine to inline the `isa<NoneType>` test like we already do in many places.

### DIFF
--- a/src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.td
+++ b/src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.td
@@ -54,7 +54,7 @@ class VariadicSizeIsGT<int N> : Constraint<
 >;
 
 // Create a unit constant that will be used as none input.
-def CreateUnitConstant : NativeCodeCall<"$_builder.create<ONNXNoneOp>($_loc)">;
+def CreateNoneConstant : NativeCodeCall<"$_builder.create<ONNXNoneOp>($_loc)">;
 
 class getNth<int n, string x> : NativeCodeCall<x # "[" # n # "]">;
 
@@ -437,7 +437,7 @@ def GetMatMulBiasLayoutStringAttr : NativeCodeCall<
 //                          (ZHighMatMulOp
 //                              (ZHighStickOp %X),
 //                              (ZHighStickOp %Y),
-//                              CreateUnitConstant)
+//                              CreateNoneConstant)
 //===----------------------------------------------------------------------===//
 
 def replaceONNXMatMulPattern : Pat<
@@ -446,7 +446,7 @@ def replaceONNXMatMulPattern : Pat<
      (ZHighMatMulOp
         (ZHighStickOp $x, (GetMatMulLayoutStringAttr (GetRank $x))),
         (ZHighStickOp $y, (GetMatMulLayoutStringAttr (GetRank $y))),
-        (CreateUnitConstant)))
+        (CreateNoneConstant)))
 >;
 
 //===----------------------------------------------------------------------===//
@@ -567,7 +567,7 @@ def replaceONNXGemmBias2DPattern : Pat<
        (ZHighMatMulOp
           (ZHighStickOp $a, (_2DLayoutAttr)),
           (ZHighStickOp $b, (_2DLayoutAttr)),
-          (CreateUnitConstant)),
+          (CreateNoneConstant)),
        (returnType $res)),
     $c),
   [(HasRankOf<2> $c)],
--- a/src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.td
+++ b/src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.td
@@ -54,7 +54,7 @@ class VariadicSizeIsGT<int N> : Constraint<
 >;
 
 // Create a unit constant that will be used as none input.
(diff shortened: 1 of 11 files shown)


src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.td
```diff
@@ -54,7 +54,7 @@ class VariadicSizeIsGT<int N> : Constraint<
 >;
 
 // Create a unit constant that will be used as none input.
-def CreateUnitConstant : NativeCodeCall<"$_builder.create<ONNXNoneOp>($_loc)">;
+def CreateNoneConstant : NativeCodeCall<"$_builder.create<ONNXNoneOp>($_loc)">;
```
tungld, line 57: It makes a lot of sense. How about using `CreateNoneValue` instead of `CreateNoneConstant`? as you commented its does not matter whether it's a constant or not. Furthermore, it would be consistent with the check function `isNoneValue`.

sorenlassen: good idea

done

sorenlassen:
@chentong319 @AlexandreEichenberger this PR follows up on [our discussion](https://github.com/onnx/onnx-mlir/pull/1965#issuecomment-1396413519) in PR #1965

please let me know what you think

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.td
```diff
@@ -54,7 +54,7 @@ class VariadicSizeIsGT<int N> : Constraint<
 >;
 
 // Create a unit constant that will be used as none input.
-def CreateUnitConstant : NativeCodeCall<"$_builder.create<ONNXNoneOp>($_loc)">;
+def CreateNoneValue : NativeCodeCall<"$_builder.create<ONNXNoneOp>($_loc)">;
```
AlexandreEichenberger, line 57: Thanks for simplifying the language...`CreateUnitConstant` is confusing to me, `CreateNoneValue` is not.

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -1065,9 +1065,7 @@ class FrontendGenImpl {
       if (v.empty()) {
         // Missing (optional) parameter.
         operandOnnxTypes.push_back(unspecifiedType);
-        auto no_value = builder_.create<ONNXNoneOp>(UnknownLoc());
-
-        operands.push_back(no_value);
+        operands.push_back(none());
```
AlexandreEichenberger, line 1068: Where does the `none()` definition comes from? Since `none` is such a generic term, and if a none value is created here, maybe a `createNoneValue()` might be a bit more explicit

src/Dialect/ONNX/ONNXOps/OpHelper.hpp
```diff
@@ -164,13 +164,14 @@ mlir::ElementsAttr getElementAttributeFromONNXValue(mlir::Value value);
...
 
-// Test if the value is from a NoneType or tensor<0 x ELEMENTARY_TYPE>
-bool isNoneValue(mlir::Value value);
+// Test if the value is none. Since none is a unit value it never makes a
+// difference whether it's a constant (the result of ONNXNoneOp) or the
+// optional result of some other op (e.g. ONNXDropoutOp mask result).
+// Note: It's ok to inline the isa<NoneType> test and not call this function.
+inline bool isNoneValue(mlir::Value value) {
+  return llvm::isa<mlir::NoneType>(value.getType());
+}
```
AlexandreEichenberger, line 174: inlining is good, tx

AlexandreEichenberger (approved):
LGTM,

would it make sense to convert all `ConstantUnit` and have a verifier that make sure we have none of them? I realize that this might not be a pattern that can be detected... so if that is the case, no need to spend resources doing so.

sorenlassen:
> would it make sense to convert all `ConstantUnit` and have a verifier that make sure we have none of them?

sorry, I don't understand what you mean - what is `ConstantUnit`?

src/Builder/FrontendDialectTransformer.cpp
sorenlassen: good idea, I renamed it as you suggested

it's method in the enclosing class FrontendGenImpl

AlexandreEichenberger:
> sorry, I don't understand what you mean - what is `ConstantUnit`?

Didn't you switch the `CreateUnitConstant` to `CreateNoneValue`? I was saying, if there was a way to verify that no UnitConstant escaped...

sorenlassen:
> > sorry, I don't understand what you mean - what is `ConstantUnit`?
> 
> Didn't you switch the `CreateUnitConstant` to `CreateNoneValue`? I was saying, if there was a way to verify that no UnitConstant escaped...

I see. I did a global search and replace and I'm confident that I got them all.

chentong319 (approved):
LGTM!

PR #2134, opened 2023-04-02
simplify AreTheSameAxes logic and createArrayAttrFromConstantOp

sorenlassen:
I studied AreTheSameConstantOpDenseAttr() to see if it was a candidate for using the ElementsAttr equal() function in PR #2114. I discovered that it isn't. I found it a little bit confusing and this PR attempts to to simplify it and clarify how it's used.

In more detail: Renamed `AreTheSameAxisArray` and `AreTheSameConstantOpDenseAttr` to `AreTheSameAxesArrayAttr` and `AreTheSameAxesConstant` and cleaned up their comments, contracts, and implementation. Moved them from OpHelper to Rewrite.cpp because they are only used in Rewrite.td. Also simplified the implementation of `createArrayAttrFromConstantOp`.

### DIFF
--- a/src/Dialect/ONNX/ONNXOps/OpHelper.cpp
+++ b/src/Dialect/ONNX/ONNXOps/OpHelper.cpp
@@ -13,7 +13,6 @@
 //===----------------------------------------------------------------------===//
 
 #include "mlir/IR/TypeUtilities.h"
-#include "llvm/ADT/SmallSet.h"
 #include "llvm/ADT/TypeSwitch.h"
 
 #include "src/Dialect/Mlir/IndexExpr.hpp"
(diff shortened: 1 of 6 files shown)


philass (approved):
Nice. Does look like an improvement.

Its worth noting that because the function is in `OpHelper.hpp` there is a chance that downstream users depend on the functions.

I think we should should discuss this in the next `onnx-mlir` meeting. Is there an implication that functions defined in certain files are used downstream. We need to formalize how we handle that going forward (i.e. which files/functions do downstream users depend on / how do we communicate changes when we do change these, etc...). 

The commentary above is not really specific to this PR. I thought I'd just leave the comment here because it came to mind.

LGTM!

sorenlassen:
> Its worth noting that because the function is in `OpHelper.hpp` there is a chance that downstream users depend on the functions.
> 
> I think we should should discuss this in the next `onnx-mlir` meeting. Is there an implication that functions defined in certain files are used downstream. We need to formalize how we handle that going forward (i.e. which files/functions do downstream users depend on / how do we communicate changes when we do change these, etc...).

good point, I agree that it would be good to discuss this and maybe formulate a policy

PR #2136, opened 2023-04-03
Debug isinf_pos and isinf_neg test

hamptonm1:


AlexandreEichenberger:
@hamptonm1 can you indicates if this work will continue somewhere else, or if the problem is now solved? Tx

PR #2140, opened 2023-04-04
replace createONNXConstantOpWithDenseAttr with OnnxBuilder

sorenlassen:
Addresses code review feedback from PR #2132.

Also changed createReplacingConstantOp() to use OnnxBuilder and return Value.

### DIFF
--- a/src/Transform/ONNX/ConstProp.cpp
+++ b/src/Transform/ONNX/ConstProp.cpp
@@ -27,7 +27,6 @@
 #include "src/Dialect/ONNX/ONNXOps.hpp"
 #include "src/Dialect/ONNX/ONNXOps/OpHelper.hpp"
 #include "src/Dialect/ONNX/ONNXOps/ShapeHelper.hpp"
-#include "src/Dialect/ONNX/DialectBuilder.hpp"
 #include "src/Dialect/ONNX/OnnxElementsAttrBuilder.hpp"
 #include "src/Pass/Passes.hpp"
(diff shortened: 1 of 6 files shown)


src/Transform/ONNX/ConstProp.cpp
```diff
@@ -93,12 +94,10 @@ ElementsAttr getConstValueElements(Value constValue) {
...
-// Creates ONNXConstantOp with the location and result type from replacingValue.
-ONNXConstantOp createReplacingConstantOp(
+// Creates ONNXConstantOp with the location from replacingValue.
+Value createReplacingConstantOp(
     PatternRewriter &rewriter, Value replacingValue, ElementsAttr elements) {
-  return rewriter.create<ONNXConstantOp>(replacingValue.getLoc(),
-      replacingValue.getType(), Attribute(), elements, FloatAttr(), ArrayAttr(),
-      IntegerAttr(), ArrayAttr(), StringAttr(), ArrayAttr());
+  return OnnxBuilder(rewriter, replacingValue.getLoc()).constant(elements);
 }
```
sorenlassen, line 101: Note that this change means that the Constant op gets the type from elements attribute instead of replacingValue. In the past we had a couple of cases where these types disagreed (the Constant had unranked type or a dynamic dim) but I think we've cleared up all those cases. We need the Constant op type to agree with its attribute to make constant folding work going forward.

tungld (approved):
LGTM. Thanks for following up!

PR #2142, opened 2023-04-05
Update RunONNXModel.py to allow verification using softmax

tungld:
Sometimes we want to verify the inference output by applying softmax to normalize the output. This patch introduces an additional option `--verify-with-softmax` to `RunONNXModel.py` to verify using softmax.

Before this patch, we always verify every value in the output. Now, we add an option `--verify-every-value` to turn it on/off.


AlexandreEichenberger (approved):
LGTM, please add before merging in the help output which options are default? By using the `[ ]` in the top level usage, it does not seems to imply one of them is mandatory.

utils/RunONNXModel.py
```diff
@@ -129,6 +138,13 @@ def valid_onnx_input(fname):
                     " uint64, int64, float16, float32, float64")
 
 args = parser.parse_args()
+if (args.verify and not (args.verify_with_softmax or args.verify_every_value)):
+    raise RuntimeError("Choose verification mode: --verify-with-softmax or "
+                       "--verify-every-value or both")
+if (args.verify_with_softmax and (not args.verify)):
+    raise RuntimeError("Must specify --verify to use --verify-with-softmax")
+if (args.verify_every_value and (not args.verify)):
+    raise RuntimeError("Must specify --verify to use --verify-every-value")
```
tungld, line 147: > LGTM, please add before merging in the help output which options are default? By using the [ ] in the top level usage, it does not seems to imply one of them is mandatory.

I do the check here since I haven't found ways to make dependent arguments, say, once `--verify` is given, one must have `--verify-with-softmax` or `--verify-every-value` or both.

utils/RunONNXModel.py
```diff
@@ -74,10 +74,19 @@ def valid_onnx_input(fname):
 parser.add_argument('--verify',
                     choices=['onnxruntime', 'ref'],
                     help="Verify the output by using onnxruntime or reference"
-                    " inputs/outputs. By default, no verification")
+                    " inputs/outputs. By default, no verification. When being"
+                    " enabled, --verify-with-softmax or --verify-every-value"
+                    " must be used to specify verification mode")
```
tungld, line 79: Explain more about `-verify` that need to specify verification mode via `--verify-with-softmax` or `--verify-every-value`

PR #2145, opened 2023-04-06
Elementwise simd: generate simd code with broadcast from a scalar

AlexandreEichenberger:
When we have broadcast of a single tensor to one or more scalar, we can generate SIMD code without any issues.

e.g.
```
  %0 = "onnx.Add"(%arg0, %arg1) : (tensor<?x?x64xf32>, tensor<f32>) -> tensor<*xf32>
```

This PR add this with two lit-tests.

This PR also consolidate the SIMD code generation from unary/binary/variadic patterns into a single function, to cut down on redundancy.


AlexandreEichenberger:
Has code generation issues that are fixed in #2161. #2161 is also more general and subsumes new features being added here

PR #2146, opened 2023-04-07
Add sameDynShape and broadcastLastDim in DimAnalysis and re-word unknown to dynamic

tungld:
This patch adds some additional functions to DimAnalysis:
- `sameDynShape` to check if dynamic dimensions from two tensors are the same or not by ignoring static dimensions
- `broadcastLastDim` to check if one tensor is broadcasting to the other tensor using **the last dimension**

For example:
- For `tensor<?x1xf32>` and `tensor<?x5xf32`:  `sameDynShape` and `broadcastLastDim` return `true` if their dynamic dimensions are the same.
- For `tensor<?x2xf32>` and `tensor<?x5xf32`, `sameDynShape` returns `true` but `broadcastLastDim` return `false`  if their dynamic dimensions are the same.
- For `tensor<?x2x1xf32>` and `tensor<?x?x5xf32`, `sameDynShape` and `broadcastLastDim` return `false` because it cannot check equality of the second dimension: `2` vs `?`.

This patch also re-words `unknown` to `dynamic` in DimAnalysis to make it consistent with `ShapedType::kDynamic`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,7 +21,7 @@ A comprehensive list of documents is found [here](docs/DocumentList.md).
 * To test new code, see [here](docs/Testing.md) for instructions.
 * A guide on how to do constant propagation for ONNX operations is found
   [here](docs/ConstPropagationPass.md).
-* A guide on how to analyze unknown dimensions and query their equality at compile time is found [here](docs/UnknownDimensionAnalysis).
+* A guide on how to analyze dynamic dimensions and query their equality at compile time is found [here](docs/DynamicDimensionAnalysis).
 * To build and test for specialized accelerator, see [here](docs/AccelNNPAHowToUseAndTest.md)
 
 ## ONNX-MLIR specific dialects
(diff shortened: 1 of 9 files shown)


CONTRIBUTING.md
```diff
@@ -21,7 +21,7 @@ A comprehensive list of documents is found [here](docs/DocumentList.md).
 * To test new code, see [here](docs/Testing.md) for instructions.
 * A guide on how to do constant propagation for ONNX operations is found
   [here](docs/ConstPropagationPass.md).
-* A guide on how to analyze unknown dimensions and query their equality at compile time is found [here](docs/UnknownDimensionAnalysis).
+* A guide on how to analyze dynamic dimensions and query their equality at compile time is found [here](docs/DynamicDimensionAnalysis).
```
AlexandreEichenberger, line 24: +1 on standardizing the terms that we use. This greatly help our users long term. We should do more of that, thanks for fixing this here.

AlexandreEichenberger (approved):

PR #2147, opened 2023-04-07
Lowering ONNXMatMulInteger to Krnl dialect

tungld:
This patch adds the lowering of `ONNXMatMulInteger` to Krnl dialect.

The lowering of `KrnlMatMul` to affine now supports integers for SIMD code. This patch also adds unsigned integer support for `add` and `mul` in MathBuilder.

This patch enables `bertsquad-12-int8` in Jenkins since it now passes the test.


src/Conversion/ONNXToKrnl/Math/MatMulInteger.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+    Value AUInt32, BUInt32;
+    // Use i32 for onnx.Sub because onnx.Sub does not support ui32.
+    // It's because `arith` dialect does not have `subui` though it has `addui`
+    // and `mului`.
+    if (!isNoneValue(aZeroPoint)) {
+      Value AInt32 = create.onnx.cast(A, i32Ty);
+      Value aZeroPointInt32 = create.onnx.cast(aZeroPoint, i32Ty);
+      // Note: `sub` is not true if aZeroPoint shape is [M], M != 1,
+      // because [MxK] - [M] cannot be expressed by `sub`.
```
sorenlassen, line 78: > [MxK] - [M] cannot be expressed by `sub`
we could unsqueeze aZeroPoint like we do in MatMulInteger const prop: https://github.com/onnx/onnx-mlir/blob/main/src/Transform/ONNX/ConstProp.cpp#L438-L443

tungld: I updated this to handle this case. Do you know if we have this situation: A is [D1, D2, M, K] and aZeroPoint is [D1, D2, M]? I am not sure but handle such a situation too? Thanks!

src/Conversion/ONNXToKrnl/Math/MatMulInteger.cpp
```diff
@@ -0,0 +1,112 @@
...
+    if (!isNoneValue(bZeroPoint)) {
+      Value BInt32 = create.onnx.cast(B, i32Ty);
+      // K is the broadcating dim: [KxN] - [N] = [KxN] - [1xN]
+      Value bZeroPointInt32 = create.onnx.cast(bZeroPoint, i32Ty);
+      BInt32 = create.onnx.sub(BInt32, bZeroPointInt32);
+      BUInt32 = create.onnx.cast(BInt32, ui32Ty);
+    } else
+      BUInt32 = create.onnx.cast(B, ui32Ty);
+
+    // Emit MatMul for ui32.
```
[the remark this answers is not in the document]
tungld: No particular reason as MatMulInteger's input is i8 or ui8, so I used unsigned integer here. Perhaps, we should check input's type to decide which one to use. For example, 
- Both A and B is i8 -> use i32
- Both A and B is ui8 -> use ui32
- A is i8, B is ui8 (and vice versa) -> not sure which one is better: i32 or ui32. Looks like i32 is convenient.

What do you think?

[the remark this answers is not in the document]
sorenlassen: I recommend to always use i32 because the ranges of i8 and u8 (after taking zero inputs into account) both fit in i32, whereas u32 cannot represent negative values, and not that even u8 input can express negative values if a non-zero zero input is given

the spec also says i32 and that's what I used for constant propagation

src/Conversion/ONNXToKrnl/Math/MatMulInteger.cpp
sorenlassen: The spec says:
> If the input is N-D tensor with shape [D1, D2, M, K] then zero point tensor may have shape [D1, D2, M, 1].

In the constant propagation implementation I chose that to mean that a zero point tensor with rank > 1 must broadcast to the corresponding multidimensional matrix input shape. See https://github.com/onnx/onnx-mlir/blob/main/src/Transform/ONNX/ConstProp.cpp#L446-L450

src/Conversion/ONNXToKrnl/Math/MatMulInteger.cpp
```diff
@@ -0,0 +1,114 @@
...
+    shapeHelper.computeShapeAndAssertOnFailure();
+
+    Type i32Ty = rewriter.getIntegerType(32);
+    Type ui32Ty = rewriter.getIntegerType(32, /*isSigned=*/false);
+    assert(resElementType == i32Ty && "Output type is not i32");
+
+    Value AUInt32, BUInt32;
+    // Use i32 for onnx.Sub because onnx.Sub does not support ui32.
+    // It's because `arith` dialect does not have `subui` though it has `addui`
+    // and `mului`.
```
sorenlassen, line 65: as discussed in another comment thread, i32 also seems like the better choice because we want to support negative values

A and AZeroPoint can represent negative numbers after subtraction even if they are both u8 to begin with

src/Conversion/ONNXToKrnl/Math/MatMulInteger.cpp
tungld: [D1, D2, M, K] and [D1, D2, M, 1] is correct for broadcasting, so no problem to do subtraction. What I meant is [D1, D2, M, K] and [D1, D2, M], where we should unsqueeze. I seems the spec does not mention that case, but only ([M, K], [M]) and ([D1, D2, M, K], [D1, D2, M, 1]).

sorenlassen: what I meant is that I interpret the specification
> If the input is 2D of shape [M, K] then zero point tensor may be an M element vector [zp_1, zp_2, ..., zp_M]. If the input is N-D tensor with shape [D1, D2, M, K] then zero point tensor may have shape [D1, D2, M, 1].
to mean that a zero point with rank > 1 must broadcast to the N-D tensor shape as in the example [D1, D2, M, K] and [D1, D2, M, 1], so we should only unsqueeze zero point if it is has rank 1 as in the the example [M, K] and [zp_1, zp_2, ..., zp_M]

I hope that's clear from the implementation, asserts, and comments in ConstProp.cpp - the asserts are meant to alert us to any model that disagrees with this interpretation of the spec

maybe we can factor out the logic that decides how to unsqueeze a_zero_point into a function we can share between constant propagation and lowering

src/Conversion/ONNXToKrnl/Math/MatMulInteger.cpp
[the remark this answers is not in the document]
tungld: It makes sense. I moved to using i32 for computation. Thanks for the discussion!

src/Conversion/ONNXToKrnl/Math/MatMulInteger.cpp
tungld: Totally agree. It would be a good idea to put these constraints to the verifier for MatMulInteger so that it would be valid for both ConstProp.cpp and lowering.

src/Dialect/ONNX/ONNXOps/Math/MatMul.cpp
```diff
@@ -189,6 +189,100 @@ LogicalResult ONNXMatMulIntegerOp::inferShapes(
...
+    return success();
+
+  // If AZeroPoint is [M] (M != 1), A must be [M, K]
+  // If AZeroPoint rank is > 1, A must have the same rank, e.g.
+  // - [D1, D2, ..., DN, M, 1] if A is [D1, D2, ..., DN, M, K]
+  Value A = operandAdaptor.getA();
+  Value aZeroPoint = this->getAZeroPoint();
+  if (!isNoneValue(aZeroPoint)) {
+    auto aType = dyn_cast<ShapedType>(A.getType());
+    auto aZeroPointType = dyn_cast<ShapedType>(aZeroPoint.getType());
```
sorenlassen, line 204: since you rely on aType and aZeroPointType being non-null below, it's clearer to use `cast` instead of `dyn_cast`: both clearer to read the intention and also gives a clearer cast failure at runtime

alternatively, you can omit the explicit cast:
```
ShapedType aType, aZeroPointType = A.getType(), aZeroPointType.getType();
```
because ShapedType is a type interface: you can assign any Type to it and the constructor/assignment operator will check that the type is a ShapedType
https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Support/InterfaceSupport.h#L93-L99

src/Dialect/ONNX/ONNXOps/Math/MatMul.cpp
```diff
@@ -189,6 +189,100 @@ LogicalResult ONNXMatMulIntegerOp::inferShapes(
...
+      for (uint64_t i = 0; i < aRank - 1; ++i) {
+        if (!aType.isDynamicDim(i) && !aZeroPointType.isDynamicDim(i) &&
+            (aShape[i] != aZeroPointShape[i])) {
+          return onnx_mlir::Diagnostic::emitDimensionsMustHaveSameValueError(
+              *this->getOperation(), "A", i, aShape[i], "aZeroPoint", i,
+              aZeroPointShape[i]);
+        }
+      }
+      uint64_t lastAxis = aRank - 1;
+      if (!aType.isDynamicDim(lastAxis) &&
```
sorenlassen, line 230: I think you can leave out the check `!aType.isDynamicDim(lastAxis)` because aZeroPointType shouldn't have a non-1 static dim size in the last axis regardless of whether the dim size of a's last dimension is dynamic

src/Conversion/ONNXToKrnl/Math/MatMulInteger.cpp
```diff
@@ -0,0 +1,101 @@
...
+    if (!isNoneValue(aZeroPoint)) {
+      Value aZeroPointInt32 = create.onnx.cast(aZeroPoint, resElementType);
+      int64_t aRank = aType.getRank();
+      int64_t aZeroPointRank = aZeroPointType.getRank();
+      // If broadcasting, e.g. A is [MxK], zeroPoint is [M], M != 1.
+      // Unsqueeze zeroPoint to [Mx1] to make shapes compatible.
+      // There is no need to handle scalar zeroPoint (e.g. tensor<dtype> or
+      // tensor<1xdtype>), which is always true for broadcasting.
+      if (((aZeroPointRank == 1) && (aZeroPointType.getShape()[0] != 1)) ||
+          ((aZeroPointRank > 1) && (aZeroPointRank = aRank - 1))) {
```
sorenlassen, line 69: given that verify() checks that aZeroPointRank == rank when aZeroPointRank > 1, I think you should only unsqueeze when aZeroPointRank == 1

otherwise the logic here disagrees with constant propagation which is inconsistent

src/Dialect/ONNX/ONNXOps/Math/MatMul.cpp
```diff
@@ -189,6 +189,100 @@ LogicalResult ONNXMatMulIntegerOp::inferShapes(
...
+  if (!isNoneValue(bZeroPoint)) {
+    auto bType = dyn_cast<ShapedType>(B.getType());
+    auto bZeroPointType = dyn_cast<ShapedType>(bZeroPoint.getType());
+    uint64_t bRank = bType.getRank();
+    uint64_t bZeroPointRank = bZeroPointType.getRank();
+    ArrayRef<int64_t> bShape = bType.getShape();
+    ArrayRef<int64_t> bZeroPointShape = bZeroPointType.getShape();
+
+    // If BZeroPoint is [N], B must be [K, N]
+    if ((bZeroPointRank == 1) && (bRank != 2))
```
sorenlassen, line 254: we should allow bZeroPointRank == 1 for any shape if bZeroPointShape[0] == 1 or is dynamic (and could become 1 at runtime) because scalar zero can be expressed with a singleton rank 1 shape [1], according to the example in the spec https://github.com/onnx/onnx/blob/main/docs/Operators.md#examples-84

src/Dialect/ONNX/ONNXOps/Math/MatMul.cpp
```diff
@@ -189,6 +189,100 @@ LogicalResult ONNXMatMulIntegerOp::inferShapes(
...
+        if (i == bRank - 2)
+          continue;
+        if (!bType.isDynamicDim(i) && !bZeroPointType.isDynamicDim(i) &&
+            (bShape[i] != bZeroPointShape[i])) {
+          return onnx_mlir::Diagnostic::emitDimensionsMustHaveSameValueError(
+              *this->getOperation(), "B", i, bShape[i], "bZeroPoint", i,
+              bZeroPointShape[i]);
+        }
+      }
+      uint64_t kAxis = bRank - 1;
```
sorenlassen, line 274: I believe this should be `bRank - 2`

src/Conversion/ONNXToKrnl/Math/MatMulInteger.cpp
chentong319: In bumpy.matmul:
```
Matrix product of two arrays.

The behavior depends on the arguments in the following way.

- If both arguments are 2-D they are multiplied like conventional matrices.
- If either argument is N-D, N > 2, it is treated as a stack of matrices residing in the last two indexes and broadcast accordingly.
- If the first argument is 1-D, it is promoted to a matrix by prepending a 1 to its dimensions. After matrix multiplication the prepended 1 is removed.
- If the second argument is 1-D, it is promoted to a matrix by appending a 1 to its dimensions. After matrix multiplication the appended 1 is removed.

```

sorenlassen: @chentong319 I believe all of that is dealt with in `onnx.matmul()`

matching the zero point shapes in MatMulInteger is separate issue, I believe

src/Conversion/ONNXToKrnl/Math/MatMulInteger.cpp
tungld: if there is only one element in the tensor (`aZeroPointRank == 1` and `aZeroPointShape[0] == 1`), we don't need to unsqueeze too. It's similar to scalar tensor.

src/Dialect/Mlir/DialectBuilder.cpp
```diff
@@ -112,8 +112,18 @@ Value MathBuilder::xori(Value lhs, Value rhs) const {
 
 Value MathBuilder::add(Value lhs, Value rhs) const {
```
sorenlassen, line 113: are the changes to MathBuilder::add/mul/cast necessary for MatMulInteger? please consider adding an explanation of them in the PR description

src/Conversion/ONNXToKrnl/Math/MatMulInteger.cpp
```diff
@@ -0,0 +1,100 @@
...
+    Value AInt32 = create.onnx.cast(A, resElementType);
+    if (!isNoneValue(aZeroPoint)) {
+      Value aZeroPointInt32 = create.onnx.cast(aZeroPoint, resElementType);
+      int64_t aRank = aType.getRank();
+      int64_t aZeroPointRank = aZeroPointType.getRank();
+      // If broadcasting, e.g. A is [MxK], zeroPoint is [M], M != 1.
+      // Unsqueeze zeroPoint to [Mx1] to make shapes compatible.
+      // There is no need to handle scalar zeroPoint (e.g. tensor<dtype> or
+      // tensor<1xdtype>), which is always true for broadcasting.
+      if ((aZeroPointRank == 1) && (aZeroPointType.getShape()[0] != 1)) {
```
sorenlassen, line 68: LGTM, this matches the logic in `reshapeMatMulIntegerLhsZero()` in ConstProp.cpp

src/Conversion/ONNXToKrnl/Math/MatMulInteger.cpp
```diff
@@ -0,0 +1,100 @@
...
+    Value B = adaptor.getB();
+    Value aZeroPoint = mmiOp.getAZeroPoint(); // Optional input.
+    Value bZeroPoint = mmiOp.getBZeroPoint(); // Optional input.
+
+    // Common types.
+    auto aType = A.getType().dyn_cast<ShapedType>();
+    auto aZeroPointType = aZeroPoint.getType().dyn_cast<ShapedType>();
+    auto resMemRefType = dyn_cast<MemRefType>(
+        typeConverter->convertType(mmiOp.getResult().getType()));
+    Type resElementType = resMemRefType.getElementType();
```
sorenlassen, line 52: suggestion:
`assert(resMemRefType.getElementType() == rewriter.getI32Type())`
and rename resElementType to I32, then it's easier for the reader to understand the Int32 names below: AInt32, aZeroPointInt32, etc

sorenlassen (approved):

src/Dialect/ONNX/ONNXOps/Math/MatMul.cpp
tungld: Thanks! Now I learned a new thing in MLIR.

PR #2098, opened 2023-03-22
ONNXHybridTransformPass

sorenlassen:
Includes a new pattern-based shape inference implementation.

So far only combines shape inference and canonicalization but the plan is to include constant propagation and decomposition so that a single pass with all the patterns can cascade shape inference and all the transforms in a single pass.

I encountered a model with a long dependency chain between shape inference, canonicalization, and constant propagation which requires an incredible number of repetitive passes with the current pass infrastructure.

See PR #1770 for an earlier version of this PR.

PR #2148, opened 2023-04-07
make inferShapes() not crash on unranked inputs

sorenlassen:
Fixed inferShapes() crashes on unranked inputs in LoopOp, ScanOp, ONNXGlobalAveragePoolOp, ONNXGlobalLpPoolOp, ONNXGlobalMaxPoolOp. Added lit tests which would crash before these fixes. These crashes were discovered in draft PR #2098.

Cleaned up the LoopOp and ScanOp inferShapes() implementations to make them easier to read and compare.

Also fixed ill-formed loops in test_loop lit test is onnx_lowering.mlir and onnx_lowering_O3.mlir which crashed on a new assert 
```
assert(loopBody.getNumArguments() == 2 + getVInitial().size())
```
in ONNXLoopOp::inferShapes(). I added a TODO to add a verify() method to check this.

### DIFF
--- a/src/Dialect/ONNX/ONNXOps/ControlFlow/Loop.cpp
+++ b/src/Dialect/ONNX/ONNXOps/ControlFlow/Loop.cpp
@@ -18,14 +18,6 @@ using namespace mlir;
 using namespace mlir::OpTrait::util;
 using namespace onnx_mlir;
 
-namespace {
-template <typename Range>
-Range take_begin(Range range, size_t N) {
-  assert(range.size() >= N);
-  return llvm::make_range(range.begin(), range.begin() + N);
-}
-} // namespace
-
 //===----------------------------------------------------------------------===//
 // Verifier
 //===----------------------------------------------------------------------===//
@@ -43,6 +35,7 @@ LogicalResult ONNXLoopOp::inferShapes(
 
   auto &loopBody = getRegion();
   // Body inputs: trip count, termination condition, loop carried dependencies.
+  // TODO: Add verifier to check this.
   assert(loopBody.getNumArguments() == 2 + numCarried &&
          "Loop body must take at least 2 inputs.");
 
@@ -104,11 +97,14 @@ LogicalResult ONNXLoopOp::inferShapes(
 // Helper function to obtain subset of op results corresponding to the final
 // value of loop carried dependencies.
 Operation::result_range ONNXLoopOp::v_final() {
-  return take_begin(getVFinalAndScanOutputs(), getVInitial().size());
+  size_t numCarried = getVInitial().size();
+  auto outputs = getVFinalAndScanOutputs();
+  return llvm::make_range(outputs.begin(), outputs.begin() + numCarried);
 }
 
 // Helper function to obtain subset of op results corresponding to the scan
 // outputs.
 Operation::result_range ONNXLoopOp::scan_outputs() {
-  return llvm::drop_begin(getVFinalAndScanOutputs(), getVInitial().size());
+  size_t numCarried = getVInitial().size();
+  return llvm::drop_begin(getVFinalAndScanOutputs(), numCarried);
 }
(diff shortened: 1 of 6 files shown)


src/Dialect/ONNX/ONNXOps/ControlFlow/Loop.cpp
```diff
@@ -30,8 +30,13 @@ using namespace onnx_mlir;
...
   auto builder = Builder(getContext());
+
+  size_t numCarried = getVInitial().size();
+
   auto &loopBody = getRegion();
-  assert(loopBody.getNumArguments() >= 2 &&
+  // Body inputs: trip count, termination condition, loop carried dependencies.
+  // TODO: Add verifier to check this.
+  assert(loopBody.getNumArguments() == 2 + numCarried &&
          "Loop body must take at least 2 inputs.");
```
chentong319, line 40: update the assertion failure message according to new condition

src/Dialect/ONNX/ONNXOps/ControlFlow/Loop.cpp
```diff
@@ -42,56 +47,44 @@ LogicalResult ONNXLoopOp::inferShapes(
...
 
-  // Now we have modified loop body function input signatures according to
-  // the knowledge we have on the inputs we pass to this function. Dispatch
-  // shape inference to obtain body function output types.
+  // Now we have modified loop body input types according to
+  // the knowledge we have on the initial inputs. Dispatch
+  // shape inference to obtain body output types.
   doShapeInference(loopBody);
+  Operation *terminator = loopBody.back().getTerminator();
+  assert(terminator->getNumOperands() == 1 + getVFinalAndScanOutputs().size());
```
chentong319, line 61: Add the assertion error message
`&& "number of Values returned from LoopOp body and Loop output does not match"`

src/Dialect/ONNX/ONNXOps/ControlFlow/Loop.cpp
```diff
@@ -42,56 +47,44 @@ LogicalResult ONNXLoopOp::inferShapes(
...
-    // input to Loop operation, but we need to eliminate the possibility of
-    // early termination to be sure.
-    unsqueezedShape.insert(unsqueezedShape.begin(), ShapedType::kDynamic);
-    updateType(std::get<0>(vScanOutputValToTy), unsqueezedShape,
-        rankedScanTy.getElementType(), /*encoding=*/nullptr,
-        /*refineShape=*/false);
+  auto bodyScanOutputTys = llvm::drop_begin(bodyOuputTys, numCarried);
+  for (auto [opScanOutput, ty] : llvm::zip(scan_outputs(), bodyScanOutputTys)) {
+    // TODO: Handle SeqType, OptType.
+    if (auto rankedTy = ty.dyn_cast<RankedTensorType>()) {
```
chentong319, line 77: I guess that this IF is the fix to avoid crash on unranked tensor. The PR is more precisely to fix unranked return, isn't it

sorenlassen: yes, in PR #2098 Loop and Range crashes on unranked return types

the unranked lit tests in this PR also caused a crash in Scan's input scan types

src/Dialect/ONNX/ONNXOps/ControlFlow/Scan.cpp
```diff
@@ -28,77 +40,69 @@ using namespace onnx_mlir;
...
-  }
+  // Set state variable ScanOp output types (and shapes) to the types
+  // and inferred shapes of their counterparts in the body output.
+  // zip() runs through the shortest range which is v_final().
+  for (auto [val, ty] : llvm::zip(v_final(), bodyOuputTys))
+    val.setType(ty);
 
   // For scan outputs, we set their shape to be the shape of the return
-  // values of the loop body function corresponding to scan outputs, but
+  // values of the body corresponding to scan outputs, but
```
chentong319, line 97: Like this change to remove "function" in  the original comment.

chentong319 (approved):
LGTM!

Small changes if you want to spend time.

src/Dialect/ONNX/ONNXOps/ControlFlow/Loop.cpp
sorenlassen: changed to "LoopOp outputs count must match body results count"

also added a message in the corresponding assert in Scan.cpp

src/Dialect/ONNX/ONNXOps/ControlFlow/Loop.cpp
sorenlassen: changed to "LoopOp inputs count must match body operands count"

also added a message to the corresponding assert in Scan.cpp

PR #2149, opened 2023-04-07
lit test fixes

sorenlassen:
These are fixes to lit test issues discovered in draft PR #2098.

In some cases scalar element attributes' shapes were fixed to match the onnx.Constant's shapes, e.g. in onnx_lowering.mlir:
```
 func.func @test_sequence_erase(%arg0: !onnx.Seq<tensor<?x4x5xf32>>) -> tensor<3xi64>  {
-  %0 = onnx.Constant {value = dense<0> : tensor<1xi64>} : tensor<i64>
+  %0 = onnx.Constant {value = dense<0> : tensor<i64>} : tensor<i64>
```
This causes problems in PR #2098 and will also be bad for constant folding, so it's best to fix that in the lit tests.

In some other cases lit tests have dead code and common subexpressions which create problems in PR #2098 where shape inference can perform dead code and common subexpression elimination. The fixes in this PR make dead code un-dead by connecting results with return statements and make common subexpressions (constants with identical values) un-common by making subexpressions unique.

In some cases constants are hoisted out of nested regions to make it easier make some small follow-up changes in PR #2098.


sorenlassen:
surprising build failure https://www.onnxmlir.xyz/jenkinp/job/ONNX-MLIR-Pipeline-Docker-Build/9647/console
```
In file included from /workdir/onnx-mlir/src/Accelerators/NNPA/Transform/ZLow/ZLowRewrite.cpp:26:
/workdir/onnx-mlir/src/Accelerators/NNPA/Support/LayoutHelper.hpp:17:10: fatal error: zdnn.h: No such file or directory
```

chentong319 (approved):
LGTM!

PR #2152, opened 2023-04-08
fix compiler warnings

sorenlassen:
Fixes compiler warnings:
```
ONNXDimAnalysis.cpp:240:21: warning: unused variable 'shape2' [-Wunused-variable]
  ArrayRef<int64_t> shape2 = tensor2.getType().cast<ShapedType>().getShape();

TestLoop.cpp:126:23: warning: unqualified call to 'std::move' [-Wunqualified-std-cast-call]
  inputs.emplace_back(move(yInitTensor));

OMTensorHelper.hpp:31:9: warning: unqualified call to 'std::move' [-Wunqualified-std-cast-call]
  s = move(r);
```
Observed with Apple clang version 14.0.3 on MacOS.

Also removed a single (unneeded) occurence of `using namespace std`.


tungld (approved):
LGTM. Thanks!

PR #2156, opened 2023-04-11
Update resize for opset 18

philass:


chentong319:
The failure of the backend is caused by the attributes added in the new version of onnx. I used a generic way to transform all attributes to  external library calls. One option is that you comment out the resize test cases for backend in this PR, and I open another PR to fix that and bring those test cases back.

philass:
Hey @chentong319,

Thanks for the feedback. I appreciate that. I've commented out the tests that were effected. Please take a look at the PR when you get time

chentong319 (approved):
LGTM!

Issue #2154, opened 2023-04-10
Error for compiling (similar to #


negiyas:
How to reproduce this issue.
(1) Patch the following diff to avoid issue#2153.
```
$ git diff main
diff --git a/src/Conversion/ONNXToKrnl/ML/CategoryMapper.cpp b/src/Conversion/ONNXToKrnl/ML/CategoryMapper.cpp
index 0d63032e..5b3b3a64 100644
--- a/src/Conversion/ONNXToKrnl/ML/CategoryMapper.cpp
+++ b/src/Conversion/ONNXToKrnl/ML/CategoryMapper.cpp
@@ -87,7 +87,6 @@ struct ONNXCategoryMapperOpLowering

     // Basic information.
     int64_t rank = memRefType.getShape().size();
-    assert(((rank == 1) || (rank == 2)) && "Invalid rank of input");
     ShapedType inputType = X.getType().cast<ShapedType>();
     Type elementType = inputType.getElementType();

diff --git a/src/Dialect/ONNX/ONNXOps/ML/CategoryMapper.cpp b/src/Dialect/ONNX/ONNXOps/ML/CategoryMapper.cpp
index 623b84f4..1148ef8c 100644
--- a/src/Dialect/ONNX/ONNXOps/ML/CategoryMapper.cpp
+++ b/src/Dialect/ONNX/ONNXOps/ML/CategoryMapper.cpp
@@ -47,8 +47,6 @@ LogicalResult ONNXCategoryMapperOp::verify() {
   }

   ShapedType inputType = X.getType().cast<ShapedType>();
-  if ((inputType.getRank() != 1) && (inputType.getRank() != 2))
-    return emitOpError("input rank must be one or two");
   Type elementType = inputType.getElementType();
   if (!elementType.isInteger(64) && !elementType.isa<ONNXStringType>())
     return emitOpError("input must be a tensor of int64 or string");
```
(2) Compile bidaf-9.onnx ( https://github.com/onnx/models/blob/main/text/machine_comprehension/bidirectional_attention_flow/model/bidaf-9.onnx ) as follows.
```
$ ./build/Debug/bin/onnx-mlir -mcpu=z14 --EmitMLIR  bidaf-9.onnx
[...]

Issue #2150, opened 2023-04-08
Compile error for Roberta-base-11 when input shape 1x1 is provided


chentong319:
Helping @hamptonm1 debug the problem, I got some clue for this issue. My finding is that the error came from broadcast handling.

Compile with `onnx-mlir --O3 --EmitLLVMIR --maccel=NNPA roberta-base-11.onnx --onnx-op-stats TXT --shapeInformation 0:1x1`

I got the error message 
```
Assertion failed: (map.getNumInputs() == mapOperands.size() && "inconsistent index info"), function build, file /Users/tongchen/Projects/llvm-project/mlir/lib/Dialect/Affine/IR/AffineOps.cpp, line 2981.
```
The problematic krnl loop is 
```
    krnl.iterate(%927#0, %927#1, %927#2) with (%927#0 -> %arg1 = 0 to 1, %927#1 -> %arg2 = 0 to 1, %927#2 -> %arg3 = 0 to 768){
      %5216:3 = krnl.get_induction_var_value(%927#0, %927#1, %927#2) : (!krnl.loop, !krnl.loop, !krnl.loop) -> (index, index, index)
      %5217 = krnl.load %alloc_5932[%5216#0, %5216#1, %5216#2] : memref<1x1x768xf32>
      %5218 = krnl.load %66[%5216#0, %5216#1, %5216#2] : memref<768xf32>
      %5219 = arith.addf %5217, %5218 : f32
      krnl.store %5219, %alloc_5938[%5216#0, %5216#1, %5216#2] : memref<1x1x768xf32>
    }
```
The index and tensor dimension does not match for this load `%5218 = krnl.load %66[%5216#0, %5216#1, %5216#2] : memref<768xf32>`.

This load came from this onnx.Add with ops around it
```
    %66 = onnx.Constant dense_resource<__elided__> : tensor<768xf32>
...
    %777 = "zhigh.Add"(%775, %776) : (tensor<1x1x768xf32, #zhigh.layout<{dataLayout = "3D"}>>, tensor<1x1x768xf32, #zhigh.layout<{dataLayout = "3D"}>>) -> tensor<1x1x768xf32, #zhigh.layout<{dataLayout = "3D"}>>
    %778 = "zhigh.Unstick"(%777) : (tensor<1x1x768xf32, #zhigh.layout<{dataLayout = "3D"}>>) -> tensor<1x1x768xf32>
[...]

Issue #2137, opened 2023-04-03
Fix lagging when compiling roberta-base-11 model


hamptonm1:
It seems when testing roberta-base-11 model it compiles within a minute dynamically but statically it is taking a long time to compile as if there was an infinite loop.

tungld:
@hamptonm1 could you provide the command used to compile? Thanks!

cjvolzka:
Same flags I used on https://github.com/onnx/onnx-mlir/issues/1887.

If I use `--shapeInformation` I get this and then aborted it after 76 min.
```
--O3 --EmitLib --mtriple=s390x-ibm-loz --mcpu=z16 --maccel=NNPA roberta-base-11.onnx --onnx-op-stats TXT --shapeInformation 0:1x1 -v
Operations encountered:
-----------------------
   func.func               , 1
   func.return             , 1
   onnx.Add                , 58
   onnx.Cast               , 2
   onnx.Concat             , 80
   onnx.Constant           , 71
   onnx.CumSum             , 1
   onnx.Dim                , 116
   onnx.Div                , 41
   onnx.Equal              , 1
   onnx.Erf                , 12
   onnx.Gather             , 3
   onnx.LayoutTransform    , 56
   onnx.Mul                , 32
   onnx.NoValue            , 1
   onnx.Not                , 1
   onnx.ReduceMeanV13      , 50
   onnx.Reshape            , 139
   onnx.Sqrt               , 25
   onnx.Sub                , 25
   onnx.Transpose          , 48
  zhigh.Add                , 73
  zhigh.Div                , 8
  zhigh.MatMul             , 97
  zhigh.Mul                , 43
  zhigh.Softmax            , 12
  zhigh.Stick              , 164
  zhigh.StickifiedConstant , 173
  zhigh.Tanh               , 1
  zhigh.Unstick            , 149

```

If I compile dynamically, it compiles fine in a bit under two min:
```
time podman run ... --O3 --EmitLib --mtriple=s390x-ibm-loz --mcpu=z16 --maccel=NNPA roberta-base-11.onnx --onnx-op-stats TXT -v
Operations encountered:
-----------------------
   func.func               , 1
   func.return             , 1
   onnx.Add                , 99
   onnx.Cast               , 2
   onnx.Concat             , 146
   onnx.Constant           , 92
   onnx.ConstantOfShape    , 2
   onnx.CumSum             , 1
   onnx.Dim                , 218
   onnx.Div                , 49
[...]

Issue #1571, opened 2022-08-02
Failed to compile ONNX to .so file: "Assertion `map.getNumInputs() == mapOperands.size() && "inconsistent index info"' failed."


gcunhase:
## How to reproduce error
1. Download onnx file here: https://drive.google.com/file/d/1b1Zq2EJ0KWyCUu3L1XCSDtRdjwYIBq77/view?usp=sharing

2. Install ONNX-MLIR:
```
# Pull the latest docker image:
$ docker pull onnxmlirczar/onnx-mlir

# Test that it's working
$ git clone https://github.com/onnx/onnx-mlir.git
$ cd onnx-mlir/docker
$ python onnx-mlir.py --help
```

3. Compile ONNX file with ONNX-MLIR:
```
$ python onnx-mlir.py --O3 --EmitLib resnet_50v1.onnx
```

## Observed behavior
```
onnx-mlir: /llvm-project/mlir/lib/Dialect/Affine/IR/AffineOps.cpp:2530: static void mlir::AffineLoadOp::build(mlir::OpBuilder&, mlir::OperationState&, mlir::Value, mlir::AffineMap, mlir::ValueRange): Assertion `map.getNumInputs() == mapOperands.size() && "inconsistent index info"' failed.
Unknown error 139
```

## Expected behavior
Generate `.so` compiled file from `.onnx` file.

## System info
- Ubuntu 18.04
- ONNX-MLIR:
```
$ python onnx-mlir.py --version
onnx-mlir version 0.3.0, onnx version 1.11.0 (https://github.com/onnx/onnx-mlir.git 9c6b1ea393685eda3f78ea064fcb358b5091d778)
  LLVM version 15.0.0git (https://github.com/llvm/llvm-project.git f2b94bd7eaa83d853dc7568fac87b1f8bf4ddec6)
  Optimized build with assertions.
  Default target: x86_64-unknown-linux-gnu
  Host CPU: cascadelake
```

gcunhase:
@AlexandreEichenberger after issue https://github.com/onnx/onnx-mlir/issues/1549 was fixed, this issue happened.

PR #2158, opened 2023-04-11
Fix wrong krnl.load when broadcasting with all prefix 1s

tungld:
The lowering of binary ops wrongly generates `krnl.load` in case of broadcasting, which is reported by issues #2154, #2150, #2137.

The issue is that the lowering failed to detect broadcasting in this case:
```mlir
%0 = "onnx.Add"(%arg0, %arg1) : (tensor<1x1x5xf32>, tensor<5xf32>) -> tensor<1x1x5xf32>
```
While computing shape, `tensor<5xf32>` is prepended with `1x1`, which turns out to be similar to the first input `tensor<1x1x5>`, then `hasNoBroadcast` function thinks they are the same and there is no broadcasting.

This patch fixes the issue by keeping the original shape during shape inference, and using the original shape to check broadcasting.


hamptonm1 (approved):
LGTM! Fixes endless loop but I spotted another problem.

hamptonm1:
> I confirmed that Issue#2154 was solved by this PR. Thanks!

@negiyas Did you test using the docker image or did you build onnx-mlir yourself? I use the docker image....so not sure if maybe that is why we have different outcomes. I can also test the model you used to see what I get back.

tungld:
> But our hasNoBroadcast has different meaning. What does hasNoBroadcast really mean to us?

It should be consistent with numpy's spec finally I think.

chentong319:
> 
> It should be consistent with numpy's spec finally I think.

The bumpy's spec says when the two inputs are `multiple directional broadcastable`, namely the requirement for the inputs.  Look at the first condition, for example, `The tensors all have exactly the same shape.`. I do think we want `hasNoBroadcast()` returns true for this case.
we do not to check braodcastable if the model is assumed to be correct. What we need from hasNoBroadcast() is related to how we generate the load. For example, we may implement the load without `hasNoBroacast()`. The loop can be generated as the current implementation. The affine map from iteration vector is to 1) mask out the none existing dimensions and 2) replace the iteration variable with constant 1 if the size of that dimension is 1. For the example the Roberta error, for tensor<768xf32>, the map will be `#map<(d0, d1, d2)[] = (d2)>'. For case of tensor<2x3x4xf32> and tensor<2x1x4xf32>, the map for the tensors is 
`#map<(d0, d1, d2)[] = (d0, d1, d2)>`  and '#map<(d0, d1, d2)[] = (d0, 1, d2)>'
Again, I think we need to clearly define the meaning of `hasNoBroadcast()` while they are all `multiple directional broadcastable`.

tungld:
> The tensors all have exactly the same shape.. I do think we want hasNoBroadcast() returns true for this case.

Do you mean the current `hasNoBroadcast()` returns `false` for this case? and it is a bug and it should return `true`.

@AlexandreEichenberger any thought? I think you wrote the function for SIMD.

AlexandreEichenberger:
I am modifying this very code to get partial SIMD execution. What is the pattern that causes the issue? This one?

```
  %0 = "onnx.Add"(%arg0, %arg1) : (tensor<1x1x5xf32>, tensor<5xf32>) -> tensor<1x1x5xf32>
```

tungld:
> What is the pattern that causes the issue? This one?
> %0 = "onnx.Add"(%arg0, %arg1) : (tensor<1x1x5xf32>, tensor<5xf32>) -> tensor<1x1x5xf32>

Yes, it is.

AlexandreEichenberger:
From a code pattern perspective, `1x1x5` and `5` have no broadcasting, in the sense that there are 5 values loaded for the first argument, and 5 loaded for the second, and each one is paired with each other. 

While simd code will not be generated for such a short sequence, say it was `1x1x1024` and `1024`, we should and we will.

Please @negiyas give me the test case you want me to investigate (e.g. the original benchmark with the specific commands to force arguments to 1) and I will make sure they work.

AlexandreEichenberger (changes requested):
I would like for this PR not to go through, as the issue is not with the broadcasting per se, but with the code we generate for that pattern. 

Please give me the test cases you want me to verify, and I will make sure we gen the right code.

Tx

chentong319:
The original test case is Roberta. I think a simple test case should be OK.

chentong319:
> From a code pattern perspective, `1x1x5` and `5` have no broadcasting, in the sense that there are 5 values loaded for the first argument, and 5 loaded for the second, and each one is paired with each other.
> 
> While simd code will not be generated for such a short sequence, say it was `1x1x1024` and `1024`, we should and we will.
> 
The error is related to how to generate the reference. As I said in my previous comments, the meaning of hasNoBroadcast is not clear. From the discussion, I found that hasNoBroadcast serves two purposes in our code: 1) check the reference is contiguous for SIMD, and 2) how to generate the reference.  They are not the same.  I made comments on how to generate reference in my previous comments.

tungld:
@AlexandreEichenberger this is the test case (also in the lit test of this patch):
```
func.func @add_1_1(%arg0 : tensor<1x1x5xf32>, %arg1: tensor<5xf32>) -> tensor<1x1x5xf32> {
 %0 = "onnx.Add"(%arg0, %arg1) : (tensor<1x1x5xf32>, tensor<5xf32>) -> tensor<1x1x5xf32>
 return %0 : tensor<1x1x5xf32>
}
```
When you run `onnx-mlir-opt --convert-onnx-to-krnl --convert-krnl-to-affine`, you will get error with the main branch.

AlexandreEichenberger:
> Totally agree. They are not the same, and I think this hasNoBroadcast is now mainly used for SIMD but it is also passed to shapeHelper.getAccessExprs to compute reference indices, which causes confusion if its semantics is vague.

Agreed, they need to work in sink. Fixing both of them at this time. The solution I am after is a bit more complicated yet as it handles partial simd, as I am collapsing subsets of inner loops and their memory accesses.

To give you a taste, `?x?x4x5` and `?x4x5` can have a collapse of inner loop of 2, to become `?x?x20" and `?x20` where we handle possible broadcast in, respectively, the outer 2 and 1 dims. But `?x?x4x5` and `1` becomes `?` and `1` as the single scalar is splatted with any (regardless of how many) dynamic dimensions...

hamptonm1:
> The original test case is Roberta. I think a simple test case should be OK.

I was testing using Roberta-base-11 model but this also impacts resnet_50v1. @tungld and @AlexandreEichenberger please see this issue as well: https://github.com/onnx/onnx-mlir/issues/1571. Issue 1571 has been open since August of last year with the exact same error message.

AlexandreEichenberger:
#2161 fixes the issues listed here; please double check that it works for you as well.

tungld:
@AlexandreEichenberger could you add the lit test here into #2161 to make sure it works?

By the way, in #2161, `hasNoBroadcasting` is still confusing, I don't see how it solves the semantics issue @chentong319 mentioned here. For example, `hasNoBroadcasting` still returns `true` for the case `%0 = "onnx.Add"(%arg0, %arg1) : (tensor<1x1x5xf32>, tensor<5xf32>) -> tensor<1x1x5xf32>`, doesn't it?

hamptonm1:
> #2161 fixes the issues listed here; please double check that it works for you as well.

Okay I can test it now

AlexandreEichenberger:
Using code in PR #2161 

```mlir
func.func @add_1_1(%arg0 : tensor<1x1x5xf32>, %arg1: tensor<5xf32>) -> tensor<1x1x5xf32> {
 %0 = "onnx.Add"(%arg0, %arg1) : (tensor<1x1x5xf32>, tensor<5xf32>) -> tensor<1x1x5xf32>
 return %0 : tensor<1x1x5xf32>
}
```

get lowered to

```mlir
  func.func @add_1_1(%arg0: memref<1x1x5xf32>, %arg1: memref<5xf32>) -> memref<1x1x5xf32> attributes {llvm.emit_c_interface} {
    %alloc = memref.alloc() {alignment = 16 : i64} : memref<1x1x5xf32>
    affine.for %arg2 = 0 to 1 {
      affine.for %arg3 = 0 to 1 {
        affine.for %arg4 = 0 to 5 {
          %0 = affine.load %arg0[0, 0, %arg4] : memref<1x1x5xf32>
          %1 = affine.load %arg1[%arg4] : memref<5xf32>
          %2 = arith.addf %0, %1 : f32
          affine.store %2, %alloc[%arg2, %arg3, %arg4] : memref<1x1x5xf32>
        }
      }
    }
    return %alloc : memref<1x1x5xf32>
  }
```

you see the 
* 3 dim load: ` %0 = affine.load %arg0[0, 0, %arg4] : memref<1x1x5xf32>` where `0` are directly inserted (new pattern)
* 1 dim load: `%1 = affine.load %arg1[%arg4] : memref<5xf32>`

I think that is exactly what is expected. Same example run in SIMD too.

There are several lit tests checking for this pattern in the new PR too.

AlexandreEichenberger:
I also need to set the record straight about broadcasting and `hasNoBroadcast()` function. 

To recap, here is what numpy say about broadcasting:

---
###General Broadcasting Rules
When operating on two arrays, NumPy compares their shapes element-wise. It starts with the trailing (i.e. rightmost) dimension and works its way left. Two dimensions are compatible when

* they are equal, or
* one of them is 1.

If these conditions are not met, a `ValueError: operands could not be broadcast together exception is thrown`, indicating that the arrays have incompatible shapes.

Input arrays do not need to have the same number of dimensions. The resulting array will have the same number of dimensions as the input array with the greatest number of dimensions, where the size of each dimension is the largest size of the corresponding dimension among the input arrays. Note that missing dimensions are assumed to have size one.
---
https://numpy.org/doc/stable/user/basics.broadcasting.html#general-broadcasting-rules

ONNX itself fully agrees with NumPy

---
In ONNX, a set of tensors are multidirectional broadcastable to the same shape if one of the following is true:

1. The tensors all have exactly the same shape.
2. The tensors all have the same number of dimensions and the length of each dimensions is either a common length or 1.
3. The tensors that have too few dimensions can have their shapes prepended with a dimension of length 1 to satisfy property 2.
---
https://github.com/onnx/onnx/blob/main/docs/Broadcasting.md

See rule 3, which will prepend `1` before trying to satisfy rules 1 or 2 for dimensions that are too short.

The `hasNoBroadcast()` is perfectly in line with both definitions.

Incidentally, the bug came from the `getAccessFunction` which took a shortcut when `hasNoBroadcast()` was true. The new PR #2161 fixes that (and some more).

tungld:
I think all confusion comes from the term `broadcasting` and `broadcastable`. We think there is `broadcasting` when we broadcast a single value to multiple values. It means `tensor<3x5xf32>` and `tensor<3x5xf32` is not broadcasting.

But ONNX defines `broadcastable`:

In ONNX, a set of tensors are multidirectional broadcastable to the same shape if one of the following is true:

- The tensors all have exactly the same shape.
- The tensors all have the same number of dimensions and the length of each dimensions is either a common length or 1.
- The tensors that have too few dimensions can have their shapes prepended with a dimension of length 1 to satisfy property 2.

Looking Rule 1, `tensor<3x5xf32` and `tensor<3x5xf32>` are broadcastable but for use there is no broadcasting.

tungld:
Closing this since #2161 can fix the model.

AlexandreEichenberger:
This should have been fixed by #2161.

PR #2159, opened 2023-04-12
FuseAddConvPattern: skip if returnType is unranked

sorenlassen:
instead of crashing in (RankXMinusRankYIs<1> $res, $y)

Added lit test test_fuse_add_conv_bias for FuseAddConvPattern by adapting the existing lit test test_fuse_add_conv for FuseAddConvNullBiasPattern, and then added unranked variants of the lit tests which crashed without the fixes to FuseAddConvPattern and FuseAddConvNullBiasPattern in Rewrite.td.

This problem was discovered in PR #2098.


chentong319:
This issue is exposed by hybrid pass because we do not have control over the order of patterns to be applied to an op. Is it true? Can we control the order? Will the order of adding rules to the pattern matter?

chentong319 (approved):
LGTM!

PR #2160, opened 2023-04-13
Support code generation for onnx.Erf

negiyas:
Support code generation for onnx.Erf based on approximate function described at https://www.johndcook.com/blog/2009/01/19/stand-alone-error-function-erf/ . Simdized code generation is also supported.

This PR includes the conversion code, its test, and its numerical test.
The following graph shows the difference between onxx.Erf function values by the main branch (using erff in libm for zLinux) and this PR. (c.f. The blue line for pr2160 is overlapped by the orange line for main branch) The numerical difference is 2.24E-7 at maximum. 

I hope that this PR can be a baseline for optimizing onnx.Erf function.

image


negiyas:

negiyas:

imaihal:
I applied this PR and measured performance of roberta-base-11. The onnx.ERF became around 2 times faster in  input 8x256 (batch size x sequence_len).

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -769,6 +763,88 @@ Value emitScalarOpFor<ONNXSignOp>(ConversionPatternRewriter &rewriter,
...
+  Value minusone = create.math.constant(elementType, -1);
+  Value a1 = create.math.constant(elementType, 0.254829592);
+  Value a2 = create.math.constant(elementType, -0.284496736);
+  Value a3 = create.math.constant(elementType, 1.421413741);
+  Value a4 = create.math.constant(elementType, -1.453152027);
+  Value a5 = create.math.constant(elementType, 1.061405429);
+  Value p = create.math.constant(elementType, 0.3275911);
+  Value absx = create.math.abs(operand);
+  Value t =
+      create.math.div(one, create.math.add(one, create.math.mul(p, absx)));
```
AlexandreEichenberger, line 820: can use FMA for add/mul pattern?

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -769,6 +763,88 @@ Value emitScalarOpFor<ONNXSignOp>(ConversionPatternRewriter &rewriter,
...
+  //   y = 1.0 -
+  //       (((((a5 * t + a4) * t) + a3) * t + a2) * t + a1) * t *
+  //       exp(-absx*absx)
+  Value y = create.math.sub(
+      one, create.math.mul(
```
AlexandreEichenberger, line 825: please use fma.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -769,6 +763,88 @@ Value emitScalarOpFor<ONNXSignOp>(ConversionPatternRewriter &rewriter,
...
+                                               create.math.mul(a5, t), a4),
+                                           t),
+                                       a3),
+                                   t),
+                               a2),
+                           t),
+                       a1),
+                   t),
+               create.math.exp(
+                   create.math.mul(create.math.mul(minusone, absx), absx))));
```
AlexandreEichenberger, line 843: I think the backend will optimize this `-1 * absx` as `-0-absx`.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -769,6 +763,88 @@ Value emitScalarOpFor<ONNXSignOp>(ConversionPatternRewriter &rewriter,
...
+};
+
+template <>
+double analyzeSimdFor<ONNXErfOp>(Type t, int64_t &von, int64_t &son) {
+  return simdAnalysis(
+      {GenericOps::ArithmeticGop, GenericOps::CompareGop, GenericOps::DivGop,
+          GenericOps::MulGop, GenericOps::SelectGop, GenericOps::AbsGop,
+          GenericOps::ExpGop},
+      {6, 2, 1, 11, 2, 1, 1}, t, von, son);
+}
```
AlexandreEichenberger, line 782: I would use FMAs in the code and here:

arithmetic 2
compare 1
div 1
mul 4
FMA 5
select 1
abs 1
exp 1

AlexandreEichenberger (commented):
minor changes, mainly requesting the use of FMA, which I am pretty sure the backend will generate, but its always best to generate as good as we can, especially when its easy :-)

Also, I am trying to get an expert to provide us insight on the algo used here for ERF, for example if we could use a lower precision version when used in conjunction with accelerators that uses a lower precision than fp32 to begin with.

Eventually, we should also look into a SIMD version of `exp` which is used here, and for which I am not sure we have a SIMD approximation.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
negiyas: compare 1 arith.cmpf
     dievp 1 arith.divf
      6 vector.fma
      4 arith.mulf
      1 arith.select
      1 math.absf
      1 math.exp

negiyas:
> minor changes, mainly requesting the use of FMA, which I am pretty sure the backend will generate, but its always best to generate as good as we can, especially when its easy :-)

I hope that I have fixed the code to use fma as much as possible, but please let me know any issues.
 
> Also, I am trying to get an expert to provide us insight on the algo used here for ERF, for example if we could use a lower precision version when used in conjunction with accelerators that uses a lower precision than fp32 to begin with. 
> Eventually, we should also look into a SIMD version of `exp` which is used here, and for which I am not sure we have a SIMD approximation.

Please ask! This PR uses fifth-order Taylor expansion (using exp) as the approximation function, but it may be too much.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
negiyas: done. Thanks.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
negiyas: done. Thanks.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
negiyas: I modified a little bit to use fma for the exp result.

src/Dialect/Mlir/DialectBuilder.cpp
```diff
@@ -205,6 +205,15 @@ Value MathBuilder::floorDiv(Value lhs, Value rhs) const {
   llvm_unreachable("expected int");
 }
 
+// return (lhs * rhs) + acc
+Value MathBuilder::fma(Value lhs, Value rhs, Value acc) const {
+  assert((lhs.getType() == rhs.getType()) && (rhs.getType() == acc.getType()) &&
+         "expected same type");
+  if (lhs.getType().isa<VectorType>())
```
tungld, line 212: Thanks for adding this! `FMAOp` is for floating-point only, so element type should be checked.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -772,6 +766,79 @@ Value emitScalarOpFor<ONNXSignOp>(ConversionPatternRewriter &rewriter,
...
+  //            exp(-absx*absx)
+  //           + (-1.0)
+  Value minusy = create.math.fma(
+      create.math.mul(
+          create.math.fma(
+              create.math.fma(
+                  create.math.fma(create.math.fma(a5, t, a4), t, a3), t, a2),
+              t, a1),
+          t),
+      create.math.exp(create.math.mul(create.math.mul(minusone, absx), absx)),
```
tungld, line 836: `-1 * abs(x)` can be `0 - abs(x)` as @AlexandreEichenberger suggested.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -772,6 +766,79 @@ Value emitScalarOpFor<ONNXSignOp>(ConversionPatternRewriter &rewriter,
...
+  using IOp = NotSuportedScalarOp;
+};
+
+template <>
+double analyzeSimdFor<ONNXErfOp>(Type t, int64_t &von, int64_t &son) {
+  return simdAnalysis(
+      {GenericOps::CompareGop, GenericOps::DivGop, GenericOps::FmaGop,
+          GenericOps::MulGop, GenericOps::SelectGop, GenericOps::AbsGop,
+          GenericOps::ExpGop},
+      {1, 1, 6, 4, 1, 1, 1}, t, von, son);
```
tungld, line 784: `Fma` will be `Mul` and `Add` in case of IntegerType. How do we count them in such a situation?

@AlexandreEichenberger any though.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
negiyas: Two mul are change to arithmetic ops (by Fix  with (-1.0 * x) to (0.0 -x)).

arithmetic 2 arith.subf
compare 1 arith.cmpf
div 1 arith.divf
fma 6 vector.fma
mul 2 arith.mulf
select 1 arith.select
abs 1 math.absf
exp 1 math.exp

src/Dialect/Mlir/DialectBuilder.cpp
negiyas: Added type check logic. Thanks.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
negiyas: I changed! Thanks.

src/Dialect/Mlir/DialectBuilder.cpp
tungld: `isFloatWithVector` is enough.

```c
if (isFloatWithVector(lhs.getType()))
  return b().create<vector::FMAOp>(loc(), lhs, rhs, acc);
return add(mul(lhs, rhs), acc);
```

negiyas: Fixed with small modification. Thanks!

negiyas:

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
negiyas: @AlexandreEichenberger Thanks for the answer. 
I understand that we can simply count the operation number without considering the target type is integer or float.
Please let me know if my understanding is not correct.

AlexandreEichenberger (approved):
LGTM big thanks.

negiyas:

PR #2161, opened 2023-04-13
Elementwise simd

AlexandreEichenberger:
Implement a new method `hasManageableBroadcast` that returns if there is any partial SIMD that can be exploited. Partial starts from the innermost dimension, looks if there is manageable simd, if there is, it then check the next innermost dim, until it runs out of dimensions (can be fully simdized) or it finds a condition that is not manageable.

The big picture is that manageable broadcast is either a scalar that is broadcasted to any other dims, or no broadcast that can be demonstrated at compile time.

Some of the patterns that are partially simdizable are listed below:
* (?,?, 768) and (?, ?, 768): manageable at level 1 (innermost)
* (?,?,768) and (768): manageable at level 1 (innermost)
* (?, ?, 768) and (1): manageable at level 3 (full)
* (?,?, 3, 256) and (?, ?, 3, 256): manageable at level 2 (innermost and next innermost)
* (4, 3, 1, 2) and (?, 1, 1, 1): manageable at level 3 (all but outermost unless dim analysis can show ? to be 1)

Other changes.
* DimAnalysis: allows now negative indices to determine which dimension to analyze in `SameDim` and `SameDynDim`; negative is offset from the innermost dim. When out-of-bound, simply return false.
* Bug fixes for handling of '1' dims.
* There were times where the old algorithm generated trivial reshape that were not needed; removed.
* Code pattern change: when generating broadcast operations access functions, for dim of size 1, the code now simply generate `0` constant as opposed to a reference to the loop index that iterates from `0` inclusive to `1` exclusive.


src/Dialect/ONNX/ONNXOps/ShapeHelper.cpp
```diff
@@ -378,49 +380,256 @@ bool ONNXBroadcastOpShapeHelper::hasNoBroadcast(DimAnalysis *dimAnalysis) {
...
+  // The ruledOutBroadcast pattern can be established by shape inference using
   // DimAnalysis. If that is available, and broadcasting was ruled out, then
   // more efficient code can be generated.
-  if (hasNoBroadcast || (hasUniBroadcasting && operandIndex == 0)) {
-    for (IndexExpr ie : outputAccessExprs)
-      operandAccessExprs.emplace_back(ie);
-    return success();
-  }
+  bool noBroadcasting =
+      ruledOutBroadcast || (hasUniBroadcasting && operandIndex == 0);
```
tungld, line 576: It looks to me `ruledOutBroadcast` is the same as the previous `hasNoBroadcast`, and the semantics of `hasNoBroadcast` is unchanged in this patch and still vague.

src/Dialect/ONNX/ONNXOps/ShapeHelper.cpp
```diff
@@ -378,49 +380,256 @@ bool ONNXBroadcastOpShapeHelper::hasNoBroadcast(DimAnalysis *dimAnalysis) {
...
-      IndexExpr dim = inputsDims[i][dimIndex];
-      if (!dim.isLiteralAndIdenticalTo(1)) {
-        allOtherInputDimsAreOne = false;
-        break;
+    // Take loopDepth instead of outputRank as loopDepth reflect the
+    // (possible) flattening of the loops.
+    int64_t dimIndex = loopDepth - operandRank + r;
+    SymbolIndexExpr operandDim(inputsDims[operandIndex][dimIndex]);
+
+    bool guaranteedBroadcasting = false;
```
tungld, line 585: Now we have another one about broadcasting, which makes me more confused ... Is this additional `guaranteedBroadcasting` necessary for SIMD?

src/Dialect/ONNX/ONNXOps/ShapeHelper.cpp
```diff
@@ -327,13 +330,12 @@ LogicalResult ONNXBroadcastOpShapeHelper::customComputeShape(
   return success();
 }
 
+// Attempt to rule out broadcasting at compile time, using dim analysis when
+// available (i.e. nonnull). Must be called after computeShape.
 bool ONNXBroadcastOpShapeHelper::hasNoBroadcast(DimAnalysis *dimAnalysis) {
```
tungld, line 335: @AlexandreEichenberger could you add the lit test from #2158 into this patch to make sure it works?

By the way, `hasNoBroadcasting` is still confusing, I don't see how it solves the semantics issue @chentong319 mentioned in #2158. For example, `hasNoBroadcasting` still returns `true` for the case `%0 = "onnx.Add"(%arg0, %arg1) : (tensor<1x1x5xf32>, tensor<5xf32>) -> tensor<1x1x5xf32>`, doesn't it?

docs/Testing.md
```diff
@@ -380,6 +382,8 @@ We provide a Python script [RunONNXModelZoo.py](../utils/RunONNXModelZoo.py) to
 ```bash
 $ git clone https://github.com/onnx/models
 $ cd models
-$ ONNX_MLIR_HOME=/onnx-mlir/build/Release/ /onnx-mlir/utils/RunONNXModelZoo.py -m mnist-8 -compile-args="-O3"
+$ ONNX_MLIR_HOME=/onnx-mlir/build/Release/ /onnx-mlir/utils/RunONNXModelZoo.py -m mnist-8 -c "-O3"
 ```
-Run the script with `-h` to see all the options.
+Run the script with `-h` to see all the options. In addition to the `-m` flag to specify a model and `-c` flag to specify the compile options, useful options are the `-k` flag to leave the onnx model in the current directory as a `.tgz` file, and the `-l debug` flag to print lots of debugging info.
+
+If you are unsure of the names available models, you can always fire a `find . -name "*.onnx" -print` command to find which ones are available.
```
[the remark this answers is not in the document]
AlexandreEichenberger: great, much better than my method

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -12,11 +12,13 @@
...
 //===----------------------------------------------------------------------===//
 
+#include "llvm/Support/Debug.h"
+
 #include "src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.hpp"
 #include "src/Dialect/Krnl/DialectBuilder.hpp"
 #include "src/Dialect/ONNX/ONNXOps/ShapeHelper.hpp"
 
-#define DEBUG 0 /* Log which functions are simdized. */
+#define DEBUG_TYPE "convert-onnx-to-krnl-simd"
```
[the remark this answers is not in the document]
AlexandreEichenberger: ok

src/Dialect/ONNX/ONNXOps/ShapeHelper.cpp
```diff
@@ -378,49 +380,256 @@ bool ONNXBroadcastOpShapeHelper::hasNoBroadcast(DimAnalysis *dimAnalysis) {
...
+    bool ruledOutBroadcast) {
+  // Get info.
+  int64_t loopDepth = loopAccessExprs.size();
+  int64_t inputSize = inputsDims.size();
+  int64_t operandRank = operand.getType().cast<ShapedType>().getRank();
+  // Flattened? no more than one loop per dim in output (aka output rank).
+  // Not flattened? one loop per dim in output (aka output rank).
+  if (flattenedInnerDims)
+    assert(loopDepth <= (int64_t)outputRank &&
+           "without flattening, expect one loop iter variable per output rank");
```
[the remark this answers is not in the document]
AlexandreEichenberger: copied and pasted without editing the error message, thanks for catching it

src/Dialect/ONNX/ONNXOps/ShapeHelper.cpp
AlexandreEichenberger: I added an example in the code. Repeating here 
```
      // For example we have 5x?x3xf32 and 5x1x3xf32. We don't know at compile
      // time if the `?` will be 1 or (assuming without loss of generality) 10.
      // Normal code is to generate for `?` for that access dimension:
      //   select(dim==1 ? 0 : loop-var)
      // so that if this access needs to be broadcasted, we will only access the
      // `0` value of this broadcasted value; and if it is not broadcasted, we
      // will access each `loop-var` value for this access.
      //
      // But since all of the other dimensions are '1', they cannot generate
      // a broadcasting situation. Thus, for this access function, we can
      // generate the access `loop-var` regardless of whether the `?` will
      // evaluate to 1 or 10 at runtime. It will be correct no matter which
      // situation we will be in.
```

AlexandreEichenberger: I can see how the name is confusing... let me rename it

src/Dialect/ONNX/ONNXOps/ShapeHelper.cpp
AlexandreEichenberger: There is a lit test that check for that specific pattern, and make sure that it get simdized as it detects no broadcast.

AlexandreEichenberger:
Thanks for the thorough comments @tungld . Believe I responded to all of them.

src/Dialect/ONNX/ONNXOps/ShapeHelper.hpp
```diff
@@ -221,17 +221,79 @@ struct ONNXBroadcastOpShapeHelper : public ONNXOpShapeHelper {
...
+  // Examples with broadcasts:
+  // * 2x5xf32 and 1x5xf32 has broadcast as the second input's value are used
+  //   for each of the two instances of 5xf32 in the first input.
+  // * Same holds for 2x5xf32 and 5xf32 as shorter ranked
+  //   inputs are extended by prepending 1x.
+  //
+  // Example without broadcast:
+  // * 3x5xf32 and 3x5xf32 have no broadcast.
+  // * 1x5xf32 and 5xf32 have also no broadcast as prepending 1x results as
+  //   comparing 1x5xf32 with 1x5xf32.
```
tungld, line 251: Thanks for the explanation! It's very helpful.

tungld (approved):
LGTM. Thanks for the great work!

hamptonm1:
@AlexandreEichenberger I tested and I am seeing the same behavior you do which is the loop. The loop has nothing to do with what my assumption of what is causing it but your code fixes the other issue and the backend test pass. So this looks good to me.

src/Dialect/ONNX/ONNXOps/ShapeHelper.cpp
AlexandreEichenberger: Changed the names to be a bit clearer
```C++
  bool noBroadcasting =
      hasNoBroadcast || (hasUniBroadcasting && operandIndex == 0);
```

PR #2163, opened 2023-04-14
OperationFingerPrint instead of printing to sha1_ostream

### LINKED ISSUE #1963: improve createTagForIR() convergence test in ONNXOpTransformPass
PSA: **run onnx-mlir with `--mlir-elide-elementsattrs-if-larger=1`** on performance sensitive workloads, without it the frontend is slow and wastes memory in the ONNXOpTransformPass

createTagForIR() computes the tags in the `currentTag != previousTag` convergence test in ONNXOpTransformPass by hashing the mlir assembly textual printout

see the NOTE and TODO in ONNXOpTransformPass.cpp:
```
    // NOTE: This is slow for real models because they contain large
    // constant tensors that are expensive to print. A workaround is to
    // elide them from printing with --mlir-elide-elementsattrs-if-larger=1
    //
    // TODO: Hash without printing to speed this up, e.g. along the lines of
    // how blocks are hashed in mlir/lib/Transforms/Utils/RegionUtils.cpp
```

eliding ElementsAttr from printing with `--mlir-elide-elementsattrs-if-larger=1` has an extra significance after the introduction of DisposableElementsAttr in PR #1874 because we also waste memory in the current `DisposableElementsAttr::printAsDenseElementsAttr()` implementation which is used in ONNXConstantOp::print (see TODO in DisposableElementsAttr.cpp)

we should also fix the speed and memory waste in `DisposableElementsAttr::printAsDenseElementsAttr()`, I'll create a separate issue for that

sorenlassen:
fixes #1963

Tested on the arcfaceresnet100-8.onnx model. On my MacBook when I run:
```
RelWithDebInfo/bin/onnx-mlir arcfaceresnet100-8.onnx --mlir-timing
```
ONNXOpTransformPass takes 2.7387s before the change and 0.0032s after the change.

### DIFF
--- a/src/Transform/ONNX/ONNXOpTransformPass.cpp
+++ b/src/Transform/ONNX/ONNXOpTransformPass.cpp
@@ -13,9 +13,9 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include "mlir/IR/OperationSupport.h"
 #include "mlir/Pass/PassManager.h"
 #include "mlir/Transforms/Passes.h"
-#include "llvm/Support/raw_sha1_ostream.h"
 
 #include "src/Dialect/ONNX/ONNXOps.hpp"
 #include "src/Pass/Passes.hpp"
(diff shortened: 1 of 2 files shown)


src/Transform/ONNX/ONNXOpTransformPass.cpp
```diff
@@ -101,18 +86,20 @@ void ONNXOpTransformPass::runOnOperation() {
         onnx_mlir::createConstPropONNXToONNXPass());
     if (failed(runPipeline(dynamicPM, module)))
       return signalPassFailure();
-    currentTag = createTagForIR(module);
-  } while (currentTag != previousTag && --n > 0);
-  if (currentTag != previousTag) {
+    OperationFingerPrint after(module);
```
tungld, line 89: Cool, it's great to know MLIR has a class to support this kind of hash.

tungld (approved):
LGTM. 

Would like to hear comments from @chentong319 as he originally wrote the code.

chentong319 (approved):
LGTM!

PR #2167, opened 2023-04-14
constant propagate in lowering_with_canonicalize lit tests

sorenlassen:
The lowering_with_canonicalize lit tests lowered test_if_sign differently than the real compiler because it skipped constant propagation which hoists constants out of the if and else branches and somehow leads to different lowered code.


sorenlassen:
abandoning this PR in favor of the simpler solution in PR #2168

PR #2168, opened 2023-04-14
hoist constants in test_if_sign lit test

sorenlassen:
The lowering_with_canonicalize lit tests lowered test_if_sign differently than the real compiler because it skipped constant propagation which hoists constants out of the if and else branches and somehow leads to different lowered code.

Now the constants are hoisted in the lit test just like constant propagation would do.

This was discovered in PR #2098 where the new pattern-based shape inference hoists constants like propagation does.

### DIFF
--- a/test/mlir/onnx/onnx_lowering_with_canonicalize.mlir
+++ b/test/mlir/onnx/onnx_lowering_with_canonicalize.mlir
@@ -155,21 +155,21 @@ func.func @test_sequence_insert(%arg0: tensor<?x4x5xf32>, %arg1:tensor<3x4x5xf32
 
 // Check nested if lowering (function computes scalar Sign).
 func.func @test_if_sign(%arg0: tensor<f32>) -> tensor<i32> {
+  %zero = onnx.Constant {value = dense<0> : tensor<i32>} : tensor<i32>
+  %plus = onnx.Constant {value = dense<1> : tensor<i32>} : tensor<i32>
+  %minus = onnx.Constant {value = dense<-1> : tensor<i32>} : tensor<i32>
   %0 = onnx.Constant {value = dense<0.0> : tensor<f32>} : tensor<f32>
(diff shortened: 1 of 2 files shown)


chentong319:
The constant is hoisted to the beginning of functions by constant folding. Could you paste the result of PR2098? We always have the question: constant folding is invoked automatically by canonicalization pass. But when we add the canonicalization rules to our pattern to the hybrid transformation, will constant folding be invoked? From this example, it seems not.
This PR is just hide the difference. Perhaps, our expectation is wrong.

sorenlassen:
> The constant is hoisted to the beginning of functions by constant folding. Could you paste the result of PR2098? We always have the question: constant folding is invoked automatically by canonicalization pass. But when we add the canonicalization rules to our pattern to the hybrid transformation, will constant folding be invoked? From this example, it seems not.
> This PR is just hide the difference. Perhaps, our expectation is wrong.

I believe the hoisting of constants is a side effect of using `applyPatternsAndFoldGreedily` to drive the constant propagation pass in ConstPropONNXToONNXPass::runOnOperation(). I also use `applyPatternsAndFoldGreedily` to drive shape inference in ONNXHybridTransformPass::runOnOperation() in PR #2098 and therefore it happens with the new shape inference in that PR and you can see the resulting lit test changes in PR #2098.

With the fix to the lit test in this PR this difference between the old shape inference and the new pattern-based shape inference in PR #2098 no longer shows up in the lowering_with_canonicalize lit tests.

sorenlassen:
> This PR is just hide the difference. Perhaps, our expectation is wrong.

Note that the lowering_with_canonicalize lit tests are unrealistic: when we run the compiler in real life we always perform constant propagation and therefore hoist the constants before lowering to Krnl. (See PR #2167)

chentong319 (approved):
LGTM!

PR #2169, opened 2023-04-15
attempt to fix Jenkins Linux ppc64le warning

### LINKED ISSUE #2027: Warning as error
New issue that continues #936 .

I believe the warning due to the unused parameter in LLVM header files (isPossible) is now fixed in LLVM:
https://reviews.llvm.org/rG82ae83a8487df0747195d1abbda11920fda11eb8

However there remain some other unused parameters warnings (but much much fewer).

```
/workdir/llvm-project/mlir/include/mlir/IR/AttrTypeSubElements.h:412:37: warning: parameter 'derived' set but not used [-Wunused-but-set-parameter]
  412 | void walkImmediateSubElementsImpl(T derived,
      |                                   ~~^~~~~~~
```

LLVM indicates that it supports our gcc level
https://llvm.org/docs/GettingStarted.html#host-c-toolchain-both-compiler-and-standard-library
but obviously their tests may not test our specific compiler and/or do not flag warning as error.

@sstamenova suggested that we add a builedbot to mlir to force MLIR developers to address these issues upfront.
@gongsu832 was going to look into this.

sorenlassen:
Fixes #2027

This works around this Linux compiler warning:
```
/workdir/onnx-mlir/src/Transform/ONNX/Decompose.cpp: In copy constructor '{anonymous}::DecomposeONNXToONNXPass::DecomposeONNXToONNXPass(const {anonymous}::DecomposeONNXToONNXPass&)':
/workdir/onnx-mlir/src/Transform/ONNX/Decompose.cpp:736:25: warning: implicitly-declared 'mlir::Pass::Option<std::__cxx11::basic_string<char> >& mlir::Pass::Option<std::__cxx11::basic_string<char> >::operator=(const mlir::Pass::Option<std::__cxx11::basic_string<char> >&)' is deprecated [-Wdeprecated-copy]
  736 |     this->target = pass.target;
In file included from /workdir/llvm-project/mlir/include/mlir/Pass/PassRegistry.h:17,
                 from /workdir/llvm-project/mlir/include/mlir/Pass/Pass.h:13,
                 from /workdir/onnx-mlir/src/Transform/ONNX/Decompose.cpp:25:
/workdir/llvm-project/mlir/include/mlir/Pass/PassOptions.h:176:13: note: because 'mlir::Pass::Option<std::__cxx11::basic_string<char> >' has user-provided 'mlir::detail::PassOptions::
Option<DataType, OptionParser>& mlir::detail::PassOptions::Option<DataType, OptionParser>::operator=(const mlir::detail::PassOptions::Option<DataType, OptionParser>&) [with DataType =
 std::__cxx11::basic_string<char>; OptionParser = llvm::cl::parser<std::__cxx11::basic_string<char> >]'
  176 |     Option &operator=(const Option &other) {
```

It looks like the compiler warns that it uses an implicit copy constructor and not the specified one:
https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Pass/Pass.h#L100

I don't understand why, but if it doesn't use the intended `operator=` then that might be a real issue and the workaround in this PR should fix it.


sorenlassen:
I'm not seeing this warning in the logs anymore, so this workaround should no longer be needed

AlexandreEichenberger:
I still see it when I compile in my virtual env... maybe my compiler is too old.
I like your solution too @sorenlassen; do you want to use yours?

sorenlassen:
@AlexandreEichenberger yes, if you don't mind, let's go with my solution, please approve

AlexandreEichenberger:

AlexandreEichenberger (approved):
LGTM passing tests

PR #2172, opened 2023-04-17
Missing separator in a lit test

tungld:


sorenlassen (approved):
good catch!

AlexandreEichenberger:
note that you can feed a lit test file to `utils/fixLitTest.py` and it will complain if there are 2 functions not separated by a `// ---` pattern if you suspect this.

PR #2173, opened 2023-04-17
Add a case about Reshape to DimAnalysis

tungld:
This patch improves DimAnalysis to handle a case about Reshape that was found in the T5 model in onnx model zoo.


AlexandreEichenberger (approved):
Cool new pattern, LGTM

PR #2175, opened 2023-04-17
factor shape inference patterns into a library

sorenlassen:
And added flag `--pattern-shape-inference` to control whether to use new pattern based shape inference.


src/Compiler/CompilerOptions.cpp
```diff
@@ -208,6 +208,11 @@ llvm::cl::opt<bool> enableSimdDataLayout("simd-data-layout",
                    "Set to 'true' if you want to enable SIMD optimizations."),
     llvm::cl::init(false), llvm::cl::cat(OnnxMlirOptions));
 
+llvm::cl::opt<bool> enablePatternShapeInference("pattern-shape-inference",
+    llvm::cl::desc("Enable pattern based shape inference (default=false)\n"
+                   "Set to 'true' to enable."),
+    llvm::cl::init(false), llvm::cl::cat(OnnxMlirCommonOptions));
```
sorenlassen, line 214: I tried `init(true)` and verified that `cmake --build . --target check-onnx-lit` passed

src/Transform/ONNX/ShapeInferencePass.cpp
```diff
@@ -60,42 +51,41 @@ static SmallVector<func::FuncOp, 4> lookUpFuncsMatching(
  * purposes.
  */
 class ShapeInferencePass
-    : public PassWrapper<ShapeInferencePass, OperationPass<ModuleOp>> {
-private:
-  bool analyzeAllFunctions;
-
+    : public PassWrapper<ShapeInferencePass, OperationPass<func::FuncOp>> {
```
sorenlassen, line 54: it's debatable whether it's correct to make shape inference a function pass (both in ShapeInferencePass and ONNXHybridTransformPass) because of the modification of return shapes in the function signature in `inferFunctionReturnShapes()`

in principle this is unsafe if the function pass transforms functions in parallel and they look up each others' signatures in the module symbol table, but in reality shape inference doesn't look up function signatures (and likely doesn't work if there is any func::CallOp because it would probably break when the function signature changes)

chentong319: Agree. It seems that shape inference should go top-down in the call-graph. But the semantics and usage of function call in onnx model is not clear to me.

chentong319:
The PR itself is Okay. Do you want include the following change in this PR?
1. Change `ShapeInference` to `ShapeInferenceOpInterface`, like the bufferization example: ` BufferizableOpInterface : OpInterface<"BufferizableOpInterface">`. I got a little confused by `ShapeInference` when reviewing the code.
2. Pass nullptr to inferShape(), instead of a function do nothing. We can check the null ptr in inferShape(). I am thinking that we can use that check to put the code for ONNXReturn into the inferShape of ONNXReturn. Then the special rule for the ONNXReturnOp is not needed. No sure whether such implementation is better or not.

chentong319 (approved):
LGTM!

src/Transform/ONNX/ShapeInferencePass.cpp
sorenlassen: I created issue #2185 to record this problem

Issue #2171, opened 2023-04-16
Memory Leak with Java client


cjvolzka:
Applications using models compiled with --EmitJNI slowly leak memory after ever mainGraph() call.

## Reproduce:
 
1. Compile [mnist-12.onnx](https://github.com/onnx/models/blob/main/vision/classification/mnist/model/mnist-12.onnx) model using onnx-mlir using 
    * CPU: `--O3 --EmitLib --mtriple=s390x-ibm-loz --mcpu=z14 --onnx-op-stats TXT`
    * NNPA: `--O3 --EmitLib --mtriple=s390x-ibm-loz --mcpu=z16 --maccel=NNPA --onnx-op-stats TXT` 
2. Build the [Leaker.java.txt](https://github.com/onnx/onnx-mlir/files/11243310/Leaker.java.txt) Java application: `MODEL=mnist-12.jar; javac -cp mnist-12.jar Leaker.java`
    * Calls `mainGraph()` continuously for 5 min. Change `run_max_sec` to a different value if desired.
4. Run the Leaker Java application: `java -cp .:$MODEL Leaker`
5. Watch java RES memory usage: `top -d1 -u $USER

## Expected results:
Java does it's own memory management so usage may fluctuate, but it should remain about the same after 10-30 seconds.

## Actual Results
The program continues to slowly uses more and more memory, 

## Notes:
* Affects both CPU and NNPA. Leak is faster with NNPA but that is likely due to the model running faster.
* Issue is specific to Java clients. The same test on the same .so has no leak.
    * Extract the .so can from the .jar file. 
        * `unzip mnist-12.jar -d unzipped_jar`
    * Run with [leaker.cpp.txt](https://github.com/onnx/onnx-mlir/files/11243314/leaker.cpp.txt)
        * `MODEL=unzipped_jar/libmodel.so; g++ -g --std=c++11 -O3 leaker.cpp $MODEL -o leaker -I include && time ./leaker`
    *  C++ does not continuously use more memory
    * using `valgrind` says no leakers are possible

cjvolzka:
Notes from an investigation by another internal team member (Thanks Tabari!)

> Based on my quick reading of the code, the JNI never frees the buffer that is created and associated with the `DirectByteBuffer` that is passed to the OMTensor constructor.

> Java doesn't know how to clean up directbytebuffers that are created via JNI (it doesn't know how to) so the wrapper needs to figure out when to do it as a part of the OMTensor lifetime.

[...]

PR #2176, opened 2023-04-18
Fix memory leak due to unallocated tensor data

tehbone:
Fixes https://github.com/onnx/onnx-mlir/issues/2171. The Java `OMTensor` object was not taking full ownership duties of the backing data within the native `OMTensor` object, leading to a memory leak. This fix solves the issue by allocating a new `ByteBuffer` in the JNI `OMTensor` constructor with a copy, leaving the original data to be freed along with the jni `OMTensor`.

Some initial timings suggest that there isn't any meaningful difference in timing due to the copy in the YOLOv3-12 and MNIST models.

### DIFF
--- a/src/Runtime/jni/jniwrapper.c
+++ b/src/Runtime/jni/jniwrapper.c
@@ -504,6 +504,10 @@ OMTensorList *omtl_java_to_native(
       omTensorListCreate(jni_omts, (int64_t)jomtl_omtn), jni_omtl != NULL, env,
       japi->jecpt_cls, "jni_omtl=%p", jni_omtl);
 
+  /* omTensorListCreate makes a copy of the tensor list, so we must free the
+     list that we allocateed. */
+  free(jni_omts);
+
   return jni_omtl;
 }
 
--- a/src/Runtime/jni/jniwrapper.c
+++ b/src/Runtime/jni/jniwrapper.c
@@ -595,37 +595,12 @@ jobject omtl_native_to_java(
 
     /* Create direct byte buffer Java object from native data buffer.
      *
-     * If jni_owning is true, we take ownership by setting owner flag
-     * to false. This means that when we call omTensorListDestroy
-     * the data buffer will not be freed since it has been given to
-     * the Java direct byte buffer and the Java GC will be responsible
-     * for freeing the data buffer. This way we avoid copying the data
-     * buffer.
-     *
-     * If jni_owning is false, it means the data buffer is not freeable
-     * due to one of the two following cases:
-     *
-     *   - user has malloc-ed the data buffer so the user is
-     *     responsible for freeing it
-     *   - the data buffer is static
-     *
-     * Either way, since the data buffer will be given to Java and is
-     * subject to GC, we must make a copy of the data buffer.
+     * When the Java OMTensor object is created, it will create a copy of the
+     * tensor data to ensure Java ownership of the underlying tensor data.
+     * Creation of the DirectByteBuffer here is merely a passthrough mechanism
+     * to allow for the copy to occur.
      */
     void *jbytebuffer_data = jni_data;
-    if (jni_owning) {
-      LIB_CALL(omTensorSetOwning(jni_omts[i], (int64_t)0), 1, env,
-          japi->jecpt_cls, "");
-      LOG_PRINTF(LOG_DEBUG, "omt[%d]:%p data %p ownership taken", i,
-          jni_omts[i], jni_data);
-    } else {
-      LIB_VAR_CALL(jbytebuffer_data, malloc(jni_bufferSize),
-          jbytebuffer_data != NULL, env, japi->jecpt_cls, "jbytebuffer_data=%p",
-          jbytebuffer_data);
-      memcpy(jbytebuffer_data, jni_data, jni_bufferSize);
-      LOG_PRINTF(LOG_DEBUG, "omt[%d]:%p data %p copied into %p", i, jni_omts[i],
-          jni_data, jbytebuffer_data);
-    }
     JNI_TYPE_VAR_CALL(env, jobject, jomt_data,
         (*env)->NewDirectByteBuffer(env, jbytebuffer_data, jomt_bufferSize),
         jomt_data != NULL, japi->jecpt_cls, "omt[%d]:jomt_data=%p", i,
--- a/src/Runtime/jni/jniwrapper.c
+++ b/src/Runtime/jni/jniwrapper.c
@@ -603,6 +603,13 @@ jobject omtl_native_to_java(
      * tensor data to ensure Java ownership of the underlying tensor data.
      * Creation of the DirectByteBuffer here is merely a passthrough mechanism
      * to allow for the copy to occur.
+     *
+     * If jni_owning is false, it means the data buffer is not freeable
+     * due to one of the two following cases:
+     *
+     *   - user has malloc-ed the data buffer so the user is
+     *     responsible for freeing it
+     *   - the data buffer is static
      */
     void *jbytebuffer_data = jni_data;
     JNI_TYPE_VAR_CALL(env, jobject, jomt_data,
--- a/src/Runtime/jni/jniwrapper.c
+++ b/src/Runtime/jni/jniwrapper.c
@@ -497,8 +497,7 @@ OMTensorList *omtl_java_to_native(
   free(jobj_omts);
 
   /* Create OMTensorList to be constructed and passed to the model
-   * shared library. Note that we do own the pointers to the native
-   * OMTensor structs, jni_omts.
+   * shared library.
    */
   LIB_TYPE_VAR_CALL(OMTensorList *, jni_omtl,
       omTensorListCreate(jni_omts, (int64_t)jomtl_omtn), jni_omtl != NULL, env,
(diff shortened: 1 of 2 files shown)


gongsu832:
We made a lot of effort trying not to copy the input tensor data so I prefer we don't take the copy route as much as possible. In fact, I remember a while ago @AlexandreEichenberger said onnxruntime claimed that one of their advantages is that they don't copy the tensor data and asked if we can make the same claim and I told him yes. So we should keep that promise as much as possible. As mentioned in the slack conversation, I'm working on a solution for Java 9+ using the cleaner API that does not require copy. For Java 8, we will have to use the finalize() to avoid copy. This does have the unfortunate consequence that we will have to maintain two implementations. But the code involved is relatively simple and I'm hopeful that WMLz will eventually upgrade from Java 8.

cjvolzka:
I'm all for avoiding the extra copy if we can. However as a side note, don't we already do a copy whenever someone tries to call the get*Data() like at https://github.com/onnx/onnx-mlir/blob/7e49205dcd6c818bc92aba0abf13280f719fda54/src/Runtime/jni/src/com/ibm/onnxmlir/OMTensor.java#L392-L395?

If we use finalize()/Cleaner, do we need to do anything about https://github.com/onnx/onnx-mlir/pull/2174#issuecomment-1513486700? Do we even allow changing the data after it's initially set? If we don't support that, we could make it explicitly a one time write (ie throw an exception in `setData` if `_data` is not `null`) then we'll know we're safe.

I have no idea when WMLz will drop Java 8 support but it'll be Sept 2026 at the latest since that's when IBM's Java goes out of support for zOS (https://www.ibm.com/support/pages/java-sdk-lifecycle-dates)

gongsu832:
Yes the `fb.get(f)` copy is unavoidable (at least I don't know how) since Java cannot simply convert a FloatBuffer into float[]. So we are trying to avoid a 2nd copy. And sadly, our claim of not copying tensor data currently only applies to C/C++ case.

We do have public `setXXXData` to allow changing tensor data. The issue mentioned in https://github.com/onnx/onnx-mlir/pull/2174#issuecomment-1513486700 isn't too hard to deal with. Basically, with our internal flag, we know if the data is backed by native buffer generated by model runtime. We just need to make the JNI call to free it before setting the new data.

And some good news and bad news from my testing so far. Good news is that `finalize` does appear to fix the leak for Java 8 (it does NOT work for Java 11 if I just ignore the compile time warning and the Leaker still leaks). Bad news is that `cleaner` doesn't appear to be working. The RES size of the Leaker still creeps up although very very slowly. I ran it for about an hour and it creeps up from roughly 50MB to 70MB, whereas the finalize-fixed version hovers around 53MB. So I must be doing something wrong with my cleaner implementation.

gongsu832:
> > If we use finalize()/Cleaner, do we need to do anything about [#2174 (comment)](https://github.com/onnx/onnx-mlir/pull/2174#issuecomment-1513486700)?
> 
> It'll be a minor bit of tweaking to ensure that all of the setters deallocate the data, if necessary. @gongsu832 there is a third option to support all releases. Utilizing a combination of `WeakReference`, `RefrenceQueue` and a cleanup task, one could approximate the `Cleanup` functionality and ensure it is compatible on Java8+.

I was looking into that but it appears to be overly complicated for what we want to do. And I seem to recall reading that things on the WeakReference (or is it Phantom Reference, can't remember) has changed between 8 and 9. So it doesn't inspire a lot of confidence for me that the same code will work for both 8 and 9+. I could be wrong though.

src/Runtime/jni/jniwrapper.c
```diff
@@ -595,37 +595,12 @@ jobject omtl_native_to_java(
...
-     * for freeing the data buffer. This way we avoid copying the data
-     * buffer.
-     *
-     * If jni_owning is false, it means the data buffer is not freeable
-     * due to one of the two following cases:
-     *
-     *   - user has malloc-ed the data buffer so the user is
-     *     responsible for freeing it
-     *   - the data buffer is static
-     *
```
gongsu832, line 611: I wouldn't remove these comments. It serves as a reminder why the output tensor data buffer could have owning=false (due to compiler generating static data).

src/Runtime/jni/jniwrapper.c
```diff
@@ -595,37 +595,12 @@ jobject omtl_native_to_java(
...
-      LOG_PRINTF(LOG_DEBUG, "omt[%d]:%p data %p ownership taken", i,
-          jni_omts[i], jni_data);
-    } else {
-      LIB_VAR_CALL(jbytebuffer_data, malloc(jni_bufferSize),
-          jbytebuffer_data != NULL, env, japi->jecpt_cls, "jbytebuffer_data=%p",
-          jbytebuffer_data);
-      memcpy(jbytebuffer_data, jni_data, jni_bufferSize);
-      LOG_PRINTF(LOG_DEBUG, "omt[%d]:%p data %p copied into %p", i, jni_omts[i],
-          jni_data, jbytebuffer_data);
-    }
```
gongsu832, line 628: At the end of the loop, do we need to remove local references to the various java objects created?
```
diff --git a/src/Runtime/jni/jniwrapper.c b/src/Runtime/jni/jniwrapper.c
index 5fc6ee99..c703c4a0 100644
--- a/src/Runtime/jni/jniwrapper.c
+++ b/src/Runtime/jni/jniwrapper.c
@@ -658,6 +658,11 @@ jobject omtl_native_to_java(
     /* Set the OMTensor object in the object array */
     JNI_CALL(env, (*env)->SetObjectArrayElement(env, jobj_omts, i, jobj_omt), 1,
         NULL, "");
+
+    JNI_CALL(env, (*env)->DeleteLocalRef(env, jomt_data), 1, NULL, "");
+    JNI_CALL(env, (*env)->DeleteLocalRef(env, jomt_shape), 1, NULL, "");
+    JNI_CALL(env, (*env)->DeleteLocalRef(env, jomt_strides), 1, NULL, "");
+    JNI_CALL(env, (*env)->DeleteLocalRef(env, jobj_omt), 1, NULL, "");
   }

   /* Create the OMTensorList java object */
```
Normally local references are automatically released when the function goes out of scope. But here we are in a loop reusing all the object pointers for each tensor so I think we need these explicit DeleteLocalRef?

tehbone: While the C object pointers are being reused the refs are still tied to the scope of the function. Once the function returns back to Java, _any_ object that was created will have its local reference popped unless it was returned. Since these are transient objects I could add deletelocalref but given the size of the java objects they represent it only would matter if dealing with thousands of these within a single JNI call.

gongsu832: OK I just want to make sure there is no leakage of the local reference. The loop could theoretically iterate over thousands of tensors (i.e., output tensor list has thousands of tensors). Would that be a problem without the DeleteLocalRef?

tehbone: It shouldn't be a problem. Remember, you're talking about saving maybe 1K per tensor, if even that much

src/Runtime/jni/jniwrapper.c
```diff
@@ -504,6 +504,10 @@ OMTensorList *omtl_java_to_native(
       omTensorListCreate(jni_omts, (int64_t)jomtl_omtn), jni_omtl != NULL, env,
```
gongsu832, line 504: Can you remove the comment "Note that we do own..."? It no longer applies since now omTensorListCreate copies jni_omts.

Also, you need to replace `omTensorCreate` with `omTensorCreateWithOwnership(owning=1)` to free the input tensors themselves.

tehbone: I can remove the comment but the `omTensorCreate` is still valid. input tensors' data is still owned by Java (or to be more specific, the caller), so we shouldn't mark the tensor as needing to free the data.

gongsu832: Ah yes indeed...mind glitch.

cjvolzka:
Not that watching `top` RES is a perfect measure, but here's what I got on a 120min mnist-12 loop run on LoZ Java 11:

For this bytbuffer branch version, the first 30 sec RES fluctuates and stabilizes at 58.1m. Between min 4-7 RES moves up to 61.5m. It stays stable until min 47 where it goes to 62.5m. After that RES doesn't move again and is still there at the 120min mark. In the end, mainGraph ran 1,951,483 times.

I did a similar test for the ["reaper" version](https://github.com/onnx/onnx-mlir/pull/2182#issuecomment-1516826477) and the results here appear slightly better. Although it's hard to say how much of the RES difference is the the code vs JVM doing it's thing.

cjvolzka:
@gongsu832 So which version of the fix do you want to see go in? I'd like to start rallying around one so we can merge in a fix early next week if possible. I lean towards bytebuffer as it seems the simplist codewise and the most stable from my RES testing. However I'm not really set one way or the other other than the fix working for 1.8 and 11+

gongsu832:
As I mentioned in the other PR https://github.com/onnx/onnx-mlir/pull/2182#issuecomment-1515631129, if we really don't want to have two separate versions for 8 and 11, then my vote would go to the copy fix in this PR for now. I'm still testing my cleaner fix to make sure it works correctly but only for the purpose of getting it ready so that when WLMz finally upgrades from Java 8 we can then switch to the cleaner fix.

cjvolzka (approved):
From my testing, I'm good approving this PR. I'd still like another +1 on the code itself though because I'm not a JNI expert.

gongsu832 (approved):
LGTM

PR #2178, opened 2023-04-19
Fix JNI wrapper memory leak

gongsu832:
Fix JNI wrapper memory leak due to NewDirectByteBuffer not subjecting to Java GC. Use the Cleaner API requiring Java 9+ to avoid copying tensor data

### DIFF
--- a/src/Runtime/jni/jniwrapper.c
+++ b/src/Runtime/jni/jniwrapper.c
@@ -598,9 +598,11 @@ jobject omtl_native_to_java(
      * If jni_owning is true, we take ownership by setting owner flag
      * to false. This means that when we call omTensorListDestroy
      * the data buffer will not be freed since it has been given to
-     * the Java direct byte buffer and the Java GC will be responsible
-     * for freeing the data buffer. This way we avoid copying the data
-     * buffer.
+     * the Java direct byte buffer. However, the direct byte buffer
+     * created by NewDirectByteBuffer is not subject to Java GC so we
+     * pass clean=true to the OMTensor construct to initializ a cleaner
+     * to manually free the data buffer. This way we avoid copying the
+     * data buffer.
      *
      * If jni_owning is false, it means the data buffer is not freeable
      * due to one of the two following cases:
@@ -609,8 +611,9 @@ jobject omtl_native_to_java(
      *     responsible for freeing it
      *   - the data buffer is static
      *
-     * Either way, since the data buffer will be given to Java and is
-     * subject to GC, we must make a copy of the data buffer.
+     * Either way, since the data buffer given to Java is not subject
+     * to GC, we can simply pass it through with clean=false and no
+     * cleaner will be initialized.
      */
     void *jbytebuffer_data = jni_data;
     if (jni_owning) {
@@ -618,14 +621,8 @@ jobject omtl_native_to_java(
           japi->jecpt_cls, "");
       LOG_PRINTF(LOG_DEBUG, "omt[%d]:%p data %p ownership taken", i,
           jni_omts[i], jni_data);
-    } else {
-      LIB_VAR_CALL(jbytebuffer_data, malloc(jni_bufferSize),
-          jbytebuffer_data != NULL, env, japi->jecpt_cls, "jbytebuffer_data=%p",
-          jbytebuffer_data);
-      memcpy(jbytebuffer_data, jni_data, jni_bufferSize);
-      LOG_PRINTF(LOG_DEBUG, "omt[%d]:%p data %p copied into %p", i, jni_omts[i],
-          jni_data, jbytebuffer_data);
     }
+
     JNI_TYPE_VAR_CALL(env, jobject, jomt_data,
         (*env)->NewDirectByteBuffer(env, jbytebuffer_data, jomt_bufferSize),
         jomt_data != NULL, japi->jecpt_cls, "omt[%d]:jomt_data=%p", i,
@@ -652,7 +649,8 @@ jobject omtl_native_to_java(
     /* Create the OMTensor Java object */
     JNI_TYPE_VAR_CALL(env, jobject, jobj_omt,
         (*env)->NewObject(env, japi->jomt_cls, japi->jomt_constructor,
-            jomt_data, jomt_shape, jomt_strides, jomt_dataType),
+            jomt_data, jomt_shape, jomt_strides, jomt_dataType,
+            (jboolean)jni_owning),
         jobj_omt != NULL, japi->jecpt_cls, "omt[%d]:jobj_omt=%p", i, jobj_omt);
 
     /* Set the OMTensor object in the object array */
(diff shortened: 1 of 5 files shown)


src/Runtime/jni/jniwrapper.c
```diff
@@ -910,3 +909,23 @@ JNIEXPORT jstring JNICALL Java_com_ibm_onnxmlir_OMModel_output_1signature_1jni(
...
+  JNI_TYPE_VAR_CALL(env, jclass, jecpt_cls,
+      (*env)->FindClass(env, jnistr[CLS_JAVA_LANG_EXCEPTION]),
+      jecpt_cls != NULL, NULL, "Class java/lang/Exception not found");
+
+  /* Get native data buffer backing direct byte buffer */
+  JNI_TYPE_VAR_CALL(env, void *, jni_data,
+      (*env)->GetDirectBufferAddress(env, java_data), jni_data != NULL,
+      jecpt_cls, "jni_data=%p", jni_data);
+
+  free(jni_data);
```
tehbone, line 928: I ran into this during testing of my older `finalize()`-based PR, but `jni_data` points to the aligned allocation, not to the start of the allocation.

cjvolzka:
I tried this in our build container but I got errors trying to build onnx-mlir
```
src/com/ibm/onnxmlir/OMTensor.java:5: error: cannot find symbol
import java.lang.ref.Cleaner;
                    ^
  symbol:   class Cleaner
  location: package java.lang.ref
src/com/ibm/onnxmlir/OMTensor.java:113: error: cannot find symbol
    private static final Cleaner _cleaner = Cleaner.create();
                         ^
  symbol:   class Cleaner
  location: class OMTensor
src/com/ibm/onnxmlir/OMTensor.java:114: error: package Cleaner does not exist
    private final Cleaner.Cleanable _cleanable;
                         ^
src/com/ibm/onnxmlir/OMTensor.java:113: error: cannot find symbol
    private static final Cleaner _cleaner = Cleaner.create();
                                            ^
  symbol:   variable Cleaner
  location: class OMTensor

4 errors
gmake[2]: *** [src/Runtime/jni/CMakeFiles/javaruntime.dir/build.make:90: src/Runtime/jni/CMakeFiles/javaruntime.dir/java_compiled_javaruntime] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:5928: src/Runtime/jni/CMakeFiles/javaruntime.dir/all] Error 2
gmake[1]: *** Waiting for unfinished jobs....
```

I'm guessing it's because in our builder image we only install Java 1.8 since that's what we ultimately target running the models on. I can move that to Java 11. However, will that cause any issues then with the compiled model .jar files being compatible to run on Java 8?

src/Runtime/jni/jniwrapper.c
gongsu832: I see what you are saying but I'm a bit surprised why this would happen. Basically, when you call `NewDirectByteBuffer`, you give it the pointer to the memory block. Now when you call `GetDirectBufferAddress`, it's going to give you a different (aligned) pointer? I mean I gave it the pointer, now I want it back, why would it change it behind my back?!

gongsu832:
> I'm guessing it's because in our builder image we only install Java 1.8 since that's what we ultimately target running the models on. I can move that to Java 11. However, will that cause any issues then with the compiled model .jar files being compatible to run on Java 8?

I think you should keep your 1.8 and use finalize if that's what you are targeting. The finalize code is actually very simple. You can replace OMTensor.java with this.

[OMTensor.java.txt](https://github.com/onnx/onnx-mlir/files/11275436/OMTensor.java.txt)

src/Runtime/jni/jniwrapper.c
gongsu832: OK maybe you mean onnx-mlir might give you an aligned pointer that's not the start of the allocation? OK in that case we need fiddle a bit like the offset thing you did. Itchy.

cjvolzka:
> > I'm guessing it's because in our builder image we only install Java 1.8 since that's what we ultimately target running the models on. I can move that to Java 11. However, will that cause any issues then with the compiled model .jar files being compatible to run on Java 8?
> 
> I think you should keep your 1.8 and use finalize if that's what you are targeting. The finalize code is actually very simple. You can replace OMTensor.java with this.
> 
> [OMTensor.java.txt](https://github.com/onnx/onnx-mlir/files/11275436/OMTensor.java.txt)

I'd greatly prefer to not have to force source code changes into our build that's not in the community. It creates tech debt and becomes a maintenance nightmare in the long term.

If that's the only way to do have the code "support" both, then I'd greatly lean towards either only supporting `finalize()` and not doing Cleaner until Java 1.8 support is no longer needed; or take the extra copy approach which avoids both finalize and Cleaner.

src/Runtime/jni/jniwrapper.c
gongsu832: OK I looked at the conversion on your finalize PR and you said:

>I thought about doing that but I wanted to avoid a limitation with java - all of the integral based arguments are signed. With an offset, which can be signed, I can keep the semantics more or less the same. With a Cleaner based approach I can create another DirectByteBuffer holding the actual allocation variable which the cleaner can use to clean up - no offset calculation required.

So I think instead of passing in the offset, we could just pass in the aligned pointer as jlong, right? As you said later, it doesn't really matter if it's signed or not. 64-bit is 64-bit. It only matters how you interpret it. Doing that would save us from having to keep another flag.

gongsu832:
> > > I'm guessing it's because in our builder image we only install Java 1.8 since that's what we ultimately target running the models on. I can move that to Java 11. However, will that cause any issues then with the compiled model .jar files being compatible to run on Java 8?
> > 
> > 
> > I think you should keep your 1.8 and use finalize if that's what you are targeting. The finalize code is actually very simple. You can replace OMTensor.java with this.
> > [OMTensor.java.txt](https://github.com/onnx/onnx-mlir/files/11275436/OMTensor.java.txt)
> 
> I'd greatly prefer to not have to force source code changes into our build that's not in the community. It creates tech debt and becomes a maintenance nightmare in the long term.
> 
> If that's the only way to do have the code "support" both, then I'd greatly lean towards either only supporting `finalize()` and not doing Cleaner until Java 1.8 support is no longer needed; or take the extra copy approach which avoids both finalize and Cleaner.

I thought you just want to do a quick test on 1.8 yourself. That's certainly not the only way to support both. We can have cmake pick the right version depending on which Java version you are using.

As I mentioned in another thread, finalize does not work on Java 11 (compile gives warning, runtime has no effect) so we can't really have only finalize.

AlexandreEichenberger:
@gongsu832 do you need help from MS folks for the Window failure; or maybe just re-update to the latest and see if this now works?

AlexandreEichenberger:
@gongsu832 @cjvolzka is there a resolution for this PR?

gongsu832:
I'd like to keep it open just so we don't forget and also I can rebase the patch to the latest code. So I converted it draft.

gongsu832:
> @gongsu832 Still desirable? If so can we reactualizes it? Or should we close?
This PR serves as a reminder that when Java 1.8 goes out of service we should revisit the issue and converge on a single and simpler direct byte buffer cleaner. So I'd keep it open unless there is some problem with doing that.

PR #2179, opened 2023-04-19
Add a pattern for Tile op in DimAnalysis 

imaihal:
This patch adds a pattern for Tile op in DimAnalysis. 

Dimension in `output` tensor is dimension in `input` tensor multiplied by value in `repeats` tensor. (output_dim[i] = input_dim[i] * repeats[i])
So, if input_dim[i] is 1, output_dim[i]  and repeats[i] are equal.

Example
```
%0 = "onnx.Dim"(%arg0) {axis = 0 : si64} : (tensor<?x?xi64>) -> tensor<1xi64>
%1 = "onnx.Dim"(%arg0) {axis = 1 : si64} : (tensor<?x?xi64>) -> tensor<1xi64>
%2 = "onnx.Concat"(%0, %1) {axis = 0 : si64} : (tensor<1xi64>, tensor<1xi64>) -> tensor<2xi64>
%3 = "onnx.Tile"(%arg1, %3) : (tensor<1x1xi64>, tensor<2xi64>) -> tensor<?x?xi64>
```
This pattern appears in `t5-decoder-with-lm-head.onnx` in model zoo.

### DIFF
--- a/src/Dialect/ONNX/ONNXDimAnalysis.cpp
+++ b/src/Dialect/ONNX/ONNXDimAnalysis.cpp
@@ -594,9 +594,6 @@ void DimAnalysis::visitDim(
 
   // TileOp
   if (auto tileOp = dyn_cast<ONNXTileOp>(op)) {
-    if (!areDimsFromConcat(tileOp.getRepeats()))
-      return;
-
     // Special case:
     // output_dim[i] = input_dim[i] * repeats[i]
     //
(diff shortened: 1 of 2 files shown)


imaihal:
@tungld PR #2173 works well in `t5-encoder`. `t5-decoder` was bit different. I would like to your comments.

src/Dialect/ONNX/ONNXDimAnalysis.cpp
```diff
@@ -592,6 +592,37 @@ void DimAnalysis::visitDim(
     return;
   }
 
+  // TileOp
+  if (auto tileOp = dyn_cast<ONNXTileOp>(op)) {
+    if (!areDimsFromConcat(tileOp.getRepeats()))
```
tungld, line 597: Early `return` may block general case for tileOp.

We should call `exploreSameInputDims(dim, op, sameDims);` at the beginning to capture all general cases for tileOp, then go to special case. See MatMul as an example.

Another way is only use `return` inside the `if` of the special case, to let other cases go through (See ExpandOp as an example):
```
if (auto tileOp = dyn_cast<ONNXTileOp>(op)) {
   if (areDimsFromConcat(tileOp.getRepeats()) && inputShape[dimIndex] == 1) {
      do something;
     return;
  }
}
```

tungld (commented):
Nice, it's a good pattern. You should have a lit test for it too.

src/Dialect/ONNX/ONNXDimAnalysis.cpp
tungld: I recommend the latter way.

imaihal: @tungld I used the latter way. Thanks for your comment!

tungld (approved):
LGTM. Thanks!

PR #2182, opened 2023-04-19
Fix JNI memory leak for tensor data

### LINKED ISSUE #2171: Memory Leak with Java client

tehbone:
Fixes #2171. This is yet another approach that is compatible with Java 8+. It mirrors a Java 9 `Cleaner`-based approach, but utilizing `WeakReference`s to keep track of the lifetime of the object. A `ByteBuffer` is used to keep track of the underlying allocation, and when the reference to the data expires, the allocation will be freed. It is also tied to the lifetime of the `ByteBuffer` created by the JNI instead of the `OMTensor` to handle scenarios where the `OMTensor` is updated. In such scenarios, the GC will grab the original `ByteBuffer` well before the `OMTensor` is destroyed, allowing for the allocation to be more aggressively reclaimed.

This fix also removes the extra copy that would occur if the model did _not_ allocate storage for the returned `OMTensor` object.

### DIFF
--- a/src/Runtime/jni/jniwrapper.c
+++ b/src/Runtime/jni/jniwrapper.c
@@ -28,9 +28,11 @@
 
 #include "OnnxMlirRuntime.h"
 #include "com_ibm_onnxmlir_OMModel.h"
+#include "com_ibm_onnxmlir_OMTensor.h"
 #include "jnilog.h"
 
 extern OMTensorList *run_main_graph(OMTensorList *);
+void *omTensorGetAllocatedPtr(const OMTensor *);
 
 /* Declare type var, make call and assign to var, check condition.
  * It's assumed that a Java exception has already been thrown so
@@ -175,7 +177,7 @@ const char *jnistr[] = {
     "com/ibm/onnxmlir/OMTensor",       /* 3  CLS_COM_IBM_ONNXMLIR_OMTENSOR    */
     "com/ibm/onnxmlir/OMTensorList",   /* 4  CLS_COM_IBM_ONNXMLIR_OMTENSORLIST*/
     "<init>",                          /* 5  CTOR_INIT                        */
-    "(Ljava/nio/ByteBuffer;[J[JI)V",   /* 6  CTOR_OMTENSOR                    */
+    "(Ljava/nio/ByteBuffer;[J[JILjava/nio/ByteBuffer;)V", /* 6  CTOR_OMTENSOR */
     "([Lcom/ibm/onnxmlir/OMTensor;)V", /* 7  CTOR_OMTENSORLIST */
     "()Ljava/nio/ByteBuffer;",         /* 8  SIG_GET_DATA                     */
     "(Ljava/nio/ByteBuffer;)V",        /* 9  SIG_SET_DATA                     */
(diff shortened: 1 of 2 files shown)


tehbone:
@gongsu832 I know I have PR #2176 out there, but I think this approach satisfies all of the concerns.

gongsu832:
@tehbone As much as I hate copying, between this and copying, I'd just go for copying then switch to Cleaner when WMLz upgrades from Java 8 if we don't want to maintain two sources. I think this is just too complicated for what we want to do. Eventually Cleaner will be the way to go and it provides both auto close and explicit user close.

Moot point, I think you can just pass the jni_alloc as a jlong to the OMTensor constructor without using a DirectByteBuffer.

cjvolzka:
Not that watching `top` RES is a perfect measure, but here's what I got on a 120min mnist-12 loop run on LoZ Java 11:

This "reaper" branch  version the first 30 sec RES fluctuates and stabilizes at 60.5m at 41 seconds. Goes to 60.7m at 1min. Slowly moves to 61.1m at 5min and is at 62.2 at 7 min and stays there until 47 min. Then it goes to 63.3m until 55min. Finally over min 55-57 it goes to 92.4 where it stays until the run ends at 120 minutes. In the end, mainGraph ran 1,926,597 times.

I did a similar test for the "bytebuffer" version and the results there appear slightly better. Although it's hard to say how much of the RES difference is the the code vs JVM doing it's thing.

tehbone:
Closing, as it has been superseded by #2176

PR #2186, opened 2023-04-21
NonZero constant propagation

sorenlassen:
This PR illustrates the problem with changing shape in constant propagation, which doesn't work with `func.return` and was solved with `onnx.Return` in PR #2239. See the lit test in onnx_constprop.mlir:
```
func.func @test_nonzero() -> tensor<2x?xi64> {
  %0 = "onnx.Constant"() {value = dense<[[2, 1], [0, 2], [0, 1]]> : tensor<3x2xi8>} : () -> tensor<3x2xi8>
  %1 = "onnx.NonZero"(%0) : (tensor<3x2xi8>) -> tensor<2x?xi64>
  // NOTE: We cannot use return (i.e. "func.return") because the change of shape makes
  // func::ReturnOp::verify() fail with:
  //
  //   error: type of return operand 0 ('tensor<2x4xi64>') doesn't match function result type ('tensor<2x?xi64>') in function @test_nonzero
  //
  onnx.Return %1 : tensor<2x?xi64>
}
```

### DIFF
--- a/src/Transform/ONNX/ConstProp.cpp
+++ b/src/Transform/ONNX/ConstProp.cpp
@@ -869,6 +869,19 @@ Value ConstPropRange(PatternRewriter &rewriter, Value replacingValue,
 }
 
 //===----------------------------------------------------------------------===//
+// Code to perform constant propagation for NonZero.
+//===----------------------------------------------------------------------===//
+
+Value ConstPropNonZero(
+    PatternRewriter &rewriter, Value replacingValue, Value constValue) {
+  ConstPropCounters::count("NonZero", {constValue});
+  ElementsAttr constElements = getConstValueElements(constValue);
+  OnnxElementsAttrBuilder elementsBuilder(rewriter.getContext());
+  ElementsAttr nonZeroElements = elementsBuilder.nonZero(constElements);
+  return createReplacingConstantOp(rewriter, replacingValue, nonZeroElements);
+}
+
+//===----------------------------------------------------------------------===//
 // Pattern definition.
 //===----------------------------------------------------------------------===//
 
--- a/test/mlir/onnx/onnx_constprop.mlir
+++ b/test/mlir/onnx/onnx_constprop.mlir
@@ -1349,7 +1349,6 @@ func.func @test_nonzero() -> tensor<2x?xi64> {
   //
   //   error: type of return operand 0 ('tensor<2x4xi64>') doesn't match function result type ('tensor<2x?xi64>') in function @test_nonzero
   //
-  // Therefore we use onnx.Return instead, although it isn't meant for function return.
   onnx.Return %1 : tensor<2x?xi64>
 
 // CHECK-LABEL:  func.func @test_nonzero
--- a/test/mlir/onnx/onnx_constprop.mlir
+++ b/test/mlir/onnx/onnx_constprop.mlir
@@ -1338,3 +1338,23 @@ func.func @test_range_fp() -> tensor<3xf32> {
 // CHECK:           onnx.Return [[VAR]] : tensor<3xf32>
 // CHECK:         }
 }
+
+// -----
+
+func.func @test_nonzero() -> tensor<2x?xi64> {
+  %0 = "onnx.Constant"() {value = dense<[[2, 1], [0, 2], [0, 1]]> : tensor<3x2xi8>} : () -> tensor<3x2xi8>
+  %1 = "onnx.NonZero"(%0) : (tensor<3x2xi8>) -> tensor<2x?xi64>
+  // NOTE: We cannot use return (i.e. "func.return") because the change of shape makes
+  // func::ReturnOp::verify() fail with:
+  //
+  //   error: type of return operand 0 ('tensor<2x4xi64>') doesn't match function result type ('tensor<2x?xi64>') in function @test_nonzero
+  //
+  // Therefore we use onnx.Return instead, although it isn't meant for function return.
+  onnx.Return %1 : tensor<2x?xi64>
+
+// CHECK-LABEL:  func.func @test_nonzero
+// CHECK-SAME:   () -> tensor<2x?xi64> {
+// CHECK:           [[VAR:%.+]] = onnx.Constant dense<{{\[}}[0, 0, 1, 2], [0, 1, 1, 1]{{\]}}> : tensor<2x4xi64>
+// CHECK:           onnx.Return [[VAR]] : tensor<2x4xi64>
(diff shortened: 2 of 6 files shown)


sorenlassen:
the CI failure is because of the test_nonzero_example_cpu backend test. It can be reproduced with
`cmake --build . --target check-onnx-backend-constant` or just the one test:
```
TEST_VERBOSE=1 TEST_CASE_BY_USER=test_nonzero_example_cpu TEST_CONSTANT=true python3 test/backend/RelWithDebInfo/test.py
```
which outputs
```
RelWithDebInfo/bin/onnx-mlir -O3 --EmitLib ./test_nonzero_example/test_nonzero_example.onnx -o=./test_nonzero_example/test_nonzero_example
IMPORTER FORCE DYNAMIC  None
error: type of return operand 0 ('tensor<2x3xi64>') doesn't match function result type ('tensor<2x?xi64>') in function @main_graph
```

sorenlassen:
I just noticed that the SimplifyShapeRelatedOpsPass patterns has a function `updateFunctionSignature()` to solve the problem of func::ReturnOp::verify() failing when return type shapes change, see PR #1695

cc @tungld

tungld:
> I just noticed that the SimplifyShapeRelatedOpsPass patterns has a function updateFunctionSignature() to solve the problem of func::ReturnOp::verify() failing when return type shapes change, see PR https://github.com/onnx/onnx-mlir/pull/1695

Yes, same problem I think.

chentong319:
What's our decision, to keep the function signature unchanged or to update the function signature? I thought we wanted to introduce a special Op to isolate the return from shape change. 
```
%0 = "onnx.Constant"() {value = dense<[[2, 1], [0, 2], [0, 1]]> : tensor<3x2xi8>} : () -> tensor<3x2xi8>
%1 = "onnx.NonZero"(%0) : (tensor<3x2xi8>) -> tensor<2x?xi64>
%2 = "onnx.returnTensorCast(%1) : (tensor<2x4xT>) -> tensor<2x?xT>
onnx.Return %2 : tensor<2x?xi64>
```
All the transformation does not need to worry about updating function signature, and the better shape information for the output of function can be retrieved if needed.

sorenlassen:
> What's our decision, to keep the function signature unchanged or to update the function signature? I thought we wanted to introduce a special Op to isolate the return from shape change.
> All the transformation does not need to worry about updating function signature, and the better shape information for the output of function can be retrieved if needed.

This option was suggested but I disagree. I believe it's better to introduce an ONNXFunctionReturnOp which permits more specific operand shapes than the function signature return types, because:
1. I don't see significant downside to having an ONNX-dialect specific return op. We already have an ONNX-dialect specific terminator ONNXReturnOp for If/Scan/Loop which works fine.
2. It will be symmetrical to ONNXCallOp with respect to shape inference. (For shape inference to work well, ONNXCallOp needs to allow more specific operand shapes than the callee function signature argument shapes, and callee function signature return shapes to be more specific than ONNXCallOp return shapes.)
3. Then we will be in a situation where we can always change any operation return shape to become more specific, including function return values, without making the IR invalid. And we can always invoke the shape inference pass to propagate shapes (including making function return shapes more specific, plus clone functions with more specific argument shapes if we want to).

Inserting and `onnx.returnTensorCast` operation between the function body and return operation is a bad idea because it creates a convention that will creep into rewriting and conversion passes which will need special cases to deal with this convention. I've experienced this in one of Groq's internal dialects which has ops inserted between function bodies and function returns (and other ops between function arguments and function bodies) and they create a "tax" on development and creates ever-growing complexity throughout the code base over time.

chentong319:
To my understanding, the suggested ONNXFunctionReturnOp will be the combination of the ONNXReturnCast and current function return Op. Can ONNXFunctionReturnOp replace function returnOp for function dialect requirement?

chentong319 (commented):
What we discussed is a long-term solution.
This PR fixed the problem for constant propagation. I am Okay with it.

src/Transform/ONNX/ConstProp.cpp
```diff
@@ -1008,6 +1008,19 @@ Value ConstPropRange(PatternRewriter &rewriter, Value replacingValue,
...
+// Code to perform constant propagation for NonZero.
+//===----------------------------------------------------------------------===//
+
+Value ConstPropNonZero(
+    PatternRewriter &rewriter, Value replacingValue, Value constValue) {
+  ConstPropCounters::count("NonZero", {constValue});
+  ElementsAttr constElements = getConstValueElements(constValue);
+  OnnxElementsAttrBuilder elementsBuilder(rewriter.getContext());
+  ElementsAttr nonZeroElements = elementsBuilder.nonZero(constElements);
+  return createReplacingConstantOp(rewriter, replacingValue, nonZeroElements);
```
tungld, line 1021: @sorenlassen @chentong319 to make progress, could we do function type update inside `createReplacingConstantOp`? I know it's not the final solution, just would like to have NonZero constant prop landed since we see it in GPT-2 models.

sorenlassen: we could go ahead and introduce the new onnx function return op, that we've discussed, see PR #2239

sorenlassen:
@tungld this PR is now ready for review, I rebased it after PR #2239 was merged,

test/mlir/onnx/onnx_constprop.mlir
```diff
@@ -1338,3 +1338,22 @@ func.func @test_range_fp() -> tensor<3xf32> {
...
+// -----
+
+func.func @test_nonzero() -> tensor<2x?xi64> {
+  %0 = "onnx.Constant"() {value = dense<[[2, 1], [0, 2], [0, 1]]> : tensor<3x2xi8>} : () -> tensor<3x2xi8>
+  %1 = "onnx.NonZero"(%0) : (tensor<3x2xi8>) -> tensor<2x?xi64>
+  // NOTE: We cannot use return (i.e. "func.return") because the change of shape makes
+  // func::ReturnOp::verify() fail with:
+  //
+  //   error: type of return operand 0 ('tensor<2x4xi64>') doesn't match function result type ('tensor<2x?xi64>') in function @test_nonzero
+  //
```
tungld, line 1351: Looks like this note can be removed now, it is not about this test anymore.

tungld (approved):
LGTM with cleaning up some old comments.

test/mlir/onnx/onnx_constprop.mlir
sorenlassen: good point

removed

Issue #2143, opened 2023-04-05
Constant folding


sorenlassen:
This issue is meant to track the implementation of [constant folding](https://github.com/llvm/llvm-project/blob/main/mlir/docs/Canonicalization.md#canonicalizing-with-the-fold-method) for the ONNX dialect.

So far we have made ONNXConstantOp and ONNXNoneOp ConstantLike and we have implemented ONNXDialect::materializeConstant() and implement fold() for ONNXNoneOp, ONNXConstantOp, ONNXSqueeze[V11]Op. See PR #2009 which contains further insights on the topic. See also issue #1718.

Next steps include moving logic from ConstProp.cpp to fold() methods for different ops. Hopefully, this is facilitated by the move of some constant propagation logic into reusable ElementsAttrBuilder methods.

chentong319:
Agree on this direction.
 I moved ConstProp, and its prerequisite ShapeInference for SqueezeOp into its fold in a previous PR.
There is a small issue: the current interface of ShapeInference requires a function pointer parameter, which is used only by control flow Ops. It is OK to just provide a NULL pointer. I am thinking of remove it after the hybrid transformation is done. Actually the parameter is not needed if shape inference is driven with hybrid transformation.

PR #2187, opened 2023-04-23
move ConstProp implementations to ElementsAttrBuilder methods

sorenlassen:
This addresses leftover TODOs from PR #1874 and will make it easier to migrate current constant propagation for ScatterND, Slice, Concat, and Gather to constant folding in the future, see issue #2143.

Also made IndexExpr::getLiteral/Shape/DynSymbols/OpOrFoldResults use ArrayRef like IndexExpr::getValues.

### DIFF
--- a/src/Dialect/ONNX/ElementsAttr/ElementsAttrBuilder.cpp
+++ b/src/Dialect/ONNX/ElementsAttr/ElementsAttrBuilder.cpp
@@ -551,6 +551,46 @@ ElementsAttr ElementsAttrBuilder::gather(
   });
 }
 
+ElementsAttr ElementsAttrBuilder::scatterND(
+    ElementsAttr input, ElementsAttr indices, ElementsAttr updates) {
+  return fromWideNums(input.getType(), [&](MutableArrayRef<WideNum> dst) {
+    // numpy implementation:
+    //
+    //   dst = np.copy(input)
+    //   outer = indices.shape[:-1]
+    //   for idx in np.ndindex(outer):
+    //     dst[indices[idx]] = updates[idx]
+
+    readElementsWideNums(input, dst);
+    ArrayBuffer<int64_t> indicesBuffer = getElementsArray<int64_t>(indices);
+    auto indicesArray = castArrayRef<uint64_t>(indicesBuffer.get());
+    ArrayBuffer<WideNum> updatesBuffer = getElementsWideNums(updates);
+    ArrayRef<WideNum> updatesArray = updatesBuffer.get();
+
+    auto dataShape = input.getType().getShape();
+    auto indicesShape = indices.getType().getShape();
+    auto updatesShape = updates.getType().getShape();
+
+    int64_t indices_nd = indicesShape.back();
+    auto outer = indicesShape.drop_back();
+    int64_t n_slices = ShapedType::getNumElements(outer);
+    int64_t slice_size =
+        ShapedType::getNumElements(updatesShape.drop_front(outer.size()));
+    auto dataStrides = getDefaultStrides(dataShape);
+    auto sliceStrides = llvm::ArrayRef(dataStrides).take_front(indices_nd);
+
+    auto indicesIter = indicesArray.begin();
+    auto updatesIter = updatesArray.begin();
+    for (int64_t i = 0; i < n_slices; ++i) {
+      ArrayRef<uint64_t> idxs(indicesIter, indices_nd);
+      int64_t pos = getStridesPosition(idxs, sliceStrides);
+      std::copy_n(updatesIter, slice_size, dst.begin() + pos);
+      indicesIter += indices_nd;
+      updatesIter += slice_size;
+    }
+  });
+}
+
 ElementsAttr ElementsAttrBuilder::reduce(ElementsAttr elms,
     ArrayRef<unsigned> axes, bool keepdims,
     WideNum (*reducer)(WideNum, WideNum)) {
--- a/src/Dialect/ONNX/ElementsAttr/ElementsAttrBuilder.cpp
+++ b/src/Dialect/ONNX/ElementsAttr/ElementsAttrBuilder.cpp
@@ -502,31 +502,20 @@ ElementsAttr ElementsAttrBuilder::concat(
   });
 }
 
-namespace {
-void sliceImpl(ShapedType outputType, ArrayRef<int64_t> starts,
-    ArrayRef<int64_t> steps, ElementsAttr inputElements,
-    MutableArrayRef<WideNum> outputData) {
-  size_t rank = outputType.getRank();
-  auto outputShape = outputType.getShape();
-  auto inputStrides = getDefaultStrides(inputElements.getType().getShape());
-  ArrayBuffer<WideNum> inputBuffer = getElementsWideNums(inputElements);
-  const WideNum *start = inputBuffer.get().begin();
-  SmallVector<int64_t> stepsStrides;
-  for (size_t axis = 0; axis < rank; ++axis) {
-    start += starts[axis] * inputStrides[axis];
-    stepsStrides.push_back(steps[axis] * inputStrides[axis]);
-  }
-  for (auto &idxpos : StridesRange<1>(outputShape, {stepsStrides}))
-    outputData[idxpos.flattenedIndex] = *(start + idxpos[0]);
-}
-} // namespace
-
 ElementsAttr ElementsAttrBuilder::slice(ElementsAttr elms,
     ArrayRef<int64_t> shape, ArrayRef<int64_t> starts,
     ArrayRef<int64_t> steps) {
-  ShapedType outputType = elms.getType().clone(shape);
-  return fromWideNums(outputType, [&](MutableArrayRef<WideNum> dst) {
-    sliceImpl(outputType, starts, steps, elms, dst);
+  ShapedType outType = elms.getType().clone(shape);
+  return fromWideNums(outType, [&](MutableArrayRef<WideNum> dst) {
+    SmallVector<int64_t> strides;
+    ArrayBuffer<WideNum> src = getWideNumsAndStrides(elms, strides);
+    const WideNum *start = src.get().begin();
+    for (size_t axis = 0; axis < shape.size(); ++axis) {
+      start += starts[axis] * strides[axis];
+      strides[axis] *= steps[axis];
+    }
+    for (auto &idxpos : StridesRange<1>(shape, {strides}))
+      dst[idxpos.flattenedIndex] = *(start + idxpos[0]);
   });
 }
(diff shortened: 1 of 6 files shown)


src/Dialect/ONNX/ElementsAttr/ElementsAttrBuilder.cpp
```diff
@@ -450,6 +450,153 @@ std::vector<ElementsAttr> ElementsAttrBuilder::split(
...
+      size_t inputBlockLen = inputShape[axis] * postAxisNumElements;
+      SmallVector<int64_t> strides;
+      ArrayBuffer<WideNum> src = getWideNumsAndStrides(input, strides);
+      StridesRange<1> range(inputShape, {strides});
+      auto it = range.begin();
+      for (size_t offset = start; offset < dst.size(); offset += outBlockLen) {
+        for (size_t pos = 0; pos < inputBlockLen; ++pos) {
+          dst[offset + pos] = src.get()[it->pos[0]];
+          ++it;
+        }
```
tungld, line 497: Is it possible to do a memcpy for a block?

tungld: > that won't work in the cases where the values in input are not contiguous, namely when input is splat or a DisposableElementsAttr with non-default strides

Can we check `splat` or `non-default` strides? It looks like your DisposableElementsAttr can do it.

tungld: In case of `splat`, we can do `memset` starting from `dst[offset]`

sorenlassen: > Can we check splat or non-default strides? It looks like your DisposableElementsAttr can do it.

yes, we can test whether the ElementsAttr is splat or whether the strides describe a contiguous row-major layout, but it adds more conditional logic and it will require a lot of tests to get good test coverage

it it ok if I try to add these optimizations in a follow up PR? I can create an issue to track it

src/Transform/ONNX/ConstProp.cpp
```diff
@@ -781,79 +720,35 @@ Value ConstPropCast(
...
-    return nullptr;
-  }
+  auto outcome = shapeHelper.computeShape();
+  assert(succeeded(outcome) && "Failed to scan slice op parameters");
+  auto toLiterals = [](ArrayRef<IndexExpr> ies) {
+    return llvm::map_range(ies, [](IndexExpr ie) { return ie.getLiteral(); });
+  };
+  SmallVector<int64_t> shape(toLiterals(shapeHelper.getOutputDims()));
+  SmallVector<int64_t> starts(toLiterals(shapeHelper.starts));
+  SmallVector<int64_t> steps(toLiterals(shapeHelper.steps));
```
tungld, line 739: You can use `IndexExpr`, e.g.
```
SmallVector<int64_t> shape;
IndexExpr::getLiteral(shapeHelper.getOutputDims(), shape);
```

src/Dialect/ONNX/ElementsAttr/ElementsAttrBuilder.cpp
```diff
@@ -450,6 +450,153 @@ std::vector<ElementsAttr> ElementsAttrBuilder::split(
   return results;
 }
 
+ElementsAttr ElementsAttrBuilder::concat(
+    ArrayRef<ElementsAttr> elms, unsigned axis) {
+  assert(elms.size() >= 1 && "concat tensors must be non-empty");
+
+  ElementsAttr first = elms.front();
+  auto firstShape = first.getType().getShape();
```
tungld, line 458: Please use `ArrayRef<int64_t>` instead of auto to make it clear.

src/Dialect/ONNX/ElementsAttr/ElementsAttrBuilder.cpp
```diff
@@ -450,6 +450,153 @@ std::vector<ElementsAttr> ElementsAttrBuilder::split(
...
+  if (elms.size() == 1)
+    return first;
+
+  // Check elms are compatible and construct outShape:
+  SmallVector<int64_t> outShape(firstShape);
+  for (size_t i = 1; i < elms.size(); ++i) {
+    ElementsAttr next = elms[i];
+    assert(next.getElementType() == first.getElementType() &&
+           "concat tensors element types must agree");
+    auto nextShape = next.getType().getShape();
```
tungld, line 470: Please use `ArrayRef<int64_t>` instead of `auto`.

src/Dialect/ONNX/ElementsAttr/ElementsAttrBuilder.cpp
```diff
@@ -450,6 +450,153 @@ std::vector<ElementsAttr> ElementsAttrBuilder::split(
...
+    }
+    for (auto &idxpos : StridesRange<1>(shape, {strides}))
+      dst[idxpos.flattenedIndex] = *(start + idxpos[0]);
+  });
+}
+
+ElementsAttr ElementsAttrBuilder::gather(
+    ElementsAttr input, ElementsAttr indices, unsigned axis) {
+  ShapedType inputType = input.getType();
+  auto inputShape = inputType.getShape();
```
tungld, line 525: Please use `ArrayRef<int64_t>` instead of `auto`.

src/Dialect/ONNX/ElementsAttr/ElementsAttrBuilder.cpp
```diff
@@ -450,6 +450,153 @@ std::vector<ElementsAttr> ElementsAttrBuilder::split(
...
+    // numpy implementation:
+    //
+    //   dst = np.copy(input)
+    //   outer = indices.shape[:-1]
+    //   for idx in np.ndindex(outer):
+    //     dst[indices[idx]] = updates[idx]
+
+    auto inputShape = input.getType().getShape();
+    auto indicesShape = indices.getType().getShape();
+    auto updatesShape = updates.getType().getShape();
```
tungld, line 566: Please use `ArrayRef<int64_t>` instead of `auto`.

src/Dialect/ONNX/ElementsAttr/ElementsAttrBuilder.cpp
```diff
@@ -450,6 +450,153 @@ std::vector<ElementsAttr> ElementsAttrBuilder::split(
...
+    ArrayBuffer<WideNum> src = getElementsWideNums(input);
+    ArrayBuffer<int64_t> indicesArray = getElementsArray<int64_t>(indices);
+    size_t axisInputSize = inputShape[axis];
+    size_t inputBlockLen = axisInputSize * postAxisNumElements;
+    size_t outBlockLen = indicesArray.get().size() * postAxisNumElements;
+    size_t start = 0;
+    auto out = dst.begin();
+    for (int64_t idx : indicesArray.get()) {
+      int64_t adjustedIdx = idx < 0 ? idx + axisInputSize : idx;
+      auto in = src.get().begin() + adjustedIdx * postAxisNumElements;
```
tungld, line 544: Is it `int64_t` as the result type?

src/Dialect/ONNX/ElementsAttr/ElementsAttrBuilder.cpp
```diff
@@ -450,6 +450,153 @@ std::vector<ElementsAttr> ElementsAttrBuilder::split(
...
+    auto inputShape = input.getType().getShape();
+    auto indicesShape = indices.getType().getShape();
+    auto updatesShape = updates.getType().getShape();
+
+    int64_t indices_nd = indicesShape.back();
+    auto outer = indicesShape.drop_back();
+    int64_t n_slices = ShapedType::getNumElements(outer);
+    int64_t slice_size =
+        ShapedType::getNumElements(updatesShape.drop_front(outer.size()));
+    auto inputStrides = getDefaultStrides(inputShape);
```
tungld, line 573: What is the output type? The use of `auto` makes it difficult to understand/debug.

src/Dialect/ONNX/ElementsAttr/ElementsAttrBuilder.cpp
```diff
@@ -450,6 +450,153 @@ std::vector<ElementsAttr> ElementsAttrBuilder::split(
...
+  assert(elms.size() >= 1 && "concat tensors must be non-empty");
+
+  ElementsAttr first = elms.front();
+  auto firstShape = first.getType().getShape();
+  size_t rank = firstShape.size();
+  assert(axis < rank && "concat axis out of range");
+  if (elms.size() == 1)
+    return first;
+
+  // Check elms are compatible and construct outShape:
```
tungld, line 464: It looks to me, the check here is redundant and perhaps overlapped with the concat op's verifier. It could be ok if the use of this function is out of scope of ONNXConcatOp, but I am not sure when it happens.

src/Dialect/ONNX/ElementsAttr/ElementsAttrBuilder.cpp
tungld: Sure, it's totally fine.

src/Transform/ONNX/ConstProp.cpp
sorenlassen: done, thanks for the suggestion

I noticed that the getLiteral() implementation was missing a `const` qualifier for the first argument and went ahead and made IndexExpr::getLiteral/Shape/DynSymbols/OpOrFoldResults use ArrayRef like getValues

src/Dialect/ONNX/ElementsAttr/ElementsAttrBuilder.cpp
sorenlassen: I created issue #2197

src/Transform/ONNX/ConstProp.cpp
tungld: Thanks!

src/Dialect/ONNX/ElementsAttr/ElementsAttrBuilder.cpp
tungld: `foldOrEmitONNXSplit/Transpose/Squeeze/UnsqueezeOp` functions was created at the time when we didn't have a way to fold ops. I believe the functions will be finally replaced by folding. So if we need the asserts for these functions, I am fine to keep them for a while.

tungld:

src/Dialect/ONNX/ElementsAttr/ElementsAttrBuilder.cpp
tungld: Thank you!

tungld (approved):
LGTM!

PR #2192, opened 2023-04-25
Update dependencies to conform with LFX security scan

hamptonm1:
It looks like all the open Critical / High hits from the LFX Security Scan relate to the version of torch and pillow used by the requirement.txt file. Here is more information with regards to the security results from the scan:
https://security.lfx.linuxfoundation.org/#/a092M00001JWs71QAD/vulnerabilities 

I will be upgrading the following for compliance:
```
1. Torch 1.13.0
2. Torchvision 1.14.0
3. Pillow 9.0
```

### DIFF
--- a/docs/mnist_example/requirements.txt
+++ b/docs/mnist_example/requirements.txt
@@ -1,2 +1,2 @@
-torch>=1.9.1, <=1.11.0
-torchvision>=0.10.1, <=1.12.0
+torch~=1.13.0
+torchvision~=1.14.0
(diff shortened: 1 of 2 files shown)


hamptonm1:
It seems like `pillow` is an additional `python` package which is included upon installation. I figured to directly specify the version of pillow we need for the `python requirements.txt` file.

hamptonm1:
@AlexandreEichenberger Do you know if there is a way to run the LFX security scan manually? This way I can confirm that my updates will satisfy the scanner before merging.

AlexandreEichenberger:
I am not aware on how to do that.

hamptonm1:
> I am not aware on how to do that.

Okay no problem...we should upgrade regardless. I can always merge this PR and then check the LFX scan to confirm.

PR #2195, opened 2023-04-27
Use DimAnalysis in lowering MatMul

tungld:
This patch uses DimAnalysis to detect broadcasting when lowering `N-D x N-D` ONNXMatMul to krnl dialect. If there is no broadcasting, SIMD path will be used for the lowering.


AlexandreEichenberger (approved):
LGTM, thanks for adding this case

PR #2200, opened 2023-05-01
Shape Inference now succeeds for unimplemented ops

AlexandreEichenberger:
Before, calling shape inference on ops that were unimplemented failed. This caused an issue as some ONNX operations may be lowered after shape inference to other ONNX ops for which shape inference is known.

Thus the "default" shape inference for unimplemented ops `ONNXUnimplementedOpShapeHelper` now succeeds when calling `computeShape`. At times, we still need to know if compute shape is implemented (e.g. for Dim Analysis). Thus `isImplemented()` can be used on all shape inference helper to determine whether there is an actual implementation. In case of Dim Analysis, ops without `isImplemented() == true` are conservatively approximated.

I didn't implement a new pass to verify that all ops (after a certain point) have implemented shape inference. Instead, I just added an assert on using the actual results (namely `getOutputDims`); that way, even if an op is unimplemented but does not need shape inference, we can still perform the actual operations as normal.

### DIFF
--- a/src/Dialect/ONNX/ONNXDimAnalysis.cpp
+++ b/src/Dialect/ONNX/ONNXDimAnalysis.cpp
@@ -96,7 +96,8 @@ bool exploreSameInputDims(const onnx_mlir::DimAnalysis::DimT &dim,
   // Get its shape interface.
   onnx_mlir::ONNXOpShapeHelper *shapeHelper =
       shape_op.getShapeHelper(op, {}, nullptr, nullptr);
-  if (!shapeHelper)
+  // If no shape helper, or unimplemented, just abort.
+  if (!shapeHelper || !shapeHelper->isImplemented())
     return false;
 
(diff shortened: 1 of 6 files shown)


philass (approved):
Short and elegant! 

Thanks @AlexandreEichenberger!

philass:
Could we also get a lit test as well? Just a check that shape-inference for an unsupported op doesn't result in an error

AlexandreEichenberger:
> Could we also get a lit test as well? Just a check that shape-inference for an unsupported op doesn't result in an error

It is unfortunately not that easy, as it would fail only if we have code that tries lowering by using a call that gets the output shape from the shape helper. 

Is that ok if it remains untested?

philass:
Thats the exact lit test I had in mind!

Thanks @AlexandreEichenberger

PR #2205, opened 2023-05-03
Specify attributes for Krnl.Call

chentong319:
This PR adds another builder of krnl.Call. Instead of the boolean choice of adding or not adding all the attributes of an Op to its Krnl.Call, the new builder allows user to specify a list of attributes to be added. Therefore, when new attributes or user-defined attribute are added to an Op, the Krnl.Call can remain the same.
With this fix, the backend test cases for ResizeOp, disabled by PR #2156,  are enabled by this PR.
I also cleaned the code a little bit and added some comments.


src/Runtime/OMResize.inc
```diff
@@ -306,10 +306,13 @@ static void interpolate_nd_OMTensor(OMTensor *output_OMT, OMTensor *data,
...
 
+// The parameters that are not used are commented out.
+// I, Tong Chen, forgot why only the default value for them are used.
+// ToFix: add the support for other value if needed
 void Resize_Scales(
     OMTensor *output, OMTensor *data, OMTensor *scales,
-    char *coordinate_transformation_mode_str, float coeff_a,
-    int exclude_outside, float extrapolation_value, char *mode_str,
+    /* char *coordinate_transformation_mode_str, float coeff_a,*/
+    /* int exclude_outside, float extrapolation_value, */ char *mode_str,
```
AlexandreEichenberger, line 315: Would you be able to investigate with Blame who added them originally?

AlexandreEichenberger (commented):
LGTM, please find out why the param were added and if there are really needed.

src/Runtime/OMResize.inc
chentong319: I should be blamed for these code. The previous implementation was rewritten from a python implementation. I inherited the interface but not all the implementation.

AlexandreEichenberger: so should we just remove them?

chentong319:
I removed the unused parameters.

AlexandreEichenberger (approved):
LGTM, thanks for the updates

PR #2208, opened 2023-05-04
Bump to LLVM green commit from 2023-05-01

gargaroff:
Draft PR for the bump. Issue with the `FoldAdaptor::getData()` returning an `ArrayRef` must be resolved


gargaroff:
The pipeline fails with CMake failing to find LibEdit. I can't reproduce this on my local machine. Does anyone have an idea of what to do with this?

sorenlassen:
> The pipeline fails with CMake failing to find LibEdit. I can't reproduce this on my local machine. Does anyone have an idea of what to do with this?

not sure but maybe something to do with this https://reviews.llvm.org/D147153

gargaroff:
Thanks for the review!
I didn't even realize that `x.cast` was deprecated. Thanks for letting me know

philass (commented):
It looks like IBM's NNPA code isn't compiling.

You can iterate on it locally by building with the following command

```
  cmake -G Ninja \
        -DCMAKE_CXX_COMPILER=/usr/bin/c++ \
        -DMLIR_DIR=${MLIR_DIR} \
        -DONNX_MLIR_ENABLE_MHLO=OFF \
        -DONNX_MLIR_ACCELERATORS=NNPA \
        ..
  ```

**Comment by gargaroff** [comment]
I updated NNPA to fix the build issues there.

As for LibEdit, I set in the `Dockerfile.llvm-project` to disable LibEdit. I hope that helps, but I'm not sure if all build bots are using this Dockerfile as a base.

**Comment by gargaroff** [comment]
I'm now running into issues around the removal of `preloadDialectsInContext`. @imaihal I saw a TODO addressed to you in the code to make onnx work without relying on this. Could this be fast-tracked, such that the bump can be completed? At the moment this is blocking it.

**Comment by gargaroff** [comment]
@tungld @AlexandreEichenberger 
Thanks for trying to help move this forward. As @hamptonm1 probably pointed out to you, I am only allowed to work on this part-time and therefore don't have the resources to fix bigger TODOs that pop up as a consequence from bumping.

With this latest bump, MLIR has removed the option `preloadDialectInContext` from mlir-opt like tools. Details can be found under https://mlir.llvm.org/deprecation/ and https://discourse.llvm.org/t/psa-preloaddialectincontext-has-been-deprecated-for-1y-and-will-be-removed/68992.

I am not really sure what that option was doing under the hood, but the original description says the following:

```c++
/// - preloadDialectsInContext will trigger the upfront loading of all
///   dialects from the global registry in the MLIRContext. This option is
///   deprecated and will be removed soon.
```

You can trigger the same error as this bump PR is currently seeing by modifying [this line of code](https://github.com/onnx/onnx-mlir/blob/558a512d55e969c4c9e8edbef7f0afb59fd2f86c/src/Tools/onnx-mlir-opt/onnx-mlir-opt.cpp#L206) in onnx-mlir-opt to:

```c++
              allowUnregisteredDialects, /*preloadDialectsInContext*/ false,
```

From looking at the [commit that removed the option](https://github.com/llvm/llvm-project/commit/9c8db444bc85) it looks like this was telling MLIR to load every available dialect instead of only the dialects that have been explicitly mentioned as being required. I assume that you will have to go through every pass that is causing a crash and double-check the list of dependent dialects for that pass. This can be fixed independently of this bump on the main branch.

This is all the information I currently have about this issue. Let me know if you need more information or if you feel like this should be part of the bump. I'm open to open the discussion internally in our team as well.

tungld:
@gargaroff thank you for your effort and explanation! 

It's a pity that MLIR completely removed this flag and we have no transition time.

It looks to me the finally solution should be implementing `getDependentDialects` for every pass. And even when we do that, I am not sure whether it affects `DiposablePool` proposed by @sorenlassen, because `DiposablePool` looks to be relevant to Dialect. @sorenlassen might have more comments on `DiposablePool`.

Be said that, it might take time to update our code. Do you have any recommendation on a short term solution? In one of the threads you cited, they mentioned `If you will want to still preload dialects in the future, the only workaround will likely be to fork MlirOptMain.cpp in your own project.`. Do you think it is possible?

@imaihal please give your comments on this, it seems you did something about this code in the past. Thanks!

sorenlassen:
@tungld I think we can just register all the dialects up front, see PR #2233

gargaroff:
@AlexandreEichenberger Looks like this fixed the issue. I'm glad this turned out to be a super easy fix.

hamptonm1:
Now you can merge this in @gargaroff. Hard work paid off! 😊

gargaroff:
Thanks @hamptonm1 for raising this internally in your team. I'm glad we were able to start this regular rotation

hamptonm1:
> Thanks @hamptonm1 for raising this internally in your team. I'm glad we were able to start this regular rotation

All checks passed @gargaroff …. Do you have permissions to merge in this repository? If not I can merge it.

hamptonm1:
> @hamptonm1 Unfortunately I don't, so please go ahead and merge it

Sure I can merge it!
@AlexandreEichenberger Can we get him write access for this repo?

AlexandreEichenberger:
> Unfortunately I don't, so please go ahead and merge it

@gargaroff granted now. please use wisely.

AlexandreEichenberger:

PR #2212, opened 2023-05-04
Pow for onnx

AlexandreEichenberger:
There was an optimization that changed `Pow(x, 2)` into `Mul(x, x)` for NNPA accelerator; this is useful too for CPU as `Pow` is a function that is not simdized, whereas `Mul` is. 

So under the advice of Tong, I migrated that opt to canonizalize. The upper limit (64 for NNPA) remains at 64, but I wonder if this should be smaller and/or customizable per target.


src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
```diff
@@ -180,38 +180,12 @@ bool isUniBroadcatableFirstToSecond(Value A, Value B) {
 
 /// Check a value is defined by ONNXConstantOp or not.
 bool isDefinedByONNXConstantOp(Value v) {
+  // @Tung: why does it matters? Isn't a constant a constant?
```
sorenlassen, line 183: this is needed because `v.getDefiningOp() `returns null if v is a BlockArgument and `isa<ONNXConstantOp>(x)` crashes if x is null

but it's not needed if you use `isa_and_present`

src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -589,6 +597,87 @@ class RNNOpRewriteLayoutPattern : public OpRewritePattern<ONNXOp> {
...
+      // calculate pow(input,exponent) with "exponentiation by squaring" method
+      bool result_initialized = false;
+      while (exponent > 0) {
+        if (exponent & 1) {
+          result = result_initialized
+                       ? create.onnx.mul(resultType, result, input)
+                       : input;
+          result_initialized = true;
+        }
+        input = create.onnx.mul(resultType, input, input);
```
sorenlassen, line 646: would it be worthwhile avoiding the extra unused multiplication in the last loop iteration?

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
tungld: Yes, a block argument does not have a defining op, so it would be null.

src/Dialect/ONNX/Rewrite.cpp
tungld: @sorenlassen why is it unused? The number of `mul` must be the same as `exponent`

tungld (approved):
LGTM. Thanks for moving the code out of zhigh!

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
AlexandreEichenberger: Changing to isa_and_present as it would appear to cover more cases (all where its null)

src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -589,6 +597,87 @@ class RNNOpRewriteLayoutPattern : public OpRewritePattern<ONNXOp> {
...
+
+    Value result;
+    Type resultType = powOp.getZ().getType();
+    Type elementType = getElementTypeOrSelf(resultType);
+    if (exponent == 0) {
+      DenseElementsAttr valAttr;
+      if (elementType.isa<FloatType>())
+        valAttr = DenseElementsAttr::get(resultType, ArrayRef<float>({1.0}));
+      else if (elementType.isa<IntegerType>())
+        valAttr = DenseElementsAttr::get(resultType, ArrayRef<int64_t>({1}));
```
sorenlassen, line 632: I believe this will only work if elementType is F32 or I64, see https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/IR/BuiltinAttributes.h#L108-L120

I suggest you create a 1 attribute similar to `getIdentity()` in ConstProp.cpp:
```
Attribute one = isa<FloatType>(elementType)
  ? rewriter.getFloatAttr(elementType, 1.0)
  : rewriter.getIntegerAttr(elementType, 1);
result = create.onnx.constant(DenseElementsAttr::get(resultType, one));
```

src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -589,6 +597,87 @@ class RNNOpRewriteLayoutPattern : public OpRewritePattern<ONNXOp> {
...
+    // Test legality
+    if (!CanExpandPowOpToMul(powOp, exponent))
+      return failure();
+
+    // Rewrite
+    MultiDialectBuilder<OnnxBuilder> create(rewriter, loc);
+    Value input = powOp.getX();
+
+    Value result;
+    Type resultType = powOp.getZ().getType();
```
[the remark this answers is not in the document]
AlexandreEichenberger: Changed it to shaped type

AlexandreEichenberger:
Updated the code to do a `while(true) { .... if (cons) break ... }` pattern to avoid an unused mul. Added one more test.
Believe I replied/addressed all suggestions @tungld @sorenlassen . Thanks for the feedback

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
sorenlassen: please remove the comment

sorenlassen (approved):

AlexandreEichenberger:
Grrr, forgot to remove the comment. Good catch, will do.

src/Dialect/ONNX/Rewrite.cpp
tungld: I see. Good catch. Thanks for the explaination!

PR #2215, opened 2023-05-07
fix StridesIterator offsets to properly support negative values

sorenlassen:
Changed `pos` to `offsets` and the type from `size_t` to `int64_t` to match how negative values are used with negative strides in the case of negative steps in `ElementsAttrBuilder::slice()`


philass (approved):
LGTM!

Issue #2203, opened 2023-05-02
Add Support for ONNXCustomOp


chentong319:
The current ONNXCallOp is used to call a function defined within ONNX model. When the function is not defined, namely is an external function, ONNXCustomOp is used. It turned out the ONNXCustomOp is quite useful to represent special operations when onnx flow graph is manipulated without defining a particular new Op. Easy for experiment and use of library.
However,  currently there is no support of shape inference and lowering for ONNXCustomOp.

### Shape inference
Without the function body, we need extra info to perform shape inference. I think there are two possible approaches:

1. Handle the common cases. Define an attribute to describe a common pattern, such as multi-dimension broadcast, same shape with a list of inputs, same or a particular type for element type. We can write the shape inference for ONNXCustomOp based on this specification. Hopefully, the common pattern can cover our use cases.
2. Allow user to define particular shape inference function to associate it with the function name attribute in ONNXCustomOp. This approach is able to handle any case, but requires compiler modification.

### Call convention
I found that there are two ways to pass the tensor to library call:

1. Pass tensor as OMTensor data structure, which contains the data pointer, element type, and shape (dynamic) info.
2. Pass only the data pointer for a tensor, and user will explicitly pass the shape info as extra parameters. User can use ONNXSizeOp or ONNXShapeOp to retrieve the shape info.

An attribute for the call convention will be added to the ONNXCustomOp and will be used when ONNXCustomOp is lowered to llvm IR.

chentong319:
onnx has a set of shape inference rules and defined in the onnx model. For example, `TypeAndShapeInferenceFunction(propagateShapeAndTypeFromFirstInput)`. Such info is now used only in importer. We can inherit such info to onnx op.

chentong319:
### ONNX
onnx defines TypeAndShapeInferenceFunction for each onnx op.  onnx-mlir separates the type inference and shape inference.  The type inference is down in Builder and use the type string info imported from onnx. I think we should call the TypeAndShapeInferenceFunction in builder instead.

The basic utility for shape inference supported by onnx is prop  .  Common patterns are propagateShapeAndTypeFromFirstInput, bidirectionalBroadcastShapeInference, multidirectionalBroadcastShapeInference, and matmulShapeInference_opset_9. 

 ### ONNXRuntime
[...]

PR #2216, opened 2023-05-07
fix CustomFnTest

sorenlassen:
Reverts some changes from PR #1494 which broke CustomFnTest. Now allows ONNX ops with non-standard op_type to pass through to `ImportCustomNode()` in FrontendDialectTransformer.cpp where it can either translate to a function call to a registered schema or become an ONNXCustomOp.

Also changes `ONNXCustomOp::inferShapes()` to always succeed with no effect, similar to how shape inference for other ops succeeds with no effect when there is insufficient information to infer shapes.

With these changes the test now succeeds:
```
cmake --build . --target CustomFnTest && Debug/bin/CustomFnTest
echo $?
```

Also added parse lit test to test FusedMatMul from onnx through decomposition in CustomOpFuseMatMulPattern in Decompose.cpp.


src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -949,53 +949,38 @@ class FrontendGenImpl {
   }
 
   std::string GetImportVersionOfNode(const onnx::NodeProto &node) {
-    auto schema = GetOpSchema(node);
-    // Assume the top version
-    if (schema == nullptr) {
-      return std::string("");
-    }
```
sorenlassen, line 956: this seemed unnecessary

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -949,53 +949,38 @@ class FrontendGenImpl {
...
-      }
-    } else {
-      llvm::errs() << node.op_type();
-      if (op_dialect_top_version_map_.find(node.op_type()) !=
-          op_dialect_top_version_map_.end()) {
-        llvm_unreachable(
-            " this Op is not found in the onnx version being used");
-      } else {
-        llvm_unreachable(
-            " this Op is not supported by onnx-mlir's onnx dialect");
```
sorenlassen, line 998: these checks crashed before ImportCustomNode could process custom ops, like FusedMatMul, and non-standard ops with schemas, like TwiceFn in CustomFnTest

utils/gen_onnx_mlir.py
```diff
@@ -1441,7 +1432,6 @@ def main(args):  # type: (Type[Args]) -> None
                     op_def.write(r)
                     previous_name = schema.name
 
-    gen_op_new_version(op_importer, new_version_dict)
```
sorenlassen, line 1444: removed this because I couldn't see any other uses after I removed its use in GetImportVersionOfNode in FrontendDialectTransformer.cpp

chentong319:
The source of the error, I think, is that the version of CustomOp is not described in the gen_onnx_mlir.py. Could we fix the code from Line 1195 `void ImportNode(const onnx::NodeProto &node)`?  It is a custom op if `import_handler_map_.find(node.op_type())` is not found. The versionStr is only needed for non custom Op.
I can not remember the logic of the bulk of code you deleted. Not sure whether they can be deleted. Those code is used when we are partially moved to a new version of onnx.
For the shape inference, I am working on the PR to support the custom Op if the pattern is provided as an attribute. Currently, I will support the "SameAs" and "MDBroadcast".

chentong319:
The versionStr for custom op should always be empty string.

sorenlassen:
> The source of the error, I think, is that the version of CustomOp is not described in the gen_onnx_mlir.py. Could we fix the code from Line 1195 `void ImportNode(const onnx::NodeProto &node)`? It is a custom op if `import_handler_map_.find(node.op_type())` is not found. The versionStr is only needed for non custom Op.
> I can not remember the logic of the bulk of code you deleted. Not sure whether they can be deleted. Those code is used when we are partially moved to a new version of onnx.
> For the shape inference, I am working on the PR to support the custom Op if the pattern is provided as an attribute. Currently, I will support the "SameAs" and "MDBroadcast".

I'm aware that you're working on doing new things with CustomOp, as you described in issue #2203, but the current implementation in ImportCustomNode() was clearly intended to support the two uses exercised in CustomFnTest. I took pains to just roll back the changes from PR #1494 that broke the test and the logic in ImportCustomNode(), plus I added a parse lit test to demonstrate how FusedMatMul will work end-to-end from import of a `.onnx` model. I'm open to that we could do other things with CustomOp going forward, but let's first get the original logic in ImportCustomNode() working again as intended. Please take a close look at that test and see how it exercises the two code paths in ImportCustomNode(), namely with TwiceFn and TwiceTwiceFn as registered functions and CustomFn1 and CustomFn2 as non-functions.

I separated out the changes in this PR from my ongoing effort to get the function inlining from PR #1785 into a form that we can land.

> The versionStr for custom op should always be empty string.

yes, that should be the case after the changes in this PR

chentong319 (approved):
LGTM!

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -949,53 +949,38 @@ class FrontendGenImpl {
...
-        return std::string("");
-      for (int i = opset_list.size() - 1; i > 0; i--) {
-        if (current_opset < opset_list[i - 1]) {
-          LLVM_DEBUG(llvm::dbgs() << DEBUG_TYPE << ":   - use Opset "
-                                  << opset_list[i] << "\n");
-          return "V" + std::to_string(opset_list[i]);
-        }
-      }
-    } else {
-      llvm::errs() << node.op_type();
```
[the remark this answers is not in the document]
sorenlassen: yes, the `return ""` for that case now happens in line 963 above

PR #2219, opened 2023-05-08
Update opset 19 ops with only type changes

philass:


philass:
Hey @chentong319,

I created an issue to track the problem https://github.com/onnx/onnx-mlir/issues/2220. I believe @gongsu832 is taking a look at this.

hamptonm1:
@philass This should not be merged until the upgrade for ONNX 1.14.0 has been done. I am taking on this task of doing the upgrade but want the communities input first.

philass:
@hamptonm1 This PR does the upgrade as well

hamptonm1:
> @hamptonm1 This PR does the upgrade as well

There is more to it then what you have but @gongsu832 actually has a PR to do the upgrade of onnx 1.14.0 and update the backend test and then your PR should pass smoothly once his PR is merged first.

philass:
@chentong319 Could you take a look?

AlexandreEichenberger (commented):
@philass if we updated the version of ops to 19 but don't support the new op features for lowering to CPU, we should update the ==LIMIT== fields in the gen_onnx_mlir.pl, don't you think?

philass:
@AlexandreEichenberger we do support these ops in 19. We just don't have support for fp8 types

Do you think we could get this in? I'd like to move on to the more involved tasks for opset 19

sorenlassen (approved):
LGTM

AlexandreEichenberger:
@philass thanks for the update. We are coming with a revision of the ==MIN== and ==MAX== (which we may rename ==UNSUPPORTED==) which should also come with a clearer explanation of how to fill these fields.

Issue #2218, opened 2023-05-08
Is there any way to return multiple values by krnl.call op?


negiyas:
I am developing an onnx.Unique op lowering path by using krnl.call op.
It seems that only one output is supported by current onnx-mlir implementation of krnl.call ( https://github.com/onnx/onnx-mlir/blob/main/src/Dialect/Krnl/Krnl.td#L50-L71 )

Since the Unique op returns multiple values, (e.g. Y, indices, inverse_indices, counts) with different shapes, I want to call C function returning multiple values. Any methods to get return multiple values with different shapes are acceptable, and API of the callee C function can be changed to deal with the issue.

Are there any recommended (or suggested) ways to return multiple values by krnl.call op according to the current onnx-mlir implementation?

AlexandreEichenberger:
LSTM have multiple outputs... check how they implement returning multiple values
https://onnx.ai/onnx/operators/onnx__LSTM.html

negiyas:
@AlexandreEichenberger Thanks for the answer, and sorry for my ambiguous question. 
Since onnx.Unique op is being developed using krnl.call op calling a C function returning multiple values, but krnl.call does not support C function returning multiple values at now. I am happy if krnl.call op can return multiple values in any way. 
I suppose that we can call the C function multiple times to get one return value by each call, but this way may not be smart and efficient.

chentong319:
Krnl.Call does not support variadic output. I think it is doable to add that support. Krnl.Call add the return MemRef as parameter. We can discuss the detail if you'd like to add the code.

chentong319:
I can add the support after PR #2205 is merged. I am working on another PR may need some change for Krnl.Call. I can postpone that change after the support of multiple output, if it is urgent.

negiyas:
@chentong319 Thanks for the answer.
My request is not urgent. I am happy if you will support this function when it's good to support it.
> Krnl.Call does not support variadic output. 

Let me just confirm my issue. The argument number is not variadic, but optional arguments are included.
The following is the API of C function I want to use. I can change the API.
I hope that I can use this C function (or similar) with four (optional) arguments to store the output values. 

![image](https://github.com/onnx/onnx-mlir/assets/18544851/84e45943-7289-41da-8100-a240a42e7b66)

chentong319:
#2251 added the support for multiple return values.

PR #2222, opened 2023-05-08
Support onnx.Unique op lowering

negiyas:
Support onnx.Unique op lowering.
This PR includes an issue to use krnl.call returning multiple return values, which is not supported currently.
This issue is reported as Issue #2218 , and is need to be fixed to finalize this PR.

All TODOs were cleared as follows. 
[X] Fix issues to get multiple return values from krnl.call (Issue #2218).
[X] Prepare lit tests, and confirm the results
[X] Confirm the results with backend tests.


AlexandreEichenberger:
Thanks for the high level info

negiyas:

chentong319:
It is a common case that the size of the returned MemRef is determined by the library. The current limitation of krnl.Call forced you to get the size with the first krnl.call, and then get the result with the second krnl.call. It is fine for this PR and we can think of a better solution for future.
If my high-level understanding of this PR is correct, I will review the details of this PR.

chentong319:
Should emitArgUnique be added into Tensor/Unqiue.cpp, instead of ONNXToKrnlCommon?

negiyas:
@chentong319 Thanks for the comments.
Your comments are correct.  The first krnl.call is to get the size, and the second krnl.call is to get all necessary outputs.

negiyas:
> Should emitArgUnique be added into Tensor/Unqiue.cpp, instead of ONNXToKrnlCommon?



> Should emitArgUnique be added into Tensor/Unqiue.cpp, instead of ONNXToKrnlCommon?
Moved emitArgUnique  from ONNXToKrnlCommon to ensor/Unqiue.cpp, since emitArgUnique is not common to other operations.
Thanks.

src/Conversion/ONNXToKrnl/Tensor/Unique.cpp
```diff
@@ -0,0 +1,194 @@
...
+    ONNXUniqueOpShapeHelper shapeHelper(op, operands, &create.krnlIE);
+    Value X = operandAdaptor.getX();
+    ArrayRef<int64_t> xShape = getShape(X.getType());
+    Type elementType = X.getType().cast<MemRefType>().getElementType();
+    int64_t rank = create.krnlIE.getShapedTypeRank(X);
+    int64_t sorted = operandAdaptor.getSorted();
+    std::optional<int64_t> optionalAxis = uniqueOp.getAxis();
+    //
+    // Get axis value as a positive integer if axis attribute is specified.
+    // If set "-1" if the axis attribute is not specified.
```
chentong319, line 111: Remove the leading `If'

src/Conversion/ONNXToKrnl/Tensor/Unique.cpp
```diff
@@ -0,0 +1,194 @@
...
+  Type int_type = rewriter.getIntegerType(64);
+  Value val_axis = create.math.constant(int_type, axis);
+  Value val_sorted = create.math.constant(int_type, sorted);
+  if (count_only) {
+    SmallVector<Value, 4> operands = {total, input, val_axis, val_sorted};
+    rewriter.create<KrnlCallOp>(loc, "omTensorUniqueCount", 1, operands);
+  } else {
+    SmallVector<Value, 8> operands = {total, input, val_axis, val_sorted, Y,
+        indices, inverse_indices, counts};
+    rewriter.create<KrnlCallOp>(loc, "omTensorUnique", 8, operands);
```
chentong319, line 51: Is `8` the correct number for outputs?

src/Conversion/ONNXToKrnl/Tensor/Unique.cpp
```diff
@@ -0,0 +1,194 @@
...
+  LogicalResult matchAndRewrite(Operation *op, ArrayRef<Value> operands,
+      ConversionPatternRewriter &rewriter) const final {
+    ONNXTopKOpAdaptor operandAdaptor(operands, op->getAttrDictionary());
+    ONNXUniqueOp uniqueOp = llvm::cast<ONNXUniqueOp>(op);
+    Location loc = op->getLoc();
+    MultiDialectBuilder<KrnlBuilder, IndexExprBuilderForKrnl, MathBuilder,
+        MemRefBuilder>
+        create(rewriter, loc);
+    IndexExprScope scope(create.krnl);
+    ONNXUniqueOpShapeHelper shapeHelper(op, operands, &create.krnlIE);
```
chentong319, line 102: Why not call ShapeHelper.computeShape()?

src/Conversion/ONNXToKrnl/Tensor/Unique.cpp
```diff
@@ -0,0 +1,194 @@
...
+      outputIndexDims.emplace_back(totalDimExpr);
+      DimIndexExpr inputDimExpr = LiteralIndexExpr(xShape[0]);
+      for (int64_t i = 1; i < rank; i++) {
+        inputDimExpr = inputDimExpr * LiteralIndexExpr(xShape[i]);
+      }
+      outputInverseIndexDims.emplace_back(inputDimExpr);
+    } else {
+      for (int64_t i = 0; i < rank; i++) {
+        DimIndexExpr tDimExpr = LiteralIndexExpr(xShape[i]);
+        if (i == axis)
```
chentong319, line 149: To me, most the work should be done in computeShape. Only thing needed to be fixed here is to set the tDimExpr to the right dimension. You didn't do that way because there is no way to represent the unknown value (uniqueCount) with IndexExpr. Is that the reason?

src/Conversion/ONNXToKrnl/Tensor/Unique.cpp
```diff
@@ -0,0 +1,194 @@
...
+    Value inverse_indices =
+        create.mem.alignedAlloc(memrefType, outputInverseIndexDims);
+    Value counts = create.mem.alignedAlloc(memrefType, outputIndexDims);
+    //
+    // Emit a Unique call to get the outputs
+    //
+    create.krnl.store(iZero, uniqueCount, {});
+    emitArgUnique(rewriter, loc, uniqueCount, X, axis, /*sorted=*/sorted,
+        outputY, indices, inverse_indices, counts, /*count_only=*/false);
+    rewriter.replaceOp(op, {outputY, indices, inverse_indices, counts});
```
chentong319, line 184: Some of the outputs are optional. How do you handle that? Need to check the getResults().size()?

src/Conversion/ONNXToKrnl/Tensor/Unique.cpp
```diff
@@ -0,0 +1,194 @@
...
+    // Emit a Unique call to get the outputs to count unique
+    // slices(subtensors) of X along axis at first.
+    //
+    Type indexTy = rewriter.getIndexType();
+    Value iZero = create.math.constantIndex(0);
+    Value uniqueCount = create.mem.alloca(MemRefType::get({}, indexTy));
+    create.krnl.store(iZero, uniqueCount, {});
+    Value noneValue;
+    emitArgUnique(rewriter, loc, uniqueCount, X, axis, /*sorted=*/sorted,
+        noneValue, noneValue, noneValue, noneValue, /*count_only=*/true);
```
chentong319, line 129: I guess the noneValue will be translated into null pointer. Right?

negiyas: Yes. Correct.

src/Conversion/ONNXToKrnl/Tensor/Unique.cpp
negiyas: Types of optional outputs becomes noneType if they are not required, passed as NULL pointers in OMTensorUnique,
and OMTensorUnique prepares necessary outputs only.

src/Conversion/ONNXToKrnl/Tensor/Unique.cpp
negiyas: Done. Thanks!

src/Conversion/ONNXToKrnl/Tensor/Unique.cpp
negiyas: I changed the number from 8 to 5.  I changed the argument order of omTensorUnique to gather output arguments at the beginning. Thanks.

src/Conversion/ONNXToKrnl/Tensor/Unique.cpp
negiyas: > You didn't do that way because there is no way to represent the unknown value (uniqueCount) with IndexExpr. Is that the reason?

Probably yes according to my understanding.
This part generates code to calculate shapes of output buffers to be used for their buffer allocation before calling the second krnl.call according to the actual number of unique slices (uniqueCount), which is counted by the first krnl.call.  I am not very sure if the current way is good, and your comments are very welcome.

```
    //
    // Calculate shapes of output Tensors
    //
    Value total = create.krnl.load(uniqueCount, {});
    NonAffineIndexExpr totalDimExpr = DimIndexExpr(total);
    DimsExpr outputYDims;
    DimsExpr outputIndexDims;
    DimsExpr outputInverseIndexDims;
    if (axis < 0) {
      outputYDims.emplace_back(totalDimExpr);
      outputIndexDims.emplace_back(totalDimExpr);
      DimIndexExpr inputDimExpr = LiteralIndexExpr(xShape[0]);
      for (int64_t i = 1; i < rank; i++) {
        inputDimExpr = inputDimExpr * LiteralIndexExpr(xShape[i]);
      }
      outputInverseIndexDims.emplace_back(inputDimExpr);
    } else {
      for (int64_t i = 0; i < rank; i++) {
        DimIndexExpr tDimExpr = LiteralIndexExpr(xShape[i]);
        if (i == axis)
          tDimExpr = totalDimExpr;
        outputYDims.emplace_back(tDimExpr);
      }
      outputIndexDims.emplace_back(totalDimExpr);
      outputInverseIndexDims.emplace_back(LiteralIndexExpr(xShape[axis]));
    }
```

src/Conversion/ONNXToKrnl/Tensor/Unique.cpp
negiyas: Added hapeHelper.computeShape() to calculate the shape here.

chentong319:
I am almost done with the compiler part, except for the replacement for the variadic output. For the runtime part, could you add a high level description? Or we can schedule a meeting.

negiyas:
> I am almost done with the compiler part, except for the replacement for the variadic output. For the runtime part, could you add a high level description? Or we can schedule a meeting.

@chentong319 Thank you for reviewing this PR! Let me explain a high level description of the runtime part.

The basic onnx.Unique specification is https://github.com/onnx/onnx/blob/main/docs/Operators.md#Unique.
Let me explain the high level concepts of main functions used for implementing the runtime part.

### **How omTensorUnique is called**
As you know, this function called twice to process a single onnx.Unique op.
The first call just counts the unique number (int count), and the second call generates necessary outputs, such as indices, inverse_indices and counts. 

### **Descriptions of important functions** 
c.f. The following comments for descriptions of functions are inserted to codes of the functions, too.

**(A) getSliceData**
If the axis attr is given in onnx.Unique, it should compare slices according to the specified axis. This function is called to get a slice according to the given axis if the axis attr is given. This function gets slice data according to the specified slice axis(sliceAxis), and fills them to the specified pointer. Slice data's shape is the following.
- Input data: [x, y, z] (asumming sliceAxis= 1, the "y" axis is the slice.)
- Sliced data: [x, z, count] (s.t. count is numer of unique slice shape)

**(B) sliceTableRegister**
This function registers (1) a given slice returned by getSliceData if the axis attr is given or (2) a single element if the axis atter is not given. It compares the current slice with all of the registered slices to investigate an identical slice is registered or not, and returns a bool if it is found or not.  If the sorted attr is given, this function keeps slices in sorted order. Otherwise it keeps them in the coming order. 
This function also generates outputs of onnx.Unique, such as indices, inverse_indeces and counts. However it does not generate Y if the axis attr is specified, since this function simply keeps registered slices in the table. The output Y is generated from input Tensor, indices and axis in another function.

**(C) produceY**
This function produces output Y from input tensor, indices and sliceAxis if the axis attr is given and output Y is required.
If the axis attr is not given, sliceTableRegister generates the output Y directly, since the output is a simple array of unique elements.

**(D) omTensorUnique**
This function handles two cases, where are (1) case with no axis given (2) case with axis given.
-  case with no axis given
In this case, omTensorUnique gets one element in the input Tensor in turn, call sliceTableRegister using an element as a slice. The registered elements are used as the output Y.
- case with axis given
In this case, omTensorUnique gets a slice with getSliceData, registers it with getSliceData, The outputs indices, inverse_indices and counts are set by getSliceData. If the output Y is necessary, it is created by produceY.

negiyas:
> 1. My previous question: how to handle the variadic output? I saw you always replace UniqueOp with 4 outputs.

Oh, I did not understand your comments correctly.
Yes, the current code always calls omTensorUnique with all 4 outputs. I am fixing this issue now. Thanks for the comments.
 
>     2. One discussion (no need to change): can we use hash function of the char[] (not the data type) instead of comparison(isLessNum)?

Yes, I suppose that it's possible to introduce hash function of the char[] to reduce the comparison costs if they are not small.
The current code manages elements and all necessary outputs in sorted order to support the sort attribute of onnx.Unique, because all outputs (Y, indices and counts) need to be sorted with the element order if the sort attribute is specified.

To introduce hash function, the omTensorUnique function registers elements with hash function without considering sorting at first, then it can sort necessary outputs (e.g. with quick sort) with the element order at last before returning them. This way is a little bit complicated, but it should be faster if the number of elements is large.

negiyas:
@chentong319 I fixed this PR not to generate unnecessary outputs of onnx.Unique, and I suppose that I finished to apply all comments for this PR.

>  Since the backend tests have been passed, I did not go into the details of runtime library. The code seems good to me. 

I agree that it is difficult to verify the runtime code. Because the onnx.Unique specification (https://github.com/onnx/onnx/blob/main/docs/Operators.md#Unique) does not explicitly explain the element order of the outputs, all I can do is just to follow their examples and backend tests.

chentong319:
Can you add a lit test in which an UniqueOp  has less than 4 outputs?

negiyas:
@chentong319 Thanks for the comments.

> Can you add a lit test in which an UniqueOp has less than 4 outputs?

I suppose that test/mlir/onnx/onnx_lowering_with_canonicalize.mlir includes several tests with less than 4 outputs.
For example, the following test is returning %Y only. Please let me know if other tests are necessary.

```
func.func @unique_without_axis(%arg0: tensor<2x2xi64>) -> tensor<*xi64> {
  %Y, %indices, %inverse_indices, %counts = "onnx.Unique"(%arg0) : (tensor<2x2xi64>) -> (tensor<*xi64>, none, none, none)
  return %Y : tensor<*xi64>
}
```

chentong319 (approved):
LGTM!

AlexandreEichenberger:
@negiyas I see a few warnings, generated during the compilation for s390... Do you mind providing a new PR to remove them? That would be much appreciated...

```
In file included from /workdir/onnx-mlir/src/Runtime/OMUnique.c:15:
/workdir/onnx-mlir/src/Runtime/OMUnique.inc:110:12: warning: implicit conversion turns string literal into bool: 'const char[34]' to 'bool' [-Wstring-conversion]
    assert("Unsupported ONNX type in OMTensor");
    ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/assert.h:108:20: note: expanded from macro 'assert'
  ((void) sizeof ((expr) ? 1 : 0), __extension__ ({                     \
                   ^~~~  ~
In file included from /workdir/onnx-mlir/src/Runtime/OMUnique.c:15:
/workdir/onnx-mlir/src/Runtime/OMUnique.inc:110:12: warning: implicit conversion turns string literal into bool: 'const char[34]' to 'bool' [-Wstring-conversion]
    assert("Unsupported ONNX type in OMTensor");
    ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/assert.h:109:11: note: expanded from macro 'assert'
      if (expr)                                                         \
      ~~  ^~~~
In file included from /workdir/onnx-mlir/src/Runtime/OMUnique.c:15:
/workdir/onnx-mlir/src/Runtime/OMUnique.inc:154:11: warning: variable 'sliceStride' set but not used [-Wunused-but-set-variable]
  int64_t sliceStride = 1;
```

```
In file included from /workdir/onnx-mlir/src/Runtime/OMUnique.cpp:15:
/workdir/onnx-mlir/src/Runtime/OMUnique.inc:154:11: warning: variable 'sliceStride' set but not used [-Wunused-but-set-variable]
  int64_t sliceStride = 1;
```

Thanks

src/Runtime/OMUnique.inc
```diff
@@ -0,0 +1,422 @@
...
+  uint64_t dataSize = OM_DATA_TYPE_SIZE[dataType];
+
+  int count = 0;
+  sliceTable sliceTable;
+  if (sliceAxis < 0) { // if slice attribute is not specified
+    // manage the inputTensor as flatten one
+    uint64_t elementNum = 1;
+    for (int64_t i = 0; i < inputRank; i++)
+      elementNum *= inputShape[i];
+    void *sliceDataPtr = (YPtr == NULL) ? alloca(dataSize * elementNum) : YPtr;
```
sorenlassen, line 363: it seems unsafe to use alloca() for things of unbounded size

would it be better to change this to malloc()?

AlexandreEichenberger: @negiyas please address @sorenlassen point. For large data, `alloc` is safer. Please open a new PR for this. Thanks

PR #2223, opened 2023-05-08
Count the number of parameters in a model

tungld:
It looks like there is no information in a trained ONNX model to know how many learnable parameters are there in the model. Learnable parameters are constants stored in initializers of the model, but not all initializers are learnable parameters. 

It's often the case that most of initializers are learnable parameters, so we count the number of elements in all initializers to have a rough idea of how many learnable parameters in the mode.

`-v` will print out the number of elements, say `The ONNX model has ...`. For example, with GPT-2 model:
```sh
tungld@53315fb59bc7:~/dl/onnx-mlir/build$ ./Debug/bin/onnx-mlir ~/dl/models/GPT2/model.onnx -v
The ONNX model has 137022892 elements in its initializers. This value would be close to and greater than the number of parameters in the model. Because there is no way to exactly count the number of parameters, this value can be used to have a rough idea of the number of parameters in the model.
[/home/tungld/dl/onnx-mlir/build/]/home/tungld/dl/llvm-project/build/bin/opt: opt -O0 --mtriple=s390x-ibm-linux -o /home/tungld/dl/models/GPT2/model.bc /home/tungld/dl/models/GPT2/model.unoptimized.bc
[/home/tungld/dl/onnx-mlir/build/]/home/tungld/dl/llvm-project/build/bin/llc: llc -O0 --mtriple=s390x-ibm-linux -filetype=obj -relocation-model=pic -o /home/tungld/dl/models/GPT2/model.o /home/tungld/dl/models/GPT2/model.bc
[/home/tungld/dl/onnx-mlir/build/]/usr/bin/c++: c++ /home/tungld/dl/models/GPT2/model.o -o /home/tungld/dl/models/GPT2/model.so -shared -fPIC -L/home/tungld/dl/onnx-mlir/build/Debug/lib -lcruntime
Shared library /home/tungld/dl/models/GPT2/model.so has been compiled.
```


AlexandreEichenberger (approved):
LGTM, great new feature, integrating in the `-v` make perfect sense.

PR #2224, opened 2023-05-08
Resizev13 fix

philass:
Add decompose rule for ResizeV13

### DIFF
--- a/test/mlir/onnx/onnx_decompose.mlir
+++ b/test/mlir/onnx/onnx_decompose.mlir
@@ -268,15 +268,25 @@ func.func @test_resizev10(%arg0: tensor<1x2x3x4xf32>, %arg1: tensor<4xf32>) -> t
 
 // -----
 
-func.func @test_resizev11(%arg0: tensor<*xf32>, %arg1: tensor<*xi64>) -> tensor<*xf32> {
-  %0 = "onnx.NoValue"() {value} : () -> none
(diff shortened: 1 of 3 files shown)


test/mlir/onnx/onnx_decompose.mlir
```diff
@@ -268,15 +268,25 @@ func.func @test_resizev10(%arg0: tensor<1x2x3x4xf32>, %arg1: tensor<4xf32>) -> t
 
 // -----
 
-func.func @test_resizev11(%arg0: tensor<*xf32>, %arg1: tensor<*xi64>) -> tensor<*xf32> {
+func.func @test_resizev11(%arg0: tensor<*xf32>, %arg1: tensor<*xf32>, %arg2: tensor<*xf32>, %arg3: tensor<*xi64>) -> tensor<*xf32> {
+  %0 = "onnx.ResizeV11"(%arg0, %arg1, %arg2, %arg3) {coordinate_transformation_mode = "half_pixel", cubic_coeff_a = -7.500000e-01 : f32, exclude_outside = 0 : si64, extrapolation_value = 0.000000e+00 : f32, mode = "nearest", nearest_mode = "floor"} : (tensor<*xf32>, tensor<*xf32>, tensor<*xf32>, tensor<*xi64>) -> tensor<*xf32>
+  return %0 : tensor<*xf32>
```
tungld, line 273: Not related to this patch: I think some operations do conversions to the closest version instead of the latest version, for example, V11 to V13, V13 to the latest. That way would be easier when we have a new version, we only need to update the old latest instead of all versions to the new version.

tungld (approved):
LGTM. Thanks!

test/mlir/onnx/onnx_decompose.mlir
philass: Thats a great suggestion

PR #2226, opened 2023-05-08
Remove warning

AlexandreEichenberger:
There was one warning remaining on Linux:

```
/workdir/onnx-mlir/src/Transform/ONNX/Decompose.cpp:737:25: warning: implicitly-declared 'mlir::Pass::Option<std::__cxx11::basic_string<char> >& mlir::Pass::Option<std::__cxx11::basic_string<char> >::operator=(const mlir::Pass::Option<std::__cxx11::basic_string<char> >&)' is deprecated [-Wdeprecated-copy]
  737 |     this->target = pass.target;
```

I looked at other similar passes that have options, and none updated their options in the constructor. Thus removed the copy.

MHLO, since this is an option that is used by you I believe, please check that it works.
Thanks

### DIFF
--- a/src/Transform/ONNX/Decompose.cpp
+++ b/src/Transform/ONNX/Decompose.cpp
@@ -730,13 +730,16 @@ struct DecomposeONNXToONNXPass
   MLIR_DEFINE_EXPLICIT_INTERNAL_INLINE_TYPE_ID(DecomposeONNXToONNXPass)
 
   DecomposeONNXToONNXPass(const std::string &target) { this->target = target; }
-#if 1
+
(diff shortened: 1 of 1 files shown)


AlexandreEichenberger:
MHLO: @chenchongsong or colleagues, make sure that works for you; or we can put it down and if it stops working, we can revert it or find other means to use the options.

sorenlassen:
fwiw, I wrote a different fix in PR #2169 but abandoned it because I no longer saw the warning in the CI logs

AlexandreEichenberger:
I still see it when I compile in my virtual env... maybe my compiler is too old.
I like your solution too @sorenlassen; do you want to use yours?

sorenlassen:
solved with PR #2169

PR #2227, opened 2023-05-08
Add shape inference support for custom Op

chentong319:
As discussed in  #2203, custom op needs extra information for shape inference and lowering.

Extra attributes are added. All these attributes are optional to make them less intrusive.

**Attributes for shape inference**
- 'output_element_type': The element type for the output tensor. Optional. If not specified, follow the shape infer.
- 'shape_infer_pattern': Specify how to propagate the shape info from inputs to output. Supported patterns are `SameAs`, `MDBroadcast`.
- 'inputs_for_infer': which inputs are used for shape inference. When multiple inputs are specified for 'SameAs', all these inputs should have the same shape, and element type if output_type is not specified. When this optional attribute is not specified, all the inputs will be used for shape inference.

**Attributes for lowering**

- 'call_convention': specify how the tensor will be passed to external function.
    'OMTensor' will use OMTensor structure. refer to src/Runtime/OMTensor.inc. This data structure contains data pointer and shape info.
    'DataPointer' will pass only the data pointer. User has to manage the shape explicitly. Possibly, SizeOp or ShapeOp may be used and pass the result as extra parameters of CustomOp.
-  'output_position': To simply the memory management, the memory object for the output of CustomOp will be allocated by onnx-mlir, and then passed to the extern function to write.  This attribute specify the position of  the output. 0 means that  the output will be the first parameter in the function call, while  -1 means after the last.

Shape inference has been added and I am working on the  lowering.


AlexandreEichenberger:
Can you give a couple of examples of the shape inference patterns? What are the limitations, target patterns that you aim at solving.

chentong319:
> Can you give a couple of examples of the shape inference patterns? What are the limitations, target patterns that you aim at solving.

Two patterns are supported.
`SameAs`: the output has the same shape of the specified inputs. It is similar to ONNX's propagating shape and type. The UnaryOpShapeHelper is used for implementation.
'MDBroadcast`: the output is the multi-dimension broadcast of the specified inputs. This pattern follows ONNX specification of multi-dimension broadcast. The BroadcastOpShapeHelper is used for implementation.
The limitation is that other patterns are not supported directly. But user can add the pattern and its ShapeHelper.
I think the two pattern can cover most of the cases. Another pattern to support is flattened tensor for a given size (one of the input). Since the shape info is only used for memory allocation, only the size, not the shape, of the output really matters. We can use the dynamic size pattern to cover the rest. The dynamic size can be expression with math operation on tensor dialect.

sorenlassen:
I don't 100% understand how your ideas for custom ops will work all the way from importing an op from a .onnx model. In issue #2203 you write:
> the ONNXCustomOp is quite useful to represent special operations when onnx flow graph is manipulated without defining a particular new Op

It would be ideal if we can see how the custom ops will be imported in FrontendDialectTransformer and the new attributes will be set, with a test like CustomFnTest or a parse lit test like fusedmatmul.json in PR #2216.

src/Dialect/ONNX/AdditionalONNXOps.td
```diff
@@ -140,16 +140,51 @@ def ONNXCustomOp:ONNX_Op<"Custom",
...
+           specified, follow the shape infer.
+      'shape_infer_pattern':
+           Optional. Specify how to propagate the shape info from inputs to
+           output. Supported patterns are `SameAs`, `MDBroadcast`.
+      'inputs_for_infer':
+           which inputs are used for shape inference. When mulitple
+           inputs are specified for 'SameAs', all these inputs should
+           have the same shape, and element type if output_type is
+           not specified. When this optional attribute is not specified,
+           all the inputs will be used for shape inference.
```
AlexandreEichenberger, line 162: `inputs_for_infer`: what is the format?

src/Dialect/ONNX/AdditionalONNXOps.td
```diff
@@ -140,16 +140,51 @@ def ONNXCustomOp:ONNX_Op<"Custom",
...
+    Some attributes are introduced to control the lowering of CustomOp:
+      'call_convention':
+           specify how the tensor will be passed to external function.
+           'OMTensor' will use OMTensor structure. refer to
+           src/Runtime/OMTensor.inc. This data structure contains data pointer
+           and shape info.
+           'DataPointer' will pass only the data pointer. User has to manage
+           the shape explicitly. Possibly, SizeOp or ShapeOp may be used and
+           pass the result as extra parameters of CustomOp.
```
AlexandreEichenberger, line 171: How are the `SizeOp` or `ShapeOp` being transferred? Are they added to the function as parameter, and the user is responsible to add ONNX computations to compute them and  pass them?
If there is a scalar, say a 'float`, is an OMTensor created for it, or is it passed as a scalar? Is the call_convention a list (with one entry per input parameter)?

Having a couple of examples after the def of the inputs would further help.

src/Dialect/ONNX/ONNXOps/Additional/Custom.cpp
```diff
@@ -28,12 +28,40 @@ using namespace onnx_mlir;
 
 LogicalResult ONNXCustomOp::inferShapes(
```
AlexandreEichenberger, line 29: Is there any reason not to reuse the "ShapeHelper" infra? Since shape helpers are used, among others, for dynamic analysis, there would be great benefit in integrating it within that framework.

@tungld would you concur?

AlexandreEichenberger (commented):
There are still details to be ironed out and/or better explained.

A concrete example would be great. For example, could you do a `BlasGemm` example that would lower to a BLAS routine? Showing also that the attribute set that you propose would work well with the BLAS type interface would go a long way showing that this approach has all of the necessary support for such common interface.

```
/* Subroutine */ int dgemm_(char *transa, char *transb, integer *m, integer *
	n, integer *k, doublereal *alpha, doublereal *a, integer *lda, 
	doublereal *b, integer *ldb, doublereal *beta, doublereal *c, integer 
	*ldc)
```

Do you think that, in general, it is likely that all instances of a `CustomOp` (say `BlasGemm`) could share the same attributes for each instantiation?

Would it make sense to add 'auto-generated` parameters, such as the sizes of the I, J, K dims of a matrix? Or better use ONNX operations to add such computations?

If the goal is also to target things like "CustomOps" used by ORT, then we should show that we can now ingest their models without manual intervention (otherwise, we may as well transform the custom op in a function or a regular sequence of ONNX operation).

If that is not the goal, then better explaining what the actual target for this functionality would be good (and I think you are looking more for calling BLAS like functions). Then showing that this functionality works well with these would be useful.

chentong319:
> It would be ideal if we can see how the custom ops will be imported in FrontendDialectTransformer and the new attributes will be set, with a test like CustomFnTest or a parse lit test like fusedmatmul.json in PR https://github.com/onnx/onnx-mlir/pull/2216

The PR contains two lit test cases. I put one here. This test case is in onnx_shape_inference.mlir:
```
func.func @test_custom1(%arg0: tensor<1024xf32>, %arg1: tensor<4xf32>) -> tensor<*xf32> {
  %0 = "onnx.Custom"(%arg0, %arg1) {function_name = "testcall", shape_infer_pattern = "SameAs", inputs_for_infer = [1] } : (tensor<1024xf32>, tensor<4xf32>) -> tensor<*xf32>
  return %0 : tensor<*xf32>
}
```
The extra attributes specifies that the output is the same as the second input (inputs_for_infer=[1]).

Therefore `onnx-mlir-opt --shape-inference custom1.mlir` generates
```
module {
  func.func @test_custom1(%arg0: tensor<1024xf32>, %arg1: tensor<4xf32>) -> tensor<4xf32> {
    %0 = "onnx.Custom"(%arg0, %arg1) {function_name = "testcall", inputs_for_infer = [1], shape_infer_pattern = "SameAs"} : (tensor<1024xf32>, tensor<4xf32>) -> tensor<4xf32>
    return %0 : tensor<4xf32>
  }
}
```

How to add those attributes to Custom Op? To me, there are three ways:
1. If CustomOp is generated by certain onnx-mlir transformation, that transformation is responsible for adding the shape inference specification with attributes.
2. In onnx model, the custom ops could have the extra attributes as defined in this PR. I can try to modify the CustomFnTest to illustrate that. This approach may require change of converter. Better to push our specification of attribute to onnx community.   
3.  Provide a file (e.g. json file) to specify  the extra attributes for custom ops (indexed with function name). Such info can be used to modify either onnx model (.onnx file) or onnx-mlir (.mlir file) ( as suggested by @philass at the meeting) 

This PR is just the initial step to specify what attributes may be defined and how they are used by shape inference. I may not include the support of lowering to function call in this PR: too much content in one PR.

I'd like to point out several of my design goals:
1. Support the common cases without changing the compiler.  Extra info, other than just function name, is needed for shape inference in onnx-mlir.  It is Okay to provide extra info from data file or command line. Best to encode the info into .onnx file. We may have to develop tools to manipulate the .onnx file for this purpose.  
2. The extra attributes for shape inference or lowering are absolutely optional. I avoided using attribute with default value because it will change .mlir file.  onnx-mlir should work (with some limitation though) if they are not provided. For example, shape inference can just return success without doing anything. The bufferization can only handle output tensor with static shape ,since there is no way to generate the code the dynamic dimension.  Lowering to function call will just implicitly follow the presumed behavior.jj 
3. If the extension of onnx with custom Op is really important, they may be standardized and used by onnx community.

chentong319:
@AlexandreEichenberger Lots of your questions are on lowering to library call. Let's move our general discussion on this topic to issue #2203. This PR is just the initial step to support shape inference for CustomOp. I will remove the adding of the attributes for lowering from this PR and create another PR for lowering.

src/Dialect/ONNX/ONNXOps/Additional/Custom.cpp
chentong319: @AlexandreEichenberger I put in a new version that supports ShapeHelper interface. 
Though the UnaryOpShapeHelper is used, the semantics is similar to onnx's propagating shape info. I used attribute value `SameAs` for multiple input. The reason for multiple inputs is that we can bind the symbol of multiple inputs for dim analysis (@tungld ). It may not be a good idea to make UnaryOpShapeHelper a subclass of BroadcastOpShapeHelper.
The special case for CustomOp is that it behaves differently based on the attribute value of this Op. In this implementation, I used multiple inheritance and call the corresponding subclass based on the attribute value. In order to do that, the UnaryOpShapeHelper and BroadcastOpShapeHelper has to be virtual and constructor has to initialize all the subclasses explicitly.

src/Dialect/ONNX/AdditionalONNXOps.td
```diff
@@ -140,16 +140,51 @@ def ONNXCustomOp:ONNX_Op<"Custom",
...
+      'output_position':
+           To simply the memory management, the memory object for the output
+           of CustomOp will be allocated by onnx-mlir, and then passed to the
+           extern function to write.  This attribute specify the position of
+           the output. 0 means that  the output will be the first
+           parameter in the function call, while  -1 means after the last.
```
AlexandreEichenberger, line 177: is `0` or `-1` the only allowed values? If so, say so in the text. If not, add a "for example". Also use the standard text "positive values are index positions from the left hand side; negative values are from the right hand side" type of comments.

src/Dialect/ONNX/AdditionalONNXOps.td
AlexandreEichenberger: Please provide an answer.

src/Dialect/ONNX/ONNXOps/Additional/Custom.cpp
```diff
@@ -24,6 +24,25 @@ using namespace onnx_mlir;
...
+  if (!getShapeInferPattern().has_value()) {
+    // No shape inference pattern provided. Do not know how to infer shape
+    return success();
+  }
+
+  std::optional<ArrayAttr> inputIndexAttrs = getInputsForInfer();
+  // Use the first input because their element type should be same
+  int64_t inputIdx = 0;
+  if (inputIndexAttrs.has_value())
+    (inputIndexAttrs->getValue()[0]).cast<IntegerAttr>().getInt();
```
AlexandreEichenberger, line 41: shouldn't the above statement's getInt value be assigned to a variable?

src/Dialect/ONNX/ONNXOps/ShapeHelper.cpp
```diff
@@ -750,6 +750,58 @@ void resetTypesShapeToQuestionmarks(Operation *op) {
...
+    bool hasUniBroadcasting)
+    : ONNXOpShapeHelper(op, operands, ieBuilder, scope),
+      ONNXUnaryOpShapeHelper(op, operands, ieBuilder, scope),
+      ONNXBroadcastOpShapeHelper(op, operands, ieBuilder, scope),
+      operandsVector() {
+  ONNXCustomOp customOp = cast<ONNXCustomOp>(op);
+  if (!customOp.getShapeInferPattern().has_value()) {
+    pattern = 0;
+    return;
+  } else if (customOp.getShapeInferPattern() == "SameAs") {
```
AlexandreEichenberger, line 768: recommended format: no `else` after an `if() {... return ... }`

src/Dialect/ONNX/ONNXOps/ShapeHelper.cpp
```diff
@@ -750,6 +750,58 @@ void resetTypesShapeToQuestionmarks(Operation *op) {
...
+  ValueRange usedOperands;
+  for (auto indexAttr : inputIndexAttrs.value()) {
+    operandsVector.push_back(inputs[indexAttr.cast<IntegerAttr>().getInt()]);
+  }
+  ONNXOpShapeHelper::operands = ArrayRef(operandsVector);
+}
+
+LogicalResult ONNXCustomOpShapeHelper::computeShape() {
+  if (pattern == 1) {
+    return ONNXUnaryOpShapeHelper::computeShape();
```
AlexandreEichenberger, line 798: When the "SameAs" pattern was defined, I thought that there was an option to define the index of the input it was the same as. Is this supported here?

src/Dialect/ONNX/AdditionalONNXOps.td
```diff
@@ -140,16 +140,51 @@ def ONNXCustomOp:ONNX_Op<"Custom",
...
     The number of inputs and outputs can vary.
-
-    This operation is not part of the standard and was added to assist onnx-mlir.
+    Since the semantics of this op is not defined to compiler, some attributes
+    are introduced to guide shape inference:
+      'output_type':
+           Optional. The element type for the output tensor. Optional. If not
+           specified, follow the shape infer.
+      'shape_infer_pattern':
+           Optional. Specify how to propagate the shape info from inputs to
```
AlexandreEichenberger, line 155: This param is reported as optional here. But in shape inference, there seems to be an assert (line 773). 
We are now also supporting a new approach where shape inference is always fine (aways return success) because we may do shape inference and then change an op to something else. It's only when the shape inference result is actually used that we test if it is implemented. Unimplemented shape inference are tested using the 
```
  virtual bool isImplemented() { return true; }
```
which is default true, but should obviously be redefined here to satisfy the condition that `shape_infer_pattern` is defined, since it appears to be necessary to lowering.

src/Dialect/ONNX/AdditionalONNXOps.td
AlexandreEichenberger: this option does not seems to be used in shape inference.

AlexandreEichenberger (commented):
new batch of comments, thanks for refactoring shape inference to use shape helper.

sorenlassen:
> Let's move our general discussion on this topic to issue #2203.

before we decide on approving and merging this PR can we work through an example in issue #2203 of what we want to do end to end for custom ops for an example model, e.g. gpt2?

src/Dialect/ONNX/ONNXOps/ShapeHelper.cpp
AlexandreEichenberger: In fact, there is no need to do double inheritance. You can have the custom op as a child of Broadcast shape inference, as the call that you really need is this one
```
LogicalResult ONNXOpShapeHelper::setOutputDimsFromOperand(
    Value operand, int n, bool refineShape) {
```
which is defined on the parent class already. So you can simply pick up one of the "id" that reflect the same as shape, andset the output shape using this. And when its is a broadcast, then use the super (which is broadcast shape inf).

src/Dialect/ONNX/AdditionalONNXOps.td
chentong319: This option is intended to specify which inputs are used in shape inference, and is used to construct operands for ShapeHelper. For example, the broadcast pattern may be just on the first two input tensors. Then inputs_for_infer = [0, 1], and the other inputs will not be used in shape inference.

chentong319:
> Do you think that, in general, it is likely that all instances of a CustomOp (say BlasGemm) could share the same attributes for each instantiation?

In common case, the instances of a CustomOp with the same function_name is supposed to have the same attributes, though the attributes can be different.  The shape inference and lowering passes just works as told be the attributes.

src/Dialect/ONNX/ONNXOps/ShapeHelper.cpp
chentong319: The reason for double inheritance is that the computeShape() is a pure virtual function. I think that our code pattern is always call computeShape() first. And the computeShape for UnaryOp is just setOutputDimsFromOperand(operands[0]).
My early version of this PR did not use computeShape() and exposed the logic of computeShape in the code. I was asked by @AlexandreEichenberger to follow the existing code pattern. Now should I just change back to my previous implementation?
Another point is that the SameAs is not the same as UnaryOpShapeHelper.  The SameAs with more than one inputs could choose to propagate shape info from the best one and allow dim symbolic analysis to unify symbols. 
@AlexandreEichenberger @tungld Should we schedule a meeting to discuss this topic?

src/Dialect/ONNX/AdditionalONNXOps.td
chentong319: SizeOp or ShapeOp could be explicitly added to function as parameter by user. We can start with this approach. An alternative approach is to define another attribute to guide compiler to add extra parameter for SizeOp or ShapeOp..  
In current krnl.call lowering, scalars are passed by value. The call convention string is for all parameters, and we can define the call convention on each parameter if needed.

src/Dialect/ONNX/AdditionalONNXOps.td
chentong319: I'd like customOp to pass shape inference if shape inference pattern is not provided, and only get assert error when tensor with dynamic shape is to be allocated at lowering. This check should be in lowering Onnx.Call to krill.call. 
Back to shape inference, customOp may behave as three types, ONNXUnaryOpShapeHelper, ONNXBroadcastOpShapeHelper and ShapeHelper that simply pass (I do not see this type of subclass of ShapeHelper yet). In this PR, I will remove that assertion.

src/Dialect/ONNX/AdditionalONNXOps.td
```diff
@@ -140,16 +140,51 @@ def ONNXCustomOp:ONNX_Op<"Custom",
...
+    is a string which names the operation, other inputs are passed to the user
+    operation.
     The number of inputs and outputs can vary.
-
-    This operation is not part of the standard and was added to assist onnx-mlir.
+    Since the semantics of this op is not defined to compiler, some attributes
+    are introduced to guide shape inference:
+      'output_type':
+           Optional. The element type for the output tensor. Optional. If not
+           specified, follow the shape infer.
```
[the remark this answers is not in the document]
chentong319: The code for inferShape is using computeShapeAndUpdateType. Currently, the shape of the first input is used, but the element type is the given by attribute if specified.
```
  ONNXCustomOpShapeHelper shapeHelper(getOperation(), {});
```

src/Dialect/ONNX/AdditionalONNXOps.td
AlexandreEichenberger: please address comment above

src/Dialect/ONNX/AdditionalONNXOps.td
AlexandreEichenberger: please respond to the comment above

AlexandreEichenberger (commented):
@chentong319 please respond  to the comments I left in the code? First 3 were not addressed; will review once they are answered, Thanks.

src/Dialect/ONNX/ONNXOps/ShapeHelper.cpp
```diff
@@ -810,6 +810,55 @@ void resetTypesShapeToQuestionmarks(Operation *op) {
     resetTypeShapeToQuestionmarks(op->getResult(i));
 }
 
+//===----------------------------------------------------------------------===//
+// ONNX Custom Op Shape Helper
+//===----------------------------------------------------------------------===//
+
+ONNXCustomOpShapeHelper::ONNXCustomOpShapeHelper(Operation *op,
```
AlexandreEichenberger, line 817: This code does not respect the protocol, and will make this shape helper not compatible with other analysis phases. The protocol is as follows:

constructor -> save all the info you need from the op; the constructor cannot have any additional parameters
computeShape -> perform the computation, cannot accept any additional parameters.

In the current form, you compute the shape inside the constructor... and provide no new function for computeShape.

Here is the refactoring that you need to perform;
save `pattern` as you do
update the list of operands. your ToFix comment is correct; if you update the list of operands, you need to see how this is done in the constructor of OnnxShapeHelper (ShapeHelper.cpp, line 97).

then provide a custom `computeShape` which will use `pattern` and the (possibly pruned) list of operands, and will use one of the two shapes depending on pattern.

AlexandreEichenberger (changes requested):
Please update the CustomOp shape helper to comply with the required usage pattern.

src/Dialect/ONNX/ONNXOps/ShapeHelper.cpp
chentong319: 1. ShapeHelper::privateOperandsCache can be used to replace the ONNXCustomOpShapeHelper::operandsVector. However, they are implementation support of their own class. To combine them, ShapeHelper::privateOperandsCache has to be read/written. And ShapeHelper::operands will become invalid when the privateOeprandsCache is cleared to construct the new Ops. I'd like to keep the code as it is. If you insist, I will change it.
2. computeShape is already there.

chentong319: Added setOperands to use privateOperandsCache
computeShape is already there.

src/Interface/ShapeHelperOpInterface.hpp
```diff
@@ -137,6 +137,9 @@ struct ONNXOpShapeHelper {
   IndexExprScope *getScope() { return scope; }
   mlir::Operation *getOp() { return op; }
 
+  // Set the operands with a vector of Value
+  void setOperands(std::vector<mlir::Value>);
```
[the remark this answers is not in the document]
chentong319: Fixed.

src/Dialect/ONNX/AdditionalONNXOps.td
```diff
@@ -1,4 +1,4 @@
-// SPDX-License-Identifier: Apache-2.0
+
```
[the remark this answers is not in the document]
chentong319: fixed

AlexandreEichenberger (approved):
LGTM, thanks for all the changes

PR #2228, opened 2023-05-08
Splitv13 decomp

philass:
It seems I forgot to add a decomposition for SplitV13 when I upgraded Split for opset 18


sorenlassen (approved):
LGTM

PR #2232, opened 2023-05-10
fix resultTypeInference for constant

chentong319:
resultTypeInference is used in Builder. At this stage, ONNXConstantOp has not been normalized and may have different kinds of attributes. This PR adds the handling for other attributes. The ArrayAttr is not TypedAttr and has to create the type explicitly.


utils/gen_onnx_mlir.py
```diff
@@ -447,6 +447,26 @@
         resultTypes.push_back(attr.cast<TypedAttr>().getType());
       } else if (auto attr = getSparseValueAttr()) {
         resultTypes.push_back(attr.cast<TypedAttr>().getType());
+      } else if (auto attr = getValueFloatAttr()) {
+        resultTypes.push_back(attr.cast<TypedAttr>().getType());
```
sorenlassen, line 451: ONNXConstantOp must have tensor type, so I think you should create a scalar tensor and, moreover, the type should be f32 as for value_floats below
```
RankedTensorType::get({}, FloatType::getF32(attr.getContext()))
```

something similar applies to value_int and value_string

utils/gen_onnx_mlir.py
```diff
@@ -447,6 +447,26 @@
...
+      } else if (auto attr = getValueFloatsAttr()) {
+        SmallVector<int64_t> shape;
+        shape.push_back(attr.size());
+        resultTypes.push_back(RankedTensorType::get(shape, FloatType::getF32(attr.getContext())));
```
sorenlassen, line 455: nit: a simpler way to construct the shape is
```
int64_t size = attr.size();
RankedTensorType::get({size}, ...);
```

sorenlassen (approved):
LGTM

utils/gen_onnx_mlir.py
chentong319: Agree.

PR #2237, opened 2023-05-11
Shape inference continues when Slice's axes has not yet been a constant

tungld:
I see this pattern in a GPT-2 model generated from HuggingFace:
```
%155 = onnx.Constant dense<2> : tensor<1xi64>
%348 = "onnx.Identity"(%155) {onnx_node_name = "Identity_189"} : (tensor<1xi64>) -> tensor<*xi64>
%675 = "onnx.Slice"(%379, %672, %673, %348, %674) {onnx_node_name = "Slice_516"} : (tensor<*xui8>, tensor<*xi64>, tensor<*xi64>, tensor<*xi64>, tensor<1xi64>) ->     tensor<*xui8>
```
where `Identity` of a constant is imported and Slice's shape inference failed because it requires axes (`%348`) to be a constant.

This patch makes Slice's shape inference continue when axes is not a constant. Once `Identity` is removed by canonicalization, the shape inference will success.

A side note: here onnx-mlir imported `Identity` with an unranked output type, but we can import it using the input type.


sorenlassen:
> A side note: here onnx-mlir imported `Identity` with an unranked output type, but we can import it using the input type.

does it help to use `-useOnnxModelTypes`?

sorenlassen (approved):
LGTM

tungld:
> does it help to use -useOnnxModelTypes? 

Yes, it may help depending if the op has ranked type or not.

PR #2238, opened 2023-05-11
Use math.erf

tungld:
MLIR's math dialect has Erf function that supports vector, and it is faster than the current Taylor-based approximation in onnx-mlir:

With -O3, for a tensor of 16384 elements:
- Taylor-based approximation Erf (current implementation in onnx-mli): 0.00026 (s)
- `math.erf`: 0.00018 (s)


src/Dialect/Mlir/VectorMachineSupport.cpp
```diff
@@ -135,6 +135,7 @@ int64_t Z16VectorMachineSupport::getVectorLength(
     case GenericOps::ConversionGop:
     case GenericOps::CopySignGop:
     case GenericOps::DivGop:
+    case GenericOps::ErfGop:
```
tungld, line 138: @AlexandreEichenberger please check if this place for erf is good or not. Thanks!

tungld:
@AlexandreEichenberger this patch is ready. Please let me know if you need more information. Thanks!

AlexandreEichenberger (approved):
LGTM, tx for the good find.

PR #2240, opened 2023-05-11
make -pattern-shape-inference=true the default

sorenlassen:
I propose that we switch our default to use pattern based shape inference, which reached parity with the older implementation a while ago. Moreover, I propose that we remove the old implementation (as in PR #2243) in a month or so, provided pattern based shape inference creates no problems in the wild.


chentong319:
LGTM!

AlexandreEichenberger (approved):
LGTM, please make sure to remove in a month as we don't want too much dead code.

PR #2241, opened 2023-05-11
don't treat empty tensor as none

sorenlassen:
Follow up to PR #2133 to distinguish `none` from empty tensors.

Also removed unnecessary `EmitInitializerForInputTensor()` function and cleaned up the use of namespaces in FrontendDialectHelper.cpp.

### DIFF
--- a/src/Builder/FrontendDialectHelper.cpp
+++ b/src/Builder/FrontendDialectHelper.cpp
@@ -12,20 +12,24 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include "src/Builder/FrontendDialectHelper.hpp"
+
+#include "mlir/IR/BuiltinAttributeInterfaces.h"
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/Support/Endian.h"
 #include "llvm/Support/MemoryBuffer.h"
 #include "llvm/Support/Path.h"
 #include "llvm/Support/SwapByteOrder.h"
 
-#include "src/Builder/FrontendDialectHelper.hpp"
 #include "src/Dialect/ONNX/ElementsAttr/Arrays.hpp"
 #include "src/Dialect/ONNX/ElementsAttr/BType.hpp"
 #include "src/Dialect/ONNX/ONNXOps/OpHelper.hpp"
 #include "src/Dialect/ONNX/OnnxElementsAttrBuilder.hpp"
 #include "src/Support/FloatingPoint16.hpp"
 
-// TODO: put everything in namespace onnx_mlir, and be using namespace mlir
+using namespace mlir;
+
+namespace onnx_mlir {
 
 namespace {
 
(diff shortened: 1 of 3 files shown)


chentong319:
How will the empty tensor be treated after importer?

sorenlassen:
> How will the empty tensor be treated after importer?

I believe empty tensors work in most situations and most tensor operations operate on them, e.g. if you Add two tensors and one is empty and the other broadcasts to its shape (e.g. is scalar) then the result is an empty tensor

sorenlassen:
> According to my memory, empty tensor was used for the not-provided optional input for onnx Ops.

I changed all that to use 'none' of NoneType everywhere in PR #2133 but I overlooked this builder code.

> For AddOp, the input is not optional, and how to broadcast an empty tenor?

Right, empty tensor should not be used to represent that an optional parameter is absent. Empty tensors are valid tensors, in general. The rules for broadcast etc support empty tensors. Zero dim size works just like other dim sizes. It's the same in numpy, pytorch, etc. E.g. the result of the following is a tensor `sum` with shape `(0,)` and `np.array_equal(sum, empty)` is True:
```
import numpy as np
empty = np.arrray([])
scalar = np.array(42)
sum = np.add(empty, scalar)
```

sorenlassen:
> According to my memory, empty tensor was used for the not-provided optional input for onnx Ops.

I changed all that to use 'none' of NoneType everywhere in PR #2133 but I overlooked this builder code.

> For AddOp, the input is not optional, and how to broadcast an empty tenor?

Right, empty tensor should not be used to represent that an optional parameter is absent. Empty tensors are valid tensors, in general. The rules for broadcast etc support empty tensors. Zero dim size works just like other dim sizes. It's the same in numpy, pytorch, etc. E.g. the result of the following is a tensor `sum` with shape `(0,)` and `np.array_equal(sum, empty)` is True:
```
import numpy as np
empty = np.arrray([])
scalar = np.array(42)
sum = np.add(empty, scalar)
```

AlexandreEichenberger:
Can we document somewhere what the policy should be? And maybe document it somewhere where its easily found?

Optional not present:
empty tensor:
tensor of empty shape:

The last one comes from the specs, for example for clip
> Minimum value, under which element is replaced by min. It must be a scalar(tensor of empty shape).
it seems that scalars are tensor of empty shapes...

chentong319 (approved):
LGTM!

sorenlassen:
> Can we document somewhere what the policy should be? And maybe document it somewhere where its easily found?

@AlexandreEichenberger Is it ok if I merge this PR as is and create an issue to improve the documentation?

FWIW, the three things you mention are now completely separate things, given the changes in this PR and in PR #2133 to use 'none' of NoneType everywhere to represent optional-not-present in mlir:

> Optional not present:

In the `.onnx` protobuf representation: "An empty string may be used in the place of an actual argument's name to indicate a missing argument. Trailing optional arguments (those not followed by an argument that is present) may also be simply omitted." See https://github.com/onnx/onnx/blob/main/docs/IR.md#optional-inputs-and-outputs which is referenced in ops with optional inputs/outputs, e.g. https://github.com/onnx/onnx/blob/main/docs/Operators.md#gemm

In mlir we now use 'none' of NoneType to represent optional inputs/outputs everywhere.

> empty tensor:

An empty tensor (with a zero dim size) is a valid tensor, like in numpy and pytorch etc.

> tensor of empty shape:

Is a scalar. (As discussed elsewhere, unfortunately the onnx spec sometimes specifies that scalars should be represent this way and sometimes specifies that scalars should be represented as rank 1 tensors with dim size 1. Therefore we accept either representation of scalars for some ops.)

AlexandreEichenberger (approved):
LGTM, thanks for the explanations of the differences between none/empty/empty shape

Issue #1841, opened 2022-11-07
Support for model-local functions?


gramalingam:
ONNX supports model-local functions: these are functions defined within a model. (See: https://github.com/onnx/onnx/blob/main/onnx/onnx.in.proto#L411 ). It would be great if the ONNX-MLIR converted supported translation of these functions into MLIR.

The current translator does support translation of ONNX FunctionProto into a MLIR function (see: https://github.com/onnx/onnx-mlir/blob/main/src/Builder/FrontendDialectTransformer.cpp#L1053 ). However, this is currently used only for schema-defined functions (these are ONNX ops registered as functions in the schema). Hopefully extending this to handle model-local functions won't be much work.

I think this would require extending `ImportONNXModel` which currently translates only the top-level graph of the model ( https://github.com/onnx/onnx-mlir/blob/main/src/Builder/FrontendDialectTransformer.cpp#L81 ), to also translated the functions contained in the model.

### Rationale

Model-local functions provide an extensibility mechanism. It allows users to use "custom ops" that can be defined as functions, and embed these function-definitions in the ONNX model (without requiring access to the schema-definitions of such custom ops, etc.)

### Examples

(a) See https://github.com/onnx/onnx/blob/main/onnx/test/cpp/parser_test.cc#L457 for a toy example.

(b) See https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/test/framework/function_test.cc#L76 (and other examples in the same file) for more examples.

sorenlassen:
model local functions support was added in PR #2242

Issue #1787, opened 2022-10-13
inline function calls during onnx import?


sorenlassen:
it seems like we can inline all function calls during onnx import

this would simplify the import logic a little bit and will make it possible to remove ONNXCallOp giving us one less thing to analyze, transform, and convert in the compiler

see proof of concept implementation in PR #1785

question: should we support function protos local to a model ([link](https://github.com/onnx/onnx/blob/96046b8ccfb8e6fa82f6b2b34b3d56add2e8849c/onnx/onnx.proto#L412))? does anybody know a model that uses that onnx feature?

sorenlassen:
> Can we use inlining provided by MLIR after the function is imported into onnx-mlir?

maybe, but I think the protobuf-to-mlir import logic in FrontendDialectTransformer is simplest if we inline there instead of constructing the function call in mlir

> To me, the priority of issue is low.

right, it's not a priority right now but that's partly because we haven't started implementing shape inference and lowering for function calls yet - if we inline now then we won't have to do any of that

AlexandreEichenberger:
High level comments: I know that the onnx standard wants to implement some of their newer operators as a function of existing ONNX ops. Not sure if it comes as a function or not, but clearly we may want to implement some of these higher level ops as is. For example, the LSTM operator could be represented as a function of smaller onnx ops, and we want to transform that LSTM into a LSTM op that we support as a single op, esp when pushing them to our NNPA accelerator which support some LSTM ops.

Again, apologies because I don't know how these ONNX complex ops comes into our protobufs, but if they come as a function, we will have to be selective as which one to inline. We also had a request for supporting non-standard ONNX ops that are implemented in ORT... so that is also a second source of such functions that may/may not be expanded.

Maybe someone could look into this issue and provide an explanation here for the benefit of the others. Any takers?

sorenlassen:
Note that the lack of polymorphic type/shape system and given the way we lower with fixed shapes means that it will only be practical to share functions between call sites with identical types and shapes. I expect that that will make it impractical to do anything other than inlining once we get around to do type inference and lowering for functions and function calls. Therefore I propose that we inline now. If we figure out polymorphic types and shapes and all that, then we can revert the change to inlining in the future.

sorenlassen:
> remove ONNXCallOp giving us one less thing to analyze, transform, and convert in the compiler

[...]

Issue #2185, opened 2023-04-20
function call shape inference


sorenlassen:
As discussed in PR #2175 shape inference doesn't work for function calls: ONNXCallOp doesn't implement ShapeInferenceOpInterface and if we modify a function's return shapes during shape inference then we'll break all call ops to that function. Given that this doesn't work, ShapeInferencePass (and ONNXHybridTransformPass which also does shape inference) is currently a function pass even if it modifies function signatures, which is probably improper behavior of a function pass but doesn't matter since a function's signature is not currently used outside the function in any meaningful way.

If we want to implement function calls properly, we'll need to come up with a strategy for shape inference to propagate updates to return shapes to callers, maybe with some kind of module-level fixpoint iteration.

As mentioned in issue #1787 there are a number of other problems with the implementation of function calls in the ONNX dialect. One solution is to inline function calls during onnx import. Any solution we choose should ideally support model-local functions, see issue #1841.

cc @chentong319

chentong319:
First, I need some clarification on ONNXCallOp and functions for a model. I never handled a model using ONNXCallOp.
By reading the [onnx document](https://onnx.ai/onnx/api/classes.html#functionproto) and our builder code (FrontendDialectTransformer.cpp:TryImportFunctionCallNode), I feel the key question is whether a function body is called only by a ONNXCallOp. This means a function body has only one calling context. It will simplify our work a lot. The first approach occurred to me is to use MLIR inlining function method to eliminate ONNXCallOp and functions, leaving only the top functions.

If the functions may be called from different ONNXCallOps, it will be a different story. We have to propagate shape info in the call graph/tree in both directions.

sorenlassen:
I believe FrontendDialectTransformer.cpp:TryImportFunctionCallNode only creates one caller for each function, which is why I suggested in issue #1787 that we just inline in FrontendDialectTransformer and get rid of ONNXCallOp (proof of concept implementation in PR #1785), ONNXCallOp doesn't have implementation of shape inference or lowering so it's not useful currently

but if we want to support model-local functions (issue #1841) then maybe it's worthwhile having real functions and function calls, and then we'd need to figure out shape inference (and lowering)

chentong319:
Should we go ahead with PR#1785 for schema-based function first? 
I deleted my previous comment by mistake.

chentong319:
About lowering ONNXCallOp:
[...]

PR #2242, opened 2023-05-12
inline functions in FrontendDialectTransformer

sorenlassen:
This PR
- inlines function calls during onnx import in FrontendDialectTransfomer,
- implements function attribute parameters passing,
- supports function opset imports,
- removes the ONNXCallOp,
- and adds support for model local functions.

Resolves #1787, resolves #1841.

Supports both model local functions and “function decompositions” where the operation’s OpSchema specifies a function, which can be either context independent or context dependent.

Note that the decomposition logic is only exercised in CustomFnTest and not for regular ONNX operations because they all have handlers in `import_handler_map_` that take precedence over decomposition. PR #2213 introduces a flag `-functions-to-decompose` to control the choice of handler versus decomposition.

Added parse lit tests with the new ONNX textual syntax from PR #2262.

The reasons for abandoning ONNXCallOp and instead inline functions are discussed in issues #2185 and #1787. In summary: a function depends on the  caller’s attributes, types, and shapes in ways that don’t map directly to MLIR functions, so a separate MLIR function more or less needs to be created for each call site, and then inlining is a simpler implementation. This obviates implementing lowering and shape inference for ONNXCallOp which would be quite involved.


sorenlassen:

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -1003,212 +1107,220 @@ class FrontendGenImpl {
...
-      }
+    // Substitute caller attributes in graph nodes:
+    AttrMap caller_attr_map;
+    for (const onnx::AttributeProto &attr : node.attribute())
+      caller_attr_map[attr.name()] = &attr;
+    AttrMap attr_map;
+    for (const std::string &attrName : functionProto.attribute()) {
+      auto it = caller_attr_map.find(attrName);
+      if (it != caller_attr_map.end())
+        attr_map[attrName] = it->second;
```
chentong319, line 1202: I am not familiar with the attribute substitution. Any link to the relevant document?

sorenlassen: there isn't a lot of documentation

FunctionProto has attributes: https://github.com/onnx/onnx/blob/ad834eb73ee0cd9b6fa9ea892caeed5fa17d7dc0/onnx/onnx-ml.proto#L798-L799

and onnx's shape inference implementation illustrates how they are used:
https://github.com/onnx/onnx/blob/ad834eb73ee0cd9b6fa9ea892caeed5fa17d7dc0/onnx/shape_inference/implementation.cc#L524-L563

sorenlassen: `code_function_with_attributes` in FunModelTest in onnx's ParserTest is a good illustration:
https://github.com/onnx/onnx/blob/main/onnx/test/cpp/parser_test.cc#L500-L524

chentong319: It seems to me that the attributes with the call site is always used and the attributes attached with the Function is useless.

chentong319: Can the attribute of a function be used by ops inside function body? Or it is actually used for the call site? I am ponding on the comment

>  so a separate MLIR function more or less needs to be created for each call site,

sorenlassen: if a FunctionProto has an attribute gamma then it will be referenced as @gamma in the function's body and will be substituted by the caller's gamma attribute value when invoked

in the example below 3.0 is substituted for @gamma in foo's body in `out = custom_domain.foo<alpha=2.0, gamma=3.0>(x)` and if foo is called elsewhere with a different gamma then a different substitution needs to happen, and mlir functions don't support the substitution of values for attributes and therefore foo cannot be represented by a single mlir function

```
<
  ir_version: 9,
  opset_import: [ "" : 15, "custom_domain" : 1]
>
agraph (float[N] x) => (float[N] out)
{
  out = custom_domain.foo<alpha=2.0, gamma=3.0>(x)
}

<
domain: "custom_domain",
opset_import: [ "" : 15],
doc_string: "function foo"
>
  foo
  <alpha: float=4.0, gamma>
  (X) => (C)
  {
      constant_alpha = Constant<value_float: float=@alpha>()
      constant_gamma = Constant<value_float: float=@gamma>()
      constant_alpha_x = Mul(constant_alpha, X)
      C = Add(constant_alpha_x, constant_gamma)
  }
```

sorenlassen:
> Does this PR also remove the code to generate CustomOp if function body is not provided?

no, CustomOp support is unchanged

src/Builder/FrontendDialectTransformer.cpp
sorenlassen: > It seems to me that the attributes with the call site is always used and the attributes attached with the Function is useless.

the "attributes" attached with the Function are just attribute names

sorenlassen:
@chentong319 are you ok approving this PR or would you like me to make any changes?

src/Builder/FrontendDialectTransformer.cpp
chentong319: Will the attribute binding also use this mapping? string to Value?
```
void BindOnnxName(const std::string &onnx_name, Value symbol) {
     frontend_symbols_.AddMapping(onnx_name, symbol);
   }
```

chentong319 (approved):
LGTM!

src/Builder/FrontendDialectTransformer.cpp
sorenlassen: no, attributes names map to constants (onnx::AttributeProto which will be translated to mlir::NamedAttribute in convertOnnxAttributeProtoToMlirNamedAttribute) not mlir::Value, and attribute `@` references (like `@gamma` in the example) are replaced by substitution in one shot in `replaceAttrRefs(graph, attr_map)`, there is no FrontendGenImpl instance member for attributes like frontend_symbols_ for values

sorenlassen:

sorenlassen:

PR #2243, opened 2023-05-12
remove old non-pattern-based shape inference

sorenlassen:
The comments and logic removed in ShapeInferencePass.cpp are quite old and don't make 100% sense anymore.

Unless anybody protests within the next month I intend to merge this PR around June 18. By then hopefully we will have surfaced any issues with the new pattern based shape inference which became the default in PR #2240 which was merged today.


hamptonm1:
> I am hesitating in remove the old code.

I agree, I think we should keep it until we resolve the onnx model failures.

sorenlassen:
> > I am hesitating in remove the old code.
> 
> I agree, I think we should keep it until we resolve the onnx model failures.

OK to go ahead and remove the old code now? The comments and logic removed in ShapeInferencePass.cpp are quite old and don't make 100% sense anymore.

AlexandreEichenberger:
@chentong319 please let us know if this is ok by you. 
If there is no consequences, I am in favor of removing unused passes

chentong319 (approved):
LGTM!

PR #2247, opened 2023-05-16
Unary shape helper under broadcast shape helper

AlexandreEichenberger:
Changed the ShapeHelper hierarchy. Before, Unary and Binary shape helper were unrelated. This caused code duplication as we had to generate two different ways to do SIMD (and possibly kernel merging). 

Now, Unary is a specialization (subclass) of Binary, where calls such as "Is broadcast present" always return false as there is never any broadcasting for unary operations.

As a result, the code generating the SIMD code was reduced to one case.


tungld (approved):
LGTM. What a huge reduction in code! Thanks!

PR #2248, opened 2023-05-16
Enable IsInf Positive and Negative Backend Tests

hamptonm1:
I was able to write the code to ensure the backend tests for both `isinf_positive` and `isinf_negative` pass. Also ensured the original path for simply `isinf` does not fail as well. Thank you so much to all who have helped me in this process. I have learned so much and very thankful that I continue to grow. Please provide your valuable feedback. Thanks!!!


src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -322,6 +316,55 @@ struct ScalarOp<ONNXTanOp> {
...
+
+  // Three different cases: Infinity, Negative Infinity and Positive Infinity
+  bool detectInf = detectPosAttribute == 1 && detectNegAttribute == 1;
+  bool detectNeg = detectNegAttribute == 1 && detectPosAttribute == 0;
+  bool detectPos = detectPosAttribute == 1 && detectNegAttribute == 0;
+
+  Value equPos = create.math.eq(operand, posInf);
+  Value equNeg = create.math.eq(operand, negInf);
+
+  if (detectInf) {
```
hamptonm1, line 353: I can do an `if-else if` here too but from my experience in the community I recall sticking with just `if`

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -322,6 +316,55 @@ struct ScalarOp<ONNXTanOp> {
...
+  Value negInf = create.math.negativeInf(inputElemType);
+  Value posInf = create.math.positiveInf(inputElemType);
+
+  double detectNegAttribute = dyn_cast<ONNXIsInfOp>(op).getDetectNegative();
+  double detectPosAttribute = dyn_cast<ONNXIsInfOp>(op).getDetectPositive();
+
+  // Three different cases: Infinity, Negative Infinity and Positive Infinity
+  bool detectInf = detectPosAttribute == 1 && detectNegAttribute == 1;
+  bool detectNeg = detectNegAttribute == 1 && detectPosAttribute == 0;
+  bool detectPos = detectPosAttribute == 1 && detectNegAttribute == 0;
```
tungld, line 348: Do we have `detectPosAttribute == 0 && detectNegAttribute == 0`? If not, we can verify it inside the op's verifier.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -322,6 +316,55 @@ struct ScalarOp<ONNXTanOp> {
...
+// Scalar unary ops for lowering ONNXIsInfOp
+//===----------------------------------------------------------------------===//
+template <>
+struct ScalarOp<ONNXIsInfOp> {
+  using FOp = CustomScalarOp;
+  using IOp = NotSuportedScalarOp;
+};
+
+template <>
+Value emitScalarOpFor<ONNXIsInfOp>(ConversionPatternRewriter &rewriter,
```
tungld, line 329: It seems computation here can be simdized. Could you define `simdAnalysis` for this op for SIMD?

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -322,6 +316,55 @@ struct ScalarOp<ONNXTanOp> {
...
+  double detectNegAttribute = dyn_cast<ONNXIsInfOp>(op).getDetectNegative();
+  double detectPosAttribute = dyn_cast<ONNXIsInfOp>(op).getDetectPositive();
+
+  // Three different cases: Infinity, Negative Infinity and Positive Infinity
+  bool detectInf = detectPosAttribute == 1 && detectNegAttribute == 1;
+  bool detectNeg = detectNegAttribute == 1 && detectPosAttribute == 0;
+  bool detectPos = detectPosAttribute == 1 && detectNegAttribute == 0;
+
+  Value equPos = create.math.eq(operand, posInf);
+  Value equNeg = create.math.eq(operand, negInf);
```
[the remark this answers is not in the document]
hamptonm1: I have it outside of the `if` to avoid redundancy. I use equPos and equNeg in the first `if` which is to detect regular infinity. So I guess either way there will be redundancy.  It is used here:

```
 if (detectInf) {
    // If infinity return true for both positive and negative infinity
    result = create.math.ori(equPos, equNeg);
  }
```

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
hamptonm1: I have never done computation for SIMD but I can try and you all can let me know if what I have makes sense. Thanks

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
tungld: Checking it inside IsInf's verifier would be good enough I believe, and you can assert it here if necessary.

hamptonm1: Okay thank you for your feedback!

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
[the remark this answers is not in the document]
hamptonm1: Okay I updated this...let me know your thoughts.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
hamptonm1: I added the verifier @tungld

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
hamptonm1: Okay added `simdAnalysis`...let me know if this looks good. Thanks

src/Dialect/ONNX/ONNXOps/Math/ElementwiseUnary.cpp
```diff
@@ -280,6 +280,32 @@ LogicalResult ONNXInstanceNormalizationOp::inferShapes(
...
+  if (!hasShapeAndRank(operandAdaptor.getX()))
+    return success(); // Won't be able to do any checking at this stage.
+
+  int64_t detectPosAttribute = getDetectPositive();
+  int64_t detectNegAttribute = getDetectNegative();
+  
+  // One of the values for detectPosAttribute and detectNegAttribute must be 1. 
+  // If not, then this will result in an error. 
+  if (detectPosAttribute == 0 && detectNegAttribute == 0)
+    return emitOpError("This variation is currently unsupported. One or both of the attributes must be a value of 1 to ensure mapping to infinity.");
```
hamptonm1, line 297: @AlexandreEichenberger As I mentioned to @tungld `IsInf` only supports 3 cases. The fourth scenario which is `detectPosAttribute == 0 && detectNegAttribute == 0` is not mentioned in the documentation. So I added code to the verifier to basically say we do not support that option especially because we have no backend test that even covers it. Having `detectPosAttribute == 0 && detectNegAttribute == 0` would defeat the purpose of the `IsInf` test because everything would be false (nothing will ever be infinity). Are you okay with me adding this verifier here or do you have another suggestion?

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
AlexandreEichenberger: it seems that the code is "equal" and "or"... so why `genericops::IsInfGop`

hamptonm1: > it seems that the code is "equal" and "or"... so why `genericops::IsInfGop`

Learning... so from looking at the code, I would use`LogicalGop`... /* All logical ops: and, or, xor, not, nor, nand,... */

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
tungld: Thanks!

BTW, you need to add `IsInf` to the verifier list in the file `utils/gen_onnx_mlir.py` and run `make OMONNXOpsIncTranslation` to update the ONNX definition file, which will solve the issue in Jenkins.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -322,6 +316,58 @@ struct ScalarOp<ONNXTanOp> {
...
+//===----------------------------------------------------------------------===//
+template <>
+struct ScalarOp<ONNXIsInfOp> {
+  using FOp = CustomScalarOp;
+  using IOp = NotSuportedScalarOp;
+};
+
+template <>
+double analyzeSimdFor<ONNXIsInfOp>(Type t, int64_t &von, int64_t &son) {
+  return simdAnalysis({GenericOps::LogicalGop}, {1}, t, von, son);
```
tungld, line 330: I am afraid that we need to have `if/then/else` here to count the number of ops exactly for each case. For example, `detectInf` has two `eq` and one `or`, while `detectPos` and `detectNeg` have one `eq`.

hamptonm1: Okay let me try and figure it out :)

utils/gen_onnx_mlir.py
```diff
@@ -369,6 +369,7 @@
     'GreaterOrEqual',
     'Hardmax',
     'If',
+    'IsInf'
```
hamptonm1, line 372: @tungld I did add it to `utils/gen_onnx_mlir.py` and ran the `make OMONNXOpsIncTranslation` command but that did not work.

hamptonm1: I am going to try to rebuild the entire build directory to see if the changes will pick up then

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
hamptonm1: Fixed it!

utils/gen_onnx_mlir.py
hamptonm1: Okay I was missing a `,` and that is why the ONNXOps.td.inc file was not updating. Fixed now :)

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
AlexandreEichenberger: nit: I understand they are mutually exclusive. Probably the preferred way would be to return the value directly.
```
if (detectInf) 
  return  create.math.ori(
         create.math.eq(operand, posInf), create.math.eq(operand, negInf));
```
and so on, since result is not used in any ways

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -322,6 +316,58 @@ struct ScalarOp<ONNXTanOp> {
...
+  MultiDialectBuilder<MathBuilder> create(rewriter, loc);
+  Value negInf = create.math.negativeInf(inputElemType);
+  Value posInf = create.math.positiveInf(inputElemType);
+
+  double detectNegAttribute = dyn_cast<ONNXIsInfOp>(op).getDetectNegative();
+  double detectPosAttribute = dyn_cast<ONNXIsInfOp>(op).getDetectPositive();
+
+  // Three different cases: Infinity, Negative Infinity and Positive Infinity
+  bool detectInf = detectPosAttribute == 1 && detectNegAttribute == 1;
+  bool detectNeg = detectNegAttribute == 1 && detectPosAttribute == 0;
```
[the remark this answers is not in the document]
hamptonm1: Okay I can update that

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
hamptonm1: Thanks @AlexandreEichenberger

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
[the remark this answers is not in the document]
hamptonm1: Done!

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
hamptonm1: Done!

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
hamptonm1: @AlexandreEichenberger So I wrote skeleton code for SIMD as shown below:

```
// pos or neg: 1 compare (1 eq)
// pos and neg: 2 compares (2 eq's) and 1 logical (1 or)

template <>
double analyzeSimdFor<ONNXIsInfOp>(Type t, int64_t &von, int64_t &son) {
  if (pos || neg) {
    return simdAnalysis({GenericOps::CompareGop}, {1}, t, von, son);
  } else {
    return simdAnalysis(
        {GenericOps::CompareGop, GenericOps::LogicalGop}, {2, 1}, t, von, son);
  }
}
```

However, I ran into a slight issue.... the way I distinguish between the positive and negative cases are based on the attribute values which we would retrieve from the `Operation *op`. @tungld suggested we should pass `Operation *op` as a parameter in `analyzeSimdFor` so that we can ensure we have SIMD for the IsInf operation. What are your thoughts? If we add this parameter for `analyzeSimdFor` we would have to change it for every single operation that utilizes the template that was created in `Elementwise.cpp`. If this is a bigger issue, can we merge this PR and I can further work on the enhancement of SIMD for IsInf?

AlexandreEichenberger: @hamptonm1 please add the generic op as parameter. will undoubtedly be useful for other ops as well. that would be a worthy upgrade.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
hamptonm1: @AlexandreEichenberger I had to change this back to return the value `result` only because `Windows` check fails with the following error: 

`D:\a\1\onnx-mlir\src\Conversion\ONNXToKrnl\Math\Elementwise.cpp(361) : warning C4715: ‘onnx_mlir::emitScalarOpFor<mlir::ONNXIsInfOp>’: not all control paths return a value`

Since we are directly returning the values in the if's... it is expecting a finally return value.

chentong319 (approved):
LGTM!

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
hamptonm1: @AlexandreEichenberger Okay I added the op as a parameter... let me know if this looks good in `ElementWise.cpp`

Issue #2164, opened 2023-04-14
utils/RunONNXModel.py does not support string-type model inputs


negiyas:
"bidaf" model in the model zoo uses string-type inputs, so that we need string-type inputs in order to execute the model.
If possible, I want to use utils/RunONNXModel.py to execute the model, but it does not support string-type inputs, even if the following ad-hoc patch is used to get input_names and output_names correctly without errors.
It seems that we need (1) "generator of random string array" and (2) "loader of protobuf file (e.g. input_0.pb) supporting string type".

```
$ git diff
diff --git a/utils/RunONNXModel.py b/utils/RunONNXModel.py
index f2b77ad4..d28386d8 100755
--- a/utils/RunONNXModel.py
+++ b/utils/RunONNXModel.py
@@ -552,8 +552,8 @@ def main():
         print("  took ", end - start, " seconds.\n")

         # Get the input and output signature.
-        input_signature = sess.input_signature()
-        output_signature = sess.output_signature()
+        input_signature = sess.input_signature().replace('!krnl.strin
g', '\"string\"')
+        output_signature = sess.output_signature().replace('!krnl.str
ing', '\"string\"')
         input_names = get_names_in_signature(input_signature)
         output_names = get_names_in_signature(output_signature)
```

The following errors occur with and without specifying the reference holder.

**Use reference folder specified by "--load-ref"**
```
$ ONNX_MLIR_HOME=/Dreamer/trl/negishi/src/dlc.git/onnx-mlir/build/Debug ./utils/RunONNXModel.py --compile-args='--O3 --EmitLib --mtriple=s390x-ibm-loz --mcpu=z16 --onnx-op-stats TXT' --verify ref --verify-every-value --load-ref bidaf/test_data_set_0 --model bidaf/bidaf.onnx --load-so bidaf/bidaf-noshape-cpu.so
Numpy type not supported: object.
```

**Use random data without using "--load-ref"**
```
[...]

PR #2255, opened 2023-05-18
Allow rank-3 onnx.ml.CategoryMapper

negiyas:
Since the Bidaf model in the model zoo (https://github.com/onnx/models/tree/main/text/machine_comprehension/bidirectional_attention_flow) uses CategoryMapper op with rank-3 input, and onnx-mlir can support the rank-3 CategoryMapper op naturally.
 
This PR includes simple removal of the input rank check part, and its lit tests.


negiyas:

tungld:
LGTM. I believe you enable not only rank-3 but arbitrary rank. Can we make the numerical test TestCategoryMapper cover rank-3? And it seems TestCategoryMapper is not called by Jenkins CI:
```
      Start  1: TestConv
      Start  2: TestMatMul2D
      Start  3: TestMatMulBroadcast
      Start  4: TestGemm
      Start  5: TestLSTM
      Start  6: TestRNN
      Start  7: TestGRU
      Start  8: TestLoop
      Start  9: TestScan
      Start 10: TestElementwise
```

Could you enable TestCategoryMapper also? I am sorry for many requests. Thanks!

negiyas:
> LGTM. I believe you enable not only rank-3 but arbitrary rank. Can we make the numerical test TestCategoryMapper cover rank-3? And it seems TestCategoryMapper is not called by Jenkins CI:

The latest TestCategoryMapper  (at test/backend-cpp/TestCategoryMapper.cpp)  in the main branch causes errors. (I am not sure why the test run with the previous onnx-mlir before... The situation might be different to the current one.)
According to my understanding, the situation to use string-type inputs should be the same to the bidaf model case reported in Issuehttps://github.com/onnx/onnx-mlir/issues/2164 .

If you agree with the following proposal, I will modify the numerical test for testing two connected onnx.CategoryMapper ops.
The first op gets integer inputs and generates string outputs, and the second op gets string inputs and generates integer outputs. Accordingly we need not to use string-type inputs and outputs for the model.

I am happy to hear your comments.

negiyas:
> Looks good to me, and as an additional way, you can use a constant string as an input for CategoryMapper.

I have realized that TestCategoryMapper (at test/backend-cpp/TestCategoryMapper.cpp) is a very good test case to enable string input data, so that I am enabling string input data in this PR if the modification is enough small. (Otherwise I will create another PR for string input enablement.)

Since the impact to enable string input data, is larger, It may take time a little bit longer time.

negiyas:
> Looks good to me, and as an additional way, you can use a constant string as an input for CategoryMapper.

I fixed an issue in src/Dialect/ONNX/ONNXOps/OpHelper.cpp to enable string type in the mlirTypeToOnnxType function.
and supported rank2 and rank3 inputs tests in test/backend-cpp/TestCategoryMapper.cpp .

I will enable ./build/Debug/bin/TestCategoryMapper to the Jenkins test.

negiyas:
I confirmed that /build/Debug/bin/TestCategoryMapper run in the Jenkins test.
I suppose(hope) that this PR is ready to be merged, if it passes all tests.

tungld (approved):
LGTM. Please do check the errors on Windows. Thanks!

PR #2258, opened 2023-05-19
An option to specify axis for verification with softmax in RunONNXModel.py

imaihal:
Added an argument in `--verify-with-softmax` option to specify axis along which Softmax is performed. -1 indicates last dimension.
ex.
--verify ref --verify-with-softmax -1

### DIFF
--- a/utils/RunONNXModel.py
+++ b/utils/RunONNXModel.py
@@ -85,7 +85,7 @@ parser.add_argument('--verify-with-softmax',
                     default=None,
                     help="Verify the result obtained by applying softmax along with"
                     " specific axis. The axis can be specified"
-                    " by --verify-softmax-axis=<axis>.")
+                    " by --verify-with-softmax=<axis>.")
 parser.add_argument('--verify-every-value',
                     action='store_true',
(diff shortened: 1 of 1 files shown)


utils/RunONNXModel.py
```diff
@@ -623,13 +625,33 @@ def main():
...
+                        mismatched_elements += 1
+                        print("  at {}".format(index),
+                              "mismatch {} (actual)".format(actual_val),
+                              "vs {} (reference)".format(ref_val))
+                    if mismatched_elements == 0:
+                        print("  correct.\n")
+                    else:
+                        raise AssertionError(
+                            "  mismatched elements {}/{}.\n".format(
+                                mismatched_elements, total_elements))
```
tungld, line 654: Looks like the verification here is quite simlar to the one in `verify_every_value`, could you extract the common code into a function, then call the function? Thanks!

imaihal: Done. Thanks!

tungld (approved):
LGTM. Thanks for updating the script!

PR #2264, opened 2023-05-24
regenerated docs

sorenlassen:
with the command `cmake --build . --target onnx-mlir-docs`

### DIFF
--- a/docs/Dialects/krnl.md
+++ b/docs/Dialects/krnl.md
@@ -161,7 +161,8 @@ is lowered. Attributes will be converted to parameters too (To be Added).
 The function signature will be determined with the types of parameters.
 An LLVM::CallOp to either a runtime library or a llvm intrinsic function
 will be generated.
-The krnl.call op will be lowered to llvm at krnl-to-llvm conversion.
+The krnl.call op will be lowered to llvm at krnl-to-llvm conversion in which
+OMTensor is used as a container for MemRef arguments.
 
 Interfaces: MemoryEffectOpInterface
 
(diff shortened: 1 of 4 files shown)


sorenlassen:
The instructions in the comments in ONNX.td
https://github.com/onnx/onnx-mlir/blob/main/src/Dialect/ONNX/ONNX.td#L15
say to run `make onnx-mlir-docs` but that doesn't work for me and instead I ran `cmake --build . --target onnx-mlir-docs`

Is that correct? Should I edit the instructions?

The same goes for `make OMONNXOpsIncTranslation`: should we change that to `cmake --build . --target OMONNXOpsIncTranslation`?

sorenlassen:
@tungld can I ask you to review this PR? among other things it includes the documentation for ONNXShapeTransformOp

tungld (approved):
LGTM.

tungld:
I always use `make onnx-mlir-docs` and `make OMONNXOpsIncTranslation`, and haven't seen any issue. I call them inside `build` folder.

sorenlassen:
> I always use `make onnx-mlir-docs` and `make OMONNXOpsIncTranslation`, and haven't seen any issue. I call them inside `build` folder.

I figured out that it's because I build with ninja instead of make on OSX, following the instructions here:
https://github.com/onnx/onnx-mlir/blob/main/docs/BuildOnLinuxOSX.md
I can use `ninja onnx-mlir-docs`

I can use make if I build with `-G"Unix Makefiles"` instead of `-G Ninja`

we might consider changing the instructions to use `cmake --build . --target onnx-mlir-docs` which should work across ninja and make

sorenlassen:
Now also regenerated krnl.md, zhigh.md, zlow.md

PR #2265, opened 2023-05-24
[NNPA] Disable N-D softmax because of NaNs output

tungld:
For the GPT2 XL model on NNPA, if N-D softmax is run on NNPA, it produces NaNs because its input is out of range of dlfloat16. So, we let N-D softmax run on CPU instead to get correct results for the GPT2 XL model.


AlexandreEichenberger (approved):
LGTM, wonder if we would not need to have a flag to disable it, as it may be ok for some models.

imaihal:
> wonder if we would not need to have a flag to disable it, as it may be ok for some models.

Is `--execNodesOnCpu` option useful? Current option might not be easy to use because all node name need to be set, though.

tungld:

PR #2266, opened 2023-05-24
Upgrade torch and torchvision

hamptonm1:
We have to upgrade torch and torchvision to avoid vulnerabilities mentioned in the LFX and snyk scans. 

`torchvision 0.15.1 depends on torch==2.0.0`

### DIFF
--- a/docs/mnist_example/requirements.txt
+++ b/docs/mnist_example/requirements.txt
@@ -1,4 +1,4 @@
 numpy~=1.22.2
 pillow~=9.4.0
-torch~=2.0.0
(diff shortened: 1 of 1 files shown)


AlexandreEichenberger (commented):
please scan for torch and torch vision; there is also a reference to it in the docker info under docs

hamptonm1:
> there is also a ref for torch and torch vision in the docker info under docs; I will update that independently as I have a PR for that file

Oh perfect!! Thanks @AlexandreEichenberger

PR #2268, opened 2023-05-25
rename ONNXReturnOp to ONNXYieldOp

sorenlassen:
to follow the MLIR naming scheme in the SCF dialect:
https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Dialect/SCF/IR/SCFOps.td

follows the suggestion by @AlexandreEichenberger: https://github.com/onnx/onnx-mlir/pull/2239#discussion_r1200101718


AlexandreEichenberger (approved):
LGTM, thanks for standardizing the naming scheme.

PR #2273, opened 2023-05-25
fix inferShapes() for inputs without static shapes

### LINKED ISSUE #2271: Investigate why pattern-based shape inference fails when setting the default to True
This issue was created based on the following PR: https://github.com/onnx/onnx-mlir/pull/2270

FYI- @sorenlassen @tungld

sorenlassen:
Don't run inferShapes() on inputs the shape helper computeShape() cannot handle.

Fixes #2271.

### DIFF
--- a/src/Dialect/ONNX/ONNXOps/Tensor/ConstantOfShape.cpp
+++ b/src/Dialect/ONNX/ONNXOps/Tensor/ConstantOfShape.cpp
@@ -59,8 +59,9 @@ LogicalResult ONNXConstantOfShapeOp::verify() {
   auto inputShape = input.getType().cast<RankedTensorType>().getShape();
   if (inputShape.size() != 1)
     return emitOpError("Input tensor must be a 1D tensor");
+
   if (ShapedType::isDynamic(inputShape[0]))
-    return emitOpError("Input tensor must have static shape");
+    return success();
 
   // Calculate output dimensions.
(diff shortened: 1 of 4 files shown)


src/Dialect/ONNX/ONNXOps/Tensor/ConstantOfShape.cpp
```diff
@@ -59,8 +59,9 @@ LogicalResult ONNXConstantOfShapeOp::verify() {
   auto inputShape = input.getType().cast<RankedTensorType>().getShape();
   if (inputShape.size() != 1)
     return emitOpError("Input tensor must be a 1D tensor");
+
   if (ShapedType::isDynamic(inputShape[0]))
-    return emitOpError("Input tensor must have static shape");
```
sorenlassen, line 63: it seems wrong for verify() to require the input shape to be static - it could start dynamic and become static during shape inference

tungld: It makes sense.

tungld (approved):
LGTM. Thanks for fixing them!

hamptonm1:
Can we have @cjvolzka and @negiyas confirm on their end that they no longer see it as well? Just so we can avoid another potential issue.

cjvolzka:
I tried `--pattern-shape-inference=true` using the latest onnx-mlir build and it compiled with no problems :)

PR #2275, opened 2023-05-26
Add --modelSize to handle large models

tungld:
Add `--modelSize={small,medium,large,huge}` to handle different model sizes. Currently only `small` and `large` are supported in which `large` will place global constants into `.lrodata` section instead of `.rodata` section.


src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
```diff
@@ -162,7 +162,7 @@ void addPassesNNPA(mlir::OwningOpRef<mlir::ModuleOp> &module,
 
   if (emissionTarget >= EmitLLVMIR)
     // Lower the remaining Krnl and all ZLow ops to LLVM dialect.
-    addKrnlToLLVMPasses(pm, /*enableCSE=*/true, verifyInputTensors);
+    addKrnlToLLVMPasses(pm, /*enableCSE=*/true);
```
tungld, line 165: Not the main change in this patch. `verifyInputTensors` is redundant to pass in because we can get it directly from the options.

src/Compiler/CompilerOptions.cpp
```diff
@@ -95,6 +95,21 @@ llvm::cl::opt<std::string> march("march",
     llvm::cl::value_desc("Target a specific architecture type"),
     llvm::cl::cat(OnnxMlirOptions), llvm::cl::ValueRequired);
 
+llvm::cl::opt<MCModel> mcmodel("mcmodel",
```
AlexandreEichenberger, line 98: Can you provide an explanation about the option name? maybe `-model-size=small/large` The '-m' flags are related to machine parameters/arch. I think the concept here is more generic; we are telling the compiler to accommodate the output for large model, which may result In many different optimizations. It is perfectly fine that different arch may react to the model size differently, but that does not make this flag a "machine related flag"

Another example (maybe for a `huge`) would be to store values into files (if there is a max binary size), in the future.

Medium is said to be unsupported... then no need to add it.

AlexandreEichenberger (approved):
LGTM pending option name clarification/changes.

src/Compiler/CompilerOptions.cpp
tungld: I was inspired by `--mcmodel` in gcc: https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html. But you are right, we can use a neural name instead. LLVM uses `code-model` in `opt` and `llc`, which can be a good name.

tungld: @AlexandreEichenberger now I change to use `-modelSize=small/large`. Thanks!

tungld:

PR #2277, opened 2023-05-26
interpret empty tensors as absent Resize inputs

### LINKED ISSUE #2276: yolov4 failed
@sorenlassen it seems yolov4 has failed  since this PR:
Screenshot 2023-05-26 at 17 20 27

sorenlassen:
The yolo4 model uses a float tensor with shape [0] to represent that roi
or scales is absent in accordance with the Resize v11 spec. This violates
the spec from v13 onwards which says that empty string
inputs represents absent arguments in the protobuf model representation.
We work around this by interpreting a tensor with empty shape as an
alternative way to express that an input is absent.

Fixes #2276

### DIFF
--- a/src/Dialect/ONNX/Rewrite.cpp
+++ b/src/Dialect/ONNX/Rewrite.cpp
@@ -264,18 +264,23 @@ public:
 
   LogicalResult matchAndRewrite(
       ONNXResizeOp onnxResizeOp, PatternRewriter &rewriter) const override {
-    OnnxBuilder createONNX(rewriter, onnxResizeOp.getLoc());
-
-    if (isEmptyTensor(onnxResizeOp.getRoi()))
-      onnxResizeOp.getRoiMutable().assign(createONNX.none());
-
-    if (isEmptyTensor(onnxResizeOp.getScales()))
-      onnxResizeOp.getScalesMutable().assign(createONNX.none());
-
-    if (isEmptyTensor(onnxResizeOp.getSizes()))
-      onnxResizeOp.getSizesMutable().assign(createONNX.none());
-
-    return success();
+    bool emptyRoi = isEmptyTensor(onnxResizeOp.getRoi());
+    bool emptyScales = isEmptyTensor(onnxResizeOp.getScales());
+    bool emptySizes = isEmptyTensor(onnxResizeOp.getSizes());
+    if (emptyRoi || emptyScales || emptySizes) {
+      rewriter.updateRootInPlace(onnxResizeOp, [&] {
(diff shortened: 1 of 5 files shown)


src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -179,6 +179,45 @@ bool AreTheSameAxesConstant(int64_t rank, Value lhs, Value rhs) {
...
+// =============================================================================
+// Rewrite pattern for Resize (not handled in Rewrite.td).
+// =============================================================================
+
+// The yolo4 model uses a float tensor with shape [0] to represent that roi
+// or scales is absent (in violation of the spec which says that empty string
+// inputs represents absent arguments in the protobuf model representation).
+// We work around this by interpreting a tensor with empty shape as an
+// alternative way to express that an input is absent.
+class EmptyTensorInputsResizePattern : public OpRewritePattern<ONNXResizeOp> {
```
tungld, line 191: Could you please add lit tests for this rewriting? Thanks!

src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -179,6 +179,45 @@ bool AreTheSameAxesConstant(int64_t rank, Value lhs, Value rhs) {
...
 
+// =============================================================================
+// Rewrite pattern for Resize (not handled in Rewrite.td).
+// =============================================================================
+
+// The yolo4 model uses a float tensor with shape [0] to represent that roi
+// or scales is absent (in violation of the spec which says that empty string
+// inputs represents absent arguments in the protobuf model representation).
+// We work around this by interpreting a tensor with empty shape as an
+// alternative way to express that an input is absent.
```
tungld, line 190: Got it. Thanks for the investigation!

tungld (approved):
LGTM. Please wait for Tong's comment before merging as he has done many works on this op.

@chentong319 could you take a look at this patch? Thanks!

src/Dialect/ONNX/Rewrite.cpp
sorenlassen: good suggestion

done

chentong319:
> uses a float tensor with shape [0] to represent that roi or scales is absent

Is this a bug or just legacy representation?
If we are sure that such representation is illegal or deprecated, we may try to update the model instead of modifying the compiler.

sorenlassen:
update: I wrapped the logic to mutate the Resize operands in rewriter.updateRootInPlace() because I think the rewriter is supposed to be informed about changes to the operation graph in that way

chentong319 (approved):
Thanks for digging out the details.

PR #2280, opened 2023-05-28
reduce repeated pattern-based shape inference computations

sorenlassen:
Made matchAndRewrite() in shape inference patterns InferShapesPattern and YieldShapesPattern return failure() when no op changes are made. Otherwise applyPatternsAndFoldGreedily() would repeat GreedyRewriteConfig.maxIterations==10 times.

Also copied the optimization from the old shape inference to not (re)infer shapes for simple ops (without subgraphs) with known and static result shapes.

### DIFF
--- a/src/Transform/ONNX/ShapeInference.cpp
+++ b/src/Transform/ONNX/ShapeInference.cpp
@@ -51,14 +51,21 @@ LogicalResult verifyOp(Operation *op) {
 // Returns failure if the op and its results are unchanged,
 // like RewritePattern::matchAndRewrite().
 // Calls shapeInfOp.emitOpError() if there is any actual failure.
-LogicalResult inferShapes(ShapeInferenceOpInterface shapeInfOp) {
-  OperationFingerPrint before(shapeInfOp.getOperation());
-  if (failed(shapeInfOp.inferShapes([](Region &region) {})))
+LogicalResult inferShapes(
+    ShapeInferenceOpInterface shapeInfOp, PatternRewriter &rewriter) {
+  Operation *op = shapeInfOp.getOperation();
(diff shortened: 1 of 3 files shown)


src/Transform/ONNX/ShapeInference.cpp
```diff
@@ -21,31 +22,88 @@ namespace onnx_mlir {
...
+  }
+}
+
+// Returns failure if the op and its results are unchanged,
+// like RewritePattern::matchAndRewrite().
+// Calls shapeInfOp.emitOpError() if there is any actual failure.
+LogicalResult inferShapes(
+    ShapeInferenceOpInterface shapeInfOp, PatternRewriter &rewriter) {
+  return tryUpdateRootInPlace(shapeInfOp, rewriter, [&]() -> LogicalResult {
+    OperationFingerPrint before(shapeInfOp);
```
philass, line 72: You seem to be getting a lot of milage out of OperationFingerPrint. Its a handy trick

src/Transform/ONNX/ShapeInference.cpp
```diff
@@ -21,31 +22,88 @@ namespace onnx_mlir {
...
+    LogicalResult outcome = shapeInfOp.inferShapes([](Region &region) {});
+    OperationFingerPrint after(shapeInfOp);
+    if (failed(outcome)) {
+      assert(after == before && "op must be unchanged on failure");
+      return shapeInfOp.emitOpError("shape inference failed");
+    }
+    // succeed only shapeInfOp or its result types changed
```
philass, line 79: Nit: Forget an if
```
// succeed only if* shapeInfOp or its result types changed
```

philass (approved):
LGTM

PR #2281, opened 2023-05-29
better error messages and error handling for incompatible types

sorenlassen:
_Update: the ssd_mobilenet_v1_10 and inception-v2-6 examples below are no longer reproducible because the failures were solved in PRs #2285 and #2282_

changed verifyShapeForBroadcastingOps() to return failure for incompatible types and display the types in the error message

ssd_mobilenet_v1_10 in the model zoo went from an assertion failure to a verify failure:
```
loc("Preprocessor/mul"): error: Broadcast op with incompatible dimensions
onnx-mlir: /workdir/onnx-mlir/src/Dialect/Mlir/IndexExprBuilder.cpp:84: void onnx_mlir::IndexExprBuilder::assertHasShapeAndRank(mlir::Value): Assertion `hasShapeAndRank(value) && "expected value with shape and rank"' failed.
```
to:
```
loc("Preprocessor/mul"): error: 'onnx.Mul' op Broadcast op with incompatible types: 'tensor<f32>', 'tensor<i32>'
```

inception-v2-6 in the model zoo went from a shape inference failure to a verify failure:
```
loc(""): error: Broadcast op with incompatible dimensions
loc(""): error: Incompatible broadcast matching 112 with 112
loc(""): error: Failed to scan parameters successfully
loc(""): error: shape inference failed
```
to: 
```
loc(""): error: 'onnx.Mul' op Broadcast op with incompatible types: 'tensor<1x64x112x112xf32>', 'tensor<64xf32>'
loc(""): error: verification failed
```

also fixed ShapeHelper 'Incompatible broadcast matching' error message from `Incompatible broadcast matching 112 with 112` to `Incompatible broadcast matching 112 with 64`

PR #2285, opened 2023-05-31
use If/Loop/Scan type inference to set result types in builder

sorenlassen:
Fixes failure for ssd_mobilenet_v1_10 from the model zoo where `onnx-mlir ssd_mobilenet_v1_10.onnx` failed with:
```
loc("Preprocessor/mul"): error: Broadcast op with incompatible dimensions
loc("add"): error: Broadcast op with incompatible dimensions
```
The error message was misleading (see the fix in PR #2281). The actual problem was that we inferred the wrong element type for a Loop result. With the fix in this PR `onnx-mlir ssd_mobilenet_v1_10.onnx --EmitONNXIR` succeeds. (But it fails with `--EmitMLIR` because some shapes are dynamic.)

The PR also replaces ImportScan() with type inference, and it moves all the type inference implementations out of gen_onnx_mlir.py (instead of python code generating tablegen code generating c++ code).


src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -616,57 +615,44 @@ class FrontendGenImpl {
...
-            Type type = funcType.getResults()[i];
-            if ((num_use_inference_outputs < 0) ||
-                (i < num_use_inference_outputs)) {
-              genericOp->getOpResult(i).setType(type);
-            } else {
-              TensorType tensorType = type.cast<TensorType>();
-              Type extendedType =
-                  UnrankedTensorType::get(tensorType.getElementType());
-              genericOp->getOpResult(i).setType(extendedType);
-            }
```
sorenlassen, line 645: I suspect that this didn't work in all cases: ImportScan() called with num_use_inference_outputs set to the value of the "num_scan_inputs" attribute, but that counts trailing scan inputs while the logic here counts leading non-scan outputs

test/mlir/onnx/parse/test_range_float_type_positive_delta_expanded.json
```diff
@@ -273,6 +273,6 @@
 // CHECK-DAG:         [[VAR_10_:%.+]] = "onnx.Add"([[PARAM_5_]], [[PARAM_2_]]) : (tensor<*xf32>, tensor<f32>) -> tensor<*xf32>
 // CHECK-DAG:         [[VAR_11_:%.+]] = "onnx.Identity"([[PARAM_5_]]) : (tensor<*xf32>) -> tensor<*xf32>
 // CHECK:             onnx.Yield [[VAR_9_]], [[VAR_10_]], [[VAR_11_]] : tensor<i1>, tensor<*xf32>, tensor<*xf32>
-// CHECK:           }) {{.*}} : (tensor<i64>, tensor<i1>, tensor<f32>) -> (tensor<i1>, tensor<2xf32>)
```
sorenlassen, line 276: this parse lit test showed that the old import of Loop was wrong: it incorrectly used the condition type `i1` as the first output type

chentong319 (approved):
LGTM!
I appreciate the change to move the type inference code away from the gen_onnx_mlir.py and into dialect definition.

PR #2288, opened 2023-05-31
enable -hybrid-transform-pass by default

sorenlassen:
inception-v2-6 in the model zoo now passes with `-hybrid-transform-pass` (after the fixes in #2280 and #2282)

by enabling the flag by default we should see this model pass in the automated model zoo regression tests

added onnx_hybrid_transform.mlir lit test with a snippet of the inception-v2-6 model to illustrate how the hybrid pass goes back and forth between shape inference and canonicalization

### DIFF
--- a/src/Compiler/CompilerOptions.cpp
+++ b/src/Compiler/CompilerOptions.cpp
@@ -227,9 +227,9 @@ llvm::cl::opt<bool> enablePatternShapeInference("pattern-shape-inference",
     llvm::cl::init(false), llvm::cl::cat(OnnxMlirCommonOptions));
 
 llvm::cl::opt<bool> enableONNXHybridPass("onnx-hybrid-pass",
-    llvm::cl::desc("Enable ONNX hybrid pass (default=false)\n"
-                   "Set to 'true' if you want to enable ONNX hybrid pass."),
-    llvm::cl::init(false), llvm::cl::cat(OnnxMlirCommonOptions));
+    llvm::cl::desc("Enable ONNX hybrid pass (default=true)\n"
+                   "Set to 'false' if you want to disable ONNX hybrid pass."),
+    llvm::cl::init(true), llvm::cl::cat(OnnxMlirCommonOptions));
 
 llvm::cl::opt<bool> verifyInputTensors("verifyInputTensors",
     llvm::cl::desc(
--- a/test/mlir/onnx/onnx_hybrid_transform.mlir
+++ b/test/mlir/onnx/onnx_hybrid_transform.mlir
@@ -1,0 +1,197 @@
+// RUN: onnx-mlir-opt -onnx-hybrid-transform %s -split-input-file | FileCheck %s
+
+// -----
+
+// Illustrates the back and forth between shape inference and the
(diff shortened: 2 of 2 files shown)


src/Dialect/ONNX/ONNXOps/OpHelper.cpp
```diff
@@ -337,6 +337,8 @@ ArrayAttr CombinedTransposePattern(PatternRewriter &rewriter,
 /// Test if the permute pattern correspond to an identity pattern.
 /// Identity patterns are {0, 1, 2, ... , rank -1}.
 bool IsIdentityPermuteVector(ArrayAttr permAttr) {
+  if (!permAttr)
+    return false;
```
sorenlassen, line 341: the test_transpose_default_cpu backend test resulted in an invocation of IsIdentityPermuteVector with a null permAttr from RemoveIdentityTransposePattern in Rewrite.td

this fix seems reasonable because clearly the pattern needs to support the case when the attribute is null because the ONNXTransposeOp can omit the attribute: see the "default" example in [the spec](https://github.com/onnx/onnx/blob/main/docs/Operators.md#Transpose)

chentong319: The spec says: `By default, reverse the dimensions`. Such default value is not represented in our dialect table gen (ONNXOps.td.inc) because, I guess, it is depended on the rank of the input, not the same as other attribute with default value that can be represented by a constant. In [shape inference for transpose]( https://github.com/onnx/onnx-mlir/blob/3c8655b01e6ea0177e62ffc79ad1f6f8a414c9db/src/Dialect/ONNX/ONNXOps/Tensor/Transpose.cpp#L38) , the perm attribute is added if it is not specified. Again, the pass ordering caused the problem. The attribute adding should be moved to the first pass.
I think it is okay to return false here.

chentong319 (approved):
LGTM!

src/Dialect/ONNX/ONNXOps/OpHelper.cpp
sorenlassen: > The spec says: `By default, reverse the dimensions`. Such default value is not represented in our dialect table gen (ONNXOps.td.inc) because, I guess, it is depended on the rank of the input, not the same as other attribute with default value that can be represented by a constant.

It seems like a bug that it is not labeled optional in the spec. How else should the "By default" text be understood. And, in fact, it's absent in the inception-v2-6 model. I should probably open an onnx issue to report this.

> In [shape inference for transpose](https://github.com/onnx/onnx-mlir/blob/3c8655b01e6ea0177e62ffc79ad1f6f8a414c9db/src/Dialect/ONNX/ONNXOps/Tensor/Transpose.cpp#L38) , the default value for perm attribute is added if it is not specified. Again, the pass ordering caused the problem. The attribute adding should be moved to the first pass.

Good observation that there's logic in ONNXTransposeOpShapeHelper::computeShape() to add permAttr if it is not specified, but it only runs if `data` has rank so it's not guaranteed to solve the problem even if shape inference runs first because `data` could be unranked until more transformation passes have run.

PR #2290, opened 2023-05-31
Clarify freeing memory in How To

cjvolzka:
# Description
* Add more clarity around destroying objects in "Freeing Tensor Memory" at http://onnx.ai/onnx-mlir/doxygen_html/OnnxMlirRuntime/index.html
* Update How To example to include destroying objects
* Update main.c example to include destroying objects


include/OnnxMlirRuntime.h
```diff
@@ -123,13 +146,21 @@
...
- * When creating an OMTensor with a specified tensor array, a user has the option to
- * create a tensor with or without ownership (omTensorCreateWithOwnership).
- *
- * If the ownership flag is set to "false", then a user is responsible for freeing memory until the last use.
- * Otherwise, if the flag is set to "true", then the destruction of the tensor (omTensorDestroy)
- * will also free any associated memory and data buffers.
+ * In general, if a caller creates a tensor object (omTensorCreate), they are
+ * responsible for deallocating the data buffer separately after the tensor is
+ * destroyed. If onnx-mlir creates the tensor (run_main_graph), then the
+ * tensor object owns the data buffer and its freed automatically when the
```
chentong319, line 152: its --> it is

include/OnnxMlirRuntime.h
```diff
@@ -123,13 +146,21 @@
...
+ * tensor is destroyed.
```
chentong319, line 153: It is a little vague to me: `when it is destroyed`. Call to OMTensorDestroy or destructor of class?

include/OnnxMlirRuntime.h
```diff
@@ -123,13 +146,21 @@
...
+ *
+ * This default behavior can be changed. When creating a tensor, a user may use
+ * omTensorCreateWithOwnership to explicitly set data buffer ownership. Otherwise,
+ * after a tenor is created, omTensorSetOwning can be used to change
+ * the ownership setting.
+ *
+ * When omTensorDestroy is called, if the ownership flag is set to "true",
+ * then the destruction of the tensor will also free any associated data buffer
+ * memory. If the ownership flag is set to "false", then the user is responsible
+ * for freeing the data buffer memory after destroying the tensor.
```
chentong319, line 163: Need any clarification on OMTensorList?

include/OnnxMlirRuntime.h
cjvolzka: Will fix on nest commit

include/OnnxMlirRuntime.h
cjvolzka: To answer the question, technically either. `OMTensorDestroy(tensor)` or `delete tensor` are built so they wind up doing the same thing.

Overall, I was intentionally vague here. `OMTensorListDestroy()` (or `delete tensor_list` indirectly calls `OMTensorDestroy()` which (depending on the tensor's data buffer ownership) may or may not free the data buffer. 

This happens in the main.c example actually. OMTensorDestroy() is never directly called but everything is freed correctly, which I checked with valgrind.

include/OnnxMlirRuntime.h
cjvolzka: Based on the subsection heading I was focused on Tensor objects. However I'd agree a summary on how TensorLists are freed would be useful. It's related enough that it make sense to include under this heading. I'll add a few sentences on the next commit.

chentong319 (approved):
LGTM!

PR #2184, opened 2023-04-20
Set node name from loc in instrumentation tool.

imaihal:
As described in PR #2115, node name attribute is often lost especially when the operation is rewritten. So, node name was not often shown in output of instrumentation tool.

This PR fixes it by setting node name from location implemented in PR #2115.

- Example in roberta-base-11 model
  - onnx_node_name attribute in SoftmaxV11 is lost by rewriting to Softmax. So, node name was lost in output, but it appears after this PR.

Input (.onnx or .onnx.mlir)
- SoftmaxV11 has onnx_node_name.
```
 %373 = "onnx.Div"(%371, %372) {onnx_node_name = "Div_90"} : (tensor<*xf32>, tensor<f32>) -> tensor<*xf32>
 %374 = "onnx.Add"(%373, %303) {onnx_node_name = "Add_91"} : (tensor<*xf32>, tensor<*xf32>) -> tensor<*xf32>
 %375 = "onnx.SoftmaxV11"(%374) {axis = 3 : si64, onnx_node_name = "Softmax_92"} : (tensor<*xf32>) -> tensor<*xf32>
 %376 = "onnx.MatMul"(%375, %369) {onnx_node_name = "MatMul_93"} : (tensor<*xf32>, tensor<*xf32>) -> tensor<*xf32>
```

Output(Main branch):
- The onnx_node_name is lost when SoftmaxV11 is rewritten to Softmax. So, node name is not shown.
```
#678) before onnx.Div Time elapsed: 0.000001 accumulated: 1680759914.813203 (Div_90)
#679) after  onnx.Div Time elapsed: 0.000670 accumulated: 1680759914.813873 (Div_90)
#680) before onnx.Softmax Time elapsed: 0.000001 accumulated: 1680759914.813874
#681) after  onnx.Softmax Time elapsed: 0.006557 accumulated: 1680759914.820431
#682) before onnx.MatMul Time elapsed: 0.000001 accumulated: 1680759914.820432 (MatMul_93)
#683) after  onnx.MatMul Time elapsed: 0.004311 accumulated: 1680759914.824743 (MatMul_93)
```
Output(PR) (.onnx)
- Node name is set from NameLoc.
```
#678) before onnx.Div Time elapsed: 0.000001 accumulated: 1681981634.161860 (Div_90)
#679) after  onnx.Div Time elapsed: 0.001742 accumulated: 1681981634.163602 (Div_90)
[...]

PR #2292, opened 2023-05-31
Fixed a bug in instrumention that caused corruption of file names

imaihal:
(This PR updates https://github.com/onnx/onnx-mlir/pull/2184)

I found filename extracted in instrumentation was broken.  For example, when I compiled following command and ran, filename for instrumentation was broken (Long file name seemed to be problem) 
`$onnx-mlir --instrument-stage=Onnx --instrument-ops=onnx.* --InstrumentBeforeOp --InstrumentAfterOp --InstrumentReportTime instrumentation_test_aaaaaaa.onnx.mlir`

This PR fixed the issue.


sorenlassen (approved):
good catch

src/Conversion/KrnlToLLVM/KrnlInstrument.cpp
```diff
@@ -72,10 +72,11 @@ class KrnlInstrumentOpLowering : public ConversionPattern {
...
         else if (auto fileLineColLoc = locIt.dyn_cast<FileLineColLoc>()) {
-          StringRef filename =
-              llvm::sys::path::filename(fileLineColLoc.getFilename().str());
-          name += filename.str() + ":" +
-                  std::to_string(fileLineColLoc.getLine()) + "-";
+          std::string filename =
+              llvm::sys::path::filename(fileLineColLoc.getFilename().str())
+                  .str();
+          name +=
+              filename + ":" + std::to_string(fileLineColLoc.getLine()) + "-";
```
sorenlassen, line 79: alternatively
```
StringRef filename = llvm::sys::path::filename(fileLineColLoc.getFilename().strref());
```
would result in a stringref pointing into the `fileLineColLoc.getFilename()` StringAttr which should work (and be more efficient although that doesn't matter in this context)

imaihal: @sorenlassen Thanks for your comments! I found this new code works through trial and error, but honestly speaking, I'm still not sure why old code didn't work, but new code worked. 

From your alternative code, should I use `llvm::sys::path::filename(fileLineColLoc.getFilename().strref()) ` also in new code?

sorenlassen: the old code didn't work because `fileLineColLoc.getFilename().str()` created a temporary string object with a lifetime until the end of the expression
```
path::filename(fileLineColLoc.getFilename().str())
```
and `path::filename(..)` returned a StringRef pointing into the temporary object (StringRef points into memory that it doesn't own, like ArrayRef or a C-style `char*` string)

the new code
```
path::filename(fileLineColLoc.getFilename().str()).str()
```
works because the final `.str()` constructs a string object with a copy of the StringRef memory returned by `path::filename(..)` before the end of the lifetime of the innermost temporary object (its lifetime extends to the end of the lifetime of the enclosing expression)

imaihal: Thanks for the details! I understood well.

PR #2294, opened 2023-06-02
Expand's shape inference returns success when shape input is unknown

tungld:
We let shape inference of Expand pass when its shape input is still unknown (say, `tensor<*xdtype>`). Otherwise, it causes shape inference pass failed.


sorenlassen (approved):
LGTM

PR #2296, opened 2023-06-02
make DequantizeLinear i32 zero check work for non-splat zeros

sorenlassen:
squeezenet1.0-13-qdq and ssd_mobilenet_v1_13-qdq from the model zoo hit this error

with this PR these models now make it through the frontend passes and fail in lowering to Krnl instead:
```
$ onnx-mlir /tmp/squeezenet1.0-13-qdq.onnx
Assertion failed: (isScalarValue(operands[i]) && "unary expected scalar additional values"), function operator(), file Elementwise.cpp, line 1991.
zsh: abort      Debug/bin/onnx-mlir /tmp/squeezenet1.0-13-qdq.onnx
```
```
$ onnx-mlir /tmp/ssd_mobilenet_v1_13-qdq.onnx
Scan output for while loop is not supported
UNREACHABLE executed at /Users/slassen/git/onnx-mlir-einsum/src/Conversion/ONNXToKrnl/ControlFlow/Loop.cpp:392!
zsh: abort      Debug/bin/onnx-mlir /tmp/ssd_mobilenet_v1_13-qdq.onnx
```

### DIFF
--- a/src/Dialect/ONNX/ONNXOps/Quantize/DequantizeLinear.cpp
+++ b/src/Dialect/ONNX/ONNXOps/Quantize/DequantizeLinear.cpp
@@ -127,9 +127,11 @@ LogicalResult ONNXDequantizeLinearOp::verify() {
     //   return emitOpError("x and x_zero_point must have the same data type");
 
     if (getElementType(zero.getType()).isInteger(32) && zeroLen != 0)
-      if (auto values = getElementAttributeFromONNXValue(zero))
-        if (!values.isSplat() || !values.getSplatValue<APInt>().isZero())
+      if (auto values = getElementAttributeFromONNXValue(zero)) {
+        WideNum zero = WideNum::widen<BType::INT32>(0);
+        if (!ElementsAttrBuilder::allEqual(values, zero))
           return emitOpError("x_zero_point must be 0 for data type int32");
+      }
   }
 
   if (scaleLen == ShapedType::kDynamic && zeroLen == ShapedType::kDynamic) {


src/Dialect/ONNX/ONNXOps/Quantize/DequantizeLinear.cpp
```diff
@@ -127,9 +127,11 @@ LogicalResult ONNXDequantizeLinearOp::verify() {
     //   return emitOpError("x and x_zero_point must have the same data type");
 
     if (getElementType(zero.getType()).isInteger(32) && zeroLen != 0)
-      if (auto values = getElementAttributeFromONNXValue(zero))
-        if (!values.isSplat() || !values.getSplatValue<APInt>().isZero())
+      if (auto values = getElementAttributeFromONNXValue(zero)) {
+        WideNum zero = WideNum::widen<BType::INT32>(0);
```
chentong319, line 131: I am not sure why x_zero_point should be compared to constant 0. I did not see that requirement in the specification.
Another comment is about allEqual: if the element type of 'values' is not INT32, will allEqual return false?

sorenlassen: > I am not sure why x_zero_point should be compared to constant 0. I did not see that requirement in the specification.

The spec says "In the case of dequantizing int32, there's no zero point (zero point is supposed to be 0)"
https://github.com/onnx/onnx/blob/main/docs/Operators.md#DequantizeLinear

> Another comment is about allEqual: if the element type of 'values' is not INT32, will allEqual return false?

WideNum is an untagged union so allEqual(values, zero) assumes zero has the same type as values' element type, it cannot check that it does. Here the if condition ensures values has element type i32 and correspondingly we encode zero as INT32.

chentong319 (approved):
LGTM!

sorenlassen:

PR #2297, opened 2023-06-02
use graph nodes with substituted attributes

sorenlassen:
This PR fixes a bug in function inlining where the substituted attributes were lost (accidentally imported the nodes before substitution).

I discovered the bug with @MikeHolman's  model maf_gpt2_tiny_display_name.onnx with model local function.

Before the fix in this PR compilation failed with:
```
$ onnx-mlir maf_gpt2_tiny_display_name.onnx
Unsupported data type encountered.
UNREACHABLE executed at /Users/slassen/git/onnx-mlir-einsum/src/Dialect/ONNX/ONNXOps/OpHelper.cpp:619!
```
after the fix the compilation fails with:
```
Assertion failed: (isa<To>(Val) && "cast<Ty>() argument of incompatible type!"), function cast, file Casting.h, line 566.
```
because there is an If op where the then branch has a tensor type `tensor<?xf32>` and the else branch has NoneType - let's get back to fixing that after this PR

### DIFF
--- a/src/Builder/FrontendDialectTransformer.cpp
+++ b/src/Builder/FrontendDialectTransformer.cpp
@@ -1207,7 +1207,7 @@ private:
         BindOnnxName(name, value);
       }
 
-      for (auto &fb_node : functionProto.node()) {
+      for (auto &fb_node : graph.node()) {
         ImportNode(fb_node);
       }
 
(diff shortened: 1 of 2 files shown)


philass (approved):
LGTM

sorenlassen:

sorenlassen:
I filed issue #2298 to track the remaining obstacles to compile maf_gpt2_tiny_display_name.onnx

PR #2302, opened 2023-06-06
Gemm const prop

sorenlassen:


sorenlassen:
> Did you have a benchmark where GEMM was calculated on constants? Interested to see where this use case came.

no, I just implemented this because it was easy, since ElementsAttrBuilder::matMul() already does most of the work, and with a view to adapt this to get constant folding for Gemm in the future

PR #2303, opened 2023-06-06
MatMul const prop

sorenlassen:


AlexandreEichenberger (approved):
LGTM

Do you see this optimization also with float? If so, and it has a significant perf impact, should we consider a mode where we play fast and loose with NaN?

AlexandreEichenberger:
Thanks for the clarification

PR #2305, opened 2023-06-07
separate krnl logic out of ONNXOps/OpHelper

sorenlassen:
Ideally the dependency between the Krnl and ONNX dialects should be one-way since the ONNX dialect is used with other targets than Krnl. Therefore I did this refactor to not include the krnl header in ONNXOps/OpHelper.cpp.


chentong319 (dismissed):
LGTM!

chentong319 (approved):
LGTM!

PR #2301, opened 2023-06-06
reenable pattern-based shape inference by default

sorenlassen:
Reverting PR #2270 which disabled pattern-based shape inference by default.

The problems that caused us to disable it in PR #2270 were all shape inference bugs that were hidden by the legacy shape inference (see the bug fixes in PR #2273).

Let's go back to shape-based shape inference by default so we can surface and fix any other hidden bugs and ultimately phase out the legacy shape inference which has bit rotted, see PR #2243.

PR #2307, opened 2023-06-07
implement ONNXReshapeOp::verify() and don't fail in inferShapes()

sorenlassen:
with this change tiny-yolov3-11.onnx compiles without error message and no longer fails shape inference with `-pattern-shape-inference=false`

see the last comments in PR #2301


chentong319 (approved):
LGTM!

PR #2311, opened 2023-06-08
only decompose ConvTranspose when spatial dims are static

sorenlassen:
addresses failure of MaskRCNN-10.onnx from model zoo

before it failed with:
```
Assertion failed: (!inputType.isDynamicDim(axis) && "Spatial dimensions for input data tensor need to be static."), function emitSplitAxisOutputLength1, file Decompose.cpp, line 326
```
after it fails with:
```
loc("1140"): error: 'arith.extui' op result #0 must be signless-fixed-width-integer-like, but got 'ui8'
```


sorenlassen:
@imaihal can I ask you to review this PR?

imaihal (approved):
@sorenlassen LGTM. Thanks for updating the code.

PR #2315, opened 2023-06-09
float8 support

sorenlassen:
Created float8 data types float_8e4m3fn, float_8e4m3fnuz, float_8e5m2, float_8e5m2fnuz to represent the 4 new float8 data types introduced in onnx 1.14.0: https://onnx.ai/onnx/technical/float8.html

Upgraded ONNX Constant op to v19 so we can represent float8 types in constants and added fp8 parse lit tests to demonstrate that we can import float8 tensors from onnx models.


utils/testing/floatingpoint8_data.py
philass, line 1: Nit: The name of this file `floatingpoint8_data.py` doesn't seem to be consistent with naming conventions used in other places. Maybe `fp8_data.py` is better? 

Feel free to ignore this comment as well

philass (approved):
LGTM! feel free to ignore my suggestion as well

utils/testing/floatingpoint8_data.py
sorenlassen: good suggestion

done

PR #2317, opened 2023-06-13
override float_16::to/fromFloat() with per-CPU optimized versions 

sorenlassen:
When the CPU is known to support native conversion between float and float_16, use them in class float_16 below to override the slow default APFloat-based conversions.

Implemented for ARM64 when `__ARM_FP16_FORMAT_IEEE` is defined and for x86-64 when `__F16C__` is defined.

Added benchmarks of conversion between float and float_16 in TestFloatingPoint16.cpp.

Measured the benchmarks on a MacBook pro with Apple M1 Pro cpu. Before the per-CPU optimization:
```
Benchmark                          Time             CPU   Iterations
--------------------------------------------------------------------
BM_F32_TO_FP16<float_16>     2655030 ns      2654890 ns          263
BM_F32_TO_FP16<bfloat_16>    2741322 ns      2741320 ns          259
BM_FP16_TO_F32<float_16>     1824425 ns      1824426 ns          383
BM_FP16_TO_F32<bfloat_16>    1817702 ns      1817688 ns          385
```
After:
```
BM_F32_TO_FP16<float_16>      142300 ns       142280 ns         4855
BM_F32_TO_FP16<bfloat_16>    2312391 ns      2312290 ns          300
BM_FP16_TO_F32<float_16>      224026 ns       224026 ns         3124
BM_FP16_TO_F32<bfloat_16>    1832315 ns      1832315 ns          381
```

On another MacBook Pro with Intel Core i7-7820HQ cpu I had to configure the build with `-DCMAKE_CXX_FLAGS=-march=native` to define `__F16C__`. Before the per-CPU optimization:
```
BM_F32_TO_FP16<float_16>     4313245 ns      4304093 ns          140
BM_F32_TO_FP16<bfloat_16>    4427924 ns      4423289 ns          159
BM_FP16_TO_F32<float_16>     3426035 ns      3422583 ns          206
BM_FP16_TO_F32<bfloat_16>    3532166 ns      3517064 ns          204
```
After:
```
BM_F32_TO_FP16<float_16>       99106 ns        99069 ns         6525
BM_F32_TO_FP16<bfloat_16>    4358961 ns      4351302 ns          162
BM_FP16_TO_F32<float_16>      162100 ns       161966 ns         4204
BM_FP16_TO_F32<bfloat_16>    3459624 ns      3453559 ns          202
```

### DIFF
--- a/docs/BuildOnLinuxOSX.md
+++ b/docs/BuildOnLinuxOSX.md
@@ -42,7 +42,7 @@ The `MLIR_DIR` cmake variable must be set before building onnx-mlir. It should p
 
 This project uses lit ([LLVM's Integrated Tester](https://llvm.org/docs/CommandGuide/lit.html)) for unit tests. When running cmake, we can also specify the path to the lit tool from LLVM using the `LLVM_EXTERNAL_LIT` variable but it is not required as long as MLIR_DIR points to a build directory of llvm-project. If `MLIR_DIR` points to an install directory of llvm-project, `LLVM_EXTERNAL_LIT` is required.
 
-To build ONNX-MLIR, use the following commands:
+To build ONNX-MLIR, use the following commands (maybe with additional `-DCMAKE_CXX_FLAGS` argument described [below](#enable-cpu-optimizations)):
 
 [same-as-file]: <> ({"ref": "utils/install-onnx-mlir.sh", "skip-doc": 2})
 ```bash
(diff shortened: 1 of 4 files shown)


AlexandreEichenberger:
Can there be some CMAKE magic that set these flags automatically when building the compiler on a machine that supports it? Or under a uniform flag that is common to all architectures (and default to nothing for arch not supporting it)?

Or if too complicated/not appropriate, can there be documentation (either in the cmakefile or in the .md doc) on how to enable this?

sorenlassen:
> Can there be some CMAKE magic that set these flags automatically when building the compiler on a machine that supports it? Or under a uniform flag that is common to all architectures (and default to nothing for arch not supporting it)?
>
> Or if too complicated/not appropriate, can there be documentation (either in the cmakefile or in the .md doc) on how to enable this?

since there's a tradeoff between compile speed and portability, it may be best to make users make an explicit choice to target the faster and less portable CPU features

I added an explanation of how to set the CXX_FLAGS in BuildOnLinuxOSX.md

I can't test on Windows and don't know the right MSVC flags, so I'll leave that to anyone who wants to put that to use on Windows, if that's ok

docs/BuildOnLinuxOSX.md
```diff
@@ -75,6 +75,10 @@ The environment variable `$pythonLocation` may be used to specify the base direc
 
 After the above commands succeed, an `onnx-mlir` executable should appear in the `Debug/bin` or `Release/bin` directory.
 
+### Enable CPU Optimizations
```
AlexandreEichenberger, line 78: maybe just clarify that CPU optimization are here to make the compiler run faster. That may clarify some confusion about what is accelerated

AlexandreEichenberger (approved):
LGTM

docs/BuildOnLinuxOSX.md
sorenlassen: good point, I added "To make the compiler run faster (without any affect on the generated code)"

Issue #2279, opened 2023-05-26
SIMD is not working for IsInf


hamptonm1:
When running the lit test for operation IsInf using SIMD, the program crashes with the following error:

`lhs.getType() == rhs.getType() && "expected same type"` for line 387. After talking to @AlexandreEichenberger he helped me realize that the types are different we are trying to compare a scalar to a vector and that is why it fails. So we need to utilize `create.vec.splat(vecType,constant)` to ensure the conversion to vector.

However, when I attempted to do so I continued to receive an error. So I am handing this over to @AlexandreEichenberger for further investigation. 

Here is skeleton code I wrote to use `splat`:

```
  // For vectors, need to splat the constant.
  VectorType vecType = inputElemType.dyn_cast<VectorType>();
  Value negInfVec = create.vec.splat(vecType, negInf);
  Value posInfVec = create.vec.splat(vecType, posInf);
```

hamptonm1:
Also for the IsInf operation.... I was using inputElemType opposed to elementType because for IsInf, when I use elementType it gives me the output value of `i1*` which corresponds to the below error message. It is complaining that the input is an integer and that this custom operation does not support it. This occurs because elementType is the output value which is an integer and the input value is supposed to be a float. 

```
this op does not supports custom scalar for integers
UNREACHABLE executed at /workdir/onnx-mlir/src/Conversion/ONNXToKrnl/Math/Elementwise.cpp:45!
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
Stack dump:
0.      Program arguments: /workdir/onnx-mlir/build/Debug/bin/onnx-mlir-opt --shape-inference --convert-onnx-to-krnl /workdir/onnx-mlir/test/mlir/onnx/onnx_lowering.mlir -split-input-file
 #0 0x0000004004d25eff llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) (/workdir/onnx-mlir/build/Debug/bin/onnx-mlir-opt+0x4d25eff)
 #1 0x0000004004d238f4 SignalHandler(int) Signals.cpp:0:0
 #2 0x0000004008de3420 __restore_rt (/lib/x86_64-linux-gnu/libpthread.so.0+0x14420)
 #3 0x00000040091cf00b raise /build/glibc-SzIz7B/glibc-2.31/signal/../sysdeps/unix/sysv/linux/raise.c:51:1
[...]

PR #2319, opened 2023-06-14
Support for SIMD improved

AlexandreEichenberger:
Unfortunately, MLIR does not lend itself to easily simdized code that compress results to different bit width. 

For `IsInf`, the inputs are float, and the output are bits. Now most SIMD arch would represent "true" or "false" using the same number of repeating bits as in the input (e.g. 32 bit float, IsInf would result in 32 bit of zero or 32 bits of 1). That is relatively easy to generate code for. But MLIR generates a single bit vector. So a 4x 32 bit SIMD unit of 128 bits would need to generate a 4x1 bit result... but SIMD units typically only generate 128 bit results, not a subset of it.

Thus, element wise operations that reduces (or increases) the sizes of their output with respect of their inputs cannot be easily handled at this time.

This PR nevertheless enable SIMD for INF constants, so that if their outputs are used in a select, they would be able to be used as SIMD operations.

This PR responds to issue #2279


hamptonm1 (approved):
LGTM! Thank you for looking into this.

PR #2321, opened 2023-06-15
Add --store-constants-to-file that puts constants to a binary file

tungld:
With option `-store-constants-to-file`, constants will be stored on a single binary file that has the same name as `model.so` but is postfixed by `.constants.bin`.

In other words, two files are generated for a model:
- `.so` file for the model
- `.constants.bin` for constants

`constants-to-file-single-threshold` (value in KB, default to 1) and `constants-to-file-total-threshold` (Value in GB, default to 2) to control the size of a single constant and all satisfied constants.

For example, `constants-to-file-single-threshold=1` and `constants-to-file-total-threshold=1.5` means store to file if the total size of all `>1 KB` constants is greater than `1.5GB`.

Added two functions to load and free constant buffers: `omLoadConstantsFromFile(filename)` and `omFreeBuffersForConstants()`. C++/Python OMSession will call them at the session's constructor and destructor. C users have to call them explicitly.


tungld:
@gongsu832 do you have any suggestion on why `make check-onnx-backend-numerical` failed on Jenkins? 

I included a new one `check-onnx-backend-constants-to-file` into `check-onnx-backend-numerical`. Currently, only `resnet50` is tested by `check-onnx-backend-constants-to-file`. Both `check-onnx-backend-numerical` and `check-onnx-backend-constants-to-file` passed on my local z14 machine.

gongsu832:
```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
Debug/inference_backend.py:1262: in assert_similar_outputs
    super(InferenceBackendTest, cls).assert_similar_outputs(ref_outputs, outputs, rtol, atol)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

cls = <class 'inference_backend.InferenceBackendTest'>
ref_outputs = [array([[0.001, 0.001, 0.001, 0.001, 0.001, 0.001, 0.001, 0.001, 0.001,
        0.001, 0.001, 0.001, 0.001, 0.001, 0.0..., 0.001, 0.001,
        0.001, 0.001, 0.001, 0.001, 0.001, 0.001, 0.001, 0.001, 0.001,
        0.001]], dtype=float32)]
outputs = [array([[nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan,
        nan, nan, nan, nan, nan, nan, nan, n...n, nan, nan, nan, nan, nan,
        nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan]],
      dtype=float32)]
rtol = 0.001, atol = 1e-07

    @classmethod
    def assert_similar_outputs(
        cls,
        ref_outputs: Sequence[Any],
        outputs: Sequence[Any],
        rtol: float,
        atol: float,
    ) -> None:
        np.testing.assert_equal(len(outputs), len(ref_outputs))
        for i in range(len(outputs)):
            if isinstance(outputs[i], (list, tuple)):
                for j in range(len(outputs[i])):
                    cls.assert_similar_outputs(
                        ref_outputs[i][j], outputs[i][j], rtol, atol
                    )
            else:
                np.testing.assert_equal(outputs[i].dtype, ref_outputs[i].dtype)
                if ref_outputs[i].dtype == object:  # type: ignore[attr-defined]
                    np.testing.assert_array_equal(outputs[i], ref_outputs[i])
                else:
>                   np.testing.assert_allclose(
                        outputs[i], ref_outputs[i], rtol=rtol, atol=atol
                    )
E                   AssertionError: 
E                   Not equal to tolerance rtol=0.001, atol=1e-07
E                   
E                   x and y nan location mismatch:
E                    x: array([[nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan,
E                           nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan,
E                           nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan,...
E                    y: array([[0.001, 0.001, 0.001, 0.001, 0.001, 0.001, 0.001, 0.001, 0.001,
E                           0.001, 0.001, 0.001, 0.001, 0.001, 0.001, 0.001, 0.001, 0.001,
E                           0.001, 0.001, 0.001, 0.001, 0.001, 0.001, 0.001, 0.001, 0.001,...

/usr/local/lib/python3.10/dist-packages/onnx/backend/test/runner/__init__.py:212: AssertionError
```
It appears that the model outputs are all nans.

tungld:
Yes, I guess somehow the file for constants, `resnet50.constants.bin`, was not loaded or it was deleted or not found. Not sure what is the difference between Jenkins and my local environment.

tungld:
Perhaps, I am using onnx 1.13.1 and the model was downloaded differently.

tungld:
@gongsu832 when I updated to onnx 1.14.0, I got the same error (a lot of NaN) as Jenkins. After that, I remove all downloaded models in the folder `build/test`, and `check-onnx-backend-constants-to-file` passed. I wonder if there is any mix of models or test data on Jenkins when we updated to onnx 1.14.0.

gongsu832:
> @gongsu832 when I updated to onnx 1.14.0, I got the same error (a lot of NaN) as Jenkins. After that, I remove all downloaded models in the folder `build/test`, and `check-onnx-backend-constants-to-file` passed. I wonder if there is any mix of models or test data on Jenkins when we updated to onnx 1.14.0.

If there is any code change in the onnx-mlir repo, the Jenkins CI will build a fresh docker image from scratch. So there shouldn't be any "leftover" from a previously failed build.

gongsu832:
@tungld looks like the tests are passing now.

tungld:
Thanks @gongsu832 for looking at this! I seemed a bug in my code caused the issue.

test/backend/common.py
```diff
@@ -142,4 +148,5 @@ def compile_model(model, emit):
     # Check if specific instruction are included in the compiled model.
     check_instruction(name + "_cpu", exec_name)
 
+    print("hitung", exec_name, file=sys.stderr)
```
sorenlassen, line 151: remove?

sorenlassen (commented):
I haven't reviewed the PR in detail but just want to mention that it looks good to me to do this separately from how we work with constants on disk with mlir::ElementsAttr for onnx external_data, since this topic came up in the onnx-mlir meetings this month.

src/Runtime/OMExternalConstant.inc
```diff
@@ -0,0 +1,157 @@
...
+      // No alignment.
+      free(ptr);
+    } else {
+      // Get the original malloc'd address from where we put it and free it.
+      void *original_ptr = ((void **)ptr)[-1];
+      free(original_ptr);
+    }
+  }
+}
+
```
gongsu832, line 94: There are already aligned memory allocation functions: `posix_memalign`, on windows it's `_aligned_malloc`.

src/Runtime/OMExternalConstant.inc
```diff
@@ -0,0 +1,157 @@
...
+#define XOR(a, b) (!(a) != !(b))
+
+void checkEndianness(const char constPackIsLE) {
+  if (XOR(IS_SYSTEM_LE(), constPackIsLE)) {
+    fprintf(stderr, "Constant pack is stored in a byte order that is not "
+                    "native to this current system.");
+    exit(1);
+  }
+}
+
```
gongsu832, line 33: In general, at least for ONNX, the convention is that everything written out on the disk is Little Endian. So if we are running on Big Endian, shouldn't we just do an endian conversion instead of exiting?

test/backend/common.py
tungld: Done. Thanks!

src/Compiler/CompilerUtils.cpp
```diff
@@ -429,7 +429,8 @@ static int genLLVMBitcode(const mlir::OwningOpRef<ModuleOp> &module,
   // Use the LLVM's 'opt' command to optimize the bitcode.
   std::string optPath = getToolPath("opt", kOptPath);
   Command optBitcode(/*exePath=*/optPath);
-  setXoptOption({"--code-model", modelSizeStr[modelSize]});
+  if (modelSize == ModelSize::small || modelSize == ModelSize::large)
+    setXoptOption({"--code-model", modelSizeStr[modelSize]});
```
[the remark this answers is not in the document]
tungld: Oops, these were my old code, removed it.

src/Runtime/OMExternalConstant.inc
tungld: Use them now. Thanks for the suggestion!

src/Runtime/OMExternalConstant.inc
tungld: ONNX is an neural exchange format and it's reasonable to use Little Endian. But for a compiled model, I am afraid that it's highly depending on the target architecture, e.g. LLVMIR contains information for a specific architecture.

src/Compiler/CompilerOptions.cpp
```diff
@@ -104,7 +104,10 @@ llvm::cl::opt<ModelSize> modelSize("modelSize",
                          "default code model"),
         clEnumVal(large,
             "Generate code for the large model. "
-            "Global constants are put into large read-only data section.")),
+            "Global constants are put into large read-only data section."),
+
+        clEnumVal(huge, "Generate code for the huge model. "
+                        "Global constants are put into files.")),
```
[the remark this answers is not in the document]
tungld: Sure, added more explanation to the option `--store-constants-to-file`

src/Runtime/OMExternalConstant.inc
```diff
@@ -0,0 +1,157 @@
...
+/// Open a file in binary mode
+///
+/// \param[in] fname File path
+/// \param[in] isLE Data in the file is little-endian or not
+///
+/// \return A FILE pointer
+/// TODO: find a new home for this function.
+void *omOpenBinaryFile(char *fname, int64_t isLE) {
```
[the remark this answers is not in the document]
tungld: These are internal to onnx-mlir, they will be called from `omLoadConstantsFromFile` and `omFreeBuffersForConstants` that are generated by the krnl-to-llvm lowering and not thread-safe. I added an explanation about this into the option `--store-constants-to-file`. Hope it is clearer now.

src/Runtime/OMExternalConstant.inc
```diff
@@ -0,0 +1,157 @@
...
+/// \param[in] ptr Pointer returned by omLoadExternalConstant()
+/// \param[in] fname File name at the current folder
+/// \param[in] offset Offset from the beginning of file
+/// \param[in] size Size in bytes to copy data from the binary file
+/// \param[in] align Alignment
+/// \param[in] isLE Data in the binary file is little-endian or not
+///
+/// \return None
+///
+void omLoadExternalConstant(
```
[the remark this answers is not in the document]
tungld: I see. I added  `set the point to NULL` after a buffer is freed in the function `omFreeBuffer`.

test/backend/variables.py
```diff
@@ -78,6 +79,19 @@ def get_args_from_env():
...
+        help="whether store constants to file or not, passed to the compiler",
+    )
+    parser.add_argument(
+        "--constants_to_file_total_threshold",
+        type=float,
+        default=(0.000001 if TEST_CONSTANTS_TO_FILE else 2),
+        help="total threshold to trigger constants to file. Set it to a small "
+        "value, 1024 bytes, since models in the model zoo are small",
+    )
     parser.add_argument(
```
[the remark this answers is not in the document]
tungld: Let me keep `--constants_to_file`. The reason is when we check it, we use `args.constants_to_file`, which is consistent to the argument name.

If we use `--constants-to-file`, the check is still using underscore `args.constants_to_file`. There is no `args.constants-to-file`. So, it's quite confusing, for example when we search places that use `--constants-to-file`.

src/Runtime/OMExternalConstant.inc
```diff
@@ -0,0 +1,157 @@
...
+/// \param[in] isLE Data in the file is little-endian or not
+///
+/// \return A FILE pointer
+/// TODO: find a new home for this function.
+void *omOpenBinaryFile(char *fname, int64_t isLE) {
+  checkEndianness(isLE);
+  FILE *fileptr;
+  fileptr = fopen(fname, "rb"); // Open the file in binary mode.
+  if (fileptr == NULL) {
+    perror("Error: OpenBinaryFile");
```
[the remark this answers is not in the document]
tungld: Now the file name is printed out. Thanks!

src/Runtime/OMExternalConstant.inc
gongsu832: It's in general a good idea to have a FIXED endianness (in this case little endian) when the data is written out on the disk. Because this way even without any metadata to indicate the endianness of the file, you can move the data file around and use it. That was how we did it on DLC Classic.

test/backend/variables.py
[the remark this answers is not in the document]
gongsu832: That's fine.

src/Runtime/OMExternalConstant.inc
AlexandreEichenberger: @tungld is this issue resolved?

AlexandreEichenberger (approved):
LGTM, would be good to push this, and possibly open an issue if there are unresolved issues. We do want to gen very large models as early as possible using the normal build.

src/Runtime/OMExternalConstant.inc
tungld: I don't see any benefit of using a default (little) endian, e.g. do you often exchange a compile model between little and big endian machines?  How often do we do this? If we only compile and run a model on a big endian machine, why we need to do byteswap for gigabytes of constants?

How about discussing about this more, and possibly another PR to solve this?

gongsu832: We have no control over whether people will move their model from one arch to another or not. Without metadata, it's difficult to do things without an agreed-upon endianness for constants written on the disk. Unfortunately, the rest of the world besides z pretty much defaults to little endian on disk. I think the proper solution is that the constants should always be written out in a format that has metadata to indicate its endianness.

tungld: Yes, totally agree with having metadata to indicate endianness. Let me do it in a follow-up PR. I'll create an issue to track some remaining works after we land this patch.

PR #2323, opened 2023-06-16
const prop Less, Greater, LessOrEqual, GreaterOrEqual

sorenlassen:


src/Transform/ONNX/ConstProp.td
```diff
@@ -440,16 +452,63 @@ def DivOnesOnRhs : Pat<
...
-// Constant Propagation for Equal
+//===----------------------------------------------------------------------===//
+// Constant propagate ONNXEqualOp
+//===----------------------------------------------------------------------===//
+
 def EqualConstProp : Pat<
     // From equal(c1, c2).
-    (ONNXEqualOp:$equalOp (ONNXConstantOp:$lhs $_, $_, $_, $_, $_, $_, $_, $_),
-                          (ONNXConstantOp:$rhs $_, $_, $_, $_, $_, $_, $_, $_)),
+    (ONNXEqualOp:$equalOp
```
philass, line 461: Below you use `$result` instead of `$equalOp`

src/Transform/ONNX/ConstProp.td
```diff
@@ -440,16 +452,63 @@ def DivOnesOnRhs : Pat<
...
     // To c1 == c2
     (CreateEqualOfTwoConst $equalOp, $lhs, $rhs),
     // constraints
     [(IsFromDenseONNXConstantOp:$lhs), (IsFromDenseONNXConstantOp:$rhs)]>;
 
+//===----------------------------------------------------------------------===//
+// Constant propagate ONNXLessOp
+//===----------------------------------------------------------------------===//
+
+def LessConstPropPattern: Pat<
```
philass, line 473: For `Equal`, `Less` and `Greater` I don't think you need the `$result` for the builder, because of these lines https://github.com/onnx/onnx-mlir/blob/1278d70de9c03fdeebe8298c0c8c1c474764835c/utils/gen_onnx_mlir.py#L506C1-L509

But I understand why it might be worth leaving as is for now, until `GreaterOrEqual` and `LessOrEqual` are added to that list. (I can do that in a separate PR).

philass (approved):

src/Transform/ONNX/ConstProp.td
sorenlassen: good point

ok, I'll leave as is for now

src/Transform/ONNX/ConstProp.td
sorenlassen: ok, changed to `$result` here for consistency

PR #2324, opened 2023-06-16
remove unsafe LessOpSameCastPattern canonicalization pattern

sorenlassen:
the LessOpSameCastPattern canonicalization pattern
```
def LessOpSameCastPattern: Pat<
   (ONNXLessOp (ONNXCastOp:$cast_a $a, $to1), (ONNXCastOp $b, $to2)),
   (ONNXLessOp $a, $b),
   [(HaveSameElementType $a, $b), (Equal $to1, $to2),
    (HaveSameElementTypeBitWidth $a, $cast_a),
    (ElementTypeIsNotUnsigned:$cast_a)]
>;
```
is quite limited in which types it applies to, it seems like the only practical uses are to convert between i64 and f64, i32 and f32, or i16 and f16 (or between i8 and the upcoming 8 bit floating point types)

in all cases, it's unsafe because there are always different numbers in one data type that get rounded or saturated to the same number

for instance, this lit test in onnx_canonicalization.mlir
```
func.func @test_less(%arg0 : tensor<i32>, %arg1 : tensor<i32>) -> tensor<i1> {
  %0 = "onnx.Cast"(%arg0) {to = f32} : (tensor<i32>) -> tensor<f32>
  %1 = "onnx.Cast"(%arg1) {to = f32} : (tensor<i32>) -> tensor<f32>
  %2 = "onnx.Less"(%0, %1) : (tensor<f32>, tensor<f32>) -> tensor<i1>
  onnx.Return %2 : tensor<i1>
  // CHECK-LABEL: test_less
  // CHECK: [[RES:%.]] = "onnx.Less"(%arg0, %arg1) : (tensor<i32>, tensor<i32>) -> tensor<i1>
  // CHECK: onnx.Return [[RES]] : tensor<i1>
}
```
returns `false` instead of the correct `true` if arg0 is N and arg1 is N+1 for most N greater than 2^24 or smaller than -2^24 because N and N+1 typically get rounded to the same f32 numbers given that f32 only has 24 bit mantissa

### DIFF
--- a/src/Dialect/ONNX/Rewrite.cpp
+++ b/src/Dialect/ONNX/Rewrite.cpp
@@ -904,7 +904,6 @@ void ONNXLayoutTransformOp::getCanonicalizationPatterns(
 /// on the ONNXLessOp.
 void ONNXLessOp::getCanonicalizationPatterns(
     RewritePatternSet &results, MLIRContext *context) {
-  results.insert<LessOpSameCastPattern>(context);
   results.insert<BinaryOpBroadcastAxisPattern<ONNXLessOp>>(context);
 }
 
--- a/src/Dialect/ONNX/Rewrite.td
+++ b/src/Dialect/ONNX/Rewrite.td
@@ -931,21 +931,8 @@ def ConstantOpNormalizationPattern6: Pat<
 >;
 
 //===----------------------------------------------------------------------===//
-// Canonicalization for ONNXLessOp
-//===----------------------------------------------------------------------===//
-
-def LessOpSameCastPattern: Pat<
-   (ONNXLessOp (ONNXCastOp:$cast_a $a, $to1), (ONNXCastOp $b, $to2)),
-   (ONNXLessOp $a, $b),
-   [(HaveSameElementType $a, $b), (Equal $to1, $to2),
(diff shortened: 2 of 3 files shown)


sorenlassen:
@tungld can I ask you to review this PR? I believe you introduced the pattern in PR #1449

sorenlassen:
just now I realize that this question was discussed in depth in PR #1449 with @etiotto which helps me understand why you chose this set of constraints on the pattern

nonetheless, as explained in the PR description, this pattern is not safe in all cases so I wonder if we should remove it

@tungld do you have the model that motivated you to add LessOpSameCastPattern? or can we express it as a lit test like `@test_loop_derive_max_trip_count`?

maybe we can solve the problem by adding support for floating point numbers in LoopOpRewriteMaxTripCountPattern::matchOp()

for what it's worth LessOpSameCastPattern came to my attention when I upgraded Cast op to v19 in PR #2322 and I wondered how to update this rule to add the new 'saturate' attribute

sorenlassen:
after giving this more thought I've concluded that the pattern is not getting in the way of PR #2322 so I'm closing this PR

(but I still wonder if we should try to find a safer alternative and may revisit this question in the future)

PR #2325, opened 2023-06-17
support f16 in entry point input/output signature

sorenlassen:
f16 in entry point input/output signatures is now written with double quotes like other data types.

Also added f16 support to OMExecutionService and enabled float16 backend tests for PowerPC64 and MacOS platforms (both x84_64 and Arm). The float16 tests fails on s390x, as previously noted in backend test comments. They fail on linux amd64 and I couldn't figure out why. I'll open an issue about that.

Left a TODO to make omTensorSort() support float16. This will be needed to make TopK lowering to LLVM work for float16. I'll open an issue about that.


src/Runtime/PyExecutionSessionBase.cpp
```diff
@@ -21,6 +22,36 @@ SUPPRESS_WARNINGS_POP
...
 #include "PyExecutionSessionBase.hpp"
 
+namespace pybind11 {
+namespace detail {
+
+// Note: Since float16 is not a builtin type in C++, we register
+// onnx_mlir::float_16 as numpy.float16.
+// Ref: https://github.com/pybind/pybind11/issues/1776
+//
+// This implementation is copied from https://github.com/PaddlePaddle/Paddle
```
AlexandreEichenberger, line 32: I assume that since you included this code in a file with Apache license header, the code copied here is compatible with all its requirements.

test/backend/inference_backend.py
```diff
@@ -1182,6 +1179,22 @@ def get_test_models():
         model_test_to_enable = variables.model_test_for_constants_to_file
         test_to_enable = variables.test_for_constants_to_file
 
+    if args.nofloat16:
```
AlexandreEichenberger, line 1182: Can you add a comment that indicates what is used to gate out the f16 tests?

PR #2328, opened 2023-06-20
fix check-onnx-backend-input-verification on MacOS

sorenlassen:
before this PR it failed with:
```
Undefined symbols for architecture x86_64 (or arm64):
  "___errno_location", referenced from:
      _run_main_graph in add_wrong_data_type.o
```
on both x86_64 and arm64

as well as:
```
  File "/Users/sorenlassen/git/onnx-mlir-einsum/build/test/backend/RelWithDebInfo/input_verification_backend.py", line 170, in _redirect_stdout
    c_stdout = ctypes.c_void_p.in_dll(libc, 'stdout')
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: dlsym(RTLD_DEFAULT, stdout): symbol not found
```
which is fixed with the solution from stackoverflow: https://stackoverflow.com/a/47616945


sorenlassen:
I'm trying a simpler fix for the `__errno_location()` problem in PR #2330

PR #2267, opened 2023-05-25
[WIP] Check return values of zdnn calls.

negiyas:
Objective of this PR is to check return values of zDNN APIs (e.g. zdnn_softmax), and print error messages with the return values and API names. However this PR is still work in progress, because;

(1) This PR checks return value of zdnn_softmax only (to minimize the code modifications and errors while debugging).
(2) This PR causes the following errors in krnl-to-llvm conversion.

```
$ ./build/Debug/bin/onnx-mlir-opt --maccel=NNPA --convert-krnl-to-llvm="use-opaque-pointers=false" ./test/mlir/accelerators/nnpa/conversion/lower-all-to-llvm-constant-shape-typed-pointer.mlir -split-input-file
within split at ./test/mlir/accelerators/nnpa/conversion/lower-all-to-llvm-constant-shape-typed-pointer.mlir:1 offset :14:3: error: 'llvm.return' op must be the last operation in the parent block
  "zlow.softmax"(%input, %work_area, %shape, %res) {act_func = "ACT_NONE"} : (memref<1x1x1x1x32x64xf16>, memref<8192xi8>, memref<3xi64>, memref<1x1x1x1x32x64xf16>) -> ()
  ^
within split at ./test/mlir/accelerators/nnpa/conversion/lower-all-to-llvm-constant-shape-typed-pointer.mlir:1 offset :14:3: note: see current operation: "llvm.return"(%167) : (!llvm.ptr<i8>) -> ()
```

This PR simply uses a copy of the equalOrFailed function in "./src/Conversion/KrnlToLLVM/KrnlEntryPoint.cpp" to follow its way to check the error codes for function calls, but still causes errors. I am fixing them, and your comments or suggestions are very welcome.

PR #2332, opened 2023-06-22
Fix ifThenElse function of LLVMBuilder.

imaihal:
When I fix PR #2267, I found simple ifThenElse didn't work. So I investigated it.

In original code, original Block is split into `ifBlock` and `thenBlock`, and create additional blocks for `elseBlock` and `endBlock`. In this PR, `thenBlock` and `elseBlock` are inserted between `ifBlock` and `endBlock`.

Following figure shows my understanding. if there is "instruction A" in ifBlock, it seems to put unexpected place, but in this PR, the instruction put in expected place. I referred to [llvm-project/mlir/lib/Conversion/MemRefToLLVM/MemRefToLLVM.cpp](https://github.com/llvm/llvm-project/blob/main/mlir/lib/Conversion/MemRefToLLVM/MemRefToLLVM.cpp#L177-L184)

I want to hear your comments. 

image

### DIFF
--- a/src/Dialect/Mlir/DialectBuilder.cpp
+++ b/src/Dialect/Mlir/DialectBuilder.cpp
@@ -1717,14 +1717,14 @@ void LLVMBuilder::ifThenElse(
   // Split the current block into IF, THEN, ELSE and END blocks.
   Block *ifBlock, *thenBlock, *elseBlock, *endBlock;
   ifBlock = b().getInsertionBlock();
-  thenBlock = ifBlock->splitBlock(b().getInsertionPoint());
-  elseBlock = b().createBlock(
-      thenBlock->getParent(), std::next(Region::iterator(thenBlock)));
+  endBlock = ifBlock->splitBlock(b().getInsertionPoint());
+  thenBlock = b().createBlock(
+      ifBlock->getParent(), std::next(Region::iterator(ifBlock)));
   if (elseFn)
-    endBlock = b().createBlock(
-        elseBlock->getParent(), std::next(Region::iterator(elseBlock)));
+    elseBlock = b().createBlock(
+        thenBlock->getParent(), std::next(Region::iterator(thenBlock)));
   else
-    endBlock = elseBlock;
+    elseBlock = endBlock;
 
   // Emit code for the IF block.
   b().setInsertionPointToEnd(ifBlock);


AlexandreEichenberger (approved):
Do you have a case with a lit tests that could verify that we are doing the right thing?

That would ensure that we maintain this good behavior.

Tx

tungld:
It looks like the new code changes the logic of if-then-else.

Is the code we want to generate is:
```c
if (conditionA) {
   instruction A 
}
```

The new code generates:
```c
if (conditionA) {
} else {
}
instructionA
```
I wonder if the new generated code reflects the logic we want to generate.

imaihal:
@tungld Sorry `ifBlock` in original figure was misleading. So, I updated it.

I want to insert if-then-else clause before `instruction A`, but when I investigated the behavior of `LLVMBuilder::ifThenElse` in PR #2267, the `instruction A` appeared in if sentence. So, I'm trying to fix it.
I'm still not sure why existing code is no problem.

AlexandreEichenberger:
@imaihal I believe a small example that would test this logic would be useful, but I am not really sure how to gen the pattern. Can you look for usages of the builder, and see if we can trigger its use for an existing op?

imaihal:
@AlexandreEichenberger Thanks for your comments. The `LLVMBuilder::ifThenElse` is currently used only in EntryPoint related code, but. it seems to work fine there with main branch. So, I added `LLVMBuilder::ifThenElse` to [KrnlStrlenOpLowering](https://github.com/onnx/onnx-mlir/blob/main/src/Conversion/KrnlToLLVM/KrnlStrlen.cpp) for testing as follows. 

- Added LLVMBuilder::ifThenElse in KrnlStrlenOpLowering
```
      ConversionPatternRewriter &rewriter) const override {                                                                                                                                                 
     : 
     :                                                                                                                                                   
    // Strlen call.                                                                                                                                                                                         
    Type retType = IntegerType::get(context, 64);                                                                                                                                                           
    auto funcCall = rewriter.create<func::CallOp>(                                                                                                                                                          
        loc, strlenRef, retType, ArrayRef<Value>({strPtr}));                                                                                                                                                
                                                                                                                                                                                                            
    // ifThenElse Test START                                                                                                                                                                                
    MultiDialectBuilder<LLVMBuilder, MathBuilder> create(rewriter, loc);                                                                                                                                    
    std::string error_message = "Test message";                                                                                                                                                             
    auto int32Ty = IntegerType::get(context, 32);                                                                                                                                                           
    Value zero = create.math.constant(int32Ty, 0);                                                                                                                                                          
    Value one = create.math.constant(int32Ty, 1);                                                                                                                                                           
    create.llvm.ifThenElse(/*cond=*/                                                                                                                                                                        
                               [&](LLVMBuilder &createLLVM) {                                                                                                                                               
        return createLLVM.icmp(LLVM::ICmpPredicate::ne, zero, one);                                                                                                                                         
      }, /*then=*/                                                                                                                                                                                          
      [&](LLVMBuilder &createLLVM) {                                                                                                                                                                        
        MultiDialectBuilder<LLVMBuilder, KrnlBuilder, MathBuilder> create(                                                                                                                                  
            createLLVM);                                                                                                                                                                                    
        // Print test  message.                                                                                                                                                                             
        create.krnl.printf(StringRef(error_message + "\n"));                                                                                                                                                
      });                                                                                                                                                                                                   
    // ifThenElse Test END                                                                                                                                                                                  
                                                                                                                                                                                                            
    rewriter.replaceOp(op, funcCall.getResults()[0]);                                                                                                                                                       
    return success();                                                                                                                                                                                       
  }                                                                                                                                                                                                         
```

In this case, LLVM IR generated by main branch is not correct, but the LLVM IR generated by this PR is correct.
- MLIR for testing (The first function in `krnl_C_library_function_lowering.mlir`)
  - command: onnx-mlir-opt --convert-krnl-to-affine --convert-krnl-to-llvm="use-opaque-pointers=true"  --mlir-print-ir-after-all krnl_C_library_function_lowering.mlir
```
func.func private @test_krnl_strlen1(%arg0: !krnl.string) -> i64  {
  %len = "krnl.strlen"(%arg0) : (!krnl.string) -> i64
  return %len : i64
}
```
- Result (Main branch)
  - Got an error. 
     within split at krnl_C_library_function_lowering.mlir:3 offset :5:10: error: operation with block successors must terminate its parent block
    From generated code, this seems to be  because `llvm.return"(%1) : (i64) -> ()`  in then block. 
   
```
:
    %2 = "llvm.mlir.constant"() <{value = 0 : i32}> : () -> i32
    %3 = "llvm.mlir.constant"() <{value = 1 : i32}> : () -> i32
    %4 = "llvm.icmp"(%2, %3) <{predicate = 1 : i64}> : (i32, i32) -> i1
    "llvm.cond_br"(%4)[^bb1, ^bb2] <{operand_segment_sizes = array<i32: 1, 0, 0>}> : (i1) -> ()
  ^bb1:  // pred: ^bb0
    %5 = "llvm.mlir.addressof"() <{global_name = @"om_Test message\0A"}> : () -> !llvm.ptr<array<13 x i8>>
    %6 = "llvm.bitcast"(%5) : (!llvm.ptr<array<13 x i8>>) -> !llvm.ptr
    "llvm.call"(%6) <{callee = @printf, fastmathFlags = #llvm.fastmath<none>}> : (!llvm.ptr) -> ()
    "llvm.br"()[^bb2] : () -> ()
    "llvm.return"(%1) : (i64) -> ()
  ^bb2:  // 2 preds: ^bb0, ^bb1
:
```
- Result (This PR)
   This is expected LLVM IR.
```
    %2 = llvm.mlir.constant(0 : i32) : i32
    %3 = llvm.mlir.constant(1 : i32) : i32
    %4 = llvm.icmp "ne" %2, %3 : i32
    llvm.cond_br %4, ^bb1, ^bb2
  ^bb1:  // pred: ^bb0
    %5 = llvm.mlir.addressof @"om_Test message\0A" : !llvm.ptr<array<13 x i8>>
    %6 = llvm.bitcast %5 : !llvm.ptr<array<13 x i8>> to !llvm.ptr
    llvm.call @printf(%6) : (!llvm.ptr) -> ()
    llvm.br ^bb2
  ^bb2:  // 2 preds: ^bb0, ^bb1
    llvm.return %1 : i64
```

imaihal:
@AlexandreEichenberger 
> Do you have a case with a lit tests that could verify that we are doing the right thing?

Current usage in EntryPoint looks no problem. So, it is difficult to add lit test in this PR. So, I want to add tests in subsequent PR which newly add LLVMBuilder::ifThenElse.

PR #2334, opened 2023-06-22
Upgrade dependencies to 'green commits' for week of 6-12-2023

christopherbate:
Upgrades dependencies to the green commits as described here: 
https://github.com/llvm/torch-mlir/issues/1178#issuecomment-1587540534

Changes to this code base were to accomodate the following upstream changes:
- `llvm::Optional` and associated headers have been eliminated upstream, downstream projects should use `std::optional`.
- Minor changes to dialect registration and loading patterns exposed some issues where dialects needed to be explicitly declared as dependent (in a pass) or registered with the context.

Also updated the `install-protobuf.sh` script version number to match the protobuf verison installed in the docker files.


hamptonm1:
@christopherbate Clang format failed.... a few files were impacted according to the check.

AlexandreEichenberger:
@christopherbate looks like there is a failure in the windows... they are more strict about warnings, hopefully a small issue :-)

christopherbate:
> @christopherbate looks like there is a failure in the windows... they are more strict about warnings, hopefully a small issue :-)

When I checked last time, it was a job timeout issue. How do we proceed in this case?

AlexandreEichenberger:
@christopherbate aha, that is ok then, restart with a dummy change. I recall the MS folks indicating that when rebuilding an LLVM, it pushes the limit to the max. Since the LLVM version is then cached, the second run is likely to be clean as it then only use the time for the onnx-mlir build and tests

ashay:
Hi @christopherbate, I wonder if the error in the macOS build is either because of a build type mismatch between protobuf+onnx (which are built in Release mode) and onnx-mlir (which is built in Debug mode).  My reason for suspecting this is that protobuf v3.20 seems to have changed the location of the `InternalMetadata` destructor (see https://github.com/protocolbuffers/protobuf/blame/3.20.x/src/google/protobuf/message_lite.cc#L523-L534).  Specifically, based on whether the code is compiled with `NDEBUG` or not, the the destructor is present either in the compiled library or the header file.  Btw, the macOS build mode is (seemingly inadvertently) getting set to Debug, since we don't explicitly specify Release mode (the CI log contains the line: "No build type selected, default to Debug").

In any case, the reason I started poking around is because I am on the rotation to update LLVM this week.  I tried your patch with the most recent green commits and all tests pass at my end.  I am happy to submit a separate PR to bump the LLVM and MHLO commit hashes, but it looks like the new green commits should cleanly apply to your PR as well.

AlexandreEichenberger:
@christopherbate @ashay 

It would be nice if the new LLVM build cleanly on Mac as some of us are starting to rely more on this build as some of us develop on M1 machines for which docker x86 virtual machine are quite a bit slower due to emulation.

christopherbate:
Looks like it's all working @AlexandreEichenberger

philass (approved):
Thanks @christopherbate, we really appreciate it!

LGTM!

AlexandreEichenberger (approved):
LGTM, Big thanks.

PS: don't you love these capitalization changes? At least it was easy :-)

AlexandreEichenberger:
This appears good to go; let me know if you need one of us to squash and merge it

hamptonm1:
@ashay Since this is outside of the two week window. Can @christopherbate still merge this PR before the next set of updates are applied for llvm and mhlo? Making sure it does not break existing functionality.

christopherbate:
@AlexandreEichenberger can you help merge me merge this I mistakenly thought it was already merged, didn't see the recent comments until yesterday. I keep hitting "update branch", but the Jenkins is slow so by the time it finishes running, I'll need to hit the button again.

PR #2336, opened 2023-06-26
Added canonicalize to the lowering tests

AlexandreEichenberger:
as not having the canonicalization during lowering make the test very susceptible to spurious changes


AlexandreEichenberger:
There seems to be slight issues with dependences, which run differently on Linux/amd64. It may not be worth it.

AlexandreEichenberger:
issues with the order of ops... abort for now

PR #2337, opened 2023-06-27
Preparation for Simd reduction, removal of redundant code

AlexandreEichenberger:
The current reduction code to lower ONNX to Krill used 2 methods, one for legacy and one for current ops. As new versions will be needed to handle SIMD code, the first step is to unify these two methods into a single code generation scheme.

The code also use old methods to generate Krill iterate loops; upgraded to use the builder based on lambda function.

Code generated is slightly more compact, functionally equivalent.

Once this is in, I can start the SIMD code gen.


tungld (approved):
LGTM!

Please consider changing the title since there is no SIMD code.

PR #2350, opened 2023-06-29
add --make_raw flag to onnxExternalizeData.py

sorenlassen:
This makes it possible to turn large constants which are not raw into external data.

I tested this on arcfaceresnet100-8.onnx (size 249MB) from the model zoo.

Without --make_raw it produced no external data, because none its constants are raw.

With --make_raw almost all the 249MB went into a flat .ext file and the .onnx protobuf file was reduced 1345x to 189KB.


AlexandreEichenberger (approved):
LGTM, could add an example of how it is used in the top comments.

sorenlassen:
> could add an example of how it is used in the top comments

ok, done

PR #2351, opened 2023-06-29
Lowering ONNXCustomOp

chentong319:
ONNXCustomOp is lowered to krnl.Call when onnx is converted to krnl.
In the conversion:
1. Allocate MemRef for the output, using the shape inference of the custom op.
2. Both outputs and inputs of ONNXCustomOp will become the parameters of krnl.call
3. Attributes: pass the function_name of custom Op to the functionName of krnl.call. Filter out the attributes for shape inference. Propagate other attributes.


src/Conversion/ONNXToKrnl/Additional/Custom.cpp
```diff
@@ -0,0 +1,82 @@
...
+      Value alloc = create.mem.alignedAlloc(
+          outputMemRefType, shapeHelper.getOutputDims(idx));
+      outputAllocs.emplace_back(alloc);
+    }
+
+    // Handle the attributes: exclude the attributes used for analysis
+    // function_name is passed explicitly. Others may include shape inference
+    std::vector<std::string> excludeStrings = {"function_name",
+        "shape_infer_pattern", "inputs_for_infer", "output_element_type"};
+    std::vector<std::string> attributeNames;
```
tungld, line 61: Please consider using `SmallVector` instead of `std::vector` for a small number of elements.

tungld (approved):
LGTM!

PR #2352, opened 2023-07-01
parallel scrub

sorenlassen:
ScrubDisposablePass can be slow if there are large constants to rewrite from DisposableElementsAttr to DenseElementsAttr. This change parallelizes the work with each worker rewriting a subset of the constants. This speeds things up provided that the work is not all in a single large constant.

Tested it on arcfaceresnet100-8.onnx from the model zoo on a MacBook M1
```
RelWithDebInfo/bin/onnx-mlir arcfaceresnet100-8.onnx -mlir-timing --EmitMLIR
```
and saw the time spent in ScrubDisposablePass drop by 45% from 85ms to 47ms.

Also fixed MLIRContext creation in onnx-mlir.cpp to make the `--mlir-disable-threading` flag work and added debug logging, so you can see a debug log message confirming that threading is disabled if you call
```
onnx-mlir --mlir-disable-threading -debug-only=onnx_mlir_main
```
With this fix it's easy to see the effect of parallel scrub or not by calling the above arcfaceresnet100-8.onnx example with and without `--mlir-disable-threading`

### DIFF
--- a/src/Dialect/ONNX/ElementsAttr/DisposablePool.cpp
+++ b/src/Dialect/ONNX/ElementsAttr/DisposablePool.cpp
@@ -10,6 +10,7 @@
 
 #include "src/Dialect/ONNX/ElementsAttr/DisposablePool.hpp"
 
+#include "mlir/IR/Threading.h"
 #include "llvm/ADT/DenseMap.h"
 
 #include <atomic>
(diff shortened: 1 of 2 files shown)


philass (approved):
LGTM!

It would be interesting to see if some of the even bigger onnx models have even more extreme speedups

PR #2354, opened 2023-07-03
F16 fix for pow canonicalization pattern

philass:


src/Dialect/ONNX/ONNXOps/OpHelper.cpp
```diff
@@ -554,6 +554,9 @@ RESULT_TYPE getScalarValue(ElementsAttr denseAttr, Type type) {
       elementaryType.isInteger(64)) {
     auto valueIt = denseAttr.getValues<IntegerAttr>().begin();
     return (RESULT_TYPE)(*valueIt).cast<IntegerAttr>().getInt();
+  } else if (elementaryType.isF16()) {
+    auto valueIt = denseAttr.getValues<APFloat>().begin();
+    return (RESULT_TYPE)(*valueIt).convertToFloat();
```
sorenlassen, line 559: suggestion: collapse the 3 `isF16/32/64` cases into a single `isa<FloatType>`

sorenlassen (approved):

src/Dialect/ONNX/ONNXOps/OpHelper.cpp
philass: Done

Issue #2344, opened 2023-06-28
Datatype not correctly parsed for Const-prop pass


hamptonm1:
When compiling a customer model, we see the following errors:

```
root@7bc1ee526dda:~/onnx-mlir/build# onnx-mlir --O3 --EmitLib --maccel=NNPA model.onnx --onnx-op-stats TXT -v
The ONNX model has 25071 elements in its initializers. This value would be close to and greater than the number of parameters in the model. Because there is no way to exactly count the number of parameters, this value can be used to have a rough idea of the number of parameters in the model.
loc("StatefulPartitionedCall/sequential/dense/MatMul"): error: reduction dimension must be the same
loc("StatefulPartitionedCall/sequential/dense/MatMul"): error: Failed to scan parameters successfully
loc("StatefulPartitionedCall/sequential/dense/MatMul"): error: 'onnx.MatMul' op shape inference failed
loc(fused["StatefulPartitionedCall/sequential/dense/BiasAdd", "StatefulPartitionedCall/sequential/dense/MatMul"]): error: Gemm 2nd dim of A is different than 1st dim of B
loc(fused["StatefulPartitionedCall/sequential/dense/BiasAdd", "StatefulPartitionedCall/sequential/dense/MatMul"]): error: Failed to scan parameters successfully
loc(fused["StatefulPartitionedCall/sequential/dense/BiasAdd", "StatefulPartitionedCall/sequential/dense/MatMul"]): error: 'onnx.Gemm' op shape inference failed
loc(fused["StatefulPartitionedCall/sequential/dense/BiasAdd", "StatefulPartitionedCall/sequential/dense/MatMul"]): error: Gemm 2nd dim of A is different than 1st dim of B
loc(fused["StatefulPartitionedCall/sequential/dense/BiasAdd", "StatefulPartitionedCall/sequential/dense/MatMul"]): error: Failed to scan parameters successfully
loc(fused["StatefulPartitionedCall/sequential/dense/BiasAdd", "StatefulPartitionedCall/sequential/dense/MatMul"]): error: 'onnx.Gemm' op shape inference failed
loc(fused["StatefulPartitionedCall/sequential/dense/BiasAdd", "StatefulPartitionedCall/sequential/dense/MatMul"]): error: Gemm 2nd dim of A is different than 1st dim of B
[...]

PR #2355, opened 2023-07-04
limit onnx constant value attribute to dense or disposable ElementsAttr

sorenlassen:
changed all non-test code paths to create onnx constants using OnnxBuilder::constant()

added assert in OnnxBuilder::constant() and isDenseONNXConstant() to enforce the invariant that the value attribute is always a DenseElementsAttr or DisposableElementsAttr, since the way we use the value attribute of onnx constants throughout the code base generally assumes that it behaves like those, namely that getValues<T>() works for all data types T and that you can use ElementsAttrBuilder for constant propagation and constant folding

this PR could also help track down how a DenseResourceElementsAttr ends up in a constant operand to onnx.Reshape in issue #2344


tungld (approved):
LGTM!

Thanks for updating some old code also!

PR #2356, opened 2023-07-04
miscellaneous docs fixes

sorenlassen:
Run quick check-docs or check-doc-example before slow numerical tests on all platforms.

Made OMRuntimeTest print and check the result tensor, like in OMRuntimeCppTest, and remove duplicate cruntime linkage.

Regenerated SupportedONNXOps-cpu.md. Changed documentOps.py and gen_onnx_mlir.py to use python3, otherwise 'ninja onnx_mlir_supported_ops' failed for me because I only have python3 and not python.


sorenlassen:

philass (approved):
LGTM

PR #2359, opened 2023-07-06
[NNPA] Clip very small f32 values into the range of dlfloat to avoid out-of-range

### LINKED ISSUE #2312: bertsquad-12.onnx using CPU instead of NNPA for Softmax
Using the latest commit 8185fc83b96eba25a905aa8e1f7e2aeffa0c9e0c [bertsquad-12.onnx ](https://github.com/onnx/models/blob/main/text/machine_comprehension/bert-squad/model/bertsquad-12.onnx) now takes 202 ms to run. PR #2250 reduced this back down from 2000 ms but the new results is still slower than the 160 ms of the 0.4.0 release.

At a glance, I see all the Softmax Ops are going to the CPU instead of NNPA now.

To reproduce
`--O3 --EmitLib --mtriple=s390x-ibm-loz --mcpu=z16 --maccel=NNPA --onnx-op-stats TXT`

## current 8185fc83b96eba25a905aa8e1f7e2aeffa0c9e0c
```
Operations encountered:
-----------------------
   func.func               , 1
   func.return             , 1
   onnx.Add                , 62
   onnx.Cast               , 31
   onnx.Concat             , 44
   onnx.Constant           , 77
   onnx.ConstantOfShape    , 1
   onnx.Dim                , 29
   onnx.Gather             , 1
   onnx.Mul                , 149
   onnx.NoValue            , 1
   onnx.OneHot             , 1
   onnx.Reciprocal         , 25
   onnx.ReduceMeanV13      , 50
   onnx.Reshape            , 142
   onnx.Softmax            , 12
   onnx.Split              , 1
   onnx.Sqrt               , 25
   onnx.Squeeze            , 4
   onnx.Sub                , 62
   onnx.Transpose          , 49
   onnx.Unsqueeze          , 13
  zhigh.Add                , 50
  zhigh.MatMul             , 98
  zhigh.Mul                , 86
  zhigh.Stick              , 238
  zhigh.StickifiedConstant , 147
  zhigh.Tanh               , 12
  zhigh.Unstick            , 198
```

## 0.4.0 release
```
Operations encountered:
-----------------------
   func.func               , 1
   func.return             , 1
   onnx.Add                , 62
   onnx.Cast               , 18
   onnx.Concat             , 56
   onnx.Constant           , 77
   onnx.ConstantOfShape    , 1
   onnx.Dim                , 41
   onnx.Gather             , 1
   onnx.Mul                , 149
   onnx.NoValue            , 1
   onnx.OneHot             , 1
   onnx.Reciprocal         , 25
   onnx.ReduceMeanV13      , 50
   onnx.Reshape            , 166
   onnx.Split              , 1
   onnx.Sqrt               , 25
   onnx.Squeeze            , 4
   onnx.Sub                , 62
   onnx.Transpose          , 49
   onnx.Unsqueeze          , 13
  zhigh.Add                , 50
  zhigh.MatMul             , 98
  zhigh.Mul                , 86
  zhigh.Softmax            , 12
  zhigh.Stick              , 250
  zhigh.StickifiedConstant , 147
  zhigh.Tanh               , 12
  zhigh.Unstick            , 210
```

There's also some minor changes in the output values between them but I expect that's because more functions are going to CPU over the 16 bit NNPA.

tungld:
This patch inserts `ONNXMax` before `ZHighStick` in order to clip very small out-of-range into the dlfloat range, so that NNPA operates on valid values.

It's a bit smart in the sense that there is not need to clip if the input of `ZHighStick` transitively comes from a zTensor. It's because values in a zTensor must be in the dlfloat range already. 

Added an option `nnpa-clip-to-dlfloat-range` to turn on/off this feature. It's ON by default.

Re-enabled softmax on NNPA.

Resolves #2312


src/Accelerators/NNPA/Transform/ZHigh/ZHighClipToDLFloat.cpp
```diff
@@ -0,0 +1,134 @@
...
+/// modification.
+bool valueFromZTensor(Value tensor) {
+  if (!tensor.dyn_cast<BlockArgument>()) {
+    Operation *op = tensor.getDefiningOp();
+    // From a zTensor.
+    if (isa<ZHighUnstickOp>(op))
+      return true;
+    // Concat/Transpose does not change the input precision. So we can consider
+    // that the input is already in the dlfloat range if it comes from zTensor.
+    if (isa<ONNXConcatOp, ONNXTransposeOp>(op)) {
```
AlexandreEichenberger, line 49: There are many more that falls in that category: expand/flatten/gather/reshape/squeeze/unsqueeze/pad.
Then there are also ops that are out of bound friendly: Max (of any kinds), like ReduceMax, Relu... If their input where high enough, their output is guaranteed to be high enough too.

src/Accelerators/NNPA/Transform/ZHigh/ZHighClipToDLFloat.cpp
```diff
@@ -0,0 +1,134 @@
...
+public:
+  using OpRewritePattern<ZHighStickOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(
+      ZHighStickOp stickOp, PatternRewriter &rewriter) const override {
+    Operation *genericOp = stickOp.getOperation();
+    Location loc = genericOp->getLoc();
+
+    Value input = stickOp.getIn();
+    Type inputElementType = getElementType(input.getType());
```
AlexandreEichenberger, line 67: Is there a need to explicitly test if the value is a float, and if not, not add the max?

AlexandreEichenberger (commented):
What is the performance hit of this?
I can see how doing this to constants is free, what about the stick in the critical path?
Would it not make more sense to have a clip and stickily unified operation? I can see how doing this here is useful for quick testing of whether it is useful or not; but if this really become necessary, I believe that integrating it with zDNN stick would be much better.

src/Dialect/ONNX/DialectBuilder.cpp
```diff
@@ -134,6 +134,19 @@ Value OnnxBuilder::matmul(Type Y, Value A, Value B, bool useGemm) const {
...
+  assert(inputs.size() >= 2 && "Expect at least two inputs");
+  Type elementType = getElementType(inputs[0].getType());
+  assert(llvm::all_of(inputs, [elementType](Value v) {
+    return (v.getType().cast<ShapedType>().getElementType() == elementType);
+  }) && "All inputs must have the same element type");
+  Type outputType = inputs[0].getType();
+  for (uint64_t i = 1; i < inputs.size(); ++i)
+    outputType = mlir::OpTrait::util::getBroadcastedType(
+        toTensor(outputType), inputs[i].getType());
+  return createTypedOpAndInferShapes<ONNXMaxOp>(toTensor(outputType), inputs);
```
sorenlassen, line 147: given that createTypedOpAndInferShapes calls inferShapes, will it work to just pass an unranked tensor output type and skip the calls to getBroadcastedType?

tungld:
> Would it not make more sense to have a clip and stickily unified operation? 

zDNN plans to do so, but it may take a bit of time, so in this PR, I created a new pass for this clip and added an option to turn on/off it. When zDNN supports this, we can easily turn it off or remove it.

src/Dialect/ONNX/DialectBuilder.cpp
tungld: Good catch. Will check this.

src/Accelerators/NNPA/Transform/ZHigh/ZHighClipToDLFloat.cpp
tungld: The input of stick is float itself. However, we can have a check or an assert here. I'll update this.

tungld:
> What is the performance hit of this?
> I can see how doing this to constants is free, what about the stick in the critical path?

Sure, I have to measure the performance difference and will let you know. ONNXMax's second operand is a scalar, I hope it has a good performance thanks to SIMD.

src/Accelerators/NNPA/Transform/ZHigh/ZHighClipToDLFloat.cpp
tungld: Added a check for float, only clip if bitwidth is > 16. Thanks!

tungld:
@AlexandreEichenberger @sorenlassen I addressed your comments. Thanks!

AlexandreEichenberger (approved):
LGTM, so if I understood you correctly, this is enabled only for the inputs to softmax? What about log, would that help too?

tungld:
> if I understood you correctly, this is enabled only for the inputs to softmax? What about log, would that help too?

Yes, only for the inputs to softmax. It does nothing to log where we need to check zero values.

PR #2360, opened 2023-07-06
run unit tests in CI

sorenlassen:
created check-unittest target to run all test/unit tests

also moved CustomFnTest from test/onnx2mlir to test/unit

renamed add_numerical_unittest to add_numerical_test to avoid confusion with the new add_unittest

added check-unittest to all CI pipelines, so we don't overlook when a change breaks the unit tests

the runtime of check-unittest is less than 7 seconds on my laptop


AlexandreEichenberger (approved):
LGTM

PR #2361, opened 2023-07-07
Constant propagation for ONNXMin, ONNXMax and ONNXSum

tungld:
This patch adds a template for constant propagation for variadic element-wise operations, and applies the template to ONNXMin, ONNXMax and ONNXSum.


src/Transform/ONNX/ConstProp.cpp
```diff
@@ -238,6 +238,33 @@ Value ConstPropElementwiseBinary(PatternRewriter &rewriter,
   return createReplacingConstantOp(rewriter, replacingValue, resultElements);
 }
 
+/// Do element-wise binary calculation of a variadic value and create an
+/// ONNXConstantOp for the result.
+template <typename ElementwiseBinaryOp>
+Value ConstPropVariadicElementwiseBinary(
+    PatternRewriter &rewriter, Value replacingValue, ValueRange inputList) {
+  assert(inputList.size() > 0 && "The variadic input is emply");
```
philass, line 246: "emply" -> "empty"

philass (approved):
LGTM!

It seems that you can also get const prop for ONNXSumOp using the same boiler plate https://github.com/onnx/onnx/blob/main/docs/Changelog.md#Sum-13.

src/Transform/ONNX/ConstProp.cpp
tungld: Thanks! I fixed it.

tungld:
> It seems that you can also get const prop for ONNXSumOp using the same boiler plate https://github.com/onnx/onnx/blob/main/docs/Changelog.md#Sum-13.

Yes, I added ONNXSumOp.

sorenlassen (approved):
LGTM

PR #2363, opened 2023-07-11
Fix searching entry point names on z/OS

tungld:
Fix an error in searching entry point names on z/OS when supporting `--store-constants-to-file`


cjvolzka (commented):
+1 from me. I ran the code and it addressed the initial issue. However I'll let someone else do the final approval since I'm not too familiar with the conversion code yet.

tungld:
>Technically that's a separate issue but it still blocks the end goal (getting the model to compile on z/OS). Let me know if you want me to create a different Issue here or if we want to cover it under this PR.

Yes, we can cover it under this PR, a single PR to fix the remaining issues on z/OS if possible.

AlexandreEichenberger (approved):
LGTM

tungld:
I think we have fixes at compiler level, this is from @mikeessen :
```
Hi Tung,
 
Here are the updates made so far to OMExternalConstant.inc for z/OS with gpt2.
 
The changes in the diff below are for:
1. Converting the input File name to omMMapBinaryFile to EBCDIC.
2. Invoking mmap() with __MAP_MEGA.
3. Using cds instead of __sync_bool_compare_and_swap since it not available.
 
Best regards,
Mike
 ```
I will pull Mike's changes to this patch.

Still need a change at z/OS level and Mike has successfully ran a big model on a z/OS image without NNPA. Mike is waiting for a change in another z/OS machine for checking NNPA.

tungld:
Integrated Mike's changes to this patch.

cjvolzka:
We're going to give this a smoke test and see what happens. Assuming it gets us closer, would there be any objection to merging this even if it doesn't fully get the large models compiling on z/OS? 

I know previously we said:
> Yes, we can cover it under this PR, a single PR to fix the remaining issues on z/OS if possible.

But whats here should fix a few issues toward that goal and the PR has been open for a while.

tungld:
> Assuming it gets us closer, would there be any objection to merging this even if it doesn't fully get the large models compiling on z/OS?

I don't have and we can merge this.

cjvolzka:
I confirmed with this change I can still compile the onnx-gpt2-xl-causal-lm model and run it successfully on Linux on Z. It also resolves the originally reported issues on z/OS. However we are encountering new issues after this change.

Since this resolves the original issue and has been open for a while, I'll merge this PR and create a new issue for what I'm encountering now.

Sunny-Anand:
@cjvolzka @mikeessen what was the original reported issue for this PR? What were the new issues and was there a follow-up PR for those issues @tungld?

tungld:
@Sunny-Anand this would be the latest PR related to external constant files in z/OS.

mikeessen:
@Sunny-Anand Here's the issue in Jira: https://jsw.ibm.com/browse/ZAIDLC-930

PR #2366, opened 2023-07-12
Bump to llvm commit 8b76aea8d8b1b71f6220bc2845abc749f18a19b7

flemairen6:
Green LLVM commit: 8b76aea8d8b1b71f6220bc2845abc749f18a19b7
Green MHLO commit: cae5349f85b3b2f58d80ca788e9a8695d896c3a8


maxbartel:

maxbartel:

ttjost:
@AlexandreEichenberger I am seeing an issue with the `powerpc` pipeline that relates to a recent change in `llvm-exegesis`. Therefore, we don't even build LLVM for it. =(
LLVM buildbots for `ppc` pass the test.  Do you have any idea how to solve this?
The error says that `__builtin_thread_pointer' is not supported on this target`, but I don't have a way to reproduce this.
Thank you!

philass:
Ahh that looks really frustrating.

I wonder if we can play with CMake flags to not build with `llvm-exegesis`

ttjost:
> Ahh that looks really frustrating.
> 
> I wonder if we can play with CMake flags to not build with `llvm-exegesis`

I tried doing that. This led to disabling `llvm-mca`, `obj2yaml`, and `yaml2obj`, but `lit tests` all seem to depend on them, and I get a bunch of CMake dependency errors between these tools and tests.
 
In any case, it turns out `__builtin_thread_pointer` is not supported by `gcc` (https://godbolt.org/z/chEnh7Ton), but works with `clang` (https://godbolt.org/z/c9bdTqv5f), that is why LLVM buildbots for PPC work and onnx-mlir pipelines don't.

I am somewhat blocked now, since this only happens on PPC + GCC (it seems), and everything else is green.
Should we try building an in-tree `clang` for it?
Could you anyone check this, please?

sorenlassen:
maybe the problem has to do with this llvm code change:
https://github.com/llvm/llvm-project/commit/46b5b855489f87aac2487d099175807880486041

but I can't think of any workaround

AlexandreEichenberger:
@flemairen6 @ttjost What about the s390 error, is it unrelated?

sorenlassen:
@jenkins-droid test this please

sorenlassen:
@ttjost please sign-off on the last commit to fix the DCO failure

I'm not 100% sure why the jenkins pipelines fail with `ERROR: Exception: java.lang.Exception: admin authentication failed`, maybe it's blocked by the DCO failure?

sorenlassen:
@jenkins-droid test this please

sorenlassen:
@AlexandreEichenberger can you tell why the jenkins pipelines are failing with `ERROR: Exception: java.lang.Exception: admin authentication failed`?

gongsu832:
```
/workdir/llvm-project/llvm/tools/llvm-exegesis/lib/BenchmarkRunner.cpp: In member function 'llvm::Error llvm::exegesis::{anonymous}::SubProcessFunctionExecutorImpl::createSubProcessAndRunBenchmark(llvm::StringRef, llvm::SmallVectorImpl<long int>&) const':
/workdir/llvm-project/llvm/tools/llvm-exegesis/lib/BenchmarkRunner.cpp:336:61: error: '__builtin_thread_pointer' is not supported on this target
  336 |         syscall(SYS_rseq, (intptr_t)__builtin_thread_pointer() + __rseq_offset,
      |                                     ~~~~~~~~~~~~~~~~~~~~~~~~^~
```
Apparently ppc64le does not support `__builtin_thread_pointer`

maxbartel (approved):
LGTM, the clang PR is also already up https://github.com/onnx/onnx-mlir/pull/2379

maxbartel:
@AlexandreEichenberger @philass I will merge this if the pipelines are green, the clang fix landed

PR #2368, opened 2023-07-12
demo onnx f32->f8 conversion diff between spec and ref impl

sorenlassen:
Added TestFPConversion.py test showing that for the "UZ" f8 types the onnx reference implementation of conversion from f32 +/-INF differs from the spec: reference implementation converts +/-INF to +/-MAX whereas the spec says they should convert to NaN.

See issue #2369.


sorenlassen:
closing this PR because I don't plan to merge it

PR #2373, opened 2023-07-12
Fix the compile and run execution session in Python

AlexandreEichenberger:
The error was due to the fact that the constructor of `PyOMCompileExecutionSession` used the constructor of `OMExecutionSession` which required a compiled .so. Have now added a method to first compile and then initialize the .so in the super class OMExecutionSession after its constructor is called

Cleaned up also a bit the code


src/Runtime/PyOMCompileExecutionSession.cpp
```diff
@@ -23,10 +23,14 @@ SUPPRESS_WARNINGS_POP
 
 namespace onnx_mlir {
 
+// =============================================================================
+// Constructor
+
 PyOMCompileExecutionSession::PyOMCompileExecutionSession(
     std::string inputFileName, std::string sharedLibPath, std::string flags,
```
tungld, line 30: Do we always compile a model? Does it make sense to skip compilation when `sharedLibPath` exists? Users may want to use the same PyOMCompileExecutionSession and expect the compilation happens only once.

If the .so file is not reused, we perhaps remove it at the end of the session.

AlexandreEichenberger: I see what you are saying, but then we have to look into what the compile options were... 
2 solutions:
* add a flag to say "reuse if here is already a so"
* no changes:  compile the first time, and then we can use the same .so with a regular ExecutionSession.

tungld: Option 1 looks flexible, we can enable "reuse" by default, and for debugging, not "reuse" is useful.

AlexandreEichenberger: ok, will do

AlexandreEichenberger:
As requested, the extended `OMCompileExecutionSession` interface now first check if the file that would be created by the compiler already exists, and if it does not, compile the model using the flags passed when building the python object.

Technically, the code parse the flags to detect possible `-o filename` or `-o=filename` options to determine if the user is requesting a custom output name, and check also the flags to determine if a custom target is requested.

An optional parameter, `reuse_compiled_model=0` can be passed to disable reuse; by default compiled models are reused.

AlexandreEichenberger:
@tungld , does this work for you now?

src/Runtime/PyOMCompileExecutionSession.cpp
```diff
@@ -23,40 +23,60 @@ SUPPRESS_WARNINGS_POP
...
+    if (rc != 0) {
+      // Compilation failure: save error message.
+      errorMessage = std::string(errorMsg);
+      // Empty output file name.
+      this->outputFileName = std::string();
+      throw std::runtime_error(reportCompilerError(errorMessage));
+    }
+  }
+  // Compilation success: save output file name.
+  this->outputFileName = std::string(outputName);
```
tungld, line 63: According to your comment for `omCompileOutputFileName`, do we need to free `outputName` here?

tungld (approved):
LGTM.

Please remove `third_party/benchmark` update since I guest it's not your intention in this patch.

PR #2379, opened 2023-07-18
Build llvm with clang since gcc on ppc64le doesn't support __builtin_thread_pointer

gongsu832:


AlexandreEichenberger (approved):
LGTM, please just confirm that we now build LLVM using clang on all CIs using docker, namely s390, power, and linux x86.

AlexandreEichenberger:
@gongsu832 thanks for the quick turnaround!

gongsu832:
LLVM on all archs are built with clang. onnx-mlir must also be built with clang. Otherwise, a lot of lit tests crash.

PR #2380, opened 2023-07-18
Simd for horizontal reductions

AlexandreEichenberger:
This PR create a horizontal reduction in simd mode when all of the reductions are on the innermost dimensions of the input arrays, and are a multiple of the SIMD vector length.

This PR also enables SIMD for ARM64 / AARCH64 which are used for the Mac M1/M2. Use -march=arm64 or aarch64 to enable SIMD code gen for that platform too.

Various different code cleanup where also performed, notably pulling code that was used for some lowering into builders.


test/mlir/onnx/onnx_lowering_with_canonicalize_O3.mlir
```diff
@@ -6640,3 +6793,339 @@ func.func @test_custom_dynamic1(%arg0: tensor<1024xi32>, %arg1: tensor<?x4xf32>)
 // CHECK:           return [[RES_]] : memref<?x4xf32>
 // CHECK:         }
 
+// -----
+
+// Original gpt2 reduction
```
tungld, line 6798: Thanks for adding these GPT2 tests!

src/Dialect/Mlir/DialectBuilder.cpp
```diff
@@ -1481,13 +1505,74 @@ Value VectorBuilder::mergeLow(Value lhs, Value rhs, int64_t step) const {
...
+    return b().create<vector::ReductionOp>(
+        loc(), vector::CombiningKind::MUL, value);
+  }
+  case CombiningKind::MAX: {
+    if (MathBuilder::isUnsignedIntegerWithVector(type))
+      return b().create<vector::ReductionOp>(
+          loc(), vector::CombiningKind::MAXUI, value);
+    if (MathBuilder::isIntegerWithVector(type))
+      return b().create<vector::ReductionOp>(
+          loc(), vector::CombiningKind::MAXSI, value);
```
tungld, line 1526: So if the input type is `tensor<10xi64>` we will use signed integers? Could you check with a simple example of reducemax to see if it really works or not? Just wanted to confirm.

tungld (approved):
LGTM.

Just wanted to confirm that it works for the integer type.

src/Dialect/Mlir/DialectBuilder.cpp
AlexandreEichenberger: added an int test, verified with the new `CheckONNXModel.py -m reducemeanint.mlir -t "-O3 -march=arm64"    ` command... really useful

PR #2381, opened 2023-07-19
Support multiple compiled models

tungld:
Issue: .so libraries generated by `onnx-mlir` contains the same entry point `run_main_graph`, so there are conflicts in entry points if multiple compiled .so libraries are loaded in the same program. Actually, all symbols (e.g. global variables, run_main_graph, omQueryEntryPoints, etc. ) in a compiled .so are potentially conflict.

To avoid the conflicts, this patch appends a tag into symbols to make the symbols unique across multiple .so files.
- For global variables, their names will be postfixed by the tag, e.g. `constant_0` becomes `constant_0_tag`.
- For functions (e.g. run_main_graph, omQueryEntryPoints), we generate two functions, e.g. `run_main_graph` and `run_main_graph_tag`, in which `run_main_graph` is a wrapper of `run_main_graph_tag` with the same signature and will calls `run_main_graph_tag`. Users can use one of them and get the same result. In case of loading multiple .so files, users must use the function with tags, e.g. `run_main_graph_tag` to avoid conflicts.

By default, the tag is the same as the .so filename without extension. For example, if the compiler generates `model.so`, then the tag is `model` and the tagged entry point is `run_main_graph_model`.

Users can set a tag when compiling a model by using `--tag=mytag`.

This patch also updates ExecutionSession and PyExecutionSession to support tags, so that we can run multiple models in the same python script. For example:
```python
import numpy as np
from PyRuntime import OMExecutionSession

encoder_model = 'encoder/model.so' # Assumed that the model was compiled using `--tag=encoder`
decoder_model = 'decoder/model.so' # Assumed that the model was compiled using `--tag=decoder`

# Create a session for the encoder model.
encoder_sess = OMExecutionSession(shared_lib_path=encoder_model, tag="encoder")
# Create a session for the decoder model.
decoder_sess = OMExecutionSession(shared_lib_path=decoder_model, tag="decoder")

```
If there is no tag, PyExecutionSession will use the .so filename as a tag.
```python
import numpy as np
from PyRuntime import OMExecutionSession

encoder_model = 'my_encoder.so'
decoder_model = 'my_decoder.so'

# Create a session for the encoder model.
encoder_sess = OMExecutionSession(shared_lib_path=encoder_model) # tag will be `my_encoder` by default.
# Create a session for the decoder model.
decoder_sess = OMExecutionSession(shared_lib_path=decoder_model) # tag will be `my_decoder` by default.
```

There is a special value `--tag=NONE` that totally disables all tag-related stuffs, which works as if this patch is not applied.


src/Runtime/PyExecutionSessionBase.cpp
```diff
@@ -55,8 +55,8 @@ struct npy_format_descriptor<onnx_mlir::float_16> {
 namespace onnx_mlir {
 
 PyExecutionSessionBase::PyExecutionSessionBase(
-    std::string sharedLibPath, bool defaultEntryPoint)
-    : onnx_mlir::ExecutionSession(sharedLibPath, defaultEntryPoint) {}
+    std::string sharedLibPath, std::string tag, bool defaultEntryPoint)
+    : onnx_mlir::ExecutionSession(sharedLibPath, tag, defaultEntryPoint) {}
```
AlexandreEichenberger, line 59: This code will change significantly in #2373, my recommendation will be to find out the tags from the compile option strings.

src/Runtime/PyExecutionSession.hpp
```diff
@@ -21,20 +21,24 @@ namespace onnx_mlir {
...
+      bool defaultEntryPoint = true);
 };
 } // namespace onnx_mlir
 
 PYBIND11_MODULE(PyRuntime, m) {
   py::class_<onnx_mlir::PyExecutionSession>(m, "OMExecutionSession")
       .def(py::init<const std::string &>(), py::arg("shared_lib_path"))
-      .def(py::init<const std::string &, const bool>(),
-          py::arg("shared_lib_path"), py::arg("use_default_entry_point"))
+      .def(py::init<const std::string &, const std::string &>(),
```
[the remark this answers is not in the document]
tungld: Great! I will use it.

src/Runtime/PyExecutionSessionBase.cpp
tungld: Sure, will update this to get from the compiler options once #2373 is landed.

tungld:
@AlexandreEichenberger the patch includes all I have for supporting multiple compiled models except the change to PyCompileAndRunExecusionSession to use tag. I will make the change once #2373 is landed.

tungld:
@AlexandreEichenberger I updated PyCompileAndRun to use a model tag where the tag is obtained from the compile option.  The patch is ready now.

src/Runtime/PyExecutionSessionBase.cpp
tungld: Done.

docs/UsingPyRuntime.md
```diff
@@ -92,16 +92,53 @@ for entry_point in entry_points:
...
 However, using the constructor and run method is enough to perform inferences.
 
 ```python
-def __init__(self, shared_lib_path: str, use_default_entry_point: bool):
+def __init__(self, shared_lib_path: str, tag: str, use_default_entry_point: bool):
     """
     Args:
         shared_lib_path: relative or absolute path to your .so model.
-        use_default_entry_point: use the default entry point that is `run_main_graph` or not. Set to True by default.
+        tag: a string that was passed to `--tag` when compiling the .so model. By default, it is `filename` in `filename.so`
```
AlexandreEichenberger, line 140: By default, it is the output file name without its extension, namely `filename` in `filename.so`

docs/doc_example/main.cpp
```diff
@@ -21,7 +21,7 @@ int main(int argc, char *argv[]) {
   char *errorMessage = nullptr;
   char *compiledFilename = nullptr;
   std::string flags = readArgs(argc, argv);
-  flags += "-o add-cpp-interface";
+  flags += "-o add-cpp-interface --tag=NONE";
```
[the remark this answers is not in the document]
tungld: The thing is for a string option passing to `onnx-mlir`, there is no way to distinguish the absence of the option or an empty string. We get an empty string for `tag` in both calls `onnx-mlir` and `onnx-mlir --tag=""`. Other options `maccel` and `instrument-ops` also use `NONE` to know disable the options.
 
> In any case, would it be easier here to give it an actual tag value?

By default, the tag value is the model's filename. If it requires to give an actual tag value, do users always have to pass the filename?

src/Runtime/PyOMCompileExecutionSession.hpp
```diff
@@ -47,7 +47,7 @@ PYBIND11_MODULE(PyCompileAndRuntime, m) {
       m, "OMCompileExecutionSession")
       .def(py::init<const std::string &, const std::string &, const bool,
                const bool>(),
-          py::arg("input_model_name"), py::arg("flags"),
+          py::arg("input_model_name"), py::arg("flags") = "",
```
tungld, line 50: @AlexandreEichenberger, I set a default value for `flags`. Let me know if it is ok or not. Thanks!

docs/UsingPyRuntime.md
```diff
@@ -92,16 +92,53 @@ for entry_point in entry_points:
...
+decoder_model = 'decoder/model.so' # Assumed that the model was compiled using `--tag=decoder`
+
+# Create a session for the encoder model.
+encoder_sess = OMExecutionSession(shared_lib_path=encoder_model, tag="encoder")
+# Create a session for the decoder model.
+decoder_sess = OMExecutionSession(shared_lib_path=decoder_model, tag="decoder")
+```
+
+In case two models were NOT compiled by using `--tag`, they must be compiled
+with different .so filenames so that they can be used in the same python script.
```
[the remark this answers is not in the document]
tungld: Done. Thanks!

docs/UsingPyRuntime.md
tungld: Done. Thanks!

docs/doc_example/main.cpp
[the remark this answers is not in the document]
AlexandreEichenberger: Thanks for the explanations...  So this example is trying to show that if the user really does not want the file extension to be added, user must use the NONE... If it's something that you want to be exposed to the user, then it should be explained in the options. 

So do you think there are users that don't want any suffixes?

AlexandreEichenberger (approved):
LGTM, suggested a few additional documentation/explanation requests which I believe would help

docs/doc_example/main.cpp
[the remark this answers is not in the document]
tungld: I removed the NONE tag here.

src/Compiler/CompilerOptions.cpp
```diff
@@ -284,6 +284,22 @@ llvm::cl::list<std::string> functionsToDecompose("functions-to-decompose",
...
+        "LLVMIR to make the symbols unique across multiple generated models. "
+        "By default, use the filename (without extension) of the input onnx "
+        "model or the value passed to `-o`. The tag will be appended to "
+        "global variable and function names. For backward compability, each "
+        "function has two versions with the same signature and doing the same "
+        "computation. For example, we will have two entry points: "
+        "`run_main_graph` and `run_main_graph_tag`, where `run_main_graph` "
+        "is just a wrapper of `run_main_graph_tag`. Users can call one of "
+        "the entry points and expect the same result. Passing `NONE` to "
+        "`--tag` will disable tag completely."),
```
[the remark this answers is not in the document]
tungld: I added one sentence to make `--tag=NONE` more clear.

src/Runtime/PyOMCompileExecutionSession.hpp
tungld: I restored this change. So I used an empty string for flags in the python script for `make check-multiple-models` instead of changing it here.

Issue #2375, opened 2023-07-13
IndexExprBuilder fails shape assertion for tiny-yolov3-11.onnx


nicholaiTukanov:
Hi,

I am trying to experiment with the `onnx-mlir` tools and I have received the following error when trying to emit MLIR for the `tiny-yolov3-11.onnx` (taken from the model zoo):

```bash
root@e2546c03e68f:/workdir/models# onnx-mlir --EmitMLIR ./tiny-yolov3-11.onnx 
onnx-mlir: /workdir/nico-onnx-mlir/src/Dialect/Mlir/IndexExprBuilder.cpp:120: int64_t onnx_mlir::IndexExprBuilder::getArraySize(mlir::Value, bool): Assertion `shape != ShapedType::kDynamic && "expected static size"' failed.
Aborted
```

Is this error expected? What should be done to fix it?

tungld:
@nicholaiTukanov thanks for reporting the issue!

@AlexandreEichenberger this PR caused tiny-yolo failed.
<img width="624" alt="Screenshot 2023-07-14 at 10 13 51" src="https://github.com/onnx/onnx-mlir/assets/1230113/6d66bb0d-a62d-405e-bb2a-539cda217503">

nicholaiTukanov:
@tungld and @AlexandreEichenberger, using defe402aafe3f2b6d714a4a908934d1180797213, I am also getting the following errors from all of the [yolov3 models from the model zoo](https://github.com/onnx/models/tree/main/vision/object_detection_segmentation/yolov3):

```bash
root@3bc258df3a1f:/workdir/nico-onnx-mlir/executor# onnx-mlir yolov3-<op-set>-<type>.onnx 
Onnx Model Parsing Failed on yolov3-<op-set>-<type>.onnx 
```

Is there a list of all supported models from the model zoo?

AlexandreEichenberger:
#2389 is aiming to fix this issue

PR #2389, opened 2023-07-21
Fixing the tiny-yolo regression

AlexandreEichenberger:
Fixed the issue in tiny-yolo. It was due to the fact that, even though `repeat` value  has an unknown dimension, it is defined in the standard as having the same number of inputs as the `input` value has dimensions. 

Using that, and modifying the logic a bit, let us process tiny-yolo again.


AlexandreEichenberger:
fix for #2375

AlexandreEichenberger:
@tungld passes the tests, and running tiny yolo works. Please look into it when you have a chance

tungld (approved):
LGTM. Thanks!

PR #2391, opened 2023-07-22
reuse dialect registration/loading

sorenlassen:
Replaced duplicated dialect regitration and loading with the `registerDialects(maccel)` and `loadDialects(context)` functions.

Removed accelerator support from onx-mlir-reduce because it didn't seem to be working since there is no call to `accel::initAccelerators(maccel)` and there are no command line options to specify `maccel`.

### DIFF
--- a/src/Compiler/CompilerDialects.cpp
+++ b/src/Compiler/CompilerDialects.cpp
@@ -1,0 +1,48 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//===------------------------ CompilerDialects.cpp ------------------------===//
+
+#include "CompilerDialects.hpp"
+
+#include "src/Dialect/Krnl/KrnlOps.hpp"
+#include "src/Dialect/ONNX/ONNXDialect.hpp"
+
+#include "mlir/InitAllDialects.h"
+
+using namespace mlir;
+
+namespace onnx_mlir {
+
+DialectRegistry registerDialects(ArrayRef<accel::Accelerator::Kind> accels) {
+  DialectRegistry registry;
+
+  // Note that we cannot consult command line options because they have not yet
+  // been parsed when registerDialects() is called.
+
+  registry.insert<arith::ArithDialect>();
+  registry.insert<linalg::LinalgDialect>();
+  registry.insert<affine::AffineDialect>();
+  registry.insert<LLVM::LLVMDialect>();
+  registry.insert<scf::SCFDialect>();
+  registry.insert<func::FuncDialect>();
+  registry.insert<vector::VectorDialect>();
+  registry.insert<shape::ShapeDialect>();
+  registry.insert<math::MathDialect>();
+  registry.insert<memref::MemRefDialect>();
+  registry.insert<ONNXDialect>();
+  registry.insert<KrnlDialect>();
+  registry.insert<cf::ControlFlowDialect>();
+
+  // Initialize accelerator(s) if required.
+  accel::initAccelerators(accels);
+
+  // Register dialects for accelerators.
+  for (auto *accel : accel::Accelerator::getAccelerators())
+    accel->registerDialects(registry);
+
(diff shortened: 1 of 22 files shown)


src/Compiler/CompilerDialects.cpp
```diff
@@ -0,0 +1,48 @@
...
+  registry.insert<affine::AffineDialect>();
+  registry.insert<LLVM::LLVMDialect>();
+  registry.insert<scf::SCFDialect>();
+  registry.insert<func::FuncDialect>();
+  registry.insert<vector::VectorDialect>();
+  registry.insert<shape::ShapeDialect>();
+  registry.insert<math::MathDialect>();
+  registry.insert<memref::MemRefDialect>();
+  registry.insert<ONNXDialect>();
+  registry.insert<KrnlDialect>();
```
AlexandreEichenberger, line 35: In the past, we have tried to make it possible for folks to build onnx-mlir without needing Krnl, as it is somewhat specialized for our use. Would it make sense to keep this separation?

src/Compiler/CompilerDialects.cpp
```diff
@@ -0,0 +1,48 @@
...
+  registry.insert<cf::ControlFlowDialect>();
+
+  // Initialize accelerator(s) if required.
+  accel::initAccelerators(accels);
+
+  // Register dialects for accelerators.
```
AlexandreEichenberger, line 41: Thanks for maintaining the separation between normal code and accelerator.

I saw that loading dialects in the NNPA code were removed, so they can just be registered?
@tungld maybe you could check that this works for our usage?

AlexandreEichenberger (commented):
Generally a good move, added a question about separation between krnl and the rest of the ONNX + MLIR dialects.

Also, requested @tungld to just verify that this work well for NNPA.

src/Compiler/CompilerDialects.cpp
sorenlassen: > I saw that loading dialects in the NNPA code were removed, so they can just be registered?

yes, getOrLoadDialects() is no longer needed because we now call registerDialects() and register the dialects with the context and then load them with `ctx->loadAllAvailableDialects()`

src/Compiler/CompilerDialects.cpp
sorenlassen: this PR doesn't change this, right?

I don't think the restructuring in this PR would make it harder to implement the separation if we want to do that in the future

sorenlassen: > build onnx-mlir without needing Krnl

if you want to use onnx-mlir as a library and use the ONNX dialect without Krnl, then I believe that the recommended way (at least what Groq does) is to write your own version of the onnx-mlir/onnx-mlir-opt executables and also your own registerDialects(), registerPasses(), addPasses(), etc; that is, just link things like OMONNXOps (dialect, canonicalization) and OMONNXRewrite (shape inference, const prop, decomposition) and build up things from there

in other words, registerDialects() is only meant to be reusable between onnx-mlir and onnx-mlir-opt

AlexandreEichenberger (approved):
LGTM, thanks for the restructuring

src/Tools/onnx-mlir-opt/onnx-mlir-opt.cpp
```diff
@@ -170,24 +150,23 @@ int main(int argc, char **argv) {
 
   // Set up the input file.
   std::string error_message;
-  auto file = mlir::openInputFile(input_filename, &error_message);
+  auto file = openInputFile(input_filename, &error_message);
   if (!error_message.empty()) {
     llvm::errs() << "Failure to open file; " << error_message << "\n";
-    return failed(LogicalResult::failure());
```
sorenlassen, line 176: fyi, I changed some convoluted return code expressions to 1 and 0

PR #2392, opened 2023-07-22
restructure initOMPasses() into registerPasses()

sorenlassen:
Renamed `initPasses` to `registerPasses` to follow MLIR naming convention and grouped all the related logic into new RegisterPasses.hpp/cpp source files in the Tools/onnx-mlir-opt directory where they are used.

Added comments to explain that passes are registered before command line options are parsed, because this has confused me in the past.

Also made some unrelated cleanups of dependencies in InstrumentPass and Pass/Passes.hpp.


src/Tools/onnx-mlir-opt/RegisterPasses.cpp
```diff
@@ -2,22 +2,36 @@
...
+// the functions below because they are called before command line options are
+// parsed, because passes must be registered first as they instruct command
+// line parsing: registered passes can be expressed as command line flags.
+//
+// In particular the --O flag value is passed as a function argument optLevel
+// so we can avoid reading the OptimizationLevel command-line option here.
+//
 //===----------------------------------------------------------------------===//
 
-#pragma once
```
AlexandreEichenberger, line 13: no need for the once?

src/Tools/onnx-mlir-opt/onnx-mlir-opt.cpp
```diff
@@ -147,20 +148,9 @@ int main(int argc, char **argv) {
...
-  registerSCFPasses();
-  bufferization::registerBufferizationPasses();
-
-  onnx_mlir::initOMPasses(OptimizationLevel);
-  onnx_mlir::initMLIRPasses();
-
-  // Initialize passes for accelerators.
-  for (auto *accel : onnx_mlir::accel::Accelerator::getAccelerators())
-    accel->initPasses(OptimizationLevel);
+  // Registered passes can be expressed as command line flags, so they must
```
AlexandreEichenberger, line 151: that is really good, much more succinct and isolated

AlexandreEichenberger (approved):

PR #2393, opened 2023-07-24
Add header guard for SmallFPConversion

mikeessen:
The '#pragma once' is not supported on z/OS and causes a warning during compile.  This warning shows up when compiling the Runtime code to be used when linking a z/OS shared object model.

This PR changes the  '#pragma once' to a header guard in SmallFPConversion.h.


sorenlassen (approved):
thanks for fixing

PR #2400, opened 2023-07-26
support function attributes with default values

sorenlassen:
Function attributes with default values were introduced in onnx opset 19.


chentong319 (approved):
LGTM!

sorenlassen:
thank you for the review!

PR #2401, opened 2023-07-26
Lowering to KRNL Equal Operation

hamptonm1:
I added the lowering to krnl for Equal Operation (v19) that now supports a string input. I added three lit tests which are to test the input of a string, float and integer. I also updated the backend test to match the available test offered for onnx 1.14.0. 

NOTE: I did disable the `test_equal_string_cpu` and `test_equal_string_broadcast_cpu` tests based on the information below.


hamptonm1:
Okay so thanks @tungld for helping me understand why the backend tests are failing.... for OMTensor.inc we currently do not support input type string and that is why I was getting the error `Numpy type not supported: object`. There is an issue that @negiyas created to also add this support for `CategoryMapper`: https://github.com/onnx/onnx-mlir/issues/2164

hamptonm1:
However, I did manually test that my code works by using the `/onnx-mlir/third_party/onnx/onnx/backend/test/data/node/test_equal_string/model.onnx` . Here are my results below:

```
meganhampton@megans-mbp-2 build % ./Debug/bin/onnx-mlir -EmitLib model.onnx 

meganhampton@megans-mbp-2 build % ./Debug/bin/onnx-mlir-opt -convert-onnx-to-krnl -convert-krnl-to-affine model.onnx.mlir
module attributes {llvm.data_layout = "e-m:o-i64:64-i128:128-n32:64-S128", llvm.target_triple = "arm64-apple-darwin22.5.0", "onnx-mlir.symbol-postfix" = "model"} {
  func.func @main_graph(%arg0: memref<2x!krnl.string>, %arg1: memref<2x!krnl.string>) -> memref<2xi1> attributes {input_names = ["x", "y"], llvm.emit_c_interface, output_names = ["z"]} {
    %c1 = arith.constant 1 : index
    %c2 = arith.constant 2 : index
    %c2_0 = arith.constant 2 : index
    %c1_1 = arith.constant 1 : index
    %alloc = memref.alloc() {alignment = 16 : i64} : memref<2xi1>
    %c0 = arith.constant 0 : index
    %c2_2 = arith.constant 2 : index
    %c2_3 = arith.constant 2 : index
    %c2_4 = arith.constant 2 : index
    %c0_i32 = arith.constant 0 : i32
    affine.for %arg2 = 0 to 2 {
      %0 = affine.load %arg0[%arg2] : memref<2x!krnl.string>
      %1 = affine.load %arg1[%arg2] : memref<2x!krnl.string>
      %2 = "krnl.strlen"(%0) : (!krnl.string) -> i64
      %3 = "krnl.strncmp"(%0, %1, %2) : (!krnl.string, !krnl.string, i64) -> i32
      %4 = arith.cmpi ne, %3, %c0_i32 : i32
      affine.store %4, %alloc[%arg2] : memref<2xi1>
    }
    return %alloc : memref<2xi1>
  }
  "krnl.entry_point"() {func = @main_graph, numInputs = 2 : i32, numOutputs = 1 : i32, signature = "[    { \22type\22 : !krnl.string , \22dims\22 : [2] , \22name\22 : \22x\22 }\0A ,    { \22type\22 : !krnl.string , \22dims\22 : [2] , \22name\22 : \22y\22 }\0A\0A]\00@[   { \22type\22 : \22i1\22 , \22dims\22 : [2] , \22name\22 : \22z\22 }\0A\0A]\00"} : () -> ()
}
```

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1050,11 +1051,27 @@ template <>
...
+
+  // If the two input values are a string then we want to use the krnlstrncmp.
+  // However, if the input values are a float or an int we can simply use the
+  // equal function.
+  if (inputElemType.isa<krnl::StringType>()) {
+    Value strlenRes = create.krnl.strlen(lhs);
+    Value strncmpRes = create.krnl.strncmp(lhs, rhs, strlenRes);
+    // We need to convert the results to return *i1 since krnlstrncmp returns
+    // i32.
+    results = create.math.cast(rewriter.getI1Type(), strncmpRes);
```
hamptonm1, line 1070: Here we can use `rewriter.getI1Type()` or `elementType` since they will both yield *i1 Type. Let me know if anyone has a preference. Thanks!

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1050,11 +1051,27 @@ template <>
...
-  return create.math.eq(lhs, rhs);
+  MultiDialectBuilder<KrnlBuilder, MathBuilder> create(rewriter, loc);
+  Type inputElemType = getElementType(lhs.getType());
+
+  // If the two input values are a string then we want to use the krnlstrncmp.
+  // However, if the input values are a float or an int we can simply use the
+  // equal function.
+  if (inputElemType.isa<krnl::StringType>()) {
+    Value strlenRes = create.krnl.strlen(lhs);
+    Value strncmpRes = create.krnl.strncmp(lhs, rhs, strlenRes);
```
tungld, line 1067: Just to confirm one thing here: when two strings are equal, `strncmp` returns `zero` (https://en.cppreference.com/w/c/string/byte/strncmp), it means ONNXEqual returns `zero`. But I guess MLIR interprets `zero` as `false`, doesn't it?

hamptonm1: @tungld Is this update better?

AlexandreEichenberger:
Why are the backend test disabled; we generate the lit tests, so they are not a guarantee that the code actually works. A working backend test is a better guarantee, as they know the expected result and check the compiled model run against that expected result.

hamptonm1:
> Why are the backend test disabled; we generate the lit tests, so they are not a guarantee that the code actually works. A working backend test is a better guarantee, as they know the expected result and check the compiled model run against that expected result.

@AlexandreEichenberger please look at the comment above... I explain in detail why.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1050,11 +1051,32 @@ template <>
...
+    // We need to convert the results to return *i1 since krnlstrncmp returns
+    // i32.
+    Value convertedRes = create.math.cast(rewriter.getI1Type(), strncmpRes);
```
tungld, line 1070: this looks redundant, we can create a zeroVal with `strncmpRes.getType()` then compare it to `convertedRes'.

hamptonm1: You are right... let me fix it. Thanks!!

hamptonm1: FYI- we do not even need `convertedRes`.... `convertedRes` is `i1` and `strncmpRes.getType()` is `i32`.

hamptonm1: I have an easy solution.... I changed it to `elementType` which is `i1`.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1050,11 +1051,32 @@ template <>
...
+
+  // If the two input values are a string then we want to use the krnlstrncmp.
+  // However, if the input values are a float or an int we can simply use the
+  // equal function.
+  if (inputElemType.isa<krnl::StringType>()) {
+    Value strlenRes = create.krnl.strlen(lhs);
+    Value strncmpRes = create.krnl.strncmp(lhs, rhs, strlenRes);
+    // We need to convert the results to return *i1 since krnlstrncmp returns
+    // i32.
+    Value convertedRes = create.math.cast(elementType, strncmpRes);
```
tungld, line 1070: I mean this cast is redundant. It's enough to have:
```c
 Value zeroVal = create.math.constant(convertedRes.getType(), 0);
 results = create.math.eq(convertedRes, zeroVal);
```

tungld (changes requested):

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
hamptonm1: Okay I can remove it.... you said `this looks redundant, we can create a zeroVal with strncmpRes.getType() then compare it to `convertedRes'.` and convertedRes is the cast so I was confused, I think that is why I put we do not need it.

hamptonm1: I removed it.

AlexandreEichenberger:
Thanks, then would you think that opening an issue to add the support that is needed here to support the backend tests is appropriate?

hamptonm1:
> Thanks, then would you think that opening an issue to add the support that is needed here to support the backend tests is appropriate?

@AlexandreEichenberger Sure! First I talked to @gongsu832 about the matter and he mentioned maybe @chentong319 would know how to provide the support but unfortunately he does not. So Gong mentioned Tian, I can reach out to him separately to see if I could get his help. Seems like `CategoryMapper` and `Equal` string type backend test need this support.

hamptonm1:
@AlexandreEichenberger Here is the issue I opened: **https://github.com/onnx/onnx-mlir/issues/2416**

chentong319:
> node = onnx.helper.make_node(
    "Equal",
    inputs=["x", "y"],
    outputs=["z"],
)
x = np.array(["string1", "string2"], dtype=np.dtype(object))
y = np.array(["string1", "string3"], dtype=np.dtype(object))
z = np.equal(x, y)
expect(node, inputs=[x, y], outputs=[z], name="test_equal_string")

The `np.dtype(object)` means the type as it is. I don't think it is the issue causing the failure of backend test.  From the error message you showed, I think the error is from line 125 of src/Runtime/PyExecutionSessionBase.cpp: the string type is not handled.

hamptonm1:
> > node = onnx.helper.make_node(
> > "Equal",
> > inputs=["x", "y"],
> > outputs=["z"],
> > )
> > x = np.array(["string1", "string2"], dtype=np.dtype(object))
> > y = np.array(["string1", "string3"], dtype=np.dtype(object))
> > z = np.equal(x, y)
> > expect(node, inputs=[x, y], outputs=[z], name="test_equal_string")
> 
> The `np.dtype(object)` means the type as it is. I don't think it is the issue causing the failure of backend test, since you can get the model and compile it. From the error message you showed, I think the error is from line 125 of src/Runtime/PyExecutionSessionBase.cpp: the string type is not handled.
> 
> Since you can compile the model to shared lib, try to run it with c driver within debugger. You will locate the error. Note: add "-g" into build-run-onnx-lib.sh.

@chentong319 Thanks for your review. I did some research to address the following:

np.dtype(object) for numpy version 1.23 and lower does not mean "the type as it is". According to the documentation it actually means the memory taken by the array which is now filled with pointers to Python objects. When working with strings it is best to use dtype=object and not dtype=str because the operations are not optimized otherwise.  I guess there is a difference in behavior from dtype=string vs. dtype=object.

This definition `means the type as it is.` is for version 1.24 and up of numpy as the usage of object has changed. 

However, I did look into the `src/Runtime/PyExecutionSessionBase.cpp` and added the following:

```
// string type missing
    else if (py::isinstance<py::array_t<const char*>>(inputPyArray))
      dtype = ONNX_TYPE_STRING;
```

When building onnx-mlir and then running the backend test now that string type has been defined I get a new error:

`RuntimeError: NumPy type info missing for PKc`

PKc essentially maps to string....so although the code is already implemented on the backend by Gong. It seems that numpy is not recognizing string type. I think that is why we had the comment "string type missing"

hamptonm1:
I did some further research on this error and found this:https://github.com/pybind/pybind11/issues/2226

Solution: `You may want try the arrow string array. Arrow can zero copy to numpy`

So we can try the suggestion found in this issue but I think for now I have the issue open for someone who has the knowledge to further look into implementing type string. I have tried on my own but have not figured it out and I do not think the PR should be blocked because of that.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
tungld: Thanks! and sorry for the confusion.

tungld (approved):

tungld:
LGTM!

chentong319:
Could you tried to create a test case to perform Equal on two constant string to avoid the string type input? Such test can really verify the runtime behavior of this PR. Make sure that the Equal of constants are not folded at compile time.

hamptonm1:
> Could you tried to create a test case to perform Equal on two constant string to avoid the string type input? Such test can really verify the runtime behavior of this PR. Make sure that the Equal of constants are not folded at compile time.

@chentong319 Okay...you want me to test on my personal workstation right and not add this to the source code?

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
sorenlassen: will this say the strings are equal when lhs is a prefix of rhs? I wonder if we should call `strcmp` instead of `strncmp`, or call `strncmp` with `strlenRes+1`

hamptonm1: @sorenlassen in `onnx-mlir/src/Dialect/Krnl/DialectBuilder.cpp` the only method we have defined is `strncmp` when we are doing the lowering to krnl. So that is why I used it. According to my search:

`The basic difference between these two are : strcmp compares both the strings till null-character of either string comes whereas strncmp compares at most num characters of both strings.`

So I do not see a huge advantage in terms of performance but if you see the benefit of making the change...let me know. Thanks!

sorenlassen: since you only only compare at most strlenRes then I think you don't distinguish between the cases where lhs is only a prefix of rhs or they are equal

a quick fix could be to pass strlenRes+1 instead of strlenRes as the 3rd argument to strncmp

PR #2402, opened 2023-07-27
Fix error related to tag on windows

tungld:
Though we append a tag to the symbols in a generated .so library, we use the entry point without tags **by default** on Windows. This is because we cannot define at compile time tagged entry point in header files in `include/onnx-mlir/Runtime` to make the tagged entry point visible.


src/Conversion/KrnlToLLVM/KrnlToLLVMHelper.cpp
```diff
@@ -192,8 +192,8 @@ LLVM::GlobalOp getOrCreateGlobalString(StringRef str, Location loc,
     OpBuilder &builder, ModuleOp module, LLVMTypeConverter *typeConverter) {
   MultiDialectBuilder<LLVMBuilder> create(builder, loc);
   assert(typeConverter && "Expecting a valid LLVM type converter");
-  LLVM::GlobalOp global =
-      module.lookupSymbol<LLVM::GlobalOp>("om_" + str.str());
+  LLVM::GlobalOp global = module.lookupSymbol<LLVM::GlobalOp>(
+      LLVMBuilder::SymbolPostfix(module, "om_" + str.str()));
```
tungld, line 196: This is not related to Windows, but to add a missing modification in appending tag.

AlexandreEichenberger (approved):
LGTM

PR #2403, opened 2023-07-27
Simd reduction for machines without horizontal sum/ sum across support

AlexandreEichenberger:
Perform simd for reductions where the reductions occurs in the inner dimensions of the tensors. It collapse multiple reductions dimensions into 1, perform blocking of the inner non-reduction dimension. The scheme generates 2 code pattern, one for full blocks and one for partial blocks. Full block use a N-Way permute and operate pattern to generate N results in parallel.

Note that the reduction dimensions are expected to be multiple of the SIMD Vector Length.

Performance on z16 shows a 5x reduction in execution time for the GPT2 reduction pattern (`tensor<?x?x768xf32>`)


tungld:
> Performance on z16 shows a 5x reduction in execution time for the GPT2 reduction pattern (tensor<?x?x768xf32>)

Great achivement!

tungld (approved):
LGTM! Please remove `AEE` in several comments :)

AlexandreEichenberger:
> Please remove AEE in several comments :)

It is because I am not 100% they are correct; but close enough, will remove.

PR #2404, opened 2023-07-28
Replace sequence_lens with none in RNN ops when bs is 1

tungld:
In RNN ops, sequence_lens is defined as: 

```
sequence_lens (optional, heterogeneous) - T1: Optional tensor specifying lengths of the sequences in a batch. If not specified - assumed all sequences in the batch to have length seq_length. It has shape [batch_size].
```

If batchsize is 1, it's safe to remove `sequence_lens` (set to `none`) since all sequences (one here) have the same length.


src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -687,7 +687,7 @@ class RNNOpRewriteLayoutPattern : public OpRewritePattern<ONNXOp> {
   LogicalResult matchAndRewrite(
       ONNXOp onnxOp, PatternRewriter &rewriter) const override {
     if (onnxOp.getLayout() == 0) {
-      return success();
+      return failure();
```
tungld, line 690: This is a bug causing infinite loop in applying this rule, and the pass terminated without any error, which causes other rules have no chance to be applied. It must be `failure` when the pattern does not match.

tungld:

AlexandreEichenberger (approved):
LGTM, outstanding, I was going to ask you for the same for LSTM and RNN, see that you already did it. Big thanks

PR #2407, opened 2023-07-28
add options to specify library

chentong319:
This PR actually does two things:
1. Add options for user to specify the library and library path when some op is translated into call to external library.
2. Add an option to control the ConvOptPass. Some user may want to keep the ConvOp with the similar reason that ConvOp is kept for accelerator. By default, the ConvOptPass will be invoked for CPU (unchanged).


src/Compiler/CompilerOptions.cpp
```diff
@@ -309,6 +309,22 @@ llvm::cl::opt<std::string> modelTag("tag",
     llvm::cl::value_desc("a string that matches regex ([0-9a-z_.-]+)"),
     llvm::cl::init(""), llvm::cl::cat(OnnxMlirOptions));
 
+llvm::cl::opt<bool> enableConvOptPass("enable-conv-opt-pass",
+    llvm::cl::desc("Enable the ConvOptPass. Default is true."),
+    llvm::cl::init(true), llvm::cl::cat(OnnxMlirOptions));
+
+llvm::cl::list<std::string> extraLibDirs("extra-lib-dirs",
```
AlexandreEichenberger, line 316: Love this addition, was much needed. And thanks for making it a list, that is precisely what we need.

Can we call it "-L" and "-l"? As a group, we have added a lot of options. If there is a need to distinguish, then that is fine. Standard names require no new learning from normal users.

chentong319: Changed to use "-L" and "-l"

src/Compiler/CompilerUtils.cpp
```diff
@@ -490,6 +490,16 @@ static int genSharedLib(std::string sharedLibNameWithExt,
...
+  // Add user specified libs and their path
+  // Multiple lib or directory can be specified with multiple options.
+  // For example, --extr-libs="extra1", --extra-libs="extra2"
+
+  for (std::string lib : extraLibs)
+    libs.push_back(lib);
+
+  for (std::string dir : extraLibDirs)
+    libDirs.push_back(dir);
+
```
gongsu832, line 502: Instead of changing `genSharedLib`, try to use the existing `addCompilerConfig`/`getCompilerConfig` routines like this: https://github.com/onnx/onnx-mlir/blob/76a3f070282806cbd4b6468c2b8f9ebafece554b/src/Compiler/CompilerUtils.cpp#L761

The `CCM_SHARED_LIB_DEPS` key already covers your `-l` option. For `-L`, you can define another key, for example,
```
#define CCM_SHARED_LIB_DEPS "sharedLibDirs"
```
to collect all the directories specified.

chentong319: @gongsu832 This seems to be a better solution. I made the change and please review it.

gongsu832: LGTM. Thanks.

PR #2410, opened 2023-07-31
remove --onnx-const-prop-report and ConstPropCounters

sorenlassen:
They were tedious to maintain and we didn't use them


AlexandreEichenberger (approved):
@sorenlassen : Checked with blame and it looks like you added that report; if that is the case, the LGTM. Otherwise, if it was added by someone else, please check with the folks that added it as it might be used for statistics run from time to time.

sorenlassen:
yes, I added that report in the first place and I'm not aware that anyone else has used it so I'm pretty sure no one will miss it

thanks for the review!

PR #2414, opened 2023-08-02
fix two OMUnique.inc warnings

sorenlassen:
```
In file included src/Runtime/OMUnique.c:15:
OMUnique.inc:110:12: warning: implicit conversion turns string literal into bool: 'const char[34]' to 'bool' [-Wstring-conversion]
    assert("Unsupported ONNX type in OMTensor");

In file included from src/Runtime/OMUnique.c:15:
OMUnique.inc:154:11: warning: variable 'sliceStride' set but not used [-Wunused-but-set-variable]
  int64_t sliceStride = 1;
```

### DIFF
--- a/src/Runtime/OMUnique.inc
+++ b/src/Runtime/OMUnique.inc
@@ -107,7 +107,7 @@ int isLessNum(void *arg1, void *arg2, OM_DATA_TYPE dataType) {
   // case ONNX_TYPE_COMPLEX64:
   // case ONNX_TYPE_COMPLEX128:
   default:
-    assert("Unsupported ONNX type in OMTensor");
+    assert(false && "Unsupported ONNX type in OMTensor");
   }
   return 0;
 }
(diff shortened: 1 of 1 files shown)


src/Runtime/OMUnique.inc
```diff
@@ -107,7 +107,7 @@ int isLessNum(void *arg1, void *arg2, OM_DATA_TYPE dataType) {
   // case ONNX_TYPE_COMPLEX64:
   // case ONNX_TYPE_COMPLEX128:
   default:
-    assert("Unsupported ONNX type in OMTensor");
+    assert(false && "Unsupported ONNX type in OMTensor");
```
AlexandreEichenberger, line 110: please use the recommended `llvm_unreachable`

AlexandreEichenberger (approved):
LGTM, please change to `llvm_unreachable` before merging the changes

src/Runtime/OMUnique.inc
sorenlassen: that was also my first thought until I realized that it's impossible because it must compile with C and cannot depend on llvm

AlexandreEichenberger: Got it, apologies. Should have remembered from the `.inc` file name. Please merge as is then

PR #2417, opened 2023-08-04
Do not report memory usage with --profile-ir

tungld:
Just found that `--InstrumentReportMemory` introduces a big overhead, so turn it off by default in `--profile-ir`. We can optionally enable it by using `--InstrumentReportMemory`.

For example, for mnist inference time on CPU:
- without `--InstrumentReportMemory`: 0.005210593342781067 secs
- with `--InstrumentReportMemory`: 0.11060499399900436 secs


AlexandreEichenberger (approved):
LGTM

PR #2418, opened 2023-08-07
Add option to control onnx to krnl lowering

chentong319:
Add this option to select onnx ops to be lowered to krnl.Call, instead of krnl loops. This PR only use the new option on onnx.Conv op. If the implementation is approved, more ops will be modified.


AlexandreEichenberger (commented):
Thanks for the quick update on this. 

Quick question: do you believe that there is a need for each call to result in code? Would we not be able to have a default, say `--op-for-call="ONNXGemmOp=blas_gem_from_my_favorite_lib;ONNXMatMulOp=matmul_from_my_favorite_lib"`.

chentong319:
The current option is the basic support: make decision when an op is lowered from ONNX to Krnl. Your example will be a higher level of abstraction. We can have a description file for a library, specifying the function name, parameter order, calling convention, and etc for each Op. Then we can choose which library to use for an op. 
I expect this PR to provide a preliminary hook and @NiklasGustafsson can further customize the code for his purpose.

NiklasGustafsson:
>> I expect this PR to provide a preliminary hook and @NiklasGustafsson can further customize the code for his purpose.

Huh? Did you mean to mention someone else, perhaps?

src/Conversion/ONNXToKrnl/NN/Conv.cpp
```diff
@@ -249,6 +250,18 @@ struct ONNXConvOpLowering : public OpConversionPattern<ONNXConvOp> {
     Value alloc =
         create.mem.alignedAlloc(memRefType, shapeHelper.getOutputDims());
 
+    if (checkOpToCall(op, opsForCall)) {
```
AlexandreEichenberger, line 253: I may need more of a big picture, it feels like changing onnx ops to call should be done in a separate pass, not inside each operations that might be converted. What do you think @chentong319 ?

chentong319: The big picture is to break down the current onnx to krnl lowering into several steps:
1. separate out the memref creation into a independent pass as the bufferization pass. The one-shot pass will be used.
2. A pass to lower some onnx ops to calls
3. A pass to lower the rest onnx ops to krnl dialect.

AlexandreEichenberger: Generally agree, and as an initial step, we should not conflate 2 &3 given that 3 is new. Please add a separate pass for this now.

chentong319:
The implementation of this PR is changed. Two pattern for lowering are introduced: one pattern is the original one that lowers the op to loops of krnl, and the other one lowers the op to krnl.Call. Which pattern to use is determined by the command line option. Currently the decision is only based on the name of the operation. 
The two transformations share the common code that allocates the buffer for the Op. I defined the utility function within the transformation file (NN/Conv.cpp). There are obstacles that prevent generalizing that utility
1. For most op, the buffer allocation code is similar: initialize its shapeHelper, compute the shape, and then create buffer according to the output shape. However, ops have their type for ShapeHelper and need code to specify that.
2.  Some op has special code to generate its buffer. Not sure whether we can put those code into ShapeHelper or not. 
Overall, the current PR is intended to provide a way to generate function call, not to create a separate pass to do buffer allocation for onnx Ops.

src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
```diff
@@ -240,6 +240,28 @@ Value getDimOrConstant(ConversionPatternRewriter &rewriter, Location loc,
              : create.math.constant(type, shape[axis]);
 }
 
+/// Check whether this op should be lowered to Krnl.Call according to option
+/// opsToCall. The op name is used for matching
+bool checkOpToCall(Operation *op, std::string opsForCall) {
```
AlexandreEichenberger, line 245: That implementation is fine by now; eventually if we have many entries, we may want to create a dictionary. If you make another pass at this PR, maybe add a comment indicating that if in the future we have many calls, we may want to revise this implementation.

src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.hpp
```diff
@@ -458,6 +463,90 @@ void populateLoweringONNXShapeTransformOpPattern(
...
     mlir::RewritePatternSet &, mlir::TypeConverter &, mlir::MLIRContext *);
 
+// Utilities for generating krnl.call for ONNX Ops
+
+// Create allocate based on COMPUTED shapeHelper.
+// The generic computed shapehelper avoids the specific type of shape helper
+// for each op, and shape helper may be used in krnl loop generation, too.
+// Use template in case some op has special allocation. For the generic cases,
+// the typename is only used in location, which is not absolutely needed
+template <typename OP_TYPE>
```
AlexandreEichenberger, line 473: Very generic implementation, should promote using its use in our future ops implementation.

AlexandreEichenberger (approved):
LGTM, great work.

My only suggestion is that we now start to have quite a bit of code in the `ONNXToKrnlCommon.hpp`; maybe time to migrate this into an `ONNXToKrnlCommon.inc`. Can be done in a successive PR.

Issue #2326, opened 2023-06-17
compile error


wm901115nwpu:
╰─ cmake --build . -j 1                                                                                                                                                                 ─╯
[1/696] Running gen_proto.py on onnx/onnx.in.proto
Processing /Users/unicorn/workspace/onnx-mlir/third_party/onnx/onnx/onnx.in.proto
Writing /Users/unicorn/workspace/onnx-mlir/build/third_party/onnx/onnx/onnx-ml.proto
Writing /Users/unicorn/workspace/onnx-mlir/build/third_party/onnx/onnx/onnx-ml.proto3
generating /Users/unicorn/workspace/onnx-mlir/build/third_party/onnx/onnx/onnx_pb.py
[3/696] Running gen_proto.py on onnx/onnx-operators.in.proto
Processing /Users/unicorn/workspace/onnx-mlir/third_party/onnx/onnx/onnx-operators.in.proto
Writing /Users/unicorn/workspace/onnx-mlir/build/third_party/onnx/onnx/onnx-operators-ml.proto
Writing /Users/unicorn/workspace/onnx-mlir/build/third_party/onnx/onnx/onnx-operators-ml.proto3
generating /Users/unicorn/workspace/onnx-mlir/build/third_party/onnx/onnx/onnx_operators_pb.py
[5/696] Running gen_proto.py on onnx/onnx-data.in.proto
Processing /Users/unicorn/workspace/onnx-mlir/third_party/onnx/onnx/onnx-data.in.proto
Writing /Users/unicorn/workspace/onnx-mlir/build/third_party/onnx/onnx/onnx-data.proto
Writing /Users/unicorn/workspace/onnx-mlir/build/third_party/onnx/onnx/onnx-data.proto3
generating /Users/unicorn/workspace/onnx-mlir/build/third_party/onnx/onnx/onnx_data_pb.py
[7/696] Building CXX object third_party/onnx/CMakeFiles/onnx_proto.dir/onnx/onnx-ml.pb.cc.o
FAILED: third_party/onnx/CMakeFiles/onnx_proto.dir/onnx/onnx-ml.pb.cc.o
[...]

PR #2423, opened 2023-08-08
Protobuf 3.20.3 instructions to fix compilation on M1+ Macs.

AlexandreEichenberger:
Left instructions that allowed me to fix a bug for protobuf on M1, which is partly due to brew / pip issues and partly due to the currently "mandated" protobuf version leading to a missing link issue when compiling `onnx-mlir` on Macs M1.

Big thanks to @sorenlassen for indicating in an issue here the root of the solution. https://github.com/onnx/onnx-mlir/issues/2326

Thought that as long as we are stuck on an older unfixed protobuf version (version is linked to the one used in the ONNX repo), it might be helpful to have help a bit more prominent that in an issue. Hope this will help further folks using M1 Macs


sorenlassen:
I can see that the MacOS amd64 CI pipeline failed.

I'm guessing you need to update utils/install-protobuf.sh (called from .github/workflows/macos-amd64-build.yml)

utils/install-protobuf.sh
```diff
@@ -1,5 +1,5 @@
 # Check out protobuf v3.18.3
```
sorenlassen, line 1: please update the comment too

docs/BuildOnLinuxOSX.md
```diff
@@ -97,6 +99,22 @@ make: *** [Makefile:146: all] Error 2.
 
 The suggested workaround until jsoniter is fixed is as follows: install maven (e.g. `brew install maven`) and run `alias nproc="sysctl -n hw.logicalcpu"` in your shell.
 
+#### Protobuf issue (Mac M1, specific to protobuf 3.20.3 which is currently required)
+
+On Mac M1, you may have some issues building protobuf. In particular, you may fail to install onnx (via `pip install -e third_party/onnx`) or you may fail to compile `onnx-mlir` (no arm64 symbol for `InternalMetadata::~InternalMetadata`).
+
+The first failure is likely an issue with having multiple versions of protobuf.
+Installing a version with `brew` was not helpful (version 3.20.3 because of a known bug that can be corrected with a patch below).
+Uninstall the brew version, and make sure you install the right one with pip: `pip install protobuf==3.20.3`.
```
sorenlassen, line 108: is it counterproductive to mention how to install 3.20.3 if it doesn't work to compile onnx-mlir and the real recommendation is to build from source?

AlexandreEichenberger: Surprisingly, I had to have it installed via pip because otherwise `pip install` would automatically download via pip the latest (3.4.*), which is incompatible. So it needs to have 3.20.3 installed via pip to make the `pip install -e thrid_party/onnx` install happy, but having the code version on the path for the building of onyx-mlir compiler.

Would you like me to be more explicit about this point?

utils/install-protobuf.sh
AlexandreEichenberger: thanks

docs/BuildOnLinuxOSX.md
sorenlassen: I see, now I see your description makes good sense

however, I think it's actually simpler to just "pip install onnx" (or "pip instlal onnx==$(cat third_party/onnx/VERSION_NUMBER)" if pip doesn't pick the right version automatically), then you don't need pip's protobuf

sorenlassen (approved):

docs/BuildOnLinuxOSX.md
AlexandreEichenberger: I have not tried this. For some reason, I assumed that we might need the third_party implementation... but if that is not the case, then it's even easier.

I may have to uninstall protobuf to try it out. Is that all you did? Namely no installation of protobuf, just `pip install onnnx=3.20.3` ?

AlexandreEichenberger: If so, I can write it as an alternative to try. Our CIs definitely don't go that route.

sorenlassen: > Is that all you did? Namely no installation of protobuf, just `pip install onnnx=3.20.3` ?

it's the onnx version (1.14.0), not protobuf version

I got the python onnx module with `pip install onnx` (it happened to pick the correct version, I think pip install onnx==1.14.0` would also have worked)

AlexandreEichenberger:

Issue #2422, opened 2023-08-08
Undefined symbol when running inference on NNPA


tungld:
With the latest commit, I got an error when running a model on NNPA:
```
python: symbol lookup error: /tmp/tmpqjy1e8mn/model.so: undefined symbol: OMInitCompatibleAccelNNPA`
```

It seems it's related to this PR #2407 that changed the way libraries are linked.

tungld:
@chentong319 any thought?

gongsu832:
It's also missing `-lzdnn`. @chentong319 it seems that the options for the accelerators are not being collected.

chentong319:
The library for NNPA is added in the constructor of NNPAAccelerator. The fix for repeatedly adding lib to configure by compiling module cleared those library. I will try to fix it.

PR #2424, opened 2023-08-08
Fix issue 2422

chentong319:
Problem with PR2407: the libraries for accelerator NPAA is added in the constructor, while other library configure is added in CompileModule(). When the CompileModule is repeatedly called, the library config should not be added again and again. PR2407 simply clear the configure. That caused problem for NPAA.
This PR implements a stack for the configuration. The configuration is push at the begin of CompileModule() and pop at the end of the procedure.  It is a little bit overkill for issue #2422 .


gongsu832:
After thinking a bit more, I think it's wrong to have `xxxCompilerConfig()` stuff in `compileModule()` and `emitOutputFiles()` functions. All the options related stuff should have been processed before we call `compileModule()`.

We have this code in `onnx-mlir.cpp`:
```
  if (!parseCustomEnvFlagsCommandLineOption(argc, argv, &llvm::errs()) ||
      !llvm::cl::ParseCommandLineOptions(argc, argv,
          getVendorName() + " - A modular optimizer driver\n", &llvm::errs(),
          customEnvFlags.c_str())) {
    llvm::errs() << "Failed to parse options\n";
    return 1;
  }
  // Test option requirements.
  if (!ONNXOpStats.empty() && emissionTarget <= EmitONNXIR)
    llvm::errs()
        << "Warning: --onnx-op-stats requires targets like --EmitMLIR, "
           "--EmitLLVMIR, or binary-generating emit commands.\n";

```
So I think `addCompilerConfig()` should be done here based on the `emissionTarget`. Our options handling has always been a bit "messy". For example, we have a `CompilerOptions.cpp` but in `onnx-mlir.cpp` we also have a few options there. Maybe we should take this opportunity to clean things up a bit.

Ideally, I think the above code should be in a general function (e.g., processOptions) that can be called by others like onnx-mlir.cpp or the ModelLib.cpp (for tests). What this function does is to parse the options and set the relevant option states. But it sounds like `llvm::cl::ParseCommandLineOptions` should only be called from main.

chentong319:
@gongsu832 I agree that the options/config should be better organized. Currently, the addCompilerConfig goes along the action. We can separate addCompilerConfig out with some effort.  Should I do it in this PR?

gongsu832:
@chentong319 yes at a minimum we should move the `xxxCompilerConfig` stuff out of `compileModule()` and `emitOutputFiles()`

tungld:
It would be helpful if we can add some lit tests for checking linking libraries, so that we can avoid regression.

chentong319:
> @chentong319 yes at a minimum we should move the `xxxCompilerConfig` stuff out of `compileModule()` and `emitOutputFiles()`

When I tried this, I found that the numerical tests need change too. I am not familiar with that part of code. Could we just keep the current local change inside compileModule() for this PR?

chentong319:
> It would be helpful if we can add some lit tests for checking linking libraries, so that we can avoid regression.

Done. It turned out that the error of link command does not effect the lit test.

chentong319:
@gongsu832 I looked into how to move the addCompilerConfig out of compileModule further. It's complicated how the options are gotten and interpreted from string to enum type, and how the env variable is read for calling compiler from different invocation methods. My current change is limited inside compileModule and is a safe, though not the best, solution. How do think?

gongsu832:
> @gongsu832 I looked into how to move the addCompilerConfig out of compileModule further. It's complicated how the options are gotten and interpreted from string to enum type, and how the env variable is read for calling compiler from different invocation methods. My current change is limited inside compileModule and is a safe, though not the best, solution. How do think?

Sure. I can put out a PR to consolidate the options and hopefully do things the proper way.

tungld:
@gongsu832 if it's ok, could you approve it? this issue is blocking NNPA inference. Thanks!

chentong319:
@tungld could you review at least the lit test case I added? I am not sure about the lit test configuration.

gongsu832 (approved):
Temporary workaround to unblock NNPA compilation.

test/mlir/accelerators/nnpa/driver/compiler-config.mlir
```diff
@@ -0,0 +1,16 @@
+
+// RUN: onnx-mlir --maccel=NNPA -v -tag="test" %s -o %t 2>&1 | FileCheck %s
+
+// -----
+
+// REQUIRES: system-s390
```
tungld, line 6: Where did you know to use "system-s390"? I could not find it in llvm repository. 
When I check "system-s390" on z14, the lit test is excluded. You can check on your side, by looking the result of `make check-onnx-lit`:
```
Testing Time: 4.54s
  Unsupported:   5
  Passed     : 235
```
It a lit test is excluded, it will go to "Unsupported"

tungld: You can use `// REQUIRES: system-linux` instead to run the test on linux. Or use `lit.local.cfg` to test on big endian machine only: https://github.com/onnx/onnx-mlir/blob/main/test/mlir/accelerators/nnpa/module_op_be/lit.local.cfg

chentong319: I thought system take the name of the machine from `uname` command. This test can only work on machine with zDNN library, not just zDNN.h. I will try system-linux plus big endian. Now I recalled that I wanted to add my test into module_op_be directory. That's the right place to share the local config.

tungld: I guess "system-s390" does not exist, so there is no system satisfying the requirement, so the test will be disabled in all environment.

tungld (approved):
LGTM.

chentong319:
@tungld You may merge the PR after it passes the test.

tungld:
> @tungld You may merge the PR after it passes the test.

OK.

PR #2427, opened 2023-08-09
Use DimAnalysis when lowering ONNXReshape to krnl

tungld:
This patch is to use DimAnalysis when lowering ONNXReshape. In particular, DimAnalysis is used to detect "identity" reshape, so that we can remove reshape.

### DIFF
--- a/src/Conversion/ONNXToKrnl/ConvertONNXToKrnl.cpp
+++ b/src/Conversion/ONNXToKrnl/ConvertONNXToKrnl.cpp
@@ -211,7 +211,7 @@ void populateONNXToKrnlConversionPattern(RewritePatternSet &patterns,
   // Tensor
   populateLoweringONNXArgMinMaxOpPattern(patterns, typeConverter, ctx);
   populateLoweringONNXDimOpPattern(patterns, typeConverter, ctx);
-  populateLoweringONNXReshapeOpPattern(patterns, typeConverter, ctx);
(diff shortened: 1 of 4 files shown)


AlexandreEichenberger (commented):
Smart. 

Quick question about where to have this optimization. If we had this as a onnx-to-onnx optimization, wouldn't we have better shape inference, because this to-be-removed reshape would not cloud the shape inference analysis?

tungld:
> If we had this as a onnx-to-onnx optimization, wouldn't we have better shape inference, because this to-be-removed reshape would not cloud the shape inference analysis?

Absolutely, I intended to do so, and we might need to have a separate onnx-to-onnx pass that utilizes DimAnalysis. Currently, onnx-to-onnx is done via canonicalization of each op, but I haven't found a way to use DimAnalaysis in canonicalization pattern. Anyway, using DimAnalysis in onnx-to-krnl is orthogonal to using it in onnx-to-onnx, so I make a change in onnx-to-krnl first.

AlexandreEichenberger (approved):
Sure, I buy this. Maybe open an issue for the onnx-to-onnx pass, with a note that if the case solved here if caught earlier, maybe the code here can then be omitted.

tungld:
> Maybe open an issue for the onnx-to-onnx pass, with a note that if the case solved here if caught earlier, maybe the code here can then be omitted.

Yes, created an issue #2437

Issue #2425, opened 2023-08-08
Update onnx-mlir/utils/documentOps.py to include min and max version 


hamptonm1:
We want to correctly capture the minimum and maximum supported operation version for `onnx-mlir`.

hamptonm1:
The PR is merged... so we can close this out. Thanks!

PR #2428, opened 2023-08-09
Added logic for simple min/max when generating the md files that describe our support

AlexandreEichenberger:
Can add `==MIN== <num>` or `==MAX== <num>` comments to specify min/max onset supported. 

Otherwise, the min/max is defined by a default min/max value in conjunction with the values of `version_dict` in `gen_onnx_mlir.pl` 

```
        # ==OP== Abs
        # ==MIN== 8
        "test_abs_cpu": {STATIC_SHAPE:{}, DYNAMIC_SHAPE:{-1:{-1}}, CONSTANT_INPUT:{-1}},
```

The resulting table is as below:

== Table ===========================

Onnx-mlir currently supports ONNX operations targeting up to opset 19. Limitations are listed when applicable.

* Operations are defined by the [ONNX Standard](https://github.com/onnx/onnx/blob/main/docs/Operators.md).
* Opset indicates, for each operation, the minimum and the maximum ONNX opset that is supported by the current version of onnx-mlir. Limitation field indicates limitations in the current implementation, if any.

| Op |Min Opset |Max Opset |Limitations |
| --- |--- |--- |--- |
| **Abs** |9 |18 | |
| **Acos** |7 |18 | |
| **Acosh** |9 |18 | |
| **Add** |9 |18 |No support for short integers. |

### DIFF
--- a/utils/documentOps.py
+++ b/utils/documentOps.py
@@ -26,6 +26,12 @@ import json
 import subprocess
 
 ################################################################################
+# Default min/max opset supported (when not explicitly specified).
+# Default values are used when no explicit ==MIN==/==MAX== values are used.
+min_opset_default = 9 
+max_opset_default = 18
+
+################################################################################
 # SEMANTIC for LABELING (one line per directive)
 #
 # ==ARCH== <arch>
@@ -42,19 +48,15 @@ import subprocess
 #   where <text> add "private" info about what needs to be fixed. 
 #
 # ==MIN== <num>
-#   where <num> is the minimum release version supported
+#   where <num> is the minimum release version supported.
 #
 # ==MAX== <num>
-#   where <num> is the maximum release version supported
+#   where <num> is the maximum release version supported.
 #
 ################################################################################
-# Usage.
 
 ################################################################################
-# Default min/max opset supported (when not explicitly specified).
-min_opset_default = "9" # Use that value if no ==MIN== value is given.
-max_opset_default = "19" # Use that val if no ==MAX== value and no version are given
-
+# Usage.
 def print_usage():
     print('\nGenerate MD document tables for the supported ops using the labeling left in files.')
     print("For labeling format, consult the python script directly.")
@@ -120,7 +122,7 @@ def parse_file(file_name):
             assert op not in opset_dict, "Redefinition of op " + op
             assert op in list_op_version, "Define an op " + op + " that is not listed in the ops we currently handle."
             versions = list_op_version[op]
-            opset_dict[op] = ', '.join(map(lambda x: str(x), versions))
+            opset_dict[op] = versions
             m = max(versions)
             if m > hightest_opset:
                 hightest_opset = m
@@ -149,7 +151,7 @@ def parse_file(file_name):
         if p is not None:
             assert op is not None, "Min without op."
             assert op not in min_dict, "Redefinition of min for op " + op
-            min_dict[op] = p[1]
+            min_dict[op] = int(p[1])
             if debug:
                 print("Got min for op", op, ":", min_dict[op])
             continue
@@ -158,7 +160,7 @@ def parse_file(file_name):
         if p is not None:
             assert op is not None, "Max without op."
             assert op not in max_dict, "Redefinition of max for op " + op
-            max_dict[op] = p[1]
+            max_dict[op] = int(p[1])
             if debug:
                 print("Got max for op", op, ":", max_dict[op])
             continue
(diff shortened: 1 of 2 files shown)


hamptonm1:
@AlexandreEichenberger Here is the associated issue I opened : https://github.com/onnx/onnx-mlir/issues/2425

utils/documentOps.py
```diff
@@ -142,6 +174,26 @@ def print_row(array):
...
+    if op in opset_dict:
+        min_list.extend(opset_dict[op])
+    return min(min_list)
+
+# Max opset: if explicit ==MAX== value, use as is. Otherwise max of default and
+# custom versions delivered in opset_dict.
+def compute_max_opset(op):
+    if op in max_dict:
+        return max_dict[op]
+    max_list = [max_opset_default]
```
hamptonm1, line 192: Instead of setting a hardcoded value for the max_opest... why not use the largest or first element of that operation in the `version_dict` which is located in  onnx-mlir/utils/gen_onnx_mlir.py. This value should be accurate and would essentially reflect what is currently being supported. Not every op has support for 18, so I do not think this should be hardcoded in.

utils/documentOps.py
```diff
@@ -142,6 +174,26 @@ def print_row(array):
         str += a + " |"
     print(str)
 
+# Min opset: if explicit ==MIN== value, use as is. Otherwise min of default and
+# custom versions delivered in opset_dict.
+def compute_min_opset(op):
+    if op in min_dict:
+        return min_dict[op]
+    min_list = [min_opset_default]
```
hamptonm1, line 182: My thoughts are the same for min_opset... I do not think we should hardcode it, it would be safer to use the smallest or last number in the list for `version_dict`. From my exercise, not all operations even support a version 9, so this would be misleading.

utils/documentOps.py
```diff
@@ -142,6 +174,26 @@ def print_row(array):
         str += a + " |"
     print(str)
 
+# Min opset: if explicit ==MIN== value, use as is. Otherwise min of default and
+# custom versions delivered in opset_dict.
+def compute_min_opset(op):
+    if op in min_dict:
```
hamptonm1, line 180: What about the operations that only support one version? For example `MatMulInteger` only has version 10 available in `ONNX`. So we should put MIN=NA and MAX=10. We should account for this situation where only one version is givenl and put NA for the MIN versions.

utils/documentOps.py
hamptonm1: However, if you think this would be too much to do the above... I would prefer using NA as the default values then hardcoding values we can not be sure of. NA is a given standard and we can easily make the updates once min/max are explicitly defined.

hamptonm1 (commented):
@AlexandreEichenberger Thanks for working on this! I have a few comments :)

utils/documentOps.py
AlexandreEichenberger: @hamptonm1, the value in `version_dict` only say "we now support an op that is incompatible with prior ops"... Most op that get updated from one onset to the next do not break compatibility, and thus don't show up in the version_dict even though we did quite a bit of work to make them work.

My suggestion: I would put the default min to a set that we largely support for most ops, and default max to the current onset (say 19). And then whenever there is an op that we don't support in the latest, we say MAX=13 (if we stop supporting at 14 and higher). Then if we fix this op to support 19, we simply remove the MAX. 

If you prefer to rename "MAX 13" to "NO_SUPPORT 14" (as last supported of an op at 13 means that we don't support 14...) that is fine by me too.

utils/documentOps.py
hamptonm1: You are not understanding my example..... MatMulInteger ONLY has version 10. There is no 9 or 11. As of today there is only one version available. If you search MatMulInteger for ONNX via the documentation you will see what I mean. I think the point I am saying is if for specific operation there is only support for one version (no previous version and no newer version). Should we make the MIN=10 and MAX=NA and if there is a new version that eventually comes out then we just update the MAX to the newer version.

utils/documentOps.py
hamptonm1: > @hamptonm1, the value in `version_dict` only say "we now support an op that is incompatible with prior ops"... Most op that get updated from one onset to the next do not break compatibility, and thus don't show up in the version_dict even though we did quite a bit of work to make them work. If it is confusing, I would suggest to remove looking at these values and simply rely on the default min/max and the explicit min/max.
> 
> My suggestion: I would put the default min to a set that we largely support for most ops, and default max to the current onset (say 19). And then whenever there is an op that we don't support in the latest, we say MAX=13 (if we stop supporting at 14 and higher). Then if we fix this op to support 19, we simply remove the MAX.
> 
> If you prefer to rename "MAX 13" to "NO_SUPPORT 14" (as last supported of an op at 13 means that we don't support 14...) that is fine by me too.

This is what is commented in the script above `version_dict`

```
# Record the version of each operation that is treated as the current version.
# To check whether the onnx package being used has newer version operation,
# run this script with --check-operation-version flag.
# Update this dictionary when a newer version is implemented
# TODO: how to keep the old version
```

I am assuming the exercise for this is to specify the minimum version we do support and the max version we support which varies based on operation. So it does not make sense to hardcode a value that may not be applicable to said operation. That I am not understanding. 

From your example of MAX.... I would be very clear in terms of what is supported here. So let's say minimum version support for MAX is 6 and the max is 13 as you mentioned 14 is not supported. The MIN should be 1 and MAX 13. This makes more sense to me than having "NO_SUPPORT_14".Min and max should be the inclusive range we support. So the lowest value up until the highest value. Just because we have Opset 19 available does not mean every operation is compatible with version 19. So this would create more confusion than understanding if we were to hard code values.

hamptonm1 (approved):
LGTM.... we were able to further clarify things via a meeting.

PR #2432, opened 2023-08-10
[NNPA] Remove consecutive zlow.unstick and zlow.stick ops if they use the same layout

tungld:
This patch is to remove consecutive zlow.unstick and zlow.stick ops if they use the same layout.

For example,
```mlir
zlow.unstick(%x, %y) {layout = 3DS}
zlow.stick(%y, %z) {layout = 3DS}
zlow.relu(%z)
```

will be replaced by
```mlir
zlow.relu(%x)
```
where `zlow.unstick` and `zlow.stick` are removed if they don't have any other use.

### DIFF
--- a/src/Accelerators/NNPA/Transform/ZLow/ZLowRewrite.cpp
+++ b/src/Accelerators/NNPA/Transform/ZLow/ZLowRewrite.cpp
@@ -33,6 +33,84 @@ using namespace mlir;
 namespace onnx_mlir {
 namespace zlow {
 
+/// Remove unstick if there is no use of its second operand except itself.
+class UnstickRemovalPattern : public OpRewritePattern<ZLowUnstickOp> {
+public:
+  using OpRewritePattern<ZLowUnstickOp>::OpRewritePattern;
+  LogicalResult matchAndRewrite(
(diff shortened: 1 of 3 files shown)


chentong319:
A naive question: why this optimization can not be done at ZHigh level?

tungld:
> A naive question: why this optimization can not be done at ZHigh level?

I have seen `zhigh.unstick {layout = 3DS} -> Reshape (%input, %shape) -> zhigh.stick {layout = 3DS}`, and we have no optimization that checks if Reshape is identity or not given dynamic dimensions so that it can become `zhigh.unstick {layout = 3DS} -> zhigh.stick {layout = 3DS}` and be removed at ZHigh.

PR #2427 removes Reshape during onnx-to-krnl by utilizing DimAnalysis, then at ZLow, we have  `zlow.unstick(%x, %y) {layout = 3DS} -> zlow.stick(%y, %z) {layout = 3DS}` that is removed by this patch. Issue #2437 is to propose a onnx-to-onnx optimization that utilizes DimAnalysis.

Besides, without knowledge of high-level optimizations (ONNX, ZHigh, or potentially another new dialect), it would be beneficial to implement all possible optimizations at a level (say, ZLow).

AlexandreEichenberger (approved):
LGTM

PR #2434, opened 2023-08-10
support string input for c interface

chentong319:
To me, it is easier to debug to run a model with the c interface. I found some issues in supporting string type input. 

1. The EntryPoint signature can not be parsed by llvm::Jason::parser because of the missing quotation marks. I also simplified the type from !krnl.StringType to just "string"
2. In the driver, utils/RunONNXLib.cpp, string type inputs are not handled. I added a simple workaround, which need further work. I think that the OMTensor utilities should be modified for string type, and the driver should use omTensorCreateWithRandomData().

With this PR, the test_equal_string can be ran with a tensor containing the same string for every element.


AlexandreEichenberger (approved):
LGTM

PR #2438, opened 2023-08-14
Options handling cleanup

gongsu832:
- Consolidate all options handling (except individual pass options) into CompilerOptions.{h|c}pp
- Use external storage for options so they are globally accessible
- initCompilerConfig performs all options related configuration state generation, with "hook" to allow caller specific processing
- Localize ExternalUtil.hpp so that it's only directly accessible by CompilerOptions.cpp. Others access by getToolPath
- push/popCompilerConfig no longer needed and all xxxCompilerConfig calls have been moved out of compileModule call tree
- support model file input from stdin, e.g., one can do cat model.onnx | onnx-mlir --maccel=NNPA --EmitLib


src/Compiler/CompilerOptions.hpp
```diff
@@ -161,10 +173,14 @@ std::string getCompilerOption(const onnx_mlir::OptionKind kind);
...
-// CompilerConfig may be set at initialization or inside CompileModule().
-// Since CompileModule() may be called repeated, push is used to mark
-// the current config and pop is used to restore the previously mark config
-void pushCompilerConfig(std::string k);
-void popCompilerConfig(std::string k);
 
+// Functions related to initializing compiler configuration states based on
+// parsed options
+std::optional<std::string> getEnvVar(std::string name);
+std::string getExecPath(void);
```
sorenlassen, line 180: nit: we generally omit `void` from C++ function signatures without arguments (try `grep -Er '^[a-zA-Z].*\((void)?\);$' src`)

src/Compiler/CompilerOptions.cpp
```diff
@@ -709,21 +888,160 @@ void delCompilerConfig(std::string k, std::vector<std::string> v) {
...
+// initCompilerConfig in a single function is that according to llvm doc
+// llvm::cl::ParseCommandLineOptions should be called from main.
+//
+// The caller can pass in a "hook" function to perform special handling
+// of the options after they are parsed. For example, onnx-mlir must
+// have the input filename specified and it derives the output filename
+// from the input filename if -o is not specified. onnx-mlir-opt, on the
+// other hand, defaults to stdin/stdout if the input/output filenames
+// are not specified.
+void initCompilerConfig(void (*hook)(void) /* nullptr by default */) {
```
sorenlassen, line 999: suggestion: skip the hook argument and instead change the caller to call the hook after calling initCompilerConfig - that seems a little bit simpler and I'm skeptical whether we'll use the hook mechanism for many other things in the future if you can only pass a single hook

sorenlassen (commented):
the changes generally look good to me

just two minor comments

I'm not familiar with all the code that is affected by the changes so I hope someone else will review it too

src/Compiler/CompilerOptions.hpp
gongsu832: Then I suggest we change all function prototypes without arguments to `func(void)`.

`func(void)` and `func()` mean different things. `func(void)` means the function takes no argument. `func()` means there is no information about the arguments so you can call `func(a, b, c)` and the compiler will not generate a warning.

sorenlassen: no, that's C and not C++ https://stackoverflow.com/a/51080

in C++ it's most standard to use `()` to mean no arguments

gongsu832: Quote from the link you provided:

By writing foo(void), therefore, we achieve the same interpretation across both languages and make our headers multilingual (though we usually need to do some more things to the headers to make them truly cross-language; namely, wrap them in an extern "C" if we're compiling C++).

So to me, even though the two are the same in C++, I personally would still use `func(void)`. I don't why MLIR does it the other way but fine I will change it.

src/Compiler/CompilerOptions.cpp
AlexandreEichenberger: Given the fact that the hook is called at the end, I would lean toward @sorenlassen suggestion too.

src/Compiler/CompilerOptions.cpp
```diff
@@ -303,32 +452,62 @@ llvm::cl::opt<std::string> modelTag("tag",
...
+static llvm::cl::opt<ProfileIRs, true> profileIROpt("profile-ir",
     llvm::cl::desc("Profile operations in an IR"),
+    llvm::cl::location(profileIR),
     llvm::cl::values(clEnumVal(None, "No profiling. Default value."),
         clEnumVal(
             Onnx, "Profile operations in ONNXIR generated by --EmitONNXIR.")
             APPLY_TO_ACCELERATORS(ACCEL_PROFILEIR_CL_ENUM)),
     llvm::cl::init(ProfileIRs::None), llvm::cl::cat(OnnxMlirOptions));
 
+// Options for onnx-mlir-opt only
```
AlexandreEichenberger, line 487: High level question: so some of these options were only declared in a cpp file for `onnx-mlir` and others only for `onnx-mlir-opt`. Are we now defining all of these for both? Does it means that when we type `--help` for either, we get the union of all options, including some that don't apply for that program?
If so, that would be unfortunate.

AlexandreEichenberger (commented):
@gongsu832 in general, I am very supportive of simplifying this. Just need to make sure we get it right in practice.

Personal nit: some options are defined with "thisIsTheBestEverOption" and others as "this-is-the-best-ever-option". 
My suggestion would be to go with the second style, and change it systematically except for the `-Emit`... options that are probably heavily used by our users.

src/Compiler/CompilerOptions.hpp
```diff
@@ -48,57 +50,67 @@ typedef enum {
...
-extern llvm::cl::opt<bool> enableConvOptPass;
-extern llvm::cl::list<std::string> extraLibPaths;
-extern llvm::cl::list<std::string> extraLibs;
-extern llvm::cl::opt<ProfileIRs> profileIR;
+// Options for onnx-mlir only.
+extern llvm::cl::OptionCategory OnnxMlirOptions;
+// Options for onnx-mlir-opt only.
+extern llvm::cl::OptionCategory OnnxMlirOptOptions;
+
+// Options known to onnx-mlir and/or onnx-mlir-opt
```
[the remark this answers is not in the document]
gongsu832: I just tried and it appears that the option belonging to the other command is silently ignored. #ifdef at compile time doesn't work because CompilerOptions.cpp is shared by both commands so at compile time you always have to include both sets of options. Maybe there is some way to partition the options using namespace but I'm not entirely familiar with that. Perhaps we can address the issue in a followup PR.

src/Compiler/CompilerOptions.cpp
gongsu832: `--help` will only show common and command specific options. However, as indicated above, you can still use an option of the other command which is silently ignored.

src/Compiler/CompilerOptions.hpp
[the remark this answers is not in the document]
tungld: I see that LLVM/MLIR organizes options by defining `registerXXXOptions` for different categories of options, and register ones when needed.

For example, in `mlir-opt` command:  https://github.com/llvm/llvm-project/blob/main/mlir/lib/Tools/mlir-opt/MlirOptMain.cpp#L491-L497 
```
 // Register any command line options.
  MlirOptMainConfig::registerCLOptions(registry);
  registerAsmPrinterCLOptions();
  registerMLIRContextCLOptions();
  registerPassManagerCLOptions();
  registerDefaultTimingManagerCLOptions();
  tracing::DebugCounter::registerCLOptions();
  ```
  
while in `mlir-reduce` command: https://github.com/llvm/llvm-project/blob/main/mlir/lib/Tools/mlir-translate/MlirTranslateMain.cpp#L89C2-L93C59
```
  registerAsmPrinterCLOptions();
  registerMLIRContextCLOptions();
  registerTranslationCLOptions();
  registerDefaultTimingManagerCLOptions();
```

Haven't dug into details to see how it does.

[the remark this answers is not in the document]
gongsu832: It turns out that there is a fairly simple way to make onnx-mlir not be able to use onnx-mlir-opt's options and vise versa. We call `llvm::cl::HideUnrelatedOptions` to make `--help` not print unrelated options. Those options are marked with `llvm::cl::ReallyHidden` flag. We can use `llvm::cl::getRegisteredOptions()` to get a map of the options and remove those that are marked with the `llvm::cl::ReallyHidden` flag. I've tried it and it actually works. I will update the PR with the fix.

To make the options fully dynamically constructed like what LLVM/MLIR does above likely requires major rework of the options code. Perhaps we can leave that as something for the future.

tungld (approved):
LGTM! 

The options become much organized. Thanks!

PR #2440, opened 2023-08-14
Update documentation for Min/Max support for Ops 

hamptonm1:
The following was included based on the updated `onnx-mlir/utils/documentOps.py` script:

1. Added the min and max op for NNPA documentation 
2. Added the min and max op for cpu documentation
3. Added backend test that were missing for `Slice` and `DFT`.
4. Also cleaned up a few things below. 

My process for detecting the min/max version for an op was first manually looking at the code and documentation. Next I took @chentong319 suggestion and `added a print to line 1379 of FrontendDialectTransformer.cpp`.  By adding a print to this C++ file and running the backend test, it printed all of the opset versions that are supported for each backend test. That is how I was able to deduce the correct opset version. Thanks @chentong319 for your suggestion!! Please let me know if you have further questions. Thank you!!


hamptonm1:
@AlexandreEichenberger I had a question... so in the case of `Cast`, it seems like what we support is just version 19 (based on doing a print of line 1379 of FrontendDialectTransformer.cpp) . Do we want to have min == 19 and max==19 in this case? Do we want to do something else? Let me know your thoughts and if this question even makes sense. FYI- @cjvolzka

test/backend/inference_backend.py
```diff
@@ -92,12 +95,14 @@ def get_test_models():
         # Adam
 
         # ==OP== Add
+        # ==MIN== 14
```
AlexandreEichenberger, line 98: @hamptonm1 A quick reading of the specs seems to indicate that we support add.6, and in fact we may not support add.14's bfloat16?

See https://onnx.ai/onnx/operators/text_diff_Add_6_14.html for differences.

test/backend/inference_backend.py
```diff
@@ -77,13 +77,16 @@ def get_test_models():
         # ==ARCH== cpu
 
         # ==OP== Abs
+        # ==MIN== 13
```
AlexandreEichenberger, line 80: A quick read seems to indicate that we do support abs 6, and in fact we may not support 13's float

https://onnx.ai/onnx/operators/text_diff_Abs_6_13.html

AlexandreEichenberger (commented):
A first reading appears to indicate that the MIN field may be too conservative.

AlexandreEichenberger:
> @AlexandreEichenberger I had a question... so in the case of Cast, it seems like what we support is just version 19 (based on doing a print of line 1379 of FrontendDialectTransformer.cpp) . Do we want to have min == 19 and max==19 in this case? Do we want to do something else? Let me know your thoughts and if this question even makes sense. 

There is generally no need to put a max to the current version, as this only indicates that this op will not support the future releases (20 and up).

test/backend/inference_backend.py
hamptonm1: @AlexandreEichenberger According to this PR that was merged: https://github.com/onnx/onnx-mlir/pull/1926, we have support for `bfloat16`. In fact, if we look at the documentation `onnx-mlir/src/Dialect/ONNX/ONNXOps.td.inc` it now says we have support for `TensorOf<[BF16]>]>:$C)`. I am assuming this documentation would not be updated unless `onnx-mlir/utils/gen_onnx_mlir.py` was updated with the corresponding opset number. The current version in `onnx-mlir/utils/gen_onnx_mlir.py` for the `Add` operation is 14. 

 Also when I did a print of these two tests (the trick Tong showed me). It says these tests support version 14:
 
    ```
        # ==OP== Add
        # ==LIM== No support for short integers.
        "test_add_cpu": {STATIC_SHAPE:{}, DYNAMIC_SHAPE:{-1:{-1}}, CONSTANT_INPUT:{-1}},
        "test_add_bcast_cpu": {STATIC_SHAPE:{}, DYNAMIC_SHAPE:{-1:{-1}}, CONSTANT_INPUT:{-1}},
```

```
[10/10] cd /Users/meganhampton/zDLC/onnx-mlir/build/test/backend && ONNX_...n3.11 /Users/meganhampton/zDLC/onnx-mlir/build/test/backend/Debug/test.py
opset number: 13
opset number: 7
opset number: 7
opset number: 9
opset number: 9
**opset number: 14
opset number: 14**
```

**Comment by hamptonm1** [comment]
> > @AlexandreEichenberger I had a question... so in the case of Cast, it seems like what we support is just version 19 (based on doing a print of line 1379 of FrontendDialectTransformer.cpp) . Do we want to have min == 19 and max==19 in this case? Do we want to do something else? Let me know your thoughts and if this question even makes sense.
> 
> There is generally no need to put a max to the current version, as this only indicates that this op will not support the future releases (20 and up).

So what should we put in the `max` field? Should it be `NA` then or left blank?

File: test/backend/inference_backend.py (hunk 1 above)
Comment (hamptonm1) [reply test/backend/inference_backend.py:98]: Also we have to be careful because `Add.6` supports the following attributes:

```
### Attributes
--
  |  
  | * **axis - INT** :   If set, defines the broadcast dimensions. See doc for details.
  | * **broadcast - INT** (default is '0'):   Pass 1 to enable broadcasting


```

However, these attributes were removed in `Add.7` and up. So I guess I am not sure if we really support version 6.

File: docs/SupportedONNXOps-NNPA.md [hunk 3]
```diff
@@ -12,32 +12,32 @@ Onnx-mlir currently supports ONNX operations targeting up to opset 18. Limitatio
...
-| **ReduceMean** |18, 13 |- `keepdims` must be 1.<br>- Input tensor must be 4D tensors and `axis` must be [2, 3]. | |
-| **Relu** |14 |Input tensor must be less than or equal to 4 dimensions. | |
-| **Sigmoid** |13 |Input tensor must be less than or equal to 4 dimensions. | |
-| **Softmax** |13, 11 |- `axis` must be the last dimension, i.e. `rank - 1` or -1. | |
-| **Sub** |14 |- Shape of input tensors should be the same since broadcasting is not supported.<br>- Input tensors must have static dimensions. | |
-| **Sum** |13 |- All inputs must have the same static shape (Broadcasting not supported.)<br>- Single input not supported. | |
-| **Tanh** |13 |Input tensor must be less than or equal to 4 dimensions. | |
+| Op |Min Opset |Max Opset |Limitations |Notes |
+| --- |--- |--- |--- |--- |
+| **Add** |14 |18 |- Shape of input tensors must be the same since broadcasting is not supported.<br>- Input tensors must have static dimensions. | |
```
cjvolzka, line 17: @hamptonm1 Why 18 for this? Shouldn't it either be 19 (the latest OpSet this onnx-mlir supports), 14 (the last updated version of Add) or * (if we go with my other comment).

docs/SupportedONNXOps-NNPA.md
```diff
@@ -12,32 +12,32 @@ Onnx-mlir currently supports ONNX operations targeting up to opset 18. Limitatio
...
-| **Mul** |14 |- Shape of input tensors should be the same since broadcasting is not supported.<br>- Input tensors must have static dimensions. | |
-| **Pow** |15 |- Exponent should be a scalar integer and less or equal to 64. | |
-| **ReduceMean** |18, 13 |- `keepdims` must be 1.<br>- Input tensor must be 4D tensors and `axis` must be [2, 3]. | |
-| **Relu** |14 |Input tensor must be less than or equal to 4 dimensions. | |
-| **Sigmoid** |13 |Input tensor must be less than or equal to 4 dimensions. | |
-| **Softmax** |13, 11 |- `axis` must be the last dimension, i.e. `rank - 1` or -1. | |
-| **Sub** |14 |- Shape of input tensors should be the same since broadcasting is not supported.<br>- Input tensors must have static dimensions. | |
-| **Sum** |13 |- All inputs must have the same static shape (Broadcasting not supported.)<br>- Single input not supported. | |
-| **Tanh** |13 |Input tensor must be less than or equal to 4 dimensions. | |
+| Op |Min Opset |Max Opset |Limitations |Notes |
```
cjvolzka, line 15: Nit: "Min Ver" and "Max Ver" instead of Min/Max Opset 

Basically I want to reduce the chance that someone looks at this list and thinks they have to convert their whole model to OpSet 14 because that's the "Max" (ie last updated) version of Add.

Sadly ONNX itself makes this confusing. The Op's version number is based on the OpSet where it was last updated.

test/backend/inference_backend.py
cjvolzka: We can mark it in the limitation for the Op. So we say that we support Add 6 and put in the "Limitation" field for the Op we don't support attributes X and Y.

cjvolzka:
> > > @AlexandreEichenberger I had a question... so in the case of Cast, it seems like what we support is just version 19 (based on doing a print of line 1379 of FrontendDialectTransformer.cpp) . Do we want to have min == 19 and max==19 in this case? Do we want to do something else? Let me know your thoughts and if this question even makes sense.
> > 
> > 
> > There is generally no need to put a max to the current version, as this only indicates that this op will not support the future releases (20 and up).
> 
> So what should we put in the `max` field? Should it be `NA` then or left blank?

@hamptonm1 and @AlexandreEichenberger I was going to put this in my review but it looks like there's already a thread. I'd be fine with NA or blank. Otherwise this is the idea I had written for the review....  

What would you think representing the Max something like this...
```
* Indicates we support up to the latest operator version available as of Opset 19.
| Op |Min Ver |Max Ver* |Limitations |Notes |
| --- |--- |--- |--- |--- |
| **Add** |1 | * |- ... | |
| **Fake**|6 | 12 |- ... | |
```

I'm thinking that would make it easy to discern which Ops have a "true" max vs a max of "the latest available when we wrote this doc".

So in that example, our official statement is we support up to the latest version of Add in Opset 19. When OpSet 20 comes out (assuming there isn't a change in Add) that line won't need to change. If Add *is* changed in 20, then we set Add's Max version to 14 (the last update before 20) until we support Opset 20 changes.

Say `Fake` has Op versions 6, 10, 12 and 18. However we don't support the Opset 18 changes. So the last supported version for that _Op_ is 12. The line won't need to be changed when we go to 20 unless we actually add support it.

test/accelerators/NNPA/backend/CMakeLists.txt
```diff
@@ -96,12 +96,15 @@ set(NNPA_TEST_LIST
...
 
     # ==OP== Add
+    # ==MIN== 14
     # ==LIM== - Shape of input tensors must be the same since broadcasting is not supported.<br>- Input tensors must have static dimensions.
     # Scalar tensor not supported.
     test_add_cpu,zdnn_add
     # test_add_bcast_cpu # bcast not supported
 
     # ==OP== AveragePool
+    # ==MAX== 19
```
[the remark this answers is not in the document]
hamptonm1: Please see the `onnx-mlir/utils/documentOps.py`. If the MAX or MIN are not defined that means it is defaulted to the values Alex hardcoded in the script. It should be lines 24 and 25 respectively.

test/backend/inference_backend.py
```diff
@@ -124,22 +131,28 @@ def get_test_models():
...
 
-        # ==OP==  Atanh
+        # ==OP== Atanh
+        # ==MIN== 9
         "test_atanh_cpu": {STATIC_SHAPE:{}, DYNAMIC_SHAPE:{-1:{-1}}, CONSTANT_INPUT:{-1}},
         "test_atanh_example_cpu": {STATIC_SHAPE:{}, DYNAMIC_SHAPE:{-1:{-1}}, CONSTANT_INPUT:{-1}},
 
         # ==OP== AveragePool
+        # ==MIN== 19
+        # ==MAX== 19
```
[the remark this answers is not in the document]
hamptonm1: Please see the comment above

docs/SupportedONNXOps-NNPA.md
hamptonm1: I am fine with the updates.... @AlexandreEichenberger what are your thoughts? This documented is generated when the `onnx-mlir/utils/documentOps.py` script gets run. So I will leave this to Alex :)

docs/SupportedONNXOps-NNPA.md
hamptonm1: Since I did not specify a `MAX` it defaulted to 18. @AlexandreEichenberger set the default min to 9 and the default max to 18. Given we predominately have support for those versions. From our conversations in the past.... I think we can change it to make the min 8 and the max 19. However, I will leave this to Alex as well :)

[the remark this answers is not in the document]
hamptonm1: My intent is not to be aggressive when specifying the min but rather to err on the side of caution. In my mind I would rather be safe than sorry. However, I tested `inception-v2-6.onnx` myself and it does compile. So we can still support Add.6 but not the attributes. I can do another pass at the documentation and make updates to every other operation as well. Thanks

test/backend/inference_backend.py
hamptonm1: I guess we do not have to.... I used `Netron` to analyze `inception-v2-6.onnx` and the attributes are specified and used for `Add.6`. With an `axis =1` and `broadcast=1`. But of course they no longer exist in version 7 and up.

docs/SupportedONNXOps-NNPA.md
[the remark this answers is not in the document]
hamptonm1: I also tested using `bvlcalexnet-3.onnx`, so it appears we do not support anything older than opset 6. So I can update the document to use Add.6, Conv.11 and BatchNormalization.6 :

```
meganhampton@megans-mbp-2 build % ./Debug/bin/onnx-mlir --O3 --EmitLib bvlcalexnet-3.onnx --onnx-op-stats TXT -v 
Warning: ONNX Conv in your model is using Opset 3, which is quite old. Please consider regenerating your model with a newer Opset.
Warning: ONNX Relu in your model is using Opset 3, which is quite old. Please consider regenerating your model with a newer Opset.
Warning: ONNX LRN in your model is using Opset 3, which is quite old. Please consider regenerating your model with a newer Opset.
Warning: ONNX MaxPool in your model is using Opset 3, which is quite old. Please consider regenerating your model with a newer Opset.
Warning: ONNX Conv in your model is using Opset 3, which is quite old. Please consider regenerating your model with a newer Opset.
Warning: ONNX Relu in your model is using Opset 3, which is quite old. Please consider regenerating your model with a newer Opset.
Warning: ONNX LRN in your model is using Opset 3, which is quite old. Please consider regenerating your model with a newer Opset.
Warning: ONNX MaxPool in your model is using Opset 3, which is quite old. Please consider regenerating your model with a newer Opset.
Warning: ONNX Conv in your model is using Opset 3, which is quite old. Please consider regenerating your model with a newer Opset.
Warning: ONNX Relu in your model is using Opset 3, which is quite old. Please consider regenerating your model with a newer Opset.
Warning: ONNX Conv in your model is using Opset 3, which is quite old. Please consider regenerating your model with a newer Opset.
Warning: ONNX Relu in your model is using Opset 3, which is quite old. Please consider regenerating your model with a newer Opset.
Warning: ONNX Conv in your model is using Opset 3, which is quite old. Please consider regenerating your model with a newer Opset.
Warning: ONNX Relu in your model is using Opset 3, which is quite old. Please consider regenerating your model with a newer Opset.
Warning: ONNX MaxPool in your model is using Opset 3, which is quite old. Please consider regenerating your model with a newer Opset.
Warning: ONNX Gemm in your model is using Opset 3, which is quite old. Please consider regenerating your model with a newer Opset.
Warning: ONNX Relu in your model is using Opset 3, which is quite old. Please consider regenerating your model with a newer Opset.
Warning: ONNX Dropout in your model is using Opset 3, which is quite old. Please consider regenerating your model with a newer Opset.
Warning: ONNX Gemm in your model is using Opset 3, which is quite old. Please consider regenerating your model with a newer Opset.
Warning: ONNX Relu in your model is using Opset 3, which is quite old. Please consider regenerating your model with a newer Opset.
Warning: ONNX Dropout in your model is using Opset 3, which is quite old. Please consider regenerating your model with a newer Opset.
Warning: ONNX Gemm in your model is using Opset 3, which is quite old. Please consider regenerating your model with a newer Opset.
Warning: ONNX Softmax in your model is using Opset 3, which is quite old. Please consider regenerating your model with a newer Opset.
```

test/backend/inference_backend.py
AlexandreEichenberger: Sounds good; I would then be in favor of putting it to 7 if you believe that is appropriate

docs/SupportedONNXOps-NNPA.md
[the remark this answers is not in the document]
AlexandreEichenberger: I believe opset 3 and below are antique... the return on investment to support these is very very low. We should ignore these cases.

AlexandreEichenberger: As soon as we migrate to a new level, say 19, we should put the default to 19, and every op that we don't support yet should have their max then set to 18. As we then add support for 19, removing that max (removing it entirely) will be part of the PR adding the new support.

docs/SupportedONNXOps-NNPA.md
AlexandreEichenberger: > Nit: "Min Ver" and "Max Ver" instead of Min/Max Opset

@hamptonm1 would you mind changing this as part of your PR? Search for this line`header = ["Op", "Min Opset", "Max Opset", "Limitations"]`

test/backend/inference_backend.py
hamptonm1: Okay that makes sense!

docs/SupportedONNXOps-NNPA.md
hamptonm1: @AlexandreEichenberger Yeah I can change it.... I know where it is located

test/backend/inference_backend.py
cjvolzka: > I believe most of our users will have much more recent models than Opset 6... Probably everything below 9 is very very rare.

Agreed. Outside of the modelzoo 9 is the oldest I recall seeing "in the wild" 

>I personally believe it will be less confusing to them to just say "support 7 and onwards".

13 models in the modelzoo are OpSet 6 and we support all of them so I'd go for "6 and onwards". However I'd advise caution with any "general" min statement at the OpSet level. Between ONNX's choice in how to version OpSets vs versions of the Ops themselves and mixed in with that we don't support all Ops in any OpSet version; it'll get easy to have misunderstandings fast. So I vote we keep this page focused more at the Op version level. 

For example, we'd show a min of 1 for `Conv` since it works in inception-v2-6.onnx and Conv 1 was all that was available in OpSet 6. Just because Conv supports it's version 1, doesn't mean we support OpSet 1 in general. Especially since most other Ops do have a version that we would state the minimum as 6 (OpSet 6 updated a lot of "basic" ops). At least that's my 2cents.

docs/SupportedONNXOps-NNPA.md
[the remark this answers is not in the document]
cjvolzka: Agreed. Anything less than 6 can be ignored. Six is already pretty old so we wouldn't target them either but they all already work. There was some talk of getting the zoo itself deprecate/remove models other than 6 a while back (year ago?) but I haven't heard much about it since then.

test/accelerators/NNPA/backend/CMakeLists.txt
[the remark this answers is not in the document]
cjvolzka: That makes sense. If there's a "true" max, that would show up here. If the max is just "latest currently available" then there shouldn't be a MAX value here and the script uses the default. So I'm thinking there shouldn't be a MAX value here because 19 is the latest and we support 19. (Unless by MAX, it means we _don't_ support AveragePool 19?)

test/backend/inference_backend.py
AlexandreEichenberger: @cjvolzka that make sense. You can decide 
(1) no default for min, I you want to be super explicit, or
(2) we can redefine the default min as being the floor... no number below this default.

AlexandreEichenberger:
> So what should we put in the max field? Should it be NA then or left blank?

when we have no max field, we simply don't add the `==Max==` comment. I did not provision for a NA, much simpler to just not have that field.

test/backend/inference_backend.py
cjvolzka: I'd go for option 1 then. When we add a new Op, we specify the Min version we first version we coded and/or tested for the Op. If there's no Min, that implies it's unsupported.

hamptonm1: I think we should have a default min.... because anything lower than opset 3 is not supported. There are several operations that have not changed since opset 1 for instance `Softsign`. It would not make sense to put 1 here for the minimum given anything under 6 is not supported. Having a default min just gives clear indication that the opset is still supported because there has been no changes since version 1. I think the point of this documentation is to indicate the baseline that the opset still works not when the opset was introduced like in `onnx-mlir/utils/gen_onnx_mlir.py`.

docs/SupportedONNXOps-NNPA.md
hamptonm1: @cjvolzka I think a conversation offline should be had based on the comments. I understand what you mean but @AlexandreEichenberger explained to me his reasoning and how others in the community interpret the opset list. I will schedule some time to talk but I definitely understand Alex's logic after talking with him.

hamptonm1:
> > So what should we put in the max field? Should it be NA then or left blank?
> 
> when we have no max field, we simply don't add the `==Max==` comment. I did not provision for a NA, much simpler to just not have that field.

@AlexandreEichenberger Unfortunately how you wrote the code, if I omit `==Max==` it will default to `18`. So what I can do, if you do not like `==Max==NA` I can simply add `==Max==  ` which will leave that field blank instead.

test/backend/inference_backend.py
hamptonm1: If no change was made to the version it means it still works for 6,7,8,9 and etc. The new  version just means when changes were introduced and the subsequent delta compared to the newer version.

cjvolzka: > because anything lower than opset 6 is not supported. There are several operations that have not changed since opset 1 for instance Softsign. It would not make sense to put 1 here for the minimum given anything under 6 is not supported

On this page, we are documenting specifically the version of the Ops we support. So version 1 would indeed be accurate for Ops like `Conv` and `Softsign` for the reason you pointed out. That doesn't mean we support all of Opset 1 as a whole. Just that we support Version 1 of those particular ops. 

onnx-mlir doesn't have any statements at a model's OpSet level as a whole. Theoretically If you have a model that uses Opset 1, as long as it uses only Ops that we support at V1, then it should work. 

A more likely scenario, just because you have a model that's OpSet 9, it doesn't mean we support it either. We only support it if the model only uses Ops that we document we support. 

> If not, we are frozen at the opset version right BEFORE. The new version just means when changes were introduced and the subsequent delta compared to the newer version.

Correct but again, this is at the individual Op level, not the OpSet as a whole. 

Say we didn't support `Equal` Op version 19. Your model using Opset 19 will work fine with onnx-mlir as long as it doesn't use the Equal Op. If it does use Equal, then you'd have to export the model to a max of 18 (ie OpSet right before the new Equal 19).

Right now we can't accurately make a statement like "OpSet 6 or greater" or "No later than OpSet 18". The only thing we can do accurate document the Ops Versions we do support. From there the user will have to figure out what is the min/max opset they could export their model to and be able to expect it to work with onnx-mlir.

test/accelerators/NNPA/backend/CMakeLists.txt
```diff
@@ -96,12 +96,15 @@ set(NNPA_TEST_LIST
     # ==ADDITIONAL_PARAGRAPH== NNPA has hardware limitations in dimension index size and tensor size, which are described in [NNPALimit.h](../src/Accelerators/NNPA/Conversion/ONNXToZHigh/NNPALimit.h). They are enough large for normal use cases, but if your model exceeds the limitations, CPU is used instead of NNPA.
 
     # ==OP== Add
+    # ==MIN== 14
     # ==LIM== - Shape of input tensors must be the same since broadcasting is not supported.<br>- Input tensors must have static dimensions.
```
tungld, line 100: Just my two cents. Limitation should be per opset if we added supported opsets.

test/backend/inference_backend.py
hamptonm1: Again we need to have a conversation... I do not think you are following. We can talk at the meeting later. Thanks!

test/accelerators/NNPA/backend/CMakeLists.txt
hamptonm1: @tungld Can you give an example of what you mean? Thanks!

hamptonm1:
@cjvolzka @AlexandreEichenberger I made the following updates:
1. In the code specify the min and max from a version standpoint
2. From the documentation view specify the opset number we fully support
3. I added a brief explanation of the use of an `*` and also further explaining to users what this document is for opset vs. version.

I will create an issue so we can make any other comments to change the versions if need be. Thanks!

AlexandreEichenberger:
@hamptonm1 

```
         # ==MIN== 6
         #min=1/max=13
```

Is the `#min=1/max=13` for your own consumption? I could don't find a rule in the file that used it.

hamptonm1:
> @hamptonm1
> 
> ```
>          # ==MIN== 6
>          #min=1/max=13
> ```
> 
> Is the `#min=1/max=13` for your own consumption? I could don't find a rule in the `documentOps.py` file that used it.

@AlexandreEichenberger This is what @cjvolzka asked for... this should not be consumed by `documentOps.py`. More like a reference :)

test/accelerators/NNPA/backend/CMakeLists.txt
tungld: Taking Pad as an example, the document said:
```
| **Pad** |2 |18 |axes input not supported. | |
```

Limitation `axes input not supported.` can be applied to op version 18 but not 13, 11, and 2. It's because there is no `axis` input in versions 13, 11, 2: https://onnx.ai/onnx/operators/onnx__Pad.html#pad-13.

In general, different op versions have different syntax (e.g. the number/order of inputs/attributes) and different semantics. So limitation must be specific to an op version.

test/backend/inference_backend.py
```diff
@@ -77,13 +77,17 @@ def get_test_models():
         # ==ARCH== cpu
 
         # ==OP== Abs
+        # ==MIN== 6
+        #min=1/max=13
```
cjvolzka, line 81: I thought the idea was that here in the "code" we'd have the "true" Op Versions we support. Then in the documents, we show the min model OpSet we support and the and the maximum Opset that can works with the Op (or * if we support latest). 

Also to make things clearer and easier to doc, I'd recommend we rename ==MAX== with something like ==UNSUPPORTED==. (clearer when used with `-1` in doc generator see further below)

So given the `Fake` Op example. Say we support Op V 1 up to Op V 10. Then in OpSet 19 there's a new `Fake` version 19 that we don't support. Here in the "code" that would look like:
```
# ==MIN== 1
# ==UNSUPPORTED== 19
```
(Note: There'd be no UNSUPPORTED line if we support the latest Op Version available in the latest OpSet)

Then in the document generator it would look something like: 
```
min_opset_default="6"
max_opset_default="*"

    if (==MIN==):  // No ==MIN== implies unsupported so no version to show
        opset_min = max(==MIN==, min_opset_default)
        opset_max = ==UNSUPPORTED== -1 if ==UNSUPPORTED== else max_opset_default
```

hamptonm1: @cjvolzka I explained it does not work like the way you are showing above.... what I specify in the python script WILL be generated in the document view. There is ONLY one view. That is why I put in the comments for the python script min=1/max=19 as a reference. 

When I said should we still do the `unsupported` and etc. everyone said no. What we agreed on is the minimum being set to 6 which it is and 18 as the max. For a scenario when the minimum is 18 then the max would be `*` to indicate that we are at the current version. We all agreed that we are supporting things from an opset perspective and not version. Hence why I changed the subheading back to make things clear. We agreed to change it from min/ max version to min/max opset. 

I think it should be one way or the other not both. Either a) we set the minimum to 6 and the max to 18 and use `*` for everything that is already at the max version b) just go with 1 but I think this is becoming overly complicated.......I think simplicity is key here. 

From our conversation with @AlexandreEichenberger he stated "we should not mention support for opsets we do not support". So having 1 does not really make sense which you agreed to.

AlexandreEichenberger:
### Minimums

`==MIN== x`:  we don't support `0` .. `x -1`.  Version x should be fully supported. 

For example, if op FOO 6 had support for type CRAZY, but op FOO 7 removed support for type CRAZY, then let's put `==MIN== 7`. 

However, if op FOO 6 has support for type CRAZY and that type was never deprecated all the way to our current level (say opset 25), and we support everything else from FOO 6, then lets put 
```
==MIN== 6
==LIMIT== type CRACY is unsupported
```

### Maximums/unsupported

I would like to treat `==MAX==` as `==UNSUPORTED==`. Whether we change this in the code or we just define MAX as first Opset that we don't support anymore, I am ok with either. So let me use MAX for the moment being since this is what we have in the code.

`==MAX== x ` we don't support opset x, x+1,... current opset number (say 20). If we don't have a `==MAX==` or we have a `==MAX== *`, this mean that we support the current opset number (say 20) minus any limitation we may have explicitly written down.

For example, op FOO 14 is fully supported and there is no new versions of FOO, then we have no `==MAX==` or a `==MAX== *` (they are the same).

If, however, op FOO 14 is fully supported but op FOO 18 is supported anymore, then we report `==MAX== 18` as 18 is unsupported.

Now if FOO 18 is not fully supported. Say FOO 18 added 2 new types "REASONABLE" and "CRAZY" and that we added support for REASONABLE but not CRAZY. Let's further assume that there are no new opset definition for FOO after 18 (i.e. 18 is the most recent version of this op)
```
==MAX== *
==LIMIT== Type CRAZY is not supported
```

### Observation

The key to this PR is to get the MIN and MAX/UNSUPPORTED values correct. That is the hard part. How we then present this info to the user is easy, as we can simply change a few lines in the documentOps.py to get it in the form that makes the most sense to the users. We may try it one way, ask a few folks if that make sense to them and/or how to change it, and modify the gen of the .md files accordingly

Please let me know if that make sense @hamptonm1 and @cjvolzka

hamptonm1:
> ### Minimums
> `==MIN== x`: we don't support `0` .. `x -1`. Version x should be fully supported.
> 
> For example, if op FOO 6 had support for type CRAZY, but op FOO 7 removed support for type CRAZY, then let's put `==MIN== 7`.
> 
> However, if op FOO 6 has support for type CRAZY and that type was never deprecated all the way to our current level (say opset 25), and we support everything else from FOO 6, then lets put
> 
> ```
> ==MIN== 6
> ==LIMIT== type CRACY is unsupported
> ```
> 
> ### Maximums/unsupported
> I would like to treat `==MAX==` as `==UNSUPORTED==`. Whether we change this in the code or we just define MAX as first Opset that we don't support anymore, I am ok with either. So let me use MAX for the moment being since this is what we have in the code.
> 
> `==MAX== x ` we don't support opset x, x+1,... current opset number (say 20). If we don't have a `==MAX==` or we have a `==MAX== *`, this mean that we support the current opset number (say 20) minus any limitation we may have explicitly written down.
> 
> For example, op FOO 14 is fully supported and there is no new versions of FOO, then we have no `==MAX==` or a `==MAX== *` (they are the same).
> 
> If, however, op FOO 14 is fully supported but op FOO 18 is supported anymore, then we report `==MAX== 18` as 18 is unsupported.
> 
> Now if FOO 18 is not fully supported. Say FOO 18 added 2 new types "REASONABLE" and "CRAZY" and that we added support for REASONABLE but not CRAZY. Let's further assume that there are no new opset definition for FOO after 18 (i.e. 18 is the most recent version of this op)
> 
> ```
> ==MAX== *
> ==LIMIT== Type CRAZY is not supported
> ```
> 
> ### Observation
> The key to this PR is to get the MIN and MAX/UNSUPPORTED values correct. That is the hard part. How we then present this info to the user is easy, as we can simply change a few lines in the documentOps.py to get it in the form that makes the most sense to the users. We may try it one way, ask a few folks if that make sense to them and/or how to change it, and modify the gen of the .md files accordingly
> 
> Please let me know if that make sense @hamptonm1 and @cjvolzka

Okay I am going to clean things up a bit and then we can go from there. Thanks!

PR #2441, opened 2023-08-14
ONNX Op reporting

AlexandreEichenberger:
Report on the success/failure of ONNX Op simdization during the ONNX to KRNL lowering. 

With the `-onnx-op-report=Simd`, the compiler will report data for each ONNX operation being lowered. An example of the printout is here.
```
==ONNX-SIMD-REPORT==, onnx.Sum-simd, n160, 32, 100352, variadic with manageable broadcast
```

It include the stat being reported on, the operation name, the node name, and then some stat specific info, with the last being a string explaining what happened with the optimization.

For SIMD, the first stat specific number is the SIMD vector length, and the second is the simd loop trip count. When the simd trip count is `dim(memref, 3) * 1024`, namely the simd loop was collapsed, with some of it a runtime dimension and some of it a compile time number, we report the compile time number. So it might be several time that number. If only runtime, we report `-1`.

The explanation string is expected to say
* `unsupported` if that pattern is unsupported
* ' no simd in xxx because of yyyy` if the optimization `simd` was deactivated for the operation sub-variant `xxx` because of the reason `yyy`.
* simply an explanation of why the optimization was successful.

In addition, a program `utils/make-report.py` is used to summarize this. info.

```
make-report -[h] -i <file_log> [-d <num>] [-p <op regexp>]

Usage: scan onnx op report generated by `onnx-mlir1 with `--onnx-op-report`
equal to `Simd` or `Parallel` and report statistics about the ops found
in the provided input log.

Parameters:
  -d/--detail <num>:     Print statistics:
                         0: Just count successful/unsuccessful ops.
                         1: Also count reasons for success/failure.
                         2: Also list metrics.
                         3: Also list node name.
  -h/--help:             Print usage.
  -i/--input <file_log>: Input file to be scanned.
  -p/--pattern <regexp>: Focus only on ops that match the regexp pattern.

More info:
  If you need more specific info on individual success/failure, run 
  `onnx-mlir --debug-only=lowering-to-krnl` and look at the compiler output.
  Use `-d 3` to correlate the node name printed here with compiler output.
```

Example of report at level 0 on Resnet:
```
make-report.py -i bibi.log  -d 0   
Statistics:
   onnx.Add : 5
   onnx.Add-simd : 28
   onnx.AveragePool : 1
   onnx.Conv : 20
   onnx.Gemm-simd : 1
   onnx.MatMul-simd : 33
   onnx.MaxPoolSingleOut : 1
   onnx.Relu-simd : 49
   onnx.Reshape : 66
   onnx.Softmax : 1
   onnx.Sum-simd : 16
```

More detailed info at report level 1 list also the distribution of reasons for success/failure
```
alexe@AChampignon resnet50 % make-report.py -i bibi.log  -d 1
Statistics:
   onnx.Add : 5
     *:  no simd in variadic because no beneficial VL
   onnx.Add-simd : 28
     *:  variadic with manageable broadcast
   onnx.AveragePool : 1
     *:  unsupported
   onnx.Conv : 20
     *:  unsupported
   onnx.Gemm-simd : 1
     *:  simd for copied tiles due to j dim
   onnx.MatMul-simd : 33
     *:  simd for mat * mat along j dim
   onnx.MaxPoolSingleOut : 1
     *:  unsupported
   onnx.Relu-simd : 49
     *:  unary fully flattened
   onnx.Reshape : 66
     *:  unsupported
   onnx.Softmax : 1
     *:  unsupported
   onnx.Sum-simd : 16
     *:  variadic with manageable broadcast
```

Stats specific to SIMD is obtained at level = 2
```
alexe@AChampignon resnet50 % make-report.py -i bibi.log  -d 2
Statistic legend:
   num: SIMD vector length (in elements), SIMD loop trip count (-1 is runtime), message 


Statistics:
   onnx.Add : 5
     *:  0, 49, no simd in variadic because no beneficial VL
   onnx.Add-simd : 28
     12 :  28, 196, variadic with manageable broadcast
     8 :  28, 784, variadic with manageable broadcast
     8 :  32, 3136, variadic with manageable broadcast
   onnx.AveragePool : 1
     *:  0, 0, unsupported
   onnx.Conv : 20
     *:  0, 0, unsupported
   onnx.Gemm-simd : 1
     *:  16, 1000, simd for copied tiles due to j dim
   onnx.MatMul-simd : 33
     *:  8, 0, simd for mat * mat along j dim
   onnx.MaxPoolSingleOut : 1
     *:  0, 0, unsupported
   onnx.Relu-simd : 49
     11 :  32, 100352, unary fully flattened
     13 :  32, 200704, unary fully flattened
     5 :  32, 25088, unary fully flattened
     5 :  32, 401408, unary fully flattened
     11 :  32, 50176, unary fully flattened
     4 :  32, 802816, unary fully flattened
   onnx.Reshape : 66
     *:  0, 0, unsupported
   onnx.Softmax : 1
     *:  0, 0, unsupported
   onnx.Sum-simd : 16
     3 :  32, 100352, variadic with manageable broadcast
     6 :  32, 200704, variadic with manageable broadcast
     4 :  32, 401408, variadic with manageable broadcast
     3 :  32, 802816, variadic with manageable broadcast
```

At level 3, we also list the operation node name so that we may directly correlate the observations with the listings produced by `onnx-mlir`

Similar info will be added to track the progress of parallelization in the near future.


AlexandreEichenberger:
@tungld , maybe in a second iteration, we could relate the execution time associated with an op of a given `onnx_node_name` with the stats gathered here.

utils/make-report.py
```diff
@@ -0,0 +1,185 @@
+#!/usr/bin/python3
+
+# SPDX-License-Identifier: Apache-2.0
+
+##################### fixLitTest.py ########################################
```
tungld, line 5: Thanks for this tool! I may call it from `RunONNXModel.py` sometimes in the future to show the report.

tungld (approved):
LGTM! Thank you so much for providing the report!

Just for future improvement, I am thinking whether we can do something similar to `LLVM_DEBUG` (e.g. `OM_REPORT`) instead of calling a function `onnxToKrnlSimdReport`, so that we can report in different passes/dialects with a single OM_REPORT, and enable it by `-om-report=lowering-to-krnl`.

utils/make-report.py
AlexandreEichenberger: Tx

robben225:
@AlexandreEichenberger  I find some ops(e.g. onnx.Conv) shows "unsupported" in the report of simdization during the ONNX to KRNL lowering, will these ops support simd in the future?

AlexandreEichenberger:
@robben225 yes, this is a current weakness, we are not optimizing the gen of Conv at this time. We can chat if you are interested in adding support.

PR #2442, opened 2023-08-15
[Do-not-merge][NNPA] Donot check sequence_lens in lowering ONNX to ZHigh

tungld:
This is just for testing lowering GRU to ZHigh without checking `sequence_lens`


tungld:
FYI @cjvolzka

AlexandreEichenberger:
Is there a way to determine if/when this is legal? Maybe if the size difference still fit in the 4K lower dim padded memory?

tungld:
closing this since #2479 and #2501 can handle sequence_lens.

PR #2444, opened 2023-08-15
Fix for reduction simdization in presence of constant dimensions

AlexandreEichenberger:
There was an issue with "unrequested" blocked inner loops being generated (when the control flow was simplified to fully nested) while lowering krnl to affine dialect.

The fix is relatively simple. I detect when the inner loop of a blocked loop is unused (aka its loop def is not used by any krnl operation). When that is the case, we still use the affine support to block the loop. The code simply forces the inner loop (which is not requested) to become a trivial `for(I=0; I<1; I++)` loop.

I also added a seed to generate different test case for `RunONNXModel.py` as it is useful to detect if several versions of a code always fail in the same way because of structural issues, or simply because they use the same seed.

I also added a few lit tests that cover this new issue, both to make sure the correct code pattern is generated when lowering the onnx operations to krnl, as well as when lowering the krnl to affine.

### DIFF
--- a/src/Conversion/KrnlToAffine/ConvertKrnlToAffine.cpp
+++ b/src/Conversion/KrnlToAffine/ConvertKrnlToAffine.cpp
@@ -535,11 +535,21 @@ static LogicalResult interpretOperation(Operation *op, OpBuilder &builder,
     SmallVector<AffineForOp, 1> loopsToTile = {
         llvm::cast<AffineForOp>(loopRefToOp[blockOp.getLoop()])};
 
-    if (failed(tilePerfectlyNested(
-            loopsToTile, blockOp.getTileSizeAttr().getInt(), &tiledLoops))) {
+    int64_t step = blockOp.getTileSizeAttr().getInt();
+    if (failed(tilePerfectlyNested(loopsToTile, step, &tiledLoops))) {
       return failure();
     }
 
+    if (blockOp.getResult(1).use_empty()) {
+      LLVM_DEBUG({
+        llvm::dbgs() << DEBUG_TYPE << " inner block loop unused, trivialize\n";
+        tiledLoops[1].dump();
+      });
+      tiledLoops[1].setConstantLowerBound(0);
+      tiledLoops[1].setConstantUpperBound(1);
+      tiledLoops[1].setStep(1);
+      LLVM_DEBUG(tiledLoops[1].dump());
+    }
     assert(tiledLoops.size() == 2);
     assert(blockOp.getNumResults() == 2);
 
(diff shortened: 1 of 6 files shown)


src/Conversion/ONNXToKrnl/Math/Reduction.cpp
```diff
@@ -18,7 +18,7 @@
 #include "src/Dialect/ONNX/ONNXOps/ShapeHelper.hpp"
 
 #define DEBUG_TYPE "lowering-to-krnl"
-#define DEBUG_FORCE_SHUFFLE_REDUCTION 1
+#define DEBUG_FORCE_SHUFFLE_REDUCTION 0
```
AlexandreEichenberger, line 21: This debug was accidentally left on.

utils/CheckONNXModel.py
```diff
@@ -80,34 +80,38 @@ def valid_onnx_input(fname):
...
 parser.add_argument('-l',
                      '--log-level',
                      choices=[ 'debug', 'info', 'warning', 'error', 'critical' ],
                      default='info',
-                     help="log level, default info")
+                     help="log level, default info.")
+parser.add_argument('--seed',
+                    type=str,
+                    default="42",
+                    help="seed to initialize the random num generator for inputs.")
```
tungld, line 114: Thanks for adding this!

src/Conversion/KrnlToAffine/ConvertKrnlToAffine.cpp
```diff
@@ -535,11 +535,21 @@ static LogicalResult interpretOperation(Operation *op, OpBuilder &builder,
...
         llvm::cast<AffineForOp>(loopRefToOp[blockOp.getLoop()])};
 
-    if (failed(tilePerfectlyNested(
-            loopsToTile, blockOp.getTileSizeAttr().getInt(), &tiledLoops))) {
+    int64_t step = blockOp.getTileSizeAttr().getInt();
+    if (failed(tilePerfectlyNested(loopsToTile, step, &tiledLoops))) {
       return failure();
     }
 
+    if (blockOp.getResult(1).use_empty()) {
```
tungld, line 543: Just for my learning. Is this happened because of manually unrolling when generating krnl iterations?

tungld (approved):
LGTM.

cjvolzka (approved):

src/Conversion/KrnlToAffine/ConvertKrnlToAffine.cpp
tungld: Got it. Thanks!

PR #2446, opened 2023-08-16
Call HybridTransformPass one more time to update shape if possible

### LINKED ISSUE #2443: Failure when running Roberta-base-11 model with dimension 1x1
Previously, we had an issues when compiling the Roberta model and the "work-around" was to set the pattern-shape-inference to false. However, the functionality has recently changed for pattern-shape-inference and it makes sense to get to the root of the problem. The new error I am seeing is as follows:

```
meganhampton@Megans-MBP-2 build % ./Debug/bin/onnx-mlir --O3 --EmitLib --maccel=NNPA roberta-base-11.onnx --onnx-op-stats TXT --shapeInformation 0:1x1 -v
The ONNX model has 124645716 elements in its initializers. This value would be close to and greater than the number of parameters in the model. Because there is no way to exactly count the number of parameters, this value can be used to have a rough idea of the number of parameters in the model.
Operations encountered:
-----------------------
   func.func          , 1
   func.return        , 1
   onnx.Add           , 58
   onnx.Cast          , 2
   onnx.Concat        , 77
   onnx.Constant      , 224
   onnx.CumSum        , 1
   onnx.Dim           , 116
   onnx.Div           , 41
   onnx.Equal         , 1
   onnx.Erf           , 12
   onnx.Gather        , 3
   onnx.Max           , 8
   onnx.Mul           , 32
   onnx.NoValue       , 1
   onnx.Not           , 1
   onnx.ReduceMeanV13 , 50
   onnx.Reshape       , 139
   onnx.Sqrt          , 25
   onnx.Sub           , 25
   onnx.Transpose     , 48
  zhigh.Add           , 73
  zhigh.Div           , 8
  zhigh.MatMul        , 97
  zhigh.Mul           , 43
  zhigh.Softmax       , 12
  zhigh.Stick         , 317
  zhigh.Tanh          , 1
  zhigh.Unstick       , 149
Assertion failed: (sizeInterface.isDynamicSize(unsignedIndex) && "Expected dynamic subview size"), function fold, file MemRefOps.cpp, line 1106.
zsh: abort      ./Debug/bin/onnx-mlir --O3 --EmitLib --maccel=NNPA roberta-base-11.onnx  TXT 
```

It seems like for dimensions that are `1x?` are all failing with this error (1x1, 1x2, 1x3 and so forth). Yet, if I use a `2x2` or anything else it works just fine.

tungld:
fixes #2443 

For the issue in #2443, after calling all onnx transformations, there is still the following code (extracted from the roberta model and put into a single test):
```mlir
func.func @test(%arg0 : tensor<12x5x5xf32>, %arg1: tensor<f32>) -> tensor<?x12x5x5xf32>{
  %35 = onnx.Constant dense<8.000000e+00> : tensor<f32>
  %0 = onnx.Constant dense<[1, 12, 5, 5]> : tensor<4xi64>
  %641 = "onnx.Reshape"(%arg0, %0) {allowzero = 0 : si64} : (tensor<12x5x5xf32>, tensor<4xi64>) -> tensor<?x12x5x5xf32>
  %1 = "onnx.Div"(%641, %35) {onnx_node_name = "Div_560"} : (tensor<?x12x5x5xf32>, tensor<f32>) -> tensor<?x12x5x5xf32>
  return %1 : tensor<?x12x5x5xf32>
}
```

where Reshape's output is still dynamic, but it could be static. So when lowering ONNXReshape to `ReinterpretCastOp`, `ReinterpretCastOp` complained:
```
onnx-mlir: /workdir/llvm-project/mlir/lib/Dialect/MemRef/IR/MemRefOps.cpp:1107: mlir::OpFoldResult mlir::memref::DimOp::fold(mlir::memref::DimOp::FoldAdaptor): Assertion `sizeInterface.isDynamicSize(unsignedIndex) && "Expected dynamic subview size"' failed.
```

This patch calls `createONNXHybridTransformPass` one more time so that the Reshape's output shape is updated to static ones.

### DIFF
--- a/src/Compiler/CompilerPasses.cpp
+++ b/src/Compiler/CompilerPasses.cpp
@@ -103,6 +103,18 @@ void addONNXToMLIRPasses(mlir::PassManager &pm, bool targetCPU) {
   // Simplify shape-related ops.
   pm.addPass(onnx_mlir::createSimplifyShapeRelatedOpsPass());
 
+  // One more call to ONNX shape inference/canonicalization/... to update shape
+  // if possible.
+  if (enableONNXHybridPass) {
+    // For starters only illustrating the new hybrid pass by replacing 3 passes
+    // here. The plan is to replace most of the passes in addONNXToMLIRPasses.
+    pm.addNestedPass<func::FuncOp>(onnx_mlir::createONNXHybridTransformPass());
+  } else {
+    pm.addNestedPass<func::FuncOp>(onnx_mlir::createShapeInferencePass());
+    pm.addPass(mlir::createCanonicalizerPass());
+    pm.addNestedPass<func::FuncOp>(onnx_mlir::createShapeInferencePass());
+  }
+
   // Replace ONNXReturnOp with func::ReturnOp.
   pm.addPass(onnx_mlir::createStandardFuncReturnPass());
 


sorenlassen (approved):
LGTM

thank you for solving this issue

hamptonm1:
@tungld Thanks for looking into this!

PR #2452, opened 2023-08-19
Update CPU and NNPA docs with min and max supported opsets 

cjvolzka:
# Description
* Refactor documentation to show supported model opset range for each individual operator.
* Add DeformConv to unsupported operator list (looks like we missed it on going to Opset 19)

# "Summary" / Novella
There are now two "views" of Op versions. The "code" view (ie .py and CMakeLists.txt) uses `==Min==` and `==UNSUPPORTED==` values. They are set based on the versions of the operator we know work (ie an operator centric view). 

The "documentation" view (ie *.md files) is Opset centric. The values presented there are based on model opsets and attempt to mask the complexity of how ONNX versions individual Operators. (ie Opset 19 uses the last updated version of Add, which is currently 14. So while it is technically accurate to say we only support "up to" Add 14, users may interpret that to mean their model's Opset must be less then or equal to <= 14, which isn't true.

For this PR, I set the Ops' `==MIN==` in the code files based on passing models from the onnx-model zoo. For example say a model is based on Opset 6, has Operator X, compiles, and successfully runs with onnx-mlir. Further say Operator X has versions 1, 2, 7, 10 and 19. Since the Opset 6 model works and uses X, then what we have in onnx-mlir is likely minimally compatible with X version 2 (ie 2 is the latest version of X that's <= the model's opset 6). This is not a perfect test but it is a good launching point and lines up with the models we show as supported in our model-zoo testing.

In the code view, `==UNSUPPORTED==` (if set), indicates an Op we support an earlier version of, but have not updated for a particular Opset. For example, if supported Op X was updated in Opset 19, but onnx-mlir has not been updated to handle X 19, then X's `==UNSUPORTED==` will be set to 19. If we support the latest version of the Op available, `==UNSUPPORTED==` is not set.

Since users and models are generally Opset focused, we make our documentation Opset centric. As a general statement, we only support models with a minimum Opset of 6. So continuing the example of Operator X, it has a `==MIN==` of 2 and `==UNSUPPORTED` of 19. Then our documentation (ie support statement) is that we only support models that contain Operator X if the model's Opset is between 6 and 18 (inclusvie). 

The documenation's Opset minimum for each Op is calculated as `max( Op's ==MIN==, 6)`. The Opset maximum for each Op is the Op's` ==UNSUPPORTED==` - 1 or `*`. Where `*` indicates we support the latest version available at onnx-mlir release. (The latest Opset available at release is specified at the top of the document view).


utils/documentOps.py
```diff
@@ -157,18 +154,23 @@ def parse_file(file_name):
...
         # Max release supported.
-        p = re.search(r'==MAX==\s+(\d+)\s*$', l)
+        p = re.search(r'==UNSUPPORTED==\s+(.*)\s*$', l)
         if p is not None:
-            assert op is not None, "Max without op."
-            assert op not in max_dict, "Redefinition of max for op " + op
-            max_dict[op] = int(p[1])
+            assert op is not None, "Unsupported without op."
+            assert op in min_dict and min_dict[op], "Unsupported without min for op " + op
+            assert max_dict[op] == max_opset_default, "Redefinition of Unsupported for op " + op
```
AlexandreEichenberger, line 169: This support is getting a bit funky, as you have a different behavior if no min and unsupported is given.

Maybe cleaner would be to set min/max by default when op is defined (processing of `==OP==`.
That way, as soon as an op is defined, its min and max are defined, and then we may redefine them by having `MIN` or `UNSUPPORTED`
 
For completion, you may want to check when setting either min or max that `min<=max`

utils/documentOps.py
```diff
@@ -208,40 +190,39 @@ def print_md():
...
+
     print("\n")
     # Additional top paragraph.
     if additional_top_paragraph:
         print(additional_top_paragraph)
         print("\n")
     # Table.
-    header = ["Op", "Min Opset", "Max Opset", "Limitations"]
-    separator = ["---", "---", "---", "---"]
+    header = ["Op", "Supported Opsets", "Limitations"]
```
AlexandreEichenberger, line 209: Could add "inclusive".. though its probably understood

utils/pre-onnx-mlir.py
```diff
@@ -44,7 +44,7 @@
 
 # Opset version supported by current onnx-mlir
 # Should be consistent with gen_onnx_mlir.py
-current_onnx_mlir_support_version=16
+current_onnx_mlir_support_version=19
```
AlexandreEichenberger, line 47: Could you add a stylized comments maybe such as `# ==UPDATE_ALL_VERSION_NUMBER== by scanning for this pattern` in every file where "19" appears (meaning 19 as the current version number) if it shows up in more than one place?

That will make is

AlexandreEichenberger (commented):
Just looked at this time at the code, not the actual values of min and unsupported

cjvolzka (commented):
Note to self: update Op mins based on recently landed https://github.com/onnx/onnx-mlir/pull/2219

utils/documentOps.py
cjvolzka: > This support is getting a bit funky, as you have a different behavior if no min and unsupported is given.

That was by design. Based on how it's used, `==UNSUPPORTED==` only makes sense if there's a `==MIN==`. We could see about renaming UNSUPPORTED to something that's more obvious that it shouldn't exist without a MIN (UNSUPPORTED_AT_VER?). However the end result is still the same; that key doesn't make sense without a MIN.

It's not dissimiar to how `==MIN==` is handled. It is an error if there's a `==MIN==` without an `==OP==` before it.

> Maybe cleaner would be to set min/max by default when op is defined.

I already do this for the max with `max_dict[op] = max_opset_default`. By default it'll get a `*` as the max by default. If an ==UNSUPPORTED== is specified, then the default max is overridden.

Prior to my changes, if there was an `==OP==` it would show as unsupported unless a `==MIN==` was specified. I assumed this was intentional to force a dev to specify lowest `==MIN==` their addition supported. So the error I added was to help catch a semantic error (ie `==OP==` with `==UNSUPPORTED==` but no `==MIN==`) which otherwise would just doc the Op as unsupported.

> For completion, you may want to check when setting either min or max that min<=max

Good idea. I'll include this in the next update.

utils/documentOps.py
cjvolzka: Better to be explicit. I'll add it.

utils/pre-onnx-mlir.py
cjvolzka: There's only three that I see. This file, gen_onnx_mlir.py and in OpBuildTable.inc. In gen_onnx_mlir.py, it's coding the latest updated version for each individual Op so using this variable doesn't make sense. OpBuildTable.inc also has `19` in reference to an Opset but that has "Do not modify this file directly. ... This file is automatically generated via script." so also not applicable.

In the spirit of the comment, pre-onnx-mlir.py has `current_onnx_version = "1.14.0"`. I'll see about coming up with some sort of comment tag group these together.

utils/documentOps.py
```diff
@@ -41,17 +41,17 @@
...
 # ==TODO== <text>
-#   where <text> add "private" info about what needs to be fixed. 
+#   where <text> add "private" info about what needs to be fixed.
 #
 # ==MIN== <num>
 #   where <num> is the minimum release version supported.
 #
-# ==MAX== <num>
-#   where <num> is the maximum release version supported.
+# ==UNSUPPORTED== <num>
```
AlexandreEichenberger, line 53: An assumption of the code is that we must see MIN before UNSUPPORTED. It could be removed (by putting the init code in the OP which by def needs to come first).

I am ok as is, just indicate that MIN must be precede UNSUPPORTED.

utils/documentOps.py
```diff
@@ -157,18 +154,24 @@ def parse_file(file_name):
         if p is not None:
             assert op is not None, "Min without op."
             assert op not in min_dict, "Redefinition of min for op " + op
-            min_dict[op] = int(p[1])
+            op_min = max(int(p[1]), min_opset_default)
```
AlexandreEichenberger, line 157: Note: this approach requires MIN to come before UNSUPPORTED, which is not necessary but ok.

utils/gen_onnx_mlir.py
```diff
@@ -56,7 +56,8 @@
 check_operation_version = args.check_operation_version
 list_operation_version = args.list_operation_version
 
-# Change this variable only when upgrading the ONNX support within ONNX-MLIR.
+# ==UPDATE_ONNX_VERSION_OPSET==
```
AlexandreEichenberger, line 59: Thanks

AlexandreEichenberger (approved):
LGTM, requesting an additional comment to clarify the respective order of the MIN and UNSUPPORTED annotation

utils/documentOps.py
cjvolzka: Murphy's law says that if I even try a comment update, the build system will break because it knows I'm leaving for vacation in a few hours.

I created https://github.com/onnx/onnx-mlir/issues/2460 for myself to cover this when I get back ;-)

PR #2456, opened 2023-08-21
flag -onnx-const-prop-disable-pattern

sorenlassen:
Added a flag to disable select onnx const prop patterns. This upstreams logic that proved useful at Groq on a specific model where an onnx const prop pattern was counter-productive.

Removed ConstPropSplitV11Pattern because it should be unnecessary since we canonicalize SplitV11 to Split.

Migrated ConstPropScatterNDPattern to tablegen. Couldn't migrate ConstPropSplitPattern to tablegen because it has a variadic return value but refactored it to follow the style of the tablegen const prop patterns.

For the record, I modified the tablegen file with this script:
```
sed -i 's/^def \(\w\w*\) : Pat<$/def \1 : NamedPat<"\1",/' ConstProp.td
```

### DIFF
--- a/src/Compiler/CompilerOptions.cpp
+++ b/src/Compiler/CompilerOptions.cpp
@@ -24,59 +24,60 @@
 namespace onnx_mlir {
 
 // Use external storage for the options so that they are globally accessible
-std::string inputFilename;                     // common for both
-std::string outputBaseName;                    // common for both
-std::vector<accel::Accelerator::Kind> maccel;  // common for both
-OptLevel OptimizationLevel;                    // common for both
-std::string mtriple;                           // common for both
-std::string mcpu;                              // common for both
-std::string march;                             // common for both
-InstrumentStages instrumentStage;              // common for both
-int onnxConstPropExpansionBound;               // common for both
-bool enableONNXHybridPass;                     // common for both
-std::vector<std::string> functionsToDecompose; // common for both
-EmissionTargetType emissionTarget;             // onnx-mlir only
-bool invokeOnnxVersionConverter;               // onnx-mlir only
-bool preserveLocations;                        // onnx-mlir only
-bool printIR;                                  // onnx-mlir only
-bool preserveBitcode;                          // onnx-mlir only
-bool preserveLLVMIR;                           // onnx-mlir only
-bool preserveMLIR;                             // onnx-mlir only
-bool useOnnxModelTypes;                        // onnx-mlir only
-int repeatOnnxTransform;                       // onnx-mlir only
-std::string shapeInformation;                  // onnx-mlir only
-ModelSize modelSize;                           // onnx-mlir only
-bool storeConstantsToFile;                     // onnx-mlir only
-float constantsToFileTotalThreshold;           // onnx-mlir only
-float constantsToFileSingleThreshold;          // onnx-mlir only
-bool VerboseOutput;                            // onnx-mlir only
-std::vector<std::string> Xopt;                 // onnx-mlir only
-std::vector<std::string> Xllc;                 // onnx-mlir only
-std::string mllvm;                             // onnx-mlir only
-std::string instrumentOps;                     // onnx-mlir only
-unsigned instrumentControlBits;                // onnx-mlir only
-bool instrumentONNXSignature;                  // onnx-mlir only
-std::string ONNXOpStats;                       // onnx-mlir only
-bool enableMemoryBundling;                     // onnx-mlir only
-int onnxOpTransformThreshold;                  // onnx-mlir only
-bool onnxOpTransformReport;                    // onnx-mlir only
-bool enableParallel;                           // onnx-mlir only
-bool disableSimdOption;                        // onnx-mlir only
-bool enableSimdDataLayout;                     // onnx-mlir only
-bool verifyInputTensors;                       // onnx-mlir only
-bool allowSorting;                             // onnx-mlir only
-std::string reportHeapBefore;                  // onnx-mlir only
-std::string reportHeapAfter;                   // onnx-mlir only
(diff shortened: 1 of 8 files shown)


src/Compiler/CompilerOptions.cpp
```diff
@@ -23,59 +23,66 @@
...
+std::string inputFilename;
+std::string outputBaseName;
+std::vector<accel::Accelerator::Kind> maccel;
+OptLevel OptimizationLevel;
+std::string mtriple;
+std::string mcpu;
+std::string march;
+InstrumentStages instrumentStage;
+int onnxConstPropExpansionBound;
+std::vector<std::string> onnxConstPropDisablePatterns;
```
sorenlassen, line 37: @gongsu832 fyi, I refactored the comments because the old style got in the way of this long line

gongsu832: @sorenlassen if you add the line
```
std::vector<std::string> onnxConstPropDisablePatterns; // common for both
```
then run `clang-format-9`, it will indent all the lines properly. I prefer the old style because when the list don't fit in my window I don't have to scroll back to remember which one of the 3 sections I'm currently in.

sorenlassen: ok, will do

just beware that clang-format will do something funny to avoid exceeding the line length limit (and it may break down further in the future if people add more long lines)

sorenlassen: done

gongsu832: > just beware that clang-format will do something funny to avoid exceeding the line length limit (and it may break down further in the future if people add more long lines)

When that happens you know your option name is too long and needs to be shortened. :-)

AlexandreEichenberger (commented):
Wild idea. Current code

```
  RewritePatternSet patterns(context);
  populateWithGenerated(patterns);
  patterns.insert<ConstPropSplitPattern>(&getContext());
  patterns.insert<ConstPropSplitV11Pattern>(&getContext());
  patterns.insert<ConstPropScatterNDPattern>(&getContext());
  if (failed(applyPatternsAndFoldGreedily(function, std::move(patterns))))
    signalPassFailure();
```

Can we build the normal pattern as is, then having a "GatedPattern" subclass which would extract the name (saw you added a name) and a dict of accepted name, selectively add them to the new pattern subclass, and then feed this one to the method that uses them?

May be naive, just a thought.

src/Transform/ONNX/ConstProp.cpp
```diff
@@ -1035,6 +945,74 @@ Value ConstPropNonZero(
...
+  }
+  return resValues;
+}
+
+class SplitOfConst : public OpRewritePattern<ONNXSplitOp> {
+public:
+  using OpRewritePattern<ONNXSplitOp>::OpRewritePattern;
+
+  LogicalResult match(ONNXSplitOp splitOp) const override {
+    if (!isNotDisabled("SplitOfConst"))
```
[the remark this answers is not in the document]
sorenlassen: good idea

done

src/Transform/ONNX/ConstProp.cpp
```diff
@@ -1059,17 +1037,18 @@ void ConstPropONNXToONNXPass::runOnOperation() {
 
   RewritePatternSet patterns(context);
   populateWithGenerated(patterns);
-  patterns.insert<ConstPropSplitPattern>(&getContext());
-  patterns.insert<ConstPropSplitV11Pattern>(&getContext());
-  patterns.insert<ConstPropScatterNDPattern>(&getContext());
+  patterns.insert<SplitOfConst>(context);
```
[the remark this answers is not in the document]
sorenlassen: that's a good idea if we have more patterns in the future, but there's only one non-tablegen pattern now so probably not worthwhile yet

sorenlassen:
> Can we build the normal pattern as is, then having a "GatedPattern" subclass which would extract the name (saw you added a name) and a dict of accepted name, selectively add them to the new pattern subclass, and then feed this one to the method that uses them?

I couldn't figure out how to do that with tablegen.

Alternatively, we push a change to mlir, e.g. make tablegen add the pattern name as a debug name or debug label. I checked the tablegen code and it looks pretty straightforward, e.g. in the generated populateWithGenerated() we could change 
```
patterns.add<AddConstAssociative1>(patterns.getContext());
```
to
```
patterns.addWithLabel<AddConstAssociative1>("AddConstAssociative1", patterns.getContext());
```
See: https://github.com/llvm/llvm-project/blob/main/mlir/tools/mlir-tblgen/RewriterGen.cpp#L1820

Connor-XY:
https://github.com/onnx/onnx-mlir/blob/667f0d947e4befa8a9e38b214f69088553083b3f/src/Dialect/ONNX/Rewrite.cpp#L874-L880

Actually, when I tried to filter out those patterns, I found that their debug names are already set to their type name, but I didn't really find out where that was done.

sorenlassen:
> I found that their debug names are already set to their type name, but I didn't really find out where that was done.

nice

I gather it's set here:
https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/IR/PatternMatch.h#L282
where llvm does some interesting reflection to come up with a type's name

it comes with [this disclaimer](https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/Support/TypeName.h#L19-L21):
```
/// This routine may fail on some platforms or for particularly unusual types.
/// Do not use it for anything other than logging and debugging aids. It isn't
/// portable or dependendable in any real sense.
```
but if it works then great!

@AlexandreEichenberger please take a look and let me know if you prefer I change the implementation to make use of the debug name that mlir/llvm set

AlexandreEichenberger (approved):
LGTM, thanks for the changes

PR #2457, opened 2023-08-21
Remove redundant optlevel option in numerical tests, it already defaults to 3

gongsu832:


AlexandreEichenberger:
How do you know it default to `-O3`? I was of the general impression that default is `-O0`. That is the case for llc I believe. We need to make sure that we compile these benchmarks with -O3 all the way to the backend.

gongsu832:
In `test/numerical/CMakeLists.txt`:
```
function(add_numerical_test test_name)
  add_onnx_mlir_executable(${test_name} NO_INSTALL ${ARGN})

  add_dependencies(numerical ${test_name})
  get_target_property(test_suite_folder numerical FOLDER)
  if (test_suite_folder)
    set_property(TARGET ${test_name} PROPERTY FOLDER "${test_suite_folder}")
  endif ()

  # Optimization level set by ONNX_MLIR_TEST_OPTLEVEL, defaults to 3
  add_test(NAME ${test_name}
    COMMAND ${test_name} -O${ONNX_MLIR_TEST_OPTLEVEL}
    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
    )
  set_tests_properties(${test_name} PROPERTIES LABELS numerical)
endfunction()
```
In `test/CMakeLists.txt`:
```
set(OPTLEVELS 0 1 2 3)
if (NOT DEFINED ONNX_MLIR_TEST_OPTLEVEL)
  set(ONNX_MLIR_TEST_OPTLEVEL 3 CACHE STRING "" FORCE)
endif()
```
dev image builds the tests with `-O0` and user image builds the tests with `-O3` so we cover both cases.

AlexandreEichenberger (approved):
LGTM, thanks for sharing the "proof" that we are compiling the bench at `-O3`.

Ideally, we will add the `-march=x86-64`, '-mcpu=z14`,... on the respective machines to enable the SIMD at -O3 too. Maybe another PR

PR #2459, opened 2023-08-22
[NFC] Move lit tests for onnx-to-krnl lowering into a separate folder

tungld:
This patch moves lit tests for onnx-to-krnl lowering into a separate folder, then organizes lit tests by operation categories instead of putting all lit tests into a single file.

In particular, this patch creates a folder `onnx-to-krnl` in the folder `conversion` to make it consistent with other `onnx-to-*` folders
```bash
$ ls -1 test/mlir/conversion/
instrument
krnl_to_affine
krnl_to_llvm
onnx_to_krnl
onnx_to_mhlo
onnx_to_tosa
```

The original `onnx_lowering_O3.mlir` contains no SIMD test and is basically duplicated from `onnx_lowering.mlir`, so `onnx_lowering_O3.mlir` is deleted in this patch.


AlexandreEichenberger (approved):
LGTM

PR #2461, opened 2023-08-23
associative const prop rewrite only 1-use subexpressions

sorenlassen:
The new lit test
```
func.func @test_add_const_associative2_2uses(%x: tensor<i32>, %y: tensor<i32>, %z: tensor<i32>) -> (tensor<i32>, tensor<i32>) {
  %c = onnx.Constant dense<1> : tensor<i32>
  %1 = "onnx.Add"(%x, %c) : (tensor<i32> , tensor<i32>) -> tensor<i32>
  %2 = "onnx.Add"(%1, %y) : (tensor<i32> , tensor<i32>) -> tensor<i32>
  %3 = "onnx.Add"(%1, %z) : (tensor<i32> , tensor<i32>) -> tensor<i32>
  onnx.Return %2, %3 : tensor<i32>, tensor<i32>
```
now is no longer rewritten by AddConstAssociative2 to
```
func.func @test_add_const_associative2(%arg0: tensor<i32>, %arg1: tensor<i32>, %arg2: tensor<i32>) -> (tensor<i32>, tensor<i32>) {
  %0 = onnx.Constant dense<1> : tensor<i32>
  %1 = "onnx.Add"(%arg0, %arg1) : (tensor<i32>, tensor<i32>) -> tensor<i32>
  %2 = "onnx.Add"(%1, %0) : (tensor<i32>, tensor<i32>) -> tensor<i32>
  %3 = "onnx.Add"(%arg0, %arg2) : (tensor<i32>, tensor<i32>) -> tensor<i32>
  %4 = "onnx.Add"(%3, %0) : (tensor<i32>, tensor<i32>) -> tensor<i32>
  onnx.Return %2, %4 : tensor<i32>, tensor<i32>
}
```
which lost the benefits of reusing `%1`.

I tested the change on a bunch of models we use for regression testing at Groq and it affected many of them, so evidently this condition triggers in practice, but the changes in speed and data+code footprint were minimal and mostly positive.


test/mlir/onnx/onnx_constprop.mlir
```diff
@@ -103,8 +103,9 @@ func.func @test_add_constant_4(%arg0 : tensor<3xi32>) -> tensor<3xi32> {
   %1 = onnx.Constant dense<[10, 11, 12]> : tensor<3xi32>
   %2 = "onnx.Add"(%0, %arg0) : (tensor<3xi32> , tensor<3xi32>) -> tensor<3xi32>
   %3 = "onnx.Add"(%1, %2) : (tensor<3xi32> , tensor<3xi32>) -> tensor<3xi32>
-  %4 = "onnx.Add"(%2, %3) : (tensor<3xi32> , tensor<3xi32>) -> tensor<3xi32>
```
sorenlassen, line 106: this existing lit test had two uses of `%2` and therefore no longer triggers const prop with the new 1-use constraint

I edited the lit test to remove the dual uses and added a new lit test @test_add_const_associative2_2uses below to specifically test the dual use case

philass (approved):
Nice! This was a really good catch.

PR #2462, opened 2023-08-23
Bump to LLVM green commit from 2023-08-21

Connor-XY:
Bump onnx-mlir to the latest green commit of Aug 21, 2023

Green LLVM commit: 91088978d712cd7b33610c59f69d87d5a39e3113
Green MHLO commit: 01402e75f8e8a95aa2cd9c267ce929095a2f2e54 (branch greencommit/2023-08-21-91088978)

* [mlir] add supplementalPatterns in Pattern
* [mlir] type converter made const
* [tosa] operation printing syntax prettification
* [tosa] changes axis attribute type from I64Attr to I32Attr

### DIFF
--- a/docs/BuildOnLinuxOSX.md
+++ b/docs/BuildOnLinuxOSX.md
@@ -15,7 +15,7 @@ Firstly, install MLIR (as a part of LLVM-Project):
 ``` bash
 git clone -n https://github.com/llvm/llvm-project.git
 # Check out a specific branch that is known to work with ONNX-MLIR.
-cd llvm-project && git checkout f580901d5d30e37755212f1c09e5b587587fbfeb && cd ..
+cd llvm-project && git checkout 91088978d712cd7b33610c59f69d87d5a39e3113 && cd ..
 ```
 
 [same-as-file]: <> (utils/build-mlir.sh)
--- a/docs/BuildOnWindows.md
+++ b/docs/BuildOnWindows.md
@@ -52,7 +52,7 @@ Install MLIR (as a part of LLVM-Project):
 ```shell
 git clone -n https://github.com/llvm/llvm-project.git
 # Check out a specific branch that is known to work with ONNX-MLIR.
-cd llvm-project && git checkout f580901d5d30e37755212f1c09e5b587587fbfeb && cd ..
+cd llvm-project && git checkout 91088978d712cd7b33610c59f69d87d5a39e3113 && cd ..
 ```
 
 [same-as-file]: <> (utils/build-mlir.cmd)
--- a/docs/ConstPropagationPass.md
+++ b/docs/ConstPropagationPass.md
@@ -73,6 +73,7 @@ class Pattern<
    dag sourcePattern,
    list<dag> resultPatterns,
    list<dag> additionalConstraints = [],
+   list<dag> supplementalPatterns = [],
    dag benefitsAdded = (addBenefit 0)
 >;
 ```
--- a/src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.td
+++ b/src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.td
@@ -25,6 +25,7 @@ include "src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHighCommon.td"
 /// class Pattern<
 ///    dag sourcePattern, list<dag> resultPatterns,
 ///    list<dag> additionalConstraints = [],
+///    list<dag> supplementalPatterns = [],
 ///    dag benefitsAdded = (addBenefit 0)
 /// >;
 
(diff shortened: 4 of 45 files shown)


sorenlassen:

sorenlassen:
> @jenkins-droid test this please

@gongsu832 do you know how to get the jenkins jobs to run on this PR? I gather @Connor-XY lacks sufficient committer privileges to trigger jenkins and my attempt to invoke it didn't work (I think I've experienced the same in the past, I don't really understand the mechanisms involved)

philass:
> @jenkins-droid test this please

This hasn't worked for me for a bit.

@sorenlassen What we can do to unblock this PR is to merge one of your open PRs. And then @slassen or I can merge main into this branch. That should trigger the Jenkins builds with sufficient auth permissions.

gongsu832:

gongsu832:
> > @jenkins-droid test this please
> 
> This hasn't worked for me for a bit.
> 
> @sorenlassen What we can do to unblock this PR is to merge one of your open PRs. And then @slassen or I can merge main into this branch. That should trigger the Jenkins builds with sufficient auth permissions.

This is the correct way to manually trigger the Jenkins PR. Both @sorenlassen and @philass are on the whitelist so you should be able to trigger the build. I just tried it myself and it works. I don't know why it didn't work when you tried. It should work for both of you.

sorenlassen:
> This is the correct way to manually trigger the Jenkins PR build. Both @sorenlassen and @philass are on the whitelist so you should be able to trigger the build. I just tried it myself and it works. I don't know why it didn't work when you tried. It should work for both of you.

@gongsu832 thanks for unblocking this PR.

Not important but, in case you're curious in digging deeper, I think there's some kind of pattern to jenkins not listening to Philip and me, as I've experienced this problem once before with a previous bump-llvm PR. Maybe the pattern is that if the PR is authored by a non-committer then jenkins won't listen to me because I lack some kind of elevated admin privileges? In other situations (maybe when the PR is authored by a committer) `@jenkins-droid test this please` works for me. But it's not a big problem in practice so no pressure to investigate further unless you're curious 😄

gongsu832:
> @gongsu832 thanks for unblocking this PR.
> 
> Not important but, in case you're curious in digging deeper, I think there's some kind of pattern to jenkins not listening to Philip and me, as I've experienced this problem once before with a previous bump-llvm PR. Maybe the pattern is that if the PR is authored by a non-committer then jenkins won't listen to me because I lack some kind of elevated admin privileges? In other situations (maybe when the PR is authored by a committer) `@jenkins-droid test this please` works for me. But it's not a big problem in practice so no pressure to investigate further unless you're curious 😄

People on the whitelist are all the same. There is no check for committer vs non-committer or any other github attributes. It's just a simple regex search for the "@jenkins-droid test this please" phrase. What can happen and has happened before is that sometimes the commit message may have special characters and break the regex search. But since it worked for me, it should work for you and Phil as well. So yeah I'm at a loss why you and Phil have trouble with it. It's a bit difficult to debug since when the Generic-Webhook-Trigger Plugin fails to trigger the build due to the regex filter not matching, it just fails silently without producing any diag info. I'm asking the GWT developers if they can consider adding some diag info to help debug filter failures.

sorenlassen:
@jenkins-droid test this please

sorenlassen:
@gongsu832 @AlexandreEichenberger can you give @Connor-XY committer privileges? that seems reasonable in all cases and it might unblock jenkins (which is now again refusing to run)

gongsu832:
> @gongsu832 @AlexandreEichenberger can you give @Connor-XY committer privileges? that seems reasonable in all cases and it might unblock jenkins (which is now again refusing to run)

@Connor-XY has been added to the whitelist to trigger Jenkins builds.

sorenlassen:
@Connor-XY fyi, the linux builds fail with
```
src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.td:152:40: error: Value specified for template argument 'Pat:supplemental_results' is of type dag; expected type list<dag>: (addBenefit 1)
def replaceONNXSumOpPatternRecursion : Pat<
```

Connor-XY:
> > @gongsu832 @AlexandreEichenberger can you give @Connor-XY committer privileges? that seems reasonable in all cases and it might unblock jenkins (which is now again refusing to run)
> 
> @Connor-XY has been added to the whitelist to trigger Jenkins builds.

Thanks for adding me to the whitelist! Really appreciate that!

Connor-XY:
> @Connor-XY fyi, the linux builds fail with
> 
> ```
> src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.td:152:40: error: Value specified for template argument 'Pat:supplemental_results' is of type dag; expected type list<dag>: (addBenefit 1)
> def replaceONNXSumOpPatternRecursion : Pat<
> ```

Thank you! I found that I need to add supplementalPatterns to Patterns in Accelerators.

sorenlassen (approved):
LGTM

philass (approved):
Thanks @Connor-XY!

PR #2463, opened 2023-08-24
[NNPA] Reconstruct onnx element-wise ops from zhigh ops

tungld:
This patch adds a new pass `--convert-zhigh-to-onnx` to revert a pattern `stick -> zhigh element_wise ops -> unstick` back to onnx element-wise ops. Given that stick and unstick are simdized on CPU, onnx simdized element-wise ops would provide better total performance with a smaller number of vector instructions.

In NNPA driver, `--convert-zhigh-to-onnx` pass will be called after the compiler has tried all optimizations to eliminate stick/unstick, but still having `stick -> zhigh element_wise ops -> unstick` alone.

I have tested with two models, roberta and flan-ul2, and seen performance improvement for both.


AlexandreEichenberger (approved):
LGTM,
Once we get SIMD on stickyfied data, it will be interesting to see if some of the sequences 
```
stick
nnpa-op, nnpa-op, nnpa-op, nnpa-op
unstickfy
cpu-op
stick
nnpa-op
unstick
```
will not be beneficial to simdize on CPU. For that, we will need to develop a "can-it-be-simdize-in-stickified-format" analysis of the element wise op, and decide that this rule may not apply in this circumstances...

Also, my belief is that we can count the number of "load-store" in loops to get a rough estimate of whether an op is good on NNPA or not. Say if we have "3 ops on NNPA and need 4 stick/unstick" that is 4 full-data for stick/unstick, 4 half-data for stick/unstick, plus each of the ops in half-data. Contrast this with the each op in full-data on CPU. If we can fuse the loops, on CPU, then we also reduce the full-data on CPU...

So this PR is a good first step in a longer process

chentong319:
1. Could we add the flag to disable the pass? The flag will make it possible for me to test the performance difference. 
2. The pattern for each onnx op is kind of the same. Any easy way to specify a bunch of element-wise ops? Not a hard requirement.

tungld:
> Could we add a flag to disable the pass? The flag will make it possible for me to test the performance difference.
The pattern for each onnx op is kind of the same. 

Sure, will do.

> Any easy way to specify a bunch of element-wise ops? Not a hard requirement.

The number of element-wise ops in NNPA is small (about 10), so let me keep the tablegen-based rules to make it easy to understand at this moment.

chentong319 (approved):
LGTM!

PR #2464, opened 2023-08-24
Opt reporting: add signatures to stats

AlexandreEichenberger:
Compiling with `--instrument-onnx-signature` and then gathering the output at runtime, we can then analyze ops using their runtime signature as well.

```
> roberta-base-11 % ONNX_MLIR_INSTRUMENT_FILE=run1.log RunONNXModelZoo.py -m roberta-base-11 -c="-O3 --instrument-onnx-signature"
...
> alexe@AChampignon roberta-base-11 % make-report.py -c run1.log -s sig -l 2
Statistic legend:
   num: comma separated list of shapes with inputs followed by results 

Statistics start, all ops, ordered_by name
  onnx.Add, 173
    12: sig, 1x12x1x1xfloat, 1x1x1x1xfloat, 1x12x1x1xfloat
    25: sig, 1x1x1xfloat, float, 1x1x1xfloat
    12: sig, 1x1x3072xfloat, 3072xfloat, 1x1x3072xfloat
    12: sig, 1x1x3072xfloat, float, 1x1x3072xfloat
    26: sig, 1x1x768xfloat, 1x1x768xfloat, 1x1x768xfloat
    85: sig, 1x1x768xfloat, 768xfloat, 1x1x768xfloat
    1: sig, 1x1xint64_t, int64_t, 1x1xint64_t
```

That way, we can quickly see which type of ops use which pattern.


src/Runtime/OMInstrument.inc
```diff
@@ -159,6 +159,21 @@ void ReportMemory() {
 }
 #endif
 
+FILE *getInstrumentFile() {
+  if (fout)
+    return fout;
+  fout = stdout;
+  if (getenv("ONNX_MLIR_INSTRUMENT_FILE")) {
+    char *fileName = getenv("ONNX_MLIR_INSTRUMENT_FILE");
+    FILE *newFileHandle = fopen(fileName, "w");
```
tungld, line 168: When do you close it?

tungld (commented):
Please undo deleted `.md` files. I sometimes get the same issue where `.md` files are deleted so I have to check them out manually. If you have a better solution, please share it.

src/Runtime/OMInstrument.inc
AlexandreEichenberger: yes, I don't like it. There is an OMInstrumentInit, but it is never called. Ideally we should have an OMInstrumentClose.

Would you be able to look into this? And if you do, I would consider renaming omInstrumentInit, to match all the other calls.

AlexandreEichenberger: Should also add that the normal path has no issue, it's only when we call profiling or instrumentation.

tungld:
@AlexandreEichenberger still see that this patch deletes `.md` files that are not relevant.

tungld (approved):
LGTM.

src/Runtime/OMInstrument.inc
tungld: It would be help if you can create an issue to remind us about the missing of close. Thanks!

PR #2467, opened 2023-08-30
[NNPA] Reconstruct onnx element-wise ops from zhigh ops - add a missing lit test for relu

tungld:
Just added a missing lit test for relu in PR #2463


chentong319 (approved):
LGTM!

PR #2471, opened 2023-09-01
Support signed int divisor in onnx.Mod op

negiyas:
This PR supports signed int divisor in onnx.Mod op as well as unsigned int divisor.　This PR includes the conversion pass, its lit and backend tests.  The following comments explain difference between return values of math.rem and onnx.Mod op.
 
```
    // "math.rem" returns "minus" for minus dividend and "plus or zero" for plus dividend.
    // We call the math.rem's return value "mathRemaider".
    // However onnx.ModOp should return "minus" for minus divisor and "plus or zero" for plus divisor.
    // we call the value that onnx.Mod op should return "onnxMod". The following table shows mathRemainder,
    // onnxMod and their diference (=onnxMod-mathRemainder) for some inputs.
    //
    // dividend                |  7  |  7 | -7 | -7 |  6 |  6 | -6 | -6 |
    // divisor                 |  3  | -3 |  3 | -3 |  3 | -3 |  3 | -3 |
    // ------------------------+-----+----+----+----+----+----+----+----+
    // mathRemainder           |  1  |  1 | -1 | -1 |  0 |  0 |  0 |  0 |
    // onnxMod                 |  1  | -2 |  2 | -1 |  0 |  0 |  0 |  0 |
    // onnxMod - mathRemainder |  0  | -3 |  3 |  0 |  0 |  0 |  0 |  0 |
    //
    // The following code shows logic to get onnxMod from mathRemainder
    //
    // int dividend, divisor;
    // int mathRemainder = diviend % divisor;
    // int adjustedRemainder = mathRemainder + divisor;
    //
    // if ((mathRemainder != 0) && ((dividend < 0) ^ (divisor < 0))) # c.f. "^" shows "exclusive or".
    //   return adjustedRemainder;
    // return mathRemainder;
```


chentong319:
Should we use [llvm srem](https://llvm.org/docs/LangRef.html#srem-instruction)?

negiyas:
> Should we use [llvm srem](https://llvm.org/docs/LangRef.html#srem-instruction)?
@chentong319 Thank you for useful comments. 

I investigated definitions and behavior of llvm.srem, and arith.remsi (https://mlir.llvm.org/docs/Dialects/ArithOps/#arithremsi-arithremsiop), that is converted into llvm.srem. According to the investigation, these operations support signed integers, but they cannot generate right answers when the divisor is minus. If we simply use arith.remsi for onnx.Mod, the backend tests ("test_mod_mixed_sign_int64_cpu", "test_mod_mixed_sign_int32_cpu") fail as follows.

```
 Mismatched elements: 2 / 6 (33.3%)
Max absolute difference: 3
Max relative difference: 1.5
 x: array([ 0,  1,  5,  0, -1,  3], dtype=int32)
 y: array([ 0, -2,  5,  0,  2,  3], dtype=int32)
```

Accordingly, it seems that additional logic in this PR is necessary to generate right answers for the divisor-is-minus cases.
Your comments to simplify the code are very welcome! Thank you.

chentong319 (dismissed):
LGTM!

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1136,9 +1136,25 @@ Value emitScalarOpFor<ONNXModOp>(ConversionPatternRewriter &rewriter,
...
-                     "are signless.");
+    // int diviend, divisor;
+    // rem = diviend % divisor
+    // if (rem == 0)
+    //   return rem
+    // if (diviend < 0)
+    //   rem = rem - divisor  // make rem minus
+    // if (divisor < 0)
+    //   rem = rem + divisor; // make rem minus
+    // return rem;
```
chentong319, line 1147: The code is little tricky. To understand it, I list out the case and result according to your code (NOT the comment):

1. dividend > 0 and divisor >0, rem unchanged
2. dividend < 0 and divisor <0, rem = (rem+divisor) + divisor = rem + 2*divisor
3. dividend < 0 and divisor > 0, rem = rem + divisor = rem + |divisor|
4. dividend > 0 and divisor < 0, rem = rem+divisor = rem - |divisor|

Is the second case correct? Fix the difference of the comment and code first, please.
Another question: do we need to check eq to zero again at line 1156?

chentong319 (approved):
LGTM!

PR #2475, opened 2023-09-05
Do not remove LLVM "internal" options

gongsu832:
- Do not remove LLVM "internal" options such as --debug
- Clean up some header includes and library link dependencies


sorenlassen (approved):
LGTM

PR #2476, opened 2023-09-05
[TOSA] Conversion of ONNXAveragePool2d to TOSA

maxbartel:
I have created a new function `tosa::convertPoolOp<>` which unifies the common parts of the maxpool and the averagepool conversion.


eric-k256 (approved):
Looks good to me.

maxbartel:
@AlexandreEichenberger Could I get a review, please?

AlexandreEichenberger (commented):
I checked and at this time, we don't have any code in your header file. MLIR has only very few short code segment. The patten we used is to create a `.hpp.inc` and put the code there, which is then included from the header file. That way we maintain concise header files with all the benefits of still having inlined code, if you believe that you would not like to create a new .cpp file to contain the helper functions.

Please reformat this PR to use this approach, or motivate the reason for this departure of the current code pattern. 

Otherwise, the code looks good.

AlexandreEichenberger (approved):
@maxbartel I can do the transformation after your PR goes in if you don't have the time, since its minor.

maxbartel:
I wanted to do it today anyway, but thank you for the offer! @AlexandreEichenberger

AlexandreEichenberger:
You have merge privilege, right?

maxbartel:
Yes, I have!

PR #2477, opened 2023-09-05
Some rewriting rules to simplify operation whose operands are from ConstantOfShape

tungld:
This patch adds some canonicalization rules for Unsqueeze, Add and Sub operations when their inputs are fromConstantOfShape. 

Patterns are found in the roBERTa model, that is 
```mlir
%102 = onnx.Constant dense<1.000000e+00> : tensor<f32>
%367 = onnx.ConstantOfShape(%366) {onnx_node_name = "ConstantOfShape_9", value = dense<1.000000e+00> : tensor<1xf32>} : (tensor<2xi64>) -> tensor<?x?xf32>
// These two Unsqueezes can be folded into a new onnx.ConstantOfShape with value 1.0.
%370 = "onnx.Unsqueeze"(%367, %104) : (tensor<?x?xf32>, tensor<1xi64>) -> tensor<?x1x?xf32>
%371 = "onnx.Unsqueeze"(%370, %103) : (tensor<?x1x?xf32>, tensor<1xi64>) -> tensor<?x1x1x?xf32>
// This Sub can be folded into onnx.ConstantOfShape with value 0 since %102 and %371 are 1.0.
%372 = "onnx.Sub"(%102, %371) {onnx_node_name = "Sub_18"} : (tensor<f32>, tensor<?x1x1x?xf32>) -> tensor<?x1x1x?xf32>  
// Multiply by 0
%373 = "onnx.Mul"(%372, %101) {onnx_node_name = "Mul_20"} : (tensor<?x1x1x?xf32>, tensor<f32>) -> tensor<?x1x1x?xf32> 
// Add to 0
%418 = "onnx.Add"(%417, %373) {onnx_node_name = "Add_91"} : (tensor<?x12x?x?xf32>, tensor<?x1x1x?xf32>) -> tensor<?x12x?x?xf32>
```


src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
```diff
@@ -388,6 +388,60 @@ class SplitLargeMatMulPattern : public OpRewritePattern<ONNXMatMulOp> {
   }
 };
 
+/// This pattern is to replace `C = add/sub(A, B)` by `A` when B is a zero
+/// defined by ConstantOfShape and C's shape is the same as A's shape.
+/// In other words, the output does not depend on the second operand.
+template <typename OP_TYPE>
+class AddSubWithZeroConstantOfShapePattern : public OpRewritePattern<OP_TYPE> {
```
tungld, line 395: It's better to do this pattern at ONNX level, but since it uses DimAnalysis and we have no onnx-to-onnx pass that uses DimAnalysis, so we put this pattern here.

src/Dialect/ONNX/Rewrite.cpp
sorenlassen, line 1: it seems to me that that ReplaceBinaryOpByConstantOfShapePattern and ReplaceUnsqueezeOfConstantOfShapeRewritePattern stretch the scope of canonicalization

can they go into SimplifyShapeRelatedOpsPass instead? or another existing or now rewriting pass?

AlexandreEichenberger:
Before looking at the details of this PR, if I understand well, you suggest that we propagate the "unsqueezed" into the constant-shape... namely

```
%366a = concat( %366[0], 1, 1, %366[1])
%371 = const of shape (%366a) {value = 1.0}
%372 = sub (%371, const 1.0) 
%373 = mul (%372, const 0.0)
%418 = add(%417, %373)
```

And then constant propagation should do the trick ? We cannot use our constant tensors way since I believe our  constants expect constant shapes, and here the shape is dynamic, isn't it @sorenlassen ? But we can generate a series of `const of shape` with the appropriate `value` attribute for each intermediary results.

I really think this opt should be an onnx to onnx transformation, as this really benefits all of the paths.

In addition, we may convert the `const of shape` into `const in certain conditions for a more compact representation.

If the use of %373  is by an operation supporting broadcast:
```
%373 = const of shape (%366a) { value = x }
%xxx = <elementary op supporting broadcast> (<results with shape yyy>, %373)
```

we can rely on the broadcast of the `<elementary op supporting broadcast>` to simplify to
```
%373a = const {value = x} of shape <1>
%xxx = <elementary op supporting broadcast> (<results with shape yyy>, %373)
```
when we can prove that the constant `%373` is not "expanding" any dimensions of `<results with shape yyy>`. Namely if `y` is a value of the shape in `yyy`, then if `y=1`, then the corresponding shape in `%373` must be 1; if `y=?`, then the corresponding shape in `%373` must be 1 or the same `?` value.

sorenlassen:
> We cannot use our constant tensors way since I believe our constants expect constant shapes, and here the shape is dynamic, isn't it @sorenlassen ?

right

> In addition, we may convert the const of shape into const in certain conditions for a more compact representation.

would it be helpful to rewrite ConstantOfShape to a scalar Constant followed by Expand?

src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -273,6 +273,120 @@ class BinaryOpBroadcastAxisPattern : public OpRewritePattern<OP_TYPE> {
...
+          RankedTensorType::get({1}, elementType), ArrayRef<int64_t>({res}));
+    } else if (isa<FloatType>(elementType)) {
+      double lhs = getScalarValue<double>(lhsAttr, elementType);
+      double rhs = getScalarValue<double>(rhsAttr, elementType);
+      float res =
+          (float)ElementWiseBinaryOpImpl<OP_TYPE, double>::eval(lhs, rhs);
+      resAttr = DenseElementsAttr::get(
+          RankedTensorType::get({1}, elementType), ArrayRef<float>({res}));
+    } else
+      llvm_unreachable("Unexpected type.");
```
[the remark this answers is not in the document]
tungld: Yes, it would be helpful. OpHelper.{cpp/hpp} is one of the candidate. What do you think?

tungld:
@AlexandreEichenberger your understanding is correct. @AlexandreEichenberger @sorenlassen if there is no objection, I will go with a new pass, say `onnx-rewrite`

src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -865,6 +979,84 @@ class PowToMulRewritePattern : public OpRewritePattern<ONNXPowOp> {
...
+    // Rewrite
+    MultiDialectBuilder<OnnxBuilder> create(rewriter, loc);
+    // Get the old shape.
+    SmallVector<Value, 4> oldDims;
+    getDims(constantOfShapeOp.getInput(), oldDims);
+    int64_t oldRank = oldDims.size();
+    // Get unsqueeze axes.
+    ElementsAttr axesAttrs = getElementAttributeFromONNXValue(axes);
+    SmallVector<int64_t, 4> axesI64;
+    getArrayFromElementsAttr<int64_t>(axesAttrs, axesI64);
```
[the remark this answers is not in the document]
tungld: Thanks!

src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -273,6 +273,120 @@ class BinaryOpBroadcastAxisPattern : public OpRewritePattern<OP_TYPE> {
...
+    MultiDialectBuilder<OnnxBuilder> create(rewriter, loc);
+    ElementsAttr lhsAttr, rhsAttr;
+    if (definedBy<ONNXConstantOp>(lhs))
+      lhsAttr = cast<ElementsAttr>(constantOp.getValue().value());
+    else
+      lhsAttr = cast<ElementsAttr>(constantOfShapeOp.getValue().value());
+    if (definedBy<ONNXConstantOp>(rhs))
+      rhsAttr = cast<ElementsAttr>(constantOp.getValue().value());
+    else
+      rhsAttr = cast<ElementsAttr>(constantOfShapeOp.getValue().value());
```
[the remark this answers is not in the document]
tungld: Good suggestion! Thanks!

src/Dialect/ONNX/Rewrite.cpp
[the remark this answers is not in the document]
tungld: Done. I moved them into `OpHelper.hpp`.

src/Compiler/CompilerPasses.cpp
```diff
@@ -95,15 +95,14 @@ void addONNXToMLIRPasses(mlir::PassManager &pm, bool targetCPU) {
...
+      pm.addPass(onnx_mlir::createRewriteONNXToONNXPass());
       pm.addNestedPass<func::FuncOp>(onnx_mlir::createShapeInferencePass());
       pm.addNestedPass<func::FuncOp>(
           onnx_mlir::createConstPropONNXToONNXPass());
     }
   }
 
-  // Simplify shape-related ops.
-  pm.addPass(onnx_mlir::createSimplifyShapeRelatedOpsPass());
-
```
tungld, line 106: Move `SimplifyShapeRelatedOpsPass` pass into ONNXOpTransform pass so that `SimplifyShapeRelatedOpsPass` can be called repeatedly.

tungld:
@AlexandreEichenberger @sorenlassen I changed the file `Rewrite.cpp` that contains canonicalization patterns into a new one `Canonicalize.cpp` and used `Rewrite.cpp` for a new pass `onnx-rewrite`. `onnx-rewrite` will be home for onnx-to-onnx transformations, and we will move "complex" canonicalization patterns from `Canonicalize.cpp` to this pass in the future.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
```diff
@@ -388,6 +388,60 @@ class SplitLargeMatMulPattern : public OpRewritePattern<ONNXMatMulOp> {
   }
 };
 
+/// This pattern is to replace `C = add/sub(A, B)` by `A` when B is a zero
+/// defined by ConstantOfShape and C's shape is the same as A's shape.
```
sorenlassen, line 392: I think all your rewrites might become a little bit simpler if you start by rewriting/canonicalizing `ConstantOfShape(value=x,shape=S)` to `Exand(Constant(value=x,shape=[]),shape=S)`

as @AlexandreEichenberger mentioned, here we could have a simple rule to rewrite `Add/Sub(A,Expand(B,S))` to `Add/Sub(A,B)` when shape(A)==S (and generalize it to any op that broadcasts it's operands)

then the const prop rules Add/SubZerosOnRhs would finish your rewrite here if you implement the const prop rules TODO to follow them with Expand

I think the other patterns
ReplaceBinaryOpByConstantOfShapePattern and
ReplaceUnsqueezeOfConstantOfShapeRewritePattern
also become a little bit simpler and more general if you define them for Expand after rewriting ConstantOfShape to Constant and Expand

I'll try to take a closer look at the whole PR tonight

AlexandreEichenberger: Thanks for the detailed explanation, I know understand how using the `constantOfShape` benefits from being expressed as a `expand(const(...), shape= )`.

In by original review, I was eager to see if we can do the constant propagation can happen when all constants are known but for the shape. Is there a way to sink the computations inside the extend?
```
add(const 1, expand(const 2, shape)) -> expand(add(const 1, const 2), shape) -> expand(const 3, shape)
```

tungld: I like the idea of using `Exand(Constant(value=x,shape=[]),shape=S)` because we can reuse the constant propagation, and it also means we should call the constant propagation after each op rewriting. So for a long chain of operations, we should call `onnx-rewrite -> constprop -> onnx-rewrite -> ...` repeatedly, and I am afraid that it would be slow when we add more rules to `onnx-rewrite` in the future. Meanwhile, this patch eagerly folds constants to have a new ConstantOfShape, so that all ConstantOfShape-related ops can be folded in a single pass.

sorenlassen: > for a long chain of operations, we should call `onnx-rewrite -> constprop -> onnx-rewrite -> ...` repeatedly, and I am afraid that it would be slow when we add more rules to `onnx-rewrite` in the future

we could do onnx-rewrite together with constprop in a hybrid pass

I had always planned to include constprop in the hybrid pass, see the implementation in PR #2494

tungld: It looks to me you gather all patterns to put them into a single pass HybridPass, so does it mean patterns in `onnx-rewrite` will be added to HybridPass too? and in the future all patterns related to onnx ops will go to HybridPass?

sorenlassen: > It looks to me you gather all patterns to put them into a single pass HybridPass, so does it mean patterns in `onnx-rewrite` will be added to HybridPass too? and in the future all patterns related to onnx ops will go to HybridPass?

yes, we could add onnx-rewrite to hybrid pass

you could also make another hybrid pass, e.g. with just onnx-rewrite and constprop, if that works best to achieve your goals

tungld: How about using `ONNXOpTransformPass`? I am not clear which of `ONNXOpTransformPass` or `HybridPass` is used for which situation?

sorenlassen: > Is there a way to sink the computations inside the extend?
> `add(const 1, expand(const 2, shape)) -> expand(add(const 1, const 2), shape) -> expand(const 3, shape)`

I think this could be accomplished with the rule:
`Add(X,Expand(Y,shape=S)) -> Expand(Add(X,Y),shape=S)` when `X` has scalar shape (or any shape that is known to broadcast to Y's shape, but that's tricky to check more generally if Y has dynamic shape)

it can be generalized to any operand order and any broadcastable op with 2 or more operands

sorenlassen: fwiw, ReplaceUnsqueezeOfConstantOfShapeRewritePattern could maybe also be reformulated to act on Expand instead of ConstantOfShape so it could apply to Expand in other situations than just those that arise from ConstantOfShape

sorenlassen: > How about using `ONNXOpTransformPass`? I am not clear which of `ONNXOpTransformPass` or `HybridPass` is used for which situation?

I was hoping that hybrid pass would replace ONNXOpTransformPass

it should be more efficient because it's designed do everything in a single pass

it's also potentially simpler to understand the hybrid pass at a high level as just doing all the included sets of patterns (shape inference, canonicalization, constprop) at once compared to understanding the logic in ONNXOpTransformPass (iteration until convergence, and a sequence of passes with shape inference in between each)

tungld:
@AlexandreEichenberger @sorenlassen I create a new PR that decomposes ConstantOfShape into Expand and Constant: #2499, which will replace this PR. Closing this PR.

PR #2478, opened 2023-09-05
Add support for string data type

hamptonm1:
Add surrounding support for string data type


hamptonm1:
@chentong319 What are your thoughts? Let me know if I am missing anything here.

src/Runtime/PyExecutionSessionBase.cpp
```diff
@@ -105,7 +105,8 @@ std::vector<py::array> PyExecutionSessionBase::pyRun(
       dtype = ONNX_TYPE_INT32;
     else if (py::isinstance<py::array_t<std::int64_t>>(inputPyArray))
       dtype = ONNX_TYPE_INT64;
-    // string type missing
+    // else if (py::isinstance<py::array_t<const char *>>(inputPyArray))
```
hamptonm1, line 108: @chentong319 Here I added the missing string type which uses `const char *` pybind11 datatype.

src/Runtime/OMUnique.inc
```diff
@@ -94,7 +94,8 @@ int isLessNum(void *arg1, void *arg2, OM_DATA_TYPE dataType) {
     return *((int32_t *)arg1) < *((int32_t *)arg2);
   case ONNX_TYPE_INT64:
     return *((int64_t *)arg1) < *((int64_t *)arg2);
-  // case ONNX_TYPE_STRING:
+  case ONNX_TYPE_STRING:
```
hamptonm1, line 97: I updated the missing `ONNX_TYPE_STRING` case for `[OMUnique.inc](https://github.com/onnx/onnx-mlir/pull/2478/files#diff-6406fc477efda850ee50a9eb2ba96b0a6edb3ce3e5de8b03072688c8ee29f146)`

src/Runtime/jni/src/com/ibm/onnxmlir/OMRunner.java
```diff
@@ -41,207 +41,221 @@
...
+			put(OMTensor.ONNX_TYPE_INT32, numpyEndian + "i4");
+			put(OMTensor.ONNX_TYPE_UINT32, numpyEndian + "u4");
+			put(OMTensor.ONNX_TYPE_INT64, numpyEndian + "i8");
+			put(OMTensor.ONNX_TYPE_UINT64, numpyEndian + "u8");
+			put(OMTensor.ONNX_TYPE_FLOAT16, numpyEndian + "f2");
+			put(OMTensor.ONNX_TYPE_FLOAT, numpyEndian + "f4");
+			put(OMTensor.ONNX_TYPE_DOUBLE, numpyEndian + "f8");
+			// numpy documentation: Unicode string
+			// https://numpy.org/doc/stable/reference/arrays.dtypes.html
+			put(OMTensor.ONNX_TYPE_STRING, "|U25");
```
hamptonm1, line 121: I also added `U25` which indicates a unicode string in numpy up to length 25. In the documentation, it says to avoid using S25 ->np.string_ because that maps to a byte string and it is zero-terminated bytes (not recommended).

hamptonm1: The addition is for line 101 and line 121. All of the changes in this file has to do with formatting.

src/Runtime/PyExecutionSessionBase.cpp
hamptonm1: I have to comment this out for now until we get pybind11 support for it....I get the error `runtimeerror: numpy type info missing for pkc`.

chentong319: This is the first place we got assertion error. The link you provided is really helpful.

src/Runtime/jni/src/com/ibm/onnxmlir/OMRunner.java
chentong319: Let's focus on byte string now. You can leave an assertion for unicode string.

hamptonm1:
I am closing this one out because someone else already started working on it :)

PR #2485, opened 2023-09-06
Bundling of memory is now done by MLIR only

AlexandreEichenberger:
Removed functionality that is now done by MLIR. Could not fully remove `KrnlGetRef` Op because it is used for one operation. @chentong319 maybe you could investigate if there is an alternative way to get the address of a memory, as it is used with strings. If not possible, its fine to keep `KrnlGetRef`


src/Conversion/ONNXToKrnl/ML/CategoryMapper.cpp
```diff
@@ -268,6 +268,7 @@ struct ONNXCategoryMapperOpLowering
                 (shape[i] == ShapedType::kDynamic) ? 1 : shape[i]);
           auto memRefType = MemRefType::get(
               newShape, krnl::StringType::get(elementType.getContext()));
+          // Sole use of krnl.getRef.
```
AlexandreEichenberger, line 271: @chentong319 this is the only use of `getRef` in the current code base.

chentong319 (approved):
LGTM!

chentong319:
The krnl.getRef and krnl.find_index can be removed by using the generic krnl.call. In the library of the find_index, we can get the data_ptr from the omTensor. Therefore, we do not need to get ref in onnx-mlir. Leave these change to another PR.

AlexandreEichenberger:
> The krnl.getRef and krnl.find_index can be removed by using the generic krnl.call. In the library of the find_index, we can get the data_ptr from the omTensor. Therefore, we do not need to get ref in onnx-mlir. Leave these change to another PR.

@chentong319 if you do this change, I will do a follow on PR to remove getRef if we believe that we don't need it anymore.
Thanks for the insight.

PR #2486, opened 2023-09-07
Add a pattern to remove identity onnx.Reshape

tungld:
When the output and input of Reshape have the same shape, Reshape becomes an identity op. This patch adds a pattern for such a case.

This pattern was found in the roBERTa model


src/Dialect/ONNX/ONNXOps/OpHelper.cpp
```diff
@@ -396,6 +396,27 @@ bool hasShapeAndRank(Operation *op) {
...
+  for (uint64_t i = 0; i < lhsShape.size(); ++i) {
+    int64_t x = lhsShape[i];
+    int64_t y = rhsShape[i];
+    if ((x == ShapedType::kDynamic) || (y == ShapedType::kDynamic) ||
+        (x != y)) {
+      allSame = false;
+      break;
+    }
+  }
+  return allSame;
```
sorenlassen, line 417: could you just say?
```
return hasStaticShape(lhs.getType()) && getShape(lhs) == getShape(rhs);
```

and consider putting it in Rewrite.cpp because it seems rather specialized and unlikely to be used outside Rewrite.td

sorenlassen (approved):

src/Dialect/ONNX/ONNXOps/OpHelper.cpp
tungld: Thanks! I updated the code and moved it into Rewrite.cpp.

PR #2487, opened 2023-09-08
Fix constant propagation pass issue for onnx.Gather op.

negiyas:
Fix constant propagation pass issue for onnx.Gather op.
Current constant propagation pass does not work for onnx.Gather with unranked indices (e.g. Tensor<i32>).
This PR stops constant propagation with the case. 

[X] Prepare LIT test to check errors

```
onnx-mlir: ./src/Dialect/ONNX/ElementsAttr/DisposableElementsAttr.hpp.inc:75: onnx_mlir::ArrayBuffer<X> mlir::DisposableElementsAttr::getArray() const [with X = long int]: Assertion `onnx_mlir::toBType<X> == getBType() && “X must match element type”‘ failed.
```

### DIFF
--- a/src/Transform/ONNX/ConstProp.td
+++ b/src/Transform/ONNX/ConstProp.td
@@ -84,6 +84,12 @@ def IsConstOfOnes : Constraint<
   CPred<"isDenseONNXConstant($_self) && isConstOf($_self, 1.0)">,
   "Value is an all-ones constant tensor">;
 
+// Check the rank of a value is greater than a given integer.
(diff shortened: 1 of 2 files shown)


sorenlassen:
actually, I think the real problem is that the const prop implementation only works for int64 indices and not int32, sorry about that

rank 0 indices seems to work fine for int64

please see the alternative fix in PR #2488

negiyas:
@sorenlassen Your comments are right. I am closing this PR. Thank you for your comments.

PR #2489, opened 2023-09-11
[NNPA] Do not remove unstick-view-stick in case of nchw layout

tungld:
This patch fixes an issue in the rewriting rule that removes `unstick-view-stick`. As NCHW layout unstickification and stickification internally do data transpose, we can not remove the pattern in such a case.


AlexandreEichenberger (approved):
LGTM

PR #2490, opened 2023-09-11
Ingest Opset 19 Resize

philass:
`coordinate_transformation_mode` adds a string mode for `"half_pixel_symmetric"`.

As a result there are no changes in the underlying representation of the op. Hence just a bump of `Resize` in `gen_onnx_mlir.py`


chentong319 (approved):
LGTM!

PR #2494, opened 2023-09-12
add const prop to hybrid pass

sorenlassen:
Introduced a flag `-onnx-hybrid-pass-disable-class` to disable any of shape-inference, canonicalization, and constant-propagation.

### DIFF
--- a/src/Compiler/CompilerOptions.cpp
+++ b/src/Compiler/CompilerOptions.cpp
@@ -35,6 +35,7 @@ InstrumentStages instrumentStage;                      // common for both
 int onnxConstPropExpansionBound;                       // common for both
 std::vector<std::string> onnxConstPropDisablePatterns; // common for both
 bool enableONNXHybridPass;                             // common for both
+std::vector<std::string> onnxHybridPassDisableClasses; // common for both
 std::vector<std::string> functionsToDecompose;         // common for both
 EmissionTargetType emissionTarget;                     // onnx-mlir only
 bool invokeOnnxVersionConverter;                       // onnx-mlir only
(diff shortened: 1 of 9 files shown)


sorenlassen:
I had always planned to add constprop and decomposition to the hybrid pass

this PR adds constprop

for decomposition I'm not 100% sure how to do it because it's a implemented as a partial conversion, while the hybrid pass is a "plain rewrite"

sorenlassen:
@chentong319 @tungld this PR is ready for review: I introduced a flag -onnx-hybrid-pass-disable-class to disable any of shape-inference, canonicalization, and constant-propagation, and I ran `RunONNXModelZoo.py -c="-O3"` on this branch and saw the same 119 models pass and 33 models fail as on main

src/Compiler/CompilerOptions.cpp
```diff
@@ -177,6 +178,16 @@ static llvm::cl::opt<bool, true> enableONNXHybridPassOpt("onnx-hybrid-pass",
...
 
+static llvm::cl::list<std::string, std::vector<std::string>>
+    onnxHybridPassDisableClassesOpt("onnx-hybrid-pass-disable-class",
+        llvm::cl::desc("Class of hybrid pass patterns to disable.\n"
+                       "Can be shape-inference, canonicalization, "
+                       "constant-propagation, decomposition.\n"
+                       "Repeat the flag to disable multiple classes."),
+        llvm::cl::value_desc("class of hybrid pass patterns to disable"),
+        llvm::cl::location(onnxHybridPassDisableClasses),
+        llvm::cl::cat(OnnxMlirCommonOptions));
```
tungld, line 189: I would make this option a sub-option of `-onnx-hybrid-transform` instead of a parallel option here, e.g. `--onnx-hybrid-transform=disable-class=constant-propagation,canonicalization`

chentong319 (approved):
LGTM!

src/Compiler/CompilerOptions.cpp
sorenlassen: ok

done

sorenlassen: you can call it with `--onnx-hybrid-transform="constant-propagation=false canonicalization=false"`

tungld: Thanks!

tungld (approved):
LGTM!

PR #2495, opened 2023-09-12
Add support for opset 19 AveragePool with dilations

philass:


AlexandreEichenberger (approved):
LGTM, but please update this field as appropriate before merging

onnx-mlir/test/backend/inference_backend.py
```
        # ==OP== AveragePool
        # ==MIN== 1
        # ==UNSUPPORTED== 19
```
If it is now fully supported, then remove the `==UNSUPPORTED==` line. If it is partially supported, then still remove this line and add a ` ==LIM== <explanation>` line.

PR #2496, opened 2023-09-12
add decomposition to hybrid pass

sorenlassen:
Changed the conversion patterns to plain rewrite patterns so they can be used with the greedy rewrite in the hybrid pass. Since decomposition doesn't convert types, there's no need to use conversion patterns.

This change complements PR #2494 that adds const prop to the hybrid pass.

The original motivation for the hybrid pass was an example that needed all four of (1) shape inference, (2) canonicalization, (3) constant propagation, and (4) decomposition to compile a model with a sequence of Einsum operations that needed all of the above to derive static shapes for all the Einsum operands.


sorenlassen:
@tungld @chentong319 this PR is ready for review, I have rebased it after PRs #2494 (const prop) and #2515 (Squeeze/UnsqueezeV11)

tungld:
`Decompose` pass marked operations illegal so that we make sure it does not appear because many decomposed ops don't have shape inference or lowering:

```
target.addIllegalOp<ONNXPadV11Op>();
target.addIllegalOp<ONNXPadV13Op>();
```

Should we have a function for such `addIllegalOp` and call it from HybridPass, same as what you are doing for patterns? In my recent [PR](https://github.com/onnx/onnx-mlir/blob/main/src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.cpp#L300), I also  extract both patterns and legality check from a pass to be reused.

sorenlassen:
> Should we have a function for such `addIllegalOp` and call it from HybridPass, same as what you are doing for patterns?

as mentioned in the comment in ONNXHybridTransformPass.cpp my intention was that the hybrid pass is best effort
```
// Note that the decomposition patterns are applied "best effort" with a greedy
// rewrite, not a partial conversion with "legalization" to ensure that every
// decomposable op is decomposed.
```
and that we'd still run the decomposition pass to ensure that every decomposable op is decomposed

if we want the hybrid pass to ensure that it decomposes all illegal ops then I believe that we need to change applyPatternsAndFoldGreedily() to applyPartialConversion() and I worry that that will constrain the way we can apply the patterns

tungld (approved):
LGTM. Thanks for the explanation!

PR #2498, opened 2023-09-13
Create Constant Propagation flag

hamptonm1:
Constant Folding will be enabled only for level 03 and all other levels with have constant folding disabled so this way the output when compiling the models can be consistent across the board.


src/Transform/ONNX/ConstProp.cpp
```diff
@@ -23,6 +23,8 @@
 #include "llvm/ADT/StringSet.h"
 #include "llvm/Support/Debug.h"
 
+#include "src/Compiler/CompilerOptions.hpp"
+#include "src/Compiler/CompilerPasses.hpp"
```
sorenlassen, line 27: to make it easy to use each of the dialects and passes as libraries that people can link into their own compiler binaries without using the onnx-mlir binary, and to avoid unnecessary cyclical dependencies, we should avoid making the dialects and passes depend on anything in src/Compiler (with #include or cmake LINK_LIBS)

to disable constant propagation, consider instead changing addONNXToMLIRPasses() to not call createConstPropONNXToONNXPass()

sorenlassen (commented):
should we change "constant folding" to "constant propagation"? we generally use that terminology, e.g. in the ConstPropONNXToONNXPass name and in the const-prop flags and the docs https://github.com/onnx/onnx-mlir/blob/main/docs/ConstPropagationPass.md

are you sure that it's a good idea to change the default? it seems like a big and possibly disruptive behavior change

sorenlassen:
> In light of this, do you still believe it's a mistake to change the -O0 behavior?

thanks for the explanation

I don't object to making this change

hamptonm1:
> should we change "constant folding" to "constant propagation"? we generally use that terminology, e.g. in the ConstPropONNXToONNXPass name and in the const-prop flags and the docs https://github.com/onnx/onnx-mlir/blob/main/docs/ConstPropagationPass.md
> 
> are you sure that it's a good idea to change the default? it seems like a big and possibly disruptive behavior change

Thanks for your reply @sorenlassen I changed the name to constant propagation.

hamptonm1:
> > are you sure that it's a good idea to change the default? it seems like a big and possibly disruptive behavior change
> 
> @sorenlassen, the idea is that constant propagation changes the results (because of precision, esp when runtime computations are done on reduced precision engine), so it might make sense to disable it at -O0... which is really a flag that say "don't optimize much". Some compiler have a separate flag that disable optimizations that changes values (e.g. disable FMA because of the precision changes due to the higher precision inside the M and A of the FMA).
> 
> In light of this, do you still believe it's a mistake to change the -O0 behavior? Open to suggestions.

@sorenlassen @AlexandreEichenberger This is a great point.... I think @tungld was thinking the same thing. To keep constant prop intact for the default Level which is --O0 and have it disabled only for O1 and O2. I am perfectly fine with making the needed changes if that is what everyone is in agreement to :) 

Also we do have this option `--enable-constant-prop=true ` which will essentially enable constant propagation regardless of the level. So if someone specifies this flag on the command line during compile time.... they will be able to ensure constant prop is indeed on and it will override the default behavior as shown by the test we have here. 

`meganhampton@Megans-MacBook-Pro-2 onnx-mlir % ./Debug/bin/onnx-mlir --O0 --EmitLib roberta-base-11.onnx --onnx-op-stats TXT --enable-constant-prop=true --shapeInformation 0:1x1 -v `

src/Transform/ONNX/ConstProp.cpp
hamptonm1: I originally was going to do it the way you suggested but if I was to do it that way we would need to have multiple `if-statements` in several c++ files (NNPA/CompilerPass/simplifyshaperelatedops/onnxoptransformpass) where as doing it this way requires the update of only the `ConstProp.cpp` but if everyone feels this is much better I can make the update. I am fine with whatever is best for the community :)

hamptonm1:

src/Transform/ONNX/ConstProp.cpp
hamptonm1: @sorenlassen Okay I made the updates and tested but I went back to my old way of doing things but it removes the #include "src/Compiler/CompilerOptions.hpp" and #include "src/Compiler/CompilerPasses.hpp" file as you wanted.

src/Transform/ONNX/SimplifyShapeRelatedOps.cpp
```diff
@@ -60,6 +60,7 @@ Now, it's straightforward to update the output shape of Reshape from
 #include "mlir/Transforms/Passes.h"
 #include "llvm/Support/Debug.h"
 
+#include "src/Compiler/CompilerOptions.hpp"
```
sorenlassen, line 63: please don't add this dependency on src/Compiler from this pass

src/Transform/ONNX/ONNXOpTransformPass.cpp
```diff
@@ -86,8 +86,12 @@ void ONNXOpTransformPass::runOnOperation() {
       dynamicPM.addNestedPass<func::FuncOp>(
           onnx_mlir::createShapeInferencePass());
     }
-    dynamicPM.addNestedPass<func::FuncOp>(
-        onnx_mlir::createConstPropONNXToONNXPass());
+    if ((/*enableConstantProp*/ onnx_mlir::OptimizationLevel >= 3 &&
+            !onnx_mlir::enableConstantProp) ||
```
sorenlassen, line 90: ONNXOpTransformPass has an unfortunate dependency on CompilerOptions because it reads the enableConvOptPass flag - it would be ideal if we could break that dependency, and not add another dependency on the new flags

AlexandreEichenberger:
> To keep constant prop intact for the default Level which is --O0 and have it disabled only for O1 and O2.

Sorry, that does not work for me, happy to have a conversation with @tungld , generally optimizations are incrementally added when going to increasing level of O<num>. Doing differently would be counter intuitive. 

Just like gcc/clang, we expect folks looking for performance to use `-O3`.  For example, only that option unlock simd/blocking/... optimizations.

src/Transform/ONNX/ONNXOpTransformPass.cpp
hamptonm1: Either way I need the `OptimizationLevel`... that is the main reason why. So I still need to pull it from somewhere which is `src/Compiler/CompilerOptions.hpp`. This flag is contingent upon the level being specified at compile time.

src/Transform/ONNX/SimplifyShapeRelatedOps.cpp
hamptonm1: Maybe we should have a discussion... if we can make allowances for `ONNXOpTransformPass`, I do not see any harm with doing the same thing for the c++ file above.

hamptonm1:
> > To keep constant prop intact for the default Level which is --O0 and have it disabled only for O1 and O2.
> 
> Sorry, that does not work for me, happy to have a conversation with @tungld , generally optimizations are incrementally added when going to increasing level of O. Doing differently would be counter intuitive.
> 
> Just like gcc/clang, we expect folks looking for performance to use `-O3`. For example, only that option unlock simd/blocking/... optimizations.

Okay that works for me.

tungld:
> To keep constant prop intact for the default Level which is --O0 and have it disabled only for O1 and O2.

> 
> Sorry, that does not work for me, happy to have a conversation with @tungld , generally optimizations are incrementally added when going to increasing level of O. Doing differently would be counter intuitive.
> 
> Just like gcc/clang, we expect folks looking for performance to use -O3. For example, only that option unlock simd/blocking/... optimizations.

My idea is constant propagation is by default off  at -O{0, 1, 2} and  on at -O3. But at -O{0,1,2} users can turn it on manually by `--enable-constant-propagation`.

src/Transform/ONNX/SimplifyShapeRelatedOps.cpp
sorenlassen: the dependency in ONNXOpTransformPass is just an inconsistency and can be removed

please see PR #2506

src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
```diff
@@ -79,7 +79,10 @@ void addONNXToZHighPasses(
   pm.addNestedPass<func::FuncOp>(onnx_mlir::createShapeInferencePass());
   // There are more opportunities for const propagation once all zhigh ops were
   // generated.
-  pm.addNestedPass<func::FuncOp>(onnx_mlir::createConstPropONNXToONNXPass());
+  if ((/*enableConstantProp*/ optLevel >= 3 && !enableConstantProp) ||
```
AlexandreEichenberger, line 82: Nit:
```
if (/*enableConstantProp*/ optLevel >= 3 || enableConstantProp) 
```
is sufficient and easier to read. Please change here and where this pattern repeat.

src/Compiler/CompilerPasses.cpp
```diff
@@ -84,8 +84,13 @@ void addONNXToMLIRPasses(mlir::PassManager &pm, bool targetCPU) {
   }
   // There are more opportunities for const propagation once all tensors have
   // inferred shapes.
-  pm.addNestedPass<func::FuncOp>(onnx_mlir::createConstPropONNXToONNXPass());
 
+  // We want to enable Constant Propagation only for Level O3 or when a user
+  // manually specifies the "enable-constant-prop" flag.
+  if ((/*enableConstantProp*/ optLevel >= 3 && enableConstantProp) ||
```
AlexandreEichenberger, line 90: I think you meant this:
```
if (optLevel >= 3 || enableConstantProp)
```

src/Compiler/CompilerPasses.cpp
```diff
@@ -96,19 +101,23 @@ void addONNXToMLIRPasses(mlir::PassManager &pm, bool targetCPU) {
     for (int i = 0; i < repeatOnnxTransform; i++) {
       pm.addPass(mlir::createCanonicalizerPass());
       pm.addNestedPass<func::FuncOp>(onnx_mlir::createShapeInferencePass());
-      pm.addNestedPass<func::FuncOp>(
-          onnx_mlir::createConstPropONNXToONNXPass());
+      if ((/*enableConstantProp*/ optLevel >= 3 && !enableConstantProp) ||
```
AlexandreEichenberger, line 104: please simplify as above

src/Transform/ONNX/ONNXOpTransformPass.cpp
AlexandreEichenberger: @hamptonm1 please remove direct references to `OptimizationLevel`.

You can simply add a parameter to the constructor
```
  ONNXOpTransformPass(int threshold, bool report, bool targetCPU,
      bool enableSimdDataLayoutOpt) {
```

so that that object only depend on the local variables.

I believe this is what @sorenlassen asked for.

src/Transform/ONNX/SimplifyShapeRelatedOps.cpp
```diff
@@ -517,7 +518,9 @@ void SimplifyShapeRelatedOpsPass::runOnOperation() {
   for (unsigned i = 0; i < 3; ++i) {
     topDownShapeSimplification(context, moduleOp);
     OpPassManager pm("builtin.module");
+    if ((/*enableConstantProp*/ onnx_mlir::OptimizationLevel >= 3 && !onnx_mlir::enableConstantProp) || onnx_mlir::enableConstantProp) {
```
AlexandreEichenberger, line 521: simplify the logic as mentioned before, and add a local parameter optLevel that is set when constructing this object.

src/Compiler/CompilerPasses.cpp
hamptonm1: Nope that is not what I meant... I need to update this but will explain if need be.

src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
hamptonm1: That will not cover all paths... I am happy to talk if need be.

src/Compiler/CompilerPasses.cpp
hamptonm1: The logic I have covers two cases 1) the case of when a user simply specifies the flag on the command line to enable constant propagation  and 2) when we a use Level 03 and the default flag which is set to false. So what you are proposing simply will not cover two cases but rather one.

src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
tungld: This simplified condition does not allow to manually disable constant propagation at -O3.

hamptonm1: For instance if we were to apply the simplification that means users are no longer forced to use content propagation at Level 03.... they are now able to turn it off if they want to. Please look at the explanation below:

```
meganhampton@Megans-MacBook-Pro-2 build % ./Debug/bin/onnx-mlir --O1 --EmitLib roberta-base-11.onnx --onnx-op-stats TXT --enable-constant-prop=false --shapeInformation 0:1x1 -v 
The ONNX model has 124645716 elements in 
```

This will allow a user to disable constant prorogation instead of enable.

src/Transform/ONNX/SimplifyShapeRelatedOps.cpp
hamptonm1: I can remove `onnx_mlir::OptimizationLevel` and use `optLevel` instead... I will use the enum like in NNPA.

src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
AlexandreEichenberger: I must be missing something:

```
-O3 -enable=false  -> (O3 && !enable) || enable = true  || false = true
-O3 -enable=true   -> (O3 && !enable) || enable = false || true = true
-O0 -enable=false  -> (O3 && !enable) || enable = false || false = false
-O0 -enable=true   -> (O3 && !enable) || enable = false || true = true
```

vs 

```
-O3 -enable=false  -> O3 || enable = true
-O3 -enable=true   -> O3 || enable = true
-O0 -enable=false  -> O3 || enable = false
-O0 -enable=true   -> O3 || enable = true
```

Please let me know where I went wrong @tungld @hamptonm1 .

I do find it a bit confusing that setting explicitly `-enable-constant-prop=false` will still result in constant prop to be run. Maybe a better name is `force-constant-prop`.

hamptonm1:
@sorenlassen I reverted the other files I touched before and removed the `if-conditional` logic. I left all updates in `ConstProp.cpp`. @AlexandreEichenberger I just used your simplified logic for now so we can merge this PR in.... I will explain to you offline what myself and @tungld had in mind. If this looks good... please provide feedback or approve. Thanks!

src/Compiler/CompilerPasses.cpp
```diff
@@ -45,8 +45,8 @@ namespace onnx_mlir {
 void configurePasses() {
   // Set global vector machine support.
   VectorMachineSupport::setGlobalVectorMachineSupport(march, mcpu, "");
-  configureConstPropONNXToONNXPass(
-      onnxConstPropExpansionBound, onnxConstPropDisablePatterns);
+  configureConstPropONNXToONNXPass(onnxConstPropExpansionBound,
+      onnxConstPropDisablePatterns, OptimizationLevel, enableConstantProp);
```
sorenlassen, line 49: thanks, this solution satisfies my concerns about dependencies

fwiw, configureConstPropONNXToONNXPass could potentially omit the optLevel parameter if you passed a single value `(OptimizationLevel >= 3 || enableConstProp)` for constantPropIsEnabled here - that way ConstPropONNXToONNXPass wouldn't need to know the details of the policy for enabling/disabling it

@AlexandreEichenberger please go ahead and approve if your concerns are addressed

src/Transform/ONNX/ConstProp.cpp
```diff
@@ -63,10 +63,14 @@ namespace {
...
   static int expansionBound;
   static StringSet<> disabledPatterns;
+  static int optLevel;
+  static bool constantPropIsEnabled;
 };
 
 int ConstPropONNXToONNXPassConfiguration::expansionBound = -1; // -1 == no bound
 StringSet<> ConstPropONNXToONNXPassConfiguration::disabledPatterns;
+int ConstPropONNXToONNXPassConfiguration::optLevel;
+bool ConstPropONNXToONNXPassConfiguration::constantPropIsEnabled;
```
sorenlassen, line 73: please assign explicit defaults to these, like expansionBound, so it's clear what happens if configureConstPropONNXToONNXPass() isn't called

hamptonm1: okay I can update it

AlexandreEichenberger (commented):
Please address @sorenlassen 's init request, and name change; its all good then.

src/Transform/ONNX/ConstProp.cpp
```diff
@@ -86,6 +90,16 @@ bool satisfiesExpansionBound(Value result) {
          getSizeInBytes(resultType);
 }
 
+// We want to enable Constant Propagation only for Level O3 or when a user
+// manually specifies the "enable-constant-prop" flag.
+bool enableConstantPropagation() {
```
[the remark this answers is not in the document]
hamptonm1: Okay I can change it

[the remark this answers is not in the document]
hamptonm1: Okay updates completed

src/Compiler/CompilerPasses.cpp
AlexandreEichenberger: I missed @sorenlassen suggestion, and in fact I like it. Can we remove `OptimizationLevel` from the `ConstPropONNXToONNXPassConfiguration` object, simply call  here
```
 configureConstPropONNXToONNXPass(onnxConstPropExpansionBound,
       onnxConstPropDisablePatterns, OptimizationLevel>=3 ||  enableConstantProp);
```

as really, the const propagation pass does not need to be aware of if/when -O flags matter. That way, all this logic is neatly hidden in one spot.

Last chance, promised.

hamptonm1: @AlexandreEichenberger Done

src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
AlexandreEichenberger: I agree with that @tungld , so we need to have an optional; if `enable` is given by user, use that. Otherwise, use the -O3.

Can you give an example of a boolean option that is has no default, and how to test if the option is defined, and whether its true or false?

AlexandreEichenberger (approved):
LGTM, @tungld I believe this PR does not disable const prop at O3, but maybe we can add this as a separate PR.

PR #2499, opened 2023-09-13
Decompose ConstantOfShape into Expand and some rewriting rules

tungld:
This patch decomposes `ConstantOfShape {value}` into `Expand (Constant {value})`, and some canonicalization rules for Unsqueeze, Add and Sub operations when their inputs are from Expand (I use canonicalization because rules become now simpler).

Patterns are found in the roBERTa model, that is 
```mlir
%102 = onnx.Constant dense<1.000000e+00> : tensor<f32>
%367 = onnx.ConstantOfShape(%366) {onnx_node_name = "ConstantOfShape_9", value = dense<1.000000e+00> : tensor<1xf32>} : (tensor<2xi64>) -> tensor<?x?xf32>
// These two Unsqueezes can be folded into a new onnx.ConstantOfShape with value 1.0.
%370 = "onnx.Unsqueeze"(%367, %104) : (tensor<?x?xf32>, tensor<1xi64>) -> tensor<?x1x?xf32>
%371 = "onnx.Unsqueeze"(%370, %103) : (tensor<?x1x?xf32>, tensor<1xi64>) -> tensor<?x1x1x?xf32>
// This Sub can be folded into onnx.ConstantOfShape with value 0 since %102 and %371 are 1.0.
%372 = "onnx.Sub"(%102, %371) {onnx_node_name = "Sub_18"} : (tensor<f32>, tensor<?x1x1x?xf32>) -> tensor<?x1x1x?xf32>  
// Multiply by 0
%373 = "onnx.Mul"(%372, %101) {onnx_node_name = "Mul_20"} : (tensor<?x1x1x?xf32>, tensor<f32>) -> tensor<?x1x1x?xf32> 
// Add to 0
%418 = "onnx.Add"(%417, %373) {onnx_node_name = "Add_91"} : (tensor<?x12x?x?xf32>, tensor<?x1x1x?xf32>) -> tensor<?x12x?x?xf32>
```


src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -874,6 +874,78 @@ class PowToMulRewritePattern : public OpRewritePattern<ONNXPowOp> {
   int64_t maxPower;
 };
 
+// Rewrite a pattern like the following:
+//
+// %shape = onnx.Concat(%dim1, %dim2)
```
AlexandreEichenberger, line 879: A question that I believe I know the answer, but just to double check.

The pattern that is optimized here is really `expand -> unsqueezed`. Because our current approach is to always have a concat of the dimensions (that includes one or more runtime dims), then we will always find a `concat` of the shape, correct?
And if the shape itself is a constant, then constant propagation will anyway get rid of all 3 and generate one big constant for `%u`, so the constant shape is not relevant to this pattern.

Thanks for double checking.

src/Dialect/ONNX/Rewrite.td
```diff
@@ -390,6 +394,77 @@ def FuseMulConvNullBiasPattern: Pat<
...
+>;
+def PropagateScalarConstantExpandForDiv2: Pat<
+  (ONNXDivOp (ONNXExpandOp
+                (ONNXConstantOp:$lhs $_, $_, $_, $_, $_, $_, $_, $_),
+                $shape),
+             (ONNXConstantOp:$rhs $_, $_, $_, $_, $_, $_, $_, $_)),
+  (ONNXExpandOp (ONNXDivOp $lhs, $rhs), $shape),
+  [(IsScalarTensor:$lhs), (IsScalarTensor:$rhs)]
+>;
+
```
AlexandreEichenberger, line 467: If you agree that this is true, maybe add a comment that "this pattern can be reproduce for all element wise operations".

@sorenlassen @tungld I see that you require both constants to be scalar, is that a necessary requirement?

src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -874,6 +874,78 @@ class PowToMulRewritePattern : public OpRewritePattern<ONNXPowOp> {
...
+
+    // Match
+    // 1. data is from ExpandOp, axes is from ConstantOp.
+    if (!definedBy<ONNXExpandOp>(data) || !definedBy<ONNXConstantOp>(axes))
+      return failure();
+    auto expandOp = cast<ONNXExpandOp>(data.getDefiningOp());
+    // 2. ExpandOp's input is a scalar tensor so that it's safe to use a new
+    // shape that do not violate the broadcasting rule..
+    if (!isScalarTensor(expandOp.getInput()))
+      return failure();
```
sorenlassen, line 910: good catch, I hadn't thought about the broadcasting rule for Expand

fwiw, maybe you could generalize the pattern by rewriting Unsqueeze(Expand(X,Concat(A,B)) to Expand(Unsqueeze(X),Concat(A,B,1)), and then rely on const prop to deal with Unsqueeze(X) when X is constant

src/Dialect/ONNX/Rewrite.td
sorenlassen: in, e.g., Sub(C1,Expand(C2,shape)), if C1 is not scalar then you don't know how it will play together with shape - for instance, it could be that shape broadcasts to C1's larger shape and not the other way around

it's probably possible to define more lax conditions but it makes things more complicated

it might be safe to do the transform to Expand(Sub(C1,C2),shape) if C2 is not scalar

@tungld I think you need to restrict isScalarTensor to requite rank=0 (or find another workaround), otherwise the Sub example breaks if C1 has rank 1 and C2 and shape have rank 0

fwiw, if you want to instantiate this pattern for more ops without code duplication, you could use a class like FuseTransposeAndElementwiseUnaryOpPattern in the old Rewrite.td

src/Dialect/ONNX/Rewrite.cpp
tungld: Yes, you are right, but we still double-check if the shape is really from concat (it's maybe the case when SimplifiedShapeRelatedOp has not yet called and we still have ShapeOp instead of Concat. Perhaps, in the future we will always decompose ShapeOp into Concat), so that we can get each dim and insert unsqueezing dims.

src/Dialect/ONNX/Rewrite.td
tungld: > otherwise the Sub example breaks if C1 has rank 1 and C2 and shape have rank 0

I have a lit test in this PR for this case, it works when applying constprop, e.g.

Input:
```mlir
func.func @test_replace_sub_by_expand_2(%arg0: tensor<?xi64>) -> tensor<2x?xf32> {
    %0 = onnx.Constant dense<2.000000e+00> : tensor<f32>
    %1 = onnx.Constant dense<5.000000e+00> : tensor<1xf32>
    %2 = onnx.Constant dense<2> : tensor<1xi64>
    %3 = "onnx.Dim"(%arg0) {axis = 0 : si64} : (tensor<?xi64>) -> tensor<1xi64>
    %4 = "onnx.Concat"(%2, %3) {axis = 0 : si64} : (tensor<1xi64>, tensor<1xi64>) -> tensor<2xi64>
    %5 = "onnx.Sub"(%1, %0) : (tensor<1xf32>, tensor<f32>) -> tensor<1xf32>
    %6 = "onnx.Expand"(%5, %4) : (tensor<1xf32>, tensor<2xi64>) -> tensor<2x?xf32>
    return %6 : tensor<2x?xf32>
  }
```

output when doing constprop:
```mlir
func.func @test_replace_sub_by_expand_2(%arg0: tensor<?xi64>) -> tensor<2x?xf32> {
    %0 = onnx.Constant dense<3.000000e+00> : tensor<1xf32>
    %1 = onnx.Constant dense<2> : tensor<1xi64>
    %2 = "onnx.Dim"(%arg0) {axis = 0 : si64} : (tensor<?xi64>) -> tensor<1xi64>
    %3 = "onnx.Concat"(%1, %2) {axis = 0 : si64} : (tensor<1xi64>, tensor<1xi64>) -> tensor<2xi64>
    %4 = "onnx.Expand"(%0, %3) : (tensor<1xf32>, tensor<2xi64>) -> tensor<2x?xf32>
    return %4 : tensor<2x?xf32>
  }
```

tungld: I can generalize it to a common pattern.

src/Dialect/ONNX/Rewrite.cpp
tungld: Unsqueeze has axis and I believe we need to have a new axis for new input (X in `Expand(Unsqueeze(X),Concat(A,B,1))`). Maybe we can generalize it later if there is a need.

src/Transform/ONNX/Decompose.td
```diff
@@ -511,4 +511,14 @@ def ConvTransposeOpPattern2: Pattern<
...
 >;
 
+//===----------------------------------------------------------------------===//
+// Rewrite `ONNXConstantOfShapeOp {value} (%shape)` into
+//         `ONNXExpandOp(ONNXConstantOp {value}, %shape)
+//===----------------------------------------------------------------------===//
+
+def ConstantOfShapePattern: Pat<
+  (ONNXConstantOfShapeOp $shape, $value),
+  (ONNXExpandOp (ONNXConstantOpFromDenseAttr $value), $shape)
```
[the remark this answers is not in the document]
tungld: Sure, it can be done.

src/Dialect/ONNX/Rewrite.td
sorenlassen: the example I had in mind was
```
    %1 = onnx.Constant dense<1> : tensor<1xi32>
    %2 = onnx.Constant dense<2> : tensor<i32>
    %3 = onnx.Constant dense<> : tensor<0xi64>
    %4 = "onnx.Expand"(%2, %3) : (tensor<i32>, tensor<0xi64>) -> tensor<i32>
    %5 = "onnx.Sub"(%1, %4) : (tensor<1xi32>, tensor<i32>) -> tensor<1xi32>
```
it goes wrong if the PropagateScalarConstantExpandForSub1 pattern rewrites that to
```
    %1 = onnx.Constant dense<1> : tensor<1xi32>
    %2 = onnx.Constant dense<2> : tensor<i32>
    %3 = onnx.Constant dense<> : tensor<0xi64>
    %5 = "onnx.Sub"(%1, %2) : (tensor<1xi32>, tensor<i32>) -> tensor<1xi32>
    %6 = "onnx.Expand"(%5, %3) : (tensor<1xi32>, tensor<0xi64>) -> tensor<i32>
```
because the empty pattern `%3` cannot expand the shape `[1]`

tungld: What is the semantics of expand when shape is `tensor<0xi64`?

When I run your example with this patch, the output is
```
func.func @test_replace_sub_by_expand_rank0() -> tensor<1xi32> {
    %0 = onnx.Constant dense<1> : tensor<1xi32>
    %1 = onnx.Constant dense<2> : tensor<i32>
    %2 = onnx.Constant dense<> : tensor<0xi64>
    %3 = "onnx.Sub"(%0, %1) : (tensor<1xi32>, tensor<i32>) -> tensor<1xi32>
    %4 = "onnx.Expand"(%3, %2) : (tensor<1xi32>, tensor<0xi64>) -> tensor<1xi32>
    return %4 : tensor<1xi32>
  }
```

Note that the last op is the output type of `%4` is `tensor<1xi32` not `tensor<i32` as your output. I am not sure how you know the output type of `%6` is `tensor<i32` (`%6 = "onnx.Expand"(%5, %3) : (tensor<1xi32>, tensor<0xi64>) -> tensor<i32>`)

tungld: @AlexandreEichenberger @sorenlassen I created a C++ rewriting rule for binary ops instead of using tablegen. It can be extended to support new binary ops in the future.

sorenlassen: > What is the semantics of expand when shape is `tensor<0xi64`?

I haven't tested it but I assume that you can only expand the empty shape to the empty shape, i.e., Expand(X,shape=[]) only works if X has a rank 0 type

> I am not sure how you know the output type of `%6` is `tensor<i32 (%6 = "onnx.Expand"(%5, %3) : (tensor<1xi32>, tensor<0xi64>) -> tensor<i32>)`

sorry for being unclear, I should have omitted `-> tensor<i32>`, the point of the example is that the last line is breaks down because Expand(X,shape=[]) doesn't work when X has a rank 1 type

tungld:

src/Dialect/ONNX/Rewrite.td
tungld: OK, to avoid confusion, I exclude the case where Expand's shape is `tensor<0xdtype>`.

src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -282,6 +282,95 @@ class BinaryOpBroadcastAxisPattern : public OpRewritePattern<OP_TYPE> {
   }
 };
 
+// A pattern to turn
+//   `BinaryOp(Constant_X), ExpandOp(Constant_Y))`
```
sorenlassen, line 286: typo: unbalanced parentheses

src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -282,6 +282,95 @@ class BinaryOpBroadcastAxisPattern : public OpRewritePattern<OP_TYPE> {
...
+    Type outputType = op->getResult(0).getType();
+
+    // Match
+    //  - lhs is a scalar constant, and
+    //  - rhs is ExpandOp whose input is a scalar constant, or vice versa.
+    Value expandShape = nullptr;
+    auto matchValue = [&expandShape](Value v) -> Value {
+      Value res = v;
+      if (definedBy<ONNXExpandOp>(res)) {
+        auto expandOp = cast<ONNXExpandOp>(res.getDefiningOp());
```
sorenlassen, line 345: nit: this can be done without the extra type check in `cast` in one line
```
if (auto expandOp = dyn_cast_if_present<ONNXExpandOp>(res.getDefiningOp())) {
```

src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -282,6 +282,95 @@ class BinaryOpBroadcastAxisPattern : public OpRewritePattern<OP_TYPE> {
...
+        expandShape = expandOp.getShape();
+      }
+      if (isDenseONNXConstant(res) && isScalarTensor(res))
+        return res;
+      return nullptr;
+    };
+    Value lhsConstant = matchValue(lhs);
+    Value rhsConstant = matchValue(rhs);
+    if (!expandShape || !lhsConstant || !rhsConstant)
+      return failure();
```
sorenlassen, line 356: we should also return failure if lhs and rhs are both Expand results, right? (in that case the second shape overwrote the first in expandShape)

src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -282,6 +282,95 @@ class BinaryOpBroadcastAxisPattern : public OpRewritePattern<OP_TYPE> {
...
+    if (!hasShapeAndRank(expandShape))
+      return failure();
+    ArrayRef<int64_t> dims = getShape(expandShape.getType());
+    if ((dims.size() == 1) && (dims[0] == 0))
+      return failure();
+
+    // Rewrite
+    MultiDialectBuilder<OnnxBuilder> create(rewriter, loc);
+    Value res = create.onnx.expand(outputType,
+        EmitBinaryOp<OP_TYPE>(create, lhsConstant, rhsConstant), expandShape);
```
sorenlassen, line 367: if you clone binaryOp then you don't need EmitBinaryOp:
```
auto clone = rewriter.cloneWithoutRegions<OP_TYPE>(*op);
clone->setOperands({lhsConstant, rhsConstant});
clone.inferShapes([](mlir::Region &region) {}));
```
or
```
auto clone = create.onnx.createOpAndInferShapes<OP_TYPE>(lhsConstant, rhsConstant);
```
or
```
auto clone = rewriter.create<OP_TYPE>(lhsConstant, rhsConstant);
```
and rely on that all your ops have a builder that infers the result type (custom_builder_broadcast_to_same_type_ops_list in gen_onnx_mlir.py)

sorenlassen (approved):

src/Transform/ONNX/Decompose.td
```diff
@@ -60,6 +60,14 @@ def CreateNoneValue : NativeCodeCall<"$_builder.create<ONNXNoneOp>($_loc)">;
 def createScalarDenseAttrRank0
     : NativeCodeCall<"::onnx_mlir::createScalarDenseAttr($_builder, $0)">;
 
+// Create a scalar DenseElementsAttr of tensor<dtype> from an ElementsAttr.
+// The input ElementsAttr must have only one element. Otherwise only the first
+// element is used to create the scalar DenseElementsAttr
+def ReshapeElementsAttrToRank0 : NativeCodeCall<
+    "mlir::DenseElementsAttr::get(RankedTensorType::get({},"
+    " onnx_mlir::getElementType(cast<ElementsAttr>($0).getType())),"
+    " {*cast<ElementsAttr>($0).getValues<Attribute>().begin()})">;
```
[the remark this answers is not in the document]
tungld: Oh, I see. I know DenseElementsAttr has reshape but was not aware of it in DisposableElementsAttr. Thanks!

src/Dialect/ONNX/Rewrite.cpp
tungld: As I know this does not work if `res` is a BlockArgument then `res.getDefiningOp()` failed. `definedBy` checks BlockArgument.

sorenlassen: the BlockArgument case should work because then res.getDefiningOp() and dyn_cast_if_present return nullptr

src/Dialect/ONNX/Rewrite.cpp
tungld: I use `auto clone = create.onnx.createOpAndInferShapes<OP_TYPE>(lhsConstant, rhsConstant);`. Thanks!

src/Dialect/ONNX/Rewrite.cpp
tungld: Good catch. I checked it now and added a lit test for this. Thanks!

src/Dialect/ONNX/Rewrite.cpp
tungld: I see. It works, so moving to using `dyn_cast_if_present` now.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
```diff
@@ -388,6 +388,62 @@ class SplitLargeMatMulPattern : public OpRewritePattern<ONNXMatMulOp> {
   }
 };
 
+/// This pattern is to replace `C = add/sub(A, B)` by `A` when B is a zero
```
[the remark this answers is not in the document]
sorenlassen: actually, we could move Add/SubZerosOnRhs and the similar patterns (MulOnes, MatMulZeros, MatMulIntegerConstZero) to here, as they don't do conventional constant propagation and folding

in fact, they are getting in the way in the constant propagation pass as I'm restructuring constant propagation to support "lazy" delayed, symbolic constant folding

src/Dialect/ONNX/Rewrite.td
sorenlassen: sorry, please forget my example, I misunderstood how Expand works, I thought it's shape argument was the shape of the output, but on closer inspection I found that it broadcasts together with the input shape to form the output shape

in PR #2508 I added an Expand const prop lit test to illustrate this

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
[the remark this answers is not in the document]
tungld: I see. Only one input is a constant. It looks canonicalization can be another candidate. I put this pattern here because it requires DimAnalyis, so it's beneficial to NNPA only. We should find a new home for it, e.g. a new onnx-to-onnx transformation that utilizes DimAnalysis.

tungld:

AlexandreEichenberger:
@sorenlassen you have clearly a very in-depth knowledge of this topic; so when you are fine with the PR, I am good too.

PR #2501, opened 2023-09-13
Handle sequence_lens for GRU (NNPA)

chentong319:
The NNPA library does not support sequence_lens for GRU. This PR fixes the value after the GRU library call.
1. When ONNX to ZHigh, add code to fix the value. Since the fix operation can not be expressed with ONNX ops without LoopOp, CustomOp is used.
2. Then CustomOp is lowered to Knrl IR


src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.cpp
```diff
@@ -101,6 +101,28 @@ Value getLSTMGRUGetY(
...
+
+  // Create a CustomOp to fix Y because zDNN GRU ignores the sequence_lens
+  std::vector<Value> inputs = {val, sequenceLens, initialH};
+  ONNXCustomOp customOp =
+      rewriter.create<ONNXCustomOp>(loc, resY.getType(), inputs);
+  StringAttr funcNameAttr = rewriter.getStringAttr("FixGRUY");
+  customOp->setAttr("function_name", funcNameAttr);
+  StringAttr shapeInferAttr = rewriter.getStringAttr("SameAs");
+  customOp->setAttr("shape_infer_pattern", shapeInferAttr);
+  return customOp.getResults()[0];
```
tungld, line 123: Can we provide a new ZHigh Op, e.g. `ZHighLSTMGRUGetYWithSequenceLens`, instead of calling ONNXCustomOp here? The reason is that this fix is for NNPA only, and it's weird to expose the fix inside the lowering of CustomOp.

cjvolzka:
I gave the PR a test run today. Results look good for forward GRU. I had some issues with bidir GRU though. That last half of the second output tensor (h_n) came back all zeros.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.cpp
chentong319: Sure. CustomOp is a quick way to prototype. I will fix the bidirectional first.

cjvolzka:
We should also update the limitations for GRU and LSTM at https://github.com/onnx/onnx-mlir/blob/main/docs/SupportedONNXOps-NNPA.md as a part of this PR.

chentong319:
@cjvolzka @tungld I made the changes according to your suggestions. Please review the PR again.

src/Conversion/ONNXToKrnl/Additional/Custom.cpp
```diff
@@ -20,6 +20,84 @@ using namespace mlir;
 
 namespace onnx_mlir {
 
+void FixGRUY(Location loc, ConversionPatternRewriter &rewriter,
```
tungld, line 23: Is it still necessary? I think it's redundant.

src/Conversion/ONNXToKrnl/Additional/Custom.cpp
```diff
@@ -20,6 +20,84 @@ using namespace mlir;
...
+          initial = createKrnl.load(initialH, {directionIV, bs, hs});
+        }
+        Value cond = createMath.sge(
+            createMath.cast(sequenceUB.getType(), sequenceIV), sequenceUB);
+        Value newV = createMath.select(cond, /*padding*/ initial, currentV);
+        createKrnl.store(newV, output, indices);
+      });
+}
+
+void FixGRUYh(Location loc, ConversionPatternRewriter &rewriter,
```
tungld, line 65: Same here.

tungld (changes requested):
Please undo changes in Custom.cpp since they are not necessary I believe.

src/Conversion/ONNXToKrnl/Additional/Custom.cpp
chentong319: Yes, they are redundant.

src/Accelerators/NNPA/Conversion/ZHighToZLow/ZHighToZLow.cpp
```diff
@@ -1220,6 +1220,136 @@ struct ZHighToZLowGRUOpLowering : public ConversionPattern {
   }
 };
 
+//===----------------------------------------------------------------------===//
+// Lower ZHigh FixGRUY to Krnl
+//===----------------------------------------------------------------------===//
+
+struct ZHighToZLowFixGRUYOpLowering : public ConversionPattern {
+  ZHighToZLowFixGRUYOpLowering(TypeConverter &typeConverter, MLIRContext *ctx)
```
chentong319, line 1228: I have a new implementation with better performance for this function. 
1. Reuse the input buffer, instead of creating a new buffer.
2. Change the loop structure. Iterate the batch first and load sequence_lens[bs]. Then the lower bound of the sequence loop will be min(sequence_lens[bs], 0). We will just fill the elements unconditionally with initial value.

The two optimizations are unrelated. I am testing them for performance. The current implement is correct and simple in code.

tungld: Got it. Feel free to include it here or another PR.

tungld (approved):
LGTM. Thanks for solving the issue quickly!

PR #2504, opened 2023-09-14
Bump LLVM/MHLO

philass:
The goals of the LLVM rotation were to
- Keep the LLVM dependency up to date
- Sync on a commit that matches other projects

The torch-mlir community has a weekly cadence.

Unfortunately LLVM didn't get updated for about a month. When it did get updated it was on one of the weeks that we don't uplift (this week). 

I've done this second bump for this week so we can have a matching LLVM commit between both `torch-mlir` and `onnx-mlir` for the first time in over a month.

Specifically matching this parallel commit in `torch-mlir` https://github.com/llvm/torch-mlir/pull/2454


sorenlassen (approved):
LGTM

PR #2506, opened 2023-09-14
remove src/Compiler dependencies from src/{Builder,Transform,Conversion}

sorenlassen:
to make it easy to use each of the dialects and passes as libraries that people can link into their own compiler binaries without using the onnx-mlir binary, and to avoid unnecessary cyclical dependencies, we should avoid making the dialects and passes depend on anything in src/Compiler (with #include or cmake LINK_LIBS)

### DIFF
--- a/src/Builder/CMakeLists.txt
+++ b/src/Builder/CMakeLists.txt
@@ -11,7 +11,6 @@ add_onnx_mlir_library(OMBuilder
   ModelInputShaper.cpp
 
   LINK_LIBS PUBLIC
-  OMCompilerOptions
   OMHasOnnxSubgraphOpInterface
   OMONNXOps
   OMResultTypeInferenceOpInterface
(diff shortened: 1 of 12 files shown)


sorenlassen:

AlexandreEichenberger (approved):
LGTM, please run locally before merging.
````
ONNX_MLIR_FLAGS="-O3 -march=amd64" make check-onnx-backend
```
on a m1 Mac, or `-march=x86-64` on a x86 machine just to double check that removing the init of the vector model works. I tried it sometime ago without that code sequences and it had trouble, but maybe it is now working because I tried it wrong.

That test will enable SIMD which is unfortunately not yet set in our CIs

sorenlassen:
ok, I tried
```
ONNX_MLIR_FLAGS="-O3 -march=arm64" make check-onnx-backend
```
and it ran successfully for me on a M1 Mac

PR #2507, opened 2023-09-14
ONNX to Zhigh guided by cost model

AlexandreEichenberger:
Using Cedric's cost model to make a first approximation at only migrating to zAIU the ops that appears to be beneficial according to that cost model.

Enables using the `-enable-zhigh-cost-model` flag.

### DIFF
--- a/src/Accelerators/NNPA/Compiler/NNPACompilerOptions.cpp
+++ b/src/Accelerators/NNPA/Compiler/NNPACompilerOptions.cpp
@@ -48,4 +48,11 @@ llvm::cl::opt<bool> nnpaEnableZHighToOnnx("enable-zhigh-to-onnx",
         "level. Default is true."),
     llvm::cl::init(true), llvm::cl::cat(OnnxMlirOptions));
 
+llvm::cl::opt<bool> nnpaEnableZHighCostModel("enable-zhigh-cost-model",
+    llvm::cl::desc(
+        "Enabling a performance cost model to estimate the benefit of "
+        "migrating an eligible onnx operation to a ZHigh operation. Default is "
+        "false."),
+    llvm::cl::init(false), llvm::cl::cat(OnnxMlirOptions));
+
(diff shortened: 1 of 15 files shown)


AlexandreEichenberger:

AlexandreEichenberger:

chentong319:
Since we are going to add different control on the conversion, I suggest that we clear up the program structure first. 
1. For a conversion, define an analysis set. Currently, there is only DimAnalysis.
2. Decide which op will be converted with the current AddDynamicallyLegalOpFor<>(op, analysis_set, flags).  Basically to implement addDynamicallLegalA different name for the function may be used, such as decideOpToConvert
3. In the AddDynamicallyLegalOpFor (Actually target->addDynamicallyLegalOp<OP_TYPE>), we should unify the code for different Op

-      First check whether it is legal for the conversion. This is the isSuitableForZDNN(op) return return false.  return true. This Op can not be converted. I suggest change the name isSuitableForZDNN to isLegalForZDNN.
- Then use the analysis result to decide do we want to convert the op. The analysis result can be generated with a global view, or from file or command line. Drawback is the overhead on Op eliminated by other standard.
- (not sure the order) Check locally of the current Op for decision

4. The rewriting rule just decide HOW to convert, assuming that Ops have been checked and decided.

@AlexandreEichenberger @tungld Should we restructure the code before we add new code?

tungld:
@tong fyi #2510

tong:
> @tong fyi #2510

i am not the tong you are looking for.

AlexandreEichenberger:
>  I suggest that we clear up the program structure first.

You are making excellent points @chentong319, I will transform the code to separate benefit from validity (it was kind of separate already if we disable the benefit flag, but it is indeed better to have a stronger/clearer separation.

tungld:
@AlexandreEichenberger looks like this patch is replaced by #2534 that was landed.

PR #2508, opened 2023-09-15
Expand const prop lit test with short shape operand

sorenlassen:
To remind that Expand's shape can be shorter than the data input shape.


sorenlassen:
I forgot that Expand works this way while we discussed PR #2499

tungld (approved):
LGTM

PR #2509, opened 2023-09-15
Support --shapeInformation for .mlir input files

tungld:
This patch is to make `--shapeInformation` work with .mlir input files. It was only working with .onnx input files.

### DIFF
--- a/src/Compiler/CompilerUtils.cpp
+++ b/src/Compiler/CompilerUtils.cpp
@@ -14,6 +14,7 @@
 
 #include "CompilerUtils.hpp"
 
+#include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Dialect/LLVMIR/LLVMDialect.h"
 #include "mlir/Parser/Parser.h"
 #include "mlir/Pass/PassManager.h"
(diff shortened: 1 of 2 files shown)


tungld:
@gongsu832 it looks like Jenkins AMD64 has some connection issue. Do you have any idea on what happened?

AlexandreEichenberger (approved):
LGTM, will bring parity between .onnx and .mlir files (e.g. just after a -EmitONNXIR). Thanks

AlexandreEichenberger:

src/Compiler/CompilerUtils.cpp
```diff
@@ -183,6 +185,35 @@ static void loadMLIR(std::string inputFilename, mlir::MLIRContext &context,
...
+      argTy = modelInputShaper_.reshape(i, argTy);
+      // Update the arguments.
+      funcOp.getBody().back().getArgument(i).setType(argTy);
+      newArgTypes.emplace_back(argTy);
+    }
+    // Update the function type.
+    FunctionType newType =
+        FunctionType::get(&context, newArgTypes, funcType.getResults());
+    ConversionPatternRewriter rewriter(&context);
+    rewriter.updateRootInPlace(funcOp, [&] { funcOp.setType(newType); });
```
sorenlassen, line 215: I think you can just set funcOp's type directly without ConversionPatternRewriter and updateRootInPlace

AlexandreEichenberger:
@tungld works like a charm, thanks.

Issue #2481, opened 2023-09-05
Windows CI always builds llvm-project


gongsu832:
It appears that Windows CI is failing to find the cached and therefore always rebuilds llvm-project. This makes the build take almost 4 hours to finish.

```
2023-09-05T18:03:37.0764952Z ##[section]Starting: Check for mlir artifact
2023-09-05T18:03:37.0873571Z ==============================================================================
2023-09-05T18:03:37.0873717Z Task         : PowerShell
2023-09-05T18:03:37.0873784Z Description  : Run a PowerShell script on Linux, macOS, or Windows
2023-09-05T18:03:37.0873895Z Version      : 2.226.2
2023-09-05T18:03:37.0873963Z Author       : Microsoft Corporation
2023-09-05T18:03:37.0874041Z Help         : https://docs.microsoft.com/azure/devops/pipelines/tasks/utility/powershell
2023-09-05T18:03:37.0874145Z ==============================================================================
2023-09-05T18:03:38.0697494Z Generating script.
2023-09-05T18:03:38.1116486Z ========================== Starting Command Output ===========================
2023-09-05T18:03:38.1356537Z ##[command]"C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe" -NoLogo -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -Command ". 'D:\a\_temp\c6201e30-a2ee-4f59-abe5-3fd607debe63.ps1'"
2023-09-05T18:04:53.7329503Z Found build NotFound containing artifact MLIR_Windows_91088978d712cd7b33610c59f69d87d5a39e3113
2023-09-05T18:04:53.8304662Z ##[section]Finishing: Check for mlir artifact
```

gongsu832:
@MikeHolman will you or some other Microsoft folk be able to take a look at this? Thanks.

MikeHolman:
@jcwchen, can you take a look at this?

jcwchen:
It seems that this not found issue happened occasionally somehow? I did see there are few newer commits did find the cached artifact.

gongsu832:
Yes it does seem that the problem has gone away. Thanks for looking into it. We will keep an eye on it.

sorenlassen:
[...]

PR #2511, opened 2023-09-15
grew windows CI timeout

sorenlassen:
see @jcwchen advice in issue #2481

### DIFF
--- a/.azure-pipelines/Windows-CI.yml
+++ b/.azure-pipelines/Windows-CI.yml
@@ -11,9 +11,10 @@ parameters:
 
 jobs:
 - job: Build_onnx_mlir_Windows
-  timeoutInMinutes: 240
+  # 4h timeout is sometimes a tiny bit short when llvm-project is rebuilt
+  timeoutInMinutes: 270
   pool:
-    vmImage: 'windows-2019'
+    vmImage: 'windows-2022'
 
   steps:
   - checkout: self
--- a/.azure-pipelines/Windows-CI.yml
+++ b/.azure-pipelines/Windows-CI.yml
@@ -14,7 +14,7 @@ jobs:
   # 4h timeout is sometimes a tiny bit short when llvm-project is rebuilt
   timeoutInMinutes: 270
   pool:
-    vmImage: 'windows-2022'
+    vmImage: 'windows-2019'
 
   steps:
   - checkout: self


sorenlassen:
with windows-2022 I encountered this error in step "Install onnx-mlir" in Windows-CI.yml
```
-- Build files have been written to: D:/a/1/onnx-mlir/build
ninja: error: 'C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/DIA SDK/lib/amd64/diaguids.lib', needed by 'Release/bin/onnx-mlir.exe', missing and no known rule to make it
```

any suggestions how to overcome this are welcome

jcwchen (commented):
I guess the reason is the CI directly used cached llvm with windows-2022? However, that cached llvm was built with windows-2019. Building llvm from scratch with windows-2022 in this PR might help.

sorenlassen:
> Building llvm from scratch with windows-2022 in this PR might help.

ok, I changed `$mlir_tag` to not use the windows-2019 artifacts but then it fails building llvm-project:
```
[183/4217] Building IntrinsicsLoongArch.h...
FAILED: include/llvm/IR/IntrinsicsLoongArch.h D:/a/1/llvm-project/build/include/llvm/IR/IntrinsicsLoongArch.h 
cmd.exe /C "cd /D D:\a\1\llvm-project\build && D:\a\1\llvm-project\build\bin\llvm-min-tblgen.exe -gen-intrinsic-enums -intrinsic-prefix=loongarch -I D:/a/1/llvm-project/llvm/include/llvm/IR -ID:/a/1/llvm-project/build/include -ID:/a/1/llvm-project/llvm/include D:/a/1/llvm-project/llvm/include/llvm/IR/Intrinsics.td --write-if-changed -o include/llvm/IR/IntrinsicsLoongArch.h -d include/llvm/IR/IntrinsicsLoongArch.h.d"
```
it looks like llvm-min-tblgen.exe fails but I don't know what goes wrong or how to overcome that

jcwchen (commented):
Just found the latest llvm is still using windows-2019 in its CI: https://github.com/llvm/llvm-project/blob/bce6c53464c115f5f83a57284e409b1db8c2b03f/.github/workflows/llvm-project-tests.yml#L40-L42 and so it might not be compatible with newer windows-2022. Sorry I should confirm it earlier and thank you for trying it out. I guess the least thing we can do right now is simply increasing the timeout a bit?

sorenlassen:
> Sorry I should confirm it earlier and thank you for trying it out.

no problem

> I guess the least thing we can do right now is simply increasing the timeout a bit?

ok, I reverted the vmImage upgrade from this PR - please approve if you agree with the change

I would love if you can find another solution to make the Windows CI run faster - it is typically the slowest of the CI pipelines and therefore slows down onnx-mlir development

jcwchen (approved):

PR #2512, opened 2023-09-16
fix f16 onnx.BatchNormalizationInferenceMode canonicalization

sorenlassen:
new lit test test_rewrite_batchnormtestmode_1d_f16 crashed in createDenseElementsAttrFromFloatAttr because DenseElementsAttr::get(t, apfloat) wants apfloat to have the semantics of t's element type

createDenseElementsAttrFromFloatAttr was called on the result element type and `$epsilon` in the RewriteBatchNormInferenceModeConvPattern1 pattern


philass (approved):
Thanks for this fix

PR #2515, opened 2023-09-18
don't canonicalize to SqueezeV11, UnsqueezeV11

sorenlassen:
With this change it should be the case that no pattern introduces ops that are decomposed in DecomposeONNXToONNXPass and therefore they don't undo the work of that pass.

This will avoid unnecessary chains of patterns in the hybrid pass with decomposition (PR #2496). While chains of patterns should work, without them it is simpler to understand the interplay between sets of pattens.


src/Dialect/ONNX/ONNXOps.td.inc
```diff
@@ -9520,25 +9520,35 @@ def ONNXUnsqueezeOp:ONNX_Op<"Unsqueeze",
...
-      assert(sh && "failed to allocate shape helper");
-      return sh;
-    }
-  }];
+  let builders = [
+    OpBuilder<(ins "Value":$data, "Value":$axes), [{
+      auto resultType = UnrankedTensorType::get(data.getType().cast<ShapedType>().getElementType());
+      build($_builder, $_state, resultType, data, axes);
+    }]>,
+    OpBuilder<(ins "ValueRange":$operands, "ArrayRef<NamedAttribute>":$attributes), [{
```
tungld, line 9528: It looks redundant, the last Unsqueeze does not have attributes: https://onnx.ai/onnx/operators/onnx__Unsqueeze.html

tungld (commented):
The patch looks good. I see SqueezeV11 is still used in other places such as in the lowering of RNN, in onnx-to-zhigh, and onnx-to-krnl. I wonder if you have time to remove it totally (e.g. in onnx-to-zhigh) in this patch also.

sorenlassen:
> I see SqueezeV11 is still used in other places such as in the lowering of RNN, in onnx-to-zhigh, and onnx-to-krnl. I wonder if you have time to remove it totally (e.g. in onnx-to-zhigh) in this patch also.

I don't have the time, sorry

I tried onnx-to-zhigh but the lit test changes are too much - my implementation is here if you want to take it over: https://github.com/sorenlassen/onnx-mlir-einsum/commit/bdd701fd4b25f3a1ccd4698af30a0dd26b9f519d

onnx-to-krnl is too involved for me ('emitIntermediateIR' etc)

src/Dialect/ONNX/ONNXOps.td.inc
tungld: OK. I see. It's an issue in `gen_onnx_mlir.py`. Thanks for clarification!

tungld (approved):
LGTM.

PR #2518, opened 2023-09-20
Add a pass --set-onnx-node-name to set node name if onnx-node-name is missing 

tungld:
This patch introduces a new pass `--set-onnx-node-name` to set value for `onnx-node-name` attribute of an operation if the attribute is absent.

To make onnx-node-name of an operation easy to interpret, we first try to use the onnx-node-name of the defining operation of the first input if possible. In that case, onnx-node-node will be `firstInputOp's onnx-node-name + currentOp's opName + ID`.

If we can not get the onnx-node-name from the first input, we simply use `the currentOp's opName (e.g. onnx.Add) + ID`.

We also check if there is any duplication in the existing node names. If yes, we edit the existing node names to make them unique.


AlexandreEichenberger (commented):
There is already code to make names, see this
```
// We may try to relate the node names generated by the instrumentation
// with the node names printed by opt-report. Thus it is key to keep the
// code that generates these node name in sync.
//
// The code are found here:
// 1) `matchAndRewrite` from `src/Conversion/KrnlToLLVM/KrnlInstrument.cpp`
// 2) `getNodeNameInPresenceOfOpt` from
//    `src/Dialect/ONNX/ONNXOps/OpHelper.cpp`
```

I wonder if this pass could be used to create missing names. You may also look in how the names were created there, as @chentong319 (I believe) put some good thought into it.

tungld:
> I wonder if this pass could be used to create missing names. You may also look in how the names were created there, as @chentong319 (I believe) put some good thought into it.

Thanks for the suggestion! Just for my learning of your good thought. As I understand the best idea in `getNodeNameInPresenceOfOpt` is to get NameLoc that contains `onnx-node-name` created when importing a model, and NameLoc is propagated if we use (actually we are using) `ONNXLoc()` function in rewriting rules. Tablegen-based rules has not yet been using `ONNXLoc()`.

I now use `getNodeNameInPresenceOfOpt` as the first try to set names. But I have to disable using FileLineLoc since it produces a name containing the model file name (e.g. mode.mlir) and some symbols that are not friendly to users to use for device placement, e.g. `%1 = "onnx.Relu"(%0) {onnx_node_name = "onnx_set_node_name.mlir:20 offset :5_0"} : (tensor<3x2xf32>) -> tensor<3x2xf32>`.

src/Dialect/ONNX/ONNXOps/OpHelper.cpp
```diff
@@ -701,7 +701,14 @@ bool isScalarTensor(Value v) {
...
 //    `src/Dialect/ONNX/ONNXOps/OpHelper.cpp`
 
-std::string getNodeNameInPresenceOfOpt(Operation *op) {
+std::string getNodeNameInPresenceOfOpt(Operation *op, bool useFileLine) {
+  auto getNameFromFileLineLoc = [](FileLineColLoc loc, std::string &name,
+                                    std::string postfix = "") {
+    std::string filename =
+        llvm::sys::path::filename(loc.getFilename().str()).str();
+    name += filename + ":" + std::to_string(loc.getLine()) + postfix;
+  };
```
tungld, line 710: Create this lambda function to avoid duplicated code.

AlexandreEichenberger:
So can you confirm that both code are kept in sync? Or better yet, would you be able to unify the two so that we use `getNodeNameInPresenceOfOpt` in the `matchAndRewrite` from `src/Conversion/KrnlToLLVM/KrnlInstrument.cpp`?

The reason they need to generate the same names is so that I may relate the instrumentation names and the optimization names when generating reports.

tungld:
> The reason they need to generate the same names is so that I may relate the instrumentation names and the optimization names when generating reports.

I am sure you will get the same name in instrumentation and reports, because setting onnx_node_name here is at onnx level before any instrumentation is called. It is like we import a model and setting onnx_node_name. Thus, instrumentation (or the function `getNodeNameInPresenceOfOpt`) is always getting name from onnx_node_name first if available. In other words, setting onnx_node_name has no conflicts with instrumentation and reports.

AlexandreEichenberger (approved):
Thanks, LGTM

PR #2521, opened 2023-09-21
Bump LLVM commit to f66cd9e9556a53142a26a5c21a72e21f1579217c 

flemairen6:
Bump LLVM to f66cd9e9556a53142a26a5c21a72e21f1579217c 
I haven't bumped MHLO in absence of a green commit. Should I try to bump it to a closer commit, or should we just disable MHLO as suggested in the thread?


philass (approved):
LGTM! Thanks @flemairen6

PR #2527, opened 2023-09-22
Updates to RunONNXModel.py for quartile analysis, analyze-simd.py for default name with tags

AlexandreEichenberger:
Added mesurement only on middle 2 quartile, default for simd analysis… with tags


utils/RunONNXModel.py
```diff
@@ -646,14 +652,19 @@ def main():
             end = time.perf_counter()
             elapsed = end - start
             perf_results += [elapsed]
-            print("  {} iteration: {} seconds".format(ordinal(i+1), elapsed))
+            print("  {} iteration, {}, seconds".format(ordinal(i+1), elapsed))
```
AlexandreEichenberger, line 655: added `,` to generates CSV files that can be scanned by Pandas without further preprocessing.

tungld (approved):
LGTM.

PR #2529, opened 2023-09-25
don't crash decompose pass if Einsum can't decompose

sorenlassen:
We decompose as the first compiler pass (maybe because that's sometimes needed for shape inference?) but that breaks in several cases for Einsum, which is marked illegal and this is required to decompose: it needs static shapes for all inputs and the output, which can require a prior shape inference pass.

Now that the hybrid transform can decompose after PR #2496 it's best to not fail decomposition because there's a chance that shape inference (maybe in conjunction with other transformations in the hybrid transform) can infer the shapes later, and then a later run of the hybrid transform can decompose Einsum.

With this change I was able to compile a version of XLNet (https://huggingface.co/xlnet-large-cased) with fixed batch size and sequence length. That model makes extensive use of the hybrid transform to combine shape inference, canonicalization, constant propagation, and decomposition to decompose many Einsum ops with a long chain of dependency on all these transformations.


philass (approved):
LGTM! Thanks

PR #2530, opened 2023-09-25
replace ONNXOpTransformPass with hybrid transform

sorenlassen:
replaces ONNXOpTransformPass and the repeatOnnxTransform loop in addONNXToMLIRPasses() with hybrid transform, unless disabled with `--onnx-hybrid-pass=false`

set maxNumRewrites in onnx-hybrid-transform to 300 and tested that the model https://huggingface.co/xlnet-large-cased succeeds without warning with this value

prints warning to stderr if onnx-hybrid-transform doesn't converge with the given maxNumRewrites

I tested this PR on the models in the model zoo and the same models passed and failed as with main

possible future work:
- replace const-prop, shape-inference, canonicalizer passes in SimplifyShapeRelatedOpsPass with hybrid transform
- replace some shape-inference/const-prop/canonicalizer passes in addONNXToZHighPasses() with hybrid transform

### DIFF
--- a/src/Transform/ONNX/ONNXHybridTransformPass.cpp
+++ b/src/Transform/ONNX/ONNXHybridTransformPass.cpp
@@ -23,6 +23,7 @@
 #include "src/Interface/ShapeInferenceOpInterface.hpp"
 #include "src/Pass/Passes.hpp"
 #include "src/Transform/ONNX/ConstProp.hpp"
+#include "src/Transform/ONNX/ConvOpt.hpp"
 #include "src/Transform/ONNX/Decompose.hpp"
 #include "src/Transform/ONNX/ShapeInference.hpp"
 
@@ -67,6 +68,14 @@ struct ONNXHybridTransformPass
       llvm::cl::desc("Enable decomposition in hybrid transform"),
       llvm::cl::init(true)};
 
+  Option<bool> enableConvOpt{*this, "enable-conv-opt",
+      llvm::cl::desc("Enable convolution optimization for CPU"),
+      llvm::cl::init(false)};
+
+  Option<bool> enableSimdDataLayoutOpt{*this, "simd-data-layout",
+      llvm::cl::desc("Enable SIMD data layout optimizations"),
+      llvm::cl::init(false)};
+
   FrozenRewritePatternSet patterns;
 
   ONNXHybridTransformPass() = default;
@@ -74,6 +83,11 @@ struct ONNXHybridTransformPass
   ONNXHybridTransformPass(const ONNXHybridTransformPass &pass)
       : patterns(pass.patterns) {
     copyOptionValuesFrom(&pass);
+  }
+
+  ONNXHybridTransformPass(bool enableConvOpt, bool enableSimdDataLayoutOpt) {
+    this->enableConvOpt = enableConvOpt;
+    this->enableSimdDataLayoutOpt = enableSimdDataLayoutOpt;
   }
 
   StringRef getArgument() const override { return "onnx-hybrid-transform"; }
@@ -101,6 +115,10 @@ struct ONNXHybridTransformPass
       getDecomposeONNXToONNXPatterns(cumulativePatterns);
     }
 
+    if (enableConvOpt) {
+      getConvOptONNXToONNXPatterns(enableSimdDataLayoutOpt, cumulativePatterns);
+    }
+
     patterns = FrozenRewritePatternSet(std::move(cumulativePatterns));
     return success();
   }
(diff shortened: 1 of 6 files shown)


sorenlassen:
set maxNumRewrites in onnx-hybrid-transform to 300 and tested that


I tested the model https://huggingface.co/xlnet-large-cased with different maxNumRewrites values and found that it fails with 100:
```
Warning: onnx-hybrid-transform didn't converge with max-num-rewrites=100
Warning: onnx-hybrid-transform didn't converge with max-num-rewrites=100
Assertion failed: (isRankedShapedType(ty) && "Type must be ranked"), function getShape, file TypeUtilities.cpp, line 42.
```
and with 200 it succeeds with one warning:
```
Warning: onnx-hybrid-transform didn't converge with max-num-rewrites=200
```
and with 300 it succeeds wiithout warnings

src/Transform/ONNX/ONNXHybridTransformPass.cpp
```diff
@@ -101,19 +119,27 @@ struct ONNXHybridTransformPass
...
 
     GreedyRewriteConfig config;
     config.useTopDownTraversal = true;
-    (void)applyPatternsAndFoldGreedily(body, patterns, config);
+    config.maxNumRewrites =
+        maxNumRewrites == -1 ? GreedyRewriteConfig::kNoLimit : maxNumRewrites;
+    if (failed(applyPatternsAndFoldGreedily(body, patterns, config))) {
+      llvm::errs() << "Warning: onnx-hybrid-transform didn't converge with "
+                      "max-num-rewrites="
+                   << maxNumRewrites.getValue() << "\n";
```
tungld, line 141: How did you know that it failed because of "not converge" but other errors like wrong patterns?

sorenlassen: The applyPatternsAndFoldGreedily returns false only if there are still patterns to apply when the rewrite limit is reached.

chentong319:
Question to @tungld : is the ConvOpt a really general onnx to onnx transformation that should be included into the hybrid pass? Should it be part of onnx to ZHigh lowering?

tungld:
> Question to @tungld : is the ConvOpt a really general onnx to onnx transformation that should be included into the hybrid pass? Should it be part of onnx to ZHigh lowering?

It's for CPU, and looks like a general transformation. @AlexandreEichenberger may have more comments as he wrote ConvOpt.

src/Transform/ONNX/ONNXHybridTransformPass.cpp
tungld: Thanks for the explanation! I found `applyPatternsAndFoldGreedily` internally emits the same error message using  LLVM_DEBUG: https://github.com/llvm/llvm-project/blob/main/mlir/lib/Transforms/Utils/GreedyPatternRewriteDriver.cpp#L731

tungld (approved):
LGTM. Would be good to have @AlexandreEichenberger commentsOn ConvOpt before merging.

AlexandreEichenberger:
> set maxNumRewrites in onnx-hybrid-transform to 300 and tested that
>I tested the model https://huggingface.co/xlnet-large-cased with different maxNumRewrites values and found that it fails with 100:

Can it be made proportional to the number of ops? That way if we have a very large model in the future, it should scale with it?

AlexandreEichenberger:
Info on the ConvOpt, it transforms a 1x1 Conv into a matmul, as it can be seamlessly implemented as such. I let you decide if it's appropriate for the hybrid transform or not. Its essentially a one-to-one substitution as the two are identical, and the matmul is well optimized whereas the Conv is not yet optimized (for CPU)

sorenlassen:
> > set maxNumRewrites in onnx-hybrid-transform to 300 and tested that
> > I tested the model https://huggingface.co/xlnet-large-cased with different maxNumRewrites values and found that it fails with 100:
> 
> Can it be made proportional to the number of ops? That way if we have a very large model in the future, it should scale with it?

done, it's now calculated as 20 + 0.2 * #ops

@AlexandreEichenberger @tungld I believe I've addressed all feedback and will merge the PR on Monday unless you have more input before then

AlexandreEichenberger (approved):
LGTM

chentong319 (approved):
LGTM!

src/Transform/ONNX/ONNXHybridTransformPass.cpp
```diff
@@ -67,6 +68,18 @@ struct ONNXHybridTransformPass
...
+      llvm::cl::init(false)};
+
+  Option<bool> simdDataLayout{*this, "simd-data-layout",
+      llvm::cl::desc("Enable SIMD data layout optimizations"),
+      llvm::cl::init(false)};
+
+  Option<int> maxNumRewrites{*this, "max-num-rewrites",
+      llvm::cl::desc("Limit the number of rewrites. -1 means no limit."),
+      llvm::cl::init(300)};
+
```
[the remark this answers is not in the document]
sorenlassen: no, but I welcome any ideas

fwiw, now that we agreed to remove ConvOpt we're back to the original vision of grouping the 4 classes of transformations (shape inference, canonicalization, const prop, and decomposition) originally proposed for the hybrid pass 1y ago in PR #1770, so I don't expect that we'll need to add a lot more controls going forward

PR #2533, opened 2023-09-26
Add NNPA level compatibility check

mikeessen:
Add NNPA level compatibility check to validate operations for the specified --mcpu value.


src/Accelerators/NNPA/Conversion/ONNXToZHigh/NNPALimit.h
```diff
@@ -28,3 +28,9 @@ static constexpr int64_t NNPA_MAXIMUM_TENSOR_SIZE = 4294967296;
 // See zDNN API doc
 static constexpr int64_t MAXIMUM_NUM_HIDDEN_SIZE_LSTM = 8192;
 static constexpr int64_t MAXIMUM_NUM_HIDDEN_SIZE_GRU = 10880;
+
+// The NNPA levels.
+static constexpr const char *NNPA_Z13 = "z13";
+static constexpr const char *NNPA_Z14 = "z14";
+static constexpr const char *NNPA_Z15 = "z15";
```
tungld, line 35: Do we use `NNPA_Z{13, 14, 15}` in practice?

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
```diff
@@ -15,13 +15,36 @@
...
+  } catch (...) {
+    retNNPAFloat = 0;
+  }
+  return retNNPAFloat;
+}
+
+/// A function to check whether the input NNPA level, ie. "z16", is compatible
+/// with the current NNPA level.
+bool isCompatibleWithNNPALevel(std::string inputNNPALevel) {
+  if (convertNNPALevel(inputNNPALevel) <= convertNNPALevel(mcpu)) {
```
tungld, line 42: For brevity, we can use `return convertNNPALevel(inputNNPALevel) <= convertNNPALevel(mcpu);`

BTW, a corner case, if both `convertNNPALevel(inputNNPALevel)` and `convertNNPALevel(cpu)` return 0 (because of some failure), this condition is `true` that is not what we want. Perhaps, you could check if `convertNNPALevel` succeeded or not before doing comparison.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
```diff
@@ -240,8 +263,13 @@ bool meetPoolParamRestrictions(int64_t inputShape, int64_t kernelShape,
 }
 
 /// Default legality check.
+/// This template is not directly called.
 template <typename OP_TYPE>
 bool isSuitableForZDNN(OP_TYPE op, const DimAnalysis *dimAnalysis) {
+  // Check NNPA level.
+  if (!isCompatibleWithNNPALevel(NNPA_Z16)) {
+    return false;
+  }
```
tungld, line 272: Looks redundant. Just returning `false` for unsupported ops is enough.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
```diff
@@ -250,6 +278,10 @@ bool isSuitableForZDNN(OP_TYPE op, const DimAnalysis *dimAnalysis) {
 template <>
 bool isSuitableForZDNN<ONNXAddOp>(
     ONNXAddOp op, const DimAnalysis *dimAnalysis) {
+  // Check NNPA level.
+  if (!isCompatibleWithNNPALevel(NNPA_Z16)) {
+    return false;
+  }
```
tungld, line 284: Could you please remove `{` and `}` for this simple if? Same for other locations.

tungld (commented):
@mikeessen thanks for the patch!

I see we have another locations that need `isCompatibleWithNNPALevel`, those are inside `addDynamicallyLegalOpFor` functions in https://github.com/onnx/onnx-mlir/blob/main/src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
cjvolzka: @tungld unless it's a standard in onnx-mlir to drop the `{}`'s for one line `if` statements, I lean towards keeping them. Functionally nothing changes but the braces are less error prone during maintenance or debugging scenarios.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/NNPALimit.h
mikeessen: Thanks, I removed the definitions for NNPA_Z{13, 14, 15}

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
mikeessen: Thanks, I updated for the corner case and also brevity.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
mikeessen: Thanks, I updated to just return false.

mikeessen:
@tungld I updated  the `addDynamicallyLegalOpFor` functions in https://github.com/onnx/onnx-mlir/blob/main/src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp with `isCompatibleWithNNPALevel`.  Please take a look.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
tungld: @cjvolzka we don't have a standard but we follow LLVM coding style: https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements

cjvolzka: Ok, good to know. I defer to the established norm for the repo on such things.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
```diff
@@ -499,6 +499,10 @@ void getRewriteONNXForZHighDynamicallyLegal(
   // broadcasting.
   addDynamicallyLegalOpFor<ONNXAddOp>(
       target, dimAnalysis, [](ONNXAddOp op, const DimAnalysis *dimAnalysis) {
+        // Check NNPA level.
+        if (!isCompatibleWithNNPALevel(NNPA_Z16)) {
+          return false;
```
tungld, line 504: @mikeessen I am sorry that the logic here is a bit confusing. `false` means the operation is illegal and must be rewritten to run on NNPA, while `true` means the operation is legal and will remain on CPU. So I think we should return true. Same for the other changes in this file.

It's contrary to the logic in `isSuitableForZDNN`. Actually it's like calling `! isSuitableForZDNN()`.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
mikeessen: Thanks, updated to remove { } from these statements.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
mikeessen: Thanks for the clarification.   I updated false to true.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
AlexandreEichenberger: The only places I think it's ok to have extra `{}` is in the presence of nested ifs, as dangling `else` are dangerous. Otherwise, best to keep to the LLVM formatting rules

AlexandreEichenberger:
@mikeessen Can you confirm: we now need the `-mcpu=z16` in order to have `-maccel=NNPA` take effect, or does the `-maccel=NNPA` grandfather the `-mcpu=z16` option.

cjvolzka:
> we now need the -mcpu=z16 in order to have -maccel=NNPA take effect,

This is my expectation. Somewhere llvm has to be using something as a default mcpu when one isn't specified. Something makes me think that's z13 but I might be off. Whatever that default is, it's not z16. So an unspecified mcpu means "use z13 (or whatever the llvm default is) instructions". NNPA instructions won't be available for anything but the latest processor.

So before this change, specifying only -maccel=NNPA, we would create a model that targeted (by implied default) for z13 which couldn't possibly run on z13. 

After this change, specifying only -maccel=NNPA, we compile a model that will run on z13 using all available NNPA instructions for z13 (which is zero).

AlexandreEichenberger:
@mikeessen would you mind checking the doc that all our examples that have NNPA also have the `-mcpu` option? Tx

mikeessen:
@AlexandreEichenberger Thanks, there are a couple examples which needed updating.  I created a PR to address: https://github.com/onnx/onnx-mlir/pull/2551

AlexandreEichenberger:
Thanks for the quick fixes on the doc :-)

PR #2535, opened 2023-09-27
Fix computing intermediate zero in DynamicQuantizeLinear

tungld:
@cjvolzka found that the emitted code for intermediate zero is inconsistent with its equation (https://github.com/onnx/onnx-mlir/pull/2129#discussion_r1336296760). Instead of emitting `interZeroPoint = qMin - (xMin / scale)`, the lowering emitted `interZeroPoint =  (qMin - xMin) / (scale)`.

This patch is to fix the above bug.


AlexandreEichenberger (approved):
LGTM, thanks for the fix

cjvolzka (approved):

PR #2536, opened 2023-09-27
[NNPA] Set device placement by using a JSON file

tungld:
This patch allows users to schedule device placement for NNPA by using a json file. Two new compiler options are added: 
- `--nnpa-load-device-placement-file=cfg.json` to load device placement from a json file.
- `--nnpa-save-device-placement-file=cfg.json` to save the device placement done by the compiler to a json file.

Some examples for a json file for `--nnpa-load-device-placement-file`. Note that we can use regular expression to match strings.

Input program:
```mlir
func.func @test_load_config_file_all_on_cpu(%arg0: tensor<?x?x?xf32>) -> tensor<?x?x?xf32> {
  %0 = "onnx.Relu"(%arg0) {onnx_node_name = "Relu_0"} : (tensor<?x?x?xf32>) -> tensor<?x?x?xf32>
  %1 = "onnx.Relu"(%0) {onnx_node_name = "Relu_1"} : (tensor<?x?x?xf32>) -> tensor<?x?x?xf32>
  %2 = "onnx.Relu"(%1) {onnx_node_name = "Relu_2"} : (tensor<?x?x?xf32>) -> tensor<?x?x?xf32>
  %3 = "onnx.Sigmoid"(%2) {onnx_node_name = "Sigmoid_0"} : (tensor<?x?x?xf32>) -> tensor<?x?x?xf32>
  onnx.Return %3 : tensor<?x?x?xf32>
```

1. Schedule all operations to run on CPU
```json
{
  "device_placement": [
    {
      "device": "cpu",
      "node_type": "onnx.*",
      "onnx_node_name": ".*"
    }
  ]
}
```

2. Schedule all Relu operations to run on CPU:
```json
{
  "device_placement": [
    {
      "device": "cpu",
      "node_type": "onnx.Relu",
      "onnx_node_name": ".*"
    }
  ]
}
```
3.  Schedule operations using onnx_node_name: here we use regex to chose only Relu_1 and Relu_2 operations, exact match is used for onnx.Sigmoid.
```json
{
  "device_placement": [
    {
      "device": "cpu",
      "node_type": "onnx.Relu",
      "onnx_node_name": "Relu_(1|2)"
    },
    {
      "device": "nnpa",
      "node_type": "onnx.Sigmoid",
      "onnx_node_name": "Sigmoid_0"
    }
  ]
}
```

The new change is able to handle what the old option `execNodesOnCpu` does. Thus, remove `execNodesOnCpu` in this patch.


AlexandreEichenberger:
@tungld this functionality will be a great help. Could you be a bit more specific about the regex. In your example
```
"onnx_node_name": "Relu_[1,2]"
```
that does not seems to match the normal expectations of matching either `Relu_1` or `Relu_2` as `Reel_,` seems to be also legal. Is that a typo, or you are using a different regexp here.

Also, is it a match from the beginning to the end (implied `^` and `$`) or its a match if it fits anywhere in the string.

Probably a md page to give a bit more explanation might be useful.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/DevicePlacement.cpp
```diff
@@ -13,13 +13,26 @@
...
 // - a cost model
 //
+// Device placement is done via setting `device` attribute for each operation.
+// Values for `device` attribute is one of the following strings:
+// - "": an empty string means the compiler decides whether the operation is on
+// CPU or NNPA.
+// - "nnpa": the operation may run on NNPA or CPU, and the final decision is
+// made by the compiler. If `device=nnpa` is the result of this device-placement
+// pass, then it means the compiler thinks it is suitable for NNPA.
+// - "cpu": the operation is guaranteed to run on CPU.
```
imaihal, line 23: This might not be appropriate question for this PR, but what is the difference between `device=""` and `device=nnpa`? I think compiler determines the device when either attribute is used. By specifying `device=nnpa`, can we get different results, or is this useful for users? I mean only providing `decice=""` and `device=cpu` attribute is not enough?

src/Accelerators/NNPA/Conversion/ONNXToZHigh/DevicePlacement.cpp
```diff
@@ -93,26 +181,101 @@ void DevicePlacementPass::runOnOperation() {
...
+                  .getValue()
+                  .str()
+            : "";
+    llvm::json::Value jsonObj = llvm::json::Object{
+        {DEVICE_KEY, deviceStr},
+        {NODE_TYPE_KEY, nodeTypeStr},
+        {ONNX_NODE_NAME_KEY, nodeNameStr},
+    };
+    jsonArr.emplace_back(jsonObj);
+  }
```
imaihal, line 267: Is it possible to save attributes to Json file only when the attribute is not ""? Should we save  the attribute ""?

src/Accelerators/NNPA/Conversion/ONNXToZHigh/DevicePlacement.cpp
AlexandreEichenberger: Had the same question, why do we even accept that?

tungld: - When users **setting** device placement via a JSON file, there is no difference between `device=""` and `device=nnpa`, because we are able to force to CPU at this moment only, meaning `device=cpu`. I hope users can set NNPA in the future  by `device=nnpa` and `device=""` means let the compiler decide, but I haven't had a good solution yet.

- When the compiler **prints out** its device placement, e.g. via `--EmitONNXIR` or `--save-device-placement-file`, `device=""` and `device="nnpa"` are completely different. `device="nnpa"` means the compiler decide to run the op on NNPA while `device=""` means there is no guarantee that the op runs on CPU or NNPA.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/DevicePlacement.cpp
tungld: Yes, we should because it helps users know the `node_type` and `onnx_node_name` of the operation, and use these information to decide whether they want the operation to run on CPU by changing `device` from "" to "cpu". Otherwise, users need some way to explore these information, e.g. using --EmitONNXIR.

tungld:
> that does not seems to match the normal expectations of matching either Relu_1 or Relu_2; Reel_, seems to be also legal. Is that a typo, or you are using a different regexp here.

Good catch! Thanks! My mistake, I changed the regular expression to `Relu_(1|2)`. I use C++ `std::regex_match` for string matching that uses the standard regular expression. I added a .md page to explain about device placement.

tungld:
@AlexandreEichenberger  @imaihal this patch is ready for review. Could you take another look? Thanks!

docs/DevicePlacement-NNPA.md
```diff
@@ -0,0 +1,128 @@
...
+      "onnx_node_name": "Relu_(1|2)"
+    },
+    {
+      "device": "nnpa",
+      "node_type": "onnx.Sigmoid",
+      "onnx_node_name": "Sigmoid_0"
+    }
+  ]
+}
+```
```
AlexandreEichenberger, line 128: Please states if the regex starts from the beginning, aka a `re.match` in python or equivalent of `$onnx.Relu`... you may want to check that if node type above is only `Relu` it fails but if it is `onnx.Relu` it succeeds.

Also, if you have 2 rules that fires... say one that takes all the "relu" and another one that takes all the ops on "line3"... and there are some overlap between the two sets, what happen.

AlexandreEichenberger (approved):
LGTM, maybe just add a bit more info on the regex behavior and its good to go.

docs/DevicePlacement-NNPA.md
tungld: I explained more about this in the document and added two examples for matching `Relu` and overlapping.

PR #2537, opened 2023-09-28
Remove Level Constraint for Constant Propagation 

hamptonm1:
I changed the flag from `enableConstantProp` to `disableConstantProp` so from now on the user would have to manually specify if they want to disable constant propagation. It is no longer tied to level 3 and the default behavior will be to always enable constant propagation unless a user feels otherwise. I was also able to renable the two backend tests that were failing.


AlexandreEichenberger (approved):
LGTM, thanks for the fix

sorenlassen (commented):
LGTM

PR #2538, opened 2023-09-29
Require Python code to be formatted by black, similar to clang-format for C/C++ code

gongsu832:


gongsu832:

docs/Workflow.md
```diff
@@ -295,3 +295,5 @@ Using the GitHub interface:
 Very small PRs are easy to review. Very large PRs are very difficult to review.
 
 Follow the [coding style](https://llvm.org/docs/CodingStandards.html) used by LLVM for your code. We use the `clang-format` command to get the proper style, which is also tested by our CIs. It is acceptable to run the command on all of the files that were modified by your PR. We recommend using VS code where the clang formatter will be run automatically using the clang format configuration file already present in the repository.
+
+For python code, we use the [black](https://pypi.org/project/black) code formatter. You should run the `black` command on all the python code modified by your PR, which must pass the black code formatter CI check before it can be merged.
```
chentong319, line 299: could you add the command line in the doc? With option `"--check --exclude third_party"`?

chentong319 (approved):
Thanks for your efforts!

docs/Workflow.md
gongsu832: That option is meant for the Jenkins CI to check all python files excluding third_party. Typically you only run black on the files you changed in your PR, as described. Usage of black is at the link provided.

gongsu832:
I don't use VSCode but you may be able to use this https://marketplace.visualstudio.com/items?itemName=ms-python.black-formatter

AlexandreEichenberger:
Thanks, I used a plugin and it worked perfectly. If there is a way to run the CI like the clang format where it list the failing files, that would be useful. Short of that, one can always look at the list of modified python scripts and manually format them in VSCode. Thanks for adding this new rule, it was getting a bit unruly.

gongsu832:
Not sure what you mean. The whole purpose of this PR is to have a github workflow check the python code format with black and fail if some of them are not formatted, just like the clang format.

AlexandreEichenberger:
What I mean is that the "details" of a failed clang format CI tells me which cpp file fails, but the "details" for the python just say "fail" without listing the failing files.

gongsu832:
image

Issue #2524, opened 2023-09-22
Failed to compile llama2-7b model due to "Type must be ranked" issue


chenqiny:
I tried to compile llama2-7b model.(models: https://huggingface.co/steerapi/Llama-2-7b-chat-hf-onnx)  
But I failed due to Type must be ranked isues.  
I am using ONNX-MLIR on x86 prebuild container:   
onnxmlirczar/onnx-mlir                                                             latest                 6053c88f8692   10 hours ago    1.9GB  
onnx-mlir version 0.4.2, onnx version 1.14.1 (https://github.com/onnx/onnx-mlir.git 2f1003b)  
LLVM version 18.0.0git (https://github.com/llvm/llvm-project.git f66cd9e9556a53142a26a5c21a72e21f1579217c)  
Optimized build with assertions.  
Default target: x86\_64-unknown-linux-gnu  
Host CPU: sandybridge

```
~/code/onnx-mlir/docker> ./onnx-mlir.py -O3 --EmitLib ~/code/Llama-2-7b-chat-hf-onnx/onnx/decoder_model_fp16.onnx
onnx-mlir: /workdir/onnx-mlir/src/Support/TypeUtilities.cpp:42: ArrayRef<int64_t> onnx_mlir::getShape(mlir::Type): Assertion `isRankedShapedType(ty) && "Type must be ranked"' failed.

~/code/onnx-mlir/docker> ./onnx-mlir.py -O3 --EmitLib ~/code/Llama-2-7b-chat-hf-onnx/onnx/decoder_model_merged.onnx
onnx-mlir: /workdir/onnx-mlir/src/Support/TypeUtilities.cpp:42: ArrayRef<int64_t> onnx_mlir::getShape(mlir::Type): Assertion `isRankedShapedType(ty) && "Type must be ranked"' failed.
~/code/onnx-mlir/docker> ./onnx-mlir.py -O3 --EmitLib ~/code/Llama-2-7b-chat-hf-onnx/onnx/decoder_model.onnx
onnx-mlir: /workdir/onnx-mlir/src/Support/TypeUtilities.cpp:42: ArrayRef<int64_t> onnx_mlir::getShape(mlir::Type): Assertion `isRankedShapedType(ty) && "Type must be ranked"' failed.
~/code/onnx-mlir/docker> ./onnx-mlir.py -O3 --EmitLib ~/code/Llama-2-7b-chat-hf-onnx/onnx/decoder_model_merged_quantized.onnx
[...]

PR #2539, opened 2023-09-29
Shape inference for Reshape if the data is unranked tensor

chentong319:
We can get the rank of the output tensor if the shape() has static shape. This could be useful in the middle of iteration. 
I didn't change the symbolic computation of dims in computeShape() because unranked tensor is not allowed when the expression is used in lowering.
The error is mentioned by #2524 .


src/Dialect/ONNX/ONNXOps/Tensor/Reshape.cpp
```diff
@@ -116,10 +118,12 @@ LogicalResult ONNXReshapeOp::verify() {
...
 LogicalResult ONNXReshapeOp::inferShapes(
     std::function<void(Region &)> doShapeInference) {
-  // Cannot infer shape without data rank or static shape of shape.
+  // Cannot infer shape without data rank and static shape of shape.
+  // If shape is constant shape, the rank of the output known.
+  // This step may be helpful to reach the fix point.
   // TODO: Infer shape without data rank if shape is a constant
   //       without -1 and without 0 and allowzero.
-  if (!hasShapeAndRank(getData()) || !hasStaticShape(getShape().getType()))
+  if (!hasShapeAndRank(getData()) && !hasStaticShape(getShape().getType()))
```
tungld, line 126: As the comment at L125 said, even though `shape` of `Reshape` is a constant, it may contain symbolic value, e.g. `0`, meaning taking the dimension from the input dimension. Not sure the code afterward can handle such a case: input is unranked and shape contains `0`. Otherwise, we need to check `shape` values in details.

chentong319: Add the code to handle 0 or -1 for shape. Since we just need to handle them for shape inference, those dimensions are conservatively given unknown (QuestionmarkIndexExpr). When coming to lowering, the data can not be unranked.

chentong319: Also test cases are added to lit test.

tungld (approved):
LGTM!

src/Dialect/ONNX/ONNXOps/Tensor/Reshape.cpp
tungld: Thanks!

PR #2541, opened 2023-09-30
Add a framework for NNPA op placement heuristics

AlexandreEichenberger:
Add a framework to enable device placement heuristics, populated initially by 3
* QualifyingOps: Place all qualifying ops on NNPA (default)
* FasterOps: Place qualifying ops that are faster on NNPA
* FasterOpsWSU: FasterOps with stick/unstick cost. 

Faster put only on NNPA ops that are faster than the CPU estimated time

Faster with stick unstick evaluates, if placing a given op on NNPA, the number of new sticks/unstick that would be added, and the number that would be removed. It starts by seeding operations that go 3x faster only on the NNPA, and then iterates to see if other ops nearby can be added without augmenting the cost of stick/unstick, essentially.


src/Accelerators/NNPA/Conversion/ONNXToZHigh/DevicePlacementHeuristic.hpp
```diff
@@ -0,0 +1,66 @@
...
+// Place all ops that qualify for NNPA execution on the NNPA when the operations
+// are estimated run faster on the NNPA, including the costs of Stick and
+// Unstick necessary for NNPA execution. The algorithm starts to place on the
+// CPU/NNPA operations that are significantly faster on CPU/NNPA. Then it aims
+// to add operations to the NNPA when the new operations are faster including
+// the additional (if any) stick/unstick required for these less significantly
+// faster NNPA operations. Three factors below can modify the sensitivity at
+// which ops are assigned to the NNPA.
+
+// minFactor: NNPA has to be at least minFactor times faster than CPU.
```
tungld, line 45: It's a bit difficult to know the difference between `minFactor` and  `significantNNPAFactor`.
As I understand, we use `minFactor` in `nnpaTimeWithOverheads * minFactor <= cpuTime`, so it would be better to say ` minFactor: NNPA (including stick/untick) has to be at least minFactor times faster than CPU.`

tungld (approved):
LGTM! Thanks for the great contribution!

src/Accelerators/NNPA/Conversion/ONNXToZHigh/DevicePlacementHeuristic.cpp
```diff
@@ -0,0 +1,496 @@
...
+            llvm::dbgs() << "      +1 unstick: " << costOfUnstickOp << "\n");
+        totalCostBenefit += costOfUnstickOp;
+      }
+      if (nnpaOpCount > 0) {
+        // Moving this op to NNPA will remove the need to stick this result
+        LLVM_DEBUG(
+            llvm::dbgs() << "      -1 stick: " << -costOfStickOp << "\n");
+        totalCostBenefit -= costOfStickOp;
+      }
+    }
```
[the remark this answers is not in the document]
AlexandreEichenberger: exactly

src/Accelerators/NNPA/Conversion/ONNXToZHigh/DevicePlacementHeuristic.hpp
AlexandreEichenberger: thanks for the suggested wording, much appreciated.

PR #2544, opened 2023-10-02
constant fold for IfOp

chentong319:
When the condition for IfOp is a constant, the IfOp can be replaced with either Then or Else branch.


tungld (approved):

tungld:
LGTM!

PR #2547, opened 2023-10-03
Handling the measurements associated with multiple runs

AlexandreEichenberger:


AlexandreEichenberger:
@tungld trying to average the results of multiple runs. Technical question, is there an easy way to add instruction to be the first and the last when inserting instrumentation so that I may initialize/close a file? Right now I only initialize during the first... which is ok but not great. Thanks

tungld:
> @tungld trying to average the results of multiple runs. Technical question, is there an easy way to add instruction to be the first and the last when inserting instrumentation so that I may initialize/close a file? Right now I only initialize during the first... which is ok but not great. Thanks

I don't know how to determine the last instrument point, but we can call a function to close the file at the end of the entry point `run_main_graph`. In that case, we would open the file with the append mode.

test/mlir/accelerators/nnpa/conversion/instrument/add-onnx-level.mlir
```diff
@@ -1,12 +1,14 @@
-// RUN: onnx-mlir --mcpu=z16 --maccel=NNPA  --printIR --EmitZHighIR --instrument-stage=Onnx --instrument-ops=onnx.* --InstrumentBeforeOp --InstrumentAfterOp --InstrumentReportTime -tag="test" %s | FileCheck %s
+// RUN: onnx-mlir --mcpu=z16 --maccel=NNPA  --printIR --EmitZHighIR -profile-ir=Onnx %s | FileCheck %s
```
tungld, line 1: Thanks for updating this!

utils/RunONNXModel.py
```diff
@@ -550,6 +550,9 @@ def warning(msg):
 def data_without_top_bottom_quartile(data, percent):
     data = np.array(sorted(data))
     trim = int(percent * data.size / 100.0)
+    if trim == 0 or data.size - 2 * trim < 1:
+        # Want at least one element, return as is.
+        return data
```
tungld, line 555: Thanks! I sometimes got the error here with a small number of iterations, say 3.

tungld (approved):
LGTM!

utils/RunONNXModel.py
AlexandreEichenberger: So did I :-)

PR #2551, opened 2023-10-04
Add mcpu to doc examples with maccel NNPA

mikeessen:
Update doc examples that specify maccel NNPA to include mcpu since NNPA instructions will only be used with an appropriate cpu.


AlexandreEichenberger (approved):
Thanks for the quick turnaround, much appreciated

PR #2558, opened 2023-10-16
[NNPA] Clip constant f32 values when stickifying them

tungld:
This patch is to clip constant f32 values when stickifying them for NNPA. Values will be clipped into the range of dlfloat16 before converting to dlfloat16, which avoids the "Range violation" error. The clipping is done during compilation time.


AlexandreEichenberger (approved):
LGTM, since its clipping at compile time, there is no performance degradation loss.

cjvolzka:
@tungld I just noticed it looks like there's a couple of documentation references that should be updated. They both mention the NNPALimit.h file in the old location prior to this change.

* https://github.com/onnx/onnx-mlir/blob/7f8f1d99f26e0a86479db08bf9a77c0b85797f32/docs/SupportedONNXOps-NNPA.md?plain=1#L13
* https://github.com/onnx/onnx-mlir/blob/7f8f1d99f26e0a86479db08bf9a77c0b85797f32/test/accelerators/NNPA/backend/CMakeLists.txt#L96

tungld:
Thanks @cjvolzka! I created a PR #2570 to update them.

PR #2559, opened 2023-10-16
Shape inference for Layer Normalization Op

AlexandreEichenberger:
Interested in optimizing LayerNormalization as an ONNX op in onnx-mlir, thus added shape inference for it.


src/Dialect/ONNX/ONNXOps/NN/Normalization.cpp
```diff
@@ -149,3 +149,124 @@ LogicalResult ONNXInstanceNormalizationOp::verify() {
...
+    if (bType.getElementType() != XElementType)
+      emitOpError("LayerNormalization op with incompatible B type");
+  }
+
+  // Check scale.
+  if (hasShapeAndRank(scale)) {
+    // Can check at this stage.
+    ShapedType scaleType = scale.getType().cast<ShapedType>();
+    ArrayRef<int64_t> scaleShape = scaleType.getShape();
+    SmallVector<int64_t> bcastShape;
```
tungld, line 206: it would be consistent if we name it `scaleCastShape` instead.

src/Dialect/ONNX/ONNXUnsupportedOps.hpp
```diff
@@ -44,7 +44,7 @@ UNSUPPORTED_OPS(ONNXHammingWindowOp)
 UNSUPPORTED_OPS(ONNXHannWindowOp)
 UNSUPPORTED_OPS(ONNXImputerOp)
 UNSUPPORTED_OPS(ONNXLabelEncoderOp)
-UNSUPPORTED_OPS(ONNXLayerNormalizationOp)
+// hi alex UNSUPPORTED_OPS(ONNXLayerNormalizationOp)
```
tungld, line 47: Should it be removed?

tungld (approved):
LGTM!

src/Dialect/ONNX/ONNXUnsupportedOps.hpp
AlexandreEichenberger: oops

src/Dialect/ONNX/ONNXOps/NN/Normalization.cpp
AlexandreEichenberger: `beast` stands for broadcast. I know changed it to `BBroadcastShape` and `scaleBroadcstChape` to be more explicit. Thanks for the catch.

src/Dialect/ONNX/ONNXUnsupportedOps.hpp
tungld: I think you can remove it in your later PR :)

src/Dialect/ONNX/ONNXOps/NN/Normalization.cpp
tungld: I see, I though b was B operand.

Issue #2310, opened 2023-06-07
SequenceMap should implement ResultTypeInferenceOpInterface


sorenlassen:
resnet-preproc-v1-18.onnx from model zoo fails during import in FrontendDialectTransformer because it has the operation SequenceMap which implements HasOnnxSubgraphOpInterface but not ResultTypeInferenceOpInterface, therefore it hits the assert
```
// Output types are propagated from region terminator to op results
// in opWithTypeInference logic below.
assert(opName.hasTrait<ResultTypeInferenceOpInterface::Trait>() &&
  "Subgraph ops must implement ResultTypeInferenceOpInterface");
```

the assert was introduced in PR #2285

PR #2560, opened 2023-10-16
import SequenceMap

sorenlassen:
Import SequenceMap even though the op has a region and doesn't implement ResultTypeInferenceOpInterface, provided `-useOnnxModelTypes` is true (which is the default now), thus relying on the result types inferred by `onnx::shape_inference::InferShapes()`. This makes it possible to import resnet-preproc-v1-18.onnx (see issue #2310).

With this change we can successfully import every model from the model zoo with `onnx-mlir -EmitONNXBasic`.

Also made Resize shape inference fail with a precise error message if axes is specified, since resnet-preproc-v1-18 has a Resize with axes.


AlexandreEichenberger:
@chentong319 do you want to have a look as you implemented the Sequences

chentong319 (approved):
LGTM!

PR #2561, opened 2023-10-17
Layer normalization lowering to KRNL

AlexandreEichenberger:
Lowering of ONNX LayerNormalization using the algorithm from the ONNX `--functions-to-decompose=LayerNormalization`.

Also enabled all of the backend tests. Did not add a lit test as the output is pretty long, relying on execution of the backend test should be enough at this stage.

This will corresponds to the `-O0` or fall through path. Will provide next a fused loop SIIMD implementation.

Minor upgrade: enabled model created using the `.onnxtext` framework to be compiled and executed in the `RunONNXModel.py` and `CheckONNXModel.py`.


AlexandreEichenberger:
@tung, this PR is a fallback for the `ONNX.LayerNormalization` to generate generic ONNX code when a (future) optimized code gen lowering fails (because of possibly some assumptions/sizes, ...) to use an faster optimized approach. 

I now realize that `ZHigh` would need a similar approach, namely if it decided to lower `ONNX.LayerNormalization` to something that can be then migrated to ZHigh.

So ideally we should have a function that can be called either from Conversion/ONNXToKrnl or from ConverstionONNXToZHigh.

Can you look a bit at the 
```
  LogicalResult generateONNXCode(ConversionPatternRewriter &rewriter,
      MDBuilder &create, ONNXLayerNormalizationOp lnOp, Value epsilon,
```
function in Normalization.cpp and see if we can somewhat extract it for use in both place?

We could do a separate phase that decide to do an ONNX-to-ONNX decompose too, it would be tricky to ensure that all of the conditions that may make the optimized path to fails are accurately reflected as the condition that would make the `LayerNormalization` illegal.

What do you think?

@chentong319 , your input would be very much appreciated too as you have a good picture of these challenges.

chentong319:
How about lowering LayorNormalization to onnx ops instead of krnl ops? Again the decision has to be interwound with the process of lowering to ZHigh.

AlexandreEichenberger:
Given the provided feedback, I migrated the function that transform LayerNorm to an independent function. As this get use elsewhere, we can then change its interface to what is best for both uses. It will work as is, but may be refined to be more elegant.

Thanks @tungld and @chentong319 for the feedback

tungld (approved):
LGTM.

PR #2564, opened 2023-10-17
Constant propagation for PowOp

tungld:
Found PowOp with constant inputs in GPT2, so I implemented constant propagation for it.


sorenlassen (approved):
LGTM

PR #2568, opened 2023-10-17
Initial changes for llvm uplift

philass:
The changes in this uplift are mainly related to following the migration guide in this discussion post  https://discourse.llvm.org/t/psa-bufferization-new-buffer-deallocation-pipeline/73375/11


philass:
The final issues seem to be related to https://discourse.llvm.org/t/psa-bufferization-new-buffer-deallocation-pipeline/73375/11

I'm working through them now

test/mlir/driver/buffer_loop_hoisting.mlir
philass, line 1: This is quite broken. This just disables the checks and offending ops

hamptonm1: @philass I understand your pain... the same goes for ONNX upgrade ( a lot of issues). Are we planning on just waiting for the next torch-mlir /llvm commit or are you going to work through this still? 

I noticed `torch-mlir` has upgraded to a more recent llvm commit as well: https://github.com/llvm/torch-mlir/pull/2517

philass:
Hey @AlexandreEichenberger, @tungld, @chentong319 

This should be ready for review. Only one lit test has a lingering issue. All other lit tests and end to end tests pass.

`test/mlir/driver/buffer_loop_hoisting.mlir` no longer works. I don't understand the part of the codebase too well. Everything else succeeds. How would you guys suggest I proceed.

Selfishly I would like to either
1) Delete the test
2) Leave the test changed as it is right now in the PR to get the lit tests to pass. 

The test can than be updated at a later point in time once it is fully triaged. 

Someone familiar with this part of the codebase could triage the issue.  The LLVM dialect, memref, and bufferization dialect or not things I regularly work with. So I'm likely not the best person to trouble shoot this lit test.

This would allow us to unblock the LLVM uplift which has already slipped (Really my fault there).

test/mlir/driver/buffer_loop_hoisting.mlir
philass: I will land this PR. Than we can figure out how the next bump should go

AlexandreEichenberger (approved):
LGTM, thanks for working on this, much appreciated.

Would it make sense to open an issue with the lit test you had to yank?

philass:
@AlexandreEichenberger Thanks for the review

> Would it make sense to open an issue with the lit test you had to yank?

Will do

PR #2569, opened 2023-10-17
mhlo: migrate from MLIR-HLO to StableHlo

ashay:
This patch replaces the ONNX-to-MHLO conversion pipeline with the
ONNX-to-StableHlo conversion pipeline.  Aside from the relatively few
changes to the C++ source files, most of the changes in this patch
update the lit tests, all of which were updated using the
utils/mlir2FileCheck.py script.

This patch does not rename the ONNX_MLIR_ENABLE_MHLO CMake flag nor does
it make the repository point to the StableHlo submodule.  I plan to make
these changes in a separate patch, since build-related changes are
likely to take a life of their own.

During the migration, I noticed a couple of bugs in the StableHlo
generation/lowering, specifically, in the generation/lowering of MatMul
and Convolution StableHlo operations.  I have disabled these tests for
now and if/when this patch is accepted, I'll create a GitHub issue so
that we can track the fixes.

cc: @burmako


tungld:

tungld:

philass:

ashay:
The CI error occurs because our MHLO submodule has a build bug.  I see that the bug has been fixed upstream, so we'll need to pull a new MHLO commit before retrying this patch.  I'll close this PR for now and reattempt it after the submodule update.  Thanks Phil and Tung for triggering the builds for the last few days!

ashay:
I couldn't re-open this PR after force pushing to the same branch, so here is the new PR: https://github.com/onnx/onnx-mlir/pull/2569.

PR #2571, opened 2023-10-18
Extend shapeInformation to set batch size for all inputs

tungld:
Option `--shapeInformation` currently allows setting static/dynamic dimensions for each model input. And, for each input, users need to set **all** dimensions, for example, if the first input is rank of 3, users need to pass a value say `0:-1x3x5` that must include 3 dimension sizes.

This patch extends `--shapeInformation` to allow partial setting of dimensions, counting from the outermost dimensions. For example, if users only want to set the first dimension of the first input to 5, users can use `0:5` and omit the remaining dimensions.

This new extension is useful when we set batch size for all inputs of a model. The batch size is often the first dimension (the outermost dimension), so we can use `-1:8` to set the batch size to 8 for all inputs.

### DIFF
--- a/src/Builder/ModelInputShaper.cpp
+++ b/src/Builder/ModelInputShaper.cpp
@@ -55,13 +55,22 @@ void ModelInputShaper::setShapeInformation(
   if (!shapeInformation.empty()) {
     std::stringstream shapeInfoString(shapeInformation);
     std::string shapeString;
+    bool hasAllInputSetting = false;
     while (std::getline(shapeInfoString, shapeString, ',')) {
+      assert(!hasAllInputSetting &&
+             "When -1 is used for all inputs, no other setting is allowed.");
       size_t pos = shapeString.find(':');
       std::string inputString = shapeString.substr(0, pos);
       std::string dimString = shapeString.substr(pos + 1);
 
       int64_t inputID = std::stoi(inputString);
-      assert(inputID >= 0 && "input_id must be >= 0");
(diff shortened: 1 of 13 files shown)


src/Builder/ModelInputShaper.cpp
```diff
@@ -55,13 +55,22 @@ void ModelInputShaper::setShapeInformation(
   if (!shapeInformation.empty()) {
     std::stringstream shapeInfoString(shapeInformation);
     std::string shapeString;
+    bool hasAllInputSetting = false;
     while (std::getline(shapeInfoString, shapeString, ',')) {
+      assert(!hasAllInputSetting &&
```
AlexandreEichenberger, line 60: Would we possibly create an error as opposed to an assert? Better for user-fronting issues.

AlexandreEichenberger (approved):
LGTM, great functionality.

Have you investigated if (for example when we have many model inputs having a batch size of -1), whether relying on the user to correlate them would help?

Or is dim analysis good enough to detect this on its own?

tungld:
> Have you investigated if (for example when we have many model inputs having a batch size of -1), whether relying on the user to correlate them would help?

Yes, I have thought about that and need a way for users to express the relationship between dims.

src/Builder/ModelInputShaper.cpp
tungld: Yes, will change it to an error.

tungld: Done. Thanks!

PR #2567, opened 2023-10-17
MHLO support default off

AlexandreEichenberger:
Disabling MHLO at this time due to the lack of response and engagement from the MHLO contributors, and the significant overhead that arise from maintaining this branch which does not appear to be heavily worked on anymore.

If you do rely on it, and want continued support, please comment on this PR and the linked issue #2525 

Will leave open a couple of days to gather any feedback you may have.

Thanks for your cooperation on keeping the ONNX-MLIR effort vibrant

PR #2573, opened 2023-10-19
mhlo: migrate from MLIR-HLO to StableHlo

ashay:
This patch replaces the ONNX-to-MHLO conversion pipeline with the
ONNX-to-StableHlo conversion pipeline.  Aside from the relatively few
changes to the C++ source files, most of the changes in this patch
update the lit tests, all of which were updated using the
utils/mlir2FileCheck.py script.

This patch also makes this repository point to the StableHLO submodule
(and drops the MLIR-HLO submodule) since our MHLO commit was old and
contained build bugs that were fixed upstream in the StableHLO
repository.

During the migration, I noticed that MatMul operations with dynamic             
dimensions are not lowered correctly.  Specifically, the lowering for           
matmul operations includes StableHLO's BroadcastInDimOp, which requires         
that its dimensions be static.  Fixing this requires a slightly more            
involved fix, but I will file an issue for this and fixing it in a              
separate PR.  For now, I have inserted a guard condition that skips the         
lowering for matmul with dynamic dimensions, so that the pass doesn't           
crash.


philass:

src/Conversion/ONNXToStableHlo/ONNXToStableHloCommon.cpp
```diff
@@ -2,18 +2,19 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-//====----- ONNXToMhloCommon.cpp - ONNX dialects to Mhlo lowering ---------===//
+//====----- ONNXToStableHloCommon.cpp - ONNX dialects to StableHlo lowering
+//---------===//
```
AlexandreEichenberger, line 6: nit formatting; this happens also elsewhere, please do a sweep

AlexandreEichenberger (approved):
LGTM, approving this PR as it is a step in the right direction (Stable HLO). 

Please let's continue our discussion about HLO involvement in this project and the level of support, either in the #2567 issue and preferably in a weekly onnx-mlir meeting.

src/Conversion/ONNXToStableHlo/ONNXToStableHloCommon.cpp
ashay: Thanks!  Fixed in the most recent commit.

sorenlassen:
@gongsu832 can you please make the jenkins pipelines run? they don't seem to respond to the jenkins-droid command

AlexandreEichenberger:

ashay:
I imagine that this has come up before, so pardon my ignore when I ask, would y'all be open to:
1. letting builds run on a push to branch, instead of requiring one of the admins to manually trigger builds, and
2. enabling ccache/sccache in CI builds?

tungld:

PR #2577, opened 2023-10-24
Fuse op when the shape is not static

chentong319:
Two changes in this PR
1. In fusible op selection, dimAnalysis is used to handle dynamic shape.
2. In code generation, the shapeHelper is used to get access expression from loop index set.

Lit test cases that can be fused or can not be fused are added.


src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1735,29 +1748,20 @@ bool OpFusionHelper::isControlFlowValidForFusion(
...
-  ArrayRef<int64_t> useShape = getShape(useOp->getResults()[0].getType());
+  Type defOutputType = defOp->getResultTypes()[0];
+  Type useOutputType = useOp->getResultTypes()[0];
+  ArrayRef<int64_t> defShape = getShape(defOutputType);
+  ArrayRef<int64_t> useShape = getShape(useOutputType);
   if (defShape != useShape) {
     return false;
   }
 
+  // Check the inputs in the defOp
```
AlexandreEichenberger, line 1764: Typo? "Check the inputs in the **useOp**"

AlexandreEichenberger: So one of the constraints is that all the "other" inputs of the candidate useOp are constants? If so, it would be helpful to have a description at the top of the class that lists the current limitations. What if it uses an input of an op that is already in the set of fused op, wouldn't that be ok too?

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1735,29 +1748,20 @@ bool OpFusionHelper::isControlFlowValidForFusion(
...
-  if (!hasStaticShape(defOp->getResults()[0].getType()))
-    return false;
-
-  ArrayRef<int64_t> defShape = getShape(defOp->getResults()[0].getType());
-  ArrayRef<int64_t> useShape = getShape(useOp->getResults()[0].getType());
+  Type defOutputType = defOp->getResultTypes()[0];
+  Type useOutputType = useOp->getResultTypes()[0];
+  ArrayRef<int64_t> defShape = getShape(defOutputType);
+  ArrayRef<int64_t> useShape = getShape(useOutputType);
   if (defShape != useShape) {
```
AlexandreEichenberger, line 1760: This condition seems to be stronger than the code that starts in line 1786:
```
  if (getRank(defOutputType) != getRank(useOutputType))
     return false;

   if (dimAnalysis) {
     if (!dimAnalysis->sameShape(defOp->getResult(0), useOp->getResult(0)))
       return false;
   } else {
     // If there is no dimAnalysis, check the simplest case.
     // Static and the same shape
     if (!hasStaticShape(useOutputType))
       return false;

     ArrayRef<int64_t> inputShape = getShape(useOutputType);
     if (inputShape != defShape)
       return false;
```

test/mlir/conversion/onnx_to_krnl/onnx_lowering_fuse.mlir
```diff
@@ -178,3 +178,94 @@ func.func @fuse_element_14(%arg0: tensor<5xf32>) -> tensor<*xf32> {
 // CHECK:           return [[RES_]] : memref<5xf32>
 // CHECK:         }
 }
+
+// -----
+
+func.func @fuse_element_15(%arg0: tensor<4x5xf32>, %arg1: tensor<?xf32>) -> tensor<*xf32> {
```
AlexandreEichenberger, line 184: Interesting: here the Add has a parameter that is not a constant but still is fused. Is that because it came from the args?

test/mlir/conversion/onnx_to_krnl/onnx_lowering_fuse.mlir
```diff
@@ -178,3 +178,94 @@ func.func @fuse_element_14(%arg0: tensor<5xf32>) -> tensor<*xf32> {
...
+// CHECK:             [[LOAD_PARAM_1_MEM_:%.+]] = krnl.load [[PARAM_1_]]{{.}}[[VAR_8_]]{{.}} : memref<?xf32>
+// CHECK:             [[VAR_10_:%.+]] = arith.addf [[LOAD_RES_MEM_]], [[LOAD_PARAM_1_MEM_]] : f32
+// CHECK:             krnl.store [[VAR_10_]], [[RES_1_]]{{.}}[[VAR_3_1_]]#0, [[VAR_3_1_]]#1] : memref<4x?xf32>
+// CHECK:           }
+// CHECK:           return [[RES_1_]] : memref<4x?xf32>
+// CHECK:         }
+
+// -----
+
+func.func @fuse_element_17(%arg0: tensor<?x5xf32>, %arg1: tensor<?xf32>) -> tensor<*xf32> {
```
AlexandreEichenberger, line 248: Very cool, works very well. If you have the time, I like to add a comment that indicates what the expected behavior is. For example "//Fusion will work because shape analysis can rule out broadcasting in the lowest dimension of the add"

AlexandreEichenberger (approved):
LGTM, if you don't mind adding a few clarification (as comment in the code). Otherwise, new functionality is impressive with respect to broadcasts.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
chentong319: Currently, the other inputs have to be constant or block argument. In theory, it is okay that another input comes from the set of fused op. Just need a DAG to record the fusible candidates, while a list is used now. Will consider this extension.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
chentong319: I assume that the comparison of defShape and useShape will treat the dynamic dimension as the same. The later code use DimAnalysis to analyze the dynamic dimension. Will check it.

chentong319: I tested with an example. My assumption is correct.

test/mlir/conversion/onnx_to_krnl/onnx_lowering_fuse.mlir
chentong319: yes.

PR #2579, opened 2023-10-24
Relax the condition for checking a shape defined by Concat

tungld:
This patch is a complement to #2450 to improve ShapeInference when a shape input is defined by Concat of dimensions.

Before this patch, it was required that the inputs of Concat must be defined by Constant, Dim or Cast. But now, the inputs can be defined by any operation as long as they contain only one i64 element.

For example, with this patch, we can infer shape for the following Expand:
```
func.func @test_expand_with_concat(%arg0: tensor<1xi64>, %arg1: tensor<1xi64>, %arg2: tensor<f32>) -> tensor<*xf32> { 
  %0 = onnx.Constant dense<1> : tensor<1xi64>                                                      
  %1 = "onnx.Concat"(%arg0, %0, %arg1) {axis = 0 : si64} : (tensor<1xi64>, tensor<1xi64>, tensor<1xi64>) -> tensor<3xi64>
  %2 = "onnx.Expand"(%arg2, %1) : (tensor<f32>, tensor<3xi64>) -> tensor<*xf32>
  return %2 : tensor<*xf32>
```
The result will be `%2 = "onnx.Expand"(%arg2, %1) : (tensor<f32>, tensor<3xi64>) -> tensor<?x1x?xf32>`

### DIFF
--- a/src/Dialect/ONNX/ONNXDimAnalysis.cpp
+++ b/src/Dialect/ONNX/ONNXDimAnalysis.cpp
@@ -84,7 +84,8 @@ static std::optional<DimAnalysis::DimT> insertDimWhenUseful(const Value tensor,
       // The correct axis is from ONNXDimOp.
       axis = dimOp.getAxis();
       okToInsert = true;
-    } else if (isa<ONNXCastOp>(op) || tensorType.isDynamicDim(axis))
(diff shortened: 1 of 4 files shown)


AlexandreEichenberger (approved):
LGTM

PR #2584, opened 2023-10-25
[Testing][NNPA] Generate affine code for stickification and unstickification

tungld:


src/Accelerators/NNPA/Dialect/ZLow/ZLow.td
```diff
@@ -449,4 +457,66 @@ def ZLowDummyOp:ZLow_Op<"dummy", [MemRefsNormalizable]> {
...
+  let arguments = (ins F32:$input);
+  let results = (outs DLF16:$output);
+  let builders = [
+    OpBuilder<(ins "::mlir::Value":$input), [{
+      // F16 is used as a container for DLF16.
+      build($_builder, $_state, $_builder.getF16Type(), input);
+    }]>
+  ];
+  let hasCanonicalizer = 1;
+}
```
tungld, line 488: Scalar conversion is not used at this moment, but we still support it for completeness.

src/Accelerators/NNPA/Conversion/ZHighToZLow/ZHighToZLow.cpp
```diff
@@ -1580,8 +1585,135 @@ struct ZHighToZLowStickifiedConstantOfShapeOpLowering
...
+        alloc, outputDims, flattenedOutputDims, collapsedInnermostLoops);
+
+    // Create loop iteration (flattened to 1D) and block it by VL.
+    ValueRange loopDef = create.krnl.defineLoops(1);
+    ValueRange blockedLoopDef = create.krnl.block(loopDef[0], VL);
+    SmallVector<Value, 1> optimizedLoopDef(1, blockedLoopDef[0]);
+
+    if (enableParallel) {
+      create.krnl.parallel(blockedLoopDef[0]);
+      onnxToKrnlParallelReport(op, /*successful*/ true, 0,
```
AlexandreEichenberger, line 1664: Thanks!

src/Accelerators/NNPA/Conversion/ZHighToZLow/ZHighToZLow.cpp
```diff
@@ -1580,8 +1585,135 @@ struct ZHighToZLowStickifiedConstantOfShapeOpLowering
...
+    MDBuilder create(rewriter, loc);
+
+    Operation *op = convertOp.getOperation();
+    ValueRange operands = adaptor.getOperands();
+    Value X = operands[0];
+    int64_t rank = getRank(X.getType());
+
+    // SIMD info.
+    // Fixed VL for the conversion instruction: 8 elements per instruction call.
+    int64_t VL = 8;
```
AlexandreEichenberger, line 1616: Note, we can get better loop unrolling by selecting VL as multiple of the machine VL; with normal operations, an `add` of vector length 16 becomes 2 simd instructions. So for example, you could unroll by a cache length, so that you get better register usage by having multiple instruction in flight at the same time and (for parallel) no false sharing. What I have done is estimate the register pressure... so if it's simple like here, you can definitely have 4-8 instructions in flight. If has a lot of intermediary results, then its better not to unroll too much as we don't want spilling. In any case, the backend may also unroll, if you do it here, you are sure it happens :-)

AlexandreEichenberger (approved):
LGTM, see comment on VL for possible performance tuning. 
Also, if we see performance regression with respect to zDNN, we can still apply this optimization to simplify/eliminate needless conversions (e.g. unstick -> transpose -> restock), and then regenerate zdnn stick/unstick if pattern was not optimized.

tungld:
> Also, if we see performance regression with respect to zDNN, we can still apply this optimization to simplify/eliminate needless conversions (e.g. unstick -> transpose -> restock), and then regenerate zdnn stick/unstick if pattern was not optimized.

Totally agree!

src/Accelerators/NNPA/Conversion/ZHighToZLow/ZHighToZLow.cpp
tungld: Yes, you are right. I have another round for optimization, e.g. optimizing LayoutTransform, so I will do unrolling then with a new PR.

PR #2590, opened 2023-10-26
Add --mcpu flag to ensure NNPA backend tests are executing

hamptonm1:
When testing on a z16, several backend tests were failing for NNPA because we included the `--mcpu` flag recently to check hw and so this was also missing from the backend tests.


AlexandreEichenberger (approved):
LGTM,

do you think it may make sense to add a warning if we have the NNPA `maccel` without a `mcpu=z*`?

hamptonm1:
@jenkins-droid Please test this

hamptonm1:

PR #2591, opened 2023-10-26
Add vim to docker dev file

hamptonm1:
There are other useful packages we can add to the dev docker file but I will leave at your discretion. Installing vim comes in handy when viewing a file but not a requirement of course :)

### DIFF
--- a/docker/Dockerfile.onnx-mlir-dev
+++ b/docker/Dockerfile.onnx-mlir-dev
@@ -22,6 +22,8 @@ ARG ACCEL=NNPA
 ARG TEST_NOFLOAT16
 ARG TEST_MCPU
 
+RUN apt-get install -y vim
+
 RUN LLVM_PROJECT_ROOT=${WORK_DIR}/llvm-project \
     && ONNX_MLIR_ROOT=${WORK_DIR}/onnx-mlir \
 # Build onnx-mlir and run tests


gongsu832:
The purpose of the dev image is to provide a minimal env that anybody can use to start developing for onnx-mlir. What's considered "useful" is highly subjective so it's best left to the individual developers. Not everybody uses vim. Some prefer emacs, or whatever other editor. Same goes for other "useful" things. The docker image size is already pretty bloated as is. It's 13.4GB uncompressed and takes 25+ minutes just to publish the image to dockerhub. We tried very hard to keep the image size down.

hamptonm1:
> The purpose of the dev image is to provide a minimal env that anybody can use to start developing for onnx-mlir. What's considered "useful" is highly subjective so it's best left to the individual developers. Not everybody uses vim. Some prefer emacs, or whatever other editor. Same goes for other "useful" things. The docker image size is already pretty bloated as is. It's 13.4GB uncompressed and takes 25+ minutes just to publish the image to dockerhub. We tried very hard to keep the image size down.

Thanks @gongsu832 this was not a requirement, it was simply a suggestion. I understand your sentiments and I wanted the opinions of others in the community as well.

AlexandreEichenberger:
@hamptonm1 isn't it pretty easy to add your favorite editor in an running instance of the image? If so, maybe we should refrain from adding it to the image that everyone shares.

AlexandreEichenberger:
@hamptonm1 please close if we are not going ahead with this one.

PR #2593, opened 2023-10-27
fix const prop of onnx.Cast to int types

sorenlassen:


src/Dialect/ONNX/ElementsAttr/ElementsAttrBuilder.cpp
```diff
@@ -276,80 +276,213 @@ ElementsAttr ElementsAttrBuilder::where(ElementsAttr cond, ElementsAttr lhs,
...
+    return min;
+  // static_cast<double>(max)) can round to a larger number
+  // so return max if from is greater or equal, not just if greater
+  if (from >= max)
+    return max;
+
+  if (TRUNCATE)
+    return static_cast<TO>(from);
+
+  // llrint recommendation: https://stackoverflow.com/a/47347224
```
AlexandreEichenberger, line 319: Looks like you researched this issue extensively and that it is better than simply `trunc(from+0.5)`

src/Transform/ONNX/ConstProp.cpp
```diff
@@ -664,16 +666,16 @@ Value ConstPropGemm(PatternRewriter &rewriter, Value replacingValue,
   if (hasBias) {
     ElementsAttr bias = getConstValueElements(biasMatrixValue);
     if (beta != 1.0) {
```
AlexandreEichenberger, line 668: Isn't the neutral element for bias `0`? Maybe a copy/paste error from `alpha`

AlexandreEichenberger (approved):
LGTM, please check comment of a possible copy/paste error.

src/Transform/ONNX/ConstProp.cpp
sorenlassen: I think the implementation is correct: you can skip the scalar multiplication with beta if beta==1.0

the `if (hasBias)` condition above skips adding beta*bias if bias is none - one could consider also skipping if beta==0.0, i.e., `if (hasBias && beta != 0.0)` if we ignore NaNs in the bias matrix

src/Dialect/ONNX/ElementsAttr/ElementsAttrBuilder.cpp
sorenlassen: I believe llrint's "round to nearest, ties to even" (sometimes called round-to-even or RNE) is the go to rounding mode used for numerical computing and implemented by CPUs (see [the RNE entries in the float 8 rounding tables for onnx.Cast](https://github.com/onnx/onnx/blob/main/docs/Operators.md#Cast))

you can see more details in the stackoverflow link in the code comment

sorenlassen (commented):
thank you for the review!

PR #2594, opened 2023-10-30
[NNPA] Add a lit test for the IBM CCFD model

tungld:


AlexandreEichenberger (approved):
LGTM, neat lit test with `CHECK-NOT`

PR #2597, opened 2023-10-31
Upgrade to ONNX 1.15.0

hamptonm1:
Installing `ONNX 1.15.0` was a bit tricky this time around. It required me to upgrade `pip` and `setuptools`. Also we had to manually install `libprotobuf-dev` because I received an error that the libprotobuf.so file did not exist when installing onnx. Also I added the new backend test which became available with the release of `ONNX 1.15.0`.


hamptonm1:

PR #2599, opened 2023-10-31
fix op fuse error

chentong319:
Op fuse use ONNXBroadcastOpShapeHelper to generate access expression. However, memref.dim, which is not needed, is also generated. 
Details explanation can be found in the code.
More test cases are added.

### DIFF
--- a/src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
+++ b/src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
@@ -1893,7 +1893,21 @@ Value OpFusionHelper::emitFuseOps(
     defOpResult =
         emitScalar(rewriter, loc, useOp, currentElementType, inputValues);
     defOp = useOp;
-    alloc = defOp->getResult(0);
+
+    // Explanation:
+    // The variable, alloc, is used to initialize ONNXBroadcastOpShapeHelper.
+    // The output of defOp should have the same shape as the alloc, which is
+    // one of the prerequisits for fusion.
+    // However, they may have different element type for some ops, such as
+    // comparison and cast.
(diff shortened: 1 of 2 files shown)


src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1893,7 +1893,21 @@ Value OpFusionHelper::emitFuseOps(
...
+    // However, they may have different element type for some ops, such as
+    // comparison and cast.
+    // The issue is that ONNXBroadcastOpShapeHelper.computeShape() may use
+    // this parameter to generate some code (memref.dim) that is not really
+    // needed here. Due to this live user, the original op can not be erased.
+    // This error occurs when there are more than one op with dynamic dim
+    // to be fused.
+    // Luckily, the ShapeHelper does not check the element type.
+    // Therefore, alloc is used for all the fused op.
+    // alloc = defOp->getResult(0);
```
tungld, line 1910: @chentong319 is it a temporary fix or do you think the right approach would be checking multiple ops with dynamic dim as you said `This error occurs when there are more than one op with dynamic dim to be fused`?

Also, since `alloc = defOp->getResult(0);` was removed, it's weird to have comments at the end of the function. Perhaps move the comment somewhere up.

chentong319: I think the fix is the right approach. In ONNXBroadcastOpShapeHelper, only the shape of tensor is consider, not the element type of the tensor. Therefore, alloc can be used for different op, as long as it guaranteed that it has the same shape for that operand, which has only scalar value not tensor value with fusion. That's the tricky part of the code gen.
I will move  the comment to the operands construction and add `const`.

tungld (approved):
LGTM.

PR #2600, opened 2023-11-01
[NNPA] Support scalar broadcasting binary ops on NNPA

tungld:
This patch is to optimize MatMul and Div in Attention layer, in particular to remove stick/unstick between them:
image

Two functionalities are added:
- Supporting scalar broadcasting binary ops on NNPA. It's because the second operand of Div is a scalar value.
- Since tensors are 4D, there are reshapes between MatMul and Div to convert 4D to 3D and vice versa. This patch creates a rule to propagate Reshape down in the operation chain, so that there is no Reshape in between MatMul and Div. After that only stick and unstick with the same layout are in between, and they are automatically removed by existing rules.

Though rewriting rules are templates and can be applied to other binary ops, this patch only enables Div to get better performance for Attention. If we enable Add, the scalar Add in LayerNorm computation would run on NNPA which introduces more stick/unstick and not good.


src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.td
```diff
@@ -201,7 +212,23 @@ def replaceONNXDivPattern : Pat<
   (ONNXDivOp $x, $y),
   (ZHighUnstickOp (ZHighDivOp (ZHighStickOp:$s_x $x, (NoneLayoutAttr)),
                               (ZHighStickOp $y, (NoneLayoutAttr)),
-                              (returnType $s_x)))
+                              (returnType $s_x))),
+  [], [],
+  (addBenefit 0)
+>;
```
chentong319, line 218: The condition of the pattern is implicitly mixed with the condition in LegalityCheck. Should we check the condition here redundantly, or any better approach?

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.td
```diff
@@ -201,7 +212,23 @@ def replaceONNXDivPattern : Pat<
...
+
+def replaceONNXDivBroadcastPattern : Pat<
+  (ONNXDivOp (either $x, $y)),
+  (ZHighUnstickOp
+    (ZHighDivOp
+      (ZHighStickOp:$s_x $x, (NoneLayoutAttr)),
+      (ZHighStickifiedConstantOfShapeOp
```
chentong319, line 225: Any chance to enable ZHighStickify Op to handle broadcast?

src/Accelerators/NNPA/Conversion/ZHighToZLow/ZHighToZLow.cpp
```diff
@@ -1498,13 +1499,59 @@ struct ZHighToZLowBatchNormOpLowering : public ConversionPattern {
...
+    ZHighStickifiedConstantOfShapeOpShapeHelper shapeHelper(
+        op, operands, &create.krnlIE);
+    shapeHelper.computeShapeAndAssertOnFailure();
+
+    // Convert ZTensor type to MemRefType.
+    ZMemRefType zMemRefType =
+        convertZTensorToMemRefType(*op->result_type_begin());
+
+    // Allocate a buffer for the result MemRef.
+    Value res = insertAllocAndDeallocZMemRef(
```
chentong319, line 1536: Do we add deallocZMemRef buy ourselves? Or just it's an old name for the function?

chentong319:
If the shape to broadcast to is a static shape, what will happen?

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.td
tungld: What is the dynShape approach?

tungld:
> If the shape to broadcast to is a static shape, what will happen? Handled previously with a stickified constant?

Yes, it was handled previously, expanding the scalar to a constant then stickifying it at compile time to produce a stickified constant.

src/Accelerators/NNPA/Conversion/ZHighToZLow/ZHighToZLow.cpp
tungld: I removed `dealloc` in all names and comments. Thanks!

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.td
chentong319: By redundancy, I meant to add the condition checked by LegalityCheck into the pattern so that the pattern can stand alone.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.td
chentong319: You pass the dynShape to create the broadcast constant.  Combine stickily and tensor expansion into one op.

chentong319 (approved):
LGTM!
For future:
1. This transformation has unique overhead due to the constant expansion. Should we modify the cost model?
2.  In addition to scalar, handle tensor broadcast.

tungld:
> This transformation has unique overhead due to the constant expansion. Should we modify the cost model?

Good idea. Yes, we should I think.

PR #2606, opened 2023-11-07
Fusing two Reshape operations

### LINKED ISSUE #2601: Reshape optimization shows errors
Reshape optimization is really powerful and is really helping a lot. In a new test example, from converting a GroupNorm into a LayerNorm, a complicated pattern is being solved. However, there is an error in the propagating of one reshape into another. Below is a simplified example.

```mlir
func.func @main_graph(%arg0: tensor<3x4x2x2xf32>) -> tensor<?x?x?xf32> {
  %0 = onnx.Constant dense<0.00999999977> : tensor<1xf32>
  %1 = "onnx.Cast"(%0) {saturate = 1 : si64, to = f32} : (tensor<1xf32>) -> tensor<1xf32>
  %2 = "onnx.Shape"(%arg0) {start = 0 : si64} : (tensor<3x4x2x2xf32>) -> tensor<4xi64>
  %3 = "onnx.Shape"(%arg0) {end = 2 : si64, start = 1 : si64} : (tensor<3x4x2x2xf32>) -> tensor<1xi64>
  %4 = onnx.Constant dense<2> : tensor<1xi64>
  %5 = "onnx.Div"(%3, %4) : (tensor<1xi64>, tensor<1xi64>) -> tensor<1xi64>
  %6 = "onnx.Shape"(%arg0) {end = 1 : si64, start = 0 : si64} : (tensor<3x4x2x2xf32>) -> tensor<1xi64>
  %7 = "onnx.Shape"(%arg0) {start = 2 : si64} : (tensor<3x4x2x2xf32>) -> tensor<2xi64>
  %8 = "onnx.Concat"(%6, %4, %5, %7) {axis = 0 : si64} : (tensor<1xi64>, tensor<1xi64>, tensor<1xi64>, tensor<2xi64>) -> tensor<5xi64>
  %9 = "onnx.Reshape"(%arg0, %8) {allowzero = 0 : si64} : (tensor<3x4x2x2xf32>, tensor<5xi64>) -> tensor<*xf32>
  %10 = onnx.Constant {value_ints = [0, 0, -1]} : tensor<3xi64>
  %11 = "onnx.Reshape"(%9, %10) {allowzero = 0 : si64} : (tensor<*xf32>, tensor<3xi64>) -> tensor<?x?x?xf32>
  onnx.Return %11 : tensor<?x?x?xf32>
}
```

High level, a `3x 4 x2x2` is first transformed to `3x 2x2  x2x2`; and then the `3x2x2x2x2` is transformed to `3x2x8`.

here is the output of shape inference
```mlir
module attributes {llvm.data_layout = "e-m:o-i64:64-i128:128-n32:64-S128", llvm.target_triple = "arm64-apple-darwin23.0.0", "onnx-mlir.symbol-postfix" = "reshape-opt-issue-small"} {
  func.func @main_graph(%arg0: tensor<3x4x2x2xf32>) -> tensor<3x2x8xf32> {
    %0 = onnx.Constant dense<[0, 0, -1]> : tensor<3xi64>
    %1 = "onnx.Reshape"(%arg0, %0) {allowzero = 0 : si64} : (tensor<3x4x2x2xf32>, tensor<3xi64>) -> tensor<3x2x8xf32>
    onnx.Return %1 : tensor<3x2x8xf32>
  }
}
```

The shape of %1 is correct  (because of the refine step), but it you look at the input to the reshape, its a `[0,0,-1]` vector which would result in a `3x4x8` not the expected `3x2x8`.

tungld:
This patch is to fuse two Reshape operations into a single Reshape operation.

The old fusion rule for reshape was wrong because it ignored the first Reshape's shape and directly used the second Reshape's shape as the result shape.

Resolves #2601


AlexandreEichenberger (approved):

src/Dialect/ONNX/ONNXOps/OpHelper.cpp
```diff
@@ -317,6 +317,24 @@ ONNXConstantOp getONNXConstantOp(Value value) {
   return dyn_cast_or_null<ONNXConstantOp>(value.getDefiningOp());
 }
 
+bool getI64ValuesFromSmallONNXConstantOp(
```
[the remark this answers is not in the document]
tungld: Just to confirm: does it work with DisposableElementsAttrs? I guess the issue is the value in ONNXConstantOp can be DisposableElementsAttrs or DenseElementsAttrs. That's why I convert to DisposableElementsAttrs to DenseElementsAttrs instead of reading values directly from DisposableElementsAttrs, and it's expensive.

src/Dialect/ONNX/Rewrite.td
```diff
@@ -576,12 +576,6 @@ def SwapTransposeConcatPattern: Pat<
 // Canonicalization for ONNXReshapeOp
 //===----------------------------------------------------------------------===//
 
-def FuseReshapePattern:  Pat<
```
sorenlassen, line 579: could we keep this when az2 is false?

src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -1043,6 +1043,227 @@ class ReplaceUnsqueezeOfExpandRewritePattern
...
+      ONNXReshapeOp secondReshapeOp, PatternRewriter &rewriter) const override {
+    // Second Reshape.
+    Operation *op = secondReshapeOp.getOperation();
+    Value secondData = secondReshapeOp.getData();
+    Value secondShape = secondReshapeOp.getShape();
+    int64_t secondAllowZero = secondReshapeOp.getAllowzero();
+    if (secondAllowZero != 0)
+      return rewriter.notifyMatchFailure(op, [&](::mlir::Diagnostic &diag) {
+        diag << "Does not support AllowZero != 0";
+      });
```
sorenlassen, line 1124: you can just say
`rewriter.notifyMatchFailure("Does not support AllowZero != 0");`
see https://github.com/sorenlassen/llvm-project/blob/main/mlir/include/mlir/IR/PatternMatch.h#L615-L623

src/Dialect/ONNX/ONNXOps/OpHelper.cpp
```diff
@@ -317,6 +317,24 @@ ONNXConstantOp getONNXConstantOp(Value value) {
...
+  if (!getElementType(elemsAttr.getType()).isInteger(64))
+    return false;
+
+  DenseElementsAttr denseAttr =
+      ElementsAttrBuilder::toDenseElementsAttr(elemsAttr);
+  auto valueIt = denseAttr.getValues<IntegerAttr>().begin();
+  for (unsigned int i = 0; i < denseAttr.getNumElements(); ++i) {
+    int64_t d = (*valueIt++).cast<IntegerAttr>().getInt();
+    iRes.emplace_back(d);
+  }
```
sorenlassen, line 334: `getValues<T>()` works on the ElementsAttr interface and is supported by DisposableElementsAttr, so there is no need to convert to DenseElementsAttr

we generally access ElementsAttr like this all over the onnx-mlir code base and is why all the code paths for the ONNX dialect work regardless of whether an ElementsAttr is dense or disposable

moreover, it's simpler and faster to access the elements as int64_t, see the example in ConstPropOfShape: https://github.com/onnx/onnx-mlir/blob/main/src/Transform/ONNX/ConstProp.cpp#L905

src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -1043,6 +1043,227 @@ class ReplaceUnsqueezeOfExpandRewritePattern
...
+
+    // First Reshape.
+    if (!definedBy<ONNXReshapeOp>(secondData))
+      return rewriter.notifyMatchFailure(op, [&](::mlir::Diagnostic &diag) {
+        diag << "The input data is not defined by a Reshape";
+      });
+    auto firstReshapeOp = cast<ONNXReshapeOp>(secondData.getDefiningOp());
```
sorenlassen, line 1131: I recently learned that getDefiningOp can take a template argument, so you can say
```
auto firstReshapeOp = secondData.getDefiningOp<ONNXReshapeOp>();
if (!firstReshapeOp) return ...;
```

sorenlassen (approved):
LGTM

the comments are excellent and makes it easy to understand the subtle logic

src/Dialect/ONNX/ONNXOps/OpHelper.cpp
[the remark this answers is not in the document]
tungld: ElementsAttrs works well. Thanks!

src/Dialect/ONNX/Rewrite.td
tungld: I guess you mean `az2 == 1`. I restores this and checked `az1` and `az2`. I am not sure how it is when both `az1` and `az2` are 1 because I don't understand well a tensor of a mix of zero and non-zero dimensions. So only support `az1=0` and `az2=1`.

src/Dialect/ONNX/Rewrite.cpp
tungld: Great. I used the simplified verson.

src/Dialect/ONNX/ONNXOps/OpHelper.cpp
tungld: This is what I was looking for. Changed to use `getValues<T>()`. Thanks!

src/Dialect/ONNX/Rewrite.cpp
tungld: Yeah, I saw it yesterday too.

src/Dialect/ONNX/Rewrite.td
sorenlassen: sorry, I got mixed up

I think you can drop the `IntegerAttrIsOf<0> $az1)` requirement because when $az2==1 then $s2 describes the output shape completely regardless of $s1, if I understand correctly

if a tensor mixes zero and non-zero dimensions then the tensor is just empty (if any dimension is zero) and the spec says that then there can be no -1 (spec: "If the attribute 'allowzero' is set, it is invalid for the specified shape to contain both a zero value and -1, as the value of the dimension corresponding to -1 cannot be determined uniquely.")

tungld: OK, so I removed the `az1` testing. Thanks!

PR #2607, opened 2023-11-07
[NNPA] Not using either in tablegen

tungld:
I was trying to use `either` keyword in tablegen for operations to avoid writing two patterns with swapped operands for lowering ONNXDiv to ZHigh, but it seems it did not work as expected.
So this patch removes `either` keyword and explicitly defines two patterns.


chentong319 (approved):
LGTM!

PR #2609, opened 2023-11-08
Detect LayerNorm in presence of reciprocal and div of 1

AlexandreEichenberger:
T5 model apparently use a `x * (1 / y)` pattern instead of `x / y`. The new recompose also now check for reciprocal and divide 1 by patterns.

Lit tests added.


tungld (approved):
LGTM. Thanks!

PR #2611, opened 2023-11-09
[NNPA] Use F16 as element type for zTensor

tungld:
Currently F32 is used in zTensor, e.g. `tensor<1x28x28x1xf32, #zhigh.layout<{dataLayout = "NHWC"}>>` which is not good because the element type of a zTensor is DLFloat16. Since MLIR does not support DLFLOAT16, this patch will use F16 as element type for zTensors, e.g. `tensor<1x28x28x1xf16, #zhigh.layout<{dataLayout = "NHWC"}>>`

Several type issues were exposed by using F16 and also fixed in this path.

This is the first step toward implementing stickify/unstickify in onnx-mlir.


AlexandreEichenberger (approved):
LGTM

PR #2614, opened 2023-11-10
Parse and set --mcpu in onnx-mlir-opt command

tungld:
`--maccel` needs `--mcpu` to know it is z16 or not, so this patch parses and sets `--mcpu` in the `onnx-mlir-opt` command. Otherwise, we will see a warning `Warning: No NNPA code is generated because --mcpu is not set or < z16.` when running `onnx-mlir-opt` for NNPA.


tungld:

AlexandreEichenberger (approved):
LGTM

PR #2616, opened 2023-11-10
Import dim_param for model inputs and outputs

tungld:
In ONNX specification, a dimension can be `dim_value` (static) or `dim_param` (a named dynamic dimension).
If two dynamic dimensions has the same `dim_param`, they must have the same value at runtime.

This patch will import `dim_param` of the inputs and outputs of a model, and embed such `dim_param` in argument and result attributes. This patch also moves `input_names` and `output_names` into argument and result attributes `onnx.names`. for example
```
func.func @main_graph(
         %arg0: tensor<?x?xf32> {onnx.dim_params = "0:batch_size,1:sequence_len", onnx.name = "X"}, 
         %arg1: tensor<1x?xf32> {onnx.dim_params = "1:sequence_len", onnx.name = "Y"}) 
    -> (tensor<?x?xf32> {onnx.dim_params = "0:batch_size,1:sequence_len", onnx.name = "Z"})
```

`input_dim_params` and `output_dim_params` are a list of `"dim_index:dim_param"`

`input_dim_params = ["0:batch_size,1:sequence_len", "1:sequence_len"]` so the second unknown dim of input 0 is the same as the second unknown dim of input 1 because they have the same dim_param.

`output_dim_params = ["0:batch_size,1:sequence_len", "1:sequence_len"]` so the output has the same shape as the input.

In a practical model, e.g. t5-decoder with KV cache, dim_param is like the following, and we know that all the first dimensions (batch_size) must be the same.
Screenshot 2023-11-10 at 22 19 38

These dim_param information will be used later by the dynamic dimension analysis to make the analysis stronger. Also we can check if users' inputs to the model are consistent or not.

### DIFF
--- a/src/Builder/FrontendDialectTransformer.cpp
+++ b/src/Builder/FrontendDialectTransformer.cpp
@@ -500,7 +500,6 @@ private:
         std::string dimParam = "";
         Type argTy = ImportType(input.type(), dimParam);
         argTy = modelInputShaper_.reshape(inputIndex, argTy);
-        llvm::outs() << "input dim: " << dimParam << "\n";
         inputDimParams.emplace_back(dimParam);
 
         argTypes.emplace_back(argTy);
@@ -549,7 +548,6 @@ private:
     for (const auto &output : graph.output()) {
       std::string dimParam = "";
       ImportOutputTensor(output, retTys, retVals, dimParam);
-      llvm::outs() << "output dim: " << dimParam << "\n";
       outputDimParams.emplace_back(dimParam);
     }
 
--- a/src/Builder/FrontendDialectTransformer.cpp
+++ b/src/Builder/FrontendDialectTransformer.cpp
@@ -190,6 +190,7 @@ private:
   // mapping between string name and symbol
   ValueSymbolMapping frontend_symbols_;
 
+  // Keep shape information set by users.
   ModelInputShaper modelInputShaper_;
 
   using ImportHandlerType = void (onnx_mlir::detail::FrontendGenImpl::*)(
@@ -287,7 +288,8 @@ private:
    * Import an onnx tensor type by determining and returning its type
    * @param type_proto onnx tensor TypeProto.
    */
-  Type ImportTensorType(const onnx::TypeProto &type_proto) {
+  Type ImportTensorType(
+      const onnx::TypeProto &type_proto, std::string &dimParam) {
     assert(type_proto.value_case() == onnx::TypeProto::kTensorType &&
            "expect tensor type");
     std::vector<int64_t> dims;
@@ -300,6 +302,7 @@ private:
     auto shape_proto = tensor_type.shape();
     for (int i = 0; i < shape_proto.dim_size(); i++) {
       if (shape_proto.dim()[i].dim_value()) {
+        // Dim is a constant value.
         int dim_numeric_size = shape_proto.dim()[i].dim_value();
         assert(dim_numeric_size != 0 &&
                "Parsed an tensor with a dimension size of zero");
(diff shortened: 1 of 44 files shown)


AlexandreEichenberger (approved):
LGTM, I had wanted something like this for a long time, it's great that it is part of the ONNX protocol so that we don't have to reinvent the wheel. With our efficient dimAnalysis, this should really help out disambiguating dynamic dimensions that are really identical.

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -324,7 +334,8 @@ class FrontendGenImpl {
       onnx::TypeProto elem_type = input_seq_type.elem_type();
       assert(elem_type.value_case() == onnx::TypeProto::kTensorType &&
              "expect tensor inside sequence type");
-      Type mlir_elem_type = ImportTensorType(elem_type);
+      std::string s;
```
sorenlassen, line 337: a little bit easier to read if you rename `s` to `dimParam`

or consider adding convenience methods `ImportTensorType` and `ImportType` without the dimParam parameter, or make dimParam a pointer that defaults to nullptr

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -287,7 +288,8 @@ class FrontendGenImpl {
    * Import an onnx tensor type by determining and returning its type
    * @param type_proto onnx tensor TypeProto.
```
sorenlassen, line 289: please add a @param description of dimParam

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -533,8 +557,17 @@ class FrontendGenImpl {
...
+    for (uint64_t i = 0; i < inputDimParams.size(); ++i)
+      inputDimParamsRefs.emplace_back(llvm::StringRef(inputDimParams[i]));
+    for (uint64_t i = 0; i < outputDimParams.size(); ++i)
+      outputDimParamsRefs.emplace_back(llvm::StringRef(outputDimParams[i]));
     op->setAttr("input_names", builder_.getStrArrayAttr(inputNames));
     op->setAttr("output_names", builder_.getStrArrayAttr(outputNames));
+    op->setAttr(
+        "input_dim_params", builder_.getStrArrayAttr(inputDimParamsRefs));
+    op->setAttr(
+        "output_dim_params", builder_.getStrArrayAttr(outputDimParamsRefs));
```
sorenlassen, line 570: I think it would be nicer if we made all of these argument and result attributes
https://github.com/sorenlassen/llvm-project/blob/main/mlir/include/mlir/Dialect/Func/IR/FuncOps.td#L241-L245
```
func.func @main_graph(
  %arg0: tensor<?x?xf32> {onnx.name = "X", onnx.dims = "0:batch_size,1:sequence_len"},
  %arg1: tensor<1x?xi64> {onnx.name = "Y", onnx.dims = "1:sequence_len"}
) -> (
  tensor<?x?xf32> : {onnx.name = "Z", onnx.dims = "0:batch_size,1:sequence_len"}
)
```
and then you can omit the onnx.dims attribute in the cases where it's empty

utils/testing/dim_param.py
```diff
@@ -0,0 +1,55 @@
...
+    "test-dim-param",
+    [X, Y],
+    [Z],
+)
+
+# Create the model (ModelProto)
+model_def = helper.make_model(graph_def, producer_name="onnx-mlir")
+
+onnx.checker.check_model(model_def)
+print(MessageToJson(model_def))
```
sorenlassen, line 55: if you like the more concise textual representation of the .onnxtext lit tests you can change the last line to `print(onnx.printer.to_text(model_def))` - see utils/onnx2text.py

sorenlassen (approved):
really nice PR, I had always hoped that your dynamic dimensions framework would apply to onnx's dim_params

src/Builder/FrontendDialectTransformer.cpp
tungld: Great suggestion! I moved to use argument/result attributes instead of function attributes. Thanks!

src/Builder/FrontendDialectTransformer.cpp
tungld: Done. Thanks!

utils/testing/dim_param.py
tungld: It's quite simple so let me keep the current one.

src/Builder/FrontendDialectTransformer.cpp
tungld: Done. Changed to use optional arguments.

PR #2617, opened 2023-11-13
Fix issues about "make check-onnx-backend-dynamic"

negiyas:
This PR fixes the following two issues.

(1) "make check-onnx-backend-dynamic" does not test dynamic cases but tests static cases.
"test/backend/common.py" sets the "TEST_IMPORTER_FORCE_DYNAMIC" environment variable, but "test/backend/variables.py" does not use it. "test/backend/variables.py". 
This PR fixes test/backend/common.py to set the IMPORTER_FORCE_DYNAMIC (instead of "TEST_IMPORTER_FORCE_DYNAMIC") environment variable, which is directly used in the onnx-mlir and onnx-mlir-opt.

(2) If the first issue is fixed, "make check-onnx-backend-dynamic" causes errors with some tests.
(If the first issue is fixed) For example, "test_group_normalization_epsilon_cpu" test causes the following error while compiling the generated model.
```
$ TEST_CASE_BY_USER=test_group_normalization_epsilon_cpu cmake --build . --target check-onnx-backend-dynamic
...
error: failed to materialize conversion for result #0 of operation 'onnx.GroupNormalization' that remained live after conversion
```

Generated model and compilation results are as follows.
```
$ cat test_group_normalization_epsilon.onnx.mlir
module attributes {llvm.data_layout = "E-m:e-i1:8:16-i8:8:16-i64:64-f128:64-v128:64-a:8:16-n32:64", llvm.target_triple = "s390x-ibm-linux", "onnx-mlir.symbol-postfix" = "test_group_normalization_epsilon"} {
  func.func @main_graph(%arg0: tensor<?x?x?x?xf32> {onnx.name = "x"}, %arg1: tensor<?xf32> {onnx.name = "scale"}, %arg2: tensor<?xf32> {onnx.name = "bias"}) -> (tensor<3x4x2x2xf32> {onnx.name = "y"}) {
    %0 = "onnx.GroupNormalization"(%arg0, %arg1, %arg2) {epsilon = 0.00999999977 : f32, num_groups = 2 : si64} : (tensor<?x?x?x?xf32>, tensor<?xf32>, tensor<?xf32>) -> tensor<3x4x2x2xf32>
    onnx.Return %0 : tensor<3x4x2x2xf32>
  }
}
```
```
$ ./build/Debug/bin/onnx-mlir -O3 --EmitLib test_group_normalization_epsilon.onnx.mlir                   
loc("test_group_normalization_epsilon.onnx.mlir":3:10): error: failed to materialize conversion for result #0 of operation 'onnx.GroupNormalization' that remained live after conversion
```

To avoid this compile error, the output shapes need to be set dynamic as well as input shapes.
This PR changes  the output shapes to avoid the compilation errors.

```
$ cat test_group_normalization_epsilon_fixed.onnx.mlir
module attributes {llvm.data_layout = "E-m:e-i1:8:16-i8:8:16-i64:64-f128:64-v128:64-a:8:16-n32:64", llvm.target_triple = "s390x-ibm-linux", "onnx-mlir.symbol-postfix" = "test_group_normalization_epsilon"} {
  func.func @main_graph(%arg0: tensor<?x?x?x?xf32> {onnx.name = "x"}, %arg1: tensor<?xf32> {onnx.name = "scale"}, %arg2: tensor<?xf32> {onnx.name = "bias"}) -> (tensor<?x?x?x?xf32> {onnx.name = "y"}) {
    %0 = "onnx.GroupNormalization"(%arg0, %arg1, %arg2) {epsilon = 0.00999999977 : f32, num_groups = 2 : si64} : (tensor<?x?x?x?xf32>, tensor<?xf32>, tensor<?xf32>) -> tensor<?x?x?x?xf32>
    onnx.Return %0 : tensor<?x?x?x?xf32>
  }
}
```
```
$ ./build/Debug/bin/onnx-mlir -O3 --EmitLib test_group_normalization_epsilon_fixed.onnx.mlir
<No error occurs>
```


negiyas:
I realized that this PR is necessary to solve issue #2639. See that details at issue #2639.

chentong319:
I do not understand why the more accurate output type may cause the problem.
I dig a little bit. GroupNormalization op should be rewritten in the Decompose.cpp.
For the dynamic shape, this check failed! ` if (!onnx_mlir::isRankedShapedType(input.getType()))`
`
   836 	    Value input = groupNormOp.getX();
   837 	    if (!onnx_mlir::isRankedShapedType(input.getType()))
-> 838 	      return failure();
   839 	
   840 	    // Get info.
   841 	    Value scale = groupNormOp.getScale();
Target 0: (onnx-mlir) stopped.
(lldb) p input.dump()
<block argument> of type 'tensor<?x?x?x?xf32>' at index: 0
(lldb) p input.getType().dump()
tensor<?x?x?x?xf32>
(lldb) 
` 
We should fix this issue. 
If we cannot decompose GroupNormalization with dynamic input, just give it an empty dynamic shape in test script.

negiyas:
Closing this PR, since

negiyas:
Closing this PR, since the second issue is solved by PR2725, and no modifications for testing and fixing the issue are necessary.
I created another PR (#2735) to fix the first issue only.

```
(2) If the first issue is fixed, "make check-onnx-backend-dynamic" causes errors with some tests.
(If the first issue is fixed) For example, "test_group_normalization_epsilon_cpu" test causes the following error while compiling the generated model.
```

PR #2620, opened 2023-11-14
Use dim_params in dynamic dimension analysis

tungld:
This patch improves the dynamic dimension analysis by getting the relationship between input and output dimensions via `onnx.dim_params` attribute.


AlexandreEichenberger (approved):
LGTM, if you go through one more rebuild, would just add a bit of cements for the build functions as you have now 2, one that is called for normal ops and one for the function params. Lots of useful debug printouts so that helps understand the code too.

Thanks for this great addition, looking forward to see the improvements.

AlexandreEichenberger:

tungld:
>  if you go through one more rebuild, would just add a bit of cements for the build functions as you have now 2, one that is called for normal ops and one for the function params

Yes, done.

PR #2623, opened 2023-11-15
Update rapidcheck to include the fix for missing <cstdint> include so that it can be compiled by clang 16+

gongsu832:


tungld (approved):
LGTM. Thank you so much!

PR #2626, opened 2023-11-15
Update Protobuf to v21.12

hamptonm1:
Upgrading Protobuf to match the onnx community. In order to upgrade Protobuf we have to include the abseil dependencies as well or else we will receive failures. I am hoping by upgrading Protobuf it will help us with ONNX 1.15.0 upgrade.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,52 @@ set(CMAKE_CXX_STANDARD 17)
 
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
+# Here we are installing the required dependency named "abseil" 
+# in order to upgrade protobuf version to >= 3.21.0
+if (Protobuf_FOUND)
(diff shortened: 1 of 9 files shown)


CMakeLists.txt
```diff
@@ -17,6 +17,52 @@ set(CMAKE_CXX_STANDARD 17)
...
+    absl::synchronization
+    absl::time
+    absl::type_traits
+    absl::utility
+    absl::variant
+    utf8_range::utf8_range
+    utf8_range::utf8_validity
+  )
+endif()
+
```
gongsu832, line 65: No code is setting `Protobuf_FOUND` so this block of code is never run. The successful builds means this block of code is not needed.

hamptonm1: @gongsu832 ... I removed it. Can we approve so I can merge this in?

gongsu832 (approved):
LGTM

PR #2627, opened 2023-11-16
OpenMP support in model runtime

gongsu832:
If LLVM is built with OpenMP support enabled, a corresponding `libOMomp.a` will be built in `$<CONFIG>/lib`. This library can then be embedded into the generated model.so to provide OpenMP support for the model runtime.

### DIFF
--- a/src/Compiler/CMakeLists.txt
+++ b/src/Compiler/CMakeLists.txt
@@ -188,6 +188,9 @@ add_dependencies(OMCompilerUtils cruntime)
 if (ONNX_MLIR_ENABLE_JNI)
   add_dependencies(OMCompilerUtils jniruntime)
 endif()
+if (TARGET omp)
+  add_dependencies(OMCompilerUtils libOMomp)
+endif()
 
 add_onnx_mlir_library(OMCompiler
   OnnxMlirCompiler.cpp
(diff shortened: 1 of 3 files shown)


gongsu832:

chentong319:
I added amp to my llvm-project build and got libomp.dylib, libgomp.dylib and libiomp5.dylib in my build/lib, but no .a. Is this Mac specific or need other flags in cmake configuration?

chentong319:
I used the flags in this PR to build my llvm-project and it worked.
Suggest to modify the document for OMP build, too.

gongsu832:
> I used the flags in this PR to build my llvm-project and it worked. Suggest to modify the document for OMP build, too.

There is no user visible changes to the onnx-mlir build process since it's all "automatic". If llvm-project is built with "clang;openmp", then a `libOMomp.a` is automatically built without any change to the current way onnx-mlir is built. Then in @AlexandreEichenberger 's PR, he just needs to link against `-lOMomp` when generating model.so.

chentong319:
I build natively following docs/BuildOnLinux.md.

AlexandreEichenberger (approved):
LGTM, thanks, got it to work with the additional `-lomp` and `-L .../onnx-mlir/build/src/Runtime/omp/lib/` options.

gongsu832:
> LGTM, thanks, got it to work with the additional `-lomp` and `-L .../onnx-mlir/build/src/Runtime/omp/lib/` options.

Actually, all you need is `-lOMomp`.

PR #2629, opened 2023-11-20
Support string type inputs from Python application code.

negiyas:
This PR supports string type inputs from Python application code.
- This PR supports utils/RunONNXModel.py and test/backend/inference_backend.py except the JNI cases. we assume that numpy.arrays of strings (e.g. `numpy.array([["abc", "d"], ["e", "fghi"]], dtype=np.str_)`).  
- This PR enables two backend tests with string inputs ("test_equal_string_cpu test_equal_string_broadcast_cpu") by "cmake check-onnx-backend."
- JNI interface with string type inputs is not supported, so that tests with string type inputs are skipped with the emit JNI option.
- This PR is similar to PR#2478. The objective might be the same, but the approach to use “object” type for string is different.

The basic strategies are as follows.
-  At Python side, use dtype=object instead of dtype=np.str_ to handle string values as pointers in order to avoid that they are handled as an array of constant-length strings. No limitations with string length.
- At C++ runtime side,
1. Investigate if the dtype is object in order to check the original data type is string or not. (c.f. https://github.com/pybind/pybind11/issues/1538 )
2. Convert the array to a vector of strings (e.g.  by  `inputPyArray.cast<std::vector<std::string>>();` for 1D array).
3. Generate OMTensor for each argument by allocating buffers for the strings and copying the strings to the allocated buffer (c.f. strdup is used in the code.) in order to convert c++ vector type multi-dimentional array into c-type one-dimentional array (used in the OMTensor structure).
4. Invoke the compiled code with the generated OMTensor(s) as input arguments. 
5. Free buffers for the strings in omTensorDestroy.

(Additional comments)

- Confirmed that onnx-mlir can run bidaf-9 model in the model zoo correctly with this PR. (C.f. Refence input files in the model zoo is not correct, input files of the second and third arguments (test_data_set_?/input_{1,2}.pb) should be swapped.)

Comments on the basic strategy and testing policies are very welcome.


negiyas:
@gongsu832 This PR does not support JNI. Can you kindly support to enable the JNI part? Or can you create another PR to support string inputs with JNI?

gongsu832:
> @gongsu832 This PR does not support JNI. Can you kindly give me comments to support to enable the JNI part, or create another PR to support string inputs with JNI?

Let's first decide how users would typically input a tensor of strings for the languages we support. For example, let's say a model specifies an input tensor `[d1, d2, d3]` of strings, how would a user input it?

- for C/C++, it could be a multi-dimensional array of pointers, each pointing to a string, e.g., `char *tensor[d1][d2][d3];`
- for Java, it could be a multi-dimensional array of strings, e.g., `Strings[d1][d2][d3] tensor;`
- for Python, I'm not very familiar with pybind11, how would a user input such a string tensor in Python?

Regardless of how the string tensor is inputted in various languages, ultimately they will all be converted to the `OMTensor` expected input format, which is an one-dimensional `d1*d2*d3` array with shape info. For C/C++, multi-dimensional and one-dimensional arrays don't really differ much since the array elements are all stored contineously in memory in both cases. So conversion between the two is straightforward. But for Java (and for Python also I suspect) the memory layout for multi-dimensional and one-dimensional arrays are very different. So in order to minimize copying, the choice of the input format matters quite a bit. In Java, so far for non-string data, we have used one-dimensional array input, combined with bytebuffer backing, to avoid having to copy input data when crossing Java/native boundary. But I'm not yet sure if we can do the same for string data so I'll have to think a bit. I think we need to do the same for Python, i.e., carefully consider the input format to minimize copying when crossing Python/native boundary.

Also, I would not change the meaning of the `_owning` flag because of the string tensor. When `_owning` is true, it means the API will free the tensor **data**. For string tensor, data means the pointer array plus the strings themselves. So freeing tensor data means freeing both. There shouldn't be a case where the pointer array is freed but the strings are not, or the strings are freed but the pointer array is not.

negiyas:
@gongsu832 Thanks for your variable comments.

> Let's first decide how users would typically input a tensor of strings for the languages we support. For example, let's say a model specifies an input tensor `[d1, d2, d3]` of strings, how would a user input it? 
>     * for C/C++, it could be a multi-dimensional array of pointers, each pointing to a string, e.g., `char *tensor[d1][d2][d3];`
>     * for Java, it could be a multi-dimensional array of strings, e.g., `Strings[d1][d2][d3] tensor;`
>     * for Python, I'm not very familiar with pybind11, how would a user input such a string tensor in Python?

Yes. Although this PR focuses to support Python, but other languages should be considered.
For Python, we assume that numpy.arrays of strings (e.g. `numpy.array([["abc", "d"], ["e", "fghi"]], dtype=np.str_)`). 

> Regardless of how the string tensor is inputted in various languages, ultimately they will all be converted to the OMTensor expected input format, which is an one-dimensional d1*d2*d3 array with shape info.

For Python, "pybind11" converts muti-dimentional Python arrays into the multi-dimentional std::vector types  in c++(e.g. `std::vector<std::vector<std::string>>` for 2D), which are different from the OMTensor style. 
And this PR uses pointers of strings directly for string data themselves to when crossing Python/native boundary. 

> So conversion between the two is straightforward. But for Java (and for Python also I suspect) the memory layout for multi-dimensional and one-dimensional arrays are very different.

The code copies strings in the C++ vector types into one-dimentional array with shape info (1) in order to change the array style (from C++ multi-dim vector to C one-dim array) and (2) duplicate buffers for the string data for keeping the string buffers until "omTensorDestroy."  This is only one copy for each string.  I agree that copies should be minimized, and I suppose that this copy in PR are minimum and mandatory.

> Also, I would not change the meaning of the `_owning` flag because of the string tensor. When `_owning` is true, it means the API will free the tensor **data**. For string tensor, data means the pointer array plus the strings themselves. So freeing tensor data means freeing both. There shouldn't be a case where the pointer array is freed but the strings are not, or the strings are freed but the pointer array is not.

Yes. exactly. I suppose that there are no such cases now, but I am not sure especially about other language cases.
It might to be better to use bitmasks to manage buffers for pointer arrays and strings independently.
Anyway, we need to manage string data themselves by the OMTensor structure to keep them until "omTensorDestroy", I suppose. 

Thank you for your useful comments. I am not sure my comments answers all of your questions, but more comments and suggestions are very welcome!

gongsu832:
I think we should make the string tensor behave as similar to non-string tensor as possible. Currently, C/C++ and Java APIs use the combination of an one-dimensional data array with a shape array instead of a multi-dimensional array for the input/output tensors . This has the advantage that the one-dimensional data array is tightly packed in contiguous memory and therefore they can be passed into the model runtime via `OMTensor` struct without having to copy the data array. As I mentioned before, one-dimensional vs multi-dimensional array doesn't make much difference in C/C++ but matters quite a bit in Java. I I suspect it matters too in Python. I don't know why we used a multi-dimensional array with the Python API. If I were doing it, I would have used one-dimensional data array + shape array as well for Python, if only just for being consistent with other languages.

So back to the string tensor, I think it should also use a combination of one-dimensional string array plus a shape array for input/output string tensors, i.e., something like `[ "foo", "bar", ... ]` (string data) + `[ 2, 3, 4 ]` (shape). Internally, this is represented as an one-dimensional array of pointers to string data (with a total of `2*3*4=24` pointers and strings). This makes a string tensor behave "almost identical" to an int64 tensor, with the difference being that after you get the int64 element you need to dereference it to get the actual string data. This representation of string tensor should in most cases allow passing it by manipulating the pointers therefore avoid having to copy the actual string data.

Regarding the `_owning` flag, I don't think it has anything to do with the frontend language being used. It's a flag for the model runtime API, which is in C.

negiyas:
Thank you for the comments!

> I think we should make the string tensor behave as similar to non-string tensor as possible. Currently, C/C++ and Java APIs use the combination of an one-dimensional data array with a shape array instead of a multi-dimensional array for the input/output tensors . This has the advantage that the one-dimensional data array is tightly packed in contiguous memory and therefore they can be passed into the model runtime via `OMTensor` struct without having to copy the data array. As I mentioned before, one-dimensional vs multi-dimensional array doesn't make much difference in C/C++ but matters quite a bit in Java. I I suspect it matters too in Python. I don't know why we used a multi-dimensional array with the Python API. If I were doing it, I would have used one-dimensional data array + shape array as well for Python, if only just for being consistent with other languages.

I suppose that this PR uses one-dementional data array (of pointers to strings), but does not use multi-dimentional array. 
In order to keep the one-dimentional data array structure, this PR copies string data pointers in from the multi-dimentional data array (generated by "pybind11") to the one-dimentional array (used in the original OMTensor structure). 
(I inserted some words in the description, I am sorry if my descriptions were not enough.)

In this PR, string data are kept in the one-dimentional data array, and shapes are kept in the shape structure. It is same to  numerical data cases. The only difference is that the one-dimentional data array keeps pointers instead of real values, since the data lengths of string elements in an array are different unlike numerical data cases.

> So back to the string tensor, I think it should also use a combination of one-dimensional string array plus a shape array for input/output string tensors, i.e., something like `[ "foo", "bar", ... ]` (string data) + `[ 2, 3, 4 ]` (shape). Internally, this is represented as an one-dimensional array of pointers to string data (with a total of `2*3*4=24` pointers and strings). This makes a string tensor behave "almost identical" to an int64 tensor, with the difference being that after you get the int64 element you need to deference it to get the actual string data. This representation of string tensor should in most cases allow passing it by manipulating the pointers therefore avoid having to copy the actual string data.

 I suppose that this PR keeps the one-dimentional array structure. A string tensor behaves "almost identical" to an int64 tensor as described in this comment.

> Regarding the `_owning` flag, I don't think it has anything to do with the frontend language being used. It's a flag for the model runtime API, which is in C.

I understand the idea, but onnx-mlir runtime code needs to hold the string data (not pointers) in the OMTensor structure until "omTensorDestroy" is called, because Python applications free string data used for an OMTensor for the first argument before generating another OMTensor for the next argument.

Thank you very much for the comments, and I am sorry if my explanations were not enough or I do not understand your comments. Your comments are important and very welcome to me.

gongsu832:
> I suppose that this PR uses one-dementional data array (of pointers to strings), but does not use multi-dimentional array. In order to keep the one-dimentional data array structure, this PR copies string data pointers in from the multi-dimentional data array (generated by "pybind11") to the one-dimentional array (used in the original OMTensor structure). (I inserted some words in the description, I am sorry if my descriptions were not enough.)
> 
> In this PR, string data are kept in the one-dimentional data array, and shapes are kept in the shape structure. It is same to numerical data cases. The only difference is that the one-dimentional data array keeps pointers instead of real values, since the data lengths of string elements in an array are different unlike numerical data cases.

What I meant is that the Python interface, instead of using a multi-dimensional numpy array, should have used an one dimensional array plus a shape array, like the other languages.

>I understand the idea, but onnx-mlir runtime code needs to hold the string data (not pointers) in the OMTensor structure until "omTensorDestroy" is called, because Python applications free string data used for an OMTensor for the first argument before generating another OMTensor for the next argument.
>
I'm not sure I understand this. Can you provide an example?

include/onnx-mlir/Runtime/OMTensor.h
```diff
@@ -115,11 +115,15 @@ OM_EXTERNAL_VISIBILITY OMTensor *omTensorCreate(
...
- * release the data_ptr upon destruction.
+ * @param owning whether OMTensor owns the data, if set to
+ * OMTENSOR_OWNING_DATA_PTR, OMTensor will release the data_ptr upon
+ * destruction. If set to OMTENSOR_OWNING_DATA_PTR_AND_STRING, OMTensor will
+ * free both the data_ptr and its corresponding strings upon destruction.
  * @return pointer to OMTensor created, NULL if creation failed.
  *
  */
+#define OMTENSOR_OWNING_DATA_PTR 1
+#define OMTENSOR_OWNING_DATA_PTR_AND_STRING 2
```
chentong319, line 126: Your change can support pointers to data other than string, right?

negiyas:
> What I meant is that the Python interface, instead of using a multi-dimensional numpy array, should have used an one dimensional array plus a shape array, like the other languages.

Thanks. Now, I understand your points. I just wanted to treat multi-dimentional string input data same to multi-dimentional numerical input data.

I am investigating how the current Python test codes (especially utils/RunONNXModel.py and test/backend/inference_backend.py) treats mumti-dimentional (numerical) data.

I suppose that shape array is already passed by the test codes, 



 and logic for receiving multi-dimentional data array from applications may not be necessary. 



I do not intend to change treatment of multi-dimentional array for supporting string data.


I just want to follow the way for numerical data cases, and shape info should be given




We need to manage multi-dimentional



For numerical numbers in Python applications, multi-dimentional numpy array 


Current onnx-mlir for numerical numbers compiler uses
In the Python 


> 
> > I understand the idea, but onnx-mlir runtime code needs to hold the string data (not pointers) in the OMTensor structure until "omTensorDestroy" is called, because Python applications free string data used for an OMTensor for the first argument before generating another OMTensor for the next argument.
> 
> I'm not sure I understand this. Can you provide an example?

include/onnx-mlir/Runtime/OMTensor.h
chentong319: Current implementation stores char * pointer in the Tensor of string. If a Tensor of string and the strings are created inside the compiler, we will have to allocate the space for data_ptr and for the strings.  That will be a new case for the ownership. If the Tensor of strings are just propagate pointers from one tensor to another, the ownership will be the same as the float or integer data type.

negiyas:
> What I meant is that the Python interface, instead of using a multi-dimensional numpy array, should have used an one dimensional array plus a shape array, like the other languages.

Thanks. I have got the point now. My comments are as follows.

(1) The current implementation uses multi-dimensional numpy for input/output Tensors, that is a standard array format in Python.  This PR simply follows the current one for the numerical data cases, does not change the policy including multi-dimensional array treatment.
(2) "numpy" in Python originally has the shape attribute for managing array shape explicitly as same to the OMTensor. The shape attribute defines the shape of numpy. Users gets the array's shape by referring the shape attribute, and changes the shape by modifying the shape attribute. 
(3) The "pybind" module and current onnx-mlir runtime module use the numpy shape attribute of input/output arrays to pass their shapes. To utilize them, we need to use a standard numpy shape attribute for multi-dimensional arrays.

gongsu832:
> > What I meant is that the Python interface, instead of using a multi-dimensional numpy array, should have used an one dimensional array plus a shape array, like the other languages.
> 
> Thanks. I have got the point now. My comments are as follows.
> 
> (1) The current implementation uses multi-dimensional numpy for input/output Tensors, that is a standard array format in Python. This PR simply follows the current one for the numerical data cases, does not change the policy including multi-dimensional array treatment. (2) "numpy" in Python originally has the shape attribute for managing array shape explicitly as same to the OMTensor. The shape attribute defines the shape of numpy. Users gets the array's shape by referring the shape attribute, and changes the shape by modifying the shape attribute. (3) The "pybind" module and current onnx-mlir runtime module use the numpy shape attribute of input/output arrays to pass their shapes. To utilize them, we need to use a standard numpy shape attribute for multi-dimensional arrays.

I understand you are trying to follow the current way of using numpy multi-dimensional array for the Python wrapper. But that's exactly what I'm questioning, that whether this is the best way of doing it. I think using a multi-dimensional numpy array actually makes things unnecessarily complicated because you have to "flatten" it to what `OMTensor` expects, an one-dimensional array. If you had used an one-dimensional array (plus a shape array) to begin with, pybind11 probably would have converted it to a vector instead of vector of vector. So there is no need to "flatten" it further. In addition, you would also be able to handle arbitrary number of dimensions instead of hardcode a maximum 4 dimensions in `PyExecutionSessionBase.cpp`, which looks rather hacky.

include/onnx-mlir/Runtime/OMTensor.h
negiyas: @gongsu832 Thanks for the comments. I changed the code according to your comments. I updated the code to use the OMTensor->_allocated buffer for string data themselves. And there are no modifications on src/Runtime/OMTensor.inc from the original code now.

The string data by allocated by Python applications are copied into memory area in the OMTensor->_allocated buffer, which follows the area for array of string pointers (in the PyExecutionSessionBase::pyRun function in PyExecutionSessionBase.cpp). The _allocated buffer is freed in the mTensorDestroy like other cases.

I hope that this update can answer your concerns related to the OMTensor->_owning flag.
Please let me know your additional comments and suggestions.

(Let me explain my thought about the multi-dimensional array in Python by another comments.)

negiyas:
> I understand you are trying to follow the current way of using numpy multi-dimensional array for the Python wrapper. But that's exactly what I'm questioning, that whether this is the best way of doing it. 

Because "pybind11" uses shape defined the shape attribute of numpy, we cannot flat the multi-dimensional arrays in Python code to keep the original shape in order to utilize the pybind code.  According to my understanding, all special treatments for numpy is done in runtime code for Python in "src/RuntimePyExecutionSessionBase.cpp".
In the case of Python, in comparing with other languages, it is considered that the runtime code and pybind11 module are a part of Python application to prepare shape and uni-dimensional arrays. 

>I think using a multi-dimensional numpy array actually makes things unnecessarily complicated because you have to "flatten" it to what OMTensor expects, an one-dimensional array. If you had used an one-dimensional array (plus a shape array) to begin with, pybind11 probably would have converted it to a vector instead of vector of vector.  So there is no need to "flatten" it further. In addition, you would also be able to handle arbitrary number of dimensions instead of hardcode a maximum 4 dimensions in PyExecutionSessionBase.cpp, which looks rather hacky.

Yes, exactly. As described in your comments, for the numerical data cases, we need NOT to flatten multi-dimensional array manually, I tried to follow the same way for string data cases, but I could not follow the way for string data because of a pybind11 issue.  I will update the code if we find better way to avoid the issue or pybind11 fixes the issue.

I appended the following comments in the code to explain the situation.

```
      //
      // Convert multi-dimensional array (for string data pointers) to
      // one-dimensional array to mange multi-dimensional array in an
      // integrated way.
      //
      // For numerical array, pybind11 can convert multi-dimensional array into
      // one-dimensional array without manual conversion, but pybind11 has
      // an issue about "dtype caster does not accept strings or type objects"
      // (https://github.com/pybind/pybind11/issues/1538). The issue page
      // shows a solution to avoid the bug, but it does not work for our case.
      // The following part solves the issue temporally, and will be updated
      // if we find better way to avoid the issue or pybind11 fixes the issue.
      //
```

gongsu832:
The current `PyExecutionSessionBase::pyRun` function 
```
std::vector<py::array> PyExecutionSessionBase::pyRun(const std::vector<py::array> &inputsPyArray)
```
takes a vector of `py::array`, which is converted from `numpy.ndarray` by pybind11. What I was trying to say is that there is nothing that says we have to do this. We could always make the function take two vectors, one is the `inputsPyArray` like we have now but the original `numpy.ndarray` is flattened, and the other a new `shapesPyArray` that carries the shape of the original unflattened `numpy.ndarray`.

This way you will be able to handle arbitrary dimension without having to hardcode a maximum supported dimension. That part of the code is really ugly. Yes this means that more test code will have to be changed, e.g., code like
```
outputs = session.run(inputs)
```
will need to be changed to something like
```
outputs = session.run(map(lambda t: t.flatten(), inputs), map(lambda t: t.shape, inputs))
```
But IMO this is the right way to do it.

chentong319:
A side question, for an integer or float input tensor, do we have to allocate a new buffer and copy the data if flattened? I mean to use the data pointer from np.array. Is there any copy when py.array is converted to np.array by pybind? The reason I am asking these question is that the input tensors for KV caching may be quite large.

gongsu832:
> A side question, for an integer or float input tensor, do we have to allocate a new buffer and copy the data if flattened? I mean to use the data pointer from np.array. Is there any copy when py.array is converted to np.array by pybind? The reason I am asking these question is that the input tensors for KV caching may be quite large.

Good question. When you flatten a `numpy.ndarray` with `flatten()`, it always makes a copy. So a better function is `ravel()`. According to its doc, it returns an one-dimensional "view" of the `numpy.ndarray` and only makes a copy if needed. Although it's not clear from the doc what condition makes the copy needed, my guess is that as long as the multi-dimensional array is contiguous in memory, `ravel()` doesn't make a copy.

When pybind11 converts the `numpy.ndarray` to a native array, it doesn't make a copy unless the array is immutable, according to the code in `PyExecutionSessionBase::pyRun` (https://github.com/onnx/onnx-mlir/blob/552c2c1cc0b3674f8ba3d7a836d2be7e3a1449f6/src/Runtime/PyExecutionSessionBase.cpp#L78-L89)

negiyas:

negiyas:
@gongsu832 > This way you will be able to handle arbitrary dimension without having to hardcode a maximum supported dimension. That part of the code is really ugly. Yes this means that more test code will have to be changed, e.g., code like `outputs = session.run(inputs)`  will need to be changed to something like `outputs = session.run(map(lambda t: t.flatten(), inputs), map(lambda t: t.shape, inputs))`

Thanks for the comments. I fixed this PR according to your comments, so that this PR does not have hardcoded maxmum supported dimension now.  I changed codes for calling session.run with the following three arguments. All arguments are numpy.ndarray to utilize the pybind11 module, which supports numpy.ndarray (not list of Python).

1. "data": numpy.ndarray of the flatten input data.
2.  "shape": numpy.ndarray of the original input shape.
3.  "strides": numpy.ndarray of the original input stride.

```
run_inputs = []
run_shapes = []
run_strides = []
for inp in inputs:
    run_inputs.append(inp.flatten())
    run_shapes.append(np.array(inp.shape, dtype=np.int64))
    run_strides.append(np.array(inp.strides, dtype=np.int64))
outs = sess.run(run_inputs, run_shapes, run_strides)
```
 
 I hope that updated PR resolves your concerns. Thanks.

gongsu832:
@negiyas Thanks for the updates. It's much nicer without the hardcoded dimension limit. In order to use this new Python API, I think `py_om_execution_session.py` should become part of the `src/Runtime` (perhaps renamed to something like `PyOMExecutionSession.py`) and gets installed along with other runtime libraries. Although it should probably get installed into python package directory instead of /usr/local/lib using the python setuptools (we can address that in a separate PR). So users will just do
```
from PyOMExecutionSession import PyOMExecutionSession
```
in order to use the Python API. And more Python APIs can be added to it in the future.

negiyas:
@gongsu832 Thanks for your useful comments.

> Although it should probably get installed into python package directory instead of /usr/local/lib using the python setuptools (we can address that in a separate PR).  in order to use the Python API. And more Python APIs can be added to it in the future.

I agree that 'PyOMExecutionSession.py' should be installed as a python package to enable users to use  `import PyOMExecutinoSession` simply.

If you are O.K., I would like to keep this PR without using python setuptools, and will create another PR to install Python packages. Accordingly more Python API can utilize the new APIs in the future.

Your comments are very welcome.

negiyas:

gongsu832:
> @gongsu832 Thanks for your useful comments.
> 
> > Although it should probably get installed into python package directory instead of /usr/local/lib using the python setuptools (we can address that in a separate PR).  in order to use the Python API. And more Python APIs can be added to it in the future.
> 
> I agree that 'PyOMExecutionSession.py' should be installed as a python package to enable users to use `import PyOMExecutinoSession` simply.
> 
> If you are O.K., I would like to keep this PR without using python setuptools, and will create another PR to install Python packages. Accordingly more Python API can utilize the new APIs in the future.
> 
> Your comments are very welcome.

@negiyas It's fine to have another PR to deal with the setuptools part. But I think `PyOMExecutionSession.py` should be under a more logical directory such as `src/Runtime/python`. The current directory structure is such that `src/Runtime` has the C/C++ APIs, `src/Runtime/jni` has the Java APIs (I'm thinking `src/Runtime/java` may be a better name). So it seems logical to put Python APIs under `src/Runtime/python`.

negiyas:

negiyas:
@gongsu832 > It's fine to have another PR to deal with the setuptools part. But I think `PyOMExecutionSession.py` should be under a more logical directory such as `src/Runtime/python`. The current directory structure is such that `src/Runtime` has the C/C++ APIs, `src/Runtime/jni` has the Java APIs (I'm thinking `src/Runtime/java` may be a better name). So it seems logical to put Python APIs under `src/Runtime/python`.

Thanks for your your comments. I am moving PyOMCompileExecutionSession.{hpp,cpp} and other existing Python-related codes (e.g. PyExecutionSession.{hpp,cpp}, PyExecutionSessionBase.{hpp,cpp}) under src/Runtime to src/Runtime/python.

negiyas:
@gongsu832 I am sorry, but I cannot understand your comments about python package fully.

gongsu> In order to use this new Python API, I think `py_om_execution_session.py` should become part of the `src/Runtime` (perhaps renamed to something like `PyOMExecutionSession.py`) and gets installed along with other runtime libraries. Although it should probably get installed into python package directory instead of /usr/local/lib using the python setuptools (we can address that in a separate PR). So users will just do `from PyOMExecutionSession import PyOMExecutionSession` in order to use the Python API. And more Python APIs can be added to it in the future.

I have questions about the comments.
(1) Does "python package" mean [so-called python package](https://packaging.python.org/en/latest/tutorials/installing-packages/) installed by pip.
(2) What does "python package directory" mean?

I am happy if you kindly give me more information about what should we do to apply your comments.

gongsu832:
Imagine you do a "make install" then delete the onnx-mlir source repo. In order to be able to use the python API, your python wrapper and its related pybind11 .so need to be installed into a directory searched by python. When you install them using setuptools (on top of which pip runs), they are typically installed under `/usr/local/lib/python3.xx/dist-packages`.

negiyas:
@gongsu832 Thanks for comments about python installation issue.
Since we agreed that the issue will be managed by another PR, I hope that this PR can be merged at first.
Can you kindly teach me what are necessary items for this PR to be merged?

gongsu832:
@negiyas I think the last thing left to be done for this PR is to handle `py_om_execution_session.py` a bit more gracefully without a "make install".

Currently you have `py_om_execution_session.py` in two places, `backend` and `utils`. Basically, you need `py_om_execution_session.py` in every directory where you use it. That's a bit cumbersome. Instead, I think you should have a single copy of `py_om_execution_session.py` under `src/Runtime/python` and have it copied to `CMAKE_LIBRARY_OUTPUT_DIRECTORY` (which is set to `${CMAKE_BINARY_DIR}/$<CONFIG>/lib`) during cmake. That directory is search by most of our python code by doing a `sys.path.append`.

negiyas:
@gongsu832 Thank you so much for your very helpful comments.
I updated this PR to apply your comments as follows.
1.  Integrated two py_om_execution_session.py files into "src/Runtime/python/py_om_execution_session.py".
2.  Add the following lines in "src/Runtime/python/CMakeLists.txt" to copy the python code in the target machine.
```
file(GENERATE
  OUTPUT ${CMAKE_LIBRARY_OUTPUT_DIRECTORY}/py_om_execution_session.py
  INPUT ${CMAKE_CURRENT_SOURCE_DIR}/py_om_execution_session.py
  )
```

gongsu832:
@negiyas Thanks for the updates. Nitpicking a bit...can we rename `py_om_execution_session.py` to `PyOmExecutionSession.py` to be consistent with the naming of other files?

negiyas:
@gongsu832 Thanks. I renamed py_om_execution_session.py to `PyOMExecutionSession.py`.

gongsu832 (approved):
LGTM. Thanks for the updates.

src/Runtime/python/PyExecutionSession.hpp
```diff
@@ -34,7 +34,8 @@ PYBIND11_MODULE(PyRuntime, m) {
       .def("entry_points", &onnx_mlir::PyExecutionSession::pyQueryEntryPoints)
       .def("set_entry_point", &onnx_mlir::PyExecutionSession::pySetEntryPoint,
           py::arg("name"))
-      .def("run", &onnx_mlir::PyExecutionSession::pyRun, py::arg("input"))
+      .def("run", &onnx_mlir::PyExecutionSession::pyRun, py::arg("input"),
+          py::arg("shape"), py::arg("stride"))
```
tungld, line 38: @negiyas could you update all documents about `PyExecutionSession` and `PyOMCompileExecutionSession`? since their run methods now need two additional params. Or recommend users to switch to `PyOMExecutionSession` instead.
I see several places (and maybe more):
- https://github.com/onnx/onnx-mlir/blob/main/docs/UsingPyRuntime.md
- https://github.com/onnx/onnx-mlir/blob/main/docs/mnist_example/README.md

src/Runtime/python/PyExecutionSessionBase.cpp
```diff
@@ -54,6 +54,40 @@ struct npy_format_descriptor<onnx_mlir::float_16> {
...
+    numElem *= shape[i];
+  uint64_t strLenTotal = 0;
+  uint64_t off = 0;
+  void *dataBuffer = NULL;
+  assert(pyArray.ndim() == 1 && "input pyArray should be flatten");
+  auto vec = pyArray.cast<std::vector<std::string>>();
+  for (int64_t i = 0; i < shape[0]; ++i)
+    strLenTotal += strlen(vec[i].data()) + 1;
+  dataBuffer = malloc(sizeof(char *) * numElem + strLenTotal);
+  assert(dataBuffer && "fail to alloc bufferData for string data");
```
AlexandreEichenberger, line 80: If possible when there are errors that are linked to the inputs, we try to have more friendly error messages, like throwing exceptions with a test message indicating the source of the error. Please do so if feasible.

src/Runtime/python/PyExecutionSessionBase.cpp
```diff
@@ -62,15 +96,24 @@ PyExecutionSessionBase::PyExecutionSessionBase(
...
+    const std::vector<py::array> &stridesPyArray) {
   if (!isInitialized)
     throw std::runtime_error(reportInitError());
   if (!_entryPointFunc)
     throw std::runtime_error(reportUndefinedEntryPointIn("run"));
 
   // 1. Process inputs.
   std::vector<OMTensor *> omts;
-  for (auto inputPyArray : inputsPyArray) {
+  assert((inputsPyArray.size() == shapesPyArray.size()) &&
```
AlexandreEichenberger, line 109: same here, if the inputs come from the user (which I believe they do), I suggest to have a friendlier error message via throwing an exception.

src/Runtime/python/PyExecutionSessionBase.hpp
```diff
@@ -36,7 +36,9 @@ class PyExecutionSessionBase
       bool defaultEntryPoint = true);
   std::vector<std::string> pyQueryEntryPoints();
   void pySetEntryPoint(std::string entryPointName);
-  std::vector<py::array> pyRun(const std::vector<py::array> &inputsPyArray);
+  std::vector<py::array> pyRun(const std::vector<py::array> &inputsPyArray,
```
AlexandreEichenberger, line 39: Can you add a comment here on what are the expectations on the inputs? Like each array has to have the same number of entires... Also, I understand the array (data input) and the shape (size of the data). Can you add an explanation on the strides?

AlexandreEichenberger (changes requested):
I left a few minor comments about asserts that can be easily changed. But I have a bigger question, which I would like to have addressed.

Before: we had `OMExecutionSession` and `OMCompileExecutionSession` python classes to perform execution and compilation&execution.

Now you have added (for reasons that @gongsu832 and you agreed on, I assume) to modify the run method of these 2 python class to take 3 parameters instead of 1.

Then you also added a new python class, `PyOMExecutionSession` that provides back the 1 parameter `run` method.

2 Questions:
* Why is there not also a `PyOMCompileExecutionSession`? It feels just as needed as for the execution alone?
* Do you ever anticipate the user to need to access the old classes (the one with the 3-argument run class)? To me, it looks like that is not the case. Users will want to use the `PyOMExecutionSession` (and  `PyOMCompileExecutionSession`) because they are more convenient.

So if the answer is yes and yes to these 2 questions, I would recommend to rename the `PyOM...` class to drop the `Py` so as to preserve the old name; and rename the `OM...` class as `OMInternal...` as they are only used internally to provide our support and are not meant to be used directly.

As a result of that, there is no need to change any of the existing code base as we preserve the same (external) interfaces of having only 1 parameter for the `run` operation.

Quick question: does this new 2 step interface have an impact on performance? E.g. does it increase the number of data copy (as they are possibly very large), especially for the NON-String inputs?

Please respond to these questions, and depending on the answer, change the code accordingly.

src/Runtime/python/PyExecutionSession.hpp
negiyas: @tungld I will update the documents and example codes after all APIs are finalized. Thanks.

negiyas:
@AlexandreEichenberger I have a bigger question, which I would like to have addressed.

Thanks for your comments. I am changing 

> 2 Questions:
> (Q1) Why is there not also a PyOMCompileExecutionSession? It feels just as needed as for the execution alone?
> (Q2) Do you ever anticipate the user to need to access the old classes (the one with the 3-argument run class)? To me, it looks like that is not the case. Users will want to use the PyOMExecutionSession (and PyOMCompileExecutionSession) because they are more convenient.

The answers are yes and yes.

> So if the answer is yes and yes to these 2 questions, I would recommend to rename the PyOM... class to drop the Py so as to preserve the old name; and rename the OM... class as OMInternal... as they are only used internally to provide our support and are not meant to be used directly.

Yes. I agreed with @gongsu832 to change the argument number of "OMExecutionSession" function in the "PyRuntime" module from one to three in order to keep language-independent interface consistent to other languages. 

To implement the agreement,
- At first, I changed the "OMExecutionSession" function.
- Then, I introduced "PyOMExecutionSession" function for Python programmers' convenience. 

However, I realized that 






"OMExecutionSession" module implementation at first, then I introduced "PyOMExecutionSession" for user convenience later. 

 
 
Since I kept the old module name("PyRuntime") and function name("OMExecutionSession"), and added a new module to hide the API change

As a result of that, there is no need to change any of the existing code base as we preserve the same (external) interfaces of having only 1 parameter for the run operation.

> Quick question: does this new 2 step interface have an impact on performance? E.g. does it increase the number of data copy (as they are possibly very large), especially for the NON-String inputs?

Please respond to these questions, and depending on the answer, change the code accordingly.

negiyas:
@AlexandreEichenberger Thanks for your thorough comments.
> I have a bigger question, which I would like to have addressed.

Let me answer this bigger question at first. I will answer other questions next.
Basically speaking, I updated this PR, which preserves the external Python interfaces and still has the same internal("pybind") interface with three arguments.

> Before: we had OMExecutionSession and OMCompileExecutionSession python classes to perform execution and compilation&execution.
> Then you also added a new python class, PyOMExecutionSession that provides back the 1 parameter run method.

Thanks for the summary. It is correct.

> 2 Questions:
> Why is there not also a PyOMCompileExecutionSession? It feels just as needed as for the execution alone?
> Do you ever anticipate the user to need to access the old classes (the one with the 3-argument run class)?
 
The answers are yes and yes.

> As a result of that, there is no need to change any of the existing code base as we preserve the same (external) interfaces of having only 1 parameter for the run operation.
> Now you have added (for reasons that @gongsu832 and you agreed on, I assume) to modify the run method of these 2 python class to take 3 parameters instead of 1.

Yes. I agreed with @gongsu832 to have class with 3-argument run class for execution, and developed this PR based on your summary.  Now I have realized that we can develop this PR, and have updated it, which preserves the external Python interfaces, and still has the same internal("pybind") interface with three arguments.

1. Change the PyRuntime/PyCompileAndRuntime module names to PyRuntimeC/PyCompileAndRuntimeC, which have 
OMExecutionSession/OMCompileExecutionSession class with 3-argument "run" function.
2. Introduce PyRuntime.py/PyCompileAndRuntime.py, which has OMExecutionSession/OMCompileExecutionSession class with 1-argument run function, which preserve the existing Python interfaces fully.

I hope that the latest PR answers most comments of @AlexandreEichenberger  and @gongsu832.
Further comments on the latest PR are very welcome!

> Quick question: does this new 2 step interface have an impact on performance? E.g. does it increase the number of data copy (as they are possibly very large), especially for the NON-String inputs?

I suppose that this PR has a negligible impact on performance. Although we need to pass two additional arguments (shapes and strides) though the "pybind" module, the arguments are quite small (1D with input rank size) and are passed only once at whole module execution.

src/Runtime/python/PyExecutionSession.hpp
negiyas: @tungld In the latest PR, no APIs for pythons are changed from the main branch, as you can see in the following comments.
Accordingly no documents need to be updated.

src/Runtime/python/PyExecutionSessionBase.cpp
negiyas: Updated the code to throw a python exception if memory allocation fails. Thanks.

src/Runtime/python/PyExecutionSessionBase.cpp
negiyas: Fixed in a similar way to the above.

src/Runtime/python/PyExecutionSessionBase.hpp
negiyas: Comments added. Thanks.

src/Runtime/python/PyExecutionSessionBase.hpp
```diff
@@ -21,7 +21,7 @@
 
 namespace py = pybind11;
 
-#include "ExecutionSession.hpp"
+#include "../ExecutionSession.hpp"
```
AlexandreEichenberger, line 24: Nit, I believe in most places we give the path from the `onnx-mlir` root dir. I would use that here if you don't mind.

src/Runtime/python/PyRuntime.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python3
+
+############# PyOMRuntime.py #######################################
+#
+# Copyright 2021-2023 The IBM Research Authors.
```
AlexandreEichenberger, line 5: Nit, please update all of the copyright to include year 2024, here and elsewhere.

AlexandreEichenberger (approved):
LGTM, thanks for the changes, I am grateful that you found a way to not change the external interface while satisfying the needs for the 3-arguments run method. Please confirm that it is indeed the case.

I left 2 small nits, approved the PR but believe you should correct them. One is about the relative include path, and the other is about the copyright not being updated to 2024.

Thanks for the good work.

AlexandreEichenberger:
PS: I see the python format failing for other PRs as well, so no need to worry about this.\

I asked @gongsu832 to look into why it reports a failure

gongsu832:
> LGTM, thanks for the changes, I am grateful that you found a way to not change the external interface while satisfying the needs for the 3-arguments run method. Please confirm that it is indeed the case.
> 
I think there might be some misunderstanding. The external interface will change to be 3 arguments. The original 1-argument interface uses shape info embedded in the numpy array and the python-to-native conversion is done by pybind11 directly. But it has trouble supporting string data type because the multi-dimensional array must be "flattened" to 1-dimension by copying each dimension manually in C/C++ since pybind11 currently does not directly support string data type. With the 3-argument interface, the numpy array is "flattened" in the new python wrapper (and hence the need for additional arguments to carry shape info) and this "flatten" typically is a simple "view change" without copying. This 3-argument interface is also more similar to C/C++ and Java interfaces.

gongsu832:
> PS: I see the python format failing for other PRs as well, so no need to worry about this.\
> 
> I asked @gongsu832 to look into why it reports a failure

The problem is that MacOS is now using a newer version of black 24.1.0. So on your local machine you should upgrade black to that version.

negiyas:

src/Runtime/python/PyRuntime.py
negiyas: I updated "2023" to "2024", in files included in this PR, and other files including "-2023" still remains.

src/Runtime/python/PyExecutionSessionBase.hpp
negiyas: Fixed. Thanks!

negiyas:
@gongsu832 I updated the black command, and also updated format of utils/analyze-simd.py and utils/pre-onnx-mlir.py to fix the black format errors. Thanks.

negiyas:
> @gongsu832 The external interface will change to be 3 arguments. The original 1-argument interface uses shape info embedded in the numpy array and the python-to-native conversion is done by pybind11 directly. But it has trouble supporting string data type because the multi-dimensional array must be "flattened" to 1-dimension by copying each dimension manually in C/C++ since pybind11 currently does not directly support string data type.

Yes. Exactly. Thanks for pointing out this issue. I remember this issue again now:-).
@gongsu832 Thank you again for reviewing this complicated PR!

cjvolzka:
After pulling in this change, I'm noticing that the PyRuntime files are coming out as `PyRuntimeC.cpython-*-linux-gnu.so` Also to get our Python examples to work, I have to do `from PyRuntimeC import OMExecutionSession`

As is, that would then break existing client applications. Their programs would need to be updated for the change. Also I haven't tried it yet, but I'm guessing the onnx-mlir examples at https://github.com/onnx/onnx-mlir/blob/main/docs/mnist_example/mnist-runPyRuntime.py#L4 are broken because it still has `from PyRuntime` instead of `from PyRuntimeC`

If it breaks clients, then that should be a major version change (ie 0.5.0) for onnx-mlir. Is that desired? For zDLC, after generating the `PyRuntimeC.cpython-*-linux-gnu.so`, I can probably rename the `PyRuntime.cpython-*-linux-gnu.so` to avoid breaking exploiters but I wonder if we should revert the name back in onnx-mlir as a whole.

gongsu832:
Sorry I didn't notice this when reviewing the PR. I think the change is due to the original `PyRuntime` generated by pybind11 conflicting with the new `PyRuntime.py` wrapper. @negiyas I think we should keep the original `PyRuntime` in order not to break the existing python clients. We can perhaps name the new python wrapper `PyOMRuntime.py`.

@cjvolzka keep in mind that the original `PyRuntime` will not support string data type. To use string data type, you must switch to `PyOMRuntime`.

src/Runtime/python/PyExecutionSession.hpp
cjvolzka: This breaks compatibility with existing client applications as the interface to .run() now takes three inputs instead of one.

Do we need to pass in "shape" and "stride" separately? After the "input" is passed in to the interface, internally could we not get this info from input.shape and input.strides?

cjvolzka:
> I can probably rename the PyRuntime.cpython-*-linux-gnu.so to avoid breaking exploiters but I wonder if we should revert the name back in onnx-mlir as a whole.

Actually I was wrong on this. I missed that the .run() interface changed. So that wouldn't actually work to avoid breaking exploiters. See comment at https://github.com/onnx/onnx-mlir/pull/2629#discussion_r1476594493

src/Runtime/python/PyExecutionSession.hpp
cjvolzka: I forget we can take in a list of arrays. Although I think the idea would still apply. Get the shape and stride info for each tensor in the run "input" list. That would also remove the need to check that each of the three lists (input, shape, stride) need to be the same shape and the user needing to create those additional lists when the numpy arrays already have that info. 

(also side nit, they should be called "inputs" "shapes" and "strides" since we expect lists containing 1 or more of those things.)

gongsu832: @cjvolzka The reason the shape and stride are passed in separately is because the input tensor of a multi-dimensional array of objects (pybind11 does not support string type conversion yet so we have to use generic object type) must be "flattened" to one-dimension when calling the native C/C++ runtime. If we pass in a multi-dimensional input tensor and let pybind11 do the conversion, in C++ code we would get a vector of vector of vector ... and it's very cumbersome to "flatten" this into one-dimension, requiring copying each dimension manually. Instead, doing a conversion in python (using the ravel() method of the numpy array) typically involves just a "view" change without copying at all since the original multi-dimensional input tensor is in most cases already contiguous in memory (unless you did slice etc. operations to the tensor in which case ravel() will do a copy when flattening).

So the change of the input is not so much about how many additional parameters are introduced, but rather the requirement of always passing in an one-dimensional tensor to pybind11.

tungld: @gongsu832 I just think whether pybind11 supports optional arguments so that we can pass a single input for all the cases rather than string. And, if input is string, we have an internal check to require the other two arguments.

gongsu832: With the Python wrapper, it's a lot more flexible to offer the API to the end user, as long as internally we "flatten" the input tensor before invoking pybind11. So taking a single input numpy array (which has all the shape info available) certainly works. Although it's typically a good practice to make the APIs of different languages as similar as possible. In this case, the C/C++ and Java APIs both take an one-dimensional array and a shape array (C/C++ API needs to take additional parameters such as rank and data type since it's typeless). So Python API should in general follow that theme using basic Python array because not everybody may be using numpy array for whatever reason. The single input numpy array API can be offered as an "added sugar" on top of the "normal" API.

tungld: It makes sense to unify to using one-dimensional array in pybind11, and the Python wrapper will do whatever it can to support other data types, such as ones in numpy, torch, etc. So what is missing is a convenient way to deploy the python wrapper, e.g. pip package.

tungld: FYI: this is a guideline about creating a .so file to package python scripts: https://fransiska.github.io/2023/09/14/creating-so-files-from-python 

I wonder if it is good to generate a PyRuntime*.so to contain the python wrapper. Then users will need two .so files (PyRuntime*.so and PyRuntimeC*.so) to keep their python program unchanged. Or, it is even better if we can package both the python wrapper and PyRuntimeC*.so in a single .so file, so that there is no change at the user side.

cjvolzka: > @gongsu832 I just think whether pybind11 supports optional arguments so that we can pass a single input for all the cases rather than string. And, if input is string, we have an internal check to require the other two arguments.

If this is possible, this would be the way I'd like to see this go. We keep backwards compatibility while enabling string support.

> Although it's typically a good practice to make the APIs of different languages as similar as possible. In this case, the C/C++ and Java APIs both take an one-dimensional array and a shape array (C/C++ API needs to take additional parameters such as rank and data type since it's typeless). So Python API should in general follow that theme using basic Python array because not everybody may be using numpy array for whatever reason.

I agree with this in principle but not in this implementation. Since this is an already established API, we need to plan out with backwards incompatibility a primary concern and not an afterthought, even if the new way is more consistent.

For example, we keep the interface to .run() "as is" using numpy arryies and add a deprecated warning to use a new run_model() (name tbd). Then .run_model() has whatever new requirements we want (ie always be flattened, but see below for more). Then after a release or two, we can drop the "old" .run() since it is deprecated.

Second, if we are trying to keep the interfaces consistent between languages, the .run_model() should not take an array, numpy array or a list of those. We should have OMTensor and OMTensorList Python objects. I've actually had many clients using models with a single input tensor get tripped up that the .run() input needs to be a list of numpy arrays and not just the single numpy array. If we required creating and passing in an OMTensorList Python object, that would be a lot clearer.

Additionally (ie not required, but I'd love to see this) we should add an API to all the languages that return some sort of API version (each language interface having it's own version). Then client applications (or serving environments), that support "old" and "new" models created by onnx-mlir can query the model and know what APIs are available.

gongsu832: There is no "run()" API in C/C++ and Java. It's a Python thing which unfortunately wasn't following the API theme of other languages. The current Python API is built directly on pybind11 and as a result said before couldn't support string data type. So that forces us to introduce a Python wrapper in front of pybind11. All user facing Python APIs should now be implemented in this Python wrapper rather than directly on pybind11. The Python API should implement some basic routines similar to those in C/C++ and Java. It can also add additional routines like "run()" for convenience and backward compatibility.

gongsu832: > FYI: this is a guideline about creating a .so file to package python scripts: https://fransiska.github.io/2023/09/14/creating-so-files-from-python
> 
> I wonder if it is good to generate a PyRuntime*.so to contain the python wrapper. Then users will need two .so files (PyRuntime*.so and PyRuntimeC*.so) to keep their python program unchanged. Or, it is even better if we can package both the python wrapper and PyRuntimeC*.so in a single .so file, so that there is no change at the user side.

There is no need to do that. A Python package is just a directory that can have both .py and .so files. So for us, we can have a .py wrapper for all user facing APIs and a .so for pybind11 related code.

cjvolzka: Ok, after the onnx-mlir SIG tonight I finally understood that there's also a new PyRuntime.py that acts as a backwards compatible wrapper for the updated PyRuntimeC.cpython-*-linux-gnu.so files. 

For zDLC, at build time, we also pre-build the PyRuntimeC files for all versions of Python that we support and copy them into our deliverable container image. Since I missed that PyRuntime.py exists, it wasn't copied into our container which made me think we had to update our pythonapplications to use new PyRuntimeC directly.

Now that I know it exists, I'll experiment with the PyRuntime.py later this week. Sorry for all the confusion!

PR #2635, opened 2023-11-22
Enable krnl.parallel to have multiple loops

chentong319:
Previously, only one loop ref is allowed by krnl.parallel.
This PR allows a variadic input of loop refs for krnl.prallel. For example, the krnl for an element wise op on three dimension tensor could be (if our parallelization had been modified):
```
    %0:3 = krnl.define_loops 3
    krnl.parallel %0#0, %0#1, %0#2 : !krnl.loop, !krnl.loop, !krnl.loop
    krnl.iterate(%0#0, %0#1, %0#2) with (%0#0 -> %arg2 = 0 to 3, %0#1 -> %arg3 = 0 to 4, %0#2 -> %arg4 = 0 to 5){
      %1:3 = krnl.get_induction_var_value(%0#0, %0#1, %0#2) : (!krnl.loop, !krnl.loop, !krnl.loop) -> (index, index, index)
      %2 = krnl.load %arg0[%1#0, %1#1, %1#2] : memref<3x4x5xf32>
      %3 = krnl.load %arg1[%1#0, %1#1, %1#2] : memref<3x4x5xf32>
      %4 = arith.addf %2, %3 : f32
      krnl.store %4, %alloc[%1#0, %1#1, %1#2] : memref<3x4x5xf32>
    } 
```
In the Krnl to affine, each loop in the krnl.parallel will be converted into a affine.parallel. 
```
    affine.parallel (%arg2) = (0) to (3) {
      affine.parallel (%arg3) = (0) to (4) {
        affine.parallel (%arg4) = (0) to (5) {
          %0 = affine.load %arg0[%arg2, %arg3, %arg4] : memref<3x4x5xf32>
          %1 = affine.load %arg1[%arg2, %arg3, %arg4] : memref<3x4x5xf32>
          %2 = arith.addf %0, %1 : f32
          affine.store %2, %alloc[%arg2, %arg3, %arg4] : memref<3x4x5xf32>
        }
      }
    }
```
and the omp loops will have three nested parallel, which is not we want.

This is not the final result yet. At least two PRs will be needed:
1. Create krnl.parallel with multiple loops
2. Create affine.parallel with multiple loops so that we can have multiple loops for the amp.wsloop. 
```
    affine.parallel (%arg2, %arg3, %arg4) = (0, 0, 0) to (3, 4, 5) {
          %0 = affine.load %arg0[%arg2, %arg3, %arg4] : memref<3x4x5xf32>
          %1 = affine.load %arg1[%arg2, %arg3, %arg4] : memref<3x4x5xf32>
          %2 = arith.addf %0, %1 : f32
          affine.store %2, %alloc[%arg2, %arg3, %arg4] : memref<3x4x5xf32>
    }
```

### DIFF
--- a/src/Dialect/Krnl/DialectBuilder.cpp
+++ b/src/Dialect/Krnl/DialectBuilder.cpp
@@ -126,8 +126,8 @@ ValueRange KrnlBuilder::getInductionVarValue(ValueRange loops) const {
       .getResults();
 }
 
-void KrnlBuilder::parallel(Value loop) const {
-  b().template create<KrnlParallelOp>(loc(), loop);
+void KrnlBuilder::parallel(ValueRange loops) const {
+  b().template create<KrnlParallelOp>(loc(), loops);
 }
 
 void KrnlBuilder::iterate(ValueRange originalLoops, ValueRange optimizedLoops,
--- a/src/Dialect/Krnl/Krnl.td
+++ b/src/Dialect/Krnl/Krnl.td
@@ -513,10 +513,10 @@ def KrnlParallelOp : Op<Krnl_Dialect, "parallel"> {
     ```
   }];
 
-  let arguments = (ins AnyType:$loop);
+  let arguments = (ins Variadic<AnyType>:$loops);
 
   let assemblyFormat = [{
-      $loop attr-dict `:` type($loop)
+      $loops attr-dict `:` type($loops)
   }];
 }
 
--- a/src/Dialect/Krnl/Krnl.td
+++ b/src/Dialect/Krnl/Krnl.td
@@ -509,14 +509,14 @@ def KrnlParallelOp : Op<Krnl_Dialect, "parallel"> {
     operator before krnl.iterate. Since we do not want to parallelize the loop
     until we interpret krnl.block, krnl.permute and krnl.unroll.
     ```
-    krnl.parallel %i
+    krnl.parallel (%i0, %i1)
     ```
   }];
 
   let arguments = (ins Variadic<AnyType>:$loops);
 
   let assemblyFormat = [{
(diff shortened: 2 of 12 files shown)


src/Dialect/Krnl/Krnl.td
```diff
@@ -509,14 +509,14 @@ def KrnlParallelOp : Op<Krnl_Dialect, "parallel"> {
     operator before krnl.iterate. Since we do not want to parallelize the loop
     until we interpret krnl.block, krnl.permute and krnl.unroll.
     ```
-    krnl.parallel %i
+    krnl.parallel (%i0, %i1)
```
AlexandreEichenberger, line 512: Please add a one liner, for example:

```
When multiple loop specifiers are passed as parameters, these loops must be perfectly nested; they may result in a single `OpenMP parallel for`  construct with the `collapsed` loop clause.
```

src/Dialect/Krnl/DialectBuilder.cpp
```diff
@@ -126,8 +126,8 @@ ValueRange KrnlBuilder::getInductionVarValue(ValueRange loops) const {
       .getResults();
 }
 
-void KrnlBuilder::parallel(Value loop) const {
-  b().template create<KrnlParallelOp>(loc(), loop);
+void KrnlBuilder::parallel(ValueRange loops) const {
```
AlexandreEichenberger, line 129: If it does not generate conflicts, you may want to consider a constructor that also take a single Value.

chentong319: ValueRange will be automatically constructed when a single Value is passed. You may have noticed that the current code to call this builder function remains the same.

PR #2636, opened 2023-11-22
make the OMP parallel runtime included with the `-parallel` flag

AlexandreEichenberger:
Using the '-parallel` flag, the OpenMP runtime is linked in the model so that no matter how the model is deployed, the OpenMP runtime will work.

Also moved the bufferization before OMP expansion as it seems to fail otherwise at this time. This may be reverted in the future.

### DIFF
--- a/src/Runtime/omp/CMakeLists.txt
+++ b/src/Runtime/omp/CMakeLists.txt
@@ -27,7 +27,7 @@ if(TARGET omp)
 
   add_custom_target(libOMomp
     COMMAND ${CMAKE_COMMAND} -E copy_if_different
-            ${OMP_TOPDIR}/lib/libomp.a ${ONNX_MLIR_LIBRARY_PATH}/libOMomp.a
+            ${OMP_TOPDIR}/lib/libomp.a ${ONNX_MLIR_LIBRARY_PATH}/libompruntime.a
     DEPENDS OMomp
     # BYPRODUCTS requires cmake 3.20+
     BYPRODUCTS ${OMP_TOPDIR}/lib/libomp.a
--- a/src/Compiler/CompilerOptions.cpp
+++ b/src/Compiler/CompilerOptions.cpp
@@ -1073,6 +1073,10 @@ void initCompilerConfig() {
             : std::vector<std::string>{"jniruntime", "cruntime"});
     addCompilerConfig(CCM_SHARED_LIB_PATH_DEPS, {getLibraryPath()});
 
+    // Add OpenMP LLVM library if parallel is enabled.
+    if (enableParallel)
+      addCompilerConfig(CCM_SHARED_LIB_DEPS, {"ompruntime"});
+
     // Add user specified libs and their path
     // Multiple lib or directory can be specified with multiple options.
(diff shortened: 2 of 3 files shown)


src/Runtime/omp/CMakeLists.txt
```diff
@@ -27,7 +27,7 @@ if(TARGET omp)
 
   add_custom_target(libOMomp
```
gongsu832, line 28: For consistency, this `libOMomp` and the one in `src/Compiler/CMakeLists.txt` should also be changed to `ompruntime`.

chentong319:
Is the libomp.a is copied into our lib directory as libompruntime.a to avoid the library path issue?

gongsu832:
> Is the libomp.a is copied into our lib directory as libompruntime.a to avoid the library path issue?

That's one of the reason. Others being,

- LLVM can build either `libomp.so` or `libomp.a` depending on whether `BUILD_SHARED_LIBS` is set to ON or OFF. But we must use `libomp.a` (and its objs must be compiled with -fPIC). So we cannot just simply copy over what LLVM has built. We must build `libomp.a` ourselves according to our requirement.
- The `omp` name is already taken by LLVM so we need to name ours something else, i.e., `ompruntime`. Since when people link with `-lomp`, they typically expect to link against the LLVM `libomp.so` or `libomp.a`, not accidentally link against our `libomp.a`.

chentong319 (approved):
The change of CompilePass.cpp LGTM!
I will rely on Gong for the change of cmakefile.
One comment:  some users of onnx-mlir, are building onnx-mlir natively with their own script. They may not use OpenMP. Do we have enough info for such users? Can they easily choose not include OpenMP in their onnx-mlir build?

AlexandreEichenberger:
> One comment: some users of onnx-mlir, are building onnx-mlir natively with their own script. They may not use OpenMP. Do we have enough info for such users? Can they easily choose not include OpenMP in their onnx-mlir build?

Good question, @gongsu832 what triggers the building of the lib?

```
if(TARGET omp)
```

is it triggered if we build LLVM with the OpenMP option?
```
   -DLLVM_ENABLE_PROJECTS="mlir;clang;openmp" 
```

or is it an explicit target when we build onnx-mlir?

gongsu832:
Yes it's triggered when LLVM is built with `openmp`. No config necessary on the onnx-mlir side. Of course, this only means that the `libompruntime.a` will be built. Whether it will be embedded into the model runtime depends on the `-parallel` flag.

gongsu832:

gongsu832:

AlexandreEichenberger:
At this time, the `--parallel` flag is still experimental and there are known bug around matrix multiply/gemm and a few others that require private buffers. 

At this time, we don't set a max number of threads, but our implementation is built on OpenMP. As such,

```
export OMP_THREAD_LIMIT 16
export OMP_NUM_THREADS 8
```

would be an example of setting that limit the total number of threads to 16, with at most 8 given per parallel region (in case nested parallel region are used, then each parallel thread could in turn get 1 additional thread for a nested parallel region)

PR #2640, opened 2023-11-24
A rewrite pattern to optimize constant scaling in self-attention layer

tungld:
In the self-attention layer, the output of MatMul is scaled by a constant factor via a division/multiplication operation. This patch rewrites the division/multiplication operation so that the constant input of MatMul will be scaled instead of its output. Thus, the scaling of the constant inputs can be folded at compile time. 

For example, this patch rewrites the following pattern:
```
shape_transform(X1 * A1 + B1) * shape_transform(X2 * A2 + B2) / k
```

into
```
shape_transform(X1 * A1 + B1) * shape_transform(X2 * A2/k + B2/k)
```
if A2, B2 and k are constants,

or into
```
shape_transform(X1 * A1/k + B1/k) * shape_transform(X2 * A2 + B2)
```
if A1, B1 and k are constants,

where
- `*` is matrix multiplication; `+` and `/` are element-wise addition and division
- A1, A2, B1, B2, and k are constants so that A1/k, B1/k, A2/k and B2/k can
be folded. k is a scalar constant so that it's broadcastable to all A1, A2,
B1, B2.
- shape_transform includes a sequence of operations that change the data
shape of the input but not numerical values, for example: Reshape,
Transpose, etc.

### DIFF
--- a/src/Dialect/ONNX/Rewrite.cpp
+++ b/src/Dialect/ONNX/Rewrite.cpp
@@ -218,34 +218,49 @@ bool matchShapeAddMatMul(Value v, Value &matA, Value &biasB,
     Operation *&matmulOp, Operation *&addOp) {
   if (v.isa<BlockArgument>())
     return false;
+  if (!hasOneUseExceptDimOp(v))
+    return false;
   Value origV = v;
-  // Match shape operations.
-  while (isa<ONNXReshapeOp, ONNXTransposeOp, ONNXSqueezeOp, ONNXUnsqueezeOp>(
-      origV.getDefiningOp())) {
-    origV = origV.getDefiningOp()->getOperands()[0];
-    if (origV.isa<BlockArgument>())
-      return false;
+  // Match a sequence of shape operations. Each shape operation has only one
+  // use.
+  while (auto defOp = origV.getDefiningOp()) {
+    if (!isa<ONNXReshapeOp, ONNXTransposeOp, ONNXSqueezeOp, ONNXUnsqueezeOp>(
+            defOp))
+      break;
+    origV = defOp->getOperands()[0];
+    if (!hasOneUseExceptDimOp(origV))
+      break;
   }
+  if (origV.isa<BlockArgument>() || !hasOneUseExceptDimOp(origV))
+    return false;
 
   // Match Add.
-  addOp = origV.getDefiningOp();
-  if (!addOp || !isa<ONNXAddOp>(addOp))
+  auto onnxAddOp = origV.getDefiningOp<ONNXAddOp>();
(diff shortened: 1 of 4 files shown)


tehbone:
Can this be abstracted to any pair wise operation? There could be some attention models that have a multiplication by an inverse sqrt.

tungld:
> Can this be abstracted to any pair wise operation? There could be some attention models that have a multiplication by an inverse sqrt.

Good suggestion! Yes, it is quite straightforward to support multiplication also (I don’t think it is applicable to addition and subtraction). Will add that soon. Thanks!

src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -395,6 +450,104 @@ class PropagateReshapeThroughBinaryOpPattern
...
+    if (!isScalarConstantTensor(K))
+      return rewriter.notifyMatchFailure(
+          genericOp, "The second operand of Div/Mul is not a scalar constant");
+
+    // Match lhs = shape_transform(X1*A1 + B1)
+    Value A1, B1;
+    Operation *lhsSubMatOp, *lhsAddOp;
+    bool matchLHS = matchShapeAddMatMul(lhs, A1, B1, lhsSubMatOp, lhsAddOp);
+
+    // Match rhs = shape_transform(X2*A2 + B2)
```
AlexandreEichenberger, line 510: Nit, you could test match RHS only when matchLHS fails. Less testing, same results
```
bool matchRHS = !matchRHS && matchShapeAddMatMul(rhs, A2, B2, rhsSubMatOp, rhsAddOp);
```

Then you can get rid of the case where both matches, as this will never be the case

src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -209,6 +209,61 @@ bool haveSameStaticShape(Value lhs, Value rhs) {
   return hasStaticShape(lhsT) && (getShape(lhsT) == getShape(rhsT));
 }
 
+// Match v = shape_transform(X*A + B).
```
AlexandreEichenberger, line 212: Do we cover the case where instead of `X*A+B` we have a `Gemm` op?

AlexandreEichenberger (approved):
LGTM

gongsu832:

src/Dialect/ONNX/Rewrite.cpp
tungld: Good idea. I will add Gemm.

tungld: I added the case for Gemm.

PR #2644, opened 2023-11-27
Add options to drive CheckONNXModel.py

AlexandreEichenberger:
Often one runs a test by adding a flag to the original set of flags, e.g. to test with / without parallel. This PR adds a `-a` flags to that new options are added to the reference flags, e.g.

```
CheckONNXModel.py -r="-O3 -march=arm64" -a="-parallel" -m gemm.mlir  
```

only one of `-t` or `-a` option can be used.

Also add bounds for inputs, as some tests need postitive only values...


tungld (approved):
LGTM!

PR #2646, opened 2023-11-28
ONNX-to-TOSA: Add ResizeOp Conversion

maxbartel:
This introduces the conversion of the `ResizeOp` from ONNX to TOSA.
The different modes of the op would all probably deserve their own operation instead of just an `Attribute`, so be extra careful when reviewing please. Also TOSA has the concept of using a fraction instead a float for the scale. You can look it up here: https://www.mlplatform.org/tosa/tosa_spec.html#_resize. I tried to give as much information as possible on how I arrived at the different formulas. Please let me know if something is not clear.

We also have to disallow a lot of cases because the onnx op is more powerful than the TOSA one.

### DIFF
--- a/src/Conversion/ONNXToTOSA/ONNXToTOSALegalizeUtils.cpp
+++ b/src/Conversion/ONNXToTOSA/ONNXToTOSALegalizeUtils.cpp
@@ -35,6 +35,17 @@ using namespace mlir;
 namespace onnx_mlir {
 namespace tosa {
 
+int64_t convertNegativeAxis(int64_t axis, int64_t inputRank) {
+  if (axis < 0)
+    axis += inputRank;
+
+  // Check if axis is in correct range.
+  assert(
+      (axis >= 0 && axis < inputRank) && "axis attribute not in correct range");
+
+  return axis;
+}
+
 mlir::RankedTensorType reduceAxisToOne(llvm::ArrayRef<int64_t> shape,
     mlir::Type elementType, mlir::Attribute encoding) {
   return mlir::RankedTensorType::get(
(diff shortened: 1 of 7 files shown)


src/Conversion/ONNXToTOSA/Tensor/Resize.cpp
```diff
@@ -0,0 +1,332 @@
...
+    int64_t numerator;
+    int64_t denominator;
+  };
+
+  /// ## coordinateTransformationMode ##
+  /// TOSA uses the formula ix = (ox * scale_x_d + offset_x) / scale_x_n
+  /// to find the input coordinates. In order to lower ONNX one needs to
+  /// express the modes in this context. Border is only used to check if certain
+  /// conditions in the dimensions are met, but has no actual use in calculating
+  /// something. It is probably an error in the TOSA specification. In order to
```
eric-k256, line 136: Border is used to ensure that the shape calculation is exactly defined with an integer ratio.

eric-k256 (approved):
The TOSA op is definitely more focused. It appears that you've covered the many unsupported cases and have something that looks reasonable.

src/Conversion/ONNXToTOSA/ONNXToTOSALegalizeUtils.cpp
```diff
@@ -35,6 +35,17 @@ using namespace mlir;
...
+  if (axis < 0)
+    axis += inputRank;
+
+  // Check if axis is in correct range.
+  assert(
+      (axis >= 0 && axis < inputRank) && "axis attribute not in correct range");
+
+  return axis;
+}
+
```
AlexandreEichenberger, line 48: Its ok to add that code, but note that there is already a version of this in `ShapeHelper.hpp`
```
// Check if axis is in [-rank, rank), or [-rank, rank] when includeRank is
// true.  Assert when not in range. Return positive axis.
int64_t getAxisInRange(int64_t axis, int64_t rank, bool includeRank = false);
int64_t getAxisInRange(int64_t axis, mlir::Value val, bool includeRank = false);
```

AlexandreEichenberger (approved):
LGTM

PR #2651, opened 2023-12-01
[NNPA] Disable AddConstAssociative2 and AddConstAssociative3 rules since they are not good for NNPA

tungld:
Disable two following constprop rules when compiling a model for NNPA:
- `add(add(x, c), y)` to `add(add(x, y), c)`
- `add(x, add(y, c))` to `add(add(x, y), c)`

because in foundation models we have `add(add(matmul(x,z), c), z), y)`, and we want to keep c near matmul so that `add(matmul(x,z), c)` will be run on NNPA as one call.


AlexandreEichenberger:
> because in foundation models we have add(add(matmul(x,z), c), z), y), and we want to keep c near matmul so that add(matmul(x,z), c) will be run on NNPA as one call.

@tung, is that because of the sizes of `c` and `z`? Maybe we should have an explicit rule that chante the `matmul` and `add` into a `gemm` if suitable shapes are found for `zAIU` target, that would be a more explicit solution.

tungld:
> @tung, is that because of the sizes of c and z? Maybe we should have an explicit rule that chante the matmul and add into a gemm if suitable shapes are found for zAIU target, that would be a more explicit solution.

Yes, we have a rule in Rewrite.td to change matmul+add into a gemm. But in this case, x of matmul is 3D while Gemm only support 2D. That's why Gemm is not generated.

AlexandreEichenberger (approved):
LGTM

PR #2653, opened 2023-12-01
[RunONNXModel.py] Do not use tag if loading a compiled .so model

tungld:
When using `RunONNXModel.py`, if users load a compiled model by `--load-so=model.so`, we initialize a ExecutionSession with "none" tag because users may change the filename of the `.so` file.


AlexandreEichenberger (approved):
LGTM

PR #2655, opened 2023-12-01
ONNX 1.15.0 update and misc small fixes

gongsu832:
- update ONNX to 1.15.0
- ubi8 can now install ninja-build with yum so remove manual installation
- fix target check-onnx-backend-case for generating all_test_names.txt
- make targets OMONNXOpsIncTranslation and onnx-mlir-docs to generate relevant files


chentong319:
Is the change of document caused by the docgen tool we are using?
By the way, can you change line 64 of MLIR.cmake to avoid generating the documents directly into the original directory? An undesired side effect of the current cmake configuration is that the cmake clean will remove the document files (*.md) in the original directory. Not a necessary change if too much change.

chentong319:
1. For `make OMONNXOpsIncTranslation`.  I tried with the main branch. The two targets are generated and copied to their desired location. However, a copy of file OpBuildTable.inc is left in utils. I did not figure out why. The CMAKE commands for ONNXOps.td.inc and OpBuildTable.inc are the same.
2. One necessary change: the allowed version in utils/gen_onnx_mlir.py should be 1.15.0 when the third_party/onnx is upgraded.

gongsu832:
> 1. For `make OMONNXOpsIncTranslation`.  I tried with the main branch. The two targets are generated and copied to their desired location. However, a copy of file OpBuildTable.inc is left in utils. I did not figure out why. The CMAKE commands for ONNXOps.td.inc and OpBuildTable.inc are the same.
> 
As mentioned above, I'm not familiar with how these `.inc` and `.md` files are handled so I leave it to someone who is more familiar with the matter.

>     2. One necessary change: the allowed version in utils/gen_onnx_mlir.py should be 1.15.0 when the third_party/onnx is upgraded.
The file is already updated in the PR.

chentong319 (approved):
LGTM!

PR #2656, opened 2023-12-05
[NNPA] Fix the size of the reserved field in ztensor

tungld:
The `reserved` field in ztensor has size of 31 instead of 32. This patch fixes that issue. Though the `reserved` field is not used now and it does not affect the runtime, it's better to use the correct value.


AlexandreEichenberger (approved):
LGTM

PR #2654, opened 2023-12-01
Bump LLVM to 5e5a22caf88ac1ccfa8dc5720295fdeba0ad9372 

flemairen6:
Bump LLVM to 5e5a22caf88ac1ccfa8dc5720295fdeba0ad9372 and stableHlo to 83f095e7217c897f1eccac5652600ceb944cb0e0 to align with https://github.com/llvm/torch-mlir/pull/2598 
The typed pointers seems to have been deprecated for good, so I removed stuff that was related to them. I might have missed some.
I also see some tests from `krnl_to_llvm` failing as if they could not lower some operators. I have really little experience with this area of the project, it would be great if someone could have a look at the issue with me.

PR #2658, opened 2023-12-07
Bump LLVM to 5e5a22caf88ac1ccfa8dc5720295fdeba0ad9372

tungld:
Combined PR #2654 and #2657 and fixed the remaining lit tests.


tungld:
@gongsu832 thanks for looking at this. It seems there is a patch to fix this in LLVM (just merged 20 hours ago): https://github.com/llvm/llvm-project/pull/74002. Do you think it can solve the warning here?

tungld:
So should we process with disabling the warning as a temporary solution? and reenable it once we bump to a newer LLVM. I think the warning is about SparseTensor that we haven't used in onnx-mlir.

gongsu832:
As long as you remember to reenable it once we update past the commit including the fix. :-)

AlexandreEichenberger:
So do we need to patch LLVM?

tungld:
> So do we need to patch LLVM?

I think No, we just suppress it so that Windows will not treat it as an error.

tungld:
@gongsu832 the place I inserted code to suppress C4293 did not work. Do you have any suggestion where should I put the code to suppress a warning?

gongsu832:
Unfortunately, no. I don't know much about developing on Windows. We probably need to ask the Microsoft folks.

PR #2659, opened 2023-12-08
Make ompruntime use the same CC/CXX compilers that onnx-mlir is configured with

gongsu832:
ompruntime should use the same CC/CXX compiler that was used to compile onnx-mlir instead of being hardcoded to clang/clang++.


AlexandreEichenberger (approved):
Make sense, thanks
LGTM

PR #2660, opened 2023-12-11
Modelzoo restructure breaks verifying user docker image

gongsu832:
Use docs/mnist_example/mnist.onnx for verifying user docker image so we don't depend on modelzoo, which has just been completely restructured.

This is a blocking issue for Jenkins CI so we should merge asap.


gongsu832:

hamptonm1 (approved):
I am blocked until this PR is merged in, so approving since it is passing all checks.

PR #2665, opened 2023-12-22
Allow onnx.unique with no data

negiyas:
Current onnx.Unique op implementation causes the following runtime error with a Tensor with no data.
This case should be handled as a normal case without causing errors.

```
.../src/Runtime/OMUnique.inc:271: produceY: Assertion `count > 0 && "count should be greater than 0"' failed.
```


tungld:
Does the input tensor have no data? If so, we can check it at compile time and just return an empty tensor without any computation when lowering Unique.

PR #2668, opened 2024-01-03
Implementation for Gelu Operation 

hamptonm1:
This code essentially adds support for `onnx.gelu `using the approximate attribute value.

When `approximate = none` the following output is expected:
`y = 0.5 * x * (1 + erf(x/sqrt(2)))`

When`approximate = tanh` the following output is expected:
`y = 0.5 * x * (1 + Tanh(sqrt(2/pi) * (x + 0.044715 * x^3)))`

**What this PR covers:**

- Added code to elementwise to ensure the calculations are done correctly for the output
- Added code for SIMD of onnx.gelu 
- I also added a verifier to onnx.gelu to make sure that only the string value of `tanh` or `none` are accepted values for `approximate`
- I created two lit tests
- I added the eight backend tests associated with gelu from onnx 1.15.0 but only 4/8 tests pass. I am not too familiar with what the expanded variation of this tests means so will comment out until I better understand how to debug. 
- Updated documentation and did some small housekeeping


src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -331,6 +334,74 @@ struct ScalarOp<ONNXTanOp> {
...
+  using FOp = CustomScalarOp;
+  using IOp = CustomScalarOp;
+};
+
+template <>
+double analyzeSimdFor<ONNXGeluOp>(
+    Type t, Operation *op, int64_t &von, int64_t &son) {
+  double results;
+  StringRef approximate = dyn_cast<ONNXGeluOp>(op).getApproximate();
+  if (approximate == "none")
```
hamptonm1, line 351: @AlexandreEichenberger Let me know what you think of the SIMD for Gelu... I think I added every operation but it is always good to have another set of eyes. Let me know if there is anything I am missing. Thanks :)

src/Dialect/ONNX/ONNXOps.hpp
```diff
@@ -25,7 +25,7 @@
 
 namespace mlir {
 // OpSet level supported by onnx-mlir
-static constexpr int CURRENT_ONNX_OPSET = 17;
```
hamptonm1, line 28: @AlexandreEichenberger Are we okay with increasing this value or should we keep it at 19?

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -331,6 +334,74 @@ struct ScalarOp<ONNXTanOp> {
...
+  Value div = create.math.div(operand, sqrt_two);
+  Value erfApprox = create.math.erf(div);
+  Value add = create.math.add(one, erfApprox);
+  Value dec = create.math.add(
+      operand, create.math.mul(decimal, create.math.pow(operand, three)));
+  Value tanhApprox = create.math.tanh(create.math.mul(sqrt_two_over_pi, dec));
+  // Approximate = none returns an output of y = 0.5 * x * (1 +
+  // erf(x/sqrt(2)))
+  if (approximate == "none")
+    return create.math.mul(create.math.mul(half, operand), add);
```
tungld, line 396: To avoid emitting redundant code, please move local variables for the case `approximate == "none"` into the body of this `if` since they are not used in the other `if (approximate == "tanh")`, e.g. the following local variables.
```
  Value sqrt_two = create.math.constant(elementType, sqrt(2));
  Value div = create.math.div(operand, sqrt_two);
  Value erfApprox = create.math.erf(div);
  Value add = create.math.add(one, erfApprox);
```

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -331,6 +334,74 @@ struct ScalarOp<ONNXTanOp> {
...
+  // Approximate = tanh returns an output of y = 0.5 * x * (1 + Tanh(sqrt(2/pi)
+  // * (x + 0.044715 * x^3)))
+  if (approximate == "tanh")
+    return create.math.mul(
```
tungld, line 400: Same here, please move local variables `tanhApprox`, `dec`,  ... into the body of this `if`:
```
  Value three = create.math.constant(elementType, 3);
  Value decimal = create.math.constant(elementType, 0.044715);
  Value sqrt_two_over_pi = create.math.constant(elementType, sqrt(2 / M_PI));
  Value dec = create.math.add(operand, create.math.mul(decimal, create.math.pow(operand, three)));
  Value tanhApprox = create.math.tanh(create.math.mul(sqrt_two_over_pi, dec));
```

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -331,6 +334,74 @@ struct ScalarOp<ONNXTanOp> {
...
+template <>
+double analyzeSimdFor<ONNXGeluOp>(
+    Type t, Operation *op, int64_t &von, int64_t &son) {
+  double results;
+  StringRef approximate = dyn_cast<ONNXGeluOp>(op).getApproximate();
+  if (approximate == "none")
+    results = simdAnalysis({GenericOps::ArithmeticGop, GenericOps::DivGop,
+                               GenericOps::MulGop, GenericOps::SqrtGop},
+        {1, 1, 2, 1}, t, von, son);
+  if (approximate == "tanh")
```
tungld, line 355: Use `approximate.equals_insensitive("tanh")` to cover different cases. Please make the change at other locations as well.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -331,6 +334,74 @@ struct ScalarOp<ONNXTanOp> {
...
+  // indicates the use of tanh approximation.
+  StringRef approximate = dyn_cast<ONNXGeluOp>(op).getApproximate();
+
+  // Create constants
+  Value half = create.math.constant(elementType, 0.5);
+  Value one = create.math.constant(elementType, 1);
+  Value three = create.math.constant(elementType, 3);
+  Value decimal = create.math.constant(elementType, 0.044715);
+  Value sqrt_two_over_pi = create.math.constant(elementType, sqrt(2 / M_PI));
+  Value sqrt_two = create.math.constant(elementType, sqrt(2));
```
tungld, line 384: Please use camel case for variable names, i.e. `sqrtTwoOverPi`, `sqrtTwo`, etc.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -331,6 +334,74 @@ struct ScalarOp<ONNXTanOp> {
...
+  Value tanhApprox = create.math.tanh(create.math.mul(sqrt_two_over_pi, dec));
+  // Approximate = none returns an output of y = 0.5 * x * (1 +
+  // erf(x/sqrt(2)))
+  if (approximate == "none")
+    return create.math.mul(create.math.mul(half, operand), add);
+  // Approximate = tanh returns an output of y = 0.5 * x * (1 + Tanh(sqrt(2/pi)
+  // * (x + 0.044715 * x^3)))
+  if (approximate == "tanh")
+    return create.math.mul(
+        create.math.mul(half, operand), create.math.add(one, tanhApprox));
```
tungld, line 401: I would create a global variable for `create.math.mul(half, operand)` since it's used in both cases.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
hamptonm1: Okay fixed!

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
hamptonm1: Hmm but according to the documentation it should only be `tanh` and `none`. If onnx does not support different cases, I do not think we should either or else there will be an error.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
hamptonm1: Okay when I move the Value's into the `if` I actually get the following errors:

```
[10/14] Building CXX object src/Conversion/ONNXToKrnl/CMakeFiles/OMONNXToKrnl.dir/Math/Elementwise.cpp.o
FAILED: src/Conversion/ONNXToKrnl/CMakeFiles/OMONNXToKrnl.dir/Math/Elementwise.cpp.o 
/usr/bin/c++ -DONNX_ML=1 -DONNX_MLIR_DECOMP_ONNX_CONVTRANSPOSE -DONNX_MLIR_ENABLE_STABLEHLO -DONNX_NAMESPACE=onnx -DONNX_USE_LITE_PROTO=1 -D_DEBUG -D_GLIBCXX_ASSERTIONS -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -I/Users/meganhampton/zDLC/llvm-project/llvm/include -I/Users/meganhampton/zDLC/llvm-project/build/include -I/Users/meganhampton/zDLC/llvm-project/mlir/include -I/Users/meganhampton/zDLC/llvm-project/build/tools/mlir/include -I/Users/meganhampton/zDLC/onnx-mlir -I/Users/meganhampton/zDLC/onnx-mlir/build -I/Users/meganhampton/zDLC/onnx-mlir/include -I/Users/meganhampton/zDLC/onnx-mlir/third_party/onnx -I/Users/meganhampton/zDLC/onnx-mlir/build/third_party/onnx -isystem /opt/homebrew/include -isystem /opt/homebrew/Cellar/protobuf@21/21.12/include -fPIC -fvisibility-inlines-hidden -Werror=date-time -Werror=unguarded-availability-new -Wall -Wextra -Wno-unused-parameter -Wwrite-strings -Wcast-qual -Wmissing-field-initializers -Wimplicit-fallthrough -Wcovered-switch-default -Wno-noexcept-type -Wnon-virtual-dtor -Wdelete-non-virtual-dtor -Wsuggest-override -Wstring-conversion -Wmisleading-indentation -Wctad-maybe-unsupported -fdiagnostics-color -DSUPPRESS_THIRD_PARTY_WARNINGS -g -std=gnu++17 -arch arm64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX14.2.sdk   -D_DEBUG -D_GLIBCXX_ASSERTIONS -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -MD -MT src/Conversion/ONNXToKrnl/CMakeFiles/OMONNXToKrnl.dir/Math/Elementwise.cpp.o -MF src/Conversion/ONNXToKrnl/CMakeFiles/OMONNXToKrnl.dir/Math/Elementwise.cpp.o.d -o src/Conversion/ONNXToKrnl/CMakeFiles/OMONNXToKrnl.dir/Math/Elementwise.cpp.o -c /Users/meganhampton/zDLC/onnx-mlir/src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
/Users/meganhampton/zDLC/onnx-mlir/src/Conversion/ONNXToKrnl/Math/Elementwise.cpp:390:5: warning: misleading indentation; statement is not part of the previous 'if' [-Wmisleading-indentation]
    Value erfApprox = create.math.erf(div);
    ^
/Users/meganhampton/zDLC/onnx-mlir/src/Conversion/ONNXToKrnl/Math/Elementwise.cpp:388:3: note: previous statement is here
  if (approximate == "none")
  ^
/Users/meganhampton/zDLC/onnx-mlir/src/Conversion/ONNXToKrnl/Math/Elementwise.cpp:389:11: warning: unused variable 'div' [-Wunused-variable]
    Value div = create.math.div(operand, sqrtTwo);
          ^
/Users/meganhampton/zDLC/onnx-mlir/src/Conversion/ONNXToKrnl/Math/Elementwise.cpp:390:39: error: no viable conversion from '<overloaded function type>' to 'mlir::Value'
    Value erfApprox = create.math.erf(div);
                                      ^~~
/Users/meganhampton/zDLC/llvm-project/mlir/include/mlir/IR/Value.h:96:7: note: candidate constructor (the implicit copy constructor) not viable: no overload of 'div' matching 'const mlir::Value &' for 1st argument
class Value {
      ^
/Users/meganhampton/zDLC/llvm-project/mlir/include/mlir/IR/Value.h:96:7: note: candidate constructor (the implicit move constructor) not viable: no overload of 'div' matching 'mlir::Value &&' for 1st argument
/Users/meganhampton/zDLC/llvm-project/mlir/include/mlir/IR/Value.h:98:13: note: candidate constructor not viable: no overload of 'div' matching 'detail::ValueImpl *' for 1st argument
  constexpr Value(detail::ValueImpl *impl = nullptr) : impl(impl) {}
            ^
/Users/meganhampton/zDLC/onnx-mlir/src/Dialect/Mlir/DialectBuilder.hpp:116:31: note: passing argument to parameter 'val' here
  mlir::Value erf(mlir::Value val) const;
                              ^
/Users/meganhampton/zDLC/onnx-mlir/src/Conversion/ONNXToKrnl/Math/Elementwise.cpp:397:5: warning: misleading indentation; statement is not part of the previous 'if' [-Wmisleading-indentation]
    Value tanhApprox = create.math.tanh(create.math.mul(sqrtTwoOverPi, dec));
    ^
/Users/meganhampton/zDLC/onnx-mlir/src/Conversion/ONNXToKrnl/Math/Elementwise.cpp:395:3: note: previous statement is here
  if (approximate == "tanh")
  ^
/Users/meganhampton/zDLC/onnx-mlir/src/Conversion/ONNXToKrnl/Math/Elementwise.cpp:396:11: warning: unused variable 'dec' [-Wunused-variable]
    Value dec = create.math.add(operand, create.math.mul(decimal, create.math.pow(operand, three)));
          ^
/Users/meganhampton/zDLC/onnx-mlir/src/Conversion/ONNXToKrnl/Math/Elementwise.cpp:397:72: error: use of undeclared identifier 'dec'
    Value tanhApprox = create.math.tanh(create.math.mul(sqrtTwoOverPi, dec));
                                                                       ^
4 warnings and 2 errors generated.
```

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
hamptonm1: None of these values are redundant.... I actually only use each one once. If I move it into the body I get the error as shown below :)

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
hamptonm1: kk done

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
hamptonm1: But if we see reasons to have different variations... I am open to change. Thanks for reviewing!

AlexandreEichenberger: I am fine either way, I see that the GELU verifier will bomb if the string is not `none` or `approximate`. What is done for the other strings? would be good to have a consensus.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
AlexandreEichenberger: @hamptonm1, the code as is is a bit difficult to read. What @tungld is requesting is that the values that are exclusively computed for the first case be only included in the first case, and same for the second case. 

I have a hard time verifying the cost model because not all the computation before the if are used in both cases. 

If too difficult to separate, then I would suggest to simply have all of the code in the then, and then all of the code in the else, and removing values that are unused.

src/Dialect/Mlir/DialectBuilder.cpp
```diff
@@ -218,6 +218,10 @@ Value MathBuilder::fma(Value lhs, Value rhs, Value acc) const {
   return add(mul(lhs, rhs), acc);
 }
 
+Value MathBuilder::erf(Value val) const {
```
AlexandreEichenberger, line 221: Thanks for adding, much appreciated.

src/Dialect/ONNX/ONNXOps.hpp
AlexandreEichenberger: I think that is the correct level, if one or more ops from opset 20 are supported. You can clarify this in the comments so that future developers better understand the intent.

src/Dialect/ONNX/ONNXOps/Math/ElementwiseUnary.cpp
```diff
@@ -257,6 +257,28 @@ LogicalResult ONNXFloorOp::inferShapes(
   return inferShapeForUnaryOps(this->getOperation());
 }
 
+//===----------------------------------------------------------------------===//
+// Gelu
+//===----------------------------------------------------------------------===//
+LogicalResult ONNXGeluOp::verify() {
```
AlexandreEichenberger, line 263: Thanks for the verifier, very complete implementation, appreciated.

test/mlir/conversion/onnx_to_krnl/Math/Elementwise.mlir
```diff
@@ -901,6 +901,99 @@ func.func @test_isnan(%arg0 : tensor<2x3x4xf32>) -> tensor<*xi1> {
 
 // -----
 
+func.func @test_gelu_none(%arg0 : tensor<2x3x4xf32>) -> tensor<2x3x4xf32> {
```
AlexandreEichenberger, line 904: If you don't mind, I would suggest to add this test to the related `Elementwise_with_canonicalize.mlir` file, that way it's less sensitive to the redundant operations that are generated, e.g. there are multiple instances of the constant `2`... all that is cleaned up with the additional canonicalized pass.

AlexandreEichenberger (commented):
Very complete implementations, just a few minor details to ironed out. Solid, thanks

src/Dialect/ONNX/ONNXOps/Math/ElementwiseUnary.cpp
hamptonm1: You are welcome!!

test/mlir/conversion/onnx_to_krnl/Math/Elementwise.mlir
hamptonm1: Okay sure! So remove this test from Elementwise.mlir and place it in Elementwise_with_canonicalize.mlir? Or have these two tests in both locations? Just making sure I am following here.

src/Dialect/Mlir/DialectBuilder.cpp
hamptonm1: yw!

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
hamptonm1: You are right... whatever is done throughout the code basis is what should be followed. I see it done in other places, so I do not mind adding it but yes I made sure that it will fail if it is not what onnx is expecting :)

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
hamptonm1: Ohh this is not to be dogmatic, I was saying I received an error when attempting to do things the way @Tung suggested. I will try to find a better solution... I am open at this point. I understand his request and agree but I was stating the values are not redundant just in case. Thanks!

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
hamptonm1: Okay done!

test/mlir/conversion/onnx_to_krnl/Math/Elementwise.mlir
hamptonm1: Okay fixed!

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
hamptonm1: Okay I figured out why it was failing.... it needed curly braces around the if statement or else we will get those errors. Everything is fixed now.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
hamptonm1: Fixed

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -331,6 +334,77 @@ struct ScalarOp<ONNXTanOp> {
...
+};
+
+template <>
+double analyzeSimdFor<ONNXGeluOp>(
+    Type t, Operation *op, int64_t &von, int64_t &son) {
+  double results;
+  StringRef approximate = dyn_cast<ONNXGeluOp>(op).getApproximate();
+  if (approximate.equals_insensitive("none"))
+    results = simdAnalysis(
+        {GenericOps::ArithmeticGop, GenericOps::MulGop}, {1, 3}, t, von, son);
```
AlexandreEichenberger, line 353: Please add `GenericOps:: ErfGop` with 1

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -331,6 +334,77 @@ struct ScalarOp<ONNXTanOp> {
...
+  if (approximate.equals_insensitive("tanh"))
+    results = simdAnalysis(
+        {GenericOps::ArithmeticGop, GenericOps::MulGop, GenericOps::PowGop},
+        {2, 4, 1}, t, von, son);
+  return results;
```
AlexandreEichenberger, line 358: Since pow(3) is really 2 matmul, I would remove the Pow and add 1 to MulGop. Note that we also have `TrigHyperbolicGop` so it can be added too.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -331,6 +334,77 @@ struct ScalarOp<ONNXTanOp> {
...
+  // indicates the use of tanh approximation.
+  StringRef approximate = dyn_cast<ONNXGeluOp>(op).getApproximate();
+
+  // Local constants
+  Value half = create.math.constant(elementType, 0.5);
+  Value one = create.math.constant(elementType, 1);
+  Value halfTimesOperand = create.math.mul(half, operand);
+
+  // Approximate = none returns an output of y = 0.5 * x * (1 +
+  // erf(x/sqrt(2)))
```
AlexandreEichenberger, line 382: Awesome, super clear now.

test/mlir/conversion/onnx_to_krnl/Math/Elementwise_with_canonicalize.mlir
```diff
@@ -83,6 +83,67 @@ func.func @test_variadic_elementwise_op_template_unknown_dims(%arg0: tensor<?x4x
 
 // -----
 
+func.func @test_gelu_none(%arg0 : tensor<2x3x4xf32>) -> tensor<2x3x4xf32> {
+  %0 = "onnx.Gelu"(%arg0) {approximate = "none"} : (tensor<2x3x4xf32>) -> tensor<2x3x4xf32>
+  return %0 : tensor<2x3x4xf32>
+
+// mlir2FileCheck.py
+// CHECK-LABEL:  func.func @test_gelu_none
```
AlexandreEichenberger, line 91: Tx, example is so much more concise now.

AlexandreEichenberger (approved):
LGTM, please update the approx for SIMD, which is a very minor edit.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
hamptonm1: Thanks! I did not realize we had a `GenericOps:: ErfGop`. Let me fix that now

hamptonm1: Done!

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
hamptonm1: Done!

PR #2669, opened 2024-01-04
fix broken ONNX_MLIR_FLAGS

AlexandreEichenberger:
The default ONNX_MLIR_FLAGS env variable was not scanned in `onnx-mlir`. This opinion is useful and documented. This PR restore its proper usage.

### DIFF
--- a/src/Compiler/CompilerOptions.cpp
+++ b/src/Compiler/CompilerOptions.cpp
@@ -279,11 +279,16 @@ static llvm::cl::opt<std::string, true> shapeInformationOpt("shapeInformation",
     llvm::cl::value_desc("value"), llvm::cl::location(shapeInformation),
     llvm::cl::cat(OnnxMlirOptions));
 
+// Default value is defined by the OnnxMlirEnvOptionName constant string
+// variable, but the default setting mechanism here cannot be used here as we
+// need to evaluate this value prior to the compiler options being set. Proper
+// handling of the value of this compiler option is set by the calling the
+// parseCustomEnvFlagsCommandLineOption(...) function.
 static llvm::cl::opt<std::string, true> customEnvFlagsOpt("customEnvFlags",
     llvm::cl::desc("Override default option env var OnnxMlirEnvOptionName: "
                    "ONNX_MLIR_FLAGS"),
     llvm::cl::value_desc("option env var"), llvm::cl::location(customEnvFlags),
-    llvm::cl::init(OnnxMlirEnvOptionName), llvm::cl::cat(OnnxMlirOptions));
+    llvm::cl::init(""), llvm::cl::cat(OnnxMlirOptions));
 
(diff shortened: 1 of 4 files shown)


src/Compiler/CompilerOptions.cpp
```diff
@@ -281,7 +283,7 @@ static llvm::cl::opt<std::string, true> customEnvFlagsOpt("customEnvFlags",
     llvm::cl::desc("Override default option env var OnnxMlirEnvOptionName: "
                    "ONNX_MLIR_FLAGS"),
     llvm::cl::value_desc("option env var"), llvm::cl::location(customEnvFlags),
-    llvm::cl::init("ONNX_MLIR_FLAGS"), llvm::cl::cat(OnnxMlirOptions));
+    llvm::cl::init(OnnxMlirEnvOptionName), llvm::cl::cat(OnnxMlirOptions));
```
gongsu832, line 286: The initial value for `customEnvFlags` is meaningless because it's not set until `llvm::cl::ParseCommandLineOptions` is called. But we need its initial value in `parseCustomEnvFlagsCommandLineOption` before `llvm::cl::ParseCommandLineOptions` is called. So to make things less confusing, it's probably better to simply do `llvm::cl::init("")` (and add some comments about this).

src/Compiler/CompilerOptions.cpp
```diff
@@ -607,6 +609,10 @@ bool parseCustomEnvFlagsCommandLineOption(
     }
     // The envVar is verified, use it.
     setCustomEnvVar(envVar);
+  } else {
+    // It appears that the default value of this compiler options may be set too
+    // late for where its first used. So set it explicitly here.
+    setCustomEnvVar(OnnxMlirEnvOptionName);
```
gongsu832, line 615: Instead of this, I would set `envVar` to `OnnxMlirEnvOptionName` early, i.e.,
```
std::string envVar = OnnxMlirEnvOptionName;
```
Because then you can reuse the code below that checks for recursive `-customEnvFlags`.

src/Compiler/CompilerOptions.cpp
AlexandreEichenberger: @gongsu832 good suggestion, that is exactly what I discovered empirically, good to have it confirmed.

src/Compiler/CompilerOptions.cpp
AlexandreEichenberger: Good suggestion

AlexandreEichenberger:
@gongsu832 updated the code to reflect your good suggestion. The code is more consistent and robust now, thanks.

AlexandreEichenberger:
@gongsu832 do you want to do a second round? I addressed all your concerns.

gongsu832:
LGTM

PR #2671, opened 2024-01-09
Fix NNPA tests segfault

gongsu832:
NNPA tests segfault because different set of tests such as `check-onnx-backend-nnpa`, `check-onnx-backend-dynamic-nnpa`, `check-onnx-backend-constant-nnpa`, etc. are running in parallel and generating model.so in the same directory.

### DIFF
--- a/test/backend/variables.py
+++ b/test/backend/variables.py
@@ -221,11 +221,11 @@ def get_args_from_env():
 
 def get_runtime_vars():
     TEST_CASE_BY_USER = os.getenv("TEST_CASE_BY_USER")
-    if TEST_CASE_BY_USER is not None and TEST_CASE_BY_USER != "":
-        result_dir = "./"
     # Use ONNX_HOME if it is not the default /tmp
-    elif os.environ["ONNX_HOME"] != "/tmp":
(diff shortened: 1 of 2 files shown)


tungld (approved):
LGTM. Thanks!

test/backend/variables.py
```diff
@@ -221,11 +221,11 @@ def get_args_from_env():
...
 def get_runtime_vars():
     TEST_CASE_BY_USER = os.getenv("TEST_CASE_BY_USER")
-    if TEST_CASE_BY_USER is not None and TEST_CASE_BY_USER != "":
-        result_dir = "./"
     # Use ONNX_HOME if it is not the default /tmp
-    elif os.environ["ONNX_HOME"] != "/tmp":
+    if os.environ["ONNX_HOME"] != "/tmp":
         result_dir = os.environ["ONNX_HOME"]
+    elif TEST_CASE_BY_USER is not None and TEST_CASE_BY_USER != "":
+        result_dir = "./"
```
tungld, line 228: Wow, very good finding! This would be the root cause.

PR #2673, opened 2024-01-10
Remove a Window warning suppression

tungld:
Remove a Window warning suppression because the latest LLVM has included a fix for that warning.


AlexandreEichenberger (approved):
LGTM

PR #2677, opened 2024-01-11
Fix runtime errors about onnx.Unique caused by invalid condition checking of optional parameters (e.g. indices, inverseIndices and Y).

negiyas:
Fix runtime errors about onnx.Unique caused by invalid condition checking of optional parameters (e.g. indices, inverseIndices and Y).

If an optional parameter of onnx.Unique is not given, "src/Conversion/ONNXToKrnl/Tensor/Unique.cpp" generates code to call "omTensorUnique" function with  emptyMemref generated by  `Value emptyMemref = create.mem.alignedAlloc(MemRefType::get({0}, i64Type));` for its corresponding argument.

The "omTensorUnique" function should check if the corresponding argument is empty memref or not, but the current implementation checks if the pointer is NULL or not. This is not correct.


tungld:
As I understand `indices`, `inverseIndices` and `Y` are the operation outputs not parameters. So if they are None, we don't need to do any computation, e.g. calling an external function for unique, but just return None values.

When I look at the lowering of Unique, I found that it really creates a MemRef for an empty output tensor (See https://github.com/onnx/onnx-mlir/blob/main/src/Conversion/ONNXToKrnl/Tensor/Unique.cpp#L177), which, I think, is a wrong way. That's perhaps why I see you need to check if a MemRef is empty or not.

The correct way would be checking if the outputs `indices`, `inverseIndices` and `Y` are None **at the beginging** of the lowering, then just return a None value. One of the similar examples is LSTM where we handle None outputs in that way: https://github.com/onnx/onnx-mlir/blob/main/src/Conversion/ONNXToKrnl/RNN/LSTM.cpp#L606. No MemRef is created, hence no need to handle empty output tensors in external functions in `OMUnique.inc`

negiyas:
@tungld Thanks for your useful comments. I understand your points.

Let me explain the reason why I used emptyMemref value created by ‘create.mem.alignedAlloc(MemRefType::get({0}, i64Type))’ for not-given optional values instead of noneValue.

The lowering pass generates KrnlCallop op for the “omTensorUnique,” which includes three optional arguments, but the current KenlCallOp implementation does not allow None Value for the optional values (see "KrnlCallOp::build" for the detail). If noneValue is passed as an argument, compile errors occur while generating KrnlCall op.

**src/Conversion/ONNXToKrnl/Tensor/Unique.cpp**
```
Value emitArgUnique(ConversionPatternRewriter &rewriter, Location loc,
    Value total, Value input, int64_t axis, int64_t sorted, Value Y,
    Value indices, Value inverse_indices, Value counts, bool count_only) {
...
    SmallVector<Value, 8> operands = {total, Y, indices, inverse_indices,
        counts, input, val_axis, val_sorted};
    rewriter.create<KrnlCallOp>(loc, "omTensorUnique", 5, operands);
```

**src/Dialect/Krnl/KrnlOps.cpp**
```
void KrnlCallOp::build(OpBuilder &builder, ::mlir::OperationState &odsState,
    std::string funcNameStr, ValueRange resultVals, Operation *op,
    ValueRange operands, bool copyAttrs) {
  // Creates parameters for KrnlCall for Optional input (with NoneType)
  // The semantics of optional input is ONNX Op specific and should be
  // handled when lowering ONNX Op, not lowering KrnlCall.
  // For now, None input is picked out from parameters of KrnCall.
  // The Op will decide which external function to call based on the input.
  // For future work: it might be possible to assume None type is
  // always for a tensor and implemented with a nullptr in llvm.
  // Then the None input can be handled inside the external function.
  // Currently, onnx-mlir::NoneType is not handled by typeConverter of
  // ONNXToKrnl conversion.
```

Since the function has three optional arguments, we want to use a special value for not-given arguments, otherwise we need to prepare eight (=2**3) functions with different argument patterns to support all patterns.
Accordingly the current PR uses emptyMemref value created by ‘create.mem.alignedAlloc(MemRefType::get({0}, i64Type))’ as a special value for not-given optional value instead of noneValue .

Your comments and advises to avoid this situation are very welcome!  (for example, can we generates a KrnlCall with nullptr arguments?). Thanks in advance.

src/Conversion/KrnlToLLVM/KrnlCall.cpp
```diff
@@ -119,6 +119,13 @@ class KrnlCallOpLowering : public ConversionPattern {
       parameterTypeList.emplace_back(opaquePtrTy);
       parameterList.emplace_back(omTensor);
       omTensors.emplace_back(omTensor);
+    } else if (ty.isa<NoneType>()) {
+      // generate llvm null pinter for NoneType
```
tungld, line 123: nit: `generate` -> `Generate`, and please end the sentence with a full stop.

src/Conversion/KrnlToLLVM/KrnlCall.cpp
```diff
@@ -119,6 +119,13 @@ class KrnlCallOpLowering : public ConversionPattern {
...
+      auto int8Ty = IntegerType::get(context, 8);
+      auto opaquePtrTy = getPointerType(context, int8Ty);
+      parameterTypeList.emplace_back(opaquePtrTy);
+      Value nullPtr = create.llvm.null(getI8PointerType(context));
```
tungld, line 127: You can pass `opaquePtrTy` here instead of `getI8PointerType(context)`.

src/Conversion/KrnlToLLVM/KrnlNone.cpp
```diff
@@ -0,0 +1,57 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//===------ KrnlNone.cpp - Lower KrnlNoneOp --------------------------===//
```
tungld, line 5: Please expand it to 80 characters

src/Conversion/KrnlToLLVM/KrnlNone.cpp
```diff
@@ -0,0 +1,57 @@
...
+//
+// Copyright 2019-2022 The IBM Research Authors.
```
tungld, line 7: nit: `2019-2022` -> `2024`

src/Conversion/KrnlToLLVM/KrnlNone.cpp
```diff
@@ -0,0 +1,57 @@
...
+  explicit KrnlNoneOpLowering(
+      LLVMTypeConverter &typeConverter, MLIRContext *context)
+      : ConversionPattern(
+            typeConverter, KrnlNoneOp::getOperationName(), 1, context) {}
+  LogicalResult matchAndRewrite(Operation *op, ArrayRef<Value> operands,
+      ConversionPatternRewriter &rewriter) const override {
+    Location loc = op->getLoc();
+    LLVMBuilder create(rewriter, loc);
+    ModuleOp module = op->getParentOfType<ModuleOp>();
+    MLIRContext *context = module.getContext();
```
tungld, line 44: No need to get module, you can use `MLIRContext *context = rewriter.getContext();`

src/Dialect/Krnl/Krnl.td
```diff
@@ -86,7 +86,7 @@ def KrnlCallOp : Op<Krnl_Dialect, "call",
       // Should mlir::OpTrait::AttrSizedOperandSegments be used?
       // The case for krnl.Call is simple.
       DefaultValuedAttr<SI64Attr, "1">:$numOfOutput,
-      Variadic<AnyType>:$parameters);
+      Variadic<AnyTypeOf<[AnyType, NoneType]>>:$parameters);
```
tungld, line 89: Have you tried to not add NoneType here? I am not 100% sure whether AnyType includes NoneType.

src/Dialect/Krnl/Krnl.td
```diff
@@ -1273,6 +1273,26 @@ def KrnlFindIndexOp : Op<Krnl_Dialect, "find_index",
...
+
+  let arguments = (ins UnitAttr:$value);
+  let results = (outs NoneType:$none_val);
+
+  let builders = [
+    OpBuilder<(ins),[{
+      build($_builder, $_state, $_builder.getNoneType(), $_builder.getUnitAttr());
+    }]>];
+}
+
```
tungld, line 1295: Thanks for adding this!

src/Runtime/OMUnique.inc
```diff
@@ -322,6 +321,11 @@ void produceY(const OMTensor *inputTensor, OMTensor *indices, int64_t sliceAxis,
   }
 }
 
+static bool isEmptyTensor(OMTensor *tensor) {
+  return (tensor == NULL) || (omTensorGetRank(tensor) <= 0) ||
+         (omTensorGetShape(tensor)[0] == 0);
```
tungld, line 326: By using nullptr for None, do we still need to check rank and shape?

src/Conversion/KrnlToLLVM/KrnlCall.cpp
negiyas: Fixed. Thanks.

src/Conversion/KrnlToLLVM/KrnlCall.cpp
negiyas: Fixed, and confirmed that it works. Thanks.

src/Conversion/KrnlToLLVM/KrnlNone.cpp
negiyas: Fixed. Thanks.

src/Conversion/KrnlToLLVM/KrnlNone.cpp
negiyas: Fixed. Thanks.

src/Conversion/KrnlToLLVM/KrnlNone.cpp
negiyas: Fixed. Thanks.

src/Dialect/Krnl/Krnl.td
negiyas: I have tried the original Krnl.td, and have confirmed that it works for lit tests.
Accordingly I use the original Krnl.td in this PR. Thanks for your variable comments!

src/Dialect/Krnl/Krnl.td
negiyas: Thanks.

src/Runtime/OMUnique.inc
negiyas: Removed the isEmptyTensor funciont, and fixed the code to check "pointer == NULL" directly. Thanks!

tungld (approved):
LGTM.

PR #2679, opened 2024-01-11
Add onnx-mlir command options when using -v

AlexandreEichenberger:
Lost some time debugging because I had a `ONNX_MLIR_FLAGS=xxx` that I was not aware of. 

Now if we use the `-v` flag, it prints the value of the default env variable for the flags, plus the actual options used.

```
ONNX_MLIR_FLAGS="-O0" onnx-mlir -O3 -march=arm64 -parallel -v aiu-par-test-v2.mlir 
Onnx-mlir default options from 'ONNX_MLIR_FLAGS' are '-O0'.
Onnx-mlir command: onnx-mlir -O0 -O3 -march=arm64 -parallel -v aiu-par-test-v2.mlir
...
```

so that we may more easily detect what is going on.

Also, the `-v` was using a mixture of `stdout` and `stderr` as output. Found online a recommendation to use `stdout`, so switched to use that systematically.


src/Compiler/CompilerOptions.cpp
```diff
@@ -601,6 +603,8 @@ bool parseCustomEnvFlagsCommandLineOption(
       envVar = arg.substr(sizeof("-customEnvFlags"));
       break;
     }
+    if (arg.compare("-v") == 0)
+      verbose = true;
```
gongsu832, line 607: The arguments are scanned backwards so if `-customEnvFlags` or `---customEnvFlags` is specified after `-v`, `verbose` won't be set.

Also, if `-v` is specified inside the env var, e.g., `ONNX_MLIR_FLAGS="-v -O0"`, one would expect to get the same output below.

AlexandreEichenberger: Good points which I can address.

AlexandreEichenberger: Its now working: Below and example of the `-v` caught from the env var.

```
ONNX_MLIR_FLAGS="-O0 -v" onnx-mlir -O3 -march=arm64 -parallel aiu-par-test-v2.mlir 
Onnx-mlir default options from 'ONNX_MLIR_FLAGS' are '-O0 -v'.
Onnx-mlir command: onnx-mlir -O0 -v -O3 -march=arm64 -parallel aiu-par-test-v2.mlir
[/Users/alexe/OM/atests/ops/] /Users/alexe/OM/llvm-project/build/bin/opt: opt -O3 --mtriple=arm64-apple-darwin23.2.0 --march=arm64 --code-model small -o aiu-par-test-v2.bc aiu-par-test-v2.unoptimized.bc
[/Users/alexe/OM/atests/ops/] /Users/alexe/OM/llvm-project/build/bin/llc: llc -O3 --mtriple=arm64-apple-darwin23.2.0 --march=arm64 --code-model small -filetype=obj -relocation-model=pic -o aiu-par-test-v2.o aiu-par-test-v2.bc
[/Users/alexe/OM/atests/ops/] /usr/bin/c++: c++ aiu-par-test-v2.o -o aiu-par-test-v2.so -shared -fPIC -L/Users/alexe/OM/onnx-mlir/build/Debug/lib -lcruntime -lompruntime
Shared library 'aiu-par-test-v2.so' has been compiled.
```

AlexandreEichenberger:
@gongsu832 Thanks for the comment; added a fix to cover the case you recommended. Ready for another round of review/approval.

gongsu832:
> @gongsu832 Thanks for the comment; added a fix to cover the case you recommended. Ready for another round of review/approval.

The case where `verbose` doesn't get set if `-customEnvFlags` or `--customEnvFlags` is specified after `-v` doesn't seem to be fixed yet. For example, if someone does,
```
export MY_ONNX_MLIR_FLAGS="-O3"
onnx-mlir -v -customEnvFlags=MY_ONNX_MLIR_FLAGS ...
```
`verbose` won't be set.

AlexandreEichenberger:
Right you are, this an another condition are fixed now.

src/Compiler/CompilerOptions.cpp
```diff
@@ -590,16 +590,17 @@ bool parseCustomEnvFlagsCommandLineOption(
   // Use the default ONNX MLIR Environment variable, unless specified otherwise
   // by an argument, see below.
   std::string envVar = OnnxMlirEnvOptionName;
+  // VerboseOutput is not yet set, so scan ourselves.
+  bool verbose = false;
   // Customized version? -customEnvFlags=val and save its value.
-  for (int i = argc - 1; i > 1; --i) {
+  for (int i = argc - 1; i > 0; --i) {
```
gongsu832, line 596: For command line arguments, when the same option is specified multiple times, typically the last one takes effect. That's probably why the arguments were scanned backwards originally. With the changes just made, you don't "break" at the first match (last instance) of `-customEnvFlags` or `--customEnvFlags`. Therefore, to keep the last-instance-takes-effect semantic, the arguments should now be scanned forwards instead of backwards.

AlexandreEichenberger:
@gongsu832 ready for review

gongsu832 (approved):
LGTM

AlexandreEichenberger:
Thanks @gongsu832 , that was a very thorough review that uncovered many mistake of mine, appreciated.

Issue #2667, opened 2023-12-25
Failed Test Sequence_with_dealloc


Connor-XY:
When I ran the test `conversion/onnx_to_krnl/Sequence/Sequence_with_dealloc` on my end, the test failed because two operands got inverted:

```
FAIL: Open Neural Network Frontend :: conversion/onnx_to_krnl/Sequence/Sequence_with_dealloc.mlir (1 of 324)
******************** TEST 'Open Neural Network Frontend :: conversion/onnx_to_krnl/Sequence/Sequence_with_dealloc.mlir' FAILED ********************
Exit Code: 1

Command Output (stderr):

RUN: at line 1: /data00/yan.xu/onnx-mlir/build/Debug/bin/onnx-mlir-opt -O3 --shape-inference --convert-onnx-to-krnl --buffer-deallocation-pipeline  /data00/yan.xu/onnx-mlir/test/mlir/conversion/onnx_to_krnl/Sequence/Sequence_with_dealloc.mlir -split-input-file | /data00/yan.xu/llvm-project/build/bin/FileCheck /data00/yan.xu/onnx-mlir/test/mlir/conversion/onnx_to_krnl/Sequence/Sequence_with_dealloc.mlir
+ /data00/yan.xu/onnx-mlir/build/Debug/bin/onnx-mlir-opt -O3 --shape-inference --convert-onnx-to-krnl --buffer-deallocation-pipeline /data00/yan.xu/onnx-mlir/test/mlir/conversion/onnx_to_krnl/Sequence/Sequence_with_dealloc.mlir -split-input-file
+ /data00/yan.xu/llvm-project/build/bin/FileCheck /data00/yan.xu/onnx-mlir/test/mlir/conversion/onnx_to_krnl/Sequence/Sequence_with_dealloc.mlir
/data00/yan.xu/onnx-mlir/test/mlir/conversion/onnx_to_krnl/Sequence/Sequence_with_dealloc.mlir:52:15: error: CHECK-DAG: expected string not found in input
// CHECK-DAG: [[VAR_14_:%.+]] = affine.apply [[MAP_1_]](){{.}}[[VAR_12_]], [[VAR_dim_0_]]{{.}}
              ^
<stdin>:41:43: note: scanning from here
 %12 = arith.index_cast %11 : i64 to index
                                          ^
<stdin>:41:43: note: with "MAP_1_" equal to "#map1"
 %12 = arith.index_cast %11 : i64 to index
                                          ^
[...]

PR #2680, opened 2024-01-15
[onnx-to-krnl] remove the check of operands for affine.apply

Connor-XY:
Remove the check of the order of operands in affine.apply, which is discussed in https://github.com/onnx/onnx-mlir/issues/2667.


tungld (approved):
LGTM!

PR #2684, opened 2024-01-19
Migrate adding alloca_scope to scf passes (for parallel with NNPA)

AlexandreEichenberger:
Previous approach of adding `alloca_scope` at the affine level did not work for NNPA as additional passes of `canonicalize` were called between adding `alloca_scope` and handling of buffers when generating NNPA code.

This problem is fixed here by migrating that pass to work with `scf` and calling that pass just before handling buffers.

Once this is well established, we should probably remove the old pass at the affine level as its unlikely to be useful anymore.

With this fix, Roberta passes with NNPA:
```
CheckONNXModel.py -m roberta-base-11.onnx -r="-O3 -mcpu=z16 -maccel=NNPA" -a="--parallel" --shape-info=0:8x384 

> Reference command:  /workdir/onnx-mlir/build/Debug/../../utils/RunONNXModel.py --compile-args="-O3 -mcpu=z16 -maccel=NNPA" --save-ref=check-ref --shape-info=0:8x384 --seed=42 --model=roberta-base-11.onnx
>   Successfully ran the reference example, saved refs in "check-ref".

> Test command:  /workdir/onnx-mlir/build/Debug/../../utils/RunONNXModel.py --compile-args="-O3 -mcpu=z16 -maccel=NNPA --parallel" --load-ref=check-ref --verify=ref --verify-every-value --model=roberta-base-11.onnx
>   Successfully ran the test example and verified against "check-ref".
```


tungld (approved):
LGTM.

PR #2685, opened 2024-01-19
Fixes for clean run of backend with parallelism enabled

AlexandreEichenberger:
I recently became aware of how to run `check-onnx-backend` with `-parallel` enabled on Linux-based platforms. It exposed a few bugs that are fixed in this PR. It now run cleanly.

```
OMP_NUM_THREADS=6 TEST_NOFLOAT16=true ONNX_MLIR_FLAGS="-O3 -mcpu=z16 -parallel" make -j 12 check-onnx-backend
[...]
bringing up nodes...
                                                                                                                                                                                                                                                                   [100%]
538 passed, 2268 skipped in 185.14s (0:03:05)
[100%] Built target check-onnx-backend
....

One of the issue is that `alloca_scope` is a bit fragile for some tests, but running `canonicalize` shortly after bufferization removes redundant copies, so that work well. Another other issue is that we were attempting to parallelize output loops for ops generating scalar output, so parallelism is now disabled for such loops.

This PR builds on #2684.


AlexandreEichenberger:
@tungld Note that this does not turn on parallelism by default or test it yet. It is just to enable a clean run

tungld (approved):
LGTM.

I though I reviewed this yesterday, but not :(

PR #2687, opened 2024-01-23
Update document for LLVM and StableHLO process

hamptonm1:
I pretty much kept the same outline of what was written previously and simply changed `MLIR-HLO` to `StableHLO`. I also updated the links so that we can reference the appropriate location of things. As a community we agreed on updating LLVM and StableHLO simultaneously which would be much easier than separating the two processes (It keeps everything in sync on a biweekly cadence). Please provide comments and feedback if you see any way we can improve what I have. Thanks so much for reviewing!


hamptonm1:
FYI- @Connor-XY

ashay (approved):

hamptonm1:

mikeessen (approved):
LGTM!

PR #2688, opened 2024-01-25
[NNPA] zDNN extension for splitting ztensors

tungld:
This patch is a prototype of zDNN extension that splits a zDNN call into multiple zDNN calls over chunks of the input. Splitting is done directly on ztensor instead of CPU tensor. zDNN matmul is supported in this patch.

For users, there are three variable environments to control splitting ztensors:
- OM_ZTENSOR_SPLIT_ENABLED: to turn on/off the splitting.
- OM_ZTENSOR_SPLIT_SIZE: to set the size of each chunk along an axis.
- OM_ZTENSOR_SPLIT_DEBUG: to enable debugging, that includes timing.

To compile a large matmul, users need to specify `--enable-zhigh-decompose-stick-unstick` in order to stickify/unstickify very large tensors.

For developers, this patch provides two helper functions:
- `void splitZTensor(const zdnn_ztensor *input, SplitInfo *splitInfo, bool copyData)`: to split a ztensor into multiple ztensors stored in SplitInfo.
- `void mergeZTensors(const SplitInfo *splitInfo, zdnn_ztensor *output)`: to merge multiple ztensors in SplitInfo into a single ztensor.

Once a ztensor is splitted, it is just a loop to call zdnn matmul for each chunk, e.g. 
```C
 for (uint32_t i = 0; i < splitInfoA.numOfChunks; ++i) {
    status = call_zdnn_matmul_op(splitInfoA.tensors + i, inputB, inputC, opType,
        splitInfoY.tensors + i, isBcast);
    assert(status == ZDNN_OK);
  }
```


src/Accelerators/NNPA/Runtime/zDNNExtension/MatMul.c
```diff
@@ -0,0 +1,141 @@
...
+    printf("Split zTensor A along axis %d into %d chunks of %d elements \n",
+        axis, splitInfoA.numOfChunks, splitInfoA.chunkSize);
+  initSplitInfo(output, axis, chunkSize, &splitInfoY);
+
+  double splitTime = 0.;
+  double mmTime = 0.;
+  double mergeTime = 0.;
+  clock_t start_time, end_time;
+
+  // Split input A into chunks.
```
AlexandreEichenberger, line 82: There are a lot of implicit assertions here (e.g. only A is too big/needs splitting, only in the e2 dim), which is fine. I would recommend to list them at the top of the function so that we have them explicitly listed.

src/Accelerators/NNPA/Runtime/zDNNExtension/zDNNExtension.c
```diff
@@ -0,0 +1,280 @@
...
+                         (uint64_t)shape->dim2 * (uint64_t)shape->dim1 *
+                         (uint64_t)AIU_2BYTE_CELL_SIZE;
+  uint64_t sizeFromBuffer = t->buffer_size;
+  assert(sizeFromDim == sizeFromBuffer && "buffer size mismatched");
+}
+
+zdnn_status allocZTensorChunk(const zdnn_ztensor *input, uint32_t axis,
+    uint32_t chunkSize, zdnn_ztensor *output) {
+  zdnn_status status;
+  zdnn_tensor_desc *preTransDesc = malloc(sizeof(zdnn_tensor_desc));
```
AlexandreEichenberger, line 76: Since you control the free, you could allocate a single `2*sizeof(zdnn_tensor_desc)` array and assign `[0]` to `sizeof(zdnn_tensor_desc)` and `[1]` to `transDesc` and thus have only one malloc.

I would also have an assert / return status. 

Speaking of it, I am confused because many of these functions have return of status but we also assert on failure. This may need to be cleaned up.

src/Accelerators/NNPA/Runtime/zDNNExtension/zDNNExtension.c
```diff
@@ -0,0 +1,280 @@
...
+  status = zdnn_init_ztensor_with_malloc(preTransDesc, transDesc, output);
+  return status;
+}
+
+zdnn_status freeZTensorChunk(zdnn_ztensor *t) {
+  if (t->pre_transformed_desc)
+    free(t->pre_transformed_desc);
+  if (t->transformed_desc)
+    free(t->transformed_desc);
+  zdnn_free_ztensor_buffer(t);
```
AlexandreEichenberger, line 103: Is it true that the `zdnn_free_ztensor_buffer` does not look at the `t->pre_transformed_desc` or `t->transformed_desc`?

If not sure, it might be safer to first free the buffer and then the descriptors?

AlexandreEichenberger (approved):
A very good first cut. It could be merged as is, or refined a bit before pushing it down.

Eventually, with parallelization across several NNPA / L3, we may want to move to a single loop that has the "split -> compute -> merge" inside that loop so as to use additional cache capacity of the the multiple L3 caches. This will require a bit of code reorg, nothing major.

src/Accelerators/NNPA/Runtime/zDNNExtension/MatMul.c
```diff
@@ -0,0 +1,141 @@
...
+#include <sys/time.h>
+
+#include "src/Accelerators/NNPA/Runtime/zDNNExtension/zDNNExtension.h"
+#include "zdnn.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+zdnn_status call_zdnn_matmul_op(const zdnn_ztensor *inputA,
```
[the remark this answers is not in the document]
tungld: Done.

src/Accelerators/NNPA/Runtime/zDNNExtension/MatMul.c
```diff
@@ -0,0 +1,141 @@
...
+zdnn_status zdnn_matmul_op_common(const zdnn_ztensor *inputA,
+    const zdnn_ztensor *inputB, const zdnn_ztensor *inputC, int opType,
+    zdnn_ztensor *output, bool isBcast) {
+  zdnn_status status;
+  bool isDebug = ZTensorSplitDebugFromEnv();
+
+  // For a MatMul of (M,N)*(N,P),
+  // We split M that is the third one in the original 4D tensor.
+  uint32_t axis = 2;
+  uint32_t chunkSize = ZTensorSplitSizeFromEnv();
```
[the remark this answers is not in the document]
tungld: I created a function `zDNNExtensionInit()` to initialize these variables, then call it from `OMInitCompatibleAccelNNPA`.

> Long term, do you think its more appropriate to have theses values dictated by the compiler (aka becoming a parameter to the extension) or a runtime env?

Yes, and for different tensor size, `chunkSize` would be tuned automatically.

src/Accelerators/NNPA/Runtime/zDNNExtension/MatMul.c
```diff
@@ -0,0 +1,141 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//===-------------------------- MatMul.c ----------------------------------===//
+//
+// Copyright 2024 The IBM Research Authors.
+//
+// =============================================================================
+//
+// A wrapper of zdnn_matmul_op supports very large input tensors.
```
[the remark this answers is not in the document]
tungld: Yes, updated this to `A wrapper of zdnn_matmul_op for ztensor partition and parallelism.`.

src/Accelerators/NNPA/Runtime/zDNNExtension/MatMul.c
```diff
@@ -0,0 +1,141 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+zdnn_status call_zdnn_matmul_op(const zdnn_ztensor *inputA,
+    const zdnn_ztensor *inputB, const zdnn_ztensor *inputC, int opType,
+    zdnn_ztensor *output, bool isBcast) {
+  zdnn_status status;
+  if (isBcast) {
+    status = zdnn_matmul_bcast_op(
```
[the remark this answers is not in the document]
tungld: Done.

src/Accelerators/NNPA/Runtime/zDNNExtension/MatMul.c
tungld: Now, we checked the other dimensions also.

src/Accelerators/NNPA/Runtime/zDNNExtension/zDNNExtension.c
tungld: Good suggestion. Done.

src/Accelerators/NNPA/Runtime/zDNNExtension/zDNNExtension.c
tungld: > Is it true that the zdnn_free_ztensor_buffer does not look at the t->pre_transformed_desc or t->transformed_desc?

Yes, it's true. But I still moved `free the buffer` up as your suggestion.

tungld:
@AlexandreEichenberger thank you for your comments. I addressed all of them except utilising L3 cache.

imaihal:
General question. This seems to be useful not only for onnx-mlir users, but also for other zDNN users. Will this function be moved to the zDNN library in the future?

tungld:
> General question. This seems to be useful not only for onnx-mlir users, but also for other zDNN users. Will this function be moved to the zDNN library in the future?

It's great if this can be shared among multiple frameworks. However, inside a compiler, we may have more opportunities to do optimization for multiple zdnn ops, e.g. so don't need to split/merge for consecutive zdnn ops.

PR #2689, opened 2024-01-26
Added an onnx-mlir option to select which operation to parallelize

AlexandreEichenberger:
Up to know, we either parallelized all operations (with `-parallel`) or none. This PR add the ability to define a `-parallelize-ops=regexp` to decide which operation should be parallelized. It must be used in conjunction with the `-parallel` flag.

As we already had some support for regex processing, I factorized the regex processing from InstrumentOps to compiler util so that it may be reused where ever needed.


AlexandreEichenberger:
Looks like I have a circular dependence in the makefiles that shows up when compiling under Release; will address tomorrow.

tungld (approved):
LGTM.

PR #2690, opened 2024-01-29
[NNPA] zDNN extension for splitting elementwise operations 

tungld:
This patch provides helper functions that split zdnn elementwise operations if required. I also cleaned up some function signatures a bit to make it more concise.


src/Accelerators/NNPA/Runtime/zDNNExtension/zDNNExtension.c
```diff
@@ -223,10 +270,9 @@ static void copyZTensorChunkScalar(zdnn_ztensor *output,
               // Copy 2 bytes at a time.
               uint64_t offsetSrc =
                   AIU_2BYTE_CELL_SIZE * (d1 + D1 * (d2 + D2 * SD3Offset));
-              uint64_t offsetDest =
+              uint64_t offsetDst =
                   AIU_2BYTE_CELL_SIZE * (d1 + D1 * (d2 + D2 * TD3Offset));
-              memcpy(output->buffer + offsetDest, input->buffer + offsetSrc,
-                  AIU_2BYTE_CELL_SIZE);
+              memcpy(dst + offsetDst, src + offsetSrc, AIU_2BYTE_CELL_SIZE);
```
AlexandreEichenberger, line 275: I would just check that there is not large overheads to do a memcpy for 2 bytes. Otherwise, I would just use `int16_t` pointers.

AlexandreEichenberger (approved):
LGTM

src/Accelerators/NNPA/Runtime/zDNNExtension/zDNNExtension.c
tungld: I changed to use uint16_t pointers. It would be better. Thanks!

PR #2692, opened 2024-01-31
Fix compile messages for z/OS

mikeessen:
This PR adds a header guard for zDNNExtension.h instead of using `#pragma once`.  The `#pragma once` is not supported on z/OS and causes a warning during compile. This warning shows up when compiling the Runtime code to be used when linking a z/OS shared object model.

This PR also corrects a warning message issued when compiling jniwrapper.c on z/OS.

### DIFF
--- a/src/Accelerators/NNPA/Runtime/zDNNExtension/zDNNExtension.h
+++ b/src/Accelerators/NNPA/Runtime/zDNNExtension/zDNNExtension.h
@@ -12,7 +12,8 @@
 //
 //===----------------------------------------------------------------------===//
 
-#pragma once
+#ifndef ONNX_MLIR_ZDNNEXTENSION_H
+#define ONNX_MLIR_ZDNNEXTENSION_H
 
 #include "zdnn.h"
 
@@ -180,3 +181,5 @@ zdnn_status zdnn_tanh_ext(const zdnn_ztensor *input, zdnn_ztensor *output);
 #ifdef __cplusplus
 }
 #endif
+
+#endif // ONNX_MLIR_ZDNNEXTENSION_H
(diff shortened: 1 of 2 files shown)


tungld (approved):
LGTM. Thanks!

gongsu832:
@mikeessen What compiler command are you using to compile `jniwrapper.c`? `SetLongArrayRegion` is a standard JNI function that takes a const pointer for the last argument per JNI documentation:
```
void Set<PrimitiveType>ArrayRegion(JNIEnv *env, ArrayType array, jsize start, jsize len, const NativeType *buf);
```
So passing in a const pointer is the correct thing to do unless z/OS JNI implementation does not conform to the JNI documentation. It's generally a bad idea to remove the const qualifier to fix warnings since the const qualifier helps the compiler catch unintended writes into the const buffer.

mikeessen:
> @mikeessen What compiler command are you using to compile `jniwrapper.c`? `SetLongArrayRegion` is a standard JNI function that takes a const pointer for the last argument per JNI documentation:
> 
> ```
> void Set<PrimitiveType>ArrayRegion(JNIEnv *env, ArrayType array, jsize start, jsize len, const NativeType *buf);
> ```
> 
> So passing in a const pointer is the correct thing to do unless z/OS JNI implementation does not conform to the JNI documentation. It's generally a bad idea to remove the const qualifier to fix warnings since the const qualifier helps the compiler catch unintended writes into the const buffer.

@gongsu832 The Java 8 jni.h definition of SetLongArrayRegion does not contain the const qualifier.  In Java 11 it does, but we need to continue to support Java 8.  Once we no longer need to support Java 8 we can remove the z/OS specific code for this.  Thanks!

```
``` void (JNICALL * SetLongArrayRegion)(JNIEnv *env, jlongArray array, jsize start, jsize len, jlong *buf);
```

gongsu832:
Interesting. On Linux, J9 Java 8 jni.h definition of SetLongArrayRegion does not contain const, but OpenJDK Java 8 does. Sigh.

PR #2694, opened 2024-01-31
Convert CastLikeOp to CastOp in Rewrite Rules

hamptonm1:
It appears that the handling of `CastLikeOp` was incorrect. According to the ONNX documentation, `CastLike` is not unary but rather binary (for some reason CastLike was being treated as unary). So I added code to elementwise to fully implement this operation and I also modified files this way things can work accordingly. I also added the backend tests (about 6 roughly) for `CastLikeOp` which ran successfully. 

Other things to note:

1. I added the canonicalize pass for EmitONNXIR so that this way we are utilizing the pattern for the normalization of constant op (make sure constants are dense) along with the pattern I included to eliminate castlike op's that have the same tensor input. 

1. For consistency, I moved `IsNaNOp` to the ElementUnary.cpp file since it is applicable. 

**UPDATE**

Based on the commentary, I simply converted `CastLikeOp` to `CastOp`


src/Dialect/ONNX/Rewrite.td
```diff
@@ -431,6 +439,18 @@ def CastEliminationPattern : Pat<
...
+// Canonicalization for ONNXCastLikeOp
+//===----------------------------------------------------------------------===//
+
+def CastLikeEliminationPattern:  Pat<
+  // Remove a CastLike pattern if the first input tensor and second input tensor are the same. 
+  (ONNXCastLikeOp $val, $castTo, $saturate),
+  // Remove the castlike.
+  (replaceWithValue $val),
+  // Check that the first input tensor and second input tensor are the same
+  [(HaveSameShape $val, $castTo)]>;
```
tungld, line 452: Quick comment, checking shape is not enough to know two inputs are the same. It would be "$val == $castTo"

hamptonm1: Thanks for the feedback…. I saw an ‘equal’ constraint. I can use that maybe

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -473,6 +473,33 @@ Value emitScalarOpFor<ONNXCastOp>(ConversionPatternRewriter &rewriter,
...
+Value emitScalarOpFor<ONNXCastLikeOp>(ConversionPatternRewriter &rewriter,
+    Location loc, Operation *op, Type elementType,
+    ArrayRef<Value> scalarOperands) {
+
+  Value input = scalarOperands[0];
+  Value target = scalarOperands[1];
+
+  Type targetType = target.getType();
+
+  CheckIfCustomScalarOpIsSupported<ONNXCastOp>(targetType);
```
tungld, line 495: ONNXCastLike?

src/Dialect/ONNX/ONNXOps/Math/ElementwiseUnary.cpp
```diff
@@ -334,6 +320,16 @@ LogicalResult ONNXIsInfOp::inferShapes(
...
+//===----------------------------------------------------------------------===//
+// IsNaN
+//===----------------------------------------------------------------------===//
+
+LogicalResult ONNXIsNaNOp::inferShapes(
+    std::function<void(Region &)> doShapeInference) {
+  IntegerType i1Type = IntegerType::get(getContext(), 1, IntegerType::Signless);
+  return inferShapeForUnaryOps(getOperation(), i1Type);
+}
+
```
tungld, line 332: Thanks! This looks better.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
hamptonm1: Oops...my fault. That was from copying and pasting haha

hamptonm1: Fixed!

src/Dialect/ONNX/Rewrite.td
hamptonm1: Updated!

hamptonm1: Wait nvm... I did a check for the same shape and same element type

src/Dialect/ONNX/ONNXOps/Math/ElementwiseBroadcast.cpp
```diff
@@ -161,6 +161,20 @@ LogicalResult ONNXBitShiftOp::inferShapes(
   return inferShapeForBroadcastingOps<ONNXBitShiftOp>(*this);
 }
 
+//===----------------------------------------------------------------------===//
+// CastLikeOp
+//===----------------------------------------------------------------------===//
+LogicalResult ONNXCastLikeOp::inferShapes(
```
hamptonm1, line 167: @chentong319 We previously talked about mapping the ONNXCastLike to use ONNXCast if it is possible in my next PR. Will create an issue to tag said progress :)

test/numerical/CMakeLists.txt
```diff
@@ -63,20 +63,20 @@ add_numerical_test(TestGemm
...
 
 add_numerical_test(TestRNN
   TestRNN.cpp
   LINK_LIBS PRIVATE ${TEST_LINK_LIBS}
   )
 
-add_numerical_test(TestGRU
-  TestGRU.cpp
-  LINK_LIBS PRIVATE ${TEST_LINK_LIBS}
-  )
```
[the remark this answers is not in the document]
hamptonm1: Below is the errors or click the [link](https://www.onnxmlir.xyz/jenkinx/blue/organizations/jenkins/ONNX-MLIR-Pipeline-Docker-Build/detail/ONNX-MLIR-Pipeline-Docker-Build/13960/pipeline/) :

```
 1/10 Test  #5: TestLSTM .........................***Failed   27.05 sec
Model can reuse the current seed by exporting "TEST_SEED=761153055"
Target options: "--maccel=NONE"
Using configuration: seed=3718789376380535031

- LSTM implementation correctness
error: type of return operand 0 ('tensor<?x?x2x5xf32>') doesn't match function result type ('tensor<*xf32>') in function @main_graph
error: type of return operand 0 ('tensor<?x?x2x5xf32>') doesn't match function result type ('tensor<*xf32>') in function @main_graph
error: type of return operand 0 ('tensor<?x?x2x5xf32>') doesn't match function result type ('tensor<*xf32>') in function @main_graph
attempt 1 with direction 2, S 1, B 5, I 5, H 5, isDynS 1, isDynB 1, isNoneH 0, isNoneC 0, isNoneP 0, layout 0
attempt 2 with direction 2, S 1, B 5, I 5, H 5, isDynS 0, isDynB 0, isNoneH 0, isNoneC 0, isNoneP 0, layout 0
attempt 3 with direction 2, S 1, B 5, I 5, H 5, isDynS 1, isDynB 1, isNoneH 0, isNoneC 0, isNoneP 0, layout 1
attempt 4 with direction 2, S 1, B 5, I 5, H 5, isDynS 1, isDynB 1, isNoneH 0, isNoneC 0, isNoneP 0, layout 1
attempt 5 with direction 2, S 1, B 5, I 5, H 5, isDynS 1, isDynB 1, isNoneH 0, isNoneC 0, isNoneP 0, layout 1
Falsifiable after 3 tests

int:
2

int:
1

int:
5

int:
5

int:
5

int:
1

int:
0

int:
0

int:
0

int:
0

int:
0

/workdir/onnx-mlir/test/numerical/TestLSTM.cpp:101:
RC_ASSERT(isOMLSTMTheSameAsNaiveImplFor(D, S, B, I, H, isDynS == 0, isDynB == 0, isNoneH == 1, isNoneC == 1, isNoneP == 1, layout))

Expands to:
false
Some of your RapidCheck properties had failures. To reproduce these, run with:
RC_PARAMS="reproduce=B8BTTRVTgkWbwxWZtVmb0FGdp9mbgM2byJXZjRnblN3c3DIqdwzybOz9AiaH8s8mzcPgo2BPLv5M3DIqdwzybODBAMgAAAAAAA"

 2/10 Test  #7: TestGRU ..........................***Failed   16.67 sec
Model can reuse the current seed by exporting "TEST_SEED=1168459696"
Target options: "--maccel=NONE"
Using configuration: seed=17122692856318736406

- GRU implementation correctness
error: type of return operand 0 ('tensor<5x?x2x5xf32>') doesn't match function result type ('tensor<*xf32>') in function @main_graph
error: type of return operand 0 ('tensor<5x?x2x5xf32>') doesn't match function result type ('tensor<*xf32>') in function @main_graph
error: type of return operand 0 ('tensor<5x?x2x5xf32>') doesn't match function result type ('tensor<*xf32>') in function @main_graph
attempt 1 with direction 1, S 1, B 5, I 5, H 5, linearBeforeReset 0, isDynS 1, isDynB 1, layout 0
attempt 2 with direction 2, S 1, B 5, I 5, H 5, linearBeforeReset 0, isDynS 1, isDynB 0, layout 1
attempt 3 with direction 2, S 1, B 5, I 5, H 5, linearBeforeReset 0, isDynS 1, isDynB 0, layout 1
attempt 4 with direction 2, S 1, B 5, I 5, H 5, linearBeforeReset 0, isDynS 1, isDynB 0, layout 1
Falsifiable after 2 tests

int:
2

int:
1

int:
5

int:
5

int:
5

int:
1

int:
0

int:
0

int:
1

/workdir/onnx-mlir/test/numerical/TestGRU.cpp:96:
RC_ASSERT(isOMGRUTheSameAsNaiveImplFor( D, S, B, I, H, L, isDynS == 0, isDynB == 0, layout))

Expands to:
false
Some of your RapidCheck properties had failures. To reproduce these, run with:
RC_PARAMS="reproduce=B4xRSVFIp1GcsVWbl5GdhRXav5GIj9mcyV2Y05WZzNnF48v0uZAotbBO_LtbGAa7Wgz_S7mBg2uF48v0uZAotLAACEAAAAAA"

```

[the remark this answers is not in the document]
tungld: It potentially relates to adding `canonicalization` pass. Could you try undoing adding `canonicalization` pass to see if there is the error or not?

src/Dialect/ONNX/Rewrite.td
```diff
@@ -431,6 +439,18 @@ def CastEliminationPattern : Pat<
...
+// Canonicalization for ONNXCastLikeOp
+//===----------------------------------------------------------------------===//
+
+def CastLikeEliminationPattern:  Pat<
+  // Remove a CastLike pattern if the first input tensor and second input tensor are the same. 
+  (ONNXCastLikeOp $val, $castTo, $saturate),
+  // Remove the castlike.
+  (replaceWithValue $val),
+  // Check that the first input tensor and second input tensor are the same
+  [(HaveSameElementType $val, $castTo), (HaveSameShape $val, $castTo)]>;
```
tungld, line 452: Do you think only `HaveSameElementType` is enough? `HaveSameShape` looks redundant since we use the type of the target only.

hamptonm1: @tungld The `Equal` did not work I guess because it is designed to compare to ints and not tensors... do you have another suggestion?  Maybe I can try to create another rule?

test/numerical/CMakeLists.txt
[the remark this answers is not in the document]
hamptonm1: Sure!! Let me try

src/Dialect/ONNX/Rewrite.td
tungld: Your `HaveSameElementType` looks enough. Have you tried it only? without `HaveSameShape`,

hamptonm1: Let me do so now :)

hamptonm1: It works just fine!

test/numerical/CMakeLists.txt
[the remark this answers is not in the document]
hamptonm1: @tungld You are right! it is the canonicalization pass... the tests pass once I remove it. What is the work-around in this case because we still need the pass for now?

```

[100%] Built target numerical
[100%] Running the ONNX-MLIR numerical regression tests
      Start  1: TestConv
      Start  2: TestMatMul2D
      Start  3: TestMatMulBroadcast
      Start  4: TestGemm
      Start  5: TestLSTM
      Start  6: TestRNN
      Start  7: TestGRU
      Start  8: TestLoop
      Start  9: TestScan
      Start 10: TestElementwise
.s.s 1/10 Test  #8: TestLoop .........................   Passed    2.08 sec
.s.s1/1 Test #1: TestCategoryMapper ...............   Passed    3.27 sec

```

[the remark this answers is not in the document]
tungld: One workaround is moving the canonicalization pass down, e.g. calling it after Decompose/Recompose, or even later. But I am not 100% sure if it works.

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -1582,12 +1582,12 @@ int ImportFrontendModelFile(StringRef model_fname, MLIRContext &context,
     if (ret != CompilerSuccess)
       return ret;
 
-    auto status = google::protobuf::util::JsonStringToMessage(json, &model);
-    if (!status.ok()) {
-      *errorMessage = "Json Model Parsing Failed on " + model_fname.str() +
-                      " with error '" + status.ToString() + "'";
-      return InvalidOnnxFormat;
-    }
+    // auto status = google::protobuf::util::JsonStringToMessage(json, &model);
```
AlexandreEichenberger, line 1585: Why is this removed? For debugging?

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -473,6 +473,33 @@ Value emitScalarOpFor<ONNXCastOp>(ConversionPatternRewriter &rewriter,
   return create.math.cast(elementType, scalarOperands[0]);
 }
 
+//===----------------------------------------------------------------------===//
```
AlexandreEichenberger, line 476: @hamptonm1 I thought we would avoid having CastLike because they are not optimized away in constant propagation. Isn't it too late to remove it here, during ONNX to Krnl lowering? I thought we would nuke them in Rewrite.

src/Dialect/ONNX/ONNXOps/Math/ElementwiseBroadcast.cpp
AlexandreEichenberger: If its removed before even hitting shape inference, we don't need this code at all

src/Dialect/ONNX/Rewrite.td
```diff
@@ -431,6 +435,18 @@ def CastEliminationPattern : Pat<
...
+//===----------------------------------------------------------------------===//
+
+def CastLikeEliminationPattern:  Pat<
+  // Remove a CastLike pattern if the first input tensor and second input tensor are the same. 
+  (ONNXCastLikeOp $val, $castTo, $saturate),
+  // Remove the castlike.
+  (replaceWithValue $val),
+  // Check that the first input tensor and second input tensor are the same
+  [(HaveSameElementType $val, $castTo)]>;
+
```
AlexandreEichenberger, line 449: I would add here (or probably more likely in Rewrite.cpp) a rule that transforms ALL cast like into casts; and thus this rule could be rewritten for the regular cast (if it does not already exists).

src/Builder/FrontendDialectTransformer.cpp
hamptonm1: Oh by accident! Yes sorry..ignore that :)

hamptonm1:
@AlexandreEichenberger So just clarify, in this code I did not remove all `castlike` ... I removed the ones that failed to be properly propagated which are any two input tensors of the same value. After testing with this code there are several castlike's still mentioned in the model but the two input tensor values are different and therefore work. So I just wrote code the way it made sense to me at the time, I honestly was not too sure of going about just using cast only and then adding a rewrite rule but open to trying. However, the code above I would need doing things the current way.

AlexandreEichenberger:
> @AlexandreEichenberger So just clarify, in this code I did not remove all `castlike` ... I removed the ones that failed to be properly propagated which are any two input tensors of the same value. After testing with this code there are several castlike's still mentioned in the model but the two input tensor values are different and therefore work. So I just wrote code the way it made sense to me at the time, I honestly was not too sure of going about just using cast only and then adding a rewrite rule but open to trying. However, the code above I would need doing things the current way.

Why not changing all of them? Since `castlike` is totally equivalent to `cast` in the cases that we are willing to consider (aka we choke on models for which the types are not known at compile time), I suggest to migrate all `catlike` to `cast`.

hamptonm1:
> > @AlexandreEichenberger So just clarify, in this code I did not remove all `castlike` ... I removed the ones that failed to be properly propagated which are any two input tensors of the same value. After testing with this code there are several castlike's still mentioned in the model but the two input tensor values are different and therefore work. So I just wrote code the way it made sense to me at the time, I honestly was not too sure of going about just using cast only and then adding a rewrite rule but open to trying. However, the code above I would need doing things the current way.
> 
> Why not changing all of them? Since `castlike` is totally equivalent to `cast` in the cases that we are willing to consider (aka we choke on models for which the types are not known at compile time), I suggest to migrate all `catlike` to `cast`.

I stated above why... I will copy and paste my answer below again:

`So I just wrote code the way it made sense to me at the time, I honestly was not too sure of going about just using cast only and then adding a rewrite rule but open to trying. However, the code above I would need doing things the current way.`

src/Dialect/ONNX/ONNXOps/Math/ElementwiseBroadcast.cpp
hamptonm1: I removed it

hamptonm1:
@chentong319 @AlexandreEichenberger I tried to make the updates this morning and the code compiled correctly. Let me know if this is what you are expecting. Thanks

src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -331,6 +331,41 @@ class BinaryOpBroadcastAxisPattern : public OpRewritePattern<OP_TYPE> {
...
+
+    // Replace
+    MultiDialectBuilder<OnnxBuilder> create(rewriter, loc);
+    Value res;
+    res = create.onnx.cast(input, outputType);
+    rewriter.replaceOp(castLikeOp, res);
+    return success();
+  };
+};
+
```
tungld, line 368: @hamptonm1 if you move this into Decompose pass, then we don't need its shape inference.

docs/SupportedONNXOps-cpu.md
```diff
@@ -37,7 +37,7 @@ Onnx-mlir currently supports ONNX operations targeting up to opset 19. Limitatio
 | **BitwiseXor** |18 - * | | |
 | **BlackmanWindow** |none | | | |
 | **Cast** |6 - 18 |Cast only between float and double types. Only ppc64le and MacOS platforms support float16. | |
-| **CastLike** |none | | | |
+| **CastLike** |19 - * |CastLike only between float and double types. Only ppc64le and MacOS platforms support float16. | |
```
AlexandreEichenberger, line 40: x86? Personally don't know

src/Dialect/ONNX/ONNXOps/Math/ElementwiseBroadcast.cpp
```diff
@@ -161,6 +161,20 @@ LogicalResult ONNXBitShiftOp::inferShapes(
   return inferShapeForBroadcastingOps<ONNXBitShiftOp>(*this);
 }
 
+//===----------------------------------------------------------------------===//
```
AlexandreEichenberger, line 164: Can you motivate in a sentence why "castLike" is seen as a broadcast ? There is only one input.

src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -331,6 +331,41 @@ class BinaryOpBroadcastAxisPattern : public OpRewritePattern<OP_TYPE> {
...
+    Value input = op->getOperand(0);
+
+    // The output type will be the same as the target_type or the second input
+    Type outputType = castLikeOp.getTargetType()
+                          .getType()
+                          .cast<ShapedType>()
+                          .getElementType();
+
+    // Replace
```
AlexandreEichenberger, line 360: great

docs/SupportedONNXOps-cpu.md
hamptonm1: @AlexandreEichenberger I copied the same thing that `Cast` had... look at line 39. So should be the same...tests are the same as well. So I decided to add the same comment :)

src/Dialect/ONNX/ONNXOps/Math/ElementwiseBroadcast.cpp
hamptonm1: Ohh because Castlike actually has two inputs not one.

src/Dialect/ONNX/Rewrite.cpp
hamptonm1: @tungld Ahh thanks for that!! Let me try it out :)

hamptonm1: @tungld That did not work unfortunately.... I still see the error even when I move it to decompose pass because I still specify ONNXCastLike:

```
ld: Undefined symbols:
  mlir::ONNXCastLikeOp::inferShapes(std::__1::function<void (mlir::Region&)>), referenced from:
      mlir::detail::ShapeInferenceOpInterfaceInterfaceTraits::Model<mlir::ONNXCastLikeOp>::inferShapes(mlir::detail::ShapeInferenceOpInterfaceInterfaceTraits::Concept const*, mlir::Operation*, std::__1::function<void (mlir::Region&)>) in libOMONNXOps.a[5](ONNXDialect.cpp.o)
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

src/Dialect/ONNX/ONNXOps/Math/ElementwiseBroadcast.cpp
hamptonm1: I added a sentence :)

AlexandreEichenberger: Can you just check with a small example that if the two inputs are incompatible in shape, we don't have issues?
```
castlike( 32x64xf32, 2xi32)
```

that it works well with the broadcast op? I just don't remember if the broadcast logic is not taking over something that we don't really want (i.e. a "common" size between the various inputs). For example, an `add` operation with the above inputs is illegal as we cannot broadcast the one to the other.

hamptonm1: Okay I will put it back unary... I am fine with that.

src/Dialect/ONNX/Rewrite.cpp
tungld: You should declare that this op will be illegal otherwise the rule is not applied. See https://github.com/onnx/onnx-mlir/blob/main/src/Transform/ONNX/Decompose.cpp#L949

hamptonm1: kk testing now

hamptonm1: @tungld Okay that worked but now the model does not work. I get the following error:

```
Onnx-mlir command: ./Debug/bin/onnx-mlir --O3 --EmitLib --onnx-op-stats TXT model.onnx -v
The ONNX model has 42970 elements in its initializers. This value would be close to and greater than the number of parameters in the model. Because there is no way to exactly count the number of parameters, this value can be used to have a rough idea of the number of parameters in the model.
loc("n1"): error: failed to legalize operation 'onnx.CastLike' that was explicitly marked illegal
```

src/Dialect/ONNX/ONNXOps/ShapeHelper.hpp
```diff
@@ -352,8 +352,8 @@ using ONNXAtanOpShapeHelper = ONNXUnaryOpShapeHelper;
 using ONNXAtanhOpShapeHelper = ONNXUnaryOpShapeHelper;
 using ONNXBernoulliOpShapeHelper = ONNXUnaryOpShapeHelper;
 using ONNXBitwiseNotOpShapeHelper = ONNXUnaryOpShapeHelper;
-using ONNXCastLikeOpShapeHelper = ONNXUnaryOpShapeHelper;
 using ONNXCastOpShapeHelper = ONNXUnaryOpShapeHelper;
+using ONNXCastLikeOpShapeHelper = ONNXUnaryOpShapeHelper;
```
AlexandreEichenberger, line 356: Thanks, it feel safer to me to have it as a unary. 
Nit: alphabetically listed.

AlexandreEichenberger (approved):
LGTM, thanks for all the work. 

Please also get the nod from @tungld since he had a few comments.

src/Dialect/ONNX/Rewrite.td
```diff
@@ -595,8 +599,9 @@ def RemoveIdentityReshapePattern1:  Pat<
   (ONNXReshapeOp $val, $shape, $az),
   // Remove the reshape.
   (replaceWithValue $val),
-  // Check that val has the specified shape.
-  [(HasSpecifiedConstantShape $val, $shape)]>;
+  // Check if shape has no value and that val has the specified shape.
+  [(IsNoneValue:$shape), (HasSpecifiedConstantShape $val, $shape)]
```
tungld, line 603: `HasSpecifiedConstantShape` will never run if `shape` is not None (`IsNoneValue:$shape` returns false), so it's redundant, please remove it. Also `RemoveIdentityReshapePattern2` covered the situation of `HasSpecifiedConstantShape`.

tungld: Please add a lit test for the case where `shape` is None.

src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -1516,6 +1551,7 @@ void ONNXAndOp::getCanonicalizationPatterns(
 void ONNXCastOp::getCanonicalizationPatterns(
     RewritePatternSet &result, MLIRContext *context) {
   result.insert<CastEliminationPattern>(context);
+  result.insert<PropagateCastLikeToCastPattern>(context);
```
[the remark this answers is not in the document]
tungld: Also consider rename `PropagateCastLikeToCastPattern` to `ReplaceCastLikeByCastPattern`. `Propagate` is confusing.

tungld (changes requested):

src/Dialect/ONNX/Rewrite.cpp
```diff
@@ -390,7 +425,7 @@ class PropagateScalarConstantExpandPattern : public OpRewritePattern<OP_TYPE> {
 
     rewriter.replaceOp(op, {res});
     return success();
-  }
+  };
```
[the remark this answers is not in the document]
hamptonm1: It is not... I can remove it.

src/Dialect/ONNX/Rewrite.cpp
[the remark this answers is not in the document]
hamptonm1: Okay I moved it into ONNXCastLike and renamed the pattern.

src/Dialect/ONNX/Rewrite.td
hamptonm1: @tungld We can not use the value `None` in the reshape lit test... I get the following error:

```
within split at /Users/meganhampton/zDLC/onnx-mlir/test/mlir/onnx/onnx_canonicalization.mlir:203 offset :7:8: error: 'onnx.Reshape' op operand #1 must be tensor of 64-bit signless integer values, but got 'none'
  %1 = "onnx.Reshape"(%arg0, %cst) : (tensor<10x11x12x13xf32>, none) -> tensor<10x11x12x13xf32>
```

since onnx.Reshape does not have an optional value for reshape it automatically fails. Maybe you have another idea but I did not include the lit test for that reason :)

tungld: Interesting! So by its definition onnx.Reshape does not allow None, so somehow onnx.Reshape with None existed in generated IR and passed its verification.

Do you know whether onnx.Reshape with None is in the original onnx model or generated by our compiler?

tungld (approved):
LGTM.

PR #2697, opened 2024-02-02
[NNPA] Splitting the second input in zdnn matmul 

tungld:
This patch supports splitting the second input B in zdnn matmul. So, for `A(BS, M, N) * B(N, P) + C(P)`, we split both M and/or P.
This patch supports reuse of the original buffer in some cases:
- When `BS = 1`, `N=64`, splitting and merging A need not copy data and the original ztensor is reused.
- When `BS=1`, splitting and merging B need not copy data and the original ztensor is reused.

### DIFF
--- a/src/Accelerators/NNPA/Runtime/zDNNExtension/zDNNExtension.c
+++ b/src/Accelerators/NNPA/Runtime/zDNNExtension/zDNNExtension.c
@@ -147,10 +147,13 @@ static void freeZTensorChunk(zdnn_ztensor *t) {
 
 static void copyZTensorChunk(
     const SplitInfo *splitInfo, uint32_t chunkID, bool fromChunk) {
-  // Only support the second innermost axis in the CPU tensor at this moment.
-  // axis = 2 in the CPU tensor corresponds to dim3 in zTensor.
-  if (splitInfo->axis != 2) {
-    printf("Only support the second innermost dimension at this moment.");
+  // Only support the first and second innermost axis in the CPU tensor at this
+  // moment.
+  // axis = 2 in the CPU tensor corresponds to dim2 in zTensor.
+  // axis = 3 in the CPU tensor corresponds to dim1 in zTensor.
+  if (!(splitInfo->axis == 2 || splitInfo->axis == 3)) {
+    printf("Only support the first and second innermost dimension at this "
+           "moment.");
     return;
   }
 
@@ -170,13 +173,17 @@ static void copyZTensorChunk(
   assert(dst && "Destination buffer is NULL");
 
   // Shape information.
+  // (e4, e3, e2, e1) -> (d6=e4, d5=e1/64, d4=e3, d3=e2/32, d2=32, d1=64)
   zTensorShape origShape;
   getZTensorShape(splitInfo->origZTensor, &origShape);
   zTensorShape chunkShape;
   getZTensorShape(chunk->ztensor, &chunkShape);
   assert(origShape.dim6 == chunkShape.dim6);
-  assert(origShape.dim5 == chunkShape.dim5);
+  if (splitInfo->axis != 3) // e1
+    assert(origShape.dim5 == chunkShape.dim5);
   assert(origShape.dim4 == chunkShape.dim4);
+  if (splitInfo->axis != 2) // e2
+    assert(origShape.dim3 == chunkShape.dim3);
   assert(origShape.dim2 == chunkShape.dim2);
   assert(origShape.dim1 == chunkShape.dim1);
   // Ensure that each element is 2 bytes.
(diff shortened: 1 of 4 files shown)


tungld:

tungld:
@AlexandreEichenberger this is ready to review.

src/Accelerators/NNPA/Runtime/zDNNExtension/zDNNExtension.c
```diff
@@ -103,42 +103,145 @@ static zdnn_status allocZTensorChunk(
...
+    zdnn_init_ztensor(preTransDesc, transDesc, chunkZTensor);
+    // Set a buffer size for the chunk.
+    chunkZTensor->buffer_size = zdnn_getsize_ztensor(transDesc);
+    // Set a buffer for the chunk.
+    uint64_t reuseBufferOffset;
+    if (chunkID == splitInfo->numOfChunks - 1) {
+      assert((origZTensor->buffer_size - chunkZTensor->buffer_size) %
+                 (splitInfo->numOfChunks - 1) ==
+             0);
+      reuseBufferOffset =
```
AlexandreEichenberger, line 216: would't the last chunk still start at a `0 mod 4k` page? Maybe you can provide a comment that provide insight.

src/Accelerators/NNPA/Runtime/zDNNExtension/zDNNExtension.c
```diff
@@ -103,42 +103,145 @@ static zdnn_status allocZTensorChunk(
...
+           "Chunk buffer is outside the original buffer");
+    status = ZDNN_OK;
+  } else {
+    // Init a zTensor with malloc.
+    status =
+        zdnn_init_ztensor_with_malloc(preTransDesc, transDesc, chunkZTensor);
+  }
+  chunkZTensor->is_transformed = true;
+
+  return status;
```
AlexandreEichenberger, line 239: It would be useful for parallelization if you could state at the beginning of this function that it only modifies values inside the chunkInfo data structure `splitInfo.chunk[chunkID]`, if true.

src/Accelerators/NNPA/Runtime/zDNNExtension/zDNNExtension.c
```diff
@@ -147,10 +250,13 @@ static void freeZTensorChunk(zdnn_ztensor *t) {
 
 static void copyZTensorChunk(
     const SplitInfo *splitInfo, uint32_t chunkID, bool fromChunk) {
-  // Only support the second innermost axis in the CPU tensor at this moment.
-  // axis = 2 in the CPU tensor corresponds to dim3 in zTensor.
-  if (splitInfo->axis != 2) {
-    printf("Only support the second innermost dimension at this moment.");
+  // Only support the first and second innermost axis in the CPU tensor at this
```
AlexandreEichenberger, line 253: Again, for OpenMP, would be convenient if you could define what is modified here... I think its only the private data in a single chunkID when copying to chunk, and when copying from chunk it copes into the shared "origTensor" but each buffer is guaranteed (if that is true) to write only into a distinct tile than any other chunks.

AlexandreEichenberger (commented):
First round of comments; many due to names mis match between the definitions used in zDNN and here. I think it helps the comprehension quite a bit if we find solid non-ambiguous naming first.

tungld:
@AlexandreEichenberger thank you for your comments! I updated the code according to your comments. It is ready for another round of review.

src/Accelerators/NNPA/Runtime/zDNNExtension/zDNNExtension.c
```diff
@@ -68,201 +68,378 @@ void zDNNExtensionInit() {
...
-  return;
 }
 
-static void copyZTensorChunkScalar(
-    const SplitInfo *splitInfo, uint32_t chunkID, bool fromChunk) {
-  // Only support the second innermost axis in the CPU tensor at this moment.
-  // axis = 2 in the CPU tensor corresponds to dim3 in zTensor.
-  if (splitInfo->axis != 2) {
-    printf("Only support the second innermost dimension at this moment.");
+static void copyDataForTileScalar(
```
AlexandreEichenberger, line 393: Do you mind adding a high level comment of the difference between `CopyDataForTile` and `copyDataForTileScalar`?

src/Accelerators/NNPA/Runtime/zDNNExtension/zDNNExtension.c
```diff
@@ -68,201 +68,378 @@ void zDNNExtensionInit() {
...
-  // Allocate one ztensor struct.
-  chunk->ztensor = malloc(sizeof(zdnn_ztensor));
-  if (!chunk->ztensor)
-    return ZDNN_ALLOCATION_FAILURE;
-  zdnn_ztensor *chunkZTensor = chunk->ztensor;
+  // Adjust the number of elements per tile for the last tile.
+  uint32_t numOfElemsPerTile = splitInfo->numOfElemsPerTile;
+  if (tileID == splitInfo->numOfTiles - 1) {
+    // The last tile.
+    numOfElemsPerTile = getUnmappedDim(fullZTensor, axis) -
```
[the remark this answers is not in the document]
tungld: We can use numerical tests for this purpose, but still need a NNPA machine to run the tests.

tungld:
@AlexandreEichenberger thank you for your comments! I updated the code according to your comments. I also expose functions to work with a specific tile, e.g. allocate, copy, free a single tile.
It is ready for another round of review.

src/Accelerators/NNPA/Runtime/zDNNExtension/zDNNExtension.c
```diff
@@ -68,201 +68,378 @@ void zDNNExtensionInit() {
...
+    break;
+  case E2:
+    mappedNumOfElemsPerTile =
+        CEIL(splitInfo->numOfElemsPerTile, AIU_STICKS_PER_PAGE);
+    break;
+  case E1:
+    mappedNumOfElemsPerTile =
+        CEIL(splitInfo->numOfElemsPerTile, AIU_2BYTE_CELLS_PER_STICK);
+    break;
+  default:
```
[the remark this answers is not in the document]
tungld: I used our own defined unreachable to avoid depending on llvm.

src/Accelerators/NNPA/Runtime/zDNNExtension/zDNNExtension.c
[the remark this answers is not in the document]
AlexandreEichenberger: Good idea. We may have to tweak the tests a bit; numerical where meant to test lots of combination of "very small" sizes to cover corner cases of register & cache tiling. We can reuse the infra to run larger test when targeting NNPA.

src/Accelerators/NNPA/Runtime/zDNNExtension/zDNNExtension.c
AlexandreEichenberger: Thanks, that is super useful, I was trying to understand the why, and now it makes perfect sense.

src/Accelerators/NNPA/Runtime/zDNNExtension/zDNNExtension.h
```diff
@@ -121,6 +121,17 @@ void zDNNExtensionInit();
 // Helper Functions
 // -----------------------------------------------------------------------------
 
+inline void omUnreachable() {
```
AlexandreEichenberger, line 124: Very good. The name prefix `om` made me think of 2 things: 
1) If you see this as useful also for our `cruntime`, then let's move it there. Note that we should be able to do that later too. 
2) Do we need a prefix for the entry points into this new library? We can argue that we do not, as it is only used by our compiler and is technically not exposed to the outside.

I don't have a firm opinion on either points.

AlexandreEichenberger (approved):
LGTM, thanks for this outstanding PR

src/Accelerators/NNPA/Runtime/zDNNExtension/zDNNExtension.h
tungld: Thanks for your comments! Both two things look reasonable to me. I see that cruntime is now using `assert` instead of `unreachable` and there is no a common place for shared functions. Another PR for that seems better.

PR #2698, opened 2024-02-02
Enhance cmake to fetch remote or local the dependencies.

cbalint13:
Hi folks,

This request enhances ```cmake``` options  for a better rel-eng / distro friendly purpose.

---

New options:

* ~~System library options: ```ONNX_MLIR_USE_SYSTEM_ONNX```, ```ONNX_MLIR_USE_SYSTEM_PYBIND11```~~
* Build control options: ```ONNX_MLIR_BUILD_BENCHMARKS```, ```ONNX_MLIR_BUILD_RAPIDCHECKS```

This allows to pick up local system libraries and also allows to build only some parts (no test/bench).

The original behaviour of ```cmake``` is kept.


cbalint13:

CMakeLists.txt
```diff
@@ -161,31 +165,48 @@ set(CMAKE_MESSAGE_LOG_LEVEL NOTICE)
 
 # Add third party subdirectories and define options appropriate to run their cmakes.
 set(pybind11_FIND_QUIETLY ON)
-add_subdirectory(third_party/onnx)
-add_subdirectory(third_party/pybind11)
-add_subdirectory(third_party/rapidcheck)
+
+if (ONNX_MLIR_USE_SYSTEM_ONNX)
+  find_package(ONNX REQUIRED)
```
gongsu832, line 170: onnx-mlir requires a specific version of onnx, which is the one included under third_party. So there needs some way to extract that version number and put it here.

cbalint13: @gongsu832 ,

* The ```find_package(ONNX REQUIRED)``` is filling the ```ONNIX_VERSION``` cmake var already.
* It is also [picked up and used here](https://github.com/onnx/onnx-mlir/pull/2698/files#diff-9c32b95937462105ee738c4d993a32c9669dfefb9b73742ca8021e2db41cdd53R89-R91) in the case when a local system library is found.

See a runtime sample:
```
$ rpm -q onnx-devel
onnx-devel-1.16.0-20240123.1.git3a39b0b8.fc40.x86_64

$ cmake  {..} -DONNX_MLIR_USE_SYSTEM_ONNX=ON {...}
{...}
-- Accelerators             : NONE
-- OpenMP support           : OFF
-- Found Git: /usr/bin/git (found version "2.43.0") 
-- Onnx version             : 1.16.0
{...}
```

* I did let it by default ```OFF``` these knobs (so default is to pick up ```third_party```), not enforced a specific version as ```REQUIRED```, one who want to use these is assuming to know what he is doing.

gongsu832: onnx-mlir currently does NOT support onnx 1.16.0. It only works with a specific version of onnx, currently 1.15.0. That's why we always install onnx from third_party/onnx. If you want to use a system installed onnx, it must match the version found in third_party/onnx.

cbalint13: > @gongsu832 ,
> 
> * The `find_package(ONNX REQUIRED)` is filling the `ONNIX_VERSION` cmake var already.
> * It is also [picked up and used here](https://github.com/onnx/onnx-mlir/pull/2698/files#diff-9c32b95937462105ee738c4d993a32c9669dfefb9b73742ca8021e2db41cdd53R89-R91) in the case when a local system library is found.
> 
> See a runtime sample:
> 
> ```
> $ rpm -q onnx-devel
> onnx-devel-1.16.0-20240123.1.git3a39b0b8.fc40.x86_64
> 
> $ cmake  {..} -DONNX_MLIR_USE_SYSTEM_ONNX=ON {...}
> {...}
> -- Accelerators             : NONE
> -- OpenMP support           : OFF
> -- Found Git: /usr/bin/git (found version "2.43.0") 
> -- Onnx version             : 1.16.0
> {...}
> ```
> 
> * I did let it by default `OFF` these knobs (so default is to pick up `third_party`), not enforced a specific version as `REQUIRED`, one who want to use these is assuming to know what he is doing.

@gongsu832 ,

To better reflect the origin of ONNX I [added explicit message](https://github.com/onnx/onnx-mlir/pull/2698/files#diff-9c32b95937462105ee738c4d993a32c9669dfefb9b73742ca8021e2db41cdd53R94-R98) for the case of system found library.
```
$ cmake  {..} -DONNX_MLIR_USE_SYSTEM_ONNX=ON {...}
{...}
-- Onnx version             : 1.16.0 (system)

$ cmake  {..} -DONNX_MLIR_USE_SYSTEM_ONNX=OFF {...}
-- Onnx version             : 1.16.0
```

cbalint13: @gongsu832 

> onnx-mlir currently does NOT support onnx 1.16.0. It only works with a specific version of onnx, currently 1.15.0. That's why we always install onnx from third_party/onnx. If you want to use a system installed onnx, it must match the version found in third_party/onnx.

Ok, it is [enforced now](https://github.com/onnx/onnx-mlir/pull/2698/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR170) to specific version.
```
$ cmake  {..} -DONNX_MLIR_USE_SYSTEM_ONNX=ON {...}
{...}

CMake Error at CMakeLists.txt:173 (find_package):
  Could not find a configuration file for package "ONNX" that exactly matches
  requested version "1.15.0".

  The following configuration files were considered but not accepted:

    /usr/lib64/cmake/ONNX/ONNXConfig.cmake, version: 1.16.0
    /lib64/cmake/ONNX/ONNXConfig.cmake, version: 1.16.0

```
However the idea of this permissive know was that once user enable it he *really knows* what is doing.

gongsu832: @cbalint13 As I said above, onnx-mlir only works with a specific version of onnx, currently 1.15.0. In fact, I think the presence of third_party/onnx source repo is mandatory (@chentong319 can confirm yes or no) in order to generate the operator translation table. That's precisely why we always install onnx from third_party/onnx instead of having to deal with incorrect version of system installed onnx.

cbalint13: @gongsu832 , @chentong319 

> As I said above, onnx-mlir only works with a specific version of onnx, currently 1.15.0. 

* Let's have it enforced, in case of system lib too, just like it is now: 
https://github.com/onnx/onnx-mlir/pull/2698/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR170

> In fact, I think the presence of third_party/onnx source repo is mandatory

* Works fine *without*, I can confirm from my side, one needed workaround for this is here:
 https://github.com/onnx/onnx-mlir/pull/2698/files#diff-9c32b95937462105ee738c4d993a32c9669dfefb9b73742ca8021e2db41cdd53R89-R91

gongsu832: When you say it "works fine _without_", I assume you mean that cmake runs fine. But it takes more to upgrade onnx to a new version for onnx-mlir. For example, you also need to run `make OMONNXOpsIncTranslation` from the `build/utils` directory to generate the operator translation table. I think running `make OMONNXOpsIncTranslation` requires the presence of third_party/onnx source repo. Of course, once the table has been generated, we may not need the third_party/onnx source repo anymore.

cbalint13: > When you say it "works fine _without_", I assume you mean that cmake runs fine. But it takes more to upgrade onnx to a new version for onnx-mlir. 

* Works fine *without* the ```/third_party/onnx``` (absent at all), having ```1.15.0 EXACT REQUIRED``` system version.
* By *works* I mean: configures + compiles + tests

>For example, you also need to run `make OMONNXOpsIncTranslation` from the `build/utils` directory to >generate the operator translation table. I think running `make OMONNXOpsIncTranslation` requires the >presence of third_party/onnx source repo. 

* All include headers and other ONNX things are well picked up in the context of system libraries, find_package(ONNX ...) provides all required paths to the onnx requirements.

CMakeLists.txt
```diff
@@ -161,31 +165,48 @@ set(CMAKE_MESSAGE_LOG_LEVEL NOTICE)
 
 # Add third party subdirectories and define options appropriate to run their cmakes.
 set(pybind11_FIND_QUIETLY ON)
-add_subdirectory(third_party/onnx)
-add_subdirectory(third_party/pybind11)
-add_subdirectory(third_party/rapidcheck)
+
+if (ONNX_MLIR_USE_SYSTEM_ONNX)
+  find_package(ONNX 1.15.0 EXACT REQUIRED)
```
gongsu832, line 170: I don't know why my comment disappeared so here it is again.

Instead of hardcoding 1.15.0, why not use
```
file(READ "${ONNX_MLIR_SRC_ROOT}/third_party/onnx/VERSION_NUMBER" ONNX_VERSION)
find_package(ONNX ${ONNX_VERSION} EXACT REQUIRED)
```

cbalint13:
@gongsu832 ,

I added requested changes related to concerns here.

The new changes:

 * Introduce [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html) cmake's feature requiring a minimum 3.24 version
 * Centralize all required versions in a single clear cmake entry [here](https://github.com/onnx/onnx-mlir/pull/2698/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR18-R22) , either tag or hash strings.
 * The ```FetchContent``` will **fetch/compile/use** from remote **the required versions** from remote automatically
 * If ```FetchContent``` finds on local system **the EXACT equivalent** will use that instead
 * Display a final stat telling if a "(remote)" or "(local)" was used to be clear about the origins
 * The ```third_party/``` folder can go completely away

 
I can't think for now for a more centralized thing, also if a local library exact match is found than pickup and used.

CMakeLists.txt
```diff
@@ -1,18 +1,26 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Match the minimum required version of LLVM and MLIR
-cmake_minimum_required(VERSION 3.20.0)
+cmake_minimum_required(VERSION 3.24.0)
```
gongsu832, line 4: We use ubuntu:jammy for our dev docker image. Its cmake version is up to 3.22.1. So we cannot require more than that.

cbalint13: > We use ubuntu:jammy for our dev docker image. Its cmake version is up to 3.22.1. So we cannot require more than that.

See, I leave this variation of proposal for a possible future betterment.

Thanks.

cbalint13: > > We use ubuntu:jammy for our dev docker image. Its cmake version is up to 3.22.1. So we cannot require more than that.

I was mistaken , `FetchContent` [requires](https://cmake.org/cmake/help/latest/module/FetchContent.html) minimum `3.11` not `3.24` so it fits in the actual `3.22` requirement.
I unchanged  the version requirement from proposal with a new push.

gongsu832:
> @gongsu832 ,
> 
> I added requested changes related to concerns here.
> 
> The new changes:
> 
>     * Introduce [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html) cmake's feature requiring a minimum 3.24 version
> 
>     * Centralize all required versions in a single clear cmake entry [here](https://github.com/onnx/onnx-mlir/pull/2698/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR18-R22) , either tag or hash strings.
> 
>     * The `FetchContent` will **fetch/compile/use** from remote **the required versions** from remote automatically
> 
>     * If `FetchContent` finds on local system **the EXACT equivalent** will use that instead
> 
>     * Display at the final status telling if a "(remote)" or "(local)" was used to be clear about the origins
> 
>     * The `third_party/` folder can go completely away with this
> 
> 
> I can't think for now for a more centralized thing, also if a local library exact match is found then it is picked up instead.

It's not clear to me why we introduce FetchContent unless we plan to discard third_party, which we don't. We keep submodules in third_party because occasionally we might need small changes in the submodules until the changes are upstreamed. For example, onnx 1.15.0 does not build with protobuf 25.2 out of the box. With some minor changes it will. So if onnx-mlir wants to upgrade to protobuf 25.2 before necessary changes for onnx are upstreamed, we can ship a customized onnx 1.15.0 under third_party.

cbalint13:
@gongsu832 

> It's not clear to me why we introduce FetchContent unless we plan to discard third_party, which we don't.

* The (updated) proposal came after the issue with the ONNX_VERSION being a separate file, handled separetly aside from newly introduce find_package() option with already populates such vars. Thus, to complain to all issues: separatae subfolders, separate version lookup mechanism for onnx and a new separate ```find_package()``` the ```FetchContent``` offers a complete and unified solution to all these issues.

> We keep submodules in third_party because occasionally we might need small changes in the submodules until the changes are upstreamed. For example, onnx 1.15.0 does not build with protobuf 25.2 out of the box. With some minor changes it will. So if onnx-mlir wants to upgrade to protobuf 25.2 before necessary changes for onnx are upstreamed, we can ship a customized onnx 1.15.0 under third_party.

* There is a fine grain control of steps within ```FetchContent```, i.e for [applying a patch](https://cmake.org/cmake/help/latest/module/ExternalProject.html#patch-step-options) in the preparation steps. This makes highly flexible and centralized to a single point of cmake entry. It is very similar concept to bazel's repo rules.

Thanks for your time.

PR #2699, opened 2024-02-02
Parallel tuning

AlexandreEichenberger:
- Parallel for Layout transform is now exploited only to the outer-level (before it was all levels, separately, which was slow).
- Added parallel for the onnx.Gather operation
- Ensured that parallel is only exploited when there is enough parallelism (e.g. concat had sometime short runs of 1, and was parallelized)
- replaced debug for parallel by reporting parallel (activated with the `-opt-report=Parallel`)


src/Conversion/ONNXToKrnl/Additional/LayoutTransform.cpp
```diff
@@ -83,7 +83,7 @@ struct ONNXLayoutTransformOpLowering
     // Insert loop over all inputs.
     ValueRange loopDef = create.krnl.defineLoops(rank);
     if (enableParallel) {
-      create.krnl.parallel(loopDef);
+      create.krnl.parallel(loopDef[0]);
```
tungld, line 86: Thanks for fixing this!

tungld (approved):
LGTM.

PR #2702, opened 2024-02-06
Add omp parallel for in MatMul lib (NNPA)

chentong319:
I first build llvm-project with targets of clang and omp. Then use the built clang to configure onnx-mlir, so that the omp pragma can be interpreted. 

To compile a model, I added `-lgomp` to avoid undefined symbol at runtime. `--parallel` can not solve this issue. Will look into a more elegant solution.

### DIFF
--- a/src/Accelerators/NNPA/Runtime/zDNNExtension/MatMul.c
+++ b/src/Accelerators/NNPA/Runtime/zDNNExtension/MatMul.c
@@ -17,6 +17,7 @@
 #define _OPEN_THREADS
 #endif
 #include <pthread.h>
+#include <sched.h>
 
 #include <assert.h>
 #include <math.h>
@@ -96,7 +97,9 @@ static zdnn_status zdnn_matmul_op_common(const zdnn_ztensor *inputA,
   // Call zdnn_matmul_op on each chunk.
   if (OMZTensorSplitDebug)
     start_time = clock();
+#pragma omp parallel for
   for (uint32_t i = 0; i < splitInfoA.numOfChunks; ++i) {
+    //printf("====cpu id %d=======\n", sched_getcpu());
     zdnn_ztensor *zaTensor = (splitInfoA.chunks + i)->ztensor;
     zdnn_ztensor *zyTensor = (splitInfoY.chunks + i)->ztensor;
     zdnn_status status = call_zdnn_matmul_op(
--- a/src/Accelerators/NNPA/Runtime/CMakeLists.txt
+++ b/src/Accelerators/NNPA/Runtime/CMakeLists.txt
@@ -21,6 +21,6 @@ set_target_properties(RuntimeNNPA
   PROPERTIES
   LANGUAGE C
   POSITION_INDEPENDENT_CODE TRUE
-  COMPILE_OPTIONS -O3
+  COMPILE_OPTIONS "-O3;-fopenmp"
   )
 


tungld:
> Then use the built clang to configure onnx-mlir, so that the omp pragma can be interpreted.

So you were able to build onnx-mlir with clang and omp? I had to build onnx-mlir with c++ instead.

gongsu832:
> > Then use the built clang to configure onnx-mlir, so that the omp pragma can be interpreted.
> 
> So you were able to build onnx-mlir with clang and omp? I had to build onnx-mlir with gcc instead.

omp on s390x compiles with gcc in general. With clang, you need at least clang-18. No distro comes with clang-18 yet so you need to manually install it yourself.

gongsu832:
> > omp on s390x compiles with gcc in general.
> 
> @gongsu832 do you think that we may start building `clang` when building `llvm-project` and then using it to compile `onnx-mlir`? As we start having parallelism, it becomes increasingly important to build onnx-mlir and the generated inference graph with the same compiler (esp not gcc and clang).

Switching to a new level of compiler isn't as simple as just building and installing the new compiler binary. There are typically accompanying libraries that must also be built and installed, which might conflict with what are already installed on the base system. So it's generally not advisable to build a compiler yourself unless you are willing to go through the trouble of packaging all necessary binaries and libraries and resolving potential conflicts with the base system (which is basically what the distros do when they release a new version of the compiler package).

src/Accelerators/NNPA/Runtime/zDNNExtension/MatMul.c
```diff
@@ -77,6 +92,11 @@ static zdnn_status zdnn_matmul_op_common(const zdnn_ztensor *inputA,
 
   // Call zdnn_matmul_op on each tile.
   // Iterate over the tiles along the first dim of A.
+
+  // Commented out omp_set_nested because current build cannot guarantee that
+  // OpenMP is supported. Use "OMP_SET_NESTED=TRUE" at runtime instead.
+  // omp_set_nested(true);
+#pragma omp parallel for
```
tungld, line 99: Have you tried to do `copyData, call zdnn op, copyData` in one single iteration? It would put data close to the current NNPA accelerator.

We have two kinds of parallelism: multiple CPUs and multiple NNPAs, and each one would require different number of threads. For example, we have more the number of CPUs than that of NNPAs. Does it make sense to define the number of NNPAs we'd like to use?

src/Accelerators/NNPA/Runtime/zDNNExtension/MatMul.c
```diff
@@ -86,13 +106,17 @@ static zdnn_status zdnn_matmul_op_common(const zdnn_ztensor *inputA,
...
       zdnn_ztensor *zcTensor = splitInfoC.tiles + j;
       zdnn_ztensor *zybTensor = splitInfoYB.tiles + j;
       zdnn_status status = call_zdnn_matmul_op(
           zaTensor, zbTensor, zcTensor, opType, zybTensor, isBcast);
       assert(status == ZDNN_OK);
+      if (OMZTensorSplitDebug) {
+        printf("thread [%u, %u] is on cpu %d\n", i, j, sched_getcpu());
+      }
     }
     copyData(&splitInfoYB, TILES_TO_FULL);
```
tungld, line 121: Do you plan to parallelize the data copy from full to tiles and tiles to full?

tungld:
> I can get speedup close to 2 for 2 tiles for each A and B (ideal speedup should be close to 4).

I see you measure end-to-end time that includes data copy between the full tensor and tiles. Do you measure how long the data copy takes?

src/Accelerators/NNPA/Runtime/zDNNExtension/MatMul.c
```diff
@@ -41,14 +42,24 @@ static inline zdnn_status call_zdnn_matmul_op(const zdnn_ztensor *inputA,
       inputA, inputB, inputC, (zdnn_matmul_ops)opType, output);
 }
 
+static float get_elapse(
```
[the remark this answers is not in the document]
chentong319: Modified.

src/Accelerators/NNPA/Runtime/zDNNExtension/MatMul.c
```diff
@@ -41,14 +42,24 @@ static inline zdnn_status call_zdnn_matmul_op(const zdnn_ztensor *inputA,
...
+    const struct timeval start_t, const struct timeval end_t) {
+  return (((end_t.tv_sec * 1000000.) + end_t.tv_usec) -
+             ((start_t.tv_sec * 1000000) + start_t.tv_usec)) /
+         1000;
+}
+
+// It is supposed that sched.h should have the declaration of sched_getcpu.
+// No problem when I compiled a standalone test case.
```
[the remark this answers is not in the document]
chentong319: Modified

src/Accelerators/NNPA/Runtime/zDNNExtension/MatMul.c
chentong319: This PR is just the initial enablement of omp in runtime library. We need to design a global control on how to use the thread resource.
One thing I tried is the different way of binding. According to document, we can use OMP_PLACES to describe the numa structure, and then use OMP_PROC_BIND to control the binding. For example, for parallel loop for CPU work, we may use `close` binding so that the threads can share cache. For zAIU threads we should use `spread` binding so that zAIU on different chip can be used. But my experiments did not go as I expected. Need further investigation.

chentong319:
> > I can get speedup close to 2 for 2 tiles for each A and B (ideal speedup should be close to 4).
> 
> I see you measure end-to-end time that includes data copy between the full tensor and tiles. Do you measure how long the data copy takes?

I added the measurement for the mm loop, which contains some data copy.

src/Accelerators/NNPA/Runtime/zDNNExtension/zDNNExtension.h
```diff
@@ -254,6 +257,11 @@ zdnn_status zdnn_softmax_ext(const zdnn_ztensor *input, void *save_area,
     zdnn_softmax_act act_func, zdnn_ztensor *output);
 zdnn_status zdnn_tanh_ext(const zdnn_ztensor *input, zdnn_ztensor *output);
 
+// -----------------------------------------------------------------------------
+// Misc Utility Functions
+// -----------------------------------------------------------------------------
+float GetElapseTime(const struct timeval start_t, const struct timeval end_t);
+
```
tungld, line 264: Thanks! Really appreciate.

tungld (approved):
LGTM.

PR #2703, opened 2024-02-06
Update docs/UsingPyRuntime.md to catch up PR #2629 (Support string type inputs from Python application code).

negiyas:
Update docs/UsingPyRuntime.md to catch up PR #2629 (Support string type inputs from Python application code).
Only document is updated.

image


AlexandreEichenberger (approved):
LGTM, thanks for the slide in the PR description.

@cjvolzka Does this description helps you? Need more info?

PR #2704, opened 2024-02-06
Move constant normalization rules into Decompose pass

tungld:
Move constant normalization rules from Canonicalization pass into Decompose pass, so that we always have ConstantOp with ElementAttrs in the following passes.


hamptonm1 (approved):
Yes this works perfect! I tested it with my code. Thanks so much @tungld

PR #2706, opened 2024-02-07
Do not save LLVMIR to .ll file if not requested

tungld:
There is a bug in saving LLVMIR to `.ll`. It was always saving to a .ll file (which is slow) then deleting the file if not requested from users.
This patch adds a `if` statement so that the saving to .ll is done only when users request.


imaihal (approved):
LGTM!

PR #2709, opened 2024-02-08
Optimization for Data conversions FP32 <-> DLF16

AlexandreEichenberger:
Current support for conversions associated with NNPA does not support vectors of arbitrary lengths; they must be of length 8. But the code would benefit from further unrolling. So this PR introduces explicit unrolling in the `NNPA/Conversion/ZHighToZLow` conversion operators.

I tested various unroll factors on a `1024x1024` tensor, and best results were achieved with an unroll of 8 SIMD vectors, thus processing 8x8=64 values per unrolled SIMD loop iteration.

Speedup are 1.2x for the `DCF16ToFL32` and 1.4x for the `FL32ToDCF16`.

I wanted to also optimize the LayoutTransform, as it is relatively slow; but it is difficult to do so for the specific pattern seen for the NNPA in this generic LayoutTransform. So I am planning to do a lowering of a unified Conversion and Layout transform from the High dialect where I can make explicit assumptions about the actual layouts. This will come in a forthcoming PR.


tungld (approved):
LGTM.

tungld:
> So I am planning to do a lowering of a unified Conversion and Layout transform from the High dialect where I can make explicit assumptions about the actual layouts.

Same thought. We can have say `zhigh.LayoutTransform` with supported layout and optimize it.

PR #2710, opened 2024-02-09
Replace CastLike by Cast in Decompose pass

tungld:
Replace CastLike by Cast in Decompose pass so that there is no need of shape inference and lowering for CastLike.


src/Dialect/ONNX/ONNXUnsupportedOps.hpp
```diff
@@ -76,3 +64,17 @@ UNSUPPORTED_OPS(ONNXTreeEnsembleClassifierOp)
 UNSUPPORTED_OPS(ONNXTreeEnsembleRegressorOp)
 UNSUPPORTED_OPS(ONNXUpsampleV7Op)
 UNSUPPORTED_OPS(ONNXZipMapOp)
+// These ops were decomposed into supported ops during Decompose pass.
```
AlexandreEichenberger, line 67: That is a great idea to separate the two lists. Do you mind just putting a bit more comments for both? Say

```
// Set of ONNX Ops that we currently do not support, listed alphabetically. Please remove from list when support is added.
...

// Set of ONNX Ops that are supported indirectly as every instances are converted/decomposed in one or more supported ONNX ops. Listed alphabetically. 
#define CONVERTED_TO_SUPPORTED_OPS(_x) UNSUPPORTED_OPS(_x)
#CONVERTED_TO_SUPPORTED_OPS(ONNXCastLikeOp)

...```

**Review by AlexandreEichenberger — APPROVED** [review]
LGTM, left you a comment to further clarify what it means to be "unsupported"

File: src/Dialect/ONNX/ONNXUnsupportedOps.hpp [hunk 2]
```diff
@@ -76,3 +64,17 @@ UNSUPPORTED_OPS(ONNXTreeEnsembleClassifierOp)
...
+UNSUPPORTED_OPS(ONNXCastLikeOp)
```
hamptonm1, line 68: Since `CastLikeOp` is unsupported do we still need to have the backend tests I added? Also do we need the version specified at `utils/gen_onnx_mlir.py` for `CastLikeOp`?

hamptonm1 (commented):
@tungld I have two questions

src/Dialect/ONNX/ONNXUnsupportedOps.hpp
tungld: @hamptonm1 `unsupported` is quite confusing here so I will reorganize this as Alex suggested to make it clearer. 

CastLikeOp is marked `Unsupported` meaning that there is no handling of shape inference/lowering/... inside the compiler because it was rewritten into a "supported" op (i.e. CastOp).

tungld: Backend tests are still needed. We can run a model that contains CastLikeOp. Just the way we support it is via CastOp instead of directly lowering it to krnl/llvm.

src/Dialect/ONNX/ONNXUnsupportedOps.hpp
tungld: Thanks! I changed this. It looks better now.

src/Dialect/ONNX/ONNXUnsupportedOps.hpp
hamptonm1: Understood it was my assumption that because we are converting CastLikeOp to CastOp that we will no longer need the backend tests. Especially because the backend tests I just added and they are exactly the same for CastOp.

hamptonm1 (approved):
LGTM

PR #2712, opened 2024-02-13
Constant Propagation for Mod

hamptonm1:
Add constant propagation for modulo and I updated the backend tests as well as included two new lit tests for mod.


hamptonm1:
@tungld So I noticed the backend tests are failing once I added the constant propagation.... I commented them out for now. I noticed in macOS all of the tests passed but in Linux it fails which is very weird.

hamptonm1:
@tungld I was able to figure out why the backend tests were failing.... I had to add several logic for the constant propagation of mod to handle the case when 1) we have negative int values 2) positive int values 3) when the values are floats 4) when the values are multiple of each other. I added code which successfully handles each scenario. I also added new lit tests to confirm. Please let me know your thoughts.

Also I handled `fmod` in `ConstProp.cpp`  by checking if the value is a float. This way was tested and works as expected.

src/Transform/ONNX/ConstProp.cpp
```diff
@@ -217,6 +218,33 @@ struct ElementWiseBinaryOpImpl<ONNXMaxOp, T> {
...
 };
 
+template <typename T>
+struct ElementWiseBinaryOpImpl<ONNXModOp, T, EnableNotBool<T>> {
+  static T eval(T lhs, T rhs) {
+    // This is the original calculation for mod
+    int mod = lhs - floor(lhs / rhs) * rhs;
+
+    // Check if the values are floats
+    if (lhs != floor(lhs) && rhs != floor(rhs)) {
```
tungld, line 228: Since this is a struct template on type T, it is better to have two instances:
```
template <>
struct ElementWiseBinaryOpImpl<ONNXModOp, int64, EnableNotBool<int64>>
```

and 

```
template <>
struct ElementWiseBinaryOpImpl<ONNXModOp, double, EnableNotBool<double>>
```

src/Transform/ONNX/ConstProp.cpp
```diff
@@ -217,6 +218,33 @@ struct ElementWiseBinaryOpImpl<ONNXMaxOp, T> {
...
+    // Handle the case when one of the int values are negative
+    if (lhs < 0 || rhs < 0) {
+      // Check if lhs is a multiple of rhs
+      if (remainder(lhs, rhs) == 0) {
+        return mod;
+      } else {
+        return mod + rhs;
+      }
+      // Both int values are positive, so we can calculate as normal
+    } else {
```
tungld, line 242: nit: just `return mod` without `else`.

hamptonm1: Okay fixed!

src/Transform/ONNX/ConstProp.cpp
hamptonm1: I updated the code to add two struct templates with the int and double data types

src/Transform/ONNX/ConstProp.cpp
```diff
@@ -217,6 +218,33 @@ struct ElementWiseBinaryOpImpl<ONNXMaxOp, T> {
...
+template <typename T>
+struct ElementWiseBinaryOpImpl<ONNXModOp, T, EnableNotBool<T>> {
+  static T eval(T lhs, T rhs) {
+    // This is the original calculation for mod
+    int mod = lhs - floor(lhs / rhs) * rhs;
+
+    // Check if the values are floats
+    if (lhs != floor(lhs) && rhs != floor(rhs)) {
+      // Rounding to match the results of the backend tests
+      return (std::floor(fmod(lhs, rhs) * 1000000000) / 1000000000);
```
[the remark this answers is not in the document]
hamptonm1: As we discussed `fmod == 1` forces the use of `std::fmod` which essentially calculates the mod for floats only and `fmod == 0` uses the original mod calculation which only takes integers. From my testing if you try to use floats with the original calculation of mod, we will receive an incorrect value. So I will add code to the verifier for mod to ensure the attribute is set to the correct value when doing constant propagation. Thanks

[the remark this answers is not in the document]
hamptonm1: @tungld After checking you actually wrote code for the verifier already

```
static LogicalResult verify(ONNXModOp op) {
  Type elementType;
  if (op.A().getType().isa<ShapedType>())
    elementType = op.A().getType().cast<ShapedType>().getElementType();
  else
    return op.emitError("Input type must be TensorType or MemRefType");

  // Verify that when the input type is floating point, then `fmod` attribute
  // must be set to 1.
  if (elementType.isa<FloatType>() && (op.fmod() != 1))
    return op.emitError("fmod must be 1 when the input type is floating point");

  return success();
}
```

Here is the PR you did about 3 years ago: https://github.com/onnx/onnx-mlir/commit/7ec687ef73cbdba51c92d7f3830a0c72fff23c72#diff-55c7373ec478b68740f0af3b5fdfc4fdaf9cfc696f302f10ff0fc48f687a5493R3674


It does cover our basis but if you think we need to update the verifier, please let me know. Thank you!

src/Transform/ONNX/ConstProp.cpp
```diff
@@ -217,6 +217,34 @@ struct ElementWiseBinaryOpImpl<ONNXMaxOp, T> {
   static T eval(T lhs, T rhs) { return std::max<T>(lhs, rhs); }
 };
 
+template <>
+struct ElementWiseBinaryOpImpl<ONNXModOp, int64_t, EnableNotBool<int64_t>> {
+  static int64_t eval(int64_t lhs, int64_t rhs) {
+    // The original calculation for mod
```
tungld, line 223: I guess the whole computation here can be shorten to:
```
int64_t rem = remainder(lhs, rhs);
if ((rem != 0) && ((lhs < 0) ^ (rhs < 0)))
  return (rem + rhs);
return rem;
```

tungld (commented):
Could you update the verifier of ModOp to verify that `fmod` must be 0 if the input type is integer if you think it is a correct statement?

src/Transform/ONNX/ConstProp.cpp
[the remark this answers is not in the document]
tungld: Yes, please verify the case where the input type is integer.

src/Transform/ONNX/ConstProp.cpp
hamptonm1: Thanks for the review @tungld but there is a reason why I am using `lhs - floor(lhs / rhs) * rhs;` and not `remainder(lhs, rhs);`. Remainder is not always accurate in calculating the mod and I have a feeling that is the reason why mod is not working correctly. We use remainder in `Elementwise.cpp` but from testing we are not always getting the accurate results. I have calculated mod by hand and using google search engine and at times it differs from the results when using remainder. So to be safe I would rather do the following:

```
  // The original calculation for mod
    int mod = lhs - floor(lhs / rhs) * rhs;
    // Check if lhs is a multiple of rhs
    int64_t rem = remainder(lhs, rhs);

    // Handle the case when one of the int values are negative
    // If both int values are positive or multiples of each other, we can
    // calculate as normal
    if ((rem != 0) && ((lhs < 0) ^ (rhs < 0)))
      return (mod + rhs);
    return mod;
  }
};

```
The test below fails when using remainder... it is saying 5 mod 8 should be -3 but it actually should be 5. So this is another reason why I will not use remainder... the results are off. 

```
LC/llvm-project/build/bin/FileCheck /Users/meganhampton/zDLC/onnx-mlir/test/mlir/onnx/onnx_constprop.mlir
+ /Users/meganhampton/zDLC/onnx-mlir/build/Debug/bin/onnx-mlir-opt --shape-inference --constprop-onnx /Users/meganhampton/zDLC/onnx-mlir/test/mlir/onnx/onnx_constprop.mlir -split-input-file
+ /Users/meganhampton/zDLC/llvm-project/build/bin/FileCheck /Users/meganhampton/zDLC/onnx-mlir/test/mlir/onnx/onnx_constprop.mlir
/Users/meganhampton/zDLC/onnx-mlir/test/mlir/onnx/onnx_constprop.mlir:490:12: error: CHECK: expected string not found in input
 // CHECK: [[CONST:%.+]] = onnx.Constant dense<5> : tensor<i64>
           ^
<stdin>:335:49: note: scanning from here
 func.func @test_modulo_int_pos() -> tensor<i64> {
                                                ^
<stdin>:336:2: note: possible intended match here
 %0 = onnx.Constant dense<-3> : tensor<i64>
 ^

Input file: <stdin>
Check file: /Users/meganhampton/zDLC/onnx-mlir/test/mlir/onnx/onnx_constprop.mlir

-dump-input=help explains the following input dump.

Input was:
<<<<<<
             .
             .
             .
           330: } 
           331:  
           332:  
           333: // ----- 
           334: module { 
           335:  func.func @test_modulo_int_pos() -> tensor<i64> { 
check:490'0                                                     X~~ error: no match found
           336:  %0 = onnx.Constant dense<-3> : tensor<i64> 
check:490'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
check:490'1      ?                                           possible intended match
           337:  onnx.Return %0 : tensor<i64> 
check:490'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
           338:  } 
check:490'0     ~~~
           339: } 
check:490'0     ~~
           340:  
check:490'0     ~
           341:  
check:490'0     ~
             .
             .
             .
>>>>>>

--
```

src/Transform/ONNX/ConstProp.cpp
```diff
@@ -217,6 +217,34 @@ struct ElementWiseBinaryOpImpl<ONNXMaxOp, T> {
...
+    int mod = lhs - floor(lhs / rhs) * rhs;
+
+    // Handle the case when one of the int values are negative
+    if (lhs < 0 || rhs < 0) {
```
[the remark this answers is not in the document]
hamptonm1: Okay I updated the code to use `XOR` instead of `OR`. Thanks!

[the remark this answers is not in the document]
hamptonm1: I also added another lit test to cover two negative int values.

src/Accelerators/NNPA/Runtime/CMakeLists.txt
```diff
@@ -21,6 +21,6 @@ set_target_properties(RuntimeNNPA
   PROPERTIES
   LANGUAGE C
   POSITION_INDEPENDENT_CODE TRUE
-  COMPILE_OPTIONS "-O3;-fopenmp"
+  COMPILE_OPTIONS -option O3 fopenmp
```
hamptonm1, line 24: @tungld FYI  I added the -option to bypass this error I was seeing when compiling onnx-mlir locally.

`clang: error: unsupported option '-fopenmp'`

I got the idea to use option from the following documentation: https://cmake.org/cmake/help/latest/command/target_compile_options.html


Let me know if this is suitable.

src/Transform/ONNX/ConstProp.cpp
tungld: OK, so it's because `remainder` works on float type. So it would be:
```
int64_t rem = lhs % rhs;
if ((rem != 0) && ((lhs < 0) ^ (rhs < 0)))
  return (rem + rhs);
return rem;

File: src/Accelerators/NNPA/Runtime/CMakeLists.txt (hunk 6 above)
Comment (tungld) [reply src/Accelerators/NNPA/Runtime/CMakeLists.txt:24]: @chentong319 could you take a look at this? we are using clang by default to build onnx-mlir. Thanks! Since we haven't yet used openmp in zdnn extension, one way is to remove `-fopenmp`

File: src/Transform/ONNX/ConstProp.cpp (hunk 4 above)
Comment (hamptonm1) [reply src/Transform/ONNX/ConstProp.cpp:223]: @tungld Using ‘lhs%rhs’ does not work with eval. I have the error messages to prove it. I’m not sure what is wrong with using ‘floor’. It works accurately compared to remainder and all 6 of my lit tests pass. If we don’t have a reason why we shouldn’t use what I have, I’m not understanding. Thanks.

Comment (tungld) [reply src/Transform/ONNX/ConstProp.cpp:223]: Can you elaborate on `@tungld Using ‘lhs%rhs’ does not work with eval.`?

Comment (hamptonm1) [reply src/Transform/ONNX/ConstProp.cpp:223]: I have used ‘%’ in eval and received an error message something like this operation can not be performed. I can send you the direct error message once I have my laptop.

File: src/Accelerators/NNPA/Runtime/CMakeLists.txt (hunk 6 above)
Comment (AlexandreEichenberger) [reply src/Accelerators/NNPA/Runtime/CMakeLists.txt:24]: @gongsu832 also has a mechanism to invoke OpenMP only when it detects that OpenMP was built in the LLVM project. Maybe that is the mechanism to use here.

File: src/Transform/ONNX/ConstProp.cpp [hunk 7]
```diff
@@ -217,6 +217,31 @@ struct ElementWiseBinaryOpImpl<ONNXMaxOp, T> {
   static T eval(T lhs, T rhs) { return std::max<T>(lhs, rhs); }
 };
 
+template <>
+struct ElementWiseBinaryOpImpl<ONNXModOp, int64_t, EnableNotBool<int64_t>> {
+  static int64_t eval(int64_t lhs, int64_t rhs) {
+    // The original calculation for mod
+    int mod = lhs - floor(lhs / rhs) * rhs;
```
AlexandreEichenberger, line 224: Minimally, `mod` should be an `int64_t` as returned. Explicit type lengths are preferred.

hamptonm1: Thanks for the correction.... just seeing that by accident :)

src/Accelerators/NNPA/Runtime/CMakeLists.txt
hamptonm1: @chentong319 said I can remove it for now

src/Transform/ONNX/ConstProp.cpp
hamptonm1: @tungld This was the error message:

```
/Users/meganhampton/zDLC/onnx-mlir/src/Transform/ONNX/ConstProp.cpp:230:44: error: invalid operands to binary expression ('double' and 'double')
  static T eval(T lhs, T rhs) { return lhs % rhs; }
```

But because you wanted me to switch to using two templates and the data type is specified it works. Before it was when I was using `T`. I made the updates, I think this is ready to be merged in now. Thanks

src/Transform/ONNX/ConstProp.cpp
AlexandreEichenberger: Please push your changes, I don't see them yet.

hamptonm1: @AlexandreEichenberger Thanks!! There has been some connectivity issues today... the push was stalling. Thanks for the observation. What about now?

AlexandreEichenberger:
All good on my side, I see this comment by @tungld 

> Could you update the verifier of ModOp to verify that fmod must be 0 if the input type is integer if you think it is a correct statement?

Can you confirm that you acted on it? Tx

hamptonm1:
> All good on my side, I see this comment by @tungld
> 
> > Could you update the verifier of ModOp to verify that fmod must be 0 if the input type is integer if you think it is a correct statement?
> 
> Can you confirm that you acted on it? Tx

Confirmed!

AlexandreEichenberger (approved):
LGTM, Thanks again Megan.

Issue #2713, opened 2024-02-14
Segfault when running inference with the IBM granite.20B model (the one with KV cache) at the number of input tokens of at least 1265


tungld:
Segfault when running inference with the IBM granite.20B model (the one with KV cache) at the number of input tokens of at least 1265.

Segfault does not happen if the number of input tokens is less than 1265.

tungld:
Looks like something wrong with storing bool constants to file. PR #2714 is a workaround.

PR #2714, opened 2024-02-14
Do not store bool constants to file

tungld:
This is a temporary solution to avoid segfault reported in #2713


AlexandreEichenberger (approved):
LGTM

PR #2715, opened 2024-02-14
Fix link error on z/OS

mikeessen:
The link of the runtime fails on z/OS with recent changes in zDNNExtension.

 ```
IEW2456E 9207 SYMBOL __builtin_unreachable UNRESOLVED. MEMBER COULD NOT BE
INCLUDED FROM THE DESIGNATED CALL LIBRARY.
FSUM3065 The LINKEDIT step ended with return code 8.
```

Update the` omUnreachable()` to account for `__builtin_unreachable` and `__assume(false)` not being available on z/OS.


mikeessen:
@tungld Are you good with this fix?

tungld (approved):
LGTM. Thanks for the fix!

PR #2717, opened 2024-02-16
Fix position of registry for OpenMP dialect 

imaihal:
This PR moves `registerOpenMPDialectTranslation()` into `registerDialects()`. This should be better position. The `registry.insert<mlir::omp::OpenMPDialect>()` removed in this PR is done in `mlir::registerOpenMPDialectTranslation()`.

Without this PR, I got following errors when specifying `-parallel` option in numerical tests as shown below.

Example:
```
$ export OMP_NUM_THREADS=4
$ TestMatMul2D  -O3 -mcpu=z16 --mtriple=s390x-ibm-loz -parallel

error: cannot be converted to LLVM IR: missing `LLVMTranslationDialectInterface` registration for dialect for op: omp.parallel
```

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -16,7 +16,6 @@ add_onnx_mlir_executable(onnx-mlir
   onnx-mlir.cpp
 
   LINK_LIBS PRIVATE
-  MLIROpenMPToLLVMIRTranslation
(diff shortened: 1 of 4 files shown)


tungld (approved):
LGTM. Thanks for the fix!

PR #2721, opened 2024-02-20
Support Opset20 for IsInf and IsNaN

hamptonm1:
I updated onnx.IsInf and onnx.IsNaN to be compatible with opset 20. The updates for both of these ops is supporting new input data types such as float16 and bfloat16. So I made the applicable changes in the code base as well as the documentation so that these ops are backwards compatible.


src/Accelerators/NNPA/Runtime/CMakeLists.txt
```diff
@@ -21,6 +21,6 @@ set_target_properties(RuntimeNNPA
   PROPERTIES
   LANGUAGE C
   POSITION_INDEPENDENT_CODE TRUE
-  COMPILE_OPTIONS "-O3;-fopenmp"
+  COMPILE_OPTIONS -option O3 fopenmp
```
hamptonm1, line 24: @chentong319 I added the `-option` to bypass this error I was seeing when compiling `onnx-mlir`.

`clang: error: unsupported option '-fopenmp'`

I got the idea to use option from the following documentation: https://cmake.org/cmake/help/latest/command/target_compile_options.html

hamptonm1: Or I can remove `-fopenmp` .... it is blocking my building

chentong319: Yes, you can remove fopenmp flag now.
Which build failed with this flag? This flag will be needed in future.

chentong319:
Is it true that the latest version of IsInf and IsNaN allow more data type than the previous ones? Is there any other difference? If the latest version only adds some element types, we can use the newest op for the older versions. The only loss is the precise element type check in verifier, if there is a such check.

hamptonm1:
> Is it true that the latest version of IsInf and IsNaN allow more data type than the previous ones? Is there any other difference? If the latest version only adds some element types, we can use the newest op for the older versions. The only loss is the precise element type check in verifier, if there is a such check. Multi-versioning of ops should be avoid as much as possible.

@chentong319 Yes this is true, I can do that then. Let me update the code. Thanks

src/Dialect/ONNX/ONNXOps/Math/ElementwiseUnary.cpp
```diff
@@ -326,8 +326,8 @@ LogicalResult ONNXIsInfOp::inferShapes(
 
 LogicalResult ONNXIsNaNOp::inferShapes(
     std::function<void(Region &)> doShapeInference) {
-  IntegerType i1Type = IntegerType::get(getContext(), 1, IntegerType::Signless);
-  return inferShapeForUnaryOps(getOperation(), i1Type);
+  return inferShapeForUnaryOps(this->getOperation(),
+      this->getResult().getType().cast<ShapedType>().getElementType());
```
chentong319, line 330: Can it be guaranteed the output type is ShapedType?

hamptonm1: I can change It back to be safe... no guarantee

chentong319 (approved):
LGTM!

Issue #2639, opened 2023-11-24
Backend tests with the "force-dynamic" mode are not executed by "cmake check-onnx-backend-dynamic"


negiyas:
(Updated in March 6th, 2024)
Currently "make check-onnx-backend-dynamic" does not test dynamic cases but tests static cases, because "test/backend/common.py" sets the "TEST_IMPORTER_FORCE_DYNAMIC" environment variable, but "test/backend/variables.py" does not process it.

PR#2735 fixes this issue, and dynamic backend tests are enabled.
It also disabled the following 27 backend tests to pass the Jenkins backend tests temporally.

Your supports to fix these issues are very welcome!
If you fix some issues or find some unnecessary tests, please report the results at this Issue. Thanks.

1. test_instancenorm_epsilon_cpu
2. test_instancenorm_example_cpu
3. test_gathernd_example_float32_cpu
4. test_gathernd_example_int32_batch_dim1_cpu
5. test_gathernd_example_int32_cpu
6. test_castlike_DOUBLE_to_FLOAT_cpu
7. test_castlike_FLOAT_to_DOUBLE_cpu
8. test_convtranspose_3d_cpu
9. test_convtranspose_cpu
10. test_constant_pad_cpu
11. test_convtranspose_pad_cpu
12. test_convtranspose_kernel_shape_cpu
13. test_convtranspose_output_shape_cpu
14. test_einsum_transpose_cpu
15. test_einsum_inner_prod_cpu
16. test_depthtospace_example_cpu
17. test_convtranspose_1d_cpu
18. test_convtranspose_dilations_cpu
19. test_convtranspose_autopad_same_cpu
20. test_convtranspose_pads_cpu
21. test_einsum_batch_diagonal_cpu
22. test_depthtospace_crd_mode_example_cpu
23. test_einsum_batch_matmul_cpu
24. test_einsum_sum_cpu
25. test_edge_pad_cpu
26. ~~test_onnxmlir_top_k_float16_cpu~~                # Test condition fixed. K should be a1D tensor, cannot be dynamic.
27. ~~test_onnxmlir_top_k_smallest_float16_cpu~~  # Test condition fixed. K should be a1D tensor, cannot be dynamic
28. test_reflect_pad_cpu
29. test_spacetodepth_example_cpu
30. ~~test_top_k_smallest_cpu~~                             # Test condition fixed. K should be a1D tensor, cannot be dynamic
[...]

PR #2725, opened 2024-02-21
Fix dynamic shape for GroupNormalizeOp decomposition

chentong319:
#2639  Reported the error. #2617 tried to fix this problem by changing output type.
This PR fix the decomposition of GroupNormalization. 
The fix is simple: just use the output type of the original op to create the new op.
This is a general issue in transformation: if the provided output type of the original op is not the same as the result of shape inference, we have to make sure the generated op to use the original output type.


tungld (approved):
LGTM.

PR #2726, opened 2024-02-22
[NFC] Move passes to their own dialect

tungld:
MLIR often puts transformation passes into a folder `Transforms` inside `Dialect` for all passes related to a specific dialect.

This patch moves ONNX transformation passes into ONNX dialect folder to make the ONNX dialect self-contained. Passes that work across multiple dialects are still in `src/Transform`. Next PRs will be for other dialects.

For ONNX, I changed `Rewrite.{cpp,td}` to `Canonicalize.{cpp,td}` since we define all canonicalization rules there, which avoids naming confusion. Also `Canonicalize.{cpp,td}` is moved into the `ONNXOps` folder where ShapeInference resides. So, in summary, `Canonicalize`, `ShapeInference` and `Verifier` of ONNX ops are in the ONNXOps folder.


AlexandreEichenberger (approved):
Assume no changes, just file moved.

LGTM, thanks for the reorg

PR #2733, opened 2024-02-27
Clean up krnl.dim, krnl.shape, and krnl.getref

tungld:
`krnl.dim` and `krnl.shape` are not used in any generated code since we have moved to using IndexExpr when working with dimensions. So, this patch removes code related to `krnl.dim` and `krnl.shape`.


tungld:
> if you feel like it, there is also krnl.GetRef that is only used in 2 places

I removed it and replaced it by `memref.subview`.

PR #2736, opened 2024-03-03
Fix `onnx.Gather` conversion to `stablehlo` in dynamic case

srcarroll:
The previous implementation of `ONNXGatherOpLoweringToStablehlo` didn't handle dynamic input cases correctly yielding an invalid `stablehlo.dynamic_broacast_in_dim` op.  Nor was the dynamic case even tested. This patch fixes this and adds a test.


src/Conversion/ONNXToStablehlo/Tensor/Gather.cpp
```diff
@@ -56,19 +57,16 @@ struct ONNXGatherOpLoweringToStablehlo : public ConversionPattern {
...
+    if (!inputType.isDynamicDim(axisLit)) {
       int64_t axisDimSizeLit = inputType.getShape()[axisLit];
       axisDimSize = getShapedInt(loc, rewriter, axisDimSizeLit, indices);
     } else {
-      Value inputShape = rewriter.create<shape::ShapeOfOp>(loc, data);
-      Value indicesShape = rewriter.create<shape::ShapeOfOp>(loc, indices);
-      Value axisDimSizeIndexValue =
-          rewriter.create<shape::GetExtentOp>(loc, inputShape, axisLit);
-      Value axisDimSizeValue = rewriter.create<arith::IndexCastOp>(
-          loc, indicesType.getElementType(), axisDimSizeIndexValue);
```
srcarroll, line 68: the issue was that `axisDimSizeValue` is an `i64` not a `tensor<i64>` i could just use the old method but add a `tensor.from_elements` to get the right type. but i thought `tensor.dim` + `tensor.splat` was more straight forward. however if this is against design principles I can do the former.

tungld:

tungld:
@Connor-XY could you help take a look at this patch, if you don't mind? thanks.

Connor-XY (approved):
LGTM! I really appreciate that you fix the bug and add a test for dynamic case.

tungld:
Thank you very much @Connor-XY and @srcarrol!

tungld (approved):
LGTM.

PR #2738, opened 2024-03-04
[StableHLO] Add `onnx.Dim` lowering to StableHLO

brnorris03:
Lower `onnx.Dim` to `shape.get_extent` or constant.


tungld:

tungld:
@Connor-XY could you help take a look on this patch? and also some recent PRs about stablehlo. Thanks!

src/Dialect/ONNX/ONNXOps/Additional/Dim.cpp
```diff
@@ -37,12 +37,16 @@ LogicalResult ONNXDimOpShapeHelper::computeShape() {
...
+    return emitOpError("input must have shape and rank.");
+
   int64_t axis = this->getAxis();
-  return failure((axis < 0) || (axis >= getRank(this->getData().getType())));
+  if ((axis < 0) || (axis >= getRank(this->getData().getType())))
+    return emitOpError("attribute ")
+           << ONNXDimOp::getAxisAttrName() << " value is " << axis
+           << ", accepted range is [0, "
+           << getRank(this->getData().getType()) - 1 << "].";
+  return success();
```
tungld, line 49: Thanks so much for the explicit error messages! Really appreciate it.

src/Conversion/ONNXToStablehlo/Tensor/Dim.cpp
```diff
@@ -0,0 +1,64 @@
...
+    assert(tensorArg.getType().isa<RankedTensorType>() &&
+           "Expected ranked tensor type");
+
+    RankedTensorType tensorType = tensorArg.getType().cast<RankedTensorType>();
+    int64_t rank = tensorType.getRank();
+
+    assert((axis >= 0 && axis < rank) &&
+           "Axis must be in the range [0, input tensor rank - 1]");
+
+    Value dimValue = rewriter.create<tensor::DimOp>(loc, tensorArg, axis);
```
Connor-XY, line 44: Could we use [stablehlo.get_dimension_size](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#get_dimension_size) so we don't need to use tensor dialect?

brnorris03: Possibly, but it does make this way more complex than the tensor option, primarily because `stablehlo.get_dimension_size` returns `tensor<i32>` which must then be converted to the `tensor<1x64>` expected result type (so both the shape and size of the element type change). And of course there is a chance for overflow if the results exceeds i32 (which wouldn't happen with the`tensor` conversion).

src/Dialect/ONNX/ONNXOps/Additional/Dim.cpp
brnorris03: :-) You are welcome!

src/Conversion/ONNXToStablehlo/Tensor/Dim.cpp
brnorris03: Is there a particular reason why `tensor` should be avoided? I am also able to use the `shape` dialect (a bit more verbosely) for this, but that gets lowered to the `tensor.dim` op anyway.

tungld:

src/Conversion/ONNXToStablehlo/Tensor/Dim.cpp
tungld: Perhaps, @Connor-XY would like to use stablehlo ops at this level as many as possible. Not sure why `stablehlo.get_dimension_size` returns `tensor<i32>` while its input axis is small but i64 :)

brnorris03: That's understandable, but I don't actually see how I can avoid using `tensor` dialect ops entirely here. Also I am not able to successfully do the different-size index type conversions (`i32`->`i64`, going through `index` doesn't work). Any suggestions?

Connor-XY: Yes, I would like to convert it to stablehlo ops as much as we can. It is fine to convert it to `shape` or `tensor` dialect if it is hard to do so with pure stablehlo ops. It is possible to use `stablehlo.get_dimension_size` to get `tensor<i32>`, then `stablehlo.reshape` to get `tensor<1xi32>`, and then `stablehlo.convert` to convert it from `tensor<1xi32>` to `tensor<1xi64>`.

Connor-XY (approved):
LGTM! Thanks for adding support for `Onnx.Dim` conversion. It is fine to use `shape` or `tensor` dialect. No worries!

brnorris03:
> LGTM! Thanks for adding support for `Onnx.Dim` conversion. It is fine to use `shape` or `tensor` dialect. No worries!

Thank you!

tungld (approved):
LGTM.

PR #2739, opened 2024-03-04
Elide DenseResourceElementsAttr

tungld:
DenseResourceElementsAttr that is used for big constant stickifies in NNPA is not elided in the output file `model.tmp`. 

This patch adds a flag (that is newly supported in MLIR) to elide DenseResourceElementsAttr.


AlexandreEichenberger (approved):
LGTM, tx

PR #2740, opened 2024-03-04
Clean up ElideKrnlGlobalConstants

tungld:
ElideKrnlGlobalConstants pass is obsoleted, and was replaced by MLIR elide flags. So remove it in this patch.


AlexandreEichenberger (approved):
LGTM

PR #2744, opened 2024-03-07
Support Opset 20 for ReduceMin and ReduceMax

hamptonm1:
ReduceMin and ReduceMax now support data type `tensor(boolean)` for opset20. I updated the files accordingly, so that we can get things updated. Thanks! 

We need to have legacy support for both versions 18 and 13 to bypass the errors I was receiving in the backend tests because the backend tests for ReduceMin and ReduceMax are not separated by versions. So I had to update the decompose pass for version 18 to version 20.


hamptonm1:
@chentong319 @AlexandreEichenberger Let me know what you guys think and if I should make any further updates. Thanks!

src/Dialect/ONNX/ONNXOps/Math/Reduction.cpp
```diff
@@ -267,6 +269,15 @@ LogicalResult ONNXReduceMaxOp::inferShapes(
   return inferShapeForReductionOps<ONNXReduceMaxOp>(*this);
 }
 
+//===----------------------------------------------------------------------===//
+// ReduceMax legacy: ReduceMaxV18
+//===----------------------------------------------------------------------===//
+
+LogicalResult ONNXReduceMaxV18Op::inferShapes(
+    std::function<void(Region &)> doShapeInference) {
+  return inferShapeForReductionOps_old<ONNXReduceMaxV18Op>(*this);
```
philass, line 278: Shouldn't this be 
```
return inferShapeForReductionOps<ONNXReduceMaxV18Op>(*this);
```

src/Dialect/ONNX/ONNXOps/Math/Reduction.cpp
```diff
@@ -303,6 +314,15 @@ LogicalResult ONNXReduceMinOp::inferShapes(
   return inferShapeForReductionOps<ONNXReduceMinOp>(*this);
 }
 
+//===----------------------------------------------------------------------===//
+// ReduceMin legacy: ReduceMinV18
+//===----------------------------------------------------------------------===//
+
+LogicalResult ONNXReduceMinV18Op::inferShapes(
+    std::function<void(Region &)> doShapeInference) {
+  return inferShapeForReductionOps_old<ONNXReduceMinV18Op>(*this);
```
philass, line 323: Same here

philass (approved):
Other than that LGTM

It's a shame that a new op needs to be added for this, but I see how our current infrastructure makes that a necessity. This is beyond the scope of this PR, but we should rethink our versioning. In cases where the newer version is a super set of functionality with no breaking changes, we shouldn't have to add new ops.

AlexandreEichenberger (changes requested):
Code seems fine but I would like to know why a new op is needed before we merge, as we should avoid new versions as much as possible

AlexandreEichenberger (approved):
LGTM after explanation, it makes sense.

hamptonm1:
> LGTM after explanation, it makes sense.

@AlexandreEichenberger Thanks for the talk and I am glad everything makes sense now. Side point for reference if anyone wants to understand the reason why please view the description in this PR for clarity. Thanks!!

src/Dialect/ONNX/ONNXOps/Math/Reduction.cpp
hamptonm1: @philass Thanks for catching that!

PR #2747, opened 2024-03-08
Handle more general cases in `ONNX.Expand` lowering to `stablehlo`

srcarroll:
Previously the lowering only worked if the defining op of the `shape` operand was a constant or `ONNX.Shape` op.  It is unnecessary to specialize for the latter, so this patch generalizes it.


srcarroll:
@tungld can i get a review on this?

tungld (approved):
LGTM. Thanks!

PR #2748, opened 2024-03-12
Gathernd dynamic

chentong319:
Fix the issue for dynamic shape of GatherND.
According to the onnx document, `output tensor of rank q + r - indices_shape[-1] - 1 - b`. Therefore, if the last dimension of thesecond input, indices, is dynamic, the rank of the output tensor can't be computed. This PR uses "{0:{-1}}" for the dynamic shape.
Another change in the PR is to detect such case in ONNXPreKrnlVerify pass.

This PR also include @negiyas's fix for lowering to enable all the gathered backend test case pass.


chentong319:
@negiyas I included your changes.

negiyas:
@chentong319 Thank you so much!

tungld (approved):
LGTM.

Issue #2743, opened 2024-03-07
Dynamic backend tests cause errors.


negiyas:
(Updated in 2024/05/08)
All issues errors with dynamic backend tests (on CPU) are fixed. Thanks!

(Updated in 2024/03/26)

PR #2735 enables dynamic backend tests, but the following backend tests still fail.
Accordingly PR #2735 temporally disables these failed dynamic backend tests to pass the Jenkins tests.

Please look at the failed tests if you have time. We may have to (1) fix the issues, (2) change the test conditions, or (3) remove the tests if we can confirm that they are not supported.

Please check the following marks if fixed some of them. Thanks.
PR #2753 executes the failed dynamic backend tests, and  you can look at the error log at https://www.onnxmlir.xyz/jenkins/job/ONNX-MLIR-Pipeline-Docker-Build/14379/consoleText .

[X] 1. "test_castlike_FLOAT_to_DOUBLE_cpu" # Fixed by PR#2762
[X] 2. "test_castlike_DOUBLE_to_FLOAT_cpu" # Fixed by PR#2762
[X] 3. "test_castlike_FLOAT_to_FLOAT16_cpu" # Fixed by PR#2762
[X] 4. "test_castlike_FLOAT16_to_FLOAT_cpu" # Fixed by PR#2762
[X] 5. "test_castlike_FLOAT16_to_DOUBLE_cpu" # Fixed by PR#2762
[X] 6. "test_castlike_DOUBLE_to_FLOAT16_cpu" # Fixed by PR#2762
[X] 7. "test_convtranspose_1d_cpu"
[X] 8. "test_convtranspose_3d_cpu"
[X] 9. "test_convtranspose_autopad_same_cpu"
[X] 10. "test_convtranspose_cpu"
[X] 11. "test_convtranspose_dilations_cpu"
[X] 12. "test_convtranspose_kernel_shape_cpu"
[X] 13. "test_convtranspose_output_shape_cpu"
[X] 14. "test_convtranspose_pad_cpu"
[X] 15. "test_convtranspose_pads_cpu"
[X] 16. "test_depthtospace_example_cpu"
[X] 17. "test_depthtospace_crd_mode_example_cpu"
[X] 18. "test_einsum_batch_diagonal_cpu"
[X] 19. "test_einsum_batch_matmul_cpu"
[X] 20. "test_einsum_inner_prod_cpu"
[X] 21. "test_einsum_sum_cpu"
[...]

PR #2753, opened 2024-03-14
[Only for testing] Enable all dynamic backend test temporally.

negiyas:
For testing, this PR enables all dynamic backend test to check debugging status to confirm status of Issue #2743


negiyas:
I am closing this PR, since all dynamic backend tests on CPU are fixed.

PR #2757, opened 2024-03-18
Support Opset 20 for ConstantOfShape

mikeessen:
ConstantOfShape added support for `tensor(bfloat16)` in opset 20.  I updated the files accordingly.


docs/SupportedONNXOps-cpu.md
```diff
@@ -49,7 +49,7 @@ Onnx-mlir currently supports ONNX operations targeting up to opset 20. Limitatio
 | **Concat** |6 - * | | |
 | **ConcatFromSequence** |none | | | |
 | **Constant** |6 - * | | |
-| **ConstantOfShape** |9 - 19 | | |
+| **ConstantOfShape** |20 - * | | |
```
cjvolzka, line 52: The first number is the lowest opset we support for an operator. So that should stay as `9` here unless something happened that we no longer support this operator in older Opsets. 

If that was the case (which I don't think it is), then `# ==MIN== 9` would need to be updated for this Op in inference_backend.py as well.

src/Builder/OpBuildTable.inc
```diff
@@ -40,7 +40,7 @@ op_dialect_version_map_["Compress"] = {11};
 op_dialect_version_map_["Concat"] = {13};
 op_dialect_version_map_["ConcatFromSequence"] = {11};
 op_dialect_version_map_["Constant"] = {19};
-op_dialect_version_map_["ConstantOfShape"] = {9};
+op_dialect_version_map_["ConstantOfShape"] = {20};
```
cjvolzka, line 43: I'm not sure exactly how this file works, but based on some of the other operators (like Clip above), should be `{20, 9}`?

docs/SupportedONNXOps-cpu.md
mikeessen: > The first number is the lowest opset we support for an operator. So that should stay as `9` here unless something happened that we no longer support this operator in older Opsets.
> 
> If that was the case (which I don't think it is), then `# ==MIN== 9` would need to be updated for this Op in inference_backend.py as well.

Thanks, I will open a new PR and update `inference_backend.py` and `SupportedONNXOps-cpu.md`

src/Builder/OpBuildTable.inc
mikeessen: The {20, 9} would mean there are multiple versions of the operation implemented in onnxx-mlir, for example the interface of the operation changes between versions.  In the case of a `new data type` an additional version does need to be implemented, so `{20}` is correct.

Issue #2749, opened 2024-03-12
check-onnx-backend-numerical-nnpa fails following #2735


cjvolzka:
After #2735 merged, check-onnx-backend-numerical-nnpa fail. The commit prior to this PR merges, passes without issue.

Most if not all the failures seem to have `AttributeError: /workdir/onnx-mlir/build/test/accelerators/NNPA/backend/Release/check-onnx-backend-dynamic-nnpa/test_min_float32/test_<some_test>.so: undefined symbol: zdnn_<some_api>` 

Several tests fail, see [check-onnx-backend-numerical-nnpa.failures.txt](https://github.com/onnx/onnx-mlir/files/14577310/check-onnx-backend-numerical-nnpa.failures.txt). Below is an example from `test_add_cpu`
```
____________________ OnnxBackendNodeModelTest.test_add_cpu _____________________
[gw4] linux -- Python 3.10.12 /usr/bin/python3.10

test_self = <test.OnnxBackendNodeModelTest testMethod=test_add_cpu>
device = 'CPU', kwargs = {}
model_pb_path = '/usr/local/lib/python3.10/dist-packages/onnx/backend/test/data/node/test_add/model.onnx'
model_dir = '/usr/local/lib/python3.10/dist-packages/onnx/backend/test/data/node/test_add'
use_dummy = False
model = ir_version: 7
producer_name: "backend-test"
graph {
  node {
    input: "x"
    input: "y"
    output: "sum"
    op_ty...  dim {
            dim_value: 5
          }
        }
      }
    }
  }
}
opset_import {
  domain: ""
  version: 14
}


    def run(test_self: Any, device: str, **kwargs) -> None:
        if model_test.url is not None and model_test.url.startswith(
            "onnx/backend/test/data/light/"
        ):
            # testing local files
            model_pb_path = os.path.normpath(
                os.path.join(
                    os.path.dirname(__file__),
                    "..",
                    "..",
                    "..",
                    "..",
                    model_test.url,
                )
            )
            if not os.path.exists(model_pb_path):
                raise FileNotFoundError(f"Unable to find model {model_pb_path!r}.")
            onnx_home = os.path.expanduser(
                os.getenv("ONNX_HOME", os.path.join("~", ".onnx"))
            )
            models_dir = os.getenv(
                "ONNX_MODELS", os.path.join(onnx_home, "models", "light")
            )
            model_dir: str = os.path.join(models_dir, model_test.model_name)
[...]

PR #2760, opened 2024-03-19
Fix Issue #2749 with reducing tests for check-onnx-backend-dynamic-nnpa .

negiyas:
This patch fix issue #2749 with reducing tests for check-onnx-backend-dynamic-nnpa, since NNPA cannot be used for some cases with dynamic inputs.


cjvolzka (approved):
I tried it out and it looks like this takes care of the issue so LGTM

PR #2761, opened 2024-03-19
Update LLVM to a4ca07f13b560b4f6fa5459eef7159e4f9ee9a6b

brnorris03:


hamptonm1:

AlexandreEichenberger:
FYI, look into the window error. Sometime, the build needs to be repeated as it has not enough resource to rebuild both the LLVM and  ONNX-MLIR. Other source of potential (trivial) errors are that warnings are treated as error; so check if the window compiler issued a warning. Hope this helps.

brnorris03:
> FYI, look into the window error. Sometime, the build needs to be repeated as it has not enough resource to rebuild both the LLVM and ONNX-MLIR. Other source of potential (trivial) errors are that warnings are treated as error; so check if the window compiler issued a warning. Hope this helps.

It says that it ran out of space when linking one of the tests (`TestConv.exe`). What should I do about that?

requirements.txt
```diff
@@ -2,6 +2,7 @@ lit~=15.0
 # numpy 1.24 deprecates np.object, np.bool, np.float, np.complex, np.str,
 # and np.int which are used heavily in onnx-mlir.
 numpy~=1.22.2, <=1.23.5
+onnx~=1.15.0
```
hamptonm1, line 5: @brnorris03 Why did you add onnx to the requirement.txt file? We have a step in the docker file to install onnx here: https://github.com/onnx/onnx-mlir/blob/main/docker/Dockerfile.onnx-mlir-dev#L16 and locally we would need to do `pip install -e third_party/onnx` to ensure we are installing the correct onnx commit.

brnorris03: It is required by the backend tests. I had been adding it manually to be able to test. I can change `requirements.txt` to use the local version. I don't really see a need to install manually.

hamptonm1: kk no worries, just curious! I will leave that decision to @AlexandreEichenberger :)

hamptonm1 (approved):
LGTM!! Thanks for all your hard work and help to make this happen :)

PR #2762, opened 2024-03-19
Update the dynamic shape values for onnx.CastLike

hamptonm1:
When issuing the target `check-onnx-backend-dynamic` we receive the following error for all `onnx.Castlike` tests:

`error: failed to materialize conversion for result #0 of operation 'onnx.CastLike' that remained live after conversion`

I simply update the dynamic shape values to bypass this error. Please let me know your thoughts. Thank you!


AlexandreEichenberger (approved):
LGTM

PR #2765, opened 2024-03-20
[mlir][ONNX] Implement `TileOp` canonicalizer

srcarroll:
Canonicalize `TileOp` when it is identity, i.e. when the input/result shape is same and all `repeat` values are constantly known to be 1.

### DIFF
--- a/src/Dialect/ONNX/ONNXOps/Canonicalize.td
+++ b/src/Dialect/ONNX/ONNXOps/Canonicalize.td
@@ -134,6 +134,12 @@ class HaveSameDim<int dim>: Constraint<
         "$1.getType().cast<RankedTensorType>().getShape()[" # dim # "])">,
   "Two tensors have the same specified dimension">;
 
+def HaveSameShapedType: Constraint<
+    CPred<"(isa<ShapedType>($0.getType()) &&"
+          "dyn_cast<ShapedType>($0.getType()) == "
+          "dyn_cast<ShapedType>($1.getType()))">,
+    "has same shaped type">;
+
 def HaveSameStaticShape: Constraint<
   CPred<"onnx_mlir::haveSameStaticShape($0, $1)">,
   "Two tensors have the same static shape">;
@@ -195,8 +201,8 @@ def IsNotFromONNXConstantOp: Constraint<
     "Is a value not from ONNXConstantOp">;
 
 def IsFromONNXConstantOpWithDenseElementsAttr: Constraint<
-    And<[CPred<" isa<ONNXConstantOp>($_self.getDefiningOp()) ">,
-         CPred<" onnx_mlir::getONNXConstantOp($_self).getValueAttr().isa<DenseElementsAttr>() ">
+    And<[CPred<" onnx_mlir::getONNXConstantOp($_self) ">,
+         CPred<" isa<DenseElementsAttr>(onnx_mlir::getONNXConstantOp($_self).getValueAttr()) ">
         ]>, "Value is not a ONNXConstantOp with a DenseElementsAttr">;
 
 def AreTheSameAxesConstant: Constraint<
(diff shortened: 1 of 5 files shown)


src/Dialect/ONNX/ONNXOps/Canonicalize.td
```diff
@@ -195,8 +201,8 @@ def IsNotFromONNXConstantOp: Constraint<
     "Is a value not from ONNXConstantOp">;
 
 def IsFromONNXConstantOpWithDenseElementsAttr: Constraint<
-    And<[CPred<" isa<ONNXConstantOp>($_self.getDefiningOp()) ">,
-         CPred<" onnx_mlir::getONNXConstantOp($_self).getValueAttr().isa<DenseElementsAttr>() ">
+    And<[CPred<" onnx_mlir::getONNXConstantOp($_self) ">,
+         CPred<" isa<DenseElementsAttr>(onnx_mlir::getONNXConstantOp($_self).getValueAttr()) ">
```
srcarroll, line 205: I changed this because i was getting crashes with the previous implementation. So I opted for a consistent check by using getONNXConstantOp in both.

brnorris03:

src/Dialect/ONNX/ONNXOps/Canonicalize.td
tungld: Do you have an example that can reproduce the crash?

tungld (approved):
LGTM. Thanks!

AlexandreEichenberger:
@srcarroll The PR is can be merged; since you probably don't have merge right, let me know here when you would like us to do so.

srcarroll:
that's right, i dont have permissions, so yes please merge it for me. thanks!

src/Dialect/ONNX/ONNXOps/Canonicalize.td
srcarroll: it would crash whenever the input is not from a constant op

tungld: Is the input a block argument? say an argument of a function. In such a case, `$_self.getDefiningOp()` may crash. If so, we can fix this by using `$_self.isa<BlockArgument> && isa<ONNXConstantOp>($_self.getDefiningOp())`

srcarroll: ah yes, good catch. it is specifically a problem for block arg. however i don't think `$_self.getDefiningOp()` should crash in this case, but rather just return nullptr. let me double check the error i get again to confirm, but i think it was a casting assertion error. in which case it was `isa` that was crashing and i think the proper modification to the original is to do `$_self.getDefiningOp<ONNXConstantOp>()`.

srcarroll: also, the reason `getONNXConstantOp` doesn't crash on this is because it does a `dyn_cast_or_null`, which is safer for the result of `getDefiningOp`

tungld: Yes, `$_self.getDefiningOp<ONNXConstantOp>()` would work.

srcarroll: i updated to this. thanks for the discussion

tungld:

src/Dialect/ONNX/ONNXOps/Canonicalize.td
tungld: Thanks! It seems you forgot to sign some last commits so that DCO failed.

srcarroll: oh sorry about that. let me fix

tungld:

AlexandreEichenberger:
@srcarroll thanks for your contributions, much appreciated.

PR #2766, opened 2024-03-21
Introduce krnl.get_linear_offset_index to get a linear offset from a MemRef

tungld:


AlexandreEichenberger (approved):
LGTM.

AlexandreEichenberger:
@tungld : 2 observations.

I am now using this approach and it works well for generating stickily code. But I notice that the code pattern could be optimized a bit further.

Example of generated code:
```mlir
            %7 = arith.cmpi slt, %arg2, %c0 : index
            %8 = arith.subi %c-1, %arg2 : index
            %9 = arith.select %7, %8, %arg2 : index
            %10 = arith.divsi %9, %c32 : index
            %11 = arith.subi %c-1, %10 : index
            %12 = arith.select %7, %11, %10 : index
            %13 = arith.remsi %arg2, %c32 : index
            %14 = arith.cmpi slt, %13, %c0 : index
            %15 = arith.addi %13, %c32 : index
            %16 = arith.select %14, %15, %13 : index
            %17 = arith.muli %16, %c64 : index
            %18 = arith.muli %12, %c2048 : index
            %19 = arith.addi %17, %18 : index
            %20 = arith.muli %1, %c2048 : index
            %21 = arith.muli %20, %arg4 : index
            %22 = arith.addi %19, %21 : index
            %23 = arith.muli %20, %0 : index
            %24 = arith.muli %23, %arg1 : index
            %25 = arith.addi %22, %24 : index
            %reinterpret_cast = memref.reinterpret_cast %alloc to offset: [%25], sizes: [1, 2, 64], strides: [128, 64, 1] : memref<?x?x1x?x32x64xf16> to memref<1x2x64xf16>
```

First there are selects introduced on whether indices are positive or negative. I don't know if there is away to indicate that the indices (which are array indices and should always be nonnegative) are nonnegative and thus don't need this special casing. 

Second, a more minor point. The formula you use to compute the offset is of the following type:

```
offset = ((a*u+b)*v+c)*w+d
```
since the dimensions `u, v, w` are typically loop independent, we can compute them separately:

```
w' = w
v' = v * w'
u' = u * v'

offset = a*u' + b * v' + c * w' + d
```

Then if some of the terms are loop independent, they can easily get out of the innermost loop. Say we iterate only over `b` in the innermost loop, `u' v', w', a*u', c*w'` can all be hoisted from the loop.

The hoisting would likely occur in the backend (aka llvm). I have not double checked in this example what gets hoisted.

PR #2768, opened 2024-03-22
Compiler generated Unstick

AlexandreEichenberger:
Generate compiler code to generate unstick operation. The code convert the f16 into f32 into an Nx(64M) tile of data in SIMD fashion. Then the tile is copied back to the output, which used normal data layout.
The operation supports 4D, 3DS, 3D, and 2D. 1D and 2DS are not supported on the CPU, so no support is provided here.

Data below show you the performance results for a add operation that runs on the zAIU, with zDNN vs compiler generated (`-enable-compiler-stick-unstick`) without and with parallelism (6 threads).

image

Tests on Roberta shows a 17% improvement when data size is a compile time constant. When dynamic sizes are used, it's a roughly 17% decrease in performance, mainly because the compiler generated code is much slower than the original zDNN. To the best of my knowledge, it's because the current compiler code does not generate prefetch. This issue will be investigated in a subsequent PR.


AlexandreEichenberger:
Code was verified using `CheckONNXModel.py` for 4D, 3DS, and 3D using shapes that are just the perfect size (e.g. 32x64), too big, or too small.

Code was also verified using Roberta.
```
OMP_NUM_THREADS=6 CheckONNXModel.py -m roberta-base-11.onnx -r="-O3 -mcpu=z16 -maccel=NNPA -parallel" -a="-enable-compiler-stick-unstick" -shape-info=0:6x384

> Reference command:  /workdir/onnx-mlir/build/Debug/../../utils/RunONNXModel.py --compile-args="-O3 -mcpu=z16 -maccel=NNPA -parallel" --save-ref=hape-info=0:6x384 --seed=42 --model=roberta-base-11.onnx
Filed while executing reference compile and run
Temporary directory has been created at /tmp/tmp9cbryun5
Compiling the model ...
  took  116.02507415041327  seconds.

Loading the compiled model ...
  took  0.00034797564148902893  seconds.

Generating random inputs using seed 42 ...
The shape of the 1st input is unknown. Use --shape-info to set.
 - The input signature:  {'type': 'i64', 'dims': [-1, -1], 'name': 'input_ids'}
Return code 1
root@02ca841a26f2:~/workspace/models/roberta-base-11# OMP_NUM_THREADS=6 CheckONNXModel.py -m roberta-base-11.onnx -r="-O3 -mcpu=z16 -maccel=NNPA -parallel" -a="-enable-compiler-stick-unstick" --shape-info=0:6x384

> Reference command:  /workdir/onnx-mlir/build/Debug/../../utils/RunONNXModel.py --compile-args="-O3 -mcpu=z16 -maccel=NNPA -parallel" --save-ref=check-ref --shape-info=0:6x384 --seed=42 --model=roberta-base-11.onnx
>   Successfully ran the reference example, saved refs in "check-ref".

> Test command:  /workdir/onnx-mlir/build/Debug/../../utils/RunONNXModel.py --compile-args="-O3 -mcpu=z16 -maccel=NNPA -parallel -enable-compiler-stick-unstick" --load-ref=check-ref --verify=ref --verify-every-value --model=roberta-base-11.onnx
>   Successfully ran the test example and verified against "check-ref".
```

AlexandreEichenberger:
I added some DialectBuilder support for `prefetch` for the affine and memref dialect, but it is not usable at this time when generating our ONNX-to-KRNL lowering.

src/Accelerators/NNPA/Conversion/ZHighToZLow/ZHighToZLow.cpp
```diff
@@ -32,15 +32,19 @@
 #include "src/Support/TypeUtilities.hpp"
 
 #define DEBUG_TYPE "zhigh-to-zlow"
-
+#define ENABLE_CSU_PAR true /* Allow parallel compiler gen Stick/Unstick. */
+#define CS_N 2              /* Tiling for Stick */
+#define CS_M 2              /* Tiling for Stick */
+#define CU_N 2              /* Tiling for Unstick */
+#define CU_M 2              /* Tiling for Unstick */
```
tungld, line 39: Perhaps, moving them inside onnx_mIir::zhigh namespace is better.

tungld (approved):
LGTM. Looking forward to seeing how prefetch helps boost the performance.

src/Accelerators/NNPA/Conversion/ZHighToZLow/ZHighToZLow.cpp
AlexandreEichenberger: Agreed, once I am done with parameter tuning, I will migrate the files back down in the code, its easier to edit with vi all the terms in one spot.

Cleanup will also remove the alternative algorithms. Hiroshi also told me how to use big pages, which is what they did, possibly to avoid the prefetch issues. Time will tell if prefetch can alleviate the issues with dynamic shapes (which I think results in repetitive mallocs, which may exhibit less locality as what onnx-mlir does when sizes are static). This theory should be investigated, or maybe someone already knows :-)

PR #2772, opened 2024-03-23
Relax dynamic restriction on `ScatterND` stablehlo conversion

srcarroll:
The previous implementation asserted that all dims of the `indices` operand were static. First of all, assertion isn't appropriate here as the definition of the op allows dynamic.  This is a condition for the pattern match and thus should fail the match with `notifyMatchFailure`. Second, the implementation only relies on the last dim being static, so we can relax the match condition and only fail if last dim is dynamic.

### DIFF
--- a/src/Conversion/ONNXToStablehlo/Tensor/ScatterND.cpp
+++ b/src/Conversion/ONNXToStablehlo/Tensor/ScatterND.cpp
@@ -40,8 +40,9 @@ struct ONNXScatterNDOpLoweringToStablehlo
     auto indicesType = indices.getType().cast<ShapedType>();
     int64_t dataRank = dataType.getRank();
     int64_t indicesRank = indicesType.getRank();
-    assert(indicesType.hasStaticShape() &&
-           "only support indices with static shape");
+    if (indicesType.isDynamicDim(indicesRank - 1))
+      return rewriter.notifyMatchFailure(
+          op, "only support indices with static last dim");
     int64_t partialIdxDim = indicesType.getDimSize(indicesRank - 1);
 
     assert(dataRank >= 1 && "The rank of 'data' must be >= 1");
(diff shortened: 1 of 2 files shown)


srcarroll:
I'll be honest. I don't completely understand the lowering of `ScatterND` so I'm not sure if the result is correct. But I just noticed that the assertion was too restrictive based on the fact that only the last dim of `indices` needed to be static.  So I'm assuming everything else is correct when relaxing the restriction. Please confirm.

AlexandreEichenberger:

AlexandreEichenberger:

AlexandreEichenberger (approved):
LGTM, feel free to get a second opinion from an HLO expert.

AlexandreEichenberger:
@srcarroll let me know when you are ready for me to push the code.

Connor-XY:
> I'll be honest. I don't completely understand the lowering of `ScatterND` so I'm not sure if the result is correct. But I just noticed that the assertion was too restrictive based on the fact that only the last dim of `indices` needed to be static. So I'm assuming everything else is correct when relaxing the restriction. Please confirm.

The assertion is restrictive since I would like to provide a basic implementation for static-shaped tensors first, and fail dynamic ones temporarily. I believe that you are now supporting dynamic cases and we could remove some restrictions gradually. As for the `ScatterND` op, it indeed only requires the last dim of `indices` to be static, and your test case should work as I don't see any problem with it.

srcarroll:
> > I'll be honest. I don't completely understand the lowering of `ScatterND` so I'm not sure if the result is correct. But I just noticed that the assertion was too restrictive based on the fact that only the last dim of `indices` needed to be static. So I'm assuming everything else is correct when relaxing the restriction. Please confirm.
> 
> The assertion is restrictive since I would like to provide a basic implementation for static-shaped tensors first, and fail dynamic ones temporarily. I believe that you are now supporting dynamic cases and we could remove some restrictions gradually. As for the `ScatterND` op, it indeed only requires the last dim of `indices` to be static, and your test case should work as I don't see any problem with it.

@Connor-XY  Thanks for the confirmation! @AlexandreEichenberger you can merge. I'm rebasing now just in case

AlexandreEichenberger:

PR #2775, opened 2024-03-27
[NNPA] Add one more constraint to the pattern DimUnstickRemovalPattern

tungld:
The pattern `DimUnstickRemovalPattern` is for ONNXDim to get dimension directly from a ztensor without unstick. However, we have another rule to reconstruct `stick -> zdnn op -> unstick` back to an ONNX op if NNPA is not beneficial. That situation results in a generated code where both the onnx op and zdnn op exists while the zdnn op would be redundant.
This produces the overhead caused by the redundant zdnn op.

This patch adds one more constraint to the pattern so that the pattern is applied only when the consuming ops are zdnn ops so that unstick is unnecessary, and getting dim directly from the ztensor is really beneficial.


AlexandreEichenberger (approved):
LGTM, make sense; thanks for finding out the pattern

PR #2776, opened 2024-03-27
Added krnl region to krnl dialect builder

AlexandreEichenberger:
Added a function in Krnl Dialect Builder to create a region using the standard format

``` c++
  create.krnl.region([&](KrnlBuilder &createKrnl) {
   // code inside the body of the region
  });
```

Did not test, as there is no easy place to use it in the current code base.


imaihal (approved):
LGTM. Thanks!

PR #2777, opened 2024-03-30
Fix min opset doc for IsInf and IsNaN

cjvolzka:
# Description

https://github.com/onnx/onnx-mlir/pull/2721 updated the ops to include opset 20 support but set the documented min to opset to 20. This reverts the documented min opset to what it was before.


AlexandreEichenberger (approved):
Good catch, tx

Issue #2769, opened 2024-03-22
Performance degradation in Models running on NNPA


cjvolzka:
The following models are taking longer when running on NNPA now compared to the 0.4.1 release. 

* gpt2-10.onnx 
    * about 30% worse  
    * 0.4.1 - Total runMainGraph() time over all 100 inference(s) was 823.821 ms averaging 8.23821 ms per inference.
    * main - Total runMainGraph() time over all 100 inference(s) was 1187.16 ms averaging 11.8716 ms per inference.
* roberta-sequence-classification-9.onnx 
    * about 28% worse
    * 0.4.1 - Total runMainGraph() time over all 100 inference(s) was 497.415 ms averaging 4.97415 ms per inference.
    * main - Total runMainGraph() time over all 100 inference(s) was 692.769 ms averaging 6.92769 ms per inference.
* mobilenetv2-12.onnx 
    * about 6% worse
    * 0.4.1 - Total runMainGraph() time over all 100 inference(s) was 1705.98 ms averaging 17.0598 ms per inference.
    * main - Total runMainGraph() time over all 100 inference(s) was 1809.34 ms averaging 18.0934 ms per inference.
* shufflenet-v2-12.onnx
    * about 6% worse
    * 0.4.1 - Total runMainGraph() time over all 100 inference(s) was 852.561 ms averaging 8.52561 ms per inference.
    * main - Total runMainGraph() time over all 100 inference(s) was 905.177 ms averaging 9.05177 ms per inference.

tungld:
@cjvolzka Were these two lines swapped? I saw `main` is faster than `0.4.1`
```
0.4.1 - Total runMainGraph() time over all 100 inference(s) was 1187.16 ms averaging 11.8716 ms per inference.
main - Total runMainGraph() time over all 100 inference(s) was 823.821 ms averaging 8.23821 ms per inference.
```

I tried gpt2-10 with the latest onnx-mlir, but could not get any number close to 8.2ms or 11.8 ms per inference.
```shell
$ /workdir/onnx-mlir/build/Debug/bin/onnx-mlir -O3 -mcpu=z16 -maccel=NNPA gpt2-10.onnx --EmitLib -v --onnx-op-stats TXT
```
 
I ran 3 warmup iterations, then 100 iterations on a shared machine:
```shell
[...]

PR #2778, opened 2024-04-01
Add a flag to turn on/off the lowering of scalar broadcasting binary ops to NNPA

tungld:
Add a compile flag, `--nnpa-enable-scalar-bcast-binary`, to turn on/off the lowering of scalar broadcasting binary ops to NNPA, which is flexible for debugging. Default value is off.


cjvolzka:
I gave this a try today and it looks like this closes the performance gap for roberta-sequence-classification-9 from https://github.com/onnx/onnx-mlir/issues/2769

chentong319 (approved):
LGTM!

If you pass the option into the pass, the onnx-mlir-opt should work, too.

tungld:
@chentong319 thanks for the comment! I forgot to add a lit test. Since the option is added into `OnnxMlirCommonOptions` category, so it is available in `onnx-mlir-opt` also.

PR #2780, opened 2024-04-02
Update op definitions to make them work with the new bufferization

tungld:
The new bufferization mechanism in recent LLVM was slower than the old one, and we proposed `--use-old-bufferization=true` to use the old one.

Today, I tested the new mechanism again by `--use-old-bufferization=false`, and it failed to compile a model. To fix it, we need to add  `DeclareOpInterfaceMethods<MemoryEffectsOpInterface>]>` to all operations that work with MemRef.

This PR is to add `DeclareOpInterfaceMethods<MemoryEffectsOpInterface>]>` to krnl and zlow ops.


AlexandreEichenberger (approved):
LGTM,

was it still slower?

tungld:
> was it still slower?

Unfortunately, yes, still 2x slower than the old bufferization.

PR #2781, opened 2024-04-02
Introduce --dimParams option to set onnx.dim_params attrs of dynamic model inputs, and extend dynamic backend tests on NNPA with this option to utilze NNPA

negiyas:
===== 2024/04/19 updated according to local discussions =====
This PR support the following two features.

(1) This PR introduces `--dimParams` option for onnx-mlir to set `onnx.dim_params` attrs of dynamic model inputs, which specifies relationships among dimensions of model inputs. For example, onnx-mlir cannot utilize NNPA for the following dynamic model, because broadcasting of %arg0 or %arg1 may be required and broadcasting is not supported by NNPA.

_module {
  func.func @main_graph(%arg0: tensor<?x?x?xf32> {onnx.name = "x"}, %arg1: tensor<?x?x?xf32> {onnx.name = "y"})
    -> (tensor<3x4x5xf32> {onnx.name = "sum"}) {
    %0 = "onnx.Add"(%arg0, %arg1) : (tensor<?x?x?xf32>, tensor<?x?x?xf32>) -> tensor<?x?x?xf32>
    onnx.Return %0 : tensor<?x?x?xf32>
  }
}_

We can use the `--dimParams` option to specify that all corresponding dimensions between two dynamic inputs %arg0 and %arg1 are the same by `--dimParams="0:0=a,1=b,2=c|1:0=a,1=b,2=c"`. Accordingly compiler can know no broadcasting is required and can utilize NNPA in this case.

_module {
  func.func @main_graph(%arg0: tensor<?x?x?xf32> {**onnx.dim_params = "0:a,1:b,2:c",** onnx.name = "x"}, %arg1: tensor<?x?x?xf32> {**onnx.dim_params = "0:a,1:b,2:c",** onnx.name = "y"}) -> (tensor<?x?x?xf32> {onnx.name = "sum"}) {
    %0 = "onnx.Add"(%arg0, %arg1) : (tensor<?x?x?xf32>, tensor<?x?x?xf32>) -> tensor<?x?x?xf32>
    onnx.Return %0 : tensor<?x?x?xf32>
  }
}_

(2) This PR also extends functions of dynamic backend tests for NNPA by using the `--dimParams` option
In backend tests on NNPA, support new notations for setting arguments of the option in the NNPA_TEST_LIST variable in the test/accelerators/NNPA/backend/CMakeLists.txt.  The following examples mean as follows. 
```
    test_add_cpu,zdnn_add,"0:0=a,1=b,2=c|1:0=a,1=b,2=c"
    test_averagepool_2d_default_cpu,zdnn_avgpool2d,NO_DYNAMIC_SHAPE_TEST
```
- In the "test_add_cpu" test, "zdnn_add" should be called and the first, second and third dimensions of the first and second inputs are the same.
- In the "test_averagepool_2d_default_cpu" test, dynamic shape test on NNPA should not be executed.

### DIFF
--- a/docs/Testing.md
+++ b/docs/Testing.md
@@ -49,11 +49,14 @@ Testing with dynamic tensor sizes is most easily performed by using the followin
 cmake --build . --config Release --target check-onnx-backend-dynamic[-jni]
 ``` 
 
-The onnx node tests usually have known dimension size for input tensors. So, to test tensor with unknown dimension, the model importer (Build/FrontendONNXTransformer.cpp) provides a functionality to generate such cases. When the environment variable, `IMPORTER_FORCE_DYNAMIC`, is set, the frontend import will turn the all the dimensions (by default) of all the input tensors of the model into -1. For example,
+The onnx node tests usually have known dimension size for input tensors. So, to test tensor with unknown dimension, the model importer (Build/FrontendONNXTransformer.cpp) provides a functionality to generate such cases. When the environment variable, `IMPORTER_FORCE_DYNAMIC`, is set, the frontend import will turn the all the dimensions (by default) of all the input tensors of the model into -1. As default, we assume that first dimension of all the inputs are the same, and the second dimension of all inputs are the same, and so on.
+
+For example,
 ```
 IMPORTER_FORCE_DYNAMIC='-1:-1' all dimensions of all the inputs will be changed
 IMPORTER_FORCE_DYNAMIC='0:-1' all dimensions of the first input will be changed
 IMPORTER_FORCE_DYNAMIC='0:-1|1:0,1' all dimensions of the first input and the 1st and 2nd dimensions of the second input will be changed
+IMPORTER_FORCE_DYNAMIC='0:0:a,1:b,2:c|1:0:a,1:b,2:c' the first three dimensions of the first of and the second inputs are changed. And assume that the first dimensions of the first and second arguments are the same, and same for the the second and third dimensions.
 ```
 
 The Backus-Naur Form (BNF) for `IMPORTER_FORCE_DYNAMIC` is as follows.
@@ -63,10 +66,11 @@ The Backus-Naur Form (BNF) for `IMPORTER_FORCE_DYNAMIC` is as follows.
             <inputString ::= <inputIndex> `:` <dimString>
              <dimString> ::= <dimIndex> | <dimIndex> `,` <dimString>
             <inputIndex> ::= <index>
-              <dimIndex> ::= <index>
+              <dimIndex> ::= <index> | <index> ':' <symbol>
                  <index> ::= -1 | <number>
                 <number> ::= <digit> | <digit><number>
                  <digit> ::= 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9
+                <symbol> ::= 'a', 'b', 'c', ...
 ```
 Value `-1` semantically represents all inputs or all dimensions, and it has the highest priority. E.g. `'0: -1, 0'` means all dimensions of the first input will be changed. Input and dimension indices start from 0.
 
@@ -76,15 +80,18 @@ func @main_graph(%arg0: tensor<3x4x5xf32>, %arg1: tensor<3x4x5xf32>) -> tensor<3
 ```
 with `IMPORTER_FORCE_DYNAMIC='-1:-1'`, the result is:
 ```
-func @main_graph(%arg0: tensor<?x?x?xf32>, %arg1: tensor<?x?x?xf32>) -> tensor<?x?x?xf32>
+func @main_graph(%arg0: tensor<?x?x?xf32>{onnx.name = "x“, onnx.dim_params = "0:a,1:b,2:c"}, %arg1: tensor<?x?x?xf32>{onnx.name = "y“, onnx.dim_params = "0:a,1:b,2:c"}) -> tensor<?x?x?xf32>
 ```
+The onnx.dim_params attributes are used to specify the assumptions about dimensions of the first and second arguments.
+
 with `IMPORTER_FORCE_DYNAMIC='0:-1'`, the result is:
 ```
-func @main_graph(%arg0: tensor<?x?x?xf32>, %arg1: tensor<3x4x5xf32>) -> tensor<3x4x5xf32>
+func @main_graph(%arg0: tensor<?x?x?xf32>{onnx.name = "x“, onnx.dim_params = "0:a,1:b,2:c"}, %arg1: tensor<3x4x5xf32>) -> tensor<3x4x5xf32>
 ```
-with `IMPORTER_FORCE_DYNAMIC='0:0,2|1:1'`, the result is:
+
+with `IMPORTER_FORCE_DYNAMIC='0:0:a,2:b|1:1:a'`, the result is:
 ```
-func @main_graph(%arg0: tensor<?x4x?xf32>, %arg1: tensor<3x?x5xf32>) -> tensor<3x4x5xf32>
+func @main_graph(%arg0: tensor<?x4x?xf32>{onnx.name = "x“, onnx.dim_params = "0:a,2:b"}, %arg1: tensor<3x?x5xf32>{onnx.name = "y“, onnx.dim_params = "1:a"}) -> tensor<3x4x5xf32>
 ```
 This is a way to use existing node test for dynamic tensors. Since not all test case can pass with dynamic tensor, there is a list in test/backend/test.py, test_not_for_dynamic, to specify which test can not pass with `IMPORTER_FORCE_DYNAMIC` is defined.
 
--- a/docs/Testing.md
+++ b/docs/Testing.md
@@ -54,7 +54,6 @@ The onnx node tests usually have known dimension size for input tensors. So, to 
 IMPORTER_FORCE_DYNAMIC='-1:-1' all dimensions of all the inputs will be changed
 IMPORTER_FORCE_DYNAMIC='0:-1' all dimensions of the first input will be changed
 IMPORTER_FORCE_DYNAMIC='0:-1|1:0,1' all dimensions of the first input and the 1st and 2nd dimensions of the second input will be changed
-IMPORTER_FORCE_DYNAMIC='0:0=a,1=b,2=c|1:0=a,1=b,2=c' the first three dimensions of the first of and the second inputs are changed. And assume that the first dimensions of the first and second arguments are the same, and same for the the second and third dimensions.
 ```
 
 The Backus-Naur Form (BNF) for `IMPORTER_FORCE_DYNAMIC` is as follows.
@@ -64,11 +63,10 @@ The Backus-Naur Form (BNF) for `IMPORTER_FORCE_DYNAMIC` is as follows.
             <inputString ::= <inputIndex> `:` <dimString>
              <dimString> ::= <dimIndex> | <dimIndex> `,` <dimString>
             <inputIndex> ::= <index>
-              <dimIndex> ::= <index> | <index> '=' <symbol>
+              <dimIndex> ::= <index>
                  <index> ::= -1 | <number>
                 <number> ::= <digit> | <digit><number>
                  <digit> ::= 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9
-                <symbol> ::= 'a', 'b', 'c', ..., 'z'
 ```
 Value `-1` semantically represents all inputs or all dimensions, and it has the highest priority. E.g. `'0: -1, 0'` means all dimensions of the first input will be changed. Input and dimension indices start from 0.
 
@@ -84,12 +82,10 @@ with `IMPORTER_FORCE_DYNAMIC='0:-1'`, the result is:
 ```
 func @main_graph(%arg0: tensor<?x?x?xf32>, %arg1: tensor<3x4x5xf32>) -> tensor<3x4x5xf32>
 ```
-
-with `IMPORTER_FORCE_DYNAMIC='0:0=a,2=b|1:1=a'`, the result is:
+with `IMPORTER_FORCE_DYNAMIC='0:0,2|1:1'`, the result is:
 ```
-func @main_graph(%arg0: tensor<?x4x?xf32>{onnx.name = "x“, onnx.dim_params = "0:a,2:b"}, %arg1: tensor<3x?x5xf32>{onnx.name = "y“, onnx.dim_params = "1:a"}) -> tensor<3x4x5xf32>
+func @main_graph(%arg0: tensor<?x4x?xf32>, %arg1: tensor<3x?x5xf32>) -> tensor<3x4x5xf32>
 ```
-The onnx.dim_params attributes are used to specify the assumptions about dimensions of the first and second arguments.
 This is a way to use existing node test for dynamic tensors. Since not all test case can pass with dynamic tensor, there is a list in test/backend/test.py, test_not_for_dynamic, to specify which test can not pass with `IMPORTER_FORCE_DYNAMIC` is defined.
 
 ### Tests with constant inputs
--- a/docs/Testing.md
+++ b/docs/Testing.md
@@ -49,14 +49,12 @@ Testing with dynamic tensor sizes is most easily performed by using the followin
 cmake --build . --config Release --target check-onnx-backend-dynamic[-jni]
 ``` 
 
-The onnx node tests usually have known dimension size for input tensors. So, to test tensor with unknown dimension, the model importer (Build/FrontendONNXTransformer.cpp) provides a functionality to generate such cases. When the environment variable, `IMPORTER_FORCE_DYNAMIC`, is set, the frontend import will turn the all the dimensions (by default) of all the input tensors of the model into -1. As default, we assume that first dimension of all the inputs are the same, and the second dimension of all inputs are the same, and so on.
-
-For example,
+The onnx node tests usually have known dimension size for input tensors. So, to test tensor with unknown dimension, the model importer (Build/FrontendONNXTransformer.cpp) provides a functionality to generate such cases. When the environment variable, `IMPORTER_FORCE_DYNAMIC`, is set, the frontend import will turn the all the dimensions (by default) of all the input tensors of the model into -1. For example,
 ```
 IMPORTER_FORCE_DYNAMIC='-1:-1' all dimensions of all the inputs will be changed
 IMPORTER_FORCE_DYNAMIC='0:-1' all dimensions of the first input will be changed
 IMPORTER_FORCE_DYNAMIC='0:-1|1:0,1' all dimensions of the first input and the 1st and 2nd dimensions of the second input will be changed
-IMPORTER_FORCE_DYNAMIC='0:0:a,1:b,2:c|1:0:a,1:b,2:c' the first three dimensions of the first of and the second inputs are changed. And assume that the first dimensions of the first and second arguments are the same, and same for the the second and third dimensions.
+IMPORTER_FORCE_DYNAMIC='0:0=a,1=b,2=c|1:0=a,1=b,2=c' the first three dimensions of the first of and the second inputs are changed. And assume that the first dimensions of the first and second arguments are the same, and same for the the second and third dimensions.
 ```
 
 The Backus-Naur Form (BNF) for `IMPORTER_FORCE_DYNAMIC` is as follows.
@@ -66,11 +64,11 @@ The Backus-Naur Form (BNF) for `IMPORTER_FORCE_DYNAMIC` is as follows.
             <inputString ::= <inputIndex> `:` <dimString>
              <dimString> ::= <dimIndex> | <dimIndex> `,` <dimString>
             <inputIndex> ::= <index>
-              <dimIndex> ::= <index> | <index> ':' <symbol>
+              <dimIndex> ::= <index> | <index> '=' <symbol>
                  <index> ::= -1 | <number>
                 <number> ::= <digit> | <digit><number>
                  <digit> ::= 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9
-                <symbol> ::= 'a', 'b', 'c', ...
+                <symbol> ::= 'a', 'b', 'c', ..., 'z'
 ```
 Value `-1` semantically represents all inputs or all dimensions, and it has the highest priority. E.g. `'0: -1, 0'` means all dimensions of the first input will be changed. Input and dimension indices start from 0.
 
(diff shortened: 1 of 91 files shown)


docs/Testing.md
```diff
@@ -49,11 +49,14 @@ Testing with dynamic tensor sizes is most easily performed by using the followin
...
+The onnx node tests usually have known dimension size for input tensors. So, to test tensor with unknown dimension, the model importer (Build/FrontendONNXTransformer.cpp) provides a functionality to generate such cases. When the environment variable, `IMPORTER_FORCE_DYNAMIC`, is set, the frontend import will turn the all the dimensions (by default) of all the input tensors of the model into -1. As default, we assume that first dimension of all the inputs are the same, and the second dimension of all inputs are the same, and so on.
+
+For example,
 ```
 IMPORTER_FORCE_DYNAMIC='-1:-1' all dimensions of all the inputs will be changed
 IMPORTER_FORCE_DYNAMIC='0:-1' all dimensions of the first input will be changed
 IMPORTER_FORCE_DYNAMIC='0:-1|1:0,1' all dimensions of the first input and the 1st and 2nd dimensions of the second input will be changed
+IMPORTER_FORCE_DYNAMIC='0:0:a,1:b,2:c|1:0:a,1:b,2:c' the first three dimensions of the first of and the second inputs are changed. And assume that the first dimensions of the first and second arguments are the same, and same for the the second and third dimensions.
```
tungld, line 59: I know by using `:` we can pass `dimIndex` directly to an argument attribute, but I recommend to use a different symbol rather than `:` for `dimIndex`, because `:` has already been used to separate `inputIndex` and `dimString`. It confuses the parsers. How about `-` or `=`?

This example is quite trivial. Can we shuffle the symbols a bit, e.g. `0:0=a,1=b,2=c|1:0,1=c,2=b`, and we can have a mix of `<index>` and `<index>  '=' <symbol>`

docs/Testing.md
```diff
@@ -63,10 +66,11 @@ The Backus-Naur Form (BNF) for `IMPORTER_FORCE_DYNAMIC` is as follows.
             <inputString ::= <inputIndex> `:` <dimString>
              <dimString> ::= <dimIndex> | <dimIndex> `,` <dimString>
             <inputIndex> ::= <index>
-              <dimIndex> ::= <index>
+              <dimIndex> ::= <index> | <index> ':' <symbol>
                  <index> ::= -1 | <number>
                 <number> ::= <digit> | <digit><number>
                  <digit> ::= 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9
+                <symbol> ::= 'a', 'b', 'c', ...
```
tungld, line 73: I would add `'z'` at the end to finish the list though it is still no perfect, e.g. 
`<symbol> ::= 'a', 'b', 'c', ..., 'z'`

docs/Testing.md
```diff
@@ -76,15 +80,18 @@ func @main_graph(%arg0: tensor<3x4x5xf32>, %arg1: tensor<3x4x5xf32>) -> tensor<3
 ```
 with `IMPORTER_FORCE_DYNAMIC='-1:-1'`, the result is:
 ```
-func @main_graph(%arg0: tensor<?x?x?xf32>, %arg1: tensor<?x?x?xf32>) -> tensor<?x?x?xf32>
+func @main_graph(%arg0: tensor<?x?x?xf32>{onnx.name = "x“, onnx.dim_params = "0:a,1:b,2:c"}, %arg1: tensor<?x?x?xf32>{onnx.name = "y“, onnx.dim_params = "0:a,1:b,2:c"}) -> tensor<?x?x?xf32>
```
tungld, line 83: `-1` for `dimIndex` to say all arguments have the same dynamic dims does not make sense. It is too restricted.
Let's consider the dynamic tests for CPU where it supports broadcasting, e.g `add(tensor<3x5xf32>, tensor<3x1xf32>`, using `-1:-1` will force all dynamic dimensions being the same and the lowering will generate different code, not taking care of broadcasting.

In summary, I recommend to not support this, and let users specify the same dynamic dimensions manually for the cases they need.

tungld (commented):
Some first comments.

docs/Testing.md
negiyas: @tungld I changed the notation according to this comment. Thanks for the suggestions!

docs/Testing.md
negiyas: Fixed. Thanks.

docs/Testing.md
negiyas: I removed the assumption about default relationships among input data ranks, and specified the relationship manually in the test/accelerators/NNPA/backend/CMakeLists.txt file for each dynamic backend tests on NNPA.
Thanks for the suggestions.

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -1386,6 +1388,56 @@ class FrontendGenImpl {
...
+      assert((pos > 0) && "invalid IMPORTER_FORCE_DYNAMIC environment");
+      int idx = stoi(envStr.substr(0, pos));
+      envStr = envStr.substr(pos + 1);
+      if (idx < 0) { // set all arguments
+        for (size_t i = 0; i < numOfArgs; i++) {
+          getParamStr(envStr, paramStrVec[i]);
+        }
+      } else {
+        getParamStr(envStr, paramStrVec[idx]);
+      }
```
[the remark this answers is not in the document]
negiyas: Fixed. Thanks!

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -1386,6 +1388,56 @@ class FrontendGenImpl {
...
+    }
+    return;
+  }
+
+  // Get named attributesfrom IMPORTER_FORCE_DYNAMIC environment.
+  SmallVector<NamedAttribute> getNamedAttrFromImporterForceDynamic(
+      func::FuncOp funcOp, size_t numOfArgs) {
```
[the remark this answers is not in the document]
negiyas: This code was cleaned up, Thanks.

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -1386,6 +1388,56 @@ class FrontendGenImpl {
...
+    SmallVector<NamedAttribute> argNamedAttrs;
+    const char *envInputString = std::getenv("IMPORTER_FORCE_DYNAMIC");
+    if (envInputString == NULL)
+      return argNamedAttrs;
+    SmallVector<std::string> inputDimParamStrVec(numOfArgs);
+    getInputDimParamStrVec(
+        envInputString, funcOp, numOfArgs, inputDimParamStrVec);
+    for (size_t i = 0; i < numOfArgs; ++i) {
+      if (inputDimParamStrVec[i].length() != 0) {
```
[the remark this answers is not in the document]
negiyas: This code was cleaned up. Thanks.

src/Builder/FrontendDialectTransformer.hpp
```diff
@@ -55,6 +55,16 @@ struct ImportOptions {
   //   - (arg0: tensor<3x4x5xf32>, arg1: tensor<10x5xf32>)
   //
   std::string shapeInformation = "";
+  // Custom onnx.dim_params attributes for the graph inputs for specifying
+  // relationship among their ranks.
```
tungld, line 59: `their ranks` looks confusing. To be precise, it would be `their dynamic dimensions`, given that `dim` in `dim_params` means dimension.

src/Builder/FrontendDialectTransformer.hpp
```diff
@@ -55,6 +55,16 @@ struct ImportOptions {
...
+  // Its format is 'input_id:dim=sym,dim=sym,dim=sym|input_id:dim=sym,dim=sym..'
+  // E.g. An ONNX model has two dynamic inputs
+  //   - (arg0: tensor<?x5xf32>, arg1: tensor<?x5xf32>)
+  // If we want to specify that the first ranks of the first and second
```
tungld, line 63: `If we want to specify that the first ranks of the first and second arguments are the same` => `If we want to specify that the first unknown dimension of arg0 and the first unknown dimension of arg1 are the same, we can assign the two dimensions to the same symbol "batch" as follows.`

src/Compiler/CompilerOptions.cpp
```diff
@@ -281,6 +282,19 @@ static llvm::cl::opt<std::string, true> shapeInformationOpt("shapeInformation",
     llvm::cl::value_desc("value"), llvm::cl::location(shapeInformation),
     llvm::cl::cat(OnnxMlirOptions));
 
+static llvm::cl::opt<std::string, true> dimParamsOpt("dimParams",
+    llvm::cl::desc(
+        "Custom onnx.dim_params attributes for the inputs of the ONNX model for"
+        "specifying relationship among ranks of the inputs.\n"
```
tungld, line 288: `among ranks of` => `among dynamic dimensions of `

src/Compiler/CompilerOptions.cpp
```diff
@@ -281,6 +282,19 @@ static llvm::cl::opt<std::string, true> shapeInformationOpt("shapeInformation",
...
+        "\"value\" is in the format of "
+        "\"INPUT_ID1:D1=S1,D2=S2,...,Dn=Sn|INPUT_ID2:D1=T1,D2=T2,...Dn=Tn\""
+        "where \"INPUT_ID1, INPUT_ID2, ...\" are input indices (starting from "
+        "0 or being -1 for all input indices), and\n"
+        "\"S1, S2, ...\" and \"T2, T2, ...\" are symbols to specify that same "
+        "symbols have the same value"),
```
tungld, line 294: I would change this to make it more clear.

`
"\"S1, S2, ...\" and \"T2, T2, ...\" are symbols assigned to each unknown dimension. Two dimensions with the same symbols are expected to have the same value at runtime"),
`

BTW, what is the valid format of symbols? arbitrary string?

tungld: Should we say that `this custom onnx.dim_params will override the existing onnx.dim_params if exists`?

test/accelerators/NNPA/backend/CMakeLists.txt
```diff
@@ -93,7 +93,15 @@ endif()
 # Followings are test cases of test_to_enable_dict in test/backend/inference_backend.py.
 # Only test cases for operations supported by zDNN are listed. Non-supported cases are
 # commented out. This list is set in the environment variable `TEST_CASE_BY_USER`. So,
-# instruction name is added after test case name in each test case.
+# instruction name and dimParams are added after test case name if necessary.
+# For example, the following line
+#    test_add_cpu,zdnn_add,"0:0=a,1=b,2=c|1:0=a,1=b,2=c"
+# means that (1) the test name is "test_add_cpu", (2) instruction name is
```
tungld, line 99: `means that` => `consists of three parts`

test/accelerators/NNPA/backend/CMakeLists.txt
```diff
@@ -93,7 +93,15 @@ endif()
...
+# "zdnn_add" to check the function is called, (3) when dimParams is
+# "NO_DYNAMIC_SHAPE_TEST", backend test is skipped, otherwise the string is
+# passed as --dimParams option. "0:0=a,1=b,2=c|1:0=a,1=b,2=c" means that the
+# first ranks of the first, second and third ranks of the first and second
```
tungld, line 103: Could you make this explanation easier to follow? Don't use rank but dimension. Ranks in MLIR means the number of dimensions (in this case rank = 3).

test/backend/common.py
```diff
@@ -132,6 +139,7 @@ def compile_model(model, emit):
             "--constants-to-file-total-threshold="
             + str(args.constants_to_file_total_threshold)
         )
+    command_list += get_compile_option(name + "_cpu")
```
tungld, line 142: I would have a `if` directly here, it's easy to follow and faster:
```python
if args.dynamic and test_name in variables.test_to_enable_dimparams_dict:
        command_list.append("--dimParams=" + variables.test_to_enable_dimparams_dict[test_name])
```

test/mlir/onnx/parse/test_add_dimparams.onnxtext
```diff
@@ -0,0 +1,23 @@
+// RUN: onnx-mlir --EmitONNXBasic --dimParams="0:0=a,1=b,2=c|1:0=a,1=b,2=c" --printIR %s | FileCheck %s
```
tungld, line 1: Thank you for adding this test! Really appreciate it.

src/Builder/FrontendDialectTransformer.hpp
```diff
@@ -55,6 +55,16 @@ struct ImportOptions {
   //   - (arg0: tensor<3x4x5xf32>, arg1: tensor<10x5xf32>)
   //
   std::string shapeInformation = "";
+  // Custom onnx.dim_params attributes for the graph inputs for specifying
+  // relationship among their ranks.
+  // Its format is 'input_id:dim=sym,dim=sym,dim=sym|input_id:dim=sym,dim=sym..'
```
tungld, line 60: `dim_id` would be better than `dim`

tungld: `..` at the end would be three dots `...`? Also that `...` is for `|` meaning multiple `input_id:dim=sym,dim=sym,dim=sym`, or for `,` meaning multiple `dim=sym,dim=sym`?

src/Compiler/CompilerOptions.cpp
```diff
@@ -281,6 +282,19 @@ static llvm::cl::opt<std::string, true> shapeInformationOpt("shapeInformation",
     llvm::cl::value_desc("value"), llvm::cl::location(shapeInformation),
     llvm::cl::cat(OnnxMlirOptions));
 
+static llvm::cl::opt<std::string, true> dimParamsOpt("dimParams",
+    llvm::cl::desc(
+        "Custom onnx.dim_params attributes for the inputs of the ONNX model for"
+        "specifying relationship among ranks of the inputs.\n"
+        "\"value\" is in the format of "
+        "\"INPUT_ID1:D1=S1,D2=S2,...,Dn=Sn|INPUT_ID2:D1=T1,D2=T2,...Dn=Tn\""
```
tungld, line 290: missing a `,` before that last `Dn`.

It's better to add `| ...` at the end to denote multiple INPUT_ID, e.g. 
```
INPUT_ID1:D1=S1,D2=S2,...,Dn=Sn | INPUT_ID2:D1=T1,D2=T2,...,Dn=Tn | ...
```

tungld (commented):
Thank you for the update! The PR is in a good shape. I just put some comments to make the explanation easier to follow.

src/Builder/FrontendDialectTransformer.cpp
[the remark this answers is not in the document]
negiyas: Fixed. Thanks!

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -450,6 +452,27 @@ class FrontendGenImpl {
     return attributes;
   }
 
+  // Generate a string vector from the dimParams option string
+  void getInputDimParamsVecFromOption(std::string optionStr, size_t numOfArgs,
+      SmallVector<std::string> &paramStrVec) {
+    std::stringstream paramStrStream;
+    paramStrStream << optionStr;
```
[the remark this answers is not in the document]
negiyas: Fixed. Thanks!

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -490,6 +513,10 @@ class FrontendGenImpl {
     // See https://github.com/onnx/onnx/blob/main/docs/IR.md for more
     // information about dim_param.
     llvm::SmallVector<std::string, 4> inputDimParams, outputDimParams;
+    size_t numOfArgs = graph.input().size();
+    llvm::SmallVector<std::string> inputDimParamsFromOption(numOfArgs);
```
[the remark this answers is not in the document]
negiyas: Changed the code not to use `numOfArgs` in `inputDimParamsFromOption` . Thanks!

src/Builder/FrontendDialectTransformer.hpp
negiyas: Fixed. Thanks!

src/Compiler/CompilerOptions.cpp
negiyas: Fixed. Thanks.

test/accelerators/NNPA/backend/CMakeLists.txt
negiyas: Thanks for the comments. I fixed. I confirmed that all "rank" were removed modifications by this PR.

test/backend/common.py
negiyas: Fixed. Thanks!

src/Builder/FrontendDialectTransformer.hpp
negiyas: Thanks. Fixed.

src/Compiler/CompilerOptions.cpp
negiyas: Fixed. Thanks.

src/Compiler/CompilerOptions.cpp
negiyas: I updated the help messages to clarify the meaning of the option, and comments for handing this option in src/Builder/FrontendDialectTransformer.cpp . Please review them again. Thanks.

src/Builder/FrontendDialectTransformer.hpp
negiyas: Fixed. Thanks.

src/Compiler/CompilerOptions.cpp
```diff
@@ -281,6 +282,23 @@ static llvm::cl::opt<std::string, true> shapeInformationOpt("shapeInformation",
...
+        "\"value\" is in the format of "
+        "\"INPUT_ID1:D1=S1,D2=S2,...,Dn=Sn|INPUT_ID2:D1=T1,D2=T2,...Dn=Tn|"
+        "...\" where \"INPUT_ID1, INPUT_ID2, ...\" are input indices "
+        "(starting from 0 or being -1 for all input indices), and\n"
+        "\"S1, S2, ...\" and \"T2, T2, ...\" are symbols to specify that same "
+        "symbols have the same value. "
+        "All dimensions of onnx.dim_params for a specified input index in "
+        "the original onnx model are cleared and repalced by this option. "
+        "onnx.dim_params for other input indices in the original onnx model "
+        "are not cleared"),
```
tungld, line 298: Thank you for adding the explanation!

src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -450,6 +452,28 @@ class FrontendGenImpl {
...
+      size_t pos = dimParamStr.find(':');
+      assert((pos > 0) && "invalid dimParams option string");
+      int idx = stoi(dimParamStr.substr(0, pos));
+      dimParamStr = dimParamStr.substr(pos + 1);
+      std::replace(dimParamStr.begin(), dimParamStr.end(), '=', ':');
+      if (idx < 0) // set all arguments
+        paramStrForAllArgs = dimParamStr;
+      else {
+        while ((int)paramStrVec.size() <= idx) // Expand paramStrVec
+          paramStrVec.emplace_back("");
```
tungld, line 470: In the option definition you wrote `onnx.dim_params for other input indices in the original onnx model are not cleared`, but It looks to me only inputs whose index is larger than the max `idx` here are not cleared. And, the inputs whose index <= the max `idx` will be cleared.

For example, we have 4 input tensors 
```
arg0: tensor<?x?xf32> {onnx.dim_param=0:x},
arg1: tensor<?x?xf32> {onnx.dim_param=0:y},
arg2: tensor<?x?xf32>  {onnx.dim_param=0:z}
arg3: tensor<?x?xf32>  {onnx.dim_param=0:c,1:d}
```

with `--dimParams="0:0=b,1=a|2:0=a,1=b"`, it becomes

```
arg0: tensor<?x?xf32> {onnx.dim_param=0:b,1:a},
arg1: tensor<?x?xf32>,
arg2: tensor<?x?xf32>  {onnx.dim_param=0:a,1:b}
arg3: tensor<?x?xf32>  {onnx.dim_param=0:c,1:d}
```

Is it correct?

negiyas: @tungld Thanks for the comment.
 It is not correct. `onnx.dim_param` of arg1 is not cleared with `--dimParams="0:0=b,1=a|2:0=a,1=b"`.
```
       // For each input tensor, use either all dimensions by the compiler
       // option OR all dimensions in the original onnx model. Dimensions
       // from the option and the model in a single input tensor are not
       // merged.
       if (inputIndex < (int)inputDimParamsFromOption.size() &&
           !inputDimParamsFromOption[inputIndex].empty())
         inputDimParams.emplace_back(inputDimParamsFromOption[inputIndex]);
       else if (!inputDimParamsFromOptionForAllArgs.empty())
         inputDimParams.emplace_back(inputDimParamsFromOptionForAllArgs);
       else if (!dimParams.empty())
           inputDimParams.emplace_back(dimParams);
```
The code checks if `inputDimParamsFromOption[inputIndex]`  is empty string, and if it's empty, it uses `dimParams ` (=onnx.dim_param in the original model), so that onnx.dim_param attr of arg1 in the original model is not cleared.

I added `test/mlir/onnx/parse/dim_param_and_option.json` to verify this situation.
In the test, With  `--dimParams="1:1=b"`, onnx.dim_params="0:A,1:B" of the first input argument in the original model, is not cleared with the option.

Original model
```
module {
  func.func @main_graph(%arg0: tensor<?x?xf32> {onnx.dim_params = "0:A,1:B", onnx.name = "X"}, %arg1: tensor<?x?xf32> {onnx.dim_params = "0:A,1:B", onnx.name = "Y"}) -> (tensor<?x?xf32> {onnx.dim_params = "0:A,1:B", onnx.name = "Z"}) {
    %0 = "onnx.Add"(%arg0, %arg1) : (tensor<?x?xf32>, tensor<?x?xf32>) -> tensor<?x?xf32>
    onnx.Return %0 : tensor<?x?xf32>
  }
}
```
Imported model
```
module {
  func.func @main_graph(%arg0: tensor<?x?xf32> {onnx.dim_params = "0:A,1:B", onnx.name = "X"}, %arg1: tensor<?x?xf32> {onnx.dim_params = "1:b", onnx.name = "Y"}) -> (tensor<?x?xf32> {onnx.dim_params = "0:A,1:B", onnx.name = "Z"}) {
    %0 = "onnx.Add"(%arg0, %arg1) : (tensor<?x?xf32>, tensor<?x?xf32>) -> tensor<?x?xf32>
    onnx.Return %0 : tensor<?x?xf32>
  }
}
```

tungld: I see. Thank you for the confirmation! Just a suggestion: if you use a map instead of vector for `inputDimParamsFromOption`, the code is easier to read and you don't need to expand the vector and check `if (inputIndex < (int)inputDimParamsFromOption.size() && !inputDimParamsFromOption[inputIndex].empty())`. Vector data structure looks not ideal here.

tungld (approved):
LGTM.

src/Builder/FrontendDialectTransformer.cpp
negiyas: I changed the code to use `std::map<int, std::string>` instead of `SmallVector<std::string>`
Thanks for the suggestion. It make the code simpler.

AlexandreEichenberger:
@hamptonm1 @Sunny-Anand Suggestion is to create issues so that the root cause of the model features will be evaluated with the goal of increasing our share of model zoo  models successfully running with / without dynamic bounds.

PR #2782, opened 2024-04-03
Cherry-pick Add a flag to turn on/off the lowering of scalar broadcasting binary ops to NNPA

cjvolzka:
* Add a flag to turn on/off scalar broadcasting binary op in NNPA

---------

Co-authored-by: Alexandre Eichenberger <alexe@us.ibm.com>
(cherry picked from commit 08d4fed88105cd8bba1d4ce5da1cd0ccbbf1af0d)

### DIFF
--- a/src/Accelerators/NNPA/Compiler/NNPACompilerOptions.cpp
+++ b/src/Accelerators/NNPA/Compiler/NNPACompilerOptions.cpp
@@ -55,6 +55,13 @@ llvm::cl::opt<bool> nnpaEnableCompilerStickUnstick(
                    "stick/unstick code. Default is false."),
     llvm::cl::init(false), llvm::cl::cat(OnnxMlirOptions));
 
+llvm::cl::opt<bool> nnpaEnableScalarBcastBinary(
+    "nnpa-enable-scalar-bcast-binary",
+    llvm::cl::desc("Enable the lowering to NNPA the broadcasting binary ops "
+                   "whose one of the operands is scalar. Currently support "
+                   "ONNXDiv only. Default is false."),
+    llvm::cl::init(false), llvm::cl::cat(OnnxMlirCommonOptions));
+
 llvm::cl::opt<std::string> nnpaLoadDevicePlacementFile{
     "nnpa-load-device-placement-file",
     llvm::cl::desc(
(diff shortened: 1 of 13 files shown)


src/Accelerators/NNPA/Compiler/NNPACompilerOptions.cpp
```diff
@@ -55,6 +55,13 @@ llvm::cl::opt<bool> nnpaEnableCompilerStickUnstick(
                    "stick/unstick code. Default is false."),
     llvm::cl::init(false), llvm::cl::cat(OnnxMlirOptions));
 
+llvm::cl::opt<bool> nnpaEnableScalarBcastBinary(
+    "nnpa-enable-scalar-bcast-binary",
+    llvm::cl::desc("Enable the lowering to NNPA the broadcasting binary ops "
```
AlexandreEichenberger, line 60: I would suggest a clearer wording:

```
Enable the lowering to NNPA of binary operations with broadcasting of a scalar operand. Currently only enable ONNXDiv. Default is false.
```

AlexandreEichenberger (approved):
LGTM.

I assume it is to cover a regression. Are there examples where this optimization is worth it? And others where it is not? It feel that it could use a performance model. Note that the DIV op does also work well on CPU with parallel (as there is non-negligible amount of work).

src/Accelerators/NNPA/Compiler/NNPACompilerOptions.cpp
cjvolzka: > I assume it is to cover a regression. 

Yes. This was a cherry pick of https://github.com/onnx/onnx-mlir/pull/2778 into our upcoming 0.4.2.0 release. It resolves the performance regression for roberta-sequence-classification-9 from https://github.com/onnx/onnx-mlir/issues/2769.

tungld: > Are there examples where this optimization is worth it? And others where it is not? It feel that it could use a performance model.

I was effective to roberta-base, but later a rule to propagate Div to the inputs of MatMul can remove Div.
I don't have a model to which this feature is effective now, but in general, we want to offload a broadcasting div to NNPA only if it is surrounded by NNPA ops. 

I reworded the explanation in PR #2780.

PR #2783, opened 2024-04-03
Cleanup HeapReporter command line options

philass:
This PR uses `llvm::cl::list` to handle the parsing of a list of passes past as command line options. 

This is consistent with the `onnx-const-prop-disable-pattern` and `functions-to-decompose` implementations for list command line parsing.

### DIFF
--- a/src/Compiler/HeapReporter.cpp
+++ b/src/Compiler/HeapReporter.cpp
@@ -12,6 +12,8 @@
 
 #include "mlir/Pass/Pass.h"
 #include "mlir/Support/LLVM.h"
+
+#include "llvm/ADT/StringExtras.h"
 #include "llvm/Support/FileSystem.h"
 #include "llvm/Support/raw_ostream.h"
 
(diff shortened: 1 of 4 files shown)


src/Compiler/HeapReporter.cpp
```diff
@@ -48,15 +39,23 @@ void logMessage(StringRef logFilename, StringRef msg,
...
 
-HeapReporter::HeapReporter(
-    std::string logFilename, StringRef beforePasses, StringRef afterPasses)
+HeapReporter::HeapReporter(std::string logFilename,
+    std::vector<std::string> beforePasses, std::vector<std::string> afterPasses)
     : logFilename(logFilename) {
-  splitToSet(beforePasses, this->beforePassesSet);
-  splitToSet(afterPasses, this->afterPassesSet);
+  beforePassesSet = llvm::StringSet<>(beforePasses);
+  afterPassesSet = llvm::StringSet<>(afterPasses);
```
sorenlassen, line 46: can you set before/afterPassesSet in the initialization list?
```
  : logFilename(logFilename), beforePassesSet(beforePasses), afterPassesSet(afterPasses) {}
```

src/Compiler/HeapReporter.cpp
```diff
@@ -48,15 +39,23 @@ void logMessage(StringRef logFilename, StringRef msg,
...
+  for (std::string &s : beforePasses)
+    beforePassesStr = beforePassesStr + s + ",";
+  for (std::string &s : afterPasses)
+    afterPassesStr = afterPassesStr + s + ",";
+
   reportBegin("onnx-mlir heap report"
               "\n--report-heap-before='" +
-              beforePasses.str() + "'\n--report-heap-after='" +
-              afterPasses.str() + "'");
+              beforePassesStr + "'\n--report-heap-after=" + afterPassesStr);
```
sorenlassen, line 58: can you use llvm::join() from StringExtras.h?
```
llvm::join(beforePasses, ",") + "'\n--report-heap-after=" + llvm::join(afterPasses, ",")
```

sorenlassen (approved):

src/Compiler/HeapReporter.cpp
philass: Good catch

src/Compiler/HeapReporter.cpp
philass: I knew there must be some handy function for this lying around. Thanks for the pointer

PR #2785, opened 2024-04-04
[NNPA]Update condition to apply unstickStickRemoval optimization

imaihal:
It seems we need to check the operations are in the same block before using `isBeforeInBlock()`.

Example in MLIR: https://github.com/llvm/llvm-project/blob/main/mlir/lib/Conversion/PDLToPDLInterp/PDLToPDLInterp.cpp#L940-L941

Without this, we get following error.

```
mlir::operation::isBeforeInBlock(mlir::Operation*): Assertion `other && other->block == block && "Expected other operation to have the same parent block."' failed.
```


tungld (approved):
LGTM!

PR #2788, opened 2024-04-05
On z/OS import time.h in addition to sys/time.h

cjvolzka:
# Description
* On z/OS `CLOCKS_PER_SEC` is in time.h and `timeval` is in sys/time.sh so we need both imported.


tungld (approved):
LGTM. Thanks for fixing it for z/OS!

PR #2790, opened 2024-04-08
[Cherry-pick] On z/OS import time.h in addition to sys/time.h (#2788)

cjvolzka:
* On z/OS import time.h in addition to sys/time.h

# Description
* On z/OS `CLOCKS_PER_SEC` is in time.h and `timeval` is in sys/time.sh so we need both imported.

---------


cjvolzka:

PR #2794, opened 2024-04-10
Fixing the location of DimAnalysis in onnx-to-zhigh pass and some rules in zhigh-to-onnx pass

### LINKED ISSUE #2789: Output of roberta models varies between NNPA compiles
After compiling [roberta-sequence-classification-9.onnx](https://github.com/onnx/models/blob/main/validated/text/machine_comprehension/roberta/model/roberta-sequence-classification-9.onnx) and [roberta-base-11.onnx](https://github.com/onnx/models/blob/main/validated/text/machine_comprehension/roberta/model/roberta-base-11.onnx) for NNPA and running the generate .so. The model output can vary slightly.

For example, compiling and running roberta-sequence-classification-9 twice may yield
```
1st: [[ 0.455078, -0.35498 ]]
2nd: [[ 0.457520, -0.355957 ]]
3rd: [[ 0.458984, -0.357422 ]]
...
```
The values are always very close but I wouldn't expect any differences at all.

Notes:
* Don't seem to see the issue compiling for CPU
* Multiple runs of the same .so file yeilds consistent results. I only encounter the issue if I recompile and then re-run the model.
* I use the same version of onnx-mlir and the .onnx files for all tests (see test_robert.sh script)
* I don't think this is from a recent change. I've noticed random failures in our tolerance test in the past. However I couldn't reproduce the issue I investigated those times because I would re-run the same .so for extended periods and not observe issue. Since I couldn't reproduce it, I thought I just needed to loosen our tolerance test from commits over time. It seems to be happening more lately so I tried both re-compiling and re-running the model and found I can reproduce the issue readily.
* I'm attaching [roberta-out.zip](https://github.com/onnx/onnx-mlir/files/14889319/roberta-out.zip) with a script and our s390x test client which will show the issue. To run:
    *  Extract .zip to a folder
    * Download models from the onnx model zoo and put in the same folder
    * run `./test-roberta.sh roberta-sequence-classification-9`
        * Notice the `actual` column's output varies between compile and run cycles. (Note: `exp` column are the CPU expected values based on the example values in the onnx-model-zoo )
   *  run `./test-roberta.sh roberta-base-11`
        * There's many more output values for this model. So it's easier to look for lines like`Min passing r_tol:` and `Max absolute difference (Min passing a_tol for r_tol to be 0)` and notice it keeps shifting. Those are the minimum a_tol and r_tol values for the test to pass that run.

tungld:
The onnx-to-zhigh pass has two phases: 1) converting multiple onnx ops into a single zhigh op, and 2) converting a single onnx op to a single zhigh op, where the second phase uses DimAnalysis (Patterns in the 1st phase at this moment does not use DimAnalysis)

The problem is DimAnalysis is currently called before the 1st phase, which is not good because the 1st  phase may change the IR so the information from DimAnalysis is obsoleted to the 2nd phase. Correct position for DimAnalysis would be just before the 2nd phase.

Other than that, this PR changes slightly the rules in zhigh-to-onnx pass so that for binary ops, only one input (instead of two) that is from stick would be enough to trigger the rule to convert a zhigh op back to an onnx op.

Resolves #2789

### DIFF
--- a/src/Accelerators/NNPA/Conversion/ONNXToZHigh/ZHighToONNX.td
+++ b/src/Accelerators/NNPA/Conversion/ONNXToZHigh/ZHighToONNX.td
@@ -37,37 +37,52 @@ def CreateONNXMaxOp : NativeCodeCall<"$_builder.create<ONNXMaxOp>($_loc, $0.getT
 // ONNXAddOp %X = ZHighUnstickOp (ZHighAddOp (ZHighStickOp %X),
 // (ZHighStickOp %Y))
 //===----------------------------------------------------------------------===//
-def replaceZHighAddPattern : Pat<
-  (ZHighUnstickOp (ZHighAddOp (ZHighStickOp:$s_x $x, $_), (ZHighStickOp:$s_y $y, $_))),
-  (ONNXAddOp $x, $y),
-  [(NotBlockArgument:$x), (HasOneUse:$s_x), (HasOneUse:$s_y)]
+def replaceZHighAddPattern1 : Pat<
+  (ZHighUnstickOp (ZHighAddOp (ZHighStickOp:$s_x $x, $_), $y)),
+  (ONNXAddOp $x, (ZHighUnstickOp $y)),
+  [(NotBlockArgument:$x), (HasOneUse:$s_x)]
 >;
 
+def replaceZHighAddPattern2 : Pat<
+  (ZHighUnstickOp (ZHighAddOp $x, (ZHighStickOp:$s_y $y, $_))),
+  (ONNXAddOp (ZHighUnstickOp $x), $y),
+  [(NotBlockArgument:$y), (HasOneUse:$s_y)]
+>;
 
 //===----------------------------------------------------------------------===//
 // ONNXMulOp %X = ZHighUnstickOp (ZHighMulOp (ZHighStickOp %X),
 // (ZHighStickOp %Y))
 //===----------------------------------------------------------------------===//
 def replaceZHighMulPattern1 : Pat<
-  (ZHighUnstickOp (ZHighMulOp $x, (ZHighStickOp:$s_y $y, $_))),
-  (ONNXMulOp (ZHighUnstickOp $x), $y),
-  [(NotBlockArgument:$y), (HasOneUse:$s_y)]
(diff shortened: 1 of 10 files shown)


AlexandreEichenberger:
> , only one input (instead of two) that is from stick would be enough to trigger the rule to convert a zhigh op back to an onnx op.

Do you account for constants? Meaning do we do something different if the non-stickified input is a constant or not? As stickily of constants is "free"

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ZHighToONNX.td
```diff
@@ -37,62 +37,107 @@ def CreateONNXMaxOp : NativeCodeCall<"$_builder.create<ONNXMaxOp>($_loc, $0.getT
...
 // ONNXDivOp %X = ZHighUnstickOp (ZHighDivOp (ZHighStickOp
 // %X),(ZHighStickOp %Y))
 // Note: turn off this pattern since NNPA is faster at this moment.
 //===----------------------------------------------------------------------===//
-// def replaceZHighDivPattern : Pat<
-//   (ZHighUnstickOp (ZHighDivOp (ZHighStickOp:$s_x $x, $_), (ZHighStickOp:$s_y $y, $_))),
-//   (ONNXDivOp $x, $y),
-//   [(NotBlockArgument:$x), (HasOneUse:$s_x), (HasOneUse:$s_y)]
-// >;
+//def replaceZHighDivPattern1 : Pat<
```
AlexandreEichenberger, line 93: Did you check this one? DIV on CPU is quite a bit slower in general

AlexandreEichenberger (approved):
LGTM, maybe double check that the DIV performs better with this PR than before.

cjvolzka (approved):
I extensively tested the commit from yesterday. (200 iternationsw without output variations, not performance regression in about a dozen models tested).

I spot tested the lastest commit and confirmed after a few iterations roberta output is still stable.

I'm approving from my end but @AlexandreEichenberger  or someone else should still review from a code perspective.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ZHighToONNX.td
tungld: Yes, since DIV on CPU is quite a bit slower in general so I disabled this rule to keep Div on NNPA.

PR #2807, opened 2024-04-25
Add an option to disable op fusion in ONNXToKrnl pass

chentong319:
Add an option to disable op fusion in ONNXToKrnl pass. The command option is `--disable-krnl-op-fusion`.
The flag could be used for debugging and instrumentation.


AlexandreEichenberger (approved):
LGTM, thanks

PR #2815, opened 2024-05-07
Enable store-constants-to-file by default

tungld:
Many users of onnx-mlir are not aware of the option `--store-constants-to-file` that would be used for big models. Thus, this patch enables store-constants-to-file by default so that constants will be automatically stored on file whenever the total size of the constants is >= 1.5GB.

### DIFF
--- a/src/Compiler/CompilerOptions.cpp
+++ b/src/Compiler/CompilerOptions.cpp
@@ -337,15 +337,17 @@ static llvm::cl::opt<bool, true> storeConstantsToFileOpt(
     "store-constants-to-file",
     llvm::cl::desc(
         "Constants will be stored on a binary file instead of be embedded "
-        "into the model.so when the sizes of constants satisfy thresholds "
-        "defined by options constants-to-file-single-threshold and "
-        "constants-to-file-total-threshold. The binary file is in the same "
-        "folder as the model.so and has the same name as the model with the "
-        "extension of .constants.bin. For inference, model.constants.bin must "
-        "be at the same folder as the inference program. If "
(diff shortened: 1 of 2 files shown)


src/Compiler/CompilerOptions.cpp
```diff
@@ -337,13 +337,16 @@ static llvm::cl::opt<bool, true> storeConstantsToFileOpt(
...
-        "into the model.so. The binary file is in the same folder as the "
-        "model.so and has the same name as the model with the extension of "
-        ".constants.bin. For inference, model.constants.bin must be at the "
-        "same folder as the inference program. If model.constants.bin is at "
-        "another folder, use the environment variable OM_CONSTANT_PATH to set "
-        "the constant folder. Windows will be supported soon."),
-    llvm::cl::location(storeConstantsToFile), llvm::cl::init(false),
+        "into the model.so when the sizes of constants satisfy thresholds "
+        "defined by options constants-to-file-single-threshold and "
+        "constants-to-file-total-threshold. The binary file is in the same "
```
AlexandreEichenberger, line 342: Nit: I would add "When using the model, "... to specify when that co-location and/or env variable must be defined.

AlexandreEichenberger (approved):
LGTM

src/Compiler/CompilerOptions.cpp
tungld: I added this `When using this option, two other options constants-to-file-single-threshold and constants-to-file-total-threshold can be used to finetune the amount of constants stored on the file.`. Hope it's fine. Thanks!

tungld:

tungld:

PR #2816, opened 2024-05-07
Use macos-13 for Github Action MacOS

tungld:
The macos-latest version is macos 14 that removed python 3.10, so it failed to install protobuf 4.21.12 using python 3.11 or 3.12 in macos 14. So, specifically use macos-13 to avoid unexpected failures, and we would update the macos version at a good timing in the future.

### DIFF
--- a/.github/workflows/macos-amd64-build.yml
+++ b/.github/workflows/macos-amd64-build.yml
@@ -17,8 +17,7 @@ jobs:
         sudo rm -rf /Library/Frameworks/Python.framework/Versions/3* /usr/local/Frameworks/Python.framework/Versions/3* /usr/local/bin/python3* || true
     - name: install tools that are needed for compilation
       run: |
-        brew install automake ninja pybind11
-        sudo brew libtool || true
+        brew install automake libtool ninja pybind11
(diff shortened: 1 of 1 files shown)


tungld:

tungld:
@gongsu832 do you have any advice on this macos issue? I think macos-latest is now Sonoma 14, and python 3.10 is not available in the image (https://github.com/actions/runner-images/blob/main/images/macos/macos-14-Readme.md) so it failed to install protobuf 4.21.12.

gongsu832:
@tungld Debugging github action is very cumbersome and painful. If we don't require anything particular from macos-14, the simplest workaround for now is to just force macos-13, like you just did.

tungld:
@gongsu832 thanks! I am ok to go with macos-13. By the way, do you know what happened to Jenkins Linux s390x, it does not run.

tungld:

tungld:
> By the way, do you know what happened to Jenkins Linux s390x, it does not run.

It's running now.

gongsu832:
It ran out of disk space and was fixed.

Issue #2813, opened 2024-05-03
Poor NNPA performance on xlm-roberta-base-language-detection-onnx 


cjvolzka:
Running the [HuggingFace protectai/xlm-roberta-base-language-detection-onnx](https://huggingface.co/protectai/xlm-roberta-base-language-detection-onnx/tree/main) has lower than expected NNPA performance for large inputs.

Compiling [model.onnx](https://huggingface.co/protectai/xlm-roberta-base-language-detection-onnx/blob/main/model.onnx) with  `--O3 --EmitLib --mtriple=s390x-ibm-loz --mcpu=z16 --maccel=NNPA model.onnx --onnx-op-stats TXT --profile-ir=ZHigh` yields:
```
Operations encountered:
-----------------------
   func.func               , 1
   func.return             , 1
   krnl.runtime_instrument , 2204
   onnx.Add                , 51
   onnx.Cast               , 3
   onnx.Concat             , 109
   onnx.Constant           , 71
   onnx.CumSum             , 1
   onnx.Dim                , 170
   onnx.Div                , 12
   onnx.Equal              , 2
   onnx.Erf                , 12
   onnx.Expand             , 1
   onnx.Gather             , 4
   onnx.LayerNormalization , 25
   onnx.MatMul             , 12
   onnx.Max                , 12
   onnx.Mul                , 26
   onnx.NoValue            , 1
   onnx.Not                , 1
   onnx.Reshape            , 108
   onnx.Slice              , 1
   onnx.Sub                , 1
   onnx.Transpose          , 48
   onnx.Unsqueeze          , 2
   onnx.Where              , 1
  zhigh.MatMul             , 86
  zhigh.Softmax            , 12
  zhigh.Stick              , 85
  zhigh.StickifiedConstant , 148
  zhigh.Tanh               , 1
  zhigh.Unstick            , 97
```

The performance issue primarily appears to be the 12 onnx.MatMul operators. On large inputs (first in
[Input Examples.txt](https://github.com/onnx/onnx-mlir/files/15203148/Input.Examples.txt)), this op dominates runtime. 

The operators are:
```
                      'MatMul': {'count': 12,
                                 'nodes': {'/roberta/encoder/layer.0/attention/self/MatMul_1': 0.064281,
                                           '/roberta/encoder/layer.1/attention/self/MatMul_1': 0.064978,
                                           '/roberta/encoder/layer.10/attention/self/MatMul_1': 0.064231,
[...]

PR #2819, opened 2024-05-08
[NNPA] An option to report why the operations are not run on NNPA (--opt-report=NNPAUnsupportedOps)

imaihal:
Generate report on why the operations are not run on NNPA. 
The report is displayed by using `--opt-report=NNPAUnsupportedOps`. The format is 
`==NNPA-UNSUPPORTEDOPS-REPORT==, <operation name>, <node name>, <reason>`

```
  --opt-report=<value>                                                              - Provide information on a specific compiler optimization.
    =NoReport                                                                       -   No report. Default value.
    =Parallel                                                                       -   Provide report on how OMP Parallel is applied to ONNX ops.
    =Simd                                                                           -   Provide report on how SIMD is applied to ONNX ops.
    =NNPAUnsupportedOps  
```

Example of xlm-roberta-base-language-detection-onnx  (issue https://github.com/onnx/onnx-mlir/issues/2813): 
```
$ onnx-mlir --O3 --mtriple=s390x-ibm-loz --mcpu=z16 --maccel=NNPA model.onnx  --EmitZHighIR --opt-report=NNPAUnsupportedOps

==NNPA-UNSUPPORTEDOPS-REPORT==, onnx.Mul, Mul_28, Element type is not F16 or F32.
==NNPA-UNSUPPORTEDOPS-REPORT==, onnx.Add, Add_31, Element type is not F16 or F32.
==NNPA-UNSUPPORTEDOPS-REPORT==, onnx.Div, Div_122, Rank 0 is not supported. zAIU only supports rank in range of (0, 4].
==NNPA-UNSUPPORTEDOPS-REPORT==, onnx.Add, Add_125, Rank 0 is not supported. zAIU only supports rank in range of (0, 4].
==NNPA-UNSUPPORTEDOPS-REPORT==, onnx.Mul, Mul_128, Rank 0 is not supported. zAIU only supports rank in range of (0, 4].
==NNPA-UNSUPPORTEDOPS-REPORT==, onnx.Div, Div_216, Rank 0 is not supported. zAIU only supports rank in range of (0, 4].
==NNPA-UNSUPPORTEDOPS-REPORT==, onnx.Add, Add_219, Rank 0 is not supported. zAIU only supports rank in range of (0, 4].
==NNPA-UNSUPPORTEDOPS-REPORT==, onnx.Mul, Mul_222, Rank 0 is not supported. zAIU only supports rank in range of (0, 4].
==NNPA-UNSUPPORTEDOPS-REPORT==, onnx.Div, Div_310, Rank 0 is not supported. zAIU only supports rank in range of (0, 4].
==NNPA-UNSUPPORTEDOPS-REPORT==, onnx.Add, Add_313, Rank 0 is not supported. zAIU only supports rank in range of (0, 4].
==NNPA-UNSUPPORTEDOPS-REPORT==, onnx.Mul, Mul_316, Rank 0 is not supported. zAIU only supports rank in range of (0, 4].
==NNPA-UNSUPPORTEDOPS-REPORT==, onnx.Div, Div_404, Rank 0 is not supported. zAIU only supports rank in range of (0, 4].
==NNPA-UNSUPPORTEDOPS-REPORT==, onnx.Add, Add_407, Rank 0 is not supported. zAIU only supports rank in range of (0, 4].
==NNPA-UNSUPPORTEDOPS-REPORT==, onnx.Mul, Mul_410, Rank 0 is not supported. zAIU only supports rank in range of (0, 4].


:
:
```


src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
```diff
@@ -266,6 +269,11 @@ bool meetPoolParamRestrictions(int64_t inputShape, int64_t kernelShape,
   return true;
 }
 
+void emitLegalityCheckMessage(Operation *op, StringRef message) {
+  LLVM_DEBUG(llvm::outs() << "[NNPA Legality Check] Warning: " << op->getLoc()
+                          << " runs on CPU. Reason: " << message << "\n");
+}
```
tungld, line 275: Could you print out the operation name also? Sometime the loc is just a number and it's not easy to know which operation is.

tungld (approved):
LGTM! It's useful. Thanks for adding this!

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
imaihal: Added operation name and node name. Thanks!

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
```diff
@@ -518,21 +542,34 @@ bool isSuitableForZDNN<ONNXMatMulOp>(
...
   if ((shapeA.size() == 2) && (shapeB.size() == 2)) {
     // unstacked case
-    if (aType.hasStaticShape() && bType.hasStaticShape())
-      return (shapeA[1] == shapeB[0]);
-    else
+    if (aType.hasStaticShape() && bType.hasStaticShape()) {
+      bool returnVal = (shapeA[1] == shapeB[0]);
+      if (!returnVal)
+        emitLegalityCheckMessage(
+            op.getOperation(), "Unstacked case, dim A 1 not equal dim B 0.");
```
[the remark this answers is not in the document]

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
```diff
@@ -62,10 +65,25 @@ bool isValidElementTypeAndRank(Value val, bool donotCheckRank) {
...
+        std::string message =
+            "Rank " + std::to_string(rank) +
+            " is not supported. zAIU only supports rank in range of (0, 4].";
+        emitLegalityCheckMessage(op, StringRef(message));
+        return false;
+      }
+    } else {
+      std::string message = "Element type is not F16 or F32.";
+      emitLegalityCheckMessage(op, StringRef(message));
+      return false;
```
tungld, line 86: I recommend to change `emitLegalityCheckMessage` to return `false`, e.g `bool emitLegalityCheckMessage(Operation *op, StringRef message)`, and then we can write:
```
return emitLegalityCheckMessage(op, StringRef(message));
```

imaihal: Done. Thanks. Also changed the function name.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
```diff
@@ -62,13 +65,26 @@ bool isValidElementTypeAndRank(Value val, bool donotCheckRank) {
...
-        return false;
+      if (!valueType.hasRank()) {
+        std::string message = "Value does not have rank.";
+        return emitWarningMessageNNPAUnsupported(op, StringRef(message));
+      }
       int64_t rank = valueType.getRank();
-      return ((rank > 0) && (rank <= 4));
+      if ((rank > 0) && (rank <= 4)) {
+        return true;
+      } else {
```
tungld, line 75: Since we have `return` in the `if` part, the `else` is not necessary. For example:

```c
if ((rank > 0) && (rank <= 4))
     return true;

std::string message =
            "Rank " + std::to_string(rank) +
            " is not supported. zAIU only supports rank in range of (0, 4].";
return emitWarningMessageNNPAUnsupported(op, StringRef(message));
```

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
```diff
@@ -231,41 +267,98 @@ StringRef getStrPaddingType(OP op) {
...
+                              "kernelShape + 1) / strides)) (= " +
+                              std::to_string(reqOutputShape) + ").";
+        return emitWarningMessageNNPAUnsupported(op, StringRef(message));
+      }
     }
   }
   return true;
 }
 
+bool emitWarningMessageNNPAUnsupported(Operation *op, StringRef message) {
```
tungld, line 345: Do we really need to use StringRef instead of just `const std::string&`? I don't see any operation on `message` and passing a pointer looks faster than constructing a StringRef.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
```diff
@@ -623,19 +625,35 @@ void getRewriteONNXForZHighDynamicallyLegal(
...
+                           std::to_string(i) + " of A and dim " +
+                           std::to_string(i) + " of B are the same.";
+            }
           }
-          return !sameBatchDims;
+          if (!sameBatchDims)
+            return !emitWarningMessageNNPAUnsupported(
+                op.getOperation(), message); /* true */
+
+          return false;
```
tungld, line 656: This can be shorten in this way:
```c
return (!sameBatchDims) || emitWarningMessageNNPAUnsupported(op.getOperation(), message);
```

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
imaihal: Updated. Thanks!

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
imaihal: Updated. Thanks!

imaihal:
@tungld I created an option for this reporting, not using `--debug-only`. This is the same approach as `--opt-report` option. Is this acceptable?

tungld:
> @tungld I created an option for this reporting, not using --debug-only. This is the same approach as --opt-report option. Is this acceptable?

If you take the approach of `--opt-report`, I recommend to add an enum value into the existing `--opt-report`, say `--opt-report=NNPA`. You can see `--profile-ir` to know how to mix CPU and NNPA values in a single option.

imaihal:
> If you take the approach of `--opt-report`, I recommend to add an enum value into the existing `--opt-report`,

I updated to use `--opt-report` adding new enum value (NNPAUnsupportedOps). Thanks!

tungld:
> I updated to use --opt-report adding new enum value (NNPAUnsupportedOps). Thanks!

When it's ready for another round of review, please let me know. Thanks!

imaihal:
Yes. ready for review.

src/Accelerators/Accelerator.hpp
```diff
@@ -51,6 +51,10 @@
 
 #define ACCEL_PROFILEIR_CL_ENUM(name) PROFILEIR_CL_ENUM_##name
 
+#define ACCEL_OPTREPORT_ENUM(name) OPTREPORT_ENUM_##name
+
+#define ACCEL_OPTREPORT_CL_ENUM(name) OPTREPORT_CL_ENUM_##name
```
tungld, line 56: Could you update [this document](https://github.com/onnx/onnx-mlir/blob/main/docs/AddCustomAccelerators.md#21-macro) to include the new macro? Thanks!

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.hpp
```diff
@@ -43,5 +44,11 @@ mlir::StringRef getStrPaddingType(OP op);
...
-bool meetPoolParamRestrictions(int64_t inputShape, int64_t kernelShape,
-    int64_t strides, int64_t outputShape, mlir::StringRef paddingType);
+bool meetPoolParamRestrictions(mlir::Operation *op, int64_t inputShape,
+    int64_t kernelShape, int64_t strides, int64_t outputShape,
+    mlir::StringRef paddingType);
+
+bool emitWarningMessageNNPAUnsupported(
+    mlir::Operation *op, const std::string &message);
+
+bool emitWarningMessageInCompatibility(mlir::Operation *op);
```
tungld, line 54: Would you mind if we change these function names to make it consistent with functions for other `--op-report` options? We have `onnxToKrnlSimdReport` and `onnxToKrnlParallelReport`,  so how about `onnxToZHighUnsupportedReport`? or feel free to choose another name.

tungld (commented):
Thanks for updating the patch. It's much better. Just some minor comments, but the most important thing is updating the document. And, feel free to merge the patch then.

In your example of roberta, I see this
```
==NNPA-UNSUPPORTEDOPS-REPORT== onnx.Softmax, /roberta/encoder/layer.0/attention/self/Softmax_176, The rank of input tensor (4) needs to be 2 or 3.
```

It seemed it was not applied in this model. Do you know what was the reason? Softmax with rank 4 is potentially rewritten to rank 2 by RewriteONNXForZHigh pass. Perhaps `axis` was not the last dimension in this model ...

AlexandreEichenberger (commented):
Could you add a `,`  after the option identifier, namely transforming 
```
==NNPA-UNSUPPORTEDOPS-REPORT== onnx.Mul, /roberta/embeddings/Mul, Element type is not F16 or F32.
```

to
```
==NNPA-UNSUPPORTEDOPS-REPORT==,  onnx.Mul, /roberta/embeddings/Mul, Element type is not F16 or F32.
```

so that is is like all of the other perf measurements; it let you do CSV imports more easily.

Great addition, thanks.

AlexandreEichenberger (approved):
Approving the PR, but please add the missing comma in the printout to be compatible with the other tools.

imaihal:
@tungld 
> 
> In your example of roberta, I see this
> 
> ```
> ==NNPA-UNSUPPORTEDOPS-REPORT== onnx.Softmax, /roberta/encoder/layer.0/attention/self/Softmax_176, The rank of input tensor (4) needs to be 2 or 3.
> ```
> 
> It seemed it was not applied in this model. Do you know what was the reason? Softmax with rank 4 is potentially rewritten to rank 2 by RewriteONNXForZHigh pass. Perhaps `axis` was not the last dimension in this model ...

I confirmed all softmax ops are lowered to zhigh.softmax in ZHighIR. 

I found this message was generated in `DevicePlacementPass` which also calls legality check. In the `DevicePlacementPass`, is `ONNXToZHighPass` called after `RewriteONNXForZHighPass`? If not and `ONNXToZHighPass` check original softmax op, I think this message is generated.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.hpp
imaihal: Done. Thanks!

imaihal:
> Could you add a `,` after the option identifier, namely transforming
> 
> ```
> ==NNPA-UNSUPPORTEDOPS-REPORT== onnx.Mul, /roberta/embeddings/Mul, Element type is not F16 or F32.
> ```
> 
> to
> 
> ```
> ==NNPA-UNSUPPORTEDOPS-REPORT==,  onnx.Mul, /roberta/embeddings/Mul, Element type is not F16 or F32.
> ```
 Done. Thanks!

tungld:
> I confirmed all softmax ops are lowered to zhigh.softmax in ZHighIR.

So the printed message was incorrect. I think you would fix that before merging.

> is ONNXToZHighPass called after RewriteONNXForZHighPass?

Yes, it is in the code, but they are independent calls and used for analysis. The results of each call will be merged later. You can take a look at: https://github.com/onnx/onnx-mlir/blob/main/src/Accelerators/NNPA/Conversion/ONNXToZHigh/DevicePlacement.cpp#L196

imaihal:
@tungld

Thanks for the comment!. I updated the PR to enable the reporting only in ONNXToZHighPass and disable it in DevicePlacementPass. This removes reporting about Softmax, as described in updated description of this PR.

tungld:
>I updated the PR to enable the reporting only in ONNXToZHighPass and disable it in DevicePlacementPass. This removes reporting about Softmax, as described in updated description of this PR.

Thank you for the great work! Please merge it at your convenience time.

imaihal:

imaihal:

PR #2823, opened 2024-05-09
Add more dynamic bakend tests on NNPA

negiyas:
Update test/accelerators/NNPA/backend/CMakeLists.txt to add more dynamic backend tests on NNPA.


imaihal (approved):
LGTM. 

Regarding ConvTranspose op, tests for dynamic dimension is disabled even for CPU because spatial dims must be static.

PR #2825, opened 2024-05-13
Added support for plot

AlexandreEichenberger:
Added the possibility to generate a jpg with the plot of the result of the analysis of a run.

Recall how to gather data of a run (here run Roberta, gather compile time info on parallel (`-opt-report=Parallel`, saved with `--log-to-file compile.log` ) and runtime info (`--profile-ir=ZHigh` saved with `ONNX_MLIR_INSTRUMENT_FILE=run.log`).

```bash
ONNX_MLIR_INSTRUMENT_FILE=run.log RunONNXModel.py -m roberta-base-11.onnx -c "-O3 -mcpu=z16 -maccel=NNPA -parallel -opt-report=Parallel --profile-ir=ZHigh" --shape-info=0:6x384  --log-to-file compile.log -w 3 -n 16
```

Then you may view and print statistics as follow
```bash
python3 ~/OM/onnx-mlir/utils/make-report.py -ums -r run.log -w 3 -p output -ums
```
where the raw data is taken from a run-file named `run.log` skipping the first 3 warmup runs (`-w 3`) with output in miliseconds (`-ums` for unit in ms). The resulting graph is:

image

Say you would like to focus the results on operations that take at least 1% of execution time, and report the ops in alphabetical order (default is by decreasing order of execution time). Add the `-m 1 --sort name` and you get this graph

image

Say you now want to correlate the runtime info with statistics of operations that ran in parallel or not. Update the command to provide the compile-time log:
```bash
python3 ~/OM/onnx-mlir/utils/make-report.py -ums -r run.log -c compile.log -s par -p output -m 1 --sort name
```

image


utils/make-report.py
```diff
@@ -57,11 +57,23 @@ def print_usage(msg=""):
...
-Parameters:
+Parameters on inputs:
   -c/--compile <file_log>: File name containing the compile-time statistics
                            or runtime signature statistics.
   -r/--runtime <file_log>: File name containing the runtime time statistics.
 
+Parameters to focus analysis:
+  -f/--focus <regexp>: Focus only on ops that match the regexp pattern.
+  -m/--min <num>:      Focus on operations with at least <num>% of exec time.
+  -supported:          Focus only on ops that are supported. Namely, the report
```
cjvolzka, line 68: nit: Typo, should be `--supported:`

cjvolzka (approved):
One typo but otherwise looks great. Thanks for adding the plots!

utils/make-report.py
AlexandreEichenberger: nice catch, thanks

PR #2826, opened 2024-05-14
Mismatched output type when rewriting Reshape

tungld:
I got the following error when compiling the gpt2 kv-cache model from HuggingFace:
```
loc("/transformer/Reshape_3"): error: failed to materialize conversion for result #0 of operation 'onnx.Reshape' that remained live after conversion
```

It's related to a canonicalization rule to replace the following Reshape:
```
%1410 = "onnx.Reshape"(%Y_69, %2) {allowzero = 0 : si64, device = "nnpa", onnx_node_name = "/transformer/Reshape_3"} : (tensor<?x?x768xf32>, tensor<3xi64>) -> tensor<?x1x768xf32>
```

`tensor<?x1x768xf32>` is replaced by `tensor<?x?x768xf32>` (because the 1st unknown dims are the same by using DimAnalysis, so the second dims are "we don't care") which caused the output type mismatched and then failed to materialize the output.

To fix that issue, this patch adds a condition to make sure the input and output dims at an axis must be both static or unknown.

### DIFF
--- a/src/Dialect/ONNX/ONNXOps/OpHelper.cpp
+++ b/src/Dialect/ONNX/ONNXOps/OpHelper.cpp
@@ -742,20 +742,31 @@ bool isIdentityReshape(
     Value inputTensor, Value outputTensor, const DimAnalysis *dimAnalysis) {
   if (!hasShapeAndRank(inputTensor) || !hasShapeAndRank(outputTensor))
     return false;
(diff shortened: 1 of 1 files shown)


AlexandreEichenberger (approved):
Good catch, LGTM

PR #2827, opened 2024-05-15
[NNPA] Recompose ops back to zhigh stick and unstick if they are not optimized by decomposition

tungld:
When using `--enable-zhigh-decompose-stick-unstick` to decompose stick and unstick into LayoutTransform, DLF16ToF32 and F32ToDLF16 ops, some optimizations are applied to remove DLF16ToF32 and F32ToDLF16 ops. But when they are not removed, we would recompose back to zhigh stick and unstick to utilize the optimized code generation for zhigh stick and unstick.

This patch adds a pass for recomposition. The pass is enabled if we compile a model with the option `--enable-zhigh-decompose-stick-unstick`.

Below is the result for roberta-base, where the number of stick and unstick dramatically decreased.

**without --enable-zhigh-decompose-stick-unstick**
Screenshot 2024-05-15 at 14 44 24

**with --enable-zhigh-decompose-stick-unstick**
Screenshot 2024-05-15 at 14 44 57


AlexandreEichenberger (approved):
LGTM, thanks

PR #2828, opened 2024-05-15
Forward input dims in Reshape ShapeHelper

tungld:
In this reshape op in a quantized bert model, the first two dimensions of the output are the same as the ones in the input, so we would like to forward IndexExprs of these dimensions from the input to the output so that DimAnalysis works. But it failed because of using `IndexExpr.selectOrSelf`. 
```
%184 = onnx.Constant dense<[0, 0, 12, 64]> : tensor<4xi64>
%494 = "onnx.Reshape"(%491, %184) {allowzero = 0 : si64, onnx_node_name = "/bert/encoder/layer.0/attention/self/Reshape"} : (tensor<?x?x768xf32>, tensor<4xi64>) -> tensor<?x?x12x64xf32>
```
This patch updates `selectOrSelf` to forward the input QuestionMark when the condition is literal.


tungld:
@AlexandreEichenberger if you know how to modify `selectOrSelf` to fix this issue, please let me know or please feel free to create a PR for that. Thanks!

src/Dialect/ONNX/ONNXOps/Tensor/Reshape.cpp
```diff
@@ -83,8 +83,9 @@ LogicalResult ONNXReshapeOpShapeHelper::computeShape() {
   // should be -1 (represented as QuestionmarkIndexExpr)
   for (unsigned i = 0; i < outputRank; ++i) {
     if (hasShapeAndRank(data)) {
+      IndexExpr dimShape = createIE->getIntFromArrayAsSymbol(shape, i);
       outputDims[i] = outputDims[i].selectOrSelf(
-          outputDims[i] == -1, numOfElements.floorDiv(numOfElementsFromShape));
+          dimShape == -1, numOfElements.floorDiv(numOfElementsFromShape));
```
tungld, line 88: I changed this to make the compare expression literal.

tungld:
It looks like I found the way to change `selectOrSelf` but it only works if the condition is literal at compile time.

AlexandreEichenberger:
@tungld  is there still a case to look into?

tungld:
> @tungld is there still a case to look into?

I changed `selectOrSelf` to handle the case where the comparison is literal. It works for the quantized bert model. I hope it's enough for most cases.

tungld:
@AlexandreEichenberger It's ready for review.

PR #2829, opened 2024-05-16
Update zdnn.cmake

hamptonm1:
If we still want to skip autoconf and config if the files already exist, I can add a switch but I figured this update is so small... it is not needed.


hamptonm1:
FYI- @tungld

gongsu832 (approved):
LGTM

Issue #2817, opened 2024-05-08
Dynamic backend tests on NNPA cause errors


negiyas:
Dynamic backend tests on NNPA cause runtime errors. (c.f. This Issue is similar to Issue #2743 "Dynamic backend tests cause errors on CPU," but this issue for errors on NNPA.)

PR #2781 introduces "--dimParams" option for specifying relationships among dimensions of model inputs. It also enables dynamic backend tests on NNPA with the --dimParams option, but some tests cannot be enabled because of runtime errors, and they are temporally disabled by the PR. We need to fix these runtime issues or confirm that they are reasonable.

The following is a list of backend tests on NNPA not causing errors with static shapes, but causing errors with dynamic shapes. Please fix errors or confirm that they are reasonable. Thanks.

[ ] test_averagepool_2d_default_cpu,zdnn_avgpool2d,NO_DYNAMIC_SHAPE_TEST
[ ] test_averagepool_2d_precomputed_pads_cpu,zdnn_avgpool2d,NO_DYNAMIC_SHAPE_TEST
[ ] test_averagepool_2d_precomputed_same_upper_cpu,zdnn_avgpool2d,NO_DYNAMIC_SHAPE_TEST
[ ] test_averagepool_2d_precomputed_strides_cpu,zdnn_avgpool2d,NO_DYNAMIC_SHAPE_TEST
[ ] test_averagepool_2d_same_upper_cpu,zdnn_avgpool2d,NO_DYNAMIC_SHAPE_TEST
[ ] test_averagepool_2d_strides_cpu,zdnn_avgpool2d,NO_DYNAMIC_SHAPE_TEST
[ ] test_batchnorm_epsilon_cpu,zdnn_mul,NO_DYNAMIC_SHAPE_TEST
[ ] test_batchnorm_example_cpu,zdnn_mul,NO_DYNAMIC_SHAPE_TEST
[ ] test_basic_conv_with_padding_cpu,zdnn_conv2d,NO_DYNAMIC_SHAPE_TEST
[ ] test_basic_conv_without_padding_cpu,zdnn_conv2d,NO_DYNAMIC_SHAPE_TEST
[ ] test_conv_with_strides_no_padding_cpu,zdnn_conv2d,NO_DYNAMIC_SHAPE_TEST
[ ] test_conv_with_strides_padding_cpu,zdnn_conv2d,NO_DYNAMIC_SHAPE　_TEST
[...]

PR #2831, opened 2024-05-21
[NNPA] Update configurations for test cases of backend test for dynamic dimensions

imaihal:
As written in issue https://github.com/onnx/onnx-mlir/issues/2817,  some backend tests with dynamic shapes for NNPA was disabled by using`NO_DYNAMIC_SHAPE_TEST` flag. 
This PR enables some of them by fixing legality check and using `--dimParam` option. There are still test cases with `NO_DYNAMIC_SHAPE_TEST`. 

Enabled 22 test cases in dynamic backend tests. (test cases with NO_DYNAMIC_SHAPE_TEST from 30 to 8)


src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
```diff
@@ -103,7 +103,8 @@ bool checkLegalityPoolOpsCommon(POOLOP op, Value Y) {
   // Check "MaxPool2D/AvgPool2D Parameter Restrictions". These restrictions are
   // described in "zDNN API Reference". Input tensor N(batchNum) and C(Channel)
   // dimensions must always match the output tensor's respective dimensions.
-  if (shapeInput[0] != shapeOutput[0] || shapeInput[1] != shapeOutput[1])
+  if ((inputType.hasStaticShape() && outputType.hasStaticShape()) &&
+      (shapeInput[0] != shapeOutput[0] || shapeInput[1] != shapeOutput[1]))
```
tungld, line 107: Can we use DimAnalysis to check dynamic dims here?

test/accelerators/NNPA/backend/CMakeLists.txt
```diff
@@ -152,13 +152,13 @@ set(NNPA_TEST_LIST
     # Disable backend tests for ConvTranspose, because they do not work on CPU.
     # test_convtranspose_1d_cpu,zdnn_conv1d
     # test_convtranspose_3d_cpu
-    test_convtranspose_autopad_same_cpu,zdnn_conv2d,NO_DYNAMIC_SHAPE_TEST # not work on CPU
-    test_convtranspose_cpu,zdnn_conv2d,NO_DYNAMIC_SHAPE_TEST # not work on CPU
+    test_convtranspose_autopad_same_cpu,zdnn_conv2d,NO_DYNAMIC_SHAPE_TEST # Spatial dims must be static even on CPU.
```
tungld, line 155: Can we use `DYNAMIC_SHAPE` to specify which dims are dynamic on CPU? so that spatial dims are always static.

test/accelerators/NNPA/backend/CMakeLists.txt
```diff
@@ -191,8 +191,8 @@ set(NNPA_TEST_LIST
     # ==OP== GlobalAveragePool
     # ==MIN== 1
     # ==LIM== - Input shape must be 4D tensor(NCHW).<br>- Dimensions in `H` and `W` must be static.
-    test_globalaveragepool_cpu,zdnn_meanreduce2d,NO_DYNAMIC_SHAPE_TEST
-    test_globalaveragepool_precomputed_cpu,zdnn_meanreduce2d,NO_DYNAMIC_SHAPE_TEST
+    test_globalaveragepool_cpu,zdnn_meanreduce2d,NO_DYNAMIC_SHAPE_TEST # Height and Width dimension must be static dimension.
+    test_globalaveragepool_precomputed_cpu,zdnn_meanreduce2d,NO_DYNAMIC_SHAPE_TEST # Height and Width dimension must be static dimension.
```
tungld, line 195: Comments here look the same as `==LIM==`, are they necessary?

test/accelerators/NNPA/backend/CMakeLists.txt
```diff
@@ -118,32 +118,32 @@ set(NNPA_TEST_LIST
...
+    test_batchnorm_epsilon_cpu,zdnn_mul_ext,"0:0=a,1=b,2=c,3=d|1:0=b|2:0=b|3:0=b|4:0=b"
+    test_batchnorm_example_cpu,zdnn_mul_ext,"0:0=a,1=b,2=c,3=d|1:0=b|2:0=b|3:0=b|4:0=b"
 
     # ==OP== Conv
     # ==MIN== 1
     # ==LIM== - `auto_pad` must be `NOTSET`, `VALID`, and `SAME_UPPER`. If `NOTSET` is used, `pads` must be set so that the padding valid type or same upper.<br>- Dimension in Height and weight must be static.<br>- `group` must be default value(1).<br>- `dilations` must be default value(1).<br>- Input and output tensors must have 4D (N x C x H x W).<br>- `kernel_shape` must be static.
-    test_basic_conv_with_padding_cpu,zdnn_conv2d,NO_DYNAMIC_SHAPE_TEST
-    test_basic_conv_without_padding_cpu,zdnn_conv2d,NO_DYNAMIC_SHAPE_TEST
+    test_basic_conv_with_padding_cpu,zdnn_conv2d,NO_DYNAMIC_SHAPE_TEST # Height and/or width must be static dimensions.
+    test_basic_conv_without_padding_cpu,zdnn_conv2d,NO_DYNAMIC_SHAPE_TEST # Height and/or width must be static dimensions.
```
tungld, line 143: Comments here look the same as ==LIM==, are they necessary?

test/accelerators/NNPA/backend/CMakeLists.txt
imaihal: I set dynamic shape in spacial dims for input on CPU, but it didn't work. DYNAMIC_SHAPE option modify only shape of input to dynamic, but I found we need to modify also the shape of output. 
It seems the shape mismatch between the output of rewriting and model output. There may be a bug in rewriting.

imaihal: @tungld  In dynamic backend test, input becomes dynamic, but output does not. Current rewriting fails to replace output tensor  because output tensor dimension by rewriting and existing dimension are different. Also shapeinference fails to replace output tensor. Should we be able to replace in this case?

```
 func.func @main_graph(%arg0: tensor<?x?x3x3xf32> {onnx.name = "X"}, %arg1: tensor<?x?x3x3xf32> {onnx.name = "W"}) -> (tensor<1x2x6x6xf32> {onnx.name = "Y"}) {
    %1 = "onnx.ConvTranspose"(%arg0, %arg1, %0) {auto_pad = "SAME_UPPER", group = 1 : si64, strides = [2, 2]} : (tensor<?x?x3x3xf32>, tensor<?x?x3x3xf32>, none) -> tensor<1x2x6x6xf32>
    onnx.Return %1 : tensor<1x2x6x6xf32>
```

tungld: It's the problem of dynamic tests. When changing the input tensor to dynamic, we should change the output tensor to a unranked tensor, and let the shape inference does its job so that we can check if the shape inference works for dynamic tensor also. If an output tensor has ranked, the compiler will consider its the correct shape and the shape inference may fail due to the mismatched shape.

imaihal: Removed `NO_DYNAMIC_SHAPE_TEST` because PR #2834 was merged. Thanks!

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp

test/accelerators/NNPA/backend/CMakeLists.txt
imaihal: Removed. Thanks.

test/accelerators/NNPA/backend/CMakeLists.txt
imaihal: Removed. Thanks.

tungld (approved):
LGTM. Thanks!

PR #2833, opened 2024-05-29
Fix the type of Normalization output

chentong319:
Set the type of the generated Value with the original type of the output of normalization Ops. Needed by the new llvm.


hamptonm1 (approved):
LGTM!! Especially after testing, thanks so much!

PR #2834, opened 2024-05-30
Force unranked tensor on output tensor in dynamic backend test

imaihal:
Existing dynamic backend tests only changes the shape of input tensor of static onnx models. We need to change shape of output tensor for appropriate dynamic tests. This was discussed in  
https://github.com/onnx/onnx-mlir/pull/2831#discussion_r1615516678

Currently only input tensor is updated to use unknown dimension in the tests. In this case, some tests fails in shape inference or rewriting because output tensor already has static dimension.  
```
func.func @main_graph(%arg0: tensor<?x?x3x3xf32> {onnx.name = "X"}, %arg1: tensor<?x?x3x3xf32> {onnx.name = "W"}) -> (tensor<1x2x6x6xf32> {onnx.name = "Y"}) {
    %1 = "onnx.ConvTranspose"(%arg0, %arg1, %0) {auto_pad = "SAME_UPPER", group = 1 : si64, strides = [2, 2]} : (tensor<?x?x3x3xf32>, tensor<?x?x3x3xf32>, none) -> tensor<1x2x6x6xf32>
    onnx.Return %1 : tensor<1x2x6x6xf32>
}
```
This PR also updates output tensor to unranked tensor. The shape inference set appropriate shape in the tests.
```
func.func @main_graph(%arg0: tensor<?x?x3x3xf32> {onnx.name = "X"}, %arg1: tensor<?x?x3x3xf32> {onnx.name = "W"}) -> (tensor<*xf32> {onnx.name = "Y"}) {
    %1 = "onnx.ConvTranspose"(%arg0, %arg1, %0) {auto_pad = "SAME_UPPER", group = 1 : si64, strides = [2, 2]} : (tensor<?x?x3x3xf32>, tensor<?x?x3x3xf32>, none) -> tensor<*xf32>
    onnx.Return %1 : tensor<*xf32>
}
```

To apply these change, we needed to update lowering for  `CastLike` op, `ConvTransepose` op. 
Also, thie PR updates configuration and description of dynamic tests of some models.


src/Dialect/ONNX/DialectBuilder.hpp
```diff
@@ -44,6 +44,8 @@ struct OnnxBuilder : DialectBuilder {
   mlir::Value add(mlir::Value A, mlir::Value B) const;
 
   // ONNXCastOp
+  mlir::Value cast(mlir::Value input, mlir::Value output,
+      mlir::IntegerAttr saturate, mlir::TypeAttr to) const;
```
tungld, line 48: It's weird to pass in an output value to construct a op. `mlir::Value output` should be replaced by `mlir::Type outputType`, and change the order of arguments, say `(mlir::Type outputType, mlir::Value input, mlir::IntegerAttr saturate, mlir::TypeAttr to)`

tungld (commented):
Thanks for the patch! I looks good.

AlexandreEichenberger:
Thanks for upgrading our support of dynamic testing

src/Dialect/ONNX/Transforms/Decompose.cpp
```diff
@@ -942,7 +944,7 @@ class ReplaceCastLikeByCastPattern : public OpRewritePattern<ONNXCastLikeOp> {
 
     // Replace
     Value res = onnx_mlir::OnnxBuilder(rewriter, loc)
-                    .cast(input, saturate, TypeAttr::get(outputType));
+                    .cast(input, output, saturate, TypeAttr::get(outputType));
```
hamptonm1, line 947: @imaihal Is there a reason why you want to pass in the `output` here ?

src/Dialect/ONNX/DialectBuilder.hpp
hamptonm1: @tungld @imaihal `outputType` is already defined here:

src/Dialect/ONNX/Transforms/Decompose.cpp lines 942 and 943
```
  // The output type will be the same as the target_type or the second input
    Type outputType = target.getType().cast<ShapedType>().getElementType();
```

It seems like we need outputType for src/Dialect/ONNX/DialectBuilder.cpp line 48, if that is the case please add outputType to the construct for cast and remove lines 942 and 943.

test/backend/inference_backend.py
```diff
@@ -885,7 +875,7 @@ def get_test_models():
         },
         # ==OP== Einsum
         # ==MIN== 12
-        # ==LIM== Limited to the types supported by ReduceSum and MatMul (which we decompose to in most cases) which exclude integers with width < 32
+        # ==LIM== Limited to the types supported by ReduceSum and MatMul (which we decompose to in most cases) which exclude integers with width < 32. Unknownd dimension is not supported in `inputs`.
```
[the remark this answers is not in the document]
hamptonm1: @imaihal Please remove the extra `d` in unknown

Also @tungld is referring to running the target `make onnx_mlir_supported_ops_cpu` and `make onnx_mlir_supported_ops_NNPA` so the comments are also updated for the documentation. Thanks!

test/backend/inference_backend.py
```diff
@@ -3310,6 +3306,7 @@ def get_test_models():
         "test_unsqueeze_unsorted_axes_cpu": {CONSTANT_INPUT: {1}},
         # ==OP== Upsample
         # ==MIN== 7
+        # ==LIM== Only `condition` of inputs supports Unknown dimension.
```
hamptonm1, line 3309: Small nit... can you make the `u` in `unknown` lower case please :)

hamptonm1 (commented):
Looks good but had some feedback!

src/Dialect/ONNX/DialectBuilder.hpp
imaihal: @hamptonm1 Thanks for the comments. The `outputType` is the type of second input(`target_type`) of CastLike op. It might be better to use `targetType` instead of `outputType`.  We want to use output of CastLikeOp. 
```
Type targetType = target.getType().cast<ShapedType>().getElementType();
Type outputType = output.getType().cast<ShapedType>().getElementType();
```
So I will create this outputType and pass it to builder.

imaihal: Done.

test/backend/inference_backend.py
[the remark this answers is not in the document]
imaihal: Done. Thanks!

src/Dialect/ONNX/Transforms/Decompose.cpp
imaihal: Done. Thanks.

test/backend/inference_backend.py
imaihal: Done. Thanks.

test/backend/inference_backend.py
```diff
@@ -1288,6 +1279,7 @@ def get_test_models():
         },
         # ==OP== GRU
         # ==MIN== 7
+        # ==LIM== Only `X` of inputs supports unknown dimensions.
```
tungld, line 1282: I know, in most cases, weights and biases are constants, so you are testing with X being unknown only. But it does not mean that this op does not support unknowns dimensions for weights and biases.  I though we wrote RNN ops using IndexExpr, and expected that it supports other unknown inputs also such as weights and biases. Could you please check it again? Thanks!

test/backend/inference_backend.py
```diff
@@ -3126,6 +3121,7 @@ def get_test_models():
         },
         # ==OP== TopK
         # ==MIN== 10
+        # ==LIM== Unknown dimension in `X` of inputs is supported.
```
tungld, line 3124: This message does not seem to be a `limitation`. Could you make it clearer?

src/Dialect/ONNX/DialectBuilder.cpp
```diff
@@ -43,6 +43,16 @@ Value OnnxBuilder::add(Value A, Value B) const {
...
 }
 
+Value OnnxBuilder::cast(
+    Type outputType, Value input, IntegerAttr saturate, TypeAttr to) const {
+  if (outputType.cast<ShapedType>().hasRank()) {
+    return cast(input, saturate, to);
+  } else {
+    Type resultType = UnrankedTensorType::get(to.getValue());
+    return b().create<ONNXCastOp>(loc(), resultType, input, saturate, to);
+  }
```
tungld, line 53: It's a bit confusing here to know the meaming of `outputType`. `outputType` often means the type of the output, so we can use it directly, say `return b().create<ONNXCastOp>(loc(), outputType, input, saturate, to);`. The logic of constructing `resultType` should be outside of this function.

I suggest to define this function as follows:
```
Value OnnxBuilder::cast(
    Type outputType, Value input, IntegerAttr saturate, TypeAttr to) const {
    return createTypedOpAndInferShapes<ONNXCastOp>(outputType, input, saturate, to);
}
```

src/Dialect/ONNX/Transforms/Decompose.cpp
```diff
@@ -934,15 +935,17 @@ class ReplaceCastLikeByCastPattern : public OpRewritePattern<ONNXCastLikeOp> {
...
     // The output type will be the same as the target_type or the second input
-    Type outputType = target.getType().cast<ShapedType>().getElementType();
+    Type targetType = target.getType().cast<ShapedType>().getElementType();
 
     // Replace
-    Value res = onnx_mlir::OnnxBuilder(rewriter, loc)
-                    .cast(input, saturate, TypeAttr::get(outputType));
+    Value res =
+        onnx_mlir::OnnxBuilder(rewriter, loc)
+            .cast(output.getType(), input, saturate, TypeAttr::get(targetType));
```
tungld, line 948: I think we can use an unknown result type, then DialectBuilder will infer shape when constructing the op.
```
Type resultType = UnrankedTensorType::get(targetType);
Value res = onnx_mlir::OnnxBuilder(rewriter, loc).cast(resultType, input, saturate, TypeAttr::get(targetType));
 ```

src/Dialect/ONNX/DialectBuilder.hpp
hamptonm1: @imaihal According to the documentation, `The (first) input tensor will be cast to produce a tensor of the same type as this (second input) tensor.`

So the `targetType` and the `outputType` should be the same type. That is maybe why there is confusion. The tensor types need to be the same for castLike to work correctly.

hamptonm1: But thanks so much for your update!

hamptonm1:
@imaihal It looks good to me once you address @tungld feedback :)

test/backend/inference_backend.py
imaihal: I tested RNN, LSTM, and GRU with `DYNAMIC_SHAPE{{-1}: {-1}}`, but most of tests failed(10 out of 12 failed). I have not investigated them yet, but it seems we have issues in dynamic shape.

tungld: Thanks for confirmation! You may not need to fix the issues but I would suggest to change the message to something like `W, B and R must be constants` to explicitly say the limitation applies to weights like W, B or R, not X.

test/backend/inference_backend.py
imaihal: Updated. Thanks!

src/Dialect/ONNX/DialectBuilder.cpp
imaihal: @tungld Thanks for your comments!
I tried this before, but when inferring shape here using `createTypedOpAndInferShapes`, rewriting CastLike op to Cast op failed. I think this is because tensor type mismatches since output type of CastLike op is unranked tensor type, but output of Cast op is ranked tensor type with unknown dimension after inferring shape. Is this strange behavior?

tungld: I see. So it means we have two cases: use outType as it is and inferShape. 

How about adding a flag `inferShape` (set to true by default) to the function? 

If `inferShape == true` we will return  `createTypedOpAndInferShapes<ONNXCastOp>(outputType, input, saturate, to);` otherwise `b().create<ONNXCastOp>(loc(), outputType, input, saturate, to);`

test/backend/inference_backend.py
tungld: Did you push the change to the PR? Looks like it's unchanged.

test/backend/inference_backend.py
imaihal: Done. Thanks!

test/backend/inference_backend.py
imaihal: Updated. Thanks!

src/Dialect/ONNX/DialectBuilder.cpp
imaihal: Done. Thanks!

src/Dialect/ONNX/Transforms/Decompose.cpp
imaihal: Updated. I checked if output type has rank, and if not, this code is used.

chentong319 (commented):
I suggest that the change of output type is turned on only for our dynamic backend test. I do not feel comfortable to modify the original type to a less precise type in general.

imaihal:
@chentong319 

> I suggest that the change of output type is turned on only for our dynamic backend test. I do not feel comfortable to modify the original type to a less precise type in general.

I agree. This modification is only applied when `IMPORTER_FORCE_DYNAMIC` is set. This is the same condition that input dimension is changed to unknown dimensions in backend tests.

hamptonm1 (approved):
LGTM!

hamptonm1:
@chentong319 Will you be able to remove your scaffolding in `Upsample` and `Layernorm` with these changes? If this executes the way we expect, we may no longer need that piece of code.

PR #2836, opened 2024-06-06
Update RunONNXModel.py to allow using numpy arrays as inputs and outputs

tungld:
It's often the case that we want to define custom inputs to check a model using RunONNXModel.py. It would be convenient to use numpy arrays as inputs and outputs. 

This patch adds an option `--load-ref-from-numpy` to allow such a feature in RunONNXModel.py.

For example, we define a python script, `data.py`, to create inputs and outputs as follows:
```python
import numpy as np
in1 = np.ones((3,4,5), dtype=np.float32)
in2 = np.zeros((3,4,5), dtype=np.float32)
inputs = [in1, in2]
outputs = [in1]
```
where we define `inputs` and `outputs` variables whose names must be fixed. `outputs` can be absent if we do not verify the output.

Then we can test a model using these inputs and outputs by using `--load-ref-from-numpy`:
```bash
RunONNXModel.py  -m add.onnx --load-ref-from-numpy=./data.py --verify=ref --verify-every-value
```


tungld:

utils/RunONNXModel.py
```diff
@@ -158,6 +160,14 @@ def check_non_negative(argname, value):
     help="Path to a folder containing reference inputs and outputs stored in protobuf."
     " If --verify=ref, inputs and outputs are reference data for verification",
 )
+data_group.add_argument(
+    "--load-ref-from-numpy",
+    metavar="PATH",
+    type=str,
+    help="Path to a python script that defines variables inputs and outputs that are a list of numpy arrays. "
```
AlexandreEichenberger, line 167: Great, please add that this script expect 2 lists, named 'inputs' an 'outputs' which lists numpy arrays.
Is outputs optional?
Please add a bit of details.

utils/RunONNXModel.py
```diff
@@ -158,6 +160,14 @@ def check_non_negative(argname, value):
...
+    " For example, inputs = [np.array([1], dtype=np.int64), np.array([2], dtype=np.float32]."
+    " Variable outputs can be omitted if --verify is not used.",
```
tungld, line 169: >Is outputs optional?
Please add a bit of details.

@AlexandreEichenberger is it not enough and confusing?

PR #2837, opened 2024-06-07
Parallel tune: go CPU parallel only when loops have enough work (systematically)

AlexandreEichenberger:
Added tests before setting a loop to parallel (when enabled). If not enough work is present, then no parallel is generated.

This is a first step; in a follow on PR, will try to exploit inner loop parallelism when profitable too.


tungld (approved):
LGTM.

PR #2838, opened 2024-06-10
Add Runtime bound check

chentong319:
MLIR provides a pass to insert runtime bound check for memerfs. This PR add this pass to onnx-mlir. However, the pass is disabled by default, and can be enabled by flag `enable-bound-check` to help debugging runtime issues.


AlexandreEichenberger (approved):
LGTM, please fix the DCO

PR #2840, opened 2024-06-10
Parallel tune: transformations for Roberta with BS=1

AlexandreEichenberger:
In Roberta with a BS of 1, we push all operations to NNPA and the ones executing on the CPU are typically parallelized as most Elementwise ops are first collapsed to a 1D iteration space that is then SIMDized and parallelized.

The one exception is transpose: this PR enables parallelization of transpose in the first 2 dims (aka if dim of outermost is 1, then it tries the second dim). 

With this transformation, all CPU operations are executed in parallel (when including the -enable-compiler-stick-unstick).

Overal, here is the reduction in execution time in the transpose (6 CPUs)
```
NEW:
  onnx.Transpose, 48, 0.0742153, 3.5623333, 3.4%

OLD:
  onnx.Transpose, 48, 0.3418125, 16.4070000, 14.7%
```

namely a 4.5x speedup for that operation, resulting in an overall reduction of 1.06x speedup for Robera with 1x384 input size

Note that this PR is built on top of the previous parallel tuning.


src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
```diff
@@ -617,17 +617,22 @@ bool hasNonIdentityLayout(ValueRange operands) {
...
 bool findSuitableParallelDimension(llvm::SmallVectorImpl<IndexExpr> &lb,
-    llvm::SmallVectorImpl<IndexExpr> &ub, int64_t firstDim, int64_t lastDim,
-    int64_t &parDim, int64_t minSize) {
-  for (int64_t i = firstDim; i < lastDim; ++i) {
+    llvm::SmallVectorImpl<IndexExpr> &ub, int64_t firstInclusiveDim,
+    int64_t lastExclusiveDim, int64_t &parDim, int64_t minSize) {
+  assert(lb.size() == ub.size() && "expected identical ranks for lb/ub");
+  firstInclusiveDim = firstInclusiveDim > 0 ? firstInclusiveDim : 0;
+  lastExclusiveDim =
+      lastExclusiveDim > (int64_t)lb.size() ? lb.size() : lastExclusiveDim;
```
tungld, line 631: Suggest to change to:
```C
if (firstInclusiveDim <= 0)
  firstInclusiveDim = 0;
if (lastExclusiveDim > (int64_t)lb.size())
  lastExclusiveDim = lb.size()
```

since the value is unchanged when condition is not met.

tungld (approved):
LGTM. Glad to see the performance improvement. Thanks!

Issue #2808, opened 2024-04-26
Better error when NNPA is requested on z16 machines with it disabled.


cjvolzka:
# Issue
We've recently encountered scenarios with Linux z16 machines that have the nnpa feature disabled.  The scenario occurs when hosts are part of z/VM groups with Live Guest Relocation enabled. If the group has both z15 and z16 hosts, z16 specific features are disabled by the hypervisor so guest vms/lpars can safely be migrated between group hosts.

You can tell if you are in such a scenario by calling `lscpu | grep "Flags:"` and `nnpa` will be missing from the list even though the system is a z16.

# Current Result
Currently we emit a misleading error that says either the version of zDNN they are using is wrong or they are not on a z16 host.

```
Model is running on hardware that is not compatible with the zDNN library that this model was compiled for (version num 1.0.1). Please check that the model is running on hardware with an integrated accelerator for AI (z16 +) that supports the required zDNN library version.
```

# Possible alternative
Ideally we should emit a better message in this scenario. I'm not sure off hand if we can detect if the host they are running on is a z16 or not, but we should be able to detect that nnpa isn't enabled by using `zdnn_is_nnpa_installed()` 

So first we check `zdnn_is_nnpa_installed` if it fails, we emit something like (wording TBD):
```
This model was compiled to use integrated accelerator for AI. Check the model is running on z16 or later system and that the nnpa feature is enabled for the system."
```

Then check `zdnn_is_version_runnable` and if it fails emit:
```
Model is running on hardware that is not compatible with the zDNN library that this model was compiled to use (zDNN X.Y.Z). Please ensure a compatible zDNN library is available.
```

PR #2842, opened 2024-06-11
Update error message for hardware compatibility

hamptonm1:
This PR is a solution to the issue opened up here: https://github.com/onnx/onnx-mlir/issues/2808


hamptonm1:

PR #2843, opened 2024-06-12
[FXML-4647] Add ONNX operator versions as a dialect trait

cferry-AMD:


cferry-AMD:
No need, this is not a PR for upstream.

PR #2844, opened 2024-06-12
Remove extraneous rank checks in NNPA ONNXLegalityCheck

mikeessen:
Method `isValidElementTypeAndRank` in  NNPA `ONNXLegalityCheck.cpp` validates that the input tensor is rank 4 or less.  Remove the extraneous checks for this in methods that call `isValidElementTypeAndRank`.


tungld (approved):
LGTM.

Thanks! They are really redundant.

PR #2857, opened 2024-06-27
Modify RunONNXModel.py for pip install

chentong319:
The goal is to create a python package so that we can run an inference like onnxruntime in python script. Since the RunONNXModel.py provides most of the necessary functionality, I added extra support in that script. The pip package related code will be checked later. I tested with a local pip install. We can upload the package later.

Major changes:
1. Add the function interface to call the main in RunONNXModel.py with the parameters similar to onnxruntime:
2. Accept list of arrays as input. 
3. Return the outputs

Example:
```
import numpy as np
import onnxmlirrun

a = np.zeros((3,4,5), dtype=np.float32)
b = a + 4
r = onnxmlirrun.RunONNXModel.onnxmlirrun(compiled_so="/Users/chentong/Projects/onnx-mlir/build/test_add.so", inputs=[a,b])
print(r)
r = onnxmlirrun.RunONNXModel.onnxmlirrun(onnx_model="/Users/chentong/Projects/onnx-mlir/build/test_add.onnx", inputs=[a,b])
print(r)
```

We may have different name for the functions and the package. Comments are welcome.


AlexandreEichenberger:
@chentong319 Can you add a reference to the ORT interface that this PR is imitating.

In your first example
```
r = onnxmlirrun.RunONNXModel.onnxmlirrun(compiled_so="/Users/chentong/Projects/onnx-mlir/build/test_add.so", inputs=[a,b])
```
was the  `onnx_model` parameter omitted on purpose? Just for me to understand better the interface.

chentong319:
> @chentong319 Can you add a reference to the ORT interface that this PR is imitating.
> 
> In your first example
> 
> ```
> r = onnxmlirrun.RunONNXModel.onnxmlirrun(compiled_so="/Users/chentong/Projects/onnx-mlir/build/test_add.so", inputs=[a,b])
> ```
> 
> was the `onnx_model` parameter omitted on purpose? Just for me to understand better the interface.

Yes, onnx_model is not needed if the compiled .so is provided. By the way, the two onnxmilrrun.RunONNXModel.onnxmlirrun commands are two independent ways to do inference.

AlexandreEichenberger:
Got it, so we can provide a onnx file or a pre-compiled binary, smart, thanks.

utils/RunONNXModel.py
```diff
@@ -861,7 +867,19 @@ def main():
...
+        return outs
+
+# Python function inteface for RunONNXModel
+# Arguments are passed as named parameters for the function
+# Extra functionality is to directly pass a list of arrays as inference input
+def onnxmlirrun(onnx_model=None, compiled_so=None, inputs=None):
+    if onnx_model :
+        args.model = onnx_model
+    if compiled_so :
+        args.load_so = compiled_so
```
tungld, line 879: Should we check if both `onnx_model` and `compiled_so` are not given?

AlexandreEichenberger: Make sense, and maybe if none are given, an error?

chentong319: I added the check whether both onnx_mdel and compiled_so are given or not give (XOR).

chentong319:
I changed the __init__.py for the package to simplify `onnxmlirrun.RunONNXModel.onnxmlirrun` to `onnxmlir.onnxmlir`. 
I am thinking to add the code for package into onnx-mlir in future.

AlexandreEichenberger:
Can you copy/link a description of the ORT runtime interface and update ours with the new names, so that we can see how they relate? Thanks

chentong319:
The Ort code is like
```
import onnxruntime as ort
import numpy as np
x, y = test_data[0][0], test_data[0][1]
ort_sess = ort.InferenceSession('fashion_mnist_model.onnx')
outputs = ort_sess.run(None, {'input': x.numpy()})
```

Since the current RunModelPython does not separate the session creation and run, our code will be like:
```
import onnxmlir
import numpy as np
x, y = test_data[0][0], test_data[0][1]
outputs = onnxmlir.run(onnx_model='fashion_mnist_model.onnx', inputs = [a, b])
```
It is doable to totally mimic ORT in separating the session creation and run, and using dictionary for input.

AlexandreEichenberger:
Ideally

```
import onnxmlir as ort
import numpy as np
x, y = test_data[0][0], test_data[0][1]
ort_sess = ort.InferenceSession('fashion_mnist_model.onnx')
outputs = ort_sess.run(None, {'input': x.numpy()})
```

with additional parameters ok

```
import onnxruntime as ort
import numpy as np
x, y = test_data[0][0], test_data[0][1]
ort_sess = ort.InferenceSession('fashion_mnist_model.onnx', compile_flags="-O3")
outputs = ort_sess.run(None, {'input': x.numpy()})
```

reusing flags that already exists, if any.

utils/RunONNXModel.py
```diff
@@ -868,6 +795,163 @@ def main():
...
+
+In onnxmlir, session_options and provider will be merged into kwargs, and
+ignored. onnxruntime.SessionOptions may contain some useful info,
+but onnxruntime package is needed to interpret it. Therefore, it is ignored now.
+Another argument, 'options' is added for onnxmlir to specify options for RunONNXModel.py
+"""
+
+
+def inferenceSession(model_name, **kwargs):
+    global args
```
tungld, line 817: Can we move this code into the `__init__` of `InferenceSession` class and remove `inferenceSession` method? The reason is it is a bit confusing about `InferenceSession` class and `inferenceSession` method. Other parts looks good to me.

chentong319: Fixed. Thanks for the suggestion.

utils/onnxmlir/VERSION_NUMBER
```diff
@@ -0,0 +1 @@
+1.0.0
```
AlexandreEichenberger, line 1: Version number here is different than `onnx-mlir/VERSION_NUMBER`. Is that on purpose?
Also the number here is different than the version listed in the file below.

utils/onnxmlir/src/onnxmlir/RunONNXModel.py
```diff
@@ -0,0 +1,959 @@
+#!/usr/bin/env python3
```
AlexandreEichenberger, line 1: Is this an identical copy? If so, at a minimum there should be a comment stating that they should stay in sync. Or maybe we could have a single version and we could "build" the pip structure by copying files in the right place. That would be my preference.

AlexandreEichenberger (commented):
Check error message if file does not exist.

AlexandreEichenberger (commented):
TODO: let's try to have parallel default and/or as set in onnxruntime.

AlexandreEichenberger (commented):
TODO: sync also with #2899 with respect to catching errors/crasshes

utils/onnxmlir/VERSION_NUMBER
chentong319: Fixed. I used a version number different from the onnx-mlir. The version for onnxmlir is for upstream onnxmlir to pip repo. It could be viewed as a pre-required package.

utils/onnxmlir/src/onnxmlir/RunONNXModel.py
chentong319: Added a target in CMake to check the files.

chentong319:
> Check error message if file does not exist.

onnx-mlir will report the error if file does not exist. Leave this to the TODO list to catch all the errors gracefully.

utils/onnxmlir/VERSION_NUMBER
AlexandreEichenberger: May I suggest a `0.1.0` and move to a `1.0.0` once we have the installation of the compiler with it at least one z?

utils/onnxmlir/pyproject.toml
```diff
@@ -0,0 +1,22 @@
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+[project]
+name = "onnxmlir"
+version = "1.0.0"
```
AlexandreEichenberger, line 6: Thanks for having it in sync.

chentong319:
@jenkins-droid test it!

hamptonm1:

AlexandreEichenberger (approved):
LGTM, thanks for the updates

PR #2860, opened 2024-07-03
Constant propagation for Reciprocal

tungld:
Constant propagation for Reciprocal


AlexandreEichenberger (approved):
LGTM,

out of curiosity, if we do new constant propagation, should we add also something to the handling of the constants (lazy eval) to avoid blowing up constant memory while compiling?

tungld:
> out of curiosity, if we do new constant propagation, should we add also something to the handling of the constants (lazy eval) to avoid blowing up constant memory while compiling?

our constant propagation mechanism is lazy evaluation, so nothing to do when supporting a new op.

PR #2861, opened 2024-07-03
Show the compilation phases

tungld:
This patch is to show the compilation phases which is helpful when compiling a big model that takes a long time.

This is one example of the printed information:
```shell
$ onnx-mlir -mcpu=z16 -maccel=NNPA bert_base_uncased_1.onnx
[1/5] Wed Jul  3 03:36:50 2024 Importing ONNX Model to MLIR Module
[2/5] Wed Jul  3 03:36:51 2024 Compiling and Optimizing MLIR Module
[3/5] Wed Jul  3 03:37:14 2024 Translating MLIR Module to LLVM and Generating LLVM Optimized Bitcode
[4/5] Wed Jul  3 03:37:35 2024 Generating Object from LLVM Bitcode
[5/5] Wed Jul  3 03:37:39 2024 Linking and Generating the Output Shared Library
```

Starting time is added to each phase so that users know how long the current phase takes

### DIFF
--- a/src/Compiler/CompilerUtils.cpp
+++ b/src/Compiler/CompilerUtils.cpp
@@ -180,10 +180,8 @@ void showCompilePhase(std::string msg) {
   strftime(buffer, 80, "%c", timeinfo);
   std::string currentTime(buffer);
 
-  // Use blue bold text.
-  llvm::outs() << "[\033[34;1m" << CURRENT_COMPILE_PHASE++ << "/"
-               << TOTAL_COMPILE_PHASE << "\033[0m] " << currentTime << " "
-               << msg << "\n";
+  llvm::outs() << "[" << CURRENT_COMPILE_PHASE++ << "/" << TOTAL_COMPILE_PHASE
+               << "] " << currentTime << " " << msg << "\n";
 }
 
 } // namespace onnx_mlir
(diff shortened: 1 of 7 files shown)


tungld:
FYI @tehbone

tehbone:
Is this default behavior or enabled via switch?

AlexandreEichenberger (approved):
LGTM, thanks for acting on this quickly.

@tehbone it looks like its on, no matter what, from my read of the diff. Would you like to have a switch?

src/Compiler/CompilerUtils.cpp
```diff
@@ -162,6 +169,23 @@ int Command::exec(std::string wdir) const {
...
+  char buffer[80];
+
+  // Get current date.
+  time(&rawtime);
+  timeinfo = localtime(&rawtime);
+  strftime(buffer, 80, "%c", timeinfo);
+  std::string currentTime(buffer);
+
+  // Use blue bold text.
+  llvm::outs() << "[\033[34;1m" << CURRENT_COMPILE_PHASE++ << "/"
```
gongsu832, line 184: ANSI escape sequence isn't support on all terminal types. And different terminal might even interpret the sequence differently. So unless the color adds a lot of value, I would just leave it out.

tungld: Ok. Removed it. Thanks!

tungld:
> Is this default behavior or enabled via switch?

Yes, it's on by default. Let me know if you want a switch or set it off.

PR #2862, opened 2024-07-03
Constprop: More unary/binary ops. Also don't crash in divide by Zero.

ttjost:


src/Dialect/ONNX/Transforms/ConstProp.cpp
```diff
@@ -134,6 +145,46 @@ Value ConstZeroTensor(
...
+    GetIntConstFunc intConstantFunc) {
+  auto elemType = type.getElementType();
+  if (auto floatType = dyn_cast<FloatType>(elemType)) {
+    auto fpValue =
+        fpConstantFunc(floatType.getFloatSemantics(), /*Negative=*/isNegative);
+    return rewriter.create<ONNXConstantOp>(value.getLoc(), Attribute(),
+        DenseElementsAttr::get(type, llvm::ArrayRef(fpValue)));
+  }
+  auto intValue = intConstantFunc(elemType.getIntOrFloatBitWidth());
+  return rewriter.create<ONNXConstantOp>(value.getLoc(), Attribute(),
```
AlexandreEichenberger, line 162: Just check: I don't see the `isNegative` being used on the int path. It looks ok as the `isNegative` is only used to find the minimum floating point.

AlexandreEichenberger (approved):
Thanks for adding all the constant propagation rules for more operations, very useful.
LGTM

AlexandreEichenberger:
@tungld @flemairen6 If we add new operators to generate constants (like in this PR), do we also need to add new operators to the"lazy creation" of our constant values, to ensure we don't blow up our memory requirement for large benchmarks?

It would be nice to have some clarity on this.

AlexandreEichenberger:

AlexandreEichenberger:

AlexandreEichenberger:

AlexandreEichenberger:
@flemairen6 you have rights to start the Jenkins, correct?

src/Dialect/ONNX/Transforms/ConstProp.cpp
ttjost: Exactly. APInt has different methods for Min and Max, while APFloat has only `getLargest`, which makes sense since the latter only changes one bit, while the former have different ranges for positive and negative.

flemairen6:
@AlexandreEichenberger I don't think I do

AlexandreEichenberger:
Looks line an error in this lit test
```******************** TEST 'Open Neural Network Frontend :: onnx/onnx_constprop.mlir' FAILED ********************
Exit Code: 1

Command Output (stderr):
--
RUN: at line 1: /workdir/onnx-mlir/build/Release/bin/onnx-mlir-opt --shape-inference --constprop-onnx /workdir/onnx-mlir/test/mlir/onnx/onnx_constprop.mlir -split-input-file | /workdir/llvm-project/build/bin/FileCheck /workdir/onnx-mlir/test/mlir/onnx/onnx_constprop.mlir
+ /workdir/onnx-mlir/build/Release/bin/onnx-mlir-opt --shape-inference --constprop-onnx /workdir/onnx-mlir/test/mlir/onnx/onnx_constprop.mlir -split-input-file
+ /workdir/llvm-project/build/bin/FileCheck /workdir/onnx-mlir/test/mlir/onnx/onnx_constprop.mlir
/workdir/onnx-mlir/test/mlir/onnx/onnx_constprop.mlir:390:12: error: CHECK: expected string not found in input
 // CHECK: onnx.Constant dense<{{.}}[-2.765630e+00, 0xFFC0], [0xFF80, 0x7FC0], [0x7F80, 0xFFC0]]>
           ^
<stdin>:244:43: note: scanning from here
 func.func @test_log() -> tensor<3x2xbf16> {                                     ^
```

If you don't see the error in your normal build environment, try building the RELEASE mode, as it sometimes expose errors.

AlexandreEichenberger:
> @AlexandreEichenberger I don't think I have

will ask for you to be added.

src/Dialect/ONNX/Transforms/ConstProp.cpp
```diff
@@ -134,6 +145,46 @@ Value ConstZeroTensor(
           type, rewriter.getZeroAttr(type.getElementType())));
 }
 
+template <typename GetFPConstFunc =
+              std::function<APFloat(const llvm::fltSemantics &, bool)>,
+    typename GetIntConstFunc = std::function<APInt(unsigned)>>
+Value GetClipConstantOfType(PatternRewriter &rewriter, ShapedType type,
+    Value value, GetFPConstFunc fpConstantFunc, bool isNegative,
+    GetIntConstFunc intConstantFunc) {
```
tungld, line 153: Since `value` is None in this case and it is used to get location only, how about not passing it in to avoid confusion? We can pass in a location instead.

e.g.
```c
Value GetClipConstantOfType(PatternRewriter &rewriter,  Location loc,
    ShapedType type, GetFPConstFunc fpConstantFunc, bool isNegative,
    GetIntConstFunc intConstantFunc) {
```

Then you can use OnnxBuilder to create a constant:
```c
OnnxBuilder create(rewriter, loc);
return create.constant(DenseElementsAttr::get(RankedTensorType::get({}, elemType), llvm::ArrayRef(fpValue)));
```

Also `min`, `max` in Clip are scalar, we can use a scalar type for the constant to make it consistent with the other case where `value` is not None.

src/Dialect/ONNX/Transforms/ConstProp.cpp
```diff
@@ -134,6 +145,46 @@ Value ConstZeroTensor(
           type, rewriter.getZeroAttr(type.getElementType())));
 }
 
+template <typename GetFPConstFunc =
+              std::function<APFloat(const llvm::fltSemantics &, bool)>,
+    typename GetIntConstFunc = std::function<APInt(unsigned)>>
+Value GetClipConstantOfType(PatternRewriter &rewriter, ShapedType type,
```
tungld, line 151: Could you lowercase the first character of a function? e.g. `getClipConstantOfType`. Same for the other functions:  CreateMaximumValueForClip, CreateMinimumValueForClip. Thanks!

src/Dialect/ONNX/Transforms/ConstProp.cpp
```diff
@@ -287,6 +372,30 @@ constexpr auto subCombiner(Type elemType) {
   return elementwiseBinaryOpCombiner<ONNXSubOp>(elemType);
 }
 
+/// Precondition: min values must be less than max values if both exist.
+bool satisfiesMinLessThanMaxRequirement(ShapedType type, Value min, Value max) {
```
tungld, line 376: It looks like checking this precondition in the ClipOp's verifier is more appropriate since it applies to the non-constant case also.

src/Dialect/ONNX/Transforms/ConstProp.cpp
```diff
@@ -203,7 +254,41 @@ struct ElementWiseBinaryOpImpl<ONNXMulOp, T, EnableNotBool<T>> {
 
 template <typename T>
 struct ElementWiseBinaryOpImpl<ONNXDivOp, T, EnableNotBool<T>> {
-  static T eval(T lhs, T rhs) { return lhs / rhs; }
+  static T eval(T lhs, T rhs) {
+    if constexpr (std::is_integral_v<T>) {
+      if (rhs == 0) {
+        // Undefined behavior. We can return any value.
+        // Performing the divison would crash.
+        return lhs;
```
tungld, line 262: I prefer to stop compilation here since this is actually an error. The input model is not good or the compiler produces this zero. 

@AlexandreEichenberger any thought?

test/mlir/onnx/onnx_constprop.mlir
```diff
@@ -379,6 +490,266 @@ func.func @test_div_ones(%arg0 : tensor<1x2xui8>) -> tensor<1x2xui8> {
...
+  // CHECK: {{.*}} = onnx.Constant dense<{{.}}[5.000000e+00, 2.000000e+00], [0.000000e+00, 0x7FC0], [0.000000e+00, 1.000000e+01]]>
+  // CHECK-NOT: {{.*}} = "onnx.Clip"
+}
+
+// CHECK-LABEL:  func @test_clip_multiple_max
+func.func @test_clip_multiple_max() -> tensor<3x2xbf16> {
+  // Test Positive Clamped, Negative Clamped, In range, NaN, -Inf,  +Inf
+  %cst = onnx.Constant dense<[[2.125, -2.125], [0.0, 0x7FC0], [0xFF80, 0x7F80]]> : tensor<3x2xbf16>
+  %min = onnx.Constant {value = dense<-2.0> : tensor<bf16>} : tensor<bf16>
+  %max = onnx.Constant dense<[[5.000, 2.000], [-1.0, 0.0], [0.0, 0.125]]> : tensor<3x2xbf16>
```
tungld, line 563: `min` and `max` must be a scalar according to its specs: https://onnx.ai/onnx/operators/onnx__Clip.html

It would fail once we verify this in the ClipOp's verifier.

tungld (commented):
Thank you for adding support for many ops!

tungld:
> do we also need to add new operators to the"lazy creation" of our constant values, to ensure we don't blow up our memory requirement for large benchmarks?

Existing templates in ConstProp can handle constant values in a lazy way. If we do manually, just to make sure to use the OnnxElementsAttrBuilder class and ElementsAttrHelper functions to construct and access ElementsAttr. More info can be found here: https://github.com/onnx/onnx-mlir/blob/main/docs/ConstPropagationPass.md

src/Dialect/ONNX/Transforms/ConstProp.cpp
ttjost: I guess this also relates to my previous comment when `Min > Max`.

test/mlir/onnx/onnx_constprop.mlir
ttjost: Nice catch! I forgot to check that (which makes things a lot easier).

src/Dialect/ONNX/Transforms/ConstProp.cpp
mgehre-amd: ONNX unfortunately doesn't specify what the behavior should be; whether the model is illegal and should be rejected or whether just the output of that operator is unspecified, but the rest of the model continues (and potentially conditionally doesn't use the result).

tungld: Does anyone know how onnxruntime handle this situation?

I just checked how the current generated code for non-constant inputs deals with div-by-zero on a Z machine. The program does not crash and these are the results:
- f32 data type: the output is `negative infinity` if `rhs` is negative, `positive infinity` if `rhs` is positive. This result may be different on other architectures.
- integer type: the output is kind of random. [According to MLIR documentation](https://mlir.llvm.org/docs/Dialects/ArithOps/#arithdivsi-arithdivsiop), the div-by-zero for integer is undefined behaviour.


Anyway, it's constant propagation and it's useful if we would print out a warning message when we see div-by-zero.

AlexandreEichenberger: I responded to the (https://github.com/onnx/onnx/issues/6165) PR. I believe this situation should be declared 'undefined' and we should specify in the specs that tools (such as ORT, onnx-mlir) should individually define what they do in such situations.

In our case (onnx-mlir), I think it would be reasonable to generate a user error and abort the compilation when the error can be detected at compile time, and silently ignore it if the values are not known at compile time. If desired, we could implement a "fussy" compilation mode where such assumptions are tested at runtime by adding some dynamic asserts. This would require some new infra as I don't think we have at this time a runtime method that provide a "assert if this condition is false". Obviously, it would not necessarily be much work to add such a runtime operation.

AlexandreEichenberger:
> 
> Existing templates in ConstProp can handle constant values in a lazy way. If we do manually, just to make sure to use the OnnxElementsAttrBuilder class and ElementsAttrHelper functions to construct and access ElementsAttr. More info can be found here: https://github.com/onnx/onnx-mlir/blob/main/docs/ConstPropagationPass.md

@tungld thanks for the link. The doc seems to be out of date, as a grep of `combinerOfElementwiseBinaryOp` (which is described as the method to develop a combiner, which would be presumably need to be added for the new ops) yielded only hits in the doc, and not the code.

AlexandreEichenberger:
@ttjost the PR is ready to be merged but there are some conflicts that needs to be manually merged. Do you have the time to do that last step?

ttjost:
> @ttjost the PR is ready to be merged but there are some conflicts that needs to be manually merged. Do you have the time to do that last step?

Apologies for the delay. I made some changes as suggested, and now we use the same approach numpy uses for clip (`When a_min is greater than a_max, returns an array in which all values are equal to a_max, as shown in the second example.`

ttjost:
@AlexandreEichenberger, could you please start Jenkins?

AlexandreEichenberger:

tungld:

tungld:

ttjost:
Thanks @tungld . I can't reproduce locally the issues I am seeing in the CI, but it relates to different implementations of some math functions when returning NaN values. Locally and in a local CI I get negative NaN, and in Onnx-MLIR CI it returns me a positive NaN. Hopefully, this now fixes the issues.

ttjost:
FYI, needed to update the branch but all tests passed previously.

flemairen6:

tungld:

ttjost:
could someone please merge this PR?

tungld:
> could someone please merge this PR? Thanks!

Merged. Thank you very much for the PR!

PR #2863, opened 2024-07-04
Fix crash in ReshapeOp inference when an input to the previous concat is not a valid dimension

jorickert:
The problem was that the check:
```
if ((getShape(vType)[0] == 1) && getElementType(vType).isSignlessInteger(64))
    return true;
```
can succeed even if `areDims(Value val)` is not true for all operands of Concat. 

This causes a crash in `getDims`, which recursively calls `getDims` on all operands of Concats, without checking if these operands are dims.

### DIFF
--- a/src/Dialect/ONNX/ONNXOps/Tensor/Reshape.cpp
+++ b/src/Dialect/ONNX/ONNXOps/Tensor/Reshape.cpp
@@ -36,7 +36,8 @@ LogicalResult ONNXReshapeOpShapeHelper::computeShape() {
   // Get info about shape operand.
   Value shape = operandAdaptor.getShape();
   int64_t outputRank = createIE->getShape(shape, 0);
-  assert(outputRank != -1 && "Shape tensor must have constant shape");
+  assert(outputRank != ShapedType::kDynamic &&
+         "Shape tensor must have constant shape");
 
   // Initialize context and results.
   outputDims.resize(outputRank);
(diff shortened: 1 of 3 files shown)


tungld:

src/Dialect/ONNX/ONNXOps/Tensor/Reshape.cpp
```diff
@@ -36,7 +36,8 @@ LogicalResult ONNXReshapeOpShapeHelper::computeShape() {
   // Get info about shape operand.
   Value shape = operandAdaptor.getShape();
   int64_t outputRank = createIE->getShape(shape, 0);
-  assert(outputRank != -1 && "Shape tensor must have constant shape");
+  assert(outputRank != ShapedType::kDynamic &&
```
tungld, line 39: Good catch! Thanks!

tungld (approved):
LGTM! Thanks for the patch!

tungld:
@jorickert could you update this PR with the latest main branch so that we can merge it? Somehow I cannot do it.

jorickert:
> @jorickert could you update this PR with the latest main branch so that we can merge it? Somehow I cannot do it.

I will on Monday

tungld:

AlexandreEichenberger:
@jorickert do you want this PR to be merged? Let us know and we will merge it.

jorickert:
@AlexandreEichenberger Yes, please merge

AlexandreEichenberger:
@jorickert Thanks for your contributions, much appreciated.

Issue #2822, opened 2024-05-09
Hugging Face GPT2 models segfault on CPU


cjvolzka:
Running the [HuggingFace openai-community/gpt2](https://huggingface.co/openai-community/gpt2) (and gpt2-large and gpt2-xl variants) compiles but returns a segfault when run on CPU. The model runs fine when compiled for NNPA.

### Export Model
I converted the models to onnx using the [Hugging Face optimum-cli](https://github.com/huggingface/optimum). The optimum-cli does not work on s390x so I converted the model on my Mac and then transferred the exported model to a Linux on Z host  to compile it.
```
model_name=gpt2
opset=13
task=text-generation
optimum-cli export onnx --model ${model_name} --framework pt --atol 0.001 --task ${task} --opset ${opset} ${model_name}-${task}-${opset}
```
Change `model_name` to `gpt-large`, etc to export other variants

### Compile Mode
And afterward compiled the model with 
* CPU
    * `--O3 --EmitLib --mtriple=s390x-ibm-loz --mcpu=z14 model.onnx --onnx-op-stats TXT --profile-ir=Onnx` 
* NNPA
    * `--O3 --EmitLib --mtriple=s390x-ibm-loz --mcpu=z16 --maccel=NNPA model.onnx --onnx-op-stats TXT --profile-ir=ZHigh`

### Run model
I used a C++ client to run the model. For the inputs, I encoded the second paragraph of Les Miserables (see
[inputs.txt](https://github.com/onnx/onnx-mlir/files/15256176/inputs.txt) for values used)

The models compiled for NNPA (except gpt2-xl, optset 13) run without issue. The CPU compiled version (and gpt2-xl, opset 13, NNPA) appear to fail at the same spot based on the profile output:
```
...
==PERF-REPORT==, onnx.Squeeze, /transformer/h.0/attn/Squeeze_2, after, 0.000005, 0.271208
==PERF-REPORT==, onnx.Sub, /transformer/h.0/attn/Sub, before, 0.000004, 0.271212
==PERF-REPORT==, onnx.Sub, /transformer/h.0/attn/Sub, after, 0.000004, 0.271216
==PERF-REPORT==, onnx.Unsqueeze, /transformer/h.0/attn/Unsqueeze_6, before, 0.000004, 0.271220
==PERF-REPORT==, onnx.Unsqueeze, /transformer/h.0/attn/Unsqueeze_6, after, 0.000004, 0.271224
[...]

PR #2865, opened 2024-07-05
Avoid using StringAttr when lowering dense constant with i1 type

imaihal:
By using this PR, segfault on GPT2 (https://github.com/onnx/onnx-mlir/issues/2822) can be avoided. Still not sure the reason why we can avoid it, but from the tests using following test cases, it seems we can not use the StringAttr for dense constant with i1 (or need to do something to use i1).

- Example including single ConstantOp with i1 type
[model.onnx.const-i1.mlir.txt](https://github.com/user-attachments/files/16104573/model.onnx.const-i1.mlir.txt)
This MLIR file includes only single ConstantOp with dense attribute of i1 type. When running this using following commands, the results were strange (wrong results, segfault, bus error) in main branch. This happened both MaxOS and z16. By using this PR, we got expected results.

```
ONNX_MLIR_HOME=${ONNX_MLIR_ROOT}/build/Debug python ${ONNX_MLIR_ROOT}/utils/RunONNXModel.py --model model.onnx.const-i1.mlir --compile-args "-O3 " --print-output
```

- GPT-2 model on MaxOS
   Numerical outputs when using the main branch were very different compared to the onnxruntime outputs, but using this PR generated correct outputs (using atol=0.01, rtol=0.05),

### DIFF
--- a/src/Conversion/KrnlToLLVM/KrnlGlobal.cpp
+++ b/src/Conversion/KrnlToLLVM/KrnlGlobal.cpp
@@ -177,7 +177,9 @@ private:
     uint64_t sizeInBytes = computeSizeInBytes(krnlGlobalOp);
     LLVM::GlobalOp global;
     if (!(mlir::isa<StringType>(denseAttr.getElementType())) &&
-        (!denseAttr.isSplat()) && (sizeInBytes > 1024)) {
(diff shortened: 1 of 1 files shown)


tungld (approved):
LGTM.

PR #2867, opened 2024-07-09
Set input onnx model in RunONNXModel.py when running with shared lib and verifying using onnxruntime

imaihal:
When running with shared library (`--load-so`) and verifying results with onnxruntime (`--verify onnxruntime`), input onnx model (input_model_path) was not set to run onnxruntime. This PR set it and adds error handling.


AlexandreEichenberger (approved):
LGTM, thanks

PR #2871, opened 2024-07-10
Upgraded NNPA infra for handling dimension specific max sizes

AlexandreEichenberger:
Upgraded NNPA infra for handling dimension specific max sizes. Migrated some infra around to support the new framework.


AlexandreEichenberger:
@tungld any reason that the `NNPALimit.h` is a `.h` as opposed to a `.hpp`? it seems to be only used in cpp files.

tungld:
> any reason that the NNPALimit.h is a .h as opposed to a .hpp? it seems to be only used in cpp files.

It was in a different folder that contains wrapper files for zdnn so .h. Now we can change it to .hpp. Thanks!

Issue #2821, opened 2024-05-09
Models take significant amounts of memory to compile


cjvolzka:
While compiling models like [HuggingFace protectai/xlm-roberta-base-language-detection-onnx](https://huggingface.co/protectai/xlm-roberta-base-language-detection-onnx) or [mistralai/Mistral-7B-v0.1](https://huggingface.co/mistralai/Mistral-7B-v0.1) I notice we take significantly larger amounts of memory than the entire model size during compiling. 

For example, the xlm-roberta-base-language-detection-onnx is about 1.11GB but during compile time I see peaks up to 9GB of memory used by `onnx-mlir`, `opt` and `llc` compiling with `--O3 --EmitLib --mtriple=s390x-ibm-loz --mcpu=z14 --onnx-op-stats TXT`.

The Mistral-7B-v0.1 model is about 29GB but during compile time I see peaks up to 70+Gb and sustained 58GB memory compiling with  `--O3 --EmitLib --mtriple=s390x-ibm-loz --mcpu=z14 --store-constants-to-file --onnx-op-stats TXT`

Is there anything that can be done to reduce the compile time memory required for these kind of models?

imaihal:
@cjvolzka How can we get onnx model for `Mistral-7B-v0.1` ?

cjvolzka:
@imaihal Sorry, I missed your question. Below is how I generated the Mistral onnx model. 

Notes: 
* I exported the model using my Mac as the tools don't support s390x. Afterward, I transferred the folder it created (with the onnx file and constants) to the s390x host to compile the model. 
* the `huggingface-cli` comand will ask a couple of questions:
    * Use https://huggingface.co/settings/tokens to generate the token it requests
    * You don't need to add the token as a git credential

```
pip install huggingface_cli optimum
huggingface-cli login
optimum-cli export onnx --model mistralai/Mistral-7B-v0.1 --framework pt --atol 0.001 --task text-generation Mistral-7B-v0.1-text-generation
```

PR #2874, opened 2024-07-12
Write a constant value to a file without buffering

imaihal:
This PR partially solves an issue of memory consumption reported in https://github.com/onnx/onnx-mlir/issues/2821 

We found that there is a spike in memory consumption when writing a constant into a file (`model.constants.bin`). This is because all constants into a buffer once and write it to a file at once. This PR changes to write the constant to the file without the buffering. This removes the spike in memory consumption.

### DIFF
--- a/src/Conversion/KrnlToLLVM/ConvertKrnlToLLVM.cpp
+++ b/src/Conversion/KrnlToLLVM/ConvertKrnlToLLVM.cpp
@@ -52,6 +52,7 @@
 #include "llvm/ADT/TypeSwitch.h"
 #include "llvm/Support/Debug.h"
 #include "llvm/Support/Endian.h"
+#include "llvm/Support/FileSystem.h"
 #include "llvm/Support/Path.h"
 
 #include "onnx/onnx_pb.h"
@@ -520,8 +521,11 @@ bool extractConstantsToFile(ModuleOp &module, std::string filepath,
   // The file will be mmaped later at runtime and aligned at the page boundary,
   // So every constants must be correctly aligned in the packed constant. Pads
   // are added if necessary.
-  std::vector<char> packedConst;
+  llvm::sys::fs::remove(filepath);
+  std::ofstream outfile(filepath, std::ios::app | std::ios::binary);
+  int64_t totalConstSize = 0;
   for (int64_t i = globalOfInterest.size() - 1; i >= 0; --i) {
+    std::vector<char> packedConst;
     KrnlGlobalOp op = globalOfInterest[i];
     ArrayRef<char> rawData = getRawData(op);
 
@@ -531,26 +535,24 @@ bool extractConstantsToFile(ModuleOp &module, std::string filepath,
       alignment = op.getAlignment().value();
 
     // Padding if necessary.
-    if ((alignment > 0) && (packedConst.size() % alignment != 0)) {
+    if ((alignment > 0) && (totalConstSize % alignment != 0)) {
       uint64_t padSize =
-          ((uint64_t)(packedConst.size() / alignment) + 1) * alignment -
-          packedConst.size();
+          ((uint64_t)(totalConstSize / alignment) + 1) * alignment -
+          totalConstSize;
       SmallVector<char> pads(padSize, (char)0);
       packedConst.insert(packedConst.end(), pads.begin(), pads.end());
     }
 
-    op.setOffsetAttr(b.getI64IntegerAttr(packedConst.size()));
+    op.setOffsetAttr(b.getI64IntegerAttr(totalConstSize));
     op.removeValueAttr();
     packedConst.insert(packedConst.end(), rawData.begin(), rawData.end());
+    outfile.write(packedConst.data(), packedConst.size());
+    totalConstSize += packedConst.size();
   }
 
   // No constant statisfying thresholds, do not store constants to file.
-  if (packedConst.empty())
+  if (totalConstSize == 0)
     return false;
-
-  // Save to file.
-  std::ofstream outfile(filepath, std::ofstream::binary);
-  outfile.write(packedConst.data(), packedConst.size());
 
   // Create a global op to store the filename in the IR.
   OpBuilder::InsertionGuard guard(b);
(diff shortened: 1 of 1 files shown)


AlexandreEichenberger:
It looks good, I was interested to see if using mmap would be faster that the `ofstream.write`. A quick benchmark run on z16 gave me the following.

```
Test with size: 128
File write duration: 0.0127008 seconds, BW: 10078.1 b/s
mmap write duration: 0.01229 seconds, BW: 10415 b/s

Test with size: 1024
File write duration: 0.0122701 seconds, BW: 83454.8 b/s
mmap write duration: 0.0121558 seconds, BW: 84239.6 b/s

Test with size: 262144
File write duration: 0.00024563 seconds, BW: 1.06723e+09 b/s
mmap write duration: 0.000119349 seconds, BW: 2.19645e+09 b/s

Test with size: 1048576
File write duration: 0.000903089 seconds, BW: 1.1611e+09 b/s
mmap write duration: 0.000410608 seconds, BW: 2.55372e+09 b/s

Test with size: 10485760
File write duration: 0.00851007 seconds, BW: 1.23216e+09 b/s
mmap write duration: 0.0046617 seconds, BW: 2.24934e+09 b/s

Test with size: 104857600
File write duration: 0.0864244 seconds, BW: 1.21329e+09 b/s
mmap write duration: 0.0501027 seconds, BW: 2.09285e+09 b/s
```

Where tests were done for 128B, 1KB, 256KB, 1MB, 10MB, and 100MB. For all but the smallest sizes, `mmap` is 1.8x faster.  Now this test wrote the whole buffer in one go, but it is possible to mmap a portion of the file.

We can go with this PR as of now. It would be interesting to see what are the average size of the constant being written (maybe add a printf and run it through the large benchmark), and let see if might be worthwhile to buffer the writes using mmap.

imaihal:
@AlexandreEichenberger Thanks for you suggestion! This PR focuses on peak memory reduction. May I use/try mmap in a follow-up PR focused on performance?

AlexandreEichenberger:
> May I use/try mmap in a follow-up PR focused on performance?

For sure. When I saw your PR, I just got curious.

src/Conversion/KrnlToLLVM/ConvertKrnlToLLVM.cpp
```diff
@@ -515,13 +516,16 @@ bool extractConstantsToFile(ModuleOp &module, std::string filepath,
...
-  // Pack all constants into a single buffer in order to save to file.
+  // Store each constant into single file.
   // Constants with the highest alignment will be packed first in the file.
   // The file will be mmaped later at runtime and aligned at the page boundary,
   // So every constants must be correctly aligned in the packed constant. Pads
   // are added if necessary.
-  std::vector<char> packedConst;
+  llvm::sys::fs::remove(filepath);
+  std::ofstream outfile(filepath, std::ios::app | std::ios::binary);
+  int64_t totalConstSize = 0;
```
tungld, line 526: `uint64_t` is more suitable.

src/Conversion/KrnlToLLVM/ConvertKrnlToLLVM.cpp
```diff
@@ -515,13 +516,16 @@ bool extractConstantsToFile(ModuleOp &module, std::string filepath,
...
   for (int64_t i = globalOfInterest.size() - 1; i >= 0; --i) {
+    std::vector<char> packedConst;
```
tungld, line 528: The name `packedConst` is obsoleted in the new context since we donnot pack constants anymore. Maybe change it to `paddedConstant` or it can be totally removed (See the below comment)

src/Conversion/KrnlToLLVM/ConvertKrnlToLLVM.cpp
```diff
@@ -531,27 +535,25 @@ bool extractConstantsToFile(ModuleOp &module, std::string filepath,
...
     // Padding if necessary.
-    if ((alignment > 0) && (packedConst.size() % alignment != 0)) {
+    if ((alignment > 0) && (totalConstSize % alignment != 0)) {
       uint64_t padSize =
-          ((uint64_t)(packedConst.size() / alignment) + 1) * alignment -
-          packedConst.size();
+          ((uint64_t)(totalConstSize / alignment) + 1) * alignment -
+          totalConstSize;
       SmallVector<char> pads(padSize, (char)0);
       packedConst.insert(packedConst.end(), pads.begin(), pads.end());
```
tungld, line 543: We can reduce memory more by not using a padded buffer but two writes to a file: one for padding and one for rawData. For example.

```c
// Padding the current data for correct alignment.
uint64_t padSize = 0;
if ((alignment > 0) && (totalConstSize % alignment != 0)) {
 padSize = ((uint64_t)(totalConstSize / alignment) + 1) * alignment - totalConstSize;
 SmallVector<char> pads(padSize, (char)0);
 outfile.write(pads.data(), pads.size());
}

// Write the constant to a file.
uint64_t constSize = rawData.size();
outfile.write(rawData.data(), constSize);

// Update the total size and op's offset.
totalConstSize += padSize + constSize;
op.setOffsetAttr(b.getI64IntegerAttr(totalConstSize));
op.removeValueAttr();

```

AlexandreEichenberger: Smart, please do that.

imaihal: Thanks for your comments! I updated. 

`totalConstSize` in `op.setOffsetAttr(b.getI64IntegerAttr(totalConstSize))` is offset. So the `totalConstSize` before adding `constSize` should be used.

src/Conversion/KrnlToLLVM/ConvertKrnlToLLVM.cpp
imaihal: Done. Thanks!

src/Conversion/KrnlToLLVM/ConvertKrnlToLLVM.cpp
tungld: Thanks!

tungld (approved):
LGTM.

PR #2875, opened 2024-07-12
Recompose QLinearMatMul and remove Quantize-Dequantize pairs

tungld:
Add two patterns found when doing static quantization for bert-base models using onnxruntime.

- 1st pattern: recompose `DequantizeLinear - MatMul - QuantizeLinear` back to `QLinearMatmul`
- 2nd pattern: remove `QuantizeLinear-DequantizeLinear` pairs

### DIFF
--- a/test/mlir/driver/compile_phases.mlir
+++ b/test/mlir/driver/compile_phases.mlir
@@ -1,4 +1,4 @@
-// RUN: onnx-mlir %s | FileCheck %s
+// RUN: onnx-mlir %s -o %t| FileCheck %s && rm %t.so
 
 // CHECK: [1/5] {{.*}} Importing ONNX Model to MLIR Module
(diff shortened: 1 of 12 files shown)


test/mlir/driver/compile_phases.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: onnx-mlir %s | FileCheck %s
+// RUN: onnx-mlir %s -o %t| FileCheck %s && rm %t.so
```
tungld, line 1: Remove the generated .so during `make check-onnx-lit`.

AlexandreEichenberger:
@tungld Can you look into @mgehre-amd comment and fix appropriately? Thanks

tungld:
@mgehre-amd thanks for pointing it out! I created a PR to revert that pattern: https://github.com/onnx/onnx-mlir/pull/2952. Thanks!

PR #2878, opened 2024-07-16
 Optimize compare -> select to a min / max for better SIMD performance

AlexandreEichenberger:
SIMD engines may have a vector min/max operation that is more efficient than a vector compare followed by a vector select. Some backend may have an optimization that transform the second pattern into the first one, but rather than relying on the backend, simply changed the original code pattern to the more efficient formulation.

This optimization reduced the overhead of saturation (clip) on stickify from 20% to 3%. Assume that it may similarly optimized these other patterns.

Note that this typically only help SIMD code, though some CPUs may have min/max instructions as well.

Saw that some operations were currently not simdized, such as softmax, dynamic quantized,... might be something to consider if key to performance.


tungld (approved):
LGTM. Good finding! Thanks!

PR #2879, opened 2024-07-17
[NNPA] Add a compiler option for saturation stickify inputs

tungld:
This patch adds a compiler option `--nnpa-saturation` to saturate all stickify inputs if enabled. Currently it is off by default so that it does not affect the current code. When enabled, it affects the path of compiler-generated stickification because the compiler-generated stickification supports saturation.


src/Accelerators/NNPA/Dialect/ZHigh/ZHigh.td
```diff
@@ -142,16 +142,26 @@ def ZHighStickOp:ZHigh_Op<"Stick", [Pure,
...
     If `layout`=`NHWC`, input must be in `NCHW` and output will be in `NHWC`.
+
+    Optional `saturation` indicates whether the CPU tensor is saturated before stickification
+    or not. If it is saturated, the dlfloat16 range would be used.
+    Saturation if off if `saturation == 0` or it is not given. Otherwise, it is on.
   }];
   let arguments = (ins AnyTypeOf<[TensorOf<[F32]>, NoneType]>:$In,
-                       OptionalAttr<StrAttr>:$layout);
+                       OptionalAttr<StrAttr>:$layout,
+                       OptionalAttr<SI64Attr>:$saturation);
```
tungld, line 154: Make it optional so that the lit tests are unchanged.

AlexandreEichenberger: smart

src/Accelerators/NNPA/Transform/ZLow/ZLowStickExpansion.cpp
```diff
@@ -330,12 +327,8 @@ class StickExpansionPattern : public OpRewritePattern<ZLowStickOp> {
...
     Value input = stickOp.getX();
     Value alloc = stickOp.getOut();
-
-    bool saturation = false;
-#if SATURATION_ON
-    // TODO: hook to operation's attribute.
-    saturation = true;
-#endif
+    std::optional<int64_t> saturationOpt = stickOp.getSaturation();
+    bool saturation = saturationOpt.has_value() && saturationOpt.value() != 0;
```
AlexandreEichenberger, line 331: Thanks!

AlexandreEichenberger (approved):
LGTM

PR #2881, opened 2024-07-18
Setting enable-compiler-stick-unstick default, enable also if saturation is on

AlexandreEichenberger:
Saturation of stick requires at this time compiler generated stick/unstick.

Due to the better performance of compiler generated stick/unstick, I also propose to turn it on by default.

### DIFF
--- a/src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
+++ b/src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
@@ -48,6 +48,9 @@ namespace onnx_mlir {
 
 void configurePassesNNPA() {
   configureOnnxToZHighLoweringPass(optReport == OptReport::NNPAUnsupportedOps);
+  // Compiler generated sticks supports saturation, so force its usage.
+  if (nnpaEnableSaturation)
+    nnpaEnableCompilerStickUnstick = true;
 }
 
 void addONNXToZHighPasses(mlir::PassManager &pm) {
--- a/src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
+++ b/src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
@@ -588,7 +588,7 @@ template <>
 double analyzeSimdFor<ONNXHardSigmoidOp>(
     Type t, Operation *op, int64_t &von, int64_t &son) {
   return simdAnalysis(
-      {GenericOps::ArithmeticGop, GenericOps::MulGop}, {2, 1}, t, von, son);
+      {GenericOps::ArithmeticGop, GenericOps::MulGop}, {3, 1}, t, von, son);
 }
 
 template <>
--- a/src/Accelerators/NNPA/Compiler/NNPACompilerOptions.cpp
+++ b/src/Accelerators/NNPA/Compiler/NNPACompilerOptions.cpp
@@ -49,6 +49,8 @@ llvm::cl::opt<bool> nnpaEnableZHighDecomposeStickUnstick(
         "Default is false."),
     llvm::cl::init(false), llvm::cl::cat(OnnxMlirOptions));
 
+// Enabled default now, could also enable it only if parallel is on as parallel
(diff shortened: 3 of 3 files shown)


src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -588,7 +588,7 @@ template <>
 double analyzeSimdFor<ONNXHardSigmoidOp>(
     Type t, Operation *op, int64_t &von, int64_t &son) {
   return simdAnalysis(
-      {GenericOps::ArithmeticGop, GenericOps::MulGop}, {2, 1}, t, von, son);
+      {GenericOps::ArithmeticGop, GenericOps::MulGop}, {3, 1}, t, von, son);
```
AlexandreEichenberger, line 591: realized that I mis-counted the ops in this operation.

src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
```diff
@@ -48,6 +48,9 @@ namespace onnx_mlir {
 
 void configurePassesNNPA() {
   configureOnnxToZHighLoweringPass(optReport == OptReport::NNPAUnsupportedOps);
+  // Compiler generated sticks supports saturation, so force its usage.
+  if (nnpaEnableSaturation)
+    nnpaEnableCompilerStickUnstick = true;
```
tungld, line 53: Since `--nnpa-enable-compiler-stick-unstick` is enabled by default, it seems we don't need to force this so that we can have "nnpa-saturation=true --nnpa-enable-compiler-stick-unstick=false" if NNPA supports stick/unstick.

test/mlir/accelerators/nnpa/driver/ccfd.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: ccfd=$(dirname %s)/ccfd.onnx && curl -L https://github.com/IBM/ai-on-z-fraud-detection/raw/main/onnx%20models/ccf_lstm_static_tf2onnx_OS_new.onnx -o ${ccfd} && onnx-mlir --mcpu=z16 --maccel=NNPA --EmitMLIR --printIR -tag="test" ${ccfd} | FileCheck %s && rm -rf ${ccfd}
+// RUN: ccfd=$(dirname %s)/ccfd.onnx && curl -L https://github.com/IBM/ai-on-z-fraud-detection/raw/main/onnx%20models/ccf_lstm_static_tf2onnx_OS_new.onnx -o ${ccfd} && onnx-mlir --mcpu=z16 --maccel=NNPA --enable-compiler-stick-unstick=false --EmitMLIR --printIR -tag="test" ${ccfd} | FileCheck %s && rm -rf ${ccfd}
```
tungld, line 1: I think you have to set `--enable-compiler-stick-unstick` to false because `--EmitMLIR` includes generating compiler stick/unstick. If we can emit an IR before generating compiler stick/unstick, we don't have to do that. `--EmitZLowIR` is before `--EmitMLIR`, but `--EmitZLowIR` does not include memref normalization and optimizations to product the optimized code here. Perhaps, we can introduce a new --EmitXXXIR for the IR just before converting to affine. Anyway, it would be for another PR.

AlexandreEichenberger: So what do you suggest for now?

tungld: Nothing to do at this moment.

src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
tungld: > Does it make sense to add a "TODO, remove this if zDNN supports saturation"

That sounds good if you agree.

AlexandreEichenberger: of course.

tungld (approved):
LGTM. Thanks!

PR #2882, opened 2024-07-18
Correct log2u in OMSort

mikeessen:
This PR corrects an error in OMSort where the log2u function is incorrectly computing the log base 2.


AlexandreEichenberger (approved):
LGTM, good catch

PR #2886, opened 2024-07-24
[STABLEHLO] Added Softmax Op conversion support to StableHLO dialect

Abhishek-TyRnT:
I have added support for SoftmaxOp lowering to stable hlo Dialect. Currently it only support Float32 dtype and static shapes. Can somebody please confirm please review this.

NOTE :- I as of now haven't made formatting changes, so clang-format might fail. Also suggest me some more ways to test it. As of now I have tested [softmax.mlir](https://github.com/onnx/onnx-mlir/blob/main/test/mlir/conversion/onnx_to_stablehlo/Math/Softmax.mlir) and [gptneox_Opset16](https://github.com/onnx/models/blob/main/Generative_AI/gptneox_Opset16_transformers/gptneox_Opset16.onnx)

As of now I am creating a draft PR, will create full PR after approval

### DIFF
--- a/src/Conversion/ONNXToStablehlo/Math/Softmax.cpp
+++ b/src/Conversion/ONNXToStablehlo/Math/Softmax.cpp
@@ -14,8 +14,8 @@
 #include "src/Conversion/ONNXToStablehlo/DialectBuilder.hpp"
 #include "src/Conversion/ONNXToStablehlo/ONNXToStablehloCommon.hpp"
 #include "src/Dialect/ONNX/ONNXOps/ShapeHelper.hpp"
+#include "src/Support/TypeUtilities.hpp"
 #include "stablehlo/dialect/BroadcastUtils.h"
-#include <iostream>
 
 using namespace mlir;
 
@@ -26,18 +26,11 @@ namespace {
 Value getReductionShapeValue(Location loc, PatternRewriter &rewriter,
     Value operand, llvm::SmallVector<int64_t, 4> axes, bool keepDims) {
   int64_t rank = mlir::cast<RankedTensorType>(operand.getType()).getRank();
-  // Mark reduction axes.
-  llvm::SmallVector<bool, 4> isReductionAxis;
-  for (int64_t i = 0; i < rank; ++i) {
-    if (std::find(axes.begin(), axes.end(), i) != axes.end())
-      isReductionAxis.push_back(true);
-    else
-      isReductionAxis.push_back(false);
-  }
+
   Value inputShape = rewriter.create<shape::ShapeOfOp>(loc, operand);
   SmallVector<Value> dims;
   for (int64_t i = 0; i < rank; i++) {
-    if (!isReductionAxis[i]) {
+    if (!(std::find(axes.begin(), axes.end(), i) != axes.end())) {
       Value dim = rewriter.create<shape::GetExtentOp>(loc, inputShape, i);
       dims.push_back(dim);
     } else if (keepDims) {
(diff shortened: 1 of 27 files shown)


AlexandreEichenberger:
(updated to the latest to start a run)

Abhishek-TyRnT:
One of the softmax checks fails cause I haven't added support for dyanmic shapes and commented that part in mlir, but it still tries to find the test. What should I do about it?

AlexandreEichenberger:
> One of the softmax checks fails cause I haven't added support for dyanmic shapes and commented that part in mlir, but it still tries to find the test. What should I do about it?

Is there a way to disable that specific test (aka for that operation) only for the HLO path?

test/mlir/conversion/onnx_to_stablehlo/Math/Softmax.mlir
```diff
@@ -27,10 +27,10 @@ func.func @test_softmax(%arg0 : tensor<10x20x30xf32>) -> tensor<10x20x30xf32> {
 
 // -----
 
-func.func @test_softmax_dynamic(%arg0 : tensor<?x20x30xf32>) -> tensor<?x20x30xf32> {
-  %0 = "onnx.Softmax"(%arg0) {axis = 1: si64} : (tensor<?x20x30xf32>) -> tensor<?x20x30xf32>
-  "func.return"(%0) : (tensor<?x20x30xf32>) -> ()
-}
+//func.func @test_softmax_dynamic(%arg0 : tensor<?x20x30xf32>) -> tensor<?x20x30xf32> {
```
AlexandreEichenberger, line 30: Commenting the function is not enough, you need also to "disable" the littest that follows.
Best would be to keep the actual function, but remove the `CHECK-SAME` line below... or rename that line to something like
```
// TODO, reenable the check same for the following string
// ([[PARAM_0_:%.+]]: tensor<?x20x30xf32>) -> tensor<?x20x30xf32> {
```

basically removing the `CHECK-Same` string that is parsed by the lit test software

AlexandreEichenberger:
Do you mind fixing the format first? It will be needed to merge anyway. 
https://clang.llvm.org/docs/ClangFormat.html
Most of us use a plugin in VSCode, it can also be run individually.
See here for which files need to be fixed. https://github.com/onnx/onnx-mlir/actions/runs/10078441752/job/27863357309?pr=2886

AlexandreEichenberger:

tungld:

src/Conversion/ONNXToStablehlo/Math/Softmax.cpp
```diff
@@ -0,0 +1,226 @@
...
+  llvm::SmallVector<bool, 4> isReductionAxis;
+  for (int64_t i = 0; i < rank; ++i) {
+    if (std::find(axes.begin(), axes.end(), i) != axes.end())
+      isReductionAxis.push_back(true);
+    else
+      isReductionAxis.push_back(false);
+  }
+
+  for (int64_t i = 0; i < rank; ++i) {
+    if (!isReductionAxis[i])
```
AlexandreEichenberger, line 149: nit: Would you be able to skip the `isReductionAxis` and directly put the `! (std::find(axes.begin(), axes.end(), i) != axes.end())` test here? 

See the same behavior above.

src/Conversion/ONNXToStablehlo/Math/Softmax.cpp
```diff
@@ -0,0 +1,226 @@
...
+  if (keepDims) {
+    Value reduceShapeValue =
+        getReductionShapeValue(loc, rewriter, operand, axes, true);
+    result = rewriter.create<stablehlo::DynamicReshapeOp>(
+        loc, outputType, result, reduceShapeValue);
+  }
+  return result;
+}
+
+bool hasStaticShape(Value val) {
```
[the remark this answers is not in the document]
Abhishek-TyRnT: Okay, done

src/Conversion/ONNXToStablehlo/Math/Softmax.cpp
Abhishek-TyRnT: Done, please check

AlexandreEichenberger (approved):
LGTM, I will assume that you have verified the HLO specific part with HLO experts.

AlexandreEichenberger:
Thanks @Abhishek-TyRnT for your continued contributions to HLO, much appreciated

PR #2888, opened 2024-07-25
Disable compliler stick/unstick on zOS

AlexandreEichenberger:
Currently we have issues generating the asm instructions needed for stick/unsick on zOS.

This PR disable this feature until we solve this issue

Issue is tracked in a separate issue


mikeessen (approved):
Looks good!  Thanks for the patch!

AlexandreEichenberger:
See issue #2889, once that one is done, we should revert the change here.

PR #2890, opened 2024-07-25
Remove Pragma Once

hamptonm1:
Removing Pragma Once and replacing it with include guards to follow IBM set precedent which can be found here:
https://www.ibm.com/docs/en/zos/2.1.0?topic=messages-ccn8735

```
GCC supports several types of pragmas, primarily in order to compile code originally written for other compilers. Note that in general we do not recommend the use of pragmas; See [Declaring Attributes of Functions](https://gcc.gnu.org/onlinedocs/gcc/extensions-to-the-c-language-family/declaring-attributes-of-functions.html#function-attributes), for further explanation.

The GNU C preprocessor recognizes several pragmas in addition to the compiler pragmas documented here. Refer to the CPP manual for more information.
```

https://gcc.gnu.org/onlinedocs/gcc/extensions-to-the-c-language-family/pragmas-accepted-by-gcc.html#


src/Dialect/ONNX/ONNXOps/OpHelper.hpp
```diff
@@ -4,15 +4,16 @@
 
 //===------- ONNXOpsHelper.hpp - Helper functions for ONNX dialects -------===//
```
mikeessen, line 5: This is a nit, but the prolog "ONNXOpsHelper.hpp" does not match the file name "OpHelper.hpp"

hamptonm1: Thanks I agree, that was there prior. Some files need to be cleaned up.

src/Compiler/DisposableGarbageCollector.hpp
```diff
@@ -8,7 +8,8 @@
 //
 //===----------------------------------------------------------------------===//
 
-#pragma once
+#ifndef ONNX_MLIR_GARABAGE_COLLECTOR_H
```
mikeessen, line 11: GARABAGE misspelled

hamptonm1: Thanks! Will fix

mikeessen (approved):
Approved with minor comments.  Thanks!

src/Compiler/DisposableGarbageCollector.hpp
hamptonm1: Fixed

AlexandreEichenberger:
@hamptonm1 thanks for the upgrade, much appreciated

PR #2891, opened 2024-07-29
Document for building llvm-project with omp (WIP)

chentong319:
Small PR to record how to build llvm-project with omp on.


docs/BuildOnLinuxOSX.md
```diff
@@ -23,17 +23,29 @@ cd llvm-project && git checkout 0913547d0e3939cc420e88ecd037240f33736820 && cd .
 mkdir llvm-project/build
 cd llvm-project/build
 cmake -G Ninja ../llvm \
-   -DLLVM_ENABLE_PROJECTS=mlir \
+   -DLLVM_ENABLE_PROJECTS="mlir;clang;openmp" \
```
AlexandreEichenberger, line 26: @chentong319 you cannot change info here without also changing it in the `build-mlic.sh` because of the `[same-as-file]` annotation.

I would recommend leaving both untouched, but maybe adding lines below just indicating what to change/add. That way, this would not require new testing. Or make a new `build-mlic-parallel.sh` script and add a new subsection on how to build it with parallel (text and script).

chentong319: I just modified build-mlir.sh  to see what will happen on different system. In the end, we may not want to change the build in this PR.

chentong319: I changed  doc and script. Gong said that it is not the time to change docker file.

AlexandreEichenberger (approved):

AlexandreEichenberger:
@chentong319 thanks for this

PR #2892, opened 2024-07-31
Add NHWC as target for compiler enabled stickification

AlexandreEichenberger:
Since NHWC is similar to 4D, add it. Did not add the HWCK as this is typically for constants and want to preserve the high level constant propagation of constant values into the Convolution filters.

Saw that during stckification, I was doing a minimum of 64 conversions; but convolutions sometimes have very small E1 dimensions (e.g. 3). So it was wasteful. I now check that if E1 is constant and very small, I do a maximum of `E1<U*VL` iterations where U is as small as possible.

Not needed for unstick as there we must do the exact number of conversions.


tungld (approved):
LGTM

PR #2893, opened 2024-08-01
Transpose NCHW to NHWC if compiler stick/unstick is enabled and enable model zoo tests on NNPA

tungld:
The compiler-generated stick/unstick now supports NHWC layout, but not NCHW layout. To support NCHW, we explicitly do NCHW-NHWC transformation before stick and after unstick.

To get correct results for convolutional models in the backend test, `--nnpa-saturation` is needed. So this patch also enables saturation in the NNPA backend test. We may consider to enable saturation by default.

### DIFF
--- a/src/Accelerators/NNPA/Compiler/NNPACompilerOptions.cpp
+++ b/src/Accelerators/NNPA/Compiler/NNPACompilerOptions.cpp
@@ -29,7 +29,8 @@ llvm::cl::opt<NNPAEmissionTargetType> nnpaEmissionTarget(
 llvm::cl::opt<bool> nnpaClipToDLFloatRange("nnpa-clip-to-dlfloat-range",
     llvm::cl::desc("Clip CPU tensors to dlfloat range before stickification to "
                    "avoid out-of-range. Only clip Softmax inputs at this "
-                   "moment. Default is true."),
+                   "moment. Default is true. This option will be removed and "
+                   "replaced by --nnpa-saturation in the future."),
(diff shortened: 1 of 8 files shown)


src/Accelerators/NNPA/Compiler/NNPACompilerOptions.cpp
```diff
@@ -29,7 +29,8 @@ llvm::cl::opt<NNPAEmissionTargetType> nnpaEmissionTarget(
 llvm::cl::opt<bool> nnpaClipToDLFloatRange("nnpa-clip-to-dlfloat-range",
     llvm::cl::desc("Clip CPU tensors to dlfloat range before stickification to "
                    "avoid out-of-range. Only clip Softmax inputs at this "
-                   "moment. Default is true."),
+                   "moment. Default is true. This option will be removed and "
+                   "replaced by --nnpa-saturation in the future."),
```
AlexandreEichenberger, line 33: Clip would be redundant only if the softmax is preceded by a stick-with-saturation. Maybe that is the case in the models we see... or maybe the saturation of the entire model transitively prevent the values to be out of bound once they reach softmax

AlexandreEichenberger (approved):
LGTM

hamptonm1 (approved):
LGTM! Thanks @tungld

src/Accelerators/NNPA/Compiler/NNPACompilerOptions.cpp
tungld: Thanks for the suggestion! Will investigate more.

Issue #2901, opened 2024-08-06
Mismatched name in function and entry point crash the compiler


AlexandreEichenberger:
When compiling this
```mlir
module attributes {} {
  func.func @main_graph(%arg0 : tensor<?x?xf32>, %arg1 : tensor<f32>, %arg2 : tensor<ui8>) -> tensor<*xui8> {
    %1 = "onnx.QuantizeLinear"(%arg0, %arg1, %arg2) {} : (tensor<?x?xf32>, tensor<f32>, tensor<ui8>) -> tensor<*xui8>
      return %1: tensor<*xui8>
  }
  "onnx.EntryPoint"() {func = @main_graph_different} : () -> ()
}
```
the program crashes because the `func.func` and `EntryPoint` do not use the same name.

This should not occur.

AlexandreEichenberger:
@Sunny-Anand example of segfault.

chentong319:
To my understanding, the two name should be the same.  I assume that the inconsistence is introduced by editing the .mlir file manually. We can add a check at compile time or runtime to handle the case.

AlexandreEichenberger:
> We can add a check at compile time or runtime to handle the case.

yes, it should not crash the compiler.

chentong319:
PR #2902 is created to fix this issue.

PR #2902, opened 2024-08-07
Fix issue 2901

chentong319:
Add a check for the name of EntryPointOp to fix the issue #2901. Compiler will get an assertion error, instead of crash.


AlexandreEichenberger (approved):
LGTM, thanks

PR #2904, opened 2024-08-08
Simdized quantized operations

AlexandreEichenberger:
Simdized quantized operations: DynamicQuantizeLinear, QuantizedLinear, and DequantizeLinear.

Added support for reduction to a scalar (current scheme for our tensor-only quantization), fused reduction of min an max needed for dynamic quantization, and added a generic support in KrnlBuilder to generate SIMD loops.

Also added `MathBuilder` support for `clip` and `round` so that we don't need to rely on `onnx` operators to do so when lowering to Krnl.

### DIFF
--- a/docs/BuildOnLinuxOSX.md
+++ b/docs/BuildOnLinuxOSX.md
@@ -15,7 +15,7 @@ Firstly, install MLIR (as a part of LLVM-Project):
 ``` bash
 git clone -n https://github.com/llvm/llvm-project.git
 # Check out a specific branch that is known to work with ONNX-MLIR.
-cd llvm-project && git checkout 0913547d0e3939cc420e88ecd037240f33736820 && cd ..
+cd llvm-project && git checkout 60a7d33106d3cd645d3100a8a935a1e3837f885d && cd ..
 ```
 
 [same-as-file]: <> (utils/build-mlir.sh)
--- a/docs/BuildOnLinuxOSX.md
+++ b/docs/BuildOnLinuxOSX.md
@@ -22,6 +22,22 @@ cd llvm-project && git checkout 0913547d0e3939cc420e88ecd037240f33736820 && cd .
 ``` bash
 mkdir llvm-project/build
 cd llvm-project/build
+
+cmake -G Ninja ../llvm \
+   -DLLVM_ENABLE_PROJECTS="mlir;clang;openmp" \
+   -DLLVM_TARGETS_TO_BUILD="host" \
+   -DCMAKE_BUILD_TYPE=Release \
+   -DLLVM_ENABLE_ASSERTIONS=ON \
+   -DLLVM_ENABLE_RTTI=ON \
+   -DENABLE_LIBOMPTARGET=OFF \
+   -DLLVM_ENABLE_LIBEDIT=OFF
+
+cmake --build . -- ${MAKEFLAGS}
+cmake --build . --target check-mlir
+```
+
+To enable parallelization for onnx-mlir, llvm-project should be configured as
+```
 cmake -G Ninja ../llvm \
    -DLLVM_ENABLE_PROJECTS=mlir \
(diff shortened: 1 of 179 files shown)


AlexandreEichenberger:
@chentong319 there is currently an error, working to fix it. It will only be a small change.

AlexandreEichenberger:
@chentong319 ran independent tests the fixes works. The latest commit should have a green build.

AlexandreEichenberger:
Summary of changes:

Elementwise:

RoundOp was expanded manually in elementwise, as a full loop over all operations. But I needed it as an operation performing on a scalar or a simd vector. So I pulled the implementation into MathBuilder, so that I can call it anywhere where I need to compute the Round (which is an elaborate operation rounding to even whole numbers).

Enabled simd for dequantize. The issue that prevented it was the lack of SIMD support for the MathBuilder.cast. Had to add this for quantize operations (which are now vectorized) so it now works here too.

Delayed splatting in `getPartiallyFlattenedSimdCode`. Since MathBuilder does the splatting when operations have a mixture of scalar and SIMD, no need to do it here anymore.

Reduction:
 
Migrated some list support in a separate file.

Created a new operation `emitFullSIMDReductionFor` that does a reduction to a single scalar (previous support only reduced to an array of reduction, not a single scalar). While at it, I also enable the fused reduction of 2 distinct reductions, as DynamicQuantizedLinear needs both the min and the max at the same time.

Changed the interface to know when ops need a division using a templated approach.


ONNXToKrnlCommon

For elementary, simple operations (such as Add/Sub...) don't have a custom `emitScalarOpFDor` template that use the MathBuilder, and thus they don't have the scalar/vector expansion scheme. Added it there directly.

[Dynamic] Quantize Linear

Added 2 functions to perform the Dynamic part (compute min/max to get the scale/zero point] and perform the conversion. Simply moved the methods to a new independently callable operations (as they will also be needed elsewhere in the future). Removed the `onnx.xxx` and replaced them by `math.xxx` builder as we now generate fused loops.

Krnl DialectBuilder

Generate a SIMD loop for the given kernel. See the .hpp for explanation of the scheme.

MLIR DialectBuilder

Added handling of scalar/vector for `math.select`, and `cast` (that one is tricky, I left explanation in the code, code are mostly making sure to systematically use the proper type, elementType or original possibly vector type).

I added a new `computeSuitableUnrollFactor` to guide simd. It basically look at if simd is possible from the data type, then look at the average usage of SIMD operations, decide of an additional unroll factor given the register pressure (low pressure, more unrolling; high pressure because lots of operation, less unrolling).

Code was added for `math.round` and `math.clip`

chentong319:
1. Do the terms, vector and simd, have different meaning in the code? 
2. What's the constraint for the vector length (VL)? I am not clear about the relationship of vector dialect and the final simd code for particular machine.
3. For reduction, I wonder whether the loop fusion in later pass can save us the trouble of handling multiple reductions.
4. For divide by mean, we could represent that semantics with tensor dialect and make our code cleaner. But not easy in our onnx to krnl framework. 
5. Overall, we are trying to generate the best code in performance for common patterns with complicated code.

AlexandreEichenberger:
> Do the terms, vector and simd, have different meaning in the code?

I use them interchangeably. If you feel strongly about the one or the other, I can do a cleanup in a subsequent PR. Technically vectorization does not require the use of SIMD. For example ESSL has a vector mode where instead of calling one "math" function at a time, it calls a long vector of them of arbitrary length, and use a mixture of SIMD and scalar operations to execute them as fast as possible. SIMD implies the use of SIMD instructions.

chentong319 (approved):
LGTM!

AlexandreEichenberger:
Thanks, will implement your suggestions in the next PR, namely:
* distinguish more precisely between Vector Length (VL) for hardware reason vs the additional unrolling for performance
* give an example of how to use the new simd krnl interface.

PR #2906, opened 2024-08-14
Cleanup of SIMD code (no functional changes)

AlexandreEichenberger:
In this PR, I bunched up several cleanups for better reading of SIMD:

1.  Vector Length (VL) has two components: one required by the hardware (e.g. 4 float32 in a 128 bit simd register) and a performance component (multiple of the hardware constraints, for better Instruction Level Parallelism or ILP). The hardware component is now systematically called `archVL`, the performance component is called `unrollVL` and the product of the two, which is the effective VL, is `totVL`.
2. We used to characterize the number of ops in a kernel using 2 lists of Generic ops and occurrence of that generic ops. For kernels using lots of ops, having 2 lists is hard to read. We have now one list of pairs, e.g. `{{GenericOps::DivGop, 1},{GenericOps::ArithmeticGop, 5},[...] {GenericOps::FloorGop, 2}},`
3. We used to keep pointers to the VectorMachineSupport class object that is associated with the current target machine. Now all is hidden inside the class, and all public interface are using `static` methods.
4. We used to distinguish between `VL=0` for SIMD not supported vs `VL=1` for scalar code. There was no functional advantage in distinguishing these 2 cases. Now no simd is always reported as a `VL=1`.

### DIFF
--- a/docs/BuildOnLinuxOSX.md
+++ b/docs/BuildOnLinuxOSX.md
@@ -15,7 +15,7 @@ Firstly, install MLIR (as a part of LLVM-Project):
 ``` bash
 git clone -n https://github.com/llvm/llvm-project.git
 # Check out a specific branch that is known to work with ONNX-MLIR.
-cd llvm-project && git checkout 0913547d0e3939cc420e88ecd037240f33736820 && cd ..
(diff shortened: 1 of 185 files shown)


chentong319 (approved):
LGTM!

AlexandreEichenberger:
@chentong319 thanks for the quick turnaround.

PR #2907, opened 2024-08-15
Update BuildONNX.md

hamptonm1:


hamptonm1:

hamptonm1:
@AlexandreEichenberger Is something wrong with our builds? This check for `Linux amd64` is failing but I did not change any code. This should not fail in terms of a backend test. Hmm

AlexandreEichenberger:
@gongsu832 there is CI that fails on a PR that has no functional changes. Do you know if maybe the x86 test machine has issues?

gongsu832:
> @gongsu832 there is CI that fails on a PR that has no functional changes. Do you know if maybe the x86 test machine has issues?

Unless the failure is repeatable, it typically indicates some sort of transient problem like network hiccup, running out of memory, running out of /tmp, etc. The solution is to simply run the build again.

hamptonm1:
> > @gongsu832 there is CI that fails on a PR that has no functional changes. Do you know if maybe the x86 test machine has issues?
> 
> Unless the failure is repeatable, it typically indicates some sort of transient problem like network hiccup, running out of memory, running out of /tmp, etc. The solution is to simply run the build again.

@gongsu832 Okay, I did that about 3 times and still saw the same exact failure (even in other PRs). So this behavior is odd to me.

docs/BuildONNX.md
```diff
@@ -9,29 +9,39 @@ You will need to install python 3.x if its not default in your environment, and
...
 1.	Create your own branch
 
-2.	cd into `third_party/onnx` and checkout the commit for the latest version of onnx (You can find the latest commit here: https://github.com/onnx/onnx/releases)
+2.	"cd" into `third_party/onnx` and checkout the commit for the latest version of onnx (You can find the latest commit here: https://github.com/onnx/onnx/releases)
 
-3.	pip uninstall onnx (remove older version)
+3.	"pip uninstall onnx" (remove older version)
 
-4.	In `onnx-mlir/` directory, pip install —user third_party/onnx (install onnx from the commit and not online version)
+4.	In `onnx-mlir/` directory, "pip install -e third_party/onnx" (install onnx from the commit and not online version)
```
chentong319, line 21: Is option "-e" really needed?

hamptonm1: Yes! Based on the comment @justinchuby made here: https://github.com/onnx/onnx/issues/5716

`Please make sure the setuptools in the environment is up to date. You should also use pip install -e . for editable mode installs, and use python -m build for building wheels. @liqunfu we should call this out in the release notes`

justinchuby: If you don’t intend to build in development mode then you may drop the -e. e is for editable.

chentong319: Thanks for clarification. We are not in development mode for onnx package.

chentong319 (approved):
LGTM!

PR #2908, opened 2024-08-16
Simd optimizations: partial SIMD loops for element-wise ops

AlexandreEichenberger:
Many optimizations where performed, here are the highlight
* more aggressive flattening of memory accesses and loops in presence of broadcast. This result in fewer loops and more parallel opportunities.
* enable partial SIMD for element wise ops. For example if a loop iterates over 65 values, the new scheme generates SIMD loops (maybe with totVL = 32 to get lots of ILP) and one additional scalar loop for the 65th value.
* use an improved method to decide which unroll vector length (unrollVL), namely a logical unrolling of the architectural vector length (archVL, say 4 for 128 bit vector registers for a 32 bit float). 

Other changes to support above optimizations
* changed the templates for the versioned ops to reuse as much as possible the code from the non-versioned op, to reduce redundancy.
* has template to return the operation mix (a set of pairs of generic operations and how many times these operations were found in a kernel) to be more flexible.
* added support when generating SIMD code using the KRNL interface to support the cases found in ElementWise, namely mixture of ranks, presence of `ONNXNoneOp`...

Finally, I found it convenient when running our backend test to know the file that is being compiled. So I added the file name to the log message (see `[1/5]`)

```
[1/5] Fri Aug 16 18:17:04 2024 Importing ONNX Model to MLIR Module from "squeezenet_old.onnx"
[2/5] Fri Aug 16 18:17:04 2024 Compiling and Optimizing MLIR Module
[3/5] Fri Aug 16 18:17:05 2024 Translating MLIR Module to LLVM and Generating LLVM Optimized Bitcode
[4/5] Fri Aug 16 18:17:06 2024 Generating Object from LLVM Bitcode
[5/5] Fri Aug 16 18:17:07 2024 Linking and Generating the Output Shared Library
```

Added `check-onnx-backend-dynamic` to Mac CI.


AlexandreEichenberger (commented):
comments to help the review

src/Conversion/ONNXToKrnl/Quantization/QuantizeLinear.cpp
```diff
@@ -27,65 +27,61 @@ void emitQuantizationLinearScalarParameters(ConversionPatternRewriter &rewriter,
...
 
   // Types
   Type quantizedElementType = quantizedType.getElementType();
   int64_t rank = inputType.getRank();
 
+  // Flatten the input data and outputs
+  DimsExpr inputDims, flatInputDims, flatAllocDims;
+  inputDims = allocDims; // Unput and output have the same shape.
+  Value flatInput =
+      create.mem.reshapeToFlatInnermost(input, inputDims, flatInputDims, rank);
```
chentong319, line 41: Just comment not for change: the flatInputDims and flatAllocDims are pure output. To me, the code will cleaner (but might be less efficient) if the flatInputDims is retrieved from the return of flatInput.

src/Dialect/Mlir/VectorMachineSupport.hpp
```diff
@@ -63,7 +64,11 @@ enum class GenericOps {
   TrigHyperbolicGop, /* Hyperbolic trig. */
 };
 
+// Describe the mix of Generic operations in a given kernel. Each generic
+// operation is associated with a number, which indicates the number of
+// occurrence of that generic op in the given kernel.
 using GenOpsMix = mlir::ArrayRef<std::pair<GenericOps, int64_t>>;
+using GenOpsMixList = llvm::SmallVector<std::pair<GenericOps, int64_t>, 4>;
```
chentong319, line 71: Can std::map be used for GenOpsMix?

src/Dialect/Krnl/DialectBuilder.cpp
```diff
@@ -273,12 +274,96 @@ KrnlIterateOp KrnlBuilder::iterateIE(ValueRange originalLoops,
       });
 }
 
-// TODO: once the interface is updated, provide a good example of how to use.
+/*
+Example of how to use the interface:
+Say you have a loop of i=0..128, j=0..0 and want to exploit r[i,j] = a[i,j] +
```
chentong319, line 279: j=0..0 ?

src/Conversion/ONNXToKrnl/Quantization/QuantizeLinear.cpp
AlexandreEichenberger: Good point, let me look into it.

src/Dialect/Mlir/VectorMachineSupport.hpp
AlexandreEichenberger: Another good point, as long as I can iterate through all the keys, it would work too and be more elegant.

src/Dialect/Krnl/DialectBuilder.cpp
AlexandreEichenberger: good catch, will change

src/Dialect/Mlir/DialectBuilder.cpp
```diff
@@ -2002,26 +2002,29 @@ void VectorBuilder::multiReduction(SmallVectorImpl<Value> &inputVecArray,
   }
 }
 
+// New style.
 /*static*/ int64_t VectorBuilder::computeSuitableUnrollFactor(
     MemRefType memRefType, int64_t collapsedInnermostLoops, GenOpsMix genOps,
-    int64_t &simdLoopStaticTripCount) {
-
-  simdLoopStaticTripCount = 0; // Initially assume no SIMD.
+    bool canOverCompute, int64_t &simdLoopStaticTripCount, bool &simdOnly) {
```
chentong319, line 2008: It is a little weird to put this function in DialectBuilder. How about a simd-related utils in onnx to krnl?

AlexandreEichenberger: Thanks for the nudge, I totally agree. Moved the functions to the `ONNXToKrnlCommon` files

src/Dialect/Krnl/DialectBuilder.cpp
AlexandreEichenberger: good catch, fixed.

src/Dialect/Mlir/VectorMachineSupport.hpp
AlexandreEichenberger: Good suggestion, needed a bit of code change but I like it. Done.

src/Conversion/ONNXToKrnl/Quantization/QuantizeLinear.cpp
AlexandreEichenberger: I see your point, when the dims are not compile time, it will generate a different code. Because code is generated to compute the dimensions in reshapeFlatInnermost, and then code will be generated again to retrieve the dimensions from the new `flatValue`.  Thus more code will be generated and I will also have to update the lit tests. So unless you feel strongly about it, I would prefer not to change it.

AlexandreEichenberger:
@chentong319 Except for one, I implemented all your suggestions. Left an explanation why on the one I did not implement. Thanks for the thoughtful comments.

AlexandreEichenberger:

chentong319 (approved):
LGTM!

PR #2910, opened 2024-08-19
Add asserts for snprintf, sprintf, and vsnprintf

mikeessen:
Update code to add assert statements for  snprintf, sprintf, and vsnprintf

### DIFF
--- a/src/Runtime/OMIndexLookup.inc
+++ b/src/Runtime/OMIndexLookup.inc
@@ -34,7 +34,8 @@ static inline uint32_t hash_string(uint32_t hval, const char *str) {
 // Adaptation of 32-bit FNV for int64_t values.
 static inline uint32_t hash_int64(uint32_t hval, int64_t val) {
   char str[20];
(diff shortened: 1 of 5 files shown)


mikeessen:
> You don't want to put a required statement inside `assert` because `assert` is a macro and the statement disappears in Release mode when `NDEBUG` is defined.

@gongsu832 Thanks for catching!

AlexandreEichenberger:
@gongsu832 @mikeessen 

That leads to another question: are there assert that we want under the Release mode? E.g. if a `malloc` returns no data, don't we want to have the compiler stops? Same for here, if there was a malicious strings, would we want to know about it and abort?

AlexandreEichenberger:
I did not find "assert that we would like to keep in the release mode". However some advocate to build release with asserts using ` -DLLVM_ENABLE_ASSERTIONS=true`.

AlexandreEichenberger (commented):
LGTM

AlexandreEichenberger (approved):

PR #2911, opened 2024-08-19
Timeout a build after 6 hours

gongsu832:
Current Jenkins build typically takes less than 3 hours. Timeout a build after 6 hours to account for the case when the build died badly (typically due to segfault) and didn't get a chance to report failure.


AlexandreEichenberger (approved):
LGTM, 
Do you know if the timeout issues a special message so that we know the CI was aborted due to timeout?

gongsu832:
Not sure since we never used this before but I'd be surprised that there is no indication on whether a build is aborted due to timeout or not.

PR #2912, opened 2024-08-20
Added Release with enable assertion to default build of ONNX MLIR

AlexandreEichenberger:
When we build LLVM in release mode, assertions are enabled by default. 
We should probably do the same for ONNX-MLIR.


gongsu832 (approved):
LGTM

PR #2915, opened 2024-08-23
Upgrade to ONNX 1.16.2 (opset 21)

gongsu832:
- upgrade to ONNX 1.16.2 (opset 21)
- comment out two tests (test_group_normalization_epsilon_cpu and test_group_normalization_example_cpu) until groupnorm implementation is updated (broken due to scale and bias shape change)
- update numpy version in the user image to match that in the dev image
- add missing dependency for some targets


test/backend/inference_backend.py
```diff
@@ -1257,21 +1258,21 @@ def get_test_models():
...
-        "test_group_normalization_epsilon_cpu": {
-            STATIC_SHAPE: {},
-            DYNAMIC_SHAPE: {-1: {-1}},
-            CONSTANT_INPUT: {-1},
-        },
+        # "test_group_normalization_epsilon_cpu": {
+        #    STATIC_SHAPE: {},
+        #    DYNAMIC_SHAPE: {-1: {-1}},
+        #    CONSTANT_INPUT: {-1},
+        # },
```
AlexandreEichenberger, line 1265: @hamptonm1 Please open an issue (if not already done) and/or reuse the code that you may have already developed for this.

test/backend/inference_backend.py
```diff
@@ -1257,21 +1258,21 @@ def get_test_models():
...
             STATIC_SHAPE: {},
             DYNAMIC_SHAPE: {-1: {-1}},
             CONSTANT_INPUT: {-1},
         },
-        "test_group_normalization_example_cpu": {
-            STATIC_SHAPE: {},
-            DYNAMIC_SHAPE: {-1: {-1}},
-            CONSTANT_INPUT: {-1},
-        },
+        # "test_group_normalization_example_cpu": {
```
AlexandreEichenberger, line 1271: @hamptonm1 Please open an issue (if not already done) and/or reuse the code that you may have already developed for this.

AlexandreEichenberger (approved):
LGTM, big thanks for this PR, much much appreciated

test/backend/inference_backend.py
hamptonm1: I already have a PR for this in my personal fork.

AlexandreEichenberger: Great @hamptonm1 

I also learned that the GroupNorm 18 is actually a mistake, no-one use it that way. PyTorch does not lower its group norm to ONNXs because ONNX is not conforming to the standard.

So if it's not too much work to add the new one and keep the old one, that is ok. Otherwise, it might be ok to declare the old one unsupported and only support the new one.

By the way, the current description (21) is a bit misleading. The scale and bias is "PER Channel", as described in the input (one value per C). The text above seems to imply it is per group, which is not the case. There willl be a PR soon to fix this in the specs.

hamptonm1: @AlexandreEichenberger Yeah I spoke with Justin about this already and verified GroupNorm 18 was a mistake, that is why there is not an old version of onnx backend tests to support version 18. There is no plans to continue to support the inaccurate representation.

AlexandreEichenberger: @hamptonm1 that makes sense. If we could, we should generate an error in the frontend if we see a GroupNorm 18, and have our only group norm be the 21. Probably @chentong319 is the right person to get help on that if you need help.

hamptonm1: @AlexandreEichenberger Okay thanks.... @tungld mentioned having a warning. So I am fine with generating an error or warning if someone uses GroupNorm 18.

PR #2916, opened 2024-08-24
Update llvm-project to llvm/llvm-project@f142f8afe21b

brnorris03:
Upstream update
  LLVM : f142f8afe21bceb00fb495468aa0b5043e98c419
  StableHLO: 54aa1a57178251981da616b877dda1a88d840d11

### DIFF
--- a/docs/BuildOnLinuxOSX.md
+++ b/docs/BuildOnLinuxOSX.md
@@ -15,7 +15,7 @@ Firstly, install MLIR (as a part of LLVM-Project):
 ``` bash
 git clone -n https://github.com/llvm/llvm-project.git
 # Check out a specific branch that is known to work with ONNX-MLIR.
-cd llvm-project && git checkout 60a7d33106d3cd645d3100a8a935a1e3837f885d && cd ..
+cd llvm-project && git checkout f9031f00f2c90bc0af274b45ec3e169b5250a688 && cd ..
 ```
 
 [same-as-file]: <> (utils/build-mlir.sh)
--- a/docs/BuildOnLinuxOSX.md
+++ b/docs/BuildOnLinuxOSX.md
@@ -15,7 +15,7 @@ Firstly, install MLIR (as a part of LLVM-Project):
 ``` bash
 git clone -n https://github.com/llvm/llvm-project.git
 # Check out a specific branch that is known to work with ONNX-MLIR.
-cd llvm-project && git checkout f9031f00f2c90bc0af274b45ec3e169b5250a688 && cd ..
+cd llvm-project && git checkout f142f8afe21bceb00fb495468aa0b5043e98c419 && cd ..
 ```
 
 [same-as-file]: <> (utils/build-mlir.sh)
--- a/docs/BuildOnWindows.md
+++ b/docs/BuildOnWindows.md
@@ -52,7 +52,7 @@ Install MLIR (as a part of LLVM-Project):
 ```shell
 git clone -n https://github.com/llvm/llvm-project.git
 # Check out a specific branch that is known to work with ONNX-MLIR.
-cd llvm-project && git checkout 60a7d33106d3cd645d3100a8a935a1e3837f885d && cd ..
+cd llvm-project && git checkout f142f8afe21bceb00fb495468aa0b5043e98c419 && cd ..
 ```
 
 [same-as-file]: <> (utils/build-mlir.cmd)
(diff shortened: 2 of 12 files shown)


brnorris03:
Not sure how to address the lit test failures, help would be appreciated

AlexandreEichenberger:
Will try to have a look.

chentong319:
The lit test on different platforms all failed in these three test. 
```
Failed Tests (3):
  Open Neural Network Frontend :: conversion/onnx_to_krnl/ControlFlow/If_with_canonicalize.mlir
  Open Neural Network Frontend :: conversion/onnx_to_krnl/Tensor/onnx_lowering_depth_to_space_op.mlir
  Open Neural Network Frontend :: conversion/onnx_to_krnl/Tensor/onnx_lowering_space_to_depth.mlir
```

Detailed info indicates that the errors are related to `builtin.unrealized_conversion_cast`. I'll investigate the issue.

brnorris03:
> The lit test on different platforms all failed in these three tests.
> 
> ```
> Failed Tests (3):
>   Open Neural Network Frontend :: conversion/onnx_to_krnl/ControlFlow/If_with_canonicalize.mlir
>   Open Neural Network Frontend :: conversion/onnx_to_krnl/Tensor/onnx_lowering_depth_to_space_op.mlir
>   Open Neural Network Frontend :: conversion/onnx_to_krnl/Tensor/onnx_lowering_space_to_depth.mlir
> ```
> 
> Detailed info indicates that the errors are related to `builtin.unrealized_conversion_cast`. I'll investigate the issue.

Yes, these occur for nested `onnx.If`, for example (but not when there is a single `onnx.If`)

chentong319:
I investigated onnx_lowering_depth_to_space_op.mlir. The source of different is that the simplification of unrealized_cast in MLIR is changed. We generated three unrealized_cast to cast between memref and tensor. Two of them are cancelled. But the new version of MLIR left an unused unrealized_cast.
Similar situation for onnx_lowering_space_to_depth.mlir.
A workaround is to create a new check code for these lit tests. The long term solution for onnx-mlir is to use bufferization toTensor and toMemref method, instead of using unrealized_cast directly.
I'll further check the IF test case. It is aslo related to unrealized_cast, but crashed the compiler.

brnorris03:
>I investigated onnx_lowering_depth_to_space_op.mlir and onnx_lowering_space_to_depth.mlir. They failed in result verification. In the lowering step. the new version of MLIR generated an extra redundant unrealized_cast. After canonicalization, the position of remaining unrealized_cast op changed. It's a trivial change.
A workaround is to create a new check code for these lit test cases. 

Thank you very much! I will change the `CHECK` for these two tests. I hadn't looked at them yet since I was stuck on the `If` test, which I think is a different issue.

chentong319:
I tried to get more info for the failure of If_with_canonicalize.mlir, but the print just crashed.
``` 
Debug/bin/onnx-mlir-opt --mlir-print-ir-after-failure If_with_canonicalize.mlir --shape-inference --convert-onnx-to-krnl
If_with_canonicalize.mlir:17:10: error: null operand found
    %4 = "onnx.If"(%3) ({
         ^
If_with_canonicalize.mlir:17:10: note: see current operation: %20 = "builtin.unrealized_conversion_cast"(<<NULL VALUE>>) : (<<NULL TYPE>>) -> memref<i32>
// -----// IR Dump After onnx_mlir::FrontendToKrnlLoweringPass Failed (convert-onnx-to-krnl) //----- //
I am trying to figure out how the NULL VALUE is created.
Without the IR dumped, I can not submit the error to llvm.

chentong319:
I checked the test case of If. I think that the source of error came from two factors: onnx.yield is converted inside onnx.If without calling the typeconvertor systematically; and the nested If imposed an uncommon order of rewriting.

For this PR, you can remove that lit to make it pass.
I will keep working on this issue.

brnorris03:
> I checked the test case of If. I think that the source of error came from two factors: onnx.yield is converted inside onnx.If without calling the typeconvertor systematically; and the nested If imposed an uncommon order of rewriting.
> 
> For this PR, you can remove the test case to make it pass. I will fix this issue with another PR.

Thank you very much for looking into it. Will do this and add a TODO.

AlexandreEichenberger:
Thanks @brnorris03 and @chentong319 for your help on this, much appreciated.

PR #2919, opened 2024-08-27
try to use new buffer deallocation

chentong319:
onnx-mlir is using the old buffer deallocation by default (option `--use-old-bufferization`). If that option is set to false, there is some compilation error. This PR is trying to fix this issue.
1. register the bufferizaton interface for the dialects that may be used. For NNPA, dialects, `memref` and `arith` may be used.
2. After that change, there is an error for memref.prefetch. Its memory side effect is not defined. Need further investigation with MLIR community. Current workaround is to add a flag, `disable-memref-prefetch`, in builder to generate nothing for prefetch.

With these changes, roberta model can be compiled with the new deallocation with command:
```
Debug/bin/onnx-mlir --O3 --EmitLib --mtriple=s390x-ibm-loz --mcpu=z16 --maccel=NNPA --disable-memref-prefetch=true   --use-old-bufferization=false roberta-base-11.onnx
```

Here is the execution time of for this model:
without deallocation: 10.6
old deallocation: 4.6
new deallocation: 4.8

### DIFF
--- a/src/Compiler/CompilerDialects.cpp
+++ b/src/Compiler/CompilerDialects.cpp
@@ -46,8 +46,8 @@ DialectRegistry registerDialects(ArrayRef<accel::Accelerator::Kind> accels) {
   for (auto *accel : accel::Accelerator::getAccelerators())
     accel->registerDialects(registry);
 
-  if (useOldBufferization)
-    memref::registerAllocationOpInterfaceExternalModels(registry);
+  memref::registerAllocationOpInterfaceExternalModels(registry);
+  arith::registerBufferDeallocationOpInterfaceExternalModels(registry);
 
   return registry;
 }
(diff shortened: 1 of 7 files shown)


src/Compiler/CompilerDialects.cpp
```diff
@@ -46,8 +46,8 @@ DialectRegistry registerDialects(ArrayRef<accel::Accelerator::Kind> accels) {
   for (auto *accel : accel::Accelerator::getAccelerators())
     accel->registerDialects(registry);
 
-  if (useOldBufferization)
-    memref::registerAllocationOpInterfaceExternalModels(registry);
+  memref::registerAllocationOpInterfaceExternalModels(registry);
```
AlexandreEichenberger, line 49: Can you comment on why we don't need to execute this conditionally anymore?

chentong319: fixed.

AlexandreEichenberger: Did you upload the latest version? It does not looked conditionally executed.

chentong319: These registrations are needed by both new and old deallocation. Thus, they are NOT conditionally executed.

AlexandreEichenberger: Got it, thanks for the explanation

AlexandreEichenberger (approved):
LGTM, looking forward to see if the new scheme can eventually deliver better performance.

PR #2922, opened 2024-08-29
Add check for malloc to prevent NULL pointer dereference in docs example.

Sunny-Anand:
This PR updates the code to check the result of memory allocation for input data tensors in doc example.

### DIFF
--- a/docs/doc_example/main.c
+++ b/docs/doc_example/main.c
@@ -12,11 +12,11 @@ OMTensorList *create_input_list() {
 
   // Construct float arrays filled with 1s or 2s.
   float *x1Data = (float *)malloc(sizeof(float) * num_elements);
-  assert(x1Data && "Failed to allocate memory for input tensor.");
+  assert(x1Data && "Failed to allocate memory for input tensor x1Data.");
   for (int i = 0; i < num_elements; i++)
     x1Data[i] = 1.0;
   float *x2Data = (float *)malloc(sizeof(float) * num_elements);
-  assert(x2Data && "Failed to allocate memory for input tensor.");
+  assert(x2Data && "Failed to allocate memory for input tensor x2Data.");
   for (int i = 0; i < num_elements; i++)
     x2Data[i] = 2.0;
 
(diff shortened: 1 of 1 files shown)


gongsu832:
`assert` is for catching bugs, something logically impossible to happen with your code's internal state. For example, you have a loop variable to control the termination of a loop. In the loop body you may change the variable according to various conditions. But by your design, you know that the variable should never exceed some value **no matter how other parts of the code work**. So you can use `assert(v <= maxval)` to catch potential bugs in your own code that changes the variable.

`malloc` returning null is not a bug. It's a perfectly normal condition that can happen when the system is running out of memory. In addition, `assert` goes away in release mode when NDEBUG is defined so your code will not check the returned pointer (when it still should).

AlexandreEichenberger:

AlexandreEichenberger (approved):
LGTM, thanks

AlexandreEichenberger:
@gongsu832 We recently enabled asserts in the release mode too, just like the LLVM release mode which keeps the asserts.

However, I do see your point. In this specific example, I agree that it would be better not to assert. Here is a better pattern:

```C++
OMTensorList *create_input_list() {
   var1 = malloc(...);
   if (!var1) return null;
  var2 = malloc(...);
   if (! var2) {
     free(var1)
     return null;
  }

int main() {

  OMTensorList *input_list = create_input_list(....)
  if (!input_list) 
    return "error code appropriate for running out of memory"

  ...
  return 0;
}
```   
 
I know it is only an example, but the pattern above is the better pattern to "instruct" users how to deal with our interface.

gongsu832:
Keeping asserts for release mode and using asserts correctly for bugs are two separate issues. Bugs are coding errors in the internal logic of your code regardless of external references. malloc returning null or user inputting something you didn't expect, etc., which I see a lot of people like to use assert on, are NOT bugs.

Sunny-Anand:

Sunny-Anand:
@gongsu832 I have removed the assert for the said malloc. @AlexandreEichenberger followed your suggestion for the recommended pattern. I still can't test the PR's and also don't have write access to merge in the PR's.

Sunny-Anand:

Sunny-Anand:

mikeessen (approved):
LGTM!

PR #2923, opened 2024-08-29
Parallel code for quantization

AlexandreEichenberger:
Added a parallel scheme for reduction (min & max) of a tensor to a single scalar.

the scheme is as follows
``` 
for(t=0..TNUM) { // parallel for TNUM threads
  reduce min max for chance input[lb...ub] into min & max[t] where lb=t * chunk, ub=min(UB, (t+1)* chunk
}
reduce min & max[0...TNUM] into min_scalar and max_scalar
```

Now input is flattened to one 1D. If the `dim[0]` of flattened input is runtime, then `chunk = dim[0]/TNUM` is also runtime.
Then in the `reduce min max` inside the parallel region, we compute the `lb = t * chunk` where both are runtime. This is not affine (affine is only affine under the multiplication of a dim with a literal constant).

Thus we need a `reduce SIMD` that works with SCF as well as the normal Krnl. 

For this, I needed to further simplify the Dialect builder so that they are more uniform, added a `MemRefBuilder` load and store (because we don't want krnl constructs inside the Dialect/MLIR sub directory).

I then added a `simdIterateIE` and `simdReduceIE` that works with an templated builder for the loops (Krnl or SCF) and scalar mem operations (Krnl or MemRef).

All of the simd code gen in elementwise and reduce uses this new scheme, except for one in Reduce.cpp which needs the unrolling of VL outer loop iteration so as to use the VL to VL reduction scheme necessary on machines (such as Z for floating point) that have no horizontal SIMD reduction operation.

Note that the new SIMD support works with arbitrary array length, by generating SIMD code for most loop iterations, followed by a small loop for the `trip-count % VL` leftover iterations.


AlexandreEichenberger:
Speedup achieved
old vs new-simd (sequential) ~ 1.4-1.7x depending on shape size & static/dynamic shape
old vs new-simd-paralllel ~ 3.2-4x depending on shape size  & static/dynamic shape

image

src/Conversion/ONNXToKrnl/Math/Reduction.cpp
```diff
@@ -289,11 +289,94 @@ using MDBuilder =
 // one is needed, then pass ONNXNoneOp in the second slot.
 // Return true if we can optimize the reduction, false otherwise.
 
-// TODO: alexe add support for parallel
-// TODO: alexe see if the new simd infrastructure can be used.
+template <typename BUILDER, typename ONNXReductionOp1,
+    typename ONNXReductionOp2>
```
tungld, line 293: It looks to me that we can generalize to multiple reduction ops. Same to `emitFullSIMDReductionFor`. But, it's ok for another PR.

src/Dialect/Krnl/DialectBuilder.hpp
```diff
@@ -98,40 +98,76 @@ struct KrnlBuilder : public DialectBuilder {
...
+     Two functions are given: reductionBuilderFn to perform the partial
+     reductions into the temporary values tmps, finishing with up to VL partial
+     reductions
+     The second function: postProcessingBuilderFn performs the reductions of the
+     up to VL partial reductions into a final scalar reduction to be stored into
+     the outputs (a scalar value). For some reductions, post processing is also
+     needed, for example, mean reduction divide the accumulated sum by the
+     number of elements. That step is also performed here.
+    */
+  void simdReduceIE(IndexExpr lb, IndexExpr ub, int64_t VL, bool fullySimd,
```
tungld, line 155: Thanks for adding this!

tungld (approved):
LGTM.

src/Conversion/ONNXToKrnl/Math/Reduction.cpp
```diff
@@ -326,92 +416,170 @@ bool emitFullSIMDReductionFor(ConversionPatternRewriter &rewriter, Location loc,
   int64_t totVL =
       computeSuitableUnrollFactor(inputType, collapsedInnermostLoops, mix,
           canOverCompute, simdLoopStaticTripCount, simdOnly);
-  // Current simdized loop only support SIMD only scheme.
-  if (!simdOnly) {
-    totVL = capVLForSimdOnly(inputType, totVL, simdLoopStaticTripCount);
+#if 1 // new approach that enables parallelism
```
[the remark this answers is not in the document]
AlexandreEichenberger: Sorry, I meant to eliminate the old approach. Will remove the old.

src/Conversion/ONNXToKrnl/Math/Reduction.cpp
AlexandreEichenberger: you are correct. Let's be on the lookout in models if we see opportunities.

PR #2926, opened 2024-09-02
Support i8 in lowering ONNXQuantizeLinear

tungld:
The lowering of ONNXQuantizeLinear itself supports i8, but there is an assert to check and discard the i8 case (Unfortunately, I don't know the reason for that assert).
This patch removes the assert to allow the i8 case and adds a lit test for that.


AlexandreEichenberger (approved):
LGTM

PR #2927, opened 2024-09-03
Fix importing Cast opset >=19

tungld:
ONNXCast opset >=19 introduces a new attribute `saturate` and onnx-mlir wrongly imported the attribute.

For example, onnx-mlir imported ONNXCast as follows:
```mlir
%755 = "onnx.Cast"(%746) {onnx_node_name = "/bert/Cast", saturate = f32, to = f32} : (tensor<?x?x?x?xi64>) -> tensor<?x?x?x?xf32>
```
where `saturate = f32` was wrong, it would be `saturate = 0 : si64` or `saturate = 1 : si64`.

This patch fixes the issue by fixing the importing code for Cast's attributes.


AlexandreEichenberger (approved):
LGTM, thanks

PR #2930, opened 2024-09-04
Fix architecture terminology

tehbone:
Changed the naming of P and Z which are IBM brands to the respective architectures to be consistent with listed x86 architecture example.

PR #2931, opened 2024-09-04
Fix architecture terminology

tehbone:
Changed the naming of P and Z which are IBM brands to the respective architectures to be consistent with listed x86 architecture example.

Re-implementation of #2930


tehbone:
cc: @hamptonm1 @AlexandreEichenberger

hamptonm1 (approved):
LGTM!

PR #2932, opened 2024-09-04
Change lowering of onnx.IF to Krnl

chentong319:
The previous implementation lowers onnx.Yield inside the code of lowering onnx.If. The type conversions are not systematically added through the conversion infrastructure. It caused an error with the new version of llvm in PR #2916 . Even with older version of llvm, the lowering of If_with_canonicalize.mlir generated an unused unrealized_cast before canonicalization.

This PR adds a lowering of onnx.Yield with a separate rule. It works with the old and new version of llvm in my test.

Be aware that onnx.Yield is also used by onnx.Loop. The lowering of onnx.Loop still lowers the onnx.Yield with the Loop. Need lots of work to break up the transformation code for Loop lowering. Will do that only if running into an error.

### DIFF
--- a/src/Conversion/ONNXToKrnl/ControlFlow/Yield.cpp
+++ b/src/Conversion/ONNXToKrnl/ControlFlow/Yield.cpp
@@ -2,7 +2,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-//===--------------------- Yield.cpp - Lowering Yield Op --------------------===//
+//===--------------------- Yield.cpp - Lowering Yield Op
+//--------------------===//
 //
 // Copyright 2019-2023 The IBM Research Authors.
 //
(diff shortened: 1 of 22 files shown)


src/Conversion/ONNXToKrnl/ControlFlow/Yield.cpp
```diff
@@ -0,0 +1,58 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//===--------------------- Yield.cpp - Lowering Yield Op
+//--------------------===//
```
tungld, line 6: Please fix the format.

tungld (approved):
LGTM. Yes, it is the correct way lowering onnx.Yeild separately. Thanks!

src/Conversion/ONNXToKrnl/ControlFlow/Yield.cpp
chentong319: fixed

PR #2935, opened 2024-09-10
Add coding practices for onnx-mlir

mikeessen:
This PR adds a coding practices documentation file to describe the coding practices to use when updating or adding code to onnx-mlir.


AlexandreEichenberger (approved):
LGTM, I would add a small example, e.g. for the cast as most might not be aware with C++ vs C casting. (or say grep for `static_cast` for examples).

PR #2937, opened 2024-09-10
SIMD/Loop framework upgrade

AlexandreEichenberger:
Added support for handling SIMD for loops that have loop iterations that are not multiple of the Vector Length.

Because we can now generate SIMD code using either Krnl, Affine, or SCF, it was painful to have multiple ways to generate loops. I have now a unified interface that create loops across all 3 dialects:

```C++
  void forLoopIE(IndexExpr lb, IndexExpr ub, int64_t step, bool useParallel,
      mlir::function_ref<void(SCFBuilder &, mlir::ValueRange)> bodyFn) const;
```

which takes a lower/upper bound as IndexExpr, a boolean to define if the loop is sequential or parallel, and the function to be called.

An example is shown here

```C++
    // Invocation of the (possibly parallel) SIMD loop.
     if constexpr (std::is_same<BUILDER, KrnlBuilder>::value ||
                   std::is_same<BUILDER, AffineBuilder>::value ||
                   std::is_same<BUILDER, SCFBuilder>::value)
       builder.forLoopIE(lb, simdUb, VL, useParallel, simdLoopBody);
    else
      llvm_unreachable("BUILDER type not supported\n");
```

This complement the 3 SIMD calls: `simdIterateIE`, `simdReduceIE`, and `simdReduce2DIE`. The last 2 calls both perform reductions, but the first one uses horizontal/do-across reductions (e.g. available on z16 with integer add) and the second one use shuffle to mix VL consecutive reductions.
All simd calls now work with arbitrary numbers of loop iterations (whether a multiple of the hardware vector length or not).

To better provide the same functionality to both reduce simd calls, I expect now one lambda function per output (before one lambda function to generate all outputs).

We also had different calls for memory load/store. Now a common interface is used for Krnl, Affine, and MemRef, and nearly identical for Vector (where the load operation needs the type to determine the VL).

They all use the calls below
```C++
  mlir::Value load(mlir::Value memref, mlir::ValueRange indices = {},
      mlir::ValueRange offsets = {}) const;
  mlir::Value loadIE(mlir::Value memref, mlir::ArrayRef<IndexExpr> indices = {},
      mlir::ValueRange offsets = {}) const;
  void store(mlir::Value val, mlir::Value memref, mlir::ValueRange indices = {},
      mlir::ValueRange offsets = {}) const;
  void storeIE(mlir::Value val, mlir::Value memref,
      mlir::ArrayRef<IndexExpr> indices, mlir::ValueRange offsets = {}) const;
```


tungld (approved):
LGTM. Really appreciate your effort of simplifying these interfaces!

PR #2938, opened 2024-09-11
Quantization: option to disable the calculation and use of the zero point.

AlexandreEichenberger:
Use `--disable-quantization-zero-point` to disable the computation of the zero point in dynamic linear quantization and ignore the zero point in linear quantization and dequantization.

### DIFF
--- a/src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
+++ b/src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
@@ -1282,9 +1282,15 @@ Value emitScalarOpFor<ONNXDequantizeLinearOp>(
   Value scaleFloat = scalarOperands[1];
   Value zeroPointInt = scalarOperands[2];
 
-  Value zeroPointFloat = create.math.cast(elementType, zeroPointInt);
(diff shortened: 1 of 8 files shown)


src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -1282,9 +1282,15 @@ Value emitScalarOpFor<ONNXDequantizeLinearOp>(
   Value scaleFloat = scalarOperands[1];
   Value zeroPointInt = scalarOperands[2];
 
-  Value zeroPointFloat = create.math.cast(elementType, zeroPointInt);
   Value xFloat = create.math.cast(elementType, XInt);
-  Value sub = create.math.sub(xFloat, zeroPointFloat);
+
+  Value sub;
+  if (!disableQuantZeroPoint && !isNoneValue(zeroPointInt)) {
```
AlexandreEichenberger, line 1288: Note: zero point is optional in dequantization, so added support for its value being none.

chentong319 (approved):
LGTM!

tungld (approved):
LGTM. Thanks!

PR #2939, opened 2024-09-11
Reuse input buffer in lowering to krnl

chentong319:
Previous,  a new memref is always created when onnx is lowered to krnl.
However, an input buffer could be reused as output, if  the input has only only one use and the shape is the same as the output. The PR tried this idea on element-wise operations.
by default, the buffer reuse is turned off.
This feature is tested with Roberta-base-11.onnx. About 80% of malloc (in either numbers or memory size) can be avoided with buffer reuse. The buffer reuse code can generate correct result, but did not bring notable performance improvement, with either SIMD on or off.

### DIFF
--- a/src/Compiler/CompilerOptions.cpp
+++ b/src/Compiler/CompilerOptions.cpp
@@ -216,7 +216,7 @@ static llvm::cl::opt<bool, true> disableKrnlOpFusionOpt(
 static llvm::cl::opt<bool, true> disableKrnlBufferReuseOpt(
     "disable-krnl-buffer-reuse",
     llvm::cl::desc("disable buffer reuse within an op in onnx-to-krnl pass"
-                  "(default=true)\n"
+                   "(default=true)\n"
                    "Set to 'false' if you want to enable buffer reuse."
                    "Default value will be false when the functionality becomes"
                    "stable."),
--- a/src/Compiler/CompilerOptions.cpp
+++ b/src/Compiler/CompilerOptions.cpp
@@ -42,6 +42,7 @@ bool enableONNXHybridPass;                             // common for both
 std::vector<std::string> functionsToDecompose;         // common for both
 std::string opsForCall;                                // common for both
 bool disableKrnlOpFusion;                              // common for both
+bool disableKrnlBufferReuse;                           // common for both
 EmissionTargetType emissionTarget;                     // onnx-mlir only
 bool invokeOnnxVersionConverter;                       // onnx-mlir only
 bool preserveLocations;                                // onnx-mlir only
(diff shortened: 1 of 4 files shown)


src/Compiler/CompilerOptions.cpp
```diff
@@ -212,6 +213,16 @@ static llvm::cl::opt<bool, true> disableKrnlOpFusionOpt(
     llvm::cl::location(disableKrnlOpFusion), llvm::cl::init(false),
     llvm::cl::cat(OnnxMlirCommonOptions));
 
+static llvm::cl::opt<bool, true> disableKrnlBufferReuseOpt(
```
AlexandreEichenberger, line 216: NIT: generally "disable" is for a function that is default on, "enable" is for one that is default off. I think you want "enable" here.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -28,6 +28,86 @@ using namespace mlir;
...
+// Default VL=0 is used for non SIMD allocation
+Value allocOrReuse(MemRefBuilder &create, Operation *op,
+    ValueRange generatedOperands, MemRefType outputMemRefType, DimsExprRef dims,
+    int64_t alignment, int64_t VL = 0);
+
+Value allocOrReuse(MemRefBuilder &create, Operation *op,
+    ValueRange generatedOperands, MemRefType outputMemRefType, DimsExprRef dims,
+    int64_t alignment, int64_t VL) {
+
+  // By default, disableKrnlBufferReuse is true. Simply allocate a memref.
```
AlexandreEichenberger, line 82: This code could be simplified as follows
```C++
if (!disableKrnlBufferReuse) {
  int indexToReuse = xxx
  if (indexToReuse != -1) return xxx
}
// no reuse, alloc
```

test/mlir/conversion/onnx_to_krnl/onnx_lowering_reuse.mlir
```diff
@@ -0,0 +1,11 @@
+// RUN: onnx-mlir-opt --disable-krnl-op-fusion=true --disable-krnl-buffer-reuse=false --shape-inference --convert-onnx-to-krnl --canonicalize %s -split-input-file | FileCheck %s
+
+// -----
+func.func @test_reuse(%arg0: tensor<1024xf32>, %arg1: tensor<1024xf32>) -> tensor<1024xf32> {
```
AlexandreEichenberger, line 4: Nice concise test, we should all aspire to do that!

src/Compiler/CompilerOptions.cpp
chentong319: fixed

AlexandreEichenberger (approved):
LGTM, thanks for this exploratory work.

PR #2946, opened 2024-09-19
[mlir2FileCheck tool] handle multiple output arguments

tungld:
The python script `mlir2FileCheck` does not handle multiple output arguments, for example, it prints out:
```
// CHECK:           [[VAR_Out_]], [[VAR_RecScale_]], [[VAR_Offset_:%.+]] = ....
```

where only the last output has `:%.+`. The correct one would be
```
// CHECK:           [[VAR_Out_:%.+]], [[VAR_RecScale_:%.+]], [[VAR_Offset_:%.+]] = "
```

This fixes the issue by defining a new regex that captures a repeated group that contains all output arguments, then parses them.


AlexandreEichenberger (approved):
LGTM, Big thanks, I was always manually editing the multiple output operations. This is much better.

PR #2949, opened 2024-09-23
Add support of return for KrnlCallOp

chentong319:
This PR allows KrnlCallOp to have a scalar return, AnyFloat or AnyInteger, and the lowering will use the return of func.call to implement the return. When there is no return, the previous usage does not need to change.

A lit test is added. The input is
```
%1 = "krnl.call"() {funcName = "get_omp_num_thread", numOfOutput = 0 : si64} : () -> (i32)
```
The output is 
```
 %0 = llvm.call @get_omp_num_thread() : () -> i32
```
The krnl call can be generated with 
```
rewriter.create<KrnlCallOp>(loc, TypeRange(rewriter.getIntegerType(32)), "omp_get_num_thread", 0, ValueRange());
```

### DIFF
--- a/src/Dialect/Krnl/Krnl.td
+++ b/src/Dialect/Krnl/Krnl.td
@@ -89,6 +89,13 @@ def KrnlCallOp : Op<Krnl_Dialect, "call",
       DefaultValuedAttr<SI64Attr, "1">:$numOfOutput,
       Variadic<AnyType>:$parameters);
 
+  // Return Value for the Call.
+  // No return if the type is NoneType (void in llvm)
+  // Only scalar type is supported now.
+  // In future, return of memref can be supported with pointer of OMTensor.
+  // The returned memref will be created inside the call. 
+  let results = (outs Variadic<AnyTypeOf<[F32, F64, I32, I64]>>:$returnValue);
+
   // builders to build KrnlCallOp from op and operands, helping conversion from
   // onnx to krnl.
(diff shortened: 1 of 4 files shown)


src/Dialect/Krnl/Krnl.td
```diff
@@ -89,13 +89,22 @@ def KrnlCallOp : Op<Krnl_Dialect, "call",
       DefaultValuedAttr<SI64Attr, "1">:$numOfOutput,
       Variadic<AnyType>:$parameters);
 
+  // Return Value for the Call.
+  // No return if the type is NoneType (void in llvm)
+  // Only scalar type is supported now.
+  // In future, return of memref can be supported with pointer of OMTensor.
+  // The returned memref will be created inside the call. 
+  let results = (outs Variadic<AnyTypeOf<[AnyFloat, AnyInteger]>>:$returnValue);
```
AlexandreEichenberger, line 97: Having a Variadic for zero or one feels a bit inelegant, but it woks.

AlexandreEichenberger (approved):
LGTM, big thanks for the quick PR, looking forward to use it.

src/Dialect/Krnl/Krnl.td
chentong319: Is there a way to define the maximum number of variadic?

AlexandreEichenberger: There are constraints that can be added on values, not sure about the variadic

PR #2951, opened 2024-09-24
[NNPA] Simplify rules in zhigh-to-onnx pass and change some pass order

tungld:
Simplify some rules in `zhigh-to-onnx` pass and change the order of calling `zhigh-to-onnx` and `constprop-zhigh` in which `constprop-zhigh` is called after `zhigh-to-onnx` so that `zhigh-to-onnx` can recompose `stick->nnpa_op->unstick` back into an onnx op.

### DIFF
--- a/src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
+++ b/src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
@@ -90,10 +90,12 @@ void addONNXToZHighPasses(mlir::PassManager &pm) {
 
   pm.addPass(onnx_mlir::createONNXToZHighPass());
   pm.addNestedPass<func::FuncOp>(onnx_mlir::createShapeInferencePass());
+
   // There are more opportunities for const propagation once all zhigh ops were
   // generated.
   pm.addNestedPass<func::FuncOp>(onnx_mlir::createConstPropONNXToONNXPass());
   pm.addPass(mlir::createCanonicalizerPass());
+
   // Layout propagation at ZHighIR.
   pm.addNestedPass<func::FuncOp>(
       onnx_mlir::zhigh::createZHighLayoutPropagationPass());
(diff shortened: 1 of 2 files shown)


tungld:

src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
```diff
@@ -152,6 +140,20 @@ void addONNXToZHighPasses(mlir::PassManager &pm) {
         onnx_mlir::zhigh::createZHighRecomposeToStickUnstickPass());
   }
 
+  // After all optimizations, if there are still light-weight ops (e.g. add,
+  // sub, ...) that are of `stick -> light-weight op -> unstick`, it's better to
+  // use CPU instead of NNPA to avoid stick/unstick. CPU is efficient to handle
+  // these ops, e.g vectorize the computation.
+  if (nnpaEnableZHighToOnnx)
+    pm.addNestedPass<func::FuncOp>(onnx_mlir::createZHighToONNXPass());
```
tungld, line 148: Move it here after all optimizations have been applied to clean up the IR.

src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
```diff
@@ -152,6 +140,20 @@ void addONNXToZHighPasses(mlir::PassManager &pm) {
...
+
+  // Constant propagation at ZHighIR: constant stickify.
+  // Only support BE machines.
+  bool isBE = llvm::endianness::native == llvm::endianness::big;
+  if (isBE)
+    pm.addNestedPass<func::FuncOp>(
+        onnx_mlir::zhigh::createZHighConstPropagationPass());
```
tungld, line 155: Move const propagation after recomposing `stick->nnpa_op->unstick` because once const propatation for stick is applied it's hard to recompose.

AlexandreEichenberger: NNPA is only for BE machines, so is that for having no issues on lit-tests?

AlexandreEichenberger (approved):
LGTM

src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
tungld: Yes, no issues on lit-tests. Lit tests for that constant propagation only run on BE machines.

PR #2952, opened 2024-09-25
Revert the rule that rewrites a quantize-dequantize pair to identity

tungld:
Remove the rule that rewrites a quantize-dequantize pair to identity. It's because there is a rounding in the QuantizeLinear's formulas so that the pair is mathematically not the same as identity.


AlexandreEichenberger (approved):
LGTM

mgehre-amd:
Thank you for the quick fix!

PR #2953, opened 2024-09-25
Cleanup of krnl iterate loops

AlexandreEichenberger:
The code used a mixture of constant and non-constant DialectBuilder, and that caused some strange code pattern. Now all of the lambda functions used in the Dialect Builders used constant Dialect Builder parameters.

This PR also change trivial

```
loops = krnl.defineLoops(1);
krnl.iterateIE(loops, loops, {lb}, {ub}, ...);
```

into the new pattern
```
krnl.forLoopIE(lb, ub, 1, /*parallel*/ false, ...);
```
Sometimes, blocking>1 and parallel can be true, depending on what the code was.

None of the changes here are functional.

### DIFF
--- a/docs/Dialects/krnl.md
+++ b/docs/Dialects/krnl.md
@@ -929,6 +929,35 @@ Typically it is used for optional arguments used in KrnlCallop.
 | :----: | ----------- |
 | `none_val` | none type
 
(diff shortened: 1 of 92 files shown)


tungld (approved):
LGTM!

Issue #2950, opened 2024-09-23
How to pass clang options to be applied to generated object file?


jinchenglee:
For example, I can compile my binary using `clang -target riscv64-unknown-elf -fno-builtin-printf -fno-lto -O2 -c example.c -o example.o`, how may I pass down the additional options through onnx-mlir? This might be a dumb question as I'm new to LLVM toolsets. 

I tried `$ onnx-mlir -mtriple=riscv64-unknown-elf -mllvm="-fno-builtin-printf -fno-lto -O2" --EmitObj ./nn.onnx`:
```
[1/6] Mon Sep 23 15:21:19 2024 (0s) Importing ONNX Model to MLIR Module from "nn.onnx"
[2/6] Mon Sep 23 15:21:19 2024 (0s) Compiling and Optimizing MLIR Module
[3/6] Mon Sep 23 15:21:19 2024 (0s) Translating MLIR Module to LLVM and Generating LLVM Optimized Bitcode
opt: Unknown command line argument '-fno-builtin-printf -fno-lto -O2'.  Try: 'opt --help'
opt: Did you mean '--rotation-prepare-for-lto'?
opt -O0 --mtriple=riscv64-unknown-elf --code-model small -fno-builtin-printf -fno-lto -O2 -o ./nn.bc ./nn.unoptimized.bc
Error message:
Program path: ~/git_repo/llvm-project/build/bin/opt
Command execution failed.
```

jinchenglee:
Seems only a single option is allowed/supported. 

```
$~/experimental/onnx-mlir-example$ onnx-mlir -mtriple=riscv64-unknown-elf -mllvm="--disable-simplify-libcalls -O2" --EmitObj ./nn.onnx
[1/6] Mon Sep 23 15:35:49 2024 (0s) Importing ONNX Model to MLIR Module from "nn.onnx"
[2/6] Mon Sep 23 15:35:49 2024 (0s) Compiling and Optimizing MLIR Module
[3/6] Mon Sep 23 15:35:49 2024 (0s) Translating MLIR Module to LLVM and Generating LLVM Optimized Bitcode
opt: Unknown command line argument '--disable-simplify-libcalls -O2'.  Try: 'opt --help'
opt: Did you mean '--disable-simplify-libcalls'?
opt -O0 --mtriple=riscv64-unknown-elf --code-model small --disable-simplify-libcalls -O2 -o ./nn.bc ./nn.unoptimized.bc
Error message:
[...]

PR #2955, opened 2024-09-25
Handle the options passed to llvm opt and llvm llc

chentong319:
Issue #2950 discovered the issue in passing options llvm. There are several problems:
1. The string of the options has to be split into vector of strings for the command format.
2. Conflict in options. So far, the optimization level is the only example.  The level specified with  particular step should have higher priority than that for the whole compiler. Thus, I can avoid multiple "-O#" error.
3. modify the comments for mllvm, xopt, and xllc to make it easy for user to pass multiple values to these options 

I tested with some examples.

### DIFF
--- a/src/Compiler/CompilerOptions.cpp
+++ b/src/Compiler/CompilerOptions.cpp
@@ -64,8 +64,6 @@ bool VerboseOutput;                                    // onnx-mlir only
 std::vector<std::string> Xopt;                         // onnx-mlir only
 std::vector<std::string> Xllc;                         // onnx-mlir only
 std::string mllvm;                                     // onnx-mlir only
-std::string mllvmopt;                                  // onnx-mlir only
-std::string mllvmllc;                                  // onnx-mlir only
 std::string instrumentOps;                             // onnx-mlir only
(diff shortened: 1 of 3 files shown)


tungld:
> Options of -mllvm is for both llvm opt and llc, but their options are not the same. New option, -mllvmopt and -mllvmllc, are introduced to pass specific option to opt or llc respectively.

Look like `-mllvmopt` and `-mllvmllc` are similar to two existing options `-Xopt` and `-Xllc`.

chentong319:
> > Options of -mllvm is for both llvm opt and llc, but their options are not the same. New option, -mllvmopt and -mllvmllc, are introduced to pass specific option to opt or llc respectively.
> 
> Look like `-mllvmopt` and `-mllvmllc` are similar to two existing options `-Xopt` and `-Xllc`.

That's true. The only difference is that each value of opt and xllc has the same limitation of mllvm: only one option is allowed. I removed the new llvmopt and llvmllc, and modified the option explanation for opt and xllc.

tungld (approved):
LGTM.

PR #2961, opened 2024-10-02
Java8 unstick issue

AlexandreEichenberger:
I suspect that the code generation pattern for Unstick may have been too complicated for the specific constant sizes found in the java8 program failed.

This PR simplifies the pattern for predictive constant patterns.


chentong319 (approved):
LGTM!
In future, we should introduce creator for true/false literal for IE, instead of using `1>0`.

AlexandreEichenberger:
> we should introduce greater for true/false literal for IE, instead of using 1>0.

very good suggestion, will do in a subsequent PR as this one needs to be merged asap.

PR #2963, opened 2024-10-02
Add --enable-unsafe-fp-math flag when -O3 and NNPA is used

chentong319:
Since NNPA uses less precise data, it is reasonable to enable aggressive optimization.
Add the flag only for O3, leaving O0 for debugging.

### DIFF
--- a/src/Compiler/CompilerOptions.cpp
+++ b/src/Compiler/CompilerOptions.cpp
@@ -1246,6 +1246,13 @@ void initCompilerConfig() {
     addCompilerConfig(CCM_SHARED_LIB_DEPS, extraLibs);
     addCompilerConfig(CCM_SHARED_LIB_PATH_DEPS, extraLibPaths);
   }
+
+  // Enable aggressive optimization for NNPA with -O3
+  if (OptimizationLevel == OptLevel::O3 &&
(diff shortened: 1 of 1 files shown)


src/Compiler/CompilerOptions.cpp
```diff
@@ -1246,6 +1246,13 @@ void initCompilerConfig() {
     addCompilerConfig(CCM_SHARED_LIB_DEPS, extraLibs);
     addCompilerConfig(CCM_SHARED_LIB_PATH_DEPS, extraLibPaths);
   }
+
+  // Enable aggressive optimization for NNPA with -O3
+  if (OptimizationLevel == OptLevel::O3 &&
+      getTargetAccel().find("NNPA") != std::string::npos &&
+      getLLVMOption().find("enable-unsafe-fp-math") == std::string::npos) {
```
AlexandreEichenberger, line 1253: I see, the idea is that you just check if that opt is already there. If it's there, either its `true` and there is no need to add it again, or `false` and then we don't add it also. Clever.

AlexandreEichenberger (approved):
LGTM, will check if it makes a difference for the test that I was looking at.

If it makes a difference, I almost wonder if we should not always turn it on at O3.

PR #2965, opened 2024-10-04
Recompose multiple ops into a single ONNXGelu

tungld:
This patch recomposes multiple operations into a single ONNXGelu operation.


AlexandreEichenberger (approved):
LGTM, thanks for expanding the usage of the `matchValueAndOp` in such a meaningful way. Really make the matching code more readable.

hamptonm1:
@tungld I know this was merged already but still wanted to review. Looks good to me!

PR #2969, opened 2024-10-08
Do not rewrite add/mul of scalar values in constprop

tungld:
This patch updates two associate rules in constant propagation for Add and Mul to avoid rewriting when both two inputs are scalar. In particular, these are two rules to update:
- `(x + (y + c))` to `(x + y) + c`: do not apply this when `y` and `c` are scalar
- `(x + c) + y)` to `(x + y) + c`: do not apply this when `x` and `c` are scalar


AlexandreEichenberger (approved):
LGTM, looking for scalar is a first step towards looking at broadcasting rules to help direct re-associativity rules.

PR #2970, opened 2024-10-09
Steer round to even to Z's hardware operation (for Z), and MLIR/LLVM roundEven for the other platforms.

AlexandreEichenberger:
Round to nearest even was currently implemented in software:
```
  // Use numpy algorithm for rint as follows.
  // ```
  // double y, r;
  // y = npy_floor(x);
  // r = x - y;
  //
  // if (r > 0.5) {
  //     y += 1.0;
  // }
  //
  // /* Round to nearest even */
  // if (r == 0.5) {
  //     r = y - 2.0*npy_floor(0.5*y);
  //     if (r == 1.0) {
  //         y += 1.0;
  //     }
  // }
```

This can be replaced by one instruction on many architectures (e.g. on Z by `FIEBRA` or `VFISB`). This PR does that for z machines.

For others, MLIR & LLVM actually implement a `math::round_even` that leads to efficient instructions as well. At this time, this is not enabled for Z, thus the direct use of ASM for the Z platform. If the ASM cannot be used on Z, then the `math::round_even_emulation` (algorithm above) performs better than the default LLVM function call backup. For other machines, the LLVM round even is used.

Another issue with using ASM driven directive is that we cannot use the LLVM features that let us used SIMD vectors of arbitrary lengths. I baked this support in the ONNX to Krnl lowering using `vector::ShapeCastOp` and then `extract` and `insert`. That same approach can be used elswhere. I checked the HW that the generated `shuffle` are eventually all gone (as we shuffle entire vectors). Right now, it is handled in the ONNX to KRNL. It could be migrated to Krnl to LLVM, but would require more work that I am not too familiar with.

### DIFF
--- a/src/Dialect/Krnl/DialectBuilder.cpp
+++ b/src/Dialect/Krnl/DialectBuilder.cpp
@@ -349,6 +349,16 @@ Value KrnlBuilder::constant(MemRefType type, StringRef name,
       alignment.value_or(nullptr));
 }
 
+//===----------------------------------------------------------------------===//
+// Math style functions.
+
+Value KrnlBuilder::round(Value input) const {
+  return b().create<KrnlRoundOp>(loc(), input.getType(), input);
+}
+
+//===----------------------------------------------------------------------===//
+// C library functions.
+
 void KrnlBuilder::memcpy(Value dest, Value src, Value numElems) const {
   MultiDialectBuilder<MathBuilder> create(*this);
   Value zero = create.math.constantIndex(0);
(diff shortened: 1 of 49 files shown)


src/Dialect/Krnl/DialectBuilder.cpp
```diff
@@ -349,6 +349,49 @@ Value KrnlBuilder::constant(MemRefType type, StringRef name,
...
+      Value input2D = create.vec.shapeCast(vecType2D, input);
+      Value output2D = input2D;
+      for (int64_t i = 0; i < numArchVec; ++i) {
+        Value subInput = create.vec.extractFrom2D(input2D, i);
+        Value subOutput =
+            b().create<KrnlRoundEvenOp>(loc(), subInput.getType(), subInput);
+        output2D = create.vec.insertInto2D(subOutput, output2D, i);
+      }
+      return create.vec.shapeCast(vecType, output2D);
+    } else {
```
tungld, line 382: This `else` looks redundant since we have `return` inside the `if`.

tungld (approved):
LGTM.

src/Dialect/Krnl/DialectBuilder.cpp
AlexandreEichenberger: Tx, will fix

PR #2972, opened 2024-10-11
Update c++ driver for running onnx model

chentong319:
The interface of StringRef was changed in the new version of llvm_project.  The member function `equal` is removed. This PR uses `compare` to do the job.


tungld (approved):
LGTM!

PR #2977, opened 2024-10-15
Re-enable ReduceMax/Min backend tests

hamptonm1:
Since this PR was merged: https://github.com/onnx/onnx-mlir/pull/2744, we now have the support for `keepdims` and as a result a few of the backend test can be re-enabled.


mikeessen (approved):
Looks good!  Thanks!

PR #2978, opened 2024-10-16
Remove a spike of memory usage in ScrubDisposablePass.

imaihal:
This PR removes a spike in memory usage. When checking memory usage in a LLM (Mistral-7b), ScrubDisposablePass consumes a lot of memory only for a moment(spike about 50GB). This is likely to due to the fact that having both denseElementsAttr and disposalElementsAttr are hold during conversion. 
In Mistral-7b model, the spike is about 23GB, and total memory usage including spike is about 50GB.  This PR removes the spike and total memory usage reduces to 27GB.

### DIFF
--- a/src/Dialect/ONNX/ElementsAttr/DisposablePool.cpp
+++ b/src/Dialect/ONNX/ElementsAttr/DisposablePool.cpp
@@ -122,9 +122,8 @@ void DisposablePool::scrub(ModuleOp moduleOp, OpAttrDictionary opsAttrs) {
         for (auto &[id, translation] : batch) {
           auto &[disposable, dense] = translation;
           dense = disposable.toDenseElementsAttr();
-          // TODO: Consider calling disposable.dispose() here to free up memory
-          //       on the go to make memory available to create the next
-          //       DenseElementsAttr. In that case should we lock mutex?
+          disposable.dispose();
+          // TODO: In that case should we lock mutex?
         }
       }
     };
--- a/src/Dialect/ONNX/ElementsAttr/DisposablePool.cpp
+++ b/src/Dialect/ONNX/ElementsAttr/DisposablePool.cpp
@@ -122,14 +122,8 @@ void DisposablePool::scrub(ModuleOp moduleOp, OpAttrDictionary opsAttrs) {
         for (auto &[id, translation] : batch) {
           auto &[disposable, dense] = translation;
           dense = disposable.toDenseElementsAttr();
-        }
-
-        // Delete DisposableElementsAttr for each batch with lock after
-        // conversion.
-        {
-          const std::lock_guard<std::mutex> lock(translationMutex);
-          for (auto &[id, translation] : batch) {
-            auto &[disposable, dense] = translation;
+          {
+            const std::lock_guard<std::mutex> lock(translationMutex);
             disposable.dispose();
           }
         }
--- a/src/Dialect/ONNX/ElementsAttr/DisposablePool.cpp
+++ b/src/Dialect/ONNX/ElementsAttr/DisposablePool.cpp
@@ -114,7 +114,7 @@ void DisposablePool::scrub(ModuleOp moduleOp, OpAttrDictionary opsAttrs) {
       return llvm::make_range(batchBegin, batchEnd);
     };
     // Parallel worker body: Fetch and process batches until there are no more.
-    auto work = [&fetchBatch](size_t threadNumber) {
+    auto work = [&fetchBatch, &translationMutex](size_t threadNumber) {
       for (;;) {
         auto batch = fetchBatch();
         if (batch.empty())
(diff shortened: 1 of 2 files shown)


src/Dialect/ONNX/ElementsAttr/DisposablePool.cpp
```diff
@@ -122,9 +122,8 @@ void DisposablePool::scrub(ModuleOp moduleOp, OpAttrDictionary opsAttrs) {
         for (auto &[id, translation] : batch) {
           auto &[disposable, dense] = translation;
           dense = disposable.toDenseElementsAttr();
-          // TODO: Consider calling disposable.dispose() here to free up memory
-          //       on the go to make memory available to create the next
-          //       DenseElementsAttr. In that case should we lock mutex?
+          disposable.dispose();
+          // TODO: In that case should we lock mutex?
```
AlexandreEichenberger, line 126: Any idea if the mutex is needed? If so, can you check in MLIR or IREE if they use a lock?

chentong319:
What the relationship of this PR and #PR2917? Should this PR go first? Or for onnx-mlir passes, should we always use disposableElementsAttr to reclaim the space when constants are translated into krnl.Global or LLVM constant?

src/Dialect/ONNX/ElementsAttr/DisposablePool.cpp
imaihal: I also think lock is not needed. Since I found the comment about the lock, I wanted to here your comments. Thanks!

imaihal:
> What the relationship of this PR and #PR2917? Should this PR go first? Or for onnx-mlir passes, should we always use disposableElementsAttr to reclaim the space when constants are translated into krnl.Global or LLVM constant?

This is different issue and no relationship with PR2917. Currently when converting from disposalElementsAttr to denseElementsAttr in this pass,  disposalElementsAttr is kept even after conversion.

imaihal:
@sorenlassen I think you wrote the comments about lock in this code. The comments is a bit old, but do you remember something about the lock.

imaihal:
I updated to use the lock when deleting disposalElementsAttr. They are deleted for each batch since I'm concerned the performance degradation by frequent lock. Could you review again?

src/Dialect/ONNX/ElementsAttr/DisposablePool.cpp
```diff
@@ -114,17 +114,24 @@ void DisposablePool::scrub(ModuleOp moduleOp, OpAttrDictionary opsAttrs) {
...
-          // TODO: Consider calling disposable.dispose() here to free up memory
-          //       on the go to make memory available to create the next
-          //       DenseElementsAttr. In that case should we lock mutex?
+        }
+
+        // Delete DisposableElementsAttr for each batch with lock after
+        // conversion.
+        {
+          const std::lock_guard<std::mutex> lock(translationMutex);
+          for (auto &[id, translation] : batch) {
```
tungld, line 131: Can we lock  the original `for-loop` at Line 122 instead of creating a new loop? I guess calling `dispose()` inside the original loop can help reduce the peak memory.

imaihal: Yes. I used the lock for each batch because I was concerned frequent lock might affect the compile time and I confirmed the spike can be removed with the lock for each batch.
However, from my test, it seems to be no performance difference with original loop. I will update.

src/Dialect/ONNX/ElementsAttr/DisposablePool.cpp
```diff
@@ -114,17 +114,18 @@ void DisposablePool::scrub(ModuleOp moduleOp, OpAttrDictionary opsAttrs) {
...
         auto batch = fetchBatch();
         if (batch.empty())
           break;
         for (auto &[id, translation] : batch) {
           auto &[disposable, dense] = translation;
           dense = disposable.toDenseElementsAttr();
-          // TODO: Consider calling disposable.dispose() here to free up memory
-          //       on the go to make memory available to create the next
-          //       DenseElementsAttr. In that case should we lock mutex?
+          {
```
tungld, line 125: This `{}` block looks redundant, since the scope of mutex can be the scope of for-loop.

tungld (approved):
LGTM.

src/Dialect/ONNX/ElementsAttr/DisposablePool.cpp
imaihal: I thought we need to use mutex only for `disposable.dispose()` so that `dense = disposable.toDenseElementsAttr();` runs in parallel.

tungld: It’s Ok, but be careful when creating attributes runs in parallel, we may have a spike again before the buffers are disposed.

chentong319 (approved):
LGTM!

PR #2981, opened 2024-10-17
Sorted all invocations of alloca

AlexandreEichenberger:
Alloca was shown to be an issue with large benchmarks, where too much stack alloc ended up crashing the process. Alloca are typically used for temporaries needed within the computations of an onnx operation.

Went through all instances of `alloca`, and left alone the scalars (allocating a few bytes per operations is fine). For the larger data, I migrated it to `alloc`.

Because `alloc` results in a more expensive call to allocate and free, when not using parallelism, I moved such calls outside the loops. But when using parallelism, such call must be inside the parallel region as each parallel thread must have its version of temporaries, so in such cases the `alloc` stayed inside the parallel loop.

I ensured that the buffer hoisting did not bring the alloc outside of the `scope_alloca`, which seems to also catch the `alloc` operations. I ran individual checks on several key operations, as at this time the verifying to parallel operations in our CIs are still very sketchy.

```
  CheckONNXModel.py -m gemm.mlir -r="-O3 -march=arm64" -a="-parallel"
  CheckONNXModel.py -m reduceMin1.mlir -r="-O3 -march=arm64 -shapeInformation=0:1024x1024x1024" -a="-parallel"
  CheckONNXModel.py -m layernorm4d.mlir -r="-O3 -march=arm64 " -a="-parallel"
```

### DIFF
--- a/src/Accelerators/NNPA/Conversion/ZLowToLLVM/ZLowToLLVMCommon.cpp
+++ b/src/Accelerators/NNPA/Conversion/ZLowToLLVM/ZLowToLLVMCommon.cpp
@@ -111,7 +111,8 @@ Value ZTensorHelper::getPreTransformedDescPtr(zdnn_data_types zDNNDataType,
   Type llvmZTensorDescStructTy = getZTensorDescStructTy(context);
   Value one = create.llvm.constant(llvmI64Ty, static_cast<int64_t>(1));
 
-  // TODO: evaluate if a heap alloc would not be better.
+  // Alloca is fine for LLVM structs; if we were to use alloc, we would also to
+  // manually insert free calls. So alloca makes total sense here.
   Value preTransformedDescPtr = create.llvm._alloca(
       krnl::getPointerType(context, llvmZTensorDescStructTy),
       llvmZTensorDescStructTy, one,
@@ -155,7 +156,6 @@ Value ZTensorHelper::getTransformedDescPtr(
   Type llvmZTensorDescStructTy = getZTensorDescStructTy(context);
   Value one = create.llvm.constant(llvmI64Ty, static_cast<int64_t>(1));
 
-  // TODO: evaluate if a heap alloc would not be better.
   Value transformedDescPtr = create.llvm._alloca(
       krnl::getPointerType(context, llvmZTensorDescStructTy),
       llvmZTensorDescStructTy, one,
@@ -217,7 +217,6 @@ ZTensor ZTensorHelper::getZTensor(Value bufferPtr, zdnn_data_types dataType,
   Value transformedDescPtr =
       getTransformedDescPtr(preTransformedDescPtr, isConcat, concatInfo);
   // Create the input zTensor.
-  // TODO: evaluate if a heap alloc would not be better.
   Value alloc =
       create.llvm._alloca(krnl::getPointerType(context, llvmZTensorStructTy),
           llvmZTensorStructTy, one,
@@ -253,7 +252,6 @@ ZTensor ZTensorHelper::getZTensor(Value preTransformedDescPtr,
   Type llvmZTensorStructTy = getZTensorStructTy(context);
   Value one =
       create.llvm.constant(rewriter.getI64Type(), static_cast<int64_t>(1));
-  // TODO: evaluate if a heap alloc would not be better.
   Value alloc =
       create.llvm._alloca(krnl::getPointerType(context, llvmZTensorStructTy),
           llvmZTensorStructTy, one,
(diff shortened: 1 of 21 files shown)


src/Accelerators/NNPA/Conversion/ZLowToLLVM/ZLowToLLVMCommon.cpp
```diff
@@ -215,6 +217,7 @@ ZTensor ZTensorHelper::getZTensor(Value bufferPtr, zdnn_data_types dataType,
   Value transformedDescPtr =
       getTransformedDescPtr(preTransformedDescPtr, isConcat, concatInfo);
   // Create the input zTensor.
+  // TODO: evaluate if a heap alloc would not be better.
```
tungld, line 220: I often see in MLIR they use alloca to create a LLVM struct.

src/Conversion/KrnlToAffine/KrnlMatmul.cpp
```diff
@@ -372,13 +372,7 @@ class KrnlMatmulLowering : public ConversionPattern {
...
            "alignment of buffers cannot be smaller than the default alignment "
            "(which is set for SIMD correctness");
-    // TODO: alloca is good as it help simplify away this data structures (as it
-    // is only used as local temp, basically extensions of registers). However,
-    // there might be issues with non-removed alloca when they are not in the
-    // innermost loop. Still think its worth it having alloca as we want
-    // eventually all the refs to alloca to be register/spill access, not memory
-    // load/stores.
-    Value TmpProd = create.mem.alignedAlloca(CTmpType, BUFFER_ALIGN);
+    Value TmpProd = create.mem.alignedAlloc(CTmpType, BUFFER_ALIGN);
```
tungld, line 375: If it is a good practice to use `alignedAlloca` for scalar type, should we add a verification in `create.mem.alignedAlloca` that checks type would be scalar. By that way, we can avoid the situation where users call `create.mem.alignedAlloca` for non-scalar type.

tungld (approved):
LGTM. 

I remember we encounted segfaults because of `alloca` in the past and added the hoisting pass to move it outside loops. It's a pity to hear that that pass does not work with `scope_alloca`.

I see [this pass](https://mlir.llvm.org/docs/Passes/#-promote-buffers-to-stack) in MLIR to promote small alloc to alloca, not sure if it works or not.

src/Accelerators/NNPA/Conversion/ZLowToLLVM/ZLowToLLVMCommon.cpp
AlexandreEichenberger: Make sense, it's still small. Want me to remove the comment? @tungld

tungld: Yes, it's ok to remove it. Thanks!

src/Conversion/KrnlToAffine/KrnlMatmul.cpp
AlexandreEichenberger: > I remember we encounted segfaults because of alloca in the past and added the hoisting pass to move it outside loops. It's a pity to hear that that pass does not work with scope_alloca.

`scope_alloca` stop the migration of `alloc` and `alloca` out of that scope. It is important for parallel regions, as having it stay in the parallel region means that each thread will get its own. So that works fine.

Now I also manually moved the case with no parallel out of the loops, just to make sure that is fine. Would you like me to check if it's really needed? That way we could have the call always inside, and if parallel is there, its blocked and otherwise it is not?

In one case, I really need it outside manually, because it was inside of an inner-loop if-then-else. So by having it outside of the if-then-else (allocating in all case), it allows it to migrate to the outside.

AlexandreEichenberger:
> I remember we encounted segfaults because of alloca in the past and added the hoisting pass to move it outside loops. It's a pity to hear that that pass does not work with scope_alloca.

`scope_alloca` stop the migration of `alloc` and `alloca` out of that scope. It is important for parallel regions, as having it stay in the parallel region means that each thread will get its own. So that works fine.

Now I also manually moved the case with no parallel out of the loops, just to make sure that is fine. Would you like me to check if it's really needed? That way we could have the call always inside, and if parallel is there, its blocked and otherwise it is not?

In one case, I really need it outside manually, because it was inside of an inner-loop if-then-else. So by having it outside of the if-then-else (allocating in all case), it allows it to migrate to the outside.

src/Accelerators/NNPA/Conversion/ZLowToLLVM/ZLowToLLVMCommon.cpp
AlexandreEichenberger: Done.

src/Conversion/KrnlToAffine/KrnlMatmul.cpp
AlexandreEichenberger: Left for another PR.

AlexandreEichenberger:
Responded to comments, and removed the placement of most "alloc" before/inside the main loop depending on parallel. Placing it inside works in both cases. If its parallel, it remains inside because of the inserted `scope_alloca`; if its sequential, then there is no scope and it naturally migrate outside the main loop.

tungld:
> If its parallel, it remains inside because of the inserted scope_alloca; if its sequential, then there is no scope and it naturally migrate outside the main loop.

Great, thanks! It was what I expected.

PR #2982, opened 2024-10-18
[RunONNXModel.py] Add options to print out input/output signatures and support big models

tungld:
This patch improves RunONNXModel.py script with the following changes:
- Add two options `--print-input-signature` and `--print-output-signature` to print out signatures.
- Support big models that are compiled into two output files: `model.so` and `model.constants.bin`

### DIFF
--- a/utils/RunONNXModel.py
+++ b/utils/RunONNXModel.py
@@ -769,7 +769,7 @@ class InferenceSession:
         start = time.perf_counter()
         shared_lib_path = self.model_path + f"/{self.default_model_name}.so"
         if not os.path.exists(shared_lib_path):
-            print(f"Input model {shared_lib_path} does not exists")
+            print(f"Input model {shared_lib_path} does not exist")
             exit(0)
         print("Loading the compiled model ...")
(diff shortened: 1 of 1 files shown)


AlexandreEichenberger (approved):
LGTM

Thanks, that is super useful. Is there a need to distinguish between input and output? I would have imagined that a common `--print-signatures` would have been enough. That way, it would also reduce the increasingly long list of options.

Also, a small pet peeve: the old `-save-so` accepted a directory, and would automatically "name" the .so; but the `-load-so` needed an explicit `so` file. Ideally, the behavior should be identical for both save and load, with my preference going to being able to name a directory in both cases.

tungld:
>  I would have imagined that a common --print-signatures would have been enough

I merged two options into a single one `--print-signatures`. Thanks!

>  Ideally, the behavior should be identical for both save and load, with my preference going to being able to name a directory in both cases. If the final behavior could be explicit in the "help" that would be good.

Yes, both new `--save-model` and `--load-model` accept a folder only.

PR #2984, opened 2024-10-25
RunONNXModel.py: Add a `--cache-model=path` option

AlexandreEichenberger:
Right now, we can save a model with `-save-model=x` and then reuse it with `-load-model=x`.

The current `-cache-model=x` option will save the model when the model does not exist, and reuse it when it exists.

Minor, but that way I can run a model using the same scripts many times in a row and benefit from caching,

### DIFF
--- a/utils/RunONNXModel.py
+++ b/utils/RunONNXModel.py
@@ -63,23 +63,23 @@ parser.add_argument(
     nargs="?",
     const="compilation.log",
     default=None,
-    help="Output compilation messages to file, default compilation.log",
+    help="Output compilation messages to file, default compilation.log.",
 )
 parser.add_argument(
     "-m",
     "--model",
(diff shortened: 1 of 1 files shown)


utils/RunONNXModel.py
```diff
@@ -639,6 +647,25 @@ def __init__(self, model_file, **kwargs):
...
+                )
+                args.save_model = args.cache_model
+            else:
+                print(
+                    'Cached compiled model found in "'
+                    + args.cache_model
+                    + '": load model this run.'
+                )
+                args.load_model = args.cache_model
+            args.cache_model = None
```
tungld, line 667: Got it. It is a combination of `save-model` and `load-model` so that we don't need to run twice.

tungld (approved):
LGTM. It's really helpful. Thanks!

PR #2985, opened 2024-10-25
Enable check-onnx-backend-numerical-nnpa on Jenkins s390x

tungld:
Enable `make check-onnx-backend-numerical-nnpa` on Jenkins s390x for testing NNPA.

JNI tests for NNPA failed for convolutional models, so we temporarily disable these JNI tests.


AlexandreEichenberger (approved):
LGTM, thanks

Issue #2987, opened 2024-10-25
Error in OMExternalConstants.inc when opening model_name.constants.bin file on z/OS


christopherlmunoz:
After linking the .so, run this client command resulted with an extra backtick or misc characters getting added to file name after converting from ASCII to EBCIDIC.

```
client/bin/modelzoo --lib mnist-8-with-binary.so --file mnist-8.tests --validate -m VERBOSE --fc-parms 0.01,0.158752,9,10
Saved mnist-8-with-binary.so for debug testing with C clients
Created mnist-8-with-binary.jar for Java clients
Iteration 0 dataset 0: Reading
gen_data_set_lists() loading 1 input tensor(s):
gen_next_tensor() processing tensor with shape ( 1 1 28 28 ) totaling 784 elements of type ONNX_TYPE_FLOAT
gen_data_set_lists() loading 1 expected output tensor(s):
gen_next_tensor() processing tensor with shape ( 1 10 ) totaling 10 elements of type ONNX_TYPE_FLOAT
Iteration 0 dataset 0: Running
Error while opening mnist-8.constants.bin`
Error while opening mnist-8.constants.bin�K������k�K������k`��K�����`
```

```
void omMMapBinaryFile(
    void **constAddr, char *filename, int64_t size, int64_t isLE) {
  checkEndianness(isLE);
  char *fname = filename;
#ifdef __MVS__
  // Convert the file name to EBCDIC for the open call.
  char *tPath;
  size_t tLen = strlen(fname);
  tPath = (char *)malloc(tLen);
  if (!tPath) {
    fprintf(stderr, "Error while malloc");
    return;
  }
  memcpy(tPath, fname, tLen);
  __a2e_s(tPath);
  fname = tPath;
  //fname[tLen] = '\0'; --------------------------------------------------->  possible fix. 
#endif
```

christopherlmunoz:
@tungld Hi, please let me know if you need more information. Thank you for taking a look.

tungld:
@christopherlmunoz adding the terminator looks good. Could you create a PR to fix this? Thanks!

PR #2991, opened 2024-10-30
Opening binary constants files fix on zOS

christopherlmunoz:
Addressing issue https://github.com/onnx/onnx-mlir/issues/2987.

After linking the .so, run this client command resulted with an extra backtick or misc characters getting added to file name after converting from ASCII to EBCIDIC.

```
client/bin/modelzoo --lib mnist-8-with-binary.so --file mnist-8.tests --validate -m VERBOSE --fc-parms 0.01,0.158752,9,10
Saved mnist-8-with-binary.so for debug testing with C clients
Created mnist-8-with-binary.jar for Java clients
Iteration 0 dataset 0: Reading
gen_data_set_lists() loading 1 input tensor(s):
gen_next_tensor() processing tensor with shape ( 1 1 28 28 ) totaling 784 elements of type ONNX_TYPE_FLOAT
gen_data_set_lists() loading 1 expected output tensor(s):
gen_next_tensor() processing tensor with shape ( 1 10 ) totaling 10 elements of type ONNX_TYPE_FLOAT
Iteration 0 dataset 0: Running
Error while opening mnist-8.constants.bin`
Error while opening mnist-8.constants.bin�K������k�K������k`��K�����`
```

### DIFF
--- a/src/Runtime/OMExternalConstant.inc
+++ b/src/Runtime/OMExternalConstant.inc
@@ -74,6 +74,9 @@ void omMMapBinaryFile(
   memcpy(tPath, fname, tLen);
   __a2e_s(tPath);
   fname = tPath;
+  fname[tLen] = '\0';
+
+
 #endif
 
   if (constAddr == NULL) {
@@ -168,6 +171,7 @@ void omGetExternalConstantAddr(
     return;
 
   outputAddr[0] = (char *)baseAddr[0] + offset;
+
 }
 
 #endif
--- a/src/Runtime/OMExternalConstant.inc
+++ b/src/Runtime/OMExternalConstant.inc
@@ -75,8 +75,6 @@ void omMMapBinaryFile(
   __a2e_s(tPath);
   fname = tPath;
   fname[tLen] = '\0';
-
-
 #endif
 
   if (constAddr == NULL) {
@@ -171,7 +169,6 @@ void omGetExternalConstantAddr(
     return;
 
   outputAddr[0] = (char *)baseAddr[0] + offset;
-
 }
 
 #endif
--- a/src/Runtime/OMExternalConstant.inc
+++ b/src/Runtime/OMExternalConstant.inc
@@ -64,17 +64,13 @@ void omMMapBinaryFile(
   char *fname = filename;
 #ifdef __MVS__
   // Convert the file name to EBCDIC for the open call.
-  char *tPath;
-  size_t tLen = strlen(fname);
-  tPath = (char *)malloc(tLen);
+  char *tPath = strdup(fname);
   if (!tPath) {
-    fprintf(stderr, "Error while malloc");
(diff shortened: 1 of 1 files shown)


AlexandreEichenberger:

AlexandreEichenberger:
@christopherlmunoz will add you to the list of approved Jenkins folks.

Please abide by the DCO policies, see here https://github.com/onnx/onnx-mlir/blob/main/docs/Workflow.md#step-7-commit--push

AlexandreEichenberger:

src/Runtime/OMExternalConstant.inc
```diff
@@ -74,6 +74,7 @@ void omMMapBinaryFile(
   memcpy(tPath, fname, tLen);
   __a2e_s(tPath);
   fname = tPath;
+  fname[tLen] = '\0';
```
tungld, line 77: Do we need to increase `tLen` by 1 so that the allocated buffer can hold the extra `\0`?

tungld: Looking at the generated code, `filename` already has `\0`. I guess the issue is that `strlen(fname)` does not count `\0` and when we did `memcpy`, `\0` was missed.

Could you try to increase `tLen` by one? e.g.
```
  char *tPath;
  size_t tLen = strlen(fname) + 1;
  tPath = (char *)malloc(tLen);
```

gongsu832: Using `strdup` can simplify the code. We have similar code in `src/Runtime/jni/jniwrapper.c`:
```
#ifdef __MVS__
/* On z/OS, we convert entry point name in ASCII into EBCDIC for
 * the omInputSignature/omOutputSignaturee function using __a2e_s.
 * However, __a2e_s converts in place so we need to make a copy of
 * the entry point name.
 *
 * On z/OS, we convert signature in EBCDIC into ASCII for the JNI
 * NewStringUTF function using __e2a_s. However, __e2a_s converts
 * in place so we need to make a copy of the signature.
 */
#define __a2e(s) conv(s, __a2e_s)
#define __e2a(s) conv(s, __e2a_s)
static inline char *conv(const char *str, size_t (*fptr)(char *)) {
  char *dup = strdup(str);
  if (dup == NULL)
    return NULL;

  /* Convert ASCII to EBCDIC in place */
  if (fptr(dup) < 0) {
    free(dup);
    return NULL;
  }

  return dup;
}
#endif
```

christopherlmunoz:
I used rebase to add a signoff retroactively. I will try to clean up the commits so that I leave only mine.

src/Runtime/OMExternalConstant.inc
christopherlmunoz: I followed this suggestion, thank you.

tungld:

tungld (approved):
LGTM.

PR #2994, opened 2024-10-31
RunONNXModel.py: save compilation info into a file when using --save-model or --cache-model

tungld:
When using `--save-model=save_folder` or `--cache-model=save_folder`, a file `compile.log` is created inside `save_folder` that contains all the compilation log printed out by the compiler.

This is helpful to know how a model was compiled.


AlexandreEichenberger (approved):
LGTM, 
 I was actually thinking that we could just save the content of the command (actually the content of the `-c` argument), and that way we could even "reset" the cache if the options are different.

PR #2996, opened 2024-11-01
run_gpt2_from_huggingface.py: do not download the onnx data file if it exists

tungld:
This patch updates the python script `run_gpt2_from_huggingface.py` so that it does not download `.onnx_data` files if they exist.


AlexandreEichenberger (approved):
LGTM

PR #2997, opened 2024-11-01
Option to not emit the full MLIR (only emit .tmp file)

imaihal:
When emitting MLIR by the option such as `--EmitONNXIR`etc., there are two versions of IR, `<name>.onnx.mlir` and `<name>.tmp` . Since the constant values are embedded in `<name>.onnx.mlir`,  we got memory and disk pressure especially in large models. In mistral-7b model, the MLIR size is about 50GB.
This PR provides an option to not emit full MLIR(`<name>.onnx.mlir`). This option works with emitting MLIR options such as --EmitONNXIR and --EmitMLIR.


src/Compiler/CompilerOptions.cpp
```diff
@@ -281,6 +282,13 @@ static llvm::cl::opt<bool, true> printIROpt("printIR",
     llvm::cl::desc("Print the IR to stdout:."), llvm::cl::location(printIR),
     llvm::cl::init(false), llvm::cl::cat(OnnxMlirOptions));
 
+static llvm::cl::opt<bool, true> emitElidedIROnlyOpt("emitElidedIROnly",
```
tungld, line 285: `emitElidedIROnly` is a bit confusing in the sense that it looks like other `EmitONNXIR`, `EmitMLIR`, ... How about `doNotEmitFullMLIRCode` (`--do-not-emit-full-mlir-code`)?

tungld (approved):
LGTM. It would be an useful option. Thanks!

Could you add a lit test for this option if you don't mind?

src/Compiler/CompilerOptions.cpp
imaihal: Updated the option name. Thanks!

imaihal: Also, added the lit test.

PR #2999, opened 2024-11-05
RunONNXModel.py: allow to change the default model name

tungld:
This patch adds an option `--default-model-name` or `-o` to allow changing the model name used in .so and .constants.bin files.


AlexandreEichenberger (approved):
LGTM, any reason not to just use the universally understood `-o` ?

tungld:
>  any reason not to just use the universally understood -o ?

Sure, I added `-o` as an alternative way. Thanks!

PR #3000, opened 2024-11-05
Release MLIRContext and Module before executing opt command for memory reduction

imaihal:
This PR releases MLIRContext and Module before execution opt command. They are not necessary after writing LLVM bitcode. By releasing the module, the peak memory reduced a bit.

### DIFF
--- a/src/Compiler/CompilerUtils.cpp
+++ b/src/Compiler/CompilerUtils.cpp
@@ -396,8 +396,9 @@ std::string getTargetFilename(
 
 // Write LLVM optimized bitcode.
 // Returns 0 on success, error code on failure.
-static int genLLVMBitcode(const mlir::OwningOpRef<ModuleOp> &module,
-    std::string outputNameNoExt, std::string optimizedBitcodeNameWithExt) {
+static int genLLVMBitcode(mlir::OwningOpRef<ModuleOp> &module,
+    std::string outputNameNoExt, std::string optimizedBitcodeNameWithExt,
+    mlir::MLIRContext &context) {
   std::string msg =
       "Translating MLIR Module to LLVM and Generating LLVM Optimized Bitcode";
   showCompilePhase(msg);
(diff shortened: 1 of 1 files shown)


src/Compiler/CompilerUtils.cpp
```diff
@@ -452,6 +453,11 @@ static int genLLVMBitcode(const mlir::OwningOpRef<ModuleOp> &module,
   llvm::WriteBitcodeToFile(*llvmModule, moduleBitcodeStream);
   moduleBitcodeStream.flush();
 
+  // Free memory before using LLVM `opt` command
+  llvmModule.reset();
+  module.release();
+  context.~MLIRContext();
```
tungld, line 459: Do you know what are freed/released with these statements? It looks not too much memory saved and not sure it is worth to make a change.

src/Compiler/CompilerUtils.cpp
```diff
@@ -598,10 +604,12 @@ static int genJniJar(const mlir::OwningOpRef<ModuleOp> &module,
 }
 
 // Return 0 on success, error code on failure
-static int compileModuleToObject(const mlir::OwningOpRef<ModuleOp> &module,
-    std::string outputNameWithoutExt, std::string &objectNameWithExt) {
+static int compileModuleToObject(mlir::OwningOpRef<ModuleOp> &module,
```
tungld, line 607: Removing `const` and freeing the object internally seems difficult to follow, at least by looking at the function name. Same for other signature changes in other functions.

src/Compiler/CompilerUtils.cpp
```diff
@@ -806,7 +819,7 @@ static int emitOutputFiles(std::string outputNameNoExt,
   case EmitLib: {
     std::string sharedLibNameWithExt;
     int rc = compileModuleToSharedLibrary(
-        module, outputNameNoExt, sharedLibNameWithExt);
+        module, outputNameNoExt, sharedLibNameWithExt, context);
```
tungld, line 822: @imaihal just be noticed that `module` and `context` are potentially used later by `outputCode`. So please make sure `outputCode` still works since `module` has been released inside `compileModuleToSharedLibrary`.

Issue #2993, opened 2024-10-30
ZDNN_MISALIGNED_TENSOR when invoking NNPA on zOS when using a constants binary file. 


christopherlmunoz:
This is on zOS using the --store-constants-binary-file. This issue is specifically on a zdnn_conv2d call. I can provide a full output if needed. 

```
bash-5.2$ client/bin/modelzoo --lib mnist-8-with-binary.so --file mnist-8.tests --validate -m VERBOSE --fc-parms 0.01,0.158752,9,10

ERROR: invoke_nnpa() (./zdnn.c:288): ZDNN_MISALIGNED_TENSOR: Tensor address is not on 4K-byte boundary
```

With ZDNN_ENABLE_PRECHECK enabled, the Kernel and Bias tensors are non 4K aligned addresses.

```
Contents of zdnn_ztensor: kernel
Pre-transformed Descriptor:
Outermost Innermost
Dimensions: 5 5 1 8
Layout: ZDNN_HWCK Format: ZDNN_FORMAT_4DFEATURE
Type: FP16
Transformed Descriptor:
Outermost Innermost
Dimensions: 5 5 1 8
Layout: ZDNN_HWCK Format: ZDNN_FORMAT_4DKERNEL
Type: ZDNN_DLFLOAT16
Buffer Addr: b4df2ffd363083d7 Size: 102400
Transformed: True
```

Thanks! FYI @tungld

tungld:
Created a PR #3002 to make debugging easier.

tehbone:
Now that I look at it, the buffer address looks to be out of sorts, leading to the following questions:
1. Is `__MAP_64` used? If not, the buffer address is incredibly wrong.
2. I suspect that some offset is used to calculate the tensor's buffer address. Is that correct?

tungld:
The starting address of the buffer was wrong and it is fixed by this #3002.

PR #3002, opened 2024-11-08
Test the return value of omMMapBinaryFile function and terminate the main program elegantly

tungld:
Currently the main program still runs even if `omMMapBinaryFile` fails to mmap the constant file, so it's difficult for debugging. 

This patch changes `omMMapBinaryFile` to return a boolean value indicating success and failure, so that the main program can exit elegantly.

This patch also includes a fix for z/OS we just found.


src/Conversion/KrnlToLLVM/KrnlEntryPoint.cpp
```diff
@@ -412,31 +412,6 @@ class KrnlEntryPointOpLowering : public OpRewritePattern<KrnlEntryPointOp> {
...
-                StringRef(errorMsg), rhs, rewriter.getI64Type(), true);
-          else
-            create.krnl.printf(StringRef(errorMsg + "\n"));
-          // Set errno.
-          krnl::emitErrNo(module, rewriter, loc, EINVAL);
-          // Return NULL.
-          create.llvm._return(create.llvm.null(getI8PointerType(context)));
-        });
-  }
-
```
tungld, line 439: Move this code into `KrnlToLLVMHelper.{hpp/cpp}` so that it can be reused.

tungld:
@christopherlmunoz FYI. This is not to fix issues about mmapping a constant file on z/OS, but to make debugging easier.

src/Runtime/OMExternalConstant.inc
```diff
@@ -113,15 +125,20 @@ void omMMapBinaryFile(
...
+      return false;
     }
 
     /* Prepare to compare-and-swap to setup the shared constAddr.
      * If we fail, another thread beat us so free our mmap.
      */
 #ifdef __MVS__
     void *expected = NULL;
-    if (cds((cds_t *)&expected, (cds_t *)&constAddr[0], *(cds_t *)tempAddr))
+    if (cds((cds_t *)&expected, (cds_t *)&constAddr[0], *(cds_t *)&tempAddr))
```
tungld, line 141: Include a fix for z/OS.

src/Runtime/OMExternalConstant.inc
```diff
@@ -67,15 +68,18 @@ void omMMapBinaryFile(
...
     fprintf(stderr, "Error while strdup");
-    return;
+    return false;
   }
   __a2e_s(tPath);
   fname = tPath;
 #endif
 
   if (constAddr == NULL) {
     perror("Error: null pointer");
```
gongsu832, line 78: Instead of mixing `fprintf` and `perror`, either use `fprintf` everywhere but add `strerror(errno)` for more informative printout, or use `perror` everywhere.

src/Runtime/OMExternalConstant.inc
```diff
@@ -89,7 +93,10 @@ void omMMapBinaryFile(
       filePath = (char *)malloc(filePathLen);
```
gongsu832, line 93: You probably want to +1 to `filePathLen` to account for the `\0` you add to the end later.

src/Runtime/OMExternalConstant.inc
```diff
@@ -89,7 +93,10 @@ void omMMapBinaryFile(
...
       if (!filePath) {
         fprintf(stderr, "Error while malloc");
-        return;
+#ifdef __MVS__
+        free(tPath);
+#endif
+        return false;
       }
       memcpy(filePath, basePath, baseLen);
```
gongsu832, line 101: The `memcpy` and adding `\0` can be replaced by a single
```
sprintf(filePath, filePathLen+1, "%s%s%s", basePath, DIR_SEPARATOR, fname);
```

src/Runtime/OMExternalConstant.inc
tungld: Good catch! I updated this.

src/Runtime/OMExternalConstant.inc
tungld: Thanks! I updated by using `snprintf`

src/Runtime/OMExternalConstant.inc
tungld: Sure, now using `fprintf` everywhere.

src/Runtime/OMExternalConstant.inc
```diff
@@ -158,7 +155,9 @@ bool omMMapBinaryFile(
     /* Make sure constAddr is the same as the mmap address.
     */
     if (constAddr[0] != tempAddr) {
```
gongsu832, line 157: This check is wrong. Only the first thread that successfully performs the compare-and-swap will have `constAddr[0] == tempAddr`. All other threads will have `constAddr[0] != tempAddr`. But it doesn't matter since by design they will throw away their `tempAddr` and use the `constAddr[0]` set by the first thread.

src/Runtime/OMExternalConstant.inc
```diff
@@ -181,11 +180,11 @@ bool omMMapBinaryFile(
 void omGetExternalConstantAddr(
     void **outputAddr, void **baseAddr, int64_t offset) {
   if (outputAddr == NULL) {
-    perror("Error: null pointer");
+    fprintf(stderr, "Error: null pointer: %s", strerror(errno));
```
gongsu832, line 183: `strerror(errno)` is meaningless here.

src/Runtime/OMExternalConstant.inc
```diff
@@ -181,11 +180,11 @@ bool omMMapBinaryFile(
...
     return;
   }
   if (baseAddr == NULL) {
-    perror("Error: null pointer");
+    fprintf(stderr, "Error: null pointer: %s", strerror(errno));
```
gongsu832, line 187: `strerror(errno)` is meaningless here.

src/Runtime/OMExternalConstant.inc
```diff
@@ -43,7 +43,7 @@ const int i = 1;
 void checkEndianness(const char constPackIsLE) {
   if (XOR(IS_SYSTEM_LE(), constPackIsLE)) {
     fprintf(stderr, "Constant pack is stored in a byte order that is not "
-                    "native to this current system.");
+                    "native to this current system: %s", strerror(errno));
```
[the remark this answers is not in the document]
tungld: Same thought. Will remove them.

src/Runtime/OMExternalConstant.inc
tungld: Does `tempAddr` become `NULL` after `munmap`? If so, we do the test only when `tempAddr != NULL`.

tungld: I explicitly set `tempAddr = NULL` after `munmap`, and check `constAddr[0] != tempAddr` only when `tempAddr != NULL` (meaning for the first successful thread).

gongsu832: Not sure what you mean. `munmap` cannot change `tempAddr` since C is passing by value (i.e., inside `munmap` it only has access to a copy of `tempAddr`). But even if `tempAddr` somehow becomes NULL after `munmap`, the check `constAddr[0] != tempAddr` is still wrong.

gongsu832: > I explicitly set `tempAddr = NULL` after `munmap`, and check `constAddr[0] != tempAddr` only when `tempAddr != NULL` (meaning for the first successful thread).

Not sure why you want to do that. A successful compare-and-swap guarantees that `constAddr[0] == tempAddr`.

tungld: Oh, you are right. In the latest code, I set `tempAddr = NULL` for failed threads, and check `if (tempAddr && (constAddr[0] != tempAddr))`. What do you think?

tungld: I want to avoid the situation we encountered where `cds` was written wrongly. Perhaps, `assert` is better.

tungld: OK. I removed the check. I was overthinking about having a check to debug, but since we found the issue, and `cds` would work correctly.

gongsu832: Yes you can use `assert` but it wouldn't help in this case. Because even if you write your `cds` wrong like the bug we just had, in the first thread after the `cds` you will have `constAddr[0] == tempAddr`. Because `cds` doesn't know/care if the `tempAddr` you supplied in the call is garbage. All it does is setting `constAddr[0]` to `tempAddr` atomically, regardless of what's in `tempAddr`.

tungld:
@gongsu832 @AlexandreEichenberger could you take another look at this PR? Apart from handling return value of `omMMapBinaryFile`, it contains a bug fix for the issue https://github.com/onnx/onnx-mlir/issues/2993. Thanks!

src/Conversion/KrnlToLLVM/KrnlToLLVMHelper.cpp
```diff
@@ -342,5 +343,47 @@ bool isZOS(ModuleOp module) {
...
+  MultiDialectBuilder<LLVMBuilder, KrnlBuilder> create(rewriter, loc);
+  create.llvm.ifThenElse(/*cond=*/
+      [&](const LLVMBuilder &createLLVM) {
+        return createLLVM.icmp(LLVM::ICmpPredicate::ne, lhs, rhs);
+      }, /*then=*/
+      [&](const LLVMBuilder &createLLVM) {
+        MultiDialectBuilder<LLVMBuilder, KrnlBuilder> create(createLLVM);
+        // Print an error message.
+        if (!errorMsg.empty())
+          create.krnl.printf(StringRef(errorMsg + "\n"));
```
AlexandreEichenberger, line 382: If we start to use the `krnl.printf` for more than debugging, I have been often experiencing issues with that call (in debugging situations) where it print too long a string. We have some issues with that call. 

I had similar issue in the `krnl.PrintTensor` and I have alleviated this by having a special `%e` sequence at the end of the string. For Instrumentation, I think I encoded the string length in an integer flag. All tricks to get around this issue. If someone could try to understand what is happening, it would be great!

src/Runtime/OMExternalConstant.inc
```diff
@@ -153,11 +172,11 @@ void omMMapBinaryFile(
 void omGetExternalConstantAddr(
     void **outputAddr, void **baseAddr, int64_t offset) {
   if (outputAddr == NULL) {
-    perror("Error: null pointer");
+    fprintf(stderr, "Error: null pointer.");
     return;
   }
   if (baseAddr == NULL) {
-    perror("Error: null pointer");
+    fprintf(stderr, "Error: null pointer.");
```
AlexandreEichenberger, line 179: No need to track these error by returning `false` or `true`?

AlexandreEichenberger (approved):
LGTM, please have @gongsu832 provide the feedback specific to the mmap part.

src/Conversion/KrnlToLLVM/KrnlToLLVMHelper.cpp
```diff
@@ -342,5 +343,47 @@ bool isZOS(ModuleOp module) {
   return zOS;
 }
 
+void equalOrFailed(ModuleOp &module, OpBuilder &rewriter, Location loc,
```
AlexandreEichenberger, line 346: Would these calls also be potentially useful if we wanted to fail the inference, say after a NNPA severe failure call?
If so, we may want to see if we would benefit from introducing something at a higher level dialects (say Krnl or SCF) that would then feed into this one.

tungld: Totally yes. It will be used anywhere in the main inference when we want to fail the inference. We can introduce a krnl op for this.

src/Runtime/OMExternalConstant.inc
tungld: I can be here, but I am a bit hesitated since this is called per each constant so there would have a lot of `if then else` generated in the main inference for checking the return value of this function.

src/Runtime/OMExternalConstant.inc
```diff
@@ -58,24 +59,27 @@ void checkEndianness(const char constPackIsLE) {
...
   fname = tPath;
 #endif
 
   if (constAddr == NULL) {
-    perror("Error: null pointer");
-    return;
+    fprintf(stderr, "Error: null pointer.");
+#ifdef __MVS__
+    free(tPath);
+#endif
```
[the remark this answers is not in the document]
tungld: @gongsu832 all of these issues come from the fact that `fname` is not in EBCDIC. So I modify the code generation to convert `fname` to EBCDIC during compilation so that `fname` here would be alread in EBCDIC, then there is no need to handle it here anymore.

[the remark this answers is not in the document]
gongsu832: OK that makes things simpler.

PR #3006, opened 2024-11-12
Delay scrubbing disposable elements attrs as long as possible

tungld:
Delay scrubbing disposable elements attrs as long as possible in order to reduce compilation memory consumption.


src/Dialect/ONNX/ONNXOps/Canonicalize.cpp
```diff
@@ -202,11 +202,7 @@ bool haveSameStaticShape(Value lhs, Value rhs) {
 
 /// Test if the input is a splat constant with a negative value or not.
 bool isNegativeSplatConstant(Value val) {
-  if (!isDenseONNXConstant(val))
-    return false;
-  ONNXConstantOp constOp = val.getDefiningOp<ONNXConstantOp>();
-  auto valAttr =
-      llvm::dyn_cast_or_null<DenseElementsAttr>(constOp.getValueAttr());
+  ElementsAttr valAttr = getElementAttributeFromONNXValue(val);
```
tungld, line 205: Update rules here to make them work with `DisposableElementsAttr` also.

imaihal (approved):
LGTM! 
I confirmed memory consumption is reduced in the model.

PR #3011, opened 2024-11-14
Update llvm-project to llvm/llvm-project@01d233ff403823389f848

hamptonm1:
LLVM: `01d233ff`
StableHLO: `37487a8e`

### DIFF
--- a/docs/BuildOnLinuxOSX.md
+++ b/docs/BuildOnLinuxOSX.md
@@ -15,7 +15,7 @@ Firstly, install MLIR (as a part of LLVM-Project):
 ``` bash
 git clone -n https://github.com/llvm/llvm-project.git
 # Check out a specific branch that is known to work with ONNX-MLIR.
-cd llvm-project && git checkout 00128a20eec27246719d73ba427bf821883b00b4 && cd ..
+cd llvm-project && git checkout 01d233ff403823389f8480897e41aea84ecbb3d3 && cd ..
 ```
 
 [same-as-file]: <> (utils/build-mlir.sh)
--- a/docs/BuildOnWindows.md
+++ b/docs/BuildOnWindows.md
@@ -52,7 +52,7 @@ Install MLIR (as a part of LLVM-Project):
 ```shell
 git clone -n https://github.com/llvm/llvm-project.git
 # Check out a specific branch that is known to work with ONNX-MLIR.
-cd llvm-project && git checkout 00128a20eec27246719d73ba427bf821883b00b4 && cd ..
+cd llvm-project && git checkout 01d233ff403823389f8480897e41aea84ecbb3d3 && cd ..
 ```
 
 [same-as-file]: <> (utils/build-mlir.cmd)
--- a/src/Conversion/KrnlToAffine/ConvertKrnlToAffine.cpp
+++ b/src/Conversion/KrnlToAffine/ConvertKrnlToAffine.cpp
@@ -815,9 +815,9 @@ AffineTypeConverter::AffineTypeConverter() {
 
   addSourceMaterialization([&](OpBuilder &builder, Type resultType,
                                ValueRange inputs,
-                               Location loc) -> std::optional<Value> {
+                               Location loc) -> Value {
     if (inputs.size() != 1)
-      return std::nullopt;
+      return Value();
 
     return builder.create<UnrealizedConversionCastOp>(loc, resultType, inputs)
         .getResult(0);
@@ -825,9 +825,9 @@ AffineTypeConverter::AffineTypeConverter() {
 
   addTargetMaterialization([&](OpBuilder &builder, Type resultType,
                                ValueRange inputs,
-                               Location loc) -> std::optional<Value> {
+                               Location loc) -> Value {
     if (inputs.size() != 1)
-      return std::nullopt;
+      return Value();
 
     return builder.create<UnrealizedConversionCastOp>(loc, resultType, inputs)
         .getResult(0);
--- a/src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
+++ b/src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
@@ -548,9 +548,9 @@ KrnlTypeConverter::KrnlTypeConverter() {
 
   addSourceMaterialization([&](OpBuilder &builder, Type resultType,
                                ValueRange inputs,
-                               Location loc) -> std::optional<Value> {
+                               Location loc) -> Value {
     if (inputs.size() != 1)
-      return std::nullopt;
+      return Value();
 
     return builder.create<UnrealizedConversionCastOp>(loc, resultType, inputs)
         .getResult(0);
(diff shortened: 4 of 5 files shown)


src/Dialect/Mlir/DialectBuilder.cpp
```diff
@@ -695,7 +697,7 @@ TypedAttr MathBuilder::negativeInfAttr(Type type) const {
         default:
           llvm_unreachable("unsupported element type");
         }
-        attr = b().getIntegerAttr(type, APInt(width, value));
+        attr = b().getIntegerAttr(type, APInt(width, value, false, true));
```
hamptonm1, line 700: I had to update the `APInt` with the following boolean values based on the this PR update: https://github.com/llvm/llvm-project/pull/114539

Previously truncation was set to true as default and they changed it to false, we need the truncation on for our lit test to pass and to be consistent with the previous behavior.

src/Dialect/Mlir/DialectBuilder.cpp
```diff
@@ -627,12 +627,14 @@ Value MathBuilder::constant(Type type, double val) const {
           // If unsigned, create a signless constant, then cast it to unsigned.
           if (elementType.isUnsignedInteger()) {
             Type signlessTy = b().getIntegerType(width);
-            constant = b().create<arith::ConstantOp>(loc(),
-                b().getIntegerAttr(signlessTy, APInt(width, (int64_t)val)));
+            constant = b().create<arith::ConstantOp>(
+                loc(), b().getIntegerAttr(signlessTy,
+                           APInt(width, (int64_t)val, false, true)));
```
mikeessen, line 632: Please change to c++ style cast to prevent code scan finding.

static_cast<int64_t>(val) instead of (int64_t)val

Also line 637.

src/Dialect/Mlir/DialectBuilder.cpp
```diff
@@ -2263,7 +2265,8 @@ Value LLVMBuilder::constant(Type type, int64_t val) const {
           assert(type.isSignless() &&
                  "LLVM::ConstantOp requires a signless type.");
           constant = b().create<LLVM::ConstantOp>(loc(), type,
-              b().getIntegerAttr(type, APInt(width, (int64_t)val)));
+              b().getIntegerAttr(
+                  type, APInt(width, (int64_t)val, false, true)));
```
mikeessen, line 2269: Same comment as previous file.  Please change to c++ style cast.

src/Dialect/Mlir/DialectBuilder.cpp
AlexandreEichenberger: @mikeessen Thanks for keeping track of the casts. Could you look into seeing if there is away to have the compiler issue a warning (probably clang for us since that what we are using)? I am afraid that I may not be as vigilant as we should be on this issue (speaking for me on the code I modify...).

That would be greatly appreciated. Thanks

mikeessen: > @mikeessen Thanks for keeping track of the casts. Could you look into seeing if there is away to have the compiler issue a warning (probably clang for us since that what we are using)? I am afraid that I may not be as vigilant as we should be on this issue (speaking for me on the code I modify...).
> 
> That would be greatly appreciated. Thanks

Thanks, I will take a look.

hamptonm1:
@AlexandreEichenberger So I went from 5 lit tests failing to now 2. The 2 tests ( `onnx-mlir/test/mlir/parallel/krnl_parallel_clause_to_omp.mlir` and `onnx-mlir/test/mlir/accelerators/nnpa/transform/zlow-stick-unstick-expansion.mlir`) that are failing have to do with openmp which I know you are the expert at. After doing some research, it seems like `LLVM` updated their checks for openmp and as a result the lit tests are failing. 

Here is the update:
https://github.com/llvm/llvm-project/pull/112229

Basically if we have more than one nested omp loop, it emits an error now. So that is why the lit tests are failing. I removed the tests for now but can you do me a favor and take a look. I understand the problem but not sure how to reformat the tests so that they can oblige to the new LLVM rules but also still pass :) 

errors:

```
within split at /workdir/onnx-mlir/test/mlir/parallel/krnl_parallel_clause_to_omp.mlir:3 offset :22:5: error: 'omp.wsloop' op loop wrapper does not contain exactly one nested op
    omp.wsloop {
    ^
```
```

within split at /workdir/onnx-mlir/test/mlir/accelerators/nnpa/transform/zlow-stick-unstick-expansion.mlir:3 offset :7:3: error: expected result type with offset = dynamic instead of 0
  "zlow.stick"(%arg0, %alloc) {layout = "3DS", saturation = -1 : si64} : (memref<16x8x128xf32>, memref<16x8x128xf16, #map>) -> ()
  ^
```

hamptonm1:
@AlexandreEichenberger also for s390x, 3 lit tests are failing with one of the same error messages I copied above:

`offset :7:3: error: expected result type with offset = dynamic instead of 0`

```
Failed Tests (3):
  Open Neural Network Frontend :: accelerators/nnpa/driver/matmul-div-in-attention-layer.mlir
  Open Neural Network Frontend :: accelerators/nnpa/driver/saturation.mlir
  Open Neural Network Frontend :: accelerators/nnpa/module_op_be/compiler-config.mlir

```

AlexandreEichenberger:
@hamptonm1  The CIs succeed for MacOS, linux on x86, POWER...: does it means that the errors reported for OPenMP:

> @AlexandreEichenberger So I went from 5 lit tests failing to now 2. The 2 tests ( onnx-mlir/test/mlir/parallel/krnl_parallel_clause_to_omp.mlir and onnx-mlir/test/mlir/accelerators/nnpa/transform/zlow-stick-unstick-expansion.mlir) that are failing 

fails only on s390x?

hamptonm1:
> @hamptonm1 The CIs succeed for MacOS, linux on x86, POWER...: does it means that the errors reported for OPenMP:
> 
> > @AlexandreEichenberger So I went from 5 lit tests failing to now 2. The 2 tests ( onnx-mlir/test/mlir/parallel/krnl_parallel_clause_to_omp.mlir and onnx-mlir/test/mlir/accelerators/nnpa/transform/zlow-stick-unstick-expansion.mlir) that are failing
> 
> fails only on s390x?

Yeah I think only s390x for now.

hamptonm1:
@AlexandreEichenberger One of the errors I am seeing is:

```
within split at /Users/runner/work/onnx-mlir/onnx-mlir/test/mlir/parallel/krnl_parallel_clause_to_omp.mlir:134 offset :22:5: error: 'omp.wsloop' op loop wrapper does not contain exactly one nested op
    omp.wsloop {
```

AlexandreEichenberger:
@hamptonm1 ok, will look at the PR this week.

hamptonm1:
> @hamptonm1 ok, will look at the PR this week.

@AlexandreEichenberger Thank you and thank you for your patience with me!

hamptonm1:
@AlexandreEichenberger Thanks for the fix! Can you do the honors of merging the PR for me? Thank you!

AlexandreEichenberger (approved):
LGTM, thanks for working on this @hamptonm1 , much appreciated

PR #3012, opened 2024-11-15
Fix a wrong function call

tungld:
This is to fix a wrong function call that converts an ASIIC string to EBCDIC.


AlexandreEichenberger (approved):
LGTM

PR #3020, opened 2024-11-27
Re-enable diagnostic error/warning printing

AlexandreEichenberger:
In the past, `emitError(...)` printed messages to stderr, but now we have to install a diagnostic error handler. So I added the necessary code, as otherwise we do not get error messages anymore.

Also, in onnx-mlir diagnostics, we issued an assert when a range was badly formed (`max<min`), but since that can be the reason for an `emitError` to begin with, it also prevented an error message to be printed. So now the range can be badly formed, we can test for it, and just adding an extra error/warning message.


src/Dialect/ONNX/ONNXOps/Canonicalize.cpp
```diff
@@ -743,7 +743,7 @@ class LoopOpRewriteMaxTripCountPattern : public OpRewritePattern<ONNXLoopOp> {
   bool isDefinedByIntegerConstantOp(Value v) const {
     if (mlir::isa<BlockArgument>(v))
       return false;
-    Operation *definingOp = v.getDefiningOp();
+    // Operation *definingOp = v.getDefiningOp();
```
AlexandreEichenberger, line 746: was unused.

AlexandreEichenberger:
Note we get a warning now when emitting LLVM because some of our functions have an non-MLIR approved / recognized attribute. The warning is
```
Unhandled parameter attribute 'onnx.name'
```
I verified that it is in the function
```
llvm.func @main_graph_test_add(%arg0: !llvm.ptr, %arg1: !llvm.ptr, %arg2: i64, %arg3: i64, %arg4: i64, %arg5: i64, %arg6: i64, %arg7: i64, %arg8: i64, %arg9: !llvm.ptr, %arg10: !llvm.ptr, %arg11: i64, %arg12: i64, %arg13: i64, %arg14: i64, %arg15: i64, %arg16: i64, %arg17: i64) -> (!llvm.struct<(ptr, ptr, i64, array<3 x i64>, array<3 x i64>)> {onnx.name = "sum"}) attributes {llvm.emit_c_interface} {
...
```

Since its probably ok, I currently disable the diagnostic reporting in the LLVM lowering unless the `-v` (Verbose) mode is requested.

src/Dialect/ONNX/ONNXOps/Canonicalize.cpp
tungld: Why don't you remove it completely?

tungld (approved):
LGTM.

Issue #1059, opened 2021-12-13
Default make need onnx to be installed


AlexandreEichenberger:
Our default build listed in the README.txt (https://github.com/onnx/onnx-mlir#installation-on-unix) is as follows:

```
cmake --build .
```

and this attempts to build stuff for running the `check-onnx-backend`, which requires onnx. To the best of my understanding, building onnx requires the procedure listed here (https://github.com/onnx/onnx-mlir#installing-third_party-onnx-for-backend-tests-or-rebuilding-onnx-operations).

I don't understand why our CIs don't appear to install onnx from our third_party submodule. 

From our discussions today, our default make (listed in the README instructions) should build whatever is needed for an end to end test, namely onnx-mlir. Building also onnx-mlir-opt and running the simple lit tests are probably fine (I am for it)

gongsu832:
The CI does install third_party/onnx. You can see in Dockerfile.onnx-mlir and Dockerfile.onnx-mlir-dev:
```
RUN ONNX_ROOT=${WORK_DIR}/onnx-mlir/third_party/onnx \
    && cd ${ONNX_ROOT} \
    && python3 setup.py -q install \
    && rm -rf .eggs .setuptools-cmake-build build dist onnx.egg-info
```
And the default make does produce all the binaries such as onnx-mlir and onnx-mlir-opt.

AlexandreEichenberger:
@gongsu832 , thanks, that explain the mystery (was grepping for `pip install`).

On the second point, could I ask you to look into separating the default make (gen of onnx-mlir and onnx-mlir-opt + lit tests that work without onnx) for our default users, plus a more complete target for our CIs and developers? That would be great. If that make could also install onnx, that would be the cherry on top... continuing as is with a bit of explanation in the "developer" section is ok too.

gongsu832:
@AlexandreEichenberger A seemingly simple thing like "install onnx" can turn into a nightmare simply because all the dependencies involved and there are unlimited number of host configurations that this can go run. There is simply no way to write some simple target or script and expect it to work in all these configurations. That's why we have the `onnx-mlir-dev` image for doing dev work.

And on the issue of the build system, I really don't quite understand why we keep trying to run tests in the default make target. Even MLIR doesn't do that. I understand the desire to make the system as easy to use as possible. But let's not burden the base build system for that. The base build system is for building onnx-mlir, not for teaching you how to use onnx-mlir. You can do that with scripts and docs.

[...]

PR #3021, opened 2024-12-02
Disabled building tests default (like LLVM), enabled for our CIs

AlexandreEichenberger:
As mentioned in #1059, our default builds require onnx to be installed. 

This PR implements the changes requested in that issue:
1) all extensive tests are predicated by `ONNX_MLIR_BUILD_TESTS` being `ON`
2) default `ONNX_MLIR_BUILD_TESTS` value switched to off.
3) Our CIs add the `ONNX_MLIR_BUILD_TESTS=ON` define.

That way, folks just building a default `onnx-mlir` will not build the tests, and thus we are not dependent on `onnx` third party to be installed.

For folks using their own scripts for testing, you must now add

```
        -DONNX_MLIR_BUILD_TESTS=ON \
```

to your `cmake` command.

### DIFF
--- a/docs/doc_check/CMakeLists.txt
+++ b/docs/doc_check/CMakeLists.txt
@@ -1,5 +1,9 @@
 # SPDX-License-Identifier: Apache-2.0
 
+if (NOT ONNX_MLIR_BUILD_TESTS)
+  set(EXCLUDE_FROM_ALL ON)
+endif()
+
 set(check_cmd ${Python3_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/check.py)
 
 add_custom_target(check-doc
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,8 @@ cmake_minimum_required(VERSION 3.20.0)
 
 project(onnx-mlir)
 
-option(ONNX_MLIR_BUILD_TESTS "Build ONNX-MLIR test executables. If OFF, just generate build targets." ON)
+# Like in LLVM, default tests are off. Lit-tests are always on.
+option(ONNX_MLIR_BUILD_TESTS "Build ONNX-MLIR test executables. If OFF, just generate build targets." OFF)
 option(ONNX_MLIR_CCACHE_BUILD "Set to ON for a ccache enabled build." OFF)
 option(ONNX_MLIR_ENABLE_STABLEHLO "Enable StableHLO support." ON)
 option(ONNX_MLIR_ENABLE_WERROR "Enable warnings as errors." OFF)
--- a/docker/Dockerfile.onnx-mlir
+++ b/docker/Dockerfile.onnx-mlir
@@ -43,6 +43,7 @@ RUN LLVM_PROJECT_ROOT=${WORK_DIR}/llvm-project \
     && CC=clang CXX=clang++ \
        cmake -DMLIR_DIR=${LLVM_PROJECT_ROOT}/build/lib/cmake/mlir \
              -DCMAKE_BUILD_TYPE=Release \
+             -DONNX_MLIR_BUILD_TESTS=ON \
              -DLLVM_ENABLE_ASSERTIONS=ON \
              -DCMAKE_INSTALL_MESSAGE=NEVER \
              -DONNX_MLIR_ACCELERATORS=${ACCEL} .. \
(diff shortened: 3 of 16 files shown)


AlexandreEichenberger:
From the log from linux/amd
```
[...]
[100%] Built target numerical
[100%] Running the ONNX-MLIR numerical regression tests
      Start  1: TestConv
      Start  2: TestMatMul2D
      Start  3: TestMatMulBroadcast
      Start  4: TestGemm
      Start  5: TestLSTM
      Start  6: TestRNN
      Start  7: TestGRU
      Start  8: TestLoop
      Start  9: TestScan
      Start 10: TestElementwise
      Start 11: TestSoftplus


[...]

[100%] Built target check-onnx-backend
bringing up nodes...
bringing up nodes...

530 passed, 2348 skipped in 103.80s (0:01:43)
```

So it appears to build & test the tests in the CI.

tungld:
@AlexandreEichenberger I am curious to know if without building onnx, can you compile a .onnx model or not? Numerical tests do not use .onnx files as input, so it does not tell whether we can compile a .onnx model. Just to double check.

docs/doc_check/CMakeLists.txt
```diff
@@ -1,5 +1,9 @@
 # SPDX-License-Identifier: Apache-2.0
 
+if (NOT ONNX_MLIR_BUILD_TESTS)
+  set(EXCLUDE_FROM_ALL ON)
+endif()
```
gongsu832, line 5: I don't think this does anything. `EXCLUDE_FROM_ALL` is a target property. Setting a variable called `EXCLUDE_FROM_ALL` won't do anything. The proper use of `EXCLUDE_FROM_ALL` should be like the code in `MLIR.cmake`:
```
  if (EXCLUDE_FROM_ALL)
    add_executable(${name} EXCLUDE_FROM_ALL ${ARG_UNPARSED_ARGUMENTS})
  else()
    add_executable(${name} ${ARG_UNPARSED_ARGUMENTS})
  endif()
```
The variable `EXCLUDE_FROM_ALL` is used to decide whether to add the property `EXCLUDE_FROM_ALL` to a target.

AlexandreEichenberger: @gongsu832 I reused this pattern from the current `onnx-mlir`. I know it does not make it right, but on my Mac it just seems to work. When I set the variable as `ON`, it builds all the tests as before. When I set it to off, the resulting build does not recognize the `make check-onyx-backend` as a valid target.

Let me run the experiment on LoZ too.

AlexandreEichenberger:
> I am curious to know if without building onnx, can you compile a .onnx model or not? Numerical tests do not use .onnx files as input, so it does not tell whether we can compile a .onnx model. Just to double check.

@tungld Here is the experiment I ran. I executed a `pip3 uninstall onnx` until it said there were none; then I ran with the `-DONNX_MLIR_BUILD_TESTS=OFF`, build an `onnx-mlir` and tried the compiler on an mlir and onnx input, both of which worked. When I built it with `ON`, the make process would fail due to the lack of installed `onnx`.

That was on a Mac, let me retry all this on a LoZ

docs/doc_check/CMakeLists.txt
gongsu832: I assume you meant `check-onnx-backend` since there is no `check-onyx-backend`.

Regardless of how the variable is set, `check-onnx-backend` should always be a valid target. The difference is whether it's included in the default `all` target.

tungld:
>  I ran with the -DONNX_MLIR_BUILD_TESTS=OFF, build an onnx-mlir and tried the compiler on an mlir and onnx input, both of which worked

Thank you for the confirmation! That sounds good. Let @gongsu832 give the final decision since the most changes are about CMakeFiles.

docs/doc_check/CMakeLists.txt
gongsu832: Also note that `add_executable`, `add_library`, etc. behave differently from `add_custom_target`. By default, `add_executable`, `add_library`, etc. will add the target to `all` unless you specify `EXCLUDE_FROM_ALL` on the target. `add_custom_target`, on the other hand, by default will not add the target to `all` unless you specify `ALL` on the target.

AlexandreEichenberger: @gongsu832 Are there specific steps you would like me to take? Or if it takes you less time, do you want to take over this PR to do it the right way (if the current approach is not the preferred approach).

gongsu832: I can do it but probably won't be able to get to it until some time next week.

AlexandreEichenberger: Thanks, there is no rush, glad it will be done using the proper approach

PR #3023, opened 2024-12-05
Return a failure instead of crashing if shape inference can not be run because of unraked operand types

jorickert:
Some  `computeShape` functions assume/rely on the operands being ranked.
This PR adds a check to them, so that they return `failure` in case this assumption does not hold, instead of asserting. This allows callers of this function to handle this failure

### DIFF
--- a/src/Dialect/ONNX/ONNXOps/Math/DFT.cpp
+++ b/src/Dialect/ONNX/ONNXOps/Math/DFT.cpp
@@ -33,6 +33,9 @@ LogicalResult ONNXGenericDFTOpShapeHelper<OP_TYPE>::customComputeShape(
   // Get info about input data operand.
   Value input = operandAdaptor.getInput();
   // Get the rank to compensate for N dimensions.
+  if (!hasShapeAndRank(input)) {
+    return failure();
+  }
   int64_t rank = createIE->getShapedTypeRank(input);
 
   // Check if the dimension for axis is a literal and in range.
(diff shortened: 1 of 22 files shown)


AlexandreEichenberger:

AlexandreEichenberger (approved):
LGTM,

I verified a few, let me just ask just to be sure. The `return failure()` that are added here are all in the `computeShape(...)` functions, which is fine since we expect to have successfully run shape inference before, were potentially unranked values have been at least ranked.

Thanks for adding "graceful" errors, much appreciated.

jorickert:
@AlexandreEichenberger Yes, the `return failure()` are all added in `computeShape` . This allows callers of this function to decide how to handle failures in case `computeShape` fails because the input is unranked. 

 Before this PR, the functions `computeShapeAndAssertOnFailure` and `computeShape` would behave the same in many cases, as `computeShape` would assert internally.
 
 It would be nice if you could merge this, as I do not have merge rights

AlexandreEichenberger:
Will do, for some reason I don't have the usual button to update to the latest. do you mind doing this on your end, and then I will merge. I asked for you to have access right to run the CIs. After a PR or two, will give you merge permission too.

jorickert:
@AlexandreEichenberger I rebased onto the latest main

jorickert:
> **Jenkins Linux amd64** Build [#16064](https://www.onnxmlir.xyz/jenkinx/job/ONNX-MLIR-Pipeline-Docker-Build/16064/) [push] Return a failure instead... failed after 53 min

@AlexandreEichenberger 
The CI error looks like some issue with the CI setup itself to me:
`s = '[0.002s][warning][perf,memops] Cannot use file /tmp/hsperfdata_root/232111 because it is locked by another process (e...QAAAAABAAAAAQEAAAABAQEBAAABAAABAQABAAEAAAEBAAEBAQABAQABAAAAAQEBAQEAAAABAQEBAQEAAQAAAAAAAQEAAAEAAAABAQAAAQEAAQEBAAEA"}]'
`

PR #3024, opened 2024-12-05
Add tolerance args to CheckONNXModel.py

AlexandreEichenberger:
Added `--rtol=x` and `--atol=y` to CheckONNXModel.py.

So for example, if you wanted to check the accuracy of a code running on CPU vs accelerator, you would launch the following command
```
CheckONNXModel.py -m <model> -r="-O3 -march=z16" -a="-maccel=NNPA" -atol=0.001 -rtol=0.005
```
where `-m` is to give the model name, `-r` is the reference compile parameters, `-a` is the additional parameters to be tested for the test compilation (you can use the `-t` to give an entirely different compiler options), and the last 2 parameters are used to pass the absolute and relative tolerance used to verify the equivalence of the reference-compiled-run vs test-compiled-run.


hamptonm1 (approved):
LGTM!

PR #3030, opened 2024-12-16
Fix biasScaleShape of GroupNormalizationV21 to support ranks > 4

jorickert:
Before this PR the `oneDimShape` assumed a spacial rank of two, which is only correct for rank==4.

### DIFF
--- a/test/mlir/onnx/onnx_decompose.mlir
+++ b/test/mlir/onnx/onnx_decompose.mlir
@@ -565,17 +565,17 @@ func.func @test_groupnorm_v18(%arg0: tensor<3x4x2x2xf32>, %arg1: tensor<2xf32>, 
 }
 // -----
 
-func.func @test_groupnorm_v21(%arg0: tensor<3x4x2x2xf32>, %arg1: tensor<2xf32>, %arg2: tensor<2xf32>) -> tensor<3x4x2x2xf32> {
-  %0 = "onnx.GroupNormalization"(%arg0, %arg1, %arg2) {epsilon = 0.00999999977 : f32, num_groups = 2 : si64} : (tensor<3x4x2x2xf32>, tensor<2xf32>, tensor<2xf32>) -> tensor<3x4x2x2xf32>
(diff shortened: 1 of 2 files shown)


hamptonm1:
@jorickert Hello!! Can you direct me to the link or document which validates your findings please? I just want to have a better understanding. Thank you!

jorickert:
@hamptonm1 
According to the onnx spec the input for GroupNormV21 has the dimensions `(N, C, D1,D2, Dn)`, scale and bias have the dimension `C` 

The formula for GroupNorm is:
`y = scale * (x - mean) / sqrt(variance + epsilon) + bias`  `where the mean and variance are computed per instance per group of channels`. The formula for LayerNorm is generally the same, the main difference (in onnx) being that it allows the selection of axes for the mean and variance.

Internally, the GroupNorm reshapes the input to `(N, G, C // G, D1, D2, Dn)` and then performs a LayerNorm like operation on it, with the reduction axes being `C // G to Dn`. This can be seen in the GroupNorm paper https://arxiv.org/pdf/1803.08494 figure 3.

The decomposition/conversion mimics this:

1. Manually reshape the input to `(N, G, C // G, D1, D2, Dn)`
2. Perform a LayerNorm with axes= `C // G to Dn`
3. Manually reshape to  `(N, C, D1,D2, Dn)`
Additionally, it is required to reshape the scale and bias. The input scale and bias have shape `C`. 
To make them compatible with the LayerNorm and broadcasting, the following reshape needs to be done:
`C` -> `(G, C // G, 1, 1, 1)` where the number of once equals the number of spacial dimensions.
Without my PR, the reshape is to (G, C // G, 1, 1), no matter how many spacial dimensions exist. 

While writing this down, I realized, that the lit tests are still wrong for GroupNormV21. The input scale and bias has the size `C//G = 2` , but it should be `C = 4` . I will fix this

AlexandreEichenberger (approved):
LGTM, let me know if you need me to merge it (and if so, when you are ready to do so),

test/mlir/onnx/onnx_decompose.mlir
jorickert, line 1: Reminder to update the bias and scale size in the lit test inputs

hamptonm1 (approved):
Okay it works for me then!

jorickert:
@AlexandreEichenberger It would be nice if you could merge this. Thank you

PR #3040, opened 2025-01-10
Update PyExecutionSessionBase.cpp to get right result

fairytale0828:
When I use pyruntime, the output tensor always contains the same element. I think this is because the pointer is not modified by the stride when getting the address. The pointer should move by the stride each time an element is read.

### DIFF
--- a/src/Runtime/python/PyExecutionSessionBase.cpp
+++ b/src/Runtime/python/PyExecutionSessionBase.cpp
@@ -232,52 +232,113 @@ std::vector<py::array> PyExecutionSessionBase::pyRun(
 
     // https://numpy.org/devdocs/user/basics.types.html
     py::dtype dtype;
+    // switch (omTensorGetDataType(omt)) {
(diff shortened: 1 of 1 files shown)


fairytale0828:
微信截图_20250110134825

chentong319:
When the computed output result with type OMTensor is converted to py::array in PyExecutionSessionBase.cpp, the stride is not used. 
```
    outputPyArrays.emplace_back(
        py::array(dtype, shape, omTensorGetDataPtr(omt)));
```
There are two ways to identify whether the error is in PyExecutionSeesionBase.cpp, or somewhere else:
1. Add omTensorPrint to print the generated OMTensor.
2. Use the c driver, instead of the python driver. You can find the instructions at the beginning of file utils/RunONNXLib.cpp.
I copy paste the instructions here: (replace test/backend/test_add/test_add.so with your .so)
```
cd onnx-mlir/build
sh ../utils/build-run-onnx-lib.sh test/backend/test_add/test_add.so

  On linux you run with:

Debug/bin/run-onnx-lib
```

PR #3043, opened 2025-01-15
[NNPA] Revise compiler options for quantization

tungld:
This PR changes compiler options for quantization so that they are more flexible in the future.

New compiler options for quantization are: `-nnpa-quant-dynamic` and `nnpa-quant-op-types`. The old option `nnpa-quantization` is removed.

Some examples:
1. enable dynamic quantization and let compiler decide how to quantize ops automatically: `--nnpa-quant-dynamic`
2. enable dynamic quantization with options: symmetric for activations and symmetric for weights: `--nnpa-quant-dynamic=symActivation,symWeight`.
3. enable dynamic quantization for MatMul op types only: `--nnpa-quant-dynamic --nnpa-quant-op-types=MatMul`. Currently only MatMul is supported.

Help information for the options:

```shell
  --nnpa-quant-dynamic                                                              - Enable dynamic quantization of the input model. If enabled, it only quantizes from fp32
 to i8. If an ONNX operation is already in i8, no quantization is applied to that operation. Optionally, a comma-separated list of quantization options can be specified as i
ts value, e.g. -nnpa-quant-dynamic=symActivation,symWeight.
  --nnpa-quant-dynamic=<value>                                                      - Enable dynamic quantization of the input model. If enabled, it only quantizes from fp32
 to i8. If an ONNX operation is already in i8, no quantization is applied to that operation. Optionally, a comma-separated list of quantization options can be specified as i
ts value, e.g. -nnpa-quant-dynamic=symActivation,symWeight.
    =symWeight                                                                      -   Symmetric quant for weights.
    =asymWeight                                                                     -   Asymmetric quant for weights.
    =symActivation                                                                  -   Symmetric quant for activations.
    =asymActivation                                                                 -   Asymmetric quant for activations.
    =<empty>                                                                        -   Compiler automatically finds the best options. Default option when `-nnpa-quant-dynam
ic` is specified without any value.
  --nnpa-quant-op-types[=<string>]                                                    - A comma-separated list of types of operations that are quantized. E.g. 'MatMul,Conv'.
 Strings for types are the same as ONNX operator names in https://onnx.ai/onnx/operators/. Currently, only MatMul is supported. Without specifying this option, the compiler
will determine the operation types by itself.
```


src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
```diff
@@ -49,11 +49,56 @@ using namespace onnx_mlir;
...
-  configureOnnxToZHighLoweringPass(optReport == OptReport::NNPAUnsupportedOps);
   // z16 does not support for hardware saturation.
   // So, force its usage to compiler generated sticks.
   if (nnpaEnableSaturation && isLessEqualNNPALevel(NNPALevel::M14))
     nnpaEnableCompilerStickUnstick = true;
+
+  // Configure ONNXToZHighLoweringPass.
+  bool isDynQuant = !nnpaQuantDynamic.empty();
+  bool isWeightSym = false;
+  bool isActivationSym = false;
```
AlexandreEichenberger, line 60: quick question: why is the original init false/false here, but the auto value below is true/false.

Should't the default with no values being the same as the auto? Is there a reason that we want the "no options whatsoever" be different than auto?

AlexandreEichenberger (approved):
LGTM, left one question and a small wording change for help.

Sunny-Anand (approved):
LGTM

src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
```diff
@@ -49,11 +49,56 @@ using namespace onnx_mlir;
...
+        break;
+      default:
+        llvm_unreachable("Unsupported quantization options");
+        break;
+      }
+    }
+    if (!isWeightSym) {
+      // TODO: Support asymmetric quantiation for weights.
+      llvm::outs()
+          << "Asymmetric quantization for weights hasn't yet supported. "
```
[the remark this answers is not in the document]
tungld: Thanks! I used `is not`.

src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
tungld: Yes, it should be the same. Fixed it. Thanks!

PR #3044, opened 2025-01-15
Build light weight PyRuntime without llvm or onnx-mlir

chentong319:
Motivation

 Python driver is needed to run the compiled model. Currently, the driver is built with onnx-mlir and can only be run in the env where onnx-mlir is built, typically inside the onnx-mlir docker image. When the compilation can be done by calling the onnx-mlir docker image, we'd like to run the compiled .so with python driver in the local env, so that all the packages installed in the local env can be used, rather than installing them on top of docker. 

In order to reach this goal, the PR tried to remove the unnecessary dependencies of pyruntime if the light-weight pyruntime is the  target. Otherwise, there is no change in how onnx-mlir is built and used.
Details can be found in docs/build-pyruntime-lit.md. 
I tried the build on a z16 machine: it takes less than 2 minutes.

Components in this PR
1. CMakefile and source code changes to cut the dependencies of pyruntime. An option `ONNX_MLIR_ENABLE_PYRUNTIME_LIT` is used to control Cmake, and consequently a compile definition `ENABLE_PYRUNTIME_LIT` issued to control the source code.
2. Wrap the built pyruntime driver into a python package
3. This python package use python docker package to call the compiler. This is equivalent to docker/onnx-mlir.py in functionality but with "import docker" interface. 

Test
Run successfully with utils/BuildPyRuntimeLit.sh.

Future works:
1. Support of float16
2. Not all llvm utilities are replaced. Only the essential ones have been implemented.
3. Can third_party/onnx be removed?
4. Try to integrate the precompiled lib for different os-arch into the package. Enable user to use pip install the package from pip server.
5. Try to integrate the utils/build-pyruntime-lit.sh into python package and invoke the build when pip install is executed.
6. Add the test of light weight PyRuntime into the build.


CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ option(ONNX_MLIR_ENABLE_STABLEHLO "Enable StableHLO support." ON)
 option(ONNX_MLIR_ENABLE_WERROR "Enable warnings as errors." OFF)
 option(ONNX_MLIR_SUPPRESS_THIRD_PARTY_WARNINGS "Suppress warning in third_party code." ON)
 option(ONNX_MLIR_ENABLE_JAVA "Set to ON for building the Java runtime, tools, and tests" ON)
+option(ONNX_MLIR_ENABLE_PYRUNTIME_LIT "Set to ON for building Python driver of running the compiled model without llvm-project." OFF)
```
AlexandreEichenberger, line 14: Maybe the name should be explained: if off, then no pyruntime is build? Or its build anyway, but when on, then it's only the pyruntime? Or when off, pyruntime is build one way, but when off, its build another way?

Maybe the name could be a bit more explicit, depending on what the answer is from the question above.

minor question: `_LIT` is it for "_LIGHT"?

CMakeLists.txt
```diff
@@ -207,8 +216,12 @@ if (ONNX_MLIR_ENABLE_STABLEHLO)
   add_compile_definitions(ONNX_MLIR_ENABLE_STABLEHLO)
 endif()
 
-add_subdirectory(utils)
-add_subdirectory(include)
-add_subdirectory(src)
-add_subdirectory(docs)
-add_subdirectory(test)
+if (ONNX_MLIR_ENABLE_PYRUNTIME_LIT)
```
AlexandreEichenberger, line 219: I see this (and above): there are some dir that added on both path. Is it that the order of them is important?

MLIR.cmake
```diff
@@ -1,33 +1,38 @@
...
-# won't change accordingly.
-unset(LLVM_DIR CACHE)
-if (NOT DEFINED MLIR_DIR)
-  message(FATAL_ERROR "MLIR_DIR is not configured but it is required. "
-    "Set the cmake option MLIR_DIR, e.g.,\n"
-    "    cmake -DMLIR_DIR=/path/to/llvm-project/build/lib/cmake/mlir ..\n"
-    )
-endif()
+if (ONNX_MLIR_ENABLE_PYRUNTIME_LIT)
+  function(llvm_update_compile_flags name)
```
AlexandreEichenberger, line 4: Can you add a one liner comment on why this function is defined here.

src/Runtime/python/onnxmlirdocker.py
```diff
@@ -0,0 +1,151 @@
+import numpy as np
```
AlexandreEichenberger, line 1: That file is totally new, is that right?

AlexandreEichenberger (commented):
Still very early review; let me get the overall picture.

How would the new code base change if we were to adapt it for PyTorch as well? Namely, what is common to any "docker compilation + pyruntime_lit" vs what is specialized for "ORT" and what would be specialized for PyTorch?

Do you mind giving me this high level info, and then I will be able to better understand the overall approach. Thanks

chentong319:
> Still very early review; let me get the overall picture.
> 
> How would the new code base change if we were to adapt it for PyTorch as well? Namely, what is common to any "docker compilation + pyruntime_lit" vs what is specialized for "ORT" and what would be specialized for PyTorch?
> 
> Do you mind giving me this high level info, and then I will be able to better understand the overall approach. Thanks

This docker compilation + pyruntime_lit provides the basic functionality for compilation and run. The interface for ORT or PyTorch will be in the python package built on top of them. This PR contains only the package, onnxmlir, with ORT interface. In future, another package for PyTorch interface will be provided.

CMakeLists.txt
chentong319: When this option is off, the pyruntime is built with onnx-mlir and llvm-project, as it was previously.
When this option is on, only the pyruntime is built without llvm-project.

Yes, LIT for LIGHT.

CMakeLists.txt
chentong319: Only the `src` is added on both path. I do not think the order of add_subdirectory matters. Just to keep the original add_subdirectory together.

src/Runtime/python/onnxmlirdocker.py
chentong319: Yes, it is the file to use docker package.

tungld (commented):
LGTM! 

Hope that the environment variable `ENABLE_PYRUNTIME_LIT` will be removed completely in next PRs.

LIT in `ONNX_MLIR_ENABLE_PYRUNTIME_LIT` is confusing with LIT tests. It would be better to use other word, e.g. LIGHT or to remove it.

tungld (approved):
LGTM!

chentong319:
> LGTM!
> 
> Hope that the environment variable `ENABLE_PYRUNTIME_LIT` will be removed completely in next PRs.
> 
> LIT in `ONNX_MLIR_ENABLE_PYRUNTIME_LIT` is confusing with LIT tests. It would be better to use other word, e.g. LIGHT or to remove it.
The definition, ENABLE_PYRUNTIME_LIT, will be away finally. Current version is the transition from the previous one and the new one.
Alex has the same concern about _LIT. I will change it to LIGHT in this PR.

PR #3045, opened 2025-01-16
Add a document for quantization on NNPA

tungld:
This PR adds a document file `docs/Quantization-NNPA.md` to explain how to use quantization on NNPA.

### DIFF
--- a/docs/Quantization-NNPA.md
+++ b/docs/Quantization-NNPA.md
@@ -6,17 +6,23 @@ NNPA in IBM Telum II supports 8-bit signed-integer quantized matrix multiplicati
 
 There are two approaches to using quantization in the onnx-mlir compiler, depending on the input ONNX model to the compile:
 - The input model is a quantized model that was quantized by other frameworks such as ONNX Runtime. In this case, the input ONNX model contains 8-bit operations, and the onnx-mlir compiler selects suitable 8-bit operations to run on NNPA. There is no special compile flags needed to enable quantization when compiling this quantized model. Hence, we do not discuss this case in this document.
+  - In this approach, the compiler supports both static and dynamic quantized models.
 - The input model is a non-quantized model, e.g. operations operate on float32 data types. In this case, the onnx-mlir compiler provides several quantization options in order to quantize the model during compilation, then run the compiled model on NNPA. The remaining of this document describes this approach.
+  - In this approach, the compiler only supports dynamic quantization.
 
+In both approaches, the following constraints are applied:
+- Only per-tensor quantization is supported, meaning `scale` and `zero_point` are computed per-tensor and are scalar values.
+- Target quantization data type is 8-bit signed-integer.
+ 
 Quantization requires NNPA in IBM Telum II, meaning that the following compile flags must be specified to enable quantization: `-maccel=NNPA -march=arch15`.
 
-# Quantization by the compiler
+# Dynamic quantization by the compiler
 
-It is important to note that the onnx-mlir compiler currently:
+Again, it is important to note that the onnx-mlir compiler currently:
 - supports per-tensor dynamic quantization, and
-- quantizes data tensors from float32 to 8-bit signed integer because NNPA supports 8-bit signed integers. If a data tensor in the input model is already in 8-bit singed integer, the compiler will not quantize it again.
+- quantizes data tensors from float32 to 8-bit signed integer. If a data tensor in the input model is already in 8-bit singed integer, the compiler will not quantize it again.
 
-The compiler provides two compile flags for quantizing a model at compile time:
+The compiler provides two compile flags for dynamically quantizing a model at compile time:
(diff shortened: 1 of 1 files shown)


tungld:

docs/Quantization-NNPA.md
```diff
@@ -0,0 +1,59 @@
+<!--- SPDX-License-Identifier: Apache-2.0 -->
+
+# Overview 
+ 
+NNPA in IBM Telum II supports 8-bit signed-integer quantized matrix multiplications. This document shows how to compile an ONNX model for quantization on NNPA.
+
+There are two approaches to using quantization in the onnx-mlir compiler, depending on the input ONNX model to the compile:
+- The input model is a quantized model that was quantized by other frameworks such as ONNX Runtime. In this case, the input ONNX model contains 8-bit operations, and the onnx-mlir compiler selects suitable 8-bit operations to run on NNPA. There is no special compile flags needed to enable quantization when compiling this quantized model. Hence, we do not discuss this case in this document.
```
Sunny-Anand, line 8: Suggestion: `ONNX-MLIR supports both dynamic and static quantized models as part of approach1`

docs/Quantization-NNPA.md
```diff
@@ -0,0 +1,59 @@
...
+- The input model is a non-quantized model, e.g. operations operate on float32 data types. In this case, the onnx-mlir compiler provides several quantization options in order to quantize the model during compilation, then run the compiled model on NNPA. The remaining of this document describes this approach.
+
+Quantization requires NNPA in IBM Telum II, meaning that the following compile flags must be specified to enable quantization: `-maccel=NNPA -march=arch15`.
+
+# Quantization by the compiler
+
+It is important to note that the onnx-mlir compiler currently:
+- supports per-tensor dynamic quantization, and
```
Sunny-Anand, line 16: Is this not true for static quantized models?

docs/Quantization-NNPA.md
```diff
@@ -0,0 +1,59 @@
...
+- quantizes data tensors from float32 to 8-bit signed integer because NNPA supports 8-bit signed integers. If a data tensor in the input model is already in 8-bit singed integer, the compiler will not quantize it again.
+
+The compiler provides two compile flags for quantizing a model at compile time:
```
Sunny-Anand, line 19: Since these flags are targetting only dynamic quantization, shall we specify `dynamically quantizing` here?

docs/Quantization-NNPA.md
```diff
@@ -0,0 +1,59 @@
...
+```
+
+Given `scale` and `zero_point`, the input `x` is quantized to
+```
+quantized_x = x/scale + zero_point
+```
+
+# Performance notes
+
+It is often the case that symmetric quantization leads to better inference performance but poorer accuracy than asymetric quantization.
```
Sunny-Anand, line 55: Typo: asymmetric

Sunny-Anand (approved):
LGTM, just have a few questions.

docs/Quantization-NNPA.md
AlexandreEichenberger: maybe we could use `--nnpa-dquant` for dynamic quantization, and use it systematically for both options.

AlexandreEichenberger (commented):
LGTM, will let the final word to @Sunny-Anand

tungld:
@Sunny-Anand @AlexandreEichenberger could you take another look at my new changes based on your comments? Would like to make sure everything is clear before merging. Thanks!

Sunny-Anand (approved):
LGTM. Thanks for the changes.

docs/Quantization-NNPA.md
```diff
@@ -0,0 +1,65 @@
+<!--- SPDX-License-Identifier: Apache-2.0 -->
+
+# Overview 
+ 
+NNPA in IBM Telum II supports 8-bit signed-integer quantized matrix multiplications. This document shows how to compile an ONNX model for quantization on NNPA.
```
AlexandreEichenberger, line 5: There seems to be no consensus on what quantization means. It always means going from a higher precision to a lower precision, but I don't think it necessarily implies integer representation. See here for example

https://huggingface.co/docs/optimum/en/concept_guides/quantization

So maybe we could be a bit clearer here.

NNPA in IBM Telum II supports 8-bit signed-integer quantized matrix multiplications. This document shows how to compile an ONNX model for 8-bit quantization on NNPA. When not following these steps, models will still be accelerated when targeting Telum systems using a mixture of 16-bit floating-point numbers for computations mapped to the Telum's Integrated AI accelerator and 32-bit floating-point numbers for computations mapped to the Telum CPUs.

I think that once this is out of the way, we may continue having the text below without changes. Or one could use "8-bit integer quantization" once at the beginning of sections.

tungld: Thanks! I updated it with your content.

PR #3046, opened 2025-01-17
Update the instruction for building multiple accelerators

tungld:
This PR updates `docs/AddCustomAccelerators.md` with instructions to build multiple accelerators.

### DIFF
--- a/src/Accelerators/CMakeLists.txt
+++ b/src/Accelerators/CMakeLists.txt
@@ -1,7 +1,8 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Populate the accelerator list and add the accelerator subdirectories.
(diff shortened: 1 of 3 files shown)


src/Accelerators/CMakeLists.txt
```diff
@@ -10,7 +11,8 @@ if (ONNX_MLIR_ACCELERATORS)
     add_subdirectory(${t})
 
     # If the accelerator can be built
-    if (${t}_ENABLED)
+    string(TOUPPER ${t} T)
```
tungld, line 14: Just to normalize accelerator names to upper case, since all cmake variables seem to be in upper case.

AlexandreEichenberger (approved):
LGTM

PR #3048, opened 2025-01-21
Option to set the number of threads for parallel compilation

imaihal:
All available CPUs are used in parallel compilation in current main branch.
This PR creates an option to set the number of threads for parallel compilation. By using `-j <#threads>` when compiling, thread pool with `<#threads>` threads are created, and used for parallel compilation in `parallelFor()` API. 
The default value is 0 and all the available CPUs are used as same with current main. When setting `-j 1`, use single thread without setting thread pool.


src/onnx-mlir.cpp
```diff
@@ -68,7 +69,12 @@ int main(int argc, char *argv[]) {
   }
 
   // Create context after MLIRContextCLOptions are registered and parsed.
-  mlir::MLIRContext context;
+  mlir::MLIRContext context(mlir::MLIRContext::Threading::DISABLED);
+  llvm::DefaultThreadPool pool(
```
AlexandreEichenberger, line 73: Why do we create a pool with 1 thread when we don't set it? Can we create it only when threads > 1?

src/Compiler/CompilerOptions.cpp
```diff
@@ -665,6 +666,13 @@ static llvm::cl::opt<bool, true> enable_bound_check("enable-bound-check",
     llvm::cl::location(enableBoundCheck), llvm::cl::init(false),
     llvm::cl::cat(OnnxMlirOptions));
 
+static llvm::cl::opt<int, true> compilation_num_threads("j",
```
AlexandreEichenberger, line 669: Thanks for reusing the same name as for other programs...

src/onnx-mlir.cpp
imaihal: Thanks for the comment. I thought so too, but when I created the pool within the if statement, I got an error  when using the pool(maybe it was illegal instruction). I'm not sure why this happened.
I will check again.

imaihal: @AlexandreEichenberger  Updated. I think this is better way to create the pool. I studied from LLVM repo.

src/onnx-mlir.cpp
```diff
@@ -68,7 +69,14 @@ int main(int argc, char *argv[]) {
   }
 
   // Create context after MLIRContextCLOptions are registered and parsed.
-  mlir::MLIRContext context;
+  mlir::MLIRContext context(mlir::MLIRContext::Threading::DISABLED);
+  std::unique_ptr<llvm::ThreadPoolInterface> threadPoolPtr;
```
chentong319, line 73: What's the program behavior if setThreadPool is not set? I guess that the compiler uses all the available threads. 
What's wrong with j=1? Can we specify that only one thread is used? Then comes the question: what the default value should be?
We should add these descriptions In the option definition.
By the way,  I am surprised to say `Threading::DISABLED`, not enabled.

AlexandreEichenberger: @imaihal please look into this. Ideally, we would like to get the best performance and fastest compiler (using only `-O3` with possibly the arch and the use of the NNPA).

Chatting with @chentong319 , MLIR is default multiple threads, so we want to preserve that. The way to ask for a single thread would be explicitly stating `-j 1`

imaihal: @AlexandreEichenberger @chentong319 Thanks for the comments.

I see. I updated to use 0 by default and use all the available processor for compilation. When setting `-j 1`, use single thread without setting thread pool. When `j > 1`, specified the number for threads are used.

Any other requests that I don't refrect?

src/onnx-mlir.cpp
```diff
@@ -68,7 +69,23 @@ int main(int argc, char *argv[]) {
...
   // Create context after MLIRContextCLOptions are registered and parsed.
+  // The multi-threading in MLIRContext is enabled by default. It must be
+  // disabled to control the number of threads. To use single thread, simply
+  // disable it. To use a specific number of threads, disable it once and then
+  // set a new thread pool.
   mlir::MLIRContext context;
+  std::unique_ptr<llvm::ThreadPoolInterface> threadPoolPtr;
+  assert(compilationNumThreads >= 0 &&
+         "The number of threads for compilation by "
+         "'-j' option should be positive value.");
```
tungld, line 80: You may define `compilationNumThreads` to be `uint64_t`, so that it is never negative, e.g. `%llvm::cl::opt<int, true> compilation_num_threads("j",`

src/Compiler/CompilerOptions.cpp
```diff
@@ -92,6 +92,7 @@ OptReport optReport;                                   // onnx-mlir only
 bool useOldBufferization;                              // onnx-mlir only
 bool enableTiming;                                     // onnx-mlir only
 bool enableBoundCheck;                                 // onnx-mlir only
+int compilationNumThreads;                             // onnx-mlir only
```
tungld, line 95: Can we apply this option to `onnx-mlir-opt` as well so that we can check performance by calling `onnx-mlir-opt --constprop-onnx -j8` for a single operation? It's ok if it needs another PR.

src/onnx-mlir.cpp
AlexandreEichenberger: That seems to be correct. Do you ignore (or warning) if user gave a negative value?

src/Compiler/CompilerOptions.cpp
AlexandreEichenberger: @tungld I was thinking along the same lines, also for example to test that we get the same result for a test that uses parallelism.

@imaihal I would then add a test that tries the operation that you parallelized and try it with `-j 4` for example. You may have to copy a large constant to do that... can lift it from a model.

src/onnx-mlir.cpp
chentong319: This sounds great!

src/Compiler/CompilerOptions.cpp
imaihal: @tungld @AlexandreEichenberger Thanks for the comments.

I applied this option to `onnx-mlir-opt`, but parallelization of constprop is not included in this PR (It is included in PR #3042). So I want to add the tests in PR #3042.

src/onnx-mlir.cpp
imaihal: Done. Thanks!

src/Compiler/CompilerOptions.cpp
AlexandreEichenberger: Absolutely, this PR set the options, and it will be tested once there is a new opt that can use parallelism.

AlexandreEichenberger:
@imaihal Is your PR ready for review again, meaning do you feel you acted on all the questions/requests?

imaihal:
@AlexandreEichenberger Yes. Could you review this again?

src/Compiler/CompilerOptions.cpp
tungld: @imaihal I didn't see any change to `onnx-mlir-opt.cpp`. I am curious on how this option is effective in`onnx-mlir-opt` command line?

AlexandreEichenberger (approved):
LGTM, thanks @imaihal  for incorporating the feedback from Tong and I.

src/Compiler/CompilerOptions.cpp
imaihal: @tungld New thread pool is created in `MlirOptMain.cpp` of MLIR. This is due to handle`-split-input-file` option  (https://github.com/llvm/llvm-project/blob/main/mlir/lib/Tools/mlir-opt/MlirOptMain.cpp#L609-L613). So, it seems to be difficult to set our own threadpool in `onnx-mlir-opt.cpp`. 

In lit tests, how about using`--mlir-disable-threading` for single thread and using mlir default(use all cpus) for multi-thread? Or, using pass local option might be possible to specify our own thread pool.

imaihal: @tungld Sorry, it is possible just by modifying onnx-mlir-opt.cpp. I updated.

PR #3049, opened 2025-01-22
Don't try to free static array in mnist example

Zentrik:
I don't think its correct to free a static array, it was causing ABORTS with `free(): invalid pointer` printed for me.


AlexandreEichenberger (approved):
You are absolutely correct, thanks for finding this bug.
I will merge the PR once it succeeds the CIs (which it should as the change is non-functional as it impacts only a README file.

AlexandreEichenberger:
@Zentrik sorry dropped the ball on this one. Will merge since it passed the tests.

AlexandreEichenberger:
Thanks @Zentrik for your contribution, much appreciated

PR #3051, opened 2025-01-25
Remove element type restriction in softmax lowering

srcarroll:
This patch removes the assertion that input element type is `f32` in the softmax to stablehlo lowering. There appears to be no good reason for this restriction as stablehlo supports more types.

The original `Softmax.mlir` test has been renamed to `Softmax-Decompose.mlir` and replaced by a new one the leaves out the `decompose-onnx` pass.  The primary reason for this is to be able to test `bf16` lowering which is still not supported by the decomposition. An error is thrown while computing the identity for max reduction (see https://github.com/onnx/onnx-mlir/blob/main/src/Dialect/Mlir/DialectBuilder.cpp#L698). Since downstream users don't necessarily need the decomposition it suffices to test without it.


AlexandreEichenberger (approved):
LGTM

AlexandreEichenberger:
Let me know if you want me to merge it @srcarroll

PR #3052, opened 2025-01-28
Fix ASAN/UBSAN issues in DimAnalysis

jorickert:
- Fixes a memory leak
- Fixes an integer overflow caused by a dynamic shape
- Fixes reshape to wrong type in LIT tests

### DIFF
--- a/src/Dialect/ONNX/ONNXDimAnalysis.cpp
+++ b/src/Dialect/ONNX/ONNXDimAnalysis.cpp
@@ -361,11 +361,11 @@ static bool exploreSameDimsUsingShapeHelper(const DimAnalysis::DimT &dim,
   ONNXOpShapeHelper *shapeHelper =
       shape_op.getShapeHelper(op, {}, nullptr, nullptr);
   // If no shape helper, or unimplemented, just abort.
-  if (!shapeHelper || !shapeHelper->isImplemented())
+  if (!shapeHelper)
     return false;
 
   // Compute shape.
-  if (failed(shapeHelper->computeShape())) {
+  if (!shapeHelper->isImplemented() || failed(shapeHelper->computeShape())) {
     delete shapeHelper;
     return false;
   }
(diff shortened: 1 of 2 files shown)


AlexandreEichenberger:
@tungld do you mind reviewing this PR, its in dim analysis. Thanks

src/Dialect/ONNX/ONNXDimAnalysis.cpp
```diff
@@ -361,11 +361,11 @@ static bool exploreSameDimsUsingShapeHelper(const DimAnalysis::DimT &dim,
   ONNXOpShapeHelper *shapeHelper =
       shape_op.getShapeHelper(op, {}, nullptr, nullptr);
   // If no shape helper, or unimplemented, just abort.
-  if (!shapeHelper || !shapeHelper->isImplemented())
+  if (!shapeHelper)
```
tungld, line 364: Could you explain what's wrong with calling `shapeHelper->isImplemented()`?

src/Dialect/ONNX/ONNXDimAnalysis.cpp
```diff
@@ -961,12 +961,14 @@ void DimAnalysis::visitDim(
...
-    for (int64_t i = 0; i < dataType.getRank(); ++i)
-      dataStaticSize *= dataType.getShape()[i];
-    for (int64_t i = 0; i < outputType.getRank(); ++i)
-      outputStaticSize *= outputType.getShape()[i];
+    for (int64_t i = 0; i < dataType.getRank(); ++i) {
+      dataStaticSize *= dataType.isDynamicDim(i) ? -1 : dataType.getShape()[i];
+    }
+    for (int64_t i = 0; i < outputType.getRank(); ++i) {
+      outputStaticSize *=
+          outputType.isDynamicDim(i) ? -1 : outputType.getShape()[i];
```
tungld, line 970: Thanks for fixing this!

test/mlir/onnx/onnx_dim_analysis.mlir
```diff
@@ -184,38 +184,38 @@ func.func @test_matmul_batchsize(%arg0: tensor<?x8x16x16xf32>) -> tensor<?x8x16x
 
 // -----
 
-func.func @test_matmul_batchsize_diff_rank(%arg0: tensor<8x?x16x4xf32>) -> tensor<8x?x16x32xf32> {
+func.func @test_matmul_batchsize_diff_rank(%arg0: tensor<8x?x16x4xf32>) -> tensor<8x?x16x128xf32> {
   %shape = onnx.Constant dense<[-1, 4, 128]> : tensor<3xi64>
-  %0 = "onnx.Reshape"(%arg0, %shape) {allowzero = 0 : si64} : (tensor<8x?x16x4xf32>, tensor<3xi64>) -> tensor<?x4x32xf32>
-  %1 = "onnx.MatMul"(%arg0, %0) : (tensor<8x?x16x4xf32>, tensor<?x4x32xf32>) -> tensor<8x?x16x32xf32>
-  "onnx.Return"(%1) : (tensor<8x?x16x32xf32>) -> ()
+  %0 = "onnx.Reshape"(%arg0, %shape) {allowzero = 0 : si64} : (tensor<8x?x16x4xf32>, tensor<3xi64>) -> tensor<?x4x128xf32>
```
tungld, line 189: Thanks for fixing this!

src/Dialect/ONNX/ONNXDimAnalysis.cpp
jorickert: There is nothing wrong with it, but if the shape helper is not implemented we still need to delete it or we will leak it's memory. This happens a few lines down

tungld: I got it! That's why we have to distinguish the two conditions. Thanks for the explanation!

tungld (approved):
LGTM!

PR #3054, opened 2025-01-29
Update onnx requirement to 1.17.0

jorickert:
The onnx submodule is using version 1.17.0, so I think the version used in the requirements should be the same


hamptonm1 (approved):
LGTM

PR #3056, opened 2025-01-29
Optimization for Roberta unstick->reshape->transpose->reshape->stick

AlexandreEichenberger:
In some situations, a sequence of transformations are "no-ops" under a given ztensor representation.

The pattern that is exploited here is the 3DS for `(A, B, C*D) <-> (A*C, B, D)` which are equivalent when `B%32=0` and `D%64=0`,

The pattern detected and transformed to a `high/zlow reshape` are the following

image

and

image

A high level proof is here below. For a detail proof, one has to follow every steps of the transformations above and show equality of memory accesses, namely that when accessing `(e3, e2, e1)`, we get the same memory location in the original 3DS tensor as well as the final 3DS tensor in the above examples.

image

In practice, this PR adds 2 rules to catch the above 2 patterns, replace them with a `zhigh.Reshape` which is similar to the `memref.reshape` in that it performs no "data layout transformation", just provide mapping between two equivalent shapes. The ZHigh version performs such equivalency for ZTensor formats such as 3D.

THe ZHigh reshape operation is lowered to ZLow equivalent reshape operation, which is then transformed to a `memref.reinterpret_cast` operations after all members are normalized.

PR adds littlest to catch the patterns listed above, and one for ZHigh to ZLow conversion, and one for ZLow to memref.

I checked that the values generated by Roberta with/without this PR were the same. Performance measurements show that in Roberta, the number of transpose were reduced from 48 to 12 (with a reductions in stick/unstick also by 36 operations). Speedup for the time spent in the transpose/stick/unstick were reduced by 9%, 33%, and 37%. Overall (with one NNPA and once CPU), the time was reduced by 4%.

At this time, this PR is restricted to static shapes.


src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.td
```diff
@@ -95,17 +95,6 @@ def replaceONNXBatchNormalizationInferenceModePattern : Pattern<
 //
 //===----------------------------------------------------------------------===//
 
-
```
AlexandreEichenberger, line 98: migrated the code elsewhere so that it can be reused, as it was needed to support the reshape op.

src/Accelerators/NNPA/Dialect/ZHigh/ZHighOps/OpHelper.cpp
```diff
@@ -402,8 +473,8 @@ AffineMapAttr getTiling2DTo4DMap(OpBuilder &b, Value val) {
   return AffineMapAttr::get(map);
 }
 
-AffineMapAttr getTiling3DTo4DMap(OpBuilder &b, Value val) {
-  assert(isTiling3DTo4D(val) &&
+AffineMapAttr getLeftmostTiling3DTo4DMap(OpBuilder &b, Value val) {
```
AlexandreEichenberger, line 476: Many of the prior operations where not specific if they applied to the right most or leftmost position, as only one was needed. As I added more, I made all names more explicit,

src/Accelerators/NNPA/Transform/ZLow/ZLowRewrite.cpp
```diff
@@ -33,6 +34,38 @@ using namespace mlir;
...
+public:
+  using OpRewritePattern<ZLowReshapeOp>::OpRewritePattern;
+  LogicalResult matchAndRewrite(
+      ZLowReshapeOp reshapeOp, PatternRewriter &rewriter) const override {
+    Location loc = reshapeOp.getLoc();
+    MultiDialectBuilder<MemRefBuilder> create(rewriter, loc);
+    IndexExprScope currScope(&rewriter, loc);
+    // Here, cannot use the shape found in the reshape op, as it is the original
+    // shape before memref normalization.
+    Value input = reshapeOp.getX();
```
tungld, line 51: We should check if X is normalized or not before processing further, by using something like this:
```
    // Input must have no affine layout. In other words, it has been normalized.
    if (hasNonIdentityLayout(input.getType()))
      return failure();
```

Without this check, I see that, in your lit test `zlow-rewrite.mlir`, `zlow.reshape` with affine_maps is still lowered to `memref.reinterpret_cast`

test/mlir/accelerators/nnpa/transform/zlow-rewrite.mlir
```diff
@@ -713,6 +713,43 @@ func.func @should_not_rewrite_unstick_transpose_stick_3(%arg0: memref<5x10xf16,
...
+    %0 = "krnl.global"() {alignment = 4096 : i64, name = "constant_stickify_3", shape = [96, 6, 1, 1, 32, 64], value = dense_resource<zhigh_3> : tensor<2359296xi8>} : () -> memref<96x6x1x1x32x64xf16>
+    "zlow.matmul"(%alloc, %arg1, %0, %alloc_1, %alloc_0) {is_bcast1 = 0 : si64, is_bcast23 = 0 : si64, is_stacked = -1 : si64, transposeA = 0 : si64, transposeB = 0 : si64} : (memref<96x384x64xf16, #map>, memref<96x64x384xf16, #map>, memref<96x6x1x1x32x64xf16>, memref<4xi64>, memref<96x384x384xf16, #map>) -> ()
+    return %alloc_0 : memref<96x384x384xf16, #map>
+
+// mlir2FileCheck.py
+// CHECK-DAG:   [[MAP_0_:#.+]] = affine_map<(d0, d1, d2) -> (d0, d2 floordiv 64, 0, d1 floordiv 32, d1 mod 32, d2 mod 64)>
+// CHECK-LABEL:  func.func @handle_zlow_reshape
+// CHECK-SAME:   ([[PARAM_0_:%.+]]: memref<8x384x768xf16, #map>, [[PARAM_1_:%.+]]: memref<96x64x384xf16, #map>) -> memref<96x384x384xf16, #map> {
+
+// CHECK-DAG:       [[VAR_reinterpret_cast_:%.+]] = memref.reinterpret_cast [[PARAM_0_]] to offset: [0], sizes: [96, 384, 64], strides: [24576, 64, 1] : memref<8x384x768xf16, #map> to memref<96x384x64xf16>
```
tungld, line 746: It does not look like what we are expecting since the input memref is not normalized.
To check this case, you can
- add a new check for this case where we call `--normalize-memrefs`, by adding this line to top of this file:
```
// RUN: onnx-mlir-opt --march=z16 --maccel=NNPA --normalize-memrefs --zlow-rewrite --canonicalize %s -split-input-file | FileCheck %s --check-prefix=RESHAPE
```
- then, replace `CHECK` by `RESHAPE` in `CHECK-DAG, CHECK-LABEL, ...` since we use the prefix `RESHAPE` for this check.

src/Accelerators/NNPA/Transform/ZLow/ZLowRewrite.cpp
AlexandreEichenberger: Got it, I did not realize that `zlow-rewrite` ran twice. Its now fixed.

AlexandreEichenberger:
Thanks @tungld for the feedback, implemented both suggestions.

tungld (approved):
LGTM!

Glad to see the performance improvement!

PR #3059, opened 2025-01-30
Bump various ops to opset 22, adding bf16 support

jorickert:
Bumped ops:
Acos
Acosh
Asin
Asinh
Atan
Atanh
AveragePool
Bernoulli
Conv
ConvTranspose
Cos
Cosh
DeformConv
Det
Dropout
Elu
EyeLike
GRU
GlobalAveragePool
GlobalMaxPool
HardSigmoid
HardSwish
InstanceNormalization
LSTM
LpNormalization
LpPool
MaxPool
MaxRoiPool
MaxUnpool
Mish
Multinomial
NegativeLogLikelihoodLoss
RNN
RandomNormal
RandomNormalLike
RandomUniform
RandomUniformLike
RoiAlign
Round
Selu
Sin
Sinh
Softplus
Softsign
Tan
ThresholdedRelu


hamptonm1:
@jorickert maybe there is some confusion on how to update these files but what was done is incorrect. We would not simply replace the opset version with the latest one although support is given for it. How the files would be update would be shown below in this example:

` "Clip": [13, 12, 11, 6],`

Here for Clip we specify the minimum opset version we support along with any of the newer ones. So if `"ConstantOfShape": [20]` uses bf16 support then the update would be `"ConstantOfShape": [22, 20]` which means we still keep the minimum version of support.

jorickert:
> Here for Clip we specify the minimum opset version we support along with any of the newer ones. 

This is different from how I thought the versioning works. Do really all versions that are supported need to be listed, or only if there is a breaking change between them?
My understanding was that only versions with breaking changes need to be listed.
So for example if `SomeOpV21` is a breaking change compared to `SomeOpV20`, we would need to list `[21, 20]`.
But if `SomeOpV22` is a strict superset of `SomeOpV21` I thought we can just update the version, without introducing a new copy of the version, so the version set would be `[22, 20]`. Is that not correct?

jorickert:
I tested it, and with this PR I can successful parse both Acos Opset 7 as well as Acos Opset 22. I will add a test for this

philass:
The way @jorickert implemented this is correct. Because opset 22 only adds a new type the ops are backwards compatible. This LGTM

philass (approved):

hamptonm1 (dismissed):
I am blocking until the maintainers review :)

philass:
> the precedent

@hamptonm1 What precedent?

AlexandreEichenberger (approved):
LGTM, since it's only adding support and not removing/changing any past behavior, this should be all good. Thanks for the update @jorickert

jorickert:
Before merging this I would like to understand the failure in #3065

hamptonm1:
> > the precedent
> 
> @hamptonm1 What precedent?

@philass precedent: preceding in time, order, or importance.

That is what I mean

PR #3062, opened 2025-02-03
Remove the pattern unstick_4ds_squeeze_stick_3ds

tungld:
This patch replaces the pattern
```
X -> unstick (4DS) -> Squeeze (axis=1) -> stick (3DS)
```
by
```
zhigh.Reshape from 4DS to 3DS
```

This pattern is found in the CCFD model.


src/Accelerators/NNPA/Dialect/ZHigh/ZHighOps/OpHelper.cpp
```diff
@@ -636,5 +636,25 @@ IntegerAttr getDefaultSaturation(PatternRewriter &rewriter) {
...
+  assert(shape4DS.size() == 4 && "The tensor must have rank of 4");
+  assert(shape4DS[1] == 1 && "The second dim must be 1");
+  if (hasStaticShape(val4DS.getType())) {
+    return create.constantInt64(
+        ArrayRef<int64_t>{shape4DS[0], shape4DS[2], shape4DS[3]});
+  }
+  Value dim0 = create.dim(val4DS, 0);
+  Value dim1 = create.dim(val4DS, 2);
+  Value dim2 = create.dim(val4DS, 3);
+  return create.concat(
```
AlexandreEichenberger, line 655: Awesome, in my second PR for optimizations related to high.reshape, I added a new `onnx.shape` builder where we can give the input value and a `permute` vector which let us pick (statically) which  dim to keep. So the sequence above can be reduced to a `create.onnx.shape(val4DS, [0, 2, 3])`. I use that for my opt, and if you don't mind I may also convert the pattern above once my PR is in. I think we may eventually need this elsewhere too.

PR #3064, opened 2025-02-03
Bump various ops to opset 21, adding int4/uint4 and 8 bit float support.

jorickert:
These are all backwards compatibel version bumps, only adding support for int/uint4.

Bumped ops:
Flatten
Identity
If
Loop
Pad
Reshape
Scan
Shape
Size
Squeeze
Transpose
Unsqueeze


jorickert:
Waiting on a review on https://github.com/onnx/onnx-mlir/pull/3059 first

tungld (approved):
LGTM! Thanks for the update!

PR #3065, opened 2025-02-03
Bump Upsample to Opset 10 and change the opset versioning to allow to skip over opset versions if a newer, backwards compatible one exists.

jorickert:
Upsample was marked as deprecated with Opset 10


src/Builder/OpBuildTable.inc
```diff
@@ -210,7 +210,7 @@ op_dialect_version_map_["TreeEnsembleClassifier"] = {1};
 op_dialect_version_map_["TreeEnsembleRegressor"] = {1};
 op_dialect_version_map_["Unique"] = {11};
 op_dialect_version_map_["Unsqueeze"] = {13, 11};
-op_dialect_version_map_["Upsample"] = {9, 7};
+op_dialect_version_map_["Upsample"] = {10, 7};
```
tungld, line 213: It looks like we have to keep 9 in the set, e.g. {10, 9, 7}. And defining a simple rule in `decompose.cpp` for `UpsampleV9Op`.

jorickert:
Supporting multiple different versions required to write simple decomposition/upgrade rule to rewrite the older opset version to a  newer one.
I extended this PR with a new built-time generated map that contains all versions of an operation as defined by onnx.
To determine the opset version for a node/op:
1.	Determine the latest valid opset version. This is the newest version in this opset-version-map that is older or equal to the current graph opset.
2.	Select the newest version from the versions supported by onnx-mlir that is equal or newer to the latest valid opset version. This allows it to skip over opset versions, that have a newer backwards compatible version.
Example:
	Versions in onnx and supported by onnx-mlir: [3, 5].
	Graph opset version to node version: 3 -> 3, 4 -> 3, 5 -> 5

	Versions in onnx: [7, 9, 10]. Version 10 is backwards compatible to version 9.
	Version supported by onnx-mlir: [7, 10].
	Graph opset version to node versionn: 7 -> 7, 8 -> 7, 9 -> 10, 10 -> 10
	
This changes allows it to not have duplicates of ops that changed in a backwards-compatible way, instead only versions with breaking changes need to be supported by onnx-mlir

jorickert:
@tungld @AlexandreEichenberger I extend this PR to include a mechanism to determine the opset version to use based on the opsets supported by onnx-mlir and the opsets defined by onnx. It would be great if you could comment on it and if you prefer this approach or the one of using trivial decompose/upgrade rules in onnx-mlir to handle the "skiipping" of backwards compatible opset versions.
Implementation note: I extended `gen_onnx_mlir.py` to build the opset map at build time, alternatively FronendDialectTransformer could be extended to read this information at runtime

tungld (approved):
LGTM!

Thanks @jorickert for proposing a new approach! It looks reasonable. But I would like to hear comments from @chentong319 before merging since he haven't done a lot of code for the frontend.

AlexandreEichenberger:
@chentong319 thanks for agreeing to review it. @jorickert Tong is our expert for frontend passes.

chentong319:
To my understanding, introducing op_sets_map fixed the issue of selecting dialect version. Could you move your description of the PR to docs? You can create a new document file and I will add reference to it later. Or you can add some comments in code and I can extract them out into document. Either way is fine.
Otherwise, the PR is good. Thanks.

jorickert:
I will move some comments from this PR inside the code later this/next week

chentong319 (approved):
LGTM!

PR #3066, opened 2025-02-03
Add support for ONNX.shape with permutation pattern

AlexandreEichenberger:
We are now seeing optimizations where we have equivalence between tensors of different shape. This PR enables the building of ONNX shapes for a given input tensor, where we may select arbitrary dimensions in arbitrary orders.

Examples

Say an 4D input with dims `(d0, d1, d2, d3)`. Call to `shape(input, {0, 1, 3, 2})` will produce a tensor with `[d0, d1, d3, d2]` values. Or call to `shape(input, {0, 2, 3})` will produce a shape of reduced dimensions (4D->3D) with dims `[d0, d2, d3]`.

This PR also allows shape builders to not have to provide the output type (namely a 1D tensor of int64_t with one value per rank) for more concise DDR rules.


src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHighCommon.td
```diff
@@ -58,20 +58,4 @@ def ACT_RELUAttr: NativeCodeCall<"$_builder.getStringAttr(\"ACT_RELU\")">;
 
 def GetTypeOf : NativeCodeCall<"$0.getType()" >;
 
-def GetNullAttr : NativeCodeCall<"Attribute()">;
```
AlexandreEichenberger, line 61: moved to `OpHelper.td` as eventually needed more generally.

src/Dialect/ONNX/DialectBuilder.cpp
```diff
@@ -376,6 +376,34 @@ Value OnnxBuilder::shape(
...
+  bool allStatic = true;
+  llvm::SmallVector<Value, 4> permutedDims;
+  llvm::SmallVector<int64_t, 4> permutedShapes;
+  for (int64_t p = 0; p < permRank; ++p) {
+    int64_t d = perm[p];
+    assert(
+        d >= 0 && d < inputRank && "perm values expected in [0..rank(input))");
+    if (ShapedType::isDynamic(inputShape[d]))
+      allStatic = false;
+    permutedDims.emplace_back(dim(input, d));
```
tungld, line 397: When the shape is static, these DimOp are still created. I know that they will be removed later by canonicalization, but they are quite redundant. We can have a light check like:
```c
bool isDynamic = llvm::any_of(inputShape, [](int64_t d) {return ShapedType::isDynamic(d);});
```

Then in the loop:
```
for (int64_t p = 0; p < permRank; ++p) {
   ...
   if (isDynamic)
      permutedDims.emplace_back(dim(input, d));
   ...
}

```

tungld (approved):
LGTM! Thanks for simplifying many codes!

PR #3067, opened 2025-02-05
[Test-Do-Not-Merge] Verify assert for member 

AlexandreEichenberger:
Usage of cg.assert, with how to generate the operation, which dialect to make legal,...

Cannot work with result of alloc as memref cannot be tested against a null value consistently.


AlexandreEichenberger:
Approach is not possible as there is no way to test whether a memref is null or not, at least consistently according to the doc.
Shows how to generate asserts, which could be reused for other purposes.

PR #3069, opened 2025-02-07
Add runtime check for Gather Op

chentong319:
Some onnx op, such as GatherOp and GatherElementsOp, requires the input value to be within certain range. For efficiency, onnx-mlir did not check that with the assumption that everything is correct.
This PR adds runtime check for the purpose of Debugging. The implementation is not the optimized. When the bug in Runtime bound check in llvm-project is fixed, we can use that for this purpose.
I tested this PR with RunONNXModel.py with certain inputs. Not sure how to add these test cases into standard test easily.

Future work: any other runtime check is needed?


src/Conversion/ONNXToKrnl/Tensor/Gather.cpp
```diff
@@ -122,6 +127,32 @@ struct ONNXGatherOpLowering : public OpConversionPattern<ONNXGatherOp> {
...
+            // are out of bounds.
+            // After the negative correction, the range should be [0, s-1]
+            Value upperBound = create.mem.dim(data, axisLit);
+            Value compareUpperBound =
+                create.math.slt(index.getValue(), upperBound);
+            rewriter.create<cf::AssertOp>(loc, compareUpperBound,
+                "indices of GatherOp is larger than the upper bound");
+            Value compareLowerBound =
+                create.math.sge(index.getValue(), zeroIE.getValue());
+            rewriter.create<cf::AssertOp>(loc, compareLowerBound,
```
AlexandreEichenberger, line 152: @chentong319 I know it is not optimized for speed, "anding" both condition and calling assert only once would speed the check up a bit. `create.math.andi(compareUpperBound, compareLowerBound)`.

src/Conversion/ONNXToKrnl/Tensor/GatherElements.cpp
```diff
@@ -93,6 +98,26 @@ struct ONNXGatherElementsOpLowering
...
+            // After the negative correction, the range should be [0, s-1]
+            Value upperBound = create.mem.dim(data, axis);
+            Value compareUpperBound =
+                create.math.slt(index.getValue(), upperBound);
+            rewriter.create<cf::AssertOp>(loc, compareUpperBound,
+                "indices of GatherOp is larger than the upper bound");
+            LiteralIndexExpr zero(0);
+            Value compareLowerBound =
+                create.math.sge(index.getValue(), zero.getValue());
+            rewriter.create<cf::AssertOp>(loc, compareLowerBound,
```
AlexandreEichenberger, line 117: same here, can use `andi`.

AlexandreEichenberger (approved):
LGTM, since you have to merge current changes anyway, please `&&` the two conditions for a slightly faster test.

Thanks for the quick turnaround on this.

src/Conversion/ONNXToKrnl/Tensor/Gather.cpp
chentong319: Separate the check to provide more accurate error message.

PR #3070, opened 2025-02-07
fix weak hash

Sunny-Anand:
This updates the sha algorithm for the docker images from sha1 to sha3_256. I have avoided renaming the use of sha1 in various docker files for now with something as generic as hash, but if there is feedback to change those variable names from tied to an algorithm to being generic, I can do that as feedback.


AlexandreEichenberger (approved):
LGTM

PR #3073, opened 2025-02-12
Since compiler generated stick/unstick is default on, change new option to disable it

AlexandreEichenberger:
Change option to be more in-line with default, where options that are default on are explicitly disabled using an appropriate `--disable-XXX` flag


tungld (approved):
LGTM!

PR #3074, opened 2025-02-12
ZHigh to ONNX optimization is default on. Switch flag from enable to disable

AlexandreEichenberger:
ZHigh to ONNX is default on, and should rarely be disabled. Having a compiler flag to explicitly disable it is more intuitive


tungld (approved):
LGTM!

src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
```diff
@@ -183,7 +183,7 @@ void addONNXToZHighPasses(mlir::PassManager &pm) {
   // sub, ...) that are of `stick -> light-weight op -> unstick`, it's better to
   // use CPU instead of NNPA to avoid stick/unstick. CPU is efficient to handle
   // these ops, e.g vectorize the computation.
-  if (nnpaEnableZHighToOnnx)
+  if (! nnpaDisableZHighToOnnx)
```
tungld, line 186: Looks like the space caused `clang-format check` failed.

AlexandreEichenberger: Tx

PR #3076, opened 2025-02-13
Add lit tests for KrnlMatmulOp lowering (Krnl to affine)

AlexandreEichenberger:
Add requested lit tests


tungld (approved):
LGTM! 

Thanks for adding these!!!

PR #3077, opened 2025-02-14
Handle out-of-bound value for Gather alike operation

chentong319:
Current situation: when a value of indices is out-of-bound, the execution is halted with assertion error.
This PR provides some improvements:
1. Do not generate assertion error. Instead, print out error message and replace with out-of-bound value with zero, with the hope that the execution will finish gracefully.
2. Print the out-of-bound value.
3. Share the code for GatherOp and GatherElementsOp. I did not define a common utility function because there are too many parameters. In future, we can merge all the lowering code, not just for enable-safe-code-gen, for the two ops.

### DIFF
--- a/src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
+++ b/src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
@@ -180,7 +180,7 @@ std::map<int64_t, int64_t> getReductionMapping(
   }
 
   for (decltype(inRank) inIndex = 0, outIndex = 0; inIndex < inRank;
-      ++inIndex) {
+       ++inIndex) {
     // If it is a reduction axis, there is no relationship among dimensions.
     if (isReductionAxis[inIndex]) {
       if (keepdims)
--- a/src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
+++ b/src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
@@ -925,7 +925,9 @@ void genSafeCodeForGatherAlike(mlir::ConversionPatternRewriter &rewriter,
         IndexExpr index = NonAffineIndexExpr(indexVal);
 
         // index should be in range of [-r, r-1], where r = dim size of
-        // data[axis]
+        // data[axis].
+        // Assume that the index is loaded from tensor with negative value
+        // correction.
         Value errorCondition =
             ((index < (-1) * axisDim) | (index >= axisDim)).getValue();
         rewriter.create<scf::IfOp>(
--- a/src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
+++ b/src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
@@ -14,10 +14,12 @@
 //===----------------------------------------------------------------------===//
 
 #include "src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.hpp"
+#include "mlir/Dialect/SCF/IR/SCF.h"
 #include "mlir/Dialect/Utils/ReshapeOpsUtils.h"
 #include "mlir/IR/BuiltinTypeInterfaces.h"
 
 #include "src/Accelerators/Accelerator.hpp"
+#include "src/Compiler/CompilerOptions.hpp"
 #include "src/Dialect/Krnl/DialectBuilder.hpp"
 #include "src/Dialect/Mlir/DialectBuilder.hpp"
 #include "src/Dialect/ONNX/ONNXOps/OpHelper.hpp"
@@ -178,7 +180,7 @@ std::map<int64_t, int64_t> getReductionMapping(
   }
 
   for (decltype(inRank) inIndex = 0, outIndex = 0; inIndex < inRank;
-       ++inIndex) {
+      ++inIndex) {
     // If it is a reduction axis, there is no relationship among dimensions.
     if (isReductionAxis[inIndex]) {
       if (keepdims)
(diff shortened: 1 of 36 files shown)


src/Conversion/ONNXToKrnl/Tensor/Gather.cpp
```diff
@@ -142,24 +141,36 @@ struct ONNXGatherOpLowering : public OpConversionPattern<ONNXGatherOp> {
...
+            Value compareUpperBound =
+                create.math.sge(index.getValue(), upperBound);
             Value compareLowerBound =
-                create.math.sge(index.getValue(), zeroIE.getValue());
-            rewriter.create<cf::AssertOp>(loc, compareLowerBound,
-                nodeNameStr +
-                    " indices of GatherOp is less than the lower bound");
+                create.math.slt(index.getValue(), zeroIE.getValue());
+            Value outBound =
+                create.math.ori(compareUpperBound, compareLowerBound);
```
AlexandreEichenberger, line 159: please add `SCFBuilder` to create object, and then you can use 
```
  void ifThenElse(mlir::Value cond, SCFThenElseBodyFn thenFn,
      SCFThenElseBodyFn elseFn = nullptr) const;
```
to be in line with current practices.

src/Conversion/ONNXToKrnl/Tensor/Gather.cpp
```diff
@@ -142,24 +141,36 @@ struct ONNXGatherOpLowering : public OpConversionPattern<ONNXGatherOp> {
...
+                              "The out-of-bound indices value is: ";
+            MultiDialectBuilder<KrnlBuilder, MathBuilder> create(rewriter, loc);
+            create.krnl.printf(msg, indexVal, true);
+            msg = "The out-of-bound index is replaced with zero.\n";
+            create.krnl.printf(msg);
+            rewriter.setInsertionPointAfter(ifOp);
+            // The modification of index could be put into IfOp to save
+            // some condition check, but the code will become complicated.
+            index = index.selectOrSelf(index < zeroIE, zeroIE);
+            index = index.selectOrSelf(index >= axisDim, zeroIE);
```
AlexandreEichenberger, line 173: Why not capping at  `0` on the lower side and `axisDim` on the upper size?

src/Conversion/ONNXToKrnl/Tensor/Gather.cpp
```diff
@@ -142,24 +141,36 @@ struct ONNXGatherOpLowering : public OpConversionPattern<ONNXGatherOp> {
...
             StringAttr nodeName =
                 op->getAttrOfType<mlir::StringAttr>("onnx_node_name");
             if (nodeName && !nodeName.getValue().empty()) {
-              nodeNameStr = nodeNameStr + nodeName.getValue().str();
+              nodeNameStr += nodeName.getValue().str();
             }
-            rewriter.create<cf::AssertOp>(loc, compareUpperBound,
-                nodeNameStr +
-                    " indices of GatherOp is larger than the upper bound");
+
```
AlexandreEichenberger, line 152: Since index is already an IndexExpr, and that we have the dim of axis in the loop bounds, you could use index express all the way through as it has overloaded operators.

```
  IndexExpr operator==(IndexExpr const b) const;
  IndexExpr operator==(int64_t const b) const;
  IndexExpr operator!=(IndexExpr const b) const;
  IndexExpr operator!=(int64_t const b) const;
  IndexExpr operator<=(IndexExpr const b) const;
  IndexExpr operator<=(int64_t const b) const;
  IndexExpr operator<(IndexExpr const b) const;
  IndexExpr operator<(int64_t const b) const;
  IndexExpr operator>=(IndexExpr const b) const;
  IndexExpr operator>=(int64_t const b) const;
  IndexExpr operator>(IndexExpr const b) const;
  IndexExpr operator>(int64_t const b) const;

  IndexExpr operator&(IndexExpr const b) const;
  IndexExpr operator|(IndexExpr const b) const;
  IndexExpr operator!() const;
```

AlexandreEichenberger (commented):
You could do a common function
```
generateDynBoundCheck0ToUb(rewriter, op, index, ub, message)
```
where you extract the message you want from the op, can append message, and generate the needed code.

src/Conversion/ONNXToKrnl/Tensor/Gather.cpp
chentong319: No particular reason. Just feel 0 is a safer choice.

chentong319: Modified

chentong319:
Rewrite most of the code:
1. Move the check and print message out of the lowering loop nest, into a separate loop nest.
2. The check and print code is put in a common function, which is used by both GatherOp and GatherElementOp.
3. Use IE for expression and lambda function for scf::IfOp
4. The clip of the value is still in the original lowering loop nest.
5. Test case added.

src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
```diff
@@ -882,4 +884,73 @@ void impl::onnxToKrnlSimdReport(Operation *op, bool successful,
...
+              if (nodeName && !nodeName.getValue().empty()) {
+                nodeNameStr += nodeName.getValue().str();
+              }
+              std::string msg = nodeNameStr +
+                                ": Value of indices is out of bound. " +
+                                "The out-of-bound indices value is: ";
+              create.krnl.printf(msg, indexVal, true);
+              msg = "The out-of-bound index is replaced with zero.\n";
+              create.krnl.printf(msg);
+              thenBuilder.create<scf::YieldOp>(thenLoc);
```
AlexandreEichenberger, line 950: I believe the yield is already generated by the SCF IfTHenElse; moreover, the else parameter defaults to nullptr, so that could also be skipped.

AlexandreEichenberger (approved):
LGTM, thanks for the updates

chentong319:
@jenkins test this please

chentong319:

PR #3078, opened 2025-02-14
Extend instrumentSignature to print data

chentong319:
Extend instrumentSignature to be a tool that can be used to print the data of a particular onnx op at runtime.
1. Change in the op selection. If the option instrumentSignature starts with "onnx_node_name:", only the rest of the string is used as regular expression pattern, and the attribute of "onnx_node_name", instead of op->getName(), is used to match. This change allows user to target a particular onnx op.
2. When the op is mapped with onnx_node_name, the data of the tensors will be printed. Therefore, one attribute is added to onnx.PrintSignature op.
3. Consequently, the lowering of PrintSignature op is modified.
4. Another change, which may not be needed, is that the instrumentSignature pass is moved from onnx to mlir group to onnx transformation group, right after the other onnx instrumentation. I feel it is better to put all the instrumentation passes for onnx are together. A side benefit is that we can easily check the instrumentation with "--EmitONNXIR".

Future work:
If we want to debug the crash inside the execution of an onnx op, there two changes are needed:
1. Move the print input before the onnx op.
2. The output should be easily used as input to run that op alone.

### DIFF
--- a/src/Compiler/CompilerOptions.cpp
+++ b/src/Compiler/CompilerOptions.cpp
@@ -72,6 +72,7 @@ std::string instrumentOps;                             // onnx-mlir only
 unsigned instrumentControlBits;                        // onnx-mlir only
 std::string parallelizeOps;                            // onnx-mlir only
 std::string instrumentSignatures;                      // onnx-mlir only
+std::string instrumentOnnxNode;                        // onnx-mlir only
 std::string ONNXOpStats;                               // onnx-mlir only
 int onnxOpTransformThreshold;                          // onnx-mlir only
 bool onnxOpTransformReport;                            // onnx-mlir only
@@ -503,7 +504,8 @@ static llvm::cl::opt<std::string, true> instrumentSignatureOpt(
     "instrument-signature",
     llvm::cl::desc(
         "Specify which high-level operations should be selected for printing\n"
-        "the type, shape, and data of their input/output tensors.\n"
+        "the type and shape of their input/output tensors.\n"
+        "The ops are selected by their op name.\n"
         "The instrument-signature defines the pattern to select the ops.\n"
         "\"NONE\" for no instrument (default).\n"
         "\"ALL\" or \"\" for all available operations.\n"
@@ -511,12 +513,31 @@ static llvm::cl::opt<std::string, true> instrumentSignatureOpt(
         "\"ops1,ops2, ...\" for the multiple ops.\n"
         "e.g. \"onnx.MatMul,onnx.Add\" for MatMul and Add op in onnx dialect.\n"
         "Asterisk is also available.\n"
-        "e.g. \"onnx.*\" for all onnx operations.\n"
-        "The string from op->getName() is used to match the regexp pattern.\n"
-        "Special case: if this option is started with \"onnx_node_name:\"\n"
-        "the attribute of \"onnx_node_name\", instead of the op->getName()\n"
-        "will be used to match the op, and the data value will be printed.\n"),
+        "e.g. \"onnx.*\" for all onnx operations.\n"),
     llvm::cl::location(instrumentSignatures), llvm::cl::init("NONE"),
+    llvm::cl::cat(OnnxMlirOptions));
+
+static llvm::cl::opt<std::string, true> instrumentONNXNodeOpt(
+    "instrument-onnx-node",
+    llvm::cl::desc(
+        "Specify which onnx operation node will be selected for \n"
+        "inserting a runtime call after the node to print the data of\n"
+        "their input/output tensors.\n"
+        "The ops are selected by their onnx node name, which is a string\n"
+        "attribute unique to each onnx node (most of time).\n"
+        "You can find them in the output of --EmitONNXIR\n"
+        "Other instrumentation in onnx-mlir is specified by op->getName(),\n"
+        "namely, the type of onnx operation, such Add, Matmul, and etc.\n"
+        "This option is able to pinpoint to a particular node.\n"
+        "The instrument-onnx-node defines the pattern to select.\n"
+        "\"NONE\" for no instrument (default).\n"
+        "Except for the special values, the regexp is used for matching.\n"
+        "\"/layer1/MatMul, onnx.Add_0, ...\" for the multiple nodes.\n"
+        "Asterisk is also available. For example:\n"
+        "\"onnx.Add_*\" for all AddOp. This feature allows you to specify\n"
+        "part of the target of onnx_node_name, as long as it is long enough\n"
+        "to be unique.\n"),
+    llvm::cl::location(instrumentOnnxNode), llvm::cl::init("NONE"),
     llvm::cl::cat(OnnxMlirOptions));
 
 static llvm::cl::opt<std::string, true> ONNXOpStatsOpt("onnx-op-stats",
--- a/src/Compiler/CompilerOptions.cpp
+++ b/src/Compiler/CompilerOptions.cpp
@@ -501,14 +501,18 @@ static llvm::cl::opt<std::string, true> parallelizeOpsOpt("parallelize-ops",
 
 static llvm::cl::opt<std::string, true> instrumentSignatureOpt(
     "instrument-signature",
-    llvm::cl::desc("Specify which high-level operations should print their"
-                   " input type(s) and shape(s)\n"
-                   "\"ALL\" or \"\" for all available operations.\n"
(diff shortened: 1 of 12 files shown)


src/Compiler/CompilerPasses.cpp
```diff
@@ -214,7 +214,7 @@ void addONNXToKrnlPasses(mlir::PassManager &pm, int optLevel, bool enableCSE,
 
 void addKrnlToAffinePasses(mlir::PassManager &pm) {
   pm.addNestedPass<func::FuncOp>(
-      onnx_mlir::krnl::createConvertKrnlToAffinePass(enableParallel));
```
AlexandreEichenberger, line 217: You lost some changes from dev main. Please add them back.

src/Conversion/ONNXToKrnl/Tensor/PrintSignature.cpp
```diff
@@ -50,15 +50,22 @@ struct ONNXPrintSignatureLowering
...
+    std::string printControl = "%e";
+    if (printSignatureOp.getPrintData() == 1) {
+      // The data of tensor will be printed
+      printControl = "%d";
+    }
     Value lastVal = printVal.pop_back_val();
     // Print all but the last one.
     for (Value oper : printVal) {
-      create.krnl.printTensor(msg + ", %t%e", oper);
+      create.krnl.printTensor(msg + ", %t" + printControl, oper);
```
AlexandreEichenberger, line 63: I believe the "%e" is still recommended for the end of the string. This is because we sometimes have issues with the `\0` end of string (never successfully debugged), so we recommend to add an explicit end of string.

src/Conversion/ONNXToKrnl/Tensor/PrintSignature.cpp
```diff
@@ -50,15 +50,22 @@ struct ONNXPrintSignatureLowering
...
       msg = "%i";
     }
     // Print the last one with replace with new op.
     rewriter.replaceOpWithNewOp<KrnlPrintTensorOp>(
-        op, msg + ", %t\n%e", lastVal);
+        op, msg + ", %t\n" + printControl, lastVal);
```
AlexandreEichenberger, line 68: same here, add the "%e" please

src/Dialect/ONNX/AdditionalONNXOps.td
```diff
@@ -322,15 +322,19 @@ def ONNXNoneOp : ONNX_Op<"NoValue", [ConstantLike, Pure]> {
 //===----------------------------------------------------------------------===//
 // ONNX PrintSignatureOp.
 def ONNXPrintSignatureOp:ONNX_Op<"PrintSignature", []> {
-  let summary = "ONNX Op to print type signature of its input operands";
+  let summary = "ONNX Op to print type signature or data of its input operands";
```
AlexandreEichenberger, line 325: please run `make onnx-mlir-docs` to also update the documentation accordingly.

src/Compiler/CompilerOptions.cpp
```diff
@@ -501,14 +501,18 @@ static llvm::cl::opt<std::string, true> parallelizeOpsOpt("parallelize-ops",
...
+        "Specify which high-level operations should print their"
+        " input type(s) and shape(s)\n"
+        "\"ALL\" or \"\" for all available operations.\n"
+        "\"NONE\" for no instrument (default).\n"
+        "\"ops1,ops2, ...\" for the multiple ops.\n"
+        "e.g. \"onnx.MatMul,onnx.Add\" for MatMul and Add ops.\n"
+        "Asterisk is also available.\n"
+        "e.g. \"onnx.*\" for all onnx operations.\n"
+        "If this option is started with \"onnx_node_name\"\n"
+        "the attribute of \"onnx_node_name\", instead of the op name\n"
```
AlexandreEichenberger, line 514: missing the `:` in the pattern, as you are looking for this
```
std::string header = "onnx_node_name:";
```

Also you state that "the data values" will be printed. Is that the input data values, the output data values? Please specify.

src/Dialect/ONNX/Transforms/InstrumentONNXSignaturePass.cpp
```diff
@@ -72,15 +72,34 @@ class InstrumentONNXSignaturePass
   void runOnOperation() override {
     onnx_mlir::EnableByRegexOption traceSpecificOpPattern(
         /*emptyIsNone*/ false);
+    // If the signaturePattern is started with "onnx_node_name: ",
+    // the node will be selected by the string attribute of onnx_node_name.
+    std::string header = "onnx_node_name:";
+    std::cout << signaturePattern << "\n";
+    bool useNodeName = false;
+    if (signaturePattern.rfind(header, 0) == 0) {
```
AlexandreEichenberger, line 80: Can you explain the behavior here? If there is one match of "onnx_node_name", then all we use the node name for all ops?

src/Compiler/CompilerPasses.cpp
chentong319: Fixed.

src/Conversion/ONNXToKrnl/Tensor/PrintSignature.cpp
chentong319: Fixed

src/Conversion/ONNXToKrnl/Tensor/PrintSignature.cpp
chentong319: Fixed

src/Dialect/ONNX/AdditionalONNXOps.td
chentong319: Done.

src/Compiler/CompilerOptions.cpp
chentong319: Fixed.

src/Dialect/ONNX/Transforms/InstrumentONNXSignaturePass.cpp
chentong319: When the option starts with `onnx_node_name:`, we use attribute("onnx_node_name"), instead of op->getName(), to check all the op.
I am not sure what you mean by node name. I modified CompilerOption.cpp for this.

AlexandreEichenberger: I believe the code is not doing what you think it is doing.

if the pattern passed is "onnx.Add, onnx_node_name:one_specific_op", then the `rfind` will return true and thus `useNodeName` is set to true. Then for every op in the walk, we will fetch the data from the `onnx_node_name` for alll.

I think what you need to do is:
1. use the old code to match with an op. if there is a match, then do the old print.
2. else search the onnx_node_name attribute, if non null, prefix the attribute with "onnx_node_name:", search. If hit, then do the new print including the data.

chentong319: > Then for every op in the walk, we will fetch the data from the onnx_node_name for all.

Yes, that's what I want. We could separate the pattern for onnx_node_name and getName. I did not do that.

chentong319:
About the changes:
1. Introduce a new option for instrumentation according the onnx_node_name. The original one is for opName, unchanged.
2. The two instrumentations are implemented in the same pass. User can use them in the same time if they like to.

src/Dialect/ONNX/Transforms/InstrumentONNXSignaturePass.cpp
```diff
@@ -71,30 +73,54 @@ class InstrumentONNXSignaturePass
...
+    onnx_mlir::EnableByRegexOption traceSpecificNodePattern(
         /*emptyIsNone*/ false);
+
     traceSpecificOpPattern.setRegexString(signaturePattern);
+    traceSpecificNodePattern.setRegexString(nodeNamePattern);
     // Iterate on the operations nested in this function.
     getOperation().walk([&](mlir::Operation *op) {
-      std::string opName = op->getName().getStringRef().str();
       auto dialect = op->getDialect();
+      Location loc = op->getLoc();
```
AlexandreEichenberger, line 86: Did not check the logic in great detail, I assume that you used a small example mlir file and tried both the signature and the new functionality and that you were happy with the outputs.

AlexandreEichenberger (approved):
LGTM
Maybe you could just print the output of a very small example  as comments just to show folks what the output looks like.

PR #3079, opened 2025-02-17
Make verifyInputTensor true by default

chentong319:
When the input to the model is not generated according to the model signature, the provided input may not match the signature and cause weird errors. Turning the check on by default will detect such errors. The overhead should be neglectable even for a toy model.

### DIFF
--- a/src/Compiler/CompilerOptions.cpp
+++ b/src/Compiler/CompilerOptions.cpp
@@ -577,8 +577,8 @@ static llvm::cl::opt<bool, true> verifyInputTensorsOpt("verifyInputTensors",
     llvm::cl::desc(
         "Verify input tensors whenever the entry point function is called.\n"
         "Data type and shape are verified. Enable this may introduce overhead "
-        "at runtime."),
-    llvm::cl::location(verifyInputTensors), llvm::cl::init(false),
+        "at runtime. Default is true"),
+    llvm::cl::location(verifyInputTensors), llvm::cl::init(true),
     llvm::cl::cat(OnnxMlirOptions));
 
 static llvm::cl::opt<bool, true> allowSortingOpt("allowSorting",


tungld (approved):
LGTM!

chentong319:
@tungld Thanks for approval. But the test failed. I did some investigation. The reason is for dialect defined type: !onnx.String. I will save this to issue and fix it later.

chentong319:
@tungld I reopened this PR. The previous error was caused by check-onnx-backend-constant and test_constant_cpu in check-onnx-backend: there more input parameter provided than the function definition. One possible solution is to disable verifyInputTensor in backend test. Another one is to allow extra actual parameter provided. I tried the later and now the only error is the lit test for verify input (need to be updated)
 What's your opinion?

tungld:
> there more input parameter provided than the function definition.

Let me take a look at this. When testing constant inputs, the model was modified, e.g. arguments are removed, so the signature must reflect this change too to make it consistent.

chentong319:
For AMD build, check-onnx-backend-dynamic-jni and check-onnx-backend-numerical failed too. It seems that it's simpler to only enable verify input for check-onnx-backend, except for the test_constant_cpu test case.

tungld:
I created this PR #3344 to make verifyInputTensor work well in `check-onnx-backend-constant`.

test/backend/inference_backend.py
```diff
@@ -698,7 +698,8 @@ def get_test_models():
         # ==MIN== 1
         # ==LIM== Does not support int4 and uint4.
         # By def, no dynamic shapes.
-        "test_constant_cpu": {STATIC_SHAPE: {}},
+        # Commented out because the model has no input but one is provided
+        # "test_constant_cpu": {STATIC_SHAPE: {}},
```
tungld: This should work now since we can handle functions with no inputs.

src/Conversion/KrnlToLLVM/KrnlEntryPoint.cpp
```diff
@@ -428,6 +428,7 @@ class KrnlEntryPointOpLowering : public OpRewritePattern<KrnlEntryPointOp> {
     int64_t inputNum = JSONArray->size();
 
     // Verify the number of inputs.
+    // In some case, there unused parameters provided.
```
tungld: I think this is no longer true after PR #3344. If it exists, we would fix that issue.

test/backend/CMakeLists.txt
```diff
@@ -113,7 +113,7 @@ add_custom_target(check-onnx-backend
 add_custom_target(check-onnx-backend-dynamic
   COMMAND
     ONNX_HOME=${FILE_GENERATE_DIR}/check-onnx-backend-dynamic
-    TEST_DYNAMIC=true
+    TEST_DYNAMIC=true TEST_DISABLE_INPUT_CHECK=true
```
tungld: Is it for AMD machines? I tested on s390 and did not see any issue.

test/backend/inference_backend.py
chentong319: Yes, I reversed the code but forgot the comment.

test/backend/CMakeLists.txt
chentong319: I still see about 20 test cases failed for check-onnx-backend-dynamic in the PR build and my local z machine. I haven't figured out why. I added this flag to check whether there is any other failure. That's the only failure.
On my machine, I got the same error when running with check-onnx-backend-dynamic, but no error if I ran the .so built by backend test  with standalone script. That puzzled me. I think that we only modify the model, but not actual input for the dynamic. Right?

tungld: We removed actual inputs also. In the old code before PR #3344, ther was a line to add a pseudo numpy array of 0 values when there was no actual inputs so the compiled model then had one input, which caused our issues. I fixed that, and the actual input must be consistent with the signature now. Otherwise, there is some bug we need to find out.

tungld: Perhaps, you need to rebuild PyRuntimeC to use the new changes.

chentong319: It's related to line 493 in KrnlEntryPoint.cpp. I added a print of the actualDim value. I got this message for test_gather_0
```
Wrong size for the dimension 3 of the input 0: expect a non-negative value 2
```

It seems that the actualDim is 2. Why the runtime thought it not non-negative?
My change for debug:
```
                // Print an error message.
                std::string msg = "Wrong size for the dimension " +
                                  std::to_string(d) + " of the input " +
                                  std::to_string(i) +
                                  ": expect a non-negative value %d\n";
                StringRef errorMsg(msg);
                create.krnl.printf(errorMsg, actualDim);
```
I cannot figure out what's wrong with this IF creation. I suspect it is related to the sign of integer.
You may test with the main branch with `TEST_COMPILE_ARGS="--verifyInputTensors=true" TEST_CASE_BY_USER=test_gather_0_cpu cmake --build . --target check-onnx-backend-dynamic`

PR #3081, opened 2025-02-19
Add support to use either docker or local compiler to compile a model

chentong319:
Background: onnxmlirdocker.py supports compilation of model using compiler docker through python docker package.
This PR add more parameter so that user can specify the container and the compiler path. 
```
 sess = onnxmlir.InferenceSession("test_add.onnx", compiler_container=""ghcr.io/onnxmlir/onnx-mlir-dev")
```
This session will use this container to compile. This is the default value and the default compiler path is `/workdir/onnx-mlir/build/Debug/bin/onnx-mlir`

Another example:
```
sess = onnxmlir.InferenceSession("test_add.onnx",
    compiler_container=None, compiler_path="/workdir/onnx-mlir/build/Debug/bin/onnx-mlir")
```
This is an example to use the compiler provided by container when you are inside the container.


src/Runtime/python/onnxmlirdocker.py
```diff
@@ -29,70 +42,117 @@ def __init__(self, model_path, **kwargs):
...
+        if "compile_options" in kwargs.keys():
+            self.compile_options = kwargs["compile_options"]
         else:
             self.compile_options = ""
 
-        if "onnx-mlir-container" in kwargs.keys():
-            self.compiler_container = kwargs["onnx-mlir-container"]
+        if "compiler_container" in kwargs.keys():
+            self.compiler_container = kwargs["compiler_container"]
+            if "compiler_path" not in kwargs.keys():
```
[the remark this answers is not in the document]
AlexandreEichenberger: Eventually, I think you should rename `compiler_container` maybe as `docker_compiler` or `podman_compiler` and use this as needed.

src/Runtime/python/onnxmlirdocker.py
```diff
@@ -29,70 +42,117 @@ def __init__(self, model_path, **kwargs):
...
+            self.compiler_container = "onnxmlir/onnx-mlir-dev"
+            self.compiler_path = "/workdir/onnx-mlir/build/Debug/bin/onnx-mlir"
+            """
 
-        # Assume we are using onnx-mlir-dev container
-        command_str = "/workdir/onnx-mlir/build/Debug/bin/onnx-mlir"
-        absolute_path = os.path.abspath(self.model_path)
-        self.model_basename = os.path.basename(absolute_path)
-        self.model_dirname = os.path.dirname(absolute_path)
+        if "compiler_path" in kwargs.keys():
```
AlexandreEichenberger, line 77: 1. option `--docker-compiler[= name]`. When no `name`, use default docker, otherwise use indicated docker
2. option `--podman-compiler[= name]`. When no `name`, use default podman, otherwise use indicated podman
3. if we can detect whether a system has docker or podman, then maybe using a `--remote-compiler[= name]` would be a useful option so that the user does not need to know what is available on a given system.
4. option `--local-compiler[=path]`. When path given, use it; otherwise test ONNX_MLIR_HOME.

I feel strongly that the compiler path inside the docker/podman should be known to us and should not require user to provide it... they really don't want to have to know this. If you need to scan the name (e.g. if the docker name has "dev" in it, then use `.../Debug/bin/onnx-mlir` and otherwise use the release, that is ok. Ideally in a docker env, we should have only one compiler installed, so adding a soft-link of `bin` to `Debug/bin` or `Release/bin` should be ok, or just install it in a known location regardless of debug or not. 

I know this PR does not yet support podman, I feel that using the `docker-compiler` name already will just make it easier not to change the interface on our users

src/Runtime/python/onnxmlirdocker.py
```diff
@@ -29,70 +42,117 @@ def __init__(self, model_path, **kwargs):
...
-                    "bind": self.container_model_dirname,
-                    "mode": "rw",
-                },
-                self.output_dirname: {
-                    "bind": self.container_output_dirname,
-                    "mode": "rw",
+        if self.compiler_container is None:
+            subprocess.run(command_str.split(" "))
+        else:
+            import docker
```
AlexandreEichenberger, line 140: please put it in a `try` to get a better error message.

AlexandreEichenberger (commented):
I really like the direction that you are taking for this work.

src/Runtime/python/onnxmlirdocker.py
chentong319: Our onnx-mlir-dev container can be used with either docker or podman. That's why compiler_container is used. We just need another variable to specify how the container will be used, docker or podman.

The PR  has the default compiler path for the known container. We need compiler_path to support customer's container.

AlexandreEichenberger (approved):
LGTM, please get feedback from @Sunny-Anand as well.

Sunny-Anand (approved):
LGTM. One suggestion I would agree with @AlexandreEichenberger here is that we can rename the `compiler_container` arg to something more clear to differentiate as `remote_compiler_registry` and `compiler_path` to `local_compiler_path`.

PR #3088, opened 2025-02-26
update pybind11  to version 2.12.0

chentong319:
The old version of pybind11 is too old for python version 3.12. I got error when building onnx-mlir locally with python version 3.12.2.
The PR update pybind11 to version 2.12.0, which is the version used by onnx.
With this PR, I am able to run inference correctly.


gongsu832 (approved):
LGTM

PR #3089, opened 2025-03-05
Improve scripts

AlexandreEichenberger:
Added options to CheckONNXModel.py to use the same input ref options as RunONNXModel.py

Used it to check `tabbert` which needs a python script for its inputs, as shown below.

```
CheckONNXModel.py -m prep_tabbert_rnn32.onnx -r="-O3 -march=arch15 -maccel=NNPA" -a="--nnpa-placement-heuristic=FasterOps" --load-ref-from-numpy=prep_tabbert.py
```


chentong319 (approved):
LGTM!
In future, CheckONNXModel.py may only need the option of -r and -a, and let RunONNXModel.py to parse the other parameters. Then we do not need to update CheckONNXModel.py when RunONNXModel.py changes it parameters.

AlexandreEichenberger:
Good suggestion @chentong319 , will do next time I modify this script.

PR #3091, opened 2025-03-07
Add result type inference to RandomNormalLike and fix wrong hardcodings for dtypes

jorickert:
Issue discovered by @sayeddla

### DIFF
--- a/test/mlir/onnx/onnx_shape_inference.mlir
+++ b/test/mlir/onnx/onnx_shape_inference.mlir
@@ -2899,12 +2899,12 @@ func.func @test_onehot_dynamic(%arg0: tensor<?x2xi64>, %arg1: tensor<i64>, %arg2
 
 // Test RandomNormal static
 
-func.func @test_random_normal_static_f16() -> tensor<*xf32> {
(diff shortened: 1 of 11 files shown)


test/mlir/onnx/onnx_shape_inference.mlir
```diff
@@ -2899,12 +2899,12 @@ func.func @test_onehot_dynamic(%arg0: tensor<?x2xi64>, %arg1: tensor<i64>, %arg2
 
 // Test RandomNormal static
 
-func.func @test_random_normal_static_f16() -> tensor<*xf32> {
```
jorickert, line 2902: In these tests I was not sure if the tests were intended to also check the conversion of the return type.
I changed them to have the same return type as the RandomNormal should have, because otherwise the input to the test is already invalid

sayeddla: I don't think these tests should have different/incorrect return types. I feel the return type inference should make sure we never run into a position where the return type is not correct

AlexandreEichenberger (approved):
LGTM, thanks for the update

PR #3094, opened 2025-03-11
Rewriting pattern to remove WhereOp and EqualOp. 

imaihal:
This PR remove ONNXWhereOp and ONNXEqualOp  into newly created ConcatOp.

```
    %0 = onnx.Constant dense<-1> : tensor<2xi64>
    %1 = onnx.Constant dense<1> : tensor<2xi64>
    %2 = onnx.Constant dense<-2> : tensor<1xi64>
    %3 = "onnx.Dim"(%arg0) {axis = 1 : si64} : (tensor<?x?xi64>) -> tensor<1xi64>
    %4 = "onnx.Concat"(%2, %3) {axis = 0 : si64} : (tensor<1xi64>, tensor<1xi64>) -> tensor<2xi64>
    %5 = "onnx.Equal"(%4, %0) : (tensor<2xi64>, tensor<2xi64>) -> tensor<2xi1>
    %6 = "onnx.Where"(%5, %1, %4) : (tensor<2xi1>, tensor<2xi64>, tensor<2xi64>) -> tensor<2xi64>
    onnx.Return %6 : tensor<2xi64>
```


src/Dialect/ONNX/ONNXOps/Canonicalize.cpp
```diff
@@ -220,6 +220,25 @@ bool isNegativeSplatConstant(Value val) {
...
+  //  if (auto disposable = mlir::dyn_cast<DisposableElementsAttr>(valAttr))
+  //    valAttr = disposable.toDenseElementsAttr();
+
+  Type elemTy = mlir::cast<ShapedType>(val.getType()).getElementType();
+  if (mlir::isa<IntegerType>(elemTy)) {
+    for (auto v : valAttr.getValues<APInt>()) {
+      if (v.getSExtValue() > 0)
+        return false;
+    }
+  }
```
tungld, line 238: It seems we need to return false for the `else` part. Otherwise, it returns true for the float type.

Also, I wonder should we do this check only if the number of values in `val` is small (typically < 10). Otherwise, this check runs for very large constants and it takes time.

src/Dialect/ONNX/ONNXOps/Canonicalize.cpp
```diff
@@ -1532,6 +1551,118 @@ struct PropagateBiasIntoLayerNormRewritePattern
...
+    // with scalar tensor
+    // - Operands in ONNXConcatOp need to be DimOp or ConstantOp
+    Value cond = onnxWhereOp.getCondition();
+    Value X = onnxWhereOp.getX();
+    Value Y = onnxWhereOp.getY();
+    if (mlir::isa<BlockArgument>(cond) || mlir::isa<BlockArgument>(X) ||
+        mlir::isa<BlockArgument>(Y))
+      return failure();
+
+    ONNXEqualOp equalOpCond = dyn_cast<ONNXEqualOp>(cond.getDefiningOp());
```
tungld, line 1583: Better to use `ONNXEqualOp equalOpCond = cond.getDefiningOp<ONNXEqualOp>();`

src/Dialect/ONNX/ONNXOps/Canonicalize.cpp
```diff
@@ -1532,6 +1551,118 @@ struct PropagateBiasIntoLayerNormRewritePattern
...
+        // The value defined by ConstantOp is compared with value B.
+        SmallVector<int64_t> aAttrValues;
+        if (!getI64ValuesFromONNXConstantOp(concatOperands[i], aAttrValues))
+          return failure();
+        int64_t a = aAttrValues.front();
+        int64_t b = bAttrValues[i];
+        if (a == b)
+          equalOpResults.emplace_back(true);
+        else
+          equalOpResults.emplace_back(false);
```
tungld, line 1637: for brevity:

```
equalOpResults.emplace_back(a==b);
```

src/Dialect/ONNX/ONNXOps/Canonicalize.cpp
AlexandreEichenberger: There are also functions that does many of the steps here.
```
template <typename OP>
bool operandOfOpDefinedBy(mlir::Operation *&matchOp, mlir::Operation *op,
    mlir::Value &matchOperand, int64_t matchThisOperandIndex = 0);
```

for a given `op`, test that operand `matchThisOperandIndex` is defined by an op of type `typename OP`. If that is so, set that defining op into `matchOp` and the operand into `matchOperand` and return true.

It makes the code more readable. 

There is also a call that looks for one match out of 2 in a binary operation, if that is a pattern that you are interested in.

tungld (approved):
LGTM! Thanks for reflecting the comments!

src/Dialect/ONNX/ONNXOps/Canonicalize.cpp
imaihal: Done. Thanks!

src/Dialect/ONNX/ONNXOps/Canonicalize.cpp
imaihal: Done. Thanks!

src/Dialect/ONNX/ONNXOps/Canonicalize.cpp
imaihal: Done. Thanks!

imaihal:
@tungld @AlexandreEichenberger Thanks for your reviews.

PR #3096, opened 2025-03-17
Set coordinate_transformation_mode=asymmetric when rewriting Upsample to Reshape to get same output.

jorickert:
Edit: Need to investigate why onnx-runtime and the backend tests have different results

Before the rewrite would rewrite to an reshape with the default `coordinate_transformation_mode = "half_pixel"`, but this leads to a different output when comparing with onnx runtime. 
 Tested using onnx runtime, with one model having a Reshape Op and one an Upsample. With  `coordinate_transformation_mode = "half_pixel"` the output had some differences, with `coordinate_transformation_mode = "asymmetric"` the output was the same.
 Comment in ORT about this: https://github.com/microsoft/onnxruntime/issues/1179#issuecomment-522197229
 
 IMO the onnx documentation for [Upsample](https://onnx.ai/onnx/operators/onnx__Upsample.html) is not very detailed and its hard to know how to do the Upsample -> Reshape rewrite when just looking at the documentation.

### DIFF
--- a/src/Dialect/ONNX/Transforms/Decompose.td
+++ b/src/Dialect/ONNX/Transforms/Decompose.td
@@ -239,7 +239,7 @@ def LogSoftmaxPattern
 def UpsamplePattern : Pat<
   (ONNXUpsampleOp $x, $scales, $mode),
   (ONNXResizeOp $x, (CreateNoneValue), $scales, (CreateNoneValue),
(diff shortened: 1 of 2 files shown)


jorickert:
Closing as their seems to be a mismatch between the onnx spec and onnx-runtime behavior

PR #3099, opened 2025-03-20
Including __errno_location call for MVS

christopherlmunoz:
Fixing errors when creating `libcruntime.a` on z/OS when the model includes constants.bin file.

 
```
IEW2456E 9207 SYMBOL __errno_location UNRESOLVED.  MEMBER COULD NOT BE INCLUDED
          FROM THE DESIGNATED CALL LIBRARY.
 IEW2665S 40FF MODULE *NULL*  IS NON-EXECUTABLE AND WAS NOT SAVED BECAUSE
          STORENX=NEVER.
```

### DIFF
--- a/src/Runtime/OMTensor.inc
+++ b/src/Runtime/OMTensor.inc
@@ -48,9 +48,9 @@
 #include "src/Support/SmallFPConversion.h"
 #endif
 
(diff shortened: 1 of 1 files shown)


tungld (approved):
LGTM!

PR #3101, opened 2025-03-21
Enable NNPA saturation by default and change the option to --nnpa-disable-saturation

tungld:
This patch enables NNPA saturation by default to avoid `out of range` errors in running some models.
The performance penalty is quite small, for example, for roberta base: 

Without saturation:

```
  Statistics 1 (excluding warmup), min, 8.873799e-01, max, 9.775084e-01, mean, 9.113163e-01, stdev, 1.960276e-02
  Statistics 2 (no warmup/quart.), min, 8.993821e-01, max, 9.181984e-01, mean, 9.091503e-01, stdev, 5.989598e-03
```

With saturation:
```
  Statistics 1 (excluding warmup), min, 8.923542e-01, max, 9.461253e-01, mean, 9.196692e-01, stdev, 1.480647e-02
  Statistics 2 (no warmup/quart.), min, 9.126368e-01, max, 9.281075e-01, mean, 9.210191e-01, stdev, 5.047690e-03
```


chentong319 (commented):
What if we change the attribute from "saturation" to "nonsaturation"? Will that save some modifications in the test case and code gen?

tungld:
> What if we change the attribute from "saturation" to "nonsaturation"? Will that save some modifications in the test case and code gen?

Looks a good idea. Let me give it a try. Thanks!

tungld:

Sunny-Anand:
@jenkins-droid  test this please

tungld:

tungld:
> What if we change the attribute from "saturation" to "nonsaturation"? Will that save some modifications in the test case and code gen?

@chentong319 I changed "saturation" to  nosaturation". Thanks!

PR #3103, opened 2025-03-25
removing weak attribute of errorno

christopherlmunoz:
When creating libcruntim.a on z/OS compiling with the latest version of xlc leads to this warning. 

```
xlc -I ../.. -I ../../include -q64 -Wc,DLL,EXPORTALL -qlanglvl=extc99 -D_XOPEN_SOURCE=600 -c -o OMTensor.o OMTensor.c

WARNING CCN3943 ./OMTensor.inc:54    Attribute "__weak__" is not supported on the target platform and is ignored.
INFORMATIONAL CCN4108 ./OMTensor.inc:54    The use of keyword '__attribute__' is non-portable.
```


christopherlmunoz:
Another solution attempt would be to have 2 separate OS checks, one for macOS with the weak attribute and mvs without.

tungld (approved):
LGTM!

PR #3104, opened 2025-03-26
Fix the custom build link for docs/Docker.md

qjivy:
The link should point to the correct item.


tungld (approved):
LGTM! Thank you for the contribution!

tungld:
@qjivy could you take a look at DCO failure and fix it, so that it can be merged. Usually when you do `git commit`, you need to sign by using `-s`, i.e. `git commit -s`.

tungld:

qjivy:
> @qjivy could you take a look at DCO failure and fix it, so that it can be merged. Usually when you do `git commit`, you need to sign by using `-s`, i.e. `git commit -s`.

Thx! Done.

PR #3106, opened 2025-03-28
Followups for torch model driver

chentong319:
1. Add a document
2. Remove the dependencies to package onnxmlir. Include the python file directly into onnxmlirtorch.
3. Add an example of how to print out all the inputs of granite model run with generate()
4. Small fixes.


tungld (approved):
LGTM!

PR #3108, opened 2025-04-01
Improve the c++ driver build-run-onnx-lib.sh

chentong319:
This PR add the support of flexibility of input/output of c++ driver.
1. Input: Originally, the utils/build-run-onnx-lib.sh is required to be call from onnx-mlir/build. With this PR, user can specify the path of compiler with env ONNX_MLIR_HOME.
2. Output: originally, the binary is written to onnx-mlir/build/bin/run-onnx-lib. The PR allows user to specify the binary with the extra argument.

The old way of using build-run-onnx-lib.sh will not be affected by this PR.


tungld (approved):
LGTM!

PR #3112, opened 2025-04-08
[NNPA] Fix an error in ZHighConstantPropagation for QuantizedStick

tungld:
Fix the following error when compiling a static quantization bert model for NNPA:

```
/build/llvm-project/mlir/lib/IR/PatternMatch.cpp:162: virtual void mlir::RewriterBase::eraseOp(Operation *): Assertion `op->use_empty() && "expected 'op' to have no uses"' failed.
```

The error happened because the scale and offset tensors from the inputs `zhigh.QuantizedStick` are bypassed as the outputs. So they must be checked to make sure they are safe to be disposed by constant propagation or not.


Sunny-Anand (approved):
LGTM

PR #3113, opened 2025-04-08
Add z17 for -march

chentong319:
Simple change: convert `--march=z17` to `--march=arch15`. The original `arch15` remains the same, while `z17` behaves like `arch15`.

Tested on z17 machine.

We may finally delete arch15 when everyone agree.


src/Compiler/CompilerOptions.cpp
```diff
@@ -1386,6 +1386,9 @@ void initCompilerConfig() {
     // Fast math option is enabled (in general)
     setLLVMOption(getLLVMOption() + " --enable-unsafe-fp-math");
   }
+
```
Sunny-Anand, line 1389: Probably need to add the below check in `src/Accelerators/NNPA/Support/NNPALimit.cpp` and avoid this extra check and assignment here. Referncing the old pr from Alex shared in slack, else we can have the bigger discussion later post release of 0.5.0.0 and cleanup there with all the references of arch15.
```
if (str[2] == '7') // Detect z17.
          return NNPALevel::M15;
```

Sunny-Anand (approved):
Thanks for the quick PR. LGTM

tungld:
> We may finally delete arch15 when everyone agree.

As I remember, we discussed about using `-march` instead of `-mcpu`, and values  like  `arch14`, `arch15` are more preferable since they are more consistent with clang and gcc. Using `-march=z16` for example is just for backward compatible where `-mcpu=z16` was used.

src/Compiler/CompilerOptions.cpp
chentong319: Fixed.

chentong319:
Okay, we will keep arch14, and arch15. It is really convenient for veteran users to use "z17".
By the way, do we want to support `-march=native`?

PR #3116, opened 2025-04-11
Combine Parallel Convolution optimization pass in ONNX Dialect

Arkar-Hema:
**CombineParallelConv2d** 

CombineParallelConv2D is an optimization pass designed to merge multiple parallel ONNXConvOp (Conv2D) operations into a single, more efficient convolution operation. This optimization enhances performance by reducing redundant computations and improving memory efficiency. 

The pass identifies Conv2D operations that: 
- Share the same input tensor. 
- Have identical kernel size, strides, padding, dilation, and group attributes (ensuring compatibility). 
- May have different numbers of output channels but maintain compatible shapes. 

Lets assume a input case:
- Input: (1, 3, 224, 224)
- Conv A: out_channels = 16
- Conv B: out_channels = 32
- Conv C: out_channels = 8
- Kernel size: (3, 3), stride: (1, 1)

Before Optimization (Three Parallel Conv2Ds)
Each convolution reads the entire input tensor and applies its own weight:
image
- Total MACs: ~25.2M
- Memory Reads: 3 times full input (one for each conv)
- Intermediate Buffers: 3 output tensors (N×16×224×224, N×32×224×224, N×8×224×224)
- Post-processing: A Concat(axis=1) op to merge all outputs → extra memory & ops

After Optimization (Combined Conv2D)
- Total MACs: ~25.2M (Same)
- out_channels = 16 + 32 + 8 = 56
- Input is read once
- Combined weight tensor of shape (56, 3, 3, 3)
- Optional bias is also concatenated

Improvement in performance metrics
- Latency Improvement: 10-20%
- Throughput Improvement: 7-30%
- Memory Usage Improvement: 20-35%


src/Dialect/ONNX/Transforms/Recompose.cpp
```diff
@@ -602,6 +651,176 @@ struct RecomposeQLinearMatMulFromQuantizeLinearPattern
...
+      if (auto conv = dyn_cast<ONNXConvOp>(user))
+        candidateConvs.push_back(conv);
+    }
+
+    // Must have at least two convs to combine.
+    if (candidateConvs.size() < 2)
+      return failure();
+
+    // Ensure all candidate convs are compatible (including bias check).
+    for (size_t i = 1; i < candidateConvs.size(); ++i) {
```
chentong319, line 676: Is it legal and profitable to fuse if there are at least two Convolution Ops can be combined?

Arkar-Hema: Yes, it is legal and profitable to perform the fusion when there are only two Conv ops.

I evaluated benchmark results on models with increasing Conv channels (32, 64, 128). The results show significant improvements after combining the Conv ops as it is reducing latency, memory usage and improves inference throughput, please take a look at it:

image
image
image

chentong319: I meant just merging two ConvOps. It seems to me that you requires to merge all the ConvOps on a tensor into one ConvOps.
I am thinking whether can you define the merge on just two ConvOps and then applying the rewriting pattern repeatedly to merge all the possible ConvOps.
Anyway the current PR is okay to me.

chentong319:
@jenkins-droid Please test it.

Arkar-Hema:
@chentong319 Thanks for the verification, could you please trigger the checks and merge the patch

PR #3117, opened 2025-04-11
Support QLinearMatMul on CPU

tungld:
This PR is to support QLinearMatMul on CPU. In particular it lowers ONNX.QLinearMatMul to Krnl dialect by rewriting the operation into ONNX.MatMul of i32 element type.

Currently only i8, ui8 and f32 data types are supported, and the backend tests for these data types are enabled.


src/Conversion/ONNXToKrnl/Math/QLinearMatMul.cpp
```diff
@@ -0,0 +1,149 @@
...
+    Value aScale = adaptor.getAScale();
+    Value aZeroPoint = adaptor.getAZeroPoint();
+    Value B = adaptor.getB();
+    Value bScale = adaptor.getBScale();
+    Value bZeroPoint = adaptor.getBZeroPoint();
+    Value yScale = adaptor.getYScale();
+    Value yZeroPoint = adaptor.getYZeroPoint();
+
+    // Only support integer8 and float32 now.
+    if (!getElementType(A.getType()).isInteger(8))
```
chentong319, line 52: integer8 for A and B, and flaot32 for zeroPoint?

tungld:
> To my understanding, the writing lowers the QLinearMatMul to a sequence of onnx op. Should such rewriting be put in decomposition pass?

It can be in the decomposition pass. However, we want to keep it as it is in case of NNPA (or other accelerators). So we need a way (it may be similar to turning off constant prop patterns) for turning off the decomposition when compiling for NNPA, which is currently not available. So that's why I decompose it during onnx-to-krnl.

src/Conversion/ONNXToKrnl/Math/QLinearMatMul.cpp
tungld: Yes, made it clear now. Thanks!

chentong319:
In general, onnx-mlir should provide a fine control on graph level optimization for different target architecture/accelerator. We can discuss such mechanism later.

src/Dialect/ONNX/ONNXOps/OpHelper.cpp
```diff
@@ -891,4 +891,27 @@ bool isElementAttrUninitializedDenseResource(ElementsAttr elementsAttr) {
          !denseResourceElementsAttr.getRawHandle().getBlob();
 }
 
+//===----------------------------------------------------------------------===//
+// Helper for quantization
+//===----------------------------------------------------------------------===//
+Value getOrCastToI8(
```
chentong319, line 897: Should this function be DialectBuilder?

AlexandreEichenberger (approved):
LGTM, still make "little" sense from a performance's perspective to run quantized models on CPU, compared to better supported float32. Still wonder if a warning would make sense.

src/Dialect/ONNX/ONNXOps/OpHelper.cpp
tungld: Yes, done. Thanks!

tungld:

PR #3122, opened 2025-04-16
Enhance shape inference for ONNX Reshape

tungld:
This patch is to enhance the shape inference for ONNX Reshape in the case where we can infer a static value for the output dimension at position of -1.

In Reshape, when the second input `shape` is a 1D tensor containing the dimensions of the output. `shape` can accept a value of -1, meaning that the output dimension is computed from the **remaining dimensions**.
In some situations, the remaining dimensions have unknown dimensions but **these unknown dimensions are the same as the the unknown dimensions in the input**. So we can remove these unknown dimensions when computing the dimension at -1, which increases the chance of getting a static value for position -1.

For example, in the following program, we can infer the dimension at position -1 (the second dim from the last of tensor<?x?x?x64xf32) to be `32` instead of unknown.

```mlir
  %1 = onnx.Constant dense<64> : tensor<1xi64>
  %2 = onnx.Constant dense<-1> : tensor<1xi64>
  %3 = "onnx.Dim"(%arg0) {axis = 0 : si64} : (tensor<?x?x2048xf32>) -> tensor<1xi64>
  %4 = "onnx.Dim"(%arg0) {axis = 1 : si64} : (tensor<?x?x2048xf32>) -> tensor<1xi64>
  %5 = "onnx.Concat"(%3, %4, %2, %1) {axis = 0 : si64} : (tensor<1xi64>, tensor<1xi64>, tensor<1xi64>, tensor<1xi64>) -> tensor<4xi64>
  %6 = "onnx.Reshape"(%arg0, %5) {allowzero = 0 : si64} : (tensor<?x?x2048xf32>, tensor<4xi64>) -> tensor<?x?x?x64xf32>
```

In this example, the output shape is `[%3, %4, -1, 64]`. Because `%3` and `%4` are the same as the 1st and 2nd dims of `%arg0` respectively, we can ignore them in computing dim at position -1. So the dim at position -1 is simple `2048/64` or `32`.


src/Dialect/ONNX/ONNXOps/Tensor/Reshape.cpp
```diff
@@ -48,12 +48,81 @@ LogicalResult ONNXReshapeOpShapeHelper::computeShape() {
...
+        outputIgnoredDims.clear();
+        dataIgnoredDims.clear();
+        break;
+      }
+      Value dim = shapeDimVals[i];
+      if (auto dimOp = dim.getDefiningOp<ONNXDimOp>()) {
+        if (dimOp.getData() != refData)
+          continue;
+        int64_t axis = dimOp.getAxis();
+        if (auto search = dataIgnoredDims.find(axis);
```
AlexandreEichenberger, line 106: A comment on the condition being tested would be useful.
Are you also testing the case where (though unlikely)

```
(%1, %2, 64) -> (%1, %1, %2, -1)
```
namely where `%1` appears twice, which would be clearly not bijective.

chentong319:
For readability, can we propagate 'onnx.name' attribute introduced by onnx model along onnx.dim op?

chentong319:
It seems to me that the code works on the restricted pattern with the data of reshape coming from MatMul. Could we relax the condition a little bit for a general case for data?
We need onnx.dim(%data, i) and value of %shape[i] to compute the size of unspecified dimension.
For the value of shape[I], we track back to the defining op, onnx.ConcatOp,  and then to its inputs of constant or onnx.dim op. That's the pattern appears in the model. Anyway, we may define a utility function with a general name in case we find more pattern to locate the definition from onnx.dim.
For the onnx.dim(%data, i), do we have to track back to its defining op or simply rely on DimAnalysis to check it with the onnx.dim from the value of shape?
Anyway, in the process of cancel out the same onnx.dim, I think that the canceled one should be removed from the list after matched. For example, in 'onnx.Reshape(%data, %reshape):  (tensor<?x?x?x256xf32>, tensor<4xi64>)`, there may be two symbols, one of which appears twice in the dimension.

src/Dialect/ONNX/ONNXOps/Tensor/Reshape.cpp
```diff
@@ -48,12 +48,81 @@ LogicalResult ONNXReshapeOpShapeHelper::computeShape() {
...
+    bool fromMatMul = false;
+    if (mmOp && isRankedShapedType(mmOp.getB().getType()) &&
+        getRank(mmOp.getB().getType()) == 2) {
+      refData = mmOp.getA();
+      fromMatMul = true;
+    }
+
+    bool isBijective = true;
+    for (int64_t i = 0; i < outputRank; ++i) {
+      if (!isBijective) {
```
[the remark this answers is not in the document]
tungld: Good catch! I fixed this case by moving the check outside the `for` loop.

src/Dialect/ONNX/ONNXOps/Tensor/Reshape.cpp
tungld: Added a lit test for this case.

tungld:
@chentong319 trying to understand your comment.

> For the onnx.dim(%data, i), do we have to track back to its defining op or simply rely on DimAnalysis to check it with the onnx.dim from the value of shape?

I think using DimAnalysis would be definitely better, so that we can check dim equality for two arbitrary dims without tracing back. Unfortunately, it is non-trivial to use DimAnalysis inside shape inference at this moment because 1) DimAnalysis uses ShapeHelper here so we may have a cyclic dependency, 2) DimAnalysis depends on the IR at a time, whenever the IR changes DimAnalysis needs to be updated.
 
> Anyway, in the process of cancel out the same onnx.dim, I think that the canceled one should be removed from the list after matched. For example, in 'onnx.Reshape(%data, %reshape): (tensor<?x?x?x256xf32>, tensor<4xi64>)`, there may be two symbols, one of which appears twice in the dimension.

Yes, it looks similar to Alex's example, I added a lit test for this case.

tungld:
> For readability, can we propagate 'onnx.dim_params' attribute introduced by onnx model along onnx.dim op?

It sounds a good thing. Will think about that.

tungld:
> tensor<?xf32> {onnx.dim_params = "0:dimB"}

Is it possible to define that attribute `{onnx.dim_params = "0:dimB"}` for a tensor? I know it's possible for function arguments, but not sure about a single tensor type.

PR #3123, opened 2025-04-16
Combine parallel dense Optimization pass in ONNX Dialect

Arkar-Hema:
**Combine Parallel Dense**

CombineParallelDense is an optimization pass designed to merge multiple parallel ONNXGemmOp (Dense/Fully Connected) operations into a single, more efficient Dense layer. This optimization reduces redundant computations, improves memory efficiency, and enhances hardware utilization. 

The pass identifies Dense (Gemm) operations that: 
- Share the same input tensor. 
- Have identical attributes such as alpha, beta, transA and transB (ensuring compatibility). 
- May have different output dimensions (number of neurons) but maintain compatible weight shapes for concatenation. 

Lets assume a input case:
- Input Shape: (1, 512)
- Dense A: out_features = 256
- Dense B: out_features = 128
- Dense C: out_features = 64
- Attributes: transB = 0, alpha = 1.0, beta = 1.0

Before Optimization (Three Parallel Gemms)
- Each GEMM does one full matrix multiplication (1×512 × 512×N)
- Three separate weight and bias tensors and produces three outputs
-Memory Reads: 3 times full input (one for each gemm)
- Post-processing: A Concat(axis=1) merges them into one output: Y (1×448)

After Optimization (Combined Dense)
- Total Output Features: 256 + 128 + 64 = 448
- All weights are concatenated along output channel axis → New weight shape: (512, 448)
- Biases are also concatenated
- A single ONNXGemmOp computes Y (1×448) directly

Improvement in performance metrics

Latency Improvement: 7-15%
Throughput Improvement: 8-14%
Memory Usage Improvement: 10-12%

### DIFF
--- a/src/Dialect/ONNX/Transforms/Recompose.cpp
+++ b/src/Dialect/ONNX/Transforms/Recompose.cpp
@@ -20,6 +20,8 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include <numeric>
+
 #include "mlir/IR/Matchers.h"
 #include "mlir/IR/PatternMatch.h"
 #include "mlir/Pass/Pass.h"
@@ -37,6 +39,53 @@
 #define DEBUG_TYPE "recompose"
 
 using namespace mlir;
+
+namespace onnx_mlir {
+// splits a tensor along a static axis into multiple outputs based on specified
+// channel sizes using the ONNX Split operation
+ValueRange emitSplitByChannels(PatternRewriter &rewriter, Location loc,
+                               Value input, ArrayRef<int64_t> splitSizes,
+                               int64_t axis) {
+
+  onnx_mlir::MultiDialectBuilder<onnx_mlir::OnnxBuilder> create(rewriter, loc);
+  ShapedType inputType = mlir::cast<ShapedType>(input.getType());
+  Type elementType = inputType.getElementType();
+  ArrayRef<int64_t> inputShape = inputType.getShape();
+
+  // Ensure the axis is within bounds and is a static dimension
+  assert(axis < static_cast<int64_t>(inputShape.size()) && axis >= 0 &&
+         "Axis out of bounds for input shape.");
+
+  assert(!inputType.isDynamicDim(axis) &&
+         "Channel dimension for input tensor must be static.");
+  // Validate split sizes
+  int64_t totalChannels = inputShape[axis];
+  int64_t sumSplitSizes =
+      std::accumulate(splitSizes.begin(), splitSizes.end(), 0);
+
+  assert(totalChannels == sumSplitSizes &&
+         "Split sizes must sum up to the total number of elements along the "
+         "axis.");
+
+  // Create Split Constant
+  Value splitConstant = create.onnx.constantInt64(splitSizes);
+
+  // Create output types for each split part
+  SmallVector<Type, 4> resultTypes;
+  for (int64_t size : splitSizes) {
+    SmallVector<int64_t> splitShape(inputShape.begin(), inputShape.end());
+    splitShape[axis] = size;
+    resultTypes.push_back(RankedTensorType::get(splitShape, elementType));
+  }
+  rewriter.setInsertionPointAfter(input.getDefiningOp());
+  // Perform Split Operation
+  ValueRange results =
+      create.onnx.split(ArrayRef(resultTypes), input, splitConstant, axis);
+
+  return results;
+}
+
+} // namespace onnx_mlir
 
 namespace {
 /// Include the patterns defined in the Declarative Rewrite framework.
@@ -602,6 +651,145 @@ struct RecomposeQLinearMatMulFromQuantizeLinearPattern
   }
 };
 
+struct CombineParallelDensePattern : public OpRewritePattern<ONNXGemmOp> {
+  using OpRewritePattern<ONNXGemmOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(ONNXGemmOp gemmOp1,
+                                PatternRewriter &rewriter) const final {
+    Value input = gemmOp1.getA();
+    if (!onnx_mlir::isRankedShapedType(input.getType()) ||
+        mlir::cast<ShapedType>(input.getType()).hasStaticShape() == false)
+      return failure();
+
+    SmallVector<ONNXGemmOp> parallelGemms = {gemmOp1};
+
+    for (auto user : input.getUsers()) {
+      ONNXGemmOp currentGemm = dyn_cast<ONNXGemmOp>(user);
+      if (currentGemm && currentGemm != gemmOp1 &&
+          areCompatible(gemmOp1, currentGemm)) {
+        parallelGemms.push_back(currentGemm);
+      }
+    }
+    if (parallelGemms.size() < 2)
+      return failure();
+
+    Location loc = gemmOp1.getLoc();
+    ShapedType inputType = mlir::cast<ShapedType>(input.getType());
+    Type elementType = inputType.getElementType();
+    onnx_mlir::MultiDialectBuilder<onnx_mlir::OnnxBuilder> create(rewriter,
+                                                                  loc);
+
+    // Identify axis dynamically based on Gemm shape consistency
+    ShapedType firstWeightType =
+        mlir::cast<ShapedType>(parallelGemms[0].getB().getType());
+    int64_t concatAxis =
+        gemmOp1.getTransB() ? 0 : firstWeightType.getRank() - 1;
+    int64_t Axis = firstWeightType.getRank() - 1;
+
+    // Concatenate weights
+    SmallVector<Value> weightValues;
+    SmallVector<int64_t> weightDims(firstWeightType.getShape());
+    int64_t totalOutputFeatures = 0;
+
+    for (auto gemm : parallelGemms) {
+      ShapedType weightType = mlir::cast<ShapedType>(gemm.getB().getType());
+      weightValues.push_back(gemm.getB());
+      totalOutputFeatures += weightType.getShape()[concatAxis];
+    }
+
+    weightDims[concatAxis] = totalOutputFeatures;
+    Type newWeightType = RankedTensorType::get(weightDims, elementType);
+    Value newWeight =
+        create.onnx.concat(newWeightType, weightValues, concatAxis);
+
+    // Concatenate biases (create zero constants for missing biases)
+    SmallVector<Value> biasValues;
+    for (auto gemm : parallelGemms) {
+      if (Value bias = gemm.getC()) {
+        biasValues.push_back(bias);
+      } else {
+        auto biasType =
+            RankedTensorType::get({totalOutputFeatures}, elementType);
+        Value zeroBias =
+            create.onnx.constant(DenseElementsAttr::get(biasType, 0.0));
+        biasValues.push_back(zeroBias);
+      }
+    }
+
+    SmallVector<int64_t> newBiasShape = {totalOutputFeatures};
+    Type newBiasType = RankedTensorType::get(newBiasShape, elementType);
+    Value newBias = create.onnx.concat(newBiasType, biasValues, 0);
+
+    // Create combined Gemm operation
+    ShapedType outputShape =
+        mlir::cast<ShapedType>(gemmOp1.getResult().getType()).getShape().vec();
+    outputShape[Axis] = totalOutputFeatures;
+    auto newOutputType = RankedTensorType::get(outputShape, elementType);
+
+    auto newGemm = rewriter.create<ONNXGemmOp>(
+        loc, newOutputType, input, newWeight, newBias, gemmOp1.getAlphaAttr(),
+        gemmOp1.getBetaAttr(), gemmOp1.getTransAAttr(),
+        gemmOp1.getTransBAttr());
+
+    // Check for common ConcatOp
+    ONNXConcatOp commonConcatOp = nullptr;
+    for (auto gemm : parallelGemms) {
+      for (auto user : gemm.getResult().getUsers()) {
+        if (auto concatOp = dyn_cast<ONNXConcatOp>(user)) {
+          if (!commonConcatOp) {
+            commonConcatOp = concatOp;
+          }
+          if (concatOp != commonConcatOp) {
+            commonConcatOp = nullptr;
+            break;
+          }
+        } else {
+          commonConcatOp = nullptr;
+          break;
+        }
+      }
+      if (!commonConcatOp) {
+        break;
+      }
+    }
+
+    if (commonConcatOp) {
+      commonConcatOp.getResult().replaceAllUsesWith(newGemm.getResult());
+      rewriter.eraseOp(commonConcatOp);
+    } else {
+      SmallVector<int64_t, 4> splitSizesVec;
+      for (auto gemm : parallelGemms) {
+        int64_t outputChannels =
+            mlir::cast<ShapedType>(gemm.getResult().getType()).getShape()[Axis];
+        splitSizesVec.push_back(outputChannels);
+      }
+
+      ArrayRef<int64_t> splitSizes(splitSizesVec);
+      ValueRange splitResults = onnx_mlir::emitSplitByChannels(
+          rewriter, loc, newGemm.getResult(), splitSizes, Axis);
+
+      for (size_t i = 0; i < parallelGemms.size(); ++i) {
+        parallelGemms[i].replaceAllUsesWith(splitResults[i]);
+      }
+    }
+
+    for (auto gemm : parallelGemms) {
+      rewriter.eraseOp(gemm);
+    }
+
+    return success();
+  }
+
+  static bool areCompatible(ONNXGemmOp a, ONNXGemmOp b) {
+    return a.getAlpha() == b.getAlpha() && a.getBeta() == b.getBeta() &&
+           a.getTransA() == b.getTransA() && a.getTransB() == b.getTransB() &&
+           mlir::cast<ShapedType>(a.getB().getType())
+                   .getShape()[a.getTransB() ? 1 : 0] ==
+               mlir::cast<ShapedType>(b.getB().getType())
+                   .getShape()[b.getTransB() ? 1 : 0];
+  }
+};
+
 struct RecomposeONNXToONNXPass
     : public PassWrapper<RecomposeONNXToONNXPass, OperationPass<func::FuncOp>> {
   MLIR_DEFINE_EXPLICIT_INTERNAL_INLINE_TYPE_ID(RecomposeONNXToONNXPass)
@@ -682,6 +870,7 @@ void onnx_mlir::getRecomposeONNXToONNXPatterns(
   patterns.insert<RecomposeGeluFromMulPattern>(context);
   patterns.insert<RecomposeLayerNormFromMulPattern>(context);
   patterns.insert<RecomposeQLinearMatMulFromQuantizeLinearPattern>(context);
+  patterns.insert<CombineParallelDensePattern>(context);
 }
 
 /*!
--- a/test/mlir/onnx/onnx_canonicalization.mlir
+++ b/test/mlir/onnx/onnx_canonicalization.mlir
@@ -482,6 +482,34 @@ func.func @test_size2(%arg0 : tensor<*xf32>) -> tensor<*xi64> {
 
 // -----
 
+func.func @consecutive_clips(%arg0: tensor<3x1024x1024xf32> {onnx.name = "input"}) -> (tensor<3x1024x1024xf32> {onnx.name = "output"}) {
+  %0 = onnx.Constant dense<-5.000000e-01> : tensor<f32>
+  %1 = onnx.Constant dense<5.000000e-01> : tensor<f32>
+  %2 = onnx.Constant dense<-3.000000e-01> : tensor<f32>
+  %3 = onnx.Constant dense<3.000000e-01> : tensor<f32>
+  %4 = "onnx.Clip"(%arg0, %0, %1) {onnx_node_name = "Clip_1"} : (tensor<3x1024x1024xf32>, tensor<f32>, tensor<f32>) -> tensor<3x1024x1024xf32>
+  %5 = "onnx.Clip"(%4, %2, %3) {onnx_node_name = "Clip_2"} : (tensor<3x1024x1024xf32>, tensor<f32>, tensor<f32>) -> tensor<3x1024x1024xf32>
+  onnx.Return %5 : tensor<3x1024x1024xf32>
+
+  // CHECK: module {
+  // CHECK: func.func @consecutive_clips(%[[ARG0:.*]]: tensor<3x1024x1024xf32> {{.*}}) -> (tensor<3x1024x1024xf32> {{.*}})
+  // CHECK: %[[VAR_1:.*]] = onnx.Constant dense<{{.*}}> : tensor<f32>
+  // CHECK: %[[VAR_2:.*]] = onnx.Constant dense<{{.*}}> : tensor<f32>
+  // CHECK: %[[VAR_3:.*]] = onnx.Constant dense<{{.*}}> : tensor<f32>
+  // CHECK: %[[VAR_4:.*]] = onnx.Constant dense<{{.*}}> : tensor<f32>
+  // CHECK: %[[CONST_MAX:.*]] = "onnx.Max"(%[[VAR_1]], %[[VAR_3]]) : (tensor<f32>, tensor<f32>) -> tensor<f32>
+  // CHECK: %[[CONST_MIN:.*]] = "onnx.Min"(%[[VAR_2]], %[[VAR_4]]) : (tensor<f32>, tensor<f32>) -> tensor<f32>
+
+  // CHECK: %[[FUSED_CLIP:.*]] = "onnx.Clip"(%[[ARG0]], %[[CONST_MAX]], %[[CONST_MIN]]) : (tensor<3x1024x1024xf32>, tensor<f32>, tensor<f32>) -> tensor<3x1024x1024xf32>
+
+  // CHECK: onnx.Return %[[FUSED_CLIP]] : tensor<3x1024x1024xf32>
+
+  // CHECK: }
+
+}
+
+// -----
+
 // COM: Test rewriting GlobalAveragePool into ReduceMeanV13
 func.func @test_global_average_pool(%arg0: tensor<1x3x5x5xf32>) -> tensor<1x3x1x1xf32> {
   %0 = "onnx.GlobalAveragePool"(%arg0) : (tensor<1x3x5x5xf32>) -> tensor<1x3x1x1xf32>


tungld:
@Arkar-Hema A general question: in what kind of models have you seen this kind of pattern: multiple Gemm ops followed by a Concat op? and also similar patterns you have recently created PRs for? Just curious on how practical it is. Thanks!

Arkar-Hema:
@tungld could you please verify this patch?

tungld:
@Arkar-Hema thank you for the information!

I have some general comments:
- I think that when multiple GEMM ops are followed by a concat, the performance in theory would be better. But, could you run with multiple input sizes to see how the performance benefit in practice?
- When multiple GEMM ops are NOT followed by a concat (this is the case for the models you listed), you need a split and I think the split axis is the innermost dimension. I am not sure how slow the split is and if we can get speedup or not. Could you do a performance comparison to see if you can achieve speedup in this case?
- Are you targeting optimization for CPU or it is beneficial for AI accelerators as well given that AI accelerators may use special data layout which may be not convenient for concat or split.

Thanks.

Arkar-Hema:
I ran performance benchmarks across a range of input sizes for both the GEMM → Concat and the Combined GEMM → Split cases. Results show that:
- In the Concat case, the optimization provides consistent Latency improvement of **2-7%,** and throughput improvement of **1-5%**
image
image
image
image


- In the cases where it splits, the optimization provides consistent Latency improvement of **1-7%,** and throughput improvement of **1-8%**
image
image
image
image


- I’ve currently targeted this pass primarily for CPU backends only.

src/Dialect/ONNX/Transforms/Recompose.cpp
```diff
@@ -602,6 +650,145 @@ struct RecomposeQLinearMatMulFromQuantizeLinearPattern
...
+    Type elementType = inputType.getElementType();
+    onnx_mlir::MultiDialectBuilder<onnx_mlir::OnnxBuilder> create(
+        rewriter, loc);
+
+    // Identify axis dynamically based on Gemm shape consistency
+    auto firstWeightType =
+        mlir::cast<ShapedType>(parallelGemms[0].getB().getType());
+    int64_t concatAxis =
+        gemmOp1.getTransB() ? 0 : firstWeightType.getRank() - 1;
+    int64_t Axis = firstWeightType.getRank() - 1;
```
tungld, line 696: Is it the axis for splitting? Is it always the last axis? If so, don't need to compute it, use `-1` for it.

src/Dialect/ONNX/Transforms/Recompose.cpp
```diff
@@ -602,6 +650,145 @@ struct RecomposeQLinearMatMulFromQuantizeLinearPattern
...
+    int64_t totalOutputFeatures = 0;
+
+    for (auto gemm : parallelGemms) {
+      ShapedType weightType = mlir::cast<ShapedType>(gemm.getB().getType());
+      weightValues.push_back(gemm.getB());
+      totalOutputFeatures += weightType.getShape()[concatAxis];
+    }
+
+    weightDims[concatAxis] = totalOutputFeatures;
+    Type newWeightType = RankedTensorType::get(weightDims, elementType);
```
tungld, line 710: No need to compute weightDims, use `UnRankedTensor` type and let the builder infer the correct shape. For example
```
Type newWeightType = UnRankedTensorType::get(elementType);
```

src/Dialect/ONNX/Transforms/Recompose.cpp
```diff
@@ -602,6 +650,145 @@ struct RecomposeQLinearMatMulFromQuantizeLinearPattern
...
+    auto firstWeightType =
+        mlir::cast<ShapedType>(parallelGemms[0].getB().getType());
+    int64_t concatAxis =
+        gemmOp1.getTransB() ? 0 : firstWeightType.getRank() - 1;
+    int64_t Axis = firstWeightType.getRank() - 1;
+
+    // Concatenate weights
+    SmallVector<Value> weightValues;
+    SmallVector<int64_t> weightDims(firstWeightType.getShape());
+    int64_t totalOutputFeatures = 0;
```
tungld, line 701: Just a naming issue. What are features in Gemm ops?

src/Dialect/ONNX/Transforms/Recompose.cpp
```diff
@@ -602,6 +650,145 @@ struct RecomposeQLinearMatMulFromQuantizeLinearPattern
...
+        create.onnx.concat(newWeightType, weightValues, concatAxis);
+
+    // Concatenate biases (create zero constants for missing biases)
+    SmallVector<Value> biasValues;
+    for (auto gemm : parallelGemms) {
+      if (Value bias = gemm.getC()) {
+        biasValues.push_back(bias);
+      } else {
+        auto biasType =
+            RankedTensorType::get({totalOutputFeatures}, elementType);
```
tungld, line 721: Could you explain why the shape of C here is `totalOutputFeatures` but not the dimension from B input of the same Gemm?
Also, the shape is incompatible if the other C inputs are not 1D tensor.

Please add a lit test for this case. Thanks!

src/Dialect/ONNX/Transforms/Recompose.cpp
```diff
@@ -602,6 +650,145 @@ struct RecomposeQLinearMatMulFromQuantizeLinearPattern
...
+
+    weightDims[concatAxis] = totalOutputFeatures;
+    Type newWeightType = RankedTensorType::get(weightDims, elementType);
+    Value newWeight =
+        create.onnx.concat(newWeightType, weightValues, concatAxis);
+
+    // Concatenate biases (create zero constants for missing biases)
+    SmallVector<Value> biasValues;
+    for (auto gemm : parallelGemms) {
+      if (Value bias = gemm.getC()) {
```
tungld, line 717: Are you sure this check works? If not, use  `!isNoneValue(gemm.getC())`

src/Dialect/ONNX/Transforms/Recompose.cpp
```diff
@@ -602,6 +650,145 @@ struct RecomposeQLinearMatMulFromQuantizeLinearPattern
...
+        auto biasType =
+            RankedTensorType::get({totalOutputFeatures}, elementType);
+        Value zeroBias =
+            create.onnx.constant(DenseElementsAttr::get(biasType, 0.0));
+        biasValues.push_back(zeroBias);
+      }
+    }
+
+    SmallVector<int64_t> newBiasShape = {totalOutputFeatures};
+    Type newBiasType = RankedTensorType::get(newBiasShape, elementType);
```
tungld, line 729: Use UnRankedTensorType please `Type newBiasType = UnRankedTensorType::get(elementType);`

src/Dialect/ONNX/Transforms/Recompose.cpp
```diff
@@ -602,6 +650,145 @@ struct RecomposeQLinearMatMulFromQuantizeLinearPattern
...
+    Value newBias = create.onnx.concat(newBiasType, biasValues, 0);
+
+    // Create combined Gemm operation
+    auto outputShape =
+        mlir::cast<ShapedType>(gemmOp1.getResult().getType()).getShape().vec();
+    outputShape[Axis] = totalOutputFeatures;
+    auto newOutputType = RankedTensorType::get(outputShape, elementType);
```
tungld, line 736: Use UnRankedTensorType please `Type newOutputType = UnRankedTensorType::get(elementType);`

src/Dialect/ONNX/Transforms/Recompose.cpp
```diff
@@ -602,6 +650,145 @@ struct RecomposeQLinearMatMulFromQuantizeLinearPattern
...
+
+    auto newGemm = rewriter.create<ONNXGemmOp>(loc, newOutputType, input,
+        newWeight, newBias, gemmOp1.getAlphaAttr(), gemmOp1.getBetaAttr(),
+        gemmOp1.getTransAAttr(), gemmOp1.getTransBAttr());
```
tungld, line 740: Could you add a builder for gemm into [ONNXDialectBuilder](https://github.com/onnx/onnx-mlir/blob/main/src/Dialect/ONNX/DialectBuilder.hpp)? Thanks!

src/Dialect/ONNX/Transforms/Recompose.cpp
```diff
@@ -602,6 +650,145 @@ struct RecomposeQLinearMatMulFromQuantizeLinearPattern
...
+
+    // Check for common ConcatOp
+    ONNXConcatOp commonConcatOp = nullptr;
+    for (auto gemm : parallelGemms) {
+      for (auto user : gemm.getResult().getUsers()) {
+        if (auto concatOp = dyn_cast<ONNXConcatOp>(user)) {
```
tungld, line 746: Don't you check the axis in this Concat is compatible with the axis for fusing Gemm ops?

I would not handle ONNXConcat here but writing another canonicalize rule for Concat that fuses `Split -> Concat` into an Identity op if their axes are consistent.

src/Dialect/ONNX/Transforms/Recompose.cpp
```diff
@@ -602,6 +650,145 @@ struct RecomposeQLinearMatMulFromQuantizeLinearPattern
...
+      ValueRange splitResults = onnx_mlir::emitSplitByChannels(
+          rewriter, loc, newGemm.getResult(), splitSizes, Axis);
+
+      for (size_t i = 0; i < parallelGemms.size(); ++i) {
+        parallelGemms[i].replaceAllUsesWith(splitResults[i]);
+      }
+    }
+
+    for (auto gemm : parallelGemms) {
+      rewriter.eraseOp(gemm);
```
tungld, line 785: Did you check that these Gemm ops must have one use before erasing? Otherwise, the generated IR would be broken.

test/mlir/onnx/onnx_recompose_combine_parallel_dense.mlir
```diff
@@ -0,0 +1,101 @@
...
+  %9 = "onnx.Gemm"(%arg0, %1, %5) {alpha = 1.000000e+00 : f32, beta = 1.000000e+00 : f32, onnx_node_name = "Gemm_2", transA = 0 : si64, transB = 0 : si64} : (tensor<1x4xf32>, tensor<4x3xf32>, tensor<3xf32>) -> tensor<1x3xf32>
+  %10 = "onnx.Gemm"(%arg0, %2, %6) {alpha = 1.000000e+00 : f32, beta = 1.000000e+00 : f32, onnx_node_name = "Gemm_3", transA = 0 : si64, transB = 0 : si64} : (tensor<1x4xf32>, tensor<4x3xf32>, tensor<3xf32>) -> tensor<1x3xf32>
+  %11 = "onnx.Gemm"(%arg0, %3, %7) {alpha = 1.000000e+00 : f32, beta = 1.000000e+00 : f32, onnx_node_name = "Gemm_4", transA = 0 : si64, transB = 0 : si64} : (tensor<1x4xf32>, tensor<4x3xf32>, tensor<3xf32>) -> tensor<1x3xf32>
+  %12 = "onnx.Relu"(%8) {onnx_node_name = "ReLU_1"} : (tensor<1x3xf32>) -> tensor<1x3xf32>
+  %13 = "onnx.Sigmoid"(%9) {onnx_node_name = "Sigmoid_2"} : (tensor<1x3xf32>) -> tensor<1x3xf32>
+  %14 = "onnx.Tanh"(%10) {onnx_node_name = "Tanh_3"} : (tensor<1x3xf32>) -> tensor<1x3xf32>
+  %15 = "onnx.LeakyRelu"(%11) {alpha = 0.00999999977 : f32, onnx_node_name = "LeakyReLU_4"} : (tensor<1x3xf32>) -> tensor<1x3xf32>
```
tungld, line 81: What did you check when using these ops Relu, Sigmoid, ...?

test/mlir/onnx/onnx_recompose_combine_parallel_dense.mlir
```diff
@@ -0,0 +1,101 @@
+// RUN: onnx-mlir  --useOnnxModelTypes=false --EmitONNXIR --printIR %s | FileCheck %s
+
+func.func @test_gemm_concat_simple(%arg0: tensor<1x4xf32>) -> tensor<1x6xf32> {
+  %0 = onnx.Constant dense<[[6.033820e-01, 0.874853491, 0.840596497],
```
tungld, line 4: Since we do not check the output accuracy, it's better to use splat constants to make the program clean, e.g. `%0 = onnx.Constant dense<5.5>: tensor<4x3xf32>`

tungld (commented):
Thanks @Arkar-Hema for the experiments! Did you compile your programs with -O3?

Since this parallel fusion may not work for accelerators, could you create a compile option to enable this if needed, for example `-fuse-parallel-onnx-gemm`?

I don't think you need to handle the case where there is a concat after multiple gemms. Just emit a split op, then later you can write a simple canonicalization rule for concat to fuse `Split -> Concat`.

Below are my first-round comments, most of them are for simplifying the code, making it easy to follow. However, the important thing is you need to check the input C carefully because it's broadcastable.

src/Dialect/ONNX/Transforms/Recompose.cpp
```diff
@@ -602,6 +650,145 @@ struct RecomposeQLinearMatMulFromQuantizeLinearPattern
...
+           mlir::cast<ShapedType>(a.getB().getType())
+                   .getShape()[a.getTransB() ? 1 : 0] ==
+               mlir::cast<ShapedType>(b.getB().getType())
+                   .getShape()[b.getTransB() ? 1 : 0];
+  }
+
+  LogicalResult matchAndRewrite(
+      ONNXGemmOp gemmOp1, PatternRewriter &rewriter) const final {
+    Value input = gemmOp1.getA();
+    if (!onnx_mlir::isRankedShapedType(input.getType()) ||
```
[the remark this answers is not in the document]
Arkar-Hema: Yes, it is required

src/Dialect/ONNX/Transforms/Recompose.cpp
```diff
@@ -602,6 +650,145 @@ struct RecomposeQLinearMatMulFromQuantizeLinearPattern
...
+        parallelGemms.push_back(currentGemm);
+      }
+    }
+    if (parallelGemms.size() < 2)
+      return failure();
+
+    Location loc = gemmOp1.getLoc();
+    ShapedType inputType = mlir::cast<ShapedType>(input.getType());
+    Type elementType = inputType.getElementType();
+    onnx_mlir::MultiDialectBuilder<onnx_mlir::OnnxBuilder> create(
```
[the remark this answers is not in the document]
Arkar-Hema: Yes, it is required

test/mlir/onnx/onnx_recompose_combine_parallel_dense.mlir
Arkar-Hema: I included different activation ops (Relu, Sigmoid, Tanh, LeakyRelu) in this test to validate that after combining the parallel Gemm operations and applying the Split, the split outputs are correctly ordered and preserve the expected tensor values per branch.

src/Dialect/ONNX/Transforms/Recompose.cpp
Arkar-Hema: I understand the suggestion to move the Concat handling into a canonicalize rule. However, I chose to handle it within the CombineParallelDensePattern pass because the parallel Gemm ops and their subsequent Concat are tightly coupled in this context — the Concat directly consumes the outputs of the combined Gemms, and its presence affects how we need to split the fused Gemm result afterward.

By handling both the combination of parallel Gemms and the subsequent Concat in a single pass, it ensures that the structural transformation is applied atomically and consistently. This avoids generating intermediate IR states where partially combined Gemms exist alongside their original Concat, which could lead to IR inconsistencies or redundant canonicalization work later.

That said, I do see the value in isolating local simplifications into canonicalize rules, and if this pattern were likely to occur independently of a preceding CombineParallelGemm, a separate canonicalization rule would make perfect sense. In this case, since the combined Gemm result’s splitting and the handling of its consumers are part of a tightly coupled optimization, keeping them together within the combine pass ensures correctness and maintains optimization intent.

src/Dialect/ONNX/Transforms/Recompose.cpp
Arkar-Hema: Operations like emitSplitByChannels require a known ranked shape to safely access dimensions by axis. If we use UnrankedTensorType, inputShape.size() inside those utilities will be unknown, leading to assertion failures. Therefore, we manually compute the combined output shape by summing output channel dimensions of parallel Gemms, and create a RankedTensorType for the new Gemm result to ensure downstream correctness.

src/Dialect/ONNX/Transforms/Recompose.cpp
```diff
@@ -602,6 +650,145 @@ struct RecomposeQLinearMatMulFromQuantizeLinearPattern
...
+
+  // Helper function to check if an gemm is mergeable
+  static bool areCompatible(ONNXGemmOp a, ONNXGemmOp b) {
+    return a.getAlpha() == b.getAlpha() && a.getBeta() == b.getBeta() &&
+           a.getTransA() == b.getTransA() && a.getTransB() == b.getTransB() &&
+           mlir::cast<ShapedType>(a.getB().getType())
+                   .getShape()[a.getTransB() ? 1 : 0] ==
+               mlir::cast<ShapedType>(b.getB().getType())
+                   .getShape()[b.getTransB() ? 1 : 0];
+  }
```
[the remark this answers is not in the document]
Arkar-Hema: I’ve addressed this by enforcing that C in each Gemm must be either None or a 1D tensor. I’ve explicitly disallowed rank-0 (scalar) and higher-rank tensors, even if technically broadcastable. The reason is that while ONNX allows unidirectional broadcasting, mixing different shapes for C (like scalars, 1D, or higher-rank) could result in implicit, context-dependent broadcasts when concatenating the C values for the combined Gemm, which may lead to unintended shape promotions or alignment issues.

By allowing only 1D Cs (or none), we make sure all biases have a consistent and predictable shape (N,), which is directly broadcastable to (M, N) in the Gemm operation without ambiguity, preserving correctness in both the combined Gemm and its Split outputs.

Arkar-Hema:
> Thanks @Arkar-Hema for the experiments! Did you compile your programs with -O3?
> 
> Since this parallel fusion may not work for accelerators, could you create a compile option to enable this if needed, for example `-fuse-parallel-onnx-gemm`?
> 
> I don't think you need to handle the case where there is a concat after multiple gemms. Just emit a split op, then later you can write a simple canonicalization rule for concat to fuse `Split -> Concat`.
> 
> Below are my first-round comments, most of them are for simplifying the code, making it easy to follow. However, the important thing is you need to check the input C carefully because it's broadcastable.

I have added it, Thanks

AlexandreEichenberger:

src/Dialect/ONNX/Transforms/Recompose.cpp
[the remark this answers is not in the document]
tungld: Thanks! There is a corner case as I commented above.

src/Dialect/ONNX/Transforms/Recompose.cpp
```diff
@@ -602,6 +651,163 @@ struct RecomposeQLinearMatMulFromQuantizeLinearPattern
...
+      int64_t outCh =
+          mlir::cast<ShapedType>(gemm.getResult().getType()).getShape()[Axis];
+      totalOutputChannels += outCh;
+    }
+    newOutputShape[Axis] = totalOutputChannels;
+    auto newOutputType = RankedTensorType::get(newOutputShape, elementType);
+
+    auto newGemm = rewriter.create<ONNXGemmOp>(loc, newOutputType, input,
+        newWeight, newBias, gemmOp1.getAlphaAttr(), gemmOp1.getBetaAttr(),
+        gemmOp1.getTransAAttr(), gemmOp1.getTransBAttr());
```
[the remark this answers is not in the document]
Arkar-Hema: I’ve used RankedTensorType for the new output type since we can fully determine the output shapes at compile-time. The input, combined weight, and optional C (restricted to None or 1D) make the resulting (M, total_N) shape for the combined Gemm output predictable. Using RankedTensorType here ensures stricter type checking and allows downstream passes to better reason about shapes for optimizations.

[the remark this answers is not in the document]
tungld: `create.onnx.gemm` computes the output shape for you at **compile time**, please do see its implementation you wrote:
```
Value OnnxBuilder::gemm(Type Y, Value A, Value B, Value C, FloatAttr alpha,
    FloatAttr beta, IntegerAttr transA, IntegerAttr transB) const {

  return createOpAndInferShapes<ONNXGemmOp>(
      toTensor(Y), A, B, C, alpha, beta, transA, transB);
}
```

Note the word *AndInferShapes*. We don't need to compute here to avoid boilerplate code.

AlexandreEichenberger:

src/Dialect/ONNX/Transforms/Recompose.cpp
```diff
@@ -602,6 +651,163 @@ struct RecomposeQLinearMatMulFromQuantizeLinearPattern
...
+      return false;
+
+    // Check C compatibility — only allow None or 1D
+    Value aC = a.getC();
+    Value bC = b.getC();
+    if (!onnx_mlir::isNoneValue(aC) && !onnx_mlir::isNoneValue(bC)) {
+      auto aCShape = mlir::cast<ShapedType>(aC.getType()).getShape();
+      auto bCShape = mlir::cast<ShapedType>(bC.getType()).getShape();
+      if (aCShape.size() != 1 || bCShape.size() != 1)
+        return false;
```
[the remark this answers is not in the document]
tungld: @Arkar-Hema please explain how did you solve this comment so that you marked it "solved"?

src/Dialect/ONNX/Transforms/Recompose.cpp
tungld: Have you tried to use `create.onnx.split` instead of your own `emitSplitByChannels` function. for exapmple, replace
```C

      auto newGemm = rewriter.create<ONNXGemmOp>(loc, newOutputType, input,
        newWeight, newBias, gemmOp1.getAlphaAttr(), gemmOp1.getBetaAttr(),
        gemmOp1.getTransAAttr(), gemmOp1.getTransBAttr());

      ArrayRef<int64_t> splitSizes(splitSizesVec);
      ValueRange splitResults = onnx_mlir::emitSplitByChannels(
          rewriter, loc, newGemm.getResult(), splitSizes, splitAxis);
```
by
```C
    Value newGemmOutput = create.onnx.gemm(input, newWeight, newBias, 
               gemmOp1.getAlphaAttr(), gemmOp1.getBetaAttr(), 
               gemmOp1.getTransAAttr(), gemmOp1.getTransBAttr());

    SmallVector<Type, 4> splitTypes(splitSizes.size(), mlir::UnrankedTensorType::get(elementType));
    ValueRange newOutputs = create.onnx.split(
        splitTypes, newGemmOutput, createonnx.constantInt64(splitSizes), splitAxis);
```
and does it work?

tungld:
Hi @Arkar-Hema When addressing a comment, could you please provide a brief explanation of how you did so? This will make the review process easier. Thanks!

src/Dialect/ONNX/Transforms/Recompose.cpp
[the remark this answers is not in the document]
Arkar-Hema: In the current implementation, I decided not to support the case where the bias shapes are different (e.g., tensor<1xf32> and tensor<5xf32>) since our concat operation would require explicit broadcasting to align the shapes before concatenation, and handling this properly would add additional complexity.

To handle this, I’ve added a check in areCompatible() to return false if the bias shapes differ in size when both biases are present - ensuring that we only merge Gemms where both bias tensors have the same shape. This preserves the correctness of the concat operation without requiring extra broadcasting logic.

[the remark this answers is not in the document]
tungld: Thank you!

I see you checked `aCShape[0] != bCShape[0]`. This is only valid if the shapes are static, e.g. `tensor<5xf32>`, but it does not work if the shapes are dynamic, e.g. both aC and bC have shape of `tensor<?xf32>`.

In the dynamic case, `aCShape[0] == bCShape[0]` during compile time, but at runtime, aC can be `tensor<1xf32` and bC can be `tensor<5xf32` for example.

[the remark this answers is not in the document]
tungld: For the static case, I wonder how you handle the following case where both aC and bC have shape of `tensor<1xf32>`. For example:
- gemm1: A: tensor<5x8x16xf32>, B: tensor<16x32xf32>, C: tensor<1xf32>
- gemm2: A: tensor<5x8x16xf32>, B: tensor<16x32xf32>, C: tensor<1xf32>

They satisfy your conditions here, so how do you combine them?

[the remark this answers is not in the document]
Arkar-Hema: Thanks for the thorough review and the great questions!

I’ve updated the areCompatible() function to properly handle the edge cases you pointed out:

Dynamic Bias Shapes:
If either of the bias tensors has a dynamic shape at dimension 0 (i.e., tensor<?xf32>), I now conservatively return false since we can’t guarantee at compile time whether they’ll match or require broadcasting at runtime.

Both Biases as tensor<1xf32>:
If both biases are of shape tensor<1xf32>, I now check their corresponding Gemm output shapes and ensure their output channels (last dimension) match before considering them compatible. If they differ, the function returns false, as merging them without this check would be invalid.

This ensures that both static and dynamic cases are handled correctly and conservatively avoids undefined behavior at runtime.

src/Dialect/ONNX/Transforms/Recompose.cpp
```diff
@@ -90,6 +91,183 @@ namespace {
...
+    // Concatenate weights
+    SmallVector<Value> weightValues;
+
+    for (auto gemm : parallelGemms) {
+      weightValues.push_back(gemm.getB());
+    }
+    Type unrankedTensorType = mlir::UnrankedTensorType::get(elementType);
+    Type newWeightType = unrankedTensorType;
+    Value newWeight =
+        create.onnx.concat(newWeightType, weightValues, concatWeightAxis);
```
tungld, line 176: Replace `newWeightType` by `unrankedTensorType `. It's redundant to define `newWeightType`.

src/Dialect/ONNX/Transforms/Recompose.cpp
```diff
@@ -90,6 +91,183 @@ namespace {
...
+        auto gemmShape =
+            mlir::cast<ShapedType>(gemm.getResult().getType()).getShape();
+        Value zeroBias = create.onnx.constant(DenseElementsAttr::get(
+            RankedTensorType::get({gemmShape[splitAxis]}, elementType), 0.0));
+        biasValues.push_back(zeroBias);
+      }
+    }
+
+    Type newBiasType = unrankedTensorType;
+    Value newBias = create.onnx.concat(newBiasType, biasValues, 0);
```
tungld, line 193: Replace `newBiasType ` by `unrankedTensorType `. It's redundant to define `newBiasType `.

src/Dialect/ONNX/Transforms/Recompose.cpp
```diff
@@ -90,6 +91,183 @@ namespace {
...
+      if (aCShape.size() != 1 || bCShape.size() != 1)
+        return false;
+      if (aCType.isDynamicDim(0) || bCType.isDynamicDim(0))
+        return false;
+      // check output channels match
+      if (aCShape[0] == 1 && bCShape[0] == 1) {
+        auto aOutputShape =
+            mlir::cast<ShapedType>(a.getResult().getType()).getShape();
+        auto bOutputShape =
+            mlir::cast<ShapedType>(b.getResult().getType()).getShape();
```
tungld, line 126: Replace these by:

```c
ArrayRef<int64_t> aOutputShape = getShape(a.getY().getType());
ArrayRef<int64_t> bOutputShape = getShape(b.getY().getType());
```

src/Dialect/ONNX/Transforms/Recompose.cpp
```diff
@@ -90,6 +91,183 @@ namespace {
...
+      int64_t outCh = mlir::cast<ShapedType>(gemm.getResult().getType())
+                          .getShape()[splitAxis];
+      totalOutputChannels += outCh;
+    }
+    newOutputShape[splitAxis] = totalOutputChannels;
+    auto newOutputType = RankedTensorType::get(newOutputShape, elementType);
+
+    auto newGemm = rewriter.create<ONNXGemmOp>(loc, newOutputType, input,
+        newWeight, newBias, gemmOp1.getAlphaAttr(), gemmOp1.getBetaAttr(),
+        gemmOp1.getTransAAttr(), gemmOp1.getTransBAttr());
```
tungld, line 212: Please replace this by 
```c
Value newGemmOutput = create.onnx.gemm(unrankedTensorType, input,
        newWeight, newBias, gemmOp1.getAlphaAttr(), gemmOp1.getBetaAttr(),
        gemmOp1.getTransAAttr(), gemmOp1.getTransBAttr());
```

src/Dialect/ONNX/Transforms/Recompose.cpp
```diff
@@ -90,6 +91,183 @@ namespace {
...
+    onnx_mlir::MultiDialectBuilder<onnx_mlir::OnnxBuilder> create(
+        rewriter, loc);
+
+    // Identify axis based on Gemm shape
+    int64_t concatWeightAxis = gemmOp1.getTransB() ? 0 : 1;
+    int64_t splitAxis = 1;
+
+    // Concatenate weights
+    SmallVector<Value> weightValues;
+
```
tungld, line 169: Redundant empty line?

src/Dialect/ONNX/Transforms/Recompose.cpp
```diff
@@ -90,6 +91,183 @@ namespace {
...
+        create.onnx.concat(newWeightType, weightValues, concatWeightAxis);
+
+    // Concatenate biases (create zero constants for missing biases)
+    SmallVector<Value> biasValues;
+    for (auto gemm : parallelGemms) {
+      if (!onnx_mlir::isNoneValue(gemm.getC())) {
+        biasValues.push_back(gemm.getC());
+      } else {
+        auto gemmShape =
+            mlir::cast<ShapedType>(gemm.getResult().getType()).getShape();
```
tungld, line 185: Please use: `ArrayRef<int64_t> gemmShape = getShape(gemm.getY().getType());`

src/Dialect/ONNX/Transforms/Recompose.cpp
```diff
@@ -90,6 +91,183 @@ namespace {
...
+      for (auto gemm : parallelGemms) {
+        int64_t outputChannels =
+            mlir::cast<ShapedType>(gemm.getResult().getType())
+                .getShape()[splitAxis];
+        splitSizesVec.push_back(outputChannels);
+      }
+
+      ArrayRef<int64_t> splitSizes(splitSizesVec);
+      ValueRange splitResults = onnx_mlir::emitSplitByChannels(
+          rewriter, loc, newGemm.getResult(), splitSizes, splitAxis);
```
tungld, line 254: Please replace this by
```c
    SmallVector<Type, 4> splitTypes(splitSizes.size(), unrankedTensorType);
    ValueRange splitResults = create.onnx.split(
        splitTypes, newGemmOutput, create.onnx.constantInt64(splitSizes), splitAxis);
```

src/Dialect/ONNX/Transforms/Recompose.cpp
```diff
@@ -90,6 +91,183 @@ namespace {
...
+    // Concatenate biases (create zero constants for missing biases)
+    SmallVector<Value> biasValues;
+    for (auto gemm : parallelGemms) {
+      if (!onnx_mlir::isNoneValue(gemm.getC())) {
+        biasValues.push_back(gemm.getC());
+      } else {
+        auto gemmShape =
+            mlir::cast<ShapedType>(gemm.getResult().getType()).getShape();
+        Value zeroBias = create.onnx.constant(DenseElementsAttr::get(
+            RankedTensorType::get({gemmShape[splitAxis]}, elementType), 0.0));
```
tungld, line 187: Check if `gemmShape[splitAxis]` is a static dimension in `areCompatible()` function. Otherwise, it fails to create a constant tensor here.

src/Dialect/ONNX/Transforms/Recompose.cpp
```diff
@@ -90,6 +91,183 @@ namespace {
...
+            mlir::cast<ShapedType>(b.getResult().getType()).getShape();
+        // Output channels is the last dim
+        if (aOutputShape.back() != bOutputShape.back())
+          return false;
+      }
+      // Otherwise, shapes must be equal
+      else if (aCShape[0] != bCShape[0])
+        return false;
+    }
+    return true;
```
tungld, line 135: When aC is None, do check that the last dim of aOutput is static. Otherwise, it fails when you create a constant tensor of zeros in the later code that you use the last dim of aOutput.

Check the same thing for bC.

Please add a list test for the case where aC or bC is None.

src/Dialect/ONNX/Transforms/Recompose.cpp
```diff
@@ -90,6 +91,183 @@ namespace {
...
+      totalOutputChannels += outCh;
+    }
+    newOutputShape[splitAxis] = totalOutputChannels;
+    auto newOutputType = RankedTensorType::get(newOutputShape, elementType);
+
+    auto newGemm = rewriter.create<ONNXGemmOp>(loc, newOutputType, input,
+        newWeight, newBias, gemmOp1.getAlphaAttr(), gemmOp1.getBetaAttr(),
+        gemmOp1.getTransAAttr(), gemmOp1.getTransBAttr());
+
+    // Check for common ConcatOp
```
tungld, line 214: Check this earlier just after you collect all `parallelGemms`. The reason is you have a `return failure()` here which may interrupt the whole rewriting while you created new weight, new bias, and new gemm. Moving this check earlier before creating any new ops would make the IR clean.

tungld (changes requested):

PR #3131, opened 2025-04-18
Backward fold scale axis to Gemm layer in ONNX Dialect

Arkar-Hema:
**BackwardFoldScaleAxisToGemm**

BackwardFoldScaleAxisToGemm is an optimization pass in ONNX-MLIR that merges a BatchNormalization layer into a preceding Gemm (General Matrix Multiplication) layer when specific conditions are met.
Goal: To simplify the model by reducing redundant computations and improving runtime efficiency by statically folding the BatchNormalization into the Gemm operation.

Conditions:
- The Gemm op must have transB = 0 (i.e., the weight matrix is used without transposition)
- The BatchNormalization must be in inference mode

Original computation:
Output=BatchNorm(Gemm(X,W,B)) 

After pass computation:
NewOutput=Gemm(X,W×γ,B×γ+β) 
where, 
- X = Input to Gemm 
- W = Weights in Gemm 
- B = Bias in Gemm 
- γ = Scale in BatchNormalization 
- β = Bias in BatchNormalization

### DIFF
--- a/src/Dialect/ONNX/ONNXOps/Canonicalize.td
+++ b/src/Dialect/ONNX/ONNXOps/Canonicalize.td
@@ -926,6 +926,54 @@ def RewriteBatchNormInferenceModeConvPattern2: Pat<
 >;
 
 //===----------------------------------------------------------------------===//
+// This is to fold the composition: 'BatchNormalization o Gemm' into 'Gemm'
+// by deriving new 'B' and 'C' for 'Gemm' operation after fusing scale and bias.
+//
+// Given:
+//   (Gemm)           Z = A * B + C
+//   (BatchNorm)      Y = scale * (Z - mean) / sqrt(var + eps) + bias
+//
+// This transformation corresponds to a recomposition:
+//   Y = A * (scale * B) + (scale * bias + C)
+//
+// Therefore, we rewrite:
+//   onnx.BatchNormalizationInferenceMode(
+//       onnx.Gemm(A, B, C, alpha, beta, transA, transB),
+//       scale, bias, mean, var
+//   ) {epsilon = ..., momentum = ...}
+//
+// as:
+//   onnx.Gemm(
+//       A,
+//       onnx.Mul(B, scale),
+//       onnx.Add(onnx.Mul(bias, scale), C),
+//       alpha, beta, transA, transB)
+//
+// This transformation is only valid when transB = 0
+// to maintain the correct computation shape alignment.
+//
+//===----------------------------------------------------------------------===//
+
+def isTransBFalse : Constraint<CPred<
+"onnx_mlir::isTransBFalse($0)">, "TransB is 1 not 0"
+>;
+
+def BackwardFoldScaleAxisToGemmPattern : Pat<
+  (ONNXBatchNormalizationInferenceModeOp:$res
+    (ONNXGemmOp $A, $B, $C, $alpha, $beta, $transA, $transB),
+    $scale, $bias, $_mean, $_var, $_epsilon, $_momentum),
+  (ONNXGemmOp
+    $A,
+    (ONNXMulOp $B, $scale),
+    (ONNXAddOp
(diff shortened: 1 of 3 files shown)


src/Dialect/ONNX/ONNXOps/Canonicalize.td
```diff
@@ -925,6 +925,54 @@ def RewriteBatchNormInferenceModeConvPattern2: Pat<
...
+//===----------------------------------------------------------------------===//
+// This is to fold the composition: 'BatchNormalization o Gemm' into 'Gemm'
+// by deriving new 'B' and 'C' for 'Gemm' operation after fusing scale and bias.
+//
+// Given:
+//   (Gemm)           Z = A * B + C
+//   (BatchNorm)      Y = scale * (Z - mean) / sqrt(var + eps) + bias
+//
+// This transformation corresponds to a recomposition:
+//   Y = A * (scale * B) + (scale * bias + C)
```
tungld, line 937: @Arkar-Hema could you elaborate how you derived this formula where `mean`, `var` and `eps` are canceled?

Arkar-Hema: Assume mean=0, var=1 and exp=0 (which is usually present in any pre-compiled and normalised models):

Y ≈ scale × Z + bias
Substituting Z = A x B + C
Y ≈ scale × (A × B + C) + bias
Y = A × (scale × B) + (scale × C + bias)

tungld: > Assume mean=0, var=1 and exp=0 (which is usually present in any pre-compiled and normalised models):

Then, you have to define this assumption in the constraint part of the rewriting rule. Otherwise, the rewriting rule produces a wrong result.

Anyway, my recommendation is to handle the general case where `mean`, `var`, `eps` are constants (not necessary concrete values, say, of 0, 1 and 0, respectively).  New scale and bias values for matmul can be easily computed from `mean`, `var`, `eps`, `scale` and `bias` of BatchNorm, and in the inference mode, these values are constants and will be folded automatically by the compiler into a single constant.

src/Dialect/ONNX/ONNXOps/Canonicalize.td
```diff
@@ -925,6 +925,54 @@ def RewriteBatchNormInferenceModeConvPattern2: Pat<
...
+//===----------------------------------------------------------------------===//
+// This is to fold the composition: 'BatchNormalization o Gemm' into 'Gemm'
+// by deriving new 'B' and 'C' for 'Gemm' operation after fusing scale and bias.
+//
+// Given:
+//   (Gemm)           Z = A * B + C
+//   (BatchNorm)      Y = scale * (Z - mean) / sqrt(var + eps) + bias
+//
+// This transformation corresponds to a recomposition:
+//   Y = A * (scale * B) + (scale * C + bias)
```
AlexandreEichenberger, line 937: Can you explain how you derive this result. Looking at the definitions, I derive the following:

```
Y = alpha * Z + B 
where
alpha = scale / sqrt(var + eps)
beta = (bias - scale * mean / sqrt(var + eps).

Substituting Z by A*B + C we get

Y = A * (alpha * B) + (alpha * C + beta)
```

The formula and code listed below seems to ignore the contributions of `mean`, `var` and `eps`.

AlexandreEichenberger: See `def FuseBatchNormInferenceModeConvPattern` for how this was done for BN and conv

Arkar-Hema: these are assumptions that I made
mean=0, var=1 and exp=0 (which is usually present in any pre-compiled and normalised models):

Y ≈ scale × Z + bias
Substituting Z = A x B + C
Y ≈ scale × (A × B + C) + bias
Y = A × (scale × B) + (scale × C + bias)

and I have updated this checks in the code

PR #3134, opened 2025-04-21
Support --march=native

chentong319:
Small change, use llvm::sys::getHostCPUName to support --march=native


AlexandreEichenberger:
@chentong319 

good idea. The current PR fails for the release mode

```
CompilerOptions.cpp:(.text._ZN9onnx_mlir18initCompilerConfigEv+0xba9): undefined reference to `llvm::sys::getHostCPUName()'
```
Had this type of errors before, you need to include the lib from llvm defining the `getHostCPUName` in the CMakefile target building the lib that includes the `CompilerOptions.cpp` file.

AlexandreEichenberger (approved):
LGTM, thanks for this useful addition.

PR #3140, opened 2025-04-24
Fusion of Matmul add covering the stacked/unstacked/bcast1/bcast23 patterns

AlexandreEichenberger:
This PR add support for the 4 patterns of matmul and add in NNPA.

image

As seen above, the stacked and and bcast1 take a bias (input_c) of the shape `[s, p]` which is added to the `[s, m, p]` output. Because we perform the lowering of the onnx to ZHigh in several steps, the first step is to possibly expand a 1D constant `[p]` to a  `[s, p]` which is accepted by the `zhigh.matmul` operation. Notice that the add performs an addition of `[s, m, p]` and `[s, p]` . This does not satisfy the broadcast rules under ONNX. So we expand the constant to be `[s, 1, p]` in the ONNX dialect, and when we actually lower the onnx matmul and add to a `zhigh.matmul` we actually `squeeze` the `1` away to satisfy the expectations of the zhigh matmul.

I have tested all 4 patterns manually, with/without the fusion to ensure that we get the same results.

I have also tested the tabbert model, with/without the fusion and verified that the outputs were identical.

In addition, the new PR reduced the number of `zhigh.add` from 102 to 26, and reducing the overall runtime by about 6%.

I added literal tests checking when optimization can/cannot run to the best of my ability.


src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.td
```diff
@@ -637,12 +640,13 @@ def replaceONNXMatMulAddPattern1 : Pat<
         (ZHighStickOp $b, (GetMatMulBiasLayoutStringAttr (GetRank $x),
                                                          (GetRank $y)), (GetDefaultSaturation)),
         (GetZeroI64Attr), (GetZeroI64Attr))),
-  [(IsMatMulLegalForZDNN $m), (HasRankOf<2> $y), (HasRankOf<1> $b),
-   (HaveSameLastDimR2R1 $y, $b)], [],
+  [(HasRankOf<2> $x), (HasRankOf<2> $y), (HasRankOf<1> $b), (HaveSameLastDimR2R1 $y, $b),
```
tungld, line 643: We didn't check rank of x because this pattern can cover both `unstacked` and `bcast23` cases.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
```diff
@@ -415,9 +420,165 @@ class SplitLargeMatMulPattern : public OpRewritePattern<ONNXMatMulOp> {
...
+    add a "1" in between the s and the p. If we have only a (p), it works too
+    when we expand the added constant to s replicas of (p) to become (s, 1 p).
+    Hardware op will eventually get (s, p).
+  */
+  LogicalResult matchAndRewrite(
+      ONNXAddOp addOp, PatternRewriter &rewriter) const override {
+    // For the moment, only handle constant shapes.
+    // TODO: add support for dynamic shapes that do not impact the constants.
+    if (!hasStaticShape(addOp.getA().getType()) ||
+        !hasStaticShape(addOp.getB().getType()))
```
tungld, line 455: it seems only B has static shape would be enough since we only need dimensions `s` and `p` to expand the input C from shape `(p)` to `(s, p)`. And B is often a constant.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
```diff
@@ -415,9 +420,165 @@ class SplitLargeMatMulPattern : public OpRewritePattern<ONNXMatMulOp> {
...
+      return failure();
+    // Try first matmul case, then add only.
+    if (processMatMulAddCase(addOp, rewriter) ||
+        processAddOnlyCase(addOp, rewriter))
```
tungld, line 459: I guess you wanted to say there are two patterns on AddOp and you would like to apply MatMul-Add pattern first. To do that we can set `benefit` for each pattern, and the pattern with a higher benefit will be called first.

For example, in tablegen, we can define benefit for `expandConstantOperandForAddOp2` to be `0`:
```
def expandConstantOperandForAddOp2: Pat<
  (ONNXAddOp $x, (ONNXConstantOp:$c $_, $_, $_, $_, $_, $_, $_, $_)),
  (ONNXAddOp $x, (ONNXExpandOp $c,
                               (CreateShapeOp $x),
                               (returnType $x))),
  [(IsUniBroadcastingFromFirstToSecond $c, $x)],
  [],
  (addBenefit 0)>;
>;
```

and in C++ pattern, we define benefit `1` for `expandConstantOperandForAddOp2`:
```c
class ExpandAddConstantPattern : public OpRewritePattern<ONNXAddOp> {
public:
  using OpRewritePattern<ONNXAddOp>::OpRewritePattern;
  
  ExpandAddConstantPattern(MLIRContext *context, PatternBenefit benefit = 1)
      : OpRewritePattern<ONNXAddOp>(context, benefit) {}


```

AlexandreEichenberger: Yes, that is true. I am still not sure benefits can be used to do that.

For example, fused multiply add (benefit 1) may want to keep the dim of the bias as 1D... so it succeeds while making no change.

Does that guarantee that the add alone (benefit 0) will not fire to augment the dims to match that of the input? 

If the first higher benefit pattern modified the graph so that the lower benefit pattern would not apply anymore, then I would trust it. But in this case, the higher benefit pattern may not modify the graph... and thus technically the second lower priority pattern may still apply, unless there is a rule stating that "if a pattern applied successfully on an op, then no other lower priority pattern may apply".

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.td
```diff
@@ -624,9 +627,9 @@ def HaveSameLastDimR2R1: Constraint<
   "Have the same last dimension"
 >;
 
-// Only 1D bias is suitable for this transformation since only then
+// Unstacked: only 1D bias is suitable for this transformation since only then
```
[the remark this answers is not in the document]
AlexandreEichenberger: Now that both are reenabled for z16, makes total sense to fuse them. Done

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.td
AlexandreEichenberger: got it

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.td
```diff
@@ -653,16 +657,122 @@ def replaceONNXMatMulAddPattern2 : Pat<
...
+>;
+
+//===----------------------------------------------------------------------===//
+// Replace onnx.add and onnx.matmul with stacked tensors with ZHighMatMul
+//===----------------------------------------------------------------------===//
+
+// Be careful, this check is R1 are the same and R3 are the same. Must have
+// rank of 3. Should we include dim analysis?
+def HaveSameLastDimR1AndR3: Constraint<
+  CPred<"(!mlir::cast<ShapedType>($0.getType()).isDynamicDim(1))" #
```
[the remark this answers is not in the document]
AlexandreEichenberger: Changed it to test if first and 3 dims of both tensors are static and first dims are the same, and 3rd dims are the same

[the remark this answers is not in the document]
AlexandreEichenberger: made the suggested changes which are more explicit.

AlexandreEichenberger:
@tungld thanks for your extensive suggestions; believe I responded to all and implemented all but one.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
AlexandreEichenberger: Yes, changed to only checking if B is constant. Had to modify the code slightly to accommodate this, it now works.

test/mlir/accelerators/nnpa/driver/dense-out-attention-layer.mlir
```diff
@@ -1,4 +1,8 @@
-// RUN: onnx-mlir --march=z16 --maccel=NNPA --EmitZHighIR --printIR %s | FileCheck %s
+// RUN: onnx-mlir --march=z17 --maccel=NNPA --EmitZHighIR --printIR %s | FileCheck %s
+
+// -----
+
+// Pattern: bcast23 which is supported on arch15, not arch14
```
tungld, line 5: change back to z16?

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
tungld: > If the first higher benefit pattern modified the graph so that the lower benefit pattern would not apply anymore, then I would trust it.

I think it does so. Once the higher benefit pattern is applied, the ONNXAdd is replaced and removed, so the lower benefit pattern has no chance to be applied on the same ONNXAdd.

tungld (approved):
LGTM. Thank you for a good fusion!

PR #3144, opened 2025-04-28
Fix types in RandomNormal(Like) test inputs to match the dtype

jorickert:
I missed this in https://github.com/onnx/onnx-mlir/pull/3091


AlexandreEichenberger (commented):
Quick question:  We support random normal in f32 or f64. Any reason not to test both?

jorickert:
@AlexandreEichenberger There are tests for both. The test for f32 was already correct, so only the test for f64 needed to be adjusted

PR #3145, opened 2025-04-29
Conv add const where the constant is a scalar

### LINKED ISSUE #2989: constexpr llvm::ArrayRef<T>::ArrayRef(const T*, const T*) [with T = long int]: Assertion `begin <= end' failed.
When I run the command in the terminal:    $ onnx-mlir encoder-epoch-12-avg-2-chunk-16-left-64.onnx 

This is the output:

[1/6] Wed Oct 30 15:29:25 2024 (0s) Importing ONNX Model to MLIR Module from "encoder-epoch-12-avg-2-chunk-16-left-64.onnx"
[2/6] Wed Oct 30 15:29:25 2024 (0s) Compiling and Optimizing MLIR Module
onnx-mlir: /home/tjy/project_cxy/llvm-project/llvm/include/llvm/ADT/ArrayRef.h:84: constexpr llvm::ArrayRef<T>::ArrayRef(const T*, const T*) [with T = long int]: Assertion `begin <= end' failed.
Aborted (core dumped)

This is the onnx model 
[model.zip](https://github.com/user-attachments/files/17568519/model.zip)

How to solve it?

AlexandreEichenberger:
The code seems to expect the `add ( conv, const)` where the `const` is a tensor of `rank>1`.
When it is not, a test asserts. It appears that the way the `AllDimsFromAxisToEndAre` is used, a scalar would also satisfy the conditions that is imposed by `ScalarOrAllDimsFromAxisToEndAre` on tensors of rank > 1.

Renamed the function `ScalarOrAllDimsFromAxisToEndAre` to better reflect what Is tested, and it will return true when input is a scalar or when all dims from axis (==1) to the end are 1.

This should fix #2989 . I compiled the first model given in that issue, and it compiled fine. I have no ability to verify the output, so counting on @ cxy2333-hue and @VioletsOleander to verify the results of their model with this new fix.


src/Dialect/ONNX/ONNXOps/Canonicalize.td
```diff
@@ -334,8 +334,9 @@ def FuseAddConvNullBiasPattern: Pat<
   [(HasShapeAndRank:$res),
    (HasNoneType $b),
    (AttributeIsNotNull:$denseAttr),
+   (RankXMinusRankYIs<1> $res, $y),
```
tungld, line 337: I think this happens that `$res` has rank of 4 in this case. But if `$res` has rank of 1 and `$y` is a scalar (rank 0), `RankXMinusRankYIs<1> $res, $y` is still true.

A better way would be check that `$y` is not  scalar by adding `HasRankGT<0> $y`, which says `$y` must have rank greater than 0.

tungld (approved):
LGTM. 

The current code is ok , but feel free to update if needed.

AlexandreEichenberger:
Updated as requested. Thanks for the quick review and good comments, as usual.

PR #3148, opened 2025-05-01
Instance and Group norm needs shape inference

AlexandreEichenberger:
We previously reduced instance and group norms to Layer norm in decompose, which is the first pass in onnx-mlir before shape inference is run. But these transformations required the shape inference of the norm ops to be known (could not be unranked). Thus it was too early. I migrated these transformations to canonicalization.

We don't have a test per se, but that example below would not work in the old approach.

```mlir
module attributes {llvm.data_layout = "e-m:o-i64:64-i128:128-n32:64-S128", llvm.target_triple = "arm64-apple-darwin23.0.0", "onnx-mlir.symbol-postfix" = "test_instancenorm_epsilon"} {
  func.func @main_graph(%arg0: tensor<2x3x4x5x6xf32>, %arg1: tensor<3xf32>, %arg2: tensor<3xf32>) -> tensor<*xf32> attributes {input_names = ["x", "s", "bias"], output_names = ["y"]} {
    %0 = "onnx.Add"(%arg0, %arg0) : (tensor<2x3x4x5x6xf32>, tensor<2x3x4x5x6xf32>) -> tensor<*xf32>
    %1 = "onnx.InstanceNormalization"(%0, %arg1, %arg2) {epsilon = 0.00999999977 : f32} : (tensor<*xf32>, tensor<3xf32>, tensor<3xf32>) -> tensor<*xf32>
    onnx.Return %1 : tensor<*xf32>
  }
}
```

this test now works.

This is in response to issue ##3141


src/Dialect/ONNX/ONNXOps/Canonicalize.cpp
```diff
@@ -1751,6 +1751,225 @@ class RemoveWhereEqualPattern : public OpRewritePattern<ONNXWhereOp> {
   }
 };
 
+// =============================================================================
```
AlexandreEichenberger, line 1754: this code is not new, migrated from `decompose.cpp` to here.

test/mlir/onnx/onnx_canonicalization.mlir
```diff
@@ -2005,3 +2005,117 @@ func.func @test_reorder_relu_maxpool(%arg0: tensor<1x64x32x32xf32>) -> tensor<1x
   // CHECK-NEXT:     return [[VAR_1_]] : tensor<1x64x16x16xf32>
   // CHECK:         }
 }
+
```
AlexandreEichenberger, line 2008: these tests are not new, they migrated from decompose tests to here.

tungld (approved):
LGTM.

PR #3152, opened 2025-05-09
Do not fuse locations in Maxpool-Relu canonicalization

jorickert:
As both operations still do their original functionality and did not get merged together, I think not fusing their locations is correct in this case.

I created a new test file for the location test, as `--mlir-print-debuginfo` makes it more difficult to write LIT tests and I do not want to impact the main canonicalization LIT test by enabling this option


jorickert:
@Arkar-Hema FYI

tungld (approved):
LGTM.

PR #3153, opened 2025-05-09
Support for arch native on mac

AlexandreEichenberger:
The -march=native results in a `apple-m1` arch (or derivatives) on Macs, which is not recognized by the `onnx-mlir` simd code generator as well as `llc`.

Add here the code to handle this properly. Also added a print of the native machine being detected under the verbose flag `-v` so that we may discover what machine is actually compiled for.

### DIFF
--- a/src/Compiler/CompilerOptions.cpp
+++ b/src/Compiler/CompilerOptions.cpp
@@ -926,6 +926,10 @@ std::string getTargetArchOption(bool forLLVMToolchain) {
     int64_t zArchNum = getZArchNum(march, mcpu);
     if (zArchNum != -1)
       return "--march=systemz";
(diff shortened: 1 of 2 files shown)


src/Compiler/CompilerOptions.cpp
```diff
@@ -1413,11 +1417,11 @@ void initCompilerConfig() {
     setLLVMOption(getLLVMOption() + " --enable-unsafe-fp-math");
   }
 
-  if (march == "z17")
```
AlexandreEichenberger, line 1416: zXX or archYY are equally recognized, so no need to convert the one into the other

chentong319 (approved):
LGTM!

PR #3160, opened 2025-05-13
Propagate onnx.dim_params in shape inference

chentong319:
ONNX model provides symbolic name for dynamic dimension of input tensor with argument attribute "onnx.dim_params". This PR tries to propagate this symbolic name for dynamic dimension during shape inference with extra attribute to operation "onnx.dim_params_[n]", where n is the index of the output tensor.
The major change is on IndexExpr to carry such info, if it exists. Other changes include the use of the dim_param info in shape inference and test case.
So far, the usu of dim_param info is far from complete. The PR is kind of experimental.
At least one benefit of this approach is that the readability of the output of onnx-mlir can be improved with the symbolic names.

### DIFF
--- a/src/Dialect/ONNX/ONNXOps/ShapeHelper.cpp
+++ b/src/Dialect/ONNX/ONNXOps/ShapeHelper.cpp
@@ -67,6 +67,32 @@ int64_t getAxisInRange(int64_t axis, Value val, bool includeRank) {
 // ONNX Op Shape Helper
 //===----------------------------------------------------------------------===//
 
+static void refineDimParams(Operation *op, DimsExpr &inferredDims, Value output) {
+  // Get the index of output
+  if (!llvm::isa<OpResult>(output)) {
+    // output is a block parameter. It could be Func, Loop, If and etc.
+    // Give up now due to the complicated control flow.
+    return;
+  }
+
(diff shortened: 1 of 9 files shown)


src/Dialect/ONNX/ONNXOps/ShapeHelper.cpp
```diff
@@ -67,6 +67,37 @@ int64_t getAxisInRange(int64_t axis, Value val, bool includeRank) {
...
+        dimParamsStr.append(",");
+      }
+      dimParamsStr = dimParamsStr + std::to_string(i) + ":" +
+                     inferredDims[i].getDimParam();
+    }
+  }
+  if (dimParamsStr == "")
+    return;
+  StringAttr dimParamsAttr = StringAttr::get(op->getContext(), dimParamsStr);
+  op->setAttr("onnx.dim_params_" + std::to_string(resultIndex),
```
tungld, line 97: Suggestion: how about using `onnx.out_dim_params_`? I though `onnx.dim_params_0` is for the first input, but actually it is for the first output.

tungld (approved):
LGTM with a minor change in naming.

The propagation looks effective. Let's see how we can utilize this information in future PRs.

For the changes to IndexExpr, I would like to hear comments from @AlexandreEichenberger as well.

PR #3162, opened 2025-05-14
Fuse locations when merging concats

jorickert:
Having accurate locations is helpful for tracing and debugging the compilation flow

Downstream we make extensive use of them

@Arkar-Hema  FYI


AlexandreEichenberger (approved):
LGTM

PR #3170, opened 2025-05-21
Prefetch prevent OMP pass to succeed

AlexandreEichenberger:
Prefetch prevents OMP generation (scf-to-openmp).
PR disable prefetching for stick/unstick
Also remove a pass that is redundant (and too early to succeed)


src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
```diff
@@ -282,10 +282,8 @@ void addPassesNNPA(mlir::OwningOpRef<mlir::ModuleOp> &module,
         // Optimizations at ZLow after normalizing MemRefs.
         pm.addPass(zlow::createZLowRewritePass());
         // The createZLowStickExpansion pass may create parallel constructs,
-        // they need to be handled here.
-        if (!nnpaDisableCompilerStickUnstick && enableParallel)
-          pm.addPass(mlir::createConvertSCFToOpenMPPass());
-
+        // but we don't need to handle it here as it will be processed when
+        // lowering Krnl to LLVM.
```
tungld, line 286: I think we can remove this comment completely. There is no context to understand it once we removed `createConvertSCFToOpenMPPass ()`

tungld (approved):
LGTM! Thank you so much for a quick solution!

src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
AlexandreEichenberger: Will remove as requested

Issue #3128, opened 2025-04-17
onnx model MLIR conversion failing


kraza8:
onnx model MLIR conversion is failing after successful onnx export.

onnx_model_path = "/vlaWithDynamo.onnx"

torch.onnx.export(vla, (inputs["input_ids"], inputs["attention_mask"], inputs["pixel_values"]), onnx_model_path, input_names=["input_node"], output_names=["output_node"], dynamo=True, report=True, export_params=True)

./onnx-mlir --EmitMLIR /vlaWithDynamo.onnx
[1/3] Wed Apr 16 16:00:24 2025 (0s) Importing ONNX Model to MLIR Module from "vlaWithDynamo.onnx"
[2/3] Wed Apr 16 16:00:24 2025 (0s) Compiling and Optimizing MLIR Module
'onnx.Pow' op Pow with different input type not implemented yet

System Info:

Linux Ubuntu 22.04*):
ONNX version: Version: 1.17.0
Python version: 3.10.12
Protobuf version: 6.30.2
torch: 2.8.0.dev20250415+cpu
onnx-mlir version: 0.4.2
Code:
from transformers import AutoModelForVision2Seq, AutoProcessor
from PIL import Image
import numpy as np
import torch
import time
import onnx
from onnxruntime.training import artifacts

from pathlib import Path

processor = AutoProcessor.from_pretrained("openvla/openvla-7b", trust_remote_code=True)

vla = AutoModelForVision2Seq.from_pretrained(
"openvla/openvla-7b",
#attn_implementation="flash_attention_2", # [Optional] Requires flash_attn
torch_dtype=torch.float32,
low_cpu_mem_usage=True,
trust_remote_code=True
)

onnx_model_path = "/vlaWithDynamo.onnx"

torch.onnx.export(vla, (inputs["input_ids"], inputs["attention_mask"], inputs["pixel_values"]), onnx_model_path, input_names=["input_node"], output_names=["output_node"],dynamo=True, report=True, export_params=True)

Expected Behavior:
onnx model to generate its MLIR form without any issues.

tungld:
Hi @kraza8, as the error message said, the Pow currently does not support different types for base and power. It supports the case where both base and power are float only. Need help from someone to add support for the case of different types.

srcarroll:
so i can get onnx-mlir from `torch.pow(x, y)` with `x` and `y` mixed types. it just casts one of the inputs appropriately before applying `onnx.Pow`. you get something like
```
[...]

PR #3175, opened 2025-05-23
Allows POW with different types (convert 2nd input to 1st input type)

AlexandreEichenberger:
Fix case listed in #3128 due to power with different type. Use the convention to convert the 2nd input to that of the first input (and output) type,

### DIFF
--- a/src/Dialect/ONNX/ONNXOps/Math/ElementwiseBroadcast.cpp
+++ b/src/Dialect/ONNX/ONNXOps/Math/ElementwiseBroadcast.cpp
@@ -347,8 +347,6 @@ LogicalResult ONNXPowOp::verify() {
   ShapedType rhsTy = mlir::cast<ShapedType>(getY().getType());
   Type rhsETy = rhsTy.getElementType();
   Type lhsETy = lhsTy.getElementType();
-  if (rhsETy != lhsETy)
-    return emitOpError("Pow with different input type not implemented yet");
   if (mlir::isa<IntegerType>(lhsETy) || mlir::isa<IntegerType>(lhsETy))
     return emitOpError("Integer power not implemented yet");
   return success();
--- a/src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
+++ b/src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
@@ -318,16 +318,6 @@ GenOpMix getGenOpMix<ONNXSinOp>(Type t, Operation *op) {
 }
 
 template <>
-struct ScalarOp<ONNXPowOp> {
-  using FOp = math::PowFOp;
-  using IOp = NotSuportedScalarOp;
-};
-template <>
-GenOpMix getGenOpMix<ONNXPowOp>(Type t, Operation *op) {
-  return {{GenericOps::PowGop, 1}};
-}
-
-template <>
 struct ScalarOp<ONNXIsNaNOp> {
   using FOp = KrnlIsNaNOp;
   using IOp = NotSuportedScalarOp;
@@ -1252,6 +1242,32 @@ Value emitScalarOpFor<ONNXNegOp>(ConversionPatternRewriter &rewriter,
   Value operand = scalarOperands[0];
   MultiDialectBuilder<MathBuilder> create(rewriter, loc);
   return create.math.neg(operand);
+}
+
+//===----------------------------------------------------------------------===//
+// Scalar binary ops for lowering ONNXPowOp
+//===----------------------------------------------------------------------===//
+
+template <>
+GenOpMix getGenOpMix<ONNXPowOp>(Type t, Operation *op) {
+  return {{GenericOps::PowGop, 1}};
+}
+
+template <>
+Value emitScalarOpFor<ONNXPowOp>(ConversionPatternRewriter &rewriter,
+    Location loc, Operation *op, Type elementType,
+    ArrayRef<Value> scalarOperands) {
+  CheckIfCustomScalarOpIsSupported<ONNXGeluOp>(elementType);
+  Value lhs = scalarOperands[0];
+  Value rhs = scalarOperands[1];
+  Type lhsElemType = getElementType(lhs.getType());
+  Type rhsElemType = getElementType(rhs.getType());
+  MultiDialectBuilder<MathBuilder> create(rewriter, loc);
+  if (lhsElemType != rhsElemType) {
+    // Conversions needed.
+    rhs = create.math.cast(lhs.getType(), rhs);
+  }
+  return create.math.pow(lhs, rhs);
 }
 
 //===----------------------------------------------------------------------===//
--- a/test/mlir/conversion/onnx_to_krnl/Math/Elementwise.mlir
+++ b/test/mlir/conversion/onnx_to_krnl/Math/Elementwise.mlir
@@ -712,7 +712,7 @@ func.func private @test_clip_default_min(%arg0: tensor<3xf32>, %arg1: tensor<f32
 
 // -----
 
-func.func private @test_pow(%arg0: tensor<3x4x5xf32>, %arg1: tensor<3x4x5xf32>) -> tensor<3x4x5xf32> {
+func.func @test_pow(%arg0: tensor<3x4x5xf32>, %arg1: tensor<3x4x5xf32>) -> tensor<3x4x5xf32> {
     %0 = "onnx.Pow"(%arg0, %arg1) : (tensor<3x4x5xf32>, tensor<3x4x5xf32>) -> tensor<3x4x5xf32>
     return %0 : tensor<3x4x5xf32>
 // CHECK-LABEL: test_pow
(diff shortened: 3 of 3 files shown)


srcarroll:
reposting this question from a related discussion [here](https://github.com/onnx/onnx-mlir/pull/3175)

i see that you are now allowing mixed types in the onnx.Pow definition and handling the appropriate type cast in the to-kernel conversion. but i wonder if it's more appropriate to keep the type conditions as is, and instead of handling in the to-kernel conversion handle it in the onnx to onnx-mlir conversion with the onnx-mlir tool.

I dont know what's best from a design standpoint but I think that other conversions from onnx mlir dialect will still have to handle the mixed types. for example the to-stablehlo and to-tosa conversions. So i guess my question is, should this be the responsibility of the specific backend consuming onnx-mlir? or should it be inherent to the onnx to onnx-mlir conversion from the start?

AlexandreEichenberger:
Interesting. From a performance perspective, if the power is a tensor it is better to do it when lowering (only one loop, free fusion). For a scalar, it does not matter much either way. I guess we could add a rule in decompose or canonicalization that add a onnx conversion.

Let me know if you / others feel that it should be a generic onnx->onnx transformation.

srcarroll:
> Interesting. From a performance perspective, if the power is a tensor it is better to do it when lowering (only one loop, free fusion). For a scalar, it does not matter much either way. I guess we could add a rule in decompose or canonicalization that add a onnx conversion.
> 
> Let me know if you / others feel that it should be a generic onnx->onnx transformation.

I personally think it should follow the same principles as any other simple arithmetic onnx op. So for example, what are the rules for `x * y` with mixed types?  I can't currently see any reason to treat `pow` differently.  I'd definitely like to hear what others think

tungld:
> So for example, what are the rules for x * y with mixed types?

Simple binary ops in ONNX require the same type for the two inputs as I know.

srcarroll:
> > So for example, what are the rules for x * y with mixed types?
> 
> Simple binary ops in ONNX require the same type for the two inputs as I know.

yes i know. sorry, i meant what are the rules for casting.  the ultimate question here is if we should be doing the casting in the onnx to onnx-mlir conversion, or should this be explicitly done before conversion?

AlexandreEichenberger:
@tungld do you know if the MLIR ops tolerate floats of different sizes?

From the  standard (https://onnx.ai/onnx/operators/onnx__Pow.html) we would need to support all combinations of ints and float, e.g. (int8 ^ float64).

For the ONNX -> Krnl path, we can rely on the MLIR Math to handle (int & int, float ^ float, float ^ int) of same sizes, and convert the other ones using explicit conversions.

Possibly other ONNX -> ? path may need conversions in all cases?

AlexandreEichenberger:
@srcarroll would your infra being able to handle float ^ float, float ^ int, int ^ int of the same size? That is what I would make onnx -> krnl support natively, and the other kind of conversions would happen in onnx -> onnx.

srcarroll:
> @srcarroll would your infra being able to handle float ^ float, float ^ int, int ^ int of the same size? That is what I would make onnx -> krnl support natively, and the other kind of conversions would happen in onnx -> onnx.

I use the onnx->stablehlo->linalg conversions to get to builtin MLIR,  so I would have to check to see how stablehlo handles these cases. but i'd say as long as we can get onnx->onnx-mlir then I'm pretty confident there's a way to get conversions the rest of the way. and i'd be happy to contribute to the onnx-mlir->stablehlo conversion if needed

srcarroll:
> @tungld do you know if the MLIR ops tolerate floats of different sizes?

for float,float `math.pow` requires same size. same for int, int with `math.ipow`. it looks like the float, int case can have different widths and the case that put this whole conversation in motion was a float32, int64 case.

srcarroll:
@AlexandreEichenberger sorry for the delay. looks like stablehlo has a `stablehlo.power` op that requires same-type inputs.  the onnx->stablehlo conversion uses a generic `ONNXElementwiseBinaryOpLoweringToStablehlo` pattern, which i'm not sure if it will for the mixed type ops that are being proposed by @tungld , so they might need their own patterns to handle the casting to same-type. i don't think that's a big deal though.

On the other hand, since `stablehlo.power` is single type, then that means we will always get only the `math.pow` version in builtin MLIR (linalg). so the onnx->stablehlo->linalg path would never generate the other pow variants. but i dont think that's necessarily the concern of onnx-mlir to accommodate.

AlexandreEichenberger:
@srcarroll  this new PR allows `ONNXPowOp` to have any int/float combinations as inputs.

Canonicalization will introduce conversions for all cases but the ones supported by MLIR, namely both inputs have the same width, and its either `float ^ float`, `int ^ int`, or `float ^ int`. All the other cases have cast introduced for the exponent to match the type of the input and result.

Then the lowering of ONNX to Krnl was expanded to support MLIR supported cases. Specifically, the mlir builder generate the right MLIR power operations depending on which of the 3 categories listed above we are in.

Added lit tests to test these scenarios.

src/Dialect/ONNX/ONNXOps/Canonicalize.cpp
```diff
@@ -1239,6 +1239,58 @@ class PowToMulRewritePattern : public OpRewritePattern<ONNXPowOp> {
...
+    // Detect cases that MLIR supports without conversions => failure, aka no
+    // need to do anything here.
+    if ((inputWidth == expWidth) &&
+        ((inputIsInt && expIsInt) || (inputIsFloat && expIsFloat) ||
+            (inputIsFloat && expIsInt)))
+      return failure();
+
+    // Rewrite pow with a cast of the exp type to the input type.
+    MultiDialectBuilder<OnnxBuilder> create(rewriter, loc);
+    Value newExp = create.onnx.cast(exp, inputElementType);
```
srcarroll, line 1276: is it safe enough to always cast to the base operand's type? 

side question about `onnx.Pow` definition: what should be the result type width be? is it always the same as the base operand type width? if it is, i guess semantically it makes sense to always cast the exponent to base type.

src/Dialect/ONNX/ONNXOps/Canonicalize.cpp
```diff
@@ -1239,6 +1239,58 @@ class PowToMulRewritePattern : public OpRewritePattern<ONNXPowOp> {
...
+    // Characterize element types.
+    Type inputElementType =
+        mlir::cast<ShapedType>(input.getType()).getElementType();
+    Type expElementType =
+        mlir::cast<ShapedType>(exp.getType()).getElementType();
+    bool inputIsInt = isa<IntegerType>(inputElementType);
+    bool inputIsFloat = isa<FloatType>(inputElementType);
+    bool expIsInt = isa<IntegerType>(expElementType);
+    bool expIsFloat = isa<FloatType>(expElementType);
+    int64_t inputWidth = 0, expWidth = 0;
```
srcarroll: this can be simplified with `getElementTypeBitWidth` from a `ShapedType` or `getIntOrFloatBitWidth` from a `Type` that's expected to be `FloatType` or `IntegerType`

src/Dialect/ONNX/ONNXOps/Canonicalize.cpp
```diff
@@ -1239,6 +1239,58 @@ class PowToMulRewritePattern : public OpRewritePattern<ONNXPowOp> {
...
+
+  LogicalResult matchAndRewrite(
+      ONNXPowOp powOp, PatternRewriter &rewriter) const override {
+    Operation *op = powOp.getOperation();
+    Location loc = powOp.getLoc();
+
+    Value input = powOp.getX();
+    Value exp = powOp.getY();
+    if (!hasShapeAndRank(input) || !hasShapeAndRank(exp))
+      return failure(); // Cannot apply pattern until ranks are known.
```
srcarroll: shape and rank shouldn't be required, right? all we need is the element type?

src/Dialect/ONNX/ONNXOps/Math/ElementwiseBroadcast.cpp
```diff
@@ -343,14 +343,7 @@ LogicalResult ONNXOrOp::inferShapes(
 //===----------------------------------------------------------------------===//
 
 LogicalResult ONNXPowOp::verify() {
```
srcarroll, line 345: do we need an explicit verifier anymore? or do all onnx ops opt into `hasVerifier`? (i dont remember)

src/Dialect/ONNX/ONNXOps/Canonicalize.cpp
AlexandreEichenberger: We generally don't check that the type are the ones supported by the standard, added the test as requested,

I believe matching the type of the base make sense as the type of the output is the same as the type of the base.

src/Dialect/ONNX/ONNXOps/Canonicalize.cpp
AlexandreEichenberger: good suggestion, done.

src/Dialect/ONNX/ONNXOps/Canonicalize.cpp
AlexandreEichenberger: good catch, fixed.

src/Dialect/ONNX/ONNXOps/Math/ElementwiseBroadcast.cpp
AlexandreEichenberger: I added the test for int or float input, so now we need it. In general, we encourage having a verifier, it needs to be requested by adding the ONNX Op on a python list.

AlexandreEichenberger:
@srcarroll thanks for your detailed comments. Believe I implemented all of them.

src/Dialect/ONNX/ONNXOps/Canonicalize.cpp
srcarroll: ok i think that makes sense. thanks

src/Dialect/ONNX/ONNXOps/Canonicalize.cpp
```diff
@@ -1239,6 +1239,48 @@ class PowToMulRewritePattern : public OpRewritePattern<ONNXPowOp> {
...
+    Type inputElementType = getElementTypeOrSelf(input);
+    Type expElementType = getElementTypeOrSelf(exp);
+    bool inputIsInt = isa<IntegerType>(inputElementType);
+    bool inputIsFloat = isa<FloatType>(inputElementType);
+    bool expIsInt = isa<IntegerType>(expElementType);
+    bool expIsFloat = isa<FloatType>(expElementType);
+    int64_t inputWidth =
+        mlir::cast<ShapedType>(input.getType()).getElementTypeBitWidth();
+    int64_t expWidth =
+        mlir::cast<ShapedType>(exp.getType()).getElementTypeBitWidth();
```
srcarroll: very nitpicky and what you have is perfectly fine, so feel free to ignore. since you already get the element type above, and the verifier already makes sure that it is either float or int, then you can avoid a cast by using the other method i mentioned earlier, `getIntOrFloatBitWidth`.

AlexandreEichenberger: It's fine, it is indeed more elegant (though there are more cast in the call anyway). Changed as requested.

src/Dialect/ONNX/ONNXOps/Math/ElementwiseBroadcast.cpp
```diff
@@ -343,14 +343,13 @@ LogicalResult ONNXOrOp::inferShapes(
...
-  Type rhsETy = rhsTy.getElementType();
-  Type lhsETy = lhsTy.getElementType();
-  if (rhsETy != lhsETy)
-    return emitOpError("Pow with different input type not implemented yet");
-  if (mlir::isa<IntegerType>(lhsETy) || mlir::isa<IntegerType>(lhsETy))
-    return emitOpError("Integer power not implemented yet");
+  // Removed integer restriction.
+  Type inputElementType = getElementTypeOrSelf(getX());
+  Type expElementType = getElementTypeOrSelf(getY());
+  if (!(isa<IntegerType>(inputElementType) || isa<FloatType>(inputElementType)))
```
tungld, line 349: You can use `isa<IntegerType, FloatType>(inputElementType)` for shot.

test/mlir/onnx/onnx_canonicalization.mlir
```diff
@@ -1403,6 +1403,100 @@ func.func @expand_pow_into_constant(%arg0: tensor<3x4x5xf32>) -> tensor<3x4x5xf3
...
+// CHECK:           [[VAR_0_:%.+]] = "onnx.Cast"([[PARAM_1_]]) {saturate = 1 : si64, to = f32} : (tensor<1xi64>) -> tensor<1xf32>
+// CHECK:           [[VAR_1_:%.+]] = "onnx.Pow"([[PARAM_0_]], [[VAR_0_]]) : (tensor<3x4x5xf32>, tensor<1xf32>) -> tensor<3x4x5xf32>
+// CHECK:           onnx.Return [[VAR_1_]] : tensor<3x4x5xf32>
+// CHECK:         }
+}
+
+// -----
+
+// COM: conv introduced due to unsupported mlir combination (int ^ float)
+func.func @pow_convert_size4(%arg0: tensor<3x4x5xi32>, %arg1: tensor<1xf32>) -> tensor<3x4x5xi32> {
```
tungld, line 1470: This case `base i32, power f32 `is interesting. It seems that converting the base to `f32` would give a correct result. For example `4^(0.5)`, the answer is `2`, but if we cast `0.5` to  `0` or `1`, the answer is `1` or `4`.

src/Dialect/ONNX/ONNXOps/Math/ElementwiseBroadcast.cpp
AlexandreEichenberger: sure thing

test/mlir/onnx/onnx_canonicalization.mlir
AlexandreEichenberger: Interesting observation: it's kind of a strange case.... I asked the question on the slack onyx-operator channel.

srcarroll: i wonder if the `int ^ float` case should be legal in the first place. seems like it's just not a well defined operation. after all there is no builtin MLIR op for this case anyway, so how would you lower it to krnl (without this canonicalization patter, which imo shouldn't be responsible for legalization)?

AlexandreEichenberger: > i wonder if the int ^ float case should be legal in the first place.

I tend to agree with that statement, as `sqrt `(aka `x ^ 0.5`) tends to give floating point numbers, that would then be rounded to an int? That sounds not that plausible. There is a discussion in the onnx/operator slack channel, feel free to provide your opinion there.

Maybe we could keep the current implementation as is at this time, @gramalingam seems to indicate that this policy is also the one of ORT. If/when the specs are clarified, we can then change it to the better defined standard.

What do you think @srcarroll @tungld ?

tungld: If it is consistent with the one in ORT, I think it's fine to keep it as is. If we can put a memo on this case somewhere, it would be helpful in the future.

srcarroll: I agree

AlexandreEichenberger: Did not add an explicit "TODO", can add one in the next (unrelated) PR if you feel it would be a good thing.

Issue #3169, opened 2025-05-19
Add Support for LpNormalization in Krnl Dialect


jagadeeshvx:
The **LpNormalization** operation is currently supported in the ONNX dialect, but it does not yet have support in the Krnl dialect. The goal of this task is to implement the lowering of the LpNormalization operation from the ONNX dialect to the Krnl dialect.

**LpNormalization Operation Overview:**
The LpNormalization operation applies the following element-wise formula to the input tensor:
 
```
if p == 1   ->   Y =  x / sum (abs(x), axis)
if p == 2   ->   Y =  x / sqrt(sum((x^2),axis))
```
**Where:**

Attributes
**axis - INT (default is '-1'):**
The axis on which to apply normalization, -1 mean last axis.
**p - INT (default is '2'):**
The order of the normalization, only 1 or 2 are supported.

**Inputs**
input (heterogeneous) - T:
**Outputs**
output (heterogeneous) - T:

References:
https://onnx.ai/onnx/operators/onnx__LpNormalization.html

@AlexandreEichenberger @tungld  I would like to contribute to this.

Note:
The backend node testing is missing for this op:https://github.com/onnx/onnx/tree/b8baa8446686496da4cc8fda09f2b6fe65c2a02c/onnx/backend/test/case/node

AlexandreEichenberger:
Contributions would be greatly appreciated, thanks. Please look at some of the recent PRs adding Relu-like functionality as a template.

jagadeeshvx:
PR for enabling the backend testing for LpNormalization in onnx: https://github.com/onnx/onnx/pull/6997
cc: @AlexandreEichenberger

tungld:
@jagadeeshvx thank you for your active contribution!

AlexandreEichenberger:
> The backend test cases for the LpNormalization operator appear to be missing from the [ONNX backend test directory](https://github.com/onnx/onnx/tree/main/onnx/backend/test/case/node). Is there an alternative way to perform numerical validation for the onnxToKrnl lowering of LpNormalization in ONNX-MLIR?

Indeed it is missing from the ONNX standard. I would open an issue and ping the Operator SIG lead just to get their opinions
https://github.com/onnx/sigs/tree/main/operators They are on the ONNX slack channel.

PR #3176, opened 2025-05-24
Adding Lowering for LpNormalization op from onnx to krnl

jagadeeshvx:
This pull request implements the lowering of the LpNormalization op from the ONNX dialect to the Krnl dialect, and add tests to validate it, addressing Issue: https://github.com/onnx/onnx-mlir/issues/3169

This PR is ready for review. Please let me know if there are any additional improvements or adjustments needed.

Closes https://github.com/onnx/onnx-mlir/issues/3169.

### DIFF
--- a/src/Accelerators/NNPA/Dialect/ZLow/ZLow.td
+++ b/src/Accelerators/NNPA/Dialect/ZLow/ZLow.td
@@ -132,9 +132,9 @@ def ZLowInvSqrtOp:ZLow_Op<"invsqrt", [MemRefsNormalizable]> {
   let description = [{
   ZLow operation to perform a invsqrt.
   }];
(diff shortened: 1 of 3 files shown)


tungld:

tungld:
@jagadeeshvx  I have no objection on the current lowering to krnl since we can fuse some computation into a single loop. However, it has not been simdized yet so it would be slow even when we compile with -O3. If you decompose LpNormalization into simpler ONNX ops like Div, Sum those are already simdized, we can have a better performance with -O3 compile option. What do you think?

AlexandreEichenberger (approved):
LGTM, it will probably not generate fast CPU code as the LPNorm is implemented as a series of smaller ONNX operations, so let's keep that in mind if that operation becomes a performance bottleneck.

PR #3179, opened 2025-05-27
Remove most warnings (includes building for NNPA)

AlexandreEichenberger:
I saw one warning where we return a stack object. It's in the `toTensors`, and I fixed this using Tong's suggestion of returning a `SmallVector<Types, 4>`.

Then there were also 2 warnings, associated with creating arrays on the stack with a runtime size:
```
   int num = ...
   char buff[num]
```
which is not supported by all compilers. Also, if `num` is large, we may run out of stack. Replaced both instance with dynamic allocation asserting if results of malloc are null.


tungld (approved):
LGTM. Thanks!

chentong319 (approved):
LGTM!

PR #3182, opened 2025-05-30
Don't use onnx_proto

cyyever:


AlexandreEichenberger:
@cyyever Can you provide a short explanation for the desired change? Thanks

cyyever:
@AlexandreEichenberger Sorry for missing that, I have added.

AlexandreEichenberger:

chentong319 (approved):
LGTM!

chentong319:
There is no real change to the build of onnx-mlir because onnx target is built anyway. Right?

gongsu832:
We use `make onnx_proto` because all we need is the generated `onnx-{data|ml|operator-ml}.pb.cc` files for the cppcheck upfront. Using `make onnx` just wastes time building the entire `onnx`.

chentong319:
> We use `make onnx_proto` because all we need is the generated `onnx-{data|ml|operator-ml}.pb.cc` files for the cppcheck upfront. Using `make onnx` just wastes time building the entire `onnx`.

You are right in theory. I checked the build record: build onnx is quick.
By the way, there is something weird to me in the build. I am working on building the light weight PyRuntimeC, which excludes building of other part of the compiler, but PyRuntimeC and its dependencies. In a clean build directory, I start with `cmake -DONNX_MLIR_ENABLE_PYRUNTIME_LIGHT=ON ..`. 
1. If I execute `cmake --build . --target PyRuntimeC`, I got an error for "onnx/onnx-ml.pb.h not found"
2. If I execute `cmake --build . --target onnx_proto PyRuntimeC`, only libonnx_proto.so are built for PyRuntimeC*.so
3. If I execute 'cmake --build .`, both libonnx.so and Plibonnx.so are built
For the case 1 and 2: it sees that the `add_dependencies(PyRuntimeC onnx_proto)`  does not work as expected
For case 3: onnx-mlir has been building libonnx.so before this PR. 
Any idea about why?

gongsu832:
It doesn't take a lot of time to build `onnx` but that's not the point. If we don't pay attention to a little waste here and a little waste there, things quickly add up.

The "onnx/onnx-ml.pb.h not found" error is because `PyRuntimeC` also depends on `OMPyExecutionSessionBase`. The error occurred when it's building `OMPyExecutionSessionBase`, which needs the .h file but you didn't tell cmake `OMPyExecutionSessionBase` depends on `onnx_proto`.

chentong319:
> It doesn't take a lot of time to build `onnx` but that's not the point. If we don't pay attention to a little waste here and a little waste there, things quickly add up.
> 
"Now onnx_proto is planned to be deprecated." We have no choice.

> The "onnx/onnx-ml.pb.h not found" error is because `PyRuntimeC` also depends on `OMPyExecutionSessionBase`. The error occurred when it's building `OMPyExecutionSessionBase`, which needs the .h file but you didn't tell cmake `OMPyExecutionSessionBase` depends on `onnx_proto`.

Thanks. I tried to add the dependence to OMPyExecutionSessionBase and it works.

cyyever:
Indeed the deprecation is under evaluation, because the onnx_proto target is now part of unofficial public API...

AlexandreEichenberger:

chentong319:

chentong319:

justinchuby:
@cyyever if users only require onnx protos in their projects, does it make sense to require building the whole of onnx? I find some of the arguments in this thread legitimate.

cyyever:
@justinchuby Generally speaking, it is a question of whether it's possible to use generated code alone. I think it's better to link the whole library to avoid the risk of missing key context. Also note that @xadupre has proposed to drop Protobuf in https://github.com/onnx/onnx/pull/7208.

justinchuby:
I think just using the generated code alone is very reasonable. This gives consumers the ability to read write onnx files without having to use any other features in onnx. Granted I imagine it’s also easy to just take the onnx proto and run protoc directly…

PR #3191, opened 2025-06-04
Added support for negative permute values in transform

AlexandreEichenberger:
Add support in shape inference and canonicalizer to support negative permute values, to be interpreted as indices from the right hand side of the tensor.

Added lit tests and verified separately that we can process an input file with negative permute values.


chentong319 (approved):
LGTM!

PR #3192, opened 2025-06-04
Parallel fixes v1

AlexandreEichenberger:
Speedup the transpose (non-blocked version, aka the one using individual load and store) by
* iterating over the output instead of the input dimensions 
* unrolling the inner loop

Check the pairs of bars for the ONNX Transpose op
image


tungld (approved):
LGTM. Very neat code. Thanks!

tungld:
@AlexandreEichenberger I am curious about which changes made the other ops (stick, unstick, layernorm) slower?

AlexandreEichenberger:
> @AlexandreEichenberger I am curious about which changes made the other ops (stick, unstick, layernorm) slower?

still investigating, we added saturation but when I ran without saturation it was slower. Work is ongoing

PR #3194, opened 2025-06-06
added support for uniform_random op with shape handling

### LINKED ISSUE #3193: Add Support for RandomUniform Op
This issue tracks the end-to-end implementation of a RandomUniform operation in ONNX-MLIR, including:

Defining a new RandomUniform op in the Krnl dialect .

Lowering the KrnlRandomUniformOp (Krnl → LLVM, emitting calls to a runtime uniform-distribution function).

Lowering the ONNX RandomUniform op down into KrnlRandomUniformOp (ONNX → Krnl) with shape inference and type inference handling.

**RandomUniform Operation Overview**

Generate a tensor with random values drawn from a uniform distribution within the specified range [low,high].

**Attributes**
dtype (INT, default = 1)
The element type of the output tensor. If not specified, defaults to TensorProto::FLOAT (value 1).

high (FLOAT, default = 1.0)
Upper bound (exclusive) of the uniform distribution.

low (FLOAT, default = 0.0)
Lower bound (inclusive) of the uniform distribution.

seed (FLOAT, optional)
Seed value for the random generator. If omitted, a nondeterministic/auto-generated seed is used.

shape (INTS, required)
1D list of integer dimensions specifying the shape of the output tensor (e.g., [d0, d1, …, dn]).

**Outputs**
output (tensor of type T)
A tensor of random values drawn from Uniform(low, high). The element type T corresponds to dtype.

I would like to contribute to this

LekkalaSravya3:
This PR adds support for lowering the ONNXRandomUniformOp from the ONNX dialect to the Krnl dialect in ONNX-MLIR.
Defined the KrnlRandomUnifrom Op in the Krnl Dialect 
Lowered the KrnlRandomUnifrom Op from the Krnl Dialect to the LLVM Dialect .
Added tests to validate the correctness of the lowered Krnl , LLVM and shape inference implementation.

This implementation addresses and closes  #3193


LekkalaSravya3:
Hi @tungld,
When you have some time, could you please take a look at this PR and let me know if any changes are needed in the lowering of the UniformRandom op?

Thank you!

LekkalaSravya3:
@AlexandreEichenberger, In the current test infrastructure, when a seed value is explicitly provided, the generated values are deterministic and consistent across runs. However, if the seed is not specified, in the current implementation i had choose to generate a random value at runtime .
I do agree that using a fixed default seed when none is provided would improve reproducibility, which is especially important for consistent test results. I referred to the implementation of RandomNormal to understand how seed values are handled and followed a similar approach by introducing randomness when the seed is absent.
If the preference is to instead use a fixed seed value , I will update the implementation accordingly. Please let me know what direction we’d like to take, and I’ll make the necessary adjustments.

chentong319:
@LekkalaSravya3 There is an alternative way to low an ONNX op to a call of runtime library. Will you want to try it?
ONNX Op can be lowered to a KrnlCallOp and the lowering of KrnlCallOp to llvm is already there. The benefit is that you only need to provide the lowering of onnx to krnl for uniform_random, which usually is quite straightforward. No need to introduce new krnl op, nor the lowering of that op to llvm. It is the preferred approach in onnx-mlir. Compared with your current implementation, you will need to change your runtime library to take in OMTensor as parameters for input/output. If you are interested in the approach, we can discuss in more details.

AlexandreEichenberger:
>  I referred to the implementation of RandomNormal to understand how seed values are handled and followed a similar approach by introducing randomness when the seed is absent.

Thanks, I believe we had a discussion in the ONNX Operator SIG about this and also left it unresolved. If you use the same approach as currently implemented, then that is good enough for now.

AlexandreEichenberger:

LekkalaSravya3:
> @LekkalaSravya3 There is an alternative way to low an ONNX op to a call of runtime library. Will you want to try it? ONNX Op can be lowered to a KrnlCallOp and the lowering of KrnlCallOp to llvm is already there. The benefit is that you only need to provide the lowering of onnx to krnl for uniform_random, which usually is quite straightforward. No need to introduce new krnl op, nor the lowering of that op to llvm. It is the preferred approach in onnx-mlir. Compared with your current implementation, you will need to change your runtime library to take in OMTensor as parameters for input/output. If you are interested in the approach, we can discuss in more details.

@chentong319 Yes, I’m interested in trying this approach. I’ll look into implementing the lowering of ONNXUniformRandomOp to KrnlCallOp as suggested.
Also, I noticed that the helper function omTensorCreateWithRandomData internally generates a random seed and doesn’t take a seed as a parameter. For reproducibility, would it be advisable to create a similar helper function that accepts a seed as an argument 
Thank you for the suggestion !

LekkalaSravya3:
> Thanks, I believe we had a discussion in the ONNX Operator SIG about this and also left it unresolved. If you use the same approach as currently implemented, then that is good enough for now.

Thank you for your review and feedback on the PR!

LekkalaSravya3:
Hi @chentong319, @AlexandreEichenberger , @tungld 

In the KrnlCallOp, it currently supports only scalar outputs and does not yet support memref outputs. If we use OMTensor, the output will be in tensor format. However, I’m unsure whether it is correct to create an iterator and invoke the runtime function for each element to store the scalar values individually. is that good to go with current implementation or the alternative way

chentong319:
> Hi @chentong319, @AlexandreEichenberger , @tungld
> 
> In the KrnlCallOp, it currently supports only scalar outputs and does not yet support memref outputs. If we use OMTensor, the output will be in tensor format. However, I’m unsure whether it is correct to create an iterator and invoke the runtime function for each element to store the scalar values individually. is that good to go with current implementation or the alternative way

It's my fault that the specification in Krnl.td is kind of misleading. It's true that in current implementation the return of the krnlCall can only be scalar, and it can not create a memref as return.  However, if you create the memref before the krnlcall, you can pass the result memref as the first N parameters to make sure the correctness for dataflow analysis .
You may use one of the defined builder:
```
OpBuilder<(ins "mlir::StringAttr":$funcNameStr, "IntegerAttr":$numOfOutput, "mlir::ValueRange":$operands)>,

OpBuilder<(ins "std::string":$funcNameStr, "mlir::ValueRange":$results, "mlir::Operation *":$op, "mlir::ValueRange":$operands, "std::vector<std::string>":$attributeNames)>,
```
The value of N can be explicitly pass as   "IntegerAttr":$numOfOutput, or inferred with "mlir::ValueRange":$results.

LekkalaSravya3:
Hi @chentong319,
Thank you for your time and explanation. I have updated the lowering of the UniformRandom operation using the KrnlCallOp. Please have a look when you get a chance.

AlexandreEichenberger:
@chentong319 Can you review this PR as its using your new infra? Thanks

src/Conversion/ONNXToKrnl/Math/RandomUniform.cpp
```diff
@@ -0,0 +1,68 @@
...
+    double low = adaptor.getLow().convertToDouble();
+    Value lowValue = create.math.constant(rewriter.getF32Type(), low);
+    auto seed = adaptor.getSeed();
+    srand(time(NULL));
+    double doubleSeed = rand() % 100;
+    if (seed)
+      doubleSeed = seed->convertToDouble();
+    Value seedValue = create.math.constant(rewriter.getF32Type(), doubleSeed);
+
+    SmallVector<Value, 5> operands = {alloc, lowValue, highValue, seedValue};
```
chentong319, line 54: The implementation looks correct to me. However, krnlCall builder supports automatically propagating attributes of onnx op to krnlCall. You do not need to handle high and low by yourself. I noticed that seed attribute is optional. Any other reason that the attributes have to be manually handled? Or lack of documentation?

src/Dialect/ONNX/ONNXOps.td.inc
```diff
@@ -6222,7 +6222,7 @@ def ONNXRandomNormalLikeOp:ONNX_Op<"RandomNormalLike",
 }
 
 def ONNXRandomUniformOp:ONNX_Op<"RandomUniform",
-  [Pure, DeclareOpInterfaceMethods<ShapeInferenceOpInterface>, DeclareOpInterfaceMethods<ShapeHelperOpInterface>]> {
+  [Pure, DeclareOpInterfaceMethods<ShapeInferenceOpInterface>, DeclareOpInterfaceMethods<ShapeHelperOpInterface>, DeclareOpInterfaceMethods<ResultTypeInferenceOpInterface>]> {
```
chentong319, line 6225: Need to add this Interface? onnx ops are pure functional ops.

src/Conversion/ONNXToKrnl/Math/RandomUniform.cpp
LekkalaSravya3: > @AlexandreEichenberger, In the current test infrastructure, when a seed value is explicitly provided, the generated values are deterministic and consistent across runs. However, if the seed is not specified, in the current implementation i had choose to generate a random value at runtime .
I do agree that using a fixed default seed when none is provided would improve reproducibility, which is especially important for consistent test results. I referred to the implementation of RandomNormal to understand how seed values are handled and followed a similar approach by introducing randomness when the seed is absent.
If the preference is to instead use a fixed seed value , I will update the implementation accordingly. Please let me know what direction we’d like to take, and I’ll make the necessary adjustments.

Yes, I agree 
However, as discussed earlier with Alexandre , I chose to manually handle the seed attribute .
Since the seed handling influences the behavior of the op, I ended up manually passing all relevant attributes (including high and low) through the krnl.call for completeness.

src/Dialect/ONNX/ONNXOps.td.inc
LekkalaSravya3: Yes, similar to #3091 where they added interface and improved type inference for RandomNormal Ops, the goal was to override model-specified output types using the dtype attribute, to ensure correctness. Since `RandomUniform` has the same behavior — it uses dtype to control output type — I thought to add the interface and handle type inference the same way to ensure consistency .

Let me know your thoughts !

chentong319: You are right. This interface is needed.

chentong319 (approved):
LGTM!

PR #3195, opened 2025-06-09
RunONNXModel test that saved and loaded models use the same compiler options

AlexandreEichenberger:
Just making sure that when we save a compiled model (`--save-model`) and when we reload the model (`--load-model`), we reuse the same compiler options.

A new file is created when saving a model, `compiler-option.txt` which save the options.

Optionally, we could make it that when we load a model and don't give compiler options, we could reuse the old ones.


utils/RunONNXModel.py
```diff
@@ -747,6 +747,18 @@ def __init__(self, model_file=None, **kwargs):
         # Otherwise, compile the ONNX model.
         if args.load_model:
             self.model_dir = args.load_model
+            compiler_option_file = os.path.join(self.model_dir, "compiler-option.txt")
+            with open(compiler_option_file, "r") as f:
+                options_from_file = f.read()
+                if options_from_file != args.compile_args:
```
tungld: My suggestion is to check string equality only if both `compiler-option.txt` and `args.compile_args` are not empty because:
- we sometimes compile a model manually and there is no `compiler-option.txt` but only `model.so`, and
- when we load a saved model, typically there is no need to input `--compile_args`.

AlexandreEichenberger: implemented, make sense.

utils/RunONNXModel.py
```diff
@@ -651,6 +651,10 @@ def data_without_top_bottom_quartile(data, percent):
     return data[trim:-trim]
 
 
+def cache_string(model_name, compile_option):
```
AlexandreEichenberger, line 654: Added the model name too for good measure.
I also found that its useful to look at this file to remember the compile options that were used in a measurement.

utils/make-report.py
```diff
@@ -506,7 +506,7 @@ def make_report(stat_message):
                     count_time_str += ", {:.7f}".format(det_time * time_unit)
                     count_time_str += ", {:.1f}%".format(get_percent(det_time, time))
 
-                det_output = "\n    " + count_time_str + ": " + det_key
+                det_output = "\n    " + count_time_str + ", " + det_key
```
AlexandreEichenberger, line 509: Found that having comma separated for all fields were better for inspection with excel.

utils/RunONNXModel.py
```diff
@@ -747,6 +751,23 @@ def __init__(self, model_file=None, **kwargs):
         # Otherwise, compile the ONNX model.
         if args.load_model:
             self.model_dir = args.load_model
+            compiler_option_file = os.path.join(self.model_dir, "compiler_option.txt")
+            # Verify that we have the same options:
+            #  if we have saved the options in the "compiler_option.txt" file, and
+            #  if we have provided compiler options
+            if args.compile_args and os.path.exists(compiler_option_file):
```
tungld, line 758: Thanks for checking these!

PR #3199, opened 2025-06-12
[NNPA] Support multiple NNPAs

tungld:
This patch updates the zdnn extension to support multiple NNPAs using OpenMP.

This patch adds two cmake variables: `ZDNNX_WITH_OMP` to turn on/off multiple NNPA support and `ZDNNX_DEBUG` to show debug info for the zdnn extenstion.

By default, multiple NNPA support is not enabled. To enable it, set ZDNNX_WITH_OMP to ON when building onnx-mlir, e.g. `cmake .. -DONNX_MLIR_ACCELERATORS=NNPA -DZDNNX_WITH_OMP=ON`

I compared this PR with the main branch to see if there is any performance regression when OMP is OFF (by default), the performance difference is < 0.5%. See the following figure:
image


src/Accelerators/NNPA/NNPAAccelerator.cpp
```diff
@@ -58,7 +58,13 @@ NNPAAccelerator::NNPAAccelerator() : Accelerator(Accelerator::Kind::NNPA) {
 
   acceleratorTargets.push_back(this);
   // Order is important! libRuntimeNNPA depends on libzdnn
-  addCompilerConfig(CCM_SHARED_LIB_DEPS, {"RuntimeNNPA", "zdnn"}, true);
+  std::vector<std::string> libs;
+  libs.push_back("RuntimeNNPA");
+  libs.push_back("zdnn");
+#ifdef ZDNNX_WITH_OMP
+  libs.push_back("gomp");
```
AlexandreEichenberger: why using Gnu OpenMP. It is never recommended to use two distinct OpenMP runtime at the same time. Please use libomp from LLVM

AlexandreEichenberger: It can be build in LLVM, and already get linked in onnx-mlir when libomp is built in LLVM. 

```
   -DLLVM_ENABLE_PROJECTS="mlir;clang;openmp" \
``` 
to build LLVM (think clang can be omitted).

tungld: I changed to using ompruntime built by LLVM which is used by `--parallel` flag also. Thanks for the suggestion!

src/Accelerators/NNPA/Runtime/zdnnx/omp_ops.c
```diff
@@ -0,0 +1,617 @@
...
+  printf("[zdnnx] num_zaius: %d\n", zdnnx_num_zaius);
+  return zdnnx_num_zaius;
+}
+
+uint32_t get_min_num_procs_per_zaiu() {
+  if (zdnnx_min_num_procs_per_zaius > 0)
+    return zdnnx_min_num_procs_per_zaius;
+
+  zdnnx_min_num_procs_per_zaius = (uint32_t)-1;
+  for (int i = 0; i < omp_get_num_places(); ++i) {
```
AlexandreEichenberger, line 56: Nit, if there are no places, then the return number of process per place is +inf... that seems wrong.

src/Accelerators/NNPA/Runtime/zdnnx/omp_ops.c
```diff
@@ -0,0 +1,617 @@
...
+    uint32_t num_procs = omp_get_place_num_procs(i);
+    if (num_procs == 0) {
+      zdnnx_min_num_procs_per_zaius = 1;
+      break;
+    }
+    if (num_procs < zdnnx_min_num_procs_per_zaius)
+      zdnnx_min_num_procs_per_zaius = num_procs;
+  }
+  printf(
```
AlexandreEichenberger, line 66: Does the `printf` need to be guarded by 
```
#ifdef ZDNNX_DEBUG
```

src/Accelerators/NNPA/Runtime/zdnnx/omp_ops.c
```diff
@@ -0,0 +1,617 @@
...
+  return zdnnx_min_num_procs_per_zaius;
+}
+
+uint32_t get_num_procs() {
+  if (zdnnx_num_procs > 0)
+    return zdnnx_num_procs;
+
+  zdnnx_num_procs = 0;
+  for (int i = 0; i < omp_get_num_places(); ++i)
+    zdnnx_num_procs += omp_get_place_num_procs(i);
```
AlexandreEichenberger, line 78: Nit, for some reason, min process for places with 0 process in it was set to 1 in `get_min_num_procs_per_zaiu` but 0 here.

AlexandreEichenberger: I was not aware that we can generate places with 0 props in it... seems king of a corner case.

src/Accelerators/NNPA/Runtime/zdnnx/omp_ops.h
```diff
@@ -0,0 +1,51 @@
...
+
+#include "zdnn.h"
+#include "zdnnx_ops.h"
+
+// Keep these values to avoid calling omp functions multiple times.
+extern uint32_t zdnnx_num_zaius;
+extern uint32_t zdnnx_min_num_procs_per_zaius;
+extern uint32_t zdnnx_num_procs;
+
+// The number of zaius is the number of omp places.
```
AlexandreEichenberger: Since it's a c file, we cannot use namespaces. I wonder if we should adopt a uniform `zdnnz_` prefix to avoid name conflicts?

Or if these are not exported (which I believe to be the case) but only used internally, then I wonder if we should make them C++ files and using namespaces, and only the file exporting the functionality should be a C file?

src/Accelerators/NNPA/Runtime/zdnnx/omp_ops.c
```diff
@@ -0,0 +1,617 @@
...
+      &si_x, &input_view, ts_e4, ts_e3, ts_e2, ts_e1, "UnaryElementwise X");
+  zdnnx_prepare_split_info(
+      &si_y, &output_view, ts_e4, ts_e3, ts_e2, ts_e1, "UnaryElementwise Y");
+
+  // No splitting, call the zdnn op without any changes.
+  if (zdnnx_has_one_tile(&si_x)) {
+    zdnn_status status;
+    if (op_type == ZDNNX_EXP_OP)
+      status = zdnn_exp(input, output);
+    else if (op_type == ZDNNX_GELU_OP)
```
AlexandreEichenberger, line 342: Question: I believe the GELU is only supported on the most recent hardware (zdnn info below).

```
the underlying hardware supports zDNN APIs 1.1.x or later at runtime
```

But presumably it is ok here since the compiler would only generate GELU when activated with the `-march=z17` flag?

src/Accelerators/NNPA/Runtime/zdnnx/omp_ops.c
```diff
@@ -0,0 +1,617 @@
...
+      status = ZDNN_UNAVAILABLE_FUNCTION;
+    return status;
+  }
+
+  // Call zdnn op on each tile.
+  uint32_t num_tiles_e4 = zdnnx_get_num_tiles(&si_x, E4);
+  uint32_t num_tiles_e3 = zdnnx_get_num_tiles(&si_x, E3);
+  uint32_t num_tiles_e2 = zdnnx_get_num_tiles(&si_x, E2);
+  uint32_t num_tiles_e1 = zdnnx_get_num_tiles(&si_x, E1);
+#pragma omp parallel for num_threads(num_tiles) collapse(4)
```
AlexandreEichenberger, line 366: Out of curiosity, have you tried the perf with `collapse(4)` vs `collapse(3)`?
There are some div/reminder (esp if the dims are not know to be multiple of powers of 2), so it might be faster to do a `collapse(3)`.

But I understand that the cost of the innermost loop can be significant (aka one zdnn call), so the overheads might be minimal with respect to the work actually being done by the loop.

src/Accelerators/NNPA/Runtime/zdnnx/zdnnx_ops.h
```diff
@@ -0,0 +1,119 @@
...
+zdnn_status zdnn_matmul_op_ext(const zdnn_ztensor *inputA,
+    const zdnn_ztensor *inputB, const zdnn_ztensor *inputC, int opType,
+    zdnn_ztensor *output);
+
+zdnn_status zdnn_matmul_bcast_op_ext(const zdnn_ztensor *inputA,
+    const zdnn_ztensor *inputB, const zdnn_ztensor *inputC, int opType,
+    zdnn_ztensor *output);
+
+// Elementwise Operations
+zdnn_status zdnn_add_ext(const zdnn_ztensor *inputA, const zdnn_ztensor *inputB,
```
AlexandreEichenberger: Do we want to name the extensions using the prefix `zdnn` as `znddx`,  instead of adding a postfix `_ext`?

src/Accelerators/NNPA/Runtime/zdnnx/zdnnx_ops.c
```diff
@@ -0,0 +1,479 @@
...
+zdnn_status zdnn_matmul_op_ext(const zdnn_ztensor *input_a,
+    const zdnn_ztensor *input_b, const zdnn_ztensor *input_c, int op_type,
+    zdnn_ztensor *output) {
+  zdnn_status status;
+#ifdef ZDNNX_DEBUG
+  START_TIMING;
+#endif
+
+#ifdef ZDNNX_WITH_OMP
+  if (get_num_procs() == 1)
```
AlexandreEichenberger: if an inlined `get_num_procs` where to respond `1` when `ZDNNX_WITH_OMP` if not defined, then we could avoid this special casing.

One possibility in the header file
```
int get_num_proc() {
# ifndef ZDNNZ_WITH_OMP
  return 1;
#else
  return get_num_proc_from_omp();
endif
}
```

src/Accelerators/NNPA/Runtime/zdnnx/zdnnx_ops.h
```diff
@@ -0,0 +1,119 @@
...
+//===----------------------------------------------------------------------===//
+
+#ifndef ZDNNX_ZDNNX_OPS_H
+#define ZDNNX_ZDNNX_OPS_H
+
+#include "zdnn.h"
+#include "zdnnx.h"
+
+// Keep these values to avoid calling zdnn functions multiple times.
+extern uint32_t nnpa_max_dim_size;
```
AlexandreEichenberger: Good idea; I believe z17 has 2 max dims depending on the dimension.

src/Accelerators/NNPA/Runtime/zdnnx/zdnnx_ops.c
```diff
@@ -0,0 +1,479 @@
...
+
+  CHECK_ZDNN_STATUS(status, "zdnn_matmul");
+  return status;
+}
+
+zdnn_status zdnn_matmul_bcast_op_ext(const zdnn_ztensor *input_a,
+    const zdnn_ztensor *input_b, const zdnn_ztensor *input_c, int op_type,
+    zdnn_ztensor *output) {
+  zdnn_status status;
+#ifdef ZDNNX_DEBUG
```
AlexandreEichenberger: Nit: the `START_TIMING` could become nothing when `ZDNNX_DEBUG` is not defined. I also prefer using the `()` for the macro to be more similar to a function since it may invoke code.

src/Accelerators/NNPA/Runtime/zdnnx/zdnnx_ops.h
```diff
@@ -0,0 +1,119 @@
...
+  ZDNNX_EXP_OP = 50,
+  ZDNNX_GELU_OP = 51,
+  ZDNNX_INVSQRT_OP = 52,
+  ZDNNX_LEAKY_RELU_OP = 53,
+  ZDNNX_LOG_OP = 54,
+  ZDNNX_RELU_OP = 55,
+  ZDNNX_SIGMOID_OP = 56,
+  ZDNNX_SQRT_OP = 57,
+  ZDNNX_TANH_OP = 58,
+} ElemementwiseOp;
```
AlexandreEichenberger: This does not seems to be used in the header file. Should it move to `zdnnx_private.h`

src/Accelerators/NNPA/Runtime/zdnnx/omp_ops.c
tungld: I checked the situation when OMP_PLACES is not set: `omp_get_num_places` returns 1, and `omp_get_place_num_procs(0)` (the number of process per place 0) returns the maximum number of procs in the machine. So no issue here I guess.

src/Accelerators/NNPA/Runtime/zdnnx/omp_ops.c
tungld: Yes, I added `#ifdef ZDNNX_DEBUG`.

src/Accelerators/NNPA/Runtime/zdnnx/omp_ops.c
tungld: I think so, it only generates GELU if `-march=z17`.

src/Accelerators/NNPA/Runtime/zdnnx/omp_ops.c
tungld: I did a check for tanh op, there was no difference in performance. I guess it's because I created a 1D view of the tensor so, collapse(4) and collapse(3) would be the same.

src/Accelerators/NNPA/Runtime/zdnnx/omp_ops.h
tungld: Yes, added `zdnnx_`

src/Accelerators/NNPA/Runtime/zdnnx/zdnnx_ops.c
tungld: It's a good idea, I tried it. However, I realized that the issue is in the following code
```c
  if (get_num_procs() == 1)
    status = seq_matmul(
        input_a, input_b, input_c, op_type, output, /*is_bcast=*/false);
  else
    status = omp_matmul(input_a, input_b, input_c, op_type, output,
        /*is_bcast=*/false);
```
where `omp_matmul` does not exist when `ZDNNX_WITH_OMP` is not defined.

Anyway, I solved it by defining a macro:
```c
#ifdef ZDNNX_WITH_OMP
#define CALL_ZDNNX_FUNC(msg, seq_func, omp_func, ...)                          \
  START_TIMING();                                                              \
  if (zdnnx_get_num_procs() == 1)                                              \
    status = seq_func(__VA_ARGS__);                                            \
  else                                                                         \
    status = omp_func(__VA_ARGS__);                                            \
  STOP_TIMING(msg);
#else
#define CALL_ZDNNX_FUNC(msg, seq_func, omp_func, ...)                          \
  START_TIMING();                                                              \
  status = seq_func(__VA_ARGS__);                                              \
  STOP_TIMING(msg);
#endif
```

src/Accelerators/NNPA/Runtime/zdnnx/zdnnx_ops.h
tungld: It's my mistake, forgetting to add them, now we have
```c
extern uint32_t nnpa_max_dim_size_e4;
extern uint32_t nnpa_max_dim_size_e3;
extern uint32_t nnpa_max_dim_size_e2;
extern uint32_t nnpa_max_dim_size_e1;
```

src/Accelerators/NNPA/Runtime/zdnnx/zdnnx_ops.c
tungld: Fixed it.

src/Accelerators/NNPA/Runtime/zdnnx/zdnnx_ops.h
tungld: Let me keep it here since it is used for defining sequential and omp operations, while `zdnnx_private.h` is for lowe-level defining primitive splitting interfaces.

src/Accelerators/NNPA/Runtime/zdnnx/zdnnx_ops.h
tungld: Yes, I changed them to `zdnnx_`. I had a TODO but forgot to do it. Thanks for the reminder!

src/Accelerators/NNPA/CMakeLists.txt
```diff
@@ -35,6 +35,16 @@ endif()
 include(zdnn.cmake)
 setup_zdnn(v1.1.2)
 
+option(ZDNNX_DEBUG "Enable ZDNNX debug informaiton" ON)
```
AlexandreEichenberger: For refs, at some time we may want to have a compile flag

AlexandreEichenberger: Maybe you should also indicate somewhere (if there is readme, or where its easy to find) that LLVM needs also to be built with `openmp` in its target, as otherwise we don't integrate it in onnx-mlir

src/Accelerators/NNPA/Runtime/zdnnx/zdnnx.h
```diff
@@ -0,0 +1,315 @@
...
+#ifndef ZDNNX_ZDNNX_H
+#define ZDNNX_ZDNNX_H
+
+#include "zdnn.h"
+
+// -----------------------------------------------------------------------------
+// Common structures
+// -----------------------------------------------------------------------------
+
+extern bool is_telum_1;
```
AlexandreEichenberger: since its a C file, I wonder if this should not also be predicated with a `zdnnx_`

src/Accelerators/NNPA/Runtime/zdnnx/zdnnx_ops.h
AlexandreEichenberger: I still find that it would be more elegant to only have the definitions of the ops here, as none of the definitions above are used in the op definitions. But if you feel strongly that it does not matter, its fine.

src/Accelerators/NNPA/Runtime/zdnnx/zdnnx_ops.c
```diff
@@ -0,0 +1,323 @@
...
+  }
+}
+
+static inline zdnn_status call_unary_op(const char *msg,
+    const zdnn_ztensor *input, const void *scalar_input, zdnn_ztensor *output,
+    ElemementwiseOp op_type) {
+  START_TIMING();
+
+  zdnn_status status;
+#ifdef ZDNNX_WITH_OMP
```
AlexandreEichenberger: One could get rid of these #defs by defining `zdnnx_get_num_procs` to zero when `ZDNNX_WITH_OMP` is 0. 

Maybe it would be slightly safer if testing with `<= 1` as I recall some places possibly returning 0... but its probably fine.

tungld:
> Also, we should consider our policy with dim unknown.. for that we would need to know the overheads so that we can take an informed decision.

@AlexandreEichenberger I measured the overheads and put the result into the main comment. In summary, for roberta, the performance difference is < 0.5%, sometimes one is faster, sometimes the other is faster, but the difference is very subtle.

src/Accelerators/NNPA/CMakeLists.txt
tungld: Now we check if only openmp is built in LLVM, then enable OMP here.

src/Accelerators/NNPA/Runtime/zdnnx/zdnnx.h
tungld: Sure, I changed all external functions to that they have `zdnnx_`.

src/Accelerators/NNPA/Runtime/zdnnx/zdnnx_ops.c
tungld: Done, only `zdnnx_get_num_procs` call is used now.

src/Accelerators/NNPA/Runtime/zdnnx/zdnnx_ops.h
tungld: I moved this to a private file. Thanks for the suggestion!

AlexandreEichenberger:
@tungld thanks for the measurements, much appreciated

PR #3207, opened 2025-06-24
Use greedy rewriter instead of partial conversion in recomposition pass and fuse locations when fusing convolutions. Also fix bugs in the parallel conv recomposition

jorickert:
To test the fusing of locations in the recomposition pass, I needed to convert it to a greedy rewriter. The partial conversion did not trigger as the convolutions where not marked as illegal ops.

In my experience the greedy rewriter is general easier to use then `applyPartialConversion`.

If you prefer it, I could split this PR into multiple for the different commits in it

Also fix bugs caused by missing preconditions checks and wrong insertion points

### DIFF
--- a/src/Accelerators/NNPA/Runtime/zdnnx/zdnnx.h
+++ b/src/Accelerators/NNPA/Runtime/zdnnx/zdnnx.h
@@ -178,12 +178,15 @@ void zdnnx_free_buffer(void *aligned_ptr);
  * @param split_info information for splitting
  * @param axis axis E1, E2, E3, or E4
  * @return the number of tiles.
- */
-inline uint32_t zdnnx_get_num_tiles(
+ * @note For clang compilers, this function needs to be static inline if the
+ *implementation is in a header file. Otherwise, it will cause errors when
+ *linking multiple object files that include this header file. This is a common
+ *issue with inline functions defined in header files.
+ **/
+static inline uint32_t zdnnx_get_num_tiles(
     zdnnx_split_info *split_info, zdnnx_axis axis) {
   return split_info->num_tiles[axis];
 }
-
 /**
  * \brief Check if there is only one tile in split_info.
  *
--- a/src/Compiler/CompilerPasses.cpp
+++ b/src/Compiler/CompilerPasses.cpp
@@ -255,6 +255,8 @@ void addKrnlToLLVMPasses(
   bufferization::BufferDeallocationPipelineOptions bufferDeallocOptions;
   mlir::bufferization::buildBufferDeallocationPipeline(
       pm, bufferDeallocOptions);
+  // This pass is necessary to move deallocation after the last user.
+  pm.addPass(mlir::bufferization::createOptimizeAllocationLivenessPass());
   pm.addPass(mlir::createConvertBufferizationToMemRefPass());
 
   // Late introduction of OpenMP, after bufferization.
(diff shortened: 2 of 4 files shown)


AlexandreEichenberger (commented):
LGTM, not an expert on the type of passes, will let @chentong319 or @tungld give their opinions.

jorickert:
Hello, the test TestGemmNNPA is failing in the s390x pipeline. I am a bit confused by this, as the code this PR touches, should not affect this test. As I am not familiar with these NNPA tests, it would be great if someone who is more familiar with them could take a quick lock at it

AlexandreEichenberger:

AlexandreEichenberger:
Trying to run the test again just in case it was a fluke.

src/Dialect/ONNX/Transforms/Recompose.cpp
```diff
@@ -851,44 +858,10 @@ void RecomposeONNXToONNXPass::runOnOperation() {
...
-        return !RecomposeQLinearMatMulFromQuantizeLinearPattern::
-            matchQLinearMatMulPattern(op, a, aScale, aZeroPoint, b, bScale,
-                bZeroPoint, outScale, outZeroPoint);
-      });
-
   RewritePatternSet patterns(context);
   onnx_mlir::getRecomposeONNXToONNXPatterns(patterns);
 
-  if (failed(applyPartialConversion(function, target, std::move(patterns))))
+  if (failed(applyPatternsGreedily(function, std::move(patterns))))
```
tungld, line 908: It is reasonable. Thanks!

tungld (approved):
LGTM! It is reasonable to use `applyPatternsGreedily` for rewriting patterns in the same dialect.

jorickert:
While working on this, I found out that this parallel-conv optimization pattern misses some precondition checks and sometimes generates an invalid graph/IR due to wrong insertion points. 
I will update this PR at the beginning of next week to include additional fixes

jorickert:
I updated this PR by including fixes for the following:
- Missing check for static types
- Missing check for no def-use chains between convs
- Wrong insertion point in rewrite, leading to invalid IR

jorickert:
@Arkar-Hema fyi

jorickert:

jorickert:
@AlexandreEichenberger @tungld I added additional fixes for the parallel-conv-fusion in this PR, it would be great if you could review

src/Dialect/ONNX/Transforms/Recompose.cpp
```diff
@@ -657,8 +656,13 @@ struct CombineParallelConv2DPattern : public OpRewritePattern<ONNXConvOp> {
       ONNXConvOp convOp1, PatternRewriter &rewriter) const final {
     Value input = convOp1.getX();
     if (!onnx_mlir::isRankedShapedType(input.getType()) ||
-        mlir::cast<ShapedType>(input.getType()).hasStaticShape() == false)
-      return failure();
+        !mlir::cast<ShapedType>(input.getType()).hasStaticShape())
+      return rewriter.notifyMatchFailure(
+          convOp1, "input must be a ranked tensor with static shape");
```
tungld, line 661: Thanks for adding many messages!

tungld (approved):
LGTM!

AlexandreEichenberger (approved):
LGTM

PR #3208, opened 2025-06-26
Enhance --shapeInformation for more cases

tungld:
This PR enhances `--shapeInformation` to support two cases:
1. Set a shape for all inputs and allow overriding for some inputs. For example: `-1:3x5,2:7x9x13` means all inputs except input 2 have shape of `3x5`, input 2 has shape `7x9x13`
2. Set a shape for a range of inputs. For example: `2-5:1x3x5` means that inputs 2, 3, 4, 5 have shape of `1x3x5`

The two cases can be combined and work together, for example:  `-1:3x5,2-5:7x9x13,6:5x7`


AlexandreEichenberger (approved):
LGTM, fun semantic to parse and write... Useful for models with 10+ shapes.

PR #3211, opened 2025-07-01
Add an optimization pattern to move Relu and LeakyRelu before Split operations. 

jorickert:
This potentially saves kernel invocations and makes it more likely that these activation functions can be fused with preceding ops.

Note: This optimization can be easily extended for all elementwise unary ops.

### DIFF
--- a/src/Dialect/ONNX/ONNXOps/Canonicalize.cpp
+++ b/src/Dialect/ONNX/ONNXOps/Canonicalize.cpp
@@ -2074,6 +2074,75 @@ public:
   }
 };
 
+// "Pulls" Relu-like operations up through a SplitOp
+struct PullReluLikeOpsThroughSplitPattern
+    : public OpRewritePattern<ONNXSplitOp> {
+  using OpRewritePattern<ONNXSplitOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(
+      ONNXSplitOp splitOp, PatternRewriter &rewriter) const final {
+
+    Operation *firstUser = nullptr;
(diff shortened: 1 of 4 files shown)


src/Dialect/ONNX/ONNXOps/Canonicalize.cpp
```diff
@@ -2334,6 +2403,13 @@ void ONNXSpaceToDepthOp::getCanonicalizationPatterns(
   results.insert<RemoveSpaceToDepthDepthToSpacePattern>(context);
 }
 
+/// on the ONNXSplitOp
+void ONNXSplitOp::getCanonicalizationPatterns(
+    RewritePatternSet &results, MLIRContext *context) {
+  results.insert<PullReluLikeOpsThroughSplitPattern>(context);
```
tungld: I wonder if it is easier to read if we are explicit here, say use a template instead:
```c
 results.insert<PullOpsThroughSplitPattern<ONNXReluOp>>(context);
 results.insert<PullOpsThroughSplitPattern<ONNXLeakyReluOp>>(context);
```
So that users don't need to look at `PullOpsThroughSplitPattern` to know which ops are supported. Also, there is no confusion that mix of ops is impossible.

tungld (approved):
LGTM! It looks a good rewriting. Thanks!

Just left one comment and please feel free to make a change if you think it is better.

src/Dialect/ONNX/ONNXOps/Canonicalize.cpp
jorickert: Thanks, thats a great suggestion

PR #3214, opened 2025-07-04
Fuse MatMul and Mul if one of the operands of Mul is a scalar constant

tungld:
This patch adds one optimization rule to Mul's canonicalization that fuses MatMul and Mul if one of the operands of Mul is a scalar constant.

This pattern is found in IBM granite models.


chentong319:
1. To be symmetrical, should the case in which $a is a ConstantOp be checked too?
2. I feel that this transformation will affect the precision of the computation. Right? I remember that we talked about adding a flag to control all the math unsafe transformations.

tungld:
@chentong319 I added the case where the first operand of Mul is a scalar. It seems that the case where the first operand of MatMul is a constant is not practical, I haven't seen it. But we can add if you feel it's really needed.

> I feel that this transformation will affect the precision of the computation. Right? I remember that we talked about adding a flag to control all the math unsafe transformations.

It does `x1 * C*B1 + x2 * C*A2` instead of `C * (x1 * A1 + x2 * A2)`. I think it would be safe.

chentong319 (approved):
LGTM!

PR #3218, opened 2025-07-10
Krnl iterate org indices

tungld:
This patch adds two functions in KrnlBuilder to build `krnl.iterate` with **original loop indices**, for example:

```c
create->krnl.iterateWithOrigLoop(loopDef, optimizedLoopDef, lbs, ubs,
        [&](const KrnlBuilder &createKrnl, ValueRange originalLoopIndices) {
     // Express computation using originalLoopIndices.
}

create->krnl.iterateIEWithOrigLoop(loopDef, optimizedLoopDef, lbs, ubs,
        [&](const KrnlBuilder &createKrnl, ValueRange originalLoopIndices) {
     // Express computation using originalLoopIndices.
}
``` 

By using original loop indices, we can optionally add optimizations such as blocking, unrolling **without changing the body of `krnl.iterate`**, avoiding writing the same computation twice.
This patch uses the new function for the lowering of Transpose as an example, and the generated IR is like:
```mlir
%1:4 = krnl.define_loops 4
%loop_block_1, %loop_local_2 = krnl.block %1#3 5 : (!krnl.loop) -> (!krnl.loop, !krnl.loop)
krnl.unroll %loop_local_2 : !krnl.loop
krnl.iterate(%1#0, %1#1, %1#2, %loop_block_1, %loop_local_2) with (%1#0 -> %arg1 = 0 to 64, %1#1 -> %arg2 = 0 to 10, %1#2 -> %arg3 = 0 to 30, %1#3 -> %arg4 = 0 to 20){
      %2 = krnl.load %alloc[%arg1, %arg3, %arg4, %arg2] : memref<64x30x20x10xf32>
      krnl.store %2, %alloc_0[%arg1, %arg2, %arg3, %arg4] : memref<64x10x30x20xf32>
}
```
and when it is lowered to affine, it's like this:
```mlir
    %alloc_0 = memref.alloc() {alignment = 16 : i64} : memref<64x10x30x20xf32>
    affine.for %arg1 = 0 to 64 {
      affine.for %arg2 = 0 to 10 {
        affine.for %arg3 = 0 to 30 {
          affine.for %arg4 = 0 to 20 step 5 {
            %0 = affine.load %alloc[%arg1, %arg3, %arg4, %arg2] : memref<64x30x20x10xf32>
            affine.store %0, %alloc_0[%arg1, %arg2, %arg3, %arg4] : memref<64x10x30x20xf32>
            %1 = affine.load %alloc[%arg1, %arg3, %arg4 + 1, %arg2] : memref<64x30x20x10xf32>
            affine.store %1, %alloc_0[%arg1, %arg2, %arg3, %arg4 + 1] : memref<64x10x30x20xf32>
            %2 = affine.load %alloc[%arg1, %arg3, %arg4 + 2, %arg2] : memref<64x30x20x10xf32>
            affine.store %2, %alloc_0[%arg1, %arg2, %arg3, %arg4 + 2] : memref<64x10x30x20xf32>
            %3 = affine.load %alloc[%arg1, %arg3, %arg4 + 3, %arg2] : memref<64x30x20x10xf32>
            affine.store %3, %alloc_0[%arg1, %arg2, %arg3, %arg4 + 3] : memref<64x10x30x20xf32>
            %4 = affine.load %alloc[%arg1, %arg3, %arg4 + 4, %arg2] : memref<64x30x20x10xf32>
            affine.store %4, %alloc_0[%arg1, %arg2, %arg3, %arg4 + 4] : memref<64x10x30x20xf32>
          }
        }
      }
    }
```

### DIFF
--- a/src/Dialect/Krnl/DialectBuilder.hpp
+++ b/src/Dialect/Krnl/DialectBuilder.hpp
@@ -78,6 +78,8 @@ struct KrnlBuilder : public DialectBuilder {
   using KrnlLoopBodyFn = impl::LoopBodyFn<KrnlBuilder>;
   using KrnlLoopBody2Fn = mlir::function_ref<void(
       const KrnlBuilder &, mlir::ValueRange, mlir::ValueRange)>;
+  using KrnlLoopBody3Fn = mlir::function_ref<void(const KrnlBuilder &,
+      mlir::ValueRange, mlir::ValueRange, mlir::ValueRange)>;
 
   void iterate(mlir::ValueRange originalLoops, mlir::ValueRange optimizedLoops,
       mlir::ValueRange lbs, mlir::ValueRange ubs,
@@ -95,6 +97,9 @@ struct KrnlBuilder : public DialectBuilder {
   void iterateIE(mlir::ValueRange originalLoops,
       mlir::ValueRange optimizedLoops, mlir::ArrayRef<IndexExpr> lbs,
       mlir::ArrayRef<IndexExpr> ubs, KrnlLoopBodyFn bodyBuilderFn) const;
+  void iterateIE(mlir::ValueRange originalLoops,
+      mlir::ValueRange optimizedLoops, mlir::ArrayRef<IndexExpr> lbs,
+      mlir::ArrayRef<IndexExpr> ubs, KrnlLoopBody2Fn bodyBuilderFn) const;
   // Deprecated.
   mlir::KrnlIterateOp iterateIE(mlir::ValueRange originalLoops,
       mlir::ValueRange optimizedLoops, mlir::ArrayRef<IndexExpr> lbs,
--- a/src/Dialect/Krnl/DialectBuilder.hpp
+++ b/src/Dialect/Krnl/DialectBuilder.hpp
@@ -83,9 +83,13 @@ struct KrnlBuilder : public DialectBuilder {
   void iterate(mlir::ValueRange originalLoops, mlir::ValueRange optimizedLoops,
       mlir::ValueRange lbs, mlir::ValueRange ubs,
       KrnlLoopBodyFn bodyBuilderFn) const;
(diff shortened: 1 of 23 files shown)


src/Dialect/Krnl/DialectBuilder.hpp
```diff
@@ -95,6 +100,9 @@ struct KrnlBuilder : public DialectBuilder {
   void iterateIE(mlir::ValueRange originalLoops,
       mlir::ValueRange optimizedLoops, mlir::ArrayRef<IndexExpr> lbs,
       mlir::ArrayRef<IndexExpr> ubs, KrnlLoopBodyFn bodyBuilderFn) const;
+  void iterateIE(mlir::ValueRange originalLoops,
```
AlexandreEichenberger: Because we are well acquainted with the normal function, but not so much with the new interface, I think this deserves an explanation here of how adding the second set of ValueRange to the lambda function changes things. Are there cases where both are needed? Or is it solely a matter of distinguishing the old and the new way?

tungld: The essential idea in krnl is to separate computation and schedule. I feel that the ultimate goal is to express optimizations by schedule ops like block, permute, unroll while keeping the iterate's body unchanged. However, I am not sure at this moment if SIMD can be totally expressed by schedule ops or not. So, during this transition period, I just keep both sets of ValueRange. Users can use one of them or both as long as the generated IR is valid (though I don't have an example of using both). Also, since we have two sets of ValueRange for originalLoop and optimizedLoop, it makes sense to have two sets of ValueRange for their indices, respectively.

tungld: @AlexandreEichenberger I changed the function name and use a single ValueRange for the original loop, which avoids confusion.

PR #3219, opened 2025-07-11
Fix the issue where deallocation is put at the end of a function

tungld:
When LLVM version was updated in onnx-mlir and the new bufferization was used, deallocations are put at the end of a function, causing a significant performance loss.

To move the deallocations after the last user instead of at the end of the function, we need to call `createOptimizeAllocationLivenessPass`. This patch adds that call to the pass manager, which would bring the performance back.


chentong319 (approved):
LGTM!

PR #3224, opened 2025-07-12
fix for s390x-z/OS zdnnx missing member

Sunny-Anand:
Fix for the below issue: 

```
IEW2456E 9207 SYMBOL zdnnx_get_num_tiles UNRESOLVED.  MEMBER COULD NOT BE      

          INCLUDED FROM THE DESIGNATED CALL LIBRARY.                            

 IEW2665S 40FF MODULE *NULL*  IS NON-EXECUTABLE AND WAS NOT SAVED BECAUSE       

          STORENX=NEVER.                                                        

IEW5033 The binder ended with return code 12.

ibm-clang: error: linker command failed with exit code 12 (use -v to see invocation)
```

### DIFF
--- a/src/Accelerators/NNPA/Runtime/zdnnx/zdnnx.h
+++ b/src/Accelerators/NNPA/Runtime/zdnnx/zdnnx.h
@@ -179,10 +179,7 @@ void zdnnx_free_buffer(void *aligned_ptr);
  * @param axis axis E1, E2, E3, or E4
  * @return the number of tiles.
  */
-inline uint32_t zdnnx_get_num_tiles(
-    zdnnx_split_info *split_info, zdnnx_axis axis) {
-  return split_info->num_tiles[axis];
-}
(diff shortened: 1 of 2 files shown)


src/Accelerators/NNPA/Runtime/zdnnx/zdnnx.h
```diff
@@ -179,10 +179,7 @@ void zdnnx_free_buffer(void *aligned_ptr);
  * @param axis axis E1, E2, E3, or E4
  * @return the number of tiles.
  */
-inline uint32_t zdnnx_get_num_tiles(
```
tungld, line 182: In [PR3015](https://github.com/onnx/onnx-mlir/pull/3015) @christopherlmunoz fixed the same issue by adding `static` in front of `inline`. I wonder if it can be applied here to keep `inline` which is good for better performance.

Sunny-Anand: Yeah, that seems like a better solution for the ibm-clang issue.

PR #3226, opened 2025-07-15
Explore the first two outermost dimensions for parallelism for Concat

tungld:
Currently the krnl code for Concat only supports parallelism over the outermost dimension, which is not good for Concat of KV cache in decoder models whose the 1st dimension is 1 (batchsize = 1). A good thing is that these KV cache Concat has the 2nd dimension of value of 8 or more. So, this patch explores the 2nd dimension for parallelism also.


AlexandreEichenberger (approved):
LGTM

PR #3227, opened 2025-07-16
[zdnnx] Use OMP_NUM_THREADS to control the number of threads for NNPA

tungld:
Since NNPA in z17 can be remotely accessed, the use of OMP_PLACES to manually assign CPU to threads is not necessary. This patch simplifies that by using OMP_NUM_THREADS which works with both compiler-generated OMP and zdnnx's OMP. 

Additionally, an environment variable `OM_NUM_ZAIU_THREADS` is introduced if users want to use a different number of threads for NNPA. Otherwise, the number of threads for NNPA is the same as OMP_NUM_THREADS.


src/Accelerators/NNPA/Runtime/zdnnx/omp_ops.c
```diff
@@ -73,13 +69,9 @@ uint32_t zdnnx_get_num_procs() {
...
-  zdnnx_num_procs = 0;
-  for (int i = 0; i < omp_get_num_places(); ++i)
-    zdnnx_num_procs += omp_get_place_num_procs(i);
-  if (zdnnx_num_procs == 0) {
-    // OMP_PLACES is not set. Return 1 to say using one processor.
-    zdnnx_num_procs = 1;
-  }
+  // If OMP_NUM_THREADS is not set, it is the sequential mode.
+  const char *env = getenv("OMP_NUM_THREADS");
+  zdnnx_num_procs = (env != NULL) ? atoi(env) : 1;
```
AlexandreEichenberger: What is the difference between `zdnnx_num_procs` and `zdnnx_num_zaius`. The names feel confusing and if they are two different entities, their name should be clearer.

tungld: Good catch! `zdnnx_num_procs` is not necessary any more. I just removed it and revised some variable names. Thanks!

chentong319:
Is omp_get_device_num more close to the number of NNPA in logic?

tungld:
> Is omp_get_device_num more close to the number of NNPA in logic?

I am not sure if NNPA is exposed as a OpenMP target device or not. Do you have any info or how to confirm that?

chentong319:
No, NNPA is not exposed to omp. We cannot use it. Moreover, number of devices is the hardware description, more like max_num_thread. I couldn't find a better way to describe the runtime parallelism on cores and zaiu seperately with omp. 
We may provide a driver lib for zAIUs in future for resource management if necessary.

chentong319 (approved):
LGTM!

AlexandreEichenberger (approved):
LGTM, thanks for removing the redundant variable

tungld:

tungld:

PR #3228, opened 2025-07-16
Compute static shape when collapsing innermost or outtermost dims

tungld:
The helper function 
```c
Value OnnxBuilder::reshapeToNDim(
    Value val, int64_t N, bool collapseMostSignificant) const {
```
always use -1 for the collapsed dimension.

However, when all the collapsing dimensions are static we can compute a literal for the collapsed dimension, which is good for compilation passes. 

This patch adds code to compute the static collapsed dimension whenever possible.


tungld:

AlexandreEichenberger (approved):
LGTM, thanks for the additional feature.

PR #3230, opened 2025-07-17
Add two patterns to remove unused zlow.stick and zlow.unstick operations

tungld:
ZLow operations are side-effect so they are not removed automatically if unused.

This patch adds two canonicalization patterns to remove zlow.stick and zlow.unstick operations, which has been seen in some models. 

Patterns for the other zlow operations will be added on demand, as needed.


src/Accelerators/NNPA/Dialect/ZLow/ZLowCombine.cpp
```diff
@@ -28,13 +28,66 @@ namespace {
 namespace onnx_mlir {
 namespace zlow {
 
+static LogicalResult removeUnusedOp(
+    PatternRewriter &rewriter, Operation *op, ArrayRef<int64_t> resultIndices) {
+  SmallVector<Value> results;
+  for (int64_t i : resultIndices)
+    results.emplace_back(op->getOperands()[i]);
+  bool allHasOneUse =
```
AlexandreEichenberger, line 40: Confused, why would a stick with only one use be removed?

tungld: For example:
```mlir
func.func @remove_unused_stick_op(%arg0: memref<5x10xf32>) -> memref<10xf32> {
  %0 = memref.alloc() {alignment = 4096 : i64} : memref<5x10xf16, #map>
  "zlow.stick"(%arg0, %0) {layout = "2D"} : (memref<5x10xf32>, memref<5x10xf16, #map>) -> ()
  %1 = memref.alloc() {alignment = 4096 : i64} : memref<10xf32>
  return %1: memref<10xf32>
}
```

The result `%0` of `zlow.stick` has only one use (that is `zlow.stick`), so `zlow_stick` can be removed.

AlexandreEichenberger: Ahh, are you looking at the `%0` in the `stick` as a use (it is really a write of its content). Seen as a use, you say there are no other "real" uses. 

Maybe a small comment in the code would clarify this when we read it again in 6 months

tungld: Yes, added a comment.

PR #3234, opened 2025-07-22
[RunONNXModel.py] Add an option to store profiling info to a file

tungld:
When we want to profile a model, we compile the model with option `--profile-ir` and then use the environment variable `ONNX_MLIR_INSTRUMENT_FILE` to store runtime profiling info to a file. However, it's quite difficult to remember the env name.

This patch adds an option `--write-runtime-log` into `RunONNXModel.py` that internally sets a value for `ONNX_MLIR_INSTRUMENT_FILE`.

To make the options complete, this patch also changes the old option `--log-to-file` to `--write-compile-log` that is used to output compilation info to a file.

### DIFF
--- a/utils/RunONNXModel.py
+++ b/utils/RunONNXModel.py
@@ -95,14 +95,6 @@ def check_non_negative(argname, value):
 
 parser = argparse.ArgumentParser()
 parser.add_argument(
-    "--log-to-file",
-    action="store",
-    nargs="?",
-    const="compilation.log",
-    default=None,
-    help="Output compilation messages to file, default compilation.log.",
-)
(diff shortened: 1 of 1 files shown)


chentong319:
I thought the PR will add an option to onnx-mlir. In general, I do not like env variables which are implicitly used.

AlexandreEichenberger:
> I thought the PR will add an option to onnx-mlir. In general, I do not like env variables which are implicitly used.

hardwiring output file name in the compiler would mean that each time we run an experiment, it will overwrite the previous file. That is why (unfortunately) it should probably remain a env var.

PR #3237, opened 2025-07-23
Generate parallel code for onnx.Slice

tungld:
This patch generates parallel code for onnx.Slice.

### DIFF
--- a/src/Conversion/ONNXToKrnl/Tensor/Slice.cpp
+++ b/src/Conversion/ONNXToKrnl/Tensor/Slice.cpp
@@ -65,7 +65,7 @@ struct ONNXSliceOpLowering : public OpConversionPattern<ONNXSliceOp> {
             op, true, parId, lbs[parId], ubs[parId], "slice");
       } else {
         onnxToKrnlParallelReport(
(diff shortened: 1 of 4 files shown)


src/Conversion/ONNXToKrnl/Tensor/Slice.cpp
```diff
@@ -48,7 +54,22 @@ struct ONNXSliceOpLowering : public OpConversionPattern<ONNXSliceOp> {
 
     ValueRange loopDef = create.krnl.defineLoops(outputRank);
     SmallVector<IndexExpr, 4> lbs(outputRank, LitIE(0));
-    create.krnl.iterateIE(loopDef, loopDef, lbs, shapeHelper.getOutputDims(),
+    DimsExpr ubs = shapeHelper.getOutputDims();
+
+    // Enable parallelism if required.
```
AlexandreEichenberger, line 59: This pattern is so frequent, wish we had made a function for it... maybe a later refactoring.

tungld: Yes, I think so.

Issue #3209, opened 2025-06-30
Errors in the nested create.krnl.Iterate for the new version of llvm


chentong319:
The new version of llvm added verification on affine.apply to check that the symbol parameter can not be a loop index. For example (for illustration only):
`
#map = affine_map<(d0)[s0] -> (d0 + s0)>
affine.for %arg0 =           {
    affine.for %arg1 =           {
        %1 = affine.apply #map(%arg1)[%arg0] 
    }
}
`
The result of %1 will be arg0+arg1. It is correct when lowered to scf, though it is not correct for affine specification. The new version of llvm will detect such error.
Such error may occur when the inner loop nest is created with another create.Krnl.Iterate nested inside the outer loop nest. In each create.krnl.iterate, its own loop induction variables are in parameters. But the loop induction variables are not explicitly specified. The code-get has to be careful to use DimIE or SymIE for the Values outside the current loop nest.

Sunny-Anand:
@AlexandreEichenberger, the behavior above is introduced with the new [llvm update](https://github.com/onnx/onnx-mlir/pull/3206) due to this  [pr](https://github.com/llvm/llvm-project/pull/128289). With the llvm update and with this change, when I used an in the `src/Dialect/Mlir/IndexExpr.cpp` for SymbolIndexExpr, there were around 200 such assert failures caught via the backend tests. 

```

SymbolIndexExpr::SymbolIndexExpr(Value const value) : IndexExpr() {
 assert(!mlir::affine::isValidDim(value) &&   "Creating a symbol from a dimentional index is not allowed");
  indexExprObj = new IndexExprImpl();
  assert(indexExprObj && "failed to allocate IndexExpr implementation");
  indexExprObj->initAsKind(value, IndexExprKind::Symbol);
}

SymbolIndexExpr::SymbolIndexExpr(IndexExprImpl *otherObjPtr) : IndexExpr() {
  assert(!mlir::affine::isValidDim(otherObjPtr->getValue()) && "Creating a symbol from a dimentional index is not allowed");
  // Create new IndexExpr implementation object.
  indexExprObj = new IndexExprImpl();
  assert(indexExprObj && "failed to allocate IndexExpr implementation");
  // If undefined, nothing to do.
  if (!otherObjPtr)
    return;

...
```

So this may involve even more analysis on 2 sides.

1. What, if any, performance hit we may have for the affine pass? Also does it impact SIMD performance?
2. What all ops need to be updated?

[...]

PR #3240, opened 2025-07-28
Fix dimensional operand cannot be used as a symbol in LayoutTransform

tungld:
In the new LLVM, dimensional operand cannot be used as a symbol. We had to change many symbols to dimensions when updating LLVM, but still missing one in LayoutTransform.


AlexandreEichenberger (approved):

tungld:

Sunny-Anand:
@AlexandreEichenberger, this pr is one example of the issue referenced here https://github.com/onnx/onnx-mlir/issues/3209

PR #3241, opened 2025-07-28
Remove unused function argument attributes to avoid llvm warnings

tungld:
When compiling a model with the verbose mode (-v), there are many warnings as the following:
```
Unhandled parameter attribute 'onnx.dim_params'
Unhandled parameter attribute 'onnx.name'
Unhandled parameter attribute 'onnx.dim_params'
Unhandled parameter attribute 'onnx.name'
```
These attributes are from `func.func` operations and they are unhandled when translating LLVMIR to LLVM bitcode.

This patch removes these attributes before lowering to LLVMIR so that they do not appear in LLVMIR, which avoids the warnings.


AlexandreEichenberger (approved):
LGTM, thanks

PR #3242, opened 2025-07-29
Fuse MatMul and Div if the 2nd operand of Div is a scalar constant

tungld:
This patch adds one optimization rule to Div's canonicalization that fuses MatMul and Div if the 2nd operand of Div is a scalar constant

This pattern is found in IBM granite models.

(This patch is similar to #3214 which is for MatMul and Mul)


AlexandreEichenberger (approved):
LGTM, sweet

tungld:

PR #3243, opened 2025-07-29
Use onnxmlir package inside RunONNXModel.py

chentong319:
RunONNXModel.py is a command line script to compile and run an ONNX model. This PR add the usage of onnxmlir package so that user can use this script outside of compiler container: the onnxmlir will invoke the compiler container for user. 
Changes in this PR:
1. Add an option, `--use-onnxmlir`, to enable the use of onnxmlir package. Without providing this option, user can run RunONNXModel.py in the same way as before.
2. Inside the script, I moved onnx related import closer into its usage. As a result, user does not need to install onnx package if he does not verify with onnxruntime, or read/write tensor to .pb file. We could use .pyn file instead.
3. Add an option, `--onnxmlir-args`, to pass arguments to onnxmlir.InferenceSession initialization. For example, container name and path to compiler if they are not the default value.
4. (Optional) Add an option, '--config-file`, to specify the default value for the arguments of RunONNXModel.py, if some of them are different from the default value in the script. Anyway, the command line argument has the highest priority.
5. Some changes triggered in onnxmlir package.


chentong319:
Some issues with the base of this PR. Created a new one.

PR #3245, opened 2025-07-30
Edit an error message when opening the .constants.bin file

tungld:
Edit an error message when opening the .constants.bin file by adding an instruction of how to fix.

### DIFF
--- a/src/Runtime/OMExternalConstant.inc
+++ b/src/Runtime/OMExternalConstant.inc
@@ -88,16 +88,19 @@ bool omMMapBinaryFile(
   }
   int fd = open(filePath, O_RDONLY);
   if (fd < 0) {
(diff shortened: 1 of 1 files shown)


src/Runtime/OMExternalConstant.inc
```diff
@@ -88,16 +88,19 @@ bool omMMapBinaryFile(
   }
   int fd = open(filePath, O_RDONLY);
   if (fd < 0) {
-    fprintf(stderr, "Error while opening %s: %s\n", filePath, strerror(errno));
+    fprintf(stderr,
+        "Error while opening %s: %s. Please set OM_CONSTANT_PATH to the fold "
```
AlexandreEichenberger: nit "folder"

tungld: Fixed. Thanks!

PR #3247, opened 2025-07-30
Recompose a Layer/RMSNorm even if the scale multiplication has multiple usesp…

jorickert:
This is fine as it is the last required operation in the Layer/RMSNorm and its recomposition pattern. The optional bias gets added in a separate pattern


AlexandreEichenberger (approved):
LGTM

jorickert:

PR #3248, opened 2025-07-30
Use onnxmlir package in RunONNXModel.py

chentong319:
RunONNXModel.py is a command line script to compile and run an ONNX model. This PR add the usage of onnxmlir package so that user can use this script outside of compiler container: the onnxmlir will invoke the compiler container for user.
Changes in this PR:

- Try to import onnxmlir package to use the compiler through container. Otherwise, the native compiler is used in the same way as before. The status is stored in args.use_onnxmlir.

- Inside the script, I moved onnx related import closer in its usage. As a result, user does not need to install onnx package if he does not verify with onnxruntime, or read/write tensor to .pb file. We could use python file .npy file instead.

- Add an option, --onnxmlir-args, to pass arguments to onnxmlir.InferenceSession initialization. For example, container name and path to compiler if they are not same as the default value.

- Add an option, '--config-file`, to specify the default value for the arguments of RunONNXModel.py, if some of them are different from the default value defined in the script. At last, the command line argument has the highest priority.  I like the configure file because it can record how the model is compiled and executed, and be reused. This part is optional to this PR and depends on reviewer's  opinion.

- Some changes triggered in onnxmlir package. For example, parse the compiler option for `-o`. If `-o` is not specified, temporary file is used for compilation.

### DIFF
--- a/utils/RunONNXModel.py
+++ b/utils/RunONNXModel.py
@@ -112,7 +112,8 @@ parser.add_argument(
     metavar="PATH",
     type=str,
     help="File to the configuration, a .yaml file. This file allows user to"
-    "define their own default value for arguments",
+    "define their own default value for arguments."
+    "An example can be found at utils/example-cfg.yaml",
 )
 parser.add_argument(
     "-c",
--- a/utils/example-cfg.yaml
+++ b/utils/example-cfg.yaml
@@ -1,0 +1,4 @@
+# Flags for compilation
+compile_args : -O3 --march=z16 --maccel=NNPA
+# If a native compile is use (not a container image)
+onnxmlir_args : compiler_image_name=None compiler_path=/path/to/compiler


utils/RunONNXModel.py
```diff
@@ -429,6 +453,8 @@ def read_input_from_refs(num_inputs, load_ref_filename, is_load_ref):
     inputs = []
 
     if is_load_ref:
+        from onnx import numpy_helper
```
AlexandreEichenberger, line 457: Would there be a way to store it as numpy array and thus not rely on the onnx package at all? I am not familiar with numpy serialization, but maybe there is something already built in?

utils/RunONNXModel.py
```diff
@@ -736,8 +780,17 @@ def __init__(self, model_file=None, **kwargs):
...
+        # Prepare onnx model for verifying output with onnxruntime
+        if (
+            not args.use_onnxmlir
+            and args.model
+            and args.model.endswith(".onnx")
+            and args.verify
+            and args.verify == "onnxruntime"
+            and args.verify_all_ops
+        ):
+            import onnx
```
AlexandreEichenberger, line 793: Should we have a `try` for the onnx import? Just to give a friendlier error message

utils/RunONNXModel.py
```diff
@@ -160,27 +174,37 @@ def check_non_negative(argname, value):
 parser.add_argument(
     "--atol", type=str, default="0.01", help="Absolute tolerance for verification."
 )
+parser.add_argument(
+    "-onnxmlir-args",
+    type=str,
+    default="",
+    help="Options passed to onnxmlir. Used with --use-onnxmlir",
```
AlexandreEichenberger, line 182: I did not find the option `--use-onnxmlir`, did I miss it?

Also, did you not want to change the name?

utils/RunONNXModel.py
```diff
@@ -1138,22 +1231,55 @@ def run_performance_test(self, output_name=None, input_feed=None, **kwargs):
             return outs
 
 
-################################################################################
-# Standalone driver
+def load_user_config(config_file):
```
AlexandreEichenberger, line 1235: Would you maybe include in a comment a possible config_file, so that if we want to generate one, we have an example to build from?

AlexandreEichenberger (commented):
Great new feature, added a few comments

utils/RunONNXModel.py
chentong319: An example is included.

utils/RunONNXModel.py
chentong319: logically, use-onnxmlir is determined by the availability of package, not the argument from user. I just do not want to add a separate global variable.

utils/RunONNXModel.py
chentong319: np.save/np.load can be used in future to replace .pb file.

PR #3249, opened 2025-07-30
Enable debug listing of operations that vanished

AlexandreEichenberger:
When creating `-profile-ir=xxx`, we insert instrumentation but then cleanup instrumentation around operations that vanished, so as to not clobber runtime logs with unneeded log entries.

Nearly all of the removed operations should be around Constant operations, as these operations exists in higher level dialects but then vanished when lowering.

But one bug I was tracing was spuriously removing other (needed) high level operations. 

So I added a `debug-only= print-empty-instrumentation` option that explicitly list all (non constant based operations) for which there is no code surrounding the instrumentation `before` and `after` calls.

New code also remove log generated calls that produce signatures when instrumentation is removed.


AlexandreEichenberger:

AlexandreEichenberger:

PR #3250, opened 2025-07-31
Stick/unstick fp32 data alloc at 4k pages

AlexandreEichenberger:
Cedric mentioned that 4k page alignment for the fp32 data is beneficial for performance.

This PR add 4k alignment for data used by `zlow.stick` when not implemented on CPU. For unstick, it was already the case in all but one case (which is added by this PR).

Option is default on for z17, off for z16 (as stick is always done in software).


src/Accelerators/NNPA/Transform/ZLow/ZLowStickExpansion.cpp
```diff
@@ -54,14 +58,49 @@ namespace zlow {
...
+  }
+
+  // Shared functions with other classes in this file.
+  static LogicalResult resizeAllocOfUnstickOutput(ZLowUnstickOp unstickOp) {
+    memref::AllocOp allocOfXOp =
+        unstickOp.getOut().getDefiningOp<memref::AllocOp>();
+    assert(allocOfXOp && "unstick output should always be allocated");
+    auto alignmentAttr = allocOfXOp.getAlignment();
+    int64_t intAlign = alignmentAttr ? alignmentAttr.value() : 1;
+    if (intAlign >= gAlignment) {
```
tungld: Should it be `(intAlign != 0 && intAlign % gAlignment == 0)`?

tungld (approved):
LGTM!

AlexandreEichenberger:

src/Accelerators/NNPA/Transform/ZLow/ZLowStickExpansion.cpp
AlexandreEichenberger: Ok, that would protect us if for some reason we were to align buffers to multiple of 4k. But then what happens if we want an align to 7k... should we bump it to 8k?

That is what I implemented, next 4k page given current alignment of alloc.

tungld: Thanks!

PR #3252, opened 2025-08-04
Added z17 CPU vs NNPA performance model

AlexandreEichenberger:
The heuristics that decide whether to place an op on the CPU or NNPA uses a performance model with the `-nnpa-placement-heuristic' with non default policy, like `FasterOps` for example.

Performance model was tracking only z16, I added a model that tracks z17. Added a layer that let us (easily) add new models for future NNPA implementations.

### DIFF
--- a/src/Accelerators/NNPA/Conversion/ONNXToZHigh/PerfModelArch14.inc
+++ b/src/Accelerators/NNPA/Conversion/ONNXToZHigh/PerfModelArch14.inc
@@ -9,13 +9,13 @@
 // =============================================================================
 
 // Operation Add_3ds: estimated times.
-double march14_estimatedTimeForCPU_Add_3ds(double e3, double e2, double e1) {
+double arch14_estimatedTimeForCPU_Add_3ds(double e3, double e2, double e1) {
   double complexity = e3 * e2 * e1;
   // Regression for CPU with r2 = 0.9989711028792525
   return 3.9686846353007493e-07 + 1.1794164898251022e-10 * complexity;
 }
 // Operation Add_3ds: estimated times.
-double march14_estimatedTimeForNNPA_Add_3ds(double e3, double e2, double e1) {
+double arch14_estimatedTimeForNNPA_Add_3ds(double e3, double e2, double e1) {
   double complexity = e3 * ms_ceiling(e2, 2.0) * ms_ceiling(e1, 64.0);
   double complexity2 = e3 * ms_ceiling(e2, 32.0) * ms_ceiling(e1, 64.0);
   // Regression for NNPA with r2 = 0.9994266956239162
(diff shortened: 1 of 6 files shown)


src/Accelerators/NNPA/Conversion/ONNXToZHigh/PerfModelArch15.inc
```diff
@@ -0,0 +1,349 @@
...
+}
+// Operation Unstick_3ds: estimated times.
+double arch15_estimatedTimeForNNPA_Unstick_3ds(
+    double e3, double e2, double e1) {
+  double complexity = e3 * ms_ceiling(e2, 2.0) * ms_ceiling(e1, 64.0);
+  double complexity2 = e3 * ms_ceiling(e2, 32.0) * ms_ceiling(e1, 64.0);
+  // Regression for NNPA with r2 = 0.4294645871109005
+  return 1.0784429022719318e-05 + 1.569351314487559e-10 * complexity +
+         1.5693513144875584e-10 * complexity2;
+}
```
chentong319, line 386: If the weight constants are stored in a file, will it be easier to change the value for different arch, op and system?

AlexandreEichenberger: Right now I have an auto generated script that output c code, so the effort is zero; but it does increase the number of functions (which I should make static, minimally).

At this time, it would require more work to save it in a file and load them, and then it also need to be moved around if the compiler is relocated. Maybe an array of constant might be a good compromise, it would still imply some work because each time a new arch comes, one would have to add 1 element to the arrays.

tungld (approved):
LGTM!

src/Accelerators/NNPA/Conversion/ONNXToZHigh/PerfModelArch14.inc
```diff
@@ -0,0 +1,337 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//===----------------- Auto-Generated, do not change  ---------------------===//
```
tungld, line 5: Do you put the script or command to generate this file somewhere? so that other developers can update the file for new ops.

It's also helpful if you put a comment here to tell us how to generate this file.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/PerfModelArch14.inc
```diff
@@ -0,0 +1,337 @@
...
+//
+// Copyright 2023-2025 The IBM Research Authors.
+//
+// =============================================================================
+
+// Operation Add_3ds: estimated times.
+static static double arch14_estimatedTimeForCPU_Add_3ds(
```
tungld: nit: double static

AlexandreEichenberger: fixed. I checked online, and while `static double` or `double static` are semantically identical, the preferred convention is the former rather than the latter.

Thanks for finding out that I repeated the keyword twice.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/PerfModelArch14.inc
AlexandreEichenberger: I have published the data in my own IBM GitHub repo. There are about 20 files that are needed. Would you want me to include them here (where? They are NNPA specific), or give a pointer to my repo? Only accessible within IBM.

https://github.ibm.com/alexe/support/tree/main/perf-unit-operations

So what additional comments/file/doc would you like me to include @tungld

tungld: I see, they are .sh and .mlir files, so how about put them inside a subfolder in the `utils' folder? Look like they can be used for performance regression check if we want in the future.

For additional comments, I hope that we can say something about how this file is generated or updated.
For example, in this auto-generated file: [onnx-mlir](https://github.com/onnx/onnx-mlir/tree/main)/[docs](https://github.com/onnx/onnx-mlir/tree/main/docs)
/SupportedONNXOps-cpu.md, we have:
```
<!--- Automatically generated, do not edit. -->
<!--- To update, run `make onnx_mlir_supported_ops_cpu' -->
```

src/Accelerators/NNPA/Conversion/ONNXToZHigh/PerfModelArch14.inc
tungld: Thanks! Sorry for the confusion, I meant duplicated keyword `static`, not the order of `static` and `double`.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/PerfModelArch14.inc
AlexandreEichenberger: Sounds good, will do. Its not like its my proudest script writing, but it works.
It always will need some tuning, e.g. adding new ops... but it is a good base to start from.

PR #3256, opened 2025-08-07
[zdnnx] Parallelize quantized matmul for multiple NNPAs

tungld:
This patch adds a new function `zdnnx_quantized_matmul_op` in the zdnn extension (zdnnx) to parallelize the quantized matmul on multiple NNPAs.

Currently, zdnnx only parallelizes the quantized matmul if its inputs can be fully reused for tiles.

### DIFF
--- a/src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.cpp
+++ b/src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXToZHigh.cpp
@@ -404,6 +404,14 @@ public:
     // Weight is a constant.
     if (!isDenseONNXConstant(B))
       return rewriter.notifyMatchFailure(op, "MatMul's B is not a constant.");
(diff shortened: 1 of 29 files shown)


tungld:
> With a quick reading, the normal matmul has not changed but some of the code was refactored in functions that can now be reused for quantized and unquantized mammals.

Yes, thanks for elaborating it!

> For the quantized matmul, how are the scaling/shifting done? Are they be included in the parallel loop? If they are hidden within zDNN, then they will be parallelized (as multiple threads are doing zDNN calls). If they are before/after the zdnnx_quantized_matmul_op call, then they would not be in the current code. Is that correct?

rec_scale and offset values are fields in the ztensor data structure, they are scalar. They are copied from the full ztensor to each tile during the splitting preparation. So for the computation loop, nothing needs to do with them.
The parallel loop for quantized matmul is basically the same as the one for non-quantized matmul, the only different thing is the zdnn call, it is `zdnn_quantized_matmul_op` with more inputs instead of `zdnn_matmul_op`.

PR #3259, opened 2025-08-19
Fix the bug in new RunONNXModel.py

chentong319:
The error is caused the missing `import onnx`. I tested RunONNXModelZoo.py.


tungld:

PR #3262, opened 2025-08-21
cmake error for zdnn build[NNPA]

chentong319:
When initialization cmake with NNPA on, I ran into error "ninja: error: build.ninja:9480: bad $-escape (literal $ must be written as $$", which is for `-j$(nproc)` for zdnn.cmake. 
```
COMMAND = cd /myspace/onnx-mlir/build-debug/src/Accelerators/NNPA/zDNN/src/zdnn && sh -c "export MAKEFLAGS=--no-print-directory &&                          make -q -C zdnn lib/libzdnn.so && true ||                          (MAKEFLAGS=--no-print-directory                           make -j$(nproc) -C zdnn lib/libzdnn.so &&                           ar -rc /myspace/onnx-mlir/build-debug/src/Accelerators/NNPA/zDNN/src/zdnn/zdnn/lib/libzdnn.a /myspace/onnx-mlir/build-debug/src/Accelerators/NNPA/zDNN/src/zdnn/zdnn/obj/*.o)" && /usr/bin/cmake -E touch /myspace/onnx-mlir/build-debug/src/Accelerators/NNPA/zDNN/src/zdnn-stamp/zdnn-build
```
I guess that there could be other solution. But this PR simply removed the `-j`. The only impact is that it would be slower for zdnn build.

### DIFF
--- a/src/Accelerators/NNPA/zdnn.cmake
+++ b/src/Accelerators/NNPA/zdnn.cmake
@@ -38,7 +38,7 @@ function(setup_zdnn version)
       BUILD_COMMAND sh -c "export MAKEFLAGS=--no-print-directory && \
                          make -q -C zdnn lib/libzdnn.so && true || \
                          (MAKEFLAGS=--no-print-directory \
-                          make -j$(nproc) -C zdnn lib/libzdnn.so && \
+                          make -C zdnn lib/libzdnn.so && \
                           ar -rc ${ZDNN_LIBDIR}/libzdnn.a ${ZDNN_OBJDIR}/*.o)"
 
       INSTALL_COMMAND ""


src/Accelerators/NNPA/zdnn.cmake
```diff
@@ -38,7 +38,7 @@ function(setup_zdnn version)
       BUILD_COMMAND sh -c "export MAKEFLAGS=--no-print-directory && \
                          make -q -C zdnn lib/libzdnn.so && true || \
                          (MAKEFLAGS=--no-print-directory \
-                          make -j$(nproc) -C zdnn lib/libzdnn.so && \
+                          make -C zdnn lib/libzdnn.so && \
```
tungld, line 41: I think we all have used this when building onnx-mlir for NNPA, and our jenkins uses this also and haven't seen this error. I wonder under which circumstance you got this, perhaps, the local environment is different.

Can you confirm if your machine has `nproc` program? `$(nproc)` is to invoke `nproc`.

chentong319: My machine has nproc program. It returns 106 inside or outside of the container. This issue is not new to me, and is not a persistent issue. Previously, when this issue happened, I just do not use ninja.  I just don't understand why.

gongsu832: Because the cmake files are meant to generate make rules not ninja rules which interpret $ sign differently. Slow down everyone just to make it work for ninja doesn't seem very productive.

For this particular case, if you want it to work for both make and ninja, you can do
```
diff --git a/src/Accelerators/NNPA/zdnn.cmake b/src/Accelerators/NNPA/zdnn.cmake
index ff7a280c..4d45f8f0 100644
--- a/src/Accelerators/NNPA/zdnn.cmake
+++ b/src/Accelerators/NNPA/zdnn.cmake
@@ -12,6 +12,14 @@ function(setup_zdnn version)
   set(ZDNN_OBJDIR     ${ZDNN_TOPDIR}/zdnn/obj)
   set(ZDNN_LIBDIR     ${ZDNN_TOPDIR}/zdnn/lib)

+  # ninja cannot handle $(nproc) in BUILD_COMMAND due to the $ sign
+  # so set a cmake variable ZDNN_NPROC
+  execute_process(
+    COMMAND nproc
+    OUTPUT_VARIABLE ZDNN_NPROC
+    OUTPUT_STRIP_TRAILING_WHITESPACE
+  )
+
   # Only build libzdnn on s390x
   if (${CMAKE_SYSTEM_PROCESSOR} STREQUAL "s390x")
     ExternalProject_Add(zdnn
@@ -38,7 +46,7 @@ function(setup_zdnn version)
       BUILD_COMMAND sh -c "export MAKEFLAGS=--no-print-directory && \
                          make -q -C zdnn lib/libzdnn.so && true || \
                          (MAKEFLAGS=--no-print-directory \
-                          make -j$(nproc) -C zdnn lib/libzdnn.so && \
+                          make -j${ZDNN_NPROC} -C zdnn lib/libzdnn.so && \
                           ar -rc ${ZDNN_LIBDIR}/libzdnn.a ${ZDNN_OBJDIR}/*.o)"

       INSTALL_COMMAND ""
```

chentong319: Okay. I thought that ninja is used in onnx-mlir by default. That's in the document.

PR #3265, opened 2025-08-25
Add a CSE pass before --EmitONNXIR

chentong319:
The purpose of this PR is to get a clearer output of --EmitONNXIR with unused ops removed. This is useful when a model is manually altered.
To me, the existing invocation of CSE pass in conversion from onnx to krnl could be removed. But I did not touch it in this PR.

### DIFF
--- a/src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
+++ b/src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
@@ -168,7 +168,7 @@ void addONNXToZHighPasses(mlir::PassManager &pm) {
     pm.addPass(onnx_mlir::zhigh::createZHighConstPropagationPass());
 
   // Remove common sub-expressions.
-  pm.addPass(mlir::createSymbolDCEPass());
+  pm.addPass(mlir::createCSEPass());
 
   // Clean dead code.
   pm.addPass(mlir::createSymbolDCEPass());
--- a/src/Compiler/CompilerPasses.cpp
+++ b/src/Compiler/CompilerPasses.cpp
@@ -143,6 +143,7 @@ void addONNXToMLIRPasses(mlir::PassManager &pm, bool targetCPU,
   pm.addPass(onnx_mlir::createStandardFuncReturnPass());
 
   // Clean dead code.
+  pm.addPass(mlir::createCSEPass());
   pm.addPass(mlir::createSymbolDCEPass());
 
   // Replace every DisposableElementsAttr with DenseElementsAttr.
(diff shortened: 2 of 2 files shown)


src/Compiler/CompilerPasses.cpp
```diff
@@ -153,6 +153,13 @@ void addONNXToMLIRPasses(mlir::PassManager &pm, bool targetCPU,
   // function and just before instrumentation.
   pm.addPass(createSetONNXNodeNamePass());
 
+  // Add a CSE pass to delete unreachable ops.
+  // This is useful when a model, especially manually modified one, is dumped
+  // with --EmitONNXIR.
+  // This pass is inserted before the instrumentation pass, which may add
+  // instrumentation ops with external sideeffect.
+  pm.addPass(mlir::createCSEPass());
```
tungld: I wonder if `symbol-dce` is more appropriate?: https://mlir.llvm.org/docs/Passes/#-symbol-dce

chentong319: Yes, the symbolDCE also removes the symbols, though there is no symbol declaration at this stage in onnx-mlir. 
I replaced the CSE with symbolDCE and removed the debugging parameter, which is always true now.

src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
```diff
@@ -168,7 +168,7 @@ void addONNXToZHighPasses(mlir::PassManager &pm) {
     pm.addPass(onnx_mlir::zhigh::createZHighConstPropagationPass());
 
   // Remove common sub-expressions.
-  pm.addPass(mlir::createCSEPass());
```
tungld: Was it removed by accident?

src/Compiler/CompilerPasses.cpp
```diff
@@ -178,12 +185,9 @@ void addONNXToMLIRPasses(mlir::PassManager &pm, bool targetCPU,
         onnx_mlir::createInstrumentPass(instrumentOps, instrumentActions));
 }
 
-void addONNXToKrnlPasses(mlir::PassManager &pm, int optLevel, bool enableCSE,
-    std::string ONNXOpsStatFormat) {
-  if (enableCSE)
-    // Eliminate common sub-expressions before lowering to Krnl.
-    // TODO: enable this by default when we make sure it works flawlessly.
-    pm.addPass(mlir::createCSEPass());
```
tungld, line 186: Looks like this was removed by accident since `enableCSE=true` means we call `pm.addPass(mlir::createCSEPass());`.

tungld (commented):
It seems you accidentally removed CSE? CSE is needed to merge parallel same patterns into one.

src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
chentong319: Yes, my mistake.

chentong319:
I found that this PR is not really needed. The size of remaining IR is caused by the embedding weights, simplification can not help.

PR #3266, opened 2025-08-27
[NNPA] JSON configuration file for device_placement and quantization

tungld:
This patch replaces two options `--nnpa-load-device-placement-file` and `--nnpa-save-device-placement-file` by two new options (which can be use for both device placement and quantization settings):
- `--nnpa-load-config-file` to load configuration settings for NNPA from a JSON file.
- `--nnpa-save-config-file` to save configuration settings for NNPA to a JSON file.

It is often the case that ONNX node name is used to match ONNX operators. If one does not know ONNX node names in a model, let use --nnpa-save-config-file to get a configuration file to start with.

By using a JSON file, users can currenlty control two following features:
- device placement: to decide which ONNX operators run on CPU or NNPA.
- quantization: to decide which ONNX operators are quantized to utilize i8 computation on NNPA.


Sunny-Anand:
@tungld will it be possible to enhance the examples for the 2 new settings for the above to make it more explicit by adding a model for example like bert for the above 2 settings in the docs?

tungld:
> @tungld will it be possible to enhance the examples for the 2 new settings for the above to make it more explicit by adding a model for example like bert for the above 2 settings in the docs?

@Sunny-Anand it is possible and it takes more time to play with bert. So let me add an example for bert later in another PR once we settle the approach in this PR.

docs/JsonConfigFile-NNPA.md
```diff
@@ -0,0 +1,233 @@
...
+- An ONNX operator is matched if both `node_type` AND `onnx_node_name` are matched. Once an ONNX operator is matched, its attribute `device` is updated using the JSON value.
+- The list is evaluated in sequence, with earlier items having precedence. If an ONNX operator matches an object, it does not matches against the remaining objects in the list. 
+
+## quantization[] object fields
+
+| Field          | Type              | Description                                                    |
+| -------------  | -------------     | -------------------------------------------------------------- |
+| quantize       | boolean           | Whether to apply quantization (true or false).                 |
+| node_type      | string            | ONNX operator type (e.g., "onnx.Relu", "onnx.*").              |
+| onnx_node_name | string (optional) | Specific ONNX node name to apply quantization to.              |
```
AlexandreEichenberger:
```
Specify ONNX node name (via a regexp) to apply quantization to.
```

src/Accelerators/NNPA/Conversion/ONNXToZHigh/JsonConfigFile.hpp
```diff
@@ -0,0 +1,52 @@
+#include <functional>
+#include <string>
+
+#include "llvm/Support/JSON.h"
```
AlexandreEichenberger, line 4: @chentong319 in our 'onnx-mlir-light' we were trying to reduce dependences on LLVM: is that an issue here? We had Json before for the device placement, so maybe not? Please make sure that this solution is fine with the OM-light.

AlexandreEichenberger (approved):
LGTM, thanks for merging the two placement and quantification in a unique file.

Asked @chentong319 about json and onnx-mlir-light, to see if the dependence on LLVM json is an issue or not. Please wait until he provide feedback.

docs/JsonConfigFile-NNPA.md
tungld: I updated this and another one. Thanks!

docs/JsonConfigFile-NNPA.md
```diff
@@ -0,0 +1,237 @@
+<!--- SPDX-License-Identifier: Apache-2.0 -->
+
+The compiler provides two options to allow users control how ONNX operators run on NNPA.
+- `--nnpa-load-config-file` to load configuration settings for NNPA from a JSON file.
+- `--nnpa-save-config-file` to save configuration settings for NNPA to a JSON file.
+
+It is common to use ONNX node names to match ONNX operators. If you're unsure about the node names in your model, you can use the `--nnpa-save-config-file` option to generate a configuration file as a starting point.
+
+Alternatively, you can open the ONNX model using a visualizer like Netron to inspect the node names. Note that the actual node names used by the compiler may differ slightly from those shown in Netron due to compiler optimizations, though they are usually the same.
```
tungld, line 9: @AlexandreEichenberger I added some sentences to mention Netron app when looking for node names.

tungld:

src/Accelerators/NNPA/Conversion/ONNXToZHigh/JsonConfigFile.hpp
chentong319: The onnx-mlir-light only includes code in src/Runtime. This change should not affect it.

chentong319 (approved):
LGTM!

PR #3267, opened 2025-08-28
Fusing elementwise computations with Stick/Unstick

AlexandreEichenberger:
The goal of this PR is to fuse stick and unstick into element wise operations. The PR supports element wise operation with known broadcast at compile time (on the innermost dimension). Generated code uses SIMD operations and is heavily unrolled for performance.

The general approach is that a new pass in the ONNX to ZHigh transform 
``` C++
  if (!nnpaDisableFusionOpStickUnstick)
    pm.addPass(onnx_mlir::zhigh::createFusionOpStickUnstick());
```
which search for `unstick -> elementwise-compute` or `elementwise-compute -> stick` and when all conditions are met, it then removes the stick or unstick operation and feed the ZTensor (with ZTensor data layout) directly into the inputs or outputs of the element wise operation.

Then, during the ONNX/ZHigh conversion to Krnl/ZLow, there is a new set of patterns (with higher priority) that will generate the suitable code.

The code pattern is basically to tile the innermost loop by 64 (length of a stick). Regular tensors are accessed by load-and-splat (if broadcast) or vector load (if non broadcast). ZTensor memory is accessed through a view (as vector does not do well with mapped memory) and is converted to float32. For outputs, non-tensor are simply written with regular simd stores; and ZTensors again use a view, and float32 are converted to dlfloat16, using saturation when requested.

For loops where the innermost dim is not a multiple of 64, the last suit-64 is handled first 8 values at a time (the amount of dlfloat 16 handled by a simd conversion). If the output is a ZTensor, then we know that the stick memory is tiled by 64 values, so we simply over-ride the values at the end with bad output (can be switched off with a # define). For regular tensors, the last 8 values are computed (with some bad values) and are stored in a separately allocated buffer; then a scalar loop pick the right values and save them in the output so as to never over-write any values that we should not over-write.

Because this code must reside under the `src/Accelerator/NNPA` subdirectories, I had to rip off a bit of the src/Conversion/ONNXToKrnl/Math/Elementwise` infrastructure. A new Elementwise.hpp file includes what is now needed to generate element wise operations under the regular or the NNPA subdirectories.

I have checked the results with small end-to-end examples, not included here. I have added suitable lit tests to verify the functionality.

Option can be disabled using this flag `nnpa-disable-fusion-op-stick-unstick`

Example:
```mlir
func.func @main_graph(%arg0: tensor<4x256x256xf32>, %arg1: tensor<4x256x1xf32>) -> tensor<4x256x256xf32> {
  %0 = "onnx.Add"(%arg0, %arg0) : (tensor<4x256x256xf32>, tensor<4x256x256xf32>) -> tensor<4x256x256xf32>
  %1 = "onnx.Add"(%arg1, %arg1) : (tensor<4x256x1xf32>, tensor<4x256x1xf32>) -> tensor<4x256x1xf32>

  // Broadcast pattern of 2nd input prevent assignment to NNPA, so must execute on CPU.
  %2 = "onnx.Add"(%0, %1) : (tensor<4x256x256xf32>, tensor<4x256x1xf32>) -> tensor<4x256x256xf32> 

  %3 = "onnx.Add"(%2, %arg0) : (tensor<4x256x256xf32>, tensor<4x256x256xf32>) -> tensor<4x256x256xf32>
  return %3 : tensor<4x256x256xf32>
}
```

after the removal of stick unstick resulting in ZTensor in the 3rd add (executing on CPU because of the unsupported broadcast on NNPA).

```mlir
  func.func @main_graph(%arg0: tensor<4x256x256xf32>, %arg1: tensor<4x256x1xf32>) -> tensor<4x256x256xf32> {
    %0 = "zhigh.Stick"(%arg0) {layout = "3DS"} : (tensor<4x256x256xf32>) -> tensor<4x256x256xf16, #zhigh.layout<{dataLayout = "3DS"}>>
    %1 = "zhigh.Add"(%0, %0) : (tensor<4x256x256xf16, #zhigh.layout<{dataLayout = "3DS"}>>, tensor<4x256x256xf16, #zhigh.layout<{dataLayout = "3DS"}>>) -> tensor<4x256x256xf16, #zhigh.layout<{dataLayout = "3DS"}>>
    %2 = "zhigh.Stick"(%arg1) {layout = "3DS"} : (tensor<4x256x1xf32>) -> tensor<4x256x1xf16, #zhigh.layout<{dataLayout = "3DS"}>>
    %3 = "zhigh.Add"(%2, %2) : (tensor<4x256x1xf16, #zhigh.layout<{dataLayout = "3DS"}>>, tensor<4x256x1xf16, #zhigh.layout<{dataLayout = "3DS"}>>) -> tensor<4x256x1xf16, #zhigh.layout<{dataLayout = "3DS"}>>

    // CPU Add operation with ZTensor inputs and output.
    %4 = "onnx.Add"(%1, %3) {onnx_node_name = "onnx.Add_2"} : (tensor<4x256x256xf16, #zhigh.layout<{dataLayout = "3DS"}>>, tensor<4x256x1xf16, #zhigh.layout<{dataLayout = "3DS"}>>) -> tensor<4x256x256xf16, 

#zhigh.layout<{dataLayout = "3DS"}>>
    %5 = "zhigh.Add"(%4, %0) : (tensor<4x256x256xf16, #zhigh.layout<{dataLayout = "3DS"}>>, tensor<4x256x256xf16, #zhigh.layout<{dataLayout = "3DS"}>>) -> tensor<4x256x256xf16, #zhigh.layout<{dataLayout = "3DS"}>>
    %6 = "zhigh.Unstick"(%5) : (tensor<4x256x256xf16, #zhigh.layout<{dataLayout = "3DS"}>>) -> tensor<4x256x256xf32>
    return %6 : tensor<4x256x256xf32>
  }
```

Once this settles, I will probably use this functionality to also generate code for the compiler-generated stick/unstick, as it essentially become a "copy" with input/output of ZTensor/retular tensor.


src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -2913,77 +2561,22 @@ struct ONNXWhereOpLowering : public ConversionPattern {
...
-      ONNXElementwiseUnaryOpLowering<mlir::ONNXSqrtOp>,
-      ONNXElementwiseVariadicOpLowering<mlir::ONNXSubOp>,
-      ONNXElementwiseVariadicOpLowering<mlir::ONNXSumOp>,
-      ONNXElementwiseUnaryOpLowering<mlir::ONNXTanOp>,
-      ONNXElementwiseUnaryOpLowering<mlir::ONNXTanhOp>,
-      ONNXElementwiseUnaryOpLowering<mlir::ONNXThresholdedReluOp>,
-      ONNXWhereOpLowering, ONNXElementwiseVariadicOpLowering<mlir::ONNXXorOp>>(
+// Insert pattern for all the binary, unary, and variadic operations supported
+// here. List comes from Elementwise.hpp.
+#define ELEMENTWISE_BINARY(_OP_TYPE)                                           \
```
AlexandreEichenberger, line 2566: Note: I needed in many places the list of all element wise types. So I generated macros in Elementwise.hpp to store them and then instantiate the proper pattern, using macros as shown below.

test/mlir/accelerators/nnpa/conversion/instrument/add-onnx-zhigh-level.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: onnx-mlir --march=z16 --maccel=NNPA --printIR --EmitZHighIR --profile-ir=ZHigh %s  | FileCheck %s
+// RUN: onnx-mlir --march=z16 --maccel=NNPA --printIR --EmitZHighIR --profile-ir=ZHigh  -nnpa-disable-fusion-op-stick-unstick %s  | FileCheck %s
```
AlexandreEichenberger, line 1: disabled the new opt here as irrelevant to what is tested.

src/Accelerators/NNPA/Transform/ZHigh/FusionOpStickUnstick.cpp
```diff
@@ -0,0 +1,393 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//===-------- FusionOpStickUnstick.cpp - Fuse compute op  -----------------===//
+//
+// Copyright 2023 The IBM Research Authors.
+//
+// =============================================================================
+//
+// This pass detects patterns of unstick -> op -> stick (or subset) and when
```
tungld, line 11: It looks like this can be done by the current code in onnx-mlir, which is in the layout propagation pass in zHigh. It is disabled now, but you can enable it and see check if does the same thing or not: https://github.com/onnx/onnx-mlir/blob/main/src/Accelerators/NNPA/Transform/ZHigh/ZHighLayoutPropagation.cpp#L345.

In ZHighLayoutPropagation, `unstick -> op -> stick` is rewritten by:
- propagating `unstick` down, so we have `op -> unstick -> stick`, then
- `unstick -> stick` becomes identity and is removed.

AlexandreEichenberger: Looking at the code that is in that file, I can see that it replaces

`x -> unstick -> unary`  to `x -> unary -> unstick` and leaves later pass to remove `unstick -> stick`

and for binary, the same pattern where both inputs are from unstick, namely

` binary(unstick(x), unstick(y))` to `binary(x, y) -> unstick`

My pattern is the same for unary; and for binary, I take arbitrary combinations of stick or normal for inputs and outputs (no restrictions on the layouts that we support, arbitrary combinations of 2-4D, 3DS). However, since my target is really to generate a SIMD version of the op, I need to make sure that I know if there is broadcast or not at compile time.

Also, one difference is that I only perform this operation if the unstick has only one use, because I don't want to migrate the cost of unstickify into the unary/binary op if I need to unstickify anyway for another op, or if we need to have the unstickify cost in multiple element-wise operations. [There are cases where the unstickify is used by `Dim` op... and I should tolerate this.] This could be revised at a later stage.

AlexandreEichenberger: But migrating my new code in that files make total sense.

AlexandreEichenberger:
Suggested changes have been implemented in #3271

AlexandreEichenberger:
See explanation in #3271 on why I decided to revert to this approach

AlexandreEichenberger:

AlexandreEichenberger:
Note from #3271 


> After 3 days of battling with the integration of the unstick -> op -> stick with the ZHighLayoutPropagation, I ran into a series of issues, mainly that the fusion introduces mixes of types in operations, e.g. f16 and f32. This causes issues with shape inferences that often determine the type of the output based on its inputs. I tried to disable the setting of the output types, but at the same time we also have cases where the type is set by attributes and that needs to be set (other examples are from ZHigh where the output type & layout is set from the inputs). The next steps would have been to have modes where types are set and modes where type are not set in ShapeInference, and that seemed not quite right. It also impacts Verify infra, as mixed types are not supposed to exist in most operations.


So I went back to have separate passes (this PR), where this pass is set just before lowering from ONNX/ZHigh to Krnl/ZLow.

src/Accelerators/NNPA/Transform/ZHigh/FusionOpStickUnstick.cpp
AlexandreEichenberger: Tried in a new PR but lead to problems, explained both in that other PR and below when re-opening this PR. I now removed the "dead code" as it would result in similar issues if re-activated.

src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
```diff
@@ -176,6 +181,14 @@ void addONNXToZHighPasses(mlir::PassManager &pm) {
   // Replace every DisposableElementsAttr with DenseElementsAttr.
   pm.addPass(onnx_mlir::zhigh::createZHighScrubDisposablePass());
 
+  // NOTE: Cannot have a pass of shape analysis after this pass as we add
+  // "non-standard" mixtures of types into operations, for example with "add(f16
+  // in stickified format, f23) -> f32". Shape inferences & verification
+  // requires all of the types to be identical, so we need to execute this pass
+  // JUST before lowering to KRNL/ZLow.
```
tungld, line 188: It makes sense!

tungld (approved):
LGTM! Impressive effort on fusing the SIMD codes — great work!

Thanks again for all the helpful comments; they really made the review process much smoother.

PR #3268, opened 2025-08-29
Special case for shape inference for onnx.Range operator

tungld:
There is a case in IBM granite models (or perhaps decoding models) where the output shape of Range operation can be computed at compile, that is `limit = start + constant_value`, in this case the output dim is `constant_value / delta` despite `start` being a constant or not.

This patch adds such a special case for Range's shape inference.


AlexandreEichenberger (approved):
I see, this is a case of a 
```
start = X
end = X + c
tripCount = X + c - X = c
```
makes sense, thanks. Other ops like this by any chances?

tungld:
> Other ops like this by any chances?

`onnx.Slice` also has `starts` and `ends`, however, they are 1D tensors not scalar, and `starts` is often 0s, so it looks like the pattern in this patch is not useful for `onnx.Slice`.

PR #3269, opened 2025-08-29
Improving the convtranspose decomposing for stride 2x2 

chaitanyakamarapu:
In the stride 2x2 usecase, if the weights can be divisible to four weights, stacking the weights and using the single conv in place of four convs.


chaitanyakamarapu:
incorrect merge target

PR #3271, opened 2025-09-03
Fusing Elementwise computations with Stick / Unstick v2

AlexandreEichenberger:
See description in #3267. 

Detection of `unstick -> op -> stick` has now been integrated into ZHighLayoutPropagation.cpp.

Functionality has been extended to support fusion of multiple supported ZHigh layout formats into that an elementwise operation.


AlexandreEichenberger:
After 3 days of battling with the integration of the `unstick -> op -> stick` with the ZHighLayoutPropagation, I ran into a series of issues, mainly that the fusion introduces mixes of types in operations, e.g. f16 and f32. This causes issues with shape inferences that often determine the type of the output based on its inputs. I tried to disable the setting of the output types, but at the same time we also have cases where the type is set by attributes and that needs to be set (other examples are from ZHigh where the output type & layout is set from the inputs). The next steps would have been to have modes where types are set and modes where type are not set in ShapeInference, and that seemed not quite right. It also impacts Verify infra, as mixed types are not supposed to exist in most operations.

So I went back to have separate passes, where this pass is set just before lowering from  ONNX/ZHigh to Krnl/ZLow.

So its back to #3267.

PR #3275, opened 2025-09-08
Upgrade to protobuf 6.31.1

gongsu832:
Upgrade to protobuf 6.31.1, which is the 1st version that uses rules_rust 0.56.0 with s390x support we managed to upstream. Some caveats apply:

  - bazel related toolchain support on ppc64le is even worse than that on s390x so this PR will not work on ppc64le. As a result, we will have to retire our ppc64le Jenkins CI after this PR (we can resurrect it once the bazel toolchain on ppc64le catches up). Jenkins CI is not required for merging

  - onnx cmake build file needs to be patched to add some missing dependencies on abseil, which should eventually be upstreamed


gongsu832:
OK MacOS CI build is fixed. The requirements.txt still had the old version of protobuf specified, which overrode the new version. The PR is ready to be merged.

chentong319:
Do we need to update the document(such as BuildOnLinuxOSX.md) accordingly?

gongsu832:
> Do we need to update the document(such as BuildOnLinuxOSX.md) accordingly?

There is really nothing changed in terms of building onnx-mlir itself. Most of the changes are for building protobuf with C/C++ backend on s390x. We can certainly describe what's need to be done in BuildOnLinuxOSX.md but unless you are a developer you probably don't care about that. This is also the reason we typically recommend people to use the prebuilt docker image instead of trying to do all these dirty hack themselves.

Sunny-Anand:
@gongsu832 I am running some tests based off this pr especially on the abseil dependency part. I will update the results here with my experiment. Sorry for the delay caught up with a few urgent issues. Will try to post the results and findings by next week.

utils/install-protobuf.sh
```diff
@@ -1,10 +1,47 @@
...
+cd ${ABSL_DIR}/build
+cmake -DCMAKE_INSTALL_LIBDIR=lib \
+      -DABSL_PROPAGATE_CXX_STD=ON \
+      -DBUILD_SHARED_LIBS=ON ..
+make -j${NPROC} install
+cd ../..
+rm -rf ${ABSL_DIR} ${HOME}/.cache
+
+# Install protobuf
+PROTOBUF_VERSION=6.31.1
```
Sunny-Anand, line 25: Do we want to keep the same version as 6.32 or do we want to update requirements.txt with 6.31.1

utils/install-protobuf.sh
```diff
@@ -1,10 +1,47 @@
...
+cargo install cargo-bazel --version ${CARGO_BAZEL_VERSION}
 
-cd protobuf
-./autogen.sh
-./configure --enable-static=no
-make -j$(sysctl -n hw.logicalcpu) install
-cd python
-python3 setup.py install --cpp_implementation
+# Install absl for new version of protobuf
+ABSL_VERSION=20240722.1
```
Sunny-Anand, line 11: Same as above for ABSEIL version.

utils/install-protobuf.sh
```diff
@@ -1,10 +1,47 @@
...
+# Install protobuf
+PROTOBUF_VERSION=6.31.1
+PROTOBUF_URL=https://github.com/protocolbuffers/protobuf.git
+PROTOBUF_DIR=protobuf
+git clone -b v${PROTOBUF_VERSION} --recursive ${PROTOBUF_URL} ${PROTOBUF_DIR}
+mkdir -p ${PROTOBUF_DIR}/build
+cd ${PROTOBUF_DIR}/build
+# Must specify -Dprotobuf_BUILD_TESTS=OFF otherwise find_package(absl)
+# in onnx will fail due to missing protobuf::gmock target
+cmake -DCMAKE_INSTALL_LIBDIR=lib \
```
Sunny-Anand, line 33: Same as above for using cmake and bazel together.

utils/install-onnx.sh
```diff
@@ -0,0 +1,9 @@
+git fetch --prune --unshallow --tags
```
Sunny-Anand, line 1: Same comment as above. It seems this patching is needed for all CI platforms, is this correct? Then should we get someone from onnx take a look at this?

docker/Dockerfile.onnx-mlir
```diff
@@ -21,7 +21,15 @@ COPY . onnx-mlir
 # CMAKE_INSTALL_LIBDIR to be lib.
 RUN ONNX_ROOT=${WORK_DIR}/onnx-mlir/third_party/onnx \
     && cd ${ONNX_ROOT} \
-    && CMAKE_ARGS="-DCMAKE_INSTALL_LIBDIR=lib" python3 -m pip install .
+    # Require patching until upstreamed
```
Sunny-Anand, line 24: Why is this patching needed for onnx here for abseil? Doesn't onnx as third-party install its own protoc using cmake and abseil? Is this something we need to report to onnx as a breakage?

docker/Dockerfile.llvm-project
```diff
@@ -61,16 +61,77 @@ RUN distro=$(cat /etc/os-release|grep -Po '(?<=^ID=").*(?=")|(?<=^ID=)[^"].*[^"]
...
+ARG RUST_VERSION=1.88
+ARG CARGO_BAZEL_VERSION=0.16.0
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | \
+    sh -s -- -y --default-toolchain none \
+    && . "${HOME}/.cargo/env" \
+    && rustup install ${RUST_VERSION} \
+    && cargo install cargo-bazel --version ${CARGO_BAZEL_VERSION}
+
+# Install absl for new version of protobuf
+ARG ABSL_VERSION=20240722.1
```
Sunny-Anand, line 95: Should this be updated to the latest version of ABSEIL for protobuf 32.0 https://registry.bazel.build/modules/protobuf/32.0

docker/Dockerfile.llvm-project
```diff
@@ -61,16 +61,77 @@ RUN distro=$(cat /etc/os-release|grep -Po '(?<=^ID=").*(?=")|(?<=^ID=)[^"].*[^"]
...
+             -DBUILD_SHARED_LIBS=ON .. \
+    && make -j${NPROC} install && ldconfig \
+    && cd ../.. && rm -rf ${ABSL_DIR} ${HOME}/.cache
 
 # Install protobuf
-ARG PROTOBUF_VERSION=21.12
-RUN git clone -b v${PROTOBUF_VERSION} --recursive https://github.com/protocolbuffers/protobuf.git \
-    && cd protobuf && ./autogen.sh \
-    && ./configure --enable-static=no \
+ARG PROTOBUF_VERSION=6.31.1
```
Sunny-Anand, line 109: Do we want to keep the same version as 6.32 or do we want to update requirements.txt with 6.31.1

docker/Dockerfile.llvm-project
```diff
@@ -61,16 +61,77 @@ RUN distro=$(cat /etc/os-release|grep -Po '(?<=^ID=").*(?=")|(?<=^ID=)[^"].*[^"]
...
+ARG PROTOBUF_URL=https://github.com/protocolbuffers/protobuf.git
+ARG PROTOBUF_DIR=protobuf
+RUN git clone -b v${PROTOBUF_VERSION} --recursive ${PROTOBUF_URL} ${PROTOBUF_DIR} \
+    && mkdir -p ${PROTOBUF_DIR}/build \
+    && cd ${PROTOBUF_DIR}/build \
+    # Must specify -Dprotobuf_BUILD_TESTS=OFF otherwise find_package(absl)
+    # in onnx will fail due to missing protobuf::gmock target
+    && CC=clang CXX=clang++ \
+       cmake -DCMAKE_INSTALL_LIBDIR=lib \
```
Sunny-Anand, line 118: I did try to change the build from cmake to bazel for building and installing protobuf and building the protoc; however as usual, it runs into issues with missing upstreams for rust_rules for s390x. The only comment I have here is, if we had found a way to use Bazel for this step, it would avoid needing us to build abseil as it is a dependency of protobuf and if bazel is used for building protobuf then it would also be able to fetch and build abseil own its own. https://registry.bazel.build/modules/protobuf/32.0

Sunny-Anand:
@AlexandreEichenberger @gongsu832 if you guys are ok with merging this pr, I can merge this now. Also do we want to update https://github.com/onnx/onnx-mlir/blob/main/README.md to drop ppc64le-Linux and also remove/stop the CI from running?

AlexandreEichenberger:
FYI, on my Mac I had to install the package below

``` bash
brew install abseil
```

PR #3276, opened 2025-09-09
Added missing vector model for elementwise operations

AlexandreEichenberger:
Many elementwise operations did not have vector models.. and thus were not simonized.

Added the vector operation model for these ops, and updated the impacted mlir tests

### DIFF
--- a/src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
+++ b/src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
@@ -129,13 +129,8 @@ static void CheckIfCustomScalarOpIsSupported(Type elementType) {
 }
 
 // =============================================================================
-// Scalar ops handling
+// Scalar ops handling (IN ALPHABETICAL ORDER)
 // =============================================================================
-
-template <>
-GenOpMix getGenOpMix<ONNXAddOp>(Type t, Operation *op) {
-  return {{GenericOps::ArithmeticGop, 1}};
-}
 
 template <>
 GenOpMix getGenOpMix<ONNXAbsOp>(Type t, Operation *op) {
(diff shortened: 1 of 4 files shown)


AlexandreEichenberger:
Added the new operations testing systematically against the list of all element wise operations. Most missing ones were for the ones that are handled automatically by an equivalent MLIR operation; I moved them by alphabetical order to simplify a bit.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
```diff
@@ -129,71 +129,171 @@ static void CheckIfCustomScalarOpIsSupported(Type elementType) {
...
+}
+
+template <>
+GenOpMix getGenOpMix<ONNXNotOp>(Type t, Operation *op) {
+  return {{GenericOps::ArithmeticGop, 1}};
+}
+
+template <>
+GenOpMix getGenOpMix<ONNXOrOp>(Type t, Operation *op) {
+  return {{GenericOps::ArithmeticGop, 1}};
```
tungld: `ONNXAndOp` uses `GenericOps::LogicalGop`, while this `ONNXOrOp` uses `GenericOps::ArithmeticGop`, what is the difference?

tungld (approved):
LGTM!

src/Dialect/Mlir/VectorMachineSupport.cpp
```diff
@@ -197,7 +197,11 @@ int64_t ZArch14VectorMachineSupport::computeArchVectorLength(
       return UNSUPPORTED;
     }
   }
-  // Support for integer (we consider bit-wide ops as byte wide ops).
+  // Support for integer.
```
AlexandreEichenberger, line 200: FYI, when we generated SIMD code for memref with `i1` type (boolean), the code was not good both on the arm and z machines. Thus disabled it for the moment.

If we wanted good performance, we should convert the bits (`i1`) into bytes (meaning grouping 8 bits together) and then operate on them as bytes... if would be much more efficient. Was hoping that the backend would do that, it does not appear to do so. So if we have Or/And/Not/XOR that are key to model performance, we should do that.

src/Conversion/ONNXToKrnl/Math/Elementwise.cpp
AlexandreEichenberger: Saw that too. I moved all of the "and/or/not/xor" to logical. They operate on `tensor(bool)` only in the ONNX standard.

So now its more consistant, but realistically they could all be arithmetic ops and that would be just fine too.

@tungld let me know if you want me to eliminate the `LogicalGop`.

PR #3287, opened 2025-09-25
Apply 'EXCLUDE_FROM_ALL' to the directory when ONNX_MLIR_BUILD_TESTS is not ON

### LINKED ISSUE #3285: ONNX_MLIR_BUILD_TESTS doesn't appear to be honored
The "ONNX_MLIR_BUILD_TESTS" setting is documented as controlling whether the test executables are built. But it doesn't appear to be honored. A couple of problems:

1. It's only tested in `docs/doc_example/CMakeLists.txt` and `test/multiple-models/CMakeLists.txt`
2. When it's ON, the logic sets `set(EXCLUDE_FROM_ALL ON)`. The CMake documentation for `EXCLUDE_FROM_ALL` says that it's either a property on a target ([documentation](https://cmake.org/cmake/help/latest/prop_tgt/EXCLUDE_FROM_ALL.html)), or a property on a directory ([documentation](https://cmake.org/cmake/help/latest/prop_dir/EXCLUDE_FROM_ALL.html)) - and `set(EXCLUDE_FROM_ALL ON)` sets neither.

It looks like a fix would be to:

1. Test it in `docs/doc_example/CMakeLists.txt` and `test/CMakeLists.txt`
2. Set the property on the directory with (for example):

```cmake
# If 'ONNX_MLIR_BUILD_TESTS' is ON, set 'EXCLUDE_FROM_ALL' for this directory. This will allow test targets to be
# defined, but 'all' will not depend on them by default.
if (NOT ONNX_MLIR_BUILD_TESTS)
  set_directory_properties(PROPERTIES EXCLUDE_FROM_ALL ON)
endif()
```

This problem manifests for me with a Visual Studio Generator, and the fix fixes it. I'll double check with a Ninja Generator. I'm happy to send the fix; wondering if others see this problem...?

mschofie:
Fixes #3285.

`ONNX_MLIR_BUILD_TESTS` is documented as controlling whether the test executables are built. But it doesn't appear to be honored - even when I set it to `OFF`, the tests are still built. The problem appears to be with the use of `EXCLUDE_FROM_ALL` - when attempting to opt-out of tests, the CMake variable `EXCLUDE_FROM_ALL` is set to `ON` - however CMake only documents `EXCLUDE_FROM_ALL` properties applied to Directories ([documentation](https://cmake.org/cmake/help/latest/prop_dir/EXCLUDE_FROM_ALL.html)) or Targets ([documentation](https://cmake.org/cmake/help/latest/prop_tgt/EXCLUDE_FROM_ALL.html)). This change:
1. Updates the logic that tests `ONNX_MLIR_BUILD_TESTS` to set the `EXCLUDE_FROM_ALL` property on the current Directory.
2. Moves the check from `:/test/multiple-models/CMakeLists.txt` up to `:/test/CMakeLists.txt` to apply to all of `:/test`

I've validated with "Ninja Multi-Config" and "Visual Studio 17 2022" generators on Windows. Without the fix, running a CMake build with no target specified, and `ONNX_MLIR_BUILD_TESTS` set to `OFF` will build the test targets. With the fix, the test targets aren't built by default, but will build when a target is specifically requested.

### DIFF
--- a/src/Dialect/ONNX/ONNXDimAnalysis.cpp
+++ b/src/Dialect/ONNX/ONNXDimAnalysis.cpp
@@ -941,82 +941,79 @@ void DimAnalysis::visitDim(
     Value data = reshapeOp.getData();
     Value output = reshapeOp.getReshaped();
 
-    // Special case 1: the output dimension i can be from
-    // - data[j] if
-    //    - dim j and i are the only dynamic dimension in data and output,
-    //    respectively, and
+    // Special case: when data and output have the same number of dynamic
+    // dimensions (N), the dynamic dimension output[i] is the same as the
+    // dynamic dimension data[j] if
     //    - the products of the remaining static dimensions in data and output
-    //    are equal.
+    //    are equal, and
+    //    - the remaining N-1 dynamic dimensions in data and output are the
+    //    same, respectively.
+    //
+    // Note that output[i] is the dimension this visitDim function is working
+    // on.
     //
     // It's interesting that if shape[i] is arg0[ii] and data[j] is arg1[jj],
     // we can say that arg0[ii] == arg1[jj], that can be used to verify user
     // inputs.
 
-    // Get the dynamic dimension from data.
+    // Get the dynamic dimensions from data and output.
     auto dataType = mlir::cast<RankedTensorType>(data.getType());
     auto outputType = mlir::cast<RankedTensorType>(output.getType());
-    // Check if there is only one dynamic dimension in the data and output.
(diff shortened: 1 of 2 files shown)


AlexandreEichenberger:
I see that the DCO was not satisfactorily signed. Can you confirm @mschofie that you obeyed by the DCO rules as listed in our repo. I can then override them and merge the PR. Tx

mschofie:
Apologies!

Yes. I confirm that this change conforms to the DCO rules in this repository.

PR #3288, opened 2025-09-26
add Python decoding algorithms

adrian-selk:
This PR will add new Python scripts that implement a custom way of generating text, especially decoding and selecting the next token, for inferencing Large Language Models such as GPT2 or Granite3.1, without using HF transformers. The contribution contains 3 implemented algorithms (greedy, top-k, top-p; also temperature scaling is supported), only using numpy for the calculations.
Also, a usage example `gpt2-decode.py` is provided as an updated version of `run-gpt2-from-huggingface.py`, incorporating the new classes for text generation. 
Finally, some new dependencies have been added to `requirements.txt` that are needed to run the scripts.

### DIFF
--- a/src/Compiler/CMakeLists.txt
+++ b/src/Compiler/CMakeLists.txt
@@ -210,7 +210,15 @@ add_dependencies(OMCompilerUtils cruntime)
 if (ONNX_MLIR_ENABLE_JNI)
   add_dependencies(OMCompilerUtils jniruntime)
 endif()
-if (TARGET omp)
+
+# Find OpenMP headers built by LLVM
+find_path(OPENMP_INCLUDE_DIR
+  NAMES omp.h
+  HINTS
+    ${LLVM_BUILD_BINARY_DIR}/lib/clang/*/include
+  NO_DEFAULT_PATH
+)
+if(OPENMP_INCLUDE_DIR)
   add_dependencies(OMCompilerUtils ompruntime)
 endif()
 
--- a/src/Dialect/ONNX/ONNXOps/Canonicalize.cpp
+++ b/src/Dialect/ONNX/ONNXOps/Canonicalize.cpp
@@ -1630,6 +1630,41 @@ struct RecomposeConcatPattern : public OpRewritePattern<ONNXConcatOp> {
   }
 };
 
+struct RemoveDimZeroInputInConcatPattern
+    : public OpRewritePattern<ONNXConcatOp> {
+  using OpRewritePattern<ONNXConcatOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(
+      ONNXConcatOp concatOp, PatternRewriter &rewriter) const final {
+    ValueRange inputs = concatOp.getOperands();
+    int64_t axis = concatOp.getAxis();
+
+    // Collect indices of inputs whose dim size at axis is zero.
+    SmallVector<int64_t> indices;
+    for (unsigned int i = 0; i < inputs.size(); ++i) {
+      Value inp = inputs[i];
+      if (!hasShapeAndRank(inp))
+        continue;
+      ArrayRef<int64_t> shape = getShape(inp.getType());
+      // Scalar with rank 0. Dim size is one (not zero).
+      if (shape.size() == 0)
+        continue;
+      if (shape[axis] == 0)
+        indices.emplace_back(i);
+    }
+    if (indices.empty())
+      return rewriter.notifyMatchFailure(
+          concatOp, "No operand whose dim at axis is zero");
+
+    // Rewrite: remove operands whose dim at axis is zero.
+    rewriter.modifyOpInPlace(concatOp, [&]() {
+      for (int64_t idx : indices)
+        concatOp.getOperation()->eraseOperand(idx);
+    });
+    return success();
+  }
+};
+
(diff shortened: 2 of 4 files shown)


tungld:

requirements.txt
```diff
@@ -1,8 +1,49 @@
-lit~=15.0
-# numpy 1.24 deprecates np.object, np.bool, np.float, np.complex, np.str,
-# and np.int which are used heavily in onnx-mlir.
+certifi==2025.8.3
```
tungld: Do we really need all of these packages? I see many unrelated packages such as cuda, triton, etc.
Please keep only really needed packages.

Also, it looks better to move this `requirements.txt` file into the folder `utils/python/transformers` instead of putting it in the root folder that does not make sense.

utils/python/transformers/README.md
```diff
@@ -0,0 +1,164 @@
+# Inferencing downloaded ONNX Models using Python
+
+This folder provides Python scripts for to downloading, compiling and inferencing autoregressive transformer models like GPT-2, Granite 3, or other models of the same architecture, using ONNX-MLIR.
```
tungld: nit: remove `to` in `for to downloading`.

utils/python/transformers/omdecoder.py
```diff
@@ -0,0 +1,357 @@
...
+            generation_config=generation_config,
+        )
+
+
+class OMModelForCausalLM(OMModelDecoder, OMGeneration):
+    """ONNX model with a causal language modeling head for ONNX-MLIR inference."""
+
+    main_input_name = "input_ids"
+    _supports_cache_class = False
+    _is_stateful = True
```
tungld: Where are these variables used?

utils/python/transformers/README.md
```diff
@@ -0,0 +1,164 @@
...
+export ONNX_MLIR_RUNTIME_DIR=../../build/Debug/lib
+```
+
+Alternatively, you can use a .env file. If you are using the default container setup, just copy the cell below, create a `.env`-file inside this folder, and paste the content.
+
+```bash
+ONNX_MLIR_ROOT=/workdir/onnx-mlir
+ONNX_MLIR_BIN=$ONNX_MLIR_ROOT/build/Debug/bin
+ONNX_MLIR_INCLUDE=$ONNX_MLIR_ROOT/include
+ONNX_MLIR_RUNTIME_DIR=$ONNX_MLIR_ROOT/build/Debug/lib
```
tungld: Could you check again if all these variables are needed or only `ONNX_MLIR_HOME ` is enough?

utils/python/transformers/omdecoder.py
```diff
@@ -0,0 +1,357 @@
...
+MULTI_QUERY_ATTN_MODELS = {"gpt_bigcode"}
+
+class OMConfig:
+    """Simple config class to replace transformers config dependencies."""
+    def __init__(self, config_dict: dict):
+        for key, value in config_dict.items():
+            setattr(self, key, value)
+            
+        # Add compatibility attributes for transformers
+        if not hasattr(self, '_attn_implementation'):
```
tungld: Where do we use this `_attn_implementation `?

adrian-selk: This is being added as an attribute to mitigate compability issues I have previously ran into using GPT2, since the Config JSON is expected to have certain fields (this one was missing). However, since we are now using custom classes to load the model from pretrained weights, it can be removed (it works without these lines). Thanks for the catch!

utils/python/transformers/omdecoder.py
adrian-selk: Same as with `_attn_implementation`, those were expected for the transformers pipeline. Not using transformers Config and Generation classes means they are not needed. Will be removed

utils/python/transformers/README.md
adrian-selk: It seems like only `ONNX_MLIR_HOME` is needed apart from the `PYTHONPATH` variable. Thanks for pointing out, I will remove this part from the documentation.

requirements.txt
adrian-selk: most of them have been removed, essentially the scripts only need numpy (calculations), requests (downloading from HF) and transformers (tokenization). Also the .txt has been moved

tungld:

requirements.txt
tungld: it seems you forgot removing this .txt?

utils/python/transformers/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+The `OMGeneration`-class can be used to to build own models, e.g. for ONNX-MLIR inference, by inheriting from the Mixin class:
+
+```python
+class OMModelForCausalLM(OMModelDecoder, OMGeneration):
+    """ONNX model with a causal language modeling head for ONNX-MLIR inference."""
+
+    main_input_name = "input_ids"
+    _supports_cache_class = False
+    _is_stateful = True
```
tungld: Removing these variables here as well?

adrian-selk: Yes, forgot to do that, thanks

tungld (approved):
LGTM! Congrats on the first contribution! Thanks!

tungld:

PR #3290, opened 2025-09-26
Modifying build to find openmp's runtime build files 

christopherlmunoz:
As part of the migration off `-DLLVM_ENABLE_PROJECTS` to `-DLLVM_ENABLE_RUNTIMES="openmp"` openmp build files moved, and openmp can no longer be targeted by `cmake --build . --target omp`. 

Files are now located in:
`${LLVM_BUILD_BINARY_DIR}/runtimes/runtimes-bins/openmp/runtime/src`


src/Compiler/CMakeLists.txt
```diff
@@ -210,7 +210,15 @@ add_dependencies(OMCompilerUtils cruntime)
 if (ONNX_MLIR_ENABLE_JNI)
   add_dependencies(OMCompilerUtils jniruntime)
 endif()
-if (TARGET omp)
+
+# Find OpenMP headers built by LLVM
+find_path(OPENMP_INCLUDE_DIR
+  NAMES omp.h
```
tungld, line 216: Just a confirmation: does this omp.h exist only when LLVM is built with `-DLLVM_ENABLE_RUNTIMES="openmp"`?

christopherlmunoz:
```
root@477474a1c9b3:/workdir/llvm-project/build# ls lib/clang/22/include/omp.h
ls: cannot access 'lib/clang/22/include/omp.h': No such file or directory
```
`-- OpenMP support           : OFF`

No omp.h is created, and onnx-mlir realizes that OpenMP support is OFF when -`DLLVM_ENABLE_RUNTIMES="openmp"` is not included.

PR #3292, opened 2025-09-30
Add ShapeInference for RandomUniformLike.

jorickert:
Add ResultTypeInference to Bernoulli, CastLike, EyeLike, RandomUniform, RandomUniformLike and SequenceEmpty. 
Add a result element type verifier to Bernoulli, Cast, CastLike, EyeLike, RandomNormal, RandomNormalLike, RandomUniform, RandomUniformLike.
 Refactor code between ops with `dtype` attr to have less duplication.


tungld:

src/Dialect/ONNX/ONNXOps/Math/RandomNormalLike.cpp
```diff
@@ -23,92 +24,11 @@ using namespace onnx_mlir;
...
-//===----------------------------------------------------------------------===//
-
-LogicalResult ONNXRandomNormalLikeOp::inferShapes(
-    std::function<void(Region &)> doShapeInference) {
-  if (!hasShapeAndRank(getInput()))
-    return success();
-  Type elementType = getRandomNormalLikeOutputElementType(*this);
-  return inferShapeForUnaryOps(getOperation(), elementType);
-}
+GET_SHAPE_AND_TYPE_INFERENCE_FOR_SHAPE_COPYING_OPS(ONNXRandomNormalLikeOp)
```
AlexandreEichenberger: @jorickert We had a discussion among us and to the extend possible, we would prefer avoiding macros for code. Would you mind reverting to having a normal definition of the 2 function calls, with inside a templated call. That way, we get the same benefits of reusing the same code (great contribution, thanks), and at the same time preserve the debug features of normal code.

Let me know if you feel strongly about this.

jorickert: Hey, I do not have a strong opinion about the usage of macros. I will update the PR to not use these "codegen macros"

src/Dialect/ONNX/ONNXOps/Math/ElementwiseUnary.cpp
```diff
@@ -150,6 +150,13 @@ LogicalResult ONNXBitwiseNotOp::inferShapes(
 // Cast
 //===----------------------------------------------------------------------===//
 
+LogicalResult ONNXCastOp::verify() {
+  return cast<ShapedType>(this->getResult().getType()).getElementType() ==
```
chentong319: You can use getElementTypeOrSelf(Type) defined by MLIR, which can handle ShapedType and scalar type. 
Also, getElementType(Type) defined In onnx-mlir/src/Support/TypeUtilities.hpp. In old days, this function is used to handle ShapedType and UnrankedTensor. Now it call getElementTypeOrSelf(Type) directly. Is UnrankedTensor not a problem anymore?

src/Dialect/ONNX/ONNXOps/Math/ElementwiseUnary.cpp
```diff
@@ -164,6 +171,24 @@ LogicalResult ONNXCastOp::inferShapes(
   return shapeHelper.computeShapeAndUpdateType(elementType);
 }
 
+//===----------------------------------------------------------------------===//
+// CastLike (not really unary, but similar)
+//===----------------------------------------------------------------------===//
+
+LogicalResult ONNXCastLikeOp::verify() {
+  return cast<ShapedType>(this->getResult().getType()).getElementType() ==
```
chentong319: Same here and other similar cases.

src/Dialect/ONNX/ONNXOps/OpHelper.hpp
```diff
@@ -256,6 +256,61 @@ bool isConstOf(mlir::Value constValue, double n);
 mlir::Type convertONNXTypeToMLIRType(
     mlir::Builder &builder, onnx::TensorProto_DataType onnxType);
 
+template <typename OP>
+mlir::Type getResultElementTypeFromDtypeWithFallBackToInputType(OP op) {
```
chentong319: I wonder whether template is needed here. To me, the only operation specific action is getDtype(). If we pass the getDtypeAttr() as a parameter and use the generic interface of Operation, can we avoid the template?

src/Dialect/ONNX/ONNXOps/Math/RandomNormalLike.cpp
jorickert: I replaced the macro with direct calls to the functions

src/Dialect/ONNX/ONNXOps/Math/ElementwiseUnary.cpp
jorickert: Thanks for telling me about getElementTypeOrSelf . As far as I know UnrankedTensor should not be a problem, as it also has an element type

src/Dialect/ONNX/ONNXOps/OpHelper.hpp
jorickert: I replaced the templates with functions with an explicit dtype parameter

src/Dialect/ONNX/ONNXOps/Math/ElementwiseUnary.cpp
```diff
@@ -150,6 +150,12 @@ LogicalResult ONNXBitwiseNotOp::inferShapes(
 // Cast
 //===----------------------------------------------------------------------===//
 
+LogicalResult ONNXCastOp::verify() {
+  return getElementTypeOrSelf(this->getResult().getType()) == getTo()
+             ? success()
+             : emitOpError("element type does not match the 'to' attribute");
+}
+
```
AlexandreEichenberger, line 158: thanks for adding the verifier

AlexandreEichenberger (approved):
LGTM, thanks for removing redundant code and relying on MLIR calls. Additional verifiers are also appreciated

PR #3295, opened 2025-10-01
Add sum and total option to report time in make-report.py

AlexandreEichenberger:
Added and implemented this option

```
    --reporting <name>:  quartile: average excluding outlier experiments (default).
                       all: average including every experiments.
                       sum: sum of every experiments.

```

useful for a case like Tong where he wants to sum all the times from 100 experiments.

### DIFF
--- a/utils/make-report.py
+++ b/utils/make-report.py
@@ -419,7 +419,7 @@ def parse_file_for_perf(file_name, stat_name, warmup_num=0):
     # Encountered now all the measurements, has no more.
     meas_num = start_count - warmup_num
     assert meas_num > 0, "expected at least one set of measurement after warmups"
-       
+
     discard_num = 0
     if reporting == "QUARTILE":
         discard_num = int(meas_num / 4)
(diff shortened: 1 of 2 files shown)


utils/make-report.py
```diff
@@ -432,7 +446,10 @@ def parse_file_for_perf(file_name, stat_name, warmup_num=0):
         if discard_num > 0 and meas_num - 2 * discard_num > 0:
             time_array = np.sort(time_array)
             time_array = time_array[discard_num:-discard_num]
```
chentong319, line 448: Will it be time_array[0:-0] for ALL or SUM?

AlexandreEichenberger: Note that `discard_num` is set to zero unless reporting is `QUARTILE`, and thus the test

```
        if discard_num > 0 and meas_num - 2 * discard_num > 0:
```

fails and thus the code that restricts the `time_array` values is not invoked.

chentong319 (approved):
LGTM!

PR #3296, opened 2025-10-02
Fix two issues to make omp work again with --parallel and zdnnx

tungld:
Following #3290, there is another place where you need a change so that zdnnx can find omp in llvm.

### DIFF
--- a/src/Runtime/omp/CMakeLists.txt
+++ b/src/Runtime/omp/CMakeLists.txt
@@ -1,14 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
-# Find OpenMP headers built by LLVM
-find_path(OPENMP_INCLUDE_DIR
-  NAMES omp.h
-  HINTS
-    ${LLVM_BUILD_BINARY_DIR}/lib/clang/*/include
-  NO_DEFAULT_PATH
-)
-
-if(OPENMP_INCLUDE_DIR)
+if(OMP_SUPPORTED)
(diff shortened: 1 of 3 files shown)


tungld:
@christopherlmunoz somehow I could not find you in the reviewer list to request your review ... could you take a look at this PR? Thanks!

src/Runtime/omp/CMakeLists.txt
```diff
@@ -35,21 +27,11 @@ if(OPENMP_INCLUDE_DIR)
...
-  )
-
-  if(NOT LIBOMP_PATH)
-    message(FATAL_ERROR "Could not find built OpenMP library (libomp.a or shared)")
-  endif()
-
   add_custom_target(ompruntime
     COMMAND ${CMAKE_COMMAND} -E copy_if_different
-            ${LIBOMP_PATH} ${ONNX_MLIR_LIBRARY_PATH}/libompruntime.a
+            ${OMP_TOPDIR}/openmp-build/runtime/src/libomp.a ${ONNX_MLIR_LIBRARY_PATH}/libompruntime.a
```
christopherlmunoz, line 32: Does libomp.a always get created, no matter the architecture?

tungld: Honestly, I only tested on linux environment. This code is similar to the old code for the old llvm, just different path. I hope it would work well on other archs.

chentong319 (approved):
LGTM!

PR #3298, opened 2025-10-02
Normalization with stick

AlexandreEichenberger:
Code provide fusion of stick & unstick with LayerNorm operations.

I revised the scheme to generate that code, using 2 classes that are now listed in `onnx-mlir/src/Accelerators/NNPA/Conversion/ZHighToZLow/ProcessStickDataHelper.hpp` and cpp file.

The compiler generated stick/unstick has not migrated to this new style. I am planning to migrate that code once this one is running successfully in a subsequent PR.

Also, I noticed that we added a RMS Normalization op (which we gather from a pattern of regular op). That op will be added natively in a subsequent ONNX release. The one we added manually detect and uses an optional `Bias`. That bias does not exist in the future official ONNX operation. Open question if we should "downgrade" our custom op to support no bias... and thus merge it with the future op, or should we keep this one and the future official op.

I added a bit of support (and applied it to matmul) to do a n-dim tiling and permute. These permute orders were somewhat tricky to get, so I figured I could simply add a bit of support. The builder function `blockAndPermute` simply gives a list of loop def and a list of blocking factor; it then apply the blocking to the innermost loops and permute the loop performing the blocked iterations to the innermost dims.


src/Accelerators/NNPA/Conversion/ZHighToZLow/ProcessStickData.cpp
```diff
@@ -715,8 +533,388 @@ struct ONNXElementwiseOpLoweringWithNNPALayout
...
+
+    // Outerloops (all but E1), with blocked E2 blocked by B. No E1 in lbs/ubs.
+    DimsExpr ubs(shapeHelper.getOutputDims(0));
+    ubs.pop_back();
+    DimsExpr lbs(XRank - 1, LitIE(0));
+    ValueRange loopDefs = create.krnl.defineLoops(XRank - 1);
+    SmallVector<Value, 4> outerOptLoops, innerOptLoops;
+    create.krnl.blockAndPermute(loopDefs, {B}, outerOptLoops, innerOptLoops);
+    // Handle Parallel
+    bool useParallel = false;
```
tungld, line 638: Parallel is off here, does it have any effect on multi-core performance?

src/Accelerators/NNPA/Conversion/ZHighToZLow/ProcessStickDataHelper.hpp
```diff
@@ -0,0 +1,191 @@
...
+//===----------------------------------------------------------------------===//
+
+#ifndef ONNX_MLIR_PROCESS_STICK_DATA_HELPER_H
+#define ONNX_MLIR_PROCESS_STICK_DATA_HELPER_H
+
+#include "src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.hpp"
+
+namespace onnx_mlir {
+
+// Class that supports memref that are either to normal reference or stick
```
tungld, line 22: This class looks useful. Thanks for adding it!

tungld (approved):
LGTM!

src/Accelerators/NNPA/Conversion/ZHighToZLow/ProcessStickData.cpp
AlexandreEichenberger: Right, let me add it and test.

src/Accelerators/NNPA/Conversion/ZHighToZLow/ProcessStickDataHelper.hpp
AlexandreEichenberger: Tested.

PR #3300, opened 2025-10-06
Add a compile optio --ignore-attention-mask

tungld:
This pull request adds a compile option `--ignore-attention-mask=arg-index` where `arg-index` is an integer number indicating the index of the attention_mask input in the model (the index is often 1 in bert-like models).

When this option is enabled, the masking computation in attention_layer are ignored.
For example, when using `--ignore-attention-mask=arg-index`, the self-attention `softmax(Q * K^T * scale + mask) * V` becomes `softmax(Q * K^T * scale) * V` where `+ mask` is removed.

### DIFF
--- a/src/Dialect/ONNX/Transforms/IgnoreAttentionMask.cpp
+++ b/src/Dialect/ONNX/Transforms/IgnoreAttentionMask.cpp
@@ -73,6 +73,26 @@ struct RemoveAttentionMaskPattern : public OpRewritePattern<ONNXAddOp> {
       return failure();
 
     // Check that mask comes from attention_mask.
+    bool found = false;
+    SmallVector<Value> worklist;
+    DenseSet<Value> visited;
+    worklist.emplace_back(attentionMaskArg);
+    while (!worklist.empty() && !found) {
+      Value current = worklist.back();
+      worklist.pop_back();
+      if (current == mask) {
+        found = true;
+        break;
+      }
+      for (auto *user : current.getUsers())
+        for (auto next : user->getResults()) {
+          if (visited.insert(next).second)
+            worklist.push_back(next);
+        }
+    }
+    if (!found)
+      return rewriter.notifyMatchFailure(
+          addOp, "the mask is not connected with attention_mask arg");
 
     // Rewrite: bypass the AddOp.
     rewriter.modifyOpInPlace(
@@ -100,8 +120,8 @@ struct IgnoreAttentionMaskPass
     return "Do not use attention_mask in a self_attention layer";
   }
 
-  // Usage: onnx-mlir-opt --do-not-use-attention-mask='arg-idx=1'
-  Option<bool> argIdx{*this, "arg-idx",
+  // Usage: onnx-mlir-opt --ignore-attention-mask='arg-idx=1'
+  Option<int64_t> argIdx{*this, "arg-idx",
       llvm::cl::desc("Argument index of attention_mask in the function"),
       ::llvm::cl::init(1)};
 
--- a/src/Dialect/ONNX/Transforms/IgnoreAttentionMask.cpp
+++ b/src/Dialect/ONNX/Transforms/IgnoreAttentionMask.cpp
@@ -2,13 +2,14 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-//===---- DonotUseAttentionMask.cpp - Remove masking in Attention layer
-//----===//
+//===---- IgnoreAttentionMask.cpp - Remove masking in Attention layer -----===//
 //
 // Copyright 2025 The IBM Research Authors.
 //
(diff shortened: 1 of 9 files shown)


chentong319:
Since we cannot really safely check the condition to remove the targeted AddOp, can we have a general transformation? For example,  "--replace-op-with-its-operand=0:/encoder/layer.*/attention/self/Add"where 0 is the operand to use, the string is the regular expression for op_name.

chentong319:
For multiple zaiu, the batch size has to be larger than one, this transformation cannot be used safely. Not sure the impact on application if this transformation is forced.

tungld:
> Since we cannot really safely check the condition to remove the targeted AddOp, can we have a general transformation? For example, "--replace-op-with-its-operand=0:/encoder/layer.*/attention/self/Add"where 0 is the operand to use, the string is the regular expression for op_name.

Yes, using `--replace-op-with-its-operand` would indeed be more specific and flexible. My only concern is that it might be a bit advanced for typical users, e.g. they need to know node names and how this option is related to BERT-like models. If you feel strongly that this transformation is the better approach, I'm happy to make the change. Please let me know your preference, and I’ll proceed accordingly.

> For multiple zaiu, the batch size has to be larger than one, this transformation cannot be used safely. Not sure the impact on application if this transformation is forced.

You are right, I will add a sentence in the option's comment to warn about this situation.

chentong319:
My concern of the current transformation is that the pattern does not guarantee the correctness of the transformation. How about adding a dump at compile time  when an Add is removed? Then we can use this option without worrying about unintended rewrite.
Is it true that the Op creating input of Softmax can be removed if it can be proved that op does not change the order. In this case, adding all 1 (or any same value) will not change the order of values in the tensor.

src/Dialect/ONNX/Transforms/IgnoreAttentionMask.cpp
```diff
@@ -0,0 +1,153 @@
...
+    Value mask, mmOutput;
+    if (A.getDefiningOp<ONNXMatMulOp>()) {
+      mask = B;
+      mmOutput = A;
+    } else if (B.getDefiningOp<ONNXMatMulOp>()) {
+      mask = A;
+      mmOutput = B;
+    } else
+      return failure();
+
```
chentong319: Do we need to check whether the AddOp is MD broadcast?

tungld: This op in the bert model is actually broadcasting.

tungld:
> My concern of the current transformation is that the pattern does not guarantee the correctness of the transformation. How about adding a dump at compile time when an Add is removed? Then we can use this option without worrying about unintended rewrite.

I understand your concern now. Let me switch to your suggestion of `"--replace-op-with-its-operand=0:/encoder/layer.*/attention/self/Add"` to directly rewrite the op we are targeting.

> Is it true that the Op creating input of Softmax can be removed if it can be proved that op does not change the order. In this case, adding all 1 (or any same value) will not change the order of values in the tensor.

Actually, the mask tensor here includes either 0 or inf. `1` and `0` in the model input `attention_mask` becomes `0` and `inf`, respectively, after several processing.

tungld:
@chentong319 I changed to using the option `--replace-op-with-its-operand`. Please take a look again. Thanks for your suggestion!

src/Dialect/ONNX/Transforms/ReplaceOpWithItsOperand.cpp
```diff
@@ -0,0 +1,101 @@
...
+      return WalkResult::advance();
+
+    // Match node name. Return when found the first one.
+    for (std::string IdNodenameRegex : this->nodeNameRegexList) {
+      size_t pos = IdNodenameRegex.find(":");
+      std::string inputID = IdNodenameRegex.substr(0, pos);
+      std::string nodeNameRegex = IdNodenameRegex.substr(pos + 1);
+      if (std::regex_match(opNodeName.str(), std::regex(nodeNameRegex))) {
+        int64_t id = std::stoi(inputID);
+        op->getResult(0).replaceAllUsesWith(op->getOperand(id));
```
chentong319, line 87: Do we want to replace all the Add in different layer with one regex like "/encoder/layer.*/attention/self/Add"?

tungld: Yes, the current code supports that.

chentong319:
Why your build failed in Jenkins Linux s390 again? Is the s390 Jenkins down?

tungld:
> Why your build failed in Jenkins Linux s390 again? Is the s390 Jenkins down?

Just merged in the fix for protobuf on s390x. Hope it works now.

PR #3303, opened 2025-10-08
Fix for benchmark breakage

AlexandreEichenberger:
Fixes that fix the benchmark breakage introduced by fusion of stick/unstick with operations.


AlexandreEichenberger:
Christopher Munoz ran the PR on a Z and this PR removed all the new failures.

PR #3306, opened 2025-10-15
Get dim from the input of reshape and transpose

tungld:
This patch adds two rules to get dimensions directly from the input of reshape and transpose.


AlexandreEichenberger (approved):
LGTM, thanks

PR #3309, opened 2025-10-16
New algorithm for stick/unstick decomposition

tungld:
The compile option `--enable-zhigh-decompose-stick-unstick` decomposes `zhigh.unstick` into `onnx.LayoutTransform + zhigh.DLF16ToF32` and `zhigh.stick` into `zhigh.F32ToDLF16 + onnx.LayoutTransform`, and tries to remove ` zhigh.DLF16ToF32` and `zhigh.F32ToDLF16` as many as possible. If there is no optimization, they will be recomposed back to `zhigh.stick` and `zhigh.unstick`.

However, the current implementation of decomposition may introduce redundant DLF16ToF32 ops and sometimes the operations can not be recomposed back if there is no benefit. These problems come from the fact that the current implement decomposes ALL `zhigh.unstick` and `zhigh.stick` without analyzing if there is any benefit or not.

To fix the above problems, this PR implements a simple cost model based on the number of added operations (N_a) and the number of removed operations (N_r) when decomposing a `zhigh.unstick`.  We will decompose `zhigh.unstick` only when:
- `N_r > N_a`: we can remove more added ops than removed ops.
- `N_r = N_a` and only data movement ops (e.g. transpose, expand) between stick and unstick: we can copy data with smaller data type.

Please see the lit tests for different cases.

### DIFF
--- a/docker/Dockerfile.llvm-project
+++ b/docker/Dockerfile.llvm-project
@@ -83,7 +83,7 @@ RUN curl -sL ${BAZEL_URL}/${BAZEL_VERSION}/bazel-${BAZEL_VERSION}-dist.zip \
     && cd .. && rm -rf bazel-${BAZEL_VERSION} ${HOME}/.cache
 
 # Install rust, cargo, and cargo-bazel
-ARG RUST_VERSION=1.88
+ARG RUST_VERSION=1.89
 ARG CARGO_BAZEL_VERSION=0.16.0
 RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | \
     sh -s -- -y --default-toolchain none \
--- a/utils/install-protobuf.sh
+++ b/utils/install-protobuf.sh
@@ -1,5 +1,5 @@
 # Install rust, cargo, and cargo-bazel
-RUST_VERSION=1.88
+RUST_VERSION=1.89
 CARGO_BAZEL_VERSION=0.16.0
 curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | \
      sh -s -- -y --default-toolchain none


tungld:

src/Accelerators/NNPA/Transform/ZHigh/ZHighDecomposeStickUnstick.cpp
```diff
@@ -24,11 +24,119 @@
...
+  while (!workList.empty()) {
+    Operation *current = workList.back();
+    workList.pop_back();
+    for (auto *user : current->getUsers()) {
+      bool isONNXOp = user->getDialect()->getNamespace() ==
+                      ONNXDialect::getDialectNamespace();
+      // ZHighQuantizedStickOp currently does not work with DLF16 in symmetric
+      // mode. Temporarily put it into the ONNX category. Once it supports DLF16
+      // with symmetric mode, it would be in the same category as ZHighStickOp.
+      isONNXOp |= isa<ZHighQuantizedStickOp>(user);
```
AlexandreEichenberger, line 79: I would suggest making a function in an existing helper file, with clear definition so that if we add another such ZHigh CPU operation, it will get added globally.

src/Accelerators/NNPA/Transform/ZHigh/ZHighDecomposeStickUnstick.cpp
```diff
@@ -24,11 +24,119 @@
...
+        continue;
+      }
+      // Continue if user is returned.
+      if (isa<func::ReturnOp>(user)) {
+        visited.insert(user);
+        numReturns++;
+        continue;
+      }
+      // Early stop if user is not an ONNX Op.
+      if (!isONNXOp)
```
AlexandreEichenberger, line 102: I don't understand this... is it a "I don't know what is happening so I am aborting", or is it "this is not beneficial no mater what", or "this should never happen".

Maybe you can add a comment that show an example of when this is triggered.

src/Accelerators/NNPA/Transform/ZHigh/ZHighDecomposeStickUnstick.cpp
```diff
@@ -24,11 +24,119 @@
...
+  //
+  // When decomposing a unstick op, a DLF16ToF32 op is propagated down to
+  // return, stick and compute ops.
+  // - When it meets a stick op, the DLF16ToF32 in unstick and the F32ToDLF16s
+  // in stick ops are removed.
+  // - When it meets a return or a compute op, each return or compute op needs
+  // one DLF16ToF32.
+  //
+  // So, the decomposition is only beneficial if the total number of removed
+  // F32ToDLF16/DLF16ToF32 is greater than the total number of added DLF16ToF32.
```
AlexandreEichenberger: Because of the +1 for stick (to account for the unstick), maybe you can add "strictly greater"
Maybe you can add a pointer to the lit test which details examples of when it is / is not seen as beneficial, or copy some of these example input here. I found the example in the littlest quite useful.

src/Accelerators/NNPA/Transform/ZHigh/ZHighDecomposeStickUnstick.cpp
```diff
@@ -24,11 +24,119 @@
...
+  // For N stick ops, we can remove N F32ToDLF16s in N stick ops plus 1
+  // DLF16ToF32 in the unstick op.
+  uint64_t numRemovedOps = numStickOps + 1;
+  // Each return or compute op needs one DLF16ToF32.
+  uint64_t numAddedOps = numReturns + numComputeOps;
+  // There is no benefit if the number of added DLF16ToF32 ops is more than or
+  // equal to the number of removed DLF16ToF32/F32ToDLF16.
+  if (numAddedOps > 1 && numAddedOps >= numRemovedOps)
+    return false;
+
```
AlexandreEichenberger, line 137: What is the outcome of this heuristic for an
```
u = unstick()
add(u)
t = transpose(u)
stick(t)
```
I count 1 stick, 1 cpu, 1 data op that is not a view.  So the answer is "true".  

In the current approach: that would means 4 loops (2 layout transform (LT), 1 dlf16to32, 1 transpose) vs 3 (unstick, transpose, stick).

It is getting a bit more complicated because:

-  We currently do fusion for "Unstick to an elementwise op"... but here there are 2 uses of the unstick, so it would disable the fusion. I don't think I handle this case yet, but the fused "unstick-element wise" could also save the output of the layer transform. Right now, this optimization only work from an "unstick".
- I am working on adding fused "LT [reshape that split 1 dim into 2] [transpose all but the last dim] [reshape that merge 2 dim into 1] [LT or dlf16To32]" which would also change this dynamic.  The limitation on the type of reshape is mainly for my sanity for code gen, based on what I have seen in granite. Right now this pattern is based on the presence of LT (Layout Transform op); potentially they could also be "extended" to work with stick/unstick.

Not sure yet how these fusion heuristics interact with this heuristic here.

src/Accelerators/NNPA/Transform/ZHigh/ZHighDecomposeStickUnstick.cpp
tungld: I am not clear about what is the functionality of the function you expect to have. Could you explain more about that?

src/Accelerators/NNPA/Transform/ZHigh/ZHighDecomposeStickUnstick.cpp
tungld: Sure, I added more comments to explain this.

src/Accelerators/NNPA/Transform/ZHigh/ZHighDecomposeStickUnstick.cpp
tungld: Yes, added "strickly greater". Thanks!

src/Accelerators/NNPA/Transform/ZHigh/ZHighDecomposeStickUnstick.cpp
tungld: For this example:
```
u = unstick(x)
a = add(u)
t = transpose(u)
s = stick(t)
```
the outcome is:
```
l = layout_transform(x)
u = dlf16tof32(l)
a = add(u)
t = transpose(l)
s = layout_transform(t)
```
where, we can remove f32todlf16 in stick.

I added this case into the lit test under the name: data_movement_and_stick_and_compute_ops

To make it work with your `unstick -> elementwise`, I think you can add support for `dlf16tof32 -> elementwise`.

AlexandreEichenberger:

tungld:

tungld:
@jenkindroid test this please

tungld:

PR #3312, opened 2025-10-16
Added comment to script and doc

AlexandreEichenberger:
Added suggestions (script and doc) for users that may only want to use `onnx-mlir-print.sh` where listing passes that changed state.


AlexandreEichenberger:

AlexandreEichenberger:

PR #3317, opened 2025-10-17
Only include perf tests if LLVM_INCLUDE_BENCHMARKS is enabled, as it …

jorickert:
…has dependencies that are only available if this option is set


jorickert:

PR #3322, opened 2025-10-29
Starting example of using table gen for pass

chentong319:
I am writing a new pass to hoist buffer alloc/dealloc out of OpenMP work sharing loop using the new approach: table gen for pass definition.
1. I added my transformation in src/Transform. One Passes.td can be used for multiple passes and we will have Passes.td in Dialect's Transform directory.
2. I did not implement the transformation yet, leaving the runOnOperation() empty.

### DIFF
--- a/src/Transform/BufferOMPLoopHoist.cpp
+++ b/src/Transform/BufferOMPLoopHoist.cpp
@@ -28,4 +28,4 @@ namespace onnx_mlir {
 std::unique_ptr<Pass> createBufferOMPLoopHoisting() {
   return std::make_unique<BufferOMPLoopHoistingPass>();
 };
-} // namespace
+} // namespace onnx_mlir


tungld:
Looks like you also need to register the pass so that it can be used with the tool `onnx-mlir-opt`, something like this in Passes.hpp:

```
/// Generate the code for registering transform passes.
#define GEN_PASS_REGISTRATION
#include "src/Transforms/Passes.hpp.inc"
```

chentong319:
> Looks like you also need to register the pass so that it can be used with the tool `onnx-mlir-opt`, something like this in Passes.hpp:
> 
> ```
> /// Generate the code for registering transform passes.
> #define GEN_PASS_REGISTRATION
> #include "src/Transforms/Passes.hpp.inc"
> ```

Thanks.

chentong319:
@tungld I added more changes.
1. Use  GEN_PASS_DECL_BUFFEROMPLOOPHOISTING for createBufferOMPLoopHoisting replace the our old way of createBufferOMPLoopHoistingPass (deprecated).
2. I put everything related to the pass into onnx_mlir name space to avoid any potential duplicate name in MLIR project. We may need to consider a systematic way of using name space in our project.

src/Transform/CMakeLists.txt
```diff
@@ -31,3 +37,16 @@ add_onnx_mlir_library(OMInstrument
...
   )
+
+add_onnx_mlir_library(BufferOMPLoopHoist
+  BufferOMPLoopHoist.cpp
+  
+  DEPENDS
+  OMTransformsPassIncGen
+
+  LINK_LIBS PUBLIC
+  OMONNXOps
```
tungld, line 48: I guess we don't have ONNX ops at this OMP level. But it depends on the actual implementation of this pass, so you can add a comment to remind of revising these libs when implementing the pass.

tungld:
> Use GEN_PASS_DECL_BUFFEROMPLOOPHOISTING for createBufferOMPLoopHoisting replace the our old way of createBufferOMPLoopHoistingPass (deprecated).

Yes, you are right.

src/Transform/CMakeLists.txt
chentong319: I will change this when I really implement the pass.

juhyunbae17:
@chentong319 Do we need to organize namespace in the other parts of the onnx-mlir too?

jorickert:
@chentong319 Could it be that this PR is missing a cmake edge?
If i do not build src/Transform/Passes.h.inc manaully first, I get the following error:
```
In file included from /scratch/onnx-mlir-devcontainer/src/Dialect/Krnl/DialectBuilder.cpp:18:
In file included from /scratch/onnx-mlir-devcontainer/src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.hpp:47:
/scratch/onnx-mlir-devcontainer/src/Pass/Passes.hpp:136:10: fatal error: 'src/Transform/Passes.h.inc' file not found
#include "src/Transform/Passes.h.inc"
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

PR #3328, opened 2025-11-04
Do not generate stablehlo.dynamic_broadcast_in_dim when shapes are same.

nikhil-m-k:
If the input and output have a dynamic shape such that the rank of input matches the rank of output and shapes are compatible (i.e. no broadcasting is needed), then the dynamic_broadcast_in_dim is effectively a no-op.


tungld:

PR #3329, opened 2025-11-07
Optimization of the ONNX TopK operator

adrian-selk:
This new contribution adds a new OMTopK.inc script, which can perform a partial heap-based sort to find the top-k values in a tensor / array. Before, the TopK operator was compiled to always call the generic omSort function, which is inefficient and slows down performance, especially when the K value is much smaller than the length of the array that is being processed. So now, the k value is a dynamic value we load in before performing the heap algorithm, and which is then used to determine the top-k-selection.
We also add some functionalities to comply with the ONNX TopK specification, e.g. to support a sorted and unsorted option that can be specified as a parameter. So far, this has just been ignored.
Also, the other scripts, like ONNXToKrnl.cpp and .hpp and TopK.cpp have been modified to emit this new function to the compiler instead of `omTensorSort`. 
Finally, the `TopK_with_canonicalize.mlir`-script has been updated to reflect the new logic.

### DIFF
--- a/src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
+++ b/src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.cpp
@@ -180,7 +180,7 @@ std::map<int64_t, int64_t> getReductionMapping(
   }
 
   for (decltype(inRank) inIndex = 0, outIndex = 0; inIndex < inRank;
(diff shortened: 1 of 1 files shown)


tungld:

tungld:

tungld:

PR #3331, opened 2025-11-10
Hoist alloc/dealloc for OpenMP wsloop

chentong319:
This is the implementation of the transformation.
The transformation walk to a omp.WsloopOp and then check the memref.AllocOp inside it. Check the conditions for the alloc Op:
1. Directly in the omp.LoopNest. No other parallel or loop like op.
2. Its dealloc is in the wsloop and guarantee there is no escape of the allocated space.
3. The dependent ops of the alloc are defined inside wsloop. 

If satisfied, the alloc and its dealloc are hoisted out.
Several test cases are added to lit test.

### DIFF
--- a/src/Transform/BufferOMPLoopHoist.cpp
+++ b/src/Transform/BufferOMPLoopHoist.cpp
@@ -13,14 +13,18 @@
 
 using namespace mlir;
 
-namespace mlir {
-#define GEN_PASS_DEF_BUFFEROMPLOOPHOISTING
-#include "src/Transform/Passes.h.inc"
-} // namespace mlir
-
 using namespace mlir;
 
 namespace {
+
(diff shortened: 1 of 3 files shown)


AlexandreEichenberger (approved):
LGTM

src/Transform/BufferOMPLoopHoist.cpp
```diff
@@ -1,11 +1,16 @@
...
 #include "mlir/IR/Operation.h"
+#include "mlir/Interfaces/LoopLikeInterface.h"
 
 #include "src/Pass/Passes.hpp"
 
+#define DEBUG_TYPE "buffer-omploop-hoisting"
+
 using namespace mlir;
 
 namespace mlir {
```
AlexandreEichenberger: @chentong319 , does it need to be mlir namespace, or can it be anonymous

PR #3340, opened 2025-11-27
Fix a bug in saving a json file and a bug in onnx to zhigh for quantization

tungld:
This PR fixes a bug in saving a json file, that is when the file exists, the saving does not override. The solution is just emptying the file in advance.

This PR also fixes a bug in lowering onnx to zhigh for quantization, where the matmul op instead of add op should be passed to DynQuantI8PatternHelper.

### DIFF
--- a/docs/DebuggingNumericalError.md
+++ b/docs/DebuggingNumericalError.md
@@ -72,6 +72,9 @@ Full options are listed under the `--help` flag.
 
 To gain more insight in what the compiler is doing, you may want to print the output of the compiler after each transformation pass. The flags are well documented in the `mlir` literature, and we have written a script that automatically add such flags by using a custom script.
 This script is found here `utils/onnx-mlir-print.sh`, and it is used by adding all desired compiler option, plus one last argument indicating the file in which to save the log of the compiler output.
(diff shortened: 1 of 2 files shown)


src/Accelerators/NNPA/Compiler/NNPACompilerUtils.cpp
```diff
@@ -245,6 +245,12 @@ void addPassesNNPA(mlir::OwningOpRef<mlir::ModuleOp> &module,
 
   // Override pass configurations.
   configurePasses();
+  // Empty the save json config file if it exists.
+  if (!nnpaSaveConfigFile.empty()) {
+    std::error_code EC;
+    llvm::raw_fd_ostream OS(nnpaSaveConfigFile, EC, llvm::sys::fs::OF_None);
```
AlexandreEichenberger, line 251: Fine by me, @chentong319 is the introduction of LLVM dependence here an issue for your python lightweight driver?

chentong319: No. This code is in model.so

PR #3341, opened 2025-12-02
Make dynamic dimension analysis for Reshape more general

tungld:
This PR revised the dynamic dimension analysis for Reshape by making it more general and supporting more cases.

In particular, the old analysis for Reshape handles the two following special cases:
- both input and output have **one** dynamic dimension and same static size, e.g. `tensor<?x12x64xf32>` and `tensor<?x768xf32?`
- both input and output have **exactly two** dynamic dimensions without any static dimension, e.g. `tensor<?x?xf32>` and `tensor<?x?xf32>`.

This PR generalizes the analysis algorithm to cover the two cases and to support more cases whose the number of dynamic dimensions is more than 2, for example reshape from `tensor<?x?x768xf32>` to `tensor<?x?x12x64xf32>` in granite-embedding models.

In general, this PR handles a reshape where:
- the products of the static dimensions in input and output are the same, and
- input and output have the same number of dynamic dimensions, say N, and
- N-1 dynamic dimensions in input and output are the same, respectively.

In this case, the remaining dynamic dimensions in input and output must be the same.

### DIFF
--- a/src/Runtime/OMTensorList.inc
+++ b/src/Runtime/OMTensorList.inc
@@ -77,16 +77,20 @@ OMTensorList *omTensorListCreate(OMTensor **tensors, int64_t n) {
     return NULL;
 
   TIMING_INIT_START(tensor_list_create)
-  size_t omts_bytes = sizeof(OMTensor *) * n;
   list->_size = n;
-  list->_omts = (OMTensor **)malloc(omts_bytes);
+  if (n > 0) {
+    size_t omts_bytes = sizeof(OMTensor *) * n;
+    list->_omts = (OMTensor **)malloc(omts_bytes);
 
-  if (!list->_omts) {
-    free(list); // free the previously allocated memory in case of an error
-    return NULL;
+    if (!list->_omts) {
+      free(list); // free the previously allocated memory in case of an error
+      return NULL;
+    }
+    // Copy the given OMTensors pointers to an array owned by OMTensorList
+    memcpy(list->_omts, tensors, omts_bytes);
+  } else {
+    list->_omts = NULL;
   }
-  // Copy the given OMTensors pointers to an array owned by OMTensorList
-  memcpy(list->_omts, tensors, omts_bytes);
   TIMING_STOP_PRINT(tensor_list_create);
   return list;
 }
@@ -109,7 +113,8 @@ void omTensorListDestroyShallow(OMTensorList *list) {
   if (!list)
     return;
   // Free the list as well as the pointer to the OMTensor array
-  free(list->_omts);
+  if (list->_omts)
+    free(list->_omts);
   free(list);
 }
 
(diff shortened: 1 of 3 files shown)


test/mlir/onnx/onnx_dim_analysis.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: onnx-mlir-opt --onnx-dim-analysis %s -split-input-file | FileCheck %s
+// RUN: onnx-mlir-opt --onnx-dim-analysis --canonicalize %s -split-input-file | FileCheck %s
```
tungld, line 1: Add `--canonicalize` to make the lit test output more stable.

src/Dialect/ONNX/ONNXDimAnalysis.cpp
```diff
@@ -941,82 +941,79 @@ void DimAnalysis::visitDim(
...
-    //    - dim j and i are the only dynamic dimension in data and output,
-    //    respectively, and
+    // Special case: when data and output have the same number of dynamic
+    // dimensions (N), the dynamic dimension output[i] is the same as the
+    // dynamic dimension data[j] if
     //    - the products of the remaining static dimensions in data and output
-    //    are equal.
+    //    are equal, and
+    //    - the remaining N-1 dynamic dimensions in data and output are the
+    //    same, respectively.
```
AlexandreEichenberger, line 950: Just to understand, isn't that the same as saying that all but 1 dynamic dimensions can be shown to be the same (where the other dims are either static or dynamic)? That feel clearer to me.

Maybe that is what you have in the code but is true that if we had 2 shapes (where d's with the same indices are known to be equivalent)

`(d0, 16, d1, 64, d2)` and `(d1, d2, 16/2, d3, 64*2)`, we could declare a match? 

So we can accumulate all static sizes, and pairwise match dynamic regardless of position in the rank vector, and if the last 2 are dynamic, then they must be equal.  Is that what you are doing? Above, `d1` and `d2` match, then `d0 == d3`.

Or maybe you do something more than even that?

tungld: > isn't that the same as saying that all but 1 dynamic dimensions can be shown to be the same (where the other dims are either static or dynamic)?

Yes, they are the same but for dynamic only. For static dims, we compare the accumulation of all static dims.

> So we can accumulate all static sizes, and pairwise match dynamic regardless of position in the rank vector, and if the last 2 are dynamic, then they must be equal. Is that what you are doing? Above, d1 and d2 match, then d0 == d3.

Yes, it is what I am doing. The position of dynamic dims does not matter.

AlexandreEichenberger: Awesome, thanks.

AlexandreEichenberger (approved):
LGTM, thanks for this useful addition to dynamic shape analysis.

PR #3344, opened 2025-12-04
Support empty inputs to the compiled model and remove input arguments in check-onnx-backend-constant

tungld:
This patch adds support of empty inputs passing to the inference of a compiled model. It means the model does not have inputs.

The above feature is tested by `check-onnx-backend-constant` where the model inputs are turned to constants in the model so that the model does not have inputs anymore.


tungld:

tungld:

PR #3346, opened 2025-12-09
Extend the verifyInputTensors to shape information provided with compilation option

chentong319:
1. Add two new options: shapeInformationUB and shapeInformationLB. They are in the similar format to shapeInformation but specify the upper bound and lower bound of the dimensions of input. One use case is the max sequence length. Another  possible use case is minimum batch size.
2. Insert runtime check on the actual inputs if the UB or LB is specified.
3. I tested cases. Not sure how to add such runtime test cases which may generate assertion error into our test set.

### DIFF
--- a/docker/Dockerfile.llvm-project
+++ b/docker/Dockerfile.llvm-project
@@ -91,20 +91,6 @@ RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | \
     && rustup install ${RUST_VERSION} \
     && cargo install cargo-bazel --version ${CARGO_BAZEL_VERSION}
 
-# Install absl for new version of protobuf
-ARG ABSL_VERSION=20240722.1
-ARG ABSL_URL=https://github.com/abseil/abseil-cpp.git
-ARG ABSL_DIR=abseil-cpp
-RUN git clone -b ${ABSL_VERSION} --recursive ${ABSL_URL} ${ABSL_DIR} \
-    && mkdir -p ${ABSL_DIR}/build \
-    && cd ${ABSL_DIR}/build \
-    && CC=clang CXX=clang++ \
-       cmake -DCMAKE_INSTALL_LIBDIR=lib \
-             -DABSL_PROPAGATE_CXX_STD=ON \
-             -DBUILD_SHARED_LIBS=ON .. \
-    && make -j${NPROC} install && ldconfig \
-    && cd ../.. && rm -rf ${ABSL_DIR} ${HOME}/.cache
-
 # Install protobuf
 ARG PROTOBUF_VERSION=6.31.1
 ARG PROTOBUF_URL=https://github.com/protocolbuffers/protobuf.git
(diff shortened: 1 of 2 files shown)


src/Conversion/KrnlToLLVM/KrnlEntryPoint.cpp
```diff
@@ -527,6 +537,86 @@ class KrnlEntryPointOpLowering : public OpRewritePattern<KrnlEntryPointOp> {
...
+          if (bound == -1) // No bound info for this dimension
+            continue;
+          Value actualDim = create.llvm.load(int64Ty,
+              create.llvm.getElemPtr(getPointerType(context, int64Ty), int64Ty,
+                  sizesArrayPtr,
+                  ArrayRef<LLVM::GEPArg>{static_cast<int32_t>(dimID)}));
+
+          switch (boundType) {
+          case 0: // UB
+            noLessOrFailed(module, rewriter, loc,
```
tungld, line 596: Look like this should be for LB.

src/Conversion/KrnlToLLVM/KrnlEntryPoint.cpp
```diff
@@ -527,6 +537,86 @@ class KrnlEntryPointOpLowering : public OpRewritePattern<KrnlEntryPointOp> {
...
+                create.llvm.constant(int64Ty, static_cast<int64_t>(bound)),
+                actualDim,
+                "verifyInputTensors failed: the upper bound for the input " +
+                    std::to_string(inputID) + " of dimension " +
+                    std::to_string(dimID) +
+                    " is set by --shapeInformationLB as " +
```
tungld: shapeInformationUB?

src/Conversion/KrnlToLLVM/KrnlEntryPoint.cpp
```diff
@@ -527,6 +537,86 @@ class KrnlEntryPointOpLowering : public OpRewritePattern<KrnlEntryPointOp> {
...
+                    std::to_string(bound) + ", but got ");
+            break;
+          case 1: // LB
+            noGreaterOrFailed(module, rewriter, loc,
+                create.llvm.constant(int64Ty, static_cast<int64_t>(bound)),
+                actualDim,
+                "verifyInputTensors failed: the upper bound for the input " +
```
tungld: nit: "the upper bound" -> "the lower bound"

tungld (commented):
Is it possible to put upper and lower bounds into the argument attributes when importing a model and then to the input signature when lowering onnx.EntryPoint to krnl.entry_point. 

For example, the input argument is like this:

```c
func.func @main_graph(%arg0: tensor<?xf32> {onnx.name = "x", 
   onnx.dim_params = "0:a", onnx.dim_lbs="0:5", onnx.dim_ubs="0:50"}
```

When lowering onnx.EntryPoint  to krnl, `krnl.entry_point`we will have this json info for input signature:
```json
{"type" : "f32" , "dims" : [-1] , "lbs": [5], "ubs": [50], "name": "x"}
```

We can even replace `"dims": [-1]` by `"dims": ['a']`.

chentong319:
Yes, we can transfer the UB and LB info to function op. Then we support information coming from command line and from model (.mlir or .onnx).

chentong319:
To avoid the confusion of UB or LB, I introduced a enumerate type.

tungld:
> Yes, we can transfer the UB and LB info to function op. Then we support information coming from command line and from model (.mlir or .onnx).

@chentong319 please let us know if you plan to implement this here or in another PR. Thanks!

chentong319:
In another PR.

chentong319:

chentong319:

PR #3350, opened 2025-12-12
[onnxmIirtorch] Export/import cache from disk and add config.py

tungld:
This PR improved the pytorch backend for onnx-mlir by adding these features:
- Cache the compiled model to disk and load it when starting a new process.
- Add config.py to allow users set configurations for the backend
- For each constants or parameters in the FX graph, use the first 3 values (the number can be configured via config) when hashing the graph, which makes the hash unique.
- After N times of hashing the FX graph, if the graph is unchanged, stop hashing to reduce the overhead. N can be configured via config, default is 3.


tungld:

src/Runtime/python/onnxmlirtorch/src/onnxmlirtorch/backend.py
```diff
@@ -145,22 +142,68 @@ def __init__(self, gm: torch.fx.GraphModule, compile_options) -> None:
...
+            # Use stable names for placeholders and symbolic dimensions.
             if node.op == "placeholder":
-                node_info.append(f"om_placeholder_{placeholder_counter}")
+                if "example_value" in node.meta and isinstance(
+                    node.meta["example_value"], torch.Tensor
+                ):
+                    shape = []
+                    for d in node.meta["example_value"].shape:
+                        s = str(d)
+                        if isinstance(d, torch.SymInt):
```
chentong319, line 166: Any chance of scalar of float type?

chentong319 (approved):
LGTM!

src/Runtime/python/onnxmlirtorch/src/onnxmlirtorch/backend.py
tungld: Yes, there is a SymFloat type in a fx graph, but for symbolic dimensions, only SymInt is used. So we don't need to handle SymFloat here.

PR #3352, opened 2025-12-17
remove mlir usePropertiesForAttributes option as it is deprecated with option of 0

Sunny-Anand:
https://discourse.llvm.org/t/rfc-deprecating-usepropertiesforattributes-0/87536/9 
https://mlir.llvm.org/docs/ReleaseNotes/


tungld:
This PR in LLVM did some updates when removing usePropertiesForAttributes: https://github.com/llvm/llvm-project/pull/123505, it may help us to know where we need to change in onnx-mlir also.

In my understanding, we now access attributes via properties, for example, instead of `IntegerAttr axis adaptor.getAxisAttr()`, we do `IntegerAttr axis = adaptor.getProperties().axis`

tungld:
@Sunny-Anand I created another PR #3357 to remove the flag.

Sunny-Anand:
Close this as completed via https://github.com/onnx/onnx-mlir/pull/3357

PR #3357, opened 2026-01-07
Remove mlir usePropertiesForAttributes option as it is deprecated in upcoming LLVM

tungld:
`usePropertiesForAttributes` option in dialect definition will be removed soon in LLVM. 
onnx-mlir disables "use properties for attributes" by explicitly setting `usePropertiesForAttributes=0`, but LLVM will remove the option and enable "use properties for attributes" by default.
This PR removes the option in order to update LLVM smoothly.

https://discourse.llvm.org/t/rfc-deprecating-usepropertiesforattributes-0/87536/9
https://mlir.llvm.org/docs/ReleaseNotes/

### DIFF
--- a/src/Dialect/ONNX/ONNXOps/Canonicalize.cpp
+++ b/src/Dialect/ONNX/ONNXOps/Canonicalize.cpp
@@ -1058,7 +1058,8 @@ private:
     Type elType = onnx_mlir::getElementType(input.getType());
     Type unrankedType = UnrankedTensorType::get({elType}); // placeholder
     Value transposed = create.transpose(unrankedType, input, perm);
-    auto transposeOp = mlir::dyn_cast<ONNXTransposeOp>(transposed.getDefiningOp());
+    auto transposeOp =
+        mlir::dyn_cast<ONNXTransposeOp>(transposed.getDefiningOp());
     inferShapes(transposeOp); // sets transposed's shape
     return transposed;
   }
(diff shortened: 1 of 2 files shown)


src/Accelerators/NNPA/Dialect/ZHigh/ZHighOps/QuantizedStick/QuantizedStick.cpp
```diff
@@ -82,7 +82,8 @@ void ZHighQuantizedStickOp::build(OpBuilder &builder, OperationState &state,
 //===----------------------------------------------------------------------===//
 
 LogicalResult ZHighQuantizedStickOpShapeHelper::computeShape() {
-  ZHighQuantizedStickOp::Adaptor operandAdaptor(operands);
+  auto stickOp = llvm::cast<ZHighQuantizedStickOp>(op);
```
Sunny-Anand: Should we use `llvm::dyn_cast` for all the ops?

Sunny-Anand (approved):
Thanks for a comprehensive PR. LGTM

src/Accelerators/NNPA/Dialect/ZHigh/ZHighOps/QuantizedStick/QuantizedStick.cpp
tungld: Yes, it's a good suggestion. I changed them to `mlir::dyn_cast`. Thanks!

PR #3363, opened 2026-01-16
Fix protobuf build error due to undefined reference to absl functions

gongsu832:
- let protobuf itself download and build abseil-cpp
- must specify -DCMAKE_CXX_STANDARD=17 since protobuf requires c++17 but clang 14 on Ubuntu Jammy defaults to c++14


chentong319 (approved):
LGTM!

PR #3364, opened 2026-01-18
Simplify the source code for light weight PyRuntime on Linux 

chentong319:
1. Light weight PyRuntime will have the same source code as the normal onnx-mlir on Linux. On windows, locks are needed when dynamic lib is loaded. Therefore, the llvm utility is used only for windows.
2. The only difference for Linux is the CMakefile.txt: only the a subset of code is needed to build the light weight PyRuntime.
3. Fixed the support for floating conversion issue.
4. Fixed the new issue with absl lib.
5. Tested locally. In future, we will add the running test to onnx-mlir or zdlc_pyrt.

### DIFF
--- a/src/Runtime/ExecutionSession.cpp
+++ b/src/Runtime/ExecutionSession.cpp
@@ -13,7 +13,10 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include <algorithm>
+#include <cctype>
 #include <errno.h>
+#include <filesystem>
 #include <string.h>
 
 #include <iostream>
@@ -52,7 +55,16 @@ void ExecutionSession::Init(
     std::string fname = llvm::sys::path::filename(sharedLibPath).str();
     llvm::SmallString<256> fnameWithoutExt(fname);
     llvm::sys::path::replace_extension(fnameWithoutExt, "");
-    tag = fnameWithoutExt.str().lower();
+    tag = fnameWithoutExt.lower();
+#else
+    std::string fnameWithoutExt = std::filesystem::path(sharedLibPath)
+                                      .filename()
+                                      .replace_extension("")
+                                      .string();
+    std::transform(fnameWithoutExt.begin(), fnameWithoutExt.end(),
+        fnameWithoutExt.begin(),
+        [](unsigned char c) { return std::tolower(c); });
+    tag = fnameWithoutExt;
 #endif
   }
 


src/Runtime/CMakeLists.txt
```diff
@@ -72,14 +72,15 @@ set_target_properties(OMTensorUtils
   )
 
 if (ONNX_MLIR_ENABLE_PYRUNTIME_LIGHT)
```
AlexandreEichenberger, line 74: @chentong319 is there a mismatch? In the `CmakeList.txt`, the flag that decides between the light or the heavy version is `ONNX_MLIR_ENABLE_PYRUNTIME_LIGHT`. But in the `ExecutionSession.cpp` file, the flag that determines whether llvm is used or not is `_WIN32`. 

I would have imagined that we ought to use the same flag in both places, my mistake if I misunderstand the context in which these flags are used.

src/Runtime/ExecutionSession.cpp
```diff
@@ -50,7 +48,7 @@ void ExecutionSession::Init(
   if (tag == "") {
     // ToFix: equivalent implementation of llvm utilities.
```
AlexandreEichenberger, line 52: Should the code be added, or an assert if we go to that path? It seems that we should fix this for the next release to be complete

chentong319: Sure. I will fix this.

src/Runtime/ExecutionSession.cpp
```diff
@@ -50,18 +51,27 @@ void ExecutionSession::Init(
...
+        [](unsigned char c) { return std::tolower(c); });
+    tag = fnameWithoutExt;
 #endif
   }
 
   // tag = "NONE" to use functions without tag.
   std::string lowDashTag;
   // ToFix: equivalent implementation of llv::StringRef
-#ifndef ENABLE_PYRUNTIME_LIGHT
+#if defined(_WIN32)
```
AlexandreEichenberger, line 74: the version for linux seems to be something like this:
```

#include <strings.h>  // for strcasecmp

int result = strcasecmp(tag, "NONE") != 0;
```

src/Runtime/ExecutionSession.cpp
```diff
@@ -50,18 +51,27 @@ void ExecutionSession::Init(
   if (tag == "") {
     // ToFix: equivalent implementation of llvm utilities.
```
AlexandreEichenberger, line 52: comment can be removed.

PR #3365, opened 2026-01-20
Fix python code format for the newer black version

tungld:
Recently, GitHub Action Black Format failed because a newer version of black was installed. This PR updates python codes to work with the new black version.


chentong319:
Sorry. I put a duplicated PR.

PR #3366, opened 2026-01-20
Fix errors caused by the new version of python formatter: black

chentong319:
The new version has some differences in the result.


Sunny-Anand (approved):
LGTM

PR #3369, opened 2026-01-22
[pytorch backend] Fall back to the eager mode if exporting the fx graph to onnx failed

tungld:
This PR adds a fallback in the compile mode to use the eager mode if it fails to export the fx graph to onnx.


tungld:
This code was in a merged PR #3382. Closing it.

PR #3373, opened 2026-01-26
Modernize pass

juhyunbae17:
This PR modernizes passes on src/Tranform. Other passes will be modernized later in different PRs.

### DIFF
--- a/src/Runtime/python/onnxmlirtorch/requirements.txt
+++ b/src/Runtime/python/onnxmlirtorch/requirements.txt
@@ -1,0 +1,3 @@
+onnxscript==0.5.7
+torch==2.10.0
+transformers==4.57.6
--- a/src/Runtime/python/onnxmlirtorch/src/onnxmlirtorch/backend.py
+++ b/src/Runtime/python/onnxmlirtorch/src/onnxmlirtorch/backend.py
@@ -25,6 +25,7 @@ from collections import deque
 
 import numpy as np
 import torch
+from torch.export import Dim
 from torch._inductor.codecache import (
     _ident,
     extract_tensor_metadata_for_cache_key,
@@ -38,7 +39,7 @@ from torch._subclasses.fake_tensor import (
 
 from .onnxmlirdocker import InferenceSession
 from .sessioncache import SessionCache, CacheValue
-from . import config
+from . import config, fx_utils
 
 """
 This file provides an onnx-mlir compiler backend for torch.compile().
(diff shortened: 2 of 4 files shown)


chentong319:
I saw that you are modernizing more than one pass. I suggested try only one pass in one PR? If your modification passes the test, you can finish this PR and you can have a solid correct example for your future work.
If you want keep this PR, stop at the krnl pass. 
By the way, any reason to have a separate krnlpass.td?

chentong319:

src/Pass/Passes.hpp
```diff
@@ -162,6 +158,12 @@ std::unique_ptr<mlir::Pass> createConvertONNXToLinalg();
 #define GEN_PASS_REGISTRATION
 #include "src/Transform/Passes.h.inc"
 
+// pass registration for krnl namespace
+namespace krnl {
```
juhyunbae17, line 164: This separation is to apply different namespace on the specific registration since the registration code that is generated by Tablegen is not featured by 'ifdef' bracket.

By separating them into different .td files, we can specify namespace on the specific scope.

juhyunbae17:
Since this modernization doesn't change semantics of the passes but the shape, I think they are okay to be in the same PR and they are all Transform passes. If you prefer smaller PR than this, I could do so. Perhaps the ones under the same compile target can be together in one PR(e.g., ProcessKrnlParallelClause and ProcessScfParallelPrivate together and Instrument things in another and then lowerkrnl and BufferOMPLoopHoist in each PR)

Regarding separate .td for krnl namespace, I found that the registration code generated by Tablegen is not featured by 'ifdef' so that we cannot control the namespace on these generated registration code. So, I split them into different .td files. Please also refer to [this](https://github.com/onnx/onnx-mlir/pull/3373/changes/f8070baf33360d9f6e51f9c17162de4f69b4b1b6#r2732204921)

chentong319:

chentong319:

juhyunbae17:
I have just signed off two commits

juhyunbae17:

chentong319:

chentong319:
@juhyunbae17 Can you fix the format issue? 
If you click the `...` following "GitHub Action Clang Format / check (pull_request)", you will see which lines are wrong. You can also follow the Workflow.md to reformat the files you modified.

juhyunbae17:
Yes, I am working on it. I have set up .clang-format in this repo. It will be updated soon. Then, I'll ask you to initiate the test

chentong319:
And your failure on Mac is different from other's. Need some investigation.

juhyunbae17:
@chentong319 There was compile dependency problem where referring source code was compiled before TableGen-generated .inc Now, it is fixed and onnx-mlir build itself is okay.

I also corrected format.

src/Dialect/ONNX/CMakeLists.txt
```diff
@@ -112,6 +112,7 @@ add_onnx_mlir_library(OMONNXOps
   OMResultTypeInferenceOpInterfaceIncGen
   OMShapeInferenceOpInterfaceIncGen
   OMTransformsPassIncGen
+  OMTransformsKrnlPassIncGen
```
tungld: Ideally, ONNX dialect should not depend on Krnl-related things, but Krnl does depend on ONNX. Any reason to add this dependency which potentially causes cyclic linking?

juhyunbae17: @tungld This is good point and I agree. How about moving it to src/Dialect/Krnl ? This also fixes the compile issue.

```
diff --git a/src/Dialect/Krnl/CMakeLists.txt b/src/Dialect/Krnl/CMakeLists.txt
index 683e4500..899e0ae0 100644
--- a/src/Dialect/Krnl/CMakeLists.txt
+++ b/src/Dialect/Krnl/CMakeLists.txt
@@ -18,6 +18,7 @@ add_onnx_mlir_library(OMKrnlOps
   DEPENDS
   OMKrnlIncGen
   OMSpecializedKernelOpInterface
+  OMTransformsKrnlPassIncGen
```


The original intention of dependency was to compile the TableGen (for krnl pass in Transform) before compiling files that refer to generated .inc files. Otherwise, below compile error occurs.

```
onnx-mlir/src/Pass/Passes.hpp:129:10: fatal error: 'src/Transform/PassesKrnl.h.inc' file not found
  129 | #include "src/Transform/PassesKrnl.h.inc"
```

FYI, for the reason separating Krnl part from pass.td in Transform, please refer to [this](https://github.com/onnx/onnx-mlir/pull/3373#issuecomment-3805453652)

tungld: Yes, moving it to Krnl would work. Thanks!

chentong319:

PR #3375, opened 2026-01-28
CMake of absl for light-weight PyRuntime

chentong319:
1. PyExecutionSessionBase.cpp includes code from onnx/onnx_pb.h. This introduced extra dependence, such as absl. The extra dependence complicated the light-weight PyRuntime. This PR carved out only the enum type needed from onnx and remove the dependence of onnx from PyExecutionSession. 
2. A different way to specify the dependence of onnx on absl. Previously, when Builder needs to link with onnx, we add absl into Builder library. This PR adds absl to onnx. Therefore, any component linked to onnx will be automatically linked to absl. It will easier if onnx has more dependency or more user.
3. Fix the handling of tag.

### DIFF
--- a/src/Builder/FrontendDialectTransformer.cpp
+++ b/src/Builder/FrontendDialectTransformer.cpp
@@ -539,7 +539,7 @@ private:
           inputDimParams.emplace_back(inputDimParamsFromOption[inputIndex]);
         else if (!inputDimParamsFromOptionForAllArgs.empty())
           inputDimParams.emplace_back(inputDimParamsFromOptionForAllArgs);
-        else if (!dimParams.empty())
+        else
           inputDimParams.emplace_back(dimParams);
 
         argTypes.emplace_back(argTy);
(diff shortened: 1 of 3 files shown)


chentong319:
Currently, the PR failed for 
`[test/backend/CMakeFiles/check-onnx-backend-signature.dir/build.make:71: test/backend/CMakeFiles/check-onnx-backend-signature] Error 5`.

src/Runtime/python/CMakeLists.txt
```diff
@@ -32,11 +33,28 @@ else()
...
+    # ToFix: OMMlirUtilities used for float16 is temporarily excluded because
+    # it need some support from LLVM.
+    #OMMlirUtilities
     pybind11::embed
     pybind11::python_link_helper
   )
 endif()
 
+# Refer to onnx-mlir/CMakeList.txt
+# Since PyExecutionSessionBase include onnx/onnx_pb.h, absl lib may be needed.
```
tungld, line 44: I wonder why PyExecutionSessionBase needs `onnx/onnx_pb.h`? Since we only need an .so file for the session, I hope we can remove the dependency if it exists.

chentong319: I checked. It actually only uses the enum type onnx::TensorProto indirectly defined in onnx_pb.h . I now include the definition in a separate file t cut any link to onnx or absl.

tungld: Do you need to remove the comment here? The code was commented out also.

chentong319: Yes, I will remove this comment and add some comments of the new code.

Issue #3354, opened 2025-12-17
Optimize ConvOp on CPU


tungld:
onnx.Conv has not been optimized yet for CPU. We would work on generating an optimized krnl code or utilizing Linalg dialect to generate code.

chentong319:
Are we going to call zdnn MatMul in the implementation instead of pure CPU code?

chentong319:
Yes, Linalg implementation is a good choice.

PR #3376, opened 2026-01-29
Implement ONNX Basic Conv to Linalg lowering

kimm240:
This PR adds support for lowering ONNX Conv operations to Linalg dialect. It is related to #3354.

### Features
- Lowers `onnx.Conv` to `linalg.conv_2d_nchw_fchw`
- Supports 2D convolution (4D tensors)
- Handles variable strides

### Current Limitations
- Only supports: `padding=0`, `dilation=1`, `bias=none`, `group=1`
- Static shapes only (dynamic shapes TODO)

### Changes
- Added `src/Conversion/ONNXToLinalg/NN/Conv.cpp` with Conv lowering pattern
- Added test cases in `test/mlir/conversion/onnx_to_linalg/NN/Conv.mlir`
- Fixed `ONNXNoneOp` by removing `hasFolder` trait to prevent MLIR matcher crash

### Testing
- Basic case: stride=1, padding=0, dilation=1
- Stride case: stride=2
- Rejection cases: padding>0, bias present, dilation>1


src/Dialect/ONNX/AdditionalONNXOps.td
```diff
@@ -294,7 +294,12 @@ def ONNXLayoutTransformOp:ONNX_Op<"LayoutTransform", [Pure,
 
 //===----------------------------------------------------------------------===//
 // NoneOp.
-def ONNXNoneOp : ONNX_Op<"NoValue", [ConstantLike, Pure]> {
+// Note: Do NOT mark this op as ConstantLike. It is a sentinel for "no value"
+// and does not behave like a regular foldable constant tensor. Marking it
+// ConstantLike causes MLIR's generic constant matchers (m_Constant) to assert
+// when folding fails. Keeping only Pure here avoids those issues while still
+// allowing ONNX to model optional inputs.
+def ONNXNoneOp : ONNX_Op<"NoValue", [Pure]> {
```
tungld, line 302: Thanks for fixing this!

chentong319:
Is the limitation of constant shape caused by the creation of output tensor?

kimm240:
@chentong319 
Yes, that's right. For this initial PR, I intended to keep the implementation minimal and focused on the core conversion logic for the most common cases. I plan to implement dynamic shape support, including the runtime output tensor shape calculation, in a follow-up PR.

tungld:
@kimm240 many tests failed. Could you take a look to see why they failed? Thanks!

tungld:

kimm240:
Hi @tungld,
I’ve been investigating the test regressions caused by the changes to `ONNXNoneOp`. I'd like to share the technical issue I've encountered.

- The Issue with ConstantLike
As discussed, marking `ONNXNoneOp` as ConstantLike triggers crashes in the GreedyPatternRewriter when it encounters non-ONNX patterns (like m_Constant in Arith/Tensor dialects) during the Linalg conversion.

- The materializeConstant Constraint
To avoid the crash, I initially removed the ConstantLike trait. However, this led to an assertion failure in MLIR's core infrastructure: materializeConstant is required to produce an operation with the ConstantLike trait.

- The nullptr Experiment and Regressions
I attempted to bypass this by returning nullptr in materializeConstant when the value is a UnitAttr (None) in latest commit. While this prevents the assertion crash, it introduced another failures in existing tests. It appears that returning nullptr prevents `ONNXNoneOp` from being materialized at expected stages, altering the optimization order and leading to IR that no longer matches the current FileCheck expectations.

I will update the PR as soon as I have a stable fix.

tungld:

kimm240:
@tungld 
`ONNXNoneOp` is no longer treated as a "foldable constant," so the timing of its materialization and the trajectory of canonicalization have changed slightly.

This led to minor shifts in SSA naming and attribute formatting (e.g., the appearance of <{value}> in some cases).
I have updated the FileCheck patterns in the existing tests to match this new IR state in commit 
ab313b5, 272c993.

I have verified that the semantics remain unchanged, but I would appreciate it if you could double-check to be certain.

tungld:
> I have verified that the semantics remain unchanged, but I would appreciate it if you could double-check to be certain.

@kimm240 thank you for finding a solution for this issue! It looks good and I believe that duplicated onnx.None ops can be solved by calling `--cse` pass.

PR #3377, opened 2026-01-29
Fix a bug in importing dim_param

tungld:
Importing dim_param from an onnx model was wrong if there is a mix of dynamic inputs and static inputs.

For example, the 1st input is of shape [1x2xf32] and the 2nd input is of shape [1x?xf32]. In this case the dim_param of the 2nd input was put into the 1st input, e.g. 

```
func.func @main_graph(%arg0: tensor<1x2xf32> {onnx.dim_params = "1:sequence_len", onnx.name = "X"}, %arg1: tensor<1x?xf32> {onnx.name = "Y")
```
where `onnx.dim_param` was for the 1st arg, which was wrong.

Correct output must be:
```
func.func @main_graph(%arg0: tensor<1x2xf32> {onnx.name = "X"}, %arg1: tensor<1x?xf32> {onnx.dim_params = "1:sequence_len", onnx.name = "Y")
```

### DIFF
--- a/src/Builder/FrontendDialectTransformer.cpp
+++ b/src/Builder/FrontendDialectTransformer.cpp
@@ -1506,7 +1506,7 @@ private:
       for (size_t k = 0; k < funcAttrsToMove.size(); ++k) {
         if (i < funcAttrsToMove[k].size()) {
           auto v = mlir::cast<StringAttr>(funcAttrsToMove[k].getValue()[i]);
-          if (!v.getValue().empty()) {
+          if (v && !v.getValue().empty()) {
             NamedAttribute namedAttr =
                 builder_.getNamedAttr(argAttrNames[k], v);
             argAttrs.emplace_back(namedAttr);


src/Builder/FrontendDialectTransformer.cpp
```diff
@@ -1505,10 +1505,10 @@ class FrontendGenImpl {
       SmallVector<NamedAttribute, 2> argAttrs;
       for (size_t k = 0; k < funcAttrsToMove.size(); ++k) {
         if (i < funcAttrsToMove[k].size()) {
-          auto name = mlir::cast<StringAttr>(funcAttrsToMove[k].getValue()[i]);
-          if (name) {
+          auto v = mlir::cast<StringAttr>(funcAttrsToMove[k].getValue()[i]);
```
chentong319, line 1508: Should we document that if attribute of dim_param is attached,  it will not be empty string ""?
Another solution is that "" is allowed but the user of dim_param has to check. 
I think our current one is better.

chentong319 (approved):
LGTM!

src/Builder/FrontendDialectTransformer.cpp
tungld: In this PR, "" is allowed and in that case it is the same as no `dim_param` attribute. So we cover all situations.

PR #3464, opened 2026-04-09
disable hsperfdata

gongsu832:
Disable JVM hsperfdata creation to avoid occasional locking failure warning message from clobbering up JSON test output on stdout


AlexandreEichenberger (approved):
LGTM, thanks for beefing up our infra, much appreciated

PR #3467, opened 2026-04-10
Disable dynamic size for NNPA Pool Operations

AlexandreEichenberger:
There is a dimension maximum size limit for the NNPA Pooling operations.
Before we checked only for static sizes, and assumed that dynamic shapes were fine
This led runs with errors as limit was exceeded.

This PR disable this behavior; and conservatively disable pool for dynamic shapes.

Investigation of the code also revel that we disregard shape restrictions for dynamic sized matmul.
Convolutions are fine (we check for dynamic shape).

I added a new flag `--nnpa-disable-shape-restriction` that disable the shape tests for dynamic sizes for POOL and CONV. Using this, a user can force all conv and pool operations with dynamic shapes to go to the NNPA. User can fine-tune by using the json file to then force specific ops to go to the CPU if further tuning is required.

@tungld is it true that no dynamic shape testing is needed for matrix multiply, because we use zDNNx to handle this? If so, we could potentially do the same for pool and conv, though it is a bit more difficult because when splitting the H or W dimensions, filter look at adjacent values so some replication would need to be done at the boundaries. Not impossible but a bit harder.


src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
```diff
@@ -151,30 +151,32 @@ bool checkLegalityPoolOpsCommon(
...
-    int64_t stridesW = shapeHelper.strides[1];
-    bool checkH = meetPoolParamRestrictions(op.getOperation(), inputShapeH,
-        kernelShapeH, stridesH, outputShapeH, paddingType);
-    if (!checkH)
-      return false;
-    bool checkW = meetPoolParamRestrictions(op.getOperation(), inputShapeW,
-        kernelShapeW, stridesW, outputShapeW, paddingType);
-    if (!checkW)
-      return false;
+  // Check parameter restrictions for maxpool2d/avgpool2d for each axis.
```
Sunny-Anand: Should we be checking for the strides and height dimensions here and if they satisfy as per zDNN checks, run those ops on the NNPA? I would imagine if strides >0 and height <1024 at runtime are present, they be sent to NNPA.

AlexandreEichenberger: It is checked just below with `meetPoolParamRestrictions` for both the H and the W dim. Are you referring to another test?

Sunny-Anand: I didn't see the check in the code. Yeah that was the check I was referring to. Thanks for poinint out.

tungld (approved):
LGTM!

tungld:
> is it true that no dynamic shape testing is needed for matrix multiply, because we use zDNNx to handle this?
 
 Yes, zdnnx checks the dimension sizes and splits them to make it fit into NNPA. Elementwise ops are supported also. It's great if we support Conv/Pool too.

AlexandreEichenberger:
We can look into it, handling the overlaps might be tricky in general; maybe easier with dilation and strides of 1. 

Feel free to merge, sunny wants to try it with some ocr benchmarks

PR #3471, opened 2026-04-13
Beefing up runtime for invalid inputs

AlexandreEichenberger:
RIght now, the OM Tensor and Tensor list do not have any protections against functions passing "NULL" as parameters.

This PR fixes these, returning safe values for the getters, and ignoring setters.


Sunny-Anand (approved):
LGTM

src/Runtime/OMTensor.inc
```diff
@@ -301,55 +307,81 @@ void omTensorSetDataPtr(
...
 }
 
 /* OMTensor data strides setter with PyArray stride values */
 void omTensorSetStridesWithPyArrayStrides(
     OMTensor *tensor, const int64_t *stridesInBytes) {
+  if (!tensor || !stridesInBytes)
+    return;
+  int64_t dataTypeSize = getDataTypeSize(tensor->_dataType);
+  if (dataTypeSize == 0) {
+    assert(dataTypeSize != 0);
```
tungld: When is this code called? It always fails when dataTypeSize == 0, right?

AlexandreEichenberger: I believe we build release with assert on; but if it was off, we really want to avoid divide by zero

tungld: `dataTypeSize = 0` means the type is unknown. It is a real issue, we cannot set dataTypeSize = 1 and continue. I suggest to return at this point and print out an error message for debugging.

AlexandreEichenberger: ok, implemented your suggestion

PR #3472, opened 2026-04-13
Update of pyruntime doc to reflect the new interface (no functional change)

AlexandreEichenberger:
There were slight errors in the pyruntime documentations (like using `input` instead of `inputs` for qualified parameter names for `run`.

I also surrounded the `run` commands in a `try` construct to foster better safe practices.

Also, some documentation was stale; it is now updated to reflect the latest code

### DIFF
--- a/docs/UsingPyRuntime.md
+++ b/docs/UsingPyRuntime.md
@@ -159,7 +159,7 @@ def __init__(self, shared_lib_path: str, tag: str = "", use_default_entry_point:
 def run(self, inputs: List[ndarray]) -> List[ndarray]:
     """
     Run inference on the model.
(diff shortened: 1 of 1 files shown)


third_party/onnx
Sunny-Anand, line 1: Why is there a third_party/onnx file update here? Is the pr from the main branch?

Sunny-Anand (approved):
Doc updates look LGTM

third_party/onnx
AlexandreEichenberger: That was an accident; sorry. I removed this change. It now list only one change, the `UsingPyRuntime.md` file

PR #3473, opened 2026-04-14
A lightweight DimAnalysis that can be called from ShapeHelper

tungld:
DimAnalysis is good to find out if two dynamic dimensions are equal or not at compile time. In many cases, if we know two dynamic dimensions are equal during shape inference, we can get better output shape, i.e. more static dimensions in the output shape.

For example, in this MLIR code:
```mlir
func.func @test_reshape_use_dim_analysis(%arg0: tensor<1x?x2048xf32> {onnx.dim_params = "1:s0"}, %arg1: tensor<1x?x2048xf32> {onnx.dim_params = "1:s0"}) -> tensor<*xf32> {
  %0 = onnx.Constant dense<1> : tensor<1xi64>
  %1 = "onnx.Dim"(%arg1) <{axis = 1 : si64}> : (tensor<1x?x2048xf32>) -> tensor<1xi64>
  %2 = onnx.Constant dense<-1> : tensor<1xi64>
  %3 = onnx.Constant dense<128> : tensor<1xi64>
  %4 = "onnx.Concat"(%0, %1, %2, %3) <{axis = 0 : si64}> : (tensor<1xi64>, tensor<1xi64>, tensor<1xi64>, tensor<1xi64>) -> tensor<4xi64>
  %5 = "onnx.Reshape"(%arg0, %4) <{allowzero = 0 : si64}> : (tensor<1x?x2048xf32>, tensor<4xi64>) -> tensor<*xf32>
  return %5 : tensor<*xf32>
}
```

The intention is to reshape `%arg0` from `[1, s0, 2048]` to `[1, s0, 16, 128]`.

However, since the shape input (`%4`) is `[1, s0', -1, 128]`, the current shape inference does not know `s0 = s0'` and produces an output shape of `[1, s0', s1, 128]`.

If we use DimAnalysis to analyze only **several operations** up from the current reshape, we can easily know that `s0 = s0'`, then can infer a good shape of `[1, s0, 16, 128]`.

This PR proposes a lightweight/scoped DimAnalysis to analyze dynamics dimensions of local operations (starting from the current op, up N levels):
```c
DimAnalysis(mlir::Operation *op, int64_t upwardLevel, ONNXOpShapeHelper *shapeHelper = nullptr);
```

Some key points:
- DimAnalysis can be called from ShapeHelper now without infinite recursion (ShapeHelper calls DimAnalysis, DimAnalysis calls ShapeHelper). Passing shapeHelper into DimAnalysis is important for detection of infinite recursion.
- When DimAnalysis calls ShapeHelper, make sure that `ShapeHelper->computeShape` does not change input/output dimensions (e.g. dynamic to static). Otherwise, it confuses the analysis.

Scoped DimAnalysis is used for Reshape's shape inference in this PR, for example:
```c
LogicalResult ONNXReshapeOpShapeHelper::computeShape() {
    DimAnalysis scopedAnalysis(op, /*upwardLevel*/ 3, this);
    scopedAnalysis.analyze();

    scopedAnalysis.sameDynDim(shapeDim, shapeAxis, data, k)
}
```


AlexandreEichenberger:
@tungld 

this seems quite important:

-----
Some key points:

* DimAnalysis can be called from ShapeHelper now without infinite recursion (ShapeHelper calls DimAnalysis, DimAnalysis calls ShapeHelper). Passing shapeHelper into DimAnalysis is important for detection of infinite recursion.
* When DimAnalysis calls ShapeHelper, make sure that ShapeHelper->computeShape does not change input/output dimensions (e.g. dynamic to static). Otherwise, it confuses the analysis.
----

Can you be a more explicit, maybe with an example? Infinite recursion is definitely not something we want. I would also put this warning in the code, maybe with pointers to this warning in places in DimAnalysis and ShapeHelper headers.

I like the example that you use in the code, particularly the name used:
```C++
DimAnalysis scopedAnalysis(op, /*upwardLevel*/ 3, this);
    scopedAnalysis.analyze();
```

I would suggest to make it even more explicit. Do a subclass `ScopedShapeAnalysis` that has only a builder, where you give the maximum recursion, and then make the current 
```C++
DimAnalysis(mlir::Operation *op, uint64_t upwardLevel,
      ONNXOpShapeHelper *shapeHelper = nullptr);
``` 
a protected constructor. So you don't have to change your implementation at all, all is still implemented in the `ShapeAnalysis` class but to access this functionality, we explicitly use this subclass.

src/Dialect/ONNX/ONNXOps/ShapeHelper.cpp
```diff
@@ -179,14 +179,20 @@ void ONNXOpShapeHelper::computeShapeAndAssertOnFailure() {
 void ONNXOpShapeHelper::setOutputDims(
     const DimsExpr &inferredDims, int n, bool refineShape) {
   privateOutputsDims[n] = inferredDims;
-  if (refineShape) {
+  // Donot refine shape in the analysis mode to make sure dynamic dimensions are
```
AlexandreEichenberger, line 182: Did you scan the existing code to make sure that updates to the shape is only made through these 2 calls. Namely that inserting here these `if (!isInDimAnalysisMode()` here is sufficient to fully prevent updates

src/Dialect/ONNX/ONNXDimAnalysis.hpp
```diff
@@ -127,6 +140,8 @@ class DimAnalysis {
   /// This mapping maps each dynamic dimension in the tensor to a set of same
   /// dynamic dimensions.
   DimSetMapT dimSetMap;
+  /// Set of operations for analysis.
```
AlexandreEichenberger, line 143: After reading the code, understand now that this either contains all the ops of the module or the ones included in the "upwardLevel". It would help if you added a quick comment to that effect here.

src/Dialect/ONNX/ONNXDimAnalysis.cpp
```diff
@@ -487,35 +518,164 @@ static bool exploreSameDimsUsingShapeInput(const DimAnalysis::DimT &dim,
...
+      opLevels[defOp] = currentLevel + 1;
+      worklist.push({defOp, currentLevel + 1});
+    }
+  }
+  return collectedOps;
 }
 
-DimAnalysis::DimAnalysis(ModuleOp moduleOp) {
-  moduleOp.walk([&](Operation *op) {
+/// Check if it is safe to do analysis inside the given ShapeHelper.
```
AlexandreEichenberger, line 607: Maybe giving a quick example (e.g invocation path) where it is safe and where it is not safe would help better understand this part. 

I would also suggest to avoid having "not" in the name of a function, because returning false means that it is not, not safe. So that is why when the default shapeHelper value is NULL, it is not not safe, and thus analysis proceeds normally.

Question: are there case where we use the former default constructor (namely the one without the upwardLevel) and that provides a shapeHelper?

AlexandreEichenberger (commented):
The code is useful but hard to understand as is. Made suggestions to clarify a few points.

src/Dialect/ONNX/ONNXOps/ShapeHelper.cpp
tungld: Yes, I asked Bob to help also.

src/Dialect/ONNX/ONNXDimAnalysis.cpp
tungld: > Question: are there case where we use the former default constructor (namely the one without the upwardLevel) and that provides a shapeHelper?

I removed shapeHelper for this constructor since we haven't seen a use case for it.

tungld:
@AlexandreEichenberger I hope I addressed your comments. Really appreciate your suggestions! Thanks!

src/Dialect/ONNX/ONNXDimAnalysis.cpp
```diff
@@ -486,35 +517,186 @@ static bool exploreSameDimsUsingShapeInput(const DimAnalysis::DimT &dim,
...
+      // Detected tensor<*xdtype>. Terminate now.
+      LLVM_DEBUG(llvm::dbgs()
+                 << "Detected tensor<*xdtype>. Terminate DimAnalysis now.\n");
+      return WalkResult::interrupt();
+    }
+    collectedOps.insert(op);
+    return WalkResult::advance();
+  });
+
+  if (walkResult.wasInterrupted())
```
AlexandreEichenberger, line 536: Can you leave a comments explaining why the presence of unranked clear the entire set.  This seems a new requirement (or maybe it was hidden in the old case). Just curious to better understand the issue that we try to avoid here.

src/Dialect/ONNX/ONNXDimAnalysis.cpp
AlexandreEichenberger: Thanks

AlexandreEichenberger (approved):
LGTM, thanks for all the changes, much appreciated. 
Left one new request for clarification.

src/Dialect/ONNX/ONNXDimAnalysis.cpp
tungld: Sure, I added several comments explaining the situation with unranked tensors. Basically, the analysis is not to work with unranked tensors, so we return an empty set of target operations to make the analysis no-op.

PR #3475, opened 2026-04-15
[NNPA, zdnnx] Fix issues related to handling very large tensors

### LINKED ISSUE #3468: Resolve `ZDNN_EXCEEDS_MTS` errors for long context inferences
When querying large decoder models with contexts over 30k in length, we start running into zDNN errors.
```
ERROR: invoke_nnpa() (zdnn.c:339): ZDNN_EXCEEDS_MTS: Total number of elements in tensor exceeds maximum tensor size (MTS).
```
This occurred with a Granite 3.3 Instruct 2B model that is quantized and running with batch size 1. 

There are some notes already in onnx-mlir relating to attempting to not exceeding this limit.
```c
static constexpr int64_t NNPA_MAXIMUM_TENSOR_SIZE = 4294967296;
```

But I'm unsure if these upper bounds were ever fully implemented.

tungld:
This PR fixed several issues related to handling very large tensors on NNPA:
- For single NNPA, rewrite the code that splits dimensions when the total tensor size is larger than the max tensor size on NNPA. Dimension size is decreased by half in the order of dimensions E4, E1, E3, E2 to maximize the buffer reuse.
- For multiple NNPAs, each thread will now call the sequential functions in zdnnx (that handles big tensors) instead of the original zdnn functions. This is because even when we split tensors among NNPA, one tensor may be still big for a single NNPA and need to be splitted again.

Resolves #3468 

In near future, we should add tests for zdnnx to avoid regression.


tungld:

AlexandreEichenberger:
> For multiple NNPAs, each thread will now call the sequential functions in zdnnx (that handles big tensors) instead of the original zdnn functions. This is because even when we split tensors among NNPA, one tensor may be still big for a single NNPA and need to be splitted again.

This is a bit confusing, because in general, if we have N thread for a parallel region, we can still have a loop from `i=0...m*N`, where `m` is selected so that the sizes of the partitioned data fits into a single NNPA call. 

But I think what you are actually saying, is that the tensors sent to a NNPA call can be too large along any dimensions, not just the one that is partitioned by the `i` loop index's corresponding tensor dimension. If that is true, I understand why we want to call the sequential version of the call within a thread.

AlexandreEichenberger (approved):
LGTM

tungld:
> But I think what you are actually saying, is that the tensors sent to a NNPA call can be too large along any dimensions, not just the one that is partitioned by the i loop index's corresponding tensor dimension. If that is true, I understand why we want to call the sequential version of the call within a thread.

Right, for example, we split along only batch size for multiple NNPAs. but other dimensions may be still large for a single NNPA.

PR #3476, opened 2026-04-16
Fix compile options in JsonConfigFile.md

tungld:
Option `--enable-parallel` is incorrect, it must be `--parallel`.

Removed `-mcpu` option in the document since it is not a typical use.


AlexandreEichenberger (approved):
LGTM

tungld:

Issue #2608, opened 2023-11-08
Get onnx-mlir version that was used to compile a .so file by using python


tungld:
Provide a python function to query the onnx-mlir version that was used to compile a .so file.

PR #3482, opened 2026-04-22
Embed compile info into the .so file

tungld:
Embed some compilation information into the generated .so file and the information can be queried by omCompilationInfo() function inside the .so file.

For example, with python:
```python
print(sess.compilation_info())
```
will return
```json
{
"compiler_version": "onnx-mlir version 0.4.2 (43864fda)",
"compile_options": "-O3 -march=z16 -v -o model relu.mlir",
"op_stats": {
  "func.return.3D" : 1,
  "onnx.Add.1D.0D.scalar" : 2,
  "onnx.Add.3D.0D" : 1,
  "onnx.Add.3D.1D" : 1,
  "onnx.Relu.3D" : 1
}
}
```

In `op_stats`, the operation name contains information about rank of each input, for example `onnx.Add.3D.0D` is addition of a 3D tensor (e.g. tensor<?x?x?xf32> and a scalar tensor (tensor<f32>). `.scalar` means that all inputs are scalar.

Add a new compile option `--omit-compile-info` to disable this feature. When disabled, there is no omCompilationInfo() in the .so file.

### DIFF
--- a/src/Runtime/ExecutionSession.cpp
+++ b/src/Runtime/ExecutionSession.cpp
@@ -174,6 +174,19 @@ void ExecutionSession::loadModel(
     throw ExecutionSessionException(
         "Cannot load symbol: '" + outputSignatureNameWithTag + "'.");
 
+  std::string compilationInfoNameWithTag = _compilationInfoName + lowDashTag;
+#if defined(_WIN32)
+  _compilationInfoFunc = reinterpret_cast<compilationInfoFuncType>(
+      _sharedLibraryHandle.getAddressOfSymbol(
+          compilationInfoNameWithTag.c_str()));
+#else
+  _compilationInfoFunc = reinterpret_cast<compilationInfoFuncType>(
+      dlsym(_sharedLibraryHandle, compilationInfoNameWithTag.c_str()));
+#endif
+  if (!_compilationInfoFunc)
+    throw ExecutionSessionException(
+        "Cannot load symbol: '" + compilationInfoNameWithTag + "'.");
+
 #if defined(_WIN32)
   _printInstrumentationFunc = reinterpret_cast<printInstrumentationFuncType>(
       _sharedLibraryHandle.getAddressOfSymbol(
@@ -264,6 +277,14 @@ const std::string ExecutionSession::outputSignature() const {
         "signature function.");
   errno = 0; // No errors.
   return _outputSignatureFunc(_entryPointName.c_str());
+}
+
+const std::string ExecutionSession::compilationInfo() const {
+  if (!isInitialized)
+    throw ExecutionSessionException(
+        "Execution session must be initialized once.");
+  errno = 0; // No errors.
+  return _compilationInfoFunc();
 }
 
 void ExecutionSession::printInstrumentation() {
--- a/src/Runtime/ExecutionSession.cpp
+++ b/src/Runtime/ExecutionSession.cpp
@@ -183,9 +183,6 @@ void ExecutionSession::loadModel(
   _compilationInfoFunc = reinterpret_cast<compilationInfoFuncType>(
       dlsym(_sharedLibraryHandle, compilationInfoNameWithTag.c_str()));
 #endif
-  if (!_compilationInfoFunc)
-    throw ExecutionSessionException(
-        "Cannot load symbol: '" + compilationInfoNameWithTag + "'.");
 
 #if defined(_WIN32)
   _printInstrumentationFunc = reinterpret_cast<printInstrumentationFuncType>(
@@ -284,6 +281,8 @@ const std::string ExecutionSession::compilationInfo() const {
     throw ExecutionSessionException(
         "Execution session must be initialized once.");
   errno = 0; // No errors.
+  if (_compilationInfoFunc == nullptr)
+    return "{}";
   return _compilationInfoFunc();
 }
 
--- a/src/Runtime/ExecutionSession.cpp
+++ b/src/Runtime/ExecutionSession.cpp
@@ -281,7 +281,7 @@ const std::string ExecutionSession::compilationInfo() const {
     throw ExecutionSessionException(
         "Execution session must be initialized once.");
   errno = 0; // No errors.
-  if (_compilationInfoFunc == nullptr)
+  if (!_compilationInfoFunc)
     return "{}";
   return _compilationInfoFunc();
 }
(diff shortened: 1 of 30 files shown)


src/Runtime/ExecutionSession.cpp
```diff
@@ -174,6 +174,19 @@ void ExecutionSession::loadModel(
...
+  std::string compilationInfoNameWithTag = _compilationInfoName + lowDashTag;
+#if defined(_WIN32)
+  _compilationInfoFunc = reinterpret_cast<compilationInfoFuncType>(
+      _sharedLibraryHandle.getAddressOfSymbol(
+          compilationInfoNameWithTag.c_str()));
+#else
+  _compilationInfoFunc = reinterpret_cast<compilationInfoFuncType>(
+      dlsym(_sharedLibraryHandle, compilationInfoNameWithTag.c_str()));
+#endif
+  if (!_compilationInfoFunc)
```
AlexandreEichenberger: I ran into some trouble with the `printInstrumentationFunc` before, when we tried to run the new runtime with an older .so.

I suggest that you mimic the behavior by disabling the functionality if the function is not found.... Maybe that logic could become a function to reduce redundancy of code.

The alternative would be to
1. start to have a version number so that if we have a new RT with old .so, we silently assert; if we have a new RT with a new so, then we assert because something went wrong.
2. we only support new RT with new .so (no compatibility). If so, then we should also remove the support for the print instrumentation.

@chentong319 @Sunny-Anand any suggestions?

AlexandreEichenberger (commented):
LGTM, think we should use this PR to have a short discussion about compatibility of compiled models (.so) and runtime @tungld @Sunny-Anand @chentong319 

Other thought, questions, request for feedback:
* I would suggest to add a "compiler version" string. Might be useful for debugging. That can in turn be used for compatibility (above discussion).
* Does the option string inline the options gotten from the .json config file, if provided? That might be useful as otherwise we may not really know the actual options that were used. If using the default, we may not even know any of the options used otherwise.
* Does it gather the ONNX ops (before/after decompose, recompose), or with NNPA operations? When profiling ops, we care about these. Carefully defining the purpose of this additional info would help us determine what info is needed here.
* Might some customers not want to "reveal" this info if they distribute a .so? Possibility of disabling it? Here I probably refer more to the operation mix. @Sunny-Anand ?

AlexandreEichenberger:
Also, when there are dynamic dimensions, we may have some shape computations that are done using onnx ops. For example, transforming Conv to im2Col may generate such code
```mlir
    %14 = "onnx.Add"(%8, %10) : (tensor<1xi64>, tensor<1xi64>) -> tensor<1xi64>
    %15 = "onnx.Div"(%14, %12) : (tensor<1xi64>, tensor<1xi64>) -> tensor<1xi64>
    %16 = "onnx.Add"(%9, %11) : (tensor<1xi64>, tensor<1xi64>) -> tensor<1xi64>
    %17 = "onnx.Div"(%16, %13) : (tensor<1xi64>, tensor<1xi64>) -> tensor<1xi64>
    %18 = "onnx.Concat"(%6, %7, %15, %17) <{axis = 0 : si64}> : (tensor<1xi64>, tensor<1xi64>, tensor<1xi64>, tensor<1xi64>) -> tensor<4xi64>
    %19 = "onnx.Reshape"(%5, %18) <{allowzero = 0 : si64}> : (tensor<?x64xf32>, tensor<4xi64>) -> tensor<1x64x9x9xf32>
```

So I really believe we should exclude from the operation reporting any operations on scalar...

Sunny-Anand:
> * I would suggest to add a "compiler version" string. Might be useful for debugging. That can in turn be used for compatibility (above discussion).

This was my first thought to atleast embed the compiler  version string, along with the commit id to make it easier to know which commit the user is using.
> * Does the option string inline the options gotten from the .json config file, if provided? That might be useful as otherwise we may not really know the actual options that were used. If using the default, we may not even know any of the options used otherwise.

Can this work with a mix and match of inline options and config.json - will it combine both of them together?

> * Might some customers not want to "reveal" this info if they distribute a .so? Possibility of disabling it? Here I probably refer more to the operation mix.

That's a good catch, though I would imagine that this would already be public information apart from the config, like compiler version/commit. However the compiler options and the values for them are pretty standard for all the platforms like for EmitLib, -03. I can think of -march and --mtriple to be specific to the platform, we can probably hide them or redact them from the info, or some optimization options only for NNPA.

chentong319:
I second some comments:
1. Include compiler version if there is one
2. Handle old .so
3. Do not include op stats

One more suggestion: get rid of any path info in the compilation flags.

tungld:
Thank you @AlexandreEichenberger @Sunny-Anand @chentong319 for your comments!

> I would suggest to add a "compiler version" string. Might be useful for debugging. That can in turn be used for compatibility (above discussion).

We already have this info in the .so file (embeded in to the llvm module flags). So the question is should we include it here also? or use one of the two ways only?. See #2608

> Does the option string inline the options gotten from the .json config file, if provided? That might be useful as otherwise we may not really know the actual options that were used. If using the default, we may not even know any of the options used otherwise.

Yes, it includes all options from the command line and and .json config file. I will remove path-related strings as @chentong319's suggestion.

> Does it gather the ONNX ops (before/after decompose, recompose), or with NNPA operations? When profiling ops, we care about these. Carefully defining the purpose of this additional info would help us determine what info is needed here.

The collection was called from the same place where @AlexandreEichenberger  implemented `--onnx-op-stats`, actually it called the pass `--onnx-op-stats` proposed by Alex. The printed information is the same as the output of `--onnx-op-stats` and using json format.

src/Runtime/ExecutionSession.cpp
tungld: @AlexandreEichenberger 
> I suggest that you mimic the behavior by disabling the functionality if the function is not found.... Maybe that logic could become a function to reduce redundancy of code.

I edited the code to return an empty json string, i.e. "{}" if the function not found. What do you think?

tungld:
> Might some customers not want to "reveal" this info if they distribute a .so? Possibility of disabling it? Here I probably refer more to the operation mix. @Sunny-Anand ?

@AlexandreEichenberger I added a compile option `--do-not-embed-compilation-info` to disable this feature.

src/Runtime/ExecutionSession.cpp
AlexandreEichenberger: very good; and just describe behavior in the online python documentation.

AlexandreEichenberger:
nit `--do-not-embed-compilation-info` is long; `-omit-compile-info`

AlexandreEichenberger (approved):
LGTM, thanks for responding to all the comments.

Sunny-Anand (approved):

chentong319 (approved):
LGTM!

tungld:
> nit --do-not-embed-compilation-info is long; -omit-compile-info

Yes, done.

PR #3484, opened 2026-04-22
(WIP) Enable ConvOptPass for accelerator

chentong319:
In ORC model family, there are many ConvOp can be optimized by ConvOptPass. I do not see any reason that we do not want ConvOptPass for accelerator. There is a big performance impact if those ConvOps are not offloaded to accelerator.


AlexandreEichenberger:
@chentong319 ok to close this one in light of #3499 and #3512

chentong319:
No need with the new Conv optimization.

PR #3488, opened 2026-04-27
fix cve's for pillow, update torch and torchvision

Sunny-Anand:
https://github.com/onnx/onnx-mlir/security/dependabot/22


AlexandreEichenberger (approved):
LGTM, we could drop the torch example if we think its not bringing much.

PR #3491, opened 2026-04-28
Create zizmor.yml

andife:
## What this PR does

This PR introduces [zizmor](https://woodruff.dev/zizmor/), a static analysis tool for GitHub Actions workflows, and fixes the supply chain security issues it identifies across all existing workflow files.

### New workflow: `zizmor.yml`

Adds a CI job that runs zizmor on every push to `main` and on every pull request. zizmor audits `.github/workflows/` for common security misconfigurations including:

- **Unpinned actions** — version tags like `@v4` are mutable; they can be silently updated or hijacked. zizmor requires every `uses:` reference to be pinned to an immutable SHA hash.
- **Excessive permissions** — workflows that request write permissions they do not need.
- **Script injection risks** — untrusted user input flowing into `run:` steps without sanitization.
- **Credential leaks** — credentials inadvertently persisted across steps.

Results are uploaded as SARIF to GitHub's Security tab so findings are surfaced directly alongside code review.

### Why SHA pinning?

When a workflow references `actions/checkout@v4`, GitHub resolves the `v4` tag at runtime. A compromised or mistakenly updated tag silently changes the code that executes in your CI environment — with access to secrets and the ability to modify build outputs. Pinning to a full 40-character commit SHA makes that reference immutable: the exact same code runs on every invocation, and any change to the referenced action is immediately visible in the diff.

The comment (`# v4`, `# stable`, etc.) preserved alongside each SHA makes it easy for humans to understand what version is pinned and for tools like [Dependabot](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot) or [zizmor](https://woodruff.dev/zizmor/) to suggest upgrades.

### Files changed

| File | Actions pinned |
|---|---|
| `zizmor.yml` (new) | `actions/checkout`, `zizmorcore/zizmor-action` |
| `black-format-check.yml` | `actions/checkout@v4`, `psf/black@stable` |
| `clang-format-check.yml` | `actions/checkout@v2`, `DoozyX/clang-format-lint-action@v0.5` |
| `doxygen-docs-publish.yml` | `actions/checkout@v2` (×2) |
| `macos-amd64-build.yml` | `actions/checkout@v4`, `actions/setup-python@v4`, `actions/cache/restore@v4`, `actions/cache/save@v4` |

No action versions were updated — only the reference format changed from mutable tag to immutable SHA.


AlexandreEichenberger:
@andife it passed the CIs, mind just filling the description to give a bit of feedback on what the new code is, tx

andife:
@AlexandreEichenberger I added explanations. In order to increase the supply chain security I suggest updating the github actions, with dependabot and use sha instead of version tags? I'm not familiar with the process here. Should Jenkins be working? And what about Azure Pipelines?

PR #3492, opened 2026-04-29
PYBIND signature fix

AlexandreEichenberger:
For some reason, the PYBIND signature is given for the constructors, but the current definition does not match the actual C++ signatures. This PR fixes them.

It is currently working without the fix, but I don't see why we should not have an exact match.


Sunny-Anand (approved):
LGTM

gongsu832:
Instead of removing the `const` from `py::init`, I would think adding `const` to all the C++ functions would be a more prudent choice, to let compiler catch accidental changes to the parameters passed in.

AlexandreEichenberger:
As suggested by @Sunny-Anand , to simplify the structure of the PyExecutionSession, I merged PyExecutionSessionBase code into PyExecutionSession.

As suggested by @gongsu832 I added the `const` in the PYBIND interface as well as the underlying C++ class implementations.

AlexandreEichenberger:
I am having difficulties with includes for the PyExecutionSession due to inclusion of LLVM and MLIR headers... #3494 fixes these issues by doing the right thing, namely removing such dependences.

Thus closing this PR

PR #3494, opened 2026-05-01
PY Bind signatures, and simplification of PyExecutionSession

AlexandreEichenberger:
This PR subsume #3492 in that it incorporates all the fixes there, and also eliminate the LLVM dependence for the handling of float16, which were creating issues on the z-linux build.

I incorporated @gongsu832 recommendations to use `const` in the signatures wherever possible, harmonized the parameter passing between the C++ implementations of ExecutionSession, and merged the PyExecutionSessionBase into the PyExecutionSession.


src/Runtime/python/CMakeLists.txt
```diff
@@ -10,46 +10,47 @@ file(GENERATE
...
+# 
+# if(MSVC)
+#   target_link_libraries(OMPyExecutionSessionBase
+#     PRIVATE pybind11::windows_extras
+#   )
+# endif()
+# set_target_properties(OMPyExecutionSessionBase
+#   PROPERTIES
+#   POSITION_INDEPENDENT_CODE TRUE
+#   )
```
gongsu832: nit: if you want to save some typing, you can use `if (0) ... endif()`.

src/Runtime/python/PyOMCompile.hpp
```diff
@@ -57,8 +57,8 @@ PYBIND11_MODULE(PyOMCompileC, m) {
       "    >>> compiler = OMCompile('model.onnx', '-O3 -o output')\n"
       "    >>> output_file = compiler.get_output_file_name()\n"
       "    >>> print(f'Compiled to: {output_file}')")
-      .def(py::init<const std::string &, const std::string &, const std::string &,
-               const std::string &, const bool>(),
+      .def(py::init<std::string, std::string, const std::string &,
+               const std::string &, bool>(),
```
gongsu832, line 61: Add `const` to `PyOMCompile::PyOMCompile` in `src/Runtime/python/PyOMCompile.cpp` instead?

src/Runtime/python/PyExecutionSession.cpp
```diff
@@ -14,12 +14,503 @@
...
+// better way for fixing it is found.
+//
+static void *generateOMTensorBufferForStringData(py::array pyArray) {
+  // Compute the total number of element in the pyArray, e.g. shape (3, 4) has
+  // 12 elements. While we handle here multi-dimensional arrays, we actually
+  // requires them to be flatten (aka 1 Dim) below.
+  auto shape = pyArray.shape();
+  uint64_t numElem = 1;
+  for (size_t i = 0; i < (size_t)pyArray.ndim(); ++i)
+    numElem *= shape[i];
```
gongsu832: I thought there is already `PyArray.size()`?

As a general comment, I wonder if you really must flatten and copy the array of strings. Once you convert to C++ vector<string>, you can get the [ u64 pointer, u64 size ] of each string. Can't you then just create an "auxiliary" tensor of shape n x 2 (n is the number of strings) and pass that to the `omTensorCreateWithOwnership`. Later when you process the tensor in the model runtime and see the dtype `ONNX_TYPE_STRING`, you then actually chase down the pointers to get the string data.

gongsu832: In fact, you probably don't even need the u64 size since you have the terminating `\0` in the strings.

src/Runtime/python/CMakeLists.txt
AlexandreEichenberger: Thanks for the tip. I intend to remove this once the z-folks report success in the integration issue that they appear to have.

src/Runtime/python/CMakeLists.txt
```diff
@@ -78,16 +79,24 @@ set_target_properties(OMPyExecutionSessionBase
...
   )
+
+# Link OMSmallFPConversion for float16 support.
+# This library contains only C code (SmallFPConversion.c) and has no LLVM dependencies,
+# so float16 support is now available in both full and light modes.
 target_link_libraries(PyRuntimeC
   PRIVATE
-  OMPyExecutionSessionBase
+  OMExecutionSession
+  OMSmallFPConversion
```
AlexandreEichenberger, line 99: @chentong319 is that dependence ok?

src/Runtime/python/PyOMCompile.hpp
AlexandreEichenberger: good suggestion, will implement

AlexandreEichenberger:
Recommendation was to not have redundant constructor parameter default values in the C++ Py*.hpp  and the PYBIND definition, to avoid inconsistencies. So its now only in the PYBIND interface.

src/Runtime/python/PyExecutionSession.cpp
AlexandreEichenberger: Thanks, will update with a simplified code using parry.size(). We need to keep it as one tensor because of the way our OMTensors works, namely one monolithic block of data. We would have specific handling all over for it. I am told that python strings cannot guarantee a null terminated string. Thanks for your suggestions.

src/Runtime/python/CMakeLists.txt
```diff
@@ -10,6 +10,8 @@ file(GENERATE
   INPUT ${CMAKE_CURRENT_SOURCE_DIR}/PyOMCompile.py
   )
 
+# TODO: Remove after we can build all systems using the new approach.
```
Sunny-Anand, line 13: nit: should this be removed now?

Sunny-Anand (approved):
LGTM.

src/Runtime/python/CMakeLists.txt
AlexandreEichenberger: because our CIs don't test yet for the lightweight version, will remove once we have it working for a bit.

PR #3495, opened 2026-05-01
Helper script updated (fixLitTests and RunONNXModel)

AlexandreEichenberger:
RunONNXModel had a small issue when using the `--cache` option, which is fixed. Additional functionality, if creating a new cached version of a compilation, a copy of the compile log will also be copied into the cached version if a log was requested in the first place (aka `--write-compile-log=<file>` was given).

Second, some of our lit tests have multiple `// RUN` with different `FileCheck --check-prefix` options. Our fixLitTest reported error on files with more than one `// RUN` command. This PR add functionality to be able to mention a `--prefix` option. When doing so, only the command with the matching `--check-prefix` name will be used. And if "repairs" is requested, only code with this prefix will be repaired, for functions that fail the checks. This new functionality required a bit more surgery than expected, results worked for the test cases that I was trying to fix.


tungld (approved):
LGTM.

PR #3496, opened 2026-05-04
Rename the package OMPyInfer to om_pyrt

chentong319:
The plan is to add the script of compilation into this package. The script to use compiler installed locally and in a container will be included in this package. However, the standalone compiler will remain in a separate package, an optional dependent package for this one. Therefore, a more general name is used.
This name is also similar to zdlc_pyrt.


AlexandreEichenberger (approved):
LGTM

Sunny-Anand (approved):
LGTM

Issue #3493, opened 2026-04-30
NNPA: constraints for MaxPool is not properly checked


chentong319:
In model (zdlc_modelzoo_onnx/exploiters/watsonx/iocr/detection/word_unet_v2), there are four MaxPool ops, whose input tensor size is reduced at each step. Even with constant shape provided, onnx-mlir failed to detect that the first two ops are too large for zdnn lib. I got runtime error of 'ZDNN_FUNC_RC_F004`. According to document, 
```
ZDNN_FUNC_RC_F004 - stride_height > 0 and stride_width > 0, but either input tensor's height or weight dimension is greater than 1024.
```

Here is the example of the op:
```
%73 = "onnx.MaxPoolSingleOut"(%72) <{auto_pad = "NOTSET", ceil_mode = 0 : si64, kernel_shape = [2, 2], storage_order = 0 : si64, strides = [2, 2]}> {onnx_node_name = "StatefulPartitionedCall/model/max_pooling2d_1/MaxPool"} : (tensor<1x32x1216x832xf32>) -> tensor<1x32x608x416xf32>
```
The incorrectly generated op:
```
%70 = "zhigh.MaxPool2D"(%69) <{kernel_shape = [2, 2], padding_type = "VALID_PADDING", strides = [2, 2]}> : (tensor<1x1216x832x32xf16, #zhigh.layout<{dataLayout = "NHWC"}>>) -> tensor<1x608x416x32xf16, #zhigh.layout<{dataLayout = "NHWC"}>>
```
I did some investigation. To my understanding, the problem is with the function meetPoolParamRestrictions(). The `inputShape > 1024` is only checked in the branch of `stride==0`. Should it be in the else branch? @tungld Can you confirm the issue?
A by-product discovery is that the type of parameters are different in onnx-mlir and zdnn. zdnn document specified the time to be uint32_t, and onnx-mlir provide constant of int64.

By the way, was the bug of checking also in zdnnx for maxpool?

JOSH1024:
@chentong319 is the issue solved ?

chentong319:
Yes, PR3497 fixed it.

PR #3497, opened 2026-05-05
NNPA: Add the missed check for pool ops

chentong319:
Refer to #3493 .


AlexandreEichenberger (approved):
LGTM, good catch @chentong319

PR #3498, opened 2026-05-05
Fix a bug in bound check for input

chentong319:
Should use `ShapedType::kDynamic` instead of -1.


AlexandreEichenberger (approved):
LGTM, have scanned the code at times for -1, maybe there are other cases.

src/Conversion/KrnlToLLVM/KrnlEntryPoint.cpp
```diff
@@ -584,8 +584,10 @@ class KrnlEntryPointOpLowering : public OpRewritePattern<KrnlEntryPointOp> {
         // Check each dimension
         for (uint64_t dimID = 0; dimID < pair.second.size(); ++dimID) {
           int64_t bound = pair.second[dimID];
-          if (bound == -1) // No bound info for this dimension
+          if (bound == ShapedType::kDynamic) {
```
tungld, line 587: Looks like this bound is from the option `shapeInformationUB/LB` and by its definition, its value is a positive integer or -1. So `-1` is correct, isn't it?

chentong319: When I printed the value out, it is not -1. I guess that our parser replaced -1 with the ShapedType::kDynamic.

PR #3499, opened 2026-05-05
Convolution optimizations for 1x1 (moved) and 2D conv (using im2col)

AlexandreEichenberger:
## Overview.

This PR targets optimizations for Convolutions

At a high level, the Convolutions are transformed (when suitable) to matrix multiplications. When an accelerator is used (NNPA), these matrix multiplications can then be executed on the accelerator.

First, the 1x1 conv had its own separate pass; it is now part of normal decompose. There were minor issues that were fixed. This optimization is unconditionally applied as it incurs no overhead (just reshapes) and matmul is generally always better implemented than convolutions. [There might be corner cases as reshape may not be free in NNPA stickified format, to be explored.]

Second, optimizations for 2D conv. There we have 2 distinct optimizations path, which can both be disabled with `-disable-conv-to-matmul`.

image

## First path (CPU).

CPU. For the cpu path, all 2D convolutions with Group==1 (also require weights to be constant, which is always the case) are converted to a im2col followed by a matmul. Basically, the im2col is a new additional ONNX operation that linearize all the CI  * Kh * Kw values needed to compute one output point. That is one row. And there is one row per output point. 

This decomposition is not applied to the first pass of decompose (when Hybrid is default on), because there is a small recompose optimization that combine multiple compatible convolutions into one. But this decomposition is then applied for all successive calls of Hybrid pass, which include decomposition.

## Second path (Accelerator).  

We assume that for accelerator, any convolutions that can be natively executed by the NNPA accelerator will be better of as convolutions (as converting it to matmul requires a CPU bound im2col operation). Thus, decomposition of conv to matmul is only applied for these convolutions that fail to be mapped to the accelerator.

## Implementation of im2col

There is naive N-dim implementation (currently only use 2d but the code is generic). When compiling at more than `-O0`, we have a more efficient implementation that collapse multiple outer loops (N, Oh, Ow)... namely one loop for all output pixel for one Cout. Then we check if the work may possibly includes padding; when it does not, we have quick inner loops that compute without further checking. When we may have padding, the inner loops check for each data point if we have padding or not. When padding (aka a filter point that is outside the original image boundary), we initialize with zero. Otherwise, we use the image point.

I may explore further tuning for the code gen for im2col in a subsequent PR. 

## Other unrelated changes.

* Added the possibility to generate a mlir files for the ModelLib (added `debug` option to the `compile` method) and running the model in a safe env (catching segfaults, using a `debug` in the `run` method).
* There was bad CHECK code generating by the `mlir2FileCheck` for some `VAR` patterns, which is now fixed. That python code is used in the `fixLitTest.py`.
* There was also a read after dealloc found by Valgrind in the `utils/RunONNXLib.cpp‎` code, which I fixed.


AlexandreEichenberger:
Experiment with one conv based benchmark (ckpt_iter_number_400.onnx). 
* New code: all 23 `onnx.conv` sent to 23 `onnx.Im2Col + zhigh.MatMul`. 
* Old code, all 23 `onnx.conv` implemented on CPU.

Details below. 

With new code
```
   func.func             , 1
   func.return           , 1
   onnx.Add              , 39
   onnx.Cast             , 16
   onnx.Concat           , 31
   onnx.Constant         , 56
   onnx.Dim              , 19
   onnx.Div              , 8
   onnx.Im2Col           , 22
   onnx.LayoutTransform  , 3
   onnx.Max              , 16
   onnx.MaxPoolSingleOut , 4
   onnx.Mul              , 8
   onnx.Neg              , 8
   onnx.NoValue          , 13
   onnx.Pad              , 4
   onnx.Relu             , 17
   onnx.Reshape          , 25
   onnx.Slice            , 4
   onnx.Split            , 4
   onnx.Squeeze          , 8
   onnx.Sub              , 8
   onnx.Transpose        , 1
   onnx.Unsqueeze        , 9
   onnx.UpsampleAndPad   , 4
  zhigh.DLF16ToF32       , 1
  zhigh.MatMul           , 23
  zhigh.Relu             , 1
  zhigh.Sigmoid          , 1
  zhigh.Stick            , 47
  zhigh.Unstick          , 23
```

Without new code (aka `--disable-conv-to-matmul`)
```
   func.func             , 1
   func.return           , 1
   onnx.Add              , 8
   onnx.Cast             , 16
   onnx.Concat           , 20
   onnx.Constant         , 52
   onnx.Conv             , 23
   onnx.Dim              , 16
   onnx.Div              , 8
   onnx.LayoutTransform  , 1
   onnx.Max              , 16
   onnx.MaxPoolSingleOut , 4
   onnx.Mul              , 8
   onnx.Neg              , 8
   onnx.NoValue          , 12
   onnx.Pad              , 4
   onnx.Relu             , 18
   onnx.Reshape          , 1
   onnx.Slice            , 4
   onnx.Split            , 4
   onnx.Squeeze          , 8
   onnx.Sub              , 8
   onnx.Transpose        , 1
   onnx.Unsqueeze        , 9
   onnx.UpsampleAndPad   , 4
  zhigh.DLF16ToF32       , 1
  zhigh.Sigmoid          , 1
  zhigh.Stick            , 1
```

docs/optimization-onnx-lowering/ConvIm2Col.md
```diff
@@ -0,0 +1,305 @@
+<!--- SPDX-License-Identifier: Apache-2.0 -->
+# Conv Lowering with Im2Col and MatMul
+
+**Status:** Enabled for supported [`Conv`](../onnx-mlir/src/Dialect/ONNX/ONNXOps/NN/Conv.cpp) configurations.
```
tungld: This path looks incorrect. Same in the following paragraphs.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
```diff
@@ -992,11 +999,33 @@ void getRewriteONNXForZHighDynamicallyLegal(
...
+          bool canDecompose1x1ToAMatmul = shouldDecomposeConv1x1ToMatmul(op);
+          // Rule to change to conv -> im2Col -> matmul.
+          // NNPA has fast 1xN broadcast, for M15+.
+          bool canDecomposeToIm2ColAndMatmul = shouldDecomposeConvToIm2Col(
+              op, /*hasFastBroadcast1xN=*/isCompatibleWithNNPALevel(M15));
+          bool canApplyRule = canDecomposeToPadAndConv ||
+                              canDecompose1x1ToAMatmul ||
+                              canDecomposeToIm2ColAndMatmul;
+          bool legal = suitableForZDNN || !canApplyRule;
+#if 0
```
tungld: Please push this part into the Debug mode, or remove it.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
```diff
@@ -1016,6 +1046,18 @@ struct RewriteONNXForZHighPass
...
   RewriteONNXForZHighPass() = default;
+  RewriteONNXForZHighPass(bool enableConvToMatmul) {
+    this->enableConvToMatmul = enableConvToMatmul;
+  }
+  RewriteONNXForZHighPass(const RewriteONNXForZHighPass &pass)
+      : mlir::PassWrapper<RewriteONNXForZHighPass, OperationPass<ModuleOp>>() {
+    this->enableConvToMatmul = pass.enableConvToMatmul.getValue();
+  }
+
+  Option<bool> enableConvToMatmul{*this, "enable-conv-to-matmul",
```
tungld: If possible, use `nnpa` prefix, e.g. `--nnpa-enable-conv-to-matmul`, to avoid conflict with other CPU or accelerator options.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
```diff
@@ -1016,6 +1046,18 @@ struct RewriteONNXForZHighPass
   }
 
   RewriteONNXForZHighPass() = default;
+  RewriteONNXForZHighPass(bool enableConvToMatmul) {
+    this->enableConvToMatmul = enableConvToMatmul;
+  }
+  RewriteONNXForZHighPass(const RewriteONNXForZHighPass &pass)
```
tungld, line 1056: Do we need this constructor?

src/Dialect/ONNX/ONNXOps/Additional/Im2Col.cpp
```diff
@@ -0,0 +1,271 @@
...
+LogicalResult ONNXIm2ColOp::inferShapes(
+    std::function<void(Region &)> doShapeInference) {
+  // Get input type.
+  if (!hasShapeAndRank(getX()))
+    return success();
+
+  // Create shape helper and compute shape.
+  ONNXIm2ColOpShapeHelper shapeHelper(getOperation(), {});
+  return shapeHelper.computeShapeAndUpdateType(
+      mlir::cast<ShapedType>(getX().getType()).getElementType());
```
tungld: Do not use the element type from input but output instead or do not update element type.

src/Dialect/ONNX/Transforms/ConvOpt.cpp
```diff
@@ -32,194 +32,22 @@
...
-  }
-  if (verbose)
-    printf("optimize conv 1x1 with matmul: N %d, group %d, Ci %d, Co %d, H %d, "
-           "W %d, rank %d, with%s bias\n",
-        (int)xShape[0], (int)G, (int)xShape[1], (int)wShape[0], (int)xShape[2],
-        (int)xShape[3], (int)rank, hasBias ? "" : "out");
-  return true;
-}
-
+// Conv 1x1 to MatMul optimization has been moved to Decompose.cpp.
```
tungld: Remove this comment. It's hard to understand after this PR merged.

src/Dialect/ONNX/Transforms/ConvOpt.cpp
```diff
@@ -32,194 +32,22 @@
...
-};
-
 } // namespace
 
 void onnx_mlir::getConvOptONNXToONNXPatterns(
     bool enableSimdDataLayoutOpt, RewritePatternSet &patterns) {
-  // TODO: if enable simd layout opt, we still need to determine how 1x1 and
-  // simd layout interact. Right now, only enable the one or the other. Will
-  // need to refine this later.
+  // Conv 1x1 to MatMul optimization is now handled in Decompose pass.
```
tungld: Remove this comment. It's hard to understand after this PR merged.

src/Dialect/ONNX/Transforms/Decompose.cpp
```diff
@@ -608,6 +608,100 @@ ElementsAttr reshapeElementsAttrToRank0WithDefaultValue(
...
 
+//===----------------------------------------------------------------------===//
+// Exported functions for Conv decomposition
+//===----------------------------------------------------------------------===//
+
+// Check if all values in optional array attribute equal expectedValue.
+static bool allArrayAttrValuesEqual(
+    std::optional<ArrayAttr> attr, int count, int64_t expectedValue) {
+  if (!attr.has_value())
+    return true; // No attribute means default behavior (typically 1).
```
tungld: This function is quite general so it's hard to say default behavior means `1`. It is natural to return false here.

src/Dialect/ONNX/Transforms/Decompose.cpp
```diff
@@ -608,6 +608,100 @@ ElementsAttr reshapeElementsAttrToRank0WithDefaultValue(
...
+  Value W = convOp.getW();
+  Value B = convOp.getB();
+  bool hasBias = !isNoneValue(B);
+  if (!hasShapeAndRank(X) || !hasShapeAndRank(W))
+    return false;
+  if (hasBias && !hasShapeAndRank(B))
+    return false;
+  const auto xType = mlir::cast<ShapedType>(X.getType());
+  const auto wType = mlir::cast<ShapedType>(W.getType());
+  const auto xShape = xType.getShape();
```
tungld: Avoid auto by using `ArrayRef<int64_t>`.

src/Dialect/ONNX/Transforms/Decompose.cpp
```diff
@@ -608,6 +608,100 @@ ElementsAttr reshapeElementsAttrToRank0WithDefaultValue(
...
+
+  // 2. Weight tensor must have static shape (we need to know kernel dims).
+  if (!onnx_mlir::hasStaticShape(W.getType()))
+    return false;
+
+  // 3. Must be 2D convolution (rank = 4: N x C x H x W).
+  // For now, only support 2D convolutions.
+  // Future: extend to 1D and 3D convolutions.
+  ShapedType xType = mlir::cast<ShapedType>(X.getType());
+  auto xShape = xType.getShape();
```
tungld: Avoid auto by using ArrayRef<int64_t>.

src/Dialect/ONNX/Transforms/Decompose.cpp
```diff
@@ -1134,16 +1228,385 @@ struct DecomposeHardSwishPattern : public OpRewritePattern<ONNXHardSwishOp> {
...
+    // Create ONNX builder for cleaner code.
+    onnx_mlir::MultiDialectBuilder<onnx_mlir::OnnxBuilder> create(
+        rewriter, loc);
+
+    // Get element types.
+    ShapedType xType = mlir::cast<ShapedType>(X.getType());
+    ShapedType wType = mlir::cast<ShapedType>(W.getType());
+
+    auto xShape = xType.getShape();
+    auto wShape = wType.getShape();
```
tungld: Avoid auto by using ArrayRef<int64_t>.

src/Dialect/ONNX/Transforms/Decompose.cpp
```diff
@@ -1134,16 +1228,385 @@ struct DecomposeHardSwishPattern : public OpRewritePattern<ONNXHardSwishOp> {
...
+      ShapedType bType = mlir::cast<ShapedType>(B.getType());
+      Type bReshaped = RankedTensorType::get({CO, 1}, bType.getElementType());
+      Value B_reshaped = create.onnx.unsqueeze(bReshaped, B, axes);
+      // Bias shape [CO, 1] broadcasts to [N, CO, OH*OW].
+      Y_with_bias = create.onnx.add(Y_flat, B_reshaped);
+    }
+
+    // Step 6: Reshape back to [N, CO, OH, OW].
+    // Compute output dimensions OH and OW from input dimensions and Conv
+    // parameters.
```
tungld, line 1360: Can't we use ShapeHelper to compute output dims? I wonder why we need to recompute them here.

src/Dialect/ONNX/Transforms/Decompose.cpp
```diff
@@ -1134,16 +1228,385 @@ struct DecomposeHardSwishPattern : public OpRewritePattern<ONNXHardSwishOp> {
...
+    }
+
+    // Step 5: Reshape from [N, CO, OH*OW] to [N, CO, OH, OW].
+    // Concatenate dimensions to form output shape: [N, CO, OH, OW].
+    Type shapeType = RankedTensorType::get({4}, rewriter.getI64Type());
+    Value outputShapeVals =
+        create.onnx.concat(shapeType, {N, COVal, OH, OW}, 0);
+
+    // Create reshape, using native builder when unranked (avoid shape infer).
+    Value Y;
```
tungld, line 1463: To avoid calling native builder, we can do:

```c
Value Y = create.onnx.reshape(convOutType, Y_with_bias, outputShapeVals);
Y.setType(convOutType);
```

test/mlir/accelerators/nnpa/conversion/rewrite-onnx-for-zhigh.mlir
```diff
@@ -587,42 +595,71 @@ func.func @softmax_nd_to_2d(%arg0: tensor<4x12x256x256xf32>) -> (tensor<4x12x256
...
+// CHECKZ17:           [[VAR_4_:%.+]] = "onnx.Pad"([[PARAM_0_]], [[VAR_1_]], [[VAR_2_]], [[VAR_3_]]) <{mode = "constant"}> : (tensor<5x3x32x32xf32>, tensor<8xi64>, tensor<f32>, none) -> tensor<5x3x34x34xf32>
+// CHECKZ17:           [[VAR_5_:%.+]] = "onnx.Conv"([[VAR_4_]], [[PARAM_1_]], [[VAR_0_]]) <{auto_pad = "VALID", group = 1 : si64, kernel_shape = [2, 2], pads = [0, 0, 0, 0]}> : (tensor<5x3x34x34xf32>, tensor<1024x3x2x2xf32>, none) -> tensor<5x1024x33x33xf32>
+// CHECKZ17:           return [[VAR_5_]] : tensor<5x1024x33x33xf32>
+// CHECKZ17:         }
 // -----
 
+
+
+
+
```
tungld, line 630: Too many empty lines :)

test/mlir/accelerators/nnpa/conversion/rewrite-onnx-for-zhigh.mlir
```diff
@@ -587,42 +595,71 @@ func.func @softmax_nd_to_2d(%arg0: tensor<4x12x256x256xf32>) -> (tensor<4x12x256
...
+// CHECKZ17-LABEL:  func.func @test_onnx_conv2d_with_bias_and_different_pads
+// CHECKZ17-SAME:   ([[PARAM_0_:%.+]]: tensor<1x3x224x224xf32>, [[PARAM_1_:%.+]]: tensor<64x3x7x7xf32>, [[PARAM_2_:%.+]]: tensor<64xf32>) -> tensor<1x64x112x112xf32> {
+// CHECKZ17-DAG:       [[VAR_0_:%.+]] = onnx.Constant dense<[0, 0, 3, 3, 0, 0, 3, 3]> : tensor<8xi64>
+// CHECKZ17-DAG:       [[VAR_1_:%.+]] = onnx.Constant dense<0.000000e+00> : tensor<f32>
+// CHECKZ17-DAG:       [[VAR_2_:%.+]] = "onnx.NoValue"() <{value}> : () -> none
+// CHECKZ17:           [[VAR_3_:%.+]] = "onnx.Pad"([[PARAM_0_]], [[VAR_0_]], [[VAR_1_]], [[VAR_2_]]) <{mode = "constant"}> : (tensor<1x3x224x224xf32>, tensor<8xi64>, tensor<f32>, none) -> tensor<1x3x230x230xf32>
+// CHECKZ17:           [[VAR_4_:%.+]] = "onnx.Conv"([[VAR_3_]], [[PARAM_1_]], [[PARAM_2_]]) <{auto_pad = "VALID", group = 1 : si64, kernel_shape = [7, 7], pads = [0, 0, 0, 0], strides = [2, 2]}> : (tensor<1x3x230x230xf32>, tensor<64x3x7x7xf32>, tensor<64xf32>) -> tensor<1x64x112x112xf32>
+// CHECKZ17:           return [[VAR_4_]] : tensor<1x64x112x112xf32>
+// CHECKZ17:         }
 // -----
```
tungld, line 655: Add space before `// -----`. Otherwise `-split-input-file` does not work with this separation.

test/mlir/accelerators/nnpa/conversion/rewrite-onnx-for-zhigh.mlir
```diff
@@ -351,10 +352,13 @@ func.func @test_sub_dynamic_dims(%arg0: tensor<128x?xf32>) -> (tensor<128x2xf32>
 
 // -----
 
+
 func.func @test_matmul(%arg0: tensor<4x12x256x256xf32>, %arg1: tensor<4x12x256x64xf32>) -> (tensor<4x12x256x64xf32>) {
     %0= "onnx.MatMul"(%arg0, %arg1) : (tensor<4x12x256x256xf32>, tensor<4x12x256x64xf32>) -> tensor<4x12x256x64xf32>
     return %0 : tensor<4x12x256x64xf32>
 
+}
+
```
tungld, line 361: Just curious why we need these changes: moving // CHECK out of `func.func`.

test/mlir/conversion/onnx_to_krnl/Additional/Im2Col.mlir
```diff
@@ -0,0 +1,521 @@
...
+// CHECK:             [[VAR_12_:%.+]] = arith.andi [[VAR_8_]], [[VAR_11_]] : i1
+// CHECK-DAG:         [[VAR_13_:%.+]] = arith.select [[VAR_12_]], [[VAR_4_]], [[CST_0_]] : index
+// CHECK-DAG:         [[VAR_14_:%.+]] = arith.select [[VAR_12_]], [[VAR_5_]], [[CST_0_]] : index
+// CHECK:             [[LOAD_PARAM_0_MEM_:%.+]] = krnl.load [[PARAM_0_]]{{.}}[[VAR_1_]]#0, [[VAR_1_]]#3, [[VAR_13_]], [[VAR_14_]]{{.}} : memref<1x2x5x5xf32>
+// CHECK:             [[VAR_16_:%.+]] = arith.select [[VAR_12_]], [[LOAD_PARAM_0_MEM_]], [[CST_0_dot_000000_]] : f32
+// CHECK:             krnl.store [[VAR_16_]], [[RES_]]{{.}}[[VAR_1_]]#0, [[VAR_3_]], [[VAR_2_]]{{.}} : memref<1x18x9xf32>
+// CHECK:           }
+// CHECK:           return [[RES_]] : memref<1x18x9xf32>
+// CHECK:         }
+// -----
```
tungld: The condition for `// -----` to work is one empty line before and one empty line after. Please add one empty line before `// -----`, and remove redundant empty lines after.

Please fix the same issue in the other part of this file, and another file Im2Col_O3.mlir

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.td
```diff
@@ -448,7 +448,9 @@ def ConvOpPaddingPattern: Pattern<
      (createArrayAttrOfZerosFrom $nnpapads),
      $strides)
   ],
-  [(IsConvNotLegalForZDNN:$res), (CanInferencePadsForNNPAConv:$res)]
+  [(IsConvNotLegalForZDNN:$res), (CanInferencePadsForNNPAConv:$res)],
+  [],
+  (addBenefit 10) // Higher than Conv to matmul to use convolutions first.
```
tungld: Recommend to set it higher, say 1000, just in case more conv-to-matmul pattern will be added.

tungld (commented):
The rewriting looks good. It's great that we can reuse a lot of code.

Could you add a lit test in `test/mlir/accelerator/nnpa/driver` to run our whole driver pipeline:`onnx-mlir --march=z17 --maccel=NNPA --EmitZHighIR --printIR` for the path `onnx.Conv -> zhigh.Conv -> zAIU`, that is the onnx.Conv is suitable for zAIU but it is not rewritten into matmul. It's perfect if you can add other lit tests for every path in your figure. Otherwise a single one for `onnx.Conv -> zhigh.Conv -> zAIU` would be ok. Thanks!

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
AlexandreEichenberger: moved to debug

docs/optimization-onnx-lowering/ConvIm2Col.md
AlexandreEichenberger: fixed

AlexandreEichenberger: fixed, and revised the text a bit.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
AlexandreEichenberger: good suggestion, changed it only for the rewrite pass, or did you also want the global onnx-mlir `--disable-conv-to-matmul` flag to be split into 2? Personally, I don't see why we would need 2 there, but let me know if you see a scenario where its desired there.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.td
AlexandreEichenberger: done

src/Dialect/ONNX/ONNXOps/Additional/Im2Col.cpp
AlexandreEichenberger: Done (maybe we should change the interface so that a null element type fetches the output element type automatically; another PR)

src/Dialect/ONNX/Transforms/Decompose.cpp
AlexandreEichenberger: Good suggestion.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
AlexandreEichenberger: both are used.

src/Dialect/ONNX/Transforms/Decompose.cpp
AlexandreEichenberger: I need your input on this @tungld. Shape helper create krnl ops, not onnx ops. Here we are in decompose, which is way before lowering to Krnl.

I looked at the arithmetic ops IndexExpr currently use: andi/ori, add/sub, mul/div/rem, floor div/ceil div, cast, select, clamp, compare(float/int), cast, min/max.
They all have ONNX equivalent
* used a generic compare, ONNX spells each one different,
* select is onnx.where

All the arith ops are use in-place; we would have to somewhat "template" IndexExpr to use one of two code gen scheme. Or we could create 2 sub-classes, one for each backend.  If we pursue this, I suggest to keep the current code as is and make another PR to handle this major change.

src/Dialect/ONNX/Transforms/Decompose.cpp
AlexandreEichenberger: I had one test that was unhappy with the onnx.reshape call because of the "createTypedOpAndInferShapes" was unhappy when the convolution output type was unranked.

test/mlir/accelerators/nnpa/conversion/rewrite-onnx-for-zhigh.mlir
AlexandreEichenberger: Sorry, that is a side effect of when I generate the code that support multiple prefix... had to emit the code, the non-modified tests, plus the fixed test for the targeted prefix. If it bother you, I can try to change the code.

test/mlir/accelerators/nnpa/conversion/rewrite-onnx-for-zhigh.mlir
AlexandreEichenberger: fixed

AlexandreEichenberger:
> Could you add a lit test in test/mlir/accelerator/nnpa/driver to run our whole driver pipeline:onnx-mlir --march=z17 --maccel=NNPA --EmitZHighIR --printIR for the path onnx.Conv -> zhigh.Conv -> zAIU, that is the onnx.Conv is suitable for zAIU but it is not rewritten into matmul. It's perfect if you can add other lit tests for every path in your figure. Otherwise a single one for onnx.Conv -> zhigh.Conv -> zAIU would be ok. Thanks!

I added a test that make the test for the 3 scenarios: CPU, z16, z17. I technically don't check that it becomes NNPA instructions, as this is covered elsewhere. See test/mlir/accelerators/nnpa/conversion/onnx-to-zhigh/conv-for-cpu-z16-z17.mlir file. This is a regular lit test testing one pass (hybrid for CPU, rewrite ONNX for ZHigh for accelerators). It tests the 3 machines, with all dynamic image, dynamic image with static batch size of 1 (eliminate the limitations for z16 matmul with unsupported broadcast), and all static. 9 cases for a 3x3 conv and 9 more for a 1x1 conv as 1x1 are treated differently.

Let me know if you think this is not enough, I can then try to test with onnx-mlir and the EmitZHigh. Have not done it before but I know the patterns are there.

AlexandreEichenberger:
I have a technical question about the validity of our NNPA Convolution constraints (https://github.com/ibm/zdnn#convolution-2d-requirements). 

The constraint table from zDNN has 5 columns, and the last one, to my understanding state the output shape. Right now, we treat this as a constraints, and to satisfy it it forces us to have a constant image size (aka H and W, the image width and height) is expected to be constant. However, it is my understading that the shapes mentioned there are PRECISELY the normal output shapes of convolutions for the given set of other constraints (aka dilation = 1, 2D, group=1, padding = VALID or SAME (onnx SAME_UPPER). So it is my understanding that it is not an actual constraints, just a statement of what the output shapes would be already.

So in my opinion, this should be removed. There is one valid remaining constraints on the image size, and that is for VALID... where the image cannot be smaller than the kernel spacial dimensions.

TLDR, if we can remove the constraints associated with the last column (output size), we can handle more case directly to NNPA conv as we do not need any knowledge on the image size for the SAME_UPPER pad policy.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/RewriteONNXForZHigh.cpp
tungld: Yes, `--disable-conv-to-matmul` is fine, since it's for CPU also.

tungld:
> I added a test that make the test for the 3 scenarios: CPU, z16, z17. I technically don't check that it becomes NNPA instructions, as this is covered elsewhere. See test/mlir/accelerators/nnpa/conversion/onnx-to-zhigh/conv-for-cpu-z16-z17.mlir file. This is a regular lit test testing one pass (hybrid for CPU, rewrite ONNX for ZHigh for accelerators). It tests the 3 machines, with all dynamic image, dynamic image with static batch size of 1 (eliminate the limitations for z16 matmul with unsupported broadcast), and all static. 9 cases for a 3x3 conv and 9 more for a 1x1 conv as 1x1 are treated differently.
> Let me know if you think this is not enough, I can then try to test with onnx-mlir and the EmitZHigh. Have not done it before but I know the patterns are there.

I would like to see how it is in practice using `onnx-mlir` commandline to make sure there is no regression when someone changes the order of calling passes in `onnx-mlir`. Tests for a single pass cannot check such a situation.

AlexandreEichenberger:
>Could you add a lit test in test/mlir/accelerator/nnpa/driver to run our whole driver pipeline:onnx-mlir --march=z17 --maccel=NNPA --EmitZHighIR --printIR for the path onnx.Conv -> zhigh.Conv -> zAIU, that is the onnx.Conv is suitable for zAIU but it is not rewritten into matmul. It's perfect if you can add other lit tests for every path in your figure. Otherwise a single one for onnx.Conv -> zhigh.Conv -> zAIU would be ok. Thanks!

Added this test, it makes sense. Also added a lot of "end of line" near the "// ----" strings. Have a PR that will add a script that corrects the missing ones.

test/mlir/accelerators/nnpa/driver/conv-for-z16-z17.mlir
```diff
@@ -0,0 +1,104 @@
+// RUN: onnx-mlir --march=z17 --maccel=NNPA --EmitZHighIR --printIR %s | FileCheck %s --check-prefix=CHECK-Z17
+// RUN: onnx-mlir --march=z16 --maccel=NNPA --EmitZHighIR --printIR %s | FileCheck %s --check-prefix=CHECK-Z16
+
```
tungld, line 3: Please move this file to the `driver` folder since it tests the driver `onnx-mlir`.

test/mlir/accelerators/nnpa/conversion/onnx-to-zhigh/conv-for-z16-z17.mlir
```diff
@@ -0,0 +1,104 @@
...
+// RUN: onnx-mlir --march=z16 --maccel=NNPA --EmitZHighIR --printIR %s | FileCheck %s --check-prefix=CHECK-Z16
+
+// -----
+
+func.func @conv_3x3_dyn(%arg0: tensor<?x?x?x?xf32>, %arg1: tensor<64x3x3x3xf32>) -> tensor<?x?x?x?xf32> {
+  %0 = "onnx.NoValue"() <{value}> : () -> none
+  %1 = "onnx.Conv"(%arg0, %arg1, %0) <{auto_pad = "NOTSET", dilations = [1, 1], group = 1 : si64, kernel_shape = [3, 3], pads = [0, 0, 0, 0], strides = [1, 1]}> : (tensor<?x?x?x?xf32>, tensor<64x3x3x3xf32>, none) -> tensor<?x?x?x?xf32>
+  return %1 : tensor<?x?x?x?xf32>
+
+// Cannot use conv because of the dynamic image shape, but matmul with broadcast is fine
```
tungld: Do we need to end a sentence with a period? If so, change the other comments in this file also.
By the way, the correct way to comment with FileCheck is 
```
// COM: ....
```

test/mlir/accelerators/nnpa/driver/conv-for-z16-z17.mlir
```diff
@@ -0,0 +1,104 @@
...
+  return %1 : tensor<1x?x?x?xf32>
+
+// Cannot use conv because of the dynamic image shape, but matmul with broadcast is fine
+// CHECK-Z17-LABEL:  func.func @conv_3x3_dyn_bs1
+// CHECK-Z17:        onnx.Im2Col
+// CHECK-Z17:        zhigh.MatMul
+
+// Cannot use conv because of the dynamic image shape, but matmul with broadcast is fine
+// CHECK-Z16-LABEL:  func.func @conv_3x3_dyn_bs1
+// CHECK-Z16:        onnx.Im2Col
```
tungld, line 35: Do we need to check `zhigh.MatMul` also to make sure that it goes to zAIU?

tungld (approved):
LGTM! Thanks for addressing all comments!

test/mlir/accelerators/nnpa/conversion/rewrite-onnx-for-zhigh.mlir
AlexandreEichenberger: I manually changed it back, and I also changed the script

test/mlir/accelerators/nnpa/driver/conv-for-z16-z17.mlir
AlexandreEichenberger: got it, was not aware. done

test/mlir/accelerators/nnpa/conversion/onnx-to-zhigh/conv-for-z16-z17.mlir
AlexandreEichenberger: good suggestion, done.

test/mlir/accelerators/nnpa/driver/conv-for-z16-z17.mlir
AlexandreEichenberger: it should, added and it passes

test/mlir/accelerators/nnpa/conversion/onnx-to-zhigh/conv-for-cpu-z16-z17.mlir
```diff
@@ -50,15 +50,15 @@ func.func @conv_3x3_static(%arg0: tensor<2x3x256x256xf32>, %arg1: tensor<64x3x3x
...
 // CHECK-LABEL:  func.func @conv_3x3_static
 // CHECK:        onnx.Im2Col
 
-// Can use conv
+// COM: Can use conv.
 // CHECK-Z17-LABEL:  func.func @conv_3x3_static
 // CHECK-Z17:        onnx.Conv
 
-// Can use conv
+// CCOM: an use conv.
```
tungld, line 61: Typos "CCOM"

test/mlir/accelerators/nnpa/conversion/onnx-to-zhigh/conv-for-cpu-z16-z17.mlir
```diff
@@ -93,17 +93,17 @@ func.func @conv_1x1_dyn_bs1(%arg0: tensor<1x3x?x?xf32>, %arg1: tensor<64x3x1x1xf
   return %1 : tensor<1x64x?x?xf32>
 }
 
-// Cannot use conv , but matmul is best on CPU
+// COM" Cannot use conv, but matmul is best on CPU.
```
tungld, line 96: Typo "COM""

PR #3501, opened 2026-05-08
Add compile support into om_pyrt

chentong319:
Thanks Alex for restructuring the code. Now the inference is simplified:
1. container is through the C++ interface.
2. standalone compiler package provides the path
3. Compiler session initialization and compile() function are separately.

What is not done:
1. Support the old interface which combines the constructor and compile() function. The reason is that it has two required parameters with string type, same as the existing constructor for container.
2. Update the CLI usage of this package. If this feature is not regarded to be useful, I will simply remove the file.

Status:
The new test cases in om_pyrt/tests are tested successfully.

=============== Previous comments===================================
@AlexandreEichenberger @tungld @Sunny-Anand This PR passed my initial test examples. Could you have a quick preview check or we discuss this PR at meeting next week before I spend time refining it?

### Support different ways of compilation
The current om_pyrt contains only the inference interface. This PR adds the compilation interface into this package. There are three ways to compile a model: local compiler, standalone compiler and compiler container. This PR is just wrap them into one function without doing much in integration.
When using local compiler, the compile session in C and then python interface is used. The compiler is located with parameter or PATH env variable, and the python driver is located relative to compiler or another env variable. The standalone compiler is similar to local compiler except that the path of compiler and python driver is determined by the Standalone package, making it easy to install and use. The container approach uses python package to call the compiler. Since the python code for the three approaches is already there, this PR simply put them together. 

### Integrate RunONNXModel into om_pyrt
This PR turned RunONNXMode.py as the command line interface. New way to run RunONNXModel will be `python -m om_pyrt` with the same parameters. The advantage of this change: RunONNXModel can directly use the functionalities in om_pyrt package without env variable, and some useful utilities in RunONNXModel will be moved into om_pyrt, making it easier to run/test a model in end-to-end python env. 

## Finished Test
All the test case in om_pyrt/tests. e.g. `python use_container_compiler.py`
CLI test: `python -m om_pyrt -m ./test_add.mlir`

### Argument integration
Argument from different components needs to be unified. 

### Better integration could be done with extra efforts:
we can have unified compile driver even at C level, not just python level with two changes
1. Standalone compiler package (OMPyCompile) only provides the path of compiler.
5. Use the c interface for docker/podman in OMCompile c code
However, I do not feel the urgent to do this now.


AlexandreEichenberger:
@chentong319 I like the functionality that this new PR provide and the approach looks good to me. I do think that there is a some redundant code to handle the containers. I have done a quick experiment in #3502 that illustrates the code reuse that can be achieved by migrating the functionality in the c++ code, currently a separate OMUnifiedCompile that would replace OMCompile (probably keeping the OMCompile name). Most of the new functionality provided in this PR would be similar.

Great work Tong, can't wait to see it in action.

tungld:
Hi Tong,

```python
compiled_model = om_pyrt.compile(
    "./test_add.mlir",
    "local",
    compile_options="-O3",
    compiler_path="/home/chentong/workspace/onnx-mlir/build-standalone/Debug/bin/onnx-mlir",
)
```

In torch-onnxmlir, we often compile multiple models. So with this interface, we need to input `mode` and `compiler_path` multiple times. I would like to avoid such a situation by suggesting the following changes:

- First we set compiler information by
```python
om_pyrt.set_compiler_info(mode="local", compiler_path="/home/chentong/workspace/onnx-mlir/build-standalone/Debug/bin/onnx-mlir")
```
or
```python
om_pyrt.compiler_mode("local")
om_pyrt.compiler_path("/home/chentong/workspace/onnx-mlir/build-standalone/Debug/bin/onnx-mlir")
```

- Then compile a model:

```python
compiled_model = om_pyrt.compile("model.onnx", "-O3 -march=z17")
```

For `om_pyrt.compile`, is it possible to specify an output folder, which serves as a cache? e.g.

```python
compiled_model = om_pyrt.compile("model.onnx", "-O3 -march=z17", output_dir="/cache/gpt2")
```

AlexandreEichenberger:
Tung, thanks for your suggestions. The C++ class separate the initialization and the actual compile, so it would be possible to separate when we provide the compiler path info and when we do the actual compilation (by the way, that usage pattern is closer to a "compile session"... ) It would be good too, as when we use a container, there are some overheads associated with detecting docker vs podman, plus some other container related init.

> For om_pyrt.compile, is it possible to specify an output folder, which serves as a cache? e.g.

``` python
      .def(py::init<const std::string &, const std::string &, const std::string &,
               const std::string &, bool>(),
          py::arg("input_model_path"),
          py::arg("flags"),
          py::arg("compiler_path") = "",
          py::arg("log_file_name") = "",
          py::arg("reuse_compiled_model") = false,
          "Compile an ONNX model.\n\n"
          "Args:\n"
          "    input_model_path (str): Path to the input ONNX model file (.onnx).\n"
          "        Must be a valid path to an existing ONNX or MLIR model.\n"
          "    flags (str): Compilation flags as a single string.\n"
          "        Examples: '-O3', '-O3 -o output_name', '--EmitLib'.\n"
          "        All onnx-mlir command-line options are supported.\n"
          "    compiler_path (str, optional): Path to onnx-mlir compiler binary,\n"
          "        namely path plus binary name. If empty (default), use onnx-mlir\n"
          "        at its default location.\n"
          "    log_file_name (str, optional): Path to log file for compilation output.\n"
          "        If empty (default), output goes to stdout/stderr.\n"
          "    reuse_compiled_model (bool, optional): If True, reuse existing compiled\n"
          "        model if it exists instead of recompiling. Default: False.\n\n"
...
```

When reuse is there, it will not compile if the .so already exists. For example, if you compiled once with the `-o xxx-v1`, then if you recompile with that `reuse_compiled_model=True` and the same output name, then no compilation occurs. If no `-o` is provided, matching occurs with the default output file name.

If you want something more fancy than just matching the output name, let me know what you have in mind.

tungld:
Thanks, Alex, for the explanation! The main purpose of `output_dir` is to redirect the generated .so file and related files into a specific folder. Otherwise, users need to move it manually, and need to take care of not only the .so file but also the .constants.bin file, and potentially json config files.

AlexandreEichenberger:
> The main purpose of `output_dir` is to redirect the generated .so file and related files into a specific folder. Otherwise, users need to move it manually, and need to take care of not only the .so file but also the .constants.bin file, and potentially json config files.

Right now the `-o file` can be used, but its true that this forces the user to give also a name; whereas the `output_dir` option would not.

How would you handle conflicts if the `-o` gives a fully qualified path with a different `output_dir`? Error? Precedence of the one over the other?

src/Runtime/python/PyOMCompile.hpp
```diff
@@ -28,12 +28,25 @@ namespace onnx_mlir {
...
-      const std::string &compilerPath, const std::string &logFilename,
-      bool reuseCompiledModel);
+  // Constructor for local compilation
+  PyOMCompile(const std::string &compilerPath = {}, bool verbose = false);
+
+  // Constructor for container-based compilation
+  PyOMCompile(const std::string &containerImage,
+      const std::string &compilerPathInContainer,
+      const std::string &engine = "auto", bool autoPull = true,
+      bool verbose = false);
```
AlexandreEichenberger: NIT: because this class is only ment to be used from a PYBIND interface, AI suggested to not define defaults as they are defined (as you did) below. I find it makes sense as I once changed default in the one but not the other. Thus its cleaner to only have it in the PYBIND definition.

src/Runtime/python/om_pyrt/src/om_pyrt/__init__.py
```diff
@@ -18,9 +18,11 @@
     run_model_with_input_output_arrays,
     compare_result,
 )
+from .PyOMCompile import OMCompile as CompileSession
```
AlexandreEichenberger, line 21: Question for @chentong319 @tungld @Sunny-Anand:

If we aim for good self-documentation, I have a question about the current layering, naming, and how documentation flows through it.

Right now:
- We expose `OMCompile` and `OMExecutionSession` via PyBind (`PyOMCompile.hpp`, `PyOMExecutionSession.hpp`)
- These are then wrapped/renamed in `PyOMCompile.py` and `PyRuntime.py` to add functionality
- And then renamed again at a higher level (`InferenceSession`, `CompileSession`, etc.)

This introduces multiple layers of indirection with slightly different names (e.g., `OMExecutionSession` → `InferenceSession`, `OMCompile` → `CompileSession`). I understand some of this is intentional (e.g., enabling reuse of a compile session), but the accumulation of similar names makes the API a bit harder to follow.

My question is:  
**How do we want to handle consistency across naming and documentation through these layers?**  
- Do PyBind docstrings propagate all the way through these wrappers and renamings (e.g., `help(ort.InferenceSession)`), or do we need to redefine them at each level?  
- And more broadly, should we align on a consistent naming convention (Pythonic vs CamelCase), given this is user-facing and we currently mix styles?

---

This is not specific to this PR—just something that stood out.  
Since we’re shaping a user-facing interface, it might be worth agreeing on naming, layering, and documentation conventions early.

src/Runtime/python/om_pyrt/tests/use_container_compiler.py
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+
+############# use_compiler_container.py #######################################
+#
+# Copyright 2021-2025 The IBM Research Authors.
```
AlexandreEichenberger: nit: date should be 2026.

src/Runtime/python/om_pyrt/tests/use_local_compiler.py
```diff
@@ -0,0 +1,40 @@
...
+# When compiler_image_name is None, local compiler will be used.
+# The compiler_path is used to locate the compiler.
+# compile_args is the flags passed to onnx-mlir
+import om_pyrt
+
+compile_session = om_pyrt.CompileSession(
+    compiler_path="/Users/chentong/Projects/onnx-mlir/build/Debug/bin/onnx-mlir"
+)
+
+compile_session.compile(model_file, "-O3")
```
AlexandreEichenberger: please have compile and inference session in try/except block to instill good practices.

AlexandreEichenberger (commented):
Thanks for this PR; and the examples in tests are quite useful to better understand it. I left a lengthy comments about naming conventions and documentation that, though not specific to this new PR, should be addressed, maybe here or in a subsequent PR. 

Also, I would like to better understand the distinction between `local` and `standalone` and how to expose it to the user and/or if we need to expose it to the users.

src/Runtime/python/om_pyrt/src/om_pyrt/__init__.py
chentong319: I do not have a good answer to this question.

chentong319:
The major issue I found is about the re-use. After we split the constructor and compile function. When we want to check the output file exists or not, we can not simply call getOutputFilename() because the model path and the compilation flag are needed. I added an extra function, but it is not a neat solution.
Since re-use should check model and flag used for the .so, not just the existence of the .so file. I will leave the reuse issue to future PR when the usage scenario is more clear.

chentong319:
@AlexandreEichenberger Could give another try? The recent change is to integrate with your PR for output_path. One comment: the python interface keeps the separation of constructor and compile function. The C++ interface has a special constructor for old style, which supports only the use of local compiler constructor + compile function. I did not put that into python because 1) I'd like to move to the new style, 2) that constructor has ambiguity with default parameters.

src/Runtime/python/PyOMCompile.py
```diff
@@ -27,30 +27,58 @@
...
+        compiler_path="",
+        compile_policy="local",
+        compiler_image="",
+        auto_pull=True,
+        engine="auto",
+        cache=None,
+        verbose=False,
+    ):
+
+        self.cache = cache
```
AlexandreEichenberger, line 38: I would take a slightly different approach.

```
if compiler_image and compiler_path are empty:
  if standalone can be imported:
     use compiler with path defined by the standalone
  else:
      use default container and compiler path (can be left empty, there is a list of default).
else if compiler_image is not empty:
  use provided container image, and path (if empty, default will apply
else:  #  if compiler_path is not empty:
   use compiler with given path
```

that way, if the standalone is there, nothing needs to be specified
if standalone is not there, use default compiler

If specify compile_image.... then obviously we need the container version
a user would never specify a compiler path inside a container, without giving the container.

So I don't think we need the "compile_policy".

I would put the compiler_image first, and then the compiler path. You can force the other parameters to be explicitly named (I believe by adding a `, *` before the args that must be named)

AlexandreEichenberger (approved):
LGTM, thanks for the PR. Can't wait to use it as our main delivery system

PR #3504, opened 2026-05-10
Add reshape transpose canonicalization test

KOKOSde:
### Summary

Add a canonicalization lit test for a transformer-style attention shape round trip:

`Reshape -> Transpose -> Transpose -> Reshape`

The test checks that canceling transpose permutations and the final static reshape collapse back to the original tensor. This documents an important shape-only pattern around attention head reshaping.

### Testing

Not run locally: this environment does not have `onnx-mlir-opt` / FileCheck installed. The change is a lit test only and is expected to run in ONNX-MLIR CI.


AlexandreEichenberger:

PR #3507, opened 2026-05-17
Add softmax verifier

cr-zhao:
ONNXSoftmaxOp has an `axis` attribute that must be in the range
[-rank, rank-1] of the input tensor. This constraint cannot be
expressed in the ODS type constraint system and requires a
hand-written verifier.

This PR adds a custom verifier for ONNXSoftmaxOp following the
same pattern as ONNXLogSoftmaxOp::verify().

Changes:
- Add "Softmax" to OpsWithVerifier in gen_onnx_mlir.py
- Regenerate ONNXOps.td.inc
- Implement ONNXSoftmaxOp::verify() in ElementwiseUnary.cpp


cr-zhao:

src/Dialect/ONNX/ONNXOps/Math/ElementwiseUnary.cpp
```diff
@@ -565,6 +565,24 @@ LogicalResult ONNXSinhOp::inferShapes(
 // SoftmaxOp
 //===----------------------------------------------------------------------===//
 
+LogicalResult ONNXSoftmaxOp::verify() {
+  ONNXSoftmaxOpAdaptor operandAdaptor(*this);
+  if (!hasShapeAndRank(operandAdaptor.getInput()))
+    return success();
+
+  int64_t inputRank =
```
chentong319, line 596: Should we bypass the check if the input is an unranked tensor? Could this situation happen in the iterative process of shape inference?

cr-zhao: If we return failure here, the compilation would abort before 
shape inference has a chance to determine the rank. This follows 
the same pattern as ONNXLogSoftmaxOp::verify(), which also 
bypasses the check when the input is unranked.

AlexandreEichenberger: > Should we bypass the check if the input is an unranked tensor? 

@chentong319 are you happy with the current code? Test seems to be bypassed because of the presence of the 
```C++
  if (!hasShapeAndRank(operandAdaptor.getInput()))
    return success();
``` 
just before.

chentong319 (approved):
LGTM!

PR #3508, opened 2026-05-19
Support more onnx operators for the pattern unstick -> onnx op -> stick

tungld:
This PR supports more onnx operators for the pattern `unstick -> onnx op -> stick`, where the onnx op is a data movement operator. I scanned onnx ops and added all possible operators for the optimization pattern.

In particular, new supported onnx operators are: ONNXFlattenOp, ONNXGatherOp, and ONNXSplitOp.
The pattern with ONNXGatherOp is found in one of LSTM models.

The code is reorganized so that it is easy to support a new operator if available.


src/Accelerators/NNPA/Dialect/ZHigh/ZHighOps/DLF16ToF32/DLF16ToF32.cpp
```diff
@@ -19,6 +19,57 @@ using namespace onnx_mlir;
...
+    // Match: ONNX_Op (ZHighDLF16ToF32Op X), args, attrs
+    Value onnxInput = op->getOperand(0);
+    zhigh::ZHighDLF16ToF32Op dlf16ToF32Op =
+        onnxInput.getDefiningOp<zhigh::ZHighDLF16ToF32Op>();
+    if (!dlf16ToF32Op)
+      return failure();
+    Value X = dlf16ToF32Op.getOperand();
+
+    // Rewrite
+    //  ONNX_Op (ZHighDLF16ToF32Op X), args, attrs
```
AlexandreEichenberger, line 48: Can you specify what happens to the other input operands? The `ONNX_Op` operates over a 32 bit tensor, presumably for all ints inputs. After the transformation, it then operates over one 16-bit tensors; what about the other inputs, should't they be also 16-bit tensors? If there are implicit conditions on the other inputs, or if it applies only to unary operations, then there should be an explicit warning, and maybe "forUnary" in the function name. 

If other conversions are required, for the other inputs, then there is potentially a cost benefit... potentially removing one pair of 16-32 32-16 and/or reducing overhead for data movements (transpose of 16 bits cheaper than 32 bits), against potentially adding lots of changes from 32-16 for all the other inputs. If a cost benefit is introduced, conversions against constants are free, but for other non-constant, it is not.

Same costs should be taken into account for ops that create many outputs, like split.

test/mlir/accelerators/nnpa/transform/zhigh-combine.mlir
```diff
@@ -318,18 +318,42 @@ func.func @test_dlf16_to_f32(%arg0: tensor<1x3x5x?xf16, #zhigh.layout<{dataLayou
...
+// CHECK:           [[VAR_3_:%.+]] = "onnx.Flatten"([[VAR_2_]]) <{axis = 1 : si64}> : (tensor<5x7x?xf16>) -> tensor<35x?xf16>
+// CHECK:           onnx.Return [[VAR_3_]] : tensor<35x?xf16>
+// CHECK:         }
+}
+
+// -----
+
+// COM: DLF16ToF32 is delayed and pushed down through Split that has multiple outputs,
+// and it is removed when combined with F32ToDLF16.
+func.func @test_delay_dlf16_to_f32_multiple_outputs(%arg0: tensor<2x?xf16>) -> (tensor<1x?xf16>, tensor<1x?xf16>) {
```
AlexandreEichenberger, line 344: you see, this is a good example because there aer 2 f32-to-f16 after the split. But if there were none, then it's less clear that it is worth it.

src/Accelerators/NNPA/Dialect/ZHigh/ZHighOps/DLF16ToF32/DLF16ToF32.cpp
tungld: > Can you specify what happens to the other input operands? The ONNX_Op operates over a 32 bit tensor, presumably for all ints inputs. After the transformation, it then operates over one 16-bit tensors; what about the other inputs, should't they be also 16-bit tensors? If there are implicit conditions on the other inputs, or if it applies only to unary operations, then there should be an explicit warning, and maybe "forUnary" in the function name.

These operations have ~~only one operand~~ only one main operand (others like indices). The other inputs and attributes are cloned without change. I can add an assert for that.

> If other conversions are required, for the other inputs, then there is potentially a cost benefit... potentially removing one pair of 16-32 32-16 and/or reducing overhead for data movements (transpose of 16 bits cheaper than 32 bits), against potentially adding lots of changes from 32-16 for all the other inputs. If a cost benefit is introduced, conversions against constants are free, but for other non-constant, it is not.
Same costs should be taken into account for ops that create many outputs, like split.

The cost model has already been implemented in ZHighDecomposeStickUnstick.cpp in PR #3309, which takes care of multiple outputs.

test/mlir/accelerators/nnpa/transform/zhigh-combine.mlir
tungld: The cost model has already been implemented in ZHighDecomposeStickUnstick.cpp in PR https://github.com/onnx/onnx-mlir/pull/3309, which takes care of multiple outputs. If the cost model see there is no benefit, then there is no dlf16_to_f32 and this canonicalization is not applied.

AlexandreEichenberger (approved):
LGTM, thanks for adding the informative comments

PR #3509, opened 2026-05-19
[Non functional changes] Lit test fixes (plus a script to add missing separators)

AlexandreEichenberger:
Added missing separators between two functions in our lit tests.

Script `onnx-mlir/utils/fix_mlir_separators.py` will scan all the mlir files and edit them. The script does not work in absolutely all cases, so it is important to manually verify the changes.

Also added a separator after the last RUN command, as this is expected by our fixLitTest.py script. It doesn't affect the tests.

No changes were done to the compiler in this PR


test/mlir/accelerators/nnpa/conversion/device-placement/device_placement_pass.mlir
```diff
@@ -1,5 +1,7 @@
 // RUN: onnx-mlir-opt --device-placement --march=z16 --maccel=NNPA --split-input-file %s | FileCheck %s
 
+// -----
```
tungld, line 3: Adding `// -----` after `// RUN:` doesn't fix any bug and it looks redundant. Can we remove it?

AlexandreEichenberger: I use them for the fixLitTest... I can change the scheme, but its convenient.

AlexandreEichenberger:
Reimplemented with Tung's suggestion in #3513

PR #3512, opened 2026-05-20
Removed the conditions on the output shape for Conv for NNPA

AlexandreEichenberger:
Restrictions for the output shape of the NNPA are not necessary, as they simply represent the actual shape given the inputs. I removed this restriction for CONV. I added one end-to-end example where a CONV with SAME_UPPER is now sent to NNPA even when image dims are all dynamic.


src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
```diff
@@ -1455,47 +1466,31 @@ static bool checkConv2DParamRestrictions(Operation *op, int64_t inputDim,
       return onnxToZHighUnsupportedReport(op, message);
     }
     if (paddingType == "SAME_PADDING") {
```
tungld: The `then` part is empty, so it's better to use `if (paddingType != "SAME_PADDING")`

tungld (approved):
LGTM.

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
AlexandreEichenberger: I personally don't mind this construct to highlight noting is needed for that case, but transform the case to a `if(padding == valid)`.

PR #3513, opened 2026-05-20
Fixing lit tests for the right separators between tests

AlexandreEichenberger:
New version of #3509 but without the additional separators between RUN and first test. Edited the fixLitTest.py to work accordingly.

Separators are required for proper testing.


tungld (approved):
LGTM. The new script is useful. Thanks!

PR #3516, opened 2026-05-21
Add a new pass --onnx-cse-with-node-name to do sub-common expression removal for onnx ops

### LINKED ISSUE #3515: CSE pass does not work if onnx_node_name is different
The MLIR's CSE is to remove common expression in a IR, it takes all operands and attributes to build a hash when comparing two operations.
As a result, two ONNX operations with the same semantics but different `onnx_node_name` are different, so the redundant operation is not removed.

For example:

```mlir
  func.func @test_cse_onnx_node_name(%arg0: tensor<?xf32>) -> tensor<?xf32> {
    %0 = "onnx.Relu"(%arg0) {onnx_node_name = "relu_1"} : (tensor<?xf32>) -> tensor<?xf32>
    %1 = "onnx.Relu"(%arg0) {onnx_node_name = "relu_2"} : (tensor<?xf32>) -> tensor<?xf32>
    %2 = "onnx.Add"(%0, %1) : (tensor<?xf32>, tensor<?xf32>) -> tensor<?xf32>
    onnx.Return %2 : tensor<?xf32>
```

Running `onnx-mlir-opt --cse test.mlir` does not remove the redundant "onnx.Relu".

tungld:
Resolves #3515 

CSE pass in MLIR is to remove sub-common expression. However, it uses all operands and attributes to build a hash for operation equivalence. As a result, ONNX operations that have unique onnx_node_name attributes do not benefit from CSE.

The solution proposed in this PR is as follows:
- Save onnx_node_name attributes and remove them from operations
- Run MLIR's built-in CSE pass
- Restore node names (keeping first occurrence's name for merged ops)


AlexandreEichenberger (approved):
Guess this is a way to go around the fact that node names are not defining the op. I assume that there was no easy alternative.

Out of curiosity, I see the op address being used for the hash of the map. So any rewritten op lose their node name, I assume?

tungld:
> So any rewritten op lose their node name, I assume?

In tests I used, operations not removed by CSE still have their original node name since these operations are unchanged (same address I think).

PR #3517, opened 2026-05-22
Fix a lib dependency in torch_onnxmlir and the constant path

tungld:
Fix the library name in torch_onnxmlir since the name was changed.


AlexandreEichenberger (approved):
LGTM, obviously this is calling for having this in our CIs so that when we change names, it would be immediately updated here too to pass :-)

tungld:

PR #3518, opened 2026-05-22
Added support for optional outputPath to compiler interfaces

AlexandreEichenberger:
Tung requested an easy way to add an output path for the compiler's output files. The following policies are observed:
* if a `-o path/name` is given, optional output path is ignored
* if a `-o name` is given without a path and an optional path is provided, then optional path is prepended to the name.
* if no `-o` option is provided and an optional path is provided, the the optional path is prepended to the input model name to construct a new `-o output-path/input-model-name` (without the input model name extension) 

OMCompile.hpp provides 2 set of functions to get input/output filename and tag. 
* One to be used after the compilation (without parameters as they were given during the compilation) and reflect the state after successful compilation.
* One to be used before compilation, e.g. for a caching scheme trying to predict the output file name. This one takes explicit "model name and path", "flags", and possibly "output directory" so that it may predict the answers given the input parameters.

Up to now, both types of calls had the same names, but overloaded with / without input parameters. To reduce the confusion, the first set remain "get" and the second set is now "predict" (e.g. `predictOutputFilename(model path, flags, output directory)` to better distinguish the two sets of calls.

Available in the C++ and Python interface.


src/Runtime/python/PyOMCompile.hpp
```diff
@@ -29,8 +29,8 @@ namespace onnx_mlir {
 class PyOMCompile {
 public:
   PyOMCompile(const std::string &modelPath, const std::string &flags,
-      const std::string &compilerPath, const std::string &logFilename,
-      bool reuseCompiledModel);
+      const std::string &outputPath, const std::string &compilerPath,
```
chentong319, line 32: Do we need. to add predictOutputFilename() to PyOMCompile class?

AlexandreEichenberger: Yes, but I was thinking that we don't need to make it a static class. If you modify the PyOMCompile to be created with just the path, and with the compile method doing the actual work, then we can make this predictOutputFilename(model, flag, output path) a normal method.. that is why I din't add it at this time.

chentong319 (approved):
LGTM!

PR #3519, opened 2026-05-27
[NNPA] Rewrite N-D Transpose-MatMul pattern into 3D Transpose-MatMul for IBM z17

tungld:
NNPA on IBM z17 supports transposed MatMul but with 3D tensors only.

This PR rewrites `4D transpose-matmul` pattern in to `3D transpose-matmul` so that the transpose-matmul can run on NNPA.

Example of 4D transpose-matmul:
```mlir
func.func @test_transpose_matmul_4D_transA(%arg0: tensor<1x16x64x1xf32>, %arg1: tensor<1x16x64x?xf32>) -> (tensor<1x16x1x?xf32>) {
  %0 = "onnx.Transpose"(%arg0) <{perm = [0, 1, 3, 2]}> : (tensor<1x16x64x1xf32>) -> tensor<1x16x1x64xf32>
  %1 = "onnx.MatMul"(%0, %arg1) : (tensor<1x16x1x64xf32>, tensor<1x16x64x?xf32>) -> tensor<1x16x1x?xf32>
  return %1 : tensor<1x16x1x?xf32>
```

will be rewritten into:
```
Reshape %0 to 3D
Reshape %arg1 to 3D
Transpose on 3D tensors
MatMul on 3D tensors
Reshape output to original shape.
```
Transpose-MatMul on 3D tensors will be finally lowered a to single NNPA call.

Rewrite patterns in this PR supports all transpose cases: only A or only B transposed, and both A and B transposed.


AlexandreEichenberger (approved):
LGTM, so I see that you drive the 4d->3d from the matrix multiply, to basically also reduce the dimensionality of the input to the matmul. 

Maybe we don't see this pattern, but if we add an add feeding into an matmul, or a matmul feeding into another element wise op, would it be also beneficial, when possible (aka no broadcast) to move these operations to the same 3d dim? We may do that already I believe, right?

tungld:
> Maybe we don't see this pattern, but if we add an add feeding into an matmul, or a matmul feeding into another element wise op, would it be also beneficial, when possible (aka no broadcast) to move these operations to the same 3d dim? We may do that already I believe, right?

It seems we have not yet supported such a pattern yet. For such a case, I am thinking we can detect `matmul -> reshape -> add` and rewrite into `matmul->add->reshape`.

PR #3483, opened 2026-04-22
[RFC]Implement reifyResultShapes for ONNXAddOp

kimm240:
This PR includes P0-1, P0-2, P1-1 in RFC from #3355.

### Infrastructure Setup (P0-1, P0-2) 
- Added the ReifyRankedShapedTypeOpInterface declaration to the ONNX Dialect.
- Set up the registration environment to allow ONNX operations.

### Implementation for AddOp (P1-1)
- Implemented the reifyResultShapes method for ONNXAddOp.
- ShapeHelper Integration: Leveraged the existing ShapeHelper to ensure consistency with the dialect's internal shape inference logic.
- Support for Static and Dynamic Shapes
  - Static Shapes: Materializes result shapes using arith.constant.
   - Dynamic Shapes: Tracks runtime dimensions using shape.shape_of and shape.get_extent.
### Testing
- Verified the implementation using the --test-onnx-reify-result-shapes pass.
- Added LIT tests covering both static and dynamic batch scenarios to ensure correct IR generation for result shapes.
- MLIR for test is in `onnx-mlir/test/mlir/onnx/onnx_reify_result_shapes_add.mlir`.

PR #3520, opened 2026-05-28
[RFC]Introduce Reify IndexExpr Value Provider for ONNX

kimm240:
- This PR solve [THIS](https://github.com/onnx/onnx-mlir/pull/3483#discussion_r3210697953).
- It separates provider responsibilities from IndexExprBuilder/DialectBuilder wiring.  (Inheritance -> Composition)
- It updates the related ONNX paths to use the new structure.


tungld:

src/Dialect/Mlir/IndexExprBuilder.cpp
```diff
@@ -73,6 +73,50 @@ APInt getIntValue(ElementsAttr elementsAttr, Type elType, uint64_t i) {
...
+  if (externalProvider)
+    return externalProvider->getVal(arrayVal, i);
+  return getVal(arrayVal, i);
+}
+
+mlir::Value IndexExprBuilder::queryShapeVal(
+    mlir::Value tensorOrMemrefValue, uint64_t i) {
+  if (externalProvider)
+    return externalProvider->getShapeVal(tensorOrMemrefValue, i);
+  return getShapeVal(tensorOrMemrefValue, i);
```
AlexandreEichenberger: Thanks for adding this capability. 

Can you comment on the following: is the code be simpler if we were to consider the code that does not have a external provided as a "default" external provider... Or it is simpler as is.

Thanks for you input on this @kimm240

kimm240: Good feedback @AlexandreEichenberger ! I make code in commit f3772ab for simplicity.

AlexandreEichenberger (approved):
LGTM, thanks for the new pattern

AlexandreEichenberger:

PR #3521, opened 2026-05-28
[NNPA] Enable transpose-matmul in zdnnx and introduce an environment variable ZDNNX_MAX_TILE_SIZE_IN_MB

tungld:
This pull request is
- to enable transpose-matmul in zdnnx by reorganizing the current zdnnx matmul to accep transpose_a and transpose_b, and
- to introduce an environment variable ZDNNX_MAX_TILE_SIZE_IN_MB, this is one of thresholds that affect the decision of whether a ztensor is splitted or not. If not set, the maximum tensor size in NNPA is used.


AlexandreEichenberger (approved):
LGTM, tx

PR #3525, opened 2026-06-01
Lp Normalization fix

### LINKED ISSUE #3522: LpNormalization returns the input tensor instead of the normalized output
# Description

`LpNormalization` appears to return the original input tensor rather than the normalized result.

# Environment

We used the ONNX-MLIR dev Docker image: `ghcr.io/onnxmlir/onnx-mlir-dev:latest`.

Inside the Docker image:
```
onnx-mlir version 0.4.2, onnx version 1.21.0
commit: 4a13c34aa695b228599d637cdb772c19b4b18dba
LLVM version 23.0.0git
LLVM commit: 1053047a4be7d1fece3adaf5e7597f838058c947
DEBUG build with assertions.
Default target: x86_64-unknown-linux-gnu
Host CPU: westmere
```

We also used local onnx 1.21.0 and numpy 2.4.4 to generate ONNX models before invoking Docker.

# Model, Input, and Output

The generated ONNX model contains a single `LpNormalization` node:
```python
helper.make_node("LpNormalization", ["X"], ["Y"], axis=1, p=2)
```

Input:
```python
x = np.array([[1.0, 2.0, 3.0], [2.0, 4.0, 8.0]], dtype=np.float32)
```

ONNX-MLIR output:
```python
[[1. 2. 3.]
 [2. 4. 8.]]
```

Numpy and ONNX ReferenceEvaluator output:
```python
[[0.26726124 0.5345225  0.8017837 ]
 [0.21821788 0.43643576 0.8728715 ]]
```

# Reproduction

```bash
docker pull --platform linux/amd64 ghcr.io/onnxmlir/onnx-mlir-dev:latest
python repro_lpnormalization.py
```

The reproduction script `repro_lpnormalization.py` does the following:

1. Generates a minimal ONNX model containing one LpNormalization node with axis=1 and p=2.
2. Runs the ONNX-MLIR dev Docker image.
3. Compiles the model with onnx-mlir --EmitLib model.onnx.
4. Runs the compiled model.so through PyRuntime.
5. Compares the ONNX-MLIR output with NumPy and ONNX ReferenceEvaluator.

[repro_lpnormalization.py](https://github.com/user-attachments/files/28368329/repro_lpnormalization.py)

# Additional investigation

Further investigation by @benlimpa suggests that the issue may be related to the implementation here:
https://github.com/onnx/onnx-mlir/blob/4a13c34aa695b228599d637cdb772c19b4b18dba/src/Conversion/ONNXToKrnl/Math/LpNormalization.cpp#L82

The implementation appears to return the input value (`input`) rather than the computed normalization result (`res`). We have not yet fully verified whether this is the root cause, but it seems consistent with the observed behavior.

AlexandreEichenberger:
LP Normalization was buggy and not properly implemented. This version implement the full specs, added a verifier method, and added every backend tests for this operation.

While its not super optimized, in that it generates multiple loops where in theory it could be implemented by 2 specialized loops, this can be added at a later time if this operation turns out to be critical.

Fix #3522

### DIFF
--- a/docs/SupportedONNXOps-cpu.md
+++ b/docs/SupportedONNXOps-cpu.md
@@ -112,7 +112,7 @@ Onnx-mlir currently supports ONNX operations targeting up to opset 22. Limitatio
 | **Log** |6 - * | | |
 | **LogSoftmax** |13 - * |Axis 0, 1, and default currently disabled due to changes in ONNX 1.8.1/Opset 13. |Temporally removed due to changes in onnx 1.8.1. |
 | **Loop** |6 - * |Input must have static shape. Does not support int4 and uint4. | |
-| **LpNormalization** |none | | | |
+| **LpNormalization** |6 - * | | |
 | **LpPool** |none | | | |
 | **MatMul** |6 - * | | |
 | **MatMulInteger** |10 - * | | |
--- a/src/Conversion/ONNXToKrnl/Math/LpNormalization.cpp
+++ b/src/Conversion/ONNXToKrnl/Math/LpNormalization.cpp
@@ -53,8 +53,9 @@ public:
     int64_t rank = inputType.getRank();
 
     int64_t axis = adaptor.getAxis();
-    if (axis<0) axis += rank;
-    assert(axis>=0 && axis<rank && "out of bound axis");
+    if (axis < 0)
+      axis += rank;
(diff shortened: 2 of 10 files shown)


tungld (approved):
LGTM!

PR #3526, opened 2026-06-02
Add Attention to onnx dialect

chentong319:
This PR only add Attention op into onnx dialect as an unimplemented op.


tungld (approved):
LGTM.

PR #3527, opened 2026-06-02
[NNPA] Add a new pass to replace malloc by hugepage malloc

tungld:
This PR adds a new compiler pass to control the hugepage malloc optimization for NNPA.
Calls to `malloc` are replaced by calls to `OMHugePageMalloc`, a C function to allocate a buffer with hugepage advice.
Doing experiment with a matmul of `(1,12,4096,64) * (1,12,64,4096)`, hugepage malloc makes the matmul ~5x faster.

Hugepage is enabled by default, but we can disable it by using a new boolean option `--nnpa-disable-hugepage-malloc`

### DIFF
--- a/src/Accelerators/NNPA/Compiler/NNPACompilerOptions.cpp
+++ b/src/Accelerators/NNPA/Compiler/NNPACompilerOptions.cpp
@@ -138,4 +138,9 @@ llvm::cl::opt<bool> nnpaDisableShapeRestriction(
                    "Default is false"),
     llvm::cl::init(false), llvm::cl::cat(OnnxMlirCommonOptions));
 
+llvm::cl::opt<bool> nnpaDisableHugePageMalloc("nnpa-disable-hugepage-malloc",
+    llvm::cl::desc("Disable replacement of malloc with OMHugePageMalloc. "
+                   "Default is false"),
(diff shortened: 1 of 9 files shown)


chentong319:
The pass to replace malloc with THP allocation is in NNPA. Could we call that pass outside of NNPA? Will THP affect the performance of CPU? We need a "fair" comparison between CPU and NNPA.

AlexandreEichenberger:
> The pass to replace malloc with THP allocation is in NNPA. Could we call that pass outside of NNPA? Will THP affect the performance of CPU? We need a "fair" comparison between CPU and NNPA.

good question, thanks for thinking of it. If its limited to the s390 systems, then it can be enabled when we enable `-march=z16` or `z17`

tungld:
> Could we call that pass outside of NNPA? Will THP affect the performance of CPU?

MatMul on CPU is a bit slower if using THP. So I enabled THP when using `--maccel=NNPA` path only.

AlexandreEichenberger (approved):
LGTM, please respond to @chentong319 's comments

tungld:
> please respond to @chentong319 's comments

@AlexandreEichenberger We had a chat this morning.

AlexandreEichenberger:
There is a problem when targeting NNPA on map where the huge page macro is under. #3539 has a hack to solve that problem.

Issue #3528, opened 2026-06-03
Build ONNX_MLIR_ENABLE_PYRUNTIME_LIGHT=ON fails


Andreas-Krebbel:
`cmake -DMLIR_DIR=$(readlink -f ../../llvm-project/build/lib/cmake/mlir) -DONNX_MLIR_ENABLE_PYRUNTIME_LIGHT=ON ..`

fails for me with:

```
CMake Error at /home/andreas/onnx/llvm-project/build/lib/cmake/llvm/AddLLVM.cmake:78 (target_compile_options):
  Cannot specify compile options for target "benchmark" which is not built by
  this project.
Call Stack (most recent call first):
  CMakeLists.txt:235 (llvm_update_compile_flags)


CMake Error at /home/andreas/onnx/llvm-project/build/lib/cmake/llvm/AddLLVM.cmake:78 (target_compile_options):
  Cannot specify compile options for target "benchmark_main" which is not
  built by this project.
Call Stack (most recent call first):
  CMakeLists.txt:236 (llvm_update_compile_flags)


CMake Error at /home/andreas/onnx/llvm-project/build/lib/cmake/llvm/AddLLVM.cmake:78 (target_compile_options):
  Cannot specify compile options for target "rapidcheck" which is not built
  by this project.
Call Stack (most recent call first):
  CMakeLists.txt:237 (llvm_update_compile_flags)


CMake Error at /home/andreas/onnx/llvm-project/build/lib/cmake/llvm/AddLLVM.cmake:78 (target_compile_options):
  Cannot specify compile options for target "stablehlo-opt" which is not
  built by this project.
Call Stack (most recent call first):
  CMakeLists.txt:240 (llvm_update_compile_flags)


CMake Error at /home/andreas/onnx/llvm-project/build/lib/cmake/llvm/AddLLVM.cmake:78 (target_compile_options):
  Cannot specify compile options for target "stablehlo-translate" which is
  not built by this project.
Call Stack (most recent call first):
  CMakeLists.txt:241 (llvm_update_compile_flags)


-- Accelerators             : NONE
CMake Error at src/Conversion/ONNXToStablehlo/CMakeLists.txt:35 (install):
  install TARGETS given target "ChloOps" which does not exist.


-- OpenMP support           : ON
-- Performing Test LLVM_LINKER_SUPPORTS_B_SYMBOLIC_FUNCTIONS
[...]

PR #3529, opened 2026-06-03
Fix build issue with ONNX_MLIR_ENABLE_PYRUNTIME_LIGHT=ON #3528

Andreas-Krebbel:
Gate steps affecting components that are not supposed to be built with  ONNX_MLIR_ENABLE_PYRUNTIME_LIGHT=ON anyway.


tungld:
@Andreas-Krebbel Thanks for the PR! Could you add  Signed-off-by to your commit to fix the error about DCO?

PR #3532, opened 2026-06-03
Add driver package test to Jenkins

chentong319:
1. Build and test the package: om_pyrt and torch_om. Inside and outside of the container
2. Include the package in container?


chentong319:
Replaced with a new PR.

PR #3533, opened 2026-06-04
Rewrite Reshape to change allowzero=1 to allowzero=0

tungld:
ONNXReshapeOp with allowzero = 0 is more friendly when doing dimension
analysis and optimization.

The pattern is to turn ONNXReshapeOp with allowzero = 1 to the one with
allowzero = 0 for some specific cases where the semantics are equivalent.

Case 1: All static dimensions
When both input and output have all static dimensions, allowzero can be
safely changed to 0 since there are no dynamic dimensions to consider.

Example:
```mlir
%6 = "onnx.Reshape"(%arg0, %5) <{allowzero = 1 : si64}> :
(tensor<1x4x4x8x64xf32>, tensor<4xi64>) -> tensor<1x16x8x64xf32>
```
can be rewritten into:
```mlir
%6 = "onnx.Reshape"(%arg0, %5) <{allowzero = 0 : si64}> :
(tensor<1x4x4x8x64xf32>, tensor<4xi64>) -> tensor<1x16x8x64xf32>
```

Case 2: Single dynamic dimension with reshape on static dimensions only
When there is exactly one dynamic dimension in both input and output, and
the product of static dimensions is the same, the reshape only affects
static dimensions. The dynamic dimension must remain unchanged to preserve
the total number of elements.

Example:
```mlir
%6 = "onnx.Reshape"(%arg0, %5) <{allowzero = 1 : si64}> :
(tensor<1x4x4x?x64xf32>, tensor<4xi64>) -> tensor<1x16x?x64xf32>
```
can be rewritten into:
```mlir
%6 = "onnx.Reshape"(%arg0, %5) <{allowzero = 0 : si64}> :
(tensor<1x4x4x?x64xf32>, tensor<4xi64>) -> tensor<1x16x?x64xf32>
```
where `allowzero=1` becomes `allowzero=0`.

In both cases, allowzero does not change the reshape's semantics.


AlexandreEichenberger (approved):
LGTM. Stating the obvious, this file should be split at some time. We could split alphabetically, or per operation though that might generate too many files.

tungld:
> LGTM. Stating the obvious, this file should be split at some time. We could split alphabetically, or per operation though that might generate too many files.

Our AI friends could help :)

PR #3534, opened 2026-06-04
[NNPA] A rewrite pattern for expand-reshape-stick

tungld:
This PR adds a rewrite pattern to rewrite
```
onnx.Expand -> onnx.Reshape -> zhigh.Stick
```
into 
```
zhigh.F32ToDLF16 -> onnx.Expand -> onnx.Reshape -> onnx.LayoutTransform
```
so that onnx.Expand operates on f16 instead of f32, which reduces the memory footprint. In the future we should generize this pattern to support more operations, not only expand and reshape operations

This pattern is found in decoder models, which is to handle KV cache for the decoding phase.


AlexandreEichenberger:
@tungld while this indeed work on half the data size, it also add a loop, from 2 -> 3 assuming the reshape on CPU is free. I have done this merging for the extended layout operations; at some times, I wonder if we should go to a scheme where we could capture these 3 ops at once when lowering to Krnl and generating a custom loop for it. For extended layout transform, had to add a custom operation, which is ok but maybe not the most portable approach? Or we could have an enclosing operation "ONNXFusedOps" that would just be a container for the fused op, have shape inference work transparently by "ignoring" the fused op construct; and when lowering, having a "label" in the fused op that let us select the right lowering pattern. Essentially having a generic way to fuse non-obvious ops.

tungld:
> while this indeed work on half the data size, it also add a loop, from 2 -> 3 assuming the reshape on CPU is free. 

You are right. When I did some experiment, this split did give better performance, I guess it is because f32todlfloat16 was done on a 1D array so I was fast, even though it was 3 loops overall.

> I have done this merging for the extended layout operations; at some times, I wonder if we should go to a scheme where we could capture these 3 ops at once when lowering to Krnl and generating a custom loop for it.

Yes, I would give better performance I think.

AlexandreEichenberger (approved):
LGTM, thanks for the useful feedback

PR #3536, opened 2026-06-04
Use om_pyrt build for torch_onnxmlir

chentong319:
The LIGTH weight build is having some problem even after PR3530 PR3531. I found torch_onnxmlir can use the build from om_pyrt. I modified the document and tested.
Changed some script and document related to ONNX_MLIR_ENABLE_PYRUNTIME_LIGHT.
Not the CMakeList.txt yet. The change of document is urgent because all the user of torch_onnxmlir ran into this problem. I will update CMakeList.txt later after the test is added to Jenkins.


tungld (approved):
LGTM!

tungld:
@chentong319 I updated the code in torch_onnxmlir in this PR #3541.

PR #3539, opened 2026-06-05
Disable size restrictions for convTranspose generated Conv with VALID

AlexandreEichenberger:
For NNPA, there is a constraint on CONV with VALID padding policy about the minimum size of the image, where it needs to be larger than the kernel (for each of its spatial dimensions).

It turns out that for ConvTranspose, we have a have an UpsampleAndPad of the input images, and that we can prove that regardless of the input size for the ConvTranspose, the up sampled image that is then fed to the normal convolution is such that we can guarantee that that image (spatial dimensions) will be necessarily larger than the kernel.

This enables benchmarks with ConvTranspose to have their convolution work natively as NNPA Conv operations. Without this, it can still be executed by the NNPA but after further conversion of the image as an im2col step towards a matrix multiplication.

Since it is more efficient to use Conv than the 2 step im2col and matmul, this should improve performance.

Note that for one of the WatonX benchmark, `ckpt_iter_number_400.onnx` we were able to send all Conv to the NNPA but only by using a custom flag `--nnpa-disable-shape-restriction` that forcibly asserted that all size restrictions are satisfied (even though we could not verify it by hand since there were dynamic dimensions in the spatial dimensions).  With this PR, the compiler can detect that all Conv are legally fine on the NNPA (because they are either with `SAME_UPPER` padding policy, which is fine for NNPA, or they are `VALID` but coming from a `UpsampleAndPad` generated by the transformation of `ConvTranspose`.


src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
```diff
@@ -1578,14 +1583,21 @@ bool isSuitableForZDNN<ONNXConvOp>(
...
+  // shape restriction can be skipped.
+  bool inputFromUpsampleAndPad =
+      op.getX().getDefiningOp<mlir::ONNXUpsampleAndPadOp>() != nullptr;
+
   // Check parameter restrictions for conv2d for each axis. Input/output shapes
   // may be dynamic.
-  bool isHOK = checkConv2DParamRestrictions(
-      op, inputShapeH, kernelShapeH, stridesH, outputShapeH, paddingType);
+  bool isHOK = checkConv2DParamRestrictions(op, inputShapeH, kernelShapeH,
+      stridesH, outputShapeH, paddingType, inputFromUpsampleAndPad);
```
tungld: Looks like we can check `checkConv2DParamRestrictions` inside this function since the function has `op` as input.

tungld (approved):
LGTM!

src/Accelerators/NNPA/Conversion/ONNXToZHigh/ONNXLegalityCheck.cpp
AlexandreEichenberger: Good suggestion, can now check only if flag is used. Implemented.

tungld:

PR #3541, opened 2026-06-05
Update torch_onnxmlir to use om_pyrt

tungld:
This PR updates torch_onnxmlir to use om_pyrt. Now torch_onnxmlir does not depend on any code in onnx-mlir but only the om_pyrt package to compile and run a model. It means that, once `om_pyrt` is installed, we can install torch_onnxmlir by simply calling:

```
$ pip install -e src/Runtime/python/torch_onnxmlir
```


AlexandreEichenberger (approved):
LGTM, will defer to @chentong319 to verify the correct usage of his new package. 

Love how simple it is to install the new package 
```
git clone --recursive https://github.com/onnx/onnx-mlir.git
cd onnx-mlir
pip install -e src/Runtime/python/torch_onnxmlir
```

Don't you need to compile anything? Yes because its all python interface, this step does not requires it?

Eventually, I would like that we may install the pyrt the same way from an external git where it's already compiled, and that at install command also add the prerequisite pyrt automatically. Maybe add a TODO or issue, if that makes sense.

Also, please have this or another PR remove the old way... we want only one interface.

src/Runtime/python/torch_onnxmlir/src/torch_onnxmlir/sessioncache.py
```diff
@@ -81,7 +84,14 @@ def load_from_disk(self, key):
...
             config_file = os.path.join(model_dir, "config.json")
-            sess = InferenceSession(model_so)
+            old_constant_path = os.environ.get("OM_CONSTANT_PATH")
+            os.environ["OM_CONSTANT_PATH"] = model_dir
+            sess = InferenceSession(model_so, tag=key)
+            if old_constant_path is None:
+                # Remove if it wasn't set.
+                os.environ.pop("OM_CONSTANT_PATH", None)
+            else:
+                os.environ["OM_CONSTANT_PATH"] = old_constant_path
```
chentong319, line 94: I did not think it through: Can the constant path work when the compilation is done through container and execution is done outside of container?

chentong319 (approved):
LGTM!

src/Runtime/python/torch_onnxmlir/src/torch_onnxmlir/sessioncache.py
tungld: `OM_CONSTANT_PATH` is used for inference only, it does not affect the compilation.

tungld:
> Don't you need to compile anything? Yes because its all python interface, this step does not requires it?

@AlexandreEichenberger No compilation is needed. This is a pure python package.

PR #3543, opened 2026-06-08
[Doc Only] Updated doc for lightweight build

AlexandreEichenberger:
Checked the latest building of the lightweight onnx-mlir python runtime. Tested every steps, and it works perfectly granted that we do a `git clone --recursive`. I updated the doc accordingly.

Tested instructions: onnx-mlir/docs/BuildPyRuntimeLight.md

Thanks @chentong319 for the new python runtime.


chentong319 (approved):
LGTM!

PR #3545, opened 2026-06-09
Consolidated initialization of input signatures with random inputs

AlexandreEichenberger:
For our debugging, we often need to run our models with random inputs. We had supports in various different parts, and there were driving in a different ways.

Now all this is consolidated in our onnx-mlir/Runtime, and it is exported in ExecutionSession and our Python interface. Here is the options in the python interface.

It is now used in RunONNXModel.py,  RunONNXLib.cpp, and profile-model.py. 

There is a slight increase in functionality. For dynamic bounds, one can specify one input, say `2`, all of them, say `-1`, or a range `1-2`.

Also, one can specify a specific value; for example with `--value-info=1:val1.0` which fixes the value for the first input to `1.0`. Again, ranges can also be provided.

For models with sequences parameters, there is a `--value-info=1:soz-1` where the inner most dim will have a random number of 1s followed by zeros. A positive number will provide exactly that number of ones,

This functionality is based on adding the function below:
``` C++
OMTensorList *omTensorListCreateFromInputSignature(const char *inputSignatureStr,
    const char *shapeInfo = nullptr, const char *valueInfo = nullptr,
    const char *defaultLowerBound = nullptr,
    const char *defaultUpperBound = nullptr, bool verbose = false);
```

Here is the help  for the python interface. 

```python
  |  fill_input_debug(...)
 |      fill_input_debug(self: PyRuntimeC.OMExecutionSession, shape_info: str = '', value_info: str = '', default_lower_bound: str = '', default_upper_bound: str = '', seed: int = -1, verbose: bool = False) -> list[numpy.ndarray]
 |
 |      Create input tensors populated with debug/test data.
 |
 |      Generates one numpy array per model input, using the session's input
 |      signature to determine shapes and types. Shapes are taken from the
 |      signature; dynamic dimensions must be resolved via shape_info.
 |      Data is filled with random values according to the bound parameters.
 |
 |      Args:
 |          shape_info (str, optional): Dimension overrides for dynamic axes.
 |              Format: 'INPUT_ID:D1xD2x...xDn, ...' where INPUT_ID is an integer
 |              >= 0 for a single input, a range (e.g. '2-5'), or -1 for all inputs.
 |              A dim value of -1 keeps the value from the signature.
 |              Example: '0:1x28x28,1:4' sets input 0 to shape [1,28,28] and
 |              input 1 to shape [4]. Default: '' (use the signature dims unchanged).
 |          value_info (str, optional): Per-input fill specification, overriding
 |              default_lower_bound/default_upper_bound for individual inputs.
 |              Format: 'INPUT_ID:spec1 spec2 ..., ...' where each spec is one of:
 |                min<n>  sets the lower bound for that input
 |                max<n>  sets the upper bound for that input
 |                val<n>  constant fill (equivalent to min=max=n)
 |                soz<n>  sequence-of-ones-then-zeros along the innermost
 |                        dimension: n leading ones per row, rest zeros.
 |                        n=-1 picks a random count in [1, innerDim-1] per row.
 |                        n >= innerDim is capped (all ones). Well suited for
 |                        sequence-length and attention-mask inputs.
 |              Example 1: '0:min-1.0max1.0,1:val0' fills input 0 randomly in
 |                [-1,1] and input 1 with all zeros.
 |              Example 2: '0:soz-1' fills input 0 with random soz masks.
 |              Default: '' (use type defaults).
 |          default_lower_bound (str, optional): Per-type lower bound overrides.
 |              Format: 'typename:value, typename:value, ...'
 |              Supported types: bool_ (or bool), int8, uint8, int16, uint16,
 |              int32, uint32, int64, uint64, float16, float32, float64.
 |              Built-in defaults when not overridden:
 |                floats: -0.1,  signed ints: -10,  unsigned: 0,
 |                bool_: 0 (false),  string: 0.
 |              Example: 'float32:-0.5,int64:-100'. Default: ''.
 |          default_upper_bound (str, optional): Per-type upper bound overrides.
 |              Same format as default_lower_bound.
 |              Built-in defaults when not overridden:
 |                floats: 0.1,  signed ints: 10,  unsigned: 10,
 |                bool_: 1 (true),  string: 63.
 |              Example: 'float32:0.5,int64:100'. Default: ''.
 |          verbose (bool, optional): Print shape and value range of each created
 |              tensor to stdout. Default: False.
 |          seed (int, optional): Seed for the C++ RNG for reproducibility, mirroring
 |              np.random.seed(). Pass -1 (default) to use a random seed.
 |
 |      Returns:
 |          list[numpy.ndarray]: One array per model input, ready to pass to run().
 |
 |      Raises:
 |          RuntimeError: If signature parsing fails or a dynamic dimension is
 |              not resolved.
 |
 |      Example:
 |          >>> session = OMExecutionSession('mnist.so')
 |          >>> inputs = session.fill_input_debug()
 |          >>> outputs = session.run(inputs)
``` 

Example for profile-model.py
```
 profile-model.py -m roberta-base-11.so --input-value=0:min1max500 --shape-info=0:1x384 -t 30 
```

where the values and shape are now given by the ` --input-value=0:min1max500 --shape-info=0:1x384` parameters.


utils/RunONNXModel.py
```diff
@@ -1060,13 +950,18 @@ def process_inputs(self, input_feed=None):
...
+                self.inputs = self.session.fill_input_debug(
+                    shape_info=args.shape_info or "",
+                    value_info=args.input_value or "",
+                    default_lower_bound=args.lower_bound or "",
+                    default_upper_bound=args.upper_bound or "",
+                    seed=int(float(args.seed)),
+                    verbose=True,
+                )
+            except RuntimeError:
+                # The C++ layer already printed the specific error to stderr.
```
tungld, line 963: Copilot said: 


Concern: Relying on stderr for user-facing diagnostics is risky—stderr may be redirected or lost. Consider having the C++ layer return a structured error message (via exception string) that Python can log. Example: except RuntimeError as e: print(f"Error: {e}", file=sys.stderr); sys.exit(1).

tungld (approved):
LGTM.

utils/RunONNXModel.py
AlexandreEichenberger: AI tells me that here its ok:

The specific pattern here is intentional. The except RuntimeError: sys.exit(1) at lines 890-892 and 962-964 is deliberately silent on the Python side because the C++ layer (the session_wrapper and data generation calls) already emitted a specific, detailed error message. Printing a generic f"Error: {e}" on top of that would either:

* Print a vague/empty message after a more specific one (worse UX), or
* Duplicate the same message (confusing)

AlexandreEichenberger:

PR #3546, opened 2026-06-09
Fixes for mnist for new PYRT interfaces

AlexandreEichenberger:
Mnist used the old interface prior to upgrades to the compiler interface for PYRT.

This PR fixes the issues so that it may run again properly.

### DIFF
--- a/docs/mnist_example/mnist-compile.py
+++ b/docs/mnist_example/mnist-compile.py
@@ -18,11 +18,9 @@ if False:
     flags = "-O3"
 
 try:
-    compiler = OMCompile(model, flags)
+    compiler = OMCompile()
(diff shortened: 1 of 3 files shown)


docs/mnist_example/mnist-compile.py
```diff
@@ -18,11 +18,9 @@
     flags = "-O3"
 
 try:
-    compiler = OMCompile(model, flags)
+    compiler = OMCompile()
+    compiled_model = compiler.compile(model, flags)
```
chentong319, line 22: Is there a comment somewhere saying that the onnx-mlir is assumed to be found with $PATH? Our previous docs and example assumed that everything is done inside the container.

chentong319 (approved):
LGTM!

PR #3547, opened 2026-06-15
Remove unnecessary include of Passes.hpp

chentong319:
When I built onnx-mlir from scratch on Mac, I got error for ONNXDimAnalysis.cpp: The KrnlPass.hpp.inc not found. The reason is that the dependency for KrnlPass table gen is not added. The success of build depends on how cmake schedules the tasks. Other than adding the dependency, I tried to remove the include of passes.hpp.


tungld (approved):
LGTM.

PR #3548, opened 2026-06-15
Fix the dependency of omDebugHelper

chentong319:
Two issues:
1. omDebugHelper is not added for OMPyRt build yet.
2. The dependency, cruntime, is too much. Only the OMTensorUtils is needed. The code for op runtime is not needed by driver.


AlexandreEichenberger (approved):
LGTM, thanks for the cleanup, appreciate that we only try to include what is needed.

PR #3549, opened 2026-06-16
Shape inference for Attention Op

chentong319:
1. Shape inference code
2. Lit test cases


src/Dialect/ONNX/ONNXOps/Math/Attention.cpp
```diff
@@ -0,0 +1,251 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//===---------------- Attention.cpp - ONNX Operations --------------------====//
+//
+// Copyright 2024-2025 The IBM Research Authors.
```
tungld, line 7: 2026?

tungld (approved):
LGTM!

PR #3550, opened 2026-06-16
Added compile info to profile-ir output

AlexandreEichenberger:
When doing profiling, users have requested to get some info about the compiler version/options.
I added the call to print this information directly in OMInstrument.c so that the printout info is available regardless of the programming environment (C, C++, python, java).

The new runtime log is now
```
==COMPILE-INFO-REPORT==, {
"compiler_version": "onnx-mlir version 0.5.1 (a78c43f1f)",
"compile_options": "-O3 -march=z17 -maccel=NNPA -profile-ir=ZHigh -o model extended-layout-transform-v1.mlir",
"accelerators": {"NNPA": {"nnpa_level": "--march=arch15", "zdnn_version": "1.1.1"}},
"op_stats": { [...]}
}
```

and I now also report this info in a condensed fashion in `make-report.py`
```
Report using runtime file "bibi.log"
  compiler_version: onnx-mlir version 0.5.1 (a78c43f1f)
  compile_options:  -O3 -march=z17 -maccel=NNPA -profile-ir=ZHigh -o model extended-layout-transform-v1.mlir
  accelerator NNPA: {'nnpa_level': '--march=arch15', 'zdnn_version': '1.1.1'}
[...]
```

Added info about the zDNN version as seen by the compiler.

### DIFF
--- a/src/Runtime/OMInstrument.c
+++ b/src/Runtime/OMInstrument.c
@@ -224,11 +224,17 @@ static void ProcessName(
 // =============================================================================
 // Buffer management
 
+// Weak ref: symbol is absent when compiled with --omit-compile-info.
+__attribute__((weak)) const char *omCompilationInfo(void);
+
 static inline void printStartReport() {
   if (!startReportPrinted) {
     assert(instrumentFout); // Error  "expected instrumentInitialized
                             // instrumentFout for reporting".
     fprintf(instrumentFout, "==START-REPORT==\n");
+    if (omCompilationInfo)
+      fprintf(instrumentFout, "==COMPILE-INFO-REPORT==, %s\n",
+          omCompilationInfo());
     startReportPrinted = true;
   }
 }
--- a/src/Runtime/OMInstrument.c
+++ b/src/Runtime/OMInstrument.c
@@ -224,9 +224,9 @@ static void ProcessName(
 // =============================================================================
 // Buffer management
 
-// Declared here; always defined in the model .so by emitCompilationInfo().
-// Returns "{}" when compiled with --omit-compile-info, full JSON otherwise.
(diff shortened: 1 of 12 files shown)


AlexandreEichenberger:
As @chentong319 commented, I have to use weak symbols for the getting the compile info strings because of the option that possibly remove this function.

@tungld , do you think the option to remove the `--omit-compile-info` option? What was the idea behind having it>

utils/CheckONNXModel.py
```diff
@@ -187,6 +187,20 @@ def valid_onnx_input(fname):
     " Supported types are bool, uint8, int8, uint16, int16, uint32, int32,"
     " uint64, int64, float16, float32, float64",
 )
+parser.add_argument(
```
AlexandreEichenberger, line 190: unrelated change, just added to CheckONNXModel.py the same options as existed for RunONNXModel.py

src/Runtime/OMInstrument.c
```diff
@@ -224,11 +224,18 @@ static void ProcessName(
 // =============================================================================
 // Buffer management
 
+// Declared here; always defined in the model .so by emitCompilationInfo().
+// Returns "{}" when compiled with --omit-compile-info, full JSON otherwise.
+extern const char *omCompilationInfo(void);
```
tungld: The .so model is not available when building onnx-mlir, so we may need to include `#include <onnx-mlir/Runtime/OMCompilationInfo.h>` which declares omCompilationInfo.

tungld (approved):
LGTM!

AlexandreEichenberger:
Reimplemented to avoid weak symbols, which are potentially an issue for some os that are very strict. See #3552

PR #3551, opened 2026-06-17
Support ScatterElements reduction="add" lowering

napronald:
`ScatterElements` with `reduction="add"` was previously accepted, but lowered like `reduction="none"` by directly storing updates. This was incorrect for duplicate indices, where updates should accumulate.

This PR adds the missing add-reduction path while preserving `reduction="none"` and rejecting unsupported reductions.

### DIFF
--- a/src/Conversion/ONNXToKrnl/Tensor/ScatterElements.cpp
+++ b/src/Conversion/ONNXToKrnl/Tensor/ScatterElements.cpp
@@ -30,7 +30,8 @@ struct ONNXScatterElementsOpLowering
     Operation *op = scatterElementsOp.getOperation();
     Location loc = ONNXLoc<ONNXScatterElementsOp>(op);
 
-    MultiDialectBuilder<KrnlBuilder, IndexExprBuilderForKrnl, MemRefBuilder>
+    MultiDialectBuilder<KrnlBuilder, IndexExprBuilderForKrnl, MemRefBuilder,
+        MathBuilder>
         create(rewriter, loc);
 
     // Operands and attributes.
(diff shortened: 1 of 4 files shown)


src/Conversion/ONNXToKrnl/Tensor/ScatterElements.cpp
```diff
@@ -102,7 +109,13 @@ struct ONNXScatterElementsOpLowering
             outputAccessFct.emplace_back((i == axis) ? index : accessFct[i]);
 
           // Scatter updateVal into the output tensor.
-          createKrnl.storeIE(updateVal, output, outputAccessFct);
+          if (reduction == "add") {
+            Value oldVal = createKrnl.loadIE(output, outputAccessFct);
+            Value newVal = create.math.add(oldVal, updateVal);
+            createKrnl.storeIE(newVal, output, outputAccessFct);
+          } else {
```
chentong319, line 116: Please add a comment: Previous check guarantees the reduction is none.

test/backend/inference_backend.py
```diff
@@ -3056,7 +3056,7 @@ def get_test_models():
         "test_scan9_sum_cpu": {STATIC_SHAPE: {}},
         # ==OP== ScatterElements
         # ==MIN== 11
-        # ==LIM== Does not support duplicate indices.
+        # ==LIM== Does not support reductions "max" and "min". Duplicate indices are supported for reduction="add".
```
chentong319: Can we unify the comment here and SupportedONNXOps-cpu.md?

napronald:
Addressed both comments.

chentong319:
I forgot this PR and opened a new one #3639 to support all the reduction ops for ScatterElement and ScatterND. Sorry for that. Please close this PR.

PR #3552, opened 2026-06-17
Logging compiler info in performance profile, without weak symbols.

AlexandreEichenberger:
New implementation of #3550 which does not use weak symbols but has a new call to pass compiler info directly to the profiling facility.

Previously, runtime instrumentation initialization was piggybacked onto the first real before/after KrnlInstrumentOp by setting an INIT bit in its tag — conflating "start the runtime" with "measure this op." This PR separates the two concerns by introducing `KrnlInstrumentInitOp` (krnl.runtime_instrument_init), a new Krnl dialect op emitted once per function before the first instrumented op. Its lowering calls `omCompilationInfo()` — defined in the same model .so — and passes the result to a new runtime entry point `OMInstrumentPointInit(tag, compilationInfo)`, which initializes the instrumentation runtime and stores the compilation metadata for reporting. This removes the need for the weak/strong symbol override mechanism previously used to pull `omCompilationInfo` into the runtime at print time.

On the runtime side, OMInstrument.c gains the instrumentCompilationInfo static pointer set by OMInstrumentPointInit; printStartReport uses it directly instead of calling `omCompilationInfo()`. 

As in #3550, using the `--profile-ir` option generates this in the log file (or std out)
```
==COMPILE-INFO-REPORT==, {
"compiler_version": "onnx-mlir version 0.5.1 (d648f9242)",
"compile_options": "-O3 -march=z17 -maccel=NNPA -profile-ir=ZHigh -o model extended-layout-transform-v1.mlir",
"accelerators": {"NNPA": {"nnpa_level": "--march=arch15", "zdnn_version": "1.1.1"}},
"op_stats": {
  "func.return.4D" : 1,
[...]
  "zhigh.Unstick.3D" : 1
}
}
```

and is summarized in the `make-report.py`
```
Report using runtime file "bibi.log"
  compiler_version: onnx-mlir version 0.5.1 (d648f9242)
  compile_options:  -O3 -march=z17 -maccel=NNPA -profile-ir=ZHigh -o model extended-layout-transform-v1.mlir
  accelerator NNPA: {'nnpa_level': '--march=arch15', 'zdnn_version': '1.1.1'}
```


tungld (approved):
LGTM. Thanks for quickly getting the new way work!

PR #3554, opened 2026-06-17
Added new RunONNXLib.cpp lit test

AlexandreEichenberger:
We have a test that run a model in C++ where the model is statically linked into the executable. Added a new test to ensure that we maintain this capability.


chentong319 (approved):
LGTM!

PR #3556, opened 2026-06-19
Add om_pyrt test into Jenkins

chentong319:
Two test of om_pyrt:
1. Inside the onnx-mlir-dev image. This test builds and install om_pyrt package. Then run the use_local_compiler.py to test whether the recently built onnx-mlir is usable.
2. Outside of  the onnx-mlir image. This test build s and install om_pyrt package in another independent image. Then run the use_container_compiler.py to test whether onnx-mlir the recently built onnx-mlir image can be used through docker interface.


chentong319:
@gongsu832 I have a particular question for you: do I need to do anything to clean up the image I created?

src/Runtime/python/om_pyrt/tests/use_container_compiler.py
```diff
@@ -22,15 +31,13 @@
...
+        compiler_path="onnx-mlir",
     )
 
-    compiled_model = compile_session.compile(
-        model_file, "-O3", reuse_compiled_model=True
-    )
+    compiled_model = compile_session.compile(model_file, "-O3")
 except Exception as e:
-    print("Failed to compile")
+    print("Fialed to compile")
```
tungld: Typos

.buildbot/jenkins-verify-om-package.py
```diff
@@ -0,0 +1,86 @@
...
+OM_PYRT_LIGHT_DOCKERFILE = "docker/Dockerfile.om-pyrt-light"
+
+
+def main():
+    image_name = github_repo_name + "-om-pyrt-light"
+    image_tag = pr_image_tag
+    image_full = image_name + ":" + image_tag
+
+    layer_sha256 = ""
+    for line in docker_api.build(
```
tungld, line 14: I wonder if it is better to build om-pyrt-light image once in Jenkinsfile, then multiple verifiers can reuse it. With that way, we can clean up the image using `jenkins-cleanup-build-states.py`. Also, we don't need to manually check whether the image build is successful or not here.

chentong319: If you mean to reuse the om-pyrt-light image during one Jenkins test, I can move the definition of the image name into Jenkins_common so that other stage can use it.

src/Runtime/python/om_pyrt/tests/use_container_compiler.py
chentong319: yes

.buildbot/jenkins-verify-om-package.py
chentong319: @tungld Or did you mean to use the container created by Jenkins directly without creating another container, as Jenkins-verify-user-image.py does? I tried that approach at very beginning. There are some software to install. So I borrowed the code for Docker.onnx-mlir-dev.

tungld: Hi @chentong319 sorry for the confusion. How about moving the test `use_container_compiler.py` into Dockerfile.om-pyrt-light? Similar to tests in Dockerfile.onnx-mlir-dev. So that in the future we can add more tests there. Also, when users want to build the om_pyrt image by themselves, the tests are ran to make sure the image is good.

chentong319: Oh, I see. Yes, it is natural to run use_contain_compiler.py inside Dockerfile.om-pyrt-light. In fact, I tried that and failed due to the container-in-container issue. AI told me not to do that due to the "exposure of container by Jenkins", but execute the test with a separate docker run. I can try it again to see whether it is possible. Perhaps AI is affected by the existing Jenkins-verify-user-image.py?

chentong319: Here is the reason that we can not run the test inside Dockerfile:
1. The test is a container-in-container situation.
2. docker.sock from host need to be available.
3. docker_api.build is used for the Dockerfile.*. Classic Docker build is isolated → no host volume mounts (like /var/run/docker.sock)
Therefore, we have to separate the build and the test in two steps.

tungld: OK. Thanks for checking this @chentong319! May need to note this somewhere for users of om_pyrt, e.g, how to use om_pyrt in a container.

tungld (approved):
LGTM.

PR #3557, opened 2026-06-23
Move the build of om_pyrt into pip install

chentong319:
To install om_pyrt, the current implementation requires user to perform three steps:
1. get the source with git clone onnx-mlir
2. build the required library customized to their system 
3. install the package

The PR will simplify the steps into one step: `pip install git+https://github.com/onnx/onnx-mlir.git`

Currently, I tested it with `pip install git+ssh://git@github.com/chentong319/onnx-mlir.git@pip-build`

After this PR is merged, the test can be added into Jenkins.


docs/BuildPyRuntimeLight.md
```diff
@@ -6,28 +6,44 @@ The onnx-mlir compiler can compile an ONNX model into a shared library (`.so` fi
...
-- **OMTensorUtils** (`src/Runtime`)
-- **Python driver** (`src/Runtime/python`)
-- **Utility functions**
-- **third_party/pybind11**
-
-The lightweight PyRuntimeC build is controlled by the CMake option: `ONNX_MLIR_TARGET_TO_BUILD=OMPyRt`
-
 ## Prerequisites
 
 - CMake (version 3.15 or higher recommended)
```
AlexandreEichenberger, line 11: Anyway to express these dependence inside the pip install mechanism so that it would get them automatically?

AlexandreEichenberger (approved):
LGTM (not an expert at the pip thing, so assume that its good and/or ask another person for review)

docs/BuildPyRuntimeLight.md
chentong319: Good idea. Should be consistent with onnx-mlir/requirement.txt.

PR #3558, opened 2026-06-24
Update torch_onnxmlir with several features

tungld:
This PR update torch_onnxmlir with several features:
- provide a new config option: `keep_onnx_files` if users want to keep intermediate onnx files. By default, onnx files are removed to save storage space.
- add `onnx_utils.py` for editing the exported onnx models, e.g. removing identity ops. This helps with a sliding window attention (SWA) based model where the exported onnx model has invalid identity ops (same name for both input and output)
- include the information for `example_inputs` into the hash value which is better to distinguish FX graphs.

In addition, this PR adds a new function in ExecutionSession, that is `getSharedLibPath` to return the path to .so file. At the python level, the function is `get_compiled_model_path` in PyRuntime. This function helps use get the path from a **cached** InferenceSession.


chentong319:
I remember that you created an interface to extract info from .so. Are we using that feature for caching in torch_onnxmlir? Can we use a saved cache directory for a new session?

src/Runtime/python/torch_onnxmlir/src/torch_onnxmlir/backend.py
```diff
@@ -279,7 +327,7 @@ def __init__(self, gm: torch.fx.GraphModule, *args, **kwargs):
             if len(same_hash_counter) == same_hash_size and all(same_hash_counter):
                 self.cache_key = self.gm.meta["om_hash"]
             else:
-                self.cache_key = generate_hash_key(self.gm, kwargs["options"])
+                self.cache_key = generate_hash_key(self.gm, args, kwargs["options"])
```
chentong319, line 336: Could we sort the options if it is not too much trouble?

src/Runtime/python/torch_onnxmlir/src/torch_onnxmlir/sessioncache.py
```diff
@@ -60,74 +61,92 @@ def get(self, key):
...
-            if old_constant_path is None:
-                # Remove if it wasn't set.
-                os.environ.pop("OM_CONSTANT_PATH", None)
+            if Path(model_so).exists():
+                sess = InferenceSession(model_so, tag=key)
             else:
-                os.environ["OM_CONSTANT_PATH"] = old_constant_path
+                return None
+            # Load config file.
+            config_file = os.path.join(model_dir, "config.json")
```
chentong319: Should we use a specific name for the file "config.json"?

src/Runtime/python/PyExecutionSession.cpp
```diff
@@ -466,6 +466,10 @@ std::string PyExecutionSession::pyCompilationInfo() const {
   return compilationInfo();
 }
 
+std::string PyExecutionSession::pyCompiledModelPath() const {
+  return getSharedLibPath();
```
chentong319, line 470: Any reason to use different names: CompiledModelPath and SharedLibPath?

src/Runtime/python/torch_onnxmlir/src/torch_onnxmlir/backend.py
tungld: Yes, done. I sorted it before hashing.

src/Runtime/python/torch_onnxmlir/src/torch_onnxmlir/sessioncache.py
tungld: Sure, changed to `om_backend_config.json`.

src/Runtime/python/PyExecutionSession.cpp
tungld: Good catch! I used `CompiledModelPath` at the beginning. Now move it to `SharedLibPath` for consistency.

tungld:
> I remember that you created an interface to extract info from .so. Are we using that feature for caching in torch_onnxmlir? 

Right, I just added these info into the json file.

> Can we use a saved cache directory for a new session?

Yes, new session will be created from cache if it exists so that there is no need for exporting and compilation.

chentong319 (approved):
LGTM!

PR #3560, opened 2026-06-25
Fused op support 

AlexandreEichenberger:
Support Extended Layout Transform via fused op. This is a testbed to show that this approach is viable. It does the following

1. add onnx.fused op plus support for handling shape inference through it. If we failed to lower fused op as is, the backup is to lower individual operations as is. Also profiling instrumentation will profile the fused op as is (without going further into the ops that are fused). Name is `fused.<<kind>>` where kind is the identifier with a specific fused rule that created that fused op.
2. add a support class and subclass to facilitate the creation of fused op.
3. Because some of the shape support (like concat to generate the shape of a reshape) can be an input to a fused op, but that shape may not be actually used (because the fused op "know" all it need about the reshape), this can leave some "dead code" where we have an allocated memory that is written but never used. I added a quick pass that detect "written but never read" arrays and delete them. It runs once things are lowered to affine.

Running example is ExtendedLayoutTransform.

Will add lit tests that test the new functionality. I ran Granite 3.1 with (default) and without (`-disable-fused-op`) and it yielded the same results. 

The ultimate goal is that this approach is easier to extend to different patterns, to patterns with multiple inputs/outputs,...

Relevant slides to explain the interface

image

image

image


AlexandreEichenberger:
@tungld @chentong319 fixed the failures, added lit test. ready for your formal review. Thanks

src/Accelerators/NNPA/Conversion/ZHighToZLow/ZHighToZLow.cpp
```diff
@@ -2482,6 +2483,259 @@ struct ZHighToZLowExtendedLayoutTransformLowering
   }
 };
 
+//===----------------------------------------------------------------------===//
+// ONNXFusedOp lowering for kind "zhigh.extended_layout_transform"
+//===----------------------------------------------------------------------===//
+
+struct ZHighToZLowFusedExtLayoutTransformLowering
```
AlexandreEichenberger, line 2490: FYI: this matches the previous implementation for lowering ExtendedLayoutTransform

src/Accelerators/NNPA/Dialect/ZHigh/ZHighOps/OpFusionHelper.cpp
```diff
@@ -0,0 +1,383 @@
...
+
+/// Return true if \p perm keeps the last dimension in place.
+static bool transposeKeepsLastDim(ArrayAttr perm) {
+  int64_t rank = static_cast<int64_t>(perm.size());
+  return ArrayAttrIntVal(perm, rank - 1) == rank - 1;
+}
+
+/// Try to interpret \p reshape as a split (outRank == inRank + 1).
+/// Mirrors PatternsForExtendedLayoutTransform::locateReshapeSplit exactly.
+/// On success fills \p axis and \p factor and returns true.
```
AlexandreEichenberger, line 62: FYI: this matches the previous code that detected when to generate an ExtendedLaoutTranform operation

src/Dialect/ONNX/Transforms/InstrumentONNXSignaturePass.cpp
```diff
@@ -80,58 +80,67 @@ class InstrumentONNXSignaturePass
...
-        // printing of data as it also will print the signature.
-        bool gotOne = false;
-        if (nodeNamePattern != "NONE" && nodeNamePattern != "") {
-          StringAttr onnxNodeName =
-              op->getAttrOfType<mlir::StringAttr>("onnx_node_name");
-          if (onnxNodeName && !onnxNodeName.getValue().empty()) {
-            std::string nodeNameString = onnxNodeName.getValue().str();
-            gotOne =
-                checkAndInsert(traceSpecificNodePattern, nodeNameString, 1);
+    // Pre-order walk so we can skip ONNXFusedOp bodies with WalkResult::skip().
```
AlexandreEichenberger, line 83: FYI: for signatures, we don't go into the fused operation: just print the sig of the inputs and outputs of the fused operation

src/Transform/InstrumentPass.cpp
```diff
@@ -131,51 +131,60 @@ class InstrumentPass
...
-      if (isa<KrnlInstrumentOp>(op) || isa<KrnlInstrumentInitOp>(op) ||
-          isa<ONNXPrintSignatureOp>(op))
-        return WalkResult::advance();
-
-      if (allowedOps.isEnabled(opName)) {
-        if (instrumentBefore) {
-          if (!hasInitializedRuntime) {
-            mlir::KrnlInstrumentInitOp::create(opBuilder, loc, initTag());
-            hasInitializedRuntime = true;
+    // Pre-order walk so we can skip ONNXFusedOp bodies with WalkResult::skip().
```
AlexandreEichenberger, line 134: FYI: for instrumentation, we print a before/after for the fused op, not for the ops inside.

AlexandreEichenberger:

src/Transform/EliminateWriteOnlyAlloc.cpp
```diff
@@ -0,0 +1,263 @@
+//===-- EliminateWriteOnlyAlloc.cpp - Remove write-only local allocs ------===//
```
tungld, line 5: Need to add the  header:
```
/*
 */
```

tungld (approved):
LGTM!

PR #3562, opened 2026-06-30
Removed absolute and .. path for data directory

AlexandreEichenberger:
ONNX disable extension directory to include `..`

from the specs:
> Up-directory path components such as .. are disallowed and should be stripped when parsing

ONNX also disable absolute path; they should start from the model dir.

Test example was added


test/mlir/onnx/parse/external_data_traversal.json
```diff
@@ -0,0 +1,59 @@
...
+        "attribute": [
+          {
+            "name": "value",
+            "t": {
+              "dims": ["1"],
+              "dataType": 1,
+              "name": "tensor0",
+              "externalData": [
+                {
+                  "key": "location",
```
Sunny-Anand, line 24: Might also add `a/../b` as another test for nested `..` up directory path component.

Sunny-Anand (approved):
Small suggestion on the test case. LGTM

test/mlir/onnx/parse/external_data_traversal.json
AlexandreEichenberger: Added a "truly/../evil" test case. How often do you get to do this @Sunny-Anand :-)

AlexandreEichenberger:

PR #3563, opened 2026-06-30
add dependencies to torch_onnxmlir

chentong319:
1. Fix the typo of "dependences". This entry didn't  really work previously.
2. Specify how to install om_pyrt from git in the dependencies.

It is supposed to install everything with one command.
from git:
```
pip install "git+https://github.com/onnx/onnx-mlir.git#subdirectory=src/Runtime/python/torch_onnxmlir"
```
From PyPi (Not work yet):
```
pip install torch_onnxmlir
```

There is some issues on s390x. The install of torch will fail. Instead, I have to use this command before the torch_onnxmlir:
```
pip install torch --index-url https://download.pytorch.org/whl/cpu
```


tungld (approved):
LGTM.

tungld:
> pip install troch_onnxmlir

Did you push torch_onnxmlir to PyPI?

chentong319:
> > pip install troch_onnxmlir
> 
> Did you push torch_onnxmlir to PyPI?

No, I haven't.

chentong319:
@tungld I can install the package correctly. But the onnx package cannot be imported. Even AI could not figure out why.  I tried also install onnx package locally from onnx-mlir/third_part/onnx. The installation is fine but `python -c "import onnx"` failed. I am doing this in a venv on zaiutest, not in container. Did you ever install onnx successfully outside the container?
I tried to run a test case before merging the PR. Which one in torch_onnxmlir/tests should I try if I am outside the container?

tungld:
Let me  see if we can remove the dependency on onnx or not. I was for fixing a bug which potentially didn't exists anymore.

> I tried to run a test case before merging the PR. Which one in torch_onnxmlir/tests should I try if I am outside the container?

We can use https://github.com/onnx/onnx-mlir/blob/main/src/Runtime/python/torch_onnxmlir/tests/test_add.py and set `compiler_image_name`.

chentong319:
Even with the LD_PRELOAD defined, I got error when running modified test_add.py (with docker image defined):
```
docker: symbol lookup error: /home/chentong/workspace/torch-om-env/lib/python3.12/site-packages/onnx/onnx_cpp2py_export.abi3.so: undefined symbol: PyExc_ValueError
podman: symbol lookup error: /home/chentong/workspace/torch-om-env/lib/python3.12/site-packages/onnx/onnx_cpp2py_export.abi3.so: undefined symbol: PyExc_ValueError
INFO:torch_onnxmlir.backend:Failed to compile the onnx model. Switch to the eager mode.
```
I merge the PR so that we can debug it.

PR #3564, opened 2026-07-01
[torch_onnxmlir] Remove onnx_utils

tungld:
Remove onnx_utils. It is because `onnx_utils` was to fix an issue in SWA models but another change in #3558 (about `tensor_node = graph.placeholder(f"{node.target}_omtensor")`) fixed the issue.


chentong319 (approved):
LGTM!

tungld:

PR #3566, opened 2026-07-01
Transpose mul matmul

AlexandreEichenberger:
Add a pattern that does the following (suggested optimization by @tungld )

```
//   x' = Transpose(x, perm)       (perm swaps only the last two dims)
//   x'' = Mul(x', scalar)         (scalar constant)
//   y = MatMul(x'', z)            (z not a constant)
//
// transforms to:
//   a' = Mul(x, scalar)
//   a'' = Transpose(a', perm)
//   y = MatMul(a'', z)
```

because if `z` is not a constant, then the scalar of the `mul` cannot easily be folded into `z`. In such case, it is best to have the `transpose` next to the `matmul` as it is often desirable to fold the transpose into the `matmul`.

Verified with/without on Granite 3.1, obtain the same results. I also verified that the transpose is folded into the `ZHigh.matmul`. So this should effectively remove a transpose from the execution runtime.

Added the pattern for `mul` and `div`; PR includes 3 new canonicalization lit tests.


tungld (approved):
LGTM! Thank you for adding the pattern!

AlexandreEichenberger:
Thanks @tung for merging it.

PR #3573, opened 2026-07-10
Fused op support: add reshape->expand->[mul by const]->stick support

AlexandreEichenberger:
In addition to adding support for the new pattern below (shown from Granite 4, with the mul, one of 120 instances)
```mlir
    %510 = "onnx.Fused"(%499, %501, %508) <{kind = "zhigh.expand-mul-stick"}> ({
    ^bb0(%arg83: tensor<?x4x?x128xf32>, %arg84: tensor<5xi64>, %arg85: tensor<3xi64>):
      %3877 = onnx.Constant dense<0.0883883461> : tensor<f32>
      %3878 = onnx.Constant dense<2> : tensor<1xi64>
      %3879 = "onnx.Unsqueeze"(%arg83, %3878) {onnx_node_name = "/model/layers.0/self_attn/Unsqueeze_12"} : (tensor<?x4x?x128xf32>, tensor<1xi64>) -> tensor<?x4x1x?x128xf32>
      %3880 = "onnx.Expand"(%3879, %arg84) {onnx_node_name = "/model/layers.0/self_attn/Expand"} : (tensor<?x4x1x?x128xf32>, tensor<5xi64>) -> tensor<?x4x4x?x128xf32>
      %3881 = "onnx.Mul"(%3880, %3877) {onnx_node_name = "/model/layers.0/self_attn/MatMul-/model/layers.0/self_attn/Transpose_3-/model/layers.0/self_attn/Mul_9-/model/layers.0/self_attn/Reshape_4_153"} : (tensor<?x4x4x?x128xf32>, tensor<f32>) -> tensor<?x4x4x?x128xf32>
      %3882 = "onnx.Reshape"(%3881, %arg85) <{allowzero = 0 : si64}> : (tensor<?x4x4x?x128xf32>, tensor<3xi64>) -> tensor<?x?x128xf32>
      %3883 = "zhigh.Stick"(%3882) <{layout = "3DS"}> : (tensor<?x?x128xf32>) -> tensor<?x?x128xf16, #zhigh.layout<{dataLayout = "3DS"}>>
      onnx.Yield %3883 : tensor<?x?x128xf16, #zhigh.layout<{dataLayout = "3DS"}>>
    }) {expansionN = 4 : i64, mulScalar = 0.0883883461 : f32, reshapeCollapsedCount = 3 : i64, reshapeFirstCollapsedDim = 0 : i64, stickFormat = "3DS", unsqueezedPosition = 2 : i64} : (tensor<?x4x?x128xf32>, tensor<5xi64>, tensor<3xi64>) -> tensor<?x?x128xf16, #zhigh.layout<{dataLayout = "3DS"}>>
```

This PR also streamline the infra to add new ops by having a new template for lowering to KRNL. Also, if for any reason, a detected `onnx.fused` fails its check (there is a `verify` method that ensure that the fused pattern has not been corrupted by optimizations), then the "fused" operation simply vanishes and the lowering of the individual operations within the fused are independently lowered.

This PR also fixes an issue that arises in the "verify" function; prior to this, constant op were "polluting" the verification sequence. This PR fixes that.

I have verified that this PR works with/without `-disable-fused-op` on Granite 3.1, test of Granite 3.4 also succeeded.

see docs/optimization-onnx-lowering/AddingAFusionPattern.md for more info on the new data structures


src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.hpp
```diff
@@ -620,6 +621,100 @@ struct ONNXGenericOpToCall : public mlir::OpConversionPattern<OP_TYPE> {
...
+    // (or the benefit-0 catch-all) can process the op instead.
+    if (fusedOp.getKind() != FusionT::kKind)
+      return mlir::failure();
+    // Retrieve body ops and cross-check stored params.  If the body was
+    // altered by an optimisation pass, fall back to inline.
+    fusion.retrieveOpsAndOutputValues(fusedOp);
+    if (!fusion.verifyAndRetrieveAttrs(fusedOp))
+      // Failed, use the inline fallback to allow constituent ops to be
+      // lowered on their own.
+      return FusionOpKindHelper::unFuse(rewriter, fusedOp);
```
tungld, line 665: Allowing this behaviour looks like there is something inconsistent in generating code. It is because once we generate a FusedOp and it is verified, it would be lowered successfully. It's OK to use it as a temporary solution until we find a better way.

src/Accelerators/NNPA/Conversion/ZHighToZLow/ZHighToZLow.cpp
```diff
@@ -2720,19 +2713,215 @@ struct ZHighToZLowFusedExtLayoutTransformLowering
...
+      if (d < P)
+        midDims.emplace_back(inputDims[d]);
+      else if (d == P)
+        midDims.emplace_back(LitIE(N));
+      else
+        midDims.emplace_back(inputDims[d - 1]);
+    }
+
+    // Step 2: apply the reshape's head-collapse to get the final output
+    // dims, used for allocation.
```
tungld, line 2817: We would find a way to reuse shape inference instead of recomputing the output shape here. Let's try to improve that in another PR if possible.

tungld (approved):
LGTM!

Before merging, could you please remove `--disable-fused-ops` in this test so that it is tested in our Jenkins: https://github.com/onnx/onnx-mlir/blob/main/src/Runtime/python/torch_onnxmlir/tests/test_decoder_models.py#L52?

Thank you for fixing the issue and supporting a new pattern! Do you have any information about how much performance improvement is with this PR?

src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.hpp
AlexandreEichenberger: This should never happen, unless there is some optimizations that messes up the operations. Maybe if we see it never happens (and maybe more of a warning, which I will add), we can remove the late "verify" and just automatically lower the code without looking at what is in the body of the fused op.

Let me know if that is what you have in mind @tungld

src/Accelerators/NNPA/Conversion/ZHighToZLow/ZHighToZLow.cpp
AlexandreEichenberger: @tungld I though about it, but I don't think its possible if there are dynamic shapes. Because we generate the shape calculation of very op, we would have to step through the computing of every op that we are eliminating, and in addition, we would never be able to rely on "onnx.Dim" of any of the intermediary results as they are never instantiated. 

But if you have an idea of how we could get that done, that would be great because I agree that is one step that would be nice not to have.

AlexandreEichenberger:
> Before merging, could you please remove --disable-fused-ops in this test so that it is tested in our Jenkins: https://github.com/onnx/onnx-mlir/blob/main/src/Runtime/python/torch_onnxmlir/tests/test_decoder_models.py#L52?

done

AlexandreEichenberger:
@tungld let me know if my answers satisfied your questions as far as merging this PR is concerned.

src/Conversion/ONNXToKrnl/ONNXToKrnlCommon.hpp
tungld: If it never happens, should we use `llvm_unreachable` here, according to this docs: https://github.com/onnx/onnx-mlir/blob/main/docs/ErrorHandling.md

tungld:
> @tungld let me know if my answers satisfied your questions as far as merging this PR is concerned.

Feel free to merge it, it was approved. BTW, do you have any information about how much performance improvement we get with this PR?

AlexandreEichenberger:
> Feel free to merge it, it was approved. BTW, do you have any information about how much performance improvement we get with this PR?

will start some measurements :-)

PR #3574, opened 2026-07-13
added new pattern of dim -> [seqeeze] -> range for dim analysis

AlexandreEichenberger:
In some models, I find `onnx.range` operations that create a range of `0..dim(x, axis=x) by 1`. So when that is the case, we know the shape of the output tensor of range.

It caught 4 of the 5 such pattern in Granite 4.

example pattern:
```mlir
func.func @test_range_from_dim(%arg0: tensor<?x?xi64>) -> tensor<?xi64> {
  %start = onnx.Constant dense<0> : tensor<i64>
  %delta = onnx.Constant dense<1> : tensor<i64>
  %axis = onnx.Constant dense<0> : tensor<1xi64>
  %dim = "onnx.Dim"(%arg0) {axis = 1 : si64} : (tensor<?x?xi64>) -> tensor<1xi64>
  %limit = "onnx.Squeeze"(%dim, %axis) : (tensor<1xi64>, tensor<1xi64>) -> tensor<i64>
  %0 = "onnx.Range"(%start, %limit, %delta) : (tensor<i64>, tensor<i64>, tensor<i64>) -> tensor<?xi64>
  return %0 : tensor<?xi64>
}
```

Added lit test for the new pattern.

### DIFF
--- a/src/Dialect/ONNX/ONNXDimAnalysis.cpp
+++ b/src/Dialect/ONNX/ONNXDimAnalysis.cpp
@@ -1316,6 +1316,35 @@ void DimAnalysis::visitDim(
       }
     }
   }
+
+  // RangeOp has a special case: when start == 0 and delta == 1 (scalar
+  // constants), the number of elements in the output is exactly `limit`,
+  // since num = max(ceil((limit-start)/delta), 0) = max(limit, 0). If `limit`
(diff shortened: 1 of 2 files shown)


tungld (approved):
LGTM! Thanks for adding the pattern!

PR #3575, opened 2026-07-14
Load once if the input of expand is a scalar tentor

tungld:
A small optimization for the lowering of onnx.expand: when  the input is a scalar, load its value outside the main loop of expansion.

### DIFF
--- a/src/Conversion/ONNXToKrnl/Tensor/Expand.cpp
+++ b/src/Conversion/ONNXToKrnl/Tensor/Expand.cpp
@@ -66,15 +66,22 @@ struct ONNXExpandOpLowering : public OpConversionPattern<ONNXExpandOp> {
     if (enableParallel)
       tryCreateKrnlParallel(create.krnl, op, "expand", outputLoopDef, lbs, ubs);
 
+    // If input is a scalar, load its value outside the loop.
+    Value val = nullptr;
+    if (isScalarTensor(input))
+      val = create.krnl.load(input);
+
     create.krnl.iterateIE(outputLoopDef, outputLoopDef, lbs, ubs,
         [&](const KrnlBuilder &createKrnl, ValueRange outputLoopInd) {
           IndexExprScope outputScope(createKrnl, shapeHelper.getScope());
           SmallVector<IndexExpr, 4> outputLoopIndices, lhsAccessExprs;
           getIndexExprList<DimIndexExpr>(outputLoopInd, outputLoopIndices);
-          LogicalResult res = shapeHelper.getAccessExprs(
-              input, 0, outputLoopIndices, lhsAccessExprs);
-          assert(succeeded(res) && "Could not compute access indices");
-          Value val = createKrnl.loadIE(input, lhsAccessExprs);
+          if (!val) {
+            LogicalResult res = shapeHelper.getAccessExprs(
+                input, 0, outputLoopIndices, lhsAccessExprs);
+            assert(succeeded(res) && "Could not compute access indices");
+            val = createKrnl.loadIE(input, lhsAccessExprs);
+          }
           createKrnl.store(val, alloc, outputLoopInd);
         });
 
--- a/src/Conversion/ONNXToKrnl/Tensor/Expand.cpp
+++ b/src/Conversion/ONNXToKrnl/Tensor/Expand.cpp
@@ -73,10 +73,10 @@ struct ONNXExpandOpLowering : public OpConversionPattern<ONNXExpandOp> {
 
     create.krnl.iterateIE(outputLoopDef, outputLoopDef, lbs, ubs,
         [&](const KrnlBuilder &createKrnl, ValueRange outputLoopInd) {
-          IndexExprScope outputScope(createKrnl, shapeHelper.getScope());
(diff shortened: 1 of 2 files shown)


chentong319:
Did you check whether the later optimization pass can perform such transformation automatically? I'd like to keep our lowering simple if possible.

tungld:
> Did you check whether the later optimization pass can perform such transformation automatically? I'd like to keep our lowering simple if possible.

I checked before implementing this, but unfortunately nothing happened, the load was still inside the loop. Here is the result without this PR:
```mlir
./Debug/bin/onnx-mlir -O3 -march=z17 --EmitMLIR --printIR expand-scalar.mlir
[1/3] Wed Jul 15 01:48:17 2026 (0s) Importing ONNX Model to MLIR Module from "expand-scalar.mlir"
[2/3] Wed Jul 15 01:48:17 2026 (0s) Compiling and Optimizing MLIR Module
#map = affine_map<()[s0, s1] -> (s1)>
module attributes {llvm.data_layout = "E-S64-m:e-i1:8:16-i8:8:16-i64:64-f128:64-v128:64-a:8:16-n32:64", llvm.target_triple = "s390x-unknown-linux-gnu", "onnx-mlir.compile_options" = "-O3 -march=z17 --EmitMLIR --printIR expand-scalar.mlir", "onnx-mlir.compiler_version" = "onnx-mlir version 0.5.1 (eb6fe948)", "onnx-mlir.op_stats" = "{\0A  \22func.return.2D\22 : 2,\0A  \22onnx.Expand.0D.1D\22 : 1,\0A  \22onnx.Expand.1D.1D\22 : 1\0A}\0A", "onnx-mlir.symbol-postfix" = "expand-scalar"} {
  func.func @expand_scalar_0D(%arg0: memref<f32>, %arg1: memref<2xi64>) -> memref<?x?xf32> attributes {llvm.emit_c_interface} {
    %c1 = arith.constant 1 : index
    %c0 = arith.constant 0 : index
    %0 = affine.load %arg1[%c0] : memref<2xi64>
    %1 = arith.index_cast %0 : i64 to index
    %2 = affine.load %arg1[%c1] : memref<2xi64>
    %3 = arith.index_cast %2 : i64 to index
    %alloc = memref.alloc(%1, %3) {alignment = 16 : i64} : memref<?x?xf32>
    affine.for %arg2 = 0 to %1 {
      affine.for %arg3 = 0 to #map()[%1, %3] {
        %4 = affine.load %arg0[] : memref<f32>
        affine.store %4, %alloc[%arg2, %arg3] : memref<?x?xf32>
      }
    }
    return %alloc : memref<?x?xf32>
  }
  func.func @expand_scalar_1D(%arg0: memref<1xf32>, %arg1: memref<2xi64>) -> memref<?x?xf32> attributes {llvm.emit_c_interface} {
    %c1 = arith.constant 1 : index
    %c0 = arith.constant 0 : index
    %0 = affine.load %arg1[%c0] : memref<2xi64>
    %1 = arith.index_cast %0 : i64 to index
    %2 = affine.load %arg1[%c1] : memref<2xi64>
    %3 = arith.index_cast %2 : i64 to index
    %alloc = memref.alloc(%1, %3) {alignment = 16 : i64} : memref<?x?xf32>
    affine.for %arg2 = 0 to %1 {
      affine.for %arg3 = 0 to #map()[%1, %3] {
        %4 = affine.load %arg0[%c0] : memref<1xf32>
        affine.store %4, %alloc[%arg2, %arg3] : memref<?x?xf32>
      }
    }
    return %alloc : memref<?x?xf32>
  }
}
```

AlexandreEichenberger (approved):
Why this is useful and an improvement, the code should really be reoptimized.

Ideally, we would have something like this when we know at compile time what is broadcasted
```C++
if (broadcast only one dim) {
  generate loops for iterating over the outer loop up to but excluding the broadcast dim)
    generate loops inside (starting after the broadcast dim)
       load single value
       for each broadcast val: i
          store to broadcast index i
 }
```
If there are multiple broadcasting dimensions,  we can take the innermost broadcasted dim, and it would work.

Ideally, the "non broadcasting" dim are indicated by a "1" in the expand `shape`. But I have often seen the actual dim being used in the code (e.g. granite)

eg `expand(<d1, 1, d2, d3>, shape=(d2, 4, 1, 1)` aka we need to check that the 'd1' and `d2` are identical using dim analysis

Ideally, if the broadcasting is not on the innermost dim, use SIMD load and stores.

tungld:
> Ideally, we would have something like this when we know at compile time what is broadcasted
if (broadcast only one dim) {
  generate loops for iterating over the outer loop up to but excluding the broadcast dim)
    generate loops inside (starting after the broadcast dim)
       load single value
       for each broadcast val: i
          store to broadcast index i
 }

I tried this before: flatten the outer loop but did multiple memcpy (4 in case of granite) instead of the inside loop instead. Tested with big tensors in granite, but the improvement is negligible, perhaps, it reaches the memory bandwidth limit. 
For data movement ops, looks like doing tiling is better for cache locality.

AlexandreEichenberger:
For what it is worth, in the fused op, I did not do multiple mem copy, as it would reload the data multiple times. I did a SIMD load and then multiple SIMD stores. In essence a mem-broadcast-copy (if there is such a thing).

PR #3578, opened 2026-07-16
[torch_onnxmlir] Add an option to generate sample inputs/outputs

tungld:
This PR adds two config options:
1. `generate_test_data_set`: to generate sample inputs/outputs for each generated .so file.
2. `regenerate_test_data_set`: to regenerate new sample inputs/outputs for a different model query if needed.

torch_onnxmlir may generate multiple onnx models for a single pytorch model. In such a case, it's quite difficult to know how to run the onnx models in a standalone way (without pytorch). It's because we don't know how to prepare correct inputs for these models.

This PR adds an option to generate sample inputs/outputs for every generated onnx model, so that it's ready to use with `RunONNXModel.py --load-ref=/path/to/test_data_set`.

Users enable this feature by setting `torch_onnxmlir.config.generate_test_data_set = True`. A folder `test_data_set` will be created in the cache folder, containing `input_0.pb, input_1.pb, ..., output_0.pb, output_1.pb, ...` that can be consumed by `RunONNXModel.py`.

Sample inputs/outputs are collected from the first run of the model only.


AlexandreEichenberger (approved):
Great new feature, thanks.

For the decode phase, we may want to run with multiple sequence length, any way to make this a parameter (for the gen of the inputs)?

chentong319:
> Great new feature, thanks.
> 
> For the decode phase, we may want to run with multiple sequence length, any way to make this a parameter (for the gen of the inputs)?

For the standard encoder-decoder model, the PR #3568 may address your requirements.

chentong319 (approved):
LGTM!

tungld:
> For the decode phase, we may want to run with multiple sequence length, any way to make this a parameter (for the gen of the inputs)?

We can have an option to allow overwriting the test_data_set. Rerunning with new inputs (no compilation happens) will generate new inputs/outputs and overwrite the test_data_set

tungld:
@AlexandreEichenberger Added a new option `regenerate_test_data_set` to regenerate new sample inputs/outputs for a different model query if needed.

PR #3579, opened 2026-07-21
Using .npy format instead of .pb format to store numpy array

tungld:
This PR uses .npy format instead of .pb format in torch_onnxmlir to store numpy arrays on disk, so that we don't need to `import onnx`.

It also adds a new option to RunONNXModel.py, that is `--use-npy` to use models with .npy inputs and outputs. Still keep .pb format support in RunONNXModel.py since we run models in the onnx model zoo that use .pb format.


AlexandreEichenberger (approved):
LGTM, any idea on the perf difference? Or is it only for debugging?

tungld:
> LGTM, any idea on the perf difference? Or is it only for debugging?

There may have perf difference in reading inputs between .pb and .npy, but when we measures the inference perf we don't measure the time of reading inputs. I think we shouldn't depend on onnx to just read/write data to disk.

PR #3591, opened 2026-07-30
[torch_onnxmlir] Introduce `explain_context()` to replace `explain()`

tungld:
This PR introduces `explain_context()` as the primary API for collecting compilation and inference metrics in torch_onnxmlir. 

The previous interface required manual flag management (`config.enable_explain = True/False`) and explicit `clear_metrics()` calls, which was error-prone and verbose. The new context manager automatically handles the metrics lifecycle (enable/collect/disable) and defaults to table format for better readability. 

**Old Interface (manual):**
```python
import torch_onnxmlir as om

om.config.enable_explain = True
output = compiled_model(input)
metrics = om.explain()
print(metrics)
om.config.enable_explain = False

om.clear_metrics() # for the next explain().
```

**New Interface (context manager):**
```python
import torch_onnxmlir as om

with om.explain_context() as metrics:
    output = compiled_model(input)
    print(metrics)  # Displays formatted performance metrics table

with om.explain_context() as metrics:
    output = compiled_model(new_input)
    print(metrics)  # Displays formatted performance metrics table
```

The change also reorganizes the tests, eliminating redundancy while maintaining full coverage.


AlexandreEichenberger (approved):
LGTM, thanks for streamlining the usage of metrics

